PR #748, opened 2023-03-07
revamped ArrayLike

tylerflex:
`ArrayLike` converts provided value into `np.array` and is encoded to json using `tolist()` for real and imaginary parts separately.
`constrained_array(dtype, ndim)` returns an `ArrayLike` with the dtype and number of dimensions either cast or validated.

### DIFF
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -51,6 +51,13 @@ def cached_property(cached_property_getter):
     return property(cache(cached_property_getter))
 
 
+def ndarray_encoder(val):
+    """How a ``np.ndarray`` gets handled before saving to json."""
+    if np.any(np.iscomplex(val)):
+        return dict(real=val.real.tolist(), imag=val.imag.tolist())
+    return val.real.tolist()
+
+
 class Tidy3dBaseModel(pydantic.BaseModel):
     """Base pydantic model that all Tidy3d components inherit from.
     Defines configuration for handling data structures
@@ -90,7 +97,7 @@ class Tidy3dBaseModel(pydantic.BaseModel):
         validate_assignment = True
         allow_population_by_field_name = True
         json_encoders = {
-            np.ndarray: lambda x: tuple(x.tolist()),
+            np.ndarray: ndarray_encoder,
             complex: lambda x: ComplexNumber(real=x.real, imag=x.imag),
             xr.DataArray: DataArray._json_encoder,  # pylint:disable=unhashable-member, protected-access
         }
@@ -517,6 +524,10 @@ class Tidy3dBaseModel(pydantic.BaseModel):
     def __ge__(self, other):
         """define >= for getting unique indices based on hash."""
         return hash(self) >= hash(other)
+
+    def __eq__(self, other):
+        """Define == for two Tidy3DBaseModels."""
+        return self._json_string == other._json_string
 
     @cached_property
     def _json_string(self) -> str:
--- a/tidy3d/components/geometry.py
+++ b/tidy3d/components/geometry.py
@@ -14,7 +14,7 @@ from shapely.geometry import Point, Polygon, box, MultiPolygon
 from shapely.validation import make_valid
 
 from .base import Tidy3dBaseModel, cached_property
-from .types import Bound, Size, Coordinate, Axis, Coordinate2D, ArrayLike, PlanePosition
+from .types import Bound, Size, Coordinate, Axis, Coordinate2D, ArrayFloat3D, PlanePosition
 from .types import Vertices, Ax, Shapely, annotate_type
 from .viz import add_ax_if_none, equal_aspect
 from .viz import PLOT_BUFFER, ARROW_LENGTH, arrow_style
@@ -47,6 +47,10 @@ _N_SAMPLE_POLYGON_INTERSECT = 5
 # for sampling conical frustum in visualization
 _N_SAMPLE_CURVE_SHAPELY = 40
 _IS_CLOSE_RTOL = np.finfo(float).eps
+
+
+Points = ArrayFloat3D
+
 
 # pylint:disable=too-many-public-methods
 class Geometry(Tidy3dBaseModel, ABC):
@@ -634,9 +638,7 @@ class Geometry(Tidy3dBaseModel, ABC):
         return axis, position
 
     @staticmethod
-    def rotate_points(
-        points: ArrayLike[float, 3], axis: Coordinate, angle: float
-    ) -> ArrayLike[float, 3]:
+    def rotate_points(points: Points, axis: Coordinate, angle: float) -> Points:
         """Rotate a set of points in 3D.
 
         Parameters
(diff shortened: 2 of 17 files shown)


tidy3d/components/types.py
```diff
@@ -26,71 +24,76 @@ def annotate_type(UnionType):  # pylint:disable=invalid-name
...
+        field_schema.update(schema)
 
-    def __getitem__(cls, type_ndims):
-        """Array[type, ndims] -> TypedArrayLike"""
-        desired_type, ndims = type_ndims
-        return type("Array", (TypedArrayLike,), {"inner_type": desired_type, "ndims": ndims})
 
+def constrained_array(dtype: type = None, ndim: int = None) -> type:
+    """Generate an ArrayLike sub-type with constraints built in."""
+    type_name = "ArrayLike_"
```
lucas-flexcompute, line 81: Doesn't it make more sense to add the separator in the if bodies?
```python
type_name = "ArrayLike"
if dtype is not None:
    type_name += f"_dtype={dtype}"
if ndim is not None:
    type_name += f"_ndim={ndim}"
```

tidy3d/components/base.py
```diff
@@ -517,6 +524,9 @@ def __ge__(self, other):
         """define >= for getting unique indices based on hash."""
         return hash(self) >= hash(other)
 
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
lucas-flexcompute: Currently `pydantic.BaseModel` uses a dict conversion, and it is already a [known bottleneck](https://github.com/pydantic/pydantic/issues/1494). Using a json string is more constly both due to conversion and comparison, so I'd vote not to use this.
Is there any particular reason to compare this way instead of with dicts?

tidy3d/components/geometry.py
```diff
@@ -2015,14 +2017,18 @@ def length_axis(self) -> float:
         zmin, zmax = self.slab_bounds
         return zmax - zmin
 
+    @pydantic.validator("vertices", pre=True, always=True)
+    def convert_to_numpy(cls, val):
```
lucas-flexcompute, line 2021: Why do we need this pre-conversion? Using later `copy=true` in the `ArrayLike` validator will most likely lead to 2 copies.

lucas-flexcompute (approved):

tidy3d/components/types.py
tylerflex: yea you're right. I'll make the change.

tidy3d/components/geometry.py
tylerflex: it's unfortunate but the reasoning is because a lot of other validators for "vertices" require the vertices in np.ndarray form.

Previously, they were each converting the `val` to `np.ndarray`, but I decided to just refactor in a `pre` validator so that it only occurs once. Of course, there will be an extra conversion from the `ArrayLike` validator. I'm not sure the best workaround here.

tidy3d/components/base.py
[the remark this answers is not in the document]
tylerflex: A few things.
1. the equality checking is mainly just used in tests, eg to ensure that a sim loaded from file equals the original. However, I agree we should try to use a more performant solution in case people are using this more.
2. supporting `ArrayLike` with `np.ndarray` means that simulations are no longer hashable because `np.ndarray` is not hashable. without defining my own `__eq__` for base, it was using dict comprehension and I was getting errors about `np.ndarray` not being hashable. The (hacky) fix was to convert this to json string for comparison.

I don't really see too many alternatives to this except for.
1. reverting to not store ArrayLike as `np.ndarray`, to retain hashability.
2. subclassing `np.ndarray` and providing `hash` and `eq` methods & using pydantic's dict comprehension.
3. writing our own dict comprehension that does equality checking differently if the compared values are `np.ndarray` (eg using `np.all`) 

But I'm not sure we can get around a dict comprehension here. Do you have any ideas?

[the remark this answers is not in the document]
lucas-flexcompute: I think number 3 might be the best option here, then. If that's going to be too complicated, then maybe leave it as is with a TODO note, since it is only for testing at this point.

tidy3d/components/geometry.py
lucas-flexcompute: I see… in that case I think we live with it, then.

tidy3d/components/base.py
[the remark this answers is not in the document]
tylerflex: I took a stab at subclassing [here](https://github.com/flexcompute/tidy3d/blob/84ba787958c535a16118e9b48aceda41d1cb13a2/tidy3d/components/types.py#L30-L44) and [here](https://github.com/flexcompute/tidy3d/blob/84ba787958c535a16118e9b48aceda41d1cb13a2/tidy3d/components/types.py#L76-L80). and It's a bit messy, but seems to work for hashing. not sure If this is a very good solution or we should rethink things..

tidy3d/components/types.py
```diff
@@ -26,71 +24,100 @@ def annotate_type(UnionType):  # pylint:disable=invalid-name
...
 
+""" Numpy Arrays """
+
+
+def _totuple(arr: np.ndarray) -> tuple:
+    """Convert a numpy array to a nested tuple."""
+    try:
+        return tuple(_totuple(val) for val in arr)
+    except TypeError:
+        return arr
```
lucas-flexcompute, line 35: Instead of using `try`, isn't it cleaner to use:
```python
def _to_tuple(a):
     if a.ndim > 1:
         return tuple(_to_tuple(v) for v in a)
     return tuple(a)
```

tidy3d/components/base.py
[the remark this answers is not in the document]
lucas-flexcompute: Hard to say. Numpy arrays are much easier to work with and avoid explicit loops. The drawback of sub-classing is that operation results would be ndarray, not TidyNDArray, right? But then, again the validators in ArrayLike woud convert anything the requires hashing to TidyNDArray, so that's not too bad I guess.

tidy3d/components/types.py
tylerflex: thanks, made the change

momchilmm:
I don't really understand the details but fine by me to merge. I just checked that ArrayLike is not used on the backend, so if everything works well here, we could merge.

That said, it may make sense to run backend tests with this branch anyway, since polyslab vertices are now stored as array instead of list of lists, right? Could you resolve conflicts and I can run the tests.

tylerflex:
@momchilmm I rebased this PR as it it had gotten stale. Might be good to get it in sooner than later. I dont think it's important you review all of the details, but the main decision is whether you think it's acceptable that (in it's current state), the `ArrayLike` types are stored internally as `np.ndarray` subclasses. The advantage of this is fewer internal conversions and `tolist()` calls. The disadvantage is that `np.ndarray` is not hashable, so we need to introduce a subclass of `np.ndarray` with a hash function to allow `Tidy3D` components to be hashable after the changes. Hashability isn't really used anywhere internally, but I think maybe Joaquin is using it to cache tidy3d simulations in a dict, so I feel it would be nice to support. I guess let me know if you have any objections or general thoughts, but we should probably decide something before 1.10 so this doesn't get too stale again.

tylerflex:
Just saw your message from yesterday. Sounds like a plan, I've rebased so maybe could you run the backend tests at your convenience? thanks

momchil-flex:
I think gdsfactory hashes the gdsfactory definitions that get translated to a tidy3d simulation, not the tidy3d simulation directly. But so keeping Simulation hashable is left for the future as it wasn't straightforward?

tylerflex:
Hm, I see. So the simulation is currently hashable (just checked) at least as of `pre/1.10.0`, so the stuff I've done in this PR keeps it that way while introducing an `np.ndarray` subclass to store the ArrayLike. Figure why not try to keep it.

PR #758, opened 2023-03-14
validator to error if JaxSimulation provided to SimulationTask

tylerflex:


tidy3d/web/simulation_task.py
```diff
@@ -153,6 +154,20 @@ class SimulationTask(ResourceLifecycle, Submittable, extra=Extra.allow):
...
+    def _error_if_jax_sim(cls, values):
+        """Raise error if user tries to submit simulation that's a JaxSimulation."""
+        sim = values.get("simulation")
+        if sim is None:
+            return values
+        if "JaxSimulation" in str(type(sim)):
+            raise ValueError(
+                "'JaxSimulation' not compatible with regular webapi functions. "
+                "Either convert it to Simulation with 'jax_sim.to_simulation()[0]' or use "
+                "the 'adjoint.run' function to run JaxSimulations."
```
momchil-flex, line 167: `adjoint.web.run` ?

momchil-flex (approved):
Nice, great to be able to test actually!

tidy3d/web/simulation_task.py
tylerflex: you can actually just import it directly from `adjoint` so I think it's ok.
https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/plugins/adjoint/__init__.py#L5

PR #759, opened 2023-03-15
fix task name and expose solver version

tylerflex:
task name should be fixed now. 
Solver version can be passed to `web.run()` or directly to `web.start()`

### DIFF
--- a/tidy3d/web/webapi.py
+++ b/tidy3d/web/webapi.py
@@ -9,6 +9,7 @@ import pytz
 from rich.console import Console
 from rich.progress import Progress
 
+from ..version import __version__
 from .environment import Env
 from .simulation_task import SimulationTask, SIM_FILE_HDF5, Folder
 from .task import TaskId, TaskInfo
@@ -38,6 +39,9 @@ def run(  # pylint:disable=too-many-arguments
     verbose: bool = True,
     progress_callback_upload: Callable[[float], None] = None,
     progress_callback_download: Callable[[float], None] = None,
+    solver_version: str = None,
+    worker_group: str = None,
+    protocol_version: str = __version__,
 ) -> SimulationData:
     """Submits a :class:`.Simulation` to server, starts running, monitors progress, downloads,
     and loads results as a :class:`.SimulationData` object.
@@ -61,6 +65,12 @@ def run(  # pylint:disable=too-many-arguments
         Optional callback function called when uploading file with ``bytes_in_chunk`` as argument.
     progress_callback_download : Callable[[float], None] = None
         Optional callback function called when downloading file with ``bytes_in_chunk`` as argument.
+    solver_version: str = None
+        target solver version.
+    worker_group: str = None
+        worker group
+    protocol_version: str = None
+        protocol version
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


tidy3d/web/webapi.py
```diff
@@ -152,18 +167,25 @@ def get_info(task_id: TaskId) -> TaskInfo:
...
-    solver_version : str
-        Supply or override a specific solver version to the task.
-    verbose : bool = True
-        If `True`, will print progressbars and status, otherwise, will run silently.
+    solver_version: str = None
+        target solver version.
+    worker_group: str = None
+        worker group
+    protocol_version: str = None
+        protocol version
```
momchil-flex, line 188: I am actually confused, here and in a number of other places it seems to say `protocol_version: str = None` in the docstring while `protocol_version: str = __version__` in the function definition. What I also can't follow is what logic ensures that protocol_version is passed if solver_version is None and vice versa?

tylerflex: On the first point, this is just how it was handed off to me, but I agree maybe we should make it actually `=None` in the method signature and then later set it to `__version__` if it's None?

On your second point about the logic, I dont quite follow. Is one of protocol version or solver version required? what logic should we implement because right now I dont know if there is any.

momchil-flex: So we figured out passing a `solver_version` doesn't work unless you also pass `prorocol_version=None`. I was under the impression that you would handle this internally, but I think I misunderstood that. You probably meant that you just tried passing `protocol_version=None` in web.run and that worked. In the previous webapi version, this was handled internally.

https://github.com/flexcompute/tidy3d/blob/027b422b84deb79446a013da4e6ebbb8dc0a2bdd/tidy3d/web/webapi.py#L191-L195

momchil-flex: I do feel like it doesn't hurt to handle internally here too.

tylerflex: or I guess what do we do if `solver_version` and `protocol_version` is supplied for example?

momchil-flex: That was the logic before yeah. It seems like if they are both supplied the webapi returns an error, I guess because it's not sure which one to use. We could fix this behavior but that goes through modifying the webapi, not our python code.

tylerflex: ok I changed it to do the same logic from within the simulation task submission and now all protocol versions are none by default (but set to __version__ if solver version is None)

https://github.com/flexcompute/tidy3d/blob/55d29a3e063aa61ae5978098b593a4dff1566685/tidy3d/web/simulation_task.py#L350-L353

weiliangjin2021:
The solver_version can be passed to web.run and web.start, but not to container such as Job?

momchil-flex:
> The solver_version can be passed to web.run and web.start, but not to container such as Job?

Yeah that's a good point. Also one last small thing from me - maybe just remove `protocol_version` from web.run and web.start. I can only think of very weird use cases for that. It should just be the current frontend version, or None.

tylerflex:
Ok to cover both of these points, I removed `protocol_version` from being able to be specified by user (can only be None or __version__ depending on solver_version) and added `solver_version` to job and batch.

momchil-flex:
Great, I'll give this a test now as I actually need to run with a specific solver version.

weiliangjin2021:
works for me. Maybe my jupyter issue or some settings changes: no progress bar when solver is running now. It stuck at `starting up solver` and only show progress bar when the solver finishes.

tylerflex:
seems like probably unrelated to this PR, but I can look into it tomorrow.

momchil-flex:
Looks good to me within the scope of this PR - feel free to merge.

I do see the same as Weiliang. My guess would be that it's the same issue that was introduced by the old webapi changes and then we had to work on fixing is present here, perhaps it was reverted when the webapi refactor was merged in.

The other thing is the API key. It didn't let me log in with my old style credentials. I think just having a more detailed message about where to get the API key from could be fine.

weiliangjin2021 (approved):

PR #760, opened 2023-03-15
Fix a few unstable materials in material_library

weiliangjin2021:
Performed a scan of all materials in the material library, and find that `Cu_JohnsonChristy1972` and `Ni_JohnsonChristy1972` are unstable. `TiOx_Horib` is stable with a smaller courant number since it has `eps_inf<1`. But the caveat is that no subpixel averaging will be performed for structures made of this material in the simulation. So a  `TiOx_HoribStable` of `eps_inf=1` is added.


tylerflex (approved):

lucas-flexcompute (approved):

lucas-flexcompute:
Is it wise to keep the unstable version of TiOx in the library? Same goes for Fused Silica, for example, where there seems to be a PML stable version and what I assume might be an unstable version. It's confusing from a user perspective.

weiliangjin2021:
> Is it wise to keep the unstable version of TiOx in the library? Same goes for Fused Silica, for example, where there seems to be a PML stable version and what I assume might be an unstable version. It's confusing from a user perspective.

One difference between `TiOx_Horib` and `TiOx_HoribStable` is that the former is directly from the reference, while the latter is fitted. So if we remove the former, it might also be confusing? And it's only unstable if we don't use a smaller time step size. 

As for the PML stable Fused silica, the unstable version is still stable if the material doesn't extend into PML.

weiliangjin2021:
Another reason for keeping the original version of fused silica is also that it's directly based on the reference which provides a Sellmeier formula, and the PMLStable version is fitted.

lucas-flexcompute:
In that case, we should provide this information in the documentation and also in the docstring of the material, so that it is easier to chose a variant from `help(td.material_library["TiOx"])`, for example.

weiliangjin2021:
Good point. But there is no place to add this information in the current script. How about add a field in `VariantItem` to store the info (in a separate PR)?

xin-flex (approved):

PR #762, opened 2023-03-16
Extra field plot options

lucas-flexcompute:
- Allow plotting the Poynting vector from `SimulationData.field_plot` and full vectorial magnitudes for E and H.

- Include components `"abs^2"` and `"dB"` to `val`, applicable to all components.

- Improve the selection of divergent colormap whenever the data _could_ contain positive and negative values, and centralize map around zero.

### DIFF
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -6,12 +6,12 @@ import xarray as xr
 import pydantic as pd
 import numpy as np
 
-from .monitor_data import MonitorDataTypes, MonitorDataType, AbstractFieldData
+from .monitor_data import MonitorDataTypes, MonitorDataType, AbstractFieldData, FieldTimeData
 from ..base import Tidy3dBaseModel
 from ..simulation import Simulation
 from ..boundary import BlochBoundary
 from ..source import TFSF
-from ..types import Ax, Axis, annotate_type, Literal, PlotVal
+from ..types import Ax, Axis, annotate_type, FieldVal, PlotScale, ColormapType
 from ..viz import equal_aspect, add_ax_if_none
 from ...log import DataError, log, Tidy3dKeyError, ValidationError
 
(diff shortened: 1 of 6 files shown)


lucas-flexcompute:
I did not merge the 2 Poyting vector calculations. The 2D version makes sense to be a method of `MonitorData`, whereas the 3D version makes more sense in `SimulationData`. Since they are already different calculations, I just let them be separate.

tidy3d/components/data/sim_data.py
```diff
@@ -213,13 +213,103 @@ def at_centers(self, field_monitor_name: str) -> xr.Dataset:
 
         return monitor_data.colocate(**xyz_kwargs)
 
-    def get_intensity(self, field_monitor_name: str) -> xr.DataArray:
+    def get_poynting_vector(self, field_monitor_name: str) -> xr.Dataset:
+        """return `xarray.Dataset` of the Poynting vector at Yee cell centers.
```
tylerflex, line 217: for RST, we should use double backslash here.

tylerflex: same below (and might be good to put the math bit `S = 1/2 E × conj(H)` in double backslash too

tidy3d/components/data/sim_data.py
```diff
@@ -213,13 +213,103 @@ def at_centers(self, field_monitor_name: str) -> xr.Dataset:
...
+
+        field_dataset = self.at_centers(field_monitor_name)
+
+        time_domain = isinstance(self.monitor_data[field_monitor_name], FieldTimeData)
+
+        poynting_components = {}
+
+        dims = "xyz"
+        for axis, dim in enumerate(dims):
+            dim_1 = dims[axis - 2]
```
tylerflex, line 245: our convention to get the tangential field components is to "pop" the normal dimension (eg. for normal (x, y, z), the tangential components are usually (yz), (xz), (xy) in our code.

It seems here that if, eg `dim` is `y`, then the dim1 and dim2 are `z, x`, respectively. Just want to make sure / double check that this doesn't introduce an unexpected minus sign.

tidy3d/components/data/sim_data.py
```diff
@@ -213,13 +213,103 @@ def at_centers(self, field_monitor_name: str) -> xr.Dataset:
...
+            field_value.name = f"|{field_component.name}|²"
+
+        elif val == "phase":
+            field_value = np.arctan2(field_component.imag, field_component.real)
+            field_value.name = f"∠{field_component.name}"
+
+        return field_value
+
+    def get_derived_field(self, field_monitor_name: str, field_name: str, val: FieldVal):
         """return `xarray.DataArray` of the intensity of a field monitor at Yee cell centers.
```
tylerflex, line 303: might need to update this docstring?

tidy3d/components/data/sim_data.py
```diff
@@ -228,26 +318,79 @@ def get_intensity(self, field_monitor_name: str) -> xr.DataArray:
...
-            raise DataError(
-                "Field monitor must contain 'Ex', 'Ey', and 'Ez' fields to compute intensity."
-            )
+        if field_name in ("E", "H", "S"):
+            # Gather vector components
+            required_components = [field_name + c for c in "xyz"]
+            if not all(field_cmp in dataset for field_cmp in required_components):
+                raise DataError(
+                    f"Field monitor must contain '{field_name}x', '{field_name}y', and "
+                    f"'{field_name}z' fields to compute '{field_name}."
```
tylerflex, line 338: missing trailing `'` after {field_name}

tidy3d/components/data/sim_data.py
```diff
@@ -228,26 +318,79 @@ def get_intensity(self, field_monitor_name: str) -> xr.DataArray:
...
+                )
+            field_components = (dataset[c] for c in required_components)
+
+            # Apply the requested transformation
+            if val == "real":
+                derived_data = sum(f.real**2 for f in field_components) ** 0.5
```
tylerflex, line 344: can you explain why we're taking the absolute value of the real part here? If the user wants to just plot Re{E} for example, is there a way to do it?

tidy3d/components/data/sim_data.py
```diff
@@ -298,17 +443,52 @@ def plot_field(  # pylint:disable=too-many-arguments, too-many-locals, too-many-
...
             if field_name not in field_monitor_data.field_components:
                 raise DataError(f"field_name '{field_name}' not found in data.")
-            field_data = field_monitor_data.field_components[field_name]
+            field_component = field_monitor_data.field_components[field_name]
+            field_component.name = field_name
+            field_data = self._field_component_value(field_component, val)
+
+        if scale == "dB":
+            if val == "phase":
+                log.warning("Ploting phase component in log scale masks the phase sign")
```
tylerflex, line 469: period at end of sentence

tidy3d/components/data/sim_data.py
```diff
@@ -368,39 +560,37 @@ def plot_field(  # pylint:disable=too-many-arguments, too-many-locals, too-many-
...
         field_data: xr.DataArray,
         axis: Axis,
         position: float,
-        val: Literal["real", "imag", "abs"] = "real",
         freq: float = None,
         eps_alpha: float = 0.2,
         robust: bool = True,
         vmin: float = None,
         vmax: float = None,
+        cmap_type: Literal["divergent", "sequential", "cyclic"] = "divergent",
```
tylerflex, line 593: maybe better to define the cmap type in types.py along with FieldVal and PlotScale so they exist in the same place.

tylerflex (changes requested):
Looks good overall, it will be great to have more plotting options. Just a few minor comments and a bit confused about why the absolute value of Re / Imag are plotted instead of the signed value, maybe you could explain, thanks

tylerflex:
Another thing is I'd recommend adding some calls to these new options in the tests/test_data/test_sim_data, just to run through them and ensure that they work properly. Thanks!

tidy3d/components/data/sim_data.py
```diff
@@ -213,13 +213,103 @@ def at_centers(self, field_monitor_name: str) -> xr.Dataset:
...
+            Name of field monitor used in the original :class:`Simulation`.
+
+        Returns
+        -------
+        xarray.DataArray
+            DataArray containing the Poynting vector calculated at the center locations of the Yee
+            grid.
+        """
+
+        field_dataset = self.at_centers(field_monitor_name)
```
shashwat-sh, line 237: Might be good to mention in the docstring that the Poynting vector components are computed based on colocated fields at cell centers?

tidy3d/components/data/sim_data.py
```diff
@@ -228,26 +318,79 @@ def get_intensity(self, field_monitor_name: str) -> xr.DataArray:
...
+                raise DataError(
+                    f"Field monitor must contain '{field_name}x', '{field_name}y', and "
+                    f"'{field_name}z' fields to compute '{field_name}."
+                )
+            field_components = (dataset[c] for c in required_components)
+
+            # Apply the requested transformation
+            if val == "real":
+                derived_data = sum(f.real**2 for f in field_components) ** 0.5
+                derived_data.name = f"|Re{{{field_name}}}|"
```
shashwat-sh, line 345: Related to Tyler's comment, maybe this could behave differently depending on the number of components requested? E.g. if only one is requested, then just return `real(thing)` to preserve the sign, otherwise do what you have right now?

tidy3d/components/data/sim_data.py
```diff
@@ -298,17 +443,52 @@ def plot_field(  # pylint:disable=too-many-arguments, too-many-locals, too-many-
 
         # get the DataArray corresponding to the monitor_name and field_name
 
-        # intensity
+        # deprecated intensity
         if field_name == "int":
-            field_data = self.get_intensity(field_monitor_name)
-            val = "abs"
+            log.warning(
```
shashwat-sh, line 448: I like that you did this in a way that shouldn't break any existing notebooks or tests. But I think we'll still want to update several notebooks to avoid this warning, at the time of merging this PR?

tidy3d/components/types.py
```diff
@@ -180,6 +180,8 @@ def __modify_schema__(cls, field_schema):
 
 Ax = Axes
 PlotVal = Literal["real", "imag", "abs"]
+FieldVal = Literal["real", "imag", "abs", "abs^2", "phase"]
```
shashwat-sh, line 183: In order to maintain backwards compatibility (and in keeping with that warning you have right now), we may need to temporarily add `int` here? I think that pydantic might enforce type hints (@tylerflex correct me if I'm wrong), and my concern is that the plot functions won't accept `int` as a valid argument when a `FieldVal` is expected. If that's not true, then no worries.

tylerflex: pydantic won't enforce it for methods of the class (wish it did!) so it should be ok.

tidy3d/components/data/sim_data.py
lucas-flexcompute: This is on purpose, so that we get the correct sign on the Poynting vector: `Sy = Ez Hx - Ex Hz`

tidy3d/components/data/sim_data.py
lucas-flexcompute: This function returns a scalar derived field mainly for plotting. In that case, we cannot return a vector Re{E}, that's why we use the magnitude. But that is only because we use this function for plotting. If the user wants Re{E}, can't he just get it directly?

Maybe I should change the name of function to `get_scalar_field`, what do you think?

tidy3d/components/data/sim_data.py
lucas-flexcompute: Yes, this happens in [`_field_component_value`](https://github.com/flexcompute/tidy3d/blob/f03b372cf5a5dba345b50040492a9ab4489ad45c/tidy3d/components/data/sim_data.py#L265). As I mentined in the comment to Tyler, `get_derived_field` always returns scalar values (for plotting, originally), that's why this happens.

tidy3d/components/data/sim_data.py
tylerflex: I think whether the value is a scalar field is a bit irrelevant because note that the real part of a complex-valued scalar field is still a scalar field. There's no need for taking absolute value. Maybe what I'd imagine this function doing is handling it like if val = "real" or "imag", the field component must be specified (eg. E{x,y,z}) but otherwise if val="abs" we would take the operation over the vector field. Am I missing something? including @shashwat-sh as it seemed he had a similar comment

tidy3d/components/types.py
lucas-flexcompute: `'int'` is passed as a `field_name`, not `val`, so even if pydantic were to check this, it wouldn't be a problem, as `field_name` is a general `str`.

tidy3d/components/data/sim_data.py
lucas-flexcompute: I see what you're saying, but this function is only needed as a utility for `plot_field`, which requires a scalar field. I should not have left it as an external function: it should be renamed to `_get_scalar_field` probably. It was like that because I saw no easy way for the user to get the Poynting vector, so I thought it could be useful. However, exactly because of this issue of needing a function for helping with plotting, I decided to split the Poynting stuff and created `get_poynting_vector`, which returns the whole thing. So, there's really no reason for the user to use `get_derived_field`: it should just be a helper function for plotting at this point.

shashwat-sh: Ah ok I think I misunderstood actually: so if the user requests only `real` and `Ex`, for example, then `get_derived_field` will never be called, and Re{Ex} will be plotted correctly with the sign.

shashwat-sh: I guess the sign will only be lost in the case that the user request `real` and `E`, but it so happens that the monitor is only recording `Ex`.

shashwat-sh: Is that correct?

lucas-flexcompute: That's correct. If the user requests `"E"` it has to be a magnitude (considering E is a vector regardless of available components). The choice here is to take the magnitude of the complex, real or imaginary vectors.

The case where the monitor has only `"Ex"` and `"E"` is requested will raise an error, btw, because we need all components to calculate E.

shashwat-sh: Ok, yeah that makes sense to me then.

tylerflex: As long as there's still a way to plot the (signed) real part of the vector field components, eg Ex, I'm happy.

One thing I find slightly strange is when it would come up that the user would want to plot the magnitude of E but only taking the magnitude of the real part. I would assume of "real" and "E" is provided, we'd error. Can you think of a use case? I'm struggling to find one but if you think this is reasonably to provide, that's probably fine with me.

lucas-flexcompute: I've used this type of plot before to investigate modes in lossy waveguides and leaky modes, where it isn't the case that the tangential components are real and logitudinal, imaginary (for the usual phase normalization). I wanted to look at how the real and imaginary parts compared in magnitude, regardless of component. It is indeed a very uncommon case, but I don't think it hurts to have it.

tylerflex: ok, sounds good to me then thanks for clarifying

tidy3d/components/data/sim_data.py
lucas-flexcompute: Will do!

PR #764, opened 2023-03-16
Adjust time step size with medium property for stability

### LINKED ISSUE #761: Decreasing time step size in the presence of materials with eps_inf < 1
The maximal time step size for a stable simulation is constrained to a factor that scales with sqrt(eps_inf). So to determine the time step size, we'll need to loop through all structures to find out the minimal eps_inf.

@tylerflex I can think of two ways to obtain eps_inf: 1) `.eps_model(td.inf)`, but it will warn about frequency out of range; 2) add an attribute in `AbstractMedium` to record `eps_inf`. How do you think?

weiliangjin2021:
Resolves #761

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -124,6 +124,21 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         eps = self.eps_model(frequency)
         return (eps, eps, eps)
 
+    @cached_property
+    @abstractmethod
+    def n_cfl(self):
+        """To ensure a stable FDTD simulation, it is essential to select an appropriate
+        time step size in accordance with the CFL condition. The maximal time step
+        size is inversely proportional to the speed of light in the medium, and thus
+        proportional to the index of refraction. However, for dispersive medium,
+        anisotropic medium, and other more complicated media, there are complications in
+        deciding on the choice of the index of refraction.
+
+        This property computes the index of refraction related to CFL condition, so that
+        the FDTD with this medium is stable when the time step size that doesn't take
+        material factor into account is multiplied by ``n_cfl``.
+        """
+
     @add_ax_if_none
     def plot(self, freqs: float, ax: Ax = None) -> Ax:  # pylint: disable=invalid-name
         """Plot n, k of a :class:`Medium` as a function of frequency.
@@ -281,6 +296,14 @@ class PECMedium(AbstractMedium):
         # return something like frequency with value of pec_val + 0j
         return 0j * frequency + pec_val
 
+    @cached_property
+    def n_cfl(self):
+        """This property computes the index of refraction related to CFL condition, so that
+        the FDTD with this medium is stable when the time step size that doesn't take
+        material factor into account is multiplied by ``n_cfl``.
+        """
+        return 1.0
+
 
 # PEC builtin instance
 PEC = PECMedium(name="PEC")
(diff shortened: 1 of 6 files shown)


tidy3d/components/medium.py
```diff
@@ -124,6 +124,21 @@ def eps_diagonal(self, frequency: float) -> Tuple[complex, complex, complex]:
         eps = self.eps_model(frequency)
         return (eps, eps, eps)
 
+    @cached_property
+    @abstractmethod
+    def n_cfl(self):
+        """To ensure a stable FDTD simulation, it is essential to select an appropriate
+        time step size in accordance with the CFL condition. The maximal time step
+        size is inversely propertional to the speed of light in the medium, and thus
+        propertional to the index of refraction. However, for dispersive medium,
```
momchil-flex, line 133: typo: proportional

tidy3d/components/simulation.py
```diff
@@ -2147,7 +2147,12 @@ def dt(self) -> float:
         dl_mins = [np.min(sizes) for sizes in self.grid.sizes.to_list]
         dl_sum_inv_sq = sum((1 / dl**2 for dl in dl_mins))
         dl_avg = 1 / np.sqrt(dl_sum_inv_sq)
-        return self.courant * dl_avg / C_0
+        # material factor
+        n_cfl = 1.0
+        for mat in self.mediums:
+            n_cfl = min(n_cfl, mat.n_cfl)
+        print(n_cfl)
```
momchil-flex, line 2154: remove print statement

tidy3d/components/simulation.py
```diff
@@ -2147,7 +2147,12 @@ def dt(self) -> float:
         dl_mins = [np.min(sizes) for sizes in self.grid.sizes.to_list]
         dl_sum_inv_sq = sum((1 / dl**2 for dl in dl_mins))
         dl_avg = 1 / np.sqrt(dl_sum_inv_sq)
-        return self.courant * dl_avg / C_0
+        # material factor
+        n_cfl = 1.0
+        for mat in self.mediums:
+            n_cfl = min(n_cfl, mat.n_cfl)
```
momchil-flex, line 2153: a bit pedantic but probably better and more pythonic is `n_cfl = min([mat.n_cfl for mat in self.mediums], default = 1.0)`

tidy3d/components/medium.py
```diff
@@ -1088,6 +1146,16 @@ def components(self) -> Dict[str, Medium]:
...
 
+    @cached_property
+    def n_cfl(self):
+        """This property computes the index of refraction related to CFL condition, so that
+        the FDTD with this medium is stable when the time step size that doesn't take
+        material factor into account is multiplied by ``n_cfl``.
+
+        For this medium, it takes the minimal of ``n_clf`` in all components.
+        """
+        return min((mat_component.n_cfl for mat_component in self.components.values()))
```
momchil-flex, line 1157: I think this is fine, but conceptually, do you know if this can be loosened? For example for mesh step, we take some sort of average between the mesh steps along each of the three directions.

tidy3d/components/simulation.py
momchil-flex: More importantly, shouldn't we actually be able to remove the `default = 1.0`? `self.mediums` is never empty as it contains at least the simulation medium, and if all media in the simulation have n_cfl > 1, we could actually increase the time step?

tylerflex (approved):

tidy3d/components/simulation.py
weiliangjin2021: Tried a few simulations with dispersiveless medium, and all converges. But I think it might be problematic for dispersive medium for how we decide to apply supixel or not based on permittivity > 1.

momchil-flex: Oh, very good point. Ok yeah let's leave 1 as the minimum.

tidy3d/components/medium.py
weiliangjin2021: For uniform medium, it should be "courant/sqrt[sum_i 1/(n_cfl[i]*dl_min[i])**2]/C_0". But in my test, it diverges when the simulation domain is no longer uniform.

PR #765, opened 2023-03-17
Change version log from info to debug

HelgeGehring:
It would be great if we could only output the version number if the logging level is set to debug.

### DIFF
--- a/tidy3d/__init__.py
+++ b/tidy3d/__init__.py
@@ -112,4 +112,4 @@ def set_logging_level(level: str) -> None:
     )
 
 
-log.info(f"Using client version: {__version__}")
+log.debug(f"Using client version: {__version__}")


tylerflex:
Hi @HelgeGehring we are thinking of making the default logging level "warning" but keeping the version log as info. Does this work for you? See #767

HelgeGehring:
Thanks!

But that would lead to other problems I think. If you do it like that, you hide the INFO messages from other packages as well.
Just imagine numpy, scipy and matplotlob would each decide on their own for a certain standard logging level. Suddenly your output depends on the order how you import them :/

Therefore, in order to allow usage of tidy3D along with other tools, it would be best to not make any changes to the logging and leave that to the user. 

Packages should usually use the info only for things the user should be aware of but not concerned and can fix them, i.e. if there's a more efficient way of processing data for example.

Think of it solely from the user perspective: You want things to be silent if they just work. If you get problems as the user or you just want to understand better what's going on, you change the logging level.

tylerflex:
Hey @HelgeGehring , just to keep you up to date: with a recent change, we no longer use the `logging` module and instead Tidy3d has its own, completely independent logging functionality. So any changes we (or our users) make to the tidy3d logging should not have any affect on other dependencies.

In general, agree with the sentiment that things should be silent if they just work, which is why we're considering changing just the tidy3d-wide default logging level to "warning", which would not affect other packages and things depending on tidy3d and would have the bonus of not printing out the version (as it's level info).

Hope that makes sense, thanks

HelgeGehring:
Ups, sorry got over answering

and thanks for making it silent :)

I'd guess we could do it even a bit better: look at https://docs.python.org/3/library/logging.html
you can just add `logger = logging.getLogger(__name__)` in the beginning of each file and use that logger for logging - that way there's also information about which module generated the logging entry

as described in https://docs.python.org/3/howto/logging.html#configuring-logging you can then configure the logger for your package and set the levels

you could even think of introducing an extra logging level as described in https://docs.python.org/3/howto/logging.html#logging-levels , but setting it to Warning for your package is also fine :)

I'd guess sticking with the standard python logger will make it even easier to use and you'll get more features :)

PR #767, opened 2023-03-17
defaul log level warning, and a few minor changes

tylerflex:


momchil-flex:
Are we leaving this for 2.0?

momchil-flex:
We can (and will) merge current develop into pre/1.10.0 before the rc2 release. So this can be made to use the new logger. The question is just if this is a "breaking" beahvior, it seems to be in the grey zone, where code shouldn't break but experience will change. I'm fine with either 1.10 or 2.0.

tylerflex:
I'm fine with either, maybe for purposes of clearing out PRs we can merge it now? Feel free to do so if you agree.

PR #768, opened 2023-03-17
More info in AutoGrid docstring

weiliangjin2021:
Requested by @xin-flex, explain when the two parameters will take effect.


tidy3d/components/grid/grid_spec.py
```diff
@@ -435,7 +435,8 @@ class GridSpec(Tidy3dBaseModel):
         title="Free-space wavelength",
         description="Free-space wavelength for automatic nonuniform grid. It can be 'None' "
         "if there is at least one source in the simulation, in which case it is defined by "
-        "the source central frequency.",
+        "the source central frequency. "
+        "Note: it only takes effect when at least one of the three dimensions uses 'AutoGrid`.",
```
tylerflex, line 439: use single quotes on both sides of "AutoGrid" here.  otherwise looks good.

tylerflex (approved):

tidy3d/components/grid/grid_spec.py
weiliangjin2021: :class:`.AutoGrid` seems to work in description. How about switch to this style to provide link in the doc?

tylerflex: right, that seems better actually. feel free to merge

PR #775, opened 2023-03-22
add auth.json migration to API key

An-Li-magicloud:
add migrate function and migration cli for user who has auth.json file.
It will asked user to migration from credential to API Key


tylerflex (approved):
Looks good, seems to work when I tested.
I edited the prompt instructions to add a bit more information about what this migration is doing, but from my perspective looks ready to merge after I get one other person to test. Thanks @An-Li-magicloud

momchil-flex (approved):

PR #776, opened 2023-03-22
#SCEM-2279 create .tidy3d folder after pip install

An-Li-magicloud:


An-Li-magicloud:
> @An-Li-magicloud I just tested this by removing the `.tidy3d` directory and running a local `pip install -e .` and it did not create the `~/.tidy3d` folder on my machine. I'm wondering if I did something wrong? Also, just pushed a commit to run the `black` code formatter.

I have changed approach, but it seems there is a unit test error happens in components testing which I'm not able to solve.

tylerflex (approved):

PR #779, opened 2023-03-22
fix bug in Simulation.eps_bounds with CustomMedium

tylerflex:
@weiliangjin2021 not sure how this bug got in, but it was breaking plotting with CustomMedium. I think this is the only area that needs fixing but maybe you can confirm. Thanks

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -1438,8 +1438,8 @@ class Simulation(Box):  # pylint:disable=too-many-public-methods
         for mat in [medium for medium in medium_list if isinstance(medium, CustomMedium)]:
             eps_dataset_at_freq = mat.eps_dataset_freq(freq)
             for eps_component in eps_dataset_at_freq.field_components.values():
-                eps_min = min(eps_min, eps_component.real.values.ravel())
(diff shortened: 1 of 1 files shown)


weiliangjin2021:
It seems to be removed in a commit `fix errors caught  by pylint 2.16.1`. Previously it was `eps_min = min(eps_min, min(eps_component.real.values.ravel()))`.

weiliangjin2021:
do you think `np.min(eps_component.real.values.ravel())` will return dataarray type so that it will error at `min(float, dataarray)`? It occurs here https://github.com/flexcompute/tidy3d/blob/0cf25febb41dc37b0a7991a66a005267c8fba282/tidy3d/components/medium.py#L452 when `float` is removed.

tylerflex:
> do you think np.min(eps_component.real.values.ravel()) will somtimes return dataarray type so that it will error at min(float, dataarray)?

Isn't `.values` always a `np.ndarray`? or can it be a `dataarray`?

weiliangjin2021:
You are right. Feel free to merge.

weiliangjin2021 (approved):

PR #783, opened 2023-03-22
Consider a polygon to be degenerate when its area is <fp_eps

weiliangjin2021:


weiliangjin2021:
The failure is now related to API key for material fitter @tylerflex

momchil-flex:
Now the material fitter test fails cause of API key @tylerflex ...

tylerflex:
See #784 , should fix it.

PR #790, opened 2023-03-23
Treat infinite slanted cylinder and add validators

weiliangjin2021:
For a slanted cylinder of infinite length, the reference plane can only lie in the middle. Otherwise, the behavior of cylinder at the `center` is confusing. A validator has been added.

Additionally, when `self.length_axis` is td.inf, internally in cylinder, we use `LARGE_NUMBER` instead. This is to make sure for a slanted cylinder, `radius_max` etc. are well defined. 

Finally, override the `radius` attribute in cylinder to clarify where the radius is defined.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SourceTime` plotting methods `.plot()` and `.plot_spectrum()` accept a `val` kwarg, which selects which part of the data (`'real'`, `'imag'`, or `'abs'`) to plot, rather than plotting all at once.
 - Tidy3D exceptions inherit from `ValueError` so they are handled properly by pydantic.
 - Two unstable materials in `material_library`: `Cu_JohnsonChristy1972` and `Ni_JohnsonChristy1972`. `TiOx_HoribStable` added for improved stability.
+- Bug in infinite long cylinder when the `reference_plane` is not at the bottom or the cylinder is slanted.
 
 ## [1.10.0rc1] - 2023-3-07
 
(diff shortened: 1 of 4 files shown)


lucas-flexcompute (commented):
I'm not a big fan of replacing `inf` with `LARGE_NUMBER` because there's always the chance that someone will find a situation where `LARGE_NUMBER` is not large enough. The question here is why do we need to calculate `radius_max`? If that's for bounding box calculation, for example, it doesn't really help, as the bounding box will end up large enough for the whole domain, so why not return `inf` directly? These might be much deeper changes that we might consider later on. For now, I have no simpler suggestions.

lucas-flexcompute (approved):

tylerflex (approved):

weiliangjin2021:
> I'm not a big fan of replacing `inf` with `LARGE_NUMBER` because there's always the chance that someone will find a situation where `LARGE_NUMBER` is not large enough. The question here is why do we need to calculate `radius_max`? If that's for bounding box calculation, for example, it doesn't really help, as the bounding box will end up large enough for the whole domain, so why not return `inf` directly? These might be much deeper changes that we might consider later on. For now, I have no simpler suggestions.

It's also needed in constructing the polygon for visualization in `_intersections_side`. When radius_max is inf, there will be `nan` in the polygon coordinate.

momchil-flex:
I think this looks good for a merge now, but either as part of it or separately we wanted to add a deprecation warning that reference plane will change to "middle" in 2.0. @tylerflex what do you think, should we merge this in now and then quickly add this?

tylerflex:
@momchil-flex I'm pretty sure we do already have this deprecation warning. I removed it in #793

momchil-flex:
Oh - great! Ok let's merge this in then.

tylerflex:
For reference: https://github.com/flexcompute/tidy3d/blob/pre/1.10.0/tidy3d/components/geometry.py#L994

tylerflex:
@weiliangjin2021 is the backend ready for this to be merged? if so, go ahead and merge it

PR #793, opened 2023-03-24
API changes for 2.0

tylerflex:
Removes all deprecation warnings.
Sets boundary conditions to PML by default.
Sets polyslab reference plane to "middle" by default.
Doesn't allow lowercase log levels 
Removes geometry.intersections()

Will require changes to the docs, awaiting 1.10 release to begin on that to avoid merge conflicts.

Please confirm that the changes reflect the behavior we expect for 2.0, thanks.


shashwat-sh:
https://github.com/flexcompute/tidy3d-core/pull/266
This PR changes backend tests to make all boundary specifications explicit. I'm not able to run tests yet - I think something is off with my auth setup, but just linking it here for later.

weiliangjin2021 (approved):

PR #798, opened 2023-03-25
first pass trying to fix web fitter with new webapi

tylerflex:


tylerflex:
Already fixed by #800

PR #801, opened 2023-03-27
1.10.0 version, changelog and schema

momchil-flex:


momchil-flex:
Thanks, I had actually done that but forgot to push :)

Issue #805, opened 2023-03-28
importlib-metadata version


zeyueN:
We have a minor dependency conflict from `importlib-metadata` and traced back to [this requirement](https://github.com/flexcompute/tidy3d/blob/ad1876910a1f720465989f8278928953b13ff915/requirements/basic.txt#L5). We are wondering if the comment is still true and if the locking to v4.13 can be lifted now. Thanks a lot!

tylerflex:
Front end tests all pass with `importlib-metadata==6.1.0` @momchil-flex maybe we should run backend tests just to be safe, then I have no problem bumping the requirements.

@zeyueN , just so you know `importlib-metadata==5` seemed to break `xarray`, but it must have been fixed. Maybe it's possible for you to install `importlib-metadata==6.1.0` after installing tidy3d until we merge this in. We are working on a major release now so it might not happen right away.

zeyueN:
Good to know. Thanks for the update!

tylerflex:
@zeyueN we bumped the version in our 2.0 branch and so it will be released along with 2.0 today.

https://github.com/flexcompute/tidy3d/pull/802/commits/39895dd3d86ddb0860c6d222aa1f34732fb3978d

Will close this issue

PR #812, opened 2023-03-30
Unpinning importlib-metadata version

momchil-flex:
Backend tests pass with this. Did you want to wait until after 2.0 though @tylerflex ?


tylerflex:
Also, I'd recommend making it >= 6.0.0 maybe? because technically > 4.13 will include 5.0 stuff which breaks xarray at least last I checked.

momchil-flex:
Ok. Bumped to >= 6.0 and tagged for 2.1. Fixes #805

PR #816, opened 2023-03-31
Improve web error display

An-Li-magicloud:
image

### DIFF
--- a/tidy3d/web/httputils.py
+++ b/tidy3d/web/httputils.py
@@ -1,5 +1,5 @@
 """Handles communication with server."""
-
+import json
 import os
 import time
 from enum import Enum
@@ -103,13 +103,12 @@ def handle_response(func):
         if resp.status_code == ResponseCodes.UNAUTHORIZED.value:
             raise WebError(resp.text)
 
-        resp.raise_for_status()
         json_resp = resp.json()
 
         # if the response status is still not OK, try to raise error from the json
         if resp.status_code != ResponseCodes.OK.value:
             if "error" in json_resp.keys():
-                raise WebError(json_resp["error"])
+                raise WebError(json.dumps(json_resp, indent=4))
             resp.raise_for_status()
 
         return json_resp["data"] if "data" in json_resp else json_resp


momchil-flex:
Why not just `json.dumps(json_resp)["error"]` to only show the error message?

momchil-flex:
Wait what I said doesn't make sense. I guess I don't understand why the old version doesn't work to just print the error message? It seems like it should be present in the dictionary. Maybe just removing the `resp.raise_for_status` line is enough?

An-Li-magicloud:
> Wait what I said doesn't make sense. I guess I don't understand why the old version doesn't work to just print the error message? It seems like it should be present in the dictionary. Maybe just removing the `resp.raise_for_status` line is enough?

Either way, remove `resp.raise_for_status` works but the message doesn't easy to read:

momchil-flex:
I don't know, the second option seems more standard to me, but it's true that the message is more visible in the first option. @tylerflex @xin-flex ?

tylerflex:
Hm, so I'm not sure I understand. Are you saying that in the current state of the code, it will print
image

and in the PR version it will print
image

Joaquin's message seemed to miss the cause of the error entirely, maybe something changed on the backend since then?
image

Honestly, both 1st and 2nd are probably sufficient, but it might be better to display the message more prominently. Maybe we can raise a custom exception on the front end `if code == 4000000003` `from` the WebError that is created out of the json? Not sure if that would improve things, for example:

```py
             if "error" in json_resp.keys():
                 e = WebError(json.dumps(json_resp, indent=4))
                 if "json_resp["code"] == 4000000003:
                     raise WebError("Insufficient credits.") from e
                 raise e
             resp.raise_for_status()
```
not sure if this is even valid python but if that improves readability maybe it's worth a try?

Otherwise, if the 1st screenshot is what we see currently, might want to just merge this after 2.0 in case it breaks anything.

An-Li-magicloud:
We have reformat Java response in a json, just like the first screenshot, which is not only for balance, but all other api. So If you counter any error without json format, just like me know.

So we could just print out either entire json or just ``json["error"]``, but current code won't do neither because ``resp.raise_for_status()`` prevent that happens.

tylerflex:
I see, thanks. I think maybe the 2nd option (`raise WebError(json["error"])`) might be my preference then. @An-Li-magicloud since we are releasing 2.0 today, do you think it's unlikely that this change will cause any bugs? just want to make sure because we did most of our testing earlier this week.

tylerflex:
So yea I guess just removing the `resp.raise_for_status` line (like @momchil-flex said) would do the trick. I just tested a branch with that change and it worked fine (didn't test the actual error handling because I'm not sure how) but I'd say let's just remove that line and that should be good enough.

tylerflex:
I made a PR with just this change + a changelog item so we could merge that if we go this route #817

tylerflex:
ok merged #817 and so I'm closing this. Thanks @An-Li-magicloud for helping us figure this out and fix it!

PR #821, opened 2023-04-04
filter adjoint source based on amplitude relative to max

### LINKED ISSUE #820: Avoid placing adjoint ModeSource-s for modes with very low amplitude in the forward simulation
A common usage of a `ModeMonitor` in an adjoint optimization may be to e.g. define the objective function as the power in a target mode over the sum of power in all other modes. This makes the OF dependent on the amplitudes of all outgoing modes, and places one adjoint source for each of them. This can currently make things noticeably inefficient in the pre-processing, especially when the user has selected a large number of modes in the `mode_spec`. We could improve on that in the backend, but a simple improvement is also to not add sources that will have very small amplitude compared to other ones. For example, only add a source if the associated mode_amp is larger than some cutoff threshold times the largest mode_amp recorded. Or, probably better, set the cutoff based on the gradient vector.

tylerflex:
Fixes #820

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- `adjoint` plugin now filters out adjoint sources that are below a threshold in amplitude relative to the maximum amplitude of the monitor data, reducing unnecessary processing by eliminating sources that won't contribute to the gradient.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


momchil-flex:
Looks good. Only thing is that I tried a simulation, and for many modes that I would expect no contribution, I get amplitudes above the current threshold, like on the 1e-6 scale. Note that in terms of power carried by that mode (amplitude squared), this is 1e-12, so it should still be completely negligible. My opinion is that we can set the threshold to e.g. 1e-6, but I understand if you want to be conservative about this.

momchil-flex (approved):

tylerflex:
I set it to 1e-6. Honestly, I don't have a good feeling for what it should be so I'll trust your judgement. Hoping this doesn't end up causing silent issues down the line, but we'll see.

PR #822, opened 2023-04-04
server side adjoint

tylerflex:
Here is a plan for how to change the current adjoint pipeline to do everything server side.

Basic idea: replace the `adjoint.run` and it's `fwd`, `bwd` versions with their own web API.

* on the `fwd` pass
    * store the output monitor data, associated with the task_id of the task.
    * return the original data (not including monitor data) + the task id of the task
* on the `bwd` pass
    * webAPI call passes the vjp sim_data to server, + the task_id of the fwd pass.
    * on the backend, we grab the `fwd` data using task_id
    * do the regular adjoint processing.
    * return the sim_vjp to the user.

I defined the various webapi calls and functions needed to implement in the docstrings.

Any thoughts about feasibility / edge cases or complications you can see before breaking ground on this? The main thing I'm not sure about is how feasible it is how to store the fwd data associated with a task_id. would that require new infrastructure?

Also, would the best thing be to get MC involved to implement these functions for us?


tylerflex:
A few things I did in the recent commits:
* adjoint tests save data to `tests/tmp` instead of `tidy3d` directory
* add `sim_fwd, jax_info_fwd, jax_info = JaxSimulation.to_simulation_fwd()`, which combines the `to_simulation()` with a call to process results into fwd. Declined to add a `grad_monitors: bool` flag in `to_simulation()` because the number of outputs is different, this seemed cleaner.
* cleared up a few rough edges in the code / added some comments.
* tests pass w/ and w/out monkeypatch, and tested on notebooks with a slightly more complicated treatment for async in tutorial 4 to verify that it was grabbing the proper fwd_sim data.

Think it seems basically good to go except for the following:

To do @momchil-flex 
* some comments on the hardcoded file paths for SIM_VJP and JAX_INFO.
* solver version handling (no longer hardcoded) <- merge tidy3d-core PR?
* change base of PR to pre/2.2, rebase, and fix merge conflict (either of us could do this).

Any other cleanup you think is necessary, just comment I guess.

momchil-flex:
> * some comments on the hardcoded file paths for SIM_VJP and JAX_INFO.

Yeah these are defining some file names on the server. It's like [this](https://github.com/flexcompute/tidy3d/blob/3f264364a803e04b842a554d23511d2cd7230aca/tidy3d/web/webapi.py#L30-L31) and [this](https://github.com/flexcompute/tidy3d/blob/3f264364a803e04b842a554d23511d2cd7230aca/tidy3d/web/simulation_task.py#L20-L24) in the regular tidy3d web. Which makes me think btw that in webapi.py we shouldn't be defining `SIM_FILE_JSON` but just import `SIMULATION_JSON` (renamed to `SIM_FILE_JSON`) from `simulation_task.py` (we are already importing `SIM_FILE_HDF5` from there). I'm suggesting going with `SIM_FILE_JSON` to match `SIM_FILE_HDF5`, which is the `Simulation` hdf5, and not to be confused with `SIMULATION_HDF5`, which is the `SimulationData` file. Probably best to rename that to `SIM_DATA_HDF5`. Not confusing at all, right? :)

So yeah these are similarly the file names needed for adjoint. I don't know if there's a better way to handle all this?

Issue #819, opened 2023-03-31
Group index computation from mode solver


momchil-flex:
Would be helpful to add some in-build functions to compute and plot the group index.

I am not sure if there's anything more to it than taking a numerical derivative of the dispersion?

lucas-flexcompute:
There is something already there in https://github.com/flexcompute/tidy3d/pull/743: https://github.com/flexcompute/tidy3d/blob/df4c5ce5aacbdac9c2b5ef7d36e2d2f5586fb370/tidy3d/plugins/waveguide/rectangular_dielectric.py#L650

The issue is that the numerical differentiation is sensitive to the grid changes without subpixel averaging. We also have to be careful with mode crossings, so all calculations (all required wavelengths for differentiation) have to be performed in a single run because of the mode sorting.

All that said, it should be pretty straightforward to port the code from the plugin to mode solver.

PR #828, opened 2023-04-05
Add `ModeSolver.solve_group_index` (#819)

lucas-flexcompute:
Calculation was added to `ModeSolver`, because it is not always required, so it doesn't make much sense to pay the price of the calculation if we're just interested in the effective index. I don't know if there's a better way to to it.


tidy3d/plugins/mode/mode_solver.py
```diff
@@ -116,6 +116,58 @@ def solve(self) -> ModeSolverData:
...
+        frac_step : float
+            Fractional frequency step used for numerical differentiation.
+
+        Returns
+        -------
+        :class:`ModeIndexDataArray`
+            Group indices for all waveguide wavelengths and modes indices.
+        """
+        if frac_step >= 1:
+            raise ValidationError("Argument `frac_step` must be less than 1.")
```
momchil-flex, line 137: We use single quotes, i.e. `'frac_step'` formatting in error/warning messages.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -116,6 +116,58 @@ def solve(self) -> ModeSolverData:
...
+        # of mode tracking.
+        freqs = np.outer(self.freqs, (1 - frac_step, 1, 1 + frac_step)).flatten()
+        end = freqs.size
+
+        mode_solver = self.copy(update={"freqs": freqs})
+        n_backward = mode_solver.data.n_eff.isel(f=slice(0, end, 3)).values
+        n_center = mode_solver.data.n_eff.isel(f=slice(1, end, 3)).values
+        n_forward = mode_solver.data.n_eff.isel(f=slice(2, end, 3)).values
+
+        n_group = n_center + (n_forward - n_backward) / (2 * frac_step)
```
momchil-flex, line 161: I guess if we want to save 1/3 of the solves we could also do

```
n_center = (n_forward + n_backward) / 2
n_group = n_center + (n_forward - n_backward) / (2 * frac_step)
```

I would think this should not have any noticeable effect on precision, what do you think?

momchil-flex (commented):
I think this is generally OK, but an alternative could also be to just have a `compute_group_index=False` kwarg in `ModeSolver.solve`?

Also, could you add a test?

lucas-flexcompute:
> I think this is generally OK, but an alternative could also be to just have a `compute_group_index=False` kwarg in `ModeSolver.solve`?

In that case, do you mean to add a field like `n_group` to `ModeSolverDataset`, such that when `compute_group_index == True` it gets calculated, but otherwise it is set to `None`?

tidy3d/plugins/mode/mode_solver.py
lucas-flexcompute: Honestly I'm not a big fan of this… It's hard to say how much of an impact it has (each waveguide and wavelength region would be impacted differently) but there's definitely some impact. I think a 3-point derivative is not that expensive and gives reasonable results (I actually though about using a 5-point derivative, but given the errors in the n_eff, I thought it might be too expensive for no real gain).
Now, I know I'm kind of conservative with these things and tend to err on the side of caution, so my opinion might not be what most users want…

momchil-flex: Fair enough, also it may be moot depending on what we decide about whether to have it as part of the regular `solve`, or as a standalone `solve_group_index`.

momchil-flex:
> > I think this is generally OK, but an alternative could also be to just have a `compute_group_index=False` kwarg in `ModeSolver.solve`?
> 
> In that case, do you mean to add a field like `n_group` to `ModeSolverDataset`, such that when `compute_group_index == True` it gets calculated, but otherwise it is set to `None`?

What do you think @tylerflex ? Have a separate call to compute the group index, or optionally compute it during the regular `solve`, but then have the `n_group` part of `ModeSolverDataset` be `None` by default, unless group index is requested? This doesn't sound too bad to me?

One other point: if we do it this way, we could also add the `compute_group_index` kwarg to the `ModeSolverMonitor`, and have a way to get the group index with subpixel averaging.

momchil-flex:
In some sense, I feel like we should scrap `ModeSolverMonitor` altogether, and just have `ModeMonitor` with flags for `store_amps`, `store_effective_index`, `store_fields`, `store_group_index`... Where the first two are on by default and the last two are off by default. I remember we considered something like that a while ago but opted for the ModeSolverMonitor I believe to better separate things. But I kinda like the flag method all things considered. For example, it also won't necessitate two separate mode solves for a ModeMonitor and a ModeSolverMonitor at the same place, if that's what the user wants. Am I missing some downside @tylerflex ?

lucas-flexcompute:
I can see one possibly unexpected issue. The user won't be able to change his mind after solving because we cache the data:
```python
data = mode_solver.solve()
assert data.n_group is None
new_data = mode_solver.solver(compute_group_index=True)
assert new_data.n_group is None
```

tylerflex:
As discussed, my vote is to keep the group index calculation as a separate method. I see potential pitfalls with changing the resulting data based on kwargs. It seems much more straightforward to me to add an optional method and then not have to guess / check whether the group index data is present. This was similar to the reasoning behind changing the monitors to always store the same data, instead of toggling based on the `store` keyword if you remember that @momchil-flex.

lucas-flexcompute:
If that's the final decision, I've already included a test for the calculation and fixed the warning message, so this should be ready to merge.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -71,6 +82,30 @@ def freqs_not_empty(cls, val):
...
+            return GROUP_INDEX_STEP
+        return val
+
+    def _post_init_validators(self):
+        if self.compute_group_index >= 1:
+            raise ValidationError("Argument 'compute_group_index' must be less than 1.")
+
+        if self.compute_group_index > 0:
+            if self.mode_spec.track_freq is None:
+                log.warning(
```
tylerflex, line 98: I think it would improve usability if we could tell the user specifically what field to set in these warnings. For example, by adding the sentence `Consider setting the 'ModeSpec.track_freq' field to __.` Same for the precision. `Consider setting the 'ModeSpec.precision' field to 'double'`.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -116,6 +151,7 @@ def solve(self) -> ModeSolverData:
         """
         return self.data
 
+    # pylint: disable=too-many-locals
```
tylerflex, line 154: I think it could improve readability to break this method up. or at least the group index portion of it. is it possible? Then I took a look at the full `data_raw` and noticed that it's already very long and hard to follow. @momchil-flex maybe this is pretty low priority but the mode solver code probably could be cleaned up a bit in the future. @lucas-flexcompute I'll leave it to you to decide if it's worth it, but I do think throwing a separate method to compute group index could really improve readability

```py
def data_raw(self) -> ModeSolverData:
    if self.compute_group_index > 0:
        return self._compute_group_index()
    ...
```

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -71,6 +82,30 @@ def freqs_not_empty(cls, val):
...
 
+    @pydantic.validator("compute_group_index")
+    def assign_default_on_true(cls, val):
+        """Assing the default fractional frequency step value if not provided."""
+        if val is True:
+            return GROUP_INDEX_STEP
+        return val
+
+    def _post_init_validators(self):
+        if self.compute_group_index >= 1:
```
tylerflex, line 93: might be better to throw this in the `assign_default_on_true` validator and it could be caught earlier. For example
```py
@pydantic.validator("compute_group_index")
def assign_default_on_true(cls, val):
    """Assing the default fractional frequency step value if not provided."""
    if val is True:
        val = GROUP_INDEX_STEP
    if val >= 1:
        raise ValidationError("Argument 'compute_group_index' must be less than 1.")
    return val
```

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -71,6 +82,30 @@ def freqs_not_empty(cls, val):
...
+            raise ValidationError("Argument 'compute_group_index' must be less than 1.")
+
+        if self.compute_group_index > 0:
+            if self.mode_spec.track_freq is None:
+                log.warning(
+                    "Group index calculation without mode tracking can lead to incorrect results "
+                    "around mode crossings."
+                )
+
+            if self.mode_spec.precision != "double":
```
tylerflex, line 103: might be better to do this as a root validator

```py
@pd.root_validator()
def check_precision(self, values):
     if values["compute_group_index"] and values["precision"] != "double":
         log.warning(...)
```

tylerflex (changes requested):
just a few changes, otherwise looks good. thanks!

tidy3d/plugins/mode/mode_solver.py
lucas-flexcompute: Sure thing, I've separated the group index calculation as you suggested.

tylerflex (approved):

PR #829, opened 2023-04-06
extra condition before printing early shutoff log

tylerflex:
For some reason, the run info webapi seems to have started returning: 
perc_done = 0
field_decay = 0
early in the task. (maybe before the solver progress csv is ready).

As a result, the condition for printing the early shutoff warning:
```py
if perc_done is not None and perc_done < 100:
```
was being triggered even if the task was finishing without early shutoff.

This PR changes to only trigger if 
```py
if perc_done is not None and perc_done < 100 and field_decay > 0:
```
basically if the solver progress is not present and the task has stopped running, it doesn't print this log message anymore. We should follow up with MC though.


tomflexcompute (approved):
Tested a few cases with very short, intermedium, and long run_time. All worked correctly.

PR #830, opened 2023-04-06
add finite extent to PML for dims where simulation.size == 0

tylerflex:
Makes the PML boxes (for plotting) have a size of `0.02` along any dimensions where sim.size == 0.
Otherwise, there are times where the PML will not show, eg plotting a simulation with size (1, 1, 0) along the z axis, because the PML has a size of 0 along z and the box does not get picked up by the plotter.

note, used a small number instead of td.inf to avoid potential weirdness in the 3D or cross section plots.
Also, the plotting wasn't picking up any number smaller than 0.02..

Before:
image

After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Web functions create the leading directories for the supplied filename if they don't exist.
 - Some docstring examples that were giving warnings.
 - `web.monitor()` only prints message when condition met.
+- PML boxes have non-zero extent along any dimensions where the simulation has 0 size, to fix plotting issues for 2D simulations.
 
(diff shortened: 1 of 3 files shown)


lucas-flexcompute (approved):
Is the 0.02 special in any way? From your example it looks really thin, so , if we want to have it more visible, maybe set it to 100 nm, or at least 50 nm (that's usually around the minimum feature size for most optical processes)

tylerflex:
So the 0.02 is just the height of the plotting box out of plane of the 2D plot. (see the images in the PR description). So it's not physical. The reason I chose 0.02 is because I needed some non-zero value for the shapely / plotting to register it, but if I made it too small (even 0.01) it would not show up in the plots. I wanted to keep it small enough that it would still look relatively 2D in 3D plots.

PR #833, opened 2023-04-11
robust handling for 2D medium structure validator

tylerflex:
Previous 2D medium validator for `td.Structure` grabbed `values["geometry"]` before checking if it was needed. 

The problem with this is that it `Structure.geometry` fails validation for some reason, the `values` dictionary will not contain a `geometry` key and we'll get an obscure KeyError.

This PR changes it so that:
* `values["geometry"]` is only accessed when it's needed (if the `Structure.medium` is determined to be a 2D medium).
* we use `values.get("geometry")` and raise a more helpful error if this is None, directing the user to fix the `.geometry` field.


caseyflex (approved):
looks good, thanks for the fix!

PR #835, opened 2023-04-12
revert to storing ArrayLike as np.ndarray internally

tylerflex:


lucas-flexcompute (approved):
I guess [simulation_2_1_0.json](https://github.com/flexcompute/tidy3d/pull/835/files#diff-03ac8a8820ae6894cecf38b1cfa1e9c0e87c4190748d8791cc4c7e1af9b510c6) is not part of this, is it?

tylerflex:
That json file is just generated when running the updater tests (if not present already) in test_io.py. Since it will need to be generated before the 2.1 release anyway, let's just include it.

PR #836, opened 2023-04-12
print task url for web ui after upload

tylerflex:


xin-flex (approved):
It will be better if the link is clickable, but current format is sufficiently good.

tylerflex:
@xin-flex I figured out how to make it clickable using a `rich` formatting feature. Seemed to work for me on jupyter notebooks, feel free to give it another try if you want.

PR #841, opened 2023-04-14
add python function `web.configure(apikey)` to configure

tylerflex:
In a hosted jupyter notebook environment (eg google colab), it may be more convenient to install and configure via the following lines at the top of the notebook.

```
!pip install tidy3d
import tidy3d.web as web
web.configure("XXX")
```

note, README is updated to include these instructions as well as the Quickstart in the docs.


tomflexcompute (approved):
Works very well and it's a great additional way to do authentication. 

As long as the warning message can be gotten rid of, should be good to go!

tomflexcompute:
> Thanks Tom, I asked MC to fix the warning message in my JIRA post, along with making the API key shorter, so maybe that can be a separate PR.

Sure sounds good. Thanks!

PR #842, opened 2023-04-17
more robust handling of `web.Batch.monitor()`

tylerflex:
changes
* the batch monitor makes sure the status is in the run_statuses list before indexing. If not in the list (error, diverge), it will set to 100%, no exception will be raised.
* if error or diverge, the final description in the progress bar text will be red to indicate an issue.

### DIFF
--- a/tidy3d/web/container.py
+++ b/tidy3d/web/container.py
@@ -383,7 +383,14 @@ class Batch(WebContainer):
         """
 
         def pbar_description(task_name: str, status: str) -> str:
-            return f"{task_name}: status = {status}"
+            """Make a progressbar description based on the status."""
+            description = f"{task_name}: status = {status}"
+
+            # if something went wrong, make it red
+            if "error" in status or "diverge" in status:
+                description = f"[red]{description}"
+
+            return description
 
         run_statuses = [
             "draft",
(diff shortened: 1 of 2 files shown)


tylerflex:
Just added conditions in `Batch.load` and `Batch.download` while looping through the tasks for loading into `BatchData` to check if `"error"` is in  `job.status` and if so, print a warning and exclude that task from the `BatchData`.

tidy3d/web/container.py
```diff
@@ -493,8 +508,13 @@ def download(self, path_dir: str = DEFAULT_DATA_DIR) -> None:
 
         self.to_file(self._batch_path(path_dir=path_dir))
 
-        for job in self.jobs.values():
+        for task_name, job in self.jobs.items():
             job_path = self._job_data_path(task_id=job.task_id, path_dir=path_dir)
+
+            if "error" in job.status:
+                log.warning(f"Now downloading '{task_name}' as the task errored.")
```
momchil-flex, line 515: Now -> Not

momchil-flex (approved):
Code looks good, hopefully it also works as expected :)

tidy3d/web/container.py
tylerflex: thanks, fixed in recent force push

tylerflex:
I tested it with non-erroring webAPI notebook (and also manually triggered the warning conditions as an extra test) and it worked as expected.

PR #846, opened 2023-04-18
Fully anisotropic materials

dbochkov-flexcompute:
Frontend interface for fully anisotropic materials:
- Since `AnisotropicMedium` was already taken, fully anisotropic materials are called `FullyAnisotropicMedium`
- Currently only materials for which main directions of permittivity and conductivity (symmetric part) tensors coincide (to have well-defined main complex permittivities)
- A helper class `RotationAroundAxis(AbstractRotation)` is implemented for convenient rotation of tensors in space. In future we can add more methods like `RotationEulerAngles(AbstractRotation)`, etc.
- Automatic mesher now takes into account only the max eps among all directions (previous treatment based on normal direction resulted in underresolving because tangential direction define wavelength)
- For now injection of plane waves and gaussian beams into anisotropic materials is prohibited
- When plotted, `FullyAnisotropicMedium` returns main components
- Slight modification of `ModeSolver` so that it is feeded with the full permittivity tensor
- To help that new method `eps_comp` is introduced in material classes to retrieve a specified component of a permittivity tensor

Review todo's:
- [x] Unify rotation functionality
- [x] Generalize `ArrayLike`
- [x] Standardize order of components in mode solver
- [x] changelog
- [x] unify plotting for `AnisotropicMedium` and `FullyAnisotropicMedium`
- [ ] converge on naming/unification of `AnisotropicMedium` and `FullyAnisotropicMedium`

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -1359,6 +1359,13 @@ class FullyAnisotropicMedium(AbstractMedium):
     magneto-optic effects. Note that dispersive properties and subpixel averaging are currently not
     supported for fully anisotropic materials.
 
+    Note
+    ----
+    Simulations involving fully anisotropic materials are computationally more intensive, thus,
+    they take longer time to complete. This increase strongly depends on the filling fraction of
+    the simulation domain by fully anisotropic materials, varying approximately in the range from
+    1.5 to 5. Cost of running a simulation is adjusted correspondingly.
+
     Example
     -------
     >>> perm = [[2, 0, 0], [0, 1, 0], [0, 0, 3]]
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -13,16 +13,17 @@ import xarray as xr
 from .base import Tidy3dBaseModel, cached_property
 from .grid.grid import Coords, Grid
 from .types import PoleAndResidue, Ax, FreqBound, TYPE_TAG_STR, InterpMethod, Bound, ArrayComplex4D
-from .types import Axis, Numpy, Coordinate, RealTensor
+from .types import Axis, TensorReal
 from .data.dataset import PermittivityDataset
 from .data.data_array import ScalarFieldDataArray
 from .viz import add_ax_if_none
 from .geometry import Geometry
 from .validators import validate_name_str
 from ..constants import C_0, pec_val, EPSILON_0, LARGE_NUMBER, fp_eps
-from ..constants import HERTZ, CONDUCTIVITY, PERMITTIVITY, RADPERSEC, MICROMETER, SECOND, RADIAN
+from ..constants import HERTZ, CONDUCTIVITY, PERMITTIVITY, RADPERSEC, MICROMETER, SECOND
 from ..exceptions import ValidationError, SetupError
 from ..log import log
+from .transformation import RotationType
 
 
 # evaluate frequency as this number (Hz) if inf
@@ -136,10 +137,10 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
 
         Parameters
         ----------
-        row : Axis
-            Component's row in the permittivity tensor.
-        col : Axis
-            Component's column in the permittivity tensor.
+        row : int
+            Component's row in the permittivity tensor (0, 1, or 2 for x, y, or z respectively).
+        col : int
+            Component's column in the permittivity tensor (0, 1, or 2 for x, y, or z respectively).
         frequency : float
             Frequency to evaluate permittivity at (Hz).
 
@@ -151,9 +152,8 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
 
         # This only needs to be overwritten for anisotropic materials
         if row == col:
-            eps = self.eps_model(frequency)
-            return eps
-        return 0
+            return self.eps_model(frequency)
+        return 0j
 
     @cached_property
     @abstractmethod
@@ -569,16 +569,16 @@ class CustomMedium(AbstractMedium):
         col: Axis,
         frequency: float,
         coords: Coords,
-    ) -> Numpy:
+    ) -> ArrayComplex4D:
         """Spatial profile of a single component of the complex-valued permittivity tensor at
         ``frequency`` interpolated at the supplied coordinates.
 
         Parameters
         ----------
-        row : Axis
-            Component's row in the permittivity tensor.
-        col : Axis
-            Component's column in the permittivity tensor.
+        row : int
+            Component's row in the permittivity tensor (0, 1, or 2 for x, y, or z respectively).
+        col : int
+            Component's column in the permittivity tensor (0, 1, or 2 for x, y, or z respectively).
         frequency : float
             Frequency to evaluate permittivity at (Hz).
         coords : :class:`.Coords`
@@ -586,18 +586,17 @@ class CustomMedium(AbstractMedium):
 
         Returns
         -------
-        Numpy
+        np.ndarray
             Single component of the complex-valued permittivity tensor at ``frequency`` interpolated
-            at the supplied coordinate.
+            at the supplied coordinates.
         """
 
         if row == col:
             eps_freq = self.eps_dataset_freq(frequency)
             interp_shape = [len(coord_comp) for coord_comp in coords.to_list]
             comp = ["eps_xx", "eps_yy", "eps_zz"][row]
-            eps = np.array(
-                self._interp(eps_freq.field_components[comp], coords, self.interp_method)
-            ).reshape(interp_shape)
+            eps = self._interp(eps_freq.field_components[comp], coords, self.interp_method)
+            eps = np.array(eps).reshape(interp_shape)
             return eps
         return 0
 
@@ -1304,11 +1303,28 @@ class AnisotropicMedium(AbstractMedium):
         return (eps_xx, eps_yy, eps_zz)
 
     def eps_comp(self, row: Axis, col: Axis, frequency: float) -> complex:
-        """Single component the complex-valued permittivity tensor as a function of frequency."""
+        """Single component the complex-valued permittivity tensor as a function of frequency.
 
-        if row == col:
-            return [self.xx, self.yy, self.zz][row].eps_model(frequency)
-        return 0
+        Parameters
+        ----------
+        row : int
+            Component's row in the permittivity tensor (0, 1, or 2 for x, y, or z respectively).
+        col : int
+            Component's column in the permittivity tensor (0, 1, or 2 for x, y, or z respectively).
+        frequency : float
+            Frequency to evaluate permittivity at (Hz).
+
+        Returns
+        -------
+        complex
+           Element of the relative permittivity tensor evaluated at ``frequency``.
+        """
+
+        if row != col:
+            return 0j
+        cmp = "xyz"[row]
+        field_name = cmp + cmp
+        return self.components[field_name].eps_model(frequency)
 
     @add_ax_if_none
     def plot(self, freqs: float, ax: Ax = None) -> Ax:
@@ -1337,7 +1353,11 @@ class AnisotropicMedium(AbstractMedium):
 
 
 class FullyAnisotropicMedium(AbstractMedium):
-    """Fully anisotropic medium.
+    """Fully anisotropic medium including all 9 components of the permittivity and conductivity
+    tensors. Provided permittivity tensor and the symmetric part of the conductivity tensor must
+    have coinciding main directions. A non-symmetric conductivity tensor can be used to model
+    magneto-optic effects. Note that dispersive properties and subpixel averaging are currently not
+    supported for fully anisotropic materials.
 
     Example
     -------
@@ -1348,14 +1368,14 @@ class FullyAnisotropicMedium(AbstractMedium):
     >>>)
     """
 
-    permittivity: RealTensor = pd.Field(
+    permittivity: TensorReal = pd.Field(
         [[1, 0, 0], [0, 1, 0], [0, 0, 1]],
         title="Permittivity",
         description="Relative permittivity tensor.",
         units=PERMITTIVITY,
     )
 
-    conductivity: RealTensor = pd.Field(
+    conductivity: TensorReal = pd.Field(
         [[0, 0, 0], [0, 0, 0], [0, 0, 0]],
         title="Conductivity",
         description="Electric conductivity tensor. Defined such that the imaginary part "
@@ -1369,10 +1389,10 @@ class FullyAnisotropicMedium(AbstractMedium):
         with eigenvalues >= 1.
         """
 
-        if not np.allclose(val, np.transpose(val), atol=1.0e-10):
+        if not np.allclose(val, np.transpose(val), atol=fp_eps):
             raise ValidationError("Provided permittivity tensor is not symmetric.")
 
-        if np.any(np.linalg.eigvals(val) < 1 - 1.0e-10):
+        if np.any(np.linalg.eigvals(val) < 1 - fp_eps):
             raise ValidationError("Main diagonal of provided permittivity tensor is not >= 1.")
 
         return val
@@ -1384,11 +1404,10 @@ class FullyAnisotropicMedium(AbstractMedium):
         """
 
         perm = values.get("permittivity")
-        cond_sym = 0.5 * (val + np.transpose(val))
+        cond_sym = 0.5 * (val + val.T)
+        comm_diff = np.abs(np.matmul(perm, cond_sym) - np.matmul(cond_sym, perm))
 
-        if not np.allclose(
-            np.abs(np.matmul(perm, cond_sym) - np.matmul(cond_sym, perm)), 0, atol=1.0e-10
-        ):
+        if not np.allclose(comm_diff, 0, atol=fp_eps):
             raise ValidationError(
                 "Main directions of conductivity and permittivity tensor do not coincide."
             )
@@ -1404,13 +1423,13 @@ class FullyAnisotropicMedium(AbstractMedium):
 
         Parameters
         ----------
-        xx : Medium
+        xx : :class:`.Medium`
             Medium describing the xx-component of the diagonal permittivity tensor.
-        yy : Medium
+        yy : :class:`.Medium`
             Medium describing the yy-component of the diagonal permittivity tensor.
-        zz : Medium
+        zz : :class:`.Medium`
             Medium describing the zz-component of the diagonal permittivity tensor.
-        rotation : RotationType
+        rotation : :class:`.RotationType`
                 Rotation applied to diagonal permittivity tensor.
 
         Returns
@@ -1419,17 +1438,8 @@ class FullyAnisotropicMedium(AbstractMedium):
             Resulting fully anisotropic medium.
         """
 
-        permittivity_diag = [
-            [xx.permittivity, 0, 0],
-            [0, yy.permittivity, 0],
-            [0, 0, zz.permittivity],
-        ]
-
-        conductivity_diag = [
-            [xx.conductivity, 0, 0],
-            [0, yy.conductivity, 0],
-            [0, 0, zz.conductivity],
-        ]
+        permittivity_diag = np.diag([comp.permittivity for comp in [xx, yy, zz]]).tolist()
+        conductivity_diag = np.diag([comp.conductivity for comp in [xx, yy, zz]]).tolist()
 
         permittivity = rotation.rotate_tensor(permittivity_diag)
         conductivity = rotation.rotate_tensor(conductivity_diag)
@@ -1437,18 +1447,38 @@ class FullyAnisotropicMedium(AbstractMedium):
         return cls(permittivity=permittivity, conductivity=conductivity)
 
     @cached_property
-    def eps_sigma_diag(self) -> Tuple[Tuple[float, float, float], Tuple[float, float, float]]:
-        """Main components of permittivity and conductivity tensors."""
+    def _to_diagonal(self) -> AnisotropicMedium:
+        """Construct a diagonally anisotropic medium from main components.
+
+        Returns
+        -------
+        :class:`AnisotropicMedium`
+            Resulting diagonally anisotropic medium.
+        """
+
+        perm, cond, _ = self.eps_sigma_diag
+
+        return AnisotropicMedium(
+            xx=Medium(permittivity=perm[0], conductivity=cond[0]),
+            yy=Medium(permittivity=perm[1], conductivity=cond[1]),
+            zz=Medium(permittivity=perm[2], conductivity=cond[2]),
+        )
+
+    @cached_property
+    def eps_sigma_diag(
+        self,
+    ) -> Tuple[Tuple[float, float, float], Tuple[float, float, float], TensorReal]:
+        """Main components of permittivity and conductivity tensors and their directions."""
 
         perm_diag, vecs = np.linalg.eig(self.permittivity)
         cond_diag = np.diag(np.matmul(np.transpose(vecs), np.matmul(self.conductivity, vecs)))
 
-        return (perm_diag, cond_diag)
+        return (perm_diag, cond_diag, vecs)
 
     @ensure_freq_in_range
     def eps_model(self, frequency: float) -> complex:
         """Complex-valued permittivity as a function of frequency."""
-        perm_diag, cond_diag = self.eps_sigma_diag
+        perm_diag, cond_diag, _ = self.eps_sigma_diag
 
         eps_diag = AbstractMedium.eps_sigma_to_eps_complex(perm_diag, cond_diag, frequency)
         return np.mean(eps_diag)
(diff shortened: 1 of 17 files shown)


tidy3d/components/medium.py
```diff
@@ -131,6 +131,30 @@ def eps_diagonal(self, frequency: float) -> Tuple[complex, complex, complex]:
...
+        -------
+        complex
+           Element of the relative permittivity tensor evaluated at ``frequency``.
+        """
+
+        # This only needs to be overwritten for anisotropic materials
+        if row == col:
+            eps = self.eps_model(frequency)
+            return eps
+        return 0
```
tylerflex, line 156: let's return a complex number here `0j` just to make sure it's consistent with the return type.

tidy3d/components/medium.py
```diff
@@ -539,10 +563,48 @@ def eps_diagonal_on_grid(
...
+        frequency : float
+            Frequency to evaluate permittivity at (Hz).
+        coords : :class:`.Coords`
+            The grid point coordinates over which interpolation is performed.
+
+        Returns
+        -------
+        Numpy
+            Single component of the complex-valued permittivity tensor at ``frequency`` interpolated
+            at the supplied coordinate.
```
tylerflex, line 591: pluralize "coordinate"

tidy3d/components/medium.py
```diff
@@ -539,10 +563,48 @@ def eps_diagonal_on_grid(
         ]
         return tuple(eps_list)
 
+    def eps_comp_on_grid(
+        self,
+        row: Axis,
+        col: Axis,
+        frequency: float,
+        coords: Coords,
+    ) -> Numpy:
```
tylerflex, line 572: let's use `ArrayLike` in `types.py`. Do we know the number of dimensions and dtype? For example, if a 3D complex, can use `ArrayComplex3D` defined [here](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/types.py#L109).

Change return docstring type too.

tidy3d/components/medium.py
```diff
@@ -539,10 +563,48 @@ def eps_diagonal_on_grid(
...
+        Numpy
+            Single component of the complex-valued permittivity tensor at ``frequency`` interpolated
+            at the supplied coordinate.
+        """
+
+        if row == col:
+            eps_freq = self.eps_dataset_freq(frequency)
+            interp_shape = [len(coord_comp) for coord_comp in coords.to_list]
+            comp = ["eps_xx", "eps_yy", "eps_zz"][row]
+            eps = np.array(
```
tylerflex, line 598: lets break this into two lines for readability.

tidy3d/components/medium.py
```diff
@@ -539,10 +563,48 @@ def eps_diagonal_on_grid(
         ]
         return tuple(eps_list)
 
+    def eps_comp_on_grid(
```
tylerflex, line 566: Just curious, where is this method used? in the backend?

tidy3d/components/medium.py
```diff
@@ -1241,6 +1303,13 @@ def eps_diagonal(self, frequency: float) -> Tuple[complex, complex, complex]:
         eps_zz = self.zz.eps_model(frequency)
         return (eps_xx, eps_yy, eps_zz)
 
+    def eps_comp(self, row: Axis, col: Axis, frequency: float) -> complex:
+        """Single component the complex-valued permittivity tensor as a function of frequency."""
+
+        if row == col:
+            return [self.xx, self.yy, self.zz][row].eps_model(frequency)
```
tylerflex, line 1310: could consider using `self.components` (dict of `"ii"` to `self.ii`) like so
```py
if row != col:
    return 0j
cmp = "xyz"[row]
field_name = cmp + cmp
return self.components[field_name].eps_model(frequency)
```
but what you have is also ok. Just thinking if AnisoptricMedium is changed later it's slightly nicer to use the .components property for generality

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
         return dict(xx=self.xx, yy=self.yy, zz=self.zz)
 
 
+class FullyAnisotropicMedium(AbstractMedium):
+    """Fully anisotropic medium.
```
tylerflex, line 1340: maybe add "including all 9 components of the permittivity and conductivity tensors." Also, mention whether it can model dispersive materials?

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+        units=CONDUCTIVITY,
+    )
+
+    @pd.validator("permittivity", always=True)
+    def permittivity_spd_and_ge_one(cls, val):
+        """Check that provided permittivity tensor is symmetric positive definite
+        with eigenvalues >= 1.
+        """
+
+        if not np.allclose(val, np.transpose(val), atol=1.0e-10):
```
tylerflex, line 1372: might be a bit better to define these atol and main diagonal upper limit values as constants at the top of the file. so they are better documented and easier to find and change later. Similar comment about the other validators below. (Just a suggestion, I'll let you decide if you think it's instead better left the way it is).

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+            raise ValidationError("Provided permittivity tensor is not symmetric.")
+
+        if np.any(np.linalg.eigvals(val) < 1 - 1.0e-10):
```
tylerflex, line 1375: out of curiosity, what's the reasoning behind the 1e-10 addition? is it due to numerical error in the eigvals?

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+        return val
+
+    @pd.validator("conductivity", always=True)
+    def conductivity_ge_zero_and_commutes(cls, val, values):
+        """Check that the symmetric part of conductivity tensor commutes with permittivity tensor
+        (that is, simultaneously diagonalizable) with eigenvalues >= 0.
+        """
+
+        perm = values.get("permittivity")
+        cond_sym = 0.5 * (val + np.transpose(val))
```
tylerflex, line 1387: what's the type of `val`?  `RealTensor` is a nested tuple, but this looks like it's a numpy array? Also, for future reference, you can use `val.T` as shorthand for `np.transpose(val)`.

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+
+        if not np.allclose(
```
tylerflex, line 1389: let's break this into two lines for readability (substitute `diff` for a variable describing whatever this quantity is storing.
```py
diff = np.abs(np.matmul(perm, cond_sym) - np.matmul(cond_sym, perm))
if not np.allclose(diff, 0, atol=1.0e-10):
```

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+        rotation : RotationType
+                Rotation applied to diagonal permittivity tensor.
+
+        Returns
+        -------
+        :class:`FullyAnisotropicMedium`
+            Resulting fully anisotropic medium.
+        """
+
+        permittivity_diag = [
```
tylerflex, line 1422: if we're paying by the line, this could be another option
```py
permittivity_diag = np.diag((xx.permittivity, yy.permittivity, zz.permittivity)).tolist()
```
same for conductivity. Yours is probably clear enough though.

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+
+        return val
+
+    @classmethod
+    def from_diagonal(cls, xx: Medium, yy: Medium, zz: Medium, rotation: RotationType):
+        """Construct a fully anisotropic medium by rotating a diagonally ansisotropic medium.
+
+        Parameters
+        ----------
+        xx : Medium
```
tylerflex, line 1407:
```
:class:`.Medium`
```

tylerflex: same for RotationType if it will render properly in the docs (ie has a reference in api.rst). Otherwise we'll need to probably be more specific about what a RotationType is here.

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+        return np.mean(eps_diag)
+
+    @ensure_freq_in_range
+    def eps_diagonal(self, frequency: float) -> Tuple[complex, complex, complex]:
+        """Main diagonal of the complex-valued permittivity tensor as a function of frequency."""
+
+        perm_diag, cond_diag = self.eps_sigma_diag
+
+        if np.isscalar(frequency):
+            eps_diag = AbstractMedium.eps_sigma_to_eps_complex(perm_diag, cond_diag, frequency)
```
tylerflex, line 1463: maybe more simple to do

```py
if not np.isscalar(frequency):
    perm_diag = perm_diag[:, None]
    cond_diag = cond_diag[:, None]
return AbstractMedium.eps_sigma_to_eps_complex(perm_diag, cond_diag, frequency)
```

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+        else:
+            eps_diag = AbstractMedium.eps_sigma_to_eps_complex(
+                perm_diag[:, None], cond_diag[:, None], frequency
+            )
+
+        return eps_diag
+
+    def eps_comp(self, row: Axis, col: Axis, frequency: float) -> complex:
+        """Single component the complex-valued permittivity tensor as a function of frequency."""
```
tylerflex, line 1472: copy in the docstring w. parameters and returns

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+
+        return AbstractMedium.eps_sigma_to_eps_complex(
+            self.permittivity[row][col], self.conductivity[row][col], frequency
```
tylerflex, line 1475: let's just break this into three lines
```py
eps = self.permittivity[row][col]
sig = self.conductivity[row][col]
return AbstractMedium.eps_sigma_to_eps_complex(eps, sig, frequency)
```

tidy3d/components/medium.py
```diff
@@ -1267,6 +1336,257 @@ def elements(self) -> Dict[str, IsotropicMediumType]:
...
+        material factor into account is multiplied by ``n_cfl``.
+
+        For this medium, it take the minimal of ``sqrt(permittivity)`` for main directions.
+        """
+
+        perm_diag, _ = self.eps_sigma_diag
+        return min(np.sqrt(perm_diag))
+
+    @add_ax_if_none
+    def plot(self, freqs: float, ax: Ax = None) -> Ax:
```
tylerflex, line 1491: this looks almost exactly identical to the `AnistropicMedium.plot()` is there any way to reuse the code there? for example, either:
1. have `AnistropicMedium.plot()` construct a `FullyAnisotropicMedium.from_diagonals()` and call this plotting method.
2. have `FullyAnistropicMedium.plot()` construct an `AnisotropicMedium` and call it's plotting method.

What do you think?

tidy3d/components/simulation.py
```diff
@@ -1467,6 +1478,7 @@ def eps_bounds(self, freq: float = None) -> Tuple[float, float]:
         medium_list = [self.medium] + list(self.mediums)
         medium_list = [medium for medium in medium_list if not isinstance(medium, PECMedium)]
         # regular medium
+        # wouldn't using eps_diagonal be more accurate?
```
tylerflex, line 1481: remove comment if it's addressed. otherwise, guess we should discuss more

tidy3d/components/simulation.py
```diff
@@ -2579,7 +2595,10 @@ def make_eps_data(coords: Coords):
             return xr.DataArray(eps_array, coords=coords, dims=("x", "y", "z"))
 
         # combine all data into dictionary
-        coords = grid[coord_key]
+        if coord_key[0] != "E":
```
tylerflex, line 2598: maybe a bit simpler to do
```py
if coord_key[0] == "E":
    coord_key = coord_key[0:2]
coords = grid[coord_key]
return make_eps_data(coords)
```
although I',m not exactly sure what this bit is doing exactly.

tidy3d/components/types.py
```diff
@@ -182,6 +182,18 @@ def __modify_schema__(cls, field_schema):
 FreqBoundMin = float
 FreqBound = Tuple[FreqBoundMin, FreqBoundMax]
 
+ComplexTensor = Tuple[
```
tylerflex, line 185: So there are now `ArrayLike` types that currently allow users to supply numpy arrays and have them validated (if desired) for `ndim` or `dtype`.

These arrays are stored internally as `np.ndarray`, but written to json using `.tolist()`

 I wonder if it makes sense to extend some of that to add eg `shape` and then define these tensors as a `constrained_array(dtype=float, shape=(3,3))` and just use the same machinery. Then it could simplify some of the internals eg in our validators.

See [here](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/types.py#L41-L99) for details.

Any thoughts on this? If we do this, it should be reasonably simple to extend using the existing ArrayLike as an example, but I can help out if you need. 

Note, if we do this, it might be good to rebase first to ensure that you have the most recent changes to ArrayLike introduced in either 2.1.0 or 2.0.3 (I forget exactly when).

tylerflex (changes requested):
Looks really good, thanks Daniil for this long awaited and critical feature.

A few general things to discuss:
* Add item in changelog
* compatibility with old `AnisotropicMedium`: the naming conflict is a pain. But it does seem that these two have different functionality (eg can fully anisotropic handle dispersive?) so they should be separate. However, I do wish that we could have called this one `AnisotropicMedium` and the diagonal one `DiagonallyAnisotropicMedium` or something. in the future, are we planning to keep both separate or merge them? If merge, when we get closer to that, we can add a deprecation warning to encourage people to switch to the more general syntax given what we decide.
* Can `RotationAroundAxis` be used in some of the geometry handling? We currently have some rotation code there, but I think it would be nice to unify things.

tidy3d/components/simulation.py
```diff
@@ -749,8 +750,10 @@ def _warn_grid_size_too_small(cls, val, values):  # pylint:disable=too-many-loca
         return val
 
     @pydantic.validator("sources", always=True)
-    def _source_homogeneous(cls, val, values):
-        """Error if a plane wave or gaussian beam source is not in a homogeneous region."""
+    def _source_homogeneous_isotropic(cls, val, values):
+        """Error if a plane wave or gaussian beam source is not in a homogeneous and isotropic
```
momchil-flex, line 754: TFSF should also error if the box crosses any anisotropic region?

momchil-flex (approved):

tidy3d/components/medium.py
```diff
@@ -131,6 +131,30 @@ def eps_diagonal(self, frequency: float) -> Tuple[complex, complex, complex]:
         eps = self.eps_model(frequency)
         return (eps, eps, eps)
 
+    def eps_comp(self, row: Axis, col: Axis, frequency: float) -> complex:
+        """Single component of the complex-valued permittivity tensor as a function of frequency.
+
+        Parameters
+        ----------
+        row : Axis
```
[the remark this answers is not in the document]
dbochkov-flexcompute: fixed

tidy3d/components/medium.py
```diff
@@ -131,6 +131,30 @@ def eps_diagonal(self, frequency: float) -> Tuple[complex, complex, complex]:
...
+
+        Returns
+        -------
+        complex
+           Element of the relative permittivity tensor evaluated at ``frequency``.
+        """
+
+        # This only needs to be overwritten for anisotropic materials
+        if row == col:
+            eps = self.eps_model(frequency)
```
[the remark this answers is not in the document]
dbochkov-flexcompute: fixed

tidy3d/components/medium.py
dbochkov-flexcompute: fixed

tidy3d/components/medium.py
dbochkov-flexcompute: fixed

tidy3d/components/medium.py
dbochkov-flexcompute: fixed

tidy3d/components/medium.py
dbochkov-flexcompute: fixed

tidy3d/components/medium.py
dbochkov-flexcompute: so, it's used in `Structure.eps_comp(...)` which is used in `Simulation.epsilon_on_grid(...)` which is used in `Simulation.epsilon(...)` and in mode solver

tidy3d/components/medium.py
dbochkov-flexcompute: stole this

tidy3d/components/medium.py
dbochkov-flexcompute: added more description

tidy3d/components/medium.py
dbochkov-flexcompute: switched to `fp_eps`

tidy3d/components/medium.py
dbochkov-flexcompute: yeah, exactly, sometimes values are just a little below 1

tidy3d/components/medium.py
dbochkov-flexcompute: fixed

tidy3d/components/medium.py
dbochkov-flexcompute: changed to a variation of what you suggested

tidy3d/components/medium.py
dbochkov-flexcompute: stole

tidy3d/components/medium.py
dbochkov-flexcompute: should I do the same for `eps_comp` in old `AnisotropicMedium`?

tidy3d/components/medium.py
dbochkov-flexcompute: stole

tidy3d/components/structure.py
```diff
@@ -109,6 +109,29 @@ def _check_2d_geometry(cls, val, values):
             _ = geom._normal_2dmaterial  # pylint: disable=protected-access
         return val
 
+    def eps_comp(self, row: Axis, col: Axis, frequency: float, coords: Coords) -> complex:
+        """Single component of the complex-valued permittivity tensor as a function of frequency.
+
+        Parameters
+        ----------
+        row : Axis
```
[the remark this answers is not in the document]
dbochkov-flexcompute: fixed

tidy3d/components/simulation.py
dbochkov-flexcompute: true, added

tidy3d/components/medium.py
dbochkov-flexcompute: Yeah, it is a `RealTensor`, maybe that's why I used `np.transpose` to convert it to numpy. But I should probably extend `ArrayLike` and get rid of `RealTensor`/`ComplexTensor`

tidy3d/components/medium.py
dbochkov-flexcompute: fixed

tidy3d/components/medium.py
dbochkov-flexcompute: Yeah, that's correct, the only difference is that in `AnisotropicMedium` we label components as `("xx", "yy", "zz")`, but for `FullyAnisotropicMedium` I used `("1", "2", "3")`, since main components are oriented arbitratry. I might do what you suggested as (2) and add a check inside the plotting function whether to use `("xx", "yy", "zz")` or `("1", "2", "3")` based on off-diagonal components

dbochkov-flexcompute:
@tylerflex :

> Can RotationAroundAxis be used in some of the geometry handling? We currently have some rotation code there, but I think it would be nice to unify things.

You must be talking about this https://github.com/flexcompute/tidy3d/blob/daniil/fully-anisotropic/tidy3d/components/geometry.py#L635? You're right, it's actually a specific case of `RotationAroundAxis` when it is strictly along one of the Cartesian directions. I'll think how to unify them.

> compatibility with old AnisotropicMedium: the naming conflict is a pain. But it does seem that these two have different functionality (eg can fully anisotropic handle dispersive?) so they should be separate. However, I do wish that we could have called this one AnisotropicMedium and the diagonal one DiagonallyAnisotropicMedium or something. in the future, are we planning to keep both separate or merge them? If merge, when we get closer to that, we can add a deprecation warning to encourage people to switch to the more general syntax given what we decide.

One way how `AnisotropicMedium` and `FullyAnisotropicMedium` can be unified is by just adding additional field `rotation` to `AnisotropicMedium`. In this case a user won't be able to provide permittivity and conductivity as tensors, but (1) I already impose quite strict restriction on that (only tensor with coinciding main direction are allowed, so that there is a well defined main complex epsilons), and (2) this is how some other FDTD packages do

momchil-flex:
> One way how `AnisotropicMedium` and `FullyAnisotropicMedium` can be unified is by just adding additional field `rotation` to `AnisotropicMedium`. In this case a user won't be able to provide permittivity and conductivity as tensors, but (1) I already impose quite strict restriction on that (only tensor with coinciding main direction are allowed, so that there is a well defined main complex epsilons), and (2) this is how some other FDTD packages do

A gyrotropic medium with external magnetic field will have a tensorial conductivity in every axis orientation, no?

tidy3d/components/medium.py
tylerflex: Yea I think a unified treatment of multi-dimensional arrays using ArrayLike would benefit us in the long run, so lets' try to stick with that if it's possible.

tidy3d/components/medium.py
tylerflex: Yea, basically if it's going in the docs (public method) we should include the parameters / returns

dbochkov-flexcompute:
> > One way how `AnisotropicMedium` and `FullyAnisotropicMedium` can be unified is by just adding additional field `rotation` to `AnisotropicMedium`. In this case a user won't be able to provide permittivity and conductivity as tensors, but (1) I already impose quite strict restriction on that (only tensor with coinciding main direction are allowed, so that there is a well defined main complex epsilons), and (2) this is how some other FDTD packages do
> 
> A gyrotropic medium with external magnetic field will have a tensorial conductivity in every axis orientation, no?

Yeah, I forgot to mention that we would also need a field to specify gyrotropic part of permittivity, which could be defined as gyration vector `g = (gx, gy, gz)`, which I guess typically is assumed to be proportional to external magnetic field. Then total complex permittivity becomes `eps = eps_re - 1j * sigma / omega - 1j * G`, where antisymmetric matrix `G = [[0, gz, -gy], [-gz, 0, gx], [gy, -gx, 0]]`. In terms of permittivity and conductivity: `permittivity = eps_re, conductivity = sigma + G * omega`.

So, eventually we have something like
```
medium = td.AnistropicMedium(
    xx=td.Medium(permittivity=2),
    yy=td.Medium(permittivity=3, conductivity=1),
    zz=td.Medium(permittivity=4),
    gyration=(1,2,3),
    rotation=td.RotationAroundAxis(axis=(1,1,0), angle=np.pi/3),
)
```
where we would need to decide whether gyration vector is given in coordinates before or after rotation (probably before)

tidy3d/components/medium.py
dbochkov-flexcompute: Generalized and seems to work well

tidy3d/components/medium.py
dbochkov-flexcompute: fixed

tidy3d/components/medium.py
dbochkov-flexcompute: actually, on the contrary, had to call `AnisotropicMedium.plot()` from inside of `FullyAnisotropicMedium.plot()` because `AnisotropicMedium` is more general in the sense that it allows dispersive materials, while `FullyAnisotropicMedium` currently doesn't

tidy3d/components/simulation.py
dbochkov-flexcompute: fixed

tidy3d/components/simulation.py
dbochkov-flexcompute: I changed this part a little bit to try to make it more clear.
Here I'm trying to generalize `epsilon_on_grid(..., coord_key: str, ...)` to be able to return off-diagonal components too. The way I did this is that in addition to `coord_key = "Ex", "Ey", "Ez"` we can also call `coord_key = "Exy", "Exz", "Eyx", ...` which would return a corresponding off-diagonal component but still at an Yee location (e.g., `eps_xy` at `Ex`). So, for those keys I just remove the last letter.

What I'm a bit hesitant about is that: 
- things like `Ezy` are not proper physical concepts
- from efficiency point of view, maybe it would make more sense to add a separate flag `include_off_diag`, and then `coord_key="Ex", include_off_diag=True` would return three arrays `eps_xx, eps_xy, eps_xz`

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -288,24 +288,38 @@ def _get_epsilon(self, freq: float) -> ArrayComplex4D:
         eps_yy = self.simulation.epsilon_on_grid(self._solver_grid, "Ey", freq)
         eps_zz = self.simulation.epsilon_on_grid(self._solver_grid, "Ez", freq)
 
-        return np.stack((eps_xx, eps_yy, eps_zz), axis=0)
+        eps_xy = self.simulation.epsilon_on_grid(self._solver_grid, "Exy", freq)
```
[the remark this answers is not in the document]
dbochkov-flexcompute: took this suggestion

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -288,24 +288,38 @@ def _get_epsilon(self, freq: float) -> ArrayComplex4D:
...
 
         # get rid of normal axis
-        eps_diag = np.take(eps_diag, indices=[0], axis=1 + self.normal_axis)
-        eps_diag = np.squeeze(eps_diag, axis=1 + self.normal_axis)
+        eps_tensor = np.take(eps_tensor, indices=[0], axis=1 + self.normal_axis)
+        eps_tensor = np.squeeze(eps_tensor, axis=1 + self.normal_axis)
 
         # swap axes to plane coordinates (normal_axis goes to z)
-        eps_zz, (eps_xx, eps_yy) = self.plane.pop_axis(eps_diag, axis=self.normal_axis)
+        if self.normal_axis == 0:
```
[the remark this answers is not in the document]
dbochkov-flexcompute: Tried to think about it, but couldn't think of anything straightforward. Since we use everywhere in mode solver (`compute_modes()`, `_solver_eps()`, and `_get_epsilon()`) the most common ordering (xx, xy, xz, yx, yy, etc), probably it's ok to leave it as it is? Another thing is that in future we might want to make mode solver more consistent with fdtd scheme for fully anisotropic materials, which would require changing from eps to eps^{-1} and providing more than just 9 tensorial components. So, maybe we can revisit this then?

I did reimplement this particular piece of code using permutations of a tensorial form to make it more clear and more bug-proof for now

tidy3d/components/types.py
dbochkov-flexcompute: did this, thanks!

dbochkov-flexcompute:
I think I've addressed all the comments so far, in case you want to give these corrections a look

PR #850, opened 2023-04-20
Refactoring functions to communicate through files

momchil-flex:
I managed to get a first iteration working on the backend. We need a few modifications on the frontend, I started working on those here. I think they all revolve around making everything communicate through files instead of keeping things in memory as it was set up. @tylerflex could you help with the refactor?

Some things to think about:
- Writing and reading JaxInfo to file.
- Ideally no simulation-related data should be read from the web response. This should only be used for metadata, status, etc. I have refactored things to write SimulationData to file in the tests, and then read it in, to emulate how it will actually work. For this I had to work with SimulationData and not JaxSimulationData, as writing this to hdf5 and then reading it back in caused jax to error.
- There seems to be no simple way right now to get a SimulationData without the grad monitors from a SimulationData (or JaxSimulationData) that includes the grad monitors. So I had to do two `tidy3d_run_fn` calls in the forward pass, but hopefully we can fix this.
- The full forward simulation data should also ideally be written to file and then read back out in the backward pass to emulate and test what will actually happen on the server. The two passes will be independent processes and need to communicate through file. I think we also cannot use pickle, because the file will sit in the user's S3 bucket and can be tampered with in between the two processes.
- `sim_vjp` should also be downloaded to file. It seems like we need to define more than one `path`-s, or have the `path` be a folder as opposed to a file path as it is currently.
- Not sure how to pass `fwidth_adj`. It *could* be through the response, but it would be better if it's packaged in a file, especially if in the future there could be other things that also will need to be returned by the forward pass. Maybe after the forward pass, an updated JaxInfo can be downloaded? Not sure.

### DIFF
--- a/tests/test_plugins/test_adjoint.py
+++ b/tests/test_plugins/test_adjoint.py
@@ -27,7 +27,7 @@ from tidy3d.plugins.adjoint.components.data.dataset import JaxPermittivityDatase
 from tidy3d.plugins.adjoint.web import run, run_async
 from tidy3d.plugins.adjoint.components.data.data_array import VALUE_FILTER_THRESHOLD
 
-from ..utils import run_emulated, assert_log_level, log_capture, run_async_emulated
+from ..utils import run_emulated, assert_log_level, log_capture, run_async_emulated, SIM_DATA_PATH
 
 
 EPS = 2.0
@@ -186,13 +186,13 @@ def test_adjoint_pipeline(use_emulated_run):
     """Test computing gradient using jax."""
 
     sim = make_sim(permittivity=EPS, size=SIZE, vertices=VERTICES, base_eps_val=BASE_EPS_VAL)
-    sim_data = run(sim, task_name="test")
+    sim_data = run(sim, task_name="test", path=SIM_DATA_PATH)
 
     def f(permittivity, size, vertices, base_eps_val):
         sim = make_sim(
             permittivity=permittivity, size=size, vertices=vertices, base_eps_val=base_eps_val
         )
-        sim_data = run(sim, task_name="test")
+        sim_data = run(sim, task_name="test", path=SIM_DATA_PATH)
         amp = extract_amp(sim_data)
         return objective(amp)
 
@@ -209,7 +209,7 @@ def test_adjoint_setup_fwd(use_emulated_run):
         simulation=sim,
         task_name="test",
         folder_name="default",
-        path="simulation_data.hdf5",
+        path=SIM_DATA_PATH,
         callback_url=None,
         verbose=False,
     )
@@ -233,7 +233,7 @@ def _test_adjoint_setup_adj(use_emulated_run):
         simulation=sim_orig,
         task_name="test",
         folder_name="default",
-        path="simulation_data.hdf5",
+        path=SIM_DATA_PATH,
         callback_url=None,
     )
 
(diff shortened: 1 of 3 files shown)


tylerflex:
Before doing anything, want to just discuss a bit

> Writing and reading JaxInfo to file.

Should be simple, let's just make `JaxInfo` a `Tidy3dBaseModel`.

> Not sure how to pass fwidth_adj. It could be through the response, but it would be better if it's packaged in a file, especially if in the future there could be other things that also will need to be returned by the forward pass. Maybe after the forward pass, an updated JaxInfo can be downloaded? Not sure.

Let's also add an `fwidth_adj: float` field to the `JaxInfo` as I think this is probably necessary to fully recreate the original simulation anyway and might have not been accounted for.

> Ideally no simulation-related data should be read from the web response. This should only be used for metadata, status, etc. I have refactored things to write SimulationData to file in the tests, and then read it in, to emulate how it will actually work. For this I had to work with SimulationData and not JaxSimulationData, as writing this to hdf5 and then reading it back in caused jax to error.

Yea I think for first iteration, let's do everything with JaxInfo + regular component. for simplicity.

> There seems to be no simple way right now to get a SimulationData without the grad monitors from a SimulationData (or JaxSimulationData) that includes the grad monitors. So I had to do two tidy3d_run_fn calls in the forward pass, but hopefully we can fix this.

I think we'd need a `sd, jax_info = JaxSimulationData.to_sim_data()` option, like in `sim, jax_info JaxSimulation.to_simulation()`. We'd just have to define the splitting in there but should be straightforward.

> The full forward simulation data should also ideally be written to file and then read back out in the backward pass to emulate and test what will actually happen on the server. The two passes will be independent processes and need to communicate through file. I think we also cannot use pickle, because the file will sit in the user's S3 bucket and can be tampered with in between the two processes.

so in my initial refactor, I was "writing" these to a `dict` are you just suggesting replacing this with a file for testing the front end?

> sim_vjp should also be downloaded to file. It seems like we need to define more than one path-s, or have the path be a folder as opposed to a file path as it is currently.

`sim_vjp` should be downloadable as is. Because its fields contain regular `JaxArray` not tracers (as they are prepared by us directly. So this should be simple enough to do just with a single file containing a `JaxSimulation`.

Let me know what you think and I guess the proposed refactoring based on what I said would be:
- [ ] `JaxInfo` as a tidy3d base model (with `to_file()` method)
- [ ] `fwidth_adj` in `JaxInfo`, fix the `to_simulation()` and `from_simulation()` conversion to account for this.
- [ ]  add a `JaxSimulation.to_sim_data()` to split a SimulationData into a `JaxInfo` and `SimulationData` containing no grad monitors (or we could just drop the jax info and have this drop any extra monitor data that isn't `.data`).
- [ ] write the fwd data to file instead of a global dictionary in the test implementation I have from before.
- [ ] see if `sim_vjp` can be written to file, eg in the `test_pipeline()`. In think it should be but can verify.

Let me known if I forgot something / misundersood something

momchil-flex:
I think generally this sounds good. I think the main thing this PR does is make the mock run actually write data to file, the way the real run does. I think what I'm driving at here is that the `webapi_run_...` should be mocked in a similar way to ensure that there's no issue with the file IO. Testing aside though, basically we need to implement the actual functions that will talk to the server. So these would do something like:

```
def webapi_run_adjoint_fwd(...):
    # upload task with taskType "adjoint_fwd"
    # start task
    # download results (regular "simulation_data.hdf5")
```

```
def webapi_run_adjoint_bwd(...):
    # upload task with taskType "adjoint_bwd" and parentTaskIds = ["fwd_task_id"]
    # start task
    # download results (sim_vjp file, which may need a default file name set somewhere)
```

The functions above have no return statement, and the `run_fwd` and `run_bwd` functions can load everything from files. Other ways to set it up are probably also possible (e.g. the webapi functions do the loading already and return the loaded objects), but this version makes sense to me. Coming back to testing, the `webapi_run` functions above can then be mocked similarly to how the standard run is mocked.

PR #853, opened 2023-04-23
fix(SCEM-2520) fix error message

An-Li-magicloud:


tylerflex:
Thanks @An-Li-magicloud. So this basically checks the response of each HTTP request and instead of raising a generic error, will raise a `WebError()` with the error message if there is an error in the response?

tylerflex:
Tested on a notebook and works ok. When I try with the wrong API, rather than some generic error, I get 
```
WebError("unauthorized")
```
we could perhaps strive to make the response error messages more verbose in light of these changes, but for now I think this is already a big improvement.

PR #861, opened 2023-04-26
Improvements to the mode solver for PEC regions

shashwat-sh:
- Using a high conductivity model for PEC regions in the mode solver, rather than a large negative permittivity
- Added internal functionality to apply incidence matrices or preconditioning in the mode solver for future testing (inactive by default, not accessible via the API)

### DIFF
--- a/tidy3d/plugins/mode/solver.py
+++ b/tidy3d/plugins/mode/solver.py
@@ -237,6 +237,15 @@ class EigSolver(Tidy3dBaseModel):
             Imaginary part of the effective index, shape (num_modes, ).
         """
 
+        # use a high-conductivity model for locations associated with a PEC
+        def conductivity_model_for_pec(eps, threshold=0.9 * pec_val):
+            """PEC entries associated with 'eps' are converted to a high-conductivity model."""
+            eps = eps.astype(complex)
+            eps[eps <= threshold] = 1 + 1j * np.abs(pec_val)
+            return eps
+
+        eps_tensor = conductivity_model_for_pec(eps_tensor)
+
         # Determine if ``eps`` and ``mu`` are diagonal or tensorial
         off_diagonals = (np.ones((3, 3)) - np.eye(3)).astype(bool)
         eps_offd = np.abs(eps_tensor[off_diagonals])
@@ -291,6 +300,21 @@ class EigSolver(Tidy3dBaseModel):
     ):  # pylint:disable=too-many-arguments
         """EM eigenmode solver assuming ``eps`` and ``mu`` are diagonal everywhere."""
 
+        # code associated with these options is included below in case it's useful in the future
+        enable_incidence_matrices = False
+        enable_preconditioner = False
+        analyze_conditioning = False
+
+        def incidence_matrix_for_pec(eps_vec, threshold=0.9 * np.abs(pec_val)):
+            """Incidence matrix indicating non-PEC entries associated with 'eps_vec'."""
+            nnz = eps_vec[np.abs(eps_vec) < threshold]
+            eps_nz = eps_vec.copy()
+            eps_nz[np.abs(eps_vec) >= threshold] = 0
+            rows = np.arange(0, len(nnz))
+            cols = np.argwhere(eps_nz).flatten()
+            dnz = sp.csr_matrix(([1] * len(nnz), (rows, cols)), shape=(len(rows), len(eps_vec)))
+            return dnz
+
         mode_solver_type = "diagonal"
         N = eps.shape[-1]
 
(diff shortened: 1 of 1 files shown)


tidy3d/plugins/mode/solver.py
```diff
@@ -274,6 +274,17 @@ def solver_em(
         der_mats = [cls.type_conversion(f, mat_dtype) for f in der_mats]
         neff_guess = cls.type_conversion(np.array([neff_guess]), mat_dtype)[0]
 
+        # use a high-conductivity model for locations associated with a PEC
+        # note: it's important to do this after the type checking above, so that the
+        # correct solver (tensorial / diagonal) is called
```
momchil-flex, line 279: I actually don't understand this comment. It seems this should not have anything to do with picking tensorial vs. diagonal. On the other hand I would think this should be done *before* the type setting because it makes the permittivity complex.

tidy3d/plugins/mode/solver.py
```diff
@@ -291,6 +302,21 @@ def solver_diagonal(
...
         """EM eigenmode solver assuming ``eps`` and ``mu`` are diagonal everywhere."""
 
+        # code associated with these options is included below in case it's useful in the future
+        enable_incidence_matrices = False
+        enable_preconditioner = False
+        analyze_conditioning = False
+
+        def incidence_matrix_for_pec(eps_vec, threshold=0.9 * pec_val):
+            """Incidence matrix indicating non-PEC entries associated with 'eps_vec'."""
+            nnz = eps_vec[eps_vec > threshold]
```
momchil-flex, line 312: It should now be `[np.abs(eps_vec) > threshold]` as you may have made the epsilon conductive by now?

momchil-flex (approved):

tidy3d/plugins/mode/solver.py
shashwat-sh: I think I muddled up the logic for `is_tensorial` and `complex_solver`, so yeah, that comment doesn't make sense.

> On the other hand I would think this should be done before the type setting because it makes the permittivity complex.

I thought so too, but when I try to modify the epsilon tensor earlier, so that the complex solver is used, I think it leads to a mismatch somewhere, probably because `gencoeffs` doesn't know that epsilon is complex. In particular, I get the error copied below.

I think the solution is to make the PEC value complex to begin with globally, like you'd suggested yesterday, but I worry a bit about its repercussions elsewhere. Or do you think there's a simpler fix?

```
  File "/home/shashwat/flexcompute/venvs/tidy3d-core/lib/python3.8/site-packages/multiprocess/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/home/shashwat/flexcompute/venvs/tidy3d-core/lib/python3.8/site-packages/multiprocess/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/home/shashwat/flexcompute/venvs/tidy3d-core/lib/python3.8/site-packages/pathos/helpers/mp_helper.py", line 15, in <lambda>
    func = lambda args: f(*args)
  File "/home/shashwat/flexcompute/repositories/tidy3d-core/tidy3d_backend/mode.py", line 156, in _log_solve_single_freq
    return self._solve_single_freq(freq, coords, symmetry)
  File "/home/shashwat/flexcompute/repositories/tidy3d-core/tidy3d_frontend/tidy3d/plugins/mode/mode_solver.py", line 340, in _solve_single_freq
    eps_cross=self._solver_eps(freq),
  File "/home/shashwat/flexcompute/repositories/tidy3d-core/tidy3d_frontend/tidy3d/plugins/mode/mode_solver.py", line 298, in _solver_eps
    eps_diag = self._get_epsilon(freq)
  File "/home/shashwat/flexcompute/repositories/tidy3d-core/tidy3d_backend/mode.py", line 123, in _get_epsilon
    eps_sub = get_eps_yee(
  File "/home/shashwat/flexcompute/repositories/tidy3d-core/tidy3d_backend/utils.py", line 129, in get_eps_yee
    subprocess_cmd(f"{cmd} {gencoeffs_args}")
  File "/home/shashwat/flexcompute/repositories/tidy3d-core/tidy3d_backend/utils.py", line 34, in subprocess_cmd
    subprocess.run(comm_list, stdout=None, check=True, stdin=subprocess.DEVNULL)
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/home/shashwat/flexcompute/repositories/tidy3d-core/solver/gencoeffs', './tmp/freq_1.934145e+14/eps_sub_sim.json', './tmp/freq_1.934145e+14/eps_sub_in.hdf5', './tmp/freq_1.934145e+14/eps_sub_out.hdf5', '0']' died with <Signals.SIGABRT: 6>.
```

tidy3d/plugins/mode/solver.py
shashwat-sh: Good catch, thanks

tidy3d/plugins/mode/solver.py
shashwat-sh: Or maybe it's simpler than I think and I just need to still make this change just local, and maybe revert back to real epsilon.

momchil-flex: Wait this also doesn't make sense to me. This should be way past the point at which gencoeffs is used, the output of gencoeffs is used as input to the mode solver. Note that for some reason I was also getting a gencoeffs error when trying to run from your branch locally, then I rebuilt solver and it worked, then I had to rebuild again when I changed to a different branch... which is strange because I thought there are no modifications to the actual solver.

shashwat-sh: Ok, rebuilding the solver removes the error. But now the conductivity model has no effect...I wonder actually if using an imaginary epsilon _and_ using a complex solver (therefore nudging it in the direction of complex eigvals?) just re-introduces the instability.

shashwat-sh: As for why the solver had to be rebuilt, yeah I'm not sure. There were no solver changes associated with this, but maybe there were other changes somewhere?

shashwat-sh: To clarify, the conductivity model continues to work fine if I introduce the epsilon swapping _after_ the solver data type is chosen.

shashwat-sh: https://github.com/flexcompute/tidy3d/blob/58590ae54a2628e1caaf12da7b0ead256bbe9c3e/tidy3d/plugins/mode/solver.py#L274

When the conductivity model is applied earlier, `der_mats` is set as complex, and this causes the results to still have the same spikes as the original code. If I manually force `der_mats` to be real, then the conductivity model works as usual.

However, when I apply the preconditioner, the conductivity model not only works well, but now doesn't skip that last mode!

PR #864, opened 2023-04-27
custom medium generalized, eps_dataset->dataset.

tylerflex:
* Created `SpatialDataArray` with `xyz` coords only.
* Created `Dataset` subclasses for dispersive custom mediums.
* Abstracted `CustomMedium` to derive from `AbstractCustomMedium` with `dataset` and `interp_method` fields.
* Rename `CustomMedium.eps_dataset` to `.dataset` for generality, added deprecation warning and conversion + test.
* Fixed adjoint plugin and tests to handle new `dataset` syntax.
* Added templates for `CustomDispersiveMedium` and `CustomDrude`.
* Future work can include `CustomAnisotropicMedium` with little effort.

### DIFF
--- a/tidy3d/components/data/dataset.py
+++ b/tidy3d/components/data/dataset.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Union, Dict, Callable
+from typing import Union, Dict, Callable, Tuple
 
 import xarray as xr
 import numpy as np
@@ -11,7 +11,7 @@ import pydantic as pd
 from .data_array import DataArray
 from .data_array import ScalarFieldDataArray, ScalarFieldTimeDataArray, ScalarModeFieldDataArray
 from .data_array import ModeIndexDataArray
-from .data_array import TriangleMeshDataArray
+from .data_array import TriangleMeshDataArray, SpatialDataArray
 
 from ..base import Tidy3dBaseModel
 from ..types import Axis
@@ -388,6 +388,37 @@ class PermittivityDataset(AbstractFieldDataset):
         title="Epsilon zz",
         description="Spatial distribution of the zz-component of the relative permittivity.",
     )
+
+
+class DispersiveDataset(Dataset):
+    """Dataset containing dispersive medium parameters."""
+
+    eps_inf: SpatialDataArray = pd.Field(...)
+
+    coeffs: Tuple[Tuple[SpatialDataArray, ...], ...] = pd.Field(...)
+
+    # TODO: validate that all supplied SpatialDataArrays have the same shape
+
+
+class PoleResidueDataset(Dataset):
+    """Dataset containing dispersive medium parameters."""
+
+    eps_inf: SpatialDataArray = pd.Field(...)
+
+    poles: Tuple[Tuple[SpatialDataArray, SpatialDataArray], ...] = pd.Field(...)
+
+
+class DrudeDataset(Dataset):
+    """Dataset containing dispersive medium parameters."""
+
+    eps_inf: SpatialDataArray = pd.Field(...)
+
+    coeffs: Tuple[Tuple[SpatialDataArray, SpatialDataArray], ...] = pd.Field(...)
+
+
+# TODO: if desired, subclasses of DispersiveDataset specific to Drude, etc.
+# TODO: should this all be moved to medium.py?
+# TODO: an AnisotropicDataset
 
 
 class TriangleMeshDataset(Dataset):
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -14,7 +14,7 @@ from .base import Tidy3dBaseModel, cached_property
 from .grid.grid import Coords, Grid
 from .types import PoleAndResidue, Ax, FreqBound, TYPE_TAG_STR, InterpMethod, Bound, ArrayComplex4D
 from .types import Axis
-from .data.dataset import PermittivityDataset
+from .data.dataset import PermittivityDataset, DispersiveDataset, PoleResidueDataset, DrudeDataset
 from .data.data_array import ScalarFieldDataArray
 from .viz import add_ax_if_none
 from .geometry import Geometry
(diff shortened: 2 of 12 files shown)


tidy3d/components/data/dataset.py
```diff
@@ -390,6 +390,37 @@ def symmetry_eigenvalues(self) -> Dict[str, Callable[[Axis], float]]:
...
 
 
+class DispersiveDataset(Dataset):
+    """Dataset containing dispersive medium parameters."""
+
+    eps_inf: SpatialDataArray = pd.Field(...)
+
+    coeffs: Tuple[Tuple[SpatialDataArray, ...], ...] = pd.Field(...)
+
+    # TODO: validate that all supplied SpatialDataArrays have the same shape
```
weiliangjin2021, line 400: Is the benefit of grouping `eps_inf` and `coeffs` in a data class that it will has its own validators so that it will not pollute the medium class? However, one issue is that custom dispersive medium now has a different interface from the regular medium, since in regular medium `eps_inf` and `coeffs` are separate fields.

The second concern is that `SpatialDataArray` allows for complex values, while `eps_inf` and `coeffs` are real numbers. While we can add validators to constrain them, do you think there is a better approach?

tidy3d/components/medium.py
```diff
@@ -1079,206 +736,631 @@ class Drude(DispersiveMedium):
...
+    # @cached_property
+    # def pole_residue(self):
+    #     """Representation of Medium as a pole-residue model."""
+    #     # TODO: generalize the pole_residue method of the DispersiveModel to
+    #     # # basic idea
+    #     # eps_inf = self.dataset.eps_inf
+    #     # coeffs = self.dataset.coeffs
+    #     # values = Drude.eps_model()
+    #     #     dataset = dataset.sel(points)
+    #     #     value = Drude.eps_model(dataset.values)
```
weiliangjin2021, line 1334: A general python question here: can you call `value = Drude.eps_model(frequency)` here, or is a Drude object constructed?

tylerflex: Yea I think this should be possible. But since it's a method, it expects `self` so I'm pretty sure you'd need to pass an instance that can act similar to a Drude as first argument like `value = Drude.eps_model(self, frequency)`

tidy3d/components/data/dataset.py
tylerflex: I think I see your point. In that case it would make some sense to rename `CustomMedium.eps_dataset` to `CustomMedium.permitttivity`? We could go this route (when we overload a component to make it "custom" we just replace the field with a corresponding `Datasest` if you think it makes more sense? It would probably help with some of the subclassing too?

tylerflex: I think we can just put a validator to enforce real values.

weiliangjin2021: > I think I see your point. In that case it would make some sense to rename `CustomMedium.eps_dataset` to `CustomMedium.permitttivity`? We could go this route (when we overload a component to make it "custom" we just replace the field with a corresponding `Datasest` if you think it makes more sense? It would probably help with some of the subclassing too?

That sounds like the original plan of just having SpatialDataArray in the medium's class field? There will be no need of `DispersiveDataset`?

tylerflex: Yea that's true. So let's see, I guess in its current form, the current CustomMedium is actually a CustomAnisotropicMedium isn't it?

Let's think of a general way to handle this for an AbstractMedium

say our AbstractMedium subclass has fields `f1`, `f2`, `f3` and we want to generalize these to store `DataArray` of our choosing. I think the two options are

1. Subclass from the regular component, overwrite fields diectly, eg
```py
class CustomMediumType(MediumType):
    f1 : DataArrayType
    f2: DataArrayType
    f3 : DataArrayType
   
    # validate all of the fields coordinates and shapes are equal.
    # overwrite methods that depended on f1, f2, f3...

```

2. The dataset approach we have right now
```py
class DatasetType(Dataset):
    f1 : DataArrayType
    f2: DataArrayType
    f3 : DataArrayType

class CustomMediumType(AbstractCustomMedium):
    # note, not inherited from MediumType, f1, f2, f3 are no longer fields, fi are contained in the dataset

    dataset: DatasetType

    # overwrite methods that depended on f1, f2, f3 to call from the dataset
```

is this an accurate summary? and if we indeed end up going with approach 1, we should change `CustomMedium` to be something like


```py
class CustomAnisotropicMedium(AnisotropicMedium):
    eps_xx : ScalarFieldDataArray
    eps_yy : ScalarFieldDataArray
    eps_zz : ScalarFieldDataArray

    # overwrite methods in AnisotropicMedium
```
and the custom dispersive medium would look like
```py
class CustomDrude(Drude):
        eps_inf : SpatialDataArray
        coeffs : Tuple[Tuple[SpatialDataArray, SpatialDataArray]]

        # overwrite drude methods
```
what do you think of this approach?

weiliangjin2021: That's a good summary. For approach 1, we can still subclass from a CustomAbstractMedium class that introduces `interp_method` and interpolation methods. Then e.g. for Drude model, we can 
```
class CustomDrude(CustomAbstractMedium, Drude):
        eps_inf : SpatialDataArray
        coeffs : Tuple[Tuple[SpatialDataArray, SpatialDataArray]]
```
what do you think?

tylerflex: Yea that seems reasonable. Let's do it this way, is there a way to change `CustomMedium` to be consistent with this approach while retaining backwards compatibility? It seems to me we it follows that we should have

```py
class CustomMedium(CustomAbstractMedium, Medium):
    permittivity : ScalarFieldDataArray
    
class CustomAnisotropicMedium(CustomAbstractMedium, AnisotropicMedium):
    eps_xx : ScalarFieldDataArray
    eps_yy : ScalarFieldDataArray
    eps_zz : ScalarFieldDataArray

```

and if a user tries to use `eps_dataset` syntax with `CustomMedium`, we'd need to handle this appropriately (eg return a `CustomAnisotropicMedium`? Any thoughts on making all of these consistent?

weiliangjin2021: I can think of two options. Option 1 is just an addition, while option 2 has complications.

- Leave `CustomMedium` as it is, and create a new class `CustomIsotropicMedium`. Warn in `CustomMedium`that in v3.0, it will be replaced by `CustomIsotropicMedium`.
- As what you have suggested, have both `eps_dataset` and `permittivity` fields in `CustomMedium`. When `eps_dataset` is present, `permittivity` field is ignored, and print a warning. 

Which one do you think makes more sense?

Regarding `eps_xx` in `CustomAnisotropicMedium`, it should be the union of all types of custom isotropic media?

tylerflex: Ultimately, we'd want `CustomX` to just be like medium `X` but allowing the data array fields? so in the future we'd want `CustomMedium` to accept a `permittivity : ScalarFieldDataArray` and then `CustomAnisotropicMedium` to accept `eps_xx`, `eps_yy` and `eps_zz`? Each of which should be a `CustomMedium`, or `CustomDispersiveMedium`, etc.?

With this in mind, maybe we could do as you suggest:
* make `CustomIsotropicMedium`, which accepts `permittivity`. Warn that it will be moved to `CustomMedium` in 3.0
* add `CutstomAnisotropicMedium` with `eps_ii` fields that can accept any other `CustomIsotropicMediumType` (`CustomDispersiveMedium` subclasses and `CustomIsotropicMedium`). 
* add warning in `CustomMedium` that the API will change in 3.0 and if the user truly wants fully anisotropic medium to change to `CustomAnisotropicMedium`.

Hopefully that makes sense?

weiliangjin2021: yeah, all sound good. My branch already has some of it. Maybe I'll just build upon my branch, and borrow some part from your branch?

tylerflex: Yea feel free to do it on yours or take things from this one. I just opened this PR as mostly an example. We could consider doing a refactor of the existing CustomMedium (with deprecation warnings) before implementing CustomDispersion, or do it all in one PR. Up to you, feel free to close this PR if you want to work off your own branch. this PR probably goes against the direction we've been discussing since I created it

weiliangjin2021: Regarding data array as the field in `CustomX`, we should also allow float, so it'll be `Union[data_array, float]`. I would imagine some use case where only some parameters are varying spatially. E.g. plasma frequency changes due to doping, but damping rate or eps_inf is constant. But this is not so major, so it's something to revise later on.

tylerflex: > Regarding data array as the field in CustomX, we should also allow float, so it'll be Union[data_array, float]. I 

along these lines, we could also consider simply overloading the existing classes (Medium, DispersiveMedium) to allow DataArray in the fields?

I'm not sure how this would work out in practice, but in principle, it could mean no need for CustomX specifically.

weiliangjin2021: There is no classes overloading in python? Do you mean just replacing the float to Union[data_array, float] in the existing classes?

tylerflex: yea that's what I meant. I guess it could get complicated pretty fast if we did this?

weiliangjin2021: yeah, the class will be very complicated.

PR #866, opened 2023-05-01
Find stack offset to log correct calling site

lucas-flexcompute:
Fix for logging the correct calling site.
For example, before this patch:
```sh
$ python -c 'import tidy3d; tidy3d.log.warning("Warning test"); raise tidy3d.exceptions.Tidy3dError("Error test")'
[15:53:47] WARNING: Warning test                                                                                                                                                                           log.py:50
           ERROR: Error test                                                                                                                                                                               log.py:50
  File "<string>", line 1, in <module>
tidy3d.exceptions.Tidy3dError: Error test
```

After the patch:
```sh
$ python -c 'import tidy3d; tidy3d.log.warning("Warning test"); raise tidy3d.exceptions.Tidy3dError("Error test")'
[15:54:36] WARNING: Warning test                                                                                                                                                                          <string>:1
           ERROR: Error test                                                                                                                                                                              <string>:1
  File "<string>", line 1, in <module>
tidy3d.exceptions.Tidy3dError: Error test
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Perfect electric conductors (PECs) are now modeled as high-conductivity media in both the frontend and backend mode solvers, and their presence triggers the use of a preconditioner to improve numerical stability and robustness. Consequently, the mode solver provides more accurate eigenvectors and field distributions when PEC structures are present.
 
 ### Fixed
+- Log messages provide the correct caller origin (file name and line number).
 
 ## [2.1.0] - 2023-4-18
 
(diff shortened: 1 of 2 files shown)


tylerflex:
Also, the two code examples from the description look the same, is that a typo?

tylerflex:
Actually, I had one question, it seems the fixed version prints `log.py: 50` , which is when the log handler gets called. Wouldn't we rather want it to print where the log itself gets invoked in the code?

lucas-flexcompute:
> Actually, I had one question, it seems the fixed version prints `log.py: 50` , which is when the log handler gets called. Wouldn't we rather want it to print where the log itself gets invoked in the code?

It's the other way around: before the fix it prints `log:50`. After it, it prints the calling site (`<string>:1` in my case, as I'm just calling from the command line).

tylerflex:
Ah ok, my fault, I thought the first log was from deeper in the stack trace. Makes sense, thanks.

PR #870, opened 2023-05-09
Custom dispersive medium

weiliangjin2021:
- [x] Custom dispersive medium: CustomLorentz, CustomDrdue, etc.
- [x] Consistent interface for custom non-dispersive medium
- [x] Other changes to work with pipeline

### DIFF
--- a/tests/test_components/test_custom.py
+++ b/tests/test_components/test_custom.py
@@ -6,20 +6,28 @@ import pytest
 import numpy as np
 import pydantic
 
+import numpy as np
+import xarray as xr
+import tidy3d as td
+
 from tidy3d.components.simulation import Simulation
 from tidy3d.components.geometry import Box
 from tidy3d.components.structure import Structure
 from tidy3d.components.grid.grid_spec import GridSpec
 from tidy3d import Coords
 from tidy3d.components.source import CustomFieldSource, GaussianPulse, CustomCurrentSource
-from tidy3d.components.data.data_array import ScalarFieldDataArray
+from tidy3d.components.data.data_array import ScalarFieldDataArray, SpatialDataArray
 from tidy3d.components.data.dataset import FieldDataset
 from tidy3d.exceptions import SetupError, DataError, ValidationError
 
 from ..test_data.test_monitor_data import make_field_data
 from ..utils import clear_tmp, assert_log_level, log_capture
 from tidy3d.components.data.dataset import PermittivityDataset
-from tidy3d.components.medium import CustomMedium
+from tidy3d.components.medium import CustomMedium, CustomPoleResidue, CustomSellmeier
+from tidy3d.components.medium import CustomLorentz, CustomDrude, CustomDebye, AbstractCustomMedium
+from tidy3d.components.medium import CustomIsotropicMedium, CustomAnisotropicMedium
+
+np.random.seed(4)
 
 Nx, Ny, Nz = 10, 11, 12
 X = np.linspace(-1, 1, Nx)
@@ -200,8 +208,29 @@ def test_custom_medium_simulation():
 
 def test_medium_raw():
     """Test from a raw permittivity evaluated at center."""
-    eps_raw = make_scalar_data()
+    eps_raw = make_scalar_data().real
+    eps_raw_s = SpatialDataArray(eps_raw.squeeze(dim="f", drop=True))
+
+    # lossless
     med = CustomMedium.from_eps_raw(eps_raw)
+    meds = CustomMedium.from_eps_raw(eps_raw_s)
+    assert med.eps_model(1e14) == meds.eps_model(1e14)
+
+    # lossy
+    data = np.random.random((Nx, Ny, Nz, 1)) + 1 + 1e-2 * 1j
+    eps_raw = ScalarFieldDataArray(data, coords=dict(x=X, y=Y, z=Z, f=freqs))
+    eps_raw_s = SpatialDataArray(eps_raw.squeeze(dim="f", drop=True))
+    med = CustomMedium.from_eps_raw(eps_raw)
+    meds = CustomMedium.from_eps_raw(eps_raw_s, freq=freqs[0])
+    assert med.eps_model(1e14) == meds.eps_model(1e14)
+
+    # inconsistent freq
+    with pytest.raises(SetupError):
+        med = CustomMedium.from_eps_raw(eps_raw, freq=freqs[0] * 1.1)
+
+    # missing freq
+    with pytest.raises(SetupError):
+        med = CustomMedium.from_eps_raw(eps_raw_s)
 
 
 def test_medium_interp():
@@ -273,9 +302,30 @@ def test_medium_eps_diagonal_on_grid():
 def test_medium_nk():
     """Construct custom medium from n (and k) DataArrays."""
     n = make_scalar_data().real
-    k = make_scalar_data().real * 0.01
-    med = CustomMedium.from_nk(n=n, k=k)
+    k = make_scalar_data().real * 0.001
+    ns = SpatialDataArray(n.squeeze(dim="f", drop=True))
+    ks = SpatialDataArray(k.squeeze(dim="f", drop=True))
+
+    # lossless
     med = CustomMedium.from_nk(n=n)
+    meds = CustomMedium.from_nk(n=ns)
+    assert med.eps_model(1e14) == meds.eps_model(1e14)
+    # lossy
+    med = CustomMedium.from_nk(n=n, k=k)
+    meds = CustomMedium.from_nk(n=ns, k=ks, freq=freqs[0])
+    assert med.eps_model(1e14) == meds.eps_model(1e14)
+
+    # inconsistent freq
+    with pytest.raises(SetupError):
+        med = CustomMedium.from_nk(n=n, k=k, freq=freqs[0] * 1.1)
+
+    # missing freq
+    with pytest.raises(SetupError):
+        med = CustomMedium.from_nk(n=ns, k=ks)
+
+    # inconsistent data type
+    with pytest.raises(SetupError):
+        med = CustomMedium.from_nk(n=ns, k=k)
 
 
 def test_medium_eps_model():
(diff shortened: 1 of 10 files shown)


tests/test_components/test_custom.py
```diff
@@ -10,14 +10,21 @@
...
-from tidy3d.components.data.data_array import ScalarFieldDataArray
+from tidy3d.components.data.data_array import ScalarFieldDataArray, SpatialDataArray
 from tidy3d.components.data.dataset import FieldDataset
 from tidy3d.exceptions import SetupError, DataError, ValidationError
 
 from ..test_data.test_monitor_data import make_field_data
 from ..utils import clear_tmp, assert_log_level, log_capture
 from tidy3d.components.data.dataset import PermittivityDataset
-from tidy3d.components.medium import CustomMedium
+from tidy3d.components.medium import (
```
tylerflex, line 20: break into two lines please

tidy3d/components/medium.py
```diff
@@ -12,10 +12,18 @@
 
 from .base import Tidy3dBaseModel, cached_property
 from .grid.grid import Coords, Grid
-from .types import PoleAndResidue, Ax, FreqBound, TYPE_TAG_STR, InterpMethod, Bound, ArrayComplex4D
+from .types import (
```
tylerflex, line 15: break into two lines please

tidy3d/components/medium.py
```diff
@@ -884,6 +1062,126 @@ def to_medium(self) -> Medium:
...
+    def _eps_inf_positive(cls, val):
+        """eps_inf should be positive"""
+        if np.any(val < 0):
+            raise SetupError("'eps_inf' should be positive.")
+        return val
+
+    @pd.validator("poles", always=True)
+    def _poles_correct_shape(cls, val, values):
+        """poles should have the same shape."""
+        if "eps_inf" not in values:
```
tylerflex, line 1115: "eps_inf" could be in values but None, so maybe let's make this more restrictive, like
```
if not values.get("eps_inf"):
```

tidy3d/components/medium.py
```diff
@@ -1040,14 +1391,128 @@ def pole_residue(self):
...
+                raise SetupError("We restrict :math:`\\delta_i<f_i` for now.")
+        return val
+
+    def _pole_residue_dict(self) -> Dict:
+        """Dict representation of Medium as a pole-residue model."""
+
+        poles = []
+        for (de, f, delta) in self.coeffs:
+
+            w = 2 * np.pi * f
```
tylerflex, line 1484: is there any way to reuse this from the regular `Lorentz` class? I remember that the equation was a bit error-prone so it would be good to not have separate copies.

Same comment for the other dispersive subclasses

tidy3d/components/medium.py
```diff
@@ -1105,14 +1569,117 @@ def pole_residue(self):
...
+
+        expected_shape = values["eps_inf"].shape
+        for (f, delta) in val:
+            if f.shape != expected_shape or delta.shape != expected_shape:
+                raise SetupError(
+                    "All terms in 'coeffs' should have the same dimension; "
+                    "The dimension should also be consistent with 'eps_inf'."
+                )
+            if np.any(delta < 0):
+                raise SetupError(":math:`\\delta_i` should be non-negative.")
```
tylerflex, line 1639: not sure this will render in STDOUT

tidy3d/components/medium.py
```diff
@@ -416,23 +531,79 @@ class CustomMedium(AbstractMedium):
     >>> eps = dielectric.eps_model(200e12)
     """
 
+    # permittivity: Optional[SpatialDataArray] = pd.Field(
```
tylerflex, line 534: Do we want to either:
1. Handle permittivity and conductivity  in `CustomMedium`
    a. Uncomment these fields
    b. Validate that if `eps_dataset` is provided, it gets converted to `PermittivityConductivity`. Warn user to use an anisotropic custom medium if they want this functionality. (breaks backwards compatibility?)
    c. Use `permittivity` and `conductivity` internally instead of `eps_dataset`.
    c. Provide eps_dataset as a `@property` generated from `permittivity` and `conductivity` to help with backwards compatibility.

or 
2. Provide a new `CustomIsotropicMedium` with `permittivity` and `conductivity` fields and just warn user of current `CustomMedium` that it will be renamed to `CustomAnisotropicMedium` in the future and to use `CustomIsotropicMedium` for now if that's what they want to use this class for?

tylerflex (commented):
Thanks Weliang, looks good. I think this is the right design decision. A few minor comments and we should discuss how we want to handle deprecation of CustomMedium.

tylerflex:
Also, let's add a changelog item

tidy3d/components/medium.py
weiliangjin2021: For Lorentz dispersion, one complexity is that there will be 2 poles if delta>f, and 1 pole if delta<f. 

- For custom Lorentz dispersion, it means the number of poles can vary spatially. So I temporarily restrict that delta<f at all spatial points. What do you think here?
- In regular Lorentz model, we have the conditional `d>w` etc. to treat the two. When `d` and `w` are spatial array, should i change this condition in regular Lorentz model to be `np.all(d>w)` in order to reuse?

weiliangjin2021: In Drude model, the problem of reusing comes from https://github.com/flexcompute/tidy3d/blob/58590ae54a2628e1caaf12da7b0ead256bbe9c3e/tidy3d/components/medium.py#L1096 so every `a0` needs to be converted to a dataarray.

tidy3d/components/medium.py
weiliangjin2021: Based on our previous discussion, the plan is to use 2. The commented part here will be replaced by that.

tidy3d/components/medium.py
weiliangjin2021: Just use 'delta_i' here?

tidy3d/components/medium.py
weiliangjin2021: This will error if `eps_inf` is a dataarray. How about `if values.get("eps_inf") is None`?

tidy3d/components/medium.py
tylerflex: can the issue in regular Drude be mitigated through
```py
if isinstance(c0, complex):
    a0 = 0j
else:
    a0 = 0 * c0.copy()
```
or is that making things too complex?

tidy3d/components/medium.py
tylerflex: sounds good

tidy3d/components/medium.py
tylerflex: like
```
from .types import ...
from .types import ...
```
it's just a minor style choice that we're going with for consistency.

tests/test_components/test_custom.py
tylerflex: like
```
from .types import ...
from .types import ...
```
it's just a minor style choice that we're going with for consistency.

tidy3d/components/medium.py
tylerflex: For Lorenz, I think it seems a bit too complicated to try to handle both 2 pole and 1 pole cases varying spatially using data arrays. So perhaps for now we just enforce that either all delta < f or all delta > f? Unless there is a way to unify them (eg use two poles for all, but have the 2nd poles be amplitude=0 for delta < f?

tidy3d/components/medium.py
tylerflex: ok and what was the plan to handle if the user has some legacy code using eps_dataset with diagonally anisotropic custom medium? Are we planning to just not support that until CustomAnisotropicMedium is introduced? 

in that case, we'd have to error right away and tell them that custom anisotropic medium is no longer supported?

Another option is to instead have `eps_dataset` be the internally used field, convert supplied `permittivity` and `conductivity` to an `eps_dataset` and also provide `permittivity` and `conductivity` as properties. Maybe this is the only way to avoid breaking backwards compatibility

tidy3d/components/medium.py
weiliangjin2021: > can the issue in regular Drude be mitigated through
> 
> ```python
> if isinstance(c0, complex):
>     a0 = 0j
> else:
>     a0 = 0 * c0.copy()
> ```
> 
> or is that making things too complex?

Sounds good!

weiliangjin2021: For Lorentz model, yeah, I prefer "either all delta < f or all delta > f" for now and extend later. A method is added:
```
if isinstance(coeff_a, SpatialDataArray):
    return np.all(coeff_a.data> coeff_b.data)
return coeff_a > coeff_b
```
`d>w` is then replaced by this method so that I can reuse from the regular `Lorentz` model.

tidy3d/components/medium.py
weiliangjin2021: Now I add a class `CustomIsotropicMedium` which is only internal in v2.0, and will be renamed as `CustomMedium` in v3.0. In v2.0, `CustomMedium` has a cached property `_medium` which will be a `CustomIsotropicMedium` if `permittivity` and `conductivity` are provided, and `CustomAnisotropicMedium` if `eps_dataset` is supplied. What do you think of the current structure?

One immediate improvement is to have `_medium` as CustomIstropicMedium if the components are identical in `eps_dataset`. That's an easy update that I'll do soon.

weiliangjin2021:
`test_adjoint` is expected to fail at the moment. Let me see if I can fix them later.

tidy3d/components/medium.py
tylerflex: I think it's up to you since you understand the internals a lot better. As long as the user will be able to make a `CustomMedium(eps_dataset=...)` and still access the `.eps_dataset` and have the internals work as before, I think it's fine either way. We could also just have `CustomIsotropicMedium` maintain an `eps_dataset` (or property for `CustomMedium`) and use the internals from the current `CustomMedium` class.

tylerflex:
`JaxCustomMedium` will just need similar changes to the deprecation warnings and conversions I think.

weiliangjin2021:
One question on `CustomAnisotropicMedium`: in the current implementation, one of its parent class is `AbstractCustomMedium`, which contains the attribute `interp_method`. `interp_method` is also contained in the field "xx", "yy", and "zz". As a result, `interp_method` appears 4 times in the class, and they are independent and can take different values. To me, it makes sense to have independent `interp_method` in the field "xx", "yy", and "zz", but the one directly inherited from the parent class is useless. @tylerflex Any thoughts on resolving this?

One approach that I can think of is to remove `AbstractCustomMedium` as its parent class, but this certainly has some cons.

tidy3d/components/medium.py
```diff
@@ -967,14 +1459,105 @@ def from_dispersion(cls, n: float, freq: float, dn_dwvl: float = 0, **kwargs):
...
+        Tuple[:class:`.SpatialDataArray`, :class:`.SpatialDataArray`, :class:`.SpatialDataArray`]
+            The permittivity evaluated at ``frequency``.
+        """
+        eps = Sellmeier.eps_model(self, frequency)
+        return (eps, eps, eps)
+
+    @classmethod
+    def from_dispersion(
+        cls, n: SpatialDataArray, freq: float, dn_dwvl: SpatialDataArray, **kwargs
+    ):  # pylint:disable=signature-differs
```
weiliangjin2021, line 1534: The signature of this function in `CustomSellmeier` is from the one in `Sellmeier`: "n" and "dn_dwvl" are now SpatialDataArray. So pylint will complain. Do you think it's fine? @tylerflex

tylerflex: I think it's ok.

tylerflex:
> One question on `CustomAnisotropicMedium`: in the current implementation, one of its parent class is `AbstractCustomMedium`, which contains the attribute `interp_method`. `interp_method` is also contained in the field "xx", "yy", and "zz". As a result, `interp_method` appears 4 times in the class, and they are independent and can take different values. To me, it makes sense to have independent `interp_method` in the field "xx", "yy", and "zz", but the one directly inherited from the parent class is useless. @tylerflex Any thoughts on resolving this?
> 
> One approach that I can think of is to remove `AbstractCustomMedium` as its parent class, but this certainly has some cons.

Maybe one option would be to make `interp_method` optional in the `CustomAnisotropicMedium`, while allowing it in each of the `ii` fields.  But if specified in `CustomAnisotropicMedium`  it overwrites any of the values in the fields using a validator?  (or maybe some variation of this). Would the logic still work if `interp_method` is optional in the `CustomAnisotropicMedium`?

weiliangjin2021:
> > One question on `CustomAnisotropicMedium`: in the current implementation, one of its parent class is `AbstractCustomMedium`, which contains the attribute `interp_method`. `interp_method` is also contained in the field "xx", "yy", and "zz". As a result, `interp_method` appears 4 times in the class, and they are independent and can take different values. To me, it makes sense to have independent `interp_method` in the field "xx", "yy", and "zz", but the one directly inherited from the parent class is useless. @tylerflex Any thoughts on resolving this?
> > One approach that I can think of is to remove `AbstractCustomMedium` as its parent class, but this certainly has some cons.
> 
> Maybe one option would be to make `interp_method` optional in the `CustomAnisotropicMedium`, while allowing it in each of the `ii` fields. But if specified in `CustomAnisotropicMedium` it overwrites any of the values in the fields using a validator? (or maybe some variation of this). Would the logic still work if `interp_method` is optional in the `CustomAnisotropicMedium`?

That's a good idea!

tidy3d/components/simulation.py
```diff
@@ -2622,7 +2627,7 @@ def custom_datasets(self) -> List[Dataset]:
         datasets_source = [
             src.field_dataset for src in self.sources if isinstance(src, CustomFieldSource)
         ]
-        datasets_medium = [mat.eps_dataset for mat in self.mediums if isinstance(mat, CustomMedium)]
+        datasets_medium = [mat for mat in self.mediums if isinstance(mat, AbstractCustomMedium)]
```
weiliangjin2021, line 2630: Now that custom medium no longer has a unified dataset, I'm not sure what to include here. This modification is to make sure the webapi works.

tylerflex: Maybe this is the source of the from_file code failing? Could you check if the datasets_medium are what you'd expect here? Maybe the datasets are not getting written to the file in the first place?

tylerflex: Seems like this custom_datasets property does two things:
1. triggers the upload of hdf5 files in the simulation task (if any datasets present)
2. gives something that can be easily validated to ensure that datasets are not None after loading.

tylerflex: Does something like
```py
datasets_medium = [mat.eps_dataset for mat in self.mediums if isinstance(mat, AbstractCustomMedium)]
```
work?

weiliangjin2021: `AbstractCustomMedium` itself seems to fulfill the two things, so it passes the tests. Regarding `mat.eps_dataset`, the issue is that only `CustomMedium` contains `eps_dataset`.

dbochkov-flexcompute:
Everything looks good to me. A few non-critical comments/questions:
- Do we have any simple rules where to use different types of errors? I noticed some `ValidationError`s in non-validator methods and `SetupError` in some validators, I wonder if that was intended. I think when we check those in tests it doesn't really matter because they all become `pydantic.ValidationError` and to extract exact type we would need to check `errors()["type"]`
- I noticed many randomizations in tests. Would it make sense to add fixing the random seed just to make sure it doesn't breaks accidentally?
- Is it true that in principal it would be possible to let properties and coefficients to be supplied on different meshes since they are going to be interpolated anyway? I guess, it would make things like converting from one representation to another more complicated though
- Maybe related, would it be possible eventually to make a single `AnisotropicMedium` class that allows both custom and non-custom components?

dbochkov-flexcompute:
\+ don't forget to update `CHANGELOG.md`

dbochkov-flexcompute (approved):

weiliangjin2021:
> * Do we have any simple rules where to use different types of errors? I noticed some `ValidationError`s in non-validator methods and `SetupError` in some validators, I wonder if that was intended. I think when we check those in tests it doesn't really matter because they all become `pydantic.ValidationError` and to extract exact type we would need to check `errors()["type"]`

Good point. Not intentional, I think I have messed up the two. Is the rule that `ValidationError` should be used in validator and `SetupError` in other places? Could @tylerflex confirm?

> * I noticed many randomizations in tests. Would it make sense to add fixing the random seed just to make sure it doesn't breaks accidentally?

Will do.

> * Is it true that in principal it would be possible to let properties and coefficients to be supplied on different meshes since they are going to be interpolated anyway? I guess, it would make things like converting from one representation to another more complicated though

Sorry, could you give an example on "let properties and coefficients to be supplied on different meshes"?

> * Maybe related, would it be possible eventually to make a single `AnisotropicMedium` class that allows both custom and non-custom components?

I think it's feasible. Do you think it's needed in thermal or carrier application? If not, it might not be of high priority at the moment. It should be easy to add to `AnisotropicMedium` and add deprecation warning in `CustomAnsiostropicMedium`.

> don't forget to update CHANGELOG.md

Thanks for the reminder, I'll do that in the last commit.

tylerflex:
> Good point. Not intentional, I think I have messed up the two. Is the rule that ValidationError should be used in validator and SetupError in other places? Could @tylerflex confirm?

The general idea I've been going by is that it doesn't really matter, but ValidationError is more for specific problems (something is negative when it should be positive) kind of like ValueError. SetupError is more of a high level problem when the simulation itself is set up wrong (like source outside of Simulation.bounds).

dbochkov-flexcompute:
> Sorry, could you give an example on "let properties and coefficients to be supplied on different meshes"?

Say, permittivity is provided on 100 spatial points while conductivity is provided on 10 spatial points

> I think it's feasible. Do you think it's needed in thermal or carrier application? If not, it might not be of high priority at the moment. It should be easy to add to AnisotropicMedium and add deprecation warning in CustomAnsiostropicMedium.

No, I don't think it's critical for any application. Just thinking that as we keep adding more functionality we generate more and more similarly sounding classes: `AnisotropicMedium`, `CustomAnisotropicMedium`, `FullyAnisotropicMedium`, (next?) `CustomFullyAnisotropicMedium`, ... At some point we might want to think about how to unify them

weiliangjin2021:
> > Sorry, could you give an example on "let properties and coefficients to be supplied on different meshes"?
> 
> Say, permittivity is provided on 100 spatial points while conductivity is provided on 10 spatial points

Yes, in principle it's possible. But it will require more efforts in both frontend and backend. E.g. in the frontend, it's not clear to me how to reuse regular_medium.eps_model() in that case.

> > I think it's feasible. Do you think it's needed in thermal or carrier application? If not, it might not be of high priority at the moment. It should be easy to add to AnisotropicMedium and add deprecation warning in CustomAnsiostropicMedium.
> 
> No, I don't think it's critical for any application. Just thinking that as we keep adding more functionality we generate more and more similarly sounding classes: `AnisotropicMedium`, `CustomAnisotropicMedium`, `FullyAnisotropicMedium`, (next?) `CustomFullyAnisotropicMedium`, ... At some point we might want to think about how to unify them

Right, there are some many possible combinations now.

tylerflex (approved):
Thanks @weiliangjin2021 , If you feel this is ready feel free to merge it.

tidy3d/components/medium.py
```diff
@@ -1101,13 +1719,124 @@ def pole_residue(self):
...
+            if not CustomDispersiveMedium._validate_isreal_dataarray_tuple((de, f, delta)):
+                raise SetupError("All terms in 'coeffs' should be real.")
+        return val
+
+    @pd.validator("coeffs", always=True)
+    def _coeffs_delta_all_smaller_or_larger_than_fi(cls, val):
+        """We restrict either all f>delta or all f<delta for now."""
+        for (_, f, delta) in val:
+            if not (Lorentz._all_larger(f, delta) or Lorentz._all_larger(delta, f)):
+                raise SetupError("We restrict either all 'delta<f' or all 'delta>f'.")
```
momchil-flex, line 1819: Very minor comments about error messages:
 - We don't tend to use "we" in error messages. How about "All coefficients must obey either 'delta' < 'f' everywhere, or 'delta' > 'f' everywhere." Note also the slightly different formatting of the inequality statement. I think we use `'` around variables, not around code-like statements like we would in markdown.
 - Could you change "should" to "must" in all messages?

weiliangjin2021: Thanks, both have been addressed in the latest commit.

tidy3d/components/medium.py
```diff
@@ -548,208 +1007,142 @@ def eps_diagonal_on_grid(
...
         -------
         :class:`.CustomMedium`
             Medium containing the spatially varying permittivity data.
         """
+        # lossless
         if k is None:
-            k = xr.zeros_like(n)
+            if isinstance(n, ScalarFieldDataArray):
+                n = SpatialDataArray(n.squeeze(dim="f", drop=True))
+            return cls(permittivity=n**2, interp_method=interp_method)
```
tylerflex, line 1113: is it possible to reuse the `nk_to_eps_sigma` method instead of directly applying `n**2`?

tidy3d/components/medium.py
```diff
@@ -548,208 +1007,142 @@ def eps_diagonal_on_grid(
...
-        k : :class:`.ScalarFieldDataArray` = None
+        k : Union[:class:`.SpatialDataArray`, :class:`.ScalarFieldDataArray`], optional
             Imaginary part of refrative index.
+        freq : float, optional
+            Frequency at which ``n`` and ``k`` are defined.
         interp_method : :class:`.InterpMethod`, optional
-                Interpolation method to obtain permittivity values that are not supplied
-                at the Yee grids.
+            Interpolation method to obtain permittivity values that are not supplied
+            at the Yee grids.
```
tylerflex, line 1103: I think we should add a `Note` explaining briefly how it works with `freq` and the types of `n` `k`. for example:
```
If both `n` and `k` are supplied without any frequency coordinate, the `freq` kwarg will be used to evaluate the permittivity and conductivity.
```

tidy3d/components/medium.py
```diff
@@ -548,208 +1007,142 @@ def eps_diagonal_on_grid(
...
         """
+        # lossless
         if k is None:
-            k = xr.zeros_like(n)
+            if isinstance(n, ScalarFieldDataArray):
+                n = SpatialDataArray(n.squeeze(dim="f", drop=True))
+            return cls(permittivity=n**2, interp_method=interp_method)
 
+        # lossy case
+        if np.logical_xor(*[isinstance(f, SpatialDataArray) for f in (n, k)]) or np.logical_xor(
```
tylerflex, line 1116: is it possible to replace this logic with 
```py
if not (type(n) == type(k)):
```
or even more generally 
```
if n.coords.keys() != k.coords.keys()
```
The xor logic is a bit hard to follow.

tidy3d/components/medium.py
```diff
@@ -548,208 +1007,142 @@ def eps_diagonal_on_grid(
...
-            scalar_dataset = scalar_dataset.isel(
-                {ax: [0] * len(coord_interp.to_dict[ax]) for ax in isel_ax}
-            )
-            scalar_dataset = scalar_dataset.assign_coords(
-                {ax: coord_interp.to_dict[ax] for ax in isel_ax}
+        # k is a SpatialDataArray
+        if isinstance(k, SpatialDataArray):
+            if freq is None:
+                raise SetupError(
+                    "For a lossy medium, 'freq' at which 'eps' is defined must be supplied",
```
tylerflex, line 1127:
```
"For a lossy medium, must supply 'freq' at which to convert 'n' and 'k' to a complex valued permittivity."
```

tidy3d/components/medium.py
```diff
@@ -548,208 +1007,142 @@ def eps_diagonal_on_grid(
...
+                )
+            eps_inf, sigma = CustomMedium.nk_to_eps_sigma(n, k, freq)
```
tylerflex, line 1129: minor point for clarity, but this probably isn't "eps_inf" right? just eps evaluated at f=freq?

tylerflex (approved):
few comments but in general looks good to me once they are addressed! Pre-approving

tidy3d/components/medium.py
weiliangjin2021: I initially tried `type`, but pylint complains and asks me to use `isinstance`. `keys()` probably is the answer.

tidy3d/components/medium.py
tylerflex: sure sounds good

tidy3d/components/medium.py
weiliangjin2021: How about

```
For lossy medium, if both ``n`` and ``k`` are supplied through :class:`.SpatialDataArray` that doesn't contain frequency information, the ``freq`` kwarg will be used to evaluate the permittivity and conductivity. Alternatively, ``n`` and ``k`` can be supplied through :class:`.ScalarFieldDataArray` with frequency coordinate, and leave ``freq`` kwarg as default ``None``.
```

tidy3d/components/medium.py
tylerflex: probably not but at least it's consistent so if we change that function we don't need to change everywhere else (can pass `k=0 * n`?)

tidy3d/components/medium.py
tylerflex: sounds good

tidy3d/components/medium.py
tylerflex: edited a bit but looks good overall.

```
For lossy medium, if both ``n`` and ``k`` are supplied through :class:`.SpatialDataArray`, which doesn't contain frequency information, the ``freq`` kwarg will be used to evaluate the permittivity and conductivity. Alternatively, ``n`` and ``k`` can be supplied through :class:`.ScalarFieldDataArray`, which contains a frequency coordinate. In this case, leave ``freq`` kwarg as the default of ``None``.
```

PR #871, opened 2023-05-09
retry connection in web functions

tylerflex:
Short term attempt at making the webapi more stable in connection outages.

For all major webapi functions, if connection errors, logs a warning once and retries connection for CONNECTION_RETRY_TIME=30 seconds (with REFRESH_TIME period in between tries). 

Note, not applied to `web.monitor` as this function wraps other webapi functions so it is already covered and adding it at top level would cause the warning to trigger twice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `web.test()` to simply test if the authentication is configured correctly and raise exception otherwise.
 - Add `SimulationTask.get_running_tasks()` to get a list of running tasks from the server.
 - Argument `scale` to `ModeSolver.plot_field` to control plot scaling.
+- Retry for set number of seconds in web functions if internet connection error.
 
 ### Changed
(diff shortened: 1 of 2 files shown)


lucas-flexcompute (approved):
To me, it looks good. My only suggestion would be to use the same pattern I used for the cache changes in the PR draft for the mode solver so that you can use the decorator without the function call if we're using the default values:
```python
@wait_for_connection
def upload(…):
```

Probably unrelated: Is `TOTAL_DOTS` used anywhere? I could not find references to it. If not, should it be removed?

tylerflex:
Thanks @lucas-flexcompute, I implemented that decorator pattern, could you take a quick look? I just want to make sure the connection retry time kwarg is handled correctly when called as a decorator.

Also I removed the dots constant. Turns out that was a relic from when we manually printed some dots while the simulation was running.

Thanks!

lucas-flexcompute (approved):
Yes, the implementation looks correct. That way you can call as conventional decorators or with arguments.

tylerflex:
That's a neat trick, I always wondered if there was a good way to set that up.

PR #876, opened 2023-05-15
Fix source amplitude in spectrum and source-time plotting

weiliangjin2021:
Reported in https://flow360.atlassian.net/browse/SCEM-2316

Ideally, I think we should update `amp_time`, but it will require changes in backend as well. So I simply fix it inside the plot function.

### DIFF
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -141,7 +141,6 @@ class SimulationData(Tidy3dBaseModel):
         source_time = source.source_time
         times = self.simulation.tmesh
         dt = self.simulation.dt
-        user_defined_phase = np.exp(1j * source_time.phase)
 
         # get boundary information to determine whether to use complex fields
         boundaries = self.simulation.boundary_spec.to_list
@@ -154,8 +153,11 @@ class SimulationData(Tidy3dBaseModel):
             """Source amplitude as function of frequency."""
             spectrum = source_time.spectrum(times, freqs, dt, complex_fields)
 
-            # remove user defined phase from normalization so its effect is present in the result
-            return spectrum * np.conj(user_defined_phase)
+            # Remove user defined amplitude and phase from the normalization
+            # such that they would still have an effect on the output fields.
+            # In other words, we are only normalizing out the arbitrary part of the spectrum
+            # that depends on things like freq0, fwidth and offset.
+            return spectrum / source_time.amplitude / np.exp(1j * source_time.phase)
 
         return source_spectrum_fn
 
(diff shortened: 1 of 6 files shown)


tylerflex (approved):
This seems fine to me for now, but ideally we should put the amplitude into the `self.amp_time()` directly? Let's make a note to change that (+backend fix) at some point?

momchil-flex:
No complications, should be an easy backend fix. Well, of course I'll only be sure of it once we run the tests. :)

tylerflex (approved):

tylerflex:
Looks good, I'll let you merge when the backend changes are in or whenever you think it's ready. Thanks!

momchil-flex:
Ok yeah so there was the subtlety that the user-defined amplitude needs to be taken out of the normalization spectrum. Note that this should not change anything, i.e. the normalization is still the same - it's just that previously, `source_time.amplitude` was missing from the source spectrum. Now that it's there, we have to take it out in the normalization, which makes sense, since I want the `amplitude` and `phase` I provide to have an effect on the results, I just don't want the spurious part of the source spectrum to be there.

Have a look @tylerflex @shashwat-sh @weiliangjin2021

shashwat-sh:
Makes sense to me. Do the backend tests pass now?

momchil-flex:
Yes.

tidy3d/components/data/sim_data.py
```diff
@@ -154,8 +154,11 @@ def source_spectrum_fn(freqs):
             """Source amplitude as function of frequency."""
             spectrum = source_time.spectrum(times, freqs, dt, complex_fields)
 
-            # remove user defined phase from normalization so its effect is present in the result
-            return spectrum * np.conj(user_defined_phase)
+            # Remove user defined amplitude and phase from the normalization
+            # such that they would still have an effect on the output fields.
+            # In other words, we are only normalizing out the arbitrary part of the spectrum
+            # that depends on things like freq0, fwidth and offset.
+            return spectrum / source_time.amplitude / user_defined_phase
```
tylerflex, line 161: why is the user defined phase divided out here instead of multiplied after np.conj()?

momchil-flex: It's equivalent, I just thought it makes it clear what we're doing (basically removing a user-defined complex number out of the spectrum).

It seems that `user_defined_phase` is actually not used anywhere else, so I could just make that line 

```
return spectrum / source_time.amplitude / np.exp(1j * source_time.phase)
```

momchil-flex: Note: updated the code as above.

PR #878, opened 2023-05-15
Parallelizing polyslab vjp, and using matplotlib.path.contains_points in PolySlab.inside

momchil-flex:
In the adjoint vjp timing test, this reduces the run time for 1000 vertices from 120s to 14s on my machine.

To-dos:

- [x] See if matplotlib will be what we go with for `PolySlab.inside`. Move import to top level and remove commented out old handling.
- [x] I would suggest still restricting the adjoint polyslabs to 1000 vertices for now - figure out where exactly to add this validation @tylerflex ?
- [x] I had to rearrange functions a bit in order to use multiprocessing, such that the namespace of the `store_vjp` methods couldn't be reused. Any thoughts on handling better the large list of arguments being passed around?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Perfect electric conductors (PECs) are now modeled as high-conductivity media in both the frontend and backend mode solvers, and their presence triggers the use of a preconditioner to improve numerical stability and robustness. Consequently, the mode solver provides more accurate eigenvectors and field distributions when PEC structures are present.
+- `PolySlab.inside` now uses `matplotlib.path.contains_points`.
 
 ### Fixed
 - Log messages provide the correct caller origin (file name and line number).
--- a/tests/test_plugins/test_adjoint.py
+++ b/tests/test_plugins/test_adjoint.py
@@ -17,7 +17,7 @@ from typing import Tuple, Any, List
 
 from tidy3d.exceptions import DataError, Tidy3dKeyError, AdjointError
 from tidy3d.plugins.adjoint.components.base import JaxObject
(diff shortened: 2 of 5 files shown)


tylerflex:
> [ ]  I would suggest still restricting the adjoint polyslabs to 1000 vertices for now - figure out where exactly to add this validation @tylerflex ?

Let's just throw a regular validator in `JaxPolySlab` to keep things simple. If users want to use a PolySlab without adjoint tracking, they can just make a regular `td.Structure`.

> [ ]  I had to rearrange functions a bit in order to use multiprocessing, such that the namespace of the `store_vjp` methods couldn't be reused. Any thoughts on handling better the large list of arguments being passed around?

I think how you've handled it here looks fine to me. I don't see a better way to do it if we can't capture these arguments in using closures. Only thing I could think of would be to make a big `EdgeArgs` kind of datastructure, but that seems tedious.

tylerflex:
Also, obligatory reminder about the changelog

momchil-flex:
> > [ ]  I would suggest still restricting the adjoint polyslabs to 1000 vertices for now - figure out where exactly to add this validation @tylerflex ?
> 
> Let's just throw a regular validator in `JaxPolySlab` to keep things simple. If users want to use a PolySlab without adjoint tracking, they can just make a regular `td.Structure`. Could you add that in this PR if you agree?

Added.

I also put `assume_sorted = True` in all of the adjoint plugin `interp` calls. I'll make a separate PR for interp calls in other tidy3d components.

Regarding matplotlib, I decided to stick with that. If we have a `MATPLOTLIB_INSTALLED` flag in the future, we can just revert to the old `inside` if `False`, so I left the old handling commented out. I put this change in the changelog - I don't think I need to put anything else there since this is all part of the "introducing adjoint polyslab" modification?

tylerflex:
I dont think we need another changelog item, thanks.

PR #886, opened 2023-05-17
Custom Medium Volume Integration

tylerflex:


tidy3d/plugins/adjoint/components/medium.py
```diff
@@ -206,6 +241,24 @@ class JaxCustomMedium(CustomMedium, JaxObject):
...
+        for field_dim in "xyz":
+            field_name = f"eps_{field_dim}{field_dim}"
+            data_array = val.field_components[field_name]
+            coord_lens = [len(data_array.coords[key]) for key in "xyz"]
+            dims_len1 = [val == 1 for val in coord_lens]
+            if sum(dims_len1) == 0:
+                raise SetupError(
+                    "For adjoint plugin, 'JaxCustomMedium' can't contain more than "
+                    "1 pixel along all three dimensions. "
+                    f"Detected 3D pixelated grid in '{field_name}' component of 'eps_dataset.'."
```
momchil-flex, line 257: Message is a bit confusing, namely "more than 1 pixel along all three dimensions". Rather something like `"'JaxCustomMedium' cannot be 3D, needs to be single pixel along at least one dimension."`? Also there are two periods at the end of the message, `'eps_dataset.'.`

tidy3d/plugins/adjoint/components/medium.py
```diff
@@ -75,25 +63,67 @@ def _get_volume_disc(
...
+        vol_coords: Dict[str, np.ndarray],
+        d_vol: float,
+    ) -> xr.DataArray:
+        """Get the E_fwd * E_adj * dV field distribution inside of the discretized volume."""
+
+        e_fwd = grad_data_fwd.field_components[field]
+        e_adj = grad_data_adj.field_components[field]
+        e_dotted = e_fwd * e_adj
+        e_dotted = d_vol * e_dotted.real.interp(**vol_coords).isel(f=0).values
+        return xr.DataArray(e_dotted, coords=vol_coords)
```
momchil-flex, line 81: Why do `.values` on the previous line and then convert to `DataArray` again, doesn't that make it the same thing?

tidy3d/plugins/adjoint/components/medium.py
```diff
@@ -273,40 +326,76 @@ def store_vjp(
...
+                    grad_data_adj=grad_data_adj,
+                    vol_coords=interp_coords,
+                    d_vol=d_vols,
+                ).sum(sum_axes)
+
+                # reshape values to the expected vjp shape to be more safe
+                vjp_shape = tuple(len(coord) for _, coord in coords.items())
+                vjp_values = e_dotted.real.values.reshape(vjp_shape)
+
+                vjp_component += vjp_values
```
momchil-flex, line 394: OK I am definitely confused here. So for every element of the permittivity, you sum up gradient contributions from *all three* field components? That does not make sense to me. The old handling where Ex is used for eps_xx, etc, makes more sense to me for this vjp. Maybe the problem is that the summation needs to happen at the next level, but doesn't? That is to say, when the user provides scalar epsilon, the three eps_xx, eps_yy, eps_zz gradients do need to be summed. But not here, where we still want to return the component-wise gradients, which should only depend on their corresponding field components. But maybe I'm missing something?

tidy3d/plugins/adjoint/components/medium.py
```diff
@@ -273,40 +326,76 @@ def store_vjp(
...
+                    d_sizes[dim_index] = np.array([d_len])
+
+                    # construct the interpolation coordinates along this dimension
+                    coords_interp = np.linspace(r_min + d_len / 2, r_max - d_len / 2, num_cells_dim)
+                    interp_coords[dim_pt] = coords_interp
+                    sum_axes.append(dim_pt)
+
+                # otherwise
+                else:
+                    # just evaluate at a point
```
momchil-flex, line 372: `just evaluate at the original data coords` ?

tidy3d/plugins/adjoint/components/medium.py
tylerflex: 3D is a bit confusing I think because the geometry itself can be 3D, it just can't have more than 1 pixel along each dimension. I guess maybe another way of saying this is: 
"For adjoint plugin, the 'JaxCustomMedium' is restricted to a 1D or 2D pixellated grid. It may not contain multiple pixels along all 3 dimensions. Detected 3D pixelated grid in '{field_name}' component of 'eps_dataset'."

How is that? (also resolves the double period thing)

tidy3d/plugins/adjoint/components/medium.py
tylerflex: yea this is probably more clear.

tidy3d/plugins/adjoint/components/medium.py
tylerflex: yea I think this is just a thing where I hadn't cleaned this up yet. I just replaced
```py
e_dotted = d_vol * e_dotted.real.interp(**vol_coords).isel(f=0).values
return xr.DataArray(e_dotted, coords=vol_coords)
```
with
```py
return d_vol * e_dotted.real.interp(**vol_coords).isel(f=0)
```
and it seems fine

tidy3d/plugins/adjoint/components/medium.py
tylerflex: No I think you might be right, actually. It was probably correct before but for some reason the example still was accurate with this new change. I need to think about it a bit more, but if the eps_ii only has direct influence over the E_i fields, I think the vjps should probably be component by component like they were before.

tylerflex: I reverted this and we can continue testing with the old version.

tidy3d/plugins/adjoint/components/medium.py
momchil-flex: Sounds good.

PR #887, opened 2023-05-17
Bumping maximum allowed projected data size to 50GB

### LINKED ISSUE #873: Removing/increasing monitor data size limit
Current size limit on monitor data prevents users from recording data in large simulations. We can increase/remove that limit and warn users when the monitor size is above a certain threshold. If the user is willing to download large data files, they should have the freedom to do so.

Zongfu also suggests removing the grid number limit of `Simulation`. Any concern about that?

momchil-flex:
Fixes #873 .

@tomflexcompute actually you were right, I thought that the current limit was per monitor but it was actually for the full simulation. I made the modifications as we discussed and made user messages show things in GB which I think makes more sense.


tomflexcompute:
Thanks! Sounds great.

PR #889, opened 2023-05-18
Assume sorted in MonitorData interp calls (data from the solver is always sorted)

momchil-flex:


tylerflex:
On mobile so can't formally approve but have no problem with this change so go ahead and merge if tests pass

tylerflex (approved):

PR #892, opened 2023-05-19
Fixing simulation size check to take symmetries into account

momchil-flex:


tomflexcompute (approved):
Looks good. Works well!

PR #896, opened 2023-05-22
Heat solver interface

dbochkov-flexcompute:
A very raw initial version of heat solver interface, but it would be good to discuss major design decisions. There's a short notebook that outlines implemented features and intended usage https://github.com/flexcompute/tidy3d/blob/daniil/heat-solver/tests/_test_heat.ipynb


tylerflex (commented):
Thanks Daniil, this is a really solid first pass, a few general comments and then specific ones last.

## Higher level
* TLDR: 
	* The way I am understanding this, it has a lot of parallels with the ModeSolver, especially remote mode sovler. I think it makes sense to go with that and just follow similar architecture and design principles, including making it a plugin. 
	* I think it’s worth it to start coming up with a plan / design pattern for multiple solvers inheriting from a single abstract simulation class. Including a way to do multi-physics. (Single simulation with elements of and interdependence between heat, mode, FDTD solvers). My only question is when we should introduce these changes, maybe 3.x? In which case we could follow the plugin architecture for 2.x? 

* Is it necessary to have the heat solver in the `components`? I feel like this should definitely be a plugin, just to keep separation. For example, outside of the new HeatSpec addiction to Medium, the components (Medium, Structure) don’t need any modification or dont need to know about the heat solver at all to operate, right? I feel like if we made a HeatStructure and HeatMedium and put the HeatSpec bits in there, this could be a full plugin and this would be a lot cleaner for our development and organization. For example, developing the adjoint bits as a plugin has saved a lot of headache as I dont need to worry about any of the core package affecting my adjoint features or vice versa.

* I think this is a good design based on how things are currently organized  in Tidy3D. It seems to follow the design principles of the ModeSolver, basically a wrapper on Simulation that takes extra parameters, does its own solve, and then outputs some components that can be used in the regular Tidy3D simulation. Along that line, a plugin makes sense to me. However, I think its clear we should perhaps start thinking about how to rearrange tidy3d more generally to handle different solvers that share the same components more naturally and multi-physics. Basically FDTD should not be the “first class” solver with plugins, but rather all solvers should be at the same level. I don’t know how this re-org would fit into our 2.x roadmap but it might be worth coming up with a more concrete plan for how these things will all work together. I can give it some thought. For now, I think the design here works well for purposes of getting something out there. I suggest that we follow similar conventions to the remote mode solver, I imagine @lucas-flexcompute should be able to help with some of the webapi bits. 


## TemperatureDependence
* Could it be worth utilizing the DataArray for the table version? Or in general, storing all temperature dependence as the same type of DataArray, using some classmethod constructers to generating them given whatever arguments we want to expose to the user (temperature points, custom data, or ref_temp / coeff, etc).
* minor point: names of these are quite long, maybe if we unify them, we could call it simply `TemperatureDataArray`? 

## Boundary Conditions
* Is there a medium - structure boundary condition?
* Using names as reference is one approach, but how messy is it to just pass the objects themselves in these BC objects? Just curious.

## Heat Simulation
* As discussed seems having a Simulation argument is like what we do with the mode solver. I think we will want to phase this pattern out because it creates some unnecessary coupling and also the mode solver will then contain a lot of data that is not useful. We dealt with some similar issues in the ModeSolver, for example, the Simulation and ModeSolver need to be uploaded separately to avoid unnecessary data upload and it adds some minor complications to things.
* So the heat_domain acts a bit like a monitor for where the heat distribution will be measured? Is more than one heat domain supported (ie two volumes where we want to solve the heat and apply to the FDTD simulation?)
* How do you plan to do the solver run? We’re introducing a`from tidy3d.plugins.mode.web import run` and `mode_data = run(mode_solver)` syntax in the remote mode solver, maybe we can follow a similar convention.

dbochkov-flexcompute:
> ## TemperatureDependence
> * Could it be worth utilizing the DataArray for the table version? Or in general, storing all temperature dependence as the same type of DataArray, using some classmethod constructers to generating them given whatever arguments we want to expose to the user (temperature points, custom data, or ref_temp / coeff, etc).

Yeah, using DataArray for table version sounds like a good idea. However, I am not sure if it makes sense to convert any dependence to the table representation, we would have to worry about things like what is sufficiently resolved table whenever we add non-linear dependencies (like quadratic, or with inverse dependencies `~ 1/T`,  `~ 1/(1+T)`). Feels like storing an analytical expression/fit if available is easier

> * minor point: names of these are quite long, maybe if we unify them, we could call it simply `TemperatureDataArray`?

Indeed, the length of new classes are getting out of hands a little bit. Any suggestions are appreciated

> ## Boundary Conditions
> * Is there a medium - structure boundary condition?

Not yet, but we could add it if there is interest from customers

> * Using names as reference is one approach, but how messy is it to just pass the objects themselves in these BC objects? Just curious.

What I mostly worry about is the user changing geometry of a structure and then forgetting to update that in boundary conditions

> ## Heat Simulation
> * As discussed seems having a Simulation argument is like what we do with the mode solver. I think we will want to phase this pattern out because it creates some unnecessary coupling and also the mode solver will then contain a lot of data that is not useful. We dealt with some similar issues in the ModeSolver, for example, the Simulation and ModeSolver need to be uploaded separately to avoid unnecessary data upload and it adds some minor complications to things.

I am going to try to refactor it so it inherits from Simulation, like JaxSimulation, instead of taking it as an argument. That probably makes more sense, especially, if we are going to introduce `HeatMedium` and `HeatStructure` classes

> * So the heat_domain acts a bit like a monitor for where the heat distribution will be measured? Is more than one heat domain supported (ie two volumes where we want to solve the heat and apply to the FDTD simulation?)

that could be generalized

> * How do you plan to do the solver run? We’re introducing a`from tidy3d.plugins.mode.web import run` and `mode_data = run(mode_solver)` syntax in the remote mode solver, maybe we can follow a similar convention.

yeah, no strong preference here, whatever most user-friendly

tylerflex:
## Temperature Dependence

Sounds good, let use DataArray for tabular data, I might also suggest `CustomTemperatureXXX` instead of `Table` just for consistency with tidy3d and how we define these things, unless you think table is better.

> Indeed, the length of new classes are getting out of hands a little bit. Any suggestions are appreciated

maybe `TemperatureDependenceXXX` -> `TemperatureModelXXX`? 
I was going to suggest replacing `Temperature` with `Temp` but that might be misleading (temporary)

```py
# base class td.AbstractTemperatureModel
constant = td.TemperatureModelConstant(value=2)
linear = td.TemperatureModelLinear(ref_temp=300, coeff=0.01)
table = td.CustomTemperatureModel(temp_points=[250,300,350], func_values=[4,2,5])
```
## Boundary Conditions

>> how messy is it to just pass the objects themselves in these BC objects? Just curious.

> What I mostly worry about is the user changing geometry of a structure and then forgetting to update that in boundary conditions

I see, with the structures being immutable, I can see this being a problem.. OK referencing by name makes sense to me.

## Heat Solver

I think the overall plan of 
* Inheriting HeatSimulation from Simulation
* using `tidy3d.plugins.heat.web.run(sim: HeatSimulation) -> HeatSimulationData`
Makes sense for now and follows the general framework of the other plugins (mode, fitter, adjoint) so for now let's do that and then when we move to a new organization we can hopefully just treat all the plugins the same way.

dbochkov-flexcompute:
> maybe TemperatureDependenceXXX -> TemperatureModelXXX?
I was going to suggest replacing Temperature with Temp but that might be misleading (temporary)

Yeah, I think this is better. Another consideration is that we will have similar dependencies for other solvers, so we could generalize it and remove Temperature, and have something like: `FunctionXXX`, `FuncXXX`, or `FunXXX`

dbochkov-flexcompute:
Managed to refactor as plugin and also base `HeatSimulation` off `Simulation`. One issue that I encountered is the following: I assumed I would just redefine standard fields to allow for new heat classes, for example:
```
structures: Tuple[HeatStructureType, ...] = pd.Field(...)
```
where `HeatStructureType = Union[Structure, HeatStructure]`, and this work alright overall. But when I try this for background medium
```
medium: Union[MediumType3D, HeatMediumType] = pd.Field(...)
```
it fails many of the standard `Simulation` validators. This is because some of the validators try to construct an auxilary structure based on background, e.g.:
```       
structure_bg = Structure(
    geometry=Box(
        size=values.get("size"),
        center=values.get("center"),
    ),
    medium=values.get("medium"),
)
```
where `medium` cannot be any new class.

I guess one workaround is instead of redefining standard fields, introduce new ones, for example, `heat_structures` and `heat_medium`, that would be converted to standard `structures` and `medium` when something like `to_simulation()` is called. I just think that if we go this way we should also call `to_simulation()` as part of `HeatSimulation` initialization, so that the following doesn't happen: user creates `HeatSimulation`, solves heat distribution, tries to get a new `Simulation` based on heat data but it fails because `heat_strucutres` and `heat_medium` did not go through standard `Simulation` validators, so needs to start all over again.

tylerflex:
I think we never had to worry about this before for `structures` because there is only one `Structure` type allowed, but the monitors and sources need it to work properly.

dbochkov-flexcompute:
for redefining field `structures` doing 
```
structures: Tuple[HeatStructureType, ...] = pd.Field(...)
```
doesn't seem to have issues. It's field `medium` of `HeatSimulation`/`Simulation` which is hot happy, because if set background medium to a heat medium `medium=HeatMedium(...)`, then whenever any `Simulation`'s validator tries `Structure(medium=medium, ...)` it fails

tylerflex:
Ohh I see, sorry. One way to handle this is to overwrite the offending validators in the HeatSimulation, this is what I ended up doing for some aspects of JaxSimulation. I think another approach would be to include a new `heat_medium` and `heat_structures` as you pointed out (I also had to do this for JaxSimulation). Regarding:

> user creates HeatSimulation, solves heat distribution, tries to get a new Simulation based on heat data but it fails because heat_strucutres and heat_medium did not go through standard Simulation validators, so needs to start all over again.

I'm not sure, maybe a post-init validator that simply initializes `to_simulation()` is the best approach?

PR #898, opened 2023-05-23
Fix syntax bug in mode solver

shashwat-sh:
Fixed a very silly bug in the recent mode solver improvements which was causing the preconditioner to always be applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- Bug in `ModeSolver` which was causing a preconditioner to be applied even when it is not needed.
 
 ## [2.2.1] - 2023-5-23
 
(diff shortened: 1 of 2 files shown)


tylerflex (approved):

tylerflex:
note, I think this would also work (using `any()` instead of `np.any()`)
```py
if any(any_pec(i) for i in [eps_xx, eps_yy, eps_zz]):
```

shashwat-sh:
> note, I think this would also work (using `any()` instead of `np.any()`)
> 
> ```python
> if any(any_pec(i) for i in [eps_xx, eps_yy, eps_zz]):
> ```

Yep that seems to work too. Do you think it's better one way or another?

tylerflex:
either is fine with me, just was curious about it so tried it out. Do you want to merge this for the 2.2.1 patch or wait for weiliang?

shashwat-sh:
I think we should merge it in asap, it's a tiny bug but has potentially large consequences

tylerflex:
ok, could you add a line in the changelog, also let's go with `any()` instead of `np.any()` for just 3 elements I think this makes more sense. I'll merge it after and then release 2.2.1

PR #899, opened 2023-05-23
Warning info attempt

dbochkov-flexcompute:
Wanted to get some opinions on this attempt to trace warning info during object initialization @tylerflex, @momchil-flex 

Basically, the idea is to have a dict `trace` in logger in which:
1. objects (any derivative of `Tidy3dBaseModel`) write hierarchically their info when they get initialized
2. logger write warnings (and anything else if needed)

so that it allows to trace where each of the warnings occurred exactly. It's only initialized and used during `Tidy3dBaseModel.__init__()` and cleared out after it is finished. In some more detail, for each initialized object we store hierarchically three lists/dicts: 
- "children": hashes of initialized children objects 
- "logs": warnings occured during initialization
- "fields": names of parent's fields pointing at the current object (could be multiple if fields point at the same object)

This is what I get in `tests/_test_warning_info.py`:
```
{
    "3935611705871406881": {
        "children": {
            "4411795910099000673": {
                "children": {},
                "logs": [
                    [
                        "WARNING",
                        "A large number (30) of frequency points is used in a broadband source. This can slow down simulation time and is only needed if the mode fields are expected to have a very sharp frequency dependence. 'num_freqs' < 20 is sufficient in most cases."
                    ]
                ],
                "fields": [
                    [
                        "sources",
                        0
                    ]
                ],
                "type": "GaussianBeam"
            },
            "-4149933756106994820": {
                "children": {},
                "logs": [
                    [
                        "WARNING",
                        "The ``normal_dir`` field is relevant only for surface monitors and will be ignored for monitor flux, which is a box."
                    ]
                ],
                "fields": [
                    [
                        "monitors",
                        1
                    ]
                ],
                "type": "FluxMonitor"
            }
        },
        "logs": [
            [
                "WARNING",
                "Structure at structures[0] was detected as being less than half of a central wavelength from a PML on side x-min. To avoid inaccurate results, please increase gap between any structures and PML or fully extend structure through the pml. Skipping check for structure indexes > 0."
            ]
        ],
        "fields": [],
        "type": "Simulation"
    }
}
```
With some processing we can easily reconstruct
```
sim -> sources -> 0 -> (GaussianBeam) "A large number (30) of frequency points is used in a broadband source. This can slow down simulation time and is only needed if the mode fields are expected to have a very sharp frequency dependence. 'num_freqs' < 20 is sufficient in most cases."
```
```
sim -> monitors -> 1 -> (FluxMonitor) "The ``normal_dir`` field is relevant only for surface monitors and will be ignored for monitor flux, which is a box."
```
```
sim -> (Simulation) "Structure at structures[0] was detected as being less than half of a central wavelength from a PML on side x-min. To avoid inaccurate results, please increase gap between any structures and PML or fully extend structure through the pml. Skipping check for structure indexes > 0."
```
which kind of mimics pydantic info.

Not sure exactly if this would impact performance, maybe the best to access this experimentally. If you don't see any major downsides, red flags, we could probably work on this further.

By the way, currently we don't have many warnings in initialization of Simulation components, maybe just one or two more to the two already shown above. Most of warnings are at Simulation level, i.e., based on interplay between different components.

### DIFF
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -75,8 +75,88 @@ class Tidy3dBaseModel(pydantic.BaseModel):
 
     def __init__(self, **kwargs):
         """Init method, includes post-init validators."""
+
+        # initialize warning tracing and appoint ourselves as root who strated it
+        if log.trace:
+            root = False
+        else:
+            root = True
+            log.trace = {}
+
+        # create a dict that will be filled by child objects
+        # "children": here they will put their unique id's (hashes) as they get initialized
+        # "logs": where tidy3d.log will record info about warnings that pop up during our initialization
+        # "fields": this field will be populated by parent and it will contain all fields of parent
+        #           that point to us
+        dict_tmp = {"children": {}, "logs": [], "fields": []}
+
+        # temporarily assign it as log trace dict so tht children know where to write
+        log_prev = log.trace
+        log.trace = dict_tmp
+
+        # do initialization including chidren objects
         super().__init__(**kwargs)
         self._post_init_validators()
+
+        # at this point dict_tmp["children"] and/or dict_tmp["logs"] might contain stuff
+
+        # generate a unique id for ourselves
+        self_id = self.__hash__()
+
+        if self_id not in log_prev:  # we could be identical to object already created
+
+            # populate "fields" for every child
+            # basically we try to match hashes of objects stored in our fields to hashes recorded
+            # in "children" dict
+            child_dict = dict_tmp["children"]
+
+            for key in self.__fields__:
+                field = self.__getattribute__(key)
+                if isinstance(field, Tidy3dBaseModel):
+                    child_hash = field.__hash__()
+                    if child_hash in child_dict:
+                        child_dict[child_hash]["fields"].append(key)
+
+                # our fields could be arrays
+                # is there anything else I am missing?
+                if isinstance(field, (list, tuple, np.ndarray)):
+                    for ind, sub_field in enumerate(field):
+                        if isinstance(sub_field, Tidy3dBaseModel):
+                            child_hash = sub_field.__hash__()
+                            if child_hash in child_dict:
+                                child_dict[child_hash]["fields"].append((key, ind))
+
+            # there might be children with "fields" = [], that is, created objects that are not
+            # part of our class (temporary things?); we will filter those out here
+            new_dict = {"children": {}, "logs": [], "fields": []}
+            new_dict["type"] = self.__class__.__name__
+            new_dict["logs"] = dict_tmp["logs"]
+
+            for child in child_dict:
+                if child_dict[child]["fields"] != []:
+                    new_dict["children"][child] = child_dict[child]
+
+            # finally, we put ourselves into parent's "children"
+            # but only if we have "logs" to show and/or "children" (one of which eventually has "logs")
+            # we can remove "if" to display the entire object hierarchy
+            if new_dict["logs"] != [] or new_dict["children"] != {}:
+                if root:
+                    log_prev[self_id] = new_dict
+                else:
+                    log_prev["children"][self_id] = new_dict
+
+        # point logger trace to what it was pointing before
+        log.trace = log_prev
+
+        # if this object initialized tracing, let's process and clean up
+        if root:
+            # fow now just dump everything to make sure it works
+            if log.trace != {}:
+                print("Log dump: ")
+                print(json.dumps(log.trace, indent = 4))
+
+            # clean up
+            log.trace = None
 
     def _post_init_validators(self) -> None:
         """Call validators taking ``self`` that get run after init, implement in subclasses."""
--- a/tests/_test_warning_info.py
+++ b/tests/_test_warning_info.py
@@ -1,0 +1,96 @@
+# standard python imports
+import numpy as np
+import matplotlib.pyplot as plt
+
+# tidy3d imports
+import tidy3d as td
+import tidy3d.web as web
+
+# Define material properties
+medium = td.Medium(permittivity=3)
+
+wavelength = 1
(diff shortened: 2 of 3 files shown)


tidy3d/components/base.py
```diff
@@ -75,9 +75,89 @@ def __hash__(self) -> int:
 
     def __init__(self, **kwargs):
         """Init method, includes post-init validators."""
+
+        # initialize warning tracing and appoint ourselves as root who strated it
```
tylerflex, line 79: in the future, would be best to separate this into a couple methods, so the init method would look something like
```py
    def __init__(self, **kwargs):
        """Init method, includes post-init validators."""
        self._initialize_log_cache()
        super().__init__(**kwargs)
        self._post_init_validators()
       self._consolidate_log_cache()
```

dbochkov-flexcompute:
@lucas-flexcompute

tidy3d/components/base.py
dbochkov-flexcompute: yeah, I just left it all exposed for now, if this is something we want to move forward with, then I'll definitely wrap it all into separate functions

lucas-flexcompute:
@dbochkov-flexcompute, @tylerflex I've tried to merge the capturing functionality with my caching PR over in https://github.com/flexcompute/tidy3d/pull/895. I have not tried to match ids (or hashes) to actual objects, I only logged them in a tree-like structure. I've moved the capture logic to the Logger class, because I believe that makes more sense (except for figuring out the children indexes and the parents fields). In any case, doing this map with hashes will be expensive. I'm not sure if the GUI has access to python ids, but they could use those to finish mapping when/if they need it, instead of doing in the python front-end. I've also used a try/finally block because we want to make sure we close capture contexts.

The previous caching/suppression was changed to work with the capture system. Suppressed messages on the global cache have highest priority, so if they are suppressed they wont be captured. Messages that are not displayed and added to the log consolidation are still captured (but the final consolidation summary is not). I believe this should be the desired behavior, right?

So we use global suppression for deprecation messages or things like letting the user know there is a remote mode solver the first time they use the local one. The consolidation cache is used within a context manager to control of emission of too many messages in loops: only the first is emitted and the remaining are consolidated, but all are captured if the capture functionality is enabled.

tylerflex:
Thanks Lucas, I think this is good to define the desired logging behavior (with regard to caching, suppressing, and consolidating). And then work together to combine the functionality. I think I'd defer to both of you regarding the details regarding the specific information contained in the log and how this is architected. 

Iin terms of general working behavior, my thinking is that maybe suppressed warnings should not be displayed in the python front end, but should still be captured in the log capture, that way the GUI or python front end can handle them however they like (eg, by placing an orange "warning" indicator on each structure that warns). 

Essentially this would mean something like:
1. initialize object
2. capture all warnings and log messages into a single datastructure
3. display the appropriate amount of log messages by processing this datastructure. 

Perhaps the context manager / caching could be used to flag log messages in step 2 to be handled in step 3? I'm not sure if you think this approach is feasible or potentially too difficult. 

My thinking is that it might be good to have all of the warning info, however, my worry is that for hundreds of thousands of warnings or more, the log itself could be far too large to handle efficiently and it might end up slowing down operation?

Just a few thoughts

dbochkov-flexcompute:
> In any case, doing this map with hashes will be expensive. I'm not sure if the GUI has access to python ids, but they could use those to finish mapping when/if they need it, instead of doing in the python front-end.

> @dbochkov-flexcompute What do you think about identifying the exact field and index of each children in the log tree? Is this something we could forward to GUI, since we have no use for that information on the python interface and (as it is) it seems to be an O(n²) operation?

Regarding whether it's only for GUI or not, I remember we briefly touched upon possibly providing hierarchy paths to warnings on python frontend as well. But I am leaning to agree that it wouldn't actually be that useful on python frontend, because interaction with object initialization is quite different. On python frontend we typically initialize objects one by one, so we get warnings about simulation components instantly, and when we put them together into a Simulation object, they are already initialized, so not much of hierarchy to capture. In GUI we parse the entire simulation for errors and warnings at once from a file https://github.com/flexcompute/tidy3d-client-service/blob/master/framework-1.2/service.py#L47-L62, and there it makes more sense to capture the hierarchy of objects and warnings. Having said that, maybe the hierarchy capture should not be a default option and only triggered by hands.

Regarding how to map exactly, I think matching hashes after logging is done is definitely an alternative. But I have two concerns about it:
1. Performance-wise: it feels like matching smaller sets as we go (currently) should be more optimized and less time consuming compared to if we try to match the entire list of warnings to the entire object hierarchy in the end. Maybe something like `O(k*n^2)` vs `O((k*n)^2)`. However, maybe the overhead of executing many smaller matchings compared to one big matching operations (something along the lines `set(warnings).intersection(objects)`) outweights the optimization gains. 
2. It seems like to properly reconstruct paths to warnings to display verbally (e.g.: `"simulation > structures > 2 > geometry > "something wrong with polyslab..."`), GUI part would still need to construct a hierarchy of hashes and then parse it to make matchings. Is there a better way to do construct such an hierarchy compared to doing it in `__init__` as in this PR? If not, maybe it would be better that this is provided by tidy3d team anyway since GUI team is not as familiar with Tidy3D internals as we are. For the second part, matching hashes to fields, can we pull up an object using its hash to find out what are its fields?

Given all that I am kinda leaning towards that it would be the easiest and not terribly performance degrading option just to have hierarchy capturing with field matching as a manually triggered functionality invoked only when parsing on GUI side

dbochkov-flexcompute:
So, in some sense we could make it completely orthogonal: on python frontend turn on hashing but turn off tree capturing, when parsing on GUI side do the opposite (so we can mark all warned objects)

dbochkov-flexcompute:
Closing this as it was implemented in https://github.com/flexcompute/tidy3d/pull/895

PR #901, opened 2023-05-24
fixes to `JaxDataArray` IO

tylerflex:
Adds special handling to `JaxDataArray` writing and reading from file.

Before, it was written inside of the json_string and the resulting hdf5 files were far larger than necessary.

Now, it is written and read similar to how regular `DataArray` objects.

There was significant difficulty because the `JaxDataArray` is a regular `Tidy3dBaseModel` and therefore had many snaggles compared to `xr.DataArray`, which could be handled separately. The solution finally was to overwrite the `JaxObject.json()` method to recursively remove any objects with type of "JaxDataArray".


shashwat-sh (approved):
Looks good! All backend tests passed locally btw

PR #903, opened 2023-05-25
Out-of-bounds TFSF surfaces

shashwat-sh:
Ignore TFSF surfaces outside the simulation domain in the intersections validator.


tylerflex:
looks good, can you just add a changelog item? thanks

PR #905, opened 2023-05-25
Adjoint FieldMonitor support

tylerflex:


momchil-flex (approved):
Generally looks good, just some thoughts:
- Should we validate that `colocate=False` in the monitors, or at least print a warning saying that gradients may not be as accurate if `colocate=True`? In the future if we have a user-side `colocate` working with adjoint, we can also add to the warning to use that instead of setting `colocate=True` in the monitor.
- Kinda related, what happens currently if a user tries to use things like `flux`, `dot`, `colocate`, etc? Do they get an abstruse jax error? Should we try to at least provide a more meaningful one if possible?

momchil-flex:
> > Should we validate that colocate=False in the monitors, or at least print a warning saying that gradients may not be as accurate if colocate=True? In the future if we have a user-side colocate working with adjoint, we can also add to the warning to use that instead of setting colocate=True in the monitor.
> 
> Hm, actually wouldn't colocate just work as is? The adjoint source creation doesn't care about whether colocate was on or off so I don't think we need to worry? Unless you mean specifically about the lack of reverse interpolation? We could suggest users set colocate=True for best accuracy in that case?

The colocation step that would happen on the server is not vjp-ed. The adjoint source is correct only under the assumption that the data is sampled on the Yee grid. So yeah I'm talking about the lack of reverse interpolation, but not just in 0D dimensions. The correct thing to do is not exactly reverse interpolation but rather vjp through the colocation, which is related but not exactly the same I believe.

tylerflex:
Ok I added fixes to both of these in the last two [commits](https://github.com/flexcompute/tidy3d/pull/905/commits).

1. warn if any `output_monitors` are `FieldMonitor` with `colocate=True`
2. raise `NotImplementedError` with a helpful message if `JaxFieldData.flux()` or any other methods are called (after verifying they do error with an abstruse message). 

Feel free to merge this if you approve

PR #906, opened 2023-05-25
Normalize sim data by expected dipole radiation power

dbochkov-flexcompute:
Let me know if we want to move forward with this. Some not ideal things with this is that 
- it doesn't exactly bring power very close to 1 (see https://github.com/flexcompute/tidy3d-core/issues/259#issuecomment-1556222020)
- it seems to break down for anisotropic materials (disable dipole in anisotropic or not normalize + issue a warning?)

### DIFF
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -10,10 +10,11 @@ from .monitor_data import MonitorDataTypes, MonitorDataType, AbstractFieldData, 
 from ..base import Tidy3dBaseModel
 from ..simulation import Simulation
 from ..boundary import BlochBoundary
-from ..source import TFSF
+from ..source import TFSF, PointDipole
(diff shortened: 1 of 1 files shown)


shashwat-sh:
If I understood correctly, if the dipole lies near a material interface such that some of the "equivalent" weighted dipoles are inside the material and some are outside, the normalization is still with respect to where the "real" dipole is intended to be placed? If so, it makes sense to me conceptually, but I also wonder if normalizing each equivalent dipole individually would make the flux closer to 1? Not sure which approach is better - the way it is now does seem pretty clean.

shashwat-sh:
Actually, I think my comment above is unrelated to whether there's a material interface or not

shashwat-sh:
Hmm never mind, there's actually no dependence on the grid in the normalization, so what I said shouldn't matter

shashwat-sh (approved):
Looks good to me

PR #907, opened 2023-05-26
Fixed minor plotting issue with 2d materials

caseyflex:


tylerflex (commented):
looks good to me, the only thing I am not sure about is whether we are skipping plots of `MultiPolygon` here. After thinking about it some more, I might suggest something that is potentially more backwards compatible like
```py
         if isinstance(_shape, LineString):
             xs, ys = zip(*_shape.coords)
             ax.plot(xs, ys, color=plot_params.edgecolor)
         else:
             patch = polygon_patch(_shape, **plot_params.to_kwargs())
             ax.add_artist(patch)
         return ax
```
what do you think? basically this will be guaranteed to work the same way as before except it will also handle line strings

caseyflex:
> looks good to me, the only thing I am not sure about is whether we are skipping plots of `MultiPolygon` here. After thinking about it some more, I might suggest something that is potentially more backwards compatible like
> 
> ```python
>          if isinstance(_shape, LineString):
>              xs, ys = zip(*_shape.coords)
>              ax.plot(xs, ys, color=plot_params.edgecolor)
>          else:
>              patch = polygon_patch(_shape, **plot_params.to_kwargs())
>              ax.add_artist(patch)
>          return ax
> ```
> 
> what do you think? basically this will be guaranteed to work the same way as before except it will also handle line strings

good point. I'll change it shortly

tests/test_components/test_simulation.py
```diff
@@ -1538,6 +1538,8 @@ def test_sim_volumetric_structures():
         LARGE_NUMBER * thickness / grid_dl,
         rtol=RTOL,
     )
+    # check that plotting 2d material with alpha<1 doesn't raise an error
+    _ = sim.plot_eps(ax=AX, x=0, alpha=0.2)
```
shashwat-sh, line 1542: Should we also add a `sim_data.plot_field()` test, since that's how the problem originally came to light?

shashwat-sh (approved):

tests/test_components/test_simulation.py
caseyflex: good point, added the test

tylerflex (approved):

PR #909, opened 2023-05-26
Unify dispersion fitters with a web.run interface

lucas-flexcompute:
The API for dispersion fitting is a single call, so there's not much to monitor. I've basically moved things around to mimic the `web.run` interface from other plugins.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `Medium2D` to full simulation in tests.
 - Removed `StableFitter`
 - Stable dispersion fitter is now executed through `plugins.dispersion.web.run`
+- Deprecated `StableFitter` — stable dispersion fitter is now executed through `plugins.dispersion.web.run`
 
 ### Fixed
 - Plotting 2D materials in `SimulationData.plot_field` and other circumstances.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Add `Medium2D` to full simulation in tests.
+- Removed `StableFitter`
+- Stable dispersion fitter is now executed through `plugins.dispersion.web.run`
 
 ### Fixed
 - Plotting 2D materials in `SimulationData.plot_field` and other circumstances.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Add `Medium2D` to full simulation in tests.
 - Removed `StableFitter`
-- Stable dispersion fitter is now executed through `plugins.dispersion.web.run`
-- Deprecated `StableFitter` — stable dispersion fitter is now executed through `plugins.dispersion.web.run`
+- `DispersionFitter` and `StableDispersionFitter` unified in a single `DispersionFitter` interface.
+- `StableDispersionFitter` deprecated, with stable fitter now being run instead through `plugins.dispersion.web.run(DispersionFitter)`.
 
 ### Fixed
 - Plotting 2D materials in `SimulationData.plot_field` and other circumstances.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Add `Medium2D` to full simulation in tests.
-- Removed `StableFitter`
 - `DispersionFitter` and `StableDispersionFitter` unified in a single `DispersionFitter` interface.
 - `StableDispersionFitter` deprecated, with stable fitter now being run instead through `plugins.dispersion.web.run(DispersionFitter)`.
 
--- a/tidy3d/plugins/dispersion/__init__.py
+++ b/tidy3d/plugins/dispersion/__init__.py
@@ -1,4 +1,4 @@
 """ Imports from dispersion fitter plugin. """
 
-from .fit import DispersionFitter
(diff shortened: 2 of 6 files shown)


tidy3d/plugins/dispersion/__init__.py
```diff
@@ -1,4 +1,4 @@
 """ Imports from dispersion fitter plugin. """
 
 from .fit import DispersionFitter
-from .fit_web import StableDispersionFitter, AdvancedFitterParam
+from .web import AdvancedFitterParam, StableDispersionFitter
```
tylerflex, line 4: does this break backwards compatibility for the Fitter notebook? if so, maybe we could leave `fit_web.py` with just an import of these two classes.

tylerflex (changes requested):
This looks great, appreciate it! especially all of the minor improvements everywhere. Remind me, what was the plan for adding mock tests for the http response? I remember there was something related to this in the Mode solver too. It would be nice to have as they can catch some issues, that's the only thing really. Thanks!

tidy3d/plugins/dispersion/__init__.py
lucas-flexcompute: Good catch. To be safe, we can import everything from "web.py" and issue a deprecation warning. What do you think?
The notebook itself works fine (with a deprecation warning) because it doesn't import "fit_web.py" directly, but users might.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Add `Medium2D` to full simulation in tests.
+- Removed `StableFitter`
+- Stable dispersion fitter is now executed through `plugins.dispersion.web.run`
```
tylerflex, line 13: maybe we should clarify these bullets a bit, suggestion:
```
- `Fitter` and `StableFitter` unified in a single `Fitter` interface.
- `StableFitter` deprecated, with stable fitter now being run instead through `plugins.dispersion.web.run(Fitter)`.
```
think it makes it a bit more clear how someone should use the stable fitter with the new changes.

tylerflex: sorry, I guess one other thing: I noticed the "- Removed `StableFitter`" is still in the changelog, probably this is a little misleading since it's still there, but we already mention that it's deprecated so maybe we can just remove this line?

tidy3d/plugins/dispersion/web.py
```diff
@@ -0,0 +1,362 @@
...
+        Best results of multiple fits: (dispersive medium, RMS error).
+    """
+    task = FitterData.create(fitter, num_poles, num_tries, tolerance_rms, advanced_param)
+    return task.run()
+
+
+class StableDispersionFitter(DispersionFitter):
+    """Deprecated."""
+
+    def __init__(self, *args, **kwargs):
```
tylerflex, line 347: we can also do this through a simple root_validator

```py
@pd.root_validator()
def _depredate_stable_fitter(cls, values):
        log.warning(
             "'StableDispersionFitter' has been deprecated. Use 'DispersionFitter' with "
             "'tidy3d.plugins.dispersion.web.run' to access the stable fitter from the web server."
     )
    return values

```

I think I'd prefer this to avoid messing with the init method in pydantic

lucas-flexcompute: Agreed, I don't like playing with `__init__` in those cases either.

tidy3d/plugins/dispersion/fit.py
```diff
@@ -25,20 +25,20 @@ class DispersionFitter(Tidy3dBaseModel):
...
 
-    n_data: Union[Tuple[float, ...], ArrayFloat1D] = Field(
+    n_data: ArrayFloat1D = Field(
         ...,
         title="Index of refraction data",
         description="Real part of the complex index of refraction.",
     )
 
-    k_data: Union[Tuple[float, ...], ArrayFloat1D] = Field(
+    k_data: ArrayFloat1D = Field(
```
weiliangjin2021, line 41: In principle, it should be `Optional[ArrayFloat1D]`?

tidy3d/plugins/dispersion/web.py
```diff
@@ -0,0 +1,364 @@
...
+        title="Wavelengths",
+        description="A set of wavelengths for dispersion data.",
+        units=MICROMETER,
+    )
+    n_data: Tuple[float, ...] = Field(
+        ...,
+        title="Index of refraction",
+        description="Real part of the complex index of refraction at each wavelength.",
+    )
+    k_data: Tuple[float, ...] = Field(
```
weiliangjin2021, line 124: Also adds `Optional` here.

weiliangjin2021 (approved):
Wow, looks very good! Only one minor comment regarding "Optional".

tidy3d/plugins/dispersion/fit.py
lucas-flexcompute: Good question. My understanding is that we can default to None to indicate that the value can be missing, but, if the user decides to provide it, it must be an array to pass validation. This is how it is defined in many other places, that's why I left it like this. @tylerflex, is there any reason to prefer one way or the other?

tidy3d/plugins/dispersion/web.py
lucas-flexcompute: Same as above.

tidy3d/plugins/dispersion/fit.py
tylerflex: I think technically `Optional[ArrayFloat1D]` is correct, but my understanding is that pydantic implicitly knows it's optional if the default value is `None`. So we in fact have many parts of the code that do something like this. I think in this particular case, it's fine either way. and I'd say `Optional[]` should mainly be used in method calls, or when the default value is not None (but None would be accepted). If I were writing this, I'd probably just leave it without Optional just to keep the API reference slightly more readable.

lucas-flexcompute: I'll leave it as is, then.

PR #910, opened 2023-05-26
add custom medium to SIM_FULL in tests

tylerflex:
Hey @weiliangjin2021 I noticed CustomMedium was not in SIM_FULL in our tests and so I added it. However, it seems that the tests error. Do you have any idea what's going on here?

### DIFF
--- a/tests/sims/simulation_2_2_2.json
+++ b/tests/sims/simulation_2_2_2.json
@@ -307,6 +307,82 @@
         },
         {
             "geometry": {
+                "type": "Box",
+                "center": [
+                    -1.0,
+                    0.5,
+                    0.5
+                ],
+                "size": [
+                    1.0,
+                    1.0,
+                    1.0
+                ]
+            },
+            "name": null,
+            "type": "Structure",
+            "medium": {
+                "name": null,
+                "frequency_range": null,
+                "type": "CustomMedium",
+                "eps_dataset": {
+                    "type": "PermittivityDataset",
+                    "eps_xx": "ScalarFieldDataArray",
+                    "eps_yy": "ScalarFieldDataArray",
+                    "eps_zz": "ScalarFieldDataArray"
+                },
+                "interp_method": "nearest"
+            }
+        },
+        {
+            "geometry": {
+                "type": "PolySlab",
+                "axis": 2,
+                "sidewall_angle": 0.0,
+                "reference_plane": "middle",
+                "slab_bounds": [
+                    -1.0,
+                    1.0
+                ],
+                "dilation": 0.0,
+                "vertices": [
+                    [
+                        -1.5,
+                        -1.5
+                    ],
+                    [
+                        -0.5,
+                        -1.5
+                    ],
+                    [
+                        -0.5,
+                        -0.5
+                    ]
+                ]
+            },
+            "name": null,
+            "type": "Structure",
+            "medium": {
+                "name": null,
+                "frequency_range": null,
+                "type": "PoleResidue",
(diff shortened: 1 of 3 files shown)


weiliangjin2021:
I looked into why `CustomFieldSource` works while `CustomMedium` doesn't. I believe it's due to [those lines](https://github.com/flexcompute/tidy3d/blob/10ca408fcfbbd2c809e2490e7e823102681c0b7e/tidy3d/components/source.py#L586-L593) in `CustomFieldSource`. As I added those lines to `CustomMedium`, I realize there are additional complications in CustomMedium:

- As we don't have access to the permittivity data, any validator or function that calls `eps_model` or `eps_diagonal` will fail. This includes the validator `_warn_grid_size_too_small`. `AutoGrid` will also fail for the same reason.

So in this case, shall we skip structures of custom medium in those functions when we find the data are not loaded? When will we load json when custom medium is present?

tylerflex:
> So in this case, shall we skip structures of custom medium in those functions when we find the data are not loaded? 

Yea I think that makes sense. 

> When will we load json when custom medium is present?

I'm not sure I understand. The way it works now is that when the hdf5 file is loaded, the JSON_STRING contains the json minus all the custom medium and can be loaded as a dict. Then, we go through the hdf5 file recursively and load all the data into its proper spot in the dictionary. I think this part probably doesn't need to change?

weiliangjin2021:
> > When will we load json when custom medium is present?
> 
> I'm not sure I understand. The way it works now is that when the hdf5 file is loaded, the JSON_STRING contains the json minus all the custom medium and can be loaded as a dict. Then, we go through the hdf5 file recursively and load all the data into its proper spot in the dictionary. I think this part probably doesn't need to change?

Is there a use case where one only loads the json (hdf5 is absent)?

tylerflex:
> Is there a use case where one only loads the json (hdf5 is absent)?

Yes I think this is what happens on the backend, we will just set the DataArray values to None, I believe.

This [test](https://github.com/flexcompute/tidy3d/blob/develop/tests/test_components/test_IO.py#L54-L60) is an example, but it is for yaml and I'm surprised we dont have one with json, could you add a test like this but with json and include it in the PR? Basically, we should save a CustomMedium containing Simulation (SIM_FULL) to json and load it from json and the data should be absent.  Thanks

weiliangjin2021:
> > Is there a use case where one only loads the json (hdf5 is absent)?
> 
> Yes I think this is what happens in the GUI, we will just set the DataArray values to None, I believe.
> 
> This [test](https://github.com/flexcompute/tidy3d/blob/develop/tests/test_components/test_IO.py#L54-L60) is an example, but it is for yaml and I'm surprised we dont have one with json, could you add a test like this but with json and include it in the PR? Basically, we should save a CustomMedium containing Simulation (SIM_FULL) to json and load it from json and the data should be absent. Thanks

This [test](https://github.com/flexcompute/tidy3d/blob/10ca408fcfbbd2c809e2490e7e823102681c0b7e/tests/test_components/test_IO.py#L48) does that for json already?

In GUI, what is it used for? Does it matter if the grid poits are not consistent with the one from hdf5?

tylerflex:
Ok as long as that test covers it, I think it's fine.

I'm not sure about the specifics of the GUI honestly. I just know they wanted to be able to load a simulation from json file without the data but dont know how its used. I think the custom data is just not present in the GUI.

weiliangjin2021:
Let's close this PR and I'll open a new PR?

tylerflex:
sure, feel free to close if you wish

Issue #423, opened 2022-06-23
mode overlap


joamatab:
How can I do a mode overlap from the tidy3d modesolver?

https://gdsfactory.github.io/gdsfactory/notebooks/plugins/tidy3d/01_tidy3d_modes.html#Bend-loss

CC
@momchil-flex 
@twhughes 
@flaport

flaport:
something like this should work, I think:

```python
    cross = mode1.Ex * mode2.Hy.conj() - mode1.Ey * mode2.Hx.conj()
    return 0.25*np.trapz(np.trapz(cross, y), x)
```

Just be sure to zero out the phase before doing so. Something [like this in emepy](https://github.com/BYUCamachoLab/emepy/blob/64284b46e39e1a95ade76bfb6bcc87f8e568464f/emepy/mode.py#L314) could do the job. Although it's not perfect, I think. Maybe better to find the location of maximum EM energy density in the mode and then use that phase to zero out the phase.

momchil-flex:
Yeah what @flaport describes is pretty much what we do on the backend when you have a ModeMonitor. I will think about moving this part (specifically the mode overlap function) to the frontend.

joamatab:
yes Momchil, i think the mode-overlap calculations would be very useful for the client

Hi Floris, i get a ValueError

https://github.com/gdsfactory/gdsfactory/compare/master...5112
https://github.com/gdsfactory/gdsfactory/commit/bdd15e02d22cb84add40dcd9da146627f65114d9

```
ValueError: operands could not be broadcast together with shapes (1,100) (100,9)
```

@simbilod

joamatab:
@helgegehring got it working for the latest gdsfactory version

smartalecH:
Just passing by @flaport -- but you shouldn't need the conjugate. If the materials are lossless (and the modes are normalized right) then the conjugation won't change the results anyway. But the general mode overlap integral (derived from the unconjugated reciprocity theorem in [Snyder and Love](https://link.springer.com/book/10.1007/978-1-4613-2813-1)) is an unconjugated inner product.

The conjugated inner product we have in meep is more due to convention (physicists seem to like conjugated inner products...) and it's actually a poor choice because the current formulation isn't even holomorphic with the fields, but rather conjugate holomorphic... so the corresponding adjoint is a mess and needs to be reformulated.

(we also need to fix the overlaps in EMEpy...)

PS very nice package @momchil-flex et al!

momchil-flex:
[...]

PR #924, opened 2023-06-06
Fix overlap calculation for FieldData (#423)

lucas-flexcompute:
Skip selection by mode_index when it doesn't exist in the input and remove it from the output.

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -476,7 +476,8 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
         The tangential fields from ``field_data`` are interpolated to this object's grid, so the
         data arrays don't need to have the same discretization.  The calculation is performed for
         all common frequencies between data arrays.  In the output, ``mode_index_0`` and
-        ``mode_index_1`` are the mode indices from this object and ``field_data``, respectively.
+        ``mode_index_1`` are the mode indices from this object and ``field_data``, respectively, if
+        they are instances of ``ModeSolverData``.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


tidy3d/components/data/monitor_data.py
```diff
@@ -553,7 +566,15 @@ def outer_dot(
...
-        return MixedModeDataArray(dot, coords=coords)
+        result = MixedModeDataArray(dot, coords=coords)
+
+        # Remove mode index coordinate if the input did not have it
+        if not modes_in_self:
+            result = result.isel(mode_index_0=0, drop=True)
+        if not modes_in_other:
+            result = result.isel(mode_index_1=0, drop=True)
+
+        return result
```
momchil-flex, line 577: This is a bit weird in that the result will be a `MixedModeDataArray` if both `self` and `other` are `ModeFieldData`, but `xr.DataArray` otherwise. I think there is no issue with just always returning an `xr.DataArray`? We do have some data methods that return xarray data: for example, `colocate`. More generally it's easy for a user to move from our custom data types to xarray, e.g. as soon as they do `sel` or `isel`. @tylerflex remind me if there was any specific logic to when we should use our custom data types and if you think it's fine to just return `xr.DataArray` here (so just remove MixedModeDataArray's definition entirely, as it is only used here).

tylerflex: I think if it's something stored internally, we should use our own custom data types, for the result of a method or property, using xr.DataArray or xr.Dataset is fine. and maybe preferable

momchil-flex: Sounds good, @lucas-flexcompute could you make this change? I guess it's just making the result an `xr.DataArray` from the start, and you don't need to remove the `MixedModeDataArray` even though it's only used here (maybe we'll use it in the future). I was wondering if this can also help simplify this method (a lot of `if` checks for a mode dimension currently), but I don't see an obvious way.

lucas-flexcompute: I though about it as well and the alternatives I could think of looked more cumbersome then the current ifs.

PR #932, opened 2023-06-08
Extrapolation clamping for complex fields

lucas-flexcompute:


tidy3d/components/grid/grid.py
```diff
@@ -119,10 +119,23 @@ def spatial_interp(
...
+                limit = spatial_dataarray.values.imag.max()
+                im[re > limit] = limit
+                limit = spatial_dataarray.values.imag.min()
+                im[re < limit] = limit
+                interp_dataarray.values[:] = re + 1j * im
+            else:
+                max_val = spatial_dataarray.values.max()
+                min_val = spatial_dataarray.values.min()
+                interp_dataarray = interp_dataarray.where(interp_dataarray >= min_val, min_val)
+                interp_dataarray = interp_dataarray.where(interp_dataarray <= max_val, max_val)
```
momchil-flex, line 138: How about a single function like

```
def cap_array(array):
    max_val = spatial_dataarray.values.max()
    min_val = spatial_dataarray.values.min()
    capped_array = array.where(array >= min_val, min_val)
    capped_array = capped_array.where(capped_array <= max_val, max_val)
    return capped_array
```

which is then used three times?

lucas-flexcompute: There is a function already: `numpy.clip`

lucas-flexcompute:
Added a "nearest" extrapolation option to fix issues with bad extrapolation in directions with 2 data points. I wonder if we still want to keep the previous behavior of having linear extrapolation followed by a clip operation or just use nearest extrapolation by default. Sources are current using 'nearest', but custom medium uses 'extrapolation'+clip. Would it be ok to change it to use 'nearest' extrapolation, @weiliangjin2021? If so, the function can be simplified quite a bit.

PR #933, opened 2023-06-08
Rename causal medium to stable medium in error message

weiliangjin2021:


tylerflex (approved):

momchil-flex:
Wasn't the point of the discussion yesterday to also make this a warning rather than an error if "allow_gain" is on (and maybe rename `allow_gain` to something else?)

weiliangjin2021:
> Wasn't the point of the discussion yesterday to also make this a warning rather than an error if "allow_gain" is on (and maybe rename `allow_gain` to something else?)

I'll push a change soon to make it a warning. Regarding `allow_gain` vs `allow_unstable`, personally I still prefer the former, as it indicates the physics, while `unstable` seems quite numerical. So how about adding the flag `allow_unstable` later once there is a request in simulation a non-causal medium?

PR #935, opened 2023-06-09
Backward modes

dbochkov-flexcompute:
Added direction field in `ModeSolver` and `ModeSolverMonitor`. When decomposing modes we need to also choose the right sign for the normal of the differential area. I did that by redefining `.dot()` methods for `ModeSolverData`, but maybe there is a better way? Like, making `_diff_area` negative?


tidy3d/components/data/monitor_data.py
```diff
@@ -945,6 +946,113 @@ def _reorder_modes(
...
+            modes in lossless waveguides but differs for modes in lossy materials. In that case,
+            the conjugated dot product can be interpreted as the fraction of the power of the first
+            mode carried by the second, but modes are not orthogonal with respect to that product
+            and the sum of carried power fractions exceed 1. In the non-conjugated definition,
+            orthogonal modes can be defined, but the interpretation of modal overlap as power
+            carried by a given mode is no longer valid.
+        """
+
+        # Flip integration direction for backward monitors
+        if self.monitor.direction == "-":
```
tylerflex, line 980: nitpicky but this seems a bit cleaner to avoid repetition.
```py
sign = -1 if self.monitor.direction == "-" else 1.0
return sign * super().dot(field_data, conjugate)
```

tidy3d/components/data/monitor_data.py
```diff
@@ -945,6 +946,113 @@ def _reorder_modes(
...
 
+    def dot(self, field_data: ElectromagneticFieldData, conjugate: bool = True) -> xr.DataArray:
+        """Dot product (modal overlap) with another :class:`.FieldData` object. Both datasets have
+        to be frequency-domain data associated with a 2D monitor. Along the tangential directions,
+        the datasets have to have the same discretization. Along the normal direction, the monitor
+        position may differ and is ignored. Other coordinates (``frequency``, ``mode_index``) have
+        to be either identical or broadcastable. Broadcasting is also supported in the case in
+        which the other ``field_data`` has a dimension of size ``1`` whose coordinate is not in the
+        list of coordinates in the ``self`` dataset along the corresponding dimension. In that case,
+        the coordinates of the ``self`` dataset are used in the output.
```
tylerflex, line 957: worth adding a Note about the sign depending on direction?

tidy3d/components/data/monitor_data.py
```diff
@@ -945,6 +946,113 @@ def _reorder_modes(
...
+            carried by a given mode is no longer valid.
+        """
+
+        # Flip integration direction for backward monitors
+        if self.monitor.direction == "-":
+            return -super().dot(field_data, conjugate)
+
+        return super().dot(field_data, conjugate)
+
+    def outer_dot(
```
tylerflex, line 985: same comments as `dot()`

tidy3d/components/monitor.py
```diff
@@ -489,6 +489,12 @@ class ModeSolverMonitor(AbstractModeMonitor):
     ...     name='mode_monitor')
     """
 
+    direction: Direction = pydantic.Field(
+        "+",
+        title="Propagation direction",
+        description="Direction of waveguide mode propagation.",
```
tylerflex, line 495: maybe add "along the axis defined by its normal dimension." or something to this effect.

tylerflex (commented):
looks good to me. Hard for me to see, but are there any backwards compatibility issues or conflicts with the remote mode solver? I guess one thing I'm worried about is what happens if the remote mode solver expects `direction` but an older version of the mode solver is supplied?

dbochkov-flexcompute:
A substantial revision of this PR:
- Moved logic on how to calculate backward modes inside mode solver plugin
- Switch for backward modes is based on eps tensor values
- Mode solver direction can be "+", "-", or "both". In the later case, it returns two `ModeSolverData`. And those are identical to when mode solver is called with just"+" or "-"

momchil-flex (commented):
I think this is getting close, but I'd like to think about this a little more. I am not a big fan of all the branching based on "both", and the inconsistent return of various functions based on the input. This is also problematic to incorporate with the remote mode solver, which by default returns a single ModeSolverData now (it's not impossible, but it will require changes there too).

Here's what I'm thinking, but it would require some further reorganization specifically of `EigSolver`. We could have that take the `compute_modes` input arguments as fields, instead of the current approach (which is just based on a very old legacy...) of having classmethods. Then, we could separate out the check for whether a separate solve is needed outside of `EigSolver.solver_em` to just something like `EigSolver.eps_spec` which does the epsilon transformation, etc. but just returns something like `{"tensorial": True, "complex": "False"}` without doing the actual solve. The same method would of course also be used by the actual solve when compute_modes is called, to determine the solver to use and how to compute backwards modes. For this, the epsilon transformations would also be separated in a different method, outside of `compute_modes`.

Then, we could just allow `ModeSolver.direction` to be either `"+"` or `"-"`. On the backend, if we want to reuse computation when we need both directions, we could do something like:

```
mode_solver_fwd = ModeSolver(..., direction="+")
mode_data_fwd = mode_solver_fwd.data_raw
mode_solver_bwd = ModeSolver(..., direction="-")
eps_spec = mode_solver_bwd._eps_spec # Internally, call EigSolver.eps_spec
if eps_spec["tensorial"] and eps_spec["complex"]:
    # needs a separate solve    
    mode_data_bwd = mode_solver_bwd.data_raw
else:
    # use mode_data_fwd with reflection / time reversal, replicating the login inside EigSolver
```

Maybe even this is not ideal but can you think along these lines? How does it sound?

dbochkov-flexcompute:
> Maybe even this is not ideal but can you think along these lines? How does it sound?

Yeah, I think that should be possible. I just wanted to avoid calling the eps getter another time, but it's probably not significant compared to an actual eigensolve. I thought about maybe having a flag inside ModeSolver that is set during first solve and basically carries this information, but couldn't come up with a good implementation

> I am actually still puzzled by this, namely in the limit of imag(eps) -> 0, the two possible transformations differ in the sign of the longitudinal fields. For our practical purposes, this shouldn't matter as only tangential fields enter in mode overlaps, source injection, etc. But do you have an idea of a good way to think about this discrepancy?

Isn't it that in this case (that is, $\varepsilon_{zx} = \varepsilon_{zy} = \varepsilon_{xz} = \varepsilon_{yz} = 0$ and the rest are real) tangential components are purely real, while normal are purely imaginary, so $E_{x,y}^\ast = E_{x,y}$ and $E_z^\ast = - E_z$ and similar for $H$

dbochkov-flexcompute:
Next iteration:
- moved simultaneous +/- solve out of mode solver to backend
- for now implemented it through adding to `ModeSolverData` an additional field `eps_spec` that describes characterization of eps profile in mode solver plane
- had to static'ize `ModeSolver._grid_correction()` so that it could be called without a properly defined mode solver instance

tidy3d/components/data/monitor_data.py
```diff
@@ -968,6 +973,24 @@ def _reorder_modes(
 
         return self.copy(update=update_dict)
 
+    @property
+    def time_reversed_copy(self) -> FieldData:
+        """Make a copy of the data with direction-reversed fields."""
```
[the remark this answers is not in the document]
momchil-flex: This comment is still relevant.

momchil-flex (approved):
I think this looks good now! Any known issues / inefficiencies / things you'd like to still fix some day?

dbochkov-flexcompute:
Rebased, updated docs and frontend tests, rerun backend tests. Also noticed that in group index calculation monitors frequencies were not updated. Btw, `tests/test_web/test_webapi.py` for some reason gets stuck for me on current `pre/2.3`.

Regarding "Any known issues / inefficiencies / things you'd like to still fix some day?", it looks pretty solid to me know, can't think of any immediate issues/inefficiencies

momchil-flex:
> Rebased, updated docs and frontend tests, rerun backend tests. Also noticed that in group index calculation monitors frequencies were not updated. Btw, `tests/test_web/test_webapi.py` for some reason gets stuck for me on current `pre/2.3`.

Yep, was working on fixing that, sent you a PR, could you have a look since tyler is away?

> Regarding "Any known issues / inefficiencies / things you'd like to still fix some day?", it looks pretty solid to me know, can't think of any immediate issues/inefficiencies

Awesome, let's maybe merge the PR that fixes the tests, restart the tests here, and merge everything when they pass.

momchil-flex:
Closing PR to stop tests now, will reopen to force them to restart later.

PR #938, opened 2023-06-12
Merge touching polygons in Polygon side plot

weiliangjin2021:
Fix the horizontal line issue in https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/222


lucas-flexcompute (approved):
This look correct to me. I only wonder if, performance-wise, isn't it better to perform the merge in a single call using [`unary_union`](https://shapely.readthedocs.io/en/stable/reference/shapely.unary_union.html)

The method I found most robust to weird input geometries was using the buffer trick you showed me before merging, to make sure self-intersections are healed without collapsing holes:
```python
result = shapely.unary_union([shapely.Polygon(vertices).buffer(0) for vertices in ...])
```
I don't know if it makes any difference here, though.

tomflexcompute (approved):
Tested various cases and it works well. Thanks for the fix!

weiliangjin2021:
> This look correct to me. I only wonder if, performance-wise, isn't it better to perform the merge in a single call using [`unary_union`](https://shapely.readthedocs.io/en/stable/reference/shapely.unary_union.html)
> 
> The method I found most robust to weird input geometries was using the buffer trick you showed me before merging, to make sure self-intersections are healed without collapsing holes:
> 
> ```python
> result = shapely.unary_union([shapely.Polygon(vertices).buffer(0) for vertices in ...])
> ```
> 
> I don't know if it makes any difference here, though.

Thanks for the suggestion. It does look better. Do you think we should update here as well https://github.com/flexcompute/tidy3d/blob/10ca408fcfbbd2c809e2490e7e823102681c0b7e/tidy3d/components/geometry.py#L2304-L2306

In this particular place (visualization), `buffer(0)` is perhaps not necessary, as those polygons shouldn't be self-intersecting.

lucas-flexcompute:
> Do you think we should update here as well

I think so. This should be safer than just trusting the input geometry to be perfect.

weiliangjin2021:
buffer and unary_union have been applied to the two places. @tomflexcompute, would you mind running the same tests again?

tomflexcompute:
> buffer and unary_union have been applied to the two places. @tomflexcompute, would you mind running the same tests again?

Yes it still works correctly.

PR #940, opened 2023-06-13
Simulation.allow_gain and a small update to the allow_gain docstring

momchil-flex:
Anything else that needs to be modified?

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -2769,3 +2769,15 @@ class Simulation(Box):  # pylint:disable=too-many-public-methods
         """Generate a tuple of structures wherein any 2D materials are converted to 3D
         volumetric equivalents."""
         return self._volumetric_structures_grid(self.grid)
+
+    @cached_property
+    def allow_gain(self) -> bool:
+        """``True`` if any of the mediums in the simulation allows gain."""
+
+        for medium in self.mediums:
+            if isinstance(medium, AnisotropicMedium):
+                if np.any([med.allow_gain for med in [medium.xx, medium.yy, medium.zz]]):
+                    return True
+            elif medium.allow_gain:
+                return True
+        return False
--- a/tests/sims/simulation_2_3_0rc1.json
+++ b/tests/sims/simulation_2_3_0rc1.json
@@ -14,6 +14,7 @@
     "medium": {
         "name": null,
         "frequency_range": null,
+        "allow_gain": false,
         "type": "Medium",
         "permittivity": 1.0,
         "conductivity": 0.0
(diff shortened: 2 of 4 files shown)


tidy3d/components/simulation.py
```diff
@@ -2769,3 +2769,8 @@ def volumetric_structures(self) -> Tuple[Structure]:
         """Generate a tuple of structures wherein any 2D materials are converted to 3D
         volumetric equivalents."""
         return self._volumetric_structures_grid(self.grid)
+
+    @cached_property
+    def allow_gain(self) -> bool:
+        """``True`` if any of the mediums in the simulation allow gain."""
+        return np.any([med.allow_gain for med in self.mediums])
```
weiliangjin2021, line 2776: In AnisotropicMedium, only the value of each component's `allow_gain` matters. AnisotropicMedium's `allow_gain` is dummy in the current implementation: https://github.com/flexcompute/tidy3d/blob/37aca77d86c53ad7bdad0df5fe0deebc848039d7/tidy3d/components/medium.py#L2332-L2335

So if `isinstance(med, AnisotropicMedium)`, now it requires an additional step in looking through its components.

momchil-flex: I see. I think the class tree of the Medium could use some reorganization, I think we're all aware of that, but somewhat tied up before 3.0. But generally it would seem that having separate abstract classes for an isotropic medium, diagonally anisotropic medium, and fully anisotropic medium, such that any actual medium is a subclass of one and exactly one of these, would be great. Actually I'm not sure if even that would solve the problem because of all the custom/anisotropic possibilities.

Not a huge fan on doing things based on `isinstance(med, AnisotropicMedium)` and I know that @tylerflex isn't either. But I don't see an easy way around this.

By the way what about fully anisotropic medium, no gain is allowed right now right? I think you asked about it at some point and we agreed on that.

weiliangjin2021: > By the way what about fully anisotropic medium, no gain is allowed right now right? I think you asked about it at some point and we agreed on that.

I don't think I've been involved in discussions on this. Maybe @dbochkov-flexcompute? Currently gain is allowed: https://github.com/flexcompute/tidy3d/blob/37aca77d86c53ad7bdad0df5fe0deebc848039d7/tidy3d/components/medium.py#L2484-L2498

momchil-flex: Oh, I think I'm just confused with some other discussion.

momchil-flex:
How about this @weiliangjin2021 ? Added a test too.

weiliangjin2021:
Looks good!

weiliangjin2021 (approved):

PR #943, opened 2023-06-13
Casting mode solver matrix to single precision only after it is constructed

momchil-flex:


tidy3d/plugins/mode/solver.py
```diff
@@ -259,7 +259,7 @@ def conductivity_model_for_pec(eps, threshold=0.9 * pec_val):
         if np.any(eps_offd > TOL_TENSORIAL) or np.any(mu_offd > TOL_TENSORIAL):
```
lucas-flexcompute, line 259: Setting a boolean flag from an if-clause looks like an anti-pattern:
```python
is_tensorial = np.any(eps_offd > TOL_TENSORIAL) or np.any(mu_offd > TOL_TENSORIAL)
```
Also below in `complex_solver`.

tidy3d/plugins/mode/solver.py
```diff
@@ -380,11 +383,13 @@ def any_pec(eps_vec, threshold=0.9 * np.abs(pec_val)):
             print(spl.norm(diff, ord="fro"), spl.norm(aca, ord="fro"), spl.norm(aac, ord="fro"))
 
         # Call the eigensolver. The eigenvalues are -(neff + 1j * keff)**2
+
+        print(mat.dtype, eig_guess.dtype, vec_init.dtype)
```
lucas-flexcompute, line 387: This is probably left from debugging, right?

lucas-flexcompute (approved):

tidy3d/plugins/mode/solver.py
momchil-flex: Ah yeah thanks, who knows when this was a leftover from...

tidy3d/plugins/mode/solver.py
momchil-flex: yep, removed

tidy3d/plugins/mode/solver.py
```diff
@@ -268,41 +286,32 @@ def conductivity_model_for_pec(eps, threshold=0.9 * pec_val):
...
-            )
-
-        return cls.solver_diagonal(eps_tensor, mu_tensor, der_mats, num_modes, vec_init, neff_guess)
+    @classmethod
+    def trim_small_values(cls, mat: sp.csr_matrix, tol: float) -> sp.csr_matrix:
+        """Eliminate elements of matrix ``mat`` for which ``abs(element) / abs(max_element) < tol``.
+        This operates in-place on mat so there is no return.
+        """
+        max_element = np.amax(np.abs(mat))
+        mat.data *= np.abs(mat.data) / max_element > tol
```
weiliangjin2021, line 309: Will it be safer with `mat.data *= np.logical_or(np.abs(mat.data) / max_element > tol, np.abs(mat.data) > tol)`. This doesn't change the number of nonzeros in your example.

momchil-flex: Good thinking, I updated this and added a line to changelog, looks good now?

weiliangjin2021: All look good to me

weiliangjin2021 (approved):

PR #948, opened 2023-06-15
fix callback url

tylerflex:


tidy3d/web/simulation_task.py
```diff
@@ -228,7 +228,7 @@ def create(
             f"tidy3d/projects/{folder.folder_id}/tasks",
             {
                 "taskName": task_name,
-                "call_back_url": call_back_url,
+                "callback_url": callback_url,
```
lei-flex, line 231: it should be callbackUrl not callback_url

PR #953, opened 2023-06-16
Improvements to graphene fitter

caseyflex:
- Fixes poor fitting performance under certain parameter choices
- Adds a warning if the fit is not good
- Extends the testing to more physical parameters
- Fixes the error message related to stability of intermediate calculations
- Fixes warnings related to numerical integration

Limitations:

- Fit still can be imperfect, or have a small passivity violation, under certain parameter choices. It is still always stable, so the simulation will still usually be stable.
- Need to check that the bounds on the physical parameters are reasonable @tomflexcompute 

I checked that the graphene notebook is not affected by this.


tomflexcompute:
Thanks for the improvement. Just wondering how the bounds of the physical parameters would affect the fit? In terms of realistic values, I would suggest GRAPHENE_MU_C_MAX = 3, GRAPHENE_TEMP_MIN = 0 or 1, GRAPHENE_GAMMA_MAX = 0.03. Do you think the robustness of the fit can be maintained after changing to these values?

momchil-flex (approved):
Looks good, hopefully it won't have to be revisited.

tomflexcompute (approved):
Thanks. Feel free to merge.

PR #954, opened 2023-06-16
symmetry support in JaxFieldData

tylerflex:


momchil-flex:
Looks good. A little worried about this way of handling things getting out of hand, but `FieldData` is probably the biggest thing that will need this, so if it doesn't get out of hand for that, it will probably be fine? :D

momchil-flex (approved):

PR #952, opened 2023-06-16
Align configs.

an-li-the-dev:
This is a pr to align config.
* Remove credentials authentication
* Remove config.py, use envirnoment.py instead

PR #956, opened 2023-06-16
fix dispersion fitter to use consolidated web API config

tylerflex:
we merged some [changes](https://github.com/flexcompute/tidy3d/pull/952) into develop that consolidated the webAPI config. When we merged these changes into pre/2.3, the web fitter was out of date. I tried to fix it but could you both 
(an) take a look and make sure this is the right fix and (Lucas) test it? 

Thanks!

### DIFF
--- a/tests/sims/simulation_2_3_0rc2.json
+++ b/tests/sims/simulation_2_3_0rc2.json
@@ -1,0 +1,1747 @@
+{
+    "type": "Simulation",
+    "center": [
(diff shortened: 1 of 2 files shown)


lucas-flexcompute (approved):
Looks correct to me. The tests here all worked.

tylerflex:
ok thanks, I think I'll just merge this then.

PR #957, opened 2023-06-16
Estimate cost should now always work

momchil-flex:
Now `taskInfo.estFlexUnit` is always used, and changes have been made on the server for a `taskInfo.metadataStatus` to be added. If the web api times out, a dedicated worker picks up the task to ensure that the metadata will eventually be processed.

This only works on dev env currently. Let's merge only once the server changes are in prod.

### DIFF
--- a/tidy3d/web/webapi.py
+++ b/tidy3d/web/webapi.py
@@ -686,7 +686,7 @@ def estimate_cost(task_id: str) -> float:
     Note
     ----
     Cost is calculated assuming the simulation runs for
-    the full ``run_time``. If early shut-off is triggered, the cost is adjusted proporionately.
+    the full ``run_time``. If early shut-off is triggered, the cost is adjusted proportionately.
 
     Parameters
     ----------
@@ -702,14 +702,25 @@ def estimate_cost(task_id: str) -> float:
     task = SimulationTask.get(task_id)
     if not task:
         raise ValueError("Task not found.")
-    resp = task.estimate_cost()
-    flex_unit = resp.get("flex_unit")
-    if not flex_unit:
-        log.warning(
-            "Could not get estimated cost! It will be reported in preprocessing upon "
-            "simulation run."
-        )
-    return flex_unit
+
+    task.estimate_cost()
+    task_info = get_info(task_id)
+    status = task_info.metadataStatus
+
+    # Wait for a termination status
+    while status not in ["processed", "success", "error", "failed"]:
+        time.sleep(REFRESH_TIME)
+        task_info = get_info(task_id)
+        status = task_info.metadataStatus
+
+    if status in ["processed", "success"]:
+        return task_info.estFlexUnit
+
+    log.warning(
+        "Could not get estimated cost! It will be reported during a simulation run in the "
+        "preprocessing step."
+    )
+    return None
 
 
 @wait_for_connection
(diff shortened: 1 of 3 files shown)


tidy3d/web/webapi.py
```diff
@@ -702,14 +702,24 @@ def estimate_cost(task_id: str) -> float:
...
-    resp = task.estimate_cost()
-    flex_unit = resp.get("flex_unit")
-    if not flex_unit:
-        log.warning(
-            "Could not get estimated cost! It will be reported in preprocessing upon "
-            "simulation run."
-        )
-    return flex_unit
+
+    task.estimate_cost()
```
tylerflex, line 706: does this line do something?

tidy3d/web/webapi.py
```diff
@@ -702,14 +702,24 @@ def estimate_cost(task_id: str) -> float:
...
+    task_info = get_info(task_id)
+    status = task_info.metadataStatus
+
+    # Wait for a termination status
+    while status not in ["processed", "error"]:
```
tylerflex, line 711: worried about this getting stuck in an infinite loop, eg if `status` is `None` for some reason. Do you think this is an issue and should we safeguard against it?

tidy3d/web/webapi.py
```diff
@@ -702,14 +702,24 @@ def estimate_cost(task_id: str) -> float:
...
+        time.sleep(REFRESH_TIME)
+        task_info = get_info(task_id)
+        status = task_info.metadataStatus
+
+    if status == "processed":
+        return task_info.estFlexUnit
+
+    log.warning(
+        "Could not get estimated cost! It will be reported in preprocessing upon simulation run."
```
tylerflex, line 720: maybe confusing what is meant by "reported in preprocessing". Perhaps just say how to obtain it after the simulation run?

tylerflex (approved):
a few comments, but otherwise looks good. will set this PR as a draft until we are ready to ensure it doesn't get accidentally merged in. Recently found out that you can do that through here..
image

tidy3d/web/webapi.py
momchil-flex: Yes, it kicks off a remote metadata processing call

tidy3d/web/webapi.py
tylerflex: yea I suppose it is, except that design is kind of fragile too as we've seen. I guess maybe more concretely, if the metadata processing status keys changed in the future (eg to include "errored"), we'd just want this to never enter the loop, rather than get caught infinitely in it. So maybe I'd suggest something like
```
processing_statuses = [blah, blah, blah]

while status in processing_statuses:
    ...

if status == "processed":
    ....
elif status == "error"
   ...

return None
```
kind of thing maybe?

tidy3d/web/webapi.py
momchil-flex: This message was there even before btw. The estimate cost is printed during web.monitor, but only when the task status gets to "preprocess", because that's when we know the estimated flex unit should be populated and correct. That message also says that the user can call `web.real_cost` after the simulation run to get the billed cost.

How about "It will be reported during a simulation run in the preprocessing step."? Btw hopefully this message should never be encountered anymore.

tylerflex: ok sounds good

tidy3d/web/webapi.py
momchil-flex: @lei-flex what are the currently possible values for `metadataStatus`? And do you have any suggestion for good handling?

PR #959, opened 2023-06-19
Field projection docstrings improvement

shashwat-sh:
Added more detail to the field projection docstrings based on user feedback. Thanks @tomflexcompute for pointing this out.

### DIFF
--- a/tidy3d/components/monitor.py
+++ b/tidy3d/components/monitor.py
@@ -552,7 +552,11 @@ class AbstractFieldProjectionMonitor(SurfaceIntegrationMonitor, FreqMonitor):
     far_field_approx: bool = pydantic.Field(
         True,
         title="Far field approximation",
-        description="Whether to enable the far field approximation when projecting fields.",
+        description="Whether to enable the far field approximation when projecting fields. "
+        "If ``True``, terms that decay as O(1/r^2) are ignored, as are the radial components "
+        "of fields. Typically, this should be set to ``True`` only when the projection distance "
+        "is much larger than the size of the device being modeled, and the projected points are "
+        "in the far field of the device.",
     )
 
     @property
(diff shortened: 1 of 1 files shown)


tomflexcompute (commented):
Thanks @shashwat-sh . Looks great. Could you also add some general comments about how the users should set the position and size of the projection monitors?

shashwat-sh:
> Thanks @shashwat-sh . Looks great. Could you also add some general comments about how the users should set the position and size of the projection monitors?

@tomflexcompute  I added an additional comment on the size. There is already a mention on the position. I don't think we should be too specific though, at some point the users need to understand the physics of the problem enough to know how to set this up correctly for their application. If we make it too specific, I'm worried users will take our suggestion as a "rule", apply it incorrectly, and then blame the solver. @momchil-flex  maybe you can take a look too, to see if these suggestions are general enough in your opinion?

tomflexcompute (approved):
Thanks @shashwat-sh . They look good to me. We definitely want to keep them as general as possible but provide some information at the same time.

tidy3d/components/monitor.py
```diff
@@ -552,7 +552,10 @@ class AbstractFieldProjectionMonitor(SurfaceIntegrationMonitor, FreqMonitor):
     far_field_approx: bool = pydantic.Field(
         True,
         title="Far field approximation",
-        description="Whether to enable the far field approximation when projecting fields.",
+        description="Whether to enable the far field approximation when projecting fields. "
+        "If ``True``, terms that decay as O(1/r^2) are ignored, as are the radial components "
+        "of fields. Typically, this should be set to ``True`` only when the projection distance "
+        "is much larger than the size of the device being modeled.",
```
tylerflex, line 558: out of curiosity, does the wavelength play into whether `far_field_approx` is accurate?

tylerflex (approved):

tylerflex:
could you run `black`? and then feel free to merge, thanks!

tidy3d/components/monitor.py
shashwat-sh: Yeah, in that the observations points should be in the far field of the device - I added a sentence to make this explicit. I didn't specify what number of wavelengths would be considered "far field" because I think that can be application-dependent sometimes.

tidy3d/components/monitor.py
```diff
@@ -579,7 +583,18 @@ def local_origin(self) -> Coordinate:
 
 class FieldProjectionAngleMonitor(AbstractFieldProjectionMonitor):
     """:class:`Monitor` that samples electromagnetic near fields in the frequency domain
-    and projects them at given observation angles.
+    and projects them at given observation angles. The ``center`` and ``size`` fields define
+    where the monitor will be placed in order to record near fields, typically on the order
+    of a wavelength away from the structure of interest. The near fields are then projected
```
momchil-flex, line 588: Maybe rather something like `typically very close to the structure of interest`? I'm wondering if this phrasing makes it sounds like you *should* have some spacing, like a wavelength - while in reality you can put the monitor right next to the structure.

momchil-flex (approved):

tidy3d/components/monitor.py
```diff
@@ -579,7 +583,18 @@ def local_origin(self) -> Coordinate:
...
+    to far-field locations defined by ``phi``, ``theta``, and ``proj_distance``, relative
+    to the ``custom_origin``. If the distance between the near and far field locations is
+    much larger than the size of the device, one can typically set ``far_field_approx`` to
+    ``True``, which will make use of the far-field approximation to speed up calculations.
+    If the projection distance is comparable to the size of the device, we recommend setting
+    ``far_field_approx`` to ``False``, so that the approximations are not used, and the
+    projection is accurate even just a few wavelengths away from the near field locations.
+    For most applications, it is advisable to pick the size of the monitor such that the
+    recorded near fields decay to negligible values near the edges of the monitor.
```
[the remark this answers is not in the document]
shashwat-sh: Good point, thanks, made both changes

PR #960, opened 2023-06-19
Fixes to tests broken by numpy 1.25

momchil-flex:


momchil-flex:
I keep retesting but the macos python 3.7 test keeps failing. Otherwise looks good though.

Shall we just remove this env from the github actions now?

tylerflex (approved):
yea we can probably just remove it, although I've noticed it sometimes fails on 3.9 macos so not sure. I'll merge this in because it clearly is working besides this bz2 issue.

PR #962, opened 2023-06-20
Parallelizing store_vjp over structures

momchil-flex:
This is not exactly pretty currently, but I'm not sure how to improve. Have to pass these big args lists around that can be confusing. However, when you work with a parallel pool, you can't directly use values and functions that are in the local namespace.

The other complication is that some structures need to *internally* use multi-processing (PolySlab, GeometryGroup), while others don't do that. So we split the list in two parts, structures where store_vjp is called in serial with num_proc internally, and structures for which the store_vjp calls are parallelized over num_proc, with no multi-processing inside store_vjp.

Ideally, some day we can refactor that, e.g. if PolySlab is fully vectorized, and GeometryGroup could be unraveled, maybe.

### DIFF
--- a/tidy3d/plugins/adjoint/web.py
+++ b/tidy3d/plugins/adjoint/web.py
@@ -21,9 +21,7 @@ from .components.simulation import JaxSimulation, JaxInfo
 from .components.data.sim_data import JaxSimulationData
 
 
-# TODO: confused about these paths, they aren't local but rather wherre the jax info and sim_vjp get
-# stored on the server? maybe add a comment for each?
-# when i changed them, adjoint runs errored on the solver.
+# file names and paths for server side adjoint
 SIM_VJP_FILE = "output/jax_sim_vjp.hdf5"
 JAX_INFO_FILE = "jax_info.json"
 
--- a/tidy3d/plugins/adjoint/web.py
+++ b/tidy3d/plugins/adjoint/web.py
@@ -311,6 +311,7 @@ def webapi_run_adjoint_fwd(
         verbose=verbose,
         simulation_type="adjoint_fwd",
         jax_info=jax_info,
+        solver_version="parallel-0.0.0",
     )
 
     sim_data = job.run()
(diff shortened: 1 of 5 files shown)


tidy3d/plugins/adjoint/web.py
```diff
@@ -313,6 +313,7 @@ def webapi_run_adjoint_fwd(
         verbose=verbose,
         simulation_type="adjoint_fwd",
         jax_info=jax_info,
+        # solver_version="parallel-0.0.0"
```
tylerflex, line 316: remove these?

tylerflex (changes requested):
I think this PR, while necessary, is going to lead to a lot of headaches down the line when I try to develop on top of it. I was thinking, is there a way to  reorganize it a bit so that all of the multiprocessing stuff happens completely separately from the regular pipeline? For example, keep the `store_vjp` functions as they were before but then move all of the parallel things to their own `store_vjp_parallel()` methods that wrap this? Rather than doing conditional switching based on `num_proc`.  Then, we can easily swap back and forth between the parallel and non-parallel by changing which method gets called in the base `Simulation.store_vjp`. This base method call can be controlled by an extra field `parallel:bool=True` in `JaxSimulation`, perhaps. I think this would make it a lot easier to generalize all of the vjp stuff down the line and reason about the code. It's already hard enough to reason about before these changes. what do you think?

Only complication I can see is that the PolySlab uses num_proc under the hood, but with a little refactoring there to add a parallel version of `store_vjp()` in `JaxPolySlab` that wraps the original `store_vjp`, we should be fine there as well.

momchil-flex:
I refactored things in a way that I think simplifies them a bit, and I took your suggestion of having a separate `JaxSimulation.store_vjp_parallel`. However, beyond that I'm still passing `num_proc` everywhere, because otherwise it seems to me a giant overhead of making `Structure` and all `Geometry` types to have both a `store_vjp` and a `store_vjp_parallel` method. I think this is currently not too bad actually?

tylerflex (approved):
see my PR into this branch #969 . I think as long as these changes dont break anything, we might as well merge for 2.3. My preference is to get my PR in if possible, I'm also not 100% sure about current num_proc passing logic as I mentioned.  I can pick up the pieces when I work on generalizing the geometry VJPs a a later time and it would be good to have a performance boost until then.

PR #963, opened 2023-06-20
Set max time steps to 20M and warn if more than 1M

momchil-flex:


tomflexcompute (approved):
Looks good. Would `MAX_TIME_STEPS = 2e7` ensure there is no memory issue even on a node?

PR #964, opened 2023-06-21
remove python 3.7 from tested envs

tylerflex:
To alleviate missing `bz2` package in github's Mac OS Python3.7 environments.


momchil-flex:
Could you try just excluding the macos one? https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations

Also you shouldn't have to change the tox file.

tylerflex:
k just pushed, guess we'll see.

PR #965, opened 2023-06-21
Generalized reverse interpolation for current sources

shashwat-sh:
Associated backend PR: https://github.com/flexcompute/tidy3d-core/pull/326

### DIFF
--- a/tidy3d/components/source.py
+++ b/tidy3d/components/source.py
@@ -410,7 +410,20 @@ class CurrentSource(Source, ABC):
         return pol_vec
 
 
-class UniformCurrentSource(CurrentSource):
+class ReverseInterpolatedSource(Source):
+    """Abstract source that allows reverse-interpolation along zero-sized dimensions."""
+
+    interpolate: bool = pydantic.Field(
+        True,
+        title="Enable Interpolation",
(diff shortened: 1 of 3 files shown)


tidy3d/components/source.py
```diff
@@ -410,7 +410,18 @@ def _pol_vector(self) -> Tuple[float, float, float]:
         return pol_vec
 
 
-class UniformCurrentSource(CurrentSource):
+class ReverseInterpolatedSource(Source):
+    """Abstract source that allows reverse-interpolation along zero-sized dimensions."""
+    interpolate: bool = pydantic.Field(
+        True,
+        title="Enable Interpolation",
+        description="If ``False``, the dipole is placed at the nearest Yee grid point "
```
tylerflex, line 418: I think we'll need to update the description. says "dipole" now and references "polarization".  Maybe something like
```py
Handles interpolation of size=0 dimensions of the source. if ``False``, the source data is snapped to the nearest Yee grid point. If ``True``, the source data is applied on the surrounding Yee grid points to emulate placement on the specified location using linear interpolation.
```
or something to this effect (needs refinement).

tylerflex (changes requested):

PR #966, opened 2023-06-21
Add subpixel field in custom medium

weiliangjin2021:
- Applies subpixel to a structure made of custom medium if 1) simulation's subpixel=True; 2) custom medium's subpixel=True
- For CustomAnisotropicMedium, each component has its own subpixel field. E.g. if xx-component has subpixel=True, and yy-component has subpixel=False, subpixel will be applied to xx component but not yy component.

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -355,6 +355,15 @@ class AbstractCustomMedium(AbstractMedium, ABC):
         "the extrapolated value will take the minimal (maximal) of the supplied data.",
     )
 
+    subpixel: bool = pd.Field(
+        False,
+        title="Subpixel averaging",
+        description="If ``True`` and simulation's ``subpixel`` is also ``True``, "
+        "applies subpixel averaging of the permittivity "
+        "on the interface of the structure, including exterior boundary and "
+        "intersection interfaces with other structures.",
+    )
+
     @cached_property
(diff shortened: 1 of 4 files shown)


tylerflex (approved):

tidy3d/components/medium.py
```diff
@@ -2775,6 +2788,12 @@ class CustomAnisotropicMedium(AbstractCustomMedium, AnisotropicMedium):
         description="This field is ignored. Please set ``allow_gain`` in each component",
     )
 
+    subpixel: bool = pd.Field(
```
lucas-flexcompute, line 2791: Just a suggestion if you agree: add a validator here that guarantees this value is set to `None`, so that the user is reminded that this field is ignored in case they forget and use it.

lucas-flexcompute (approved):

tylerflex:
would be good to add a changelog item too

tidy3d/components/medium.py
lucas-flexcompute: Yeah, a warning should be fine. No reason to error out.

weiliangjin2021:
> would be good to add a changelog item too

Added now. The subpixel field description has also been updated to reflect that subpixel will be applied not only on the exterior boundary, but also intersection interfaces with other structures.

PR #969, opened 2023-06-23
factor out store_vjp sequential and parallel, fix num_procs passing

tylerflex:
A few changes:

1. made NUM_PROCS_LOCAL and NUM_PROCS_SERVER for toggling these. Added into tests and default kwargs where appropriate. I think before the tests might have not been triggering the parallel vjp, at least for JaxPolySlab (not sure now) so I did some work to ensure that all pathways were tested depending on whether run_local or run were used in the tests.
2. refactored so that `store_vjp` generally toggles between a `store_local_parallel` and a `store_local_sequential` based on `num_procs`, seemed slightly cleaner (but more verbose) to handle it this way.
3. A bit more refactoring (eg separate function for making args in the jax polyslab VJP).

Let me know what you think and feel free to take any of these changes as you feel necessary.

### DIFF
--- a/tests/test_plugins/test_adjoint.py
+++ b/tests/test_plugins/test_adjoint.py
@@ -39,6 +39,7 @@ from ..utils import SIM_DATA_PATH, SIM_FULL, TMP_DIR
 FWD_SIM_DATA_FILE = TMP_DIR + "adjoint_grad_data_fwd.hdf5"
 SIM_VJP_FILE = TMP_DIR + "adjoint_sim_vjp_file.hdf5"
 RUN_PATH = TMP_DIR + "simulation.hdf5"
+NUM_PROC_PARALLEL = 2
 
 EPS = 2.0
 SIZE = (1.0, 2.0, 3.0)
@@ -95,7 +96,7 @@ def run_emulated_bwd(
     folder_name: str,
     callback_url: str,
     verbose: bool,
-    num_proc: int = 2,
+    num_proc: int = NUM_PROC_PARALLEL,
 ) -> JaxSimulation:
     """Runs adjoint simulation on our servers, grabs the gradient data from fwd for processing."""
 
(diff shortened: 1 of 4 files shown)


tidy3d/plugins/adjoint/components/simulation.py
```diff
@@ -25,6 +25,15 @@
 from .geometry import JaxPolySlab, JaxGeometryGroup
 
 
+# bandwidth of adjoint source in units of freq0 if no sources and no `fwidth_adjoint` specified
+FWIDTH_FACTOR = 1.0 / 10
+
+
+# how many processors to use for server and client side adjoint
+NUM_PROC_LOCAL = 1
+NUM_PROC_SERVER = 2
```
momchil-flex, line 34: `NUM_PROC_SERVER` is only used in test_adjoint so no need to define it here?

tests/test_plugins/test_adjoint.py
```diff
@@ -95,7 +95,7 @@ def run_emulated_bwd(
     folder_name: str,
     callback_url: str,
     verbose: bool,
-    num_proc: int = 2,
+    num_proc: int = NUM_PROC_SERVER,
```
momchil-flex, line 98: I had put `num_proc = 2` here and `num_proc = 1` in the `_async` test below so that both parallel and sequential gradient computation are tested (somewhat, as the async tests are not as full). I guess a better way would be to parametrize over one more argument, but do we want to do that?

momchil-flex (approved):
I'm surprised you hadn't also split `JaxGeometry.store_vjp` into `sequential` and `parallel`. Did you not see that part or was there a reason?

To me whatever you prefer for the future works.

tidy3d/plugins/adjoint/components/simulation.py
tylerflex: My thought was that the server side code could use it too to set default values

tests/test_plugins/test_adjoint.py
tylerflex: The paramerization over a single run local of true and false already hits both conditions. So this change just means that both conditions are tested for Async too

momchil-flex: Ohh right.

PR #970, opened 2023-06-23
restrict number of adjoint input structures

tylerflex:
feel free to change the number, as long as the Metalens Demo supports it.  I think it's currently 18x18 = 324 for the symmetry expanded one and under 100 without symmetry. However, we construct the symmetry expanded one for plotting at the end.. (even though we dont run it). Not really sure how to approach this. maybe a pre-upload validator?  Note, if we go this route, we'll need to somehow ensure that the Simulation.validate_pre_upload() is still called.


momchil-flex:
I'll test a bit when everything else is merged in and parallelization works, I'd think we can live with 1000.

momchil-flex:
Yeah no it seems like we'd have to keep it lower for now. I'm thinking 400 so your notebook still works.

PR #971, opened 2023-06-23
remove restriction on 3D pixelated grids in JaxCustomMedium

tylerflex:
Figured we might want to relax this for a potential user @xin-flex. Not sure what your thoughts are @momchil-flex or if you want to also be more strict on the total number of pixels etc. but figured I'd open the PR just to get this started.


tylerflex:
dont remember.

PR #975, opened 2023-06-27
SCEM-2949,SCEM-2969:Python client DE

magiWei:


tidy3d/web/cli/app.py
```diff
@@ -89,8 +90,13 @@ def auth(req):
         current_apikey = get_description()
         message = f"Current API key: [{current_apikey}]\n" if current_apikey else ""
         apikey = click.prompt(f"{message}Please enter your api key", type=str)
+    try:
+        resp = requests.get(
+            f"{Env.current.web_api_endpoint}/apikey", auth=auth, verify=Env.current.ssl_verify
+        )
+    except (requests.exceptions.SSLError, ssl.SSLError):
+        resp = requests.get(f"{Env.current.web_api_endpoint}/apikey", auth=auth, verify=False)
```
momchil-flex, line 98: Can this still be an issue @lei-flex ? It sounds to me like if verify fails it just goes ahead without verification anyway, which doesn't sound more secure.

momchil-flex (commented):
If this is not urgent I'd prefer leaving it for @tylerflex to test next week. If you want to get the ssl verify fix earlier we could split it from the free simulation stuff. Does the latter only work on dev env now?

magiWei:
locally running python lint.py is: 
INFO:root:PyLint Passed | Score: 10.0 | Threshold: 10.0 
,but test with tox throw out error and the code is not my push. @momchil-flex

tidy3d/web/cli/app.py
lei-flex: that's our current behavior of other webapi invoke.  let's just keep it as same as other right now. 
the long story is that we hit very painful issue at beginning on  a few of customers.  they have local SSL issues and can't connect to our services with the normal way.
so we provided the auto hack way to release user's pain part. 
When we set the verify is false, we already have the warning message to show on web. so I think it is fine.
if we actually want to block any unverifed call, let's just create another ticket to block by user's config.

magiWei:
> If this is not urgent I'd prefer leaving it for @tylerflex to test next week. If you want to get the ssl verify fix earlier we could split it from the free simulation stuff. Does the latter only work on dev env now?
> 
> Also @magiWei note that the PR is failing pylint. You can test that locally by running `python lint.py` in the root repo folder.

locally running `python lint.py` is ok, the msg:
(venv)@192 tidy3d % python lint.py
INFO:root:PyLint Starting | Path: ./tidy3d | Threshold: 10.0 
--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
INFO:root:PyLint Passed | Score: 10.0 | Threshold: 10.0 


I don't know why `Test with tox` throw out error and the code is not my push.

magiWei:
locally running python lint.py is ok, the msg:
(venv)https://github.com/192 tidy3d % python lint.py
INFO:root:PyLint Starting | Path: ./tidy3d | Threshold: 10.0
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
INFO:root:PyLint Passed | Score: 10.0 | Threshold: 10.0

I don't know why Test with tox throw out error and the code is not my push.

momchil-flex:
I put only the ssl verify part of this PR in the pre/2.3 branch as we're about to release this. For the rest, apart from the pylint errors there is also some different indentation that I would like to avoid. @magiWei could you try running black and pylint using the exact versions as defined in the requirements file?

magiWei:
I have fixed the Test error caused by indentation, and the DE means to tell the users whether the simulation currently running is free or charged, and display the corresponding information on the console. @momchil-flex According to Leixin's @xin-flex  expectations, she hopes that the DE function will be released in pre2.3.

tylerflex:
Do you still need my feedback / testing on this? What does DE stand for? Only comments I have right now are:
1. Note that if the fields in `TaskBlockInfo` change in the future, `estimate_cost()` will error for past versions. So we just need to be very careful to handle backwards compatibility.
2. putting the string formatting logic (appending "k", "M", etc) inside of the estimate_cost function makes it pretty hard to follow. I would have preferred this was factored out (and also the logic reused between time steps and grid steps)

magiWei:
DE means to tell the users whether the simulation currently running is free or charged, and display the corresponding information on the console. 

k, M is the unit, so that users can better perceive the size of the value.
@tylerflex

tylerflex:
> DE means to tell the users whether the simulation currently running is free or charged, and display the corresponding information on the console.
Yea I was more curious about what words "DE" stands for?

> k, M is the unit, so that users can better perceive the size of the value.
Yea, my problem was that the formatting of the time steps / grid cells using k, M seems logically separate from `estimate_cost()` and so it makes the function long and hard to follow. Also it's copied and pasted from the time step formatting to the grid step formatting. So I would have probably suggested a separate function `def format_number(num: int) -> str:` called inside of `estimate_cost()` that separated this logic a bit and was called on the grid cells and the time steps separately.

magiWei:
> 
Good idea. we will def one new function to do the logic on the next release.

"DE"： Desk To Experience

PR #977, opened 2023-06-27
Added fast dispersion fitter

caseyflex:
# Features:

- Passivity enforcement and lossless materials via lower and upper bounds on loss
- Weighting real / imaginary part
- Fitting of eps_inf

# Known limitations:

- Running with eps_inf=None can give worse results than running with fixed eps_inf
- Works quite well on every material in library except the following two:
    - Pd-JohnsonChristy1972, Ni-JohnsonChristy1972

### DIFF
--- a/tidy3d/plugins/dispersion/fit_fast.py
+++ b/tidy3d/plugins/dispersion/fit_fast.py
@@ -1,0 +1,858 @@
+"""Fit PoleResidue Dispersion models to optical NK data"""
+
+from __future__ import annotations
+from typing import Tuple, Optional
+
+import numpy as np
+from rich.progress import Progress
+from pydantic import Field, validator, PositiveInt, NonNegativeFloat, PositiveFloat
+import scipy
+
+from .fit import DispersionFitter
+from ...log import log
+from ...components.base import Tidy3dBaseModel, cached_property
+from ...components.medium import PoleResidue
+from ...components.types import ArrayFloat1D, ArrayComplex1D, ArrayFloat2D, ArrayComplex2D
+from ...constants import HBAR
+from ...exceptions import ValidationError
+
+# numerical tolerance for pole relocation for fast fitter
+TOL = 1e-8
+# numerical cutoff for passivity testing
+CUTOFF = np.finfo(np.float32).eps
+# range for checking and enforcing passivity, in addition to extrema method
+PASSIVITY_MIN = -10
+PASSIVITY_MAX = 4
+PASSIVITY_NUM = 1000
+# parameters for passivity optimization
+PASSIVITY_NUM_ITERS_DEFAULT = 50
+SLSQP_CONSTRAINT_SCALE_DEFAULT = 1e35
+# min value of the rms for default weights calculated based on Re[eps], Im[eps]
+RMS_MIN = 0.1
+
+DEFAULT_MAX_POLES = 5
+DEFAULT_NUM_ITERS = 20
+DEFAULT_TOLERANCE_RMS = 1e-5
+
+# this avoids divide by zero errors with lossless poles
+SCALE_FACTOR = 1.01
+
+
+class AdvancedFastFitterParam(Tidy3dBaseModel):
+    """Advanced fast fitter parameters."""
+
+    loss_bounds: Tuple[float, float] = Field(
+        (0, np.inf),
+        title="Loss bounds",
+        description="Bounds (lower, upper) on Im[eps]. Default corresponds to only passivity. "
+        "A lower bound of 0 or greater ensures passivity. To fit a gain medium without "
+        "additional constraints, use ``loss_bounds=(-np.inf, np.inf)``. "
+        "Increasing the lower bound could help with simulation stability. "
+        "A finite upper bound may be helpful when fitting lossless materials. "
+        "In this case, consider also increasing the weight for fitting the imaginary part.",
+    )
+    weights: Tuple[NonNegativeFloat, NonNegativeFloat] = Field(
+        None,
+        title="Weights",
+        description="Weights (real, imag) in objective function for fitting. The weights "
+        "are applied to the real and imaginary parts of the permittivity epsilon. The weights "
+        "will be rescaled together so they average to 1. If ``None``, the weights are calculated "
+        "according to the typical value of the real and imaginary part, so that the relative error "
+        "in the real and imaginary part of the fit should be comparable. "
+        "More precisely, the RMS value ``rms`` of the real and imaginary parts are "
+        "calculated, and the default weights are 1 / max(``rms``, ``RMS_MIN``). "
+        "Changing this can be helpful if fitting either the real or imaginary part is "
+        "more important than the other.",
+    )
+    show_progress: bool = Field(
+        True,
+        title="Show progress bar",
+        description="Whether to show progress bar during fitter run.",
+    )
+    show_unweighted_rms: bool = Field(
+        False,
+        title="Show unweighted RMS",
+        description="Whether to show unweighted RMS error in addition to the default weighted "
+        'RMS error. Requires ``td.config.logging_level = "INFO"``.',
+    )
+    relaxed: Optional[bool] = Field(
+        None,
+        title="Relaxed",
+        description="Whether to use relaxed fitting algorithm, which "
+        "has better pole relocation properties. If ``None``, will try both original and relaxed "
+        "algorithms.",
+    )
+    smooth: Optional[bool] = Field(
+        None,
+        title="Smooth",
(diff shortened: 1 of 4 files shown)


tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -0,0 +1,668 @@
...
+            return val * np.ones(len(wvl))
+        if not len(val) == len(wvl):
+            raise ValidationError("The min_loss must be same length as min_loss_wvl.")
+        return val
+
+    @cached_property
+    def _min_loss_omegas(self) -> ArrayFloat1D:
+        return 2 * np.pi * HBAR * C_0 / np.array(self.min_loss_wvl[::-1])
+
+    def fit_fast(  # pylint: disable=too-many-locals
```
lucas-flexcompute, line 229: Because it inherits from DispersionFitter, calling `FastDispersionFitter.fit` actually uses the base class. Isn't it better to override `fit` instead of creating `fit_fast`? It can be confusing as it is right now.

dbochkov-flexcompute:
some random feedback, not sure if it is already addressed: I happened to play with different fittings recently, and it seemed like `fit_fast` ignores `wvl_range` parameter, i.e. it always tries to fit all provided data?

caseyflex:
> some random feedback, not sure if it is already addressed: I happened to play with different fittings recently, and it seemed like `fit_fast` ignores `wvl_range` parameter, i.e. it always tries to fit all provided data?

thanks, should be fixed now

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -0,0 +1,803 @@
...
+            Approximated by Rational Functions," IEEE Trans. Power
+            Syst. 16, 1 (2001).
+
+        Parameters
+        ----------
+        num_poles_range : Tuple[PositiveInt, PositiveInt], optional
+            Range (min, max) for number of poles in the model.
+        eps_inf : float, optional
+            Value of eps_inf to use in fit. If None, then eps_inf is also fit.
+            Note:
```
weiliangjin2021, line 673: something to go after "Note:"?

weiliangjin2021 (commented):
Looks quite good! It now works well with several metals in my tests. Since we start to support gain medium recently, to fit a gain medium, is it by either setting `loss_bounds=(-np.inf, np.inf)` or `passivity_num_iters=0`? As for the latter, it is currently a PositiveInt, so might be good to change it to an int?

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -0,0 +1,803 @@
...
+            model = model.enforce_passivity()
+            # log.info(
+            #    f"Trying to enforce passivity. new_rms={model.rms_error}, "
+            #    f"loss_in_bounds={model.loss_in_bounds}."
+            # )
+        return model
+
+    def fit(  # pylint: disable=arguments-renamed, too-many-locals
+        self,
+        num_poles_range: Tuple[PositiveInt, PositiveInt] = (1, DEFAULT_MAX_POLES),
```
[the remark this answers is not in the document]
caseyflex: good idea!

caseyflex:
> Looks quite good! It now works well with several metals in my tests. Since we start to support gain medium recently, to fit a gain medium, is it by either setting `loss_bounds=(-np.inf, np.inf)` or `passivity_num_iters=0`? As for the latter, it is currently a PositiveInt, so might be good to change it to an int?

Thanks for the comments. It's by setting loss_bounds. I improved the docstrings to make this more clear.

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -0,0 +1,834 @@
...
+        title="Number of iterations",
+        description="Number of iterations of the fitting algorithm. Make this smaller to "
+        "speed up fitter, or make it larger to try to improve fit.",
+    )
+    passivity_num_iters: PositiveInt = Field(
+        PASSIVITY_NUM_ITERS_DEFAULT,
+        title="Number of loss bounds enforcement iterations",
+        description="Number of loss bounds enforcement iterations of the fitting algorithm. "
+        "Make this smaller to speed up fitter. There will be a warning if this value "
+        "is too small. To fit a gain medium, use the loss_bounds parameter instead.",
```
tylerflex, line 102: #can you put double backslash quotes around the variable names in these docstrings? Also for `None`, `True`, `False` instances in the strings.

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -0,0 +1,834 @@
...
+        """Evaluate model at sample frequencies."""
+        return self.evaluate(self.omega)
+
+    @cached_property
+    def imag_extrema(self) -> Tuple[ArrayFloat1D, ArrayComplex1D]:
+        """Extrema of imaginary part of eps."""
+
+        def _extrema_loss_freq_finder(
+            areal, aimag, creal, cimag
+        ):  # pylint: disable=too-many-locals
```
tylerflex, line 294: just FYI I think it should work to put the disable just above the function signature and might look cleaner.

```py
    # pylint: disable=too-many-locals
    def _extrema_loss_freq_finder(areal, aimag, creal, cimag):
```

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -0,0 +1,834 @@
...
+                    configs = [c + [r] for c in configs for r in [setting]]
+            return configs
+
+        best_model = init_model
+        warned_about_passivity_num_iters = False
+        warned_about_slsqp_constraint_scale = False
+
+        configs = make_configs()
+
+        with Progress() as progress:
```
tylerflex, line 743: we might only want progress bar if verbose, is it possible to set it up this way?

tylerflex (changes requested):
Looks good to me, I don't understand many of the internals so most of my comments are pretty surface level. More generally, are we considering adding more fitters in the future? If so, it would probably be nice to have a more unified interface in terms of parameters, the general `.fit()` method (what is displayed, how progress bars work), and the testing. Thanks Casey!

tylerflex:
Also, what I've been trying out is marking PRs as "draft" before they are ready for review and then marking "ready for review" when it is ready to review and merge.
image

lucas-flexcompute (commented):
I don't have much to add from my previous comments or Tyler's points. One thing I noticed is that `max_num_poles` and `min_num_poles` seem very similar in purpose, but are accessible in different parts of the API. Is there any deeper reason for this choice (maybe compatibility with other fitters)?

Other than that, I believe we talked about making this fitter the easy default for the user. Is this change in API supposed to be included in this PR? In that case, it should probably be followed by changes in the docs.

caseyflex:
> Looks good to me, I don't understand many of the internals so most of my comments are pretty surface level. More generally, are we considering adding more fitters in the future? If so, it would probably be nice to have a more unified interface in terms of parameters, the general `.fit()` method (what is displayed, how progress bars work), and the testing. Thanks Casey!

Thanks for the comments. I don't think we are considering adding more fitters in the future...

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -0,0 +1,836 @@
...
+        "In this case, consider also increasing the weight for fitting the imaginary part.",
+    )
+    weights: Tuple[NonNegativeFloat, NonNegativeFloat] = Field(
+        None,
+        title="Weights",
+        description="Weights (real, imag) in objective function for fitting. The weights "
+        "are applied to the real and imaginary parts of the permittivity epsilon. The weights "
+        "will be rescaled together so they average to 1. If ``None``, the weights are calculated "
+        "according to the typical value of the real and imaginary part, so that they are both "
+        "fit well. Changing this can be helpful if fitting either the real or imaginary part is "
```
weiliangjin2021, line 61: I think it would be better to explain a bit more on how the weights are computed, or what it mean by "they are both fit well".

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -0,0 +1,836 @@
...
+        "more important than the other.",
+    )
+    verbose: bool = Field(
+        True, title="Verbose", description="Whether to show progress bar during fitter run."
+    )
+    debug: bool = Field(
```
weiliangjin2021, line 67: use tidy3d's log.debug instead so that it's unified?

tylerflex (approved):

tylerflex:
@weiliangjin2021 @lucas-flexcompute if you both approve of the PR can you formally approve it through GitHub and then I'll merge it, thanks!

lucas-flexcompute (approved):

weiliangjin2021 (approved):

Issue #839, opened 2023-04-14
SimulationData.get_poynting_vector not using grid corrections for 2D datasets


lucas-flexcompute:
[`SimulationData.get_poyting_vector`](https://github.com/flexcompute/tidy3d/blob/b738ca62bde904fc0e3355ad73ea34d15b7d726f/tidy3d/components/data/sim_data.py#L234) calculates the 3D Poynting vector (time averaged or instantaneous) from the field components at the grid centers. However, for 2D monitors, this path does not include the grid corrections used in [`ElectromagneticFieldData.poynting`](https://github.com/flexcompute/tidy3d/blob/5816725fa2ee455259066c2eb7cc2c3bb49862aa/tidy3d/components/data/monitor_data.py#L357) or [`FieldTimeData.poynting`](https://github.com/flexcompute/tidy3d/blob/5816725fa2ee455259066c2eb7cc2c3bb49862aa/tidy3d/components/data/monitor_data.py#L662).

We can include a test to see if the monitor requested in `get_poynting_vector` is plane and use [`ElectromagneticFieldData._centered_fields`](https://github.com/flexcompute/tidy3d/blob/5816725fa2ee455259066c2eb7cc2c3bb49862aa/tidy3d/components/data/monitor_data.py#L311) to get the components, but that, to me, looks like a hack. It might be better to properly implement a `poynting` property for all monitor types that do the right thing.

I'm open to suggestions before starting to work on a possible fix.

PR #983, opened 2023-07-03
Grid correction to Poynting  in 2D monitor (#839)

lucas-flexcompute:
To add up to the thoughts in #839: considering the grid points where the calculation is performed comes from `SimulationData`, creating a `poyting` method directly in `MonitorData` would end up more convoluted than just exposing a method to apply the correction from the 2D monitor. That's the reson for the current implementation.

Just to make sure: @momchil-flex, the correction should be applied in the case of a 2D monitor no matter the grid points, right? Or am I misunderstanding the correction factor?

### DIFF
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -216,8 +216,23 @@ class SimulationData(Tidy3dBaseModel):
             interpolated to center locations on Yee grid.
         """
 
-        # get the data
-        monitor_data = self.load_field_monitor(field_monitor_name)
+        return self._at_centers(self.load_field_monitor(field_monitor_name))
+
+    def _at_centers(self, monitor_data: xr.Dataset) -> xr.Dataset:
+        """return xarray.Dataset representation of field monitor data
+        co-located at Yee cell centers.
+
+        Parameters
+        ----------
+        monitor_data : xr.Dataset
+            Monitor data to be co-located.
+
+        Returns
+        -------
+        xarray.Dataset
+            Dataset containing all of the fields in the data
+            interpolated to center locations on Yee grid.
+        """
 
         # discretize the monitor and get center locations
         sub_grid = self.simulation.discretize(monitor_data.monitor, extend=False)
@@ -233,6 +248,7 @@ class SimulationData(Tidy3dBaseModel):
 
         return monitor_data.colocate(**xyz_kwargs)
 
+    # pylint: disable=too-many-locals
     def get_poynting_vector(self, field_monitor_name: str) -> xr.Dataset:
         """return ``xarray.Dataset`` of the Poynting vector at Yee cell centers.
 
@@ -253,8 +269,9 @@ class SimulationData(Tidy3dBaseModel):
             DataArray containing the Poynting vector calculated based on the field components
             colocated at the center locations of the Yee grid.
         """
-
-        field_dataset = self.at_centers(field_monitor_name)
+        # Fields from 2D monitors need a correction factor
+        mon_data = self.load_field_monitor(field_monitor_name).grid_corrected_copy
+        field_dataset = self._at_centers(mon_data)
 
         time_domain = isinstance(self.monitor_data[field_monitor_name], FieldTimeData)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Source validation happens before simulation upload and raises an error if no source is present.
 - Warning instead of error if structure out of simulation bounds.
+- Cleaner display of `ArrayLike` in docs.
 
 ### Fixed
 - Point-like objects correctly appear as single points using `plt.scatter`.
-### Changed
-- Source validation happens before simulation upload and raises an error if no source is present.
-- Cleaner display of `ArrayLike` in docs.
+- Apply finite grid correction to the fields when calculating the Poynting vector from 2D monitors.
 
 ## [2.3.0] - 2023-6-30
 
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -197,20 +197,16 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
         return tangential_dims
 
     @property
-    def _grid_corrected_fields(self) -> Dict[str, DataArray]:
+    def _in_plane(self) -> Dict[str, DataArray]:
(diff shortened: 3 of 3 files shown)


momchil-flex:
> To add up to the thoughts in #839: considering the grid points where the calculation is performed comes from `SimulationData`, creating a `poyting` method directly in `MonitorData` would end up more convoluted than just exposing a method to apply the correction from the 2D monitor. That's the reson for the current implementation.
> 
> Just to make sure: @momchil-flex, the correction should be applied in the case of a 2D monitor no matter the grid points, right? Or am I misunderstanding the correction factor?

I am not sure if I understand the question exactly, but the grid correction factor depends on the specific monitor location and the specific grid used in the simulation. So even though `MonitorData` doesn't really "know" anything about the grid that was used in the normal direction, this grid correction factor carries the needed information to normalize things correctly. Here's how it arises. In `FluxMonitor`-s, to get the flux at a given location x0, we compute the flux at the two nearest cell centers x1 and x2, and then interpolate that to x0. We had some discussions where that seemed to be the correct approach (producing values that best imitate a continuous real space). Field monitors on the other hand snap each field to the x0 location. This then leads to a slightly different flux if computed afterwards. Basically the difference is `flux1 = interpolate_x0(flux(interpolate_centers(fields)))` vs `flux2 = flux(interpolate_x0(fields)))`. To make this consistent, the grid correction factor is just `sqrt(flux1 / flux2)`.

This is somewhat related to the monitor refactor I've been working on, where I have for example changed `at_centers` to use a grid provided by the `monitor_data` rather than taken to use `sim_data.simulation` (and I have also introduced `sim_data.at_boundaries`, which substitutes the typical `at_centers` use). I'll leave your PR open for now but see how this change can enter my upcoming changes.

momchil-flex:
It seems to me that within the current code, a more straightforward fix may be to just `return self` [here](https://github.com/flexcompute/tidy3d/blob/6291b4dfefe2538f7fb14270a69f9d92356e2617/tidy3d/components/data/monitor_data.py#L204) instead of raising an error, and throw in a `_grid_corrected_fields` [here](https://github.com/flexcompute/tidy3d/blob/6291b4dfefe2538f7fb14270a69f9d92356e2617/tidy3d/components/data/sim_data.py#L94)? Well with the added detail that currently `_grid_corrected_fields` returns a Dict not a FieldData. What do you think though?

lucas-flexcompute:
> It seems to me that within the current code, a more straightforward fix may be to just `return self` [here](https://github.com/flexcompute/tidy3d/blob/6291b4dfefe2538f7fb14270a69f9d92356e2617/tidy3d/components/data/monitor_data.py#L204) instead of raising an error, and throw in a `_grid_corrected_fields` [here](https://github.com/flexcompute/tidy3d/blob/6291b4dfefe2538f7fb14270a69f9d92356e2617/tidy3d/components/data/sim_data.py#L94)? Well with the added detail that currently `_grid_corrected_fields` returns a Dict not a FieldData. What do you think though?

Indeed looks like a cleaner solution that would apply the correction automatically. You understand how `SimulationData.__get_item__` is used better than I do: if there are no places where the fields are expected *not* to be corrected, great! And in that case, is there any reason not to create the data with the fields already corrected?

momchil-flex:
Ok yeah good point, it seems to me that there is no straightforward answer to what's best. Here are several options. All of them have exactly the same goal: making `FieldData.flux` the same as `FluxData` for the same size monitor, which has generally been set in a way as to hide numerical artifacts due to the finite grid as much as possible (similarly for other quantities like poynting and field1.dot(field2)). But here are various options about where exactly to apply the finite grid correction, each with its own drawback / potential source of confusion:

1. Current code: the grid correction is only applied to the fields internally e.g. when computing poynting, flux, and dot.
Drawback: If a user tries to manually compute the flux from E and H without applying the correction, they'll get a different value.

2. Code in the current PR: grid correction is applied when calling `sim_data[monitor.name]`.
Pro: fixes the drawback of 1 if user loads the field data in that way.
Drawbacks: `sim_data[monitor.name]` will differ in magnitude compared to just loading the data directly as a MonitorData, or reading the `hdf5` file directly, which may be confusing. Similarly, a 3D `FieldMonitor` that's interpolated to the same location as a 2D monitor will have different fields (the correction will be applied to the 2D monitor, but not to the 3D monitor).

3. Another option: apply grid correction already on the backend
Pro: fixes the drawback of 1 as well as the first drawback of 2.
Drawback: the inconsistency between a 3D monitor that's interp-ed afterwards, and a 2D monitors remains. I guess this could be considered a part of the interpolation, but I'm not sure if we should be modifying the raw data automatically like that.

Note that in all cases the correction is just a constant pre-factor to all the field components. It seems that probably we should do either 1. or 3 (it should be easy to reframe the current PR into 1, while 3 will require a backend change too). I would probably still choose 1 over 3, but interested in your thoughts @lucas-flexcompute @tylerflex .

lucas-flexcompute:
I believe option 1 is the one implemented by the first commit, right? I also tend to prefer it over 2 and 3. I'll wit for @tylerflex's input before reverting, though.

momchil-flex:
Yeah the first commit just replicates the behavior on the develop branch but with the fix for `get_poynting`, so it's option 1.

momchil-flex:
Sorry, seems like there's a bit of confusion here. Even if we go with 1, I didn't mean that you should revert the second commit: I thought it was cleaner, but just don't add the `grid_corrected_copy` call to `getitem`, but only call it in `get_poynting`, like in the first commit. Or is there some reason this doesn't work?

The other thing I was saying is that this also falls under the scope of the monitor refactor that I'm doing, which should hopefully get in 2.4. But I think if you can do the modification as suggested above, this should not conflict with my changes, and can indeed be put into develop directly to fix this bug in the upcoming 2.3 patch.

tidy3d/components/data/sim_data.py
```diff
@@ -253,8 +269,12 @@ def get_poynting_vector(self, field_monitor_name: str) -> xr.Dataset:
             DataArray containing the Poynting vector calculated based on the field components
             colocated at the center locations of the Yee grid.
         """
+        # Fields from 2D monitors need a correction factor
+        mon_data = self.load_field_monitor(field_monitor_name)
+        if len(mon_data.monitor.zero_dims) == 1:
```
momchil-flex, line 274: the check shouldn't be needed now?

momchil-flex (approved):
Oh yeah I see the issue. Yeah I think this looks the best now!

tidy3d/components/data/sim_data.py
lucas-flexcompute: True, I forgo we fixed that!

momchil-flex:
Looks good, could you squash commits (and fix changelog conflict) and merge? Thanks!

Issue #978, opened 2023-06-28
Make warnings more noticeable or more strongly enforced


momchil-flex:
**Is your feature request related to a problem? Please describe.**
We keep hitting issues like diverged simulations that could have been avoided if users would have heeded some of our automated warnings.

**Describe the solution you'd like**
It would be nice if we can make the warnings "stronger". One option is to have e.g. `allow_warning=False` flag in the `upload` or `run` function, and have it error if the simulation produces any warnings and it is not explicitly set to `True`. Would be useful if we can store a list of warnings in the simulation so we don't have to re-validate it.

**Describe alternatives you've considered**
Other options include asking the user interactively if they want to run it, but @xin-flex pointed out that we used to have issues with prompts  when we were using them for user name / password on different platforms and IDEs. Another option is to just make our warnings more visible by e.g. using the python `warnings` package (they're highlighed in red, an won't be turned off by our logging config level...) Or do this for the important ones at least?

lucas-flexcompute:
I'm not sure I like the idea of throwing warnings a second time if the user already ignored it. I'm more inclined to use the warnings module and maybe issue the warning in the logger automatically, so, whenever we use `log.warning` we call `warnings.warn`. It can be added [here](https://github.com/flexcompute/tidy3d/blob/pre/2.3/tidy3d/log.py#L225) to comply with the current logger functionality. As it stands, warnings would be shown twice, but I guess in this case we can change our default logging level to "ERROR".

That said, whatever the solution, maybe we should think i little more carefully on the problem: the user ignoring the warning (that is already there). Maybe we're issuing too many (unnecessary) warnings and the user becomes used to ignoring them (I can tell this is my personal experience at least). For example, the warning about no sources in the simulation is one that constantly makes me turn off all warnings. What I'm saying is that we should turn more warnings into INFO, if they are not really dangerous, and issue warnings when they really matter (simulation divergence = possible loss of many credits). We could also have a separate category: CRITICAL_WARNING between WARNING and ERROR, which could be configured to optionally raise an error (true by default).

One advantage we already have in 2.3 is the new log filtering, which should at least decrease the amount of warnings printed at once.

momchil-flex:
I think all of those are good points, we could improve significantly before forcing the user to have to explicitly allow a simulation with warnings to run. Would be great to get some of those in 2.4.

lucas-flexcompute:
[...]

PR #984, opened 2023-07-03
Style warnings to draw user attention (#978)

lucas-flexcompute:


tylerflex (approved):

tylerflex:
I also just noticed that the warnings show up with red background in the notebook, which could be a good alternative
image

lucas-flexcompute:
To be compatible with terminal color schemes (for legibility) we have to use only the first 16 color from the [color list](https://rich.readthedocs.io/en/stable/appendix/colors.html#appendix-colors). Testing the different combinations, it looked to me that red text was the best option. What do you think?

tylerflex:
Ok that's probably fine. let's try it

Issue #813, opened 2023-03-30
Better handling for "no sources" validator in Simulation


tylerflex:
Right now, we get this validator twice when constructing a sim for the mode solver and plotting the mode solver results.

It's unnecessary to print this warning in these cases as we really only want to warn the user if they are running the simulation.

We could add as a pre-upload validator (which errors) but we'd still want the user to be able to submit a simulation without sources if (for some reason) they want to view ModeMonitorData or PermittivityData results.

Not sure what the best approach is, but I might even suggest moving it from WARNING to INFO for now as a short term solution?

tylerflex:
Referenced in #978, possibly duplicate and we can close.

lucas-flexcompute:
I'm not so sure our solution for #978 will help much here. I'm proposing another solution specifically for this, just in case.

tylerflex:
perhaps we can move this to a "pre-upload" validator (and make it error) that way it doesn't get triggered for mode stuff

PR #985, opened 2023-07-03
Change missing source from warning to info (#813)

lucas-flexcompute:
Also included a flag in the pre upload validator to optionally raise an error if no sources are present in the simulation. This flag is true by default on `run` and manual job creation, but false for the mode solver. It cannot be changed from `run` (only in `Job` or `Batch`), because simulations with no sources should be very rare and we already have a lot of arguments in `run`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+# Changed
+- Source validation happens before simulation upload and raises an error if no source is present.
+
 ## [2.3.0] - 2023-6-30
 
 ### Added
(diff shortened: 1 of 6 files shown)


tylerflex (commented):
Looks good to me. only suggestion is maybe we dont need to add this to containers or web.upload basically there should never be a time that we'd want to upload without sources?

lucas-flexcompute:
In the original issue you mentioned the user might upload only to view ModeMonitorData or PermittivityData results. Is that not the case any longer?

tylerflex:
Hm, well I think that might have been something I was suggesting as a workaround to these not having their own separate web services, but my impression was that @momchil-flex didn't want to encourage this because it wastes our resources correct? so yea I think in this case it's probably good how you have it set up now.

tylerflex:
I'm still not sure we need it in containers because it seems unlikely that simulations in the containers would need to change the default behavior.

tylerflex (approved):

tylerflex:
I forgot, can we also add a "changed" item in changelog? just saying that we no longer will allow upload of simulation without sources or something?

PR #988, opened 2023-07-06
warn instead of error for structures out of sim bounds

### LINKED ISSUE #885: Change "structure completely outside of simulation domain" from error to warning
If a Structure is totally outside of the simulation domain, it will be considered as an error. It causes inconvenience especially when doing gds import where the simulation is only done on part of the layout. We can consider making it a warning and if a structure is totally outside of the domain, it will be ignored.

tylerflex:
fixes #885

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Source validation happens before simulation upload and raises an error if no source is present.
+- Warning instead of error if structure out of simulation bounds.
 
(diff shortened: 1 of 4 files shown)


tylerflex:
The way I've written this, it should be very simple to change the behavior for sources and monitors later. I added an `error` option to the validator so right now it looks like
```py
_structures_in_bounds = assert_objects_in_sim_bounds("structures", error=False)
_sources_in_bounds = assert_objects_in_sim_bounds("sources")
_monitors_in_bounds = assert_objects_in_sim_bounds("monitors")
```
but we can just add `error=False` to the other ones.

PR #990, opened 2023-07-06
internal refactor of smatrix tool

### LINKED ISSUE #717: Refactor `smatrix`
The internals of `smatrix` tool are really clunky because of how Batch is organized (a batch uploads its jobs whenever it is initialized). We should consider refactoring it to use `run_async` instead so we can more cleanly perform the internals, otherwise it is impossible to understand what's going on currently.

To do
- [ ] refactor run to use `run_async`
- [ ] Get rid of unnecessary class methods once we can run_async instead of needing to initialize a batch.
- [ ] update element mappings and run_only to use `xr.DataArray` or a datastructure more similar to `SMatrixType`
- [ ] Fix docstrings to account for the fact that S_ij maps source j to monitor i (switched currently).
- [ ] Add utility function for element mappings of reciprocal device.

tylerflex:
fixes #717 

Note: untested in the wild, will test on notebook tomorrow.

The details are probably not important, basically I changed from using a batch that was hacked together using pydantic validators to using a more direct run_async. This let me make more sensible method signatures in the main component modeler rather than having to abuse class methods everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Source validation happens before simulation upload and raises an error if no source is present.
 - Warning instead of error if structure out of simulation bounds.
+- Internal refactor of `ComponentModeler` to simplify logic.
 
 ### Fixed
 - Point-like objects correctly appear as single points using `plt.scatter`.
-### Changed
-- Source validation happens before simulation upload and raises an error if no source is present.
 - Cleaner display of `ArrayLike` in docs.
 
 ## [2.3.0] - 2023-6-30
--- a/tidy3d/plugins/smatrix/smatrix.py
+++ b/tidy3d/plugins/smatrix/smatrix.py
@@ -1,6 +1,8 @@
 """Tools for generating an S matrix automatically from tidy3d simulation and port definitions."""
+from __future__ import annotations
 
-from typing import List, Tuple, Optional, Callable, Dict
+from typing import List, Tuple, Optional, Dict
+import os
 
 import pydantic as pd
 import numpy as np
@@ -15,14 +17,14 @@ from ...components.data.sim_data import SimulationData
 from ...components.data.data_array import DataArray
 from ...components.types import Direction, Ax, Complex
 from ...components.viz import add_ax_if_none, equal_aspect
-from ...components.base import Tidy3dBaseModel
+from ...components.base import Tidy3dBaseModel, cached_property
 from ...exceptions import SetupError, Tidy3dKeyError
-from ...web.container import Batch, BatchData
-
+from ...log import log
+from ...web.container import BatchData, Batch
 
 # fwidth of gaussian pulse in units of central frequency
 FWIDTH_FRAC = 1.0 / 10
-DEFAULT_DATA_DIR = "data"
+DEFAULT_DATA_DIR = "."
 
 
 class Port(Box):
@@ -126,12 +128,17 @@ class ComponentModeler(Tidy3dBaseModel):
         title="Verbosity",
         description="Whether the :class:`.ComponentModeler` should print status and progressbars.",
     )
-    batch: Optional[Batch] = pd.Field(
+    callback_url: str = pd.Field(
         None,
-        title="Batch",
-        description="Batch of generated simulations needed for each row of the scattering matrix."
-        "Should be left ``None`` in almost all cases, which will generate the proper batch "
-        "internally and store it as ``ComponentModeler.batch``.",
+        title="Callback URL",
+        description="Http PUT url to receive simulation finish event. "
+        "The body content is a json file with fields "
+        "``{'id', 'status', 'name', 'workUnit', 'solverVersion'}``.",
+    )
+    path_dir: str = pd.Field(
+        DEFAULT_DATA_DIR,
+        title="Directory Path",
+        description="Base directory where data and batch will be downloaded.",
     )
 
     @pd.validator("simulation", always=True)
@@ -141,94 +148,59 @@ class ComponentModeler(Tidy3dBaseModel):
             raise SetupError("'ComponentModeler.simulation' must not have any sources.")
         return val
 
-    @pd.validator("batch", always=True)
-    def _set_batch(cls, val, values):
-        """Initialize the batch if not supplied."""
-
-        # if supplied, return it
-        if val is not None:
-            return val
-
-        # otherwise, generate all sims and make a new batch
-        sim_dict = cls.make_sim_dict(values=values)
-        return Batch(
-            simulations=sim_dict,
-            folder_name=values.get("folder_name"),
-            verbose=values.get("verbose"),
-        )
-
-    @classmethod
-    def make_sim_dict(cls, values: dict) -> Dict[str, Simulation]:
+    @cached_property
+    def sim_dict(self) -> Dict[str, Simulation]:
         """Generate all the :class:`Simulation` objects for the S matrix calculation."""
 
         sim_dict = {}
-        ports = values.get("ports")
-        simulation = values.get("simulation")
-        freqs = values.get("freqs")
+        mode_monitors = [self.to_monitor(port=port) for port in self.ports]
 
-        mode_monitors = [
-            cls._to_monitor(port=port, freqs=freqs) for port in ports
-        ]  # pylint:disable=protected-access
+        for (port_name, mode_index) in self.matrix_indices_run_sim:
 
-        for (port_name, mode_index) in cls.matrix_indices_run_sim(
-            ports=ports,
-            run_only=values.get("run_only"),
-            element_mappings=values.get("element_mappings"),
-        ):
+            port = self.get_port_by_name(port_name=port_name)
 
-            port = cls.get_port_by_name(port_name=port_name, ports=ports)
+            port_source = self.shift_port(port=port)
+            mode_source = self.to_source(port=port_source, mode_index=mode_index)
 
-            port_source = cls._shift_port(
-                simulation=simulation, port=port
-            )  # pylint:disable=protected-access
-            mode_source = cls._to_source(
-                port=port_source, mode_index=mode_index, freqs=freqs
-            )  # pylint:disable=protected-access
-
-            new_mnts = list(simulation.monitors) + mode_monitors
-            sim_copy = simulation.copy(update=dict(sources=[mode_source], monitors=new_mnts))
-            task_name = cls._task_name(port=port, mode_index=mode_index)
+            new_mnts = list(self.simulation.monitors) + mode_monitors
+            sim_copy = self.simulation.copy(update=dict(sources=[mode_source], monitors=new_mnts))
+            task_name = self._task_name(port=port, mode_index=mode_index)
             sim_dict[task_name] = sim_copy
         return sim_dict
 
-    @classmethod
-    def matrix_indices_monitor(cls, ports: Tuple[Port, ...]) -> Tuple[MatrixIndex, ...]:
+    @cached_property
+    def matrix_indices_monitor(self) -> Tuple[MatrixIndex, ...]:
         """Tuple of all the possible matrix indices (port, mode_index) in the Component Modeler."""
         matrix_indices = []
-        for port in ports:
+        for port in self.ports:
             for mode_index in range(port.mode_spec.num_modes):
                 matrix_indices.append((port.name, mode_index))
         return tuple(matrix_indices)
 
-    @classmethod
-    def matrix_indices_source(
-        cls, ports: Tuple[Port, ...], run_only: Tuple[MatrixIndex, ...] = None
-    ) -> Tuple[MatrixIndex, ...]:
+    @cached_property
+    def matrix_indices_source(self) -> Tuple[MatrixIndex, ...]:
         """Tuple of all the source matrix indices (port, mode_index) in the Component Modeler."""
-        return run_only if run_only is not None else cls.matrix_indices_monitor(ports=ports)
+        if self.run_only is not None:
+            return self.run_only
+        return self.matrix_indices_monitor
 
-    @classmethod
-    def matrix_indices_run_sim(
-        cls,
-        ports: Tuple[Port, ...],
-        element_mappings: Dict[Element, Dict[Element, Callable[[complex], complex]]] = None,
-        run_only: Tuple[MatrixIndex, ...] = None,
-    ) -> Tuple[MatrixIndex, ...]:
+    @cached_property
+    def matrix_indices_run_sim(self) -> Tuple[MatrixIndex, ...]:
         """Tuple of all the source matrix indices (port, mode_index) in the Component Modeler."""
 
-        if element_mappings is None or element_mappings == {}:
-            return cls.matrix_indices_source(ports=ports, run_only=run_only)
+        if self.element_mappings is None or self.element_mappings == {}:
+            return self.matrix_indices_source
 
         # all the (i, j) pairs in `S_ij` that are tagged as covered by `element_mappings`
-        elements_determined_by_map = [element_out for (_, element_out, _) in element_mappings]
+        elements_determined_by_map = [element_out for (_, element_out, _) in self.element_mappings]
 
         # loop through rows of the full s matrix and record rows that still need running.
         source_indices_needed = []
-        for col_index in cls.matrix_indices_source(ports=ports, run_only=run_only):
+        for col_index in self.matrix_indices_source:
 
             # loop through columns and keep track of whether each element is covered by mapping.
             matrix_elements_covered = []
-            for row_index in cls.matrix_indices_monitor(ports=ports):
+            for row_index in self.matrix_indices_monitor:
                 element = (row_index, col_index)
                 element_covered_by_map = element in elements_determined_by_map
                 matrix_elements_covered.append(element_covered_by_map)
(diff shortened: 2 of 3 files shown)


lucas-flexcompute (commented):
Looks good so far. Small suggestion: changing the [task name](https://github.com/flexcompute/tidy3d/pull/990/files#diff-8cfee9637dc09ebab5e95853d64a3466280e369f9560a2b807ba239da5290e18R289) to `f"smatrix_{port.name}@{mode_index}"` to be more similar to the naming convention used in Sax (unless we're planing on using other circuit simulators).

lucas-flexcompute:
Another issue that could be included in this refactor is the default data dir. The default for `BatchData` is the current working directory, which is fine, but for the smatrix plugin is "data", which becomes a problem if the directory does not exist. The (possibly long) simulation error because it cannot created the files. Either we should try to create the directory (`pathlib.Path(path_dir).mkdir(parents=True, exist_ok=True)`) or have it default to the current working directory.

lucas-flexcompute:
Yep, I've noticed the transposition when testing other components, but I though the problem might have been the changes that I was working on, but I couldn't figure them out. So maybe something really changed in the plugin at some point?

tylerflex:
Hm, that's strange. I just ran the notebook with 
- v2.2.0 (no transpose)
- v2.3.0 (has transpose)

This is the only commit between those two versions
https://github.com/flexcompute/tidy3d/commit/47b62e9038a0a8b9f5f903e90ca7151d7f5fa035

either this commit affected how the smatrix is defined? (perhaps the port name order is different?) or maybe something changed in how mode sources and monitors are computed between 2.2 and 2.3?

tylerflex:
guess the good news is that this PR seems consistent with how it is in 2.3.0 so nothing changed from it.

tylerflex:
Well, I checked that the port name ordering seems the same between the two versions so it might not be that.

@momchil-flex to keep a long story short, it seems like between v2.2.0 and v2.3.0, the scattering matrix tool started mysteriously producing the transpose of what it was before. Do you think anything might have changed between those two versions that could account for this? eg. a normalization change in the mode source / monitor?

momchil-flex:
Ehmm I have no idea. But so is the S matrix itself transposed? What you're showing is the imaginary part of S\dagger S, which should ideally be 0.

Also, small typo (it's - > its) when looking at that notebook
image

lucas-flexcompute:
> This is the only commit between those two versions [47b62e9](https://github.com/flexcompute/tidy3d/commit/47b62e9038a0a8b9f5f903e90ca7151d7f5fa035)
> 
> either this commit affected how the smatrix is defined? (perhaps the port name order is different?) or maybe something changed in how mode sources and monitors are computed between 2.2 and 2.3?

That commit seems simple enough. The order there is maintained.

tylerflex:
Here is the output from the last cell (single column of the S matrix)
from the docs (last run before 2.3)
```
0.44006653-0.54419416j 
0.54510946+0.44524725j 
0.02118845+0.00283299j
-0.01860748+0.00561448j]
```

from develop and v2.3.0
```
output from run_only port
```
0.43816309-0.5457042j 
0.54663668+0.44330396j
-0.02051844-0.00262935j
0.01856847-0.00574932j]
```

so yea in fact looks like an overall sign difference between the modes propagating left to right and right to left. Maybe it's related to the backwards mode PR @dbochkov-flexcompute ?

**Comment by tylerflex** [comment]
I vaguely remember that mode amplitudes measured with direction = "-" may have gotten an extra minus sign now?

**Comment by momchil-flex** [comment]
That's probably it yeah. I think one question that was left somewhat ambiguous is what should amplitude -1 signify. Basically, if you take a backwards mode and you `dot` it with itself, you get -1. Now, this doesn't make too much sense in terms of a dot product, but physically we define flux-like quantities with respect to the direction of power propagation through the plane, with the positive sign being power going in the positive axis direction. So if you e.g. do a simulation with a backwards-propagating mode, and you put a `FluxMonitor`, you will get `flux = -1`. Similarly, with the backward mode updates, the `ModeMonitor` amplitude in the backward direction will now be `-1`. And yeah I believe it used to be 1 instead.

Now, we could correct for this sign, but at which point do we want to do that? Should the `ModeMonitor` amplitude be 1 or -1 for perfect injection in the backward direction. 1 makes sense if you think of it as a dot product, -1 makes sense if you think of it as directed flux (actually the units are technically sqrt(W) for the amplitudes). Maybe we should go back to having 1 though.

**Comment by lucas-flexcompute** [comment]
Just to make sure: you're talking about the conjugated dot product, right? The unconjugated inner product should be zero for different modes (forwards/backwards).

**Comment by momchil-flex** [comment]
No, this also applies to lossless modes where the fields can be real and the conjugated/non-conjugated product is the same. Let me explain better. Say that we have such a system where (E, H) is a forward mode. The backward mode can be obtained by just a reflection at the plane normal to the propagation axis, and is thus just (E, -H). The dot product defined as something like int(E1 x H2 + H1 x E2) (regardless of conjugation as for now we assume these are real) is thus 0 between a forward and a backward mode, while (backward, backward) = -(forward, forward). So if modes are normalized such that (forward, forward) = 1, then (backward, backward) = -1. Previously this -1 was somehow factored out such that if did a ModeMonitor between backward-propagating simulation fields and a backward mode, you would get an amplitude of 1. Now you get -1, which makes *some* sense to me.

There is something more to the difference between conjugated and non-conjugated dot-product though. In my discussion above, I did not account for the gauge freedom. Currently, we set the gauge such that the maximum E field is real, so for purely real fields this makes the forward and backward modes as I wrote them down above, with real E and H. However, we could set a different gauge for the backward modes, namely for (E, H) forward, we could set (iE, -iH) as the backward ones. This would still mean (forward, backward) = 0 (conj or not), (backward, backward)_conj = -1, but importantly now (backward, backward)_noconj = 1. This actually makes a lot of sense the way I generally think about the conjugated vs. non-conjugated product: the conjugated one can be interpreted as power flow, but does not always preserver orthonormality. On the other hand the modes can be made orthonormal w.r.t. the non-conjugated dot product, but the amplitudes should not directly be interpreted as power flow.

Maybe from a different perspective: we currently normalize modes by the *conjugated* dot producd (mode, mode)_conj. Maybe we should be normalizing by the non-conjugated one, which should be a "true" dot product I believe (e.g. positive definite?) Furthermore, we should normalize forward and backward modes separately. In the case of real fields, this would exactly produce the (E, H), and (iE, -iH) forward and backward modes, respectively - since the (E, -H) dot product is -1 and we would then normalize by sqrt(-1).

I think physically this makes sense, I am just not sure what the implications will be for e.g. ModeMonitors in lossy waveguides.

**Comment by momchil-flex** [comment]
I feel like we shouldn't normalize by the non-conjugated product. The interpretation of amplitudes as power flow is what people are more used to. If we wanna get rid of that minus sign in the backwards direction we should just decide whether to do it at the `ModeMonitor` or at the `SMatrix` level.

**Comment by dbochkov-flexcompute** [comment]
> Similarly, with the backward mode updates, the ModeMonitor amplitude in the backward direction will now be -1. And yeah I believe it used to be 1 instead.

Actually, I think it used to be -1, but now it is 1 https://github.com/flexcompute/tidy3d-core/pull/316/files#diff-41c19f741aa3ea4ba8eaee7224f7e8ffdd42e38f27385f5f6092327ddfd7ec37R115

**Comment by dbochkov-flexcompute** [comment]
Regarding interpretation:
- another angle of looking at it is that whether -1 should be equivalent to saying that two fields are in anti-phase
- and one more angle: suppose we want to find decomposition of field F into modes u_i: F = sum(a_i u_i), then given our definition of dot product, for forward modes we have `a_i = (F, u_i)`, but for backward modes we have `a_i = - (F, u_i)`. Which also could be written as `a_i = (F, u_i) / (u_i, u_i)` in both cases.

**Comment by momchil-flex** [comment]
> Regarding interpretation:
> 
> * another angle of looking at it is that whether -1 should be equivalent to saying that two fields are in anti-phase

Ok, so this is true for forward-propagating fields, and assuming that currently we also take the -1 out of the backward-propagating ones, it's also true for those right?

> * and one more angle: suppose we want to find decomposition of field `F` into modes `u_i`: `F = sum(a_i u_i)`, then given our definition of dot product, for forward modes we have `a_i = (F, u_i)`, but for backward modes we have `a_i = - (F, u_i)`. Which also could be written as `a_i = (F, u_i) / (u_i, u_i)` in both cases.

I think this also makes perfect sense yeah.

Overall it seems to me that the current setup makes a lot of sense but yeah it has a factor of -1 difference for backward-propagating modes compared to previous releases. I'm probably OK with patching the -1 back in for compatibility, but I think I prefer just leaving this as is and maybe adding an item to our 2.3 changelog clarifying it?

**Comment by lucas-flexcompute** [comment]
My 2c: I've usually preferred the true inner product (unconjugated) for normalization, because I consider forward and backward modes as completely independent modes, each normalized by itself (bidirectional propagation is just a happy accident :) ). That way, mode decomposition becomes even simpler: `F = Σ (a_i, u_i)` and `a_i = (F, u_i)` for any mode. But then, to get the flux direction, I need to take into account the field profile, not just the amplitudes, and really calculate the Poynting vector. So it's the opposite of what we currently do: simpler mode decomposition, more complex flux calculation.

That said, I see no problem either way (normalizing by flux shouldn't be a problem unless we find modes with zero flux, but I guess we don't support those) and if our users are already used to interpreting amplitudes as directional flux, I guess it's better not to change that.

**Comment by tylerflex** [comment]
We might want to revert if nothing else so users don't get different results between minor versions of tidy3d. for example, this particular scattering matrix change took a while to figure out and I guess Lucas and I had both independently spent some time trying to figure whether it was due to bug that each of us might have introduced. Guess it might be too late for 2.3 though?

**Comment by dbochkov-flexcompute** [comment]
> We might want to revert if nothing else so users don't get different results between minor versions of tidy3d

I think another option is to treat it as a bugfix. Again depending on whether the user should expect mode decomposition to have the form of $\boldsymbol{F} = \sum a_i \boldsymbol{u}_i$ or $\boldsymbol{F} = - \sum_b a_i \boldsymbol{u}_i + \sum_f a_i \boldsymbol{u}_i$. Either way that should be clarified somewhere in the docs.

Related to that and regarding conjugated product vs non-conjugated one: now, that I think about it more, it sound like the non-conjugated form should be used during mode decomposition, since it preserves orthogonality and so that $a_i$ can always be extracted. Normalization seems to be a secondary question here, and doesn't seem to actually matter, if formula $a_i = (\boldsymbol{F}, \boldsymbol{u}_i) / (\boldsymbol{u}_i, \boldsymbol{u}_i)$ is applied correctly ( $(\cdot,\cdot)$ is the non-conjugated one here).

**Comment by momchil-flex** [comment]
> > We might want to revert if nothing else so users don't get different results between minor versions of tidy3d
> 
> I think another option is to treat it as a bugfix. Again depending on whether the user should expect mode decomposition to have the form of F=∑aiui or F=−∑baiui+∑faiui. Either way that should be clarified somewhere in the docs.

I kinda also see it as a bugfix. Basically, if I'm looking at a scattering matrix, I would probably not expect to have that extra - sign there (I would interpret it as a fields being off-phase). So the current results make more sense actually?

> Related to that and regarding conjugated product vs non-conjugated one: now, that I think about it more, it sound like the non-conjugated form should be used during mode decomposition, since it preserves orthogonality and so that ai can always be extracted. Normalization seems to be a secondary question here, and doesn't seem to actually matter, if formula ai=(F,ui)/(ui,ui) is applied correctly ( (⋅,⋅) is the non-conjugated one here).

I think to answer both this point and Lucas' comment, yeah, rigorously speaking the correct thing to do is use the non-conjugated one. However, I think that people are more used to interpreting the modal amplitudes in terms of power. Also, this is the default result returned by mode monitors in Lumerical, and they have an extra script that allows the user to get the non-conjugated product if needed (not sure how straightforward it is to use). Their discussion [here](https://optics.ansys.com/hc/en-us/articles/360034902433-Using-and-understanding-Mode-Expansion-Monitors) is pretty relevant to ours. If a user wants to get non-conjugated amplitudes in Tidy3D, they can currently place FieldMonitor and a ModeSolverMonitor and do `mode_data.dot(field_data, conjugate=False)` after the solver run.

Bottom line, my vote: keep the sign change, but add it to the changelog, and expand the ModeMonitor docstring to discuss the definition of the amplitudes as `ai=(F,ui)/(ui,ui) ` where `(., .)` is the *conjugated* product.

**Comment by dbochkov-flexcompute** [comment]
> However, I think that people are more used to interpreting the modal amplitudes in terms of power.

But isn't the modal amplitude just a coefficient in the mode expansion $\boldsymbol{F} = \sum a_i \boldsymbol{u}_i$? That is, it doesn't matter how it is calculated. If $\boldsymbol{u}_i$ is normalized to $\pm 1$, then $|a_i|^2$ gives the power carried in this mode, and the directionality of that power is embedded in the mode itself $(\boldsymbol{u}_i, \boldsymbol{u}_i) = \pm 1$ (conjugated, so it is not affected by gauge selection). Maybe there is a confusion about which dot product is used in calculating coefficients $a_i$, and which dot product in calculating flux/normalization? Sounds like for the former either, conjugated or non-conjugated, product can be used as long as it actually extracts $a_i$, while for the latter only conjugated dot product of a field with itself has valid physical sense. That is, I think we should still normalize using conjugated expression, but when calculating coefficients we have freedom to choose between the two, just have to remember to include factor $1/(u_i, u_i)$

**Comment by momchil-flex** [comment]
> > However, I think that people are more used to interpreting the modal amplitudes in terms of power.
> 
> But isn't the modal amplitude just a coefficient in the mode expansion F=∑aiui? That is, it doesn't matter how it is calculated. If ui is normalized to ±1, then |ai|2 gives the power carried in this mode

The important thing to realize in this expansion is that the $u_i$-s are not orthogonal w.r.t. the conjugated product. Using a non-orthogonal basis is unusual and breaks some standard intuition about these expansions... This is also what leads two different definitions for "power carried in this mode":

If you use conjugated dot product, $|a_i|^2$ can be interpreted as actual power in the sense that e.g. if all other coefficients are 0, a flux computation should return exactly the same as $|a_i|^2$. This is what I think people are used to. However it certainly gets messy if you have multiple modes. Because of the non-orthogonality, $sum_i(|a_i|^2)$ can be larger than the flux, for example.

If you use non-conjugated dot product, $|a_i|^2$ is something like the "amount" of the fields carried by the mode $u_i$, but we can't really call it power in the usual sense (e.g. if all other coefficients are 0, it may still not match the flux through the plane). But this is the much more standard way to do things mathematically, if you don't care about the physics.

I think the underlying issue is that the matrix is non-Hermitian and most of our intuition (or mine at least) comes from decomposing Hermitian matrices. Like associating physical power with expansion coefficients.

**Comment by dbochkov-flexcompute** [comment]
Just to point out slight differences in definitions:
In my thinking I took $\boldsymbol{F} = \sum a_i \boldsymbol{u}_i$ as the definition of $a_i$. But It think you consider definition of $a_i$ through two formulas $a_i = (\boldsymbol{F}, \boldsymbol{u}_i)_c / (\boldsymbol{u}_i, \boldsymbol{u}_i)_c$ and $a_i = (\boldsymbol{F}, \boldsymbol{u}_i)_n / (\boldsymbol{u}_i, \boldsymbol{u}_i)_n$ (?). And, I believe, the second of which coincides with the definition I consider.

But yeah, I see the issue of "decomposing power into modes" vs "decomposing fields into modes". So, maybe in other words, if you combine several (unidirectional) modes carrying powers $P_1 = |a_1|^2$, $P_2 = |a_2|^2$, $...$, and those modes are not orthogonal with respect to conjugated dot product, then the resulting field $\boldsymbol{F} = \sum a_i \boldsymbol{u}_i$ may be carrying power not equal to $\sum P_i$. While you can still decompose the total field into separate modes, you cannot decompose its power into powers of separate modes.

**Comment by momchil-flex** [comment]
> Just to point out slight differences in definitions: In my thinking I took F=∑aiui as the definition of ai. But It think you consider definition of ai through two formulas ai=(F,ui)c/(ui,ui)c and ai=(F,ui)n/(ui,ui)n (?). And, I believe, the second of which coincides with the definition I consider.

Ah yeah I see!

> But yeah, I see the issue of "decomposing power into modes" vs "decomposing fields into modes". So, maybe in other words, if you combine several (unidirectional) modes carrying powers P1=|a1|2, P2=|a2|2, ..., and those modes are not orthogonal with respect to conjugated dot product, then the resulting field F=∑aiui may be carrying power not equal to ∑Pi. While you can still decompose the total field into separate modes, you cannot decompose its power into powers of separate modes.

Right. So while maybe providing the field decomposition amplitudes would make more sense on a general level, what do people usually expect in their results? Usually, they are simulating the propagation into some target mode of interest, and what they expect is that if they somehow filter all other modes out, and they put a detector, it will measure something that's exactly given by the modal amplitude. Which is why the power definition is more commonly used in practice in my opinion.

**Comment by tylerflex** [comment]
@lucas-flexcompute, regarding the `path_dir`, it turns out that making `path_dir` an argument to `ComponentModeler.run()` causes all sorts of problems internally. Basically there's no clean way to store the path_dir passed in, so saving and loading is a headache (and not even very stable in the current version). 

So I made a change in the last commit where `path_dir` was added as a `ComponentModeler` field. The path_dir is still optional in `ComponentModeler.run()` and `.load()` but if it is provided and is different from the field value, it errors. 

Technically this breaks backwards compatibility (if the user has a script where they pass a non-default path dir, it will error). but I wonder how bad this is. What is your opinion? Another idea is to just warn that eg `.load()` will not work unless you set the same path dir as a `ComponentModeler` field.

**Comment by lucas-flexcompute** [comment]
> @lucas-flexcompute, regarding the `path_dir`, it turns out that making `path_dir` an argument to `ComponentModeler.run()` causes all sorts of problems internally. Basically there's no clean way to store the path_dir passed in, so saving and loading is a headache (and not even very stable in the current version).
> 
> So I made a change in the last commit where `path_dir` was added as a `ComponentModeler` field. The path_dir is still optional in `ComponentModeler.run()` and `.load()` but if it is provided and is different from the field value, it errors.
> 
> Technically this breaks backwards compatibility (if the user has a script where they pass a non-default path dir, it will error). but I wonder how bad this is. What is your opinion? Another idea is to just warn that eg `.load()` will not work unless you set the same path dir as a `ComponentModeler` field.

Having it as a field makes sense. We could issue a deprecation warning if it is passed to `run` or `loading` alerting the user of the new field (and later completely remove it as a parameter) and ignore the passed value instead of raising an error. We'll naturally raise the error once we deprecate the parameter anyway.

I tend to like warnings instead of errors if we can. If the user is just in a hurry and (knowingly) ignores the error just for the moment, it's like we're not getting in their way. But that's more of a personal view…

**Comment by tylerflex** [comment]
Yea that makes sense, I did some changes so now if the path_dir provided is different from the default and different from the field value, it will print a warning but still use the passed value. However, for other reasons, the batch will still be saved where the field path_dir is pointing to.

image

I think now this is finally ready for official review

File: CHANGELOG.md [hunk 1]
```diff
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Source validation happens before simulation upload and raises an error if no source is present.
 - Cleaner display of `ArrayLike` in docs.
+### Changed
```
lucas-flexcompute, line 19: Just a reminder not to forget to fix the CHANGELOG

tidy3d/plugins/smatrix/smatrix.py
```diff
@@ -126,12 +128,17 @@ class ComponentModeler(Tidy3dBaseModel):
...
-        "Should be left ``None`` in almost all cases, which will generate the proper batch "
-        "internally and store it as ``ComponentModeler.batch``.",
+        title="Callback URL",
+        description="Http PUT url to receive simulation finish event. "
+        "The body content is a json file with fields "
+        "``{'id', 'status', 'name', 'workUnit', 'solverVersion'}``.",
+    )
+    path_dir: str = pd.Field(
+        DEFAULT_DATA_DIR,
+        title="Path Directory",
```
lucas-flexcompute, line 140: Shouldn't this be "directory path"?

lucas-flexcompute (approved):
I think this looks good. Did that sign change fix the apparent transposition?

CHANGELOG.md
tylerflex: nice catch, actually develop is messed up too so I just fixed everything in the latest commit.

tylerflex:
yea looks like back to the original now I think
image

tylerflex:
actually it seems like the same as before, what changed exactly?

PR #992, opened 2023-07-07
Added NonlinearSusceptibility with chi3 support

caseyflex:
Known limitations:

- Currently only supports non-dispersive Medium

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -34,6 +34,9 @@ FREQ_EVAL_INF = 1e50
 # extrapolation option in custom medium
 FILL_VALUE = "extrapolate"
 
+# cap on number of nonlinear iterations
+NONLINEAR_MAX_NUMITERS = 100
+
 
 def ensure_freq_in_range(eps_model: Callable[[float], complex]) -> Callable[[float], complex]:
     """Decorate ``eps_model`` to log warning if frequency supplied is out of bounds."""
@@ -71,6 +74,75 @@ def ensure_freq_in_range(eps_model: Callable[[float], complex]) -> Callable[[flo
 """ Medium Definitions """
 
 
+class NonlinearSpec(ABC, Tidy3dBaseModel):
+    """Abstract specification for adding a nonlinearity to a medium.
+
+     Note
+    ----
+    The nonlinear constitutive relation is solved iteratively; it may not converge
+    for strong nonlinearities. Increasing `numiters` can help with convergence.
+    """
+
+    numiters: pd.PositiveInt = pd.Field(
+        1,
+        title="Number of iterations",
+        description="Number of iterations for solving nonlinear constitutive relation.",
+    )
+
+    @pd.validator("numiters", always=True)
+    def _validate_numiters(cls, val):
+        """Check that numiters is not too large."""
+        if val > NONLINEAR_MAX_NUMITERS:
+            raise ValidationError(
+                "'NonlinearSpec.numiters' must be less than "
+                f"{NONLINEAR_MAX_NUMITERS}, currently {val}."
+            )
+        return val
+
+
+class NonlinearSusceptibility(NonlinearSpec):
+    """Specification adding an instantaneous nonlinear susceptibility to a medium.
+
+    Note
+    ----
+    The instantaneous nonlinear polarization is given by
+    .. math::
+
+        P_{NL} = \\epsilon_0 \\chi_3 |E|^2 E
+
+    Note
+    ----
+    The nonlinear constitutive relation is solved iteratively; it may not converge
+    for strong nonlinearities. Increasing `numiters` can help with convergence.
+
+    Note
+    ----
+    For complex fields (e.g. when using Bloch boundary conditions), the nonlinearity
+    is applied separately to the real and imaginary parts, so that the above equation
+    holds when both E and :math:`P_{NL}` are replaced by their real or imaginary parts.
+    The nonlinearity is only applied to the real-valued fields since they are the
+    physical fields.
+
+    Note
+    ----
+    Different field components do not interact nonlinearly. For example,
+    when calculating :math:`P_{NL}_x`, we approximate :math:`|E|^2 \\approx |E_x|^2`.
+    This approximation is valid when the E field is predominantly polarized along one
+    of the x, y, or z axes.
+
+    Example
+    -------
+    >>> medium = Medium(permittivity=2, nonlinear_spec=NonlinearSusceptibility(chi3=1))
+    """
+
+    chi3: float = pd.Field(
+        ..., title="Chi3", description="Chi3 nonlinear susceptibility.", units="um^2 / V^2"
+    )
+
+
+NonlinearSpecType = Union[NonlinearSusceptibility]
+
+
 class AbstractMedium(ABC, Tidy3dBaseModel):
     """A medium within which electromagnetic waves propagate."""
 
@@ -92,6 +164,22 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         "diverged. Monitor data up to the divergence point will still be returned and can be "
         "useful in some cases.",
     )
+
+    nonlinear_spec: NonlinearSpecType = pd.Field(
+        None,
+        title="Nonlinear Spec",
+        description="Nonlinear spec applied on top of the base medium properties.",
+    )
+
+    @pd.validator("nonlinear_spec", always=True)
+    def _validate_nonlinear_spec(cls, val):
+        """Check compatibility with nonlinear_spec."""
+        if val is None:
+            return val
+        raise ValidationError(
+            f"A 'nonlinear_spec' of class {type(val)} is not "
+            f"currently supported for medium class {cls}."
+        )
 
     _name_validator = validate_name_str()
 
@@ -537,6 +625,16 @@ class Medium(AbstractMedium):
         units=CONDUCTIVITY,
     )
 
+    @pd.validator("nonlinear_spec", always=True)
+    def _validate_nonlinear_spec(cls, val):
+        """Check compatibility with nonlinear_spec."""
+        if val is None or isinstance(val, NonlinearSusceptibility):
+            return val
+        raise ValidationError(
+            f"A 'nonlinear_spec' of class {type(val)} is not "
+            f"currently supported for medium class {cls}."
+        )
+
     @pd.validator("conductivity", always=True)
     def _passivity_validation(cls, val, values):
         """Assert passive medium if `allow_gain` is False."""
@@ -619,6 +717,16 @@ class CustomIsotropicMedium(AbstractCustomMedium, Medium):
         "permittivity at angular frequency omega is given by conductivity/omega.",
         units=CONDUCTIVITY,
     )
+
+    @pd.validator("nonlinear_spec", always=True)
+    def _validate_nonlinear_spec(cls, val):
+        """Check compatibility with nonlinear_spec."""
+        if val is None:
+            return val
+        raise ValidationError(
+            f"A 'nonlinear_spec' of class {type(val)} is not "
+            f"currently supported for medium class {cls}."
+        )
 
     @pd.validator("permittivity", always=True)
     def _eps_inf_greater_no_less_than_one(cls, val):
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -2885,6 +2885,12 @@ class Simulation(Box):  # pylint:disable=too-many-public-methods
                     red_coords = Coords(**dict(zip("xyz", coords_reduced)))
                     eps_structure = get_eps(structure=structure, frequency=freq, coords=red_coords)
 
+                    if structure.medium.nonlinear_spec is not None:
+                        consolidated_logger.warning(
+                            "Evaluating permittivity of a nonlinear "
+                            "medium ignores the nonlinearity."
+                        )
+
                     if isinstance(structure.geometry, TriangleMesh):
                         consolidated_logger.warning(
                             "Client-side permittivity of a 'TriangleMesh' may be "
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added `hlim` and `vlim` kwargs to `Simulation.plot()` and `Simulation.plot_eps()` for setting horizontal and veritcal plot limits.
+- Added support for chi3 nonlinearity via `NonlinearSusceptibility` class.
 
 ### Changed
 - `nyquist_step` also taking the frequency range of frequency-domain monitors into account.
--- a/tests/sims/simulation_2_4_0rc2.json
+++ b/tests/sims/simulation_2_4_0rc2.json
@@ -15,6 +15,7 @@
         "name": null,
         "frequency_range": null,
         "allow_gain": false,
+        "nonlinear_spec": null,
         "type": "Medium",
         "permittivity": 1.0,
         "conductivity": 0.0
@@ -45,6 +46,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "Medium",
                 "permittivity": 2.0,
                 "conductivity": 0.0
@@ -70,6 +72,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "Medium",
                 "permittivity": 1.0,
                 "conductivity": 3.0
@@ -91,6 +94,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "Sellmeier",
                 "coeffs": [
                     [
@@ -124,6 +128,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "Lorentz",
                 "eps_inf": 2.0,
                 "coeffs": [
@@ -155,6 +160,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "Debye",
                 "eps_inf": 2.0,
                 "coeffs": [
@@ -179,6 +185,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "Debye",
                 "eps_inf": 2.0,
                 "coeffs": [
@@ -209,6 +216,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "Drude",
                 "eps_inf": 2.0,
                 "coeffs": [
@@ -239,11 +247,13 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "Medium2D",
                 "ss": {
                     "name": null,
                     "frequency_range": null,
                     "allow_gain": false,
+                    "nonlinear_spec": null,
                     "type": "PoleResidue",
                     "eps_inf": 1.0,
                     "poles": [
@@ -263,6 +273,7 @@
                     "name": null,
                     "frequency_range": null,
                     "allow_gain": false,
+                    "nonlinear_spec": null,
                     "type": "PoleResidue",
                     "eps_inf": 1.0,
                     "poles": [
@@ -305,6 +316,7 @@
                 "name": "PEC",
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "PECMedium"
             }
         },
@@ -328,11 +340,13 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": null,
+                "nonlinear_spec": null,
                 "type": "AnisotropicMedium",
                 "xx": {
                     "name": null,
                     "frequency_range": null,
                     "allow_gain": false,
+                    "nonlinear_spec": null,
                     "type": "Medium",
                     "permittivity": 1.0,
                     "conductivity": 0.0
@@ -341,6 +355,7 @@
                     "name": null,
                     "frequency_range": null,
                     "allow_gain": false,
+                    "nonlinear_spec": null,
                     "type": "Medium",
                     "permittivity": 2.0,
                     "conductivity": 0.0
@@ -349,6 +364,7 @@
                     "name": null,
                     "frequency_range": null,
                     "allow_gain": false,
+                    "nonlinear_spec": null,
                     "type": "Medium",
                     "permittivity": 3.0,
                     "conductivity": 0.0
@@ -387,6 +403,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "PoleResidue",
                 "eps_inf": 1.0,
                 "poles": [
@@ -423,6 +440,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "CustomMedium",
                 "interp_method": "nearest",
                 "subpixel": false,
@@ -451,6 +469,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "CustomDrude",
                 "eps_inf": "SpatialDataArray",
                 "coeffs": [
@@ -483,6 +502,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "CustomLorentz",
                 "eps_inf": "SpatialDataArray",
                 "coeffs": [
@@ -516,6 +536,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "CustomDebye",
                 "eps_inf": "SpatialDataArray",
                 "coeffs": [
@@ -548,6 +569,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "CustomPoleResidue",
                 "eps_inf": "SpatialDataArray",
                 "poles": [
@@ -580,6 +602,7 @@
                 "name": null,
                 "frequency_range": null,
                 "allow_gain": false,
+                "nonlinear_spec": null,
                 "type": "CustomSellmeier",
                 "coeffs": [
                     [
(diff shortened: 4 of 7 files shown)


tidy3d/components/source.py
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
         return const * offset * oscillation * amp
 
 
-SourceTimeType = Union[GaussianPulse, ContinuousWave]
+class CustomSourceTime(Pulse):
+    """Custom source time dependence.
```
tylerflex, line 305: I think this and the note require a bit more discussion. So if a user makes a CustomTimeSource with some arbitrary values `v_i` placed at times `t_i`, the source that the simulation sees at time `t` will just be the `v_i` corresponding to the nearest time coordinate? Wouldn't this have the potential to cause a source time dependence that isn't smooth (and therefore contains very high frequency components?). If so, is there a way to mitigate this? (eg by providing a `dt` that is much smaller than the simulation `dt`?  I just think we need to make it explicit what this object is doing on the solver side, eg how the source value is chosen at time `t` given the data points, and also perhaps discuss these edge cases (like where the data points are not the same as the simulation time) and potential issues.

tidy3d/components/source.py
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
...
+    Note
+    ----
+    To interpolate to simulation time steps, the value provided at the nearest time is used.
+
+    Example
+    -------
+    >>> sourcetime = CustomSourceTime.from_numpy(values=np.linspace(0, 1, 100), dt=.1)
+    """
+
+    sourcetime_dataset: Optional[TimeDataset] = pydantic.Field(
```
tylerflex, line 316: the `Source` has `source_time` as a field, so I'd vote we change this to `source_time_dataset` if it doesn't break any PEP8 convention for variable naming.

tidy3d/components/source.py
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
...
+class CustomSourceTime(Pulse):
+    """Custom source time dependence.
+
+    Note
+    ----
+    To interpolate to simulation time steps, the value provided at the nearest time is used.
+
+    Example
+    -------
+    >>> sourcetime = CustomSourceTime.from_numpy(values=np.linspace(0, 1, 100), dt=.1)
```
tylerflex, line 313: does this work? what about freq0 and fwidth?

tidy3d/components/source.py
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
...
+        ..., title="Source time dataset", description="Dataset for storing the custom source time."
+    )
+
+    _sourcetime_dataset_none_warning = warn_if_dataset_none("sourcetime_dataset")
+
+    @classmethod
+    def from_numpy(
+        cls, freq0: float, fwidth: float, values: ArrayComplex1D, dt: float
+    ) -> "CustomSourceTime":
+        """Create a CustomSourceTime from a numpy array.
```
tylerflex, line 326: I think it needs some more explicit  mention of what time coordinate are used (assume the user isn't looking at the source) for example.

```
The time coordinates used are evenly spaced between 0 and dt * (N-1) with a step size of `dt` where N is the length of the values array.

File: tidy3d/components/source.py [hunk 5]
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
...
+    """
+
+    sourcetime_dataset: Optional[TimeDataset] = pydantic.Field(
+        ..., title="Source time dataset", description="Dataset for storing the custom source time."
+    )
+
+    _sourcetime_dataset_none_warning = warn_if_dataset_none("sourcetime_dataset")
+
+    @classmethod
+    def from_numpy(
```
tylerflex, line 323: maybe `from_values` or something makes more sense here? One could imagine `from_numpy` perhaps accepting the time coordinates instead of dt, for example.

tidy3d/components/source.py
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
...
+        cls, freq0: float, fwidth: float, values: ArrayComplex1D, dt: float
+    ) -> "CustomSourceTime":
```
tylerflex, line 325: no need to put this in quotes if there's an 
```py
```
line at the top of the file

tidy3d/components/source.py
tylerflex: maybee line 342 is a good place for this.

tidy3d/components/source.py
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
...
+            Estimated frequency width of the source.
+        values: ArrayComplex1D
+            Complex values of the source amplitude.
+        dt: float
+            Time step for the `values` array.
+
+        Returns
+        -------
+        CustomSourceTime
+            `CustomSourceTime` with these values.
```
tylerflex, line 342: please use :class: linking for the CustomSourceTime

tidy3d/components/source.py
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
...
+        """
+        times = np.arange(len(values)) * dt
+        return CustomSourceTime(
+            freq0=freq0,
+            fwidth=fwidth,
+            sourcetime_dataset=TimeDataset(values=TimeDataArray(values, coords=dict(t=times))),
```
tylerflex, line 348: can we construct this variable before passing it to `CustomSourceTime` for readability?

```py
sourcetime_dataset=TimeDataArray(values, coords=dict(t=times))

return CustomSourceTime(
         freq0=freq0,
         fwidth=fwidth,
         sourcetime_dataset=sourcetime_dataset,
)

File: tidy3d/components/source.py [hunk 9]
```diff
@@ -9,10 +9,20 @@
 
 from .base import Tidy3dBaseModel, cached_property
 
-from .types import Direction, Polarization, Ax, FreqBound, ArrayFloat1D, Axis, PlotVal
+from .types import (
+    Direction,
+    Polarization,
+    Ax,
```
tylerflex, line 15: please split this import into two lines
```py
from .types import x, y, z
from .types import a, b, c
```

just an arbitrary code style thing, for consistency.

tidy3d/components/source.py
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
...
+    Example
+    -------
+    >>> sourcetime = CustomSourceTime.from_numpy(values=np.linspace(0, 1, 100), dt=.1)
+    """
+
+    sourcetime_dataset: Optional[TimeDataset] = pydantic.Field(
+        ..., title="Source time dataset", description="Dataset for storing the custom source time."
+    )
+
+    _sourcetime_dataset_none_warning = warn_if_dataset_none("sourcetime_dataset")
```
tylerflex, line 320: do we want to add any validation on the supplied sourcetime_dataset? For example, what if the time coordinates are all negative? or all extend outside of the simulation run_time? will it be handled as expected?

tests/test_components/test_IO.py
```diff
@@ -37,6 +37,8 @@ def set_datasets_to_none(sim):
             src["field_dataset"] = None
         elif src["type"] == "CustomCurrentSource":
             src["current_dataset"] = None
+        if src["source_time"]["type"] == "CustomSourceTime":
```
tylerflex, line 40: I think we also need to add this to the `Simulation.custom_datasets` property for things to work properly. For example, if not added there, the simulation may not be uploaded properly as an hdf5

https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/simulation.py#L2654-L2670

tidy3d/components/medium.py
```diff
@@ -2927,6 +2927,51 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
     )
 
 
+class AbstractNonlinearMedium(AbstractMedium, ABC):
+    """Abstract nonlinear medium.
+
+    Note
```
tylerflex, line 2933: not sure this note will propagate to the subclasses, so might need to copy it over.

tidy3d/components/medium.py
```diff
@@ -2927,6 +2927,51 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
...
+    ----
+    The instantaneous nonlinear polarization is given by
+    .. math::
+
+        P_{NL} = \\epsilon_0 \\chi_3 |E|^2 E
+
+    Note
+    ----
+    For complex fields (e.g. when using Bloch boundary conditions), the nonlinearity
+    is applied separately to the real and imaginary parts.
```
tylerflex, line 2960: could you be more explicit? so does it mean that the P_NL imaginary and real parts are computed separately from |real(E)|^3 and  |imag(E)|^3 respectively? or something else? what are the practical considerations of this? for example, is it still accurate to use KerrMedium with bloch boundaries?

tidy3d/components/medium.py
```diff
@@ -2927,6 +2927,51 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
...
+
+    Note
+    ----
+    Different field components do not interact nonlinearly. For example,
+    when calculating :math:`P_{NL}_x`, we approximate :math:`|E|^2 \\approx |E_x|^2`.
```
tylerflex, line 2965: this seems potentially quite limiting. does this mean that the nonlinearity is only accurate for 2D simulations with E field out of plane? (for example?) again, what are the practical ramifications.

tidy3d/components/medium.py
```diff
@@ -2927,6 +2927,51 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
...
+
+    Example
+    -------
+    >>> medium = KerrMedium(chi3=1)
+    """
+
+    chi3: float = pd.Field(..., title="Chi3", description="Chi3 nonlinear susceptibility.")
```
tylerflex, line 2972: can chi3 be negative? (I forget...) if not, make this a `pd.NonNegativeFloat` type.

tidy3d/components/medium.py
```diff
@@ -2927,6 +2927,51 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
...
+
+    Note
+    ----
+    The nonlinear constitutive relation is solved iteratively; it may not converge
+    for large nonlinearities. Increasing `numiters` can help with convergence.
+
+    """
+
+    numiters: pd.PositiveInt = pd.Field(
+        1,
```
tylerflex, line 2941: is 1 a reasonable number or should this be adjusted case by case? it might be good to start with a default value of something more practical for most cases (define as a constant at the top of file)? but maybe I'm not understanding something.

tylerflex (changes requested):
Thanks @caseyflex, I think both of these will be super exciting to have. I suggest having @momchil-flex chime in specifically on how we do interpolation of the custom source time as I'm a bit concerned about edge cases using nearest interpolation.  Also, is it possible to split this into two PRs? they seem quite separate no? or at least two commits if possible. Otherwise it will be a bit confusing to have both somewhat orthogonal changes in a single commit.

tylerflex:
Regarding this:

> Can we make this subclass something that could be either Medium or DispersiveMedium?

I dont see why not. currently the Kerr Medium assumes linear medium properties of vacuum? or is this a WIP?

tidy3d/components/source.py
```diff
@@ -291,7 +301,72 @@ def amp_time(self, time: float) -> complex:
...
+            Time in seconds.
+
+        Returns
+        -------
+        complex
+            Complex-valued source amplitude at that time.
+        """
+        if self.sourcetime_dataset is None:
+            return None
+        return self.sourcetime_dataset.values.sel(t=time, method="nearest").to_numpy()
```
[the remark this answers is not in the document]
caseyflex: When I was using this, I was oversampling enough in my input signal that the interpolation method didn't matter. We could change it though, if you think e.g. linear interpolation is less likely to cause problems.

tidy3d/components/source.py
caseyflex: yeah, using a small `dt` migitages this. Agreed, need more discussion here. I'll also split off the PR for CustomSourceTime

tidy3d/components/medium.py
caseyflex: 1 has been working fine for me. I think it might need to be adjusted if the nonlinearity was quite strong.

tidy3d/components/medium.py
caseyflex: It's accurate as long as the E field is predominantly polarized along one of the Cartesian axes. See also the discussion at the backend PR.

tidy3d/components/medium.py
caseyflex: yeah, it's still accurate to use KerrMedium with bloch boundaries. It wouldn't be physical do use the complex fields directly in definining the instantaneous nonlinearity, since the instantaneous physical fields are real.

tidy3d/components/medium.py
caseyflex: yeah, I think we should allow it to be negative

caseyflex:
CustomSourceTime moved here: https://github.com/flexcompute/tidy3d/pull/994

tylerflex:
Hm, maybe we can stick with `Medium` as the superclass until someone requests dispersive nonlinear? Otherwise we could I guess consider two `KerrMedium` and `KerrPoleResidue` which both inherit from the abstract nonlinear medium and then `Medium`, `PoleResidue` respectively? For now probably just `Medium` is fine.

tidy3d/components/medium.py
```diff
@@ -2927,6 +2927,61 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
...
+    ----
+    The nonlinear constitutive relation is solved iteratively; it may not converge
+    for large nonlinearities. Increasing `numiters` can help with convergence.
+
+    Note
+    ----
+    For complex fields (e.g. when using Bloch boundary conditions), the nonlinearity
+    is applied separately to the real and imaginary parts, so that the above equation
+    holds when both E and :math:`P_{NL}` are replaced by their real or imaginary parts.
+    Since the physical fields are real, it wouldn't make sense to apply the instantaneous
```
tylerflex, line 2967: maybe instead of this last sentence with "it wouldn't make sense" we can soften it with something like "The nonlinearity is only applied to the real-valued fields since they are the physical fields."

tylerflex (approved):
This looks good to me pending two minor rewordings, thanks @caseyflex!

tidy3d/components/medium.py
```diff
@@ -2927,6 +2927,61 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
...
+    Note
+    ----
+    .. math::
+
+        P_{NL} = \\epsilon_0 \\chi_3 |E|^2 E
+
+    Note
+    ----
+    The nonlinear constitutive relation is solved iteratively; it may not converge
+    for large nonlinearities. Increasing `numiters` can help with convergence.
```
tylerflex, line 2960: perhaps "strong" is better than "large"?

momchil-flex (commented):
Looks good!

> Backend supports dispersive KerrMedium but frontend currently does not. Can we make this subclass something that could be either Medium or DispersiveMedium?

This goes back to our discussion on slack today... It really does feel like our medium subclassing is becoming quite a mess. One option is to treat the nonlinear susceptibility (and the time-modulated permittivity in the future @weiliangjin2021 ) as something that can be added to any medium where it's supported, e.g.

```
td.Medium(permittivity=..., nonlinearity=KerrNonlinearity(chi3=1))
```

Then we can pick which media have a `nonlinearity` Field. In fact this could still be done through subclassing, something like

```
# this would probably still subclass from some abstract nonlinearity class
class KerrNonlinearity():
    chi3: float
    numiters: pd.PositiveInt

class AbstractNonlinearMedium(AbstractMedium, ABC):
    nonlinearity: Union[KerrNonlinearity, ...]

# now AbstractNonlinearMedium can be added anywhere we support, e.g.
class Medium(AbstractMedium, AbstractNonlinearMedium):
    ...

class DispersiveMedium(AbstractMedium, AbstractNonlinearMedium, ABC):
    ...

```

In the case of nonlinearity, can it pretty much be added on top of any medium, including custom?

momchil-flex:
Note: this is just a suggestion, not saying you should implement this right now. But we should decide on something (@tylerflex )

momchil-flex:
Whatever way we introduce the nonlinear medium, we should make sure the mode solver issues a warning that the nonlinearity is not used if someone tries a mode solve in a plane that contains a nonlinear medium. I forget, how do we handle 2D media in the mode solver now?

tylerflex:
> This goes back to our discussion on slack today... It really does feel like our medium subclassing is becoming quite a mess. One option is to treat the nonlinear susceptibility (and the time-modulated permittivity in the future @weiliangjin2021 ) as something that can be added to any medium where it's supported
> But we should decide on something (@tylerflex )

In principle it makes sense to be able to add nonlinearity to any medium. However, with the way things are set up with subclassing, I think it might make things a bit complicated. My thinking is just make KerrMedium [and eventually KerrDispersiveMedium], let the class type do the work regarding what properties the medium has, but eventually we should just separate all of the **categories** of effects in the medium:
* Linear properties
* Nonlinear properties
* Dispersion
* Time dependence
into their own fields of a Medium class, each determined by their own specs, and handle custom medium by overloading any fields with data arrays. but this seems like a 3.0 change.

I think how Casey has it here is good for now and we can always add the nonlinearity fields into the AbstracMedium class without breaking backwards compatibility if we change our minds?

caseyflex:
> Whatever way we introduce the nonlinear medium, we should make sure the mode solver issues a warning that the nonlinearity is not used if someone tries a mode solve in a plane that contains a nonlinear medium. I forget, how do we handle 2D media in the mode solver now?

We use `volumetric_structures` in `Simulation.epsilon_on_grid` and related functions, so the mode solver should work correctly for 2D materials. To handle nonlinear case, I'll put a warning in `Simulation.epsilon_on_grid`. Of course there's no way to know epsilon a priori for nonlinear materials.

caseyflex:
> > This goes back to our discussion on slack today... It really does feel like our medium subclassing is becoming quite a mess. One option is to treat the nonlinear susceptibility (and the time-modulated permittivity in the future @weiliangjin2021 ) as something that can be added to any medium where it's supported
> > But we should decide on something (@tylerflex )
> 
> In principle it makes sense to be able to add nonlinearity to any medium. However, with the way things are set up with subclassing, I think it might make things a bit complicated. My thinking is just make KerrMedium [and eventually KerrDispersiveMedium], let the class type do the work regarding what properties the medium has, but eventually we should just separate all of the **categories** of effects in the medium:
> 
> * Linear properties
> * Nonlinear properties
> * Dispersion
> * Time dependence
> 
> into their own fields of a Medium class, each determined by their own specs, and handle custom medium by overloading any fields with data arrays. but this seems like a 3.0 change.
> 
> I think how Casey has it here is good for now and we can always add the nonlinearity fields into the AbstracMedium class without breaking backwards compatibility if we change our minds?

Yeah, let's save the restructuring for 3.0. I'll leave it this way with `KerrMedium` for now, and introduce other classes like `KerrDispersiveMedium` as needed later.

tidy3d/components/medium.py
```diff
@@ -71,6 +71,53 @@ def _eps_model(self, frequency: float) -> complex:
 """ Medium Definitions """
 
 
+class Chi3Spec(Tidy3dBaseModel):
+    """Chi3 medium described by a chi3 nonlinear susceptibility.
```
tylerflex, line 75: "Specification adding chi3 nonlinear susceptibility to a medium." probably is more accurate.

tidy3d/components/medium.py
```diff
@@ -71,6 +71,53 @@ def _eps_model(self, frequency: float) -> complex:
 """ Medium Definitions """
 
 
+class Chi3Spec(Tidy3dBaseModel):
```
tylerflex, line 74: Do we want an abstract `NonlinearSpec`? or I guess maybe later?

tidy3d/components/medium.py
```diff
@@ -93,8 +140,23 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
...
+    )
+
     _name_validator = validate_name_str()
 
+    @pd.validator("nonlinear_spec", always=True)
+    def _validate_nonlinear_spec(cls, val):
+        """Validate compatibility with the nonlinear spec."""
+        if val is None:
+            return None
+        if issubclass(cls, Medium):
```
tylerflex, line 156: I think this is a problematic way to do this. It's a bit circular? if we moved `Medium` and `AbstractMedium` to their own files, it would break I think. We're dealing with some issues like this in `geometry.py` right now and it's a headache so I'd say let's keep AbstratMedium agnostic to Medium.

A few possible solutions:

1. only move `nonlinear_spec` to the `Medium` for now until we find a better way to do this.

2.  add `nonlinear_spec` to `AbstractMedium` like this, but only allow `None` values.  if it's not None, give the warning that you have here (optional).  Add `nonlinear_spec`  to `Medium` with a allowed type of `Union[Chi3Spec]` and overwrite validator if we had one.

In general, we need a simple way to validate whether we support different possible combinations of specs.  Haven't figured out a good solution yet.

tidy3d/components/medium.py
```diff
@@ -3186,6 +3244,10 @@ def perturbed_copy(
         return CustomPoleResidue.parse_obj(new_dict)
 
 
+# PEC builtin instance
```
tylerflex, line 3247: curious why did this get moved?

tidy3d/components/simulation.py
```diff
@@ -2885,6 +2885,12 @@ def make_eps_data(coords: Coords):
                     red_coords = Coords(**dict(zip("xyz", coords_reduced)))
                     eps_structure = get_eps(structure=structure, frequency=freq, coords=red_coords)
 
+                    if not structure.medium.nonlinear_spec is None:
```
tylerflex, line 2888:
```py
if not structure.medium.nonlinear_spec is None:
```
becomes
```py
if structure.medium.nonlinear_spec is not None:
```
is the convention in python.

tylerflex (changes requested):

tylerflex:
This will be a better way to do things, I can already tell. Thanks @caseyflex !

tidy3d/components/medium.py
caseyflex: maybe later I think

tidy3d/components/medium.py
```diff
@@ -71,6 +71,50 @@ def _eps_model(self, frequency: float) -> complex:
...
+    when calculating :math:`P_{NL}_x`, we approximate :math:`|E|^2 \\approx |E_x|^2`.
+    This approximation is valid when the E field is predominantly polarized along one
+    of the x, y, or z axes.
+
+    Example
+    -------
+    >>> medium = Medium(permittivity=2, nonlinear_spec=Chi3Spec(chi3=1))
+    """
+
+    chi3: float = pd.Field(..., title="Chi3", description="Chi3 nonlinear susceptibility.")
```
tylerflex, line 109: does this have units?

tidy3d/components/medium.py
```diff
@@ -93,6 +137,12 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         "useful in some cases.",
     )
 
+    nonlinear_spec: None = pd.Field(
```
dbochkov-flexcompute, line 140: just wanted to clarify how this works: would the user just get an error if they try to plug `Chi3Spec` into a dispersive medium? Ideally we should have some meaningful error message something like `f"Nonlinear feature is not currently supported for medium class {type(cls)}"`, but it's up to you

Btw, `CustomIsotropicMedium` (which will become just `CustomMedium` in 3.0) inherits from `Medium`. Do we support `Chi3Spec` for custom mediums on backend?

dbochkov-flexcompute (approved):
Looks good to me too!

caseyflex:
I added AbstractNonlinearSpec for convenience later, and I also changed AbstractMedium to take NonlinearSpecType, but then validate it as needed for different medium types. I think this addresses Daniil's concern and is more robust for the future.

tidy3d/components/medium.py
```diff
@@ -71,6 +74,75 @@ def _eps_model(self, frequency: float) -> complex:
...
+        title="Number of iterations",
+        description="Number of iterations for solving nonlinear constitutive relation.",
+    )
+
+    @pd.validator("numiters", always=True)
+    def _validate_numiters(cls, val):
+        """Check that numiters is not too large."""
+        if val > NONLINEAR_MAX_NUMITERS:
+            raise ValidationError(
+                "NonlinearSpec.numiters must be less than "
```
tylerflex, line 97: let's put anything code related in single quotes `'`. `'NonlinearSpec.numiters'`.

tidy3d/components/medium.py
```diff
@@ -93,6 +165,22 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         "useful in some cases.",
     )
 
+    nonlinear_spec: NonlinearSpecType = pd.Field(
+        None,
+        title="Nonlinear Spec",
+        description="Nonlinear spec applied on top of the base medium properties.",
+    )
+
+    @pd.validator("nonlinear_spec", always=True)
```
tylerflex, line 174: feel like we need a better way to do this more generally. so let's say we have 
```py
NonlinearSpecType = Union[None, Spec1, Spec2, ..., SpecN]
```
and then classes `ClassA`, `ClassB`, ... that each can only accept a certain subset of `NonlinearSpecType`.

Maybe the most general way is to just define in the `AbstractMedium`
```py
nonlinear_spec : None = pd.Field(None, ...)
```
and then in the individual classes
```py
nonlinear_spec: Union[None, Spec1]
```
? Which maybe becomes a bit annoying as the combinations grow but is reasonably general? Would pydantic take care of erroring if one tried to put the wrong spec in?  Basically I feel it might be nice to do this only in the type annotation rather than the validator. This also makes it more clear in the docs.

tylerflex (changes requested):

tidy3d/components/medium.py
tylerflex: On the other hand, if everything is merged into the same medium, it might be harder to keep track of the dependencies just in the type annotation. Say a non-dispersive medium can accept one type of NonlinearSpec but a dispersive can only accept another. There would be some cross coupling between the two specs that I think could only be handled through validators, so not sure. Any thoughts?

caseyflex: My opinion is that the type annotations are not the best tool for this. The incompatibilities seem to usually be from things that haven't been implemented yet, and cross coupling between different specs (for example, once pole residue is a spec, nonlinear + dispersive)

caseyflex: If we did go with the type annotation approach, like Daniil mentioned, we would still want validators that would give more helpful error messages in certain cases, like when Chi3Spec is added to CustomIsotropicMedium. So my feeling is that the type annotations add boilerplate without much content

caseyflex: but the real argument is actually like you said, the cross coupling between different specs down the line.

tylerflex: true. If we had a set of "allowed" combinations, though. we could always have a generic validator like
```
def validate_specs(cls, values):
    nonlinear_spec_type = type(values.get("nonlinear_spec"))
    dispersion_spec_type = type(values.get("dispersion_spec"))
    if (nonlinear_spec_type, dispersion_spec_type) not in ALLOWED_SPEC_COMBINATIONS:
         raise ValidationError(f"Combination of {nonlinear_spec_type} and {dispersion_spec_type} not currently supported.").
```
which could be cleaner and we could document the `ALLOWED_SPEC_COMBINATIONS`. What I worry about slightly is a set of validators with really convoluted logic and no simple way for the user to see what combinations are actually allowed. However, I guess if the allowed combinations is pretty easy to write down as a function, it might indeed be easier? @dbochkov-flexcompute any thoughts?

tidy3d/components/medium.py
```diff
@@ -71,6 +74,75 @@ def _eps_model(self, frequency: float) -> complex:
...
+    This approximation is valid when the E field is predominantly polarized along one
+    of the x, y, or z axes.
+
+    Example
+    -------
+    >>> medium = Medium(permittivity=2, nonlinear_spec=NonlinearSusceptibility(chi3=1))
+    """
+
+    chi3: float = pd.Field(
+        ..., title="Chi3", description="Chi3 nonlinear susceptibility, in units of um^2 / V^2."
```
tylerflex, line 139: we should pass the units as a keyword argument to `Field`, eg.

```py
chi3: float = pd.Field(
    ...,
    title=...,
    description=...,
    units="um^2 / V^2",
)
```
We parse this field when setting units in the API reference. Could you add this? Thanks.

tylerflex (approved):
Approve pending the units kwarg in `pd.Field()`. thanks @caseyflex

PR #994, opened 2023-07-11
Added CustomSourceTime

caseyflex:
Features:

- Validation heuristic for sufficiently fast sampling to avoid interpolation artifacts

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -26,7 +26,7 @@ from .boundary import BoundarySpec, BlochBoundary, PECBoundary, PMCBoundary, Per
 from .boundary import PML, StablePML, Absorber, AbsorberSpec
 from .structure import Structure
 from .source import SourceType, PlaneWave, GaussianBeam, AstigmaticGaussianBeam, CustomFieldSource
-from .source import CustomCurrentSource
+from .source import CustomCurrentSource, CustomSourceTime
 from .source import TFSF, Source
 from .monitor import MonitorType, Monitor, FreqMonitor, SurfaceIntegrationMonitor
 from .monitor import AbstractFieldMonitor, DiffractionMonitor, AbstractFieldProjectionMonitor
@@ -68,6 +68,10 @@ DIST_NEIGHBOR_REL_2D_MED = 1e-5
 
 # height of the PML plotting boxes along any dimensions where sim.size[dim] == 0
 PML_HEIGHT_FOR_0_DIMS = 0.02
+
+# allow some numerical flexibility before warning about CustomSourceTime
+# in units of dt
+CUSTOMSOURCETIME_TOL = 1.1
 
 
 class Simulation(Box):  # pylint:disable=too-many-public-methods
@@ -840,6 +844,7 @@ class Simulation(Box):  # pylint:disable=too-many-public-methods
         """Call validators taking z`self` that get run after init."""
         self._validate_no_structures_pml()
         self._validate_tfsf_nonuniform_grid()
+        self._validate_customsourcetime()
 
     def _validate_no_structures_pml(self) -> None:
         """Ensure no structures terminate / have bounds inside of PML."""
@@ -908,6 +913,34 @@ class Simulation(Box):  # pylint:disable=too-many-public-methods
                             "uniform grid in both directions tangential to the TFSF injection "
                             f"axis, '{'xyz'[source.injection_axis]}'."
                         )
+
+    def _validate_customsourcetime(self) -> None:
+        """Make sure custom source time is not undersampled.
+        Also, make sure that all simulation.tmesh values are covered."""
+        for source in self.sources:
+            if isinstance(source.source_time, CustomSourceTime):
+                dataset = source.source_time.source_time_dataset
+                if dataset is None:
+                    continue
+                times = dataset.values.coords["t"].values
+                if (
+                    min(times) > self.tmesh[0]
+                    or max(times) < self.tmesh[-1] - CUSTOMSOURCETIME_TOL * self.dt
+                ):
+                    raise ValidationError(
+                        "'CustomSourceTime' found with time coordinates "
+                        "'times' that do not cover the entire 'Simulation.tmesh'. Currently, "
+                        f"'(min(times), max(times)) = ({min(times)}, {max(times)})', while "
+                        f"'(min(tmesh), max(tmesh)) = ({self.tmesh[0]}, {self.tmesh[-1]}).' "
+                    )
+                max_dt = np.amax(np.diff(times))
+                if max_dt > self.dt * CUSTOMSOURCETIME_TOL:
+                    log.warning(
+                        f"'CustomSourceTime' found with time step 'max(dt) = {max_dt:.3g}', "
+                        f"while the simulation time step is 'dt={self.dt}'. "
+                        "We recommend that the largest time step of the custom source "
+                        f"be smaller than the time step of the simulation."
+                    )
 
     """ Pre submit validation (before web.upload()) """
 
@@ -2693,6 +2726,11 @@ class Simulation(Box):  # pylint:disable=too-many-public-methods
         """List of custom datasets for verification purposes. If the list is not empty, then
         the simulation needs to be exported to hdf5 to store the data.
         """
+        datasets_source_time = [
+            src.source_time.source_time_dataset
+            for src in self.sources
+            if isinstance(src.source_time, CustomSourceTime)
+        ]
         datasets_field_source = [
             src.field_dataset for src in self.sources if isinstance(src, CustomFieldSource)
         ]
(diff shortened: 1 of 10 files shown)


tidy3d/components/source.py
```diff
@@ -291,7 +294,83 @@ def amp_time(self, time: float) -> complex:
...
+            Time in seconds.
+
+        Returns
+        -------
+        complex
+            Complex-valued source amplitude at that time.
+        """
+        if self.source_time_dataset is None:
+            return None
+        return self.source_time_dataset.values.sel(t=time, method="nearest").to_numpy()
```
tylerflex, line 370: note: maybe `interp` makes more sense here?

tidy3d/components/source.py
```diff
@@ -291,7 +294,83 @@ def amp_time(self, time: float) -> complex:
...
+
+
+SourceTimeType = Union[GaussianPulse, ContinuousWave, CustomSourceTime]
```
tylerflex, line 373: note: maybe we should just take this opportunity to remove CW,  dont think it's actually supported. @momchil-flex ?

tidy3d/components/source.py
```diff
@@ -291,7 +294,83 @@ def amp_time(self, time: float) -> complex:
...
+        ..., title="Source time dataset", description="Dataset for storing the custom source time."
+    )
+
+    _source_time_dataset_none_warning = warn_if_dataset_none("source_time_dataset")
+
+    @classmethod
+    def from_values(
+        cls, freq0: float, fwidth: float, values: ArrayComplex1D, dt: float
+    ) -> CustomSourceTime:
+        """Create a CustomSourceTime from a numpy array.
```
tylerflex, line 323: add :class: link here

tidy3d/components/source.py
```diff
@@ -291,7 +294,83 @@ def amp_time(self, time: float) -> complex:
         return const * offset * oscillation * amp
 
 
-SourceTimeType = Union[GaussianPulse, ContinuousWave]
+class CustomSourceTime(Pulse):
+    """Custom source time dependence.
```
tylerflex, line 298: one thing also that might be worth mentioning explicitly is whether this can be complex-valued.

tidy3d/components/source.py
```diff
@@ -291,7 +294,83 @@ def amp_time(self, time: float) -> complex:
...
+
+    Note
```
tylerflex, line 300: is it worth validating in `Simulation` to ensure the source time is not under sampled?  What I worry about is the user ignoring this and then getting a divergent simulation. is there a way to detect under sampling? Eg by checking the minimum of the `np.diff` of the time coordinates and comparing to the `Simulation.dt`?

Also what would be the downside of just using `.interp` instead of `sel`?

tylerflex (commented):
added a few notes for discussion

tidy3d/components/source.py
caseyflex: CW works, it just isn't shown in the docs

momchil-flex: Yeah it should work, it's not tested though.

momchil-flex:
I think Tyler's question about using `interp` instead of `sel` makes sense, and also validating under-sampling.

tidy3d/components/simulation.py
```diff
@@ -908,6 +912,20 @@ def _validate_tfsf_nonuniform_grid(self) -> None:
...
+        for source in self.sources:
+            if isinstance(source.source_time, CustomSourceTime):
+                dataset = source.source_time.source_time_dataset
+                if dataset is None:
+                    continue
+                times = dataset.values.coords["t"].values
+                max_dt = np.amax(np.diff(times))
+                if max_dt > self.dt / CUSTOMSOURCETIME_SAMPLING_WARNING_FACTOR:
+                    log.warning(
+                        "'CustomSourceTime' may be undersampled relative to 'Simulation.dt'"
```
tylerflex, line 926: I'd just suggest making this warning a bit more descriptive. Printing the values of `max_dt` `self.dt ` and suggesting the solution, namely that we recommend the max dt of the custom source should be less than the dt of the simulation.

tylerflex (changes requested):

tidy3d/components/source.py
```diff
@@ -291,7 +294,81 @@ def amp_time(self, time: float) -> complex:
...
+            Time in seconds.
+
+        Returns
+        -------
+        complex
+            Complex-valued source amplitude at that time.
+        """
+        if self.source_time_dataset is None:
+            return None
+        return self.source_time_dataset.values.interp(t=time).to_numpy()
```
tylerflex, line 368: if we are going with `interp`, there are a couple edge cases brought on by xarray's API that should probably be addressed. 
1. if the source_time_dataset only has a single time value, it will raise a very obscure message.
2. if `time` is out of range of the dataset, I believe it will return nans, so we either need to 
    a. convert nans to 0 (`np.nan_to_num`)
    b. use an interpolation scheme where out of range values use the closest data point.
    c. validate (in the Simulation-level validator) that the dataset covers all of the possible `Simulation.tmesh`

So I think it would be good to

tylerflex: Is the 2nd issue resolved in the current status?

caseyflex: I added a Simulation-level validator that it covers the Simulation.tmesh, as well as a test for this error

tylerflex (approved):
looks good to me, let's wait to merge until @momchil-flex has a final look in conjunction with the backend tests, thanks @caseyflex !

caseyflex:
> looks good to me, let's wait to merge until @momchil-flex has a final look in conjunction with the backend tests, thanks @caseyflex !

I changed a minor thing -- allowing a dt of tolerance in covering the Simulation.tmesh, and just using the end values if needed in this case.

tylerflex:
Just occurred to me, eventually, it would probably be useful to provide a class method to generate this source time from data in the frequency domain as well.

tidy3d/components/simulation.py
```diff
@@ -908,6 +913,34 @@ def _validate_tfsf_nonuniform_grid(self) -> None:
...
+        """Make sure custom source time is not undersampled.
+        Also, make sure that all simulation.tmesh values are covered."""
+        for source in self.sources:
+            if isinstance(source.source_time, CustomSourceTime):
+                dataset = source.source_time.source_time_dataset
+                if dataset is None:
+                    continue
+                times = dataset.values.coords["t"].values
+                if (
+                    min(times) >= self.tmesh[0] + CUSTOMSOURCETIME_TOL * self.dt
```
tylerflex, line 926: what is the intuition behind this choice? so the user can essentially start the time coordinates at `dt` and end them at `run_time - dt` and it should pass?

caseyflex:
Yeah. I just found that without any tolerance on the ends, I would get
errors because of floating point issues, like being just under the actual
range. A dt on either side shouldn’t hurt.

caseyflex:
Could make it fp_eps I suppose, instead of dt

caseyflex:
But the tmesh already runs to run_time + dt, and it’s most natural to have
the source go only to run_time

caseyflex:
Could also have no tolerance at 0, and a dt of tolerance at run_time.

tylerflex:
A few things:
* The point of this validator is to ensure that the user-supplied data is enough to cover the tmesh right? so this change seems to reduce the size of the range from `(tmesh[0], tmesh[-1])` to `(tmesh[0] + 1.1 dt, tmesh[-1] - 1.1dt)`. Basically with this change, it seems a could not supply data for `t=0` and a `nan` would get created on our side right? That seems problematic.
* the `tmesh` is `np.arange(0.0, self.run_time + dt, dt)`, `np.arange` doesnt include the final value so the final value is actually `self.run_time`, just FYI.
image

So the problem you're having is that if a user supplies `run_time=x` to the Simulation and then makes a `t` coordinate grid for the custom time source dataset using `np.linspace(0, run_time, N)`, the validator will sometimes complain due to numerical precision?  If so, maybe one option is to relax the validator for the [0] and [-1] values by -fp_eps and +fp_eps (opposite direction as current?) and then make sure to also apply this on the solver side?

caseyflex:
> A few things:
> 
> * The point of this validator is to ensure that the user-supplied data is enough to cover the tmesh right? so this change seems to reduce the size of the range from `(tmesh[0], tmesh[-1])` to `(tmesh[0] + 1.1 dt, tmesh[-1] - 1.1dt)`. Basically with this change, it seems a could not supply data for `t=0` and a `nan` would get created on our side right? That seems problematic.
> * the `tmesh` is `np.arange(0.0, self.run_time + dt, dt)`, `np.arange` doesnt include the final value so the final value is actually `self.run_time`, just FYI.
> 
> image
> So the problem you're having is that if a user supplies `run_time=x` to the Simulation and then makes a `t` coordinate grid for the custom time source dataset using `np.linspace(0, run_time, N)`, the validator will sometimes complain due to numerical precision? If so, maybe one option is to relax the validator for the [0] and [-1] values by -fp_eps and +fp_eps (opposite direction as current?) and then make sure to also apply this on the solver side?

I changed source.py to also account for these edge cases, and produce the end values instead of nan.

Can you look at the most recent version? I think 0 tolerance at the start time and dt tolerance at the end time is reasonable, since we warn for sampling below dt anyway, it is fair to extrapolate one dt past the stop time.

tylerflex:
Alright momchil pointed me to the backend code and the handling for out of bounds and that looks fine. I thought for a second that we just evaluated the `amp` function server side and injected whatever came out of that. I think from my perspective it looks good to go.

PR #995, opened 2023-07-11
validate ArrayLike doesnt have None or nan values

tylerflex:
should fix issue where `freqs : [null]`  passed validation in UI


tomflexcompute (approved):
Thanks for the fix!

PR #996, opened 2023-07-12
Medium perturbation

dbochkov-flexcompute:
Heat and charge perturbation functionality. Currently, implemented for `Medium` and `PoleResidue`, but this is ready for an initial "big picture" review, to check if anything seems bad/needs a rework. Here's a notebook with demonstrations and explanations to help the review https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/daniil/perturbation-medium/docs/source/notebooks/PerturbationMedium.ipynb. If everything looks more or less alright, I will go ahead and finish some remaining to-do's:
- [ ] do the rest of dispersive models, that is, add `PerturbationDebye`, `PerturbationLorentz`, etc
- [x] add front-end tests
- [x] check all code documentation is in place

tagging @momchil-flex just in case

### DIFF
--- a/tidy3d/components/data/data_array.py
+++ b/tidy3d/components/data/data_array.py
@@ -9,6 +9,7 @@ import h5py
 
 from ...constants import HERTZ, SECOND, MICROMETER, RADIAN
 from ...exceptions import DataError, FileError
+from ..types import Bound
 
 # maps the dimension names to their attributes
 DIM_ATTRS = {
@@ -244,6 +245,76 @@ class SpatialDataArray(DataArray):
     _dims = ("x", "y", "z")
     _data_attrs = {"long_name": "field value"}
 
+    def sel_inside(self, bounds: Bound) -> SpatialDataArray:
+        """Return a new SpatialDataArray that contains the minimal amount data necessary to cover
+        a spatial region defined by ``bounds``.
+
+
+        Parameters
+        ----------
+        bounds : Tuple[float, float, float], Tuple[float, float float]
+            Min and max bounds packaged as ``(minx, miny, minz), (maxx, maxy, maxz)``.
+
+        Returns
+        -------
+        SpatialDataArray
+            Extracted spatial data array.
+        """
+
+        inds_list = []
+
+        for coord, smin, smax in zip(self.coords.values(), bounds[0], bounds[1]):
+
+            length = len(coord)
+
+            # if data does not cover structure at all take the closest index
+            if smax < coord[0]:  # structure is completely on the left side
+
+                # take 2 if possible, so that linear iterpolation is possible
+                comp_inds = np.arange(0, max(2, length))
+
+            elif smin > coord[-1]:  # structure is completely on the right side
+
+                # take 2 if possible, so that linear iterpolation is possible
+                comp_inds = np.arange(min(0, length - 2), length)
+
+            else:
+                if smin < coord[0]:
+                    ind_min = 0
+                else:
+                    ind_min = max(0, (coord >= smin).argmax().data - 1)
+
+                if smax > coord[-1]:
+                    ind_max = length - 1
+                else:
+                    ind_max = (coord >= smax).argmax().data
+
+                comp_inds = np.arange(ind_min, ind_max + 1)
+
+            inds_list.append(comp_inds)
+
+        return self.isel(x=inds_list[0], y=inds_list[1], z=inds_list[2])
+
+    def does_cover(self, bounds: Bound) -> bool:
+        """Check whether data fully covers specified by ``bounds`` spatial region.
+
+
+        Parameters
+        ----------
+        bounds : Tuple[float, float, float], Tuple[float, float float]
+            Min and max bounds packaged as ``(minx, miny, minz), (maxx, maxy, maxz)``.
+
+        Returns
+        -------
+        bool
+            Full cover check outcome.
+        """
+
+        return all(
+            coord[0] <= smin and coord[-1] >= smax
+            for coord, smin, smax in zip(self.coords.values(), bounds[0], bounds[1])
+        )
+
 
 class ScalarFieldDataArray(DataArray):
     """Spatial distribution in the frequency-domain.
@@ -447,6 +518,33 @@ class TriangleMeshDataArray(DataArray):
     _data_attrs = {"long_name": "surface mesh triangles"}
 
 
+class HeatDataArray(DataArray):
+    """Heat data array.
+
+    Example
+    -------
+    >>> T = [0, 1e-12, 2e-12]
+    >>> td = HeatDataArray((1+1j) * np.random.random((3,)), coords=dict(T=T))
+    """
+
+    __slots__ = ()
+    _dims = "T"
+
+
+class ChargeDataArray(DataArray):
+    """Charge data array.
+
+    Example
+    -------
+    >>> n = [0, 1e-12, 2e-12]
+    >>> p = [0, 3e-12, 4e-12]
+    >>> td = ChargeDataArray((1+1j) * np.random.random((3,3)), coords=dict(n=n, p=p))
+    """
+
+    __slots__ = ()
+    _dims = ("n", "p")
+
+
 DATA_ARRAY_TYPES = [
     SpatialDataArray,
     ScalarFieldDataArray,
@@ -465,5 +563,7 @@ DATA_ARRAY_TYPES = [
     TimeDataArray,
     FreqModeDataArray,
     TriangleMeshDataArray,
+    HeatDataArray,
+    ChargeDataArray,
 ]
 DATA_ARRAY_MAP = {data_array.__name__: data_array for data_array in DATA_ARRAY_TYPES}
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -19,12 +19,13 @@ from .data.dataset import PermittivityDataset
 from .data.data_array import SpatialDataArray, ScalarFieldDataArray, DATA_ARRAY_MAP
 from .viz import add_ax_if_none
 from .geometry import Geometry
-from .validators import validate_name_str
+from .validators import validate_name_str, validate_parameter_perturbation
 from ..constants import C_0, pec_val, EPSILON_0, LARGE_NUMBER, fp_eps
 from ..constants import HERTZ, CONDUCTIVITY, PERMITTIVITY, RADPERSEC, MICROMETER, SECOND
 from ..exceptions import ValidationError, SetupError
 from ..log import log
 from .transformation import RotationType
+from .parameter_perturbation import ParameterPerturbation
 
 
 # evaluate frequency as this number (Hz) if inf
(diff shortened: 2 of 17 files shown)


tidy3d/components/parameter_perturbation.py
```diff
@@ -0,0 +1,722 @@
...
+                    f"is outside of 'ChargePerturbation.electron_range' = {self.electron_range}"
+                )
+
+        if self.hole_range:
+            h_min, h_max = self.hole_range
+
+            # if temperature range not present just return original function
+            if np.any(hole_density < h_min) or np.any(hole_density > h_max):
+                log.warning(
+                    "Hole density values passed to 'ChargePerturbation.sample()'"
```
tylerflex, line 354: noticed some inconsistency regarding periods at end of messages and docstring descriptions. let's try to add them to all.

tidy3d/components/parameter_perturbation.py
```diff
@@ -0,0 +1,722 @@
...
+
+        return ax
+
+
+class LinearHeatPerturbation(HeatPerturbation):
+    """Linear heat perturbation."""
+
+    temperature_ref: pd.NonNegativeFloat = pd.Field(
+        ...,
+        title="Reference temperature",
```
tylerflex, line 202: I think some of the docstrings and descriptions could use a bit more expanding on. To someone not familiar with this, it could be hard to understand what these classes and their fields do. For example, the docstring "Linear heat perturbation" could be "Specification describing how heat quantities affect optical medium properties." and the description of "reference temperature" could be something like "temperature at which the thermo optic coefficient is valid. Away from this temperature ..."

Your notebook has a lot of really useful descriptions, maybe some can be copied into the docs too.

tidy3d/components/parameter_perturbation.py
```diff
@@ -0,0 +1,722 @@
...
+        description="Reference temperature.",
+        units=KELVIN,
+    )
+
+    coeff: Union[float, Complex] = pd.Field(
+        ...,
+        title="Thermo-optic Coefficient",
+        description="Thermo-optic Coefficient.",
+    )
```
tylerflex, line 211: units? seems a lot of the other ones might be missing them as well

tidy3d/components/parameter_perturbation.py
```diff
@@ -0,0 +1,722 @@
...
+        return self.perturbation_values.interp(n=e_clip, p=h_clip).drop_vars(["n", "p"])
+
+    @cached_property
+    def is_complex(self) -> bool:
+        """Whether perturbation is complex valued."""
+        return np.iscomplexobj(self.perturbation_values.data)
+
+
+ChargePerturbationType = Union[LinearChargePerturbation, CustomChargePerturbation]
+
```
tylerflex, line 611: remove extra newlines?

tidy3d/components/simulation.py
```diff
@@ -2787,3 +2790,142 @@ def allow_gain(self) -> bool:
             elif medium.allow_gain:
                 return True
         return False
+
+    @staticmethod
+    def _sel_data_inside_bounds(data: SpatialDataArray, bounds: Bound) -> Tuple[bool, SpatialDataArray]:
+
```
tylerflex, line 2796: docstring

tidy3d/components/simulation.py
```diff
@@ -2787,3 +2790,142 @@ def allow_gain(self) -> bool:
             elif medium.allow_gain:
                 return True
         return False
+
+    @staticmethod
+    def _sel_data_inside_bounds(data: SpatialDataArray, bounds: Bound) -> Tuple[bool, SpatialDataArray]:
```
tylerflex, line 2795: a few things that might help make this more 

1. Maybe this should be a DataArray method? to avoid Simulation getting even larger?
2. does it make sense to separate `is_full_cover` and `sel_data_inside_bounds` into two functions? for example `sel_data_inside_bounds` does the data selection and `is_full_cover` just looks at bounds compared to self's coordinates?

tidy3d/components/simulation.py
```diff
@@ -2787,3 +2790,142 @@ def allow_gain(self) -> bool:
...
+
+        return is_full_cover, data.isel(x=inds_list[0], y=inds_list[1], z=inds_list[2])
+
+    def apply_perturbations(
+        self,
+        temperature: SpatialDataArray = None,
+        electron_density: SpatialDataArray = None,
+        hole_density: SpatialDataArray = None,
+    ) -> Simulation:
+        """Apply heat and/or charge data to perturbation mediums contained in the Simulation
```
tylerflex, line 2850: let's make sure the users know that this returns a copy, so they dont do
```py
sim.apply_perturbations(...)
```
and not realize why things didn't change.  maybe instead of "Apply heat and/or charge data to ..." we should word it as "Return a copy of the simulation with heat and/or charge data applied to ...

tidy3d/components/simulation.py
```diff
@@ -2787,3 +2790,142 @@ def allow_gain(self) -> bool:
...
+                # for each structure select a minimal subset of data that covers it
+                t_cover, t_sub = self._sel_data_inside_bounds(temperature, bounds)
+                e_cover, e_sub = self._sel_data_inside_bounds(electron_density, bounds)
+                h_cover, h_sub = self._sel_data_inside_bounds(hole_density, bounds)
+
+                # check provided data fully cover structure
+                if not np.all([t_cover, e_cover, h_cover]):
+                    log.warning(
+                        "Provided temperature, electron density, and/or hole density data does not "
+                        f"fully cover structure {s_ind}."
```
tylerflex, line 2900: maybe
```
"fully cover 'structures[{s_ind}]'."
```

tidy3d/components/simulation.py
```diff
@@ -2787,3 +2790,142 @@ def allow_gain(self) -> bool:
...
+
+        # For each structure that contains HeatSpecSolid sample temperature field and obtain
+        # perturbed medium as a CustomMedium. Pack it into a regular Structure.
+        # Currently we perform separate interpolation inside the bounding box of each HeatStructure,
+        # a more efficient approach could be interpolating once on the entire grid and then
+        # selecting subsets of points corresponding to each structure
+        new_structures = []
+        for s_ind, s in enumerate(structures):
+            med = s.medium
+            if isinstance(med, AbstractPerturbationMedium):
```
tylerflex, line 2882: could this be factored into a `Structure.apply_perturbations` method? to simplify?

And for the `Simulation.medium` part, call `Medium.apply_perturbations`?

tidy3d/plugins/dispersion/fit.py
```diff
@@ -333,6 +333,7 @@ def _make_medium(self, coeffs):
     def _fit_single(
         self,
         num_poles: int = 3,
+        guess = None,
```
tylerflex, line 336: do we want to add this to the user-facing `fit` method?

tylerflex (changes requested):
Thanks Daniil, this is a ton of work and I think it is really well organized and covers everything I can think of very well. Most of my comments are just refining it for public consumption. The notebook is also extremely useful and will be invaluable. There were a few spelling issues, I recommend this https://pypi.org/project/jupyterlab-spellchecker/ to catch some of them before we do a more thorough proofread.

Thanks!

tidy3d/components/parameter_perturbation.py
```diff
@@ -0,0 +1,722 @@
...
+        Returns
+        -------
+        Union[ArrayLike[float], ArrayLike[Complex], SpatialDataArray]
+            Sampled perturbation value(s).
+        """
+
+        t_range = self.temperature_range
+        temperature_clip = np.clip(temperature, t_range[0], t_range[1])
+        return self.perturbation_values.interp(T=temperature_clip).drop_vars("T")
```
weiliangjin2021, line 305: shall we have "linear" and "nearest" option here?

weiliangjin2021 (commented):
Overall looks quite good! A few small comments.

tidy3d/components/medium.py
```diff
@@ -2927,6 +2928,303 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
...
+        hole_density : SpatialDataArray = None
+            Hole density field data.
+
+        Returns
+        -------
+        Union[Medium, CustomMedium]
+            Medium specification after application of heat and/or charge data.
+        """
+
+        if all([x is None for x in [temperature, electron_density, hole_density]]):
```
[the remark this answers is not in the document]
dbochkov-flexcompute: it seems to have issues when elements are numpy arrays 
```
>>> any([np.array([1, 2]), np.array([3, 4])])
  File "<stdin>", line 1, in <module>
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
thus, decided to use explicit `is not None`

tidy3d/components/medium.py
```diff
@@ -2927,6 +2928,303 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
...
+        title="Permittivity Perturbation",
+        description="List of heat and/or charge perturbations to permittivity.",
+        units=CONDUCTIVITY,
+    )
+
+    @pd.validator("permittivity_perturbation", always=True, allow_reuse=True)
+    def warn_permittivity_range(cls, val, values):
+        """Warn if permittivity can become less than one due to heat/charge perturbations."""
+
+        if val:
```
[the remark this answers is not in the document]
dbochkov-flexcompute: good idea, implemented

tidy3d/components/parameter_perturbation.py
dbochkov-flexcompute: the units are actually unknown here because it can be applied to any parameter, but I set it to `1/K` as in case of a dimensionless parameter

tidy3d/components/simulation.py
dbochkov-flexcompute: imlpemented

tidy3d/components/simulation.py
dbochkov-flexcompute: Agree. I also renamed it to `.perturbed_mediums_copy()` by analogy to `.updated_copy()`

tidy3d/plugins/dispersion/fit.py
dbochkov-flexcompute: added

tidy3d/components/simulation.py
dbochkov-flexcompute: it could be, but we would need to pass simulation bounds `sim_bounds` to it and method would have to return two values: new structure + flag whether coverage was full. So, maybe it fine to keep it all exposed

tidy3d/components/data/data_array.py
```diff
@@ -447,6 +447,33 @@ class TriangleMeshDataArray(DataArray):
     _data_attrs = {"long_name": "surface mesh triangles"}
 
 
+class HeatDataArray(DataArray):
+    """Heat data array.
+
+    Example
+    -------
+    >>> T = [0, 1e-12, 2e-12]
+    >>> td = HeatDataArray((1+1j) * np.random.random((3,)), coords=dict(T=T))
```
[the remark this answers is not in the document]
dbochkov-flexcompute: temperature here is a coordinate f=f(T), so values could be real or complex

tidy3d/components/parameter_perturbation.py
```diff
@@ -0,0 +1,722 @@
...
+
+""" Generic perturbation classes """
+
+
+class AbstractPerturbation(ABC, Tidy3dBaseModel):
+    """Abstract class for a generic perturbation."""
+
+    @cached_property
+    @abstractmethod
+    def perturbation_range(self) -> Union[Tuple[float, float], Tuple[Complex, Complex]]:
```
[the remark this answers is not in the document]
dbochkov-flexcompute: perturbations to poles can be complex

tidy3d/components/parameter_perturbation.py
```diff
@@ -0,0 +1,722 @@
...
+class CustomChargePerturbation(ChargePerturbation):
+    """Custom charge perturbation."""
+
+    perturbation_values: ChargeDataArray = pd.Field(
+        ...,
+        title="Petrubation Values",
+        description="2D array of perturbation values.",
+    )
+
+    temperature_range: Tuple[pd.NonNegativeFloat, pd.NonNegativeFloat] = pd.Field(
```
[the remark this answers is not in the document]
dbochkov-flexcompute: correct, just a copy-paste leftover

tidy3d/components/parameter_perturbation.py
```diff
@@ -0,0 +1,722 @@
...
+        return np.min(self.perturbation_values), np.max(self.perturbation_values)
+
+    @pd.root_validator(skip_on_failure=True)
+    def compute_temperature_range(cls, values):
+        """Compute and set temperature range based on provided ``perturbation_values``."""
+        if values["temperature_range"] is not None:
+            log.warning(
+                "Temperature range for 'CustomHeatPerturbation' is calculated automatically "
+                "based on provided 'perturbation_values'. Provided 'temperature_range' will be overwritten."
+            )
```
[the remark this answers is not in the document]
dbochkov-flexcompute: specified range here is not used at all, it always if overwritten by range from data

tidy3d/components/parameter_perturbation.py
dbochkov-flexcompute: yeah, added for completeness

dbochkov-flexcompute:
big cleanup, added many docstrings, added tests, and fixed many loose ends. I am considering adding rest of dispersive models `PerturbationDebye`, `PerturbationLorentz`, etc, later, and merging this without them. What do you think?

tidy3d/components/parameter_perturbation.py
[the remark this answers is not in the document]
weiliangjin2021: in that case, it should be property rather than a field?

tidy3d/components/medium.py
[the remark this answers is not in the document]
tylerflex: oh ok, sounds good

tidy3d/components/simulation.py
tylerflex: ok

tidy3d/components/parameter_perturbation.py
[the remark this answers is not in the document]
weiliangjin2021: i see, then it looks fine.

weiliangjin2021 (approved):

weiliangjin2021:
> big cleanup, added many docstrings, added tests, and fixed many loose ends. I am considering adding rest of dispersive models `PerturbationDebye`, `PerturbationLorentz`, etc, later, and merging this without them. What do you think?

Agree, as long as we can have a good heat example in the current stage.

tidy3d/components/validators.py
```diff
@@ -249,3 +251,88 @@ def _single_frequency_in_range(cls, val: FieldDataset, values: dict) -> FieldDat
...
+                    )
+                ):
+                    if perturb is not None:
+
+                        # check that base values are not none
+                        if base_value is None:
+                            raise SetupError(
+                                f"Provided '{field_name}' contains parameter perturbation  is "
+                                f"provided for a 'None' base value of '{base_field_name}'."
+                            )
```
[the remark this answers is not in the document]
dbochkov-flexcompute: Yeah, this is redundant, removed

tidy3d/components/medium.py
```diff
@@ -2927,6 +2928,256 @@ class CustomAnisotropicMediumInternal(CustomAnisotropicMedium):
...
+    )
+
+    _permittivity_perturbation_validator = validate_parameter_perturbation(
+        "permittivity_perturbation",
+        "permittivity",
+        allowed_real_range=[(1.0, None)],
+        allowed_imag_range=[None],
+        allowed_complex=False,
+    )
+    _conductivity_perturbation_validator = validate_parameter_perturbation(
```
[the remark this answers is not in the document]
dbochkov-flexcompute: so, these validators only warn if conductivity/permittivity can potentially go out of bounds (within the provided/calculated `temperature_range`, `electron_range`, and `hole_range`). The strict validation is left to the corresponding `CustomMedium`. So perhaps it's still ok to warn the user about potentially negative conductivity even if `allow_gain=True`?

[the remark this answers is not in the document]
weiliangjin2021: I see, so it will warn and then error. In that regard, there seems to be missing `allow_gain` in the return of 
```
    return CustomMedium(
            permittivity=permittivity_field,
            conductivity=conductivity_field,
            name=self.name,
            subpixel=True,
        )
```
and it seems better to add `subpixel` as a field of PerturbedMedium so that one can choose to apply subpixel or not. But warn that for a uniform medium, supixel parameter will be ignored.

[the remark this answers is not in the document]
dbochkov-flexcompute: Oh, you're right, thanks for catching that. Actually, `frequency_range` should also be passed. Fixed it.

dbochkov-flexcompute:
if there are no other comments at this point, is it ok to merge this into pre/2.4?

tylerflex:
Yea I've given all of my comments. I think it's ok if everyone else does

weiliangjin2021:
Green light from me.

PR #998, opened 2023-07-13
supports abort task

magiWei:


tidy3d/web/simulation_task.py
```diff
@@ -537,3 +537,11 @@ def get_log(
             verbose=verbose,
             progress_callback=progress_callback,
         )
+
+    def abort(self):
+        """Abort current task from server."""
+        if not self.task_id:
+            raise ValueError("Task id not found.")
+        return http.put("tidy3d/tasks/abort", json={"taskType": "NORMAL", "taskId": self.task_id})
```
momchil-flex, line 545: Can we use "FDTD" instead of "NORMAL"? This is not part of this PR as it affects multiple codebases, but could you make a JIRA ticket if needed? In the future we will have other task types too so let's be more clear about the naming.

momchil-flex (approved):
Looks good, but there are formatting issues again.

tidy3d/web/simulation_task.py
magiWei: "taskType": "NORMAL/FDTD" , taskType is just an input parameter identification bit, and will not affect the code logic of the backend, ’NORMAL‘ can be freely modified to ’FDTD‘

tidy3d/web/webapi.py
```diff
@@ -677,6 +674,27 @@ def delete_old(
     return len(tasks)
 
 
+@wait_for_connection
+def abort(task_id: TaskId) -> TaskInfo:
+    """Abort server-side data associated with task.
```
momchil-flex, line 679: Maybe "Abort the solver execution if task is currently running."

tidy3d/web/simulation_task.py
```diff
@@ -537,3 +537,9 @@ def get_log(
             verbose=verbose,
             progress_callback=progress_callback,
         )
+
+    def abort(self):
+        """Abort current task from server."""
```
momchil-flex, line 542: Same as for fdtd, "Abort the solver execution if task is currently running."

momchil-flex (commented):
What happens if user aborts a task and it is not running, e.g. if they abort a "draft", or if they abort a "success" task? What happens if a user calls the "start" api on an aborted task?

Looks good overall! Could you add a line to the CHANGELOG?

tidy3d/web/webapi.py
magiWei: Updated CHANGELOG

tidy3d/web/simulation_task.py
magiWei: Updated CHANGELOG

tidy3d/web/simulation_task.py
magiWei: changed NORMAL to FDTD

PR #1000, opened 2023-07-13
fix adjoint plugin with complex-valued permittivity inputs

tylerflex:
context: a user was constructing a complex-valued permittivity array and passing it to JaxCustomMedium. Was getting some obscure jax error saying it couldn't multiply float and complex. Seemed to be coming from the backdrop step and in the user's setting up of the complex array. After digging around a while, I realized it was because the data in the vjp of JaxCustomMedium has taken the real part. This is not only ignoring the gradient w.r.t. the complex part but also making the values dtype of float, which I guess was causing this issue. This PR fixes both things.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Support for differentiating with respect to `JaxMedium.conductivity`.
-
 - Validating that every surface (unless excluded in ``exclude_surfaces``) of a 3D ``SurfaceIntegrationMonitor`` (flux monitor or field projection monitor) is not completely outside the simulation domain.
 
 ### Changed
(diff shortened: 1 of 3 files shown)


momchil-flex:
Looks good, is this a confusion because in our usual by-hand adjoint we would take the real part of the (e * e_adj), but this is more just a consequence of the typical objective function (e.g. something like |e_output|**2), while the correct vjp shouldn't be taking that real part, and if traced correctly it will just come out in the end if needed, e.g. something like vjp + vjp* will happen somewhere for real-valued FOM-s?

tylerflex:
Kind of, so the objective function is still real valued. The way to think about is is that our typical adjoint equation is something like

```
dJ/dp = real[e_fwd * deps/dp e_adj]
```

if `eps = a + ib` is complex then we want to solve for `dJ/da` and `dJ/db` and add those contributions separately

```
dJ/dp_a = real[e_fwd * da/dp * e_adj]
dJ/dp_b = real[i * e_fwd * db/dp * e_adj] = -imag[e_fwd * db/dp * e_adj]
```

essentially, the vjp for the real and imaginary parts of `eps` can just be found by taking the real and imaginary parts of `e_fwd * e_adj` and storing those in the vjp for the complex-valued eps.

This also reminds me that I need that minus sign, which I'll add now.

PR #1010, opened 2023-07-17
fix migrate script in webapi

tylerflex:


tylerflex:
@tomflexcompute this should fix that user's error with migration. I will do a 2.3.2 patch today so I think let's try to get this in just to avoid it in the future if it comes up again.

tomflexcompute (approved):
Sounds good! Thanks.

PR #1014, opened 2023-07-17
added issue 807 functions with basic tests

alec-flexcompute:
Re-pulling to add Tyler as reviewer


tidy3d/plugins/mode/mode_solver.py
```diff
@@ -579,6 +579,104 @@ def to_mode_solver_monitor(self, name: str) -> ModeSolverMonitor:
...
+            from the ModeSolver instance and the method inputs.
+        """
+
+        mode_source = self.to_source(mode_index=mode_index, direction=direction, source_time=source_time)
+        new_sources = self.simulation.sources
+        if not new_sources:
+            new_sources = [mode_source]
+        else:
+            new_sources = new_sources + tuple([mode_source])
+        newSim = self.simulation.copy(update=dict(sources=new_sources))
```
tylerflex, line 614: in python, the convention we use (from [PEP8](https://peps.python.org/pep-0008/#method-names-and-instance-variables)) is to use lowercase + underscores, eg "`new_sim`" instead of camel case "`newSim`".

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -579,6 +579,104 @@ def to_mode_solver_monitor(self, name: str) -> ModeSolverMonitor:
...
+
+        return newSim
+
+    def add_monitor(
+        self,
+        mode_monitor_freqs: List[float],
```
tylerflex, line 620: Let's keep these arguments to be called the exact same as the wrapped methods, eg `freqs` and `name`. I think it might be confusing to change them.

For example, a user could have these two lines in their code
```py
monitor = mode_solver.to_monitor(freqs=freqs, name=name)
sim = mode_solver.simulation.updated_copy(monitors=[monitor])
```

and wouldn't need to look up the new argument names and instead just write
```
sim = mode_solver.sim_with_monitor(freqs=freqs, name=name)
```

tylerflex (changes requested):
Thanks @alec-flexcompute ! This is great and will be really useful. One last thing to note at the end of this process is that we want to have all of these changes squashed into a single commit so that we don't add a bunch of commits to tidy3d for a single feature. The reason is to keep our git history organized so we can quickly review all changes.

For this we use a feature of git called "rebasing" so this is an opportunity for you to learn more about that. We have an explanation in our wiki [here](https://www.notion.so/flexcompute/Git-Rebasing-6a7b343ee1cf4ca28fdad0a870354eee). Could you give that a read as well as this [article](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase), which is super useful for explaining what git rebasing is? You can then try to squash all of your commits into one using a rebase. I'd recommend making a backup of your code just in case, perhaps by checking out a backup branch copy of this one, because rebasing can't be undone easily. 

Feel free to take your time with this, it's mostly an opportunity for you to learn some git skills and get more familiar with that as it's a big part of our development process and is super useful in general.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -579,6 +579,92 @@ def to_mode_solver_monitor(self, name: str) -> ModeSolverMonitor:
...
+
+        Parameters
+        ----------
+        name : str
+            Name of the monitor.
+
+        Returns
+        -------
+        :class:`.Simulation`
+            Copy of the simulation with a mode monitor with specifications taken
```
tylerflex, line 660: let's update this to say :class:`.ModeSolverMonitor` instead of "mode monitor"

Let's also do this in the 2nd sentence of the docstring of this method.

And same idea for the other two methods.

Adding this makes a link to the class definition in the API, which is convenient for people who want to use this.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -579,6 +579,104 @@ def to_mode_solver_monitor(self, name: str) -> ModeSolverMonitor:
...
+        newSim = self.simulation.copy(update=dict(sources=new_sources))
+
+        return newSim
+
+    def add_monitor(
+        self,
+        mode_monitor_freqs: List[float],
+        mode_monitor_name: str,
+        ) -> Simulation:
+
```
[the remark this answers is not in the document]
tylerflex: if the black / pylons didn't complain about that extra newline (620) then let's just delete it manually to be consistent with the rest of our code. thanks

tylerflex (commented):
Thanks Alec, could you also add a line in the `CHANGELOG.md`. Basically under the most recent "Added" section, add a sentence or two saying that these three methods were added to the ModeSolver class and what they do.

tidy3d/plugins/mode/mode_solver.py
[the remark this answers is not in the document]
alec-flexcompute: Yeah I reran black/lint and those lines weren't deleted, so I deleted them. They should be fixed now

tidy3d/plugins/mode/mode_solver.py
alec-flexcompute: Ah, so that's what this does. Done and pushed with other lines deleted

tylerflex:
one other thing, when squashing commits into one, we should also rename it before the final merge to be a descriptive commit message. There is a "reword" option in rebasing that can do this for you.

tidy3d/plugins/mode/mode_solver.py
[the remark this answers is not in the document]
tylerflex: yea looks good thanks!

PR #1017, opened 2023-07-18
Updating surface integration validator to only error if all integrati…

momchil-flex:
…on surfaces are outside of the domain

It is actually convenient in some cases for the user to place a box that's partly outside of the domain without having to explicitly exclude the surfaces. The solver works as long as there's any part that's inside the domain. We could also fix the solver to work if everything is outside, but in that case probably the monitor was incorrectly set up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+### Changed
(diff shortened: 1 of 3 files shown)


shashwat-sh (approved):
This looks good but won't we get an out-of-domain error anyway for any component completely outside the domain?

momchil-flex:
The original problem was a 3D monitor that's *larger* than the simulation domain, such that it covers the simulation, but all of its surfaces are outside of it. This doesn't trigger the usual warning. This update would error in this case, as well as in more esoteric cases where e.g. some of the surfaces do cross the simulation domain, but those are excluded in `exclude_surface`.

PR #1019, opened 2023-07-18
porting to pydantic2 using the new API

joamatab:
this PR ports to pydantic2 new API instead of trying to use the old one

i used this package for this
https://github.com/pydantic/bump-pydantic#installation

which worked much better when porting gdsfactory to pydantic2

### DIFF
--- a/requirements/basic.txt
+++ b/requirements/basic.txt
@@ -9,7 +9,7 @@ h5py>=3.0.0
 rich<12.6.0 # note: rich >= 12.6 adds double progressbars
 matplotlib
 shapely>=2.0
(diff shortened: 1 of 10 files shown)


tylerflex:
Unfortunately, a lot of things weren't working so I spent a while trying to port it but no success. About 2/3 of tests are passing currently but will need a few more days to complete it. Moved this to #1030 as I dont think I can push to your fork directly. Will keep you updated on it.

joamatab:
nice Tyler,

i'll wait until you release tidy3d with pydantic>2 to make a new gdsfactory release so that both packages keep working well together

PR #1024, opened 2023-07-20
fix sign flip in adjoint mode source with backwards direction

tylerflex:
6 hours of work for 2 line fix...
Where do we think this should go? develop -> patch or pre/2.4?


momchil-flex (approved):
I'd say develop and we should probably release a patch.

PR #1030, opened 2023-07-24
pydantic v2 port 

tylerflex:


tylerflex:
@joamatab I have most of the tests finally passing, but as you can see, there were actually a **ton** of manual changes required to get this working.  This was mostly because pydantic v2 makes some very fundamental changes to how model validation and import / export work. We built many of the core parts of Tidy3D around the frameworks and assumptions made in v1 so upgrading has required a significant refactor of a lot of these internals.

In doing this refactor, I was using our tests as a guide. However, even though we have around 90-95% coverage on the front end tests, I still can't be 100% sure that these changes don't break things, especially in our docs, schema generation, or file transfer, and when we actually run Tidy3D in practice. So while most of the tests are passing now, I think we need significant more testing to ensure that users who rely on tidy3d won't get errors as a result of this update.

Because Pydantic v2 is so different from v1, there actually will be API changes for our users, for example calling `obj.dict()` will result in different behavior. (basically all of our users will need to update their scripts to use pydantic v2 features). So there is a backwards compatibility issue there. On the other side of things, I don't actually see much of a benefit to the upgrade vs. the risk of things not working properly and breaking backwards compatibility. If it were up to me, we would probably keep pydantic v1 until we release a new major version of tidy3d 3.0, where many of the internals will likely be refactored anyway.

So I just wanted to get your thoughts specifically on what is the benefit you see by upgrading gdsfactory to pydantic v2? Is it just a matter of keeping things up to date or are there new features you're using? My main motivation for trying to upgrade was simply to retain tidy3d support in gdsfactory, but after seeing all of the changes and changes in API, it will require a lot of testing to ensure that things work properly for our users and also I think some of the changes will result in breaking backwards compatibility, which we can't do unless we do a major release. I think when I started down this path, I expected that upgrading to v2 would be simple but it's unfortunately throwing a wrench in everything so we may need to come up with a better plan.

Curious about your thoughts and tagging @momchil-flex @lucas-flexcompute in case they want to chime in.

joamatab:
Hi Tyler,

i understand it's a hard migration

for us we need to choose already between using pydantic2 for the latest kfactory and fastapi integration or keeping the pydantic1 for tidy3d, so it would be easier to move everything to pydantic2

How about migrating to pydantic2 and adding deprecation warnings to the deprecated methods?

@sebastian-goeldi
@flaport
@tvt173

sebastian-goeldi:
The best way in my opinion is to make this the same way pydantic did it: Use pre-releases. Then gdsfactory (or any projects needing pydantic2) can pin the versions to the pre-release version(s), but any normal `python -m pip install tidy3d` will default to the 2.x stable version.

Additionally, with this you automatically get "betatesters" through gdsfactory and as long as @joamatab promises to put a disclaimer and religously submit bugreports, I think that would be a win-win, as you don't have to settle on a 3.x API yet but still get quite a few testers.

flaport:
From my perspective it seems like a very non-trivial migration too. I'd love to be up to date but most of my libraries use tons of workarounds and additions that only work for pydantic v1.

That being said... you *can* import pydantic v1 from pydantic v2 if I'm not mistaken. I think it's available as `pydantic.v1` or something like that. Maybe this might yield an opportunity to incrementally do the migration and have the dependency be v2 already.

tvt173:
@flaport is right, and you can see this documented [here](https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features)

though if the main goal here is to make the upgrade in a seamless way, without causing too many issues for users, i think @sebastian-goeldi's suggestion is probably the best. issue pre-release versions that the more daring users can help beta-test (i.e. via the gdsfactory plugin) until we're fairly confident everything is working, and then we can push it into the stable releases

you could also combine both suggestions to get the pre-release out ASAP by pinning the trickier portions to `pydantic.v1` for now, and incrementally upgrade them as we go 🙂

joamatab:
Nice 😊 looking forward to the new release of tidy3d with pydantic2

PR #1034, opened 2023-07-26
calculate simulation monitor data size

magiWei:


tidy3d/components/simulation.py
```diff
@@ -989,6 +991,25 @@ def _validate_monitor_size(self) -> None:
                 f"a maximum of {MAX_SIMULATION_DATA_SIZE_GB:.2f}GB are allowed."
             )
 
+    @cached_property
+    def monitors_data_size(self) -> typing.Dict:
```
tylerflex, line 995: can we annotate this? so it would be `Dict[str, float]` I think?

tidy3d/components/simulation.py
```diff
@@ -989,6 +991,25 @@ def _validate_monitor_size(self) -> None:
...
+        """calculate simulation monitors storage size"""
+        tmesh = self.tmesh
+        grid = self.grid
+        data_size = {}
+        for monitor in self.monitors:
```
tylerflex, line 1000: so we already have a very similar method used in a post-init validator here
https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/simulation.py#L959-L990

Seems like the same calculation? Rather than copy it into a new method, can we modify that `_validate_monitor_size` method to use`self.monitors_data_size` and then just iterate over the dict?

This way we only need to implement this once.

tidy3d/components/simulation.py
```diff
@@ -989,6 +991,25 @@ def _validate_monitor_size(self) -> None:
                 f"a maximum of {MAX_SIMULATION_DATA_SIZE_GB:.2f}GB are allowed."
             )
 
+    @cached_property
+    def monitors_data_size(self) -> typing.Dict:
+        """calculate simulation monitors storage size"""
```
tylerflex, line 996: Maybe it would be good to be a little more specific. 
```
"""Dictionary mapping monitor names to their estimated storage in bytes."""
```
or something

tidy3d/components/simulation.py
tylerflex: also, the convention we use throughout the rest of the code is
```py
from typing import Dict

...

... -> Dict[str, float]:
```
so this might be slightly better for consistency.

tylerflex (changes requested):
Thanks @magiWei ! just a few minor comments

tidy3d/components/simulation.py
momchil-flex: Yeah that was the whole reason I asked for this PR, let's update the validator too @magiWei.

momchil-flex (approved):
Looks good after addressing Tyler's comments. Thanks!

tidy3d/components/simulation.py
magiWei: done

tidy3d/components/simulation.py
magiWei: changed.

tylerflex:
Thanks @magiWei, looks good now. I will merge this into pre/2.4. Could you add a changelog item to "added" and then squash commits again into one?
```
- `Simulation.monitors_data_size` mapping from from monitor name to its data size in bytes.
```

tylerflex:
actually, in the interest of time, I made the change myself. Will merge this after the tests pass, thanks!

magiWei:
> actually, in the interest of time, I made the change myself. Will merge this after the tests pass, thanks!

Thanks Tyler.

PR #1041, opened 2023-07-31
fix trianglemesh custom_dataset bug

tylerflex:
The `custom_datasets` had the wrong logic. It was looping through geometry group items but not checking if they were TriangleMesh before grabbing the datasets. Was breaking Anderson localization (and probably metalens) notebooks.

For now, omitting changelog item since this bug was introduced in [this commit ](https://github.com/flexcompute/tidy3d/commit/264a745ccfc10b5ae7ed90b62b43a018b096fdc1) and doesn't affect develop.

### DIFF
--- a/tests/test_components/test_geometry.py
+++ b/tests/test_components/test_geometry.py
@@ -579,3 +579,23 @@ def test_custom_surface_geometry():
     )
     _, ax = plt.subplots()
     _ = sim.plot(y=0, ax=ax)
+
+
+def test_geo_group_sim():
+
+    geo_grp = td.TriangleMesh.from_stl("tests/data/two_boxes_separate.stl")
+    geos_orig = list(geo_grp.geometries)
+    geo_grp_full = geo_grp.updated_copy(geometries=geos_orig + [td.Box(size=(1, 1, 1))])
+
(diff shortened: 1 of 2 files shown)


caseyflex (approved):
looks good, thanks for catching and fixing this. Is there a test we can add to make sure the custom_datasets are working correctly? I guess in this case it would error whenever it was called for a GeometryGroup containing geometries that weren't TriangleMesh.

tylerflex:
Fair enough, I added a test but I'm seeing strange behavior. seems the custom_datasets is not what I'd expect. With two TriangleMesh geometries, I see 4 datasets, each of which is a tuple and two of which just contain `('type', 'TriangleMeshDataset')`. Could you look into what's going on here? I'm kind of fuzzy on the workings of the triangle mesh.

tylerflex:
Yea that does seem to do it, although I guess it's also wrong in `develop` then?

tylerflex:
oh wait, I think it's correct in develop actually since it's inside of a list comprehension.

caseyflex:
Makes sense because the notebook tests would have caught it before release
I suppose.

PR #1043, opened 2023-07-31
Improve docstrings for Gaussian beam sources to clarify the location of the waist

shashwat-sh:


tylerflex (approved):

xin-flex (changes requested):
"behind" or "in front of" are still a bit ambiguous (what is the view point?). Is it possible to use words like "to the + direction of" instead?

shashwat-sh:
> "behind" or "in front of" are still a bit ambiguous (what is the view point?). Is it possible to use words like "to the + direction of" instead?

@xin-flex thanks, updated based on the comment

shashwat-sh:
Can I go ahead and merge this?

Issue #1022, opened 2023-07-20
Add Palik models to material library


momchil-flex:

momchil-flex:
Fixed in #1047

Issue #1021, opened 2023-07-20
Replace some of the material library models with better ones found with the new fitter


momchil-flex:

tomflexcompute:
@e-g-melo I think this issue can be closed now?

momchil-flex:
@e-g-melo could you clarify status of this and #1022

e-g-melo:
They were already merged into 2.4

e-g-melo:
We can close them

momchil-flex:
fixed in #1047 

Note that you can put a comment that includes `fixes #issue_number` in a PR to automatically close an issue.

PR #1047, opened 2023-08-01
Library updated using the fast fitter and inclusion of Palik materials

e-g-melo:
Issues #1021 and #1022.

- A lot of materials were updated and improved using the new fast fitter.
- Inclusion of Palik materials: Si, SiO2, GaAs, Ge, InP, and InAs.
- The Palik materials were split into lossless and lossy wavelength ranges to improve the fits.


tomflexcompute (approved):
Thanks @e-g-melo . I randomly tested a few and they all look good. The inclusions of palik Si and SiO2 are going to be very welcomed by users in the PIC industry!

tomflexcompute:
By the way could you add it to the change log? Something like "improved fitting for materials in the material library. Added a number of common semiconductor materials such as Si, GaAs, and InP from Palik to the material library."

tidy3d/material_library/material_library.py
```diff
@@ -144,17 +132,21 @@ class MaterialItem2D(MaterialItem):
 
 Ag_RakicLorentzDrude1998 = VariantItem(
```
weiliangjin2021, line 133: The series of `RakicLorentzDrude` model is based on the analytical model from the paper. There is probably no need to update them. What improvement have you observed with the new poles?

tidy3d/material_library/material_library.py
```diff
@@ -185,17 +188,21 @@ class MaterialItem2D(MaterialItem):
 
 Ag_Yang2015Drude = VariantItem(
```
weiliangjin2021, line 189: Similarly, the original parameters seem to come from Drude model directly. Could you or @tomflexcompute who have access to the paper check if the paper provides a Drude model?

tidy3d/material_library/material_library.py
```diff
@@ -273,10 +284,20 @@ class MaterialItem2D(MaterialItem):
...
+                (22.52074625806345 + 1j * 2.0304989609387856e16),
+            ),
+            (
+                (-69676151586035.45 - 1j * 1701334260069428.2),
+                (0.010443365309052358 + 1j * 0.018058949307349314),
+            ),
+            (
+                (-0.30731881177988135 - 1j * 68198826070986.79),
+                (0.32823363278906764 + 1j * 64788884158076.12),
+            ),
```
weiliangjin2021, line 300: This is an interesting result. In the updated version, 2 out of the 3 poles resemble the original 2 poles. I think the original result is based on the Sellmeier formula in the page https://refractiveindex.info/?shelf=main&book=AlAs&page=Fern . What improvement do you see with the new poles?

weiliangjin2021 (commented):
Looks quite good. One general comment is that if the material is based on an analytical model provided in the paper, we should use the poles referred directly from the analytical modal. This is usually indicated if the variant name has the word "Drude", "Lorentz", "Sellmeier", etc. If you find the pole is imaginary, they are likely based on Sellmeier model.

tidy3d/material_library/material_library.py
e-g-melo: I've not updated all of the `RakicLorentzDrude` materials as many of them really shows no relevant improvements. But the weighted RMS error of some ones was improved by at least 1 order or magnitude using the new fitter. 

For example:

Al-RakicLorentzDrude1998
Al-RakicLorentzDrude1998

Ag-RakicLorentzDrude1998
Ag-RakicLorentzDrude1998

tidy3d/material_library/material_library.py
tomflexcompute: Yes these poles are directly converted from the Drude model with the parameter values taken from the paper directly.

weiliangjin2021:
Here is what I would suggest:

- A list of materials to remain in its original form
  - AlAs_FernOnton1971: Sellmeier, eps_inf in its original form is likely to be off
  - GaAs_Skauli2003: Sellmeier
  - Ge_Icenogle1976: Sellmeier, eps_inf in its original form is likely to be off
  - InP_Pettit1965: Sellmeier, eps_inf in its original form is likely to be off
  - Si3N4_Luke2015: Sellmeier
  - Si3N4_Philipp1973: Sellmeier
  - Y2O3-Nigara1968: Sellmeier
  - YAG_Zelmon1998: Sellmeier
  - cSi_SalzbergVilla1957: Sellmeier
  - cSi_Li1993_293K: Sellmeier
  - All variants that contains "PMLStable" keyword

e-g-melo:
Thanks Weiliang! I've changed all these materials to their original forms.

weiliangjin2021 (approved):
Green light from me to merge once the commits are squashed.

PR #1051, opened 2023-08-03
Changed CustomSourceTime to automatically modulate at freq0

caseyflex:
Changed `CustomSourceTime` to accept a (real or complex) envelope, and automatically multiply it by `e^{i \omega_0 t}`. This particular complex modulation is needed for it to accept the phase shift from the different source types (e.g. `e^{i k_0 \cdot r}` for `PlaneWave`). Currently, one could put in `cos(\omega_0 t)` time dependence (or even something stranger), which leads to inaccuracies when attempting the phase shift. Having automatic complex modulation removes these inaccuracies.

### DIFF
--- a/tidy3d/components/source.py
+++ b/tidy3d/components/source.py
@@ -296,22 +296,44 @@ class ContinuousWave(Pulse):
 
 
 class CustomSourceTime(Pulse):
-    """Custom source time dependence, real or complex valued.
+    """Custom source time dependence consisting of a real or complex envelope
+    modulated at a central frequency, as shown below.
+
+    Note
+    ----
+    .. math::
+
+        amp\\_time(t) = amplitude \\cdot \\ 
+                e^{i \\cdot phase - 2 \\pi i \\cdot freq0 \\cdot t} \\cdot \\
+                envelope(t - offset / (2 \\pi \\cdot fwidth))
 
     Note
     ----
     The source time dependence is linearly interpolated to the simulation time steps.
     To ensure that this interpolation does not introduce artifacts, it is necessary
     to use a sampling rate that is sufficiently fast relative to the simulation time step.
+    The source should also ideally start at zero and ramp up smoothly.
+    The first and last values of the envelope will be used for times that are out of range
+    of the provided data.
 
     Example
     -------
-    >>> cst = td.CustomSourceTime.from_values(freq0=1, fwidth=0.1,
-    >>>     values=np.linspace(0, 9, 10), dt=0.1)
+    >>> cst = CustomSourceTime.from_values(freq0=1, fwidth=0.1,
+    ...     values=np.linspace(0, 9, 10), dt=0.1)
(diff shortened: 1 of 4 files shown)


tidy3d/components/source.py
```diff
@@ -296,22 +296,35 @@ def amp_time(self, time: float) -> complex:
...
     Example
     -------
     >>> cst = td.CustomSourceTime.from_values(freq0=1, fwidth=0.1,
     >>>     values=np.linspace(0, 9, 10), dt=0.1)
     """
 
+    offset: float = pydantic.Field(
+        0.0,
+        title="Offset",
+        description="Time delay of the maximum value of the "
```
tylerflex, line 319: so now the source amplitude time dependences is `f(t) = real[ source_time_dataset(t) * exp(i 2 pi freq0 t) ] `? And the offset just shifts `t` -> `t - offset`? the description should probably be modified if so.

tylerflex: guess the offset is actually dimensionless so you need to incorporate fwidth but same idea?

tidy3d/components/source.py
```diff
@@ -373,18 +388,34 @@ def amp_time(self, time: float) -> complex:
...
+        # mask times that are out of range
+        mask = (time_shifted < min(data_times)) | (time_shifted > max(data_times))
+
+        # get envelope
+        envelope = np.zeros(len(time_shifted), dtype=complex)
+        values = self.source_time_dataset.values
+        envelope[mask] = values.sel(t=time_shifted[mask], method="nearest").to_numpy()
+        envelope[~mask] = values.interp(t=time_shifted[~mask]).to_numpy()
+
+        # modulation, phase, amplitude
```
tylerflex, line 412: I think it would be useful to have a formula of the pulse amplitude as a function of these variables in the docstring description

tylerflex (changes requested):
Looks fine to me, just a couple comments about clarifying in the docstrings what the source amplitude looks like.

tidy3d/components/source.py
momchil-flex: Actually the offset only shifts the envelope but not the carrier wave. The carrier wave can be shifted by `source_time.phase`.

momchil-flex (approved):
I think this makes a lot of sense now, the extra parameters like freq0 and offset are no longer decoupled from the actual signal. Only `fwidth` still kind of is right? But I guess that's fine.

caseyflex:
> I think this makes a lot of sense now, the extra parameters like freq0 and offset are no longer decoupled from the actual signal. Only `fwidth` still kind of is right? But I guess that's fine.

yeah, that was my thinking. fwidth is still just an estimate of the bandwidth, which should be fine

tylerflex (approved):

tylerflex:
feel free to rebase and merge when you are ready thanks!

PR #1053, opened 2023-08-03
pylint -> ruff

tylerflex:
change linter, why?
* ruff is way faster (< 1 second as opposed to > 10 seconds for pylint)
* plays better with pydantic v2, allowing us to update.


pyproject.toml
```diff
@@ -20,4 +20,8 @@ exclude = '''
   | build
   | dist
 )/
-'''
\ No newline at end of file
+'''
+
+[tool.ruff]
+ignore-init-module-imports = true
+typing-modules = ["tidy3d.components.types"] # without this Literal["something fails"]
```
lucas-flexcompute, line 27: Maybe add `line-length = 100` here as well, so that ruff LSP works seamlessly with this configuration.

tidy3d/components/geometry.py
```diff
@@ -35,8 +35,6 @@
     TRIMESH_AVAILABLE = False
 
 try:
-    import networkx  # pylint:disable=unused-import
-    import rtree  # pylint:disable=unused-import
```
lucas-flexcompute, line 39: Without these module imports we won't detect their existence and we'll error out later without further information to the user.

tox.ini
```diff
@@ -71,4 +71,3 @@ commands =
     pytest -rA tests/test_web/test_tidy3d_material_library.py    
     pytest -rA tests/test_web/test_tidy3d_task.py
     pytest -rA tests/test_web/test_webapi.py
```
lucas-flexcompute, line 73: Can't we just `pytest -ra tests` here? Or are we excluding a few tests?

lucas-flexcompute (approved):
+1 for ruff!

test_local.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -e
 
-black .
-python lint.py
+black . --line-length=100
```
momchil-flex, line 4: Technically this is already defined [here](https://github.com/flexcompute/tidy3d/blob/b6427aa9230192ba64ff35f769b8a7f989a7fa76/pyproject.toml#L8-L9) but I guess it doesn't hurt

tidy3d/__init__.py
```diff
@@ -114,3 +114,147 @@ def set_logging_level(level: str) -> None:
...
+    "YeeGrid",
+    "FieldGrid",
+    "Coords1D",
+    "log",
+    "set_logging_file",
+    "set_logging_console",
+    "config",
+    "__version__",
+    "Updater",
+]
```
momchil-flex, line 260: what is this about?

tidy3d/log.py
```diff
@@ -330,15 +330,15 @@ def set_logging_file(
...
+        except:  # noqa: E722
             log.warning("Log file could not be closed")
         finally:
             del log.handlers["file"]
 
     try:
         # pylint: disable=consider-using-with
         file = open(fname, filemode)
-    except:  # pylint: disable=bare-except
+    except:  # noqa: E722
```
momchil-flex, line 341: so pylint disables still work? but there's also another way? how does it work exactly?

momchil-flex (commented):
I don't see why not

pyproject.toml
tylerflex: good catch, added

tidy3d/components/geometry.py
tylerflex: thanks for noticing this, actually ruff did this automatically when I added the `--fix` option but I didn't catch it. added them back

tox.ini
tylerflex: initially we excluded some tests, but after I reorganized the tests, we put the ones we want to exclude in `_test_local` which gets ignored by pytest so yea I think we can safely do this. changed

tidy3d/__init__.py
tylerflex: something ruff was complaining about. __all__ controls what gets imported when you import * from the module. Guess it's good practice to include it

tidy3d/log.py
tylerflex: Another option is to just add ruff in addition to pylint and then talk about pydantic v2 later.. ruff might be nice to have and we could try it out a bit before transitioning completely?

momchil-flex: What is the advantage of leaving pylint but *also* using ruff?

tylerflex: * ruff catches different stuff, seemingly a wider range of issues.
* could test it out before phasing out pylint.
but yea if having two tools just complicates our lives I dont know

tylerflex:
closing as this is done in #1035

Issue #690, opened 2023-02-15
DC Component in Source Spectrum


tomflexcompute:
Currently our source spectrum doesn't contain DC component. As a result, source power is low close to zero frequency, which could lead to noisy results in a broadband simulation. Would be good to have, for example, an option in `GaussianPulse` that adds DC component to the spectrum.

momchil-flex:
@caseyflex I've assigned you here but it is low priority compared to STL and 2D materials. You can work with Tom to figure out how to set things up, probably just an extra kwarg in `GaussianPulse`.

tylerflex:
what's the status on this?

caseyflex:
Was a relatively low priority, but it’s on my to do list on the what’s new
for this month. Should be easy.

PR #1054, opened 2023-08-04
Option to allow DC component in GaussianPulse spectrum

caseyflex:
Per this issue:
https://github.com/flexcompute/tidy3d/issues/690

### DIFF
--- a/tidy3d/components/source.py
+++ b/tidy3d/components/source.py
@@ -256,6 +256,17 @@ class GaussianPulse(Pulse):
     >>> pulse = GaussianPulse(freq0=200e12, fwidth=20e12)
     """
 
+    remove_dc_component: bool = pydantic.Field(
+        True,
+        title="Remove DC Component",
+        description="Whether to remove the DC component in the Gaussian pulse spectrum. "
+        "If ``True``, the Gaussian pulse is modified at low frequencies to zero out the "
+        "DC component, which is usually desirable so that the fields will decay. However, "
+        "for broadband simulations, it may be better to have non-vanishing source power "
+        "near zero frequency. Setting this to ``False`` results in an unmodified Gaussian "
+        "pulse spectrum which can have a nonzero DC component.",
+    )
+
     def amp_time(self, time: float) -> complex:
         """Complex-valued source amplitude as a function of time."""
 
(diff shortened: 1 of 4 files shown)


caseyflex:
Yeah, it modifies the shape of the spectrum a bit too but is specifically
designed to have this effect I believe

tidy3d/components/source.py
```diff
@@ -256,14 +256,28 @@ class GaussianPulse(Pulse):
...
 
         twidth = 1.0 / (2 * np.pi * self.fwidth)
         omega0 = 2 * np.pi * self.freq0
         time_shifted = time - self.offset * twidth
 
-        const = 1j + time_shifted / twidth**2 / omega0
+        if self.allow_dc_component:
+            const = 1
+        else:
+            const = 1j + time_shifted / twidth**2 / omega0
```
tylerflex, line 280: this is just personal preference so feel free to ignore, but I feel it makes it slightly better to add the DC correction as its own block of code kind of like this.

```py
pulse_amp = offset * oscillation * amp

# subtract out DC component
if not self.allow_dc_component:
    pulse_amp *= 1j + time_shifted / twidth**2 / omega0
return pulse_amp
```

tidy3d/components/source.py
```diff
@@ -256,14 +256,28 @@ class GaussianPulse(Pulse):
     >>> pulse = GaussianPulse(freq0=200e12, fwidth=20e12)
     """
 
+    allow_dc_component: bool = pydantic.Field(
```
tylerflex, line 259: Minor point, but I think I'd slightly prefer `remove_dc_component: bool = pydantic.Field(True, ...)` because it's more in line with what we do (add a correction to remove the DC component).

tylerflex (commented):
some pretty minor suggested changes but overall looks good!

caseyflex:
> Thanks @caseyflex looks good to me, just make sure to edit the changelog too with the "remove" convention.

I'd almost rather leave the changelog as it is, since the actual functionality which was added was the ability to leave in the DC component

tylerflex:
Yea good point, might just suggest putting how to do that in the changelog, eg
```
... by setting `remove_dc_component=False` in `GaussianPulse`.
```
or something.

tomflexcompute (approved):
Looks good. Thanks for the implementation!

weiliangjin2021 (approved):

caseyflex:
I introduced an extra factor of 1j in the case where we do not remove the DC component. This extra factor makes it agree with the case where we do remove the DC component in the limit of large omega0.

PR #1055, opened 2023-08-04
Fix to mode solver with negative angle_theta

momchil-flex:
Negative theta-s were not working correctly because of this wrong check.


dbochkov-flexcompute (approved):
probably we should switch some of backend tests to negative angle

PR #1058, opened 2023-08-07
Use pytest's tmp path instead of hard-codded ones

lucas-flexcompute:
Hard-codded paths can be an issue when tests are run in parallel or unordered, particularly in Windows. This commit removes all hard-codded paths from the test suite, replacing them with pytest's tmp_path fixture, which is also automatically cleaned after run.

A few ruff warnings were also cleaned up in the process: mostly unused imports and variables, and warnings caused by a large number of matplotlib figures being created.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -8,7 +8,7 @@ from math import isclose
 import pydantic
 import numpy as np
 import xarray as xr
-import matplotlib.pylab as plt
+import matplotlib.pyplot as plt
 import matplotlib as mpl
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
--- a/tests/_test_local/_test_adjoint_performance.py
+++ b/tests/_test_local/_test_adjoint_performance.py
@@ -1,52 +1,25 @@
 import pytest
 import numpy as np
-import os
 import sys
 from memory_profiler import profile
-import matplotlib.pylab as plt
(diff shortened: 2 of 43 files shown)


tidy3d/components/simulation.py
```diff
@@ -8,7 +8,7 @@
 import pydantic
 import numpy as np
 import xarray as xr
-import matplotlib.pylab as plt
+import matplotlib.pyplot as plt
```
tylerflex, line 11: does this have any practical effects in our plotting?

lucas-flexcompute: No, `plt` is only used in `plt.colorbar(mappable, cax=cax, label=r"$\epsilon_r$")` in this file, and `colorbar` works the same way in both interfaces.

lucas-flexcompute:
> really appreciate this @lucas-flexcompute ! just wondering about the difference between pylab and pyplot and want to make sure it doesn't change our plots in unexpected way, any thoughts on that? or what was the reasoning there?

We don't use any calls that are `pylab`-specific, so the changes should not cause any impact. I made the changes because [the docs](https://matplotlib.org/stable/api/index.html#module-pylab) strongly discourage the use of `pylab`.

tylerflex:
sounds good to me! can you squash the commits into one and then "Rebase and merge" this into 2.4? thanks!

tylerflex (approved):

PR #1061, opened 2023-08-08
2d error intersections_2dbox

alec-flexcompute:


tidy3d/components/geometry.py
```diff
@@ -206,7 +206,7 @@ def intersections_2dbox(self, plane: Box) -> List[Shapely]:
         """
 
         # Verify 2D
-        if plane.size.count(0.0) != 1:
+        if plane.size.count(0.0) != 0:
```
momchil-flex, line 209: I just stumbled on this - it doesn't really make sense to me? If the goal is to verify 2D then the old version was correct. What is the issue/goal of this?

alec-flexcompute: huh that should be == 0. Lemme fix now

caseyflex (changes requested):
Looks good, thanks. Can you add a test for plotting a simulation with sim_size = (0, 0, L) in the 'y=0' direction?

tidy3d/components/geometry.py
caseyflex: The old version was giving an error when I tried to plot simulations with sim_size=(0,0,L), which had previously worked fine.

tylerflex:
@alec-flexcompute just FYI, if you get a merge conflict like 
image
the fix is to check out and pull the most recent version from the destination branch (pre/2.4) and then rebase the PR branch against the destination branch again eg.
```
git checkout pre/2.4
git pull origin pre/2.4
git checkout alec/intersections_2d
git rebase -i pre/2.4
```

alec-flexcompute:
Made changes. Let me know if anything else is needed

tylerflex:
looks like something isn't quite right with the commits. There's a "Merge branch 2.4 ..." could you try pulling the recent pre/2.4 and rebasing / squashing again?

alec-flexcompute:
I think it's good now? I'm not seeing the message after squashing

tylerflex (approved):
looks good to me.

caseyflex (approved):
thanks Alec!

PR #1070, opened 2023-08-16
Rich console width

dbochkov-flexcompute:
Here's how we can adjust width of rich output, so that it should eliminate horizontal scroll bars in web documentation. Perhaps an alternative that we can consider is to define this width in the config, so that it could be changed as needed? What do you think? @tylerflex @momchil-flex

### DIFF
--- a/tidy3d/log.py
+++ b/tidy3d/log.py
@@ -32,6 +32,9 @@ DEFAULT_LOG_STYLES = {
     "CRITICAL": "red bold",
 }
 
+# Width of the console used for rich logging (in characters).
+CONSOLE_WIDTH = 80
+
 
 def _get_level_int(level: LogValue) -> int:
     """Get the integer corresponding to the level string."""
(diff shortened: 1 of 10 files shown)


dbochkov-flexcompute:
So, I realized we already create a console instance inside our logger, and decided just to retrieve it anytime we need a rich console. Do you see any issues with that?

tylerflex:
No specifically. As long as it works properly when running different notebooks for example it should be fine.

tidy3d/log.py
```diff
@@ -32,6 +32,8 @@
     "CRITICAL": "red bold",
 }
 
+CONSOLE_WIDTH = 80
```
tylerflex, line 35: can you just add a comment above this saying that
```py
Width of the console used for rich logging (in characters).
```

tylerflex (approved):
Looks good! Thanks! just a minor comment to explain the `CONSOLE_WIDTH` variable for future reference and also could you add a CHANGELOG item? Thanks!

momchil-flex:
Can you just double check that the logging level and verbose in web functions are still independent from one another. I think they should be but just to make sure.

tylerflex:
I'm excited to see the newly rendered docs after this fix :D

PR #1083, opened 2023-08-18
Heat solver interface

dbochkov-flexcompute:
Since the old PR (https://github.com/flexcompute/tidy3d/pull/896) became very outdated, I've decided to open a new one.

Major changes/additions (directory structure is up for discussion, just whatever seemed to make sense for now):

### `components/heat_spec.py`: 
All medium classes can now have field `heat_spec` which can be either `SolidSpec` or `FuildSpec` for specifying thermal properties. To remind, there are three categories of properties:
1. optic (permittivity, conductivity, ...) - defined using standard `Medium`, `Lorentz`, etc classes
2. thermal (heat conductivity, capacity, ...) - added through field `heat_spec`.
3. response of optic properties to temperature (perturbations) - added when using `PerturbationMedium` and `PerturbationPoleResidue` classes.

And this is how different options interact with heat solver:
- Structures using medium instances with `heat_spec=None` or `heat_spec=FluidSpec()` will **not** be included in solution domain for the heat equation, and they will **not** be converted to custom mediums after applying temperature fields.
- Structures using non-perturbation medium instances with `heat_spec=SolidSpec(...)` will be included in solution domain for heat equation, but they will **not** be converted to custom mediums after applying temperature fields.
- Structures using perturbation medium instances (`PerturbationMedium` and `PerturbationPoleResidue`) with `heat_spec=SolidSpec(...)` will be included in solution domain for heat equation, and they will be converted to custom mediums after applying temperature fields. (the fullest material description)

### `components/scene.py`: 
Carved out geometric only information from `Simulation` into class `Scene` for easier transfer of geometric information between different solvers. Basically, it contains `structures`, `medium`, `center`, `size` + plotting functionality. 
- Point to discuss: plotting of custom media is dependent on simulation grid in `Simulation`. I think we should still have eps plotting in `Scene` class (along with thermal conductivity plotting, and perhaps other properties in future), so I adjusted the plotting of custom medium to use custom medium sample points + structure bounds instead of simulation grid. But now the question is whether `Simulation` in future should just directly borrow this plotting method or to have its own based on actual simulation grid?
- As discussed, for now we will not make any changes to `Simulation`, only added property `Simulation.scene` and class method `Simulation.from_scene(scene: Scene, **kwargs)` to transferring `Scene` between optic/heat simulation.
- Contrary, `HeatSimulation` does have field `scene: Scene` 

### `components/bc_placement.py`: 
This file defines classes for placement of boundary conditions on complex interfaces (`BCPlacementType=[StructureBoundary, StructureStructureInterface, MediumMediumInterface, etc]`). I realized this should be separated from heat solver components because it can be reused for other future solvers.

### `components/heat/`: 
The remaining of heat components are placed in this directory to keep everything tidy (yes, you see what I did here lol) and they include:
- `boundary.py` boundary conditions specifications: `TemperatureBC`, `HeatFluxBC`, `ConvectionBC` + `HeatBoundarySpec`
- `grid.py` heat grid specification: currently, `UniformHeatGrid` and `DistanceHeatGrid`. I can that this can go either way: we can introduce more classes to give the user more control or, contrary, unify everything into one spec so that the user don't have to think about what to select at all.
- `sim_data.py` defines `HeatSimulationData` which is currently simply `HeatSimulation` instance + temperature `SpatialDataArray`
- `simulation.py` defines `HeatSimulation`
- `source.py` heat source: currently only `UniformHeatSource`
- `viz.py` some parameters for plotting heat information

### Notebooks with demonstrations
There are two draft notebooks demonstrating:
- basic heat solver usage + comparison to analytical solutions https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/daniil/heat-solver-draft/docs/source/notebooks/HeatSolverDraft.ipynb
- usage in the context of thermally tuning a ring resonator https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/daniil/heat-solver-draft/docs/source/notebooks/ThermallyTunedRingResonatorDraft.ipynb
These are meant to provide more explanation on how every component is interacting with each other

### additional major things off the top of my head to discuss:
- Currently boundary conditions and heat source placement specifications are done through using structures and medium names. For example, `placement = td.StructureStructureInterface(structures=["box", "sphere"])`. This doesn't seem ideal. We could maybe just pass structures/mediums themselves. But that will increase size of `HeatSimulation` instances. Maybe an alternative is to just record hash of passed structures/mediums? But then it would be harder to point user which structure/medium is missing from a heat simulation.

### DIFF
--- a/tests/test_components/test_scene.py
+++ b/tests/test_components/test_scene.py
@@ -1,0 +1,274 @@
+"""Tests the scene and its validators."""
+import pytest
+import pydantic.v1 as pd
+import matplotlib.pyplot as plt
+
+import numpy as np
+import tidy3d as td
+from tidy3d.components.simulation import MAX_NUM_MEDIUMS
+from ..utils import assert_log_level, log_capture, SIM_FULL
+
+SCENE = td.Scene()
+
+SCENE_FULL = SIM_FULL.scene
+
+
+def test_scene_init():
+    """make sure a scene can be initialized"""
+
+    sim = td.Scene(
+        structures=[
+            td.Structure(
+                geometry=td.Box(size=(1, 1, 1), center=(-1, 0, 0)),
+                medium=td.Medium(permittivity=2.0),
+            ),
+            td.Structure(
+                geometry=td.Box(size=(1, 1, 1), center=(0, 0, 0)),
+                medium=td.Medium(permittivity=1.0, conductivity=3.0),
+            ),
+            td.Structure(
+                geometry=td.Sphere(radius=1.4, center=(1.0, 0.0, 1.0)), medium=td.Medium()
+            ),
+            td.Structure(
+                geometry=td.Cylinder(radius=1.4, length=2.0, center=(1.0, 0.0, -1.0), axis=1),
+                medium=td.Medium(),
+            ),
+        ],
+        medium=td.Medium(permittivity=3.0),
+    )
+
+    _ = sim.mediums
+    _ = sim.medium_map
+    _ = sim.background_structure
+
+
+def test_validate_components_none():
+
+    assert SCENE._validate_num_mediums(val=None) is None
+
+
+def test_plot_eps():
+    ax = SCENE_FULL.plot_eps(x=0)
+    SCENE_FULL._add_cbar_eps(eps_min=1, eps_max=2, ax=ax)
+    plt.close()
+
+
+def test_plot_eps_bounds():
+    _ = SCENE_FULL.plot_eps(x=0, hlim=[-0.45, 0.45])
+    plt.close()
+    _ = SCENE_FULL.plot_eps(x=0, vlim=[-0.45, 0.45])
+    plt.close()
+    _ = SCENE_FULL.plot_eps(x=0, hlim=[-0.45, 0.45], vlim=[-0.45, 0.45])
+    plt.close()
+
+
+def test_plot():
+    SCENE_FULL.plot(x=0)
+    plt.close()
+
+
+def test_plot_1d_scene():
+    s = td.Scene(structures=[td.Structure(geometry=td.Box(size=(0, 0, 1)), medium=td.Medium())])
+    _ = s.plot(y=0)
+    plt.close()
+
+
+def test_plot_bounds():
+    _ = SCENE_FULL.plot(x=0, hlim=[-0.45, 0.45])
+    plt.close()
+    _ = SCENE_FULL.plot(x=0, vlim=[-0.45, 0.45])
+    plt.close()
+    _ = SCENE_FULL.plot(x=0, hlim=[-0.45, 0.45], vlim=[-0.45, 0.45])
+    plt.close()
+
+
+def test_structure_alpha():
+    _ = SCENE_FULL.plot_structures_eps(x=0, alpha=None)
+    plt.close()
+    _ = SCENE_FULL.plot_structures_eps(x=0, alpha=-1)
+    plt.close()
+    _ = SCENE_FULL.plot_structures_eps(x=0, alpha=1)
+    plt.close()
+    _ = SCENE_FULL.plot_structures_eps(x=0, alpha=0.5)
+    plt.close()
+    _ = SCENE_FULL.plot_structures_eps(x=0, alpha=0.5, cbar=True)
+    plt.close()
+    new_structs = [
+        td.Structure(geometry=s.geometry, medium=SCENE_FULL.medium) for s in SCENE_FULL.structures
+    ]
+    S2 = SCENE_FULL.copy(update=dict(structures=new_structs))
+    _ = S2.plot_structures_eps(x=0, alpha=0.5)
+    plt.close()
+
+
+def test_filter_structures():
+    s1 = td.Structure(geometry=td.Box(size=(1, 1, 1)), medium=SCENE.medium)
+    s2 = td.Structure(geometry=td.Box(size=(1, 1, 1), center=(1, 1, 1)), medium=SCENE.medium)
+    plane = td.Box(center=(0, 0, 1.5), size=(td.inf, td.inf, 0))
+    SCENE._filter_structures_plane_medium(structures=[s1, s2], plane=plane)
+
+
+def test_get_structure_plot_params():
+    pp = SCENE_FULL._get_structure_plot_params(mat_index=0, medium=SCENE_FULL.medium)
+    assert pp.facecolor == "white"
+    pp = SCENE_FULL._get_structure_plot_params(mat_index=1, medium=td.PEC)
+    assert pp.facecolor == "gold"
+    pp = SCENE_FULL._get_structure_eps_plot_params(
+        medium=SCENE_FULL.medium, freq=1, eps_min=1, eps_max=2
+    )
+    assert float(pp.facecolor) == 1.0
+    pp = SCENE_FULL._get_structure_eps_plot_params(medium=td.PEC, freq=1, eps_min=1, eps_max=2)
+    assert pp.facecolor == "gold"
+
+
+def test_num_mediums():
+    """Make sure we error if too many mediums supplied."""
+
+    structures = []
+    for i in range(MAX_NUM_MEDIUMS):
+        structures.append(
+            td.Structure(geometry=td.Box(size=(1, 1, 1)), medium=td.Medium(permittivity=i + 1))
+        )
+    _ = td.Scene(
+        structures=structures,
+    )
+
+    with pytest.raises(pd.ValidationError):
+        structures.append(
+            td.Structure(geometry=td.Box(size=(1, 1, 1)), medium=td.Medium(permittivity=i + 2))
+        )
+        _ = td.Scene(structures=structures)
+
+
+def _test_names_default():
+    """makes sure default names are set"""
+
+    scene = td.Scene(
+        size=(2.0, 2.0, 2.0),
+        structures=[
+            td.Structure(
+                geometry=td.Box(size=(1, 1, 1), center=(-1, 0, 0)),
+                medium=td.Medium(permittivity=2.0),
+            ),
+            td.Structure(
+                geometry=td.Box(size=(1, 1, 1), center=(0, 0, 0)),
+                medium=td.Medium(permittivity=2.0),
+            ),
+            td.Structure(
+                geometry=td.Sphere(radius=1.4, center=(1.0, 0.0, 1.0)), medium=td.Medium()
+            ),
+            td.Structure(
+                geometry=td.Cylinder(radius=1.4, length=2.0, center=(1.0, 0.0, -1.0), axis=1),
+                medium=td.Medium(),
+            ),
+        ],
+    )
+
+    for i, structure in enumerate(scene.structures):
+        assert structure.name == f"structures[{i}]"
+
+
+def test_names_unique():
+
+    with pytest.raises(pd.ValidationError):
+        _ = td.Scene(
+            structures=[
+                td.Structure(
+                    geometry=td.Box(size=(1, 1, 1), center=(-1, 0, 0)),
+                    medium=td.Medium(permittivity=2.0),
+                    name="struct1",
+                ),
+                td.Structure(
+                    geometry=td.Box(size=(1, 1, 1), center=(0, 0, 0)),
+                    medium=td.Medium(permittivity=2.0),
+                    name="struct1",
+                ),
+            ],
+        )
+
+
+def test_allow_gain():
+    """Test if simulation allows gain."""
+
+    medium = td.Medium(permittivity=2.0)
+    medium_gain = td.Medium(permittivity=2.0, allow_gain=True)
+    medium_ani = td.AnisotropicMedium(xx=medium, yy=medium, zz=medium)
+    medium_gain_ani = td.AnisotropicMedium(xx=medium, yy=medium_gain, zz=medium)
+
+    # Test simulation medium
+    scene = td.Scene(medium=medium)
+    assert not scene.allow_gain
+    scene = scene.updated_copy(medium=medium_gain)
+    assert scene.allow_gain
+
+    # Test structure with anisotropic gain medium
+    struct = td.Structure(geometry=td.Box(center=(0, 0, 0), size=(1, 1, 1)), medium=medium_ani)
+    struct_gain = struct.updated_copy(medium=medium_gain_ani)
+    scene = td.Scene(
+        medium=medium,
+        structures=[struct],
+    )
+    assert not scene.allow_gain
+    scene = scene.updated_copy(structures=[struct_gain])
+    assert scene.allow_gain
+
+
+def test_perturbed_mediums_copy():
+
+    # Non-dispersive
+    pp_real = td.ParameterPerturbation(
+        heat=td.LinearHeatPerturbation(
+            coeff=-0.01,
+            temperature_ref=300,
+            temperature_range=(200, 500),
+        ),
+    )
+
+    pp_complex = td.ParameterPerturbation(
+        heat=td.LinearHeatPerturbation(
+            coeff=0.01j,
+            temperature_ref=300,
+            temperature_range=(200, 500),
+        ),
+        charge=td.LinearChargePerturbation(
+            electron_coeff=-1e-21,
+            electron_ref=0,
+            electron_range=(0, 1e20),
+            hole_coeff=-2e-21,
+            hole_ref=0,
+            hole_range=(0, 0.5e20),
+        ),
+    )
+
+    coords = dict(x=[1, 2], y=[3, 4], z=[5, 6])
+    temperature = td.SpatialDataArray(300 * np.ones((2, 2, 2)), coords=coords)
+    electron_density = td.SpatialDataArray(1e18 * np.ones((2, 2, 2)), coords=coords)
+    hole_density = td.SpatialDataArray(2e18 * np.ones((2, 2, 2)), coords=coords)
+
+    pmed1 = td.PerturbationMedium(permittivity=3, permittivity_perturbation=pp_real)
+
+    pmed2 = td.PerturbationPoleResidue(
+        poles=[(1j, 3), (2j, 4)],
+        poles_perturbation=[(None, pp_real), (pp_complex, None)],
+    )
+
+    struct = td.Structure(geometry=td.Box(center=(0, 0, 0), size=(1, 1, 1)), medium=pmed2)
+
+    scene = td.Scene(
+        medium=pmed1,
+        structures=[struct],
+    )
+
+    # no perturbations provided -> regular mediums
+    new_scene = scene.perturbed_mediums_copy()
+
+    assert isinstance(new_scene.medium, td.Medium)
+    assert isinstance(new_scene.structures[0].medium, td.PoleResidue)
+
+    # perturbations provided -> custom mediums
+    new_scene = scene.perturbed_mediums_copy(temperature)
+    new_scene = scene.perturbed_mediums_copy(temperature, None, hole_density)
+    new_scene = scene.perturbed_mediums_copy(temperature, electron_density, hole_density)
+
+    assert isinstance(new_scene.medium, td.CustomMedium)
+    assert isinstance(new_scene.structures[0].medium, td.CustomPoleResidue)
--- a/tidy3d/components/bc_placement.py
+++ b/tidy3d/components/bc_placement.py
@@ -1,0 +1,117 @@
+"""Defines placements for boundary conditions."""
+from __future__ import annotations
+
+from abc import ABC
+from typing import Union, Tuple
+
+import pydantic.v1 as pd
+
+from .types import BoxSurface
+from .base import Tidy3dBaseModel
+from ..exceptions import SetupError
+
+
+class AbstractBCPlacement(ABC, Tidy3dBaseModel):
+    """Abstract placement for boundary conditions."""
+
+
+class StructureBoundary(AbstractBCPlacement):
+    """Placement of boundary conditions on the structure's boundary.
+
+    Example
+    -------
+    >>> bc_placement = StructureBoundary(structure="box")
+    """
+
+    structure: str = pd.Field(
+        title="Structure Name",
+        description="Name of the structure.",
+    )
+
+
+class StructureStructureInterface(AbstractBCPlacement):
+    """Placement of boundary conditions between two structures.
+
+    Example
+    -------
+    >>> bc_placement = StructureStructureInterface(structures=["box", "sphere"])
+    """
+
+    structures: Tuple[str, str] = pd.Field(
+        title="Structures",
+        description="Names of two structures.",
+    )
+
+    @pd.validator("structures", always=True)
+    def unique_names(cls, val):
+        """Error if the same structure is provided twice"""
+        if val[0] == val[1]:
+            raise SetupError(
+                "The same structure is provided twice in 'StructureStructureInterface'."
+            )
+        return val
+
+
+class MediumMediumInterface(AbstractBCPlacement):
+    """Placement of boundary conditions between two mediums.
+
+    Example
+    -------
+    >>> bc_placement = MediumMediumInterface(mediums=["dieletric", "metal"])
+    """
+
+    mediums: Tuple[str, str] = pd.Field(
+        title="Mediums",
+        description="Names of two mediums.",
+    )
+
+    @pd.validator("mediums", always=True)
+    def unique_names(cls, val):
+        """Error if the same structure is provided twice"""
+        if val[0] == val[1]:
+            raise SetupError("The same medium is provided twice in 'MediumMediumInterface'.")
+        return val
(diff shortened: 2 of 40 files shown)


tidy3d/components/heat/simulation.py
```diff
@@ -0,0 +1,760 @@
...
+                add_points=False,
+                color=plot_params.edgecolor,
+                linewidth=plot_params.linewidth,
+            )
+            # ax.add_artist(patch)
+        return ax
+
+    # pylint: disable=too-many-locals,too-many-branches,too-many-statements
+    @staticmethod
+    def _construct_heat_boundaries(
```
tylerflex, line 362: this function is extremely long.. does it make sense to try to break it up into sub functions? It would be a lot of work and potentially more boilerplate / functions being used once but could make it easier to read and potentially more re-usable. I'll leave it up to you.

tidy3d/components/heat/source.py
```diff
@@ -0,0 +1,48 @@
...
+from ..data.data_array import TimeDataArray
+from ..viz import PlotParams
+
+from ...constants import VOLUMETRIC_HEAT_RATE
+
+
+class HeatSource(ABC, Tidy3dBaseModel):
+    """Abstract heat source."""
+
+    structures: Tuple[str] = pd.Field(
```
tylerflex, line 21: I might be mistaken but I think `Tuple[str]` means that it has to be a tuple of only **one** string, eg `("hello",)`.  I think what you might want is `Tuple[str, ...]` but double check me on this.

tidy3d/components/heat_spec.py
```diff
@@ -0,0 +1,51 @@
...
+    -------
+    >>> solid = SolidSpec(
+    ...     capacity=2,
+    ...     conductivity=3,
+    ... )
+    """
+
+    capacity: pd.PositiveFloat = pd.Field(
+        title="Heat capacity",
+        description=f"Volumetric heat capacity {SPECIFIC_HEAT_CAPACITY}.",
```
tylerflex, line 40: add "in units of" maybe

tidy3d/components/medium.py
```diff
@@ -181,6 +182,13 @@ def _validate_nonlinear_spec(cls, val):
             f"currently supported for medium class {cls}."
         )
 
+    heat_spec: Optional[HeatSpecType] = pd.Field(
+        None,
+        title="Heat Specification",
+        description="Specification of the medium heat properties.",
```
tylerflex, line 188: maybe we can add more description here. For example something like
```
If a heat_spec is added to a medium, it will be registered if the corresponding ``Structure`` is added to a ``HeatSimulation``.  <insert something about how to run heat sim and convert to optical sim>.  Otherwise, the ``heat_spec`` will not affect the running of an optical ``Simulation``.
```
Basically, just give a curious user looking at the medium API the opportunity to learn more about the heat solver.

tidy3d/components/scene.py
```diff
@@ -0,0 +1,1171 @@
...
+from ..constants import inf, THERMAL_CONDUCTIVITY
+from ..exceptions import SetupError
+from ..log import log
+
+# maximum number of mediums supported
+MAX_NUM_MEDIUMS = 65530
+
+
+class Scene(Box):  # pylint:disable=too-many-public-methods
+    """Contains geometric information to use in simulations.
```
tylerflex, line 39: Maybe we can make this more clear, such as
```
Contains generic information about the geometry and medium properties common to all types of simulations.
```

tidy3d/components/scene.py
```diff
@@ -0,0 +1,1171 @@
...
+                            "simulation edges. This can cause unexpected behavior. "
+                            "If intending to extend the structure to infinity along one dimension, "
+                            "use td.inf as a size variable instead to make this explicit."
+                        )
+
+        return val
+
+    """ Post-init validators """
+
+    def _post_init_validators(self) -> None:
```
tylerflex, line 123: do we need this here? or should it just be in `Simulation`? I forget how this works or if we need it to be defined inside of a scene itself.

tidy3d/web/heat.py
```diff
@@ -0,0 +1,366 @@
+"""Web API for heat simulation"""
```
tylerflex, line 1: We'll clearly need to unify the webapis between all of the solvers because this is almost a direct copy of the `Simulation` webapi right? 

I'm working with MC to provide a simple interface where they provide a template for a `Submittable` task (simulation, mode solver, heat simulation) and we implement this template in our webapi and use a single webapi to manage it. We'll see how that looks but that would hopefully eliminate the need for this file.

tylerflex (changes requested):
Looks really good @dbochkov-flexcompute. Mostly minor comments and clarifications here.

tests/test_components/test_scene.py
```diff
@@ -0,0 +1,499 @@
...
+import tidy3d as td
+from tidy3d.components.simulation import MAX_NUM_MEDIUMS
+from ..utils import assert_log_level, log_capture, STL_GEO, custom_medium, custom_drude
+from ..utils import custom_debye, custom_lorentz, custom_poleresidue, custom_sellmeier
+
+SCENE = td.Scene(size=(1, 1, 1))
+
+SCENE_FULL = td.Scene(
+    size=(8.0, 8.0, 8.0),
+    structures=[
```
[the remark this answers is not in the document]
dbochkov-flexcompute: good idea. In fact, I can just use `SCENE = SIM_FULL.scene` haha

tidy3d/components/bc_placement.py
```diff
@@ -0,0 +1,106 @@
+"""Defines heat material specifications"""
```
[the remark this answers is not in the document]
dbochkov-flexcompute: fixed

tidy3d/components/bc_placement.py
```diff
@@ -0,0 +1,106 @@
...
+        title="Structures",
+        description="Names of two structures.",
+    )
+
+    @pd.validator("structures", always=True)
+    def unique_names(cls, val):
+        """Error if the same structure is provided twice"""
+        if val[0] == val[1]:
+            raise SetupError(
+                "The same structure is provided twice in :class:.`StructureStructureInterface`."
```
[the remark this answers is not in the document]
dbochkov-flexcompute: fixed

tidy3d/components/heat/grid.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    dl: pd.PositiveFloat = pd.Field(
+        ...,
+        title="Grid Size",
+        description="Grid size for uniform grid generation.",
+        units=MICROMETER,
+    )
+
+    min_edges_per_circumference: pd.PositiveFloat = pd.Field(
```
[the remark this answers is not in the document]
dbochkov-flexcompute: it's only relevant for `Sphere` and `Cylinder`, as now explained in description

tidy3d/components/heat/sim_data.py
```diff
@@ -0,0 +1,76 @@
...
+    ...     heat_simulation=heat_sim, temperature_data=temperature_data,
+    ... )
+    """
+
+    heat_simulation: HeatSimulation = pd.Field(
+        title="Heat Simulation",
+        description="``HeatSimulation`` object describing the problem setup.",
+    )
+
+    temperature_data: SpatialDataArray = pd.Field(
```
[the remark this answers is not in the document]
dbochkov-flexcompute: resolved

tidy3d/components/heat/simulation.py
```diff
@@ -0,0 +1,760 @@
...
+    def _plot_boundary_condition(
+        self, shape: Shapely, boundary_spec: HeatBoundarySpec, ax: Ax
+    ) -> Ax:
+        """Plot a structure's cross section shape for a given boundary condition."""
+        plot_params_bc = self._get_bc_plot_params(boundary_spec=boundary_spec)
+        ax = self.plot_line(line=shape, plot_params=plot_params_bc, ax=ax)
+        return ax
+
+    # FIXME: probably needs revision
+    def plot_line(self, line: Shapely, plot_params: PlotParams, ax: Ax) -> Ax:
```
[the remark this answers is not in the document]
dbochkov-flexcompute: made this work

tidy3d/web/heat.py
dbochkov-flexcompute: let's ignore web api stuff for now until stub is merged

dbochkov-flexcompute:
Major revision taking into account all comment. Most significant additions are PolySlab support, Symmetry support, and proper monitor/simulation data classes. It seemed to accomplish that the most natural way would be to create abstract base classes for main components since their optic and heat specification will share many things, I put them in `base_sim/` for now. So, aiming at the following structure eventually for simulation classes (and similarly for other components):
```
      Scene
        |
  (composition)
        |
AbstractSimulation(ABC)------------------
        |                               |
  (inheritance)                   (inheritance)
        |                               |
Simulation(AbstractSimulation)    HeatSimulation(AbstractSimulation)
```

Demonstration notebooks have been updated too:
- https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/daniil/heat-solver-draft/docs/source/notebooks/HeatSolverDraft.ipynb
- https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/daniil/heat-solver-draft/docs/source/notebooks/ThermallyTunedRingResonatorDraft.ipynb
These are meant to provide more explanation on how every component is interacting with each other


One concerns: we chose to just copy relevant code from `Simulation` to `Scene` for now, thinking that just removing duplicating code would be easier if everything works out. But I start noticing that `Simulation` functionalities start diverging from whatever I have put in `Scene`. This could happen for other components. So, perhaps, we shouldn't wait too long until we actually reimplement `Simulation` and other optic classes based on introduced in this PR `Scene` and higher level abstractions.

momchil-flex:
Yes, that sounds important. Let's aim to have that done for 2.5 too. Maybe even do the refactor as soon as we start a 2.5 branch?

momchil-flex:
I see, not so straightforward yeah, but in terms of backwards compatibility I think we'll need to do something like point 2. for now, where simulation is still initialized with center, size, medium, and then a scence is maybe initialized as a private attribute or is stored as a cached property or something.

dbochkov-flexcompute:
there are currently 4 commits in this PR:
- "heat solver interface": last review state
- "heat solver web api": [new stuff] implementation of heat web api based on stub refactor
- "moving structures/medium back from ...": [new stuff]
-- removes `size` and `center` from `Scene` and makes them automatically calculated properties.
-- renames `UniformHeatGrid` and `DistanceHeatGrid` to `UniformUnstructureGrid` and `DistanceUnstructureGrid` since they could be reused in other solvers
-- makes `AbstractSimulation` be a derivative of `Box`, thus `domain` field is no longer needed. Simulation domains is set as usual through `size` and `center`.
-- add `structures` and `medium` to `AbstractSimulation`. + property `scene` + class method `from_scene`
- "temporarily limiting matplotlib ...": just a temporary commit to make it easier to deploy and test, can be removed at merge if no longer required

tidy3d/components/scene.py
```diff
@@ -0,0 +1,1274 @@
...
+    def _validate_num_mediums(cls, val):
+        """Error if too many mediums present."""
+
+        if val is None:
+            return val
+
+        mediums = {structure.medium for structure in val}
+        if len(mediums) > MAX_NUM_MEDIUMS:
+            raise SetupError(
+                f"Tidy3d only supports {MAX_NUM_MEDIUMS} distinct mediums."
```
tylerflex, line 85: just noticed, this should be Tidy3D (capital D)

tidy3d/components/scene.py
```diff
@@ -0,0 +1,1274 @@
...
+                f"{len(mediums)} were supplied."
+            )
+
+        return val
+
+    """ Accounting """
+
+    @cached_property
+    def bounds(self) -> Bound:
+        """Automatically defined scene's bounds based on present structures. Infinite dimensions
```
tylerflex, line 95: seems like a reasonable way to do this

tidy3d/components/scene.py
```diff
@@ -0,0 +1,1274 @@
...
+
+    @cached_property
+    def medium_map(self) -> Dict[MediumType, pd.NonNegativeInt]:
+        """Returns dict mapping medium to index in material.
+        ``medium_map[medium]`` returns unique global index of :class:`.AbstractMedium` in scene.
+
+        Returns
+        -------
+        Dict[:class:`.AbstractMedium`, int]
+            Mapping between distinct mediums to index in simulation.
```
tylerflex, line 169: simulation -> scene (everywhere in file)

tidy3d/components/scene.py
```diff
@@ -0,0 +1,1274 @@
...
+        # if the test object is a volume, test each surface recursively
+        surfaces = test_object.surfaces_with_exclusion(**test_object.dict())
+        structures_merged = []
+        for surface in surfaces:
+            structures_merged += Scene.intersecting_structures(surface, structures)
+        return structures_merged
+
+    """ Plotting General """
+
+    @staticmethod
```
tylerflex, line 257: I think it's ok for now, but eventually we could probably put a lot of these static methods in viz.py or some `components/scene/viz.py` if we wanted to separate them out.

tidy3d/components/simulation.py
```diff
@@ -3217,3 +3218,49 @@ def perturbed_mediums_copy(
             sim_dict["medium"] = med.perturbed_copy(**restricted_arrays)
 
         return Simulation.parse_obj(sim_dict)
+
+    @cached_property
+    def scene(self) -> Scene:
```
tylerflex, line 3223: is this implemented in the super class?

tylerflex: same for `from_scene`? also add from_scene return type

tidy3d/components/validators.py
```diff
@@ -182,6 +182,39 @@ def objects_in_sim_bounds(cls, val, values):
...
+def assert_objects_in_scene_bounds(field_name: str, error: bool = True):
+    """Makes sure all objects in field are at least partially inside of simulation bounds."""
+
+    @pydantic.validator(field_name, allow_reuse=True, always=True)
+    def objects_in_sim_bounds(cls, val, values):
+        """check for intersection of each structure with simulation bounds."""
+        scene = values.get("scene")
+        if scene is None:
+            raise ValidationError(
+                f"Cannot validate {field_name} because validation of 'HeatSimulation.scene' "
```
tylerflex, line 194: quotes around field name

tidy3d/web/api/tidy3d_stub.py
```diff
@@ -19,29 +20,31 @@
...
-SimulationType = Union[Simulation]
-SimulationDataType = Union[SimulationData]
+SimulationType = Union[Simulation, HeatSimulation]
+SimulationDataType = Union[SimulationData, HeatSimulationData]
 
 
 class Tidy3dStub(BaseModel, TaskStub):
 
-    simulation: SimulationType
+    simulation: SimulationType = pd.Field(discriminator="type")
```
tylerflex, line 32: eventually we should add some more description here, if only to help other developers understand

tidy3d/web/api/webapi.py
```diff
@@ -273,125 +273,159 @@ def monitor(task_id: TaskId, verbose: bool = True) -> None:
...
-            console.log(
-                f"You are running this simulation for FREE. Your current plan allows"
-                f" up to {block_info.maxFreeCount} free non-concurrent simulations per"
-                f" day (under {grid_points_str} grid points and {time_steps_str}"
-                f" time steps)"
-            )
-        else:
-            est_flex_unit = task_info.estFlexUnit
-            if est_flex_unit is not None and est_flex_unit > 0:
+        # Wait for task to finish
```
tylerflex, line 286: I won't go through all of this logic, but we should test this pretty thoroughly because the webapi logic tends to break in weird ways sometimes.

tylerflex (changes requested):
Thanks @dbochkov-flexcompute. I think this looks really good. Mostly this is just minor stuff.

tidy3d/components/heat/simulation.py
```diff
@@ -0,0 +1,815 @@
...
+    monitors: Tuple[HeatMonitorType, ...] = pd.Field(
+        (),
+        title="Monitors",
+        description="Monitors in the simulation.",
+    )
+
+    grid_spec: HeatGridType = pd.Field(
+        title="Grid Specification",
+        description="Grid specification for heat simulation.",
+        discriminator=TYPE_TAG_STR,
```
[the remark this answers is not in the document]
dbochkov-flexcompute: My understanding was that we should use discriminator whenever variable type is Union with more than 1 element, so that pydantic doesn't try to initialize provided argument with every element of that Union. Do you know if something changed during pydantic 2 transition?

tidy3d/components/heat/simulation.py
```diff
@@ -0,0 +1,815 @@
...
+                if not isinstance(geom, HeatSingleGeometryType):
+                    raise SetupError(
+                        "'HeatSimulation' does not currently support geometries of type "
+                        f"'{geom.type}'. Allowed geometries are 'Box', 'Cylinder', 'Sphere', "
+                        f"and 'PolySlab'."
+                    )
+        return val
+
+    @pd.validator("size", always=True)
+    def check_zero_dim_domain(cls, val, values):
```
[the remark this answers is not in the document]
dbochkov-flexcompute: support is planned after initial release

tidy3d/components/heat/simulation.py
```diff
@@ -0,0 +1,815 @@
...
+            source=source,
+            source_min=source_min,
+            source_max=source_max,
+            alpha=alpha,
+        )
+        ax = self.plot_shape(shape=shape, plot_params=plot_params, ax=ax)
+        return ax
+
+    @classmethod
+    def from_scene(cls, scene: Scene, **kwargs):
```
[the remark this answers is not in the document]
dbochkov-flexcompute: redefining it here to add example in docstring

tidy3d/components/simulation.py
dbochkov-flexcompute: actually, currently `Simulation` doesn't inherit from the new `AbstractSimulation` per our discussion to leave it on its own for a bit (similarly, `SimulationData`, monitors, etc). But I would like to do this transition in the next PR.

tidy3d/web/api/webapi.py
dbochkov-flexcompute: good idea, everything seemed to be working, except for download_log. but it was an issue on backend and now is fixed.

tidy3d/components/heat/simulation.py
```diff
@@ -0,0 +1,815 @@
...
+                for med in bc_place.mediums:
+                    if med in named_mediums_present:
+                        if med in med_to_bc_spec:
+                            med_to_bc_spec[med] += [bc_spec]
+                        else:
+                            med_to_bc_spec[med] = [bc_spec]
+
+        return med_to_bc_spec
+
+    @staticmethod
```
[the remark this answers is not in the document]
dbochkov-flexcompute: it will eventually go to `components/unstructured_sim`

tidy3d/components/bc_placement.py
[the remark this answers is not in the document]
dbochkov-flexcompute: it will eventually go to `components/unstructured_sim`

tidy3d/components/heat/data/sim_data.py
```diff
@@ -0,0 +1,232 @@
...
+    data: Tuple[HeatMonitorDataType, ...] = pd.Field(
+        ...,
+        title="Monitor Data",
+        description="List of :class:`.MonitorData` instances "
+        "associated with the monitors of the original :class:`.Simulation`.",
+    )
+
+    @equal_aspect
+    @add_ax_if_none
+    def plot_data(
```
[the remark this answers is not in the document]
dbochkov-flexcompute: yeah, I was hoping to just have a common implementation, but there were a bit too many specific things for each simulation type. Probably just leave like that for now and revisit after initial release

dbochkov-flexcompute:
addressed comments + added a beta stage disclaimer in heat solver webapi:
> Tidy3D's heat solver is currently in the beta stage. All heat simulations are charged a flat fee of 0.025 FlexCredit.

tidy3d/components/heat/simulation.py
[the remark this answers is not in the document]
tylerflex: I was under the impression that something changed with the discriminator recently, but I think I might be mistaken. let's leave it in.

tidy3d/components/heat/simulation.py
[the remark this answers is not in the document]
tylerflex: a lot of plotting logic probably can be separated into other files, since it seems like a lot of them are just `staticmethod`s. But we can do this at a later time, I think.

tidy3d/web/api/webapi.py
tylerflex: have you tried job and batch of heat simulation? or even mixes of simulation and heat sim? it could be interesting to see how that works too. 

In my experience, sometimes the bugs were happening in `web.monitor()` in weird edge cases, for example 
* a task errors (if that happens we generally try the task again and see if it errors twice)
* the timing is somewhat different for whatever reason (task runs before the % done is ready)

There were some improvements earlier this year to fix some of these issues but if this PR changes any of that logic, they could appear again.  Will be good to run a bunch just to ensure they don't pop up again because they can be quite tricky to reproduce.

momchil-flex (commented):
I'm pretty lost here so @dbochkov can you at least help me understand the current state and the plan for the general unification of things?

For example: HeatSimulation inherits from AbstractSimulation, but Simulation does not. Similarly HeatSimulationData inherits from AbstractSimulationData, but SimulationData does not. And while AbstractMonitorData has been introduced, both MonitorData and HeatMonitorData just inherit from Dataset.

Is there currently any duplicated code between Scene and Simulation?

I propose clarifying how everything is meant to work and putting the next few days of effort into actually merging this PR into 2.5 in a way that will avoid duplicated code and further code divergence. Then we can review/fix smaller details.

dbochkov-flexcompute:
> I'm pretty lost here so @dbochkov can you at least help me understand the current state and the plan for the general unification of things?
> 
> For example: HeatSimulation inherits from AbstractSimulation, but Simulation does not. Similarly HeatSimulationData inherits from AbstractSimulationData, but SimulationData does not. And while AbstractMonitorData has been introduced, both MonitorData and HeatMonitorData just inherit from Dataset.
> 
> Is there currently any duplicated code between Scene and Simulation?
> 
> I propose clarifying how everything is meant to work and putting the next few days of effort into actually merging this PR into 2.5 in a way that will avoid duplicated code and further code divergence. Then we can review/fix smaller details.

At this stage `HeatMonitorData` should inherit from `AbstractMonitorData`, but `MonitorData` not. I'll fix that. 

So, the idea was to only have heat classes inheriting new `Abstract*` classes to test the new code organization while keeping optic classes untounched, in case something doesn't work out and we need to rethink it. I planned to do the removal of the duplicated code and switching optic classes to `Abstract*` classes in a separate PR. Are you suggesting just to go ahead and do that in this PR? I guess that would have the benefit of double-checking that nothing weird happens when proceeding with this unification.

dbochkov-flexcompute:
Otherwise, we seemed to have converged on pretty much everything in this PR. Just need to test out `Batch` with failing heat tasks

momchil-flex:
I think separate PR is fine. Just sooner rather than later. Generally as far as I'm concerned we could merge this soon (after you rebase again and answer all current comments, etc.) And just get it out there so that we can start getting some real user feedback and start fixing things. :)

momchil-flex:
Is it currently possible for a user to have a look at the mesh that was used after a solver run?

dbochkov-flexcompute:
> Is it currently possible for a user to have a look at the mesh that was used after a solver run? I guess they should at least be able to download the mesh file even if we don't provide any gmesh-wrapped functionality because of their GPL license.

yes, it's not ideal currently, but you can directly download mesh ("mesh.cgns") or solution+mesh ("volume_proc0.vtu") using `web.core.s3utils.download_file` to visualize in thirdparty tools like paraview or pyvista. 

This user experience can definitely be improved. Probably, once we have heat monitors that return data on the original unstructured grid instead of interpolated to an equivalent Cartesian one. One idea could be to introduce something like `MeshMonitor` that only return mesh in the selected region. And if a simulation contains only such monitors, then only meshing is performed and solver step is skipped. This will provide a way to inspect and iterate on mesh before actually solving heat equation. Or just provide meshing-only "solver" similar to mode solver.

requirements/basic.txt
```diff
@@ -6,7 +6,7 @@ importlib-metadata>=6.0.0
 h5netcdf==1.0.2
 h5py>=3.0.0
 rich<12.6.0 # note: rich >= 12.6 adds double progressbars
-matplotlib
+matplotlib<=3.7.2
```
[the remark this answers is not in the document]
dbochkov-flexcompute: removed

PR #1084, opened 2023-08-21
Lucas/recursive geometry fixes

lucas-flexcompute:
This PR fixes issues raised in the discussions from flexcompute/tidy3d-core#375
Note: the value for `MAX_GEOMETRY_COUNT` comes from @weiliangjin2021 suggestion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Operations on geometry (`+`, `|`, `*`, `&`, `-`, `^`) will create the appropriate `ClipOperation` or `GeometryGroup`.
 - `Geometry.from_shapely` to extrude shapely primitives into Tidy3D geometry.
 - `Geometry.from_gds` to extrude GDS cells into geometry groups with support for holes.
+- `components.geometry.utils.traverse_geometry` used internally to traverse geometry trees.
+- `components.geometry.utils.flatten_groups` used internally to validate large geometry trees.
 
 ### Changed
 - `nyquist_step` also taking the frequency range of frequency-domain monitors into account.
--- a/tests/test_components/test_geometry.py
+++ b/tests/test_components/test_geometry.py
@@ -11,7 +11,8 @@ import trimesh
 
 import tidy3d as td
 from tidy3d.exceptions import SetupError, Tidy3dKeyError, ValidationError
-from tidy3d.components.geometry.base import Geometry, Planar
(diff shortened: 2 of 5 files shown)


weiliangjin2021:
Sorry, I just randomly suggested a value of `MAX_GEOMETRY_COUNT` as an idea. Since we see it's already slow in Anderson notebook when there are 3e3 spheres (slow down from 3s to 1 hour), perhaps setting it to be 1e2 would make more sense, then it would slow down from 3s to 2min?

momchil-flex:
Wait, do I understand correctly that this would be forcing people *not* to use geometry group if they have too many structures?

Until now, the idea has been to try and direct people to *use* that actually, as it is faster e.g. in the auto grid generation, and some validators that work with geometry bounds and don't have to iterate through e.g. 1M geometries for a metalens.

weiliangjin2021:
Yeah, we should only warn in ClipOperation except for union.

momchil-flex:
Can that even be well-defined when you have recursive operations?

Maybe we remove the validator from this PR and then think about how to handle things more generally (spped up, restrictions...)

lucas-flexcompute:
This applies the [same flattening](https://github.com/flexcompute/tidy3d/pull/1084/files#diff-7360ac4bbe331087dbf8aa41d7163c3d1f0e95986f0488ce103cf24e4520049bR451) that we use on the backend before counting. So `GeometryGroup` and union `ClipOperation` are decomposed prior to the counting. So, examples like the metalens are not affected by this number. This would error:
```python
ClipOperation(
    operation="difference",
    geometry_a=Box(...),
    geometry_b=GeometryGroup(geometries=[... too many geometries ...]),
)
```
because we cannot flatten a difference operation, but these, for example, would be fine, as they are decomposed in the backend:
```python
ClipOperation(
    operation="union",
    geometry_a=Box(...),
    geometry_b=GeometryGroup(geometries=[... no limits here ...]),
)

GeometryGroup(geometries=[... no limits here ...]))
```

tylerflex (approved):

momchil-flex:
Oh, that's great! It seems I didn't really understand what exactly the code does then.

Shall we still decrease the max number as @weiliangjin2021 is suggesting? We can increase it later if we optimize the processing.

lucas-flexcompute:
Absolutely. Does the 2 min sound like a good option? We can set the maximum to 100 until the first user complains…

lucas-flexcompute:
Done. I've also added a test to ensure the examples I gave above work (which also led to a small fix in the count).

PR #1085, opened 2023-08-21
fix lineplot zorder

tylerflex:


lucas-flexcompute (approved):
Maybe not something to change in this PR, but it would be nice to have finite z order ranks (with gaps in between) so that the user can also customize the plots by adding annotations of their own in whatever level they want to.

tylerflex:
Good idea. Could you make a front end issue for this? thanks

PR #1091, opened 2023-08-22
add uat2 env

magiWei:


magiWei:
A new environment uat2 has been added, it has been tested ok, and it needs to be released in 2.4.0

tylerflex (approved):

PR #1093, opened 2023-08-24
adjoint filter cleanup and additions

tylerflex:
* Adds a `CircularFilter`, which is just a uniform "cylinder" kernel instead of a conic kernel.
* Rename `feature_size` to `radius` in the circular filters.
* Remove extra factor of `sqrt(3)` in the radius calculation, which is inconsistent with literature.

Note: initial testing of the circular filter is showing that it's terrible compared to conic filter, so not mentioning it in the changelog for now.


momchil-flex (approved):
Looks good, just some epxlanation about the difference between conical and circular filter somehwere?

tylerflex:
Thanks, I added a bit more explanation and recommendation that the user use a conical mask (note each of them have the formula of the kernel already).

PR #1097, opened 2023-08-25
Improved warnings with Medium2D

caseyflex:
This adds a warning when a 2D geometry doesn't have a 2D medium.
@tylerflex it seems to break the adjoint tests, which has 0-thickness structures in a 2D sim. Is it possible to change these thicknesses to some other value, like td.inf? This may be more robust anyway without changing the intended behavior.

@tomflexcompute , this should also fix the plot_eps warning

### DIFF
--- a/tidy3d/components/structure.py
+++ b/tidy3d/components/structure.py
@@ -11,6 +11,7 @@ from .viz import add_ax_if_none, equal_aspect
 from .grid.grid import Coords
 from ..constants import MICROMETER
 from ..exceptions import SetupError
+from ..log import log
 
 
 class AbstractStructure(Tidy3dBaseModel):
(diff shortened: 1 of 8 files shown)


tylerflex:
@caseyflex I'm a bit confused. This seems to break both the regular and 2D adjoint tests with an error

```
E               pydantic.v1.error_wrappers.ValidationError: 1 validation error for JaxStructure
E               medium -> size -> 0
E                 ensure this value is greater than or equal to 0 (type=value_error.number.not_ge; limit_value=0)
```

Mediums shouldn't have sizes so that's strange.

Also, the adjoint tests don't have 2D structures in them (I just checked the regular one, which failed).

It seems this warning was also triggered in the tests as I see the warning message about 2D structure detected with no 2D medium.

Something seems strange here, I'm not sure what to make of it.

tidy3d/components/structure.py
```diff
@@ -99,14 +101,24 @@ def eps_diagonal(self, frequency: float, coords: Coords) -> Tuple[complex, compl
     @pydantic.validator("medium", always=True)
     def _check_2d_geometry(cls, val, values):
```
tylerflex, line 102: I can't help but wonder if this can just be simplified a lot

```py
geom = values.get("geometry")
if geom is None:
    raise SetupError("'Structure.geometry' did not pass validation, can't check 2D medium".)

if isinstance(val, Medium2D) and not geom.is_2D:
    raise SetupError("Structure with a `Medium2D` medium needs a `2D` geometry.")

if geom.is_2D and not isinstance(val, Medium2D):
    zero_axis = geom.bounding_box.size.index(0)
    log.warning("found a geometry with < 3 dimensions not associated with a Medium2D ...")
```

with a simple property `Geometry.is_2D` that is implemented in each of the geometry subclasses?

tidy3d/components/structure.py
```diff
@@ -99,14 +101,24 @@ def eps_diagonal(self, frequency: float, coords: Coords) -> Tuple[complex, compl
...
             # validate that the geometry is actually 2d
-            geom = values.get("geometry")
             if not geom:
                 raise SetupError(
                     "Can't validate 2D structure because its geometry did not pass validation."
                 )
             _ = geom._normal_2dmaterial
+        elif geom is not None:
+            if np.count_nonzero(geom.bounding_box.size) < 3:
+                zero_axis = geom.bounding_box.size.index(0)
```
tylerflex, line 114: note that `.index()` just returns the first index. if there are several dimensions of size zero, this will just return the first one. Could either change the warning message to make this more clear or use
```py
zero_axes = [ind for ind, ele in enumerate(geom.bounding_box.size) if ele == 0.0]
```

tidy3d/components/structure.py
tylerflex: Alternatives: 
* add a `Geometry.is_3D -> bool` method and check `if not geom.is_3D`
* add a `Geometry.ndims -> int` and check if `geom.ndims < 3`

tylerflex (changes requested):

tidy3d/components/structure.py
caseyflex: I think the complicated logic is a bit necessary, since we are checking different things. On the one hand, we need to check that if the medium is a Medium2D, the geometry must be a supported type and it must be 2D. Here, _normal_2dmaterial gives a detailed error message depending on the geometry and setup. On the other hand, if the bounding box has zero size in a certain direction and the medium is not Medium2D, it is helpful to give a warning. I'll add some comments to clarify this but I'm inclined to leave the logic as-is

tomflexcompute (approved):

PR #1098, opened 2023-08-25
Extended lower limit of frequency range for graphene

caseyflex:
Previously, a lower limit used for fitting the interband term was also used in frequency_range. It's better to set the lower limit of frequency_range to zero, since the Drude behavior is valid to DC


tomflexcompute (approved):
Thanks! I think the highest frequency is already sufficiently high so it should be good.

tylerflex (approved):

PR #1100, opened 2023-08-25
adapt to AWS lambda environment

magiWei:


momchil-flex:
Changes for hdf5 handling look good. Please revert all changes related to the config file, they are not needed and are also making the tests fail (can't import from tidy3d in setup.py).

magiWei:
> config file

 config file? which one?

momchil-flex:
> > config file
> 
> config file? which one?

Revert all the changes related to `~/.tidy3d` folder

magiWei:
> ~/.tidy3d

so, if user cannot write ~/.tidy3d, they use os.enviroment?

momchil-flex:
Yes.

However, this makes me realize something: when we had user name and password authentication, the user was asked whether they want to store credentials to file.

Now it seems like if user does `configure`, the API key is always written to file. This would error on lambda. But shouldn't we still ask user such that they can say no if they don't want the api key stored?

magiWei:
> 



> Yes.
> 
> However, this makes me realize something: when we had user name and password authentication, the user was asked whether they want to store credentials to file.
> 
> Now it seems like if user does `configure`, the API key is always written to file. This would error on lambda. But shouldn't we still ask user such that they can say no if they don't want the api key stored?
> 
> @lei-flex what do you think?

This is my concern, if the user needs to use username/pwd, f"{expanduser('~')}/.tidy3d" can't be writted on lambda, so I add one "/tmp/.tidy3d"

TIDY3D_DIR = f"{expanduser('~')}/.tidy3d"
TIDY3D_DIR2 = "/tmp/.tidy3d"
if not os.access(TIDY3D_DIR, os.W_OK):
    TIDY3D_DIR = TIDY3D_DIR2

momchil-flex:
> TIDY3D_DIR = f"{expanduser('~')}/.tidy3d"
> TIDY3D_DIR2 = "/tmp/.tidy3d"
> if not os.access(TIDY3D_DIR, os.W_OK):
> TIDY3D_DIR = TIDY3D_DIR2

Ok let's keep this part as long as you don't have to import anything in setup.py.

momchil-flex:
Note tests are failing currently.

magiWei:
> Note tests are failing currently.

tests are ok now, Do we need add one test about lambda function.

momchil-flex:
How can we test it? If possible to mock then yeah why not.

magiWei:
> How can we test it? If possible to mock then yeah why not.

I test it on https://github.com/flexcompute/tidy3d-python-webapi, add one new release-xxx and write one api, then deploy to lambda,

PR #1102, opened 2023-08-25
Final status update in web.monitor with the correct field decay

momchil-flex:
Previously, the final update of the progressbar could leave it at a field decay status which is not the final one and it was looking weird (e.g. saying field_decay ~1e-2 in one case while the final one was 1e-6).


tylerflex (approved):
nice 👍

PR #1103, opened 2023-08-26
ComponentModeler freqs as FreqArray

momchil-flex:
I also tried sel instead of loc but indeed breaks tests on some systems.


momchil-flex:
I'm confused, I thought this was passing on an earlier try. I'll look into it tomorrow

tylerflex (approved):

momchil-flex:
Ok seems like I finally got this to work.

- Re converting freqs to FreqArray, this stores `self.freqs` internally as a Tuple, and some versions of xarray seem to not like you passing a tuple to `sel`, which we were doing internally.
- Re using `sel` instead of `loc`, it works everywhere apart from if you need to *set* the elements, i.e. something like `data.loc(...) = something`. This happens in one place in the test, which I left as `loc`. But in other places, and in our docs notebook, `sel` should be fine.

PR #1108, opened 2023-08-29
tidy3d stub

magiWei:


stub.py
```diff
@@ -0,0 +1,94 @@
...
+    NORMAL = "NORMAL"
+    MODE_SOLVER = "MODE_SOLVER"
+    HEAT = "HEAT"
+
+
+class Stub(BaseModel):
+    @abstractmethod
+    def from_file(self, file_path: str) -> "Stub":
+        @abstractmethod
+        def from_file(self, file_path: str) -> "Stub":
```
tylerflex, line 20: is this a mistake? (two nested `from_file()` functions?)

stub.py
```diff
@@ -0,0 +1,94 @@
+"""Stub class"""
```
tylerflex, line 1: where should `stub.py` live? I think it makes sense to put it in `web/` somewhere. I think it would also be nice to separate the MC-controlled files from the EM-controlled files. Perhaps we can separate them into two directories in `web` and make sure that the EM only imports from MC.

tylerflex, line 1: We can make this module docstring more descriptive maybe 
```
"""Defines interface that can be subclassed to use with the tidy3d webapi."""
```

stub.py
```diff
@@ -0,0 +1,94 @@
...
+        """Get Stub type.
+
+        Returns
+        -------
+        :class:`TaskType`
+            An instance Type of the component class calling `load`.
+        """
+        pass
+
+    @abstractmethod
```
tylerflex, line 68: I dont think we need to enforce that this is implemented in subclasses. so maybe we can just remove `abstractmethod` but leave it as is.

stub.py
```diff
@@ -0,0 +1,94 @@
+"""Stub class"""
+
+from abc import abstractmethod
+from enum import Enum
+
+from pydantic.v1 import BaseModel
+
+
+class TaskType(str, Enum):
+    FDTD = "FDTD"
```
tylerflex, line 10: can you add descriptive docstrings to these classes?

tidy3d/components/data/monitor_data.py
```diff
@@ -827,7 +828,7 @@ def time_reversed_copy(self) -> FieldTimeData:
         return self.copy(update=new_data)
 
 
-class ModeSolverData(ModeSolverDataset, ElectromagneticFieldData):
+class ModeSolverData(ModeSolverDataset, ElectromagneticFieldData, StubData):
```
tylerflex, line 831: We won't want to import web things into the `tidy3d/components` because we enforce that the components are "pure" (to avoid circular dependencies).

So what I'm thinking is that in `web` we will need to define an (EM-controlled) class that interface between `Stub` and `StubData` and the tidy3d components. See some of my earlier comments in lark, for example

```py
class Tidy3DStub(Tidy3DBaseModel, Stub):
    component: # eg a Simulation, SimulationData, etc.

    def to_file(self, fname):
         self.components.to_file(fname, ....)
```

Then the webapi can use these interfaces. How does that sound to you?

tidy3d/web/simulation_task.py
```diff
@@ -204,10 +212,12 @@ def create(
 
         Parameters
         ----------
-        simulation: :class".Simulation"
+        stub: :class".Stub"
```
tylerflex, line 215: so we will still use this SimulationTask and other things in transition? eventually, it would be easiest to have a single `StubTask` and just use that for any Stub subclasses (eg mode solver) is that right?

tidy3d/web/webapi.py
```diff
@@ -579,27 +529,35 @@ def load(
...
+
     if not os.path.exists(path) or replace_existing:
         download(task_id=task_id, path=path, verbose=verbose, progress_callback=progress_callback)
 
     if verbose:
         console = get_logging_console()
         console.log(f"loading SimulationData from {path}")
 
-    sim_data = SimulationData.from_file(path)
+    if TaskType.FDTD == task.taskType or TaskType.NORMAL == task.taskType:
```
tylerflex, line 545: it might be nice to implement these as an `abstract method` in the `StubData` for example a `def postprocess` and then implement in the subclasses?

tylerflex (changes requested):

tylerflex (commented):
Thanks @Qingeng Wei this is getting a lot closer now, I did a review of the code with some comments. I think the major things to discuss now are:
1. The `tidy3d/components` are intended to be independent of the `web/` parts. So i would recommend creating an (EM-team controlled) interface between `Stub` and the components (like `Simulation`). For example a `Tidy3DStub` in `web/tidy3d_stub.py` that has a single field (component).  I put some details in the review and earlier in this thread.
2. There is a lot of legacy code (such as `SimulationTask`) so the idea is we will still support this in transition? It would be nice if we had a unified `StubTask` or something that we could do everything with.
3. There's some logic based on `getType` that we can maybe do more elegantly by adding abstractmethods (or just regular methods) in the stub / stub data classes and implementing them in the subclasses.
4. Let's move `stub.py` into `web/` somewhere and also separate the MC and EM parts into two separate subdirectories, ideally the MC parts will not depend on the EM parts.

stub.py
magiWei: yes, a mistake, modified.

stub.py
magiWei: modified

stub.py
magiWei: modified

stub.py
magiWei: MC : tidy3d/web/core and tidy3d/web/cli
EM:  tidy3d/web excludes of tidy3d/web/core and tidy3d/web/cli

magiWei: MC : tidy3d/web/core and tidy3d/web/cli
EM: tidy3d/web/api

tidy3d/components/data/monitor_data.py
magiWei: modified

stub.py
magiWei: removed

tidy3d/web/simulation_task.py
magiWei: SimulationTask is your StubTask by now.

tidy3d/web/webapi.py
magiWei: use function 'load_data(task_type: TaskType, path: str)'

tidy3d/web/core/http_management.py
```diff
@@ -9,9 +9,9 @@
 import toml
 from tidy3d.web.cli.constants import CONFIG_FILE
 
-from .environment import Env
-from ..exceptions import WebError
-from ..version import __version__
+from tidy3d.web.core.environment import Env
```
tylerflex, line 12: let's use relative imports, for example `from .environment import Env`

tidy3d/web/core/http_management.py
```diff
@@ -9,9 +9,9 @@
...
+from tidy3d.exceptions import WebError
```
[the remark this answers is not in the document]
tylerflex: same comment in the other files in `web/core`

tidy3d/web/core/stub.py
```diff
@@ -0,0 +1,66 @@
...
+class TaskType(str, Enum):
+    FDTD = "FDTD"
+    NORMAL = "NORMAL"
+    MODE_SOLVER = "MODE_SOLVER"
+    HEAT = "HEAT"
+
+
+class Stub(BaseModel):
+
+    def from_file(self, file_path: str) -> "Stub":
```
tylerflex, line 17: so I still think these should be `abstractmethods` (and stub should be ABC), my comment was that I think `validate_pre_upload()` could maybe just not be an abstractmethod, since we can make it optional to implement in subclasses.

tylerflex, line 17: note, if you 
```py
```
then you can use the class name in the type annotations without quotes
```
-> Stub:
```

tylerflex: Also, I think `from_file()` should be a @classmethod as well as an @abstractmethod. This way you can call it directly from the class without making an instance.

```
stub = Stub.from_file(path)
```

tylerflex (commented):
Just a few other comments after taking another look

tidy3d/web/core/http_management.py
magiWei: modified.

tidy3d/web/core/stub.py
magiWei: removed Stub

magiWei: modified

tidy3d/web/core/http_management.py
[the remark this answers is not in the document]
magiWei: modified

tests/Simulation.json
tylerflex, line 1: remove this file or move it (and the path that generated it) into `tests/tmp/` if it's required for the tests.

setup.py
```diff
@@ -1,6 +1,5 @@
 import setuptools
 from distutils.util import convert_path
-from setuptools.command.install import install
```
tylerflex, line 3: what is the reason for this? just wondering

tidy3d/web/api/asynchronous.py
```diff
@@ -2,8 +2,8 @@
 from typing import Dict, List
 
 from .container import DEFAULT_DATA_DIR, BatchData, Batch
-from ..components.simulation import Simulation
-from ..log import log
+from tidy3d.components.simulation import Simulation
```
tylerflex, line 5: we still want these to be relative imports. I think it just means adding another `.`  in front for example 

```
from ...log import log
```

tidy3d/web/api/config.py
```diff
@@ -0,0 +1,15 @@
+""" Commandline interface for tidy3d. """
+
+from tidy3d.web.cli.app import configure_fn
```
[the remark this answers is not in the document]
tylerflex: seems this comes up often, so anywhere else we import `from tidy3d import ...` let's just make them relative, thanks.

tidy3d/web/api/tidy3d_stub.py
```diff
@@ -0,0 +1,87 @@
...
+            return TaskType.FDTD
+        elif isinstance(self.component, ModeSolver):
+            return TaskType.MODE_SOLVER
+
+    def validate_pre_upload(self, source_required):
+        """Perform some pre-checks on instances of component"""
+        pass
+
+
+class Tidy3dStubData(BaseModel):
```
tylerflex, line 73: why does stub data not require to_file or from_file like the regular stub?

tidy3d/web/api/tidy3d_stub.py
```diff
@@ -0,0 +1,87 @@
...
+    """"""
+
+    data: Union[SimulationData, ModeSolverData]
+
+    @classmethod
+    def postprocess(cls, task_type: TaskType, path: str) -> Tidy3dStubData:
```
tylerflex, line 79: my idea here was rather that we provide a separate `from_file()` method and then call `self.postprocess()` 
so the generic "load" webapi would look something like
```
def load(fname):
    obj = StubData.from_file(fname)
    obj = obj.postprocess()

```

tidy3d/web/core/http_util.py
```diff
@@ -7,13 +7,18 @@
...
 
+from .constants import (
+    SIMCLOUD_APIKEY,
+    SIMCLOUD_VERSION,
+    KEY_APIKEY,
+    HEADER_APIKEY,
+    KEY_VERSION,
+    HEADER_VERSION,
+)
+from ..cli.constants import CONFIG_FILE
```
tylerflex, line 19: if this is importing from CLI, I suppose we will have the CLI be an MC controlled directory?

tidy3d/web/core/log.py
```diff
@@ -0,0 +1,364 @@
+"""Logging for core."""
```
tylerflex, line 1: is this copied from our regular `tidy3d/log.py` file? I think it might be easier to simply use the regular `logging` module for web/core errors and logging 

https://docs.python.org/3/library/logging.html

tidy3d/web/api/tidy3d_stub.py
```diff
@@ -0,0 +1,87 @@
...
+
+from pydantic.v1 import BaseModel
+
+from tidy3d import SimulationData, ModeSolverData
+from tidy3d.web.core.types import TaskType
+from tidy3d.components.simulation import Simulation
+from tidy3d.plugins.mode.mode_solver import ModeSolver
+
+
+class Tidy3dStub(BaseModel):
```
tylerflex, line 14: What happened to the `Stub` and `StubData`? I thought we were going to define an interface to define what the tidy3d side objects needed to implement to work with core?

tylerflex (commented):
Looks better, I think we are getting a bit closer but seems like we aren't in alignment on some basic things:

The code seems nicely separated into web/api and web/core, this is a good first step. 

However, there still seems to be lacking a basic interface or template on the core side defining what the EM side needs to implement to work with the webapi.

The Tidy3DStub and Tidy3DStubData should inherit from Stub and StubData ABC classes on the core side that define exactly what methods are needed to be able to run with the webapi. This part seems like the most critical and yet is still missing.

Ideally, the core should define some very generic approach to turning Stub into StubData through http requests and the EM side should simply implement a wrapper based on the tidy3d components available. If we can get the first part right first, then I think we can look in more detail at the EM / Tidy3D side.

Hopefully that is clear? Let me know if you have questions (through GitHub could be a little easier than lark). 

Thanks for your work on this!

setup.py
magiWei: added

tidy3d/web/api/container.py
```diff
@@ -9,18 +9,19 @@
...
+from tidy3d.components.data.sim_data import SimulationData
+from tidy3d.log import log, get_logging_console
 
-from ..exceptions import DataError
+from tidy3d.exceptions import DataError
 
 
 DEFAULT_DATA_PATH = "simulation_data.hdf5"
-DEFAULT_DATA_DIR = "."
+DEFAULT_DATA_DIR = ".."
```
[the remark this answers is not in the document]
magiWei: When files are moved, Pycharm automatically sets it.  Modified.

tidy3d/web/core/http_util.py
magiWei: modified

tidy3d/web/core/log.py
magiWei: modified

tidy3d/web/api/asynchronous.py
magiWei: modified

tidy3d/web/api/config.py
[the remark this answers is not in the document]
magiWei: modified

tidy3d/web/api/tidy3d_stub.py
magiWei: added

tylerflex (changes requested):
Thanks @magiWei I think this is looking pretty close. Could you make the last few changes requested here and then I think I'll ask @dbochkov-flexcompute to take a look because it will be important to make sure that we can build the heat solver on top of this too. Thank you for your help with this and patience with us as we worked on defining and communicating what we want out of this refactor. I think it's starting to look much better.

tidy3d/web/api/webapi.py
```diff
@@ -4,72 +4,35 @@
...
-    if decorated_fn:
-        return decorator(decorated_fn)
-
-    return decorator
-
 
 @wait_for_connection
 def run(
-    simulation: Simulation,
+    stub: Tidy3dStub,
```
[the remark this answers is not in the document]
dbochkov-flexcompute: to clarify: is the intended usage from user perspective eventually
```
 web.run(simulation=Tidy3dStub(component=sim))
```
or
```
 web.run(simulation=sim)
```
?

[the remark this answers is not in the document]
tylerflex: the 2nd, the Tidy3DStub is created within the webapi to interface with MC's template `StubTask`

tidy3d/web/api/webapi.py
```diff
@@ -434,7 +391,7 @@ def monitor_preprocess() -> None:
 @wait_for_connection
 def download(
     task_id: TaskId,
-    path: str = "simulation_data.hdf5",
+    path: str = "output/monitor_data.hdf5",
```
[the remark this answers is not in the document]
magiWei: modified

tidy3d/web/core/http_util.py
```diff
@@ -91,8 +133,8 @@ def wrapper(*args, **kwargs):
         resp = func(*args, **kwargs)
 
         if resp.status_code != ResponseCodes.OK.value:
-            if resp.status_code == ResponseCodes.NOT_FOUND.value:
```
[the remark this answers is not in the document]
magiWei: web.get_info(task_id), if the task_id is not exist, the API server will throw ResponseCodes.NOT_FOUND(404) exception, so if resp.status_code == ResponseCodes.NOT_FOUND.value: retun null, the usage will error with null.
I think need to throw error directly.

tidy3d/web/api/webapi.py
```diff
@@ -561,8 +516,8 @@ def load(
     replace_existing: bool = True,
     verbose: bool = True,
     progress_callback: Callable[[float], None] = None,
-) -> SimulationData:
-    """Download and Load simulation results into :class:`.SimulationData` object.
+) -> Tidy3dStubData:
+    """Download and Load simulation results into :class:`.StubData` object.
```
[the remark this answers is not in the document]
magiWei: modified

tidy3d/web/core/s3utils.py
```diff
@@ -158,7 +157,6 @@ def _get_progress(action: _S3Action):
         TransferSpeedColumn(),
         "•",
         TimeRemainingColumn(),
-        console=get_logging_console(),
```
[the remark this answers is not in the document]
magiWei: you mean: ↑ simulation.hdf5.gz ━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 1.4/1.4 kB • ? • 0:00:00

I tested, it works ok.

tidy3d/web/api/webapi.py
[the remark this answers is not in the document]
dbochkov-flexcompute: so, eventually we will change the signature of `run()` to something like
```
SimulationType = [Simulation, ModeSolver, HeatSimulation]
SimulationDataType = [SimulationData, ModeSolverData, HeatSimulationData]

def run(simulation: SimulationType, ...) -> SimulationDataType:
    ...
```
?

tylerflex:
@lucas-flexcompute when you have time could you take a look at this refactor and see if you have any comments about it, especially in the context of the mode solver webapi? The idea is to have a single template for submittable tasks controlled by MC and implement this template for various tidy3d objects (simulation, heat simulation, mode solver), each using the same webapi. Thanks

dbochkov-flexcompute:
Some more thoughts:
> SimulationTask.get_running_info(self) seems to be specific to FDTD. Other solvers will not have perc_done and field_decay values, but might have their own variables (perhaps, freqs_done for ModeSolver). Would it make sense to change it so that it gets a general dict from resp? For example, run_info = resp.get("run_info"), then for FDTD run_info will have entries run_info["perc_done"] and run_info["field_decay"]?

Maybe a temporary solution could be to change `SimulationTask.get_running_info(self)` so that it returns `{"perc_done": perc_done, "field_decay": field_decay}` for FDTD and an empty dictionary `{}` for other solvers until we establish what running info we want to get for `ModeSolver`, `Heat`, etc?

tidy3d/web/api/webapi.py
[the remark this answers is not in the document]
tylerflex: yea

tylerflex:
Yea your understanding is correct.

> * `SimulationTask.get_running_info(self)` seems to be specific to FDTD. Other solvers will not have `perc_done` and `field_decay` values, but might have their own variables (perhaps, `freqs_done` for `ModeSolver`). Would it make sense to change it so that it gets a general dict from `resp`? For example, `run_info = resp.get("run_info")`, then for FDTD `run_info` will have entries `run_info["perc_done"]` and `run_info["field_decay"]`?

I think think your suggestion to generalize so that all solvers have some kind of running info would be good. And then we can have keys specific to the individual solvers, perhaps. The ideal would be to generalize the `webapi` and `Tidy3dStub` so that there is a single logic for running and the different solvers just need to implement the various methods required by the webapi (eg. a method that defines when a solver is "done" and ready to download). A lot of work is still left to do I think defining these for all the solvers and coming up with a unified approach, but this can be done totally on our side without MC's help.

> * What is the exact purpose of `TaskStubData`? It seems to me that `TaskStub` serves as a template what needs to be passed to `SimulationTask`. But `SimulationTask` doesn't create `TaskStubData`, it just directly downloads simulation data hdf5. `TaskStubData` is only created in `web/api/...`, but do we really need `TaskStubData` at that stage?

hm, perhaps not, I think the purpose of task stub data is to do what I described in the paragraph above (generalize everything) but it's possible we don't need it. However, I wonder if on the solver side if there is some need for it to define a template for results that are to be packaged for the user to download and use. What do you think? 

> * What is the remaining plan for this PR? Currently, it seems `web/api/...` breaks backward compatibility and is also doesn't seem like something we want eventually. Is the plan for the EM team to change `web/api/...` to whatever we need once MC team is done with splitting code into `web/api/...` and `web/core/...`?

I think maybe we can add a backwards compatible `web/webapi.py` `web/asynchronous.py` `web.container.py` that just call the `web/api/` equivalents for now. I think we haven't decided on the overall end goal, whether to keep `web/api` and `web/core` and have MC develop in `core`, or move all the `web/api` stuff back into `web/` or have `core` be it's own package. Any preferences?

magiWei:
> I think maybe we can add a backwards compatible web/webapi.py web/asynchronous.py web.container.py that just call the web/api/ equivalents for now. I think we haven't decided on the overall end goal, whether to keep web/api and web/core and have MC develop in core, or move all the web/api stuff back into web/ or have core be it's own package. Any preferences?

@tylerflex 
Backwards compatible? Meaning that the users can use both methods?
`
sim = create_sim()
web.run(sim: Simulation) :
      web.run(Tidy3dStub(component= sim))
  or
stub = Tidy3dStub(component= sim)
web.run(stub)
  `    
Does this mean that users have multiple ways to run task, which conflicts with our current refactoring at the moment, because we just want Simulation, ModeSolver, HeatSolver etc. all types use `web.run( Tidy3dStub(component= sim))` instead of `web.run(sim:` Simulation) or  web.run(sim:` ModeSolver) or ...` . 

Maybe we could release a big version of tidy3d-3.0.0,  if the existing old users use the new API at the expense of their old habits.

tylerflex:
So the backwards compatibility is to ensure that any tidy3d scripts in 2.* will run with newer versions. So to summarize, we need to worry about a few things:
* the webapi functions, eg. `web.run()` must be importable from `web`.
* these functions must accept `simulation` as an argument, although we can change the type of the `simulation` to include the other objects like `Union[Simulation, ModeSolver]`, etc. main thing is that a user running `web.run(simulation=x)` will not error with the new update.
So this is what

So in this PR, the user needs to first create the `Tidy3dStub()` and then pass to the webapi functions.
```py
web.run(Tidy3dTask(simulation))
```
I think instead we want to pass the component directly (like it is currently, a type `Union[Simulation, ModeSolver, ...]`) and then construct the `Tidy3dStub` internally.
```py
web.run(simulation) #  construct `Tidy3dStub(simulation)` internally
```
It seems unnecessary for the user to have to worry about creating the stub since it is mostly just something used internally to generalize the webapi process.

I didn't realize this was how it was set up in this PR. @magiWei could you make this change when you get a chance? sorry for the confusion there. So to clarify, we keep `Tidy3DStub` in the webapi but for functions that accept `simulation` arguments, we pass the component and create the `Tidy3dStub` instance internally so the user never sees it.

Does that make sense @dbochkov-flexcompute ?

dbochkov-flexcompute:
yeah, makes sense to me. I agree that all user-facing functions/classes should either keep `simulation: Simulation` or upgrade it to `simulation: Union[Simulation, ModeSolver]`. And conversion into `Tidy3dStub` should be hidden from user. Once this PR is merged, we can further upgrade it to `simulation: Union[Simulation, ModeSolver, HeatSimulation]`.

Should we also preserve `from tidy3d.plugins.mode.web import run` for now, and maybe get rid of it at 3.0?

tidy3d/web/core/s3utils.py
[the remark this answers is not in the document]
dbochkov-flexcompute: does this keep console width of 80 characters? The reason we switched to use `get_logging_console()` is to make all rich output have the same width and display nicely in our docs

previously:
console_without

now:
console_with

tylerflex:
Yes I think we will need to still support 
```
from tidy3d.plugins.mode.web import run
```
but we could in principle just have `mode.web` import  `run` from `web/api/webapi.py` and just pass it

magiWei:
> So the backwards compatibility is to ensure that any tidy3d scripts in 2.* will run with newer versions. So to summarize, we need to worry about a few things:
> 
> * the webapi functions, eg. `web.run()` must be importable from `web`.
> * these functions must accept `simulation` as an argument, although we can change the type of the `simulation` to include the other objects like `Union[Simulation, ModeSolver]`, etc. main thing is that a user running `web.run(simulation=x)` will not error with the new update.
>   So this is what
> 
> So in this PR, the user needs to first create the `Tidy3dStub()` and then pass to the webapi functions.
> 
> ```python
> web.run(Tidy3dTask(simulation))
> ```
> 
> I think instead we want to pass the component directly (like it is currently, a type `Union[Simulation, ModeSolver, ...]`) and then construct the `Tidy3dStub` internally.
> 
> ```python
> web.run(simulation) #  construct `Tidy3dStub(simulation)` internally
> ```
> 
> It seems unnecessary for the user to have to worry about creating the stub since it is mostly just something used internally to generalize the webapi process.
> 
> I didn't realize this was how it was set up in this PR. @magiWei could you make this change when you get a chance? sorry for the confusion there. So to clarify, we keep `Tidy3DStub` in the webapi but for functions that accept `simulation` arguments, we pass the component and create the `Tidy3dStub` instance internally so the user never sees it.
> 
> Does that make sense @dbochkov-flexcompute ?

Change 
def run(
    simulation: Tidy3dStub(component=simulation),
) -> Tidy3dStubData:

to 

def run(
    simulation: Union[Simulation, ModeSolver],
) -> Union[SimulationData, ModeSolverData]:

can backwards compatibility , It's better.

tidy3d/web/core/s3utils.py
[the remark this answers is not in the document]
magiWei: @tylerflex If tidy3d/core remove `get_logging_console()` and using logging destroy `width and display nicely `, so tidy3d/web should copy tidy3d/api `get_logging_console()`

tylerflex:
> Change
> 
> ```python
> def run(
>     simulation: Tidy3dStub(component=simulation),
> ) -> Tidy3dStubData:
> ```
> 
> to
> 
> ```python
> def run(
>     simulation: Union[Simulation, ModeSolver],
> ) -> Union[SimulationData, ModeSolverData]:
> ```
> 
> can backwards compatibility , It's better.

@magiWei agreed, let's do it that way. thanks!

tests/_test_local/_test_web.py
```diff
@@ -34,17 +34,13 @@ def test_get_email_passwd_auth_key_ok(self, set_authentication_config):
 
 def test_webapi_0_run(tmp_path):
     """test complete run"""
-    _ = web.run(
-        simulation=sim_original, task_name="test_webapi", path=str(tmp_path / "sim_data.hdf5")
-    )
+    _ = web.run(stub=sim_original, task_name="test_webapi", path=str(tmp_path / "sim_data.hdf5"))
```
lucas-flexcompute, line 37: Can't we keep the name `simulation` for the arguments in the web API? `stub` doesn't indicate what the argument really is, and `simulation` should work for all kinds of simulation types in the future, I guess.

tests/test_web/test_material_fitter.py
```diff
@@ -1,27 +1,32 @@
...
     monkeypatch.setattr(http_module, "api_key", lambda: "apikey")
 
 
 @responses.activate
 def test_material_fitter(monkeypatch, set_api_key):
-    fitter = DispersionFitter.from_file("tests/data/nk_data.csv", skiprows=1, delimiter=",")
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    parent_dir = os.path.dirname(current_dir)
+    file_path = os.path.join(parent_dir, "data/nk_data.csv")
+    fitter = DispersionFitter.from_file(file_path, skiprows=1, delimiter=",")
```
lucas-flexcompute, line 27: I usually prefer to use the [`tmp_path` fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) instead of creating data in-tree and having to clean it up manually later.
Also, `pathlib` is more flexible for working with paths than `os.path`.

tests/test_web/test_webapi.py
```diff
@@ -2,26 +2,40 @@
...
 TASK_NAME = "task_name_test"
 TASK_ID = "1234"
+FOLDER_ID = "1234"
 CREATED_AT = "2022-01-01T00:00:00.000Z"
 PROJECT_NAME = "default"
 FLEX_UNIT = 1.0
 EST_FLEX_UNIT = 11.11
+FILE_SIZE_GB = 4.0
+
+path = "tidy3d.web.core.task_core"
```
lucas-flexcompute, line 37: Maybe rename to something like `task_core_path`?

tests/test_web/test_webapi.py
```diff
@@ -232,8 +278,8 @@ def mock_webapi(
 def test_source_validation(mock_upload):
     sim = make_sim().copy(update={"sources": []})
     assert upload(sim, TASK_NAME, PROJECT_NAME, source_required=False)
-    with pytest.raises(SetupError):
-        upload(sim, TASK_NAME, PROJECT_NAME)
+    # with pytest.raises(SetupError):
+    # upload(Tidy3dStub(component=sim), TASK_NAME, PROJECT_NAME)
```
lucas-flexcompute, line 282: Isn't it a problem if the user uploads a simulation without a source?

tidy3d/web/api/tidy3d_stub.py
```diff
@@ -0,0 +1,164 @@
...
+from ...components.data.sim_data import SimulationData
+from ...components.data.monitor_data import ModeSolverData
+from ..core.types import TaskType
+from ...components.simulation import Simulation
+from ...plugins.mode.mode_solver import ModeSolver
+
+
+class Tidy3dStub(BaseModel, TaskStub):
+
+    component: Union[Simulation, ModeSolver]
```
lucas-flexcompute, line 21: After the 3.0 refactor, I'm guessing `Simulation` will be renamed to `FDTD` or something like that, right? Wouldn't it be more descriptive to have this field named `simulation` than `component`?

lucas-flexcompute (commented):
Seem pretty reasonable to me, considering the comments already made by @dbochkov-flexcompute and @tylerflex. I agree that the `*Stub` classes should be hidden from the user interface. That's something that can be handled internally only. I'd go as far as saying that those classes should be in fact hidden from the user interface, so that we can adapt them in the future more freely without being afraid of breaking user code (maintaining only internal backwards compatibility).

tidy3d/web/core/environment.py
```diff
@@ -1,10 +1,9 @@
 """Environment Setup."""
 import os
+import logging as log
```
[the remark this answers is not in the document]
dbochkov-flexcompute: @tylerflex, I forgot, is `web/core` not supposed to import tidy3d components?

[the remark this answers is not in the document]
dbochkov-flexcompute: hmm, this is actually related to what we were discussing above about using `get_logging_console()`. I guess if we are adamant on not importing any tidy3d functionality in `web/core` then we have to somehow make sure that formatting (currently just console width) is consistent between tidy3d and `web/core`. Or, alternatively, we can relax this restriction and allow import of the top level tidy3d components, e.g., `tidy3d.log`, `get_logging_console()`, etc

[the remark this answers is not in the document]
lucas-flexcompute: I think I missed this discussion, so I guess my comment is incorrect in that case. However, I think top-level modules should be accessible to `web.core`, so that exceptions and logging are shared and consistent across the whole module.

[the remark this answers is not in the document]
tylerflex: the problem is that we were considering making `web/core` its own separate package potentially. However, I agree that if we keep it as part of `tidy3d`, then we could give it access to the top level stuff like exceptions, logging (even `__version__`). Having it as a separate package makes development easier between the EM and MC teams though as we can work independently and then just sync up after releases. Do you have any preferences or thoughts regarding the decision between keeping `web/core` in tidy3d vs having it separate?

tests/_test_local/_test_web.py
tylerflex: yea we agreed it should be `simulation`, I think this test maybe never got reverted (and was maybe missed since it's a local test). @magiWei could you change the `_test_web` to use `simulation` instead of `stub` as the kwarg? thanks.

tests/test_web/test_webapi.py
tylerflex: yea this should probably not be commented out. I think this is caught as a `pre_upload_validator()`, which we should still be doing. Let's add this test back @magiWei

tidy3d/web/api/tidy3d_stub.py
tylerflex: I think we haven't decided this yet but you're probably right. On the one hand, we could wait until 3.0 to change the naming, but maybe `simulation` is best also because it is consistent with the webapi as it stands so I'm ok with changing it.

tests/test_web/test_webapi.py
magiWei: Modified.

tests/test_web/test_webapi.py
magiWei: Modified.

tidy3d/web/api/tidy3d_stub.py
magiWei: There is one Enum named TaskType at GUI: 

class TaskType:
 FDTD, MODE_SOLVER, SWEEP, HEAT.
 
 we could wait until 3.0 to change the naming.

tests/_test_local/_test_web.py
magiWei: Modified.

tidy3d/web/core/environment.py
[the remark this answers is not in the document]
dbochkov-flexcompute: well, even as a separate repo, it could still import tidy3d, right? Like denormalizer, for example. Although there it is a very crucial component, not just for logging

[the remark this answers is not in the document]
magiWei: If it is possible to import tidy3d, will it create circular dependencies? If it is split into 2 repositories, how will this dependency be decoupled?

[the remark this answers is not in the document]
lucas-flexcompute: Python (pip, conda, etc.) handles cyclic dependencies between packages just fine as far I know. It wouldn't be a problem to have tidy3d and web being separate, co-dependent packages. OTOH, that's not the most elegant solution, which would be to have a 3rd base package with the core functionality (log, exceptions, etc.).

I'm not 100% aware of what would be the duplicated functionality between web and tidy3d right now, but if it is only logging, we could have the core module accept a `log` object that would be used internally and that we can set when importing it from tidy3d. For example:

```python
# In tidy3d.__init__.py
import tidy3d_core
from .log import log

tidy3d_core.set_logger(log)
```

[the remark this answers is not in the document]
tylerflex: I think I'm in favor of ensuring that `web/core` does not import from tidy3d at all just in case we want to make it a separate package eventually without worrying about circular dependency. So perhaps the best approach to logging is to handle it how Lucas suggests here and provide an API for `web/core` to set its logger.  Maybe we can do something similar for the console (`tidy3d_core.set_console_getter(get_console)`)

tidy3d/web/api/tidy3d_stub.py
tylerflex: yea let's not change this naming, but I think rather this is about what to call the `Tidy3dStub.component` field name. probably changing that to `Tidy3dStub.simulation` makes some sense, let's go with that.

tidy3d/web/core/environment.py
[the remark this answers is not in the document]
dbochkov-flexcompute: Sounds good to me. Any arguments for setting a console getter vs a console instance `.set_console(get_logging_console())`? also, should log and console have some default values if `.set_logger()` and `.set_console()` are not called?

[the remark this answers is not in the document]
tylerflex: I suggested this because we call `get_logging_console()` in the webapi instead of importing a `Console` instance. I think it needs to be this way because the `console` represents a specific output (rather than `log` which is more a handler for output). But I'm not actually sure on this, if we can get away with a specific console instance, that might be easier than setting a getter function even in our webapi.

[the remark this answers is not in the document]
magiWei: tidy3d
```python
# In tidy3d.__init__.py
from .core import core_log
from ..log import log, get_logging_console

core_log.set_logger(log, get_logging_console())
```

tidy3d/core add core_log.py

```python
logger_setting = {
    "logger": "",
    "logger_console": "",
}

def set_logger(logger, logger_console):
    logger_setting["logger"] = logger
    logger_setting["logger_console"] = logger_console
```

So, s3util can use 
```python

def _get_progress(action: _S3Action):
    """Get the progress of an action."""

    col = (
        TextColumn(f"[bold green]{_S3Action.DOWNLOADING.value}")
        if action == _S3Action.DOWNLOADING
        else TextColumn(f"[bold red]{_S3Action.UPLOADING.value}")
    )
    return Progress(
        col,
        TextColumn("[bold blue]{task.fields[filename]}"),
        BarColumn(),
        "[progress.percentage]{task.percentage:>3.1f}%",
        "•",
        DownloadColumn(),
        "•",
        TransferSpeedColumn(),
        "•",
        TimeRemainingColumn(),
        console=get_logger_console(),
    )
```
Merged the code to stub, have a check, any discuss pls let me know.

tidy3d/web/api/tidy3d_stub.py
magiWei: `Tidy3dStub.component` changed to `Tidy3dStub.simulation`.

tidy3d/web/core/environment.py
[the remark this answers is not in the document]
tylerflex: I think this looks pretty good. Only remaining question is whether it Is ok that the logging console is an instance here instead of a getter function. Can we test again running the WebAPI.ipynb notebook to ensure that it formats correctly? if so I think I'm happy with this change.

Small note: I think `get_logger_console()` will already return `None` if no logger is set, so maybe we don't need the `or None` part in `console=get_logger_console() or None`?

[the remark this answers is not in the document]
dbochkov-flexcompute: Not sure if the notebook will work until the corresponding changes on web side will be done. But yeah, passing a getter function is probably safer. In fact, `set_logging_console()` method from `tidy3d.log` changes console instance returned by `get_logging_console()`.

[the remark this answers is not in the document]
magiWei: yea, changed `console=get_logger_console() or None` to `console=get_logger_console()`

tidy3d/web/core/exceptions.py
```diff
@@ -0,0 +1,11 @@
+"""Custom Tidy3D exceptions"""
+import logging as log
```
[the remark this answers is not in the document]
magiWei: changed to 
```python
# In tidy3d.__init__.py
from .core import core_log
from ..log import log, get_logging_console

core_log.set_logger(log, get_logging_console())
```

tests/test_web/test_material_fitter.py
magiWei: Good idea.

magiWei:
@tylerflex Can you double-check if there are any disputes or changes that need to be made, if not, @dbochkov-flexcompute Daniil will co-build HeatSimulation based on the stub.

tylerflex (changes requested):
Looks almost ready to me @magiWei 

There's a comment about the `webapi.py` only exposing regular `tidy3d` components to the user (so stub and stub data are only handled internally).

Also, I think I'm not sure about the get_logging_console() because we might want that to be a getter function instead of an instance.

> Not sure if the notebook will work until the corresponding changes on web side will be done. But yeah, passing a getter function is probably safer. In fact, set_logging_console() method from tidy3d.log changes console instance returned by get_logging_console().

I guess we can try it out and see how it works.

The last thing is that it might make sense to hook `tidy3d/plugins/mode/web` up with this webapi first to ensure that it works as expected before integrating heat solver. What do you think @lucas-flexcompute @dbochkov-flexcompute ? would the mode solver be able to integrate with this webapi for testing?

lucas-flexcompute:
The newer version of the solver should work from what I can see (using a single gzipped hdf5), but I don't think it works with the older versions that required submitting simulation and mode solver separately. Would that be a problem?

dbochkov-flexcompute:
> The last thing is that it might make sense to hook tidy3d/plugins/mode/web up with this webapi first to ensure that it works as expected before integrating heat solver. What do you think @lucas-flexcompute @dbochkov-flexcompute ? would the mode solver be able to integrate with this webapi for testing?

As I understand, the plan is that mode solver will not work through stub yet, because of additional work required on web side. Instead only regular fdtd and heat will be supported initially. @magiWei please correct me if I am wrong

magiWei:
> > The last thing is that it might make sense to hook tidy3d/plugins/mode/web up with this webapi first to ensure that it works as expected before integrating heat solver. What do you think @lucas-flexcompute @dbochkov-flexcompute ? would the mode solver be able to integrate with this webapi for testing?
> 
> As I understand, the plan is that mode solver will not work through stub yet, because of additional work required on web side. Instead only regular fdtd and heat will be supported initially. @magiWei please correct me if I am wrong

Yes, stub just for FDTD and Heat in the original plan, but I have a new idea: we can add mode solver to the stub with the following logic: 

```python
def run(
    simulation: Union[Simulation, ModeSolver, HEAT],
) -> Union[SimulationData, ModeSolverData, HeatData]:

if Simulation or HEAT:
   # the same logic

elif ModeSolver:
  # ModeSolver logic
````
The code entry is the same, only the internal implementation will be differentiated. And there is another ticket for mode solver can be used independently of FDTD from  @Leixin. After finishing it, we can use the same logic:
```python
def run(
    simulation: Union[Simulation, ModeSolver, HEAT],
) -> Union[SimulationData, ModeSolverData, HeatData]:

if Simulation or HEAT or ModeSolver:
   # the same logic

````
 So, we only need to modify the internal implementation and not change the user's behavior anymore.
 
 @tylerflex @dbochkov-flexcompute @lucas-flexcompute how do you think?

**Comment by magiWei** [comment]
> The newer version of the solver should work from what I can see (using a single gzipped hdf5), but I don't think it works with the older versions that required submitting simulation and mode solver separately. Would that be a problem?

The original product design regarded the mode solver as a module of FDTD. but there is another ticket about mode solver can be used independently of FDTD from @leixin, from this perspective, it was not a problem I think.

**Comment by dbochkov-flexcompute** [comment]
> Yes, stub just for FDTD and Heat in the original plan, but I have a new idea: we can add mode solver to the stub with the following logic:
> Customers can use the same run function, only the internal implementation will be differentiated. And there is another ticket about mode solver can be used independently of FDTD from @leixin. After finishing it, we can use the same logic for all task type, we only need to modify the internal implementation and not change the user's behavior anymore.:

That could work. However, I don't think there is rush to allow running mode solver through the same `run()`, I think we can do  that in a separate PR. We just need to make sure that user still can do `from tidy3d.plugins.mode.web import run as run_mode_solver` in this PR. I believe what's more important is to work on integrating and testing heat solver web api

**Comment by magiWei** [comment]
> > Yes, stub just for FDTD and Heat in the original plan, but I have a new idea: we can add mode solver to the stub with the following logic:
> > Customers can use the same run function, only the internal implementation will be differentiated. And there is another ticket about mode solver can be used independently of FDTD from @leixin. After finishing it, we can use the same logic for all task type, we only need to modify the internal implementation and not change the user's behavior anymore.:
> 
> That could work. However, I don't think there is rush to allow running mode solver through the same `run()`, I think we can do that in a separate PR. We just need to make sure that user still can do `from tidy3d.plugins.mode.web import run as run_mode_solver` in this PR. I believe what's more important is to work on integrating and testing heat solver web api

yes, heat is a new type, but we must first agree on the stub to avoid changes in user behavior later in the heat type.

**Comment by dbochkov-flexcompute** [comment]
I added a commit with improvements here https://github.com/flexcompute/tidy3d/tree/daniil/stub, feel free to merge it fully or partially as what you think is best. Changes:
- allowed simulation types are formalized using `SimulationType = Union[Simulation, ...]`, `SimulationDataType = Union[SimulationData, ...]`
- Removed any mentions of stub in user facing docstring
- set console getter instead of console instance
- prepare `monitor()` for other simulation types

**Comment by magiWei** [comment]
> I added a commit with improvements here https://github.com/flexcompute/tidy3d/tree/daniil/stub, feel free to merge it fully or partially as what you think is best. Changes:
> 
> * allowed simulation types are formalized using `SimulationType = Union[Simulation, ...]`, `SimulationDataType = Union[SimulationData, ...]`
> * Removed any mentions of stub in user facing docstring
> * set console getter instead of console instance
> * prepare `monitor()` for other simulation types

https://github.com/flexcompute/tidy3d/blob/daniil/stub/tidy3d/web/api/webapi.py#L145-L147 There some code about Tidy3dStub

**Comment by dbochkov-flexcompute** [comment]
> There's a comment about the webapi.py only exposing regular tidy3d components to the user (so stub and stub data are only handled internally).

Do you mean that no user-facing function should be expecting `Tidy3dStub/Data` as an argument or returning it? I think they currently don't, it's just some of functions still (incorrectly) have Tidy3dStub in their docstring or return type. If that's the issue, then it is fixed here https://github.com/flexcompute/tidy3d/tree/daniil/stub

**Comment by tylerflex** [comment]
> Do you mean that no user-facing function should be expecting Tidy3dStub/Data as an argument or returning it? I think they currently don't, it's just some of functions still (incorrectly) have Tidy3dStub in their docstring or return type. If that's the issue, then it is fixed here https://github.com/flexcompute/tidy3d/tree/daniil/stub

Yea that's what I meant. sounds good.

**Comment by tylerflex** [comment]
> That could work. However, I don't think there is rush to allow running mode solver through the same run(), I think we can do that in a separate PR. We just need to make sure that user still can do from tidy3d.plugins.mode.web import run as run_mode_solver in this PR. I believe what's more important is to work on integrating and testing heat solver web api

@magiWei the solution for mode solver seems good, but maybe @dbochkov-flexcompute is right we can do this at a later PR and focus on the heat solver.

I think @dbochkov-flexcompute if this PR is in a place that you can start integrating the heat solver with, should we move to merge it soon? and into `develop`?

I just tested this PR on the `StartHere.ipynb` notebook and my task failed in `web.monitor()`. It seems related to `tidy3dMetaDataPipeline.py`. Maybe we need to rebase this against `develop` branch as `pre/2.4` is stale now.  I see there are many merge conflicts with `develop`, maybe when those are fixed it will work?

Maybe let's come up with a plan to wrap this up. proposal:
1. Rebase this PR against the destination branch (`develop` I guess, since we don't have `pre/2.5` yet) and fix merge conflicts.
2. Test on a few notebooks that use the webapi in different ways (WebAPI.ipynb, StartHere.ipynb, ParameterScan.ipynb, AdjointPluglin1...ipynb) and fix any bugs that might come up.
3. Wrap up all final issues / code formatting, final review and merge?

Then we can work on integrating heat solver and mode solvers in their own PRs? Thoughts?

**Comment by dbochkov-flexcompute** [comment]
Sounds like a good plan to me!

**Comment by magiWei** [comment]
I have merged Daniil's modifications(https://github.com/flexcompute/tidy3d/tree/daniil/stub) and develop branch to stub branch. pls have a check. @tylerflex @dbochkov-flexcompute

**Comment by tylerflex** [comment]
Thanks @magiWei. I think it looks good, I changed the base from pre/2.4 to develop on this PR.

I just tried running it using `WebAPI.ipynb` [notebook](https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/develop/docs/source/notebooks/WebAPI.ipynb) and got an error in `web.monitor()` (on the server side). Could you or @dbochkov-flexcompute look into this? I think once we clear out all of these issues when running tidy3d jobs in practice, we can merge next. Thanks!

**Comment by dbochkov-flexcompute** [comment]
> I just tried running it using WebAPI.ipynb [notebook](https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/develop/docs/source/notebooks/WebAPI.ipynb) and got an error in web.monitor() (on the server side). Could you or @dbochkov-flexcompute look into this? I think once we clear out all of these issues when running tidy3d jobs in practice, we can merge next. Thanks!

Couldn't figure out immediately what the issue is. @magiWei, will this PR work with current web endpoints or they need some additional work too?

**Comment by dbochkov-flexcompute** [comment]
Oh, didn't realize that @magiWei responded as an edit of my comment:
> I tested with tidy3d-2.4.1(solver version: release-23.3.2) on dev, it works ok. The 3 methods:
>
> ```
> #method 1
> sim = create_task_sim()
> data = web.run(sim, task_name="Simulation")
>
> #method 2
> sim = create_task_sim()
> task_id = web.upload(sim, task_name="Simulation")
> web.start(task_id)
> web.monitor(task_id)
> data = web.load(task_id)
>
> # method3
> use the code in https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/develop/docs/source/notebooks/WebAPI.ipynb
> ```
> @tylerflex so, in which environment did you test it? if you used dev, the original solver mapping is tidy3d-2.4.1 -> > parallel_mode-0.0.0, it didn't work properly. but, I modified the solver mapping to tidy3d-2.4.1 -> release-23.3.2, it works ok.

I think both Tyler and I were trying to run this in prod

**Comment by tylerflex** [comment]
Yea I was running with prod

**Comment by dbochkov-flexcompute** [comment]
Hmm, could it be that webapi sends a wrong protocol version? It seems like in prod it doesn't get picked up by the right version

Actually, I just realized that that there seems to be some breaking changes to `configure` function? That is, it now also requires specifying tidy3d version https://github.com/flexcompute/tidy3d/pull/1108/files#diff-489dddc0bbe0f0b29e418725e0bf26f06bffa9ecbebb80116861a3443d26b7a1R47 and tidy3d config file actually stores it. Does it mean that users will have to call `configure` every time they update tidy3d version? It seems very inconvenient. 

This seems to be causing issues when Tyler or I are trying to run using this branch (I assume neither of us ran `configure`). Using `export SIMCLOUD_VERSION=2.4.1` makes `WebAPI.ipynb` work as expected, but I don't we want to keep specifying tidy3d version either in env variable `SIMCLOUD_VERSION` or in `.tidy3d/config`, do we @tylerflex ? I guess this change must have come from the requirement that `web/core` doesn't import anything from `tidy3d`, but maybe importing `__version__` is OK? Or should we pass tidy3d version through stub?

**Comment by tylerflex** [comment]
Yea that's not great (regarding version) maybe instead we can set the version the same way we set the logger, which will be done at import tidy3d time? @magiWei

File: tidy3d/plugins/adjoint/web.py [hunk 32]
```diff
@@ -65,7 +65,7 @@ def _task_name_adj(task_name: str) -> str:
 
 def tidy3d_run_fn(simulation: Simulation, task_name: str, **kwargs) -> SimulationData:
     """Run a regular :class:`.Simulation` after conversion from jax type."""
-    return web_run(simulation=simulation, task_name=task_name, **kwargs)
+    return web_run(stub=simulation, task_name=task_name, **kwargs)
```
dbochkov-flexcompute, line 68: should this be `return web_run(simulation=simulation, task_name=task_name, **kwargs)`

dbochkov-flexcompute:
seems to be working for me now, I tried this notebooks so far:
- StartHere.ipynb - good
- WebAPI.ipynb - good
- AdjointPlugin1Intro.ipynb - good
- AdjointPlugin2GradientChecking.ipynb - good
- ModeSolver.ipynb - fail. This is because `from tidy3d.plugins.mode.web import run as run_mode_solver` doesn't work. I think we should just move `web/api/web.py` back into `plugins/mode/web.py` for now.

a couple of other things:
- `configure(apikey, version)` should go back to just `configure(apikey)`
- argument name `stub` -> `simulation` in tidy3d/plugins/adjoint/web.py

magiWei:
> * ModeSolver.ipynb - fail. This is because `from tidy3d.plugins.mode.web import run as run_mode_solver` doesn't work. I think we should just move `web/api/web.py` back into `plugins/mode/web.py` for now.

if we move web/api/web.py back into plugins/mode/web.py,  plugins/xx will dependency on web/xx, It goes against the very principle of dependence that we started with: just tidy3d/web can dependency on tidy3d/core and plugins/xx.

> * configure(apikey, version) should go back to just configure(apikey)
 argument name stub -> simulation in tidy3d/plugins/adjoint/web.py
 
 Modified.

tidy3d/plugins/adjoint/web.py
tylerflex: I think so

dbochkov-flexcompute:
> If we move web/api/web.py back into plugins/mode/web.py, plugins/xx will dependency on web/xx, It goes against the very principle of dependence that we started with: just tidy3d/web can dependency on tidy3d/core and plugins/xx.

agree, but we still need to preserve backward compatibility for mode solver until 3.0 at least. I added a commit here with this modifications:
- added a dummy `plugins/mode/web.py` that imports from `tidy3d/web/api`
- renamed `tidy3d/web/api/web.py` into `tidy3d/web/api/mode.py` so that there is less confusion
- removed `KEY_VERSION` related things from `migrate.py`
- removed `KEY_VERSION`  and `SIMCLOUD_VERSION` from `core/constants.py` since they are not used anywhere
- fixed a typo in `plugins/adjoint/web.py` `simulations` -> `simulation`
- removed remaining user-facing `Tidy3dStubData` mentions, fixed docstring 
@magiWei please let me know if anything doesn't look right to you

Tried running the same notebooks: everything works for me now. So, I think this should be ready to merge into pre/2.5.0 if there are no other critical concerns

dbochkov-flexcompute:
also, since we are now polishing small details, if we end up adding more commits here let's not squash them, so it's easier to track changes

magiWei:
Everything works for me now.

tests/Simulation.json
tylerflex, line 1: can we remove this file or move it to `tests/sims` if it's needed for the tests?

tidy3d/web/__init__.py
```diff
@@ -11,16 +13,24 @@
...
-from .asynchronous import run_async
-from .webapi import test
+from .api.asynchronous import run_async
+from .core import core_config
+from ..log import log, get_logging_console
+from ..version import __version__
 
 migrate()
 
+core_config.set_config(log, get_logging_console(), __version__)
```
tylerflex, line 32: @dbochkov-flexcompute does passing the `Console()` instance seem to work ok in the notebooks?

tidy3d/web/api/connect_util.py
```diff
@@ -0,0 +1,68 @@
...
+
+        return web_fn_wrapped
+
+    if decorated_fn:
+        return decorator(decorated_fn)
+
+    return decorator
+
+
+def get_time_steps_str(time_steps) -> str:
```
tylerflex, line 49: note: at a later point, we can easily refactor these two functions

tylerflex (approved):
Looks good to me. If this is able to run our notebooks, I'd say it's probably ready to merge and we can start getting the heat solver in. Thanks @magiWei @dbochkov-flexcompute

tidy3d/web/__init__.py
dbochkov-flexcompute: this is not ideal, but works ok in notebooks and I don't think we will have issues with this

tests/Simulation.json
dbochkov-flexcompute: removed

dbochkov-flexcompute:
just re-ran some notebooks again, everything is good

tylerflex:
ok, I think from my perspective, we can merge this. Let's do it into a new `pre/2.5` branch first, though.

PR #1130, opened 2023-08-29
Frequency lower bound and target_neff shift in mode solver

weiliangjin2021:


tidy3d/plugins/mode/solver.py
```diff
@@ -168,6 +170,12 @@ def compute_modes(
             target = mode_spec.target_neff
         target_neff_p = target / np.linalg.norm(kp_to_k) + fp_eps
```
momchil-flex, line 171: let's remove this `fp_eps` shift here then; it maybe was meant to play the same purpose but doesn't always?

tidy3d/plugins/mode/solver.py
```diff
@@ -18,6 +18,8 @@
 TOL_EIGS = fp_eps
 # Tolerance for deciding on the matrix to be diagonal or tensorial
 TOL_TENSORIAL = 1e-6
+# shift target neff by this value, both rel and abs, whichever results in larger shift
+TARGET_SHIFT = -10 * fp_eps
```
momchil-flex, line 22: why negative?

tidy3d/plugins/mode/solver.py
weiliangjin2021: it might be too small and overflowed.

tidy3d/plugins/mode/solver.py
weiliangjin2021: i think either positive or negative is fine. Just randomly picked one.

tidy3d/plugins/mode/solver.py
momchil-flex: right, so let's remove it in favor of the new code.

momchil-flex:
Also could you please add fixes to changelog.

tidy3d/plugins/mode/solver.py
weiliangjin2021: Positive can be safer to have a positive n_eff?

momchil-flex: I don't really have any intuition about it

momchil-flex:
Looks good, please squash and merge.

Issue #1131, opened 2023-08-29
Make default tidy3d dir configurable via environment variables


yaugenst:
I am trying to run tidy3d on serverless containers (e.g. Google Cloud Run). These containers have read-only file systems, with only `/tmp/` being writable. It would be nice if tidy3d's base/config directory could be configured via an environment variable such as `TIDY3D_DIR`.

Going a bit further, do you think it would be a good idea for tidy3d to adopt the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), which would mean putting tidy3d's configuration in `$HOME/.config/tidy3d/` by default? Resolving config directory via some precedence rules would also be cool, but probably overkill (see, e.g., [how conda does it](https://conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#searching-for-condarc)).

tylerflex:
closing this as seems fixed by #1132

PR #1132, opened 2023-08-29
Make TIDY3D_DIR configurable via environment variable

yaugenst:
Tries to read the `TIDY3D_DIR` environment variable to set tidy3d's base directory and defaults to the current behavior. Partially solves #1131 and is what I am currently using.

### DIFF
--- a/tidy3d/web/cli/constants.py
+++ b/tidy3d/web/cli/constants.py
@@ -1,7 +1,14 @@
 """Constants for the CLI."""
-from os import getenv
+
+import os
 from os.path import expanduser
 
-TIDY3D_DIR = getenv("TIDY3D_DIR", f"{expanduser('~')}/.tidy3d")
+TIDY3D_BASE_DIR = os.getenv("TIDY3D_BASE_DIR", f"{expanduser('~')}")
+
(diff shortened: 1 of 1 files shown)


momchil-flex:
Thanks. Actually we have a related PR #1100 in which the folder is set to `/tmp/.tidy3d` if `~/.tidy3d` cannot be written into. That said we can also add the environment variable option. Note that currently this is only used for the api key, which can also be set through environment variable `SIMCLOUD_APIKEY`, in which case the config folder should not be needed.

momchil-flex:
I just merged #1100 ; could you modify your PR to set the default folder as per that modification, when not provided as env variable? Thanks!

yaugenst:
Oh great thanks! I was aware of the api key variable, but thought it didn't work. Turns out I am bad at reading, I used `SIMCLOUD_API_KEY`...
That probably makes this PR obsolete for now, it might still be useful though if in the future there are more global configuration options planned.

yaugenst:
Changed the environment variable to `TIDY3D_BASE_DIR` to better reflect its purpose according to #1100.

momchil-flex:
Thanks!

PR #1134, opened 2023-08-31
Changed source spectrum calculation to work for large number of time steps

caseyflex:
This should eliminate out of memory errors associated with computing the source spectrum when the number of time steps is very large. In my testing, it also sped up the calculation by a factor of 3 or so. The speed up increases as the number of time steps increases.

The current approach is just replacing the matrix operations with a loop over time, and reducing the number of np.exp calls as much as possible.

I also tried an FFT based approach. The frequency monitors implicitly assume the signal is zero outside of the simulation run time. In order to agree with this behavior, it was necessary to zero-pad the FFT to an extent determined by the mismatch between the frequencies at which we evaluate the spectrum and the FFT frequency grid. This was messier and may reduce precision (although the FFT approach was the fastest).

### DIFF
--- a/tidy3d/components/source.py
+++ b/tidy3d/components/source.py
@@ -26,7 +26,6 @@ from ..constants import inf
 from ..exceptions import SetupError, ValidationError
 from ..log import log
 
-
 # in spectrum computation, discard amplitudes with relative magnitude smaller than cutoff
 DFT_CUTOFF = 1e-8
 # when checking if custom data spans the source plane, allow for a small tolerance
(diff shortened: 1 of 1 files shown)


momchil-flex:
Nice! I briefly thought about this approach yesterday but somehow assumed that it won't be very efficient in python for a large number of steps... But good to see that it works well actually!

I realize the `DFT_CUTOFF` was kinda weird, potentially cutting out times during the pulse window that may get close to zero just due to the carrier frequency oscillation. I hadn't appreciated that, I was rather thinking of it as just truncating to that widnow. Maybe we should keep a `DFT_CUTOFF` for speed potentially, but properly apply it to only window the pulse (i.e. only truncate up to a given index and starting from another index until the end, if possible)?

caseyflex:
> Nice! I briefly thought about this approach yesterday but somehow assumed that it won't be very efficient in python for a large number of steps... But good to see that it works well actually!
> 
> I realize the `DFT_CUTOFF` was kinda weird, potentially cutting out times during the pulse window that may get close to zero just due to the carrier frequency oscillation. I hadn't appreciated that, I was rather thinking of it as just truncating to that widnow. Maybe we should keep a `DFT_CUTOFF` for speed potentially, but properly apply it to only window the pulse (i.e. only truncate up to a given index and starting from another index until the end, if possible)?

Good point, this could greatly help with speed in a large set of use cases. Revised accordingly

momchil-flex:
Nice! By thew way you could do just something like

```
        start_stop_filter = np.where((np.abs(time_amps) / np.amax(np.abs(time_amps))) > DFT_CUTOFF)
        # find first and last index where the filter is True
        start_ind = start_stop_filter[0][0]
        stop_ind = start_stop_filter[0][-1]
```

Do you think the handling of potentially different time steps adds a noticeable overhead in the many time steps case? If so I would be down to even consider refactoring this method to assume a uniform dt...

caseyflex:
> Nice! By thew way you could do just something like
> 
> ```
>         start_stop_filter = np.where((np.abs(time_amps) / np.amax(np.abs(time_amps))) > DFT_CUTOFF)
>         # find first and last index where the filter is True
>         start_ind = start_stop_filter[0][0]
>         stop_ind = start_stop_filter[0][-1]
> ```
> 
> Do you think the handling of potentially different time steps adds a noticeable overhead in the many time steps case? If so I would be down to even consider refactoring this method to assume a uniform dt...

Sure, that way is better. It seems like when there are many different time steps, the performance is similar to how it was before this PR. So I don't know if it is necessary to refactor at this point

momchil-flex:
Sounds good, but I meant rather that we practically always have a tmesh on a uniform dt. So I was wondering if in these cases the handling incurs some overhead, but probably the bottleneck is the multiplications themselves?

caseyflex:
> Sounds good, but I meant rather that we practically always have a tmesh on a uniform dt. So I was wondering if in these cases the handling incurs some overhead, but probably the bottleneck is the multiplications themselves?

The overhead due to checking all the dts seemed negligible when I tested

caseyflex:
Kernel_indices is actually the same length as time_amps. This is what
return_inverse does for np.unique. It gives the kernel index for every dt
in dts (not just the unique ones)

tylerflex:
ok I see, feel free to disregard my comments then, that makes sense, thanks.

tylerflex (approved):

PR #1136, opened 2023-08-31
Fix warning capture

dbochkov-flexcompute:
While trying to incorporate warnings capture into GUI, MC team noticed that when we have validation errors, setting `td.log.set_capture(True)` changes returned validation errors to something not very informative. This happens because during initialization of models we use `try-finally` in `Tidy3dBaseModel.__init__()` https://github.com/flexcompute/tidy3d/pull/1136/files#diff-d331aef8c3290adc3733d4ce375c5a4b139ad41bd5634113ca404cc827e3d949L76-R82 to try to finish warning capture no matter what. But our warning capture parsing assumes the models are well-built. So, when there are validation errors the instances are incomplete and parsing produces its own errors that are not related to simulation validation. 

A seemingly straightforward way to fix this is to just remove `try-finally` construction, so that we don't attempt to built warning capture tree if there are any errors. Initially, it wouldn't work as intended, but after I added missing `discriminator=TYPE_TAG_STR` everything is fine. I believe this is because when there is no `discriminator` pydantic tries to initialize with every model possible and those that fail, exit before reaching `log.end_capture(self)` in `Tidy3dBaseModel.__init__()`. Previously `finally:` would ensure we still execure that.

Additional improvements:
- added `log.start_capture()`/`log.end_capture(self)` into `Simulation.validate_pre_upload()` to capture those warnings too
- added an optional argument `custom_loc: List = None` to `log.warning()` so that additional information can be provided about warning location. This is useful for `Simulation` validators, which otherwise would just point at the simulation itself.
- added an optional argument `capture: bool = True` to `log.warning()` to avoid capturing not very informative and somewhat arbitrary `frequency passed to 'Medium.eps_model()' is outside of 'Medium.frequency_range' = ...`,
- expanded warning capture test significantly to produce and capture pretty much every possible warning. Also added checks that `td.log.set_capture(True)` doesn't screw anything if there are validation errors.

Note that all these changes doesn't affect python client warning behavior. This only matters for output of `td.log.captured_warnings()` when `td.log.set_capture(True)`

### DIFF
--- a/tests/test_package/test_log.py
+++ b/tests/test_package/test_log.py
@@ -192,7 +192,8 @@ def test_logging_warning_capture():
         size=[domain_size, 20, 20],
         sources=[gaussian_beam, mode_source, plane_wave, tfsf],
         structures=[box, box_in_pml, box_on_boundary, box_outside],
-        monitors=[monitor_flux, mode_mnt, monitor_time, proj_mnt],
+        # monitors=[monitor_flux, mode_mnt, monitor_time, proj_mnt],
+        monitors=[monitor_flux, mode_mnt, proj_mnt],
         run_time=run_time,
         boundary_spec=bspec,
         grid_spec=gspec,
@@ -203,16 +204,17 @@ def test_logging_warning_capture():
 
     # re-add projection monitors because it has been overwritten in validators (far_field_approx=False -> True)
     monitors = list(sim_dict["monitors"])
-    monitors[3] = proj_mnt.dict()
+    monitors[2] = proj_mnt.dict()
 
     sim_dict["monitors"] = monitors
 
     td.log.set_capture(True)
     sim = td.Simulation.parse_obj(sim_dict)
+    print(sim.monitors_data_size)
     sim.validate_pre_upload()
     warning_list = td.log.captured_warnings()
     print(json.dumps(warning_list, indent=4))
-    assert len(warning_list) == 31
+    assert len(warning_list) == 30
     td.log.set_capture(False)
 
     # check that capture doesn't change validation errors
@@ -225,7 +227,8 @@ def test_logging_warning_capture():
     sim_dict_large_mnt = sim.dict()
     sim_dict_large_mnt.update({"monitors": [monitor_time.updated_copy(size=(10, 10, 10))]})
 
-    for sim_dict in [sim_dict_no_source, sim_dict_large_mnt]:
+    # for sim_dict in [sim_dict_no_source, sim_dict_large_mnt]:
+    for sim_dict in [sim_dict_no_source]:
 
         try:
             sim = td.Simulation.parse_obj(sim_dict)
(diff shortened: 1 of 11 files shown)


dbochkov-flexcompute:
Having issues with having this PR pass github tests. Everything seems fine on linux and macos, but failing on windows https://github.com/flexcompute/tidy3d/actions/runs/6086188923. There two issues:
1) monitor storage size for (atleast) `FieldTimeMonitor` seems to be different on linux vs windows. In my test simulation setup on linux I get `{'flux': 44.0, 'mode': 864.0, 'time': 19296371088.0, 'n2f_monitor': 1440.0}` for `sim.monitor_data_sizes` but in windows test runs it is `{'flux': 44.0, 'mode': 864.0, 'time': 2116501904.0, 'n2f_monitor': 1440.0}` https://github.com/flexcompute/tidy3d/actions/runs/6086188923/job/16511966069#step:5:446
2) the second type of issues is not being able to get hash of jax-related objects. weird that it shows up only in windows tests
```
          # Check the class type and its superclasses for a matching encoder
          for base in obj.__class__.__mro__[:-1]:
              try:
                  encoder = ENCODERS_BY_TYPE[base]
              except KeyError:
                  continue
              return encoder(obj)
          else:  # We have exited the for loop without finding a suitable encoder
  >           raise TypeError(f"Object of type '{obj.__class__.__name__}' is not JSON serializable")
  E           jax._src.traceback_util.UnfilteredStackTrace: TypeError: Object of type 'JVPTracer' is not JSON serializable
```

@momchil-flex @tylerflex let me know if you have any suggestions

tylerflex:
for some reason I couldn't see the error but now I do, so yea I guess
```
tidy3d\log.py:167: in end_capture
[1056](https://github.com/flexcompute/tidy3d/actions/runs/6086188923/job/16511966069?pr=1136#step:5:1057)
      model_fields = model.get_submodels_by_hash()
```
I wonder if we need to instead wrap whatever block this is in a `try` `except TypeError` and then find an alternative way to handle unhashable objects? Or we could try a different way to get the model fields, I think `model.__fields__` should work?

tests/test_package/test_log.py
```diff
@@ -89,49 +93,166 @@ def test_logging_warning_capture():
     monitor_flux = td.FluxMonitor(
         center=(0, 0, 0),
         size=(8, 8, 8),
-        freqs=freqs,
+        freqs=list(freqs),
```
tylerflex, line 96: did this not work with `freqs` passed as a numpy array? I would have thought ArrayLike would handle this?

tylerflex (approved):
Thanks @dbochkov-flexcompute . it looks good to me, I dont 100% follow all of the details so feel free to get another review if you are unsure, but given the tests are passing and the issue seems cleared up, I approve. The improvements are also quite nice , thanks 👍

tests/test_package/test_log.py
dbochkov-flexcompute: no, I was just playing around with adding frequencies outside of source range (something like `freqs=list(freqs) + [1]`), and this remained like that accidentally

Issue #1110, opened 2023-08-29
Performance optimization (testing and low hanging fruit)


tylerflex:
## Summary of findings

From one scan (N boxes in a metalens geometry)

NSide 	Nbox	Total time	Time structure VJP
2		4		3.7		        1.9
5		25		20			12
10		100		82			41
15		225		191			92

## Conclusions:

* VJP calculation takes 60% of time.
* Percentage is constant as number of structures increased.
* Total time scales linearly with number of structures.
* When structure VJP is turned off, most of the time goes to jax internals (no low hanging fruit)

### 15 x 15 metalens
```

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      8/1    0.012    0.002  190.959  190.959 traceback_util.py:162(reraise_with_filtered_traceback)
        1    0.000    0.000  190.959  190.959 api.py:644(grad_f)
        1    0.000    0.000  190.948  190.948 api.py:708(value_and_grad_f)
     24/9    0.009    0.000  190.688   21.188 linear_util.py:174(call_wrapped)
      2/1    0.000    0.000  143.153  143.153 tree_util.py:306(__call__)
        1    0.000    0.000  143.153  143.153 api.py:2092(_vjp_pullback_wrapper)
        1    0.000    0.000  143.153  143.153 ad.py:143(unbound_vjp)
      3/1    0.000    0.000  143.153  143.153 ad.py:181(backward_pass)
        1    0.000    0.000  142.901  142.901 ad.py:753(_custom_lin_transpose)
        1    0.003    0.003  142.896  142.896 custom_derivatives.py:765(<lambda>)
        1    0.000    0.000  132.483  132.483 web.py:676(run_local_bwd)
        1    0.000    0.000  108.597  108.597 simulation.py:483(store_vjp)
        1    0.000    0.000  108.597  108.597 simulation.py:506(store_vjp_sequential)
      225    0.003    0.000  107.691    0.479 simulation.py:462(_store_vjp_structure)
      225    0.012    0.000  107.668    0.479 structure.py:60(store_vjp)
      225    0.310    0.001   92.271    0.410 geometry.py:153(store_vjp)
198625/172568    3.009    0.000   62.390    0.000 base.py:77(__init__)
      3/2    0.000    0.000   60.745   30.373 custom_derivatives.py:584(__call__)
     8775    0.122    0.000   56.471    0.006 dataarray.py:2069(interp)
[...]

PR #1138, opened 2023-08-31
add a local adjoint speed test

tylerflex:
Adds a local test of adjoint processing speed on a metalens problem.
Figured it would be nice to get it into the codebase in case we needed it in the future.


tylerflex:
Fixes #1110 
Note: could not find any low hanging fruit in the way of extra `copy()` calls etc.

PR #1144, opened 2023-09-06
set colocate=False in output field monitors for backwards compatibility

tylerflex:


tidy3d/plugins/adjoint/components/simulation.py
```diff
@@ -136,15 +136,16 @@ def _output_monitors_same_freq(cls, val):
...
-            if mnt.dict().get("colocate", False):
-                raise AdjointError(
-                    "Server-side colocation ('Monitor.colocate') must be set to 'False' in the "
-                    "adjoint plugin. Instead, use 'SimulationData.at_boundaries' after the solver "
-                    "run to automatically colocate the fields to the grid boundaries, "
-                    "or 'MonitorData.colocate' if colocating to custom coordinates."
+            if mnt.colocate:
+                log.warning(
+                    "Output field monitors in the adjoint plugin require "
+                    "explicitly setting 'colocate=False'. Setting 'False' as the default value."
```
momchil-flex, line 144: I think `Setting 'False' as the default value.` might be misleading because we're also overriding the value even if it was explicitly set to 'True'. I also feel like giving a bit more hints as in the old warning message would be nice, e.g. something like

```
Output field monitors in the adjoint plugin require explicitly setting 'colocate=False'. Setting 'colocate=False' in monitor {mnt.name}. Use 'SimulationData.at_boundaries' after the solver run to automatically colocate the fields to the grid boundaries, or 'MonitorData.colocate' if colocating to custom coordinates.
```

Just to let people know they can still do that with the adjiont data.

tylerflex: ok. updated the warning message accordingly.

PR #1150, opened 2023-09-08
Fixing inf bounds handling in PolySlab

### LINKED ISSUE #1147: inf in polyslab causes warning
**Describe the bug**
Warning when td.inf is used to define a polyslab structure

**To Reproduce**
When defining a simulation with a polyslab structure that has td.inf in it, we get a warning:

tidy3d/tidy3d/components/geometry/polyslab.py:888: RuntimeWarning: invalid value encountered in scalar multiply
  max_offset += max(0, -self._tanq * self.length_axis)

**Screenshots**
Including this structure in a simulation results in the above warning when the simulation is defined.

image

weiliangjin2021:
Resolves #1147, and additionally other inf related handling in PolySlab. This issue only arises in a polyslab of vertical sidewall, as a slanted polyslab of infinite thickness will for sure self intersect.

### DIFF
--- a/tests/test_components/test_geometry.py
+++ b/tests/test_components/test_geometry.py
@@ -8,6 +8,7 @@ import matplotlib.pyplot as plt
 import gdstk
 import gdspy
 import trimesh
+import warnings
 
 import tidy3d as td
 from tidy3d.exceptions import SetupError, Tidy3dKeyError, ValidationError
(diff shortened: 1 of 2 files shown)


tylerflex (approved):

alec-flexcompute:
Works, thanks!

alec-flexcompute (approved):

Issue #1146, opened 2023-09-07
to_file and from_file support for hdf5.gz


momchil-flex:
The gzip PR introduced zipping and unzipping of `hdf5` files in a few places, e.g. https://github.com/flexcompute/tidy3d/blob/bbac22ed35acd0bf67ba2c91534649ccaaa161d0/tidy3d/web/simulation_task.py#L351-L369

It probably makes more sense to just natively support `hdf5.gz` in `from_file` and `to_file`, and use these functions wherever needed.

lucas-flexcompute:
Do we want an equivalent to `get_simulation_hdf5` for compressed data? https://github.com/flexcompute/tidy3d/blob/bbac22ed35acd0bf67ba2c91534649ccaaa161d0/tidy3d/web/simulation_task.py#L514-L516

It currently downloads and decompresses the `hdf5.gz` file. Since we're supporting compressed hdf5 directly, do we want a second function `get_simulation_hdf5_gz` or add a `decompress=True` argument to the existing one?

PR #1154, opened 2023-09-11
Add support to .hdf5.gz in base (#1146)

lucas-flexcompute:
Compression and decompression utilities have to be moved from the web module so that they can be used within components.base, but JSON_TAG creates a circular dependency, so read_simulation_from_hdf5 was kept within the simulation_task module.

This is temporarily based on `pre/2.4` for review. I'll rebase it to `pre/2.5` once it is available.


tylerflex (approved):
looks good! much cleaner. Appreciate it, @lucas-flexcompute! 

I think this might introduce some merge conflicts when #1108 (stub) is merged, so we might need your help to fix those, but hopefully it won't be too bad.

PR #1155, opened 2023-09-11
add link to GUI URL after simulation run, make links blue underline

tylerflex:
zongfu requested that the link to our web UI be more obvious


tomflexcompute (approved):
Verified that it works. 
image

One potential issue is that the GUI tidy3d version is almost always lacking behind the python frontend. Therefore, people using the latest version or any develop version will not be able to view the simulation result in the GUI.

PR #1156, opened 2023-09-12
Polarization fraction of modes; and collected information as a dataframe

momchil-flex:
We've had a `filter_pol` option in the mode spec for a while, but until now this was always computing the polarization fraction under the hood. Furthermore, this was not even taking the nonuniform grid into account, nor the fact that for angled modes, the field components need to be rotated to the frame in which the propagation axis is z in order to compute polarization fraction properly (i.e. TE/TM is defined w.r.t. the plane normal to the propagation direction, which is not the same as the mode plane normal in the case of angled modes).

I also added a `ModeSolverData.to_dataframe()` method that conveniently gathers all sorts of mode information:
image

There is one remaining problem. I added a test that compares all properties between nominally identical waveguides, one with an angle and one straight. The properties agree (up to some tolerance, because of the numerical grid), apart from the mode area. @lucas-flexcompute it is not immediately obvious to me how the modal area computation should be modified for angled modes, given that in that case the fields are recorded on a plane which is not normal to the propagation direction. Do you have any thoughts? Actually in some sense I'm not even sure if the polarization fraction computation is correct in that case, but the test at least passes for these quantities.

Meanwhile @tomflexcompute could you check again that the quantities in the dataframe agree well with Lumerical for some setup?

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -8,6 +8,7 @@ import warnings
 import xarray as xr
 import numpy as np
 import pydantic.v1 as pd
+from pandas import DataFrame
 
 from .data_array import FluxTimeDataArray, FluxDataArray
 from .data_array import MixedModeDataArray, ModeIndexDataArray, ModeAmpsDataArray
@@ -20,7 +21,7 @@ from .dataset import Dataset, AbstractFieldDataset, ElectromagneticFieldDataset
 from .dataset import FieldDataset, FieldTimeDataset, ModeSolverDataset, PermittivityDataset
 from ..base import TYPE_TAG_STR, cached_property
 from ..types import Coordinate, Symmetry, ArrayFloat1D, ArrayFloat2D, Size, Numpy, TrackFreq
-from ..types import EpsSpecType
+from ..types import EpsSpecType, Literal
 from ..grid.grid import Grid, Coords
 from ..validators import enforce_monitor_fields_present, required_if_symmetry_present
 from ..monitor import MonitorType, FieldMonitor, FieldTimeMonitor, ModeSolverMonitor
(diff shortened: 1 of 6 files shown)


momchil-flex:
> > @lucas-flexcompute it is not immediately obvious to me how the modal area computation should be modified for angled modes, given that in that case the fields are recorded on a plane which is not normal to the propagation direction. Do you have any thoughts? Actually in some sense I'm not even sure if the polarization fraction computation is correct in that case, but the test at least passes for these quantities.
> 
> This looks pretty useful! I think we can calculate the mode area (and maybe improve the fractions) by back-projecting the field components to the normal plane using the effective index + (φ, θ). I can try it out to see if it works today.

I feel like since we're always integrating something like |E|**2, it may just be some cosine factor. Let me have a look into that.

lucas-flexcompute:
Yes, looks like it is from my tests!

tidy3d/components/data/monitor_data.py
```diff
@@ -491,7 +492,7 @@ def mode_area(self) -> FreqModeDataArray:
         num = (intensity * d_area).sum(dim=d_area.dims) ** 2
         den = (intensity**2 * d_area).sum(dim=d_area.dims)
 
-        return FreqModeDataArray(num / den)
+        return FreqModeDataArray(num / den * np.cos(self.monitor.mode_spec.angle_theta))
```
lucas-flexcompute, line 495: Should we check if `isinstance(self.monitor, AbstractModeMonitor)` before accessing `.mode_spec`?

lucas-flexcompute (approved):

tidy3d/components/data/monitor_data.py
momchil-flex: Oh I didn't realize this is a property of `ElectromagneticFieldData` and not `ModeSolverData`. I guess in some cases it could be convenient to compute this for a `FieldData` that comes out of the fdtd solver, although it will be kinda up to the user to figure out if it's correct (there can be various instances in which it's not, e.g. if the data does not correspond to single-mode propagation). So in the interest of allowing for that generality, we could leave it there. On the other hand we could move it to `ModeSolverData` to be more strict and make sure that it's always a meaningful quantity? And either way should the polarization fraction computations be treated the same way?

PR #1157, opened 2023-09-13
Compute loss upper bound in PoleResidue

weiliangjin2021:
- Computes max Im[eps] of a pole residue model in its frequency_range. If frequency_range is None, over all frequencies. The extrema finder function is moved from fit_fast. @caseyflex could you please verify if anything in this PR breaks fitter?

Note that the old extrema finder has an issue: previously my focus was only to find if Im[eps] can be negative, so I wasn't paying attention if it's the extrema of Im[eps] or any quantity proportional to Im[eps]. As I checked today, the old extrema is for Im[eps]/freq, so it'll be off if you care more than just the sign of Im[eps]. It should be fixed now in this PR. 
- Fix Sellemier model `eps_model` error when wavelength ** 2 < C_i
- @e-g-melo For Palik lossless medium, please checkout the added tests. The first two materials have large loss: around 1e-5.

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -12,14 +12,15 @@ import xarray as xr
 
 from .base import Tidy3dBaseModel, cached_property
 from .grid.grid import Coords, Grid
-from .types import PoleAndResidue, Ax, FreqBound, TYPE_TAG_STR, InterpMethod, Bound, ArrayComplex3D
+from .types import PoleAndResidue, Ax, FreqBound, TYPE_TAG_STR
+from .types import InterpMethod, Bound, ArrayComplex3D, ArrayFloat1D
 from .types import Axis, TensorReal
 from .data.dataset import PermittivityDataset
 from .data.data_array import SpatialDataArray, ScalarFieldDataArray, DATA_ARRAY_MAP
 from .viz import add_ax_if_none
 from .geometry.base import Geometry
 from .validators import validate_name_str, validate_parameter_perturbation
-from ..constants import C_0, pec_val, EPSILON_0, LARGE_NUMBER, fp_eps
+from ..constants import C_0, pec_val, EPSILON_0, LARGE_NUMBER, fp_eps, HBAR
 from ..constants import HERTZ, CONDUCTIVITY, PERMITTIVITY, RADPERSEC, MICROMETER, SECOND
 from ..exceptions import ValidationError, SetupError
 from ..log import log
@@ -35,6 +36,12 @@ FILL_VALUE = "extrapolate"
 
 # cap on number of nonlinear iterations
 NONLINEAR_MAX_NUMITERS = 100
+
+# Range for checking upper bound of Im[eps], in addition to extrema method.
+# The range is in unit of eV and it's in log scale.
+LOSS_CHECK_MIN = -10
+LOSS_CHECK_MAX = 4
+LOSS_CHECK_NUM = 1000
 
 
 def ensure_freq_in_range(eps_model: Callable[[float], complex]) -> Callable[[float], complex]:
@@ -60,7 +67,7 @@ def ensure_freq_in_range(eps_model: Callable[[float], complex]) -> Callable[[flo
         # don't warn for evaluating infinite frequency
         if is_inf_scalar:
             return eps_model(self, frequency)
-        if np.any(frequency < fmin) or np.any(frequency > fmax):
+        if np.any(frequency < fmin * (1 - fp_eps)) or np.any(frequency > fmax * (1 + fp_eps)):
             log.warning(
                 "frequency passed to `Medium.eps_model()`"
                 f"is outside of `Medium.frequency_range` = {self.frequency_range}",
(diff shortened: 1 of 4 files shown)


tylerflex (approved):

e-g-melo (approved):
Hi Weiliang! Thanks for this feature.

Losses around 1e-5 are acceptable for most of the applications we have. We still can consider these Ge and GaAs mediums as low loss materials.

tidy3d/components/medium.py
```diff
@@ -1546,6 +1553,146 @@ def to_medium(self) -> Medium:
...
+            # will be ~ 1 in photonics
+            poles, residues = zip(*self._angular_freq_to_eV(np.array(self.poles)))
+            poles = np.array(poles)
+            residues = np.array(residues)
+            extrema_freq = _extrema_loss_freq_finder(
+                poles.real, poles.imag, residues.real, residues.imag
+            )
+            return PoleResidue._eV_to_angular_freq(extrema_freq)
+        except np.linalg.LinAlgError:
+            log.warning(
```
caseyflex, line 1654: The lower bound is inaccurate too, which may mean passivity is not satisfied

tidy3d/components/medium.py
```diff
@@ -1546,6 +1553,146 @@ def to_medium(self) -> Medium:
...
+                "'LinAlgError' in computing Im[eps] extrema. Im[eps] upper bound not accurate."
+            )
+            return np.array([])
+
+    def _imag_ep_extrema_with_samples(self, all_frequency_range: bool = False) -> ArrayFloat1D:
```
caseyflex, line 1659: I don't fully understand this function. It looks like it only keeps extrema within the frequency range, and also returns the frequency range in addition to those extrema. Previously, I was checking min_loss and max_loss at all extrema, regardless of whether they were in this frequency range. Then, I was also checking over the frequency range in case the extrema finder failed. While I can see only wanting to enforce max_loss over a certain frequency range, I thought we wanted passivity at all frequencies.

tidy3d/components/medium.py
```diff
@@ -1546,6 +1553,146 @@ def to_medium(self) -> Medium:
...
+
+        Parameters
+        ----------
+        f_hz : float
+            Frequency in unit of Hz
+        """
+        return f_hz * 2 * np.pi
+
+    @cached_property
+    def _imag_ep_extrema(self) -> Tuple[ArrayFloat1D]:
```
caseyflex, line 1601: We might want this function to be public. For example, it can be useful to have a certificate (in the form of the extrema frequency) for the min loss and max loss in a model.

caseyflex (changes requested):
Some of the fits seem to improve, like Au-Olmon2012stripped. However, I think this is because the new `_imag_ep_extrema_with_samples` is ignoring extrema outside of the frequency range, and this is allowing passivity violations. For example, in Au-Olmon2012stripped, when I use the fitter in pre/2.4 with Lucas's testing notebook, I get RMS error of 758, extrema frequencies of [4.68315419e+10] Hz and extrema eps values of [3022553.48615555+3043567.98870142j]. When I run on your new branch, I get RMS 17.6, which seems better, but extrema frequencies of [4.18809303e+18 1.04422636e+11] Hz and extrema eps values of [9.99999705e-01-2.47246668e-11j 1.13168850e+07+1.54442701e+07j], and the first one is a small passivity violation. This behavior is reversed when I comment out the frequency filtering in your `_imag_ep_extrema_with_samples` function.

tidy3d/components/medium.py
```diff
@@ -1546,6 +1553,146 @@ def to_medium(self) -> Medium:
             frequency_range=self.frequency_range,
         )
 
+    @staticmethod
+    def _eV_to_angular_freq(f_eV: float):
```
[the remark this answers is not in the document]
weiliangjin2021: all public now

tidy3d/components/medium.py
weiliangjin2021: What do you mean by lower bound?

tidy3d/components/medium.py
weiliangjin2021: Now fit_fast calls the extrema finder function directly. This function is only used to find max(min)_loss in the frequency_range.

tidy3d/components/medium.py
weiliangjin2021: Could you elaborate more?

weiliangjin2021:
> Some of the fits seem to improve, like Au-Olmon2012stripped. However, I think this is because the new `_imag_ep_extrema_with_samples` is ignoring extrema outside of the frequency range, and this is allowing passivity violations. For example, in Au-Olmon2012stripped, when I use the fitter in pre/2.4 with Lucas's testing notebook, I get RMS error of 758, extrema frequencies of [4.68315419e+10] Hz and extrema eps values of [3022553.48615555+3043567.98870142j]. When I run on your new branch, I get RMS 17.6, which seems better, but extrema frequencies of [4.18809303e+18 1.04422636e+11] Hz and extrema eps values of [9.99999705e-01-2.47246668e-11j 1.13168850e+07+1.54442701e+07j], and the first one is a small passivity violation. The original behavior is restored when I comment out the frequency filtering in your `_imag_ep_extrema_with_samples` function.

Thanks Casey for finding out this issue! I think I messed up the two functions when writing this part. In the revised version, the staticmethod `imag_ep_extrema` is used in fit_fast, while `_imag_ep_extrema_with_samples` is only used in finding upper/lower bound of Im[eps] in the frequency range of the medium.

tidy3d/components/medium.py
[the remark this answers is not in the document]
caseyflex: thanks!

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -291,61 +292,23 @@ def values(self) -> ArrayComplex1D:
...
-                self.residues.imag,
-            )
-        except np.linalg.LinAlgError:
-            log.warning("'LinAlgError' in passivity checking, passivity not guaranteed.")
-            return []
-
     @cached_property
     def loss_in_bounds_violations(self) -> ArrayFloat1D:
         """Return list of frequencies where model violates loss bounds."""
+        mat = PoleResidue(
```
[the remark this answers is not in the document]
caseyflex: thanks, I think this is safer!

tidy3d/components/medium.py
caseyflex: I mean that, this function is used to find all extrema (maxima and minima). So I use it to enforce min_loss and max_loss (passivity being a special case of min_loss). So, perhaps the warning message should mention that min_loss (including passivity) may not be guaranteed either if the root-finding fails.

tidy3d/components/medium.py
caseyflex: thanks!

tidy3d/components/medium.py
caseyflex: I see the static method is public now. I mean that, if you are really working hard to get a good material fit, it can be useful to relax the passivity constraint so some small negative min_loss, and see where the resulting material fit violates passivity (not just the min_loss, but also the frequencies where it can occur.) Then, you might want to change something (such as shrinking the frequency range) in order to avoid this issue.

weiliangjin2021: i see, that'll be something in the future but not in this PR?

tidy3d/components/medium.py
weiliangjin2021: Please check the revised warning msg.

caseyflex (approved):
Thanks Weiliang for this PR, it looks good to me (see my responses to your comments for one possible change in the warning message, to clarify that passivity may be violated in the case of a linear algebra error).

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -721,7 +675,7 @@ def fit(
             Best fitting result: (dispersive medium, weighted RMS error).
         """
 
-        omega = 2 * np.pi * HBAR * self.freqs[::-1]
+        omega = PoleResidue.angular_freq_to_eV(PoleResidue.Hz_to_angular_freq(self.freqs[::-1]))
```
caseyflex, line 678: One interesting thing -- this line appears to be responsible for the different results now for the following materials:


Ag-RakicLorentzDrude1998
AlAs-FernOnton1971
Al-Rakic1995
GaAs-Skauli2003
InP-Pettit1965
Si3N4-Luke2015PMLStable
Si3N4-Luke2015Sellmeier
Y2O3-Nigara1968


Because, floating-point multiplication is non-commutative. Previously, I multiplied by HBAR first, then 2 pi. The difference in results seems minor though, just a shortcoming of the approach that it is sensitive to such things.

tidy3d/components/medium.py
caseyflex: great, thanks!

tidy3d/plugins/dispersion/fit_fast.py
caseyflex: not quite, actually. I guess the previous version was multiplying 2*pi*HBAR first, then multiplying by the frequency. The new version is probably more sensible, since HBAR*frequency is of order 1

weiliangjin2021: yeah, I think the extrema finding will only make a difference when you bound the loss range by a finite nonzero value.

weiliangjin2021: so do you think i should leave this part as it is, or revert to the previous `2 * np.pi * HBAR * `?

caseyflex: so, I think that the new PR improves some of these fits by fixing this issue

caseyflex: I think you should leave it as is

PR #1158, opened 2023-09-14
Feature/ivy

magiWei:


magiWei:
@tylerflex Can you have a check and merge it to tidy3d-2.4.1, thanks.

tylerflex:
It looks ok to me, but could you explain what this change does? 
Also, could you add a line in the CHANGELOG corresponding to the change? 
thanks

tylerflex:
actually maybe this doesn't require a changelog item, it would just be good to understand what the intended change of the PR is so we can figure out whether to tell users about it through the changelog, or if it is just an internal change.

magiWei:
> actually maybe this doesn't require a changelog item, it would just be good to understand what the intended change of the PR is so we can figure out whether to tell users about it through the changelog, or if it is just an internal change.



> actually maybe this doesn't require a changelog item, it would just be good to understand what the intended change of the PR is so we can figure out whether to tell users about it through the changelog, or if it is just an internal change.

Add USER_AGENT to http header for statistics and Return empty list if there is no folder.

tylerflex (approved):
Ok, looks good to me, I modified the changelog a bit and rebased against develop. Should be ready to merge. Not sure if @momchil-flex needs to take a look but if it's not urgent we can merge in a few days after he is back. Thanks

PR #1160, opened 2023-09-15
fix bug with PolySlab in JaxSimulation.structures

tylerflex:
This should work with your particle swarm notebook now. I just tried it. 

Basically what this does is converts the `polyslab.vertices` to a list in the intermediate representation that jax handles. this was because jax was not able to handle an untracked multi-dimensional array

### DIFF
--- a/tidy3d/plugins/adjoint/components/base.py
+++ b/tidy3d/plugins/adjoint/components/base.py
@@ -37,6 +37,26 @@ class JaxObject(Tidy3dBaseModel):
             sub_children, sub_aux_data = jax_tree_flatten(field)
             children.append(sub_children)
             aux_data[field_name] = sub_aux_data
+
+        def fix_polyslab(geo_dict: dict) -> None:
+            """Recursively Fix a dictionary possibly containing a polyslab geometry."""
+            if geo_dict["type"] == "PolySlab":
+                vertices = geo_dict["vertices"]
+                geo_dict["vertices"] = vertices.tolist()
+            elif geo_dict["type"] == "GeometryGroup":
+                for sub_geo_dict in geo_dict["geometries"]:
+                    fix_polyslab(sub_geo_dict)
+            elif geo_dict["type"] == "ClipOperation":
+                fix_polyslab(geo_dict["geometry_a"])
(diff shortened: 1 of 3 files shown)


tidy3d/plugins/adjoint/components/base.py
```diff
@@ -37,6 +37,16 @@ def tree_flatten(self) -> Tuple[list, dict]:
...
             aux_data[field_name] = sub_aux_data
+
+        # fixes bug with jax handling 2D numpy array in polyslab vertices
+        if aux_data.get("type", "") == "JaxSimulation":
+            structures = aux_data["structures"]
+            for i, structure in enumerate(structures):
+                geometry = structure["geometry"]
+                if geometry["type"] == "PolySlab":
+                    vertices = geometry["vertices"]
+                    aux_data["structures"][i]["geometry"]["vertices"] = vertices.tolist()
```
momchil-flex, line 48: This probably still leaves the problem that `geometry` is a `GeometryGroup` that contains `PolySlab`-s. More generally there is also the new boolean operations that have not been tested with adjoint either. To handle this, Lucas has introduced this convenience method though: https://github.com/flexcompute/tidy3d/blob/2e07342b57a2a55efd2c54f37601cc317d8c211e/tidy3d/components/geometry/utils.py#L46-L67. I think something similar is what needs to be used here (but not sure if it can be directly used...)

tylerflex: good point. I think I'll have to pick this up on Monday unfortunately.

tylerflex: I just gave it a shot and it worked. just made the function recursive.

momchil-flex: So technically this solves geometry group but not clip operation...

tylerflex: Oh right, so I just added a special handling for clip operation. I think that covers everything?

tylerflex:
Note: tested this with TriangleMesh and it worked off the bat, so I just added it in a test but kept the actual fix the same.

PR #1164, opened 2023-09-19
fix adjoint JaxCustomMedium with truly anisoptric custom medium

tylerflex:


tylerflex:
@weiliangjin2021 I'm unsure as to whether these changes break anything when multiple frequencies are involved.

Note: there is a cast of `freq` to `float` in `eps_complex_to_eps_sigma` and then there's a selection of `freq[0]` in the next function, which follows some of the other convention for isotropic. Just want to make sure this still retains the behavior of regular custom medium.

weiliangjin2021 (approved):

PR #1165, opened 2023-09-19
added lsf converter

alec-flexcompute:
Added parser, example lsf file for testing/testing capability, and command line capability. Let me know if you want this implemented differently.

As for the converter.py file itself, let me know if you want any more comments. Most of the functions are pretty much the same, so I wasn't sure what comments/other style stuff is necessary here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ModeSolverData.pol_fraction` and `ModeSolverData.pol_fraction_waveguide` properties to compute polarization fraction of modes using two different definitions.
 - `ModeSolverData.to_dataframe()` and `ModeSolverData.modes_info` for a convenient summary of various modal properties of the computed modes.
 - Loss upper bound estimation in `PoleResidue` material model.
+- Command line tool `tidy3d convert` for conversion from `.lsf` project files into equivalent python scripts implementing the project in Tidy3D. Usage:  `tidy3d convert example.lsf tidy3d_script.py`.
 
 ### Changed
 - Output task URL before and after simulation run and make URLs blue underline formatting.
--- a/tidy3d/web/cli/app.py
+++ b/tidy3d/web/cli/app.py
@@ -12,6 +12,7 @@ import toml
 from tidy3d.web.cli.constants import TIDY3D_DIR, CONFIG_FILE, CREDENTIAL_FILE
 from tidy3d.web.cli.migrate import migrate
 from tidy3d.web.environment import Env
+from tidy3d.web.cli.converter import converter_arg
 
 if not os.path.exists(TIDY3D_DIR):
     os.mkdir(TIDY3D_DIR)
@@ -114,5 +115,14 @@ def migration():
     migrate()
 
 
+@click.command()
+@click.argument("lsf_file", type=click.Path(exists=True))
+@click.argument("new_file")
+def convert(lsf_file, new_file):
+    """Click command to convert .lsf project into Tidy3D .py file"""
+    converter_arg(lsf_file, new_file)
+
+
 tidy3d_cli.add_command(configure)
 tidy3d_cli.add_command(migration)
+tidy3d_cli.add_command(convert)
--- a/tidy3d/web/cli/converter.py
+++ b/tidy3d/web/cli/converter.py
@@ -1,0 +1,1330 @@
+from pathlib import Path
+
+
+def _remove_spaces(line: str) -> str:
+    """
+    Removes spaces from line, except spaces found between quotes. Also converts " into ' for ease of further parsing
+    """
+    quotes = False
+    comment = False
+    new_line = ""
+    for c in line:
+        if c == " " and not quotes:
+            if comment:
+                new_line += c
+        else:
+            if c == "'" or c == '"':
+                quotes = not quotes
+                c = "'"
+            elif c == "#":
+                comment = True
+            new_line += c
+    return new_line
+
+
+def _is_int(number_str: str) -> bool:
+    try:
+        int(number_str)
+        return True
+    except ValueError:
+        return False
+
+
+def _is_float(number_str: str) -> bool:
+    try:
+        float(number_str)
+        return True
+    except ValueError:
+        return False
+
+
+def _clean(sci_str: str) -> str:
+    number, power = sci_str.split("e")
+    while number[-1] == "0" or number[-1] == ".":
+        if len(number) == 1:
+            break
+        number = number[:-1]
+    while power[1] == "0" and len(power) > 2:
+        power = power[0] + power[2:]
+    if power[0] == "+":
+        power = power[1:]
+    if power == "0":
+        return number
+    if "." not in number and power == "1":
+        return number + "0"
+    return number + "e" + power
+
+
+def _to_um(array: list) -> list:
+    output_array = []
+    for number_str in array:
+        if number_str == "None":
+            output_array.append("None")
+        elif number_str == "td.inf" or number_str == "-td.inf":
+            output_array.append(number_str)
+        elif "e" in number_str:  # potentially scientific notation
+            number_power = number_str.split("e")
+            if len(number_power) != 2:  # number_str is a variable name with an 'e'
+                output_array.append(number_str + "*1e6")
+            else:
+                number, power = number_power
+                if _is_int(number):
+                    if _is_int(power):
+                        value = int(number) * 10 ** (int(power)) * 1e6
+                        value_str = _clean(f"{value:e}")
+                        output_array.append(value_str)
+                    elif _is_float(power):
+                        value = int(number) * 10 ** (float(power)) * 1e6
+                        value_str = _clean(f"{value:e}")
+                        output_array.append(value_str)
+                    else:  # number_str something else
+                        output_array.append(number_str + "*1e6")
+                elif _is_float(number):
+                    if _is_int(power):
+                        value = float(number) * 10 ** (int(power)) * 1e6
+                        value_str = _clean(f"{value:e}")
+                        output_array.append(value_str)
+                    elif _is_float(power):
+                        value = float(number) * 10 ** (float(power)) * 1e6
+                        value_str = _clean(f"{value:e}")
+                        output_array.append(value_str)
+                    else:  # number_str is something else
+                        output_array.append(number_str + "*1e6")
+                else:
+                    output_array.append(number_str + "*1e6")
+        elif _is_int(number_str):
+            value_str = _clean(f"{int(number_str):e}")
+            output_array.append(value_str)
+        elif _is_float(number_str):
+            value_str = _clean(f"{float(number_str):e}")
+            output_array.append(value_str)
+        else:  # dealing with a string representing a variable
+            output_array.append(number_str + "*1e6")
+    return output_array
+
+
+def _is_declaration(Line: str) -> bool:
+    comment = False
+    for i in range(0, len(Line)):
+        if Line[i] == "#":
+            comment = True
+        if Line[i] == "=" and not comment:
+            if "[" in Line[i] and "]" in Line[i]:  # change array syntax
+                l_bracket = Line[i].find("[")
+                new_array_string = "["
+                array = Line[l_bracket + 1 :].split(";")
+                for a in array:
+                    new_array_string += "(" + a + "),"
+                r_bracket = new_array_string.find("]")
+                new_array_string = (
+                    new_array_string[:r_bracket] + ")" + new_array_string[r_bracket:-1]
+                )
+                Line[i] = Line[i][:l_bracket] + new_array_string
+            return True
+    return False
+
+
+def _add_to_variable_dict(line: str, variable_dict: dict):
+    """
+    Takes a string which describes a variable assignment and adds the variable assignment to the variable dictionary. Supported for strings, floats, booleans, and other variables.
+
+    E.g. "string_var='str'" adds 'string_var': 'str' to the dictionary
+    """
+    variable_name = ""
+    var = False
+    for i in range(0, len(line)):
+        if line[i] != "=" and not var:
+            variable_name += line[i]
+        else:
+            if line[i] == "=":
+                var = True
+            else:
+                if line[i] == "'":
+                    # if variable represents a string
+                    variable_dict[variable_name] = line[i + 1 :].split("'")[0]
+                elif line[i : i + 4] == "True":
+                    variable_dict[variable_name] = True
+                elif line[i : i + 5] == "False":
+                    variable_dict[variable_name] = False
+                elif line[i] == "[":
+                    return  # is an array
+                else:
+                    try:  # if variable represents a float
+                        float(line[i:].split(";")[0])
+                        variable_dict[variable_name] = float(line[i:].split(";")[0])
+                    except ValueError:
+                        # if variable represents another variable
+                        variable_dict[variable_name] = variable_dict[line[i:].split(";")[0]]
+                return
+
+
+def _addrect(Lines, i: int, rect: int) -> str:
+    # https://optics.ansys.com/hc/en-us/articles/360034901473-Rectangle-Simulation-Object
+    x, y, z = "0", "0", "0"
+    xl, yl, zl = "0", "0", "0"
+    x_min, x_max = "0", "0"
+    y_min, y_max = "0", "0"
+    z_min, z_max = "0", "0"
+    name = "rect_" + str(rect)
+    medium = ""
+    while Lines[i][0] == "\n" or Lines[i][:5] == "set('" or Lines[i][0] == "#":
+        if Lines[i][5:9] == "name":
+            name = Lines[i][12:].split("'")[0]
+        elif Lines[i][5:7] == "x'":
+            x = Lines[i][8:].split(";")[0][:-1]
+        elif Lines[i][5:7] == "y'":
+            y = Lines[i][8:].split(";")[0][:-1]
+        elif Lines[i][5:7] == "z'":
+            z = Lines[i][8:].split(";")[0][:-1]
+        elif Lines[i][5:10] == "x_min":
+            x_min = Lines[i][12:].split(";")[0][:-1]
+            xl = x_max + "-" + x_min
+        elif Lines[i][5:10] == "x_max":
+            x_max = Lines[i][12:].split(";")[0][:-1]
+            xl = x_max + "-" + x_min
+        elif Lines[i][5:10] == "y_min":
+            y_min = Lines[i][12:].split(";")[0][:-1]
+            yl = y_max + "-" + y_min
+        elif Lines[i][5:10] == "y_max":
+            y_max = Lines[i][12:].split(";")[0][:-1]
+            yl = y_max + "-" + y_min
+        elif Lines[i][5:10] == "z_min":
+            z_min = Lines[i][12:].split(";")[0][:-1]
+            zl = z_max + "-" + z_min
+        elif Lines[i][5:10] == "z_max":
+            z_max = Lines[i][12:].split(";")[0][:-1]
+            zl = z_max + "-" + z_min
+        elif Lines[i][5:11] == "x span":
+            xl = Lines[i][13:].split(";")[0][:-1]
+        elif Lines[i][5:11] == "y span":
+            yl = Lines[i][13:].split(";")[0][:-1]
+        elif Lines[i][5:11] == "z span":
+            zl = Lines[i][13:].split(";")[0][:-1]
+        elif Lines[i][5:10] == "index":
+            index = Lines[i][12:].split(";")[0][:-1]
+            medium = "td.Medium(permittivity=" + index + "**2)"
+        elif Lines[i][5:13] == "material":
+            medium = Lines[i][15:].split(";")[0][:-1] + " # NOTE: check material library"
+        i += 1
+        if i == len(Lines):
+            break
+    rect_string = name.replace(" ", "_") + " = td.Structure(\n"
+    rect_string += "\tgeometry=td.Box(\n"
+    x, y, z, xl, yl, zl = _to_um([x, y, z, xl, yl, zl])
+    rect_string += "\t\tcenter=(" + x + "," + y + "," + z + "),\n"
+    rect_string += "\t\tsize=(" + xl + "," + yl + "," + zl + "),\n"
+    rect_string += "\t),\n"
+    rect_string += "\tmedium=" + medium + ",\n"
+    rect_string += "\tname='" + name + "',\n"
+    rect_string += ")\n"
+    return rect_string, name.replace(" ", "_")
+
+
+def _addsphere(Lines, i: int, sph: int):
+    # https://optics.ansys.com/hc/en-us/articles/360034901553-Sphere-Simulation-Object
+    x, y, z = "0", "0", "0"
+    radius = " # NOTE: Please specify"
+    name = "sphere_" + str(sph)
+    medium = "td.Medium()"
+    while Lines[i][0] == "\n" or Lines[i][:5] == "set('" or Lines[i][0] == "#":
+        if Lines[i][5:9] == "name":
+            name = Lines[i][12:].split("'")[0]
+        elif Lines[i][5:7] == "x'":
+            x = Lines[i][8:].split(";")[0][:-1]
+        elif Lines[i][5:7] == "y'":
+            y = Lines[i][8:].split(";")[0][:-1]
+        elif Lines[i][5:7] == "z'":
+            z = Lines[i][8:].split(";")[0][:-1]
+        elif Lines[i][5:11] == "radius":
+            radius = Lines[i][13:].split(";")[0][:-1]
+        elif Lines[i][5:10] == "index":
+            index = Lines[i][12:].split(";")[0][:-1]
+            medium = "td.Medium(permittivity=" + index + "**2)"
+        elif Lines[i][5:13] == "material":
+            medium = Lines[i][15:].split(";")[0][:-1] + " # NOTE: check material library"
+        elif Lines[i][5:19] == "make ellipsoid":
+            pass  # for now
+        i += 1
+        if i == len(Lines):
+            break
+    sphere_string = name.replace(" ", "_") + " = td.Structure(\n"
+    sphere_string += "\tgeometry=td.Sphere(\n"
+    x, y, z, radius = _to_um([x, y, z, radius])
+    sphere_string += "\t\tcenter=(" + x + "," + y + "," + z + "),\n"
+    sphere_string += "\t\tradius=" + radius + ",\n"
+    sphere_string += "\t),\n"
+    sphere_string += "\tmedium=" + medium + ",\n"
+    sphere_string += "\tname='" + name + "',\n"
+    sphere_string += ")\n"
+    return sphere_string, name.replace(" ", "_")
+
+
+def _addcircle(Lines, i: int, cyl: int):
+    # https://optics.ansys.com/hc/en-us/articles/360034901513-Circle-Simulation-Object
+    x, y, z = "0", "0", "0"
+    z_min, z_max = "-td.inf/2", "td.inf/2"
+    z_span = z_max + "-" + z_min
+    name = "cylinder_" + str(cyl)
+    radius = " # NOTE: please specify"
+    medium = "td.Medium()"
+    while Lines[i][0] == "\n" or Lines[i][:5] == "set('" or Lines[i][0] == "#":
+        if Lines[i][5:9] == "name":
+            name = Lines[i][12:].split("'")[0]
+        elif Lines[i][5:7] == "x'":
+            x = Lines[i][8:].split(";")[0][:-1]
+        elif Lines[i][5:7] == "y'":
+            y = Lines[i][8:].split(";")[0][:-1]
+        elif Lines[i][5:7] == "z'":
+            z = Lines[i][8:].split(";")[0][:-1]
+        elif Lines[i][5:10] == "z min":
+            z_min = Lines[i][12:].split(";")[0][:-1]
+            z_span = z_max + "-" + z_min
+        elif Lines[i][5:10] == "z max":
+            z_max = Lines[i][12:].split(";")[0][:-1]
+            z_span = z_max + "-" + z_min
+        elif Lines[i][5:11] == "z span":
+            z_span = Lines[i][13:].split(";")[0][:-1]
+        elif Lines[i][5:11] == "radius":
+            radius = Lines[i][13:].split(";")[0][:-1]
+        elif Lines[i][5:10] == "index":
+            index = Lines[i][12:].split(";")[0][:-1]
+            medium = "td.Medium(permittivity=" + index + "**2)"
+        elif Lines[i][5:13] == "material":
+            medium = Lines[i][15:].split(";")[0][:-1] + " # NOTE: check material library"
+        i += 1
+        if i == len(Lines):
+            break
+    cylinder_string = name.replace(" ", "_") + " = td.Structure(\n"
+    cylinder_string += "\tgeometry=td.Cylinder(\n"
+    x, y, z, radius, z_span = _to_um([x, y, z, radius, z_span])
+    cylinder_string += "\t\tcenter=(" + x + "," + y + "," + z + "),\n"
+    cylinder_string += "\t\tradius=" + radius + ",\n"
+    cylinder_string += "\t\tlength=" + z_span + ",\n"
+    cylinder_string += "\t),\n"
+    cylinder_string += "\tmedium=" + medium + ",\n"
+    cylinder_string += "\tname='" + name + "'\n"
+    cylinder_string += ")\n"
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `ModeSolverData.pol_fraction` and `ModeSolverData.pol_fraction_waveguide` properties to compute polarization fraction of modes using two different definitions.
 - `ModeSolverData.to_dataframe()` and `ModeSolverData.modes_info` for a convenient summary of various modal properties of the computed modes.
+- `tidy3d.converter.py` to convert .lsf project files into Tidy3D files, as well as `test.test_package.example.lsf` to test it.
```
tylerflex, line 11: since users will be reading this, would be best to describe how it could be used by the user. for example
```
- conversion from .lsf project files into Tidy3D files through command line: `tidy3d convert example.lsf tidy3d_script.py`
```

tests/test_package/example.lsf
tylerflex, line 1: I'd say let's throw this in `tests/data/` to keep it a bit separated.

tests/test_package/test_main.py
```diff
@@ -15,3 +15,7 @@ def test_main(extension, tmp_path):
     path = str((tmp_path / "sim").with_suffix(extension))
     save_sim_to_path(path)
     main([path, "--test_only"])
+
+
+def test_parser():
```
tylerflex, line 20: Would be good to have some documentation for how to invoke this, so as it stands now, one would need to do
```
tidy3d parsed_lsf_to_tidy3d_test.py -l example.lsf
```
?

1. I'm worried that conflicts with the previous `tidy3d simulation.json data.hdf5` command somewhat. how does it know that the `parsed_lsf_to_tidy3d_test.py` is the output file name and not the input file name for the old main function?
2. I think it might be cleaner to introduce a new command like `tidy3d convert in_file out_file` to separate functionality (same comment as 1 but a suggestion for how to resolve it)
3. It seems a bit strange from my perspective to provide the output file first and the input file second.
4. Would be good to document somewhere (eg in the docs) how to use this, maybe in our [howdoi](https://docs.flexcompute.com/projects/tidy3d/en/latest/howdoi.html) section of the docs?

tidy3d/converter.py
tylerflex, line 1: this is quite an accomplishment. what worries me is that it might be hard to modify later if something changes, but I don't see immediately how to help with that. I guess a couple questions:
1. Do we know whether lsf file syntax changes often over time?
2. is there any way to refactor some repeated code? I notice, for example, a lot of things like
```py
        elif Lines[i][5:10] == "z_min":
             z_min = Lines[i][12:].split(";")[0][:-1]
             zl = z_max + "-" + z_min
```
repeated.  generally speaking this could be contained a single function? from your perspective, how much of this could be refactored to simplify things?

tylerflex (changes requested):
Thanks @alec-flexcompute for the effort here. This will be super useful to have. Let's think about the API / command a bit more

tests/test_package/test_main.py
alec-flexcompute: Sounds good. I think I initially designed it this way because it looked like a "sim" argument was required in __main__.py, so I figured I would use that, but there is probably a way to implement it in the way you suggested in 2.

tidy3d/converter.py
alec-flexcompute: 1. I'm not sure, but given the simplicity of the syntax, and since it hasn't changed since I last worked with it in 2017, I don't think it changes that often.
2. I wanted to do this because I agree it's less simple to do it the way it currently is. The problem is that it's designed so that the "set(...)" commands need not be ordered (since I'm not sure how the commands are ordered when the lsf file is made, as there are a bunch of optional commands as well, and almost no documentation on the internet), so helper functions would need a very large number of possible variables, and I'm not sure it'd be worth it style-wise.

tests/test_package/test_main.py
tylerflex: I think so, we have a `tidy3d configure` command defined in `tidy3d/web/cli` that has its own separate functionality (setting the user API key) so we should be able to just follow that example with `tidy3d convert` and then there should be no overlap with just the `tidy3d ...` command I think (which runs simulation.json files directly, not scripts)

tylerflex (changes requested):
Thanks @alec-flexcompute There are a few issues with this still. 

First, I think the API isn't quite ideal. I just tried to run it in a few common scenarios to see what happens.

image

1. Seems to only work with `python -m tidy3d convert`. It should just work like `tidy3d convert`, which is how both the regular `tidy3d sim.json`, `tidy3d configure` and `tidy3d migrate` work as command line interfaces. Upon digging into it, we don't want `convert` to be an argument but rather a command within the `tidy3d` CLI.  See the `entry_points` in `setup.py` for an example of how these are specified and `tidy3d/web/cli` for how to set them up.
2. One side effect of doing it with `python -m` is that if the user for example forgets the output file name, the error doesn’t provide any information about how to solve the issue (missing output file).  `tidy3d —help` doesn’t display any information about this feature (what arguments exist and their descriptions). Using `argparse` you can make arguments required and make their type `source_file`, so it automatically takes care of what is done in `converter_arg`.

Next, I ran it the current way on the provided example file and tried to see what the experience would be like converting this to a working tidy3d file. There were a lot of issues that I needed to clear up that I think we could handle better. Here are some of the ones I found.

3. Units are in SI in lumerical, but we use (um, sec, Hz) in tidy3d. So fields with units of length need to be multiplied by 1e6 or they will be far too small.
4. The fields that should be strings (with quotes) were not, for example `name` fields and `polarization`.
5.  Many of the components are not imported and don’t use the `td.` namespace.
6. `PolySlab.slab_bounds()` has a backwards parentheses (like `slab_bounds(...(`)
the source `fwidth` was set with the wavelength scan but was set to `0` when the `wavelength_start == wavelength_stop` . This is not an acceptable value as we always need some `fwidth > 0`. The `fwidth` should be more reasonably set in this case, potentially at least `freq0 / 10.0` or so.
7. Mode sources in tidy3d need to be planar but an invalid line source was created.
8. Many of the sources were out of bounds of the simulation, which failed validation. The simulation has size=0 in z, as did the sources, except the sources were placed often at z !=0. I’m not sure what the idea was in this example or what the expected behavior is, but something to keep in mind.
9. `td.GridSpec` required addition of `wavelength` a few times, which is maybe expected, but it could be nice if we could handle that automatically.

I think we should clear up these details so the user doesn't get too frustrated when trying to convert their scripts.  After that, I think this feature will help people a lot, so thanks again for your efforts on it.

I've pasted the `git diff` between the original tidy3d script and the one that ran after fixing all of the issues if you want to see more specifically where they came up.

```
@@ -15,12 +15,14 @@ mesh_y=20e-9
 finer_mesh_size=2.5e-6
 mesh_accuracy=2

+freqs = [1e14]
+
 new_sphere = td.Structure(
        geometry=td.Sphere(
                center=[1e-6,2e-6,0],
                radius=5e-6,
        ),
-       medium= ## PLEASE SPECIFY,
+       medium=td.Medium(),
        name='new_sphere',
 )

@@ -47,10 +49,10 @@ plane_source_time_0 = td.GaussianPulse(
        offset=5.0
 )
 plane_source_0 = td.PlaneWave(
-       center=(0,0,3e-6),
+       center=(0,0,0e-6),
        size=(2e-6,5e-6,0),
        source_time=plane_source_time_0,
-       name=plane_source_0,
+       name="plane_source_0",
        direction='+',
        angle_theta=0,
        angle_phi=0,
@@ -79,13 +81,13 @@ output_wg_bottom = td.Structure(
 ## SOURCE
 mode_source_time_0 = td.GaussianPulse(
        freq0=td.C_0/((0.55+0.55)/2),
-       fwidth=(td.C_0/0.55)-(td.C_0/0.55),
+       fwidth=(td.C_0/0.5)-(td.C_0/0.55),
        offset=5.0
 )
 mode_spec_0 = td.ModeSpec(num_modes=1,angle_theta=0)
 mode_source_0 = td.ModeSource(
        center=[-1.25e-6,0.0,0],
-       size=[0,size_y,0],
+       size=[0,size_y,1],
        source_time=mode_source_time_0,
        name='mode_source_0',
        mode_spec=mode_spec_0,
@@ -99,10 +101,10 @@ gauss_source_time_0 = td.GaussianPulse(
        offset=5.0
 )
 gaussian_beam_0 = td.GaussianBeam(
-       center=(0,0,10e-6),
+       center=(0,0,0),
        size=(2e-6,5e-6,0),
        source_time=gauss_source_time_0,
-       name=gaussian_beam_0,
+       name="gaussian_beam_0",
        direction='+',
        angle_theta=0,
        angle_phi=0,
@@ -112,18 +114,18 @@ gaussian_beam_0 = td.GaussianBeam(
 )

 ## FDTD
-sim = Simulation(
+sim = td.Simulation(
        center=(0.0,0.0,0),
        size=(size_x,size_y,0),
-       run_time= ## PLEASE SPECIFY,
+       run_time=1e-12,
        medium=td.Medium(permittivity=1.44**2),
        symmetry=[0,-1,0],
-       boundary_spec=BoundarySpec(
-               x = Boundary(minus=td.PML(num_layers=12), plus=td.PML(num_layers=12)),
-               y = Boundary(minus=td.PML(num_layers=12), plus=td.PML(num_layers=12)),
-               z = Boundary(minus=td.PML(num_layers=12), plus=td.PML(num_layers=12)),
+       boundary_spec=td.BoundarySpec(
+               x = td.Boundary(minus=td.PML(num_layers=12), plus=td.PML(num_layers=12)),
+               y = td.Boundary(minus=td.PML(num_layers=12), plus=td.PML(num_layers=12)),
+               z = td.Boundary(minus=td.PML(num_layers=12), plus=td.PML(num_layers=12)),
        ),
-       grid_spec=td.GridSpec.auto(min_steps_per_wvl=10),
+       grid_spec=td.GridSpec.auto(min_steps_per_wvl=10, wavelength=1.0),
 )

 ## MESH IN OPTIMIZABLE REGION
@@ -138,19 +140,18 @@ opt_fields = td.FluxMonitor(
        center=(0,0,0),
        size=(finer_mesh_size,finer_mesh_size,0),
        name='opt_fields',
-       freqs= ## PLEASE SPECIFY,
+       freqs=freqs,
 )

 new_circle = td.Structure(
        geometry=td.Cylinder(
                center=(1e-6,2e-6,0),
                radius=5e-6,
+               length=td.inf,
        ),
-       medium= ## PLEASE SPECIFY,
-       name='new_circle'
+       medium=td.Medium(),
+       name='new_circle',
        )
-)
-


 vtx=[(1,0),(2,2),(4,2),(4,1),(3,1)] # comment
@@ -158,11 +159,10 @@ random_polygon = td.Structure(
        geometry=td.PolySlab(
                vertices=vtx,
                axis=2,
-               slab_bounds=(0,2e-6(,
+               slab_bounds=(0,2e-6),
        ),
-       medium= ## PLEASE SPECIFY,
+       medium=td.Medium(),
        name='random_polygon'
-       )
 )

 dipole_source_time_0 = td.GaussianPulse(
@@ -171,10 +171,10 @@ dipole_source_time_0 = td.GaussianPulse(
        offset=5.0
 )
 point_dipole_0 = td.PointDipole(
-       center=(0,-1e-6,5e-6),
+       center=(0,-1e-6,0),
        source_time=dipole_source_time_0,
-       name=point_dipole_0,
-       polarization=Ez,
+       name="point_dipole_0",
+       polarization="Ez",
 )

 ## FOM FIELDS
@@ -182,24 +182,24 @@ fom = td.FluxMonitor(
        center=(finer_mesh_size/2.0,0.0,0),
        size=(0,size_y,0),
        name='fom',
-       freqs= ## PLEASE SPECIFY,
+       freqs=freqs,
 )

 E_field = td.FieldMonitor(
        center=(0,0,0),
        size=(5e-6,5e-6,0),
        name='E_field',
-       freqs= ## PLEASE SPECIFY
+       freqs=freqs,
 )
 filename='electric_field.mat'

 # addtestUnavailableCommand; # does not yet parse to Tidy3d

-mode_spec_0 = ModeSpec(num_modes=1) # check if solving for correct mode!
+mode_spec_0 = td.ModeSpec(num_modes=1) # check if solving for correct mode!
 mode_monitor = td.ModeMonitor(
        center=(0,0,0),
        size=(5e-6,5e-6,0),
-       freqs= ## PLEASE SPECIFY
+       freqs=freqs,
        mode_spec=mode_spec_0,
        name='mode monitor',
 )
@@ -207,5 +207,5 @@ mode_monitor = td.ModeMonitor(
 sim = sim.copy(update=dict(structures=[new_sphere, input_wg, output_wg_top, output_wg_bottom, new_circle, random_polygon], # NOTE: Check order of structures for potential overlap issues
        sources=[plane_source_0, mode_source_0, gaussian_beam_0, point_dipole_0],
        monitors=[opt_fields, fom, E_field, mode_monitor],
-       grid_spec=td.GridSpec.auto(min_steps_per_wvl=10, override_structures=[refine_box_0])
-)
\ No newline at end of file
+       grid_spec=td.GridSpec.auto(min_steps_per_wvl=10, override_structures=[refine_box_0], wavelength=1)
+))
\ No newline at end of file
```

alec-flexcompute:
Thanks Tyler for the feedback. A few things:

1. does a command within the CLI also hold for `python -m`? Since "simulation" is a required argument in __main__.py I've been running into issues with defining the `convert` command.
2. Making `convert` a required argument I thought would interfere with the other commands one could run with `python -m` Another thing is that, during the checks, the file path to the example file would work with I believe python 3.7 checks, but not with 3.11 checks, and what worked with python 3.11 checks didn't work with python 3.7 checks, so I manually put in a way to handle the "file not found" error that would show up for one of them. I'm not sure it would pass this weird behavior with `source_file`
3. I did see certain objects in lsf being defined in units of microns themselves, so I left it alone, but if Lumerical officially uses SI units, I can change it.
8. This was just to test parseability, and the actual coordinates were not considered. I can change them in the example file

tylerflex:
> 1. does a command within the CLI also hold for `python -m`? Since "simulation" is a required argument in **main**.py I've been running into issues with defining the `convert` command.

So `python -m tidy3d ...` runs `tidy3d/__main__.py`. I'm a little fuzzy on the details but I think it would be the same as running `python tidy3d/__main__.py`. Right now, that syntax is dedicated to running `.json` files directly.

We have a dedicated CLI that we use for configuration and migration defined [here](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/web/cli/app.py#L43-L118). This lets users do `tidy3d configure` `tidy3d migrate` etc without conflicting with the code in `__main__.py`. I was trying to say earlier that we should introduce a new `tidy3d convert` option using the same approach:

> I think it might be cleaner to introduce a new command like tidy3d convert in_file out_file to separate functionality (same comment as 1 but a suggestion for how to resolve it)
> We have a tidy3d configure command defined in tidy3d/web/cli that has its own separate functionality (setting the user API key) so we should be able to just follow that example with tidy3d convert and then there should be no overlap with just the tidy3d ... command I think (which runs simulation.json files directly, not scripts)
There's a difference between running `python -m` and having a CLI. 

It's admittedly a bit confusing but I think it should be clear following the example in `tidy3d/web/cli/app.py`

```

@click.command()
def convert():
    convert_fn()

tidy3d_cli.add_command(convert)
```
added to that is enough to define a new `tidy3d convert` CLI and then it can be decorated with the specific arguments for this function.

> 2. Making `convert` a required argument I thought would interfere with the other commands one could run with `python -m` Another thing is that, during the checks, the file path to the example file would work with I believe python 3.7 checks, but not with 3.11 checks, and what worked with python 3.11 checks didn't work with python 3.7 checks, so I manually put in a way to handle the "file not found" error that would show up for one of them. I'm not sure it would pass this weird behavior with `source_file`

I think using the above approach should solve this. You can check out the `click` docs [here](https://click.palletsprojects.com/en/8.1.x/) for some examples. 

Regarding python 3.7, we will phase out support for that version in tidy3d 2.5.0 so if you need to implement something that doesn't work quite as expected for 3.7 that's fine.

> 3. I did see certain objects in lsf being defined in units of microns themselves, so I left it alone, but if Lumerical officially uses SI units, I can change it.

Yea see their [docs](https://optics.ansys.com/hc/en-us/articles/360034397034-Units-and-normalization-conventions-in-Lumerical-solvers#:~:text=Unless%20otherwise%20stated%2C%20Lumerical%20uses%20SI%20units.&text=This%20is%20reflected%20in%20the,coefficients%20for%20the%20material%20models.&text=The%20electron%20energy%20E%20is,)%20as%20E%20%3D%20%2DqV.) under "Units for optical solvers".  It says "If otherwise stated" , so we just need to convert length units from `m` to `um` and if they are stated as something else in the LSF file we should do the appropriate conversion.

> 4. This was just to test parseability, and the actual coordinates were not considered. I can change them in the example file

That would probably be good. When I tested the feature, I just ran the converter on the example .lsf file and tried pretending I was a user that needed to fix up that output to a working .py file. It would be good to minimize the amount of changes needed, even if it's just an example.

tidy3d/web/cli/app.py
```diff
@@ -114,5 +115,14 @@ def migration():
     migrate()
 
 
+@click.command()
+@click.argument("lsf_file")
```
tylerflex, line 119: I think we can make use of some of click's features to handle the file stuff for us.
https://click.palletsprojects.com/en/8.1.x/arguments/#file-path-arguments

tidy3d/__main__.py
tylerflex, line 1: probably we don't need these changes any more?

tidy3d/web/cli/converter.py
```diff
@@ -0,0 +1,1340 @@
...
+
+    return tidy3d_file
+
+
+# new_file_name = "new_tidy3d_file.py"
+# file_to_read = "example.lsf"
+
+
+def converter_arg(lsf_file, new_file):
+    if ".lsf" not in lsf_file:
```
tylerflex, line 1327: let's use pathlib to check explicitly that this is an extension 
`pathlib.Path('my_file.txt').suffix == ".lsf" # and import pathlib above`

same with `.py`

tidy3d/web/cli/converter.py
```diff
@@ -0,0 +1,1340 @@
...
+        print("Error:", lsf_file, "must be an .lsf file.")
```
tylerflex, line 1328: instead of this, we can just `raise Exception(f"{lsf_file} must be an .lsf file." There's no need to print and return. we can let python handle the error for us.

Note, this will also end the function execution, so no need to return.

Same for the out file handling below.

Both of these can just be regular if statements

```py
if ...:
    raise ...
if ...:
    raise ...
file_string = lsf_reader(...)

```

tidy3d/web/cli/converter.py
```diff
@@ -0,0 +1,1340 @@
...
+        return
+    elif ".py" not in new_file:
+        print("Error:", new_file, "must be a .py file")
+    else:
+        file_string = lsf_reader(lsf_file)
+        if file_string == "File not found!":
+            return
+        writer = open(new_file, "w")
```
tylerflex, line 1336: the standard approach here is
```py
with open(new_file, "w") as f:
    f.write(file_string)
```
which closes automatically even if something wrong happens in the writer.

tidy3d/web/cli/converter.py
```diff
@@ -0,0 +1,1340 @@
...
+
+    structures = "["
+    rect, sph, cyl, polyslab = 0, 0, 0, 0
+    sources, modesrc, planewv, ptdipole, gauss, tfsf = "[", 0, 0, 0, 0, 0
+    monitors = "["
+    modemon, fluxmon, fieldmon = 0, 0, 0
+    gridspec_string = "grid_spec=td.GridSpec.auto()"
+    override_structures, overstrct = "[", 0
+
+    if filename[-4:] != ".lsf":
```
tylerflex, line 1201: this should already be handled before this is called right?

tidy3d/web/cli/converter.py
```diff
@@ -0,0 +1,1340 @@
...
+        print("Needs .lsf format")
+    try:
```
tylerflex, line 1203: this can be handled simply as 

```py
with open(filename, "r") as file_lsf:
    Lines = file_lsf.readlines()
    ... # rest of code
```

it will take care of the file errors and also closing the file at the end or if anything goes wrong.

https://realpython.com/working-with-files-in-python/

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ModeSolverData.pol_fraction` and `ModeSolverData.pol_fraction_waveguide` properties to compute polarization fraction of modes using two different definitions.
 - `ModeSolverData.to_dataframe()` and `ModeSolverData.modes_info` for a convenient summary of various modal properties of the computed modes.
 - Loss upper bound estimation in `PoleResidue` material model.
+- `tidy3d/converter.py` for conversion from .lsf project files into Tidy3D files through command line, e.g. `python -m tidy3d convert example.lsf tidy3d_script.py` or `tidy3d convert example.lsf tidy3d_script.py`
```
tylerflex, line 12: let's change this to just reflect the `tidy3d convert`. I dont think we should support both. And also instead of "Tidy3D files" maybe "equivalent python scripts implementing the project in Tidy3D".

tylerflex (changes requested):
Thanks Alec, this looks much better. See the specific comments on the code for changes.

I ran the example again and noticed again a few strange things in the resulting file, mostly with the units.

1. The vertices again seem to be in units of `meters` here? ,maybe this was the one defined directly as microns in the example file? 
```py
vtx=[(1,0),(2,2),(4,2),(4,1),(3,1)] # comment
random_polygon = td.Structure(
	geometry=td.PolySlab(
		vertices=vtx*1e-6,
		axis=2,
		slab_bounds=(0,2),
	),
	medium=td.Medium(),
	name='random_polygon'
)
```

```py
mode_source_time_0 = td.GaussianPulse(
	freq0=td.C_0/(((0.55*1e6)+(0.55*1e6))/2),
	fwidth=td.C_0/(((0.55*1e6)+(0.55*1e6))/2)/10,
	offset=5.0
)
```
Also, the frequency units seem off here. `td.C_0` is defined in units of `um/sec` so that for a wavelength of 1.0, the frequency in Hz (1/sec) is just `td.C_0`. I'm assuming 0.5 is in microns? in which case we don't want to multiply by 1e6.


2. Frequency width calculation

```py
fwidth=td.C_0/(((0.55*1e6)+(0.55*1e6))/2)/10,
```

Could you explain how `fwidth` is computed (and also maybe freq0 for good measure)? before I was concerned that it was subtracting freq_min from freq_max, which is reasonable except it encounters an edge case when freq_max == freq_min. It seems now there is an averaging happening? which seems to not be the best approach if `fmax != fmin`).  I think the heuristic we want is just to make sure that `freq0 +- fwdith/2` covers all of the frequency points while ensuring that `fwidth >= freq0/10`. The idea is that we just want a gaussian centered at freq0 with width fwidth to cover the desired frequency points with some power at them.

This is just from a quick scan, I'll leave it to you to go through and test yourself. I think the way to test this is to take the output of the converter on this example, add the missing fields to generate a working tidy3d model from it, and just verify that each component works as expected and every value makes sense, especially with the units. For example, the polygon in #1 would be so small it wouldn't show up in the plot at all so something would be immediately apparent.

I think other than that, things seem to be getting closer. I just want to make sure this feature is fully wrapped up before we release it and there are no loose ends, especially on the example `.lsf` file that we are testing with. Thanks for your work in resolving all of my comments.

alec-flexcompute:
Both of these cases are an example of a mismatch between how these values are specified. For instance, in https://optics.ansys.com/hc/en-us/articles/360034924413-addplane-Script-command the wavelengths are specified in units of meters. However, in here: https://www.youtube.com/watch?v=x0r2aQODiYY the entries are given in terms of um. I figure since the actual script example is in terms of microns, I'll use that. What showed up I think was the default wavelength bounds, which I had forgotten to put in meters.

However, I think, since I haven't seen an lsf file produced, it's possible that the user will have to got through these unit issues depending on how the wavelength units are specified.

The rest of this was just a mistake on my part - I've fixed it and will be testing the output file.

alec-flexcompute:
Adjusted the example file to remove overlaps, and also added default boundary conditions for 2-dimensional simulations. For my example file, all one needs to do to get the simulation to run successfully is specify `freqs` and the runtime. 
image

When parsing in the original lsf file from which I based my example file, one needs to do this and also specify the mode source extra dimension, which is missing from this example. I think the default dimensions being 0 is the way to go, as other lsf objects have this as the default as well. The `freq0` and `fwidth` values involved match the settings from lumerical, and everything runs successfully.

tomflexcompute:
Thanks @alec-flexcompute for this great addition. Sorry for the delay due to my international travel. It looks like a very involved PR and my limited hours these days probably won't allow me to review and test it thoroughly. I can do some quick tests but maybe @momchil-flex or @weiliangjin2021 can take a closer look at this?

alec-flexcompute:
Sure, thanks for being such a big help on this! @momchil-flex and @weiliangjin2021 if you want/have time, feel free to test out this on the example.lsf file, as well as the file located here https://lumopt.readthedocs.io/en/latest/tutorial.html (you should only need to specify an extra dimension for the mode source, a frequency range, and a runtime to run), or any lsf project files you happen to have. (the internet seems to be devoid of other lsf files, so that's been a real limit to this project)

momchil-flex:
Hey, this seems to work pretty well! Hopefully it will generalize relatively well to other scripts. I suggest that we add some extra warning's in the output of the command, e.g.

```
Created new Tidy3D file: tidy3d_script.py
NOTE: This is an experimental feature. Not all scripting commands are supported and some variables may still need to be set in the script. Check the comments in the output file for hints.
```

Inside the script, could we standardize all comments to begin with `NOTE: `. Currently this is true for this one `# NOTE: Check order of structures for potential overlap issues` but not for e.g. `# check if solving for correct mode!` or `# does not yet parse to Tidy3d` (and very minor thing: we try to typeset it everywhere as Tidy3D not Tidy3d).

If you hadn't told me about the extra dimension in the mode source, and the frequency range, it would have taken me some time to figure it out (when I run the script I do get useful error messages, but stil...). Is there a way to put a comment about these too or is it not that straightforward?

> Both of these cases are an example of a mismatch between how these values are specified. For instance, in https://optics.ansys.com/hc/en-us/articles/360034924413-addplane-Script-command the wavelengths are specified in units of meters. However, in here: https://www.youtube.com/watch?v=x0r2aQODiYY the entries are given in terms of um. I figure since the actual script example is in terms of microns, I'll use that. What showed up I think was the default wavelength bounds, which I had forgotten to put in meters.

> However, I think, since I haven't seen an lsf file produced, it's possible that the user will have to got through these unit issues depending on how the wavelength units are specified.

Yes, I remember this from working with lumerical. The GUI is in um but the scripting is in m. I don't think the units can be modified, so I think it's safe to assume lsf is always in meters.

As far as I can tell the generated scripts look accurate!

alec-flexcompute:
Thanks Momchil. I've added in your comments about comments. One thing though - it isn't *quite* straightforward to add a warning to check if there's a correct number of nonzero dimensions for each object, if we're going to be robust against using a variable that was earlier defined as being 0. I can do it, but it'd add a fair amount of code to an already large amount of code (since there's a ton of possible inputs), so I'm not sure if it's worth it, style-wise/readability-wise, especially since running it results with an error that basically says what I think I would comment. LMK if you still want it implemented, though - it's definitely doable.

momchil-flex:
> I'm not sure if it's worth it, style-wise/readability-wise, especially since running it results with an error that basically says what I think I would comment.

I think this is acceptable for now then.

momchil-flex (approved):

weiliangjin2021 (approved):
This an exciting feature! It feels like beautiful soup for `.lsf`. One minor suggestion on improving the user's experience is that when we run `tidy3d convert ...`, how about printing a warning on the number of fields that need `# NOTE: please specify`? When I used it the first time, I run the converter and then `black`, and getting error messages.

alec-flexcompute:
That sounds like a good suggestion. I'll implement that. Thanks!

momchil-flex:
Should we merge this in?

alec-flexcompute:
@tylerflex @tomflexcompute any other improvements/suggestions?

tidy3d/web/cli/converter.py
```diff
@@ -0,0 +1,1332 @@
...
+    if Path(f"{lsf_file}").suffix != ".lsf":
+        raise Exception(f"{lsf_file} must be an .lsf file.")
+    elif Path(f"{new_file}").suffix != ".py":
+        raise Exception(f"{new_file} must be a .py file.")
+    else:
+        file_string = lsf_reader(lsf_file)
+        num_specs = file_string.count("NOTE: please specify")
+        with open(new_file, "w") as f:
+            f.write(file_string)
+        f.close()
```
tylerflex, line 1327: this will be closed automatically after the `with` context right above. so can remove.

tidy3d/web/cli/converter.py
```diff
@@ -0,0 +1,1332 @@
...
+    )
+    tidy3d_file += ")"
+
+    return tidy3d_file
+
+
+def converter_arg(lsf_file, new_file):
+    if Path(f"{lsf_file}").suffix != ".lsf":
+        raise Exception(f"{lsf_file} must be an .lsf file.")
+    elif Path(f"{new_file}").suffix != ".py":
```
tylerflex, line 1320: if the exception is raised, there can be no `elif` or `else`, so just
```py
if Path(...)
    raise ....
if Path(...)
    raise  ...
... # continue with no `else`
```
works and is simpler

tidy3d/web/cli/converter.py
```diff
@@ -0,0 +1,1332 @@
...
+    mode_monitor_string += "\tcenter=(" + x + "," + y + "," + z + "),\n"
+    mode_monitor_string += "\tsize=(" + xl + "," + yl + "," + zl + "),\n"
+    mode_monitor_string += "\tfreqs=" + freqs + ",\n"
+    mode_monitor_string += "\tmode_spec=" + mode_spec_name + ",\n"
+    mode_monitor_string += "\tname='" + name + "',\n"
+    mode_monitor_string += ")\n"
+    return mode_monitor_string, name.replace(" ", "_")
+
+
+def lsf_reader(filename=str) -> None:
```
tylerflex, line 1188: to annotate the `filename` as a string, one does `filename: str`. what is here currently assigns `filename` to the `str` type, if it is not provided so that would have some weird unintended consequences if called without an argument.

tidy3d/web/cli/converter.py
```diff
@@ -0,0 +1,1332 @@
...
+    sources, modesrc, planewv, ptdipole, gauss, tfsf = "[", 0, 0, 0, 0, 0
+    monitors = "["
+    modemon, fluxmon, fieldmon = 0, 0, 0
+    gridspec_string = "grid_spec=td.GridSpec.auto()"
+    override_structures, overstrct = "[", 0
+
+    with open(filename) as file_lsf:
+        Lines = file_lsf.readlines()
+    variable_dict = {}
+    for i in range(0, len(Lines)):
```
tylerflex, line 1206: a few things:
1. when iterating over a range, you dont need to supply the start if it is 0, so just 
```py
for i in range(len(Lines)):
```
is more standard
2. you can get the index `i` and the element by using `enumerate()`, which is a bit cleaner 
```py
for i, line in enumerate(Lines):
    ... # no need to `line = Lines[I]`
```

let's use #2 for the two iterations here

tylerflex (commented):
Just a few minor comments but then I think this could go in, thanks @alec-flexcompute . Let me know if something isn't clear in my suggested edits.

alec-flexcompute:
Thanks for these, they're implemented in the latest push

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ModeSolverData.pol_fraction` and `ModeSolverData.pol_fraction_waveguide` properties to compute polarization fraction of modes using two different definitions.
 - `ModeSolverData.to_dataframe()` and `ModeSolverData.modes_info` for a convenient summary of various modal properties of the computed modes.
 - Loss upper bound estimation in `PoleResidue` material model.
+- `tidy3d/converter.py` for conversion from .lsf project files into equivalent python scripts implementing the project in Tidy3D through the command line, e.g. `tidy3d convert example.lsf tidy3d_script.py`
```
tylerflex, line 12: just noticed this. I think the file is no longer here right? so I think we can just say
```
Command line tool `tidy3d convert` for conversion from `.lsf` project files into equivalent python scripts implementing the project in Tidy3D. Usage:  `tidy3d convert example.lsf tidy3d_script.py.
```

tylerflex:
@alec-flexcompute can you also re-open the docs PR with the LSF in the FAQ?

CHANGELOG.md
alec-flexcompute: Nice catch, implemented

tylerflex (approved):

alec-flexcompute:
Oh I just got done testing the generated tidy3D file. It works, so here's that comment.

PR #1166, opened 2023-09-19
remove line number in webapi output

tylerflex:
request from zongfu to remove the line number output in the webapi functions, eg the stuff on the right
image

Seems this info is irrelevant and clutters the output a bit


dbochkov-flexcompute (approved):
Looks great, more room for meaningful info. Seems like having a shared console instance between everything pays off.

I see it has tag 2.4, is it going to go in a 2.4 patch?

tylerflex:
I was going to merge it into develop and have it be released in 2.4.1

PR #1167, opened 2023-09-19
fix adjoint for very thin box

tylerflex:
It was previously trying to measure the permittivity inside by looking two grid cells inside of the boundary. This was causing errors when the PermittivityData for the box was less than 1 grid cell thick. This PR fixes this.

### DIFF
--- a/tidy3d/plugins/adjoint/components/geometry.py
+++ b/tidy3d/plugins/adjoint/components/geometry.py
@@ -237,7 +237,7 @@ class JaxBox(JaxGeometry, Box, JaxObject):
                     eps_data = grad_data_eps.field_components[eps_field_name].isel(f=0)
 
                     # get the permittivity values just inside and outside the edge
-                    n_cells_in = 3
+                    n_cells_in = 2
                     isel_out = 0 if min_max_index == 0 else -1
                     isel_ins = n_cells_in if min_max_index == 0 else -n_cells_in - 1
                     eps2 = eps_data.isel(**{dim_normal: isel_out})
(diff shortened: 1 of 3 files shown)


tidy3d/plugins/adjoint/components/geometry.py
```diff
@@ -237,7 +237,7 @@ def store_vjp(
                     eps_data = grad_data_eps.field_components[eps_field_name].isel(f=0)
 
                     # get the permittivity values just inside and outside the edge
-                    n_cells_in = 3
+                    n_cells_in = 2
```
weiliangjin2021, line 240: maybe a naive question: why is it 2 not 1?

tylerflex: it's not obvious, actually. 0 corresponds to just outside of the box (since the permittivity monitor extends a bit beyond the bounds), 1 corresponds usually to the subpixelled grid cell. we want to grab the grid cell that is completely inside of the object, which is at least 2 (at least it was). When the box is very thin, it has 3 grid cells, so it becomes complicated. 

Maybe I can ask you, if you sub pixel a very thin box (smaller than a grid cell) does the structure only influence one grid cell on the yee lattice?

tylerflex:
By the way, something else popped up here that I'm working on, this might not be ready until tomorrow.

tidy3d/plugins/adjoint/components/geometry.py
weiliangjin2021: I think it can influence 0 or 1 or 2:

- 0: if it's so thin that our sampling doesn't catch it
- 1: it's completely within a cell
- 2: it's sandwiched between 2 cells.

tylerflex:
Alright I tested this and it worked with all existing notebooks. so from my perspective should be good to go for 2.4.1.

weiliangjin2021 (approved):

PR #1168, opened 2023-09-20
add time zone to web logging output

tylerflex:


tylerflex:
didn't work. MC is looking into this.

PR #1172, opened 2023-09-25
Added two-photon absorption nonlinearity

caseyflex:
- Adds TwoPhotonAbsorption and NonlinearModel classes. Also, nonlinearities of different types can be combined. Note that the beta and n2 parameters for these classes can be complex. Also, these classes need n0, which can be provided or obtained automatically from the medium and source frequencies.
- Adds warning for normalizing against CustomSourceTime or ContinuousWave sources


tidy3d/components/medium.py
```diff
@@ -635,7 +676,7 @@ class Medium(AbstractMedium):
     @pd.validator("nonlinear_spec", always=True)
     def _validate_nonlinear_spec(cls, val):
         """Check compatibility with nonlinear_spec."""
-        if val is None or isinstance(val, NonlinearSusceptibility):
+        if val is None or isinstance(val, NonlinearSusceptibility) or isinstance(val, KerrAndTPA):
```
tylerflex, line 679: you can do 
```py
isinstance(val, (TypeA, TypeB))
```
and it will check both.

tidy3d/components/medium.py
```diff
@@ -143,11 +143,52 @@ class NonlinearSusceptibility(NonlinearSpec):
...
+    -------
+    >>> medium = Medium(permittivity=2, nonlinear_spec=KerrAndTPA(chi3=1, sigma3=1))
+    """
+
+    chi3: float = pd.Field(
+        0, title="Chi3", description="Chi3 nonlinear susceptibility.", units="um^2 / V^2"
+    )
+
+    sigma3: float = pd.Field(
+        0,
```
tylerflex, line 185: are these able to be negative? if not, can make them `pd.NonNegativeFloat`

tylerflex (commented):
just a few minor comments as I looked over this.

tidy3d/components/medium.py
caseyflex: ah, thanks

tidy3d/components/medium.py
caseyflex: chi3 can be negative. Negative sigma3 makes it active so yeah it would be good to at least warn in this case.

weiliangjin2021: Since we have `allow_gain` field in Medium, shall we only allow negative sigma3 when `allow_gain=True`? This is useful in backend to catch divergence.

tidy3d/components/medium.py
```diff
@@ -143,11 +143,52 @@ class NonlinearSusceptibility(NonlinearSpec):
...
+    The nonlinear constitutive relation is solved iteratively; it may not converge
+    for strong nonlinearities. Increasing `numiters` can help with convergence.
+
+    Note
+    ----
+    For complex fields (e.g. when using Bloch boundary conditions), the nonlinearity
+    is applied separately to the real and imaginary parts, so that the above equation
+    holds when both E and :math:`P_{NL}` are replaced by their real or imaginary parts.
+    The nonlinearity is only applied to the real-valued fields since they are the
+    physical fields.
```
momchil-flex, line 166: I'm actually confused by this note (I know that it's just copied from NonlinearSusceptibility...)

The first sentence says that the nonlinearity is applied separately to the real and imaginary parts; but the last sentence *seems* to say that the nonlinearity is only applied to the real fields. Are the "real-valued" fields referenced in the last sentence just each of Re(E) and Im(E), separately? If so can it be reworded to avoid that confusion - or maybe completely removed since the first sentence is already clear. Maybe something like "The nonlinearity is applied to the real/imaginary components separately since each of those represents physical fields."

tidy3d/components/medium.py
caseyflex: Yeah, that's a good suggestion. I'll add a validator to `AbstractMedium`, since that would have access to `allow_gain` and `nonlinear_spec`

tidy3d/components/medium.py
caseyflex: You're right, this is confusing as it stands. I'll change the wording for all such classes

caseyflex: yeah, it's each of the real/imaginary components, separately.

tidy3d/components/medium.py
```diff
@@ -140,15 +152,112 @@ class NonlinearSusceptibility(NonlinearSpec):
...
+
+    Example
+    -------
+    >>> nonlinear_susceptibility = NonlinearSusceptibility(chi3=1)
+    >>> nonlinear_spec = NonlinearSpec(models=[nonlinear_susceptibility])
+    >>> medium = Medium(permittivity=2, nonlinear_spec=nonlinear_spec)
+    """
+
+    models: Tuple[NonlinearModelType, ...] = pd.Field(
+        ...,
```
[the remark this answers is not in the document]
tylerflex: if you feel it's important to force this to be specified that makes sense to, just something that caught my eye

weiliangjin2021 (approved):

momchil-flex:
I am a tad unhappy with the naming of `numiters` as opposed to `num_iters`. Generally we follow the `num_***` convention in everything (e.g. `num_poles`, `num_layers`... and even the [num_iters](https://github.com/flexcompute/tidy3d/blob/f60bf47f84264bc622e548fa41dfc344fb85ceaf/tidy3d/plugins/dispersion/web.py#L132) in the fitter). However, we already released the first iteration of nonlinear stuff in 2.4, so I don't know if we want to make that change. That said however, are you introducing any breaking changes here anyway?

caseyflex:
> I am a tad unhappy with the naming of `numiters` as opposed to `num_iters`. Generally we follow the `num_***` convention in everything (e.g. `num_poles`, `num_layers`... and even the [num_iters](https://github.com/flexcompute/tidy3d/blob/f60bf47f84264bc622e548fa41dfc344fb85ceaf/tidy3d/plugins/dispersion/web.py#L132) in the fitter). However, we already released the first iteration of nonlinear stuff in 2.4, so I don't know if we want to make that change. That said however, are you introducing any breaking changes here anyway?

I am not introducing any breaking changes here. I can rename numiters to num_iters, and keep the old parameter in NonlinearSusceptibility and make it still work for backward compatibility.

@tylerflex I would like to move num_iters into NonlinearSpec, rather than duplicating it in every model. Would you be ok with this? I know we didn't want too many specific parameters in the general NonlinearSpec, but I feel this one is sufficiently fundamental to move there. And that would save the user from having to type it in each NonlinearModel.

tylerflex:
> @tylerflex I would like to move num_iters into NonlinearSpec, rather than duplicating it in every model. Would you be ok with this? I know we didn't want too many specific parameters in the general NonlinearSpec, but I feel this one is sufficiently fundamental to move there. And that would save the user from having to type it in each NonlinearModel.

As long as you think we can be pretty sure that this will be a field in every nonlinear model we add, specifically that future nonlinear models will not need their own `num_iters`.  If it's just about backwards compatibility, we can easily handle this I think by having `num_iters` and `numiters` but adding a validator on `numiters` which asks users to use `num_iters` and then just sets it with the supplied value. Also `alias` might be worth looking into. We might be able to set a field `numiters` with an alias of `num_iters`.

https://docs.pydantic.dev/latest/concepts/alias/

tidy3d/components/medium.py
```diff
@@ -82,24 +82,122 @@ def _eps_model(self, frequency: float) -> complex:
...
+        medium: AbstractMedium,
+        freqs: List[pd.PositiveFloat],
+    ) -> complex:
+        """Get a single value for n0."""
+        # if n0 is not specified, we need to compute it
+        if n0 is None:
+            if len(freqs) == 0:
+                raise SetupError(
+                    f"Class '{self}' cannot determine 'n0' in the absence of sources. "
+                    "Please either specify 'n0' or add sources to the simulation."
```
tylerflex, line 122: shouldn't this be asking the user to supply frequencies? could be confusing if called directly.

tidy3d/components/medium.py
```diff
@@ -82,24 +82,122 @@ def _eps_model(self, frequency: float) -> complex:
...
+
+    def _validate_medium(self, medium: AbstractMedium):
+        """Any additional validation that depends on the medium"""
+        pass
+
+    def _validate_medium_freqs(self, medium: AbstractMedium, freqs: List[pd.PositiveFloat]) -> None:
+        """Any additional validation that depends on the central frequencies of the sources."""
+        pass
+
+    def _get_n0(
```
tylerflex, line 110: I went through to try to understand this a bit better and refactored a bit, feel free to implement if it makes it clearer.

```py
        freqs = set(freqs)

        if not len(freqs):
            if n0 is None:
                raise ...
            return n0

        if len(freqs) > 1:
            if n0 is None:
                raise ...
            log.warning(...)
            return n0

        if len(freqs) == 1:
            freq = freqs[0]
            n, k = medium.nk_model(freq)
            nk = n + 1j * k

            if n0 is not None and not np.isclose(n0, nk):
                log.warning(...)
            return nk

```

tidy3d/components/medium.py
```diff
@@ -82,24 +82,122 @@ def _eps_model(self, frequency: float) -> complex:
...
+        0, title="Chi3", description="Chi3 nonlinear susceptibility.", units="um^2 / V^2"
+    )
+
     numiters: pd.PositiveInt = pd.Field(
-        1,
+        None,
         title="Number of iterations",
-        description="Number of iterations for solving nonlinear constitutive relation.",
+        description="Number of iterations for solving nonlinear constitutive relation. "
+        "Deprecated; use 'NonlinearSpec.num_iters' instead.",
```
tylerflex, line 193: this only applies if this model is added to nonlinearspec.models? It seems if we set `Medium.nonlinearspec = NonlinearSusceptibility` we should use `numiters`, could you explain? And probably a bit more discussion here could help clarify

tidy3d/components/medium.py
```diff
@@ -108,28 +206,91 @@ def _validate_numiters(cls, val):
...
+    Example
+    -------
+    >>> tpa_model = TwoPhotonAbsorption(beta=1)
+    """
+
+    beta: Complex = pd.Field(
+        0,
+        title="TPA coefficient",
+        description="Coefficient for two-photon absorption (TPA).",
+        units="um / W",
```
tylerflex, line 240: let's use the strings for constants from `constants.py` and just insert them here. Reason is if we ever want to change how they are displayed, we can do it globally by just changing `constants.py`.

tidy3d/components/medium.py
```diff
@@ -140,15 +301,114 @@ class NonlinearSusceptibility(NonlinearSpec):
...
+        if values.get("models") is not None:
+            for model in values["models"]:
+                if isinstance(model, NonlinearSusceptibility) and model.numiters is not None:
+                    log.warning(
+                        "The API for the number of nonlinear iterations has changed. "
+                        "The old usage 'NonlinearSusceptibility.numiters' is deprecated "
+                        "and will be removed in a future release. The new usage is "
+                        "'NonlinearSpec.num_iters'."
+                    )
+                    return model.numiters
```
tylerflex, line 410: so this validator will seemingly override the global `NonlinearSpec.num_iters` with the `NonlinearSusceptibility` model's `numiters`? I'm a bit confused about this.  Could you explain what this does?

tylerflex (changes requested):
Thanks @caseyflex, pretty minor comments. Still a little confused about the numiters behavior so asked for some clarification (and potentially a strange behavior spotted in a validator), but otherwise looks good.

tidy3d/components/medium.py
```diff
@@ -82,24 +82,122 @@ def _eps_model(self, frequency: float) -> complex:
...
+class NonlinearModel(ABC, Tidy3dBaseModel):
+    """Abstract model for a nonlinear material response.
+    Used as part of a :class:`.NonlinearSpec`."""
+
+    def _validate_medium_type(self, medium: AbstractMedium):
+        """Check that the model is compatible with the medium."""
+        if isinstance(medium, AbstractCustomMedium):
+            raise ValidationError(
+                f"'NonlinearModel' of class {type(self)} is not currently supported "
+                f"for medium class {type(medium)}."
```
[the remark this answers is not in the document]
caseyflex: yeah, that looks better

tidy3d/components/medium.py
caseyflex: I don't think this method will be called directly, since it is private.

caseyflex: In the only use case I have in mind, the freqs are coming from the simulation sources

tidy3d/components/medium.py
caseyflex: thanks, I used some of this in the revision

tidy3d/components/medium.py
caseyflex: this was confusing. I removed it. `AbstractMedium._nonlinear_num_iters` now handles all the backwards compatibility stuff.

tidy3d/components/medium.py
caseyflex: ok, I had to add WATT and VOLT to this file, hopefully that's ok

tylerflex: yea sounds good

tidy3d/components/medium.py
```diff
@@ -82,54 +82,218 @@ def _eps_model(self, frequency: float) -> complex:
...
 
-class NonlinearSpec(ABC, Tidy3dBaseModel):
-    """Abstract specification for adding a nonlinearity to a medium.
+class NonlinearModel(ABC, Tidy3dBaseModel):
+    """Abstract model for a nonlinear material response.
+    Used as part of a :class:`.NonlinearSpec`."""
+
+    def _validate_medium_type(self, medium: AbstractMedium):
+        """Check that the model is compatible with the medium."""
+        if isinstance(medium, AbstractCustomMedium):
```
weiliangjin2021, line 91: also not support time modulated medium: `if isinstance(medium, AbstractCustomMedium) or medium.time_modulated`?

PR #1174, opened 2023-09-25
Option to flatten all geometries in flatten_geometry

momchil-flex:


lucas-flexcompute (approved):
No problems here as far a I can see.

PR #1176, opened 2023-09-25
Warn if too many frequencies; too many modes; or too many grid points in mode source or monitor

momchil-flex:
I went with just warnings for now, including warnings that the behavior may change to a hard constraint in the future, if we decide to impose some strict limits e.g. in 2.5.

### DIFF
--- a/tidy3d/components/monitor.py
+++ b/tidy3d/components/monitor.py
@@ -20,6 +20,8 @@ from ..log import log
 
 BYTES_REAL = 4
 BYTES_COMPLEX = 8
+WARN_NUM_FREQS = 2000
+WARN_NUM_MODES = 100
 
 
 class Monitor(Box, ABC):
@@ -146,6 +148,19 @@ class FreqMonitor(Monitor, ABC):
             raise ValidationError("'freqs' must not be empty.")
         return val
 
+    @pydantic.validator("freqs", always=True)
+    def _warn_num_freqs(cls, val, values):
+        """Warn if number of frequencies is too large."""
+        if len(val) > WARN_NUM_FREQS:
+            log.warning(
+                f"A large number ({len(val)}) of frequencies detected in monitor "
+                f"'{values['name']}'. This can lead to solver slow-down and increased cost. "
+                "Consider decreasing the number of frequencies in the monitor. This may become a "
+                "hard limit in future Tidy3D versions.",
+                custom_loc=["freqs"],
+            )
+        return val
+
     @cached_property
     def frequency_range(self) -> FreqBound:
         """Frequency range of the array ``self.freqs``.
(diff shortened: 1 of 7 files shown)


tomflexcompute (approved):
Thanks. `WARN_NUM_MODES = 100` is probably much larger than necessary but I suppose it's fine if we can handle it. I can't think of any situation where someone would like to solve for 100 modes. Practically speaking something like 20 should be more than sufficient.

momchil-flex:
I think this is only relevant when using PML. I think this is equivalent to the "number of trial modes" in [Lumerical](https://optics.ansys.com/hc/en-us/articles/360034917353-FDE-solver-analysis-Modal-Analysis-Tab) and I think what's happening when you tried to compute modes of a sharp bend and you were getting no modes out was that all the modes in whatever "number of trial modes" setting you were using were spurious, and somehow automatically filtered. Technically increasing that value could help you find the physical modes.

I agree though that the warning could be maybe even at e.g. 50.

tidy3d/components/monitor.py
```diff
@@ -146,6 +148,19 @@ def _freqs_non_empty(cls, val):
...
 
+    @pydantic.validator("freqs", always=True)
+    def _warn_num_freqs(cls, val, values):
+        """Warn if number of frequencies is too large."""
+        with log as consolidated_logger:
+            if len(val) > WARN_NUM_FREQS:
+                consolidated_logger.warning(
+                    f"A large number ({len(val)}) of frequencies detected in monitor "
+                    f"'{values['name']}'. This can lead to solver slow-down and increased cost. "
+                    "Consider decreasing the number of frequencies in the monitor. This may become "
```
tylerflex, line 159: "may become a hard limit" (missing "a")

tylerflex (approved):
looks good, just a minor typo

tidy3d/components/simulation.py
```diff
@@ -1058,6 +1061,42 @@ def _validate_monitor_size(self) -> None:
...
+
+        def warn_mode_size(monitor: AbstractModeMonitor, msg_header: str, custom_loc: List):
+            """Warn if a mode component has a large number of points."""
+            num_cells = np.prod(self.discretize_monitor(monitor).num_cells)
+            if num_cells > WARN_MODE_NUM_CELLS:
+                consolidated_logger.warning(
+                    msg_header + f"has a large number ({num_cells:1.2e}) of grid points. "
+                    "This can lead to solver slow-down and increased cost. "
+                    "Consider making the size of the component smaller, as long as the modes "
+                    "of interest decay by the plane boundaries."
```
dbochkov-flexcompute, line 1075: I think you forgot `custom_loc=custom_loc` here

tidy3d/components/simulation.py
```diff
@@ -1058,6 +1061,42 @@ def _validate_monitor_size(self) -> None:
...
+                    )
+                    msg_header = f"Mode source at sources[{src_ind}] "
+                    custom_loc = ["sources", src_ind]
+                    warn_mode_size(monitor=monitor, msg_header=msg_header, custom_loc=custom_loc)
+
+        with log as consolidated_logger:
+            for monitor in self.monitors:
+                if isinstance(monitor, AbstractModeMonitor):
+                    msg_header = f"Mode monitor '{monitor.name}' "
+                    custom_loc = ["monitors", monitor.name]
```
dbochkov-flexcompute, line 1097: let's make it `custom_loc = ["monitors", mnt_ind]` so that it's consistent with pydantic style

tidy3d/components/simulation.py
momchil-flex: Ah, good catch! Is there a good way to test that the `custom_loc` is set correctly? Can it be added to the automatic tests?

tidy3d/components/simulation.py
momchil-flex: Ok

PR #1177, opened 2023-09-26
Avoid creating unecessary geometry groups of single elements

lucas-flexcompute:
Minor simplification in return from `Geometry.from_gds` to avoid creating a group with a single element.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Internal refactor of Web API functionality.
+- `Geometry.from_gds` doesn't create unecessary groups of single elements.
 
 ### Fixed
 
@@ -30,7 +31,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bug with truly anisotropic `JaxCustomMedium` in adjoint plugin.
 - Bug in adjoint plugin when `JaxBox` is less than 1 grid cell thick.
 - Bug in `adjoint` plugin where `JaxSimulation.structures` did not accept structures containing `td.PolySlab`.
-
 
 ## [2.4.0] - 2023-9-11
 
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -885,7 +885,7 @@ class Geometry(Tidy3dBaseModel, ABC):
         dilation: float = 0.0,
         sidewall_angle: float = 0,
         reference_plane: PlanePosition = "middle",
-    ) -> GeometryGroup:
+    ) -> Geometry:
         """Import a ``gdstk.Cell`` or a ``gdspy.Cell`` and extrude it into a GeometryGroup.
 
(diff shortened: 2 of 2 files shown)


tidy3d/components/geometry/base.py
```diff
@@ -885,7 +885,7 @@ def from_gds(
         dilation: float = 0.0,
         sidewall_angle: float = 0,
         reference_plane: PlanePosition = "middle",
-    ) -> GeometryGroup:
+    ) -> Geometry:
```
tylerflex, line 888: should be `Union[Geometry, GeometryGroup]`? (and in docstring)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Internal refactor of Web API functionality.
+- `Geometry.from_gds` doesn't create unecessary groups of single elements.
```
tylerflex, line 12: slight tweak, maybe just 
```
`Geometry.from_gds` returns `Geometry` instead of `GeometryGroup` for single elements.
```

tylerflex (commented):
Just some minor tweaks

Technically this breaks backwards compatibility, if the user has something set up to do something like

```
geo = PolySlab.from_gds(...)
geometries = geo.geometries
```

do you think this is ok?

tidy3d/components/geometry/base.py
lucas-flexcompute: I don't think so… `GeometryGroup` is a `Geometry` type ([see here](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/geometry/utils.py#L13-L22)).

CHANGELOG.md
tylerflex: I think probably this comment is also irrelevant because I forgot that GeometryGroup is a Geometry type.

lucas-flexcompute:
> Technically this breaks backwards compatibility, if the user has something set up to do something like
> 
> ```
> geo = PolySlab.from_gds(...)
> geometries = geo.geometries
> ```
> 
> do you think this is ok?

I don't think this would be a problem. I highly doubt any user is diving that deep into the geometry at this point, specially because it doesn't affect `PolySlab.from_gds`, only `Geometry.from_gds`, which even fewer users should be using at this point. But you are right, the return value does change in this case.

tylerflex:
ok if it doesn't change `PolySlab`, then this sounds totally fine to me. 

On that point, just want to check, do the `PolySlab.from_gds()` internals require any special handling? or is it all taken care of?

lucas-flexcompute:
No, the dependency is in the reversed direction: this function uses `PolySlab.from_gds` through [`from_shapely`](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/geometry/utils.py#L84). That's where we choose the best version (`PolySlab` or `ComplexPolySlab`) and deal with holes.

tylerflex:
got it, ok sounds good to me, thanks for clarifying.

tylerflex (approved):

PR #1180, opened 2023-09-27
A few more adjoint fixes

tylerflex:
1. When `FieldMonitor` objects were in `JaxSimulation.output_monitors` and `.flux` was called with symmetry, the symmetry covered areas were being included in the flux calculation and therefore giving `nan` as the interpolation coordinates were outside of the data range. Temporarily fixed this to avoid errors by converting `nan` values to `0`, but not sure if this is the appropriate way to handle it. For example, may undercount flux by 2x, 4x, 8x, depending on how many symmetry planes there are. Seems the deeper problem is that the fields are not getting copied over to the symmetry regions. spent a while trying to figure it out but got quite confused by the FieldData internals.
2. Same obscure `PytreeDef` error was occurring when a regular `JaxSimulation.monitor` had a `.freqs` field as a `np.ndarray`.  Added the previous fix to this and it seems to rectify issue.
3. Fix the DataArray selection to handle vectorized inputs, and also fixed the symmetry expansion to properly pass kwargs instead of dicts, which JaxDataArray can't handle currently.

thanks @e-g-melo for catching these and your patience..

### DIFF
--- a/tests/test_plugins/test_adjoint.py
+++ b/tests/test_plugins/test_adjoint.py
@@ -35,6 +35,7 @@ from tidy3d.plugins.adjoint.utils.filter import ConicFilter, BinaryProjector, Ci
 from tidy3d.web.container import BatchData
 
 from ..utils import run_emulated, assert_log_level, log_capture, run_async_emulated
+from ..test_components.test_custom import CUSTOM_MEDIUM
 
 TMP_PATH = None
 FWD_SIM_DATA_FILE = "adjoint_grad_data_fwd.hdf5"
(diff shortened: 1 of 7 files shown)


tests/test_plugins/test_adjoint.py
```diff
@@ -1482,12 +1490,26 @@ def test_polyslab_structures(use_emulated_run):
...
+        name="flux",
+    )
+
     VERTICES = np.array(
         [[-1.5, -0.5, -0.5], [-0.5, -0.5, -0.5], [-1.5, 0.5, -0.5], [-1.5, -0.5, 0.5]]
     )
     FACES = np.array([[1, 2, 3], [0, 3, 2], [0, 1, 3], [0, 2, 1]])
     STL_GEO = td.TriangleMesh.from_trimesh(trimesh.Trimesh(VERTICES, FACES))
 
+    from ..test_components.test_custom import CUSTOM_MEDIUM
```
momchil-flex, line 1506: very nit but why not import at top level?

tidy3d/plugins/adjoint/components/data/monitor_data.py
```diff
@@ -167,8 +167,8 @@ def package_flux_results(self, flux_values: JaxDataArray) -> float:
         """How to package the dictionary of fields computed via self.colocate()."""
         flux_data = flux_values
         if isinstance(flux_data, JaxDataArray):
-            return jnp.sum(flux_data.values)
-        return jnp.sum(flux_data)
+            return jnp.sum(jnp.nan_to_num(flux_data.values))
+        return jnp.sum(jnp.nan_to_num(flux_data))
```
momchil-flex, line 171: still needed?

momchil-flex (commented):
Awesome. Running backend adjoint tests with this will let you know if anything pops up.

momchil-flex (approved):
Backend tests successful.

tidy3d/plugins/adjoint/components/data/monitor_data.py
tylerflex: no but maybe doesn't hurt. what do you think?

tests/test_plugins/test_adjoint.py
tylerflex: could, I can change it.

tidy3d/plugins/adjoint/components/data/monitor_data.py
momchil-flex: I don't know, I feel like nan can alert us about something wrong, while converting to 0-s would produce wrong values and be harder to catch.

PR #1182, opened 2023-09-28
add time zone to web logging output

huwei-mc:


tylerflex:
Looks good to me. As discussed, we might want to leave this for 2.5 so we have some time to test it. Slightly concerned about edge cases if this fails due to something with `astimezone()` on some systems.

tylerflex:
After some internal discussion, we'll merge this into the 2.5 branch. Should be plenty of time to test then. Thanks @huwei-mc . I'll close this and will merge #1183, which I rebased against pre/2.5 for simplicity.

PR #1184, opened 2023-09-28
lsf - added more monitor parsing capability

alec-flexcompute:
Added capability to parse in the following (marked in green) monitors from lsf. These are permittivity monitors, field time monitors, and field monitors expressed in different solvers. I currently have it so that monitors that record the same information but in different solver regimes (e.g. addindex vs addemeindex) route to the same function in the parser.

This should thus handle every monitor instance in lumerical for which there is a Tidy3D counterpart.

Also added monitor.lsf as an example lsf file for testing that includes an example of every monitor we can handle, and a patch-up handling cases where no e.g. sources, structures, monitors, and override structures are given.

image

### DIFF
--- a/tests/data/monitors.lsf
+++ b/tests/data/monitors.lsf
@@ -1,0 +1,140 @@
+# https://optics.ansys.com/hc/en-us/articles/360034923553-Lumerical-scripting-language-Alphabetical-list
+# https://readthedocs.org/projects/lumopt/downloads/pdf/latest/
+
+switchtolayout;
+selectall;
+delete;
+
+## SIM PARAMS
+size_x=16e-6;
+size_y=18e-6;
+size_z=0;
+mesh_x=20e-9;
+mesh_y=20e-9;
+finer_mesh_size=0.5e-6;
+mesh_accuracy=2;
+
+addsphere;
+set("name","s");
+set("x",1e-6);
+set("y",2e-6);
+set("z",0);
+set("radius",0.5e-6);
+
+
+
+## FDTD
+addfdtd;
+set('dimension','2D');
+set('background index',1.44);
+set('mesh accuracy',mesh_accuracy);
+set('x',0.0);
+set('x span',size_x);
+set('y',0.0);
+set('y span',size_y);
+set('z span',size_z);
+set('force symmetric y mesh',true);
+set('y min bc','Anti-Symmetric');
+set('pml layers',12);
+
+
+## OPTIMIZATION FIELDS MONITOR IN OPTIMIZABLE REGION
+addpower;
+set('name','opt_fields');
+set('monitor type','2D Z-normal');
+set('x',-6.5e-6);
+set('x span',finer_mesh_size);
+set('y',7.5e-6);
+set('y span',finer_mesh_size);
+
+
+#adddipole;
+#set("x",0);
+#set("y",1e-6);
+#set("z",0e-6);
+
+addefieldmonitor;
+set("name","E_field");
+set("monitor type",6);  # 2D y-normal
+set("x",-4.5e-6);
+set("x span",0.5e-6);
+set("y",7.5e-6);
+set("z",0);
+set("y span",0.5e-6);
+set("record electrostatic potential",1);
+set("save data",1);
+filename = "electric_field.mat";
+set("filename",filename);
+
+addtestUnavailableCommand;
+
+addmodeexpansion;
+set("name","mode monitor");
+set("x",-3.5e-6);
+set("x span",5e-6);
+set("y",7.5e-6);
+set("z",0);
+set("y span",1e-6);
+
+addindex;
+set("name","index_monitor");
+set("x",0);
+set("x span",5e-6);
+set("z",0);
+set("y",8e-6);
+set("y span",2.5e-6);
+
+addeffectiveindex;
+set("name","neff");
+set("x",1e-6);
+set("x span",0.7e-6);
+set("y",5e-6);
+set("y span",0.7e-6);
+
+addtime;
+
+set("name","time_1");
+set("monitor type",1);  # point
+set("x",2.5e-6);
+set("z",0);
+set("y",5e-6);
+
+addmovie;
+set("name","movie_1");
+set("monitor type",3);  # 1 = 2D x-normal,  2 = 2D y-normal,  3 = 2D z-normal
+set("x",4e-6);
+set("x span",0.5e-6);
+set("y",5e-6);
+set("y span",5e-6);
+set("z",0);
+set("lock aspect ratio",1);
+set("horizontal resolution",240);
+
+addprofile;
+set("name","field_profile");
+set("monitor type",7);  # 2D z-normal
+set("x",6e-6);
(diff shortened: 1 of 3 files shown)


tests/data/monitors.lsf
tylerflex, line 1: can we add another test to run the converter on this file as well?

tidy3d/web/cli/converter.py
```diff
@@ -1093,6 +1206,9 @@ def _addefieldmonitor(Lines, i: int, fieldmon: int):
     field_monitor_string += "\tsize=(" + xl + "," + yl + "," + zl + "),\n"
     field_monitor_string += "\tname='" + name + "',\n"
     field_monitor_string += "\tfreqs=" + freqs + ",\n"
+    field_monitor_string += (
+        "\tfields=['Ez'], # NOTE: ensure these are the field(s) you want to record\n"
```
tylerflex, line 1210: probably safest is just to leave this field empty (defaults to all E and H).

tylerflex (commented):
looks good, let's just add a test for the new file. I also just tried this out and it worked as expected. thanks!

tests/data/monitors.lsf
alec-flexcompute: added a test to test_cli. LMK if that's all good

tylerflex: I thought we had a test in `tests/test_plugins/test_convert.py`?

Let's make a test there to call the regular `convert` function on the two test input files and just ensure that they run without erroring.

alec-flexcompute: Ah, I don't think we had one. Should I create a test_convert.py file, or should I add to the test_cli file? The test I wrote for test_cli.py works fine with the monitors test, though

tylerflex: sorry I meant `test_package` not `test_plugins`. I think it would probably be best in `test_package/test_convert.py` and I think let's avoid `os.system()` and just call the functions on the two data files directly just for simplicity, `os.system()` can be finicky

alec-flexcompute: Alright, this revision passed the tests (navigating the file system for the testing is finicky as well). LMK if I should change it

tests/test_package/test_convert.py
```diff
@@ -0,0 +1,16 @@
+from tidy3d.web.cli import tidy3d_cli
```
tylerflex, line 1: It's good to put a top level comment to describe what the file is testing.
```
"""Description"""
```

tests/test_package/test_convert.py
```diff
@@ -0,0 +1,16 @@
...
+from tidy3d.web.cli.converter import converter_arg
+from pathlib import Path
+
+
+def test_tidy3d_converter():
+    lsf_path = str(Path.cwd()) + "/tests/data/example.lsf"
```
tylerflex, line 7: the paths in tests assume you are in the root directory, so
```py
lsf_path = "tests/data/example.lsf"
```
should work. Let's do that for consistency.

tests/test_package/test_convert.py
```diff
@@ -0,0 +1,16 @@
...
+    converter_arg(lsf_path, "lsf_converter_test.py")
+    file_path = Path("lsf_converter_test.py")
+    assert file_path.exists()
+    Path.unlink(file_path)
+    lsf_path = str(Path.cwd()) + "/tests/data/monitors.lsf"
```
tylerflex, line 12: the 2nd half is duplicating the first. There's a pytest tool called [parameterize](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html) that we use often to test on arrays of inputs, let's implement that here. You can see an example [here](https://github.com/flexcompute/tidy3d/blob/develop/tests/test_components/test_custom.py#L77-L82). 

The idea is to simply pass the input file name to the test function and then use the same logic to test each one.

```py
@pytest.mark.parameterize(...) # set filename values here
def test_tidy3d_converter(filename):
    # construct path out of supplied filename
```

tests/test_package/test_convert.py
```diff
@@ -0,0 +1,16 @@
+from tidy3d.web.cli import tidy3d_cli
+from tidy3d.web.cli.converter import converter_arg
+from pathlib import Path
+
+
+def test_tidy3d_converter():
```
tylerflex, line 6: good to provide a docstring for every function just describing what it does.

tests/test_package/test_convert.py
```diff
@@ -0,0 +1,16 @@
...
+    lsf_path = str(Path.cwd()) + "/tests/data/example.lsf"
+    converter_arg(lsf_path, "lsf_converter_test.py")
+    file_path = Path("lsf_converter_test.py")
```
tylerflex, line 9: We've been using temporary files for things like this in tests, which helps the tests be faster and avoids unnecessary file creation bugs. There is a [tmp_path](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) tool in pytest for this, see example [here](https://github.com/flexcompute/tidy3d/blob/develop/tests/test_components/test_IO.py#L65-L69)

When you pass `tmp_path` as an argument, you can basically write to it and it will be handled virtually. Let's do something similar here for consistency with the rest of the tests.

tests/test_package/test_convert.py
```diff
@@ -0,0 +1,16 @@
...
+    assert file_path.exists()
+    Path.unlink(file_path)
```
tylerflex, line 11: note, using `tmp_path`, should be no need to do this, we can also check that the out file exists through 
```
os.path.exists(path)
```
if `path` is just a string.

tylerflex (changes requested):
Thanks @alec-flexcompute, just a few things for refining the test. It would be good for you to learn some of these advanced pytest features (parameterize, tmp_path) as we use them throughout the code and they are really useful.

tests/test_package/test_convert.py
alec-flexcompute: It appears this doesn't pass the tests. I've added os.getcwd()

tylerflex: did you try without the leading `/`? I did this and it worked for me
```py
@pytest.mark.parametrize('filename', ('example.lsf', 'monitors.lsf'))
def test_tidy3d_converter(tmp_path, filename):
    """Test tidy3d convert."""

    lsf_path = f"tests/data/{filename}"
    out_path = str(tmp_path / "out.py")
    converter_arg(lsf_path, out_path)
    assert os.path.exists(out_path)
```

tylerflex (approved):

tylerflex:
@alec-flexcompute we will merge this into `pre/2.5`. Could you please get it ready for that? (rebasing against the most up to date pre/2.5, fixing any conflicts that come up, single commit). Make sure to make a backup of the branch if you aren't comfortable with it, thanks!

alec-flexcompute:
Rebased to 2.5

tylerflex:
There should be just one commit in this PR, could you double check that you pulled the most recent pre/2.5 branch before rebasing?

tomflexcompute (approved):

PR #1185, opened 2023-09-29
merge current state of develop (2.4.2) into pre/2.5

tylerflex:
fixes some very minor merge conflicts.


momchil-flex:
I think instead of reviewing all the changed files it would be best for @dbochkov to just know what was changed when conflicts were resolved... You could try the merge locally to check that out I guess. And then the next test would be rebasing the heat on top of this and checking that everything works still.

tylerflex:
Merge conflicts were only in the webapi imports and version.py. But yea feel free to use this as needed or close

dbochkov-flexcompute:
updated this to the most recent pre/2.5.0 and develop. Indeed, all conflicts were trivial. I think we can merge this

PR #1187, opened 2023-09-29
better description for TimeMonitor.interval

tylerflex:
Requested by https://flow360.atlassian.net/browse/SCEM-2172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Warnings for too many frequencies in monitors; too many modes requested in a ``ModeSpec``; too many number of grid points in a mode monitor or mode source.
 
 ### Changed
+- Time domain monitors warn about data usage if all defaults used in time sampler specs.
 
 ### Fixed
 - Faster sorting of modes in `ModeSolverData.overlap_sort` by avoiding excessive data copying.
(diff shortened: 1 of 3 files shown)


momchil-flex (approved):
This looks good, but while we're at it I wonder if we should change the default. People tend to have very big time monitors without realizing... But I'm not sure what's a good default setting. One thing could be the nyquist sampling rate used in the frequency monitors, which is on the order of 10-20 typically. But I don't think we want to not have a clear default (i.e. to be automatically reset by the simulation). Or do we? Or can we change defaults in some other way, e.g. not have start and stop time be beginning and end of the full simulation?

tylerflex:
I don't think we should change defaults just because it will mess with people's existing scripts. 

One thing I could imagine though is setting the default value to `None` and warning in a validator that:
* we will set `interval` will be `1` by default.
* this can lead to a lot of data.
* if you really wanted this, to get rid of this warning explicitly set `interval=1` in your monitor.

momchil-flex:
I think that sounds good! You wanna do this as part of this PR?

tylerflex:
ok. what do you recommend for a good recommended `interval` value?

momchil-flex:
I feel like maybe we shouldn't recommend a value; just print the warning if `interval=None` and also `start=0` and `stop=None` (the defaults). Just say that this may produce a lot of data, and consider using these arguments to reduce that.

PR #1188, opened 2023-09-29
Space-time modulation of all types of materials

weiliangjin2021:
TODO:

- [x] classes related to space-time modulation
- [x] modulation_spec to base material class
- [x] Validate several types of materials that are yet to support time modulation
- [x] Adjust CFL number etc. when permittivity is modulation

This PR adds two fields ~~`permittivity_modulation` and `conductivity_modulation`~~ `modulation_spec` to `Medium`. 

With modulation, the permittivity (conductivity) of a medium is a linear summation of time-independent and dynamic parts. As for the dynamic part, it can be a function of both space and time: `modulation_amp(r, t)`. Similar to source, i think we can just support the case where space and time are separable: `modulation_amp(r, t) = Re[amp_time(t) amp_space(r)]`. This should cover most applications, and is more efficient to support in the backend.

- `modulation_amp(r, t)` is described by the class `SpaceTimeModulation` that has two fields: `space_modulation` and `time_modulation`. The default `space_modulation` is uniform in space 
- `SpaceModulation` class defines `amp_space(r) = amplitude(r) \\cdot \\e^{i \\cdot phase(r)}`, where amplitude and phase can be either a float, or custom data array in space
- Similarly, there is a class for time modulation. Right now, we only have `ContinuousWaveTimeModulation`. In the future, we can support custom time modulation.

Back to the medium, because of some considerations in the backend, when both permittivity and conductivity are modulated, currently we restrict that their "time modulation" must be identical.

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -27,7 +27,7 @@ from ..log import log
 from .transformation import RotationType
 from .parameter_perturbation import ParameterPerturbation
 from .heat_spec import HeatSpecType
-
+from .time_modulation import ModulationSpec
 
 # evaluate frequency as this number (Hz) if inf
 FREQ_EVAL_INF = 1e50
@@ -167,7 +167,7 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         False,
         title="Allow gain medium",
         description="Allow the medium to be active. Caution: "
-        "simulations with gain medium are unstable, and are likely to diverge."
+        "simulations with a gain medium are unstable, and are likely to diverge."
         "Simulations where 'allow_gain' is set to 'True' will still be charged even if "
         "diverged. Monitor data up to the divergence point will still be returned and can be "
         "useful in some cases.",
@@ -177,6 +177,12 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         None,
         title="Nonlinear Spec",
         description="Nonlinear spec applied on top of the base medium properties.",
+    )
+
+    modulation_spec: ModulationSpec = pd.Field(
+        None,
+        title="Modulation Spec",
+        description="Modulation spec applied on top of the base medium properties.",
     )
 
     @pd.validator("nonlinear_spec", always=True)
@@ -203,7 +209,24 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         discriminator=TYPE_TAG_STR,
     )
 
+    @pd.validator("modulation_spec", always=True)
+    def _validate_modulation_spec(cls, val, values):
+        """Check compatibility with modulation_spec."""
+        nonlinear_spec = values.get("nonlinear_spec")
+        if val is not None and nonlinear_spec is not None:
+            raise ValidationError(
+                f"For medium class {cls}, 'modulation_spec' of class {type(val)} and "
+                f"'nonlinear_spec' of class {type(nonlinear_spec)} are "
+                "not simultaneously supported."
+            )
+        return val
+
     _name_validator = validate_name_str()
+
+    @cached_property
+    def time_modulated(self) -> bool:
+        """Whether any component of the medium is time modulated."""
+        return self.modulation_spec is not None and self.modulation_spec.applied_modulation
 
     @abstractmethod
     def eps_model(self, frequency: float) -> complex:
@@ -338,7 +361,7 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         Returns
         -------
         complex
-            Complex-valued relative permittivty.
+            Complex-valued relative permittivity.
         """
         eps_real = n**2 - k**2
         eps_imag = 2 * n * k
@@ -609,6 +632,16 @@ class PECMedium(AbstractMedium):
     To avoid confusion from duplicate PECs, must import ``tidy3d.PEC`` instance directly.
     """
 
+    @pd.validator("modulation_spec", always=True)
+    def _validate_modulation_spec(cls, val):
+        """Check compatibility with modulation_spec."""
+        if val is not None:
+            raise ValidationError(
+                f"A 'modulation_spec' of class {type(val)} is not "
+                f"currently supported for medium class {cls}."
+            )
+        return val
+
     def eps_model(self, frequency: float) -> complex:
         # return something like frequency with value of pec_val + 0j
         return 0j * frequency + pec_val
@@ -663,8 +696,40 @@ class Medium(AbstractMedium):
         if not values.get("allow_gain") and val < 0:
             raise ValidationError(
                 "For passive medium, 'conductivity' must be non-negative. "
-                "To simulate gain medium, please set 'allow_gain=True'. "
-                "Caution: simulations with gain medium are unstable, and are likely to diverge."
+                "To simulate a gain medium, please set 'allow_gain=True'. "
+                "Caution: simulations with a gain medium are unstable, and are likely to diverge."
+            )
+        return val
+
+    @pd.validator("permittivity", always=True)
+    def _permittivity_modulation_validation(cls, val, values):
+        """Assert modulated permittivity cannot be <= 0."""
+        modulation = values.get("modulation_spec")
+        if modulation is None or modulation.permittivity is None:
+            return val
+
+        min_eps_inf = val if np.ndim(val) == 0 else np.min(np.array(val))
+        if min_eps_inf - modulation.permittivity.max_modulation <= 0:
+            raise ValidationError(
+                "The minimum permittivity value with modulation applied was found to be negative."
+            )
+        return val
+
+    @pd.validator("conductivity", always=True)
+    def _passivity_modulation_validation(cls, val, values):
+        """Assert passive medium if `allow_gain` is False."""
+        modulation = values.get("modulation_spec")
+        if modulation is None or modulation.conductivity is None:
+            return val
+
+        min_sigma = val if np.ndim(val) == 0 else np.min(np.array(val))
+        if not values.get("allow_gain") and min_sigma - modulation.conductivity.max_modulation < 0:
+            raise ValidationError(
+                "For passive medium, 'conductivity' must be non-negative at any time."
+                "With conductivity modulation, this medium can sometimes be active. "
+                "Please set 'allow_gain=True'. "
+                "Caution: simulations with a gain medium are unstable, "
+                "and are likely to diverge."
             )
         return val
 
@@ -676,7 +741,11 @@ class Medium(AbstractMedium):
 
         For dispersiveless medium, it equals ``sqrt(permittivity)``.
         """
-        return np.sqrt(self.permittivity)
+        permittivity = self.permittivity
+        if self.modulation_spec is not None and self.modulation_spec.permittivity is not None:
+            permittivity -= self.modulation_spec.permittivity.max_modulation
+        n, _ = self.eps_complex_to_nk(permittivity)
+        return n
 
     @ensure_freq_in_range
     def eps_model(self, frequency: float) -> complex:
@@ -787,8 +856,8 @@ class CustomIsotropicMedium(AbstractCustomMedium, Medium):
         if not values.get("allow_gain") and np.any(val.values < 0):
             raise ValidationError(
                 "For passive medium, 'conductivity' must be non-negative. "
-                "To simulate gain medium, please set 'allow_gain=True'. "
-                "Caution: simulations with gain medium are unstable, and are likely to diverge."
+                "To simulate a gain medium, please set 'allow_gain=True'. "
+                "Caution: simulations with a gain medium are unstable, and are likely to diverge."
             )
         return val
 
@@ -800,7 +869,11 @@ class CustomIsotropicMedium(AbstractCustomMedium, Medium):
 
         For dispersiveless medium, it equals ``sqrt(permittivity)``.
         """
-        return np.sqrt(np.min(self.permittivity.values))
+        permittivity = np.min(self.permittivity.values)
+        if self.modulation_spec is not None and self.modulation_spec.permittivity is not None:
+            permittivity -= self.modulation_spec.permittivity.max_modulation
+        n, _ = self.eps_complex_to_nk(permittivity)
+        return n
 
     @cached_property
     def is_isotropic(self):
@@ -845,6 +918,14 @@ class CustomMedium(AbstractCustomMedium):
     >>> eps = dielectric.eps_model(200e12)
     """
 
+    eps_dataset: Optional[PermittivityDataset] = pd.Field(
+        None,
+        title="Permittivity Dataset",
+        description="[To be deprecated] User-supplied dataset containing complex-valued "
+        "permittivity as a function of space. Permittivity distribution over the Yee-grid "
+        "will be interpolated based on ``interp_method``.",
+    )
+
     permittivity: Optional[SpatialDataArray] = pd.Field(
         None,
         title="Permittivity",
@@ -859,14 +940,6 @@ class CustomMedium(AbstractCustomMedium):
         "that the imaginary part of the complex permittivity at angular "
         "frequency omega is given by conductivity/omega.",
         units=CONDUCTIVITY,
-    )
-
-    eps_dataset: Optional[PermittivityDataset] = pd.Field(
-        None,
-        title="Permittivity Dataset",
-        description="[To be deprecated] User-supplied dataset containing complex-valued "
-        "permittivity as a function of space. Permittivity distribution over the Yee-grid "
-        "will be interpolated based on ``interp_method``.",
     )
 
     @pd.root_validator(pre=True)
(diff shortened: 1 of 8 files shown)


tidy3d/components/source.py
tylerflex, line 1: we might want to split the abstract time_dependence into its own file. I think there could be a circular dependency issue lurking by mixing the abstract time dependence / time modulation with the Source in this way. 

let's either make a new file to store the abstract time dependence, and maybe a directory or two to hold this + source time + time modulation? if it makes sense to you

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+
+from .base import Tidy3dBaseModel, cached_property
+from .types import InterpMethod
+from .source import AbstractTimeDependence
+from .data.data_array import SpatialDataArray
+from ..exceptions import ValidationError
+from ..constants import HERTZ
+
+
+class AbstractTimeModulation(AbstractTimeDependence):
```
tylerflex, line 19: inherit this from `ABC` (and other abstract ones, eg. AbstractTimeDependence, if not already)

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+    """
+
+    @cached_property
+    @abstractmethod
+    def max_modulation(self) -> float:
+        """Estimated maximal modulation amplitude."""
+
+
+class ContinuousWaveTimeModulation(AbstractTimeDependence):
+    """The modulation in time is in continuous oscillation.
```
tylerflex, line 39: Class describing modulation with a harmonic time dependence? or something like this maybe? or is "Continuous oscillation" well understood?

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+
+    """
+
+    @cached_property
+    @abstractmethod
+    def max_modulation(self) -> float:
+        """Estimated maximal modulation amplitude."""
+
+
+class ContinuousWaveTimeModulation(AbstractTimeDependence):
```
tylerflex, line 38: do we want to unify the source times with these? or not? because this looks like a CW source time. We could also have a `GaussianTimeModulation` as well. If you want to keep separate too, that's reasonable.

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+
+    """
+
+    @cached_property
+    @abstractmethod
+    def max_modulation(self) -> float:
+        """Estimated maximal modulation amplitude."""
+
+
+class SpaceModulation(AbstractSpaceModulation):
```
tylerflex, line 98: tempted to call this `SpatialModulation` for consistency with `SpatialDataArray`. Not sure, any thoughts?

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+    """The modulation profile in space with user-supplied spatial distribution of
```
tylerflex, line 99: "as a function of space" "with a user-supplied"

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+    >>> coords = dict(x=X, y=Y, z=Z)
+    >>> amp= SpatialDataArray(np.random.random((Nx, Ny, Nz)), coords=coords)
+    >>> phase= SpatialDataArray(np.random.random((Nx, Ny, Nz)), coords=coords)
+    >>> space = SpaceModulation(amplitude=amp, phase=phase)
+    """
+
+    amplitude: Union[float, SpatialDataArray] = pd.Field(
+        1,
+        title="Amplitude of modulation in space",
+        description="Amplitude of modulation that can vary spatially.",
```
tylerflex, line 128: the units are the units of whatever is being modulated? might be good to elaborate here.

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+    )
+
+    phase: Union[float, SpatialDataArray] = pd.Field(
+        0,
+        title="Phase of modulation in space",
+        description="Phase of modulation that can vary spatially.",
```
tylerflex, line 134: might be good to specify units of the phase (radians?) or at least whether this should be 2pi for a full cycle or 1.

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+    )
+
+    interp_method: InterpMethod = pd.Field(
+        "nearest",
+        title="Interpolation method",
+        description="Interpolation method to obtain space modulation values "
```
tylerflex, line 140: "Method of interpolation to use to obtain values at spatial locations other than those on the Yee grids" maybe

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+    def _real_phase(cls, val):
+        """Assert that the phase is real."""
+        if np.iscomplexobj(val):
+            raise ValidationError("'phase' must be real.")
+        return val
+
+    @cached_property
+    def max_modulation(self) -> float:
+        """Estimated maximal modulation amplitude."""
+        return np.max(np.array(self.amplitude))
```
tylerflex, line 161: what if amplitude is negative? might want to `abs` this.

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+    .. math::
+
+        \\delta \\epsilon(r, t) = \\Re[amp\\_time(t) \\cdot amp\\_space(r)]
+    """
+
+    space_modulation: SpaceModulationType = pd.Field(
+        SpaceModulation(),
+        title="Space modulation",
+        description="Space modulation from the separable space-time modulation.",
+        # discriminator=TYPE_TAG_STR,
```
tylerflex, line 186: probably best to have these for now, unless it causes errors? (just for future)

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+
+    .. math::
+
+        \\delta \\epsilon(r, t) = \\Re[amp\\_time(t) \\cdot amp\\_space(r)]
+    """
+
+    space_modulation: SpaceModulationType = pd.Field(
+        SpaceModulation(),
+        title="Space modulation",
+        description="Space modulation from the separable space-time modulation.",
```
tylerflex, line 185: Spatial dependence of a separable 'SpaceTimeModulation'.

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,207 @@
...
+        # discriminator=TYPE_TAG_STR,
+    )
+
+    time_modulation: TimeModulationType = pd.Field(
+        ...,
+        title="Space modulation",
```
tylerflex, line 191: these should be "time" and same comments as above.

tylerflex (changes requested):
Thanks @weiliangjin2021 

I think the way you have organized the spatial and time dependence seems good. 

However, off the top of my head, I think it might be nicer if we can consolidate the modulation into a single `ModulationSpec`, perhaps with `permittivity` and `conductivity` fields. I worry a little about hardcoding `permittivity_perturbation` and `conductivity_perturbation` into the `Medium` directly. Part of me feels like if we contain it in a spec, it an make it easier to change in the future as we add more features, but it's not a very specific idea at the moment. Do you or @momchil-flex  have any thoughts?

Maybe to get that discussion started: do we have (even very future) plans to modulate any other fields in the medium?  or other quantities?  For example, the individual components of anisotropic medium, geometric values, or things like this? If so, we might want to make this even more general, similar to the heat solver, so that we can attach a `SpaceTimeModulation` to an arbitrary field, or attach a `ModulationSpec` to some geometry, for example.

tidy3d/components/time_modulation.py
weiliangjin2021: Yeah, it errors when there is only one in union

tidy3d/components/time_modulation.py
weiliangjin2021: The interpolation is to obtain values on Yee grids.

tidy3d/components/time_modulation.py
weiliangjin2021: But we also have SpaceTimeModulation. SpaceModulation is more consistent with that?

tidy3d/components/time_modulation.py
tylerflex: oh ok, good to know.

tidy3d/components/time_modulation.py
tylerflex:
```
"Method of interpolation to use to obtain values at spatial locations on the Yee grids."
```

tidy3d/components/time_modulation.py
tylerflex: yea that's true, we can leave it.

weiliangjin2021:
> Maybe to get that discussion started: do we have (even very future) plans to modulate any other fields in the medium? or other quantities? For example, the individual components of anisotropic medium, geometric values, or things like this? If so, we might want to make this even more general, similar to the heat solver, so that we can attach a `SpaceTimeModulation` to an arbitrary field, or attach a `ModulationSpec` to some geometry, for example.

Modulating geometry or other parameters in the medium such as pole residue parameters are challenging to include. Individual components of anisotropic medium is enabled already, as the components can be modulated medium types. So I won't expect supporting other fields in the near future, unless strongly requested.

> However, off the top of my head, I think it might be nicer if we can consolidate the modulation into a single `ModulationSpec`, perhaps with `permittivity` and `conductivity` fields. I worry a little about hardcoding `permittivity_perturbation` and `conductivity_perturbation` into the `Medium` directly. Part of me feels like if we contain it in a spec, it an make it easier to change in the future as we add more features, but it's not a very specific idea at the moment. Do you or @momchil-flex have any thoughts?

I think either is fine. The pro of the current design is to make it obvious what one can modulate: only eps_inf and conductivity.

tidy3d/components/source.py
weiliangjin2021: It's now splitted into `time.py`

tidy3d/components/time_modulation.py
weiliangjin2021: They are slightly different: CW modulation doesn't have 'fwidth' and 'offset'. So probably better to be separate?

tidy3d/components/time_modulation.py
weiliangjin2021: Added: "It take the unit of whatever is being modulated."

tylerflex: sounds good, should be "takes the units" I think

tidy3d/components/time_modulation.py
tylerflex: ok makes sense.

tylerflex:
> I think either is fine. The pro of the current design is to make it obvious what one can modulate: only eps_inf and conductivity.

I think the main problem is that it makes it hard to generalize to other fields if we decide to later. We ran into a similar issue with the `medium.py` things. (for example, putting `chi3` in the regular `Medium` as a field vs a `NonlinearSpec`. I think the spec approach is probably better as we add new nonlinearities.

weiliangjin2021:
> I think the main problem is that it makes it hard to generalize to other fields if we decide to later. We ran into a similar issue with the `medium.py` things. (for example, putting `chi3` in the regular `Medium` as a field vs a `NonlinearSpec`. I think the spec approach is probably better as we add new nonlinearities.

Yeah, that makes sense. I did a preliminary reorganization to consolidate them into a spec now.

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,244 @@
...
+class ModulationSpec(Tidy3dBaseModel):
+    """Specification adding space-time modulation to the non-dispersive part of medium
+    including epsilon at infinity and electric conductivity.
+    """
+
+    permittivity: SpaceTimeModulation = pd.Field(
+        None,
+        title="Space-time modulation of permittivity",
+        description="Space-time modulation of permittivity at infinity frequency "
+        "applied on top of the base permittivity at infinity frequency.",
```
momchil-flex, line 220: infinity -> infinite

tidy3d/components/medium.py
```diff
@@ -663,8 +696,36 @@ def _passivity_validation(cls, val, values):
...
-                "To simulate gain medium, please set 'allow_gain=True'. "
-                "Caution: simulations with gain medium are unstable, and are likely to diverge."
+                "To simulate a gain medium, please set 'allow_gain=True'. "
+                "Caution: simulations with a gain medium are unstable, and are likely to diverge."
+            )
+        return val
+
+    @pd.validator("permittivity", always=True)
+    def _permittivity_modulation_validation(cls, val, values):
+        """Assert modulated permittivity cannot be <= 0."""
```
momchil-flex, line 706: I assume your to-do item regarding fixing the CFL number is for the cases when 0 < permittivity < 1. Practically, this sounds like some very esoteric material, I'm almost tempted to say we just validate it out right now, but it may be of interest to e.g. some academic users, so if it's not too hard to do the CFL modification then I guess it would be good.

momchil-flex (approved):
Looks great to me, just very minor comments.

tidy3d/components/medium.py
```diff
@@ -203,8 +209,25 @@ def _validate_nonlinear_spec(cls, val):
...
+                f"'nonlinear_spec' of class {type(nonlinear_spec)} are "
+                "not simultaneously supported."
+            )
+        return val
+
     _name_validator = validate_name_str()
 
+    @cached_property
+    def time_modulated(self) -> bool:
+        """Any component of the medium is time modulated."""
```
tylerflex, line 228: "Whether any component ..." (and fix in other places)

tidy3d/components/medium.py
```diff
@@ -2735,6 +2811,14 @@ def components(self) -> Dict[str, Medium]:
         """Dictionary of diagonal medium components."""
         return dict(xx=self.xx, yy=self.yy, zz=self.zz)
 
+    @cached_property
+    def time_modulated(self) -> bool:
+        """Any component of the medium is time modulated."""
+        for mat in self.components.values():
+            if mat.time_modulated:
```
tylerflex, line 2818: I think just
```py
return any(mat.time_modulated for mat in self.components.values())
```
could work too.

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def amp_time(self, time: float) -> complex:
+        """Complex-valued source amplitude as a function of time."""
+
+        omega = 2 * np.pi * self.freq0
+        return self.amplitude * np.exp(-1j * omega * time + 1j * self.phase)
+
+    @cached_property
+    def max_modulation(self) -> float:
+        """Estimated maximal modulation amplitude."""
```
tylerflex, line 73: "maximum" probably sounds more natural.

tidy3d/components/time_modulation.py
```diff
@@ -0,0 +1,244 @@
...
+        """whether the modulation is weak enough to be regarded as zero."""
+        # if isclose(np.diff(time_modulation.range), 0) and
+        if isclose(self.max_modulation, 0):
+            return True
+        return False
+
+
+class ModulationSpec(Tidy3dBaseModel):
+    """Specification adding space-time modulation to the non-dispersive part of medium
+    including epsilon at infinity and electric conductivity.
```
tylerflex, line 213: I think here and in the permittivity, we should be more clear that this is relative permittivity.  Also, maybe "epsilon at infinity" is too vague (what is epsilon? what=infinity?), maybe for this spec we can just refer to it as `relative permittivity at infinite frequency` if that is the most specific?

tidy3d/components/time_modulation.py
tylerflex: Adding to this, think we should be clear that this is relative permittivity.

tylerflex (approved):
Looks good to me, minus a few rewording comments and minor optional code reorganization.

The ModulationSpec I think looks good, this hopefully will future-proof us if we make more complicated modulations, thanks for adding it.

tidy3d/components/medium.py
weiliangjin2021: In lots of cases, it's probably 0<permittivity<1 when a dispersive medium of eps_inf=1 is modulated. But the modulation is in general very weak in EO, so the modulated permittivity is very close to one. It's easy to fix CFL number.

weiliangjin2021: Added now, together with a bunch of validators.

tidy3d/components/medium.py
```diff
@@ -976,17 +1050,40 @@ def _eps_dataset_eps_inf_greater_no_less_than_one_sigma_positive(cls, val, value
                     "Permittivity at infinite frequency at any spatial point "
                     "must be no less than one."
                 )
+
+            if modulation is not None and modulation.permittivity is not None:
+                if np.any(eps_real.values - modulation.permittivity.max_modulation <= 0):
```
dbochkov-flexcompute, line 1055: just to double check my understanding is correct: in principle this could be relaxed a bit by computing max modulation at each spatial point? But this would require either checking that eps_real/sigma and space modulation are provided on the same grid, or performing interpolation, and probably not worth doing anyway

weiliangjin2021: Yeah, probably slow as well.

PR #1190, opened 2023-10-02
2.5 changes

tylerflex:
Just a few things we were going to change for 2.5, before I forget. Might be more, so we can keep this open and add to it as needed or merge whenever.

- Python 3.7 no longer guaranteed to be supported.
- Remove colocation deprecation warning.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -15,11 +15,8 @@ jobs:
     runs-on: ${{ matrix.platform }}
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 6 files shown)


tylerflex:
@momchil-flex we should probably get this in sooner rather than later. Just want to confirm that we want to remove this colocation warning and also that this PR handles it the right way.

momchil-flex (approved):
Looks good, what do you think about officially requiring python > 3.7 since we won't be testing?

tylerflex:
yea that's a good point, I just pushed a separate commit (probably fine for them to be separate) to restrict this, and updated changelog.

tylerflex:
note: I think tests fail because I guess it's using the `pre/2.5` `tox.ini` (which includes python3.7) instead of the one from this PR.

tylerflex:
actually it was because I forgot to exclude from the workflow, should be fixed.

PR #1179, opened 2023-09-26
broadband adjoint

tylerflex:
Things to do still:
- [ ] Suppress warning about individual adjoint source spectra not covering all monitor frequencies (this is by design).
- [ ] Fix `RuntimeWarning: invalid value encountered` in `interp` in the WDM notebook.
- [ ] Fix up the WDM example text to be consistent with it using multi-frequency adjoint.
- [ ] Test multi-frequency in the other notebooks to ensure they work properly.

Docs PR: [link](https://github.com/flexcompute-readthedocs/tidy3d-docs/compare/tyler/adjoint/broadband2?expand=1)

PR #1192, opened 2023-10-04
broadband adjoint (2.5)

tylerflex:
same as #1179 but rebased against 2.5, squashed, and polished up.

Note: works fine on version 2.4.2 but fails with version 2.5.0rc1 for some reason, looking into this.

Also note: there are two added tests (`_test_adjoint_broadband ...`) that should go in core.

Summary of changes:
* users can now add multiple frequencies to `output_monitors` and also different `output_monitors` can have different frequencies.
* internally, we maintain a sorted list of `freqs_adjoint` corresponding to these monitors, which become the `freq0`s of our adjoint sources.
* if not set by `JaxSimulation.fwidth_adjoint`, the `fwidth` of the adjoint sources is chosen to be smaller than some factor (0.1) times the minimum difference between the `freqs_adjoint`. This was chosen empirically to be small enough to have low error, but not be too small. When this is set, a warning is displayed to the user outlining that an `fwidth` was chosen automatically, which could lead to large `run_time` and some mitigation approaches for tightly packed frequencies.
* if not set by a new field `JaxSimulation.run_time_adjoint`, the `run_time` of the adjoint `JaxSimulation`, will be some factor (100) divided by the adjoint `fwidth`.  Again chosen empirically to avoid early shutoff in the adjoint sims, while not being too long.
* The adjoint E fields are left unnormalized initially, but then are retroactively normalized by the adjoint source `source_time` spectrum at `fwidths_adjoint`. This is because we need to apply a different normalization frequency-by-frequency depending on the `freq0` and `fwidth` of the adjoint sources, which may introduce amplitude or phase differences.
* All of the VJP calculations are vectorized over frequency. At the end of the calculations, we sum the results over the frequency axis (consistent with derivative rule: df(x,y)/dz = df/dx + dy/dy, forget the name of it.

Corresponding notebook changes: https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/296

I think that covers it, thanks in advance for the review!

### DIFF
--- a/tidy3d/components/source.py
+++ b/tidy3d/components/source.py
@@ -98,6 +98,10 @@ class SourceTime(ABC, Tidy3dBaseModel):
         if not complex_fields:
             time_amps = np.real(time_amps)
 
+        # if all time amplitudes are zero, just return (complex-valued) zeros for spectrum
+        if np.allclose(time_amps, 0.0):
+            return (0.0 + 0.0j) * np.zeros_like(freqs)
+
         # Cut to only relevant times
         relevant_time_inds = np.where(np.abs(time_amps) / np.amax(np.abs(time_amps)) > DFT_CUTOFF)
         # find first and last index where the filter is True
--- a/tidy3d/plugins/adjoint/components/simulation.py
+++ b/tidy3d/plugins/adjoint/components/simulation.py
@@ -17,7 +17,7 @@ from ....components.simulation import Simulation
 from ....components.data.monitor_data import FieldData, PermittivityData
 from ....components.structure import Structure
 from ....components.types import Ax, annotate_type
-from ....constants import HERTZ
+from ....constants import HERTZ, SECOND
 from ....exceptions import AdjointError
 
 from .base import JaxObject
@@ -25,8 +25,14 @@ from .structure import JaxStructure
 from .geometry import JaxPolySlab, JaxGeometryGroup
 
 
-# bandwidth of adjoint source in units of freq0 if no sources and no `fwidth_adjoint` specified
+# bandwidth of adjoint source in units of freq0 if no `fwidth_adjoint`, and one output freq
 FWIDTH_FACTOR = 1.0 / 10
+
+# bandwidth of adjoint sources in units of the minimum difference between output frequencies
+FWIDTH_FACTOR_MULTIFREQ = 0.1
+
+# the adjoint run time is RUN_TIME_FACTOR / fwidth
+RUN_TIME_FACTOR = 100
 
 # how many processors to use for server and client side adjoint
 NUM_PROC_LOCAL = 1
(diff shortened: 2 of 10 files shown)


tidy3d/plugins/adjoint/components/simulation.py
```diff
@@ -227,18 +225,25 @@ def _warn_if_colocate(cls, val):
...
 
-        return output_monitors[0].freqs[0]
+        freqs = []
+        for mnt in output_monitors:
+            for freq in mnt.freqs:
+                if freq not in freqs:
+                    freqs.append(freq)
+
+        freqs.sort()
+        return freqs
```
weiliangjin2021, line 241: probably fewer lines of code by using `numpy.unique`?

weiliangjin2021 (approved):

tidy3d/plugins/adjoint/components/simulation.py
tylerflex: thanks, made this change.

tidy3d/components/source.py
```diff
@@ -105,6 +105,8 @@ def spectrum(
         stop_ind = relevant_time_inds[0][-1]
         time_amps = time_amps[start_ind:stop_ind]
         times_cut = times[start_ind:stop_ind]
+        # TODO: this fails if tim_amps is all 0 (for example if amplitude=0.0)
+        # maybe if this case is detected, we automatically return `np.zeros_like(freqs)`?
```
momchil-flex, line 109: Sounds like a good solution, just add it here so it doesn't fall through the cracks?

tidy3d/plugins/adjoint/web.py
```diff
@@ -691,6 +694,9 @@ def run_local_bwd(
         callback_url=callback_url,
         verbose=verbose,
     )
+
+    sim_data_adj = sim_data_adj.normalize_adjoint_fields()
```
momchil-flex, line 698: To fix the backend tests, I had to add this to the server-side adjoint handling too. Basically any time you add something to the internals in `run_local`, we need to make sure it is also incorporated in the server-side run, so let me know.

I've fixed the backend tests in a commit on your backend branch.

momchil-flex (approved):

tidy3d/components/source.py
tylerflex: ok added

tidy3d/plugins/adjoint/web.py
tylerflex: I see, thanks.

tylerflex:
ok I just added the source thing, fixed up changelog, squashed, etc. I think this should be ready to go assuming tests all pass.

Could I leave it to you to merge the core PR? maybe with this updated submodule commit?

momchil-flex:
Backend tests pass, merging everything!

PR #1193, opened 2023-10-06
Added PECMedium to IsotropicUniformMediumType, and added PEC2D

caseyflex:


tylerflex (approved):
seems ok to me, do we need to add a changelog item in "fixed"?

Asking @shash-sharma for a review too since I'm not totally following the original problem this is fixing.

tidy3d/components/grid/mesher.py
```diff
@@ -373,7 +373,7 @@ def structure_steps(
         min_steps = []
         for structure in structures:
             if isinstance(structure, Structure):
-                if isinstance(structure.medium, (PECMedium, Medium2D)):
+                if structure.medium.is_pec or isinstance(structure.medium, Medium2D):
```
weiliangjin2021, line 376: So this use the PEC meshing scheme even if the anisotropic is not PEC along some components?

caseyflex: good catch, I guess it should use the relevant component for meshing. Although, is this a realistic use case? one component PEC, one component not

weiliangjin2021: Maybe for some hyperbolic material that is metallic along some component and dielectric along the rest, and one models the metallic part as PEC.

momchil-flex:
Sorry, what is the status of this? @shash-sharma ?

shashwat-sh:
I think we're waiting on PEC subpixel, and some revisions on this and the associated backend PR

momchil-flex:
PEC subpixel will almost certainly not be in 2.5 - are you saying this and the backend PRs should then also not be labeled as 2.5?

caseyflex:
Weiliang had a proposal on the backend PR that could be implemented before
PEC subpixel as far as I am aware. I haven’t looked into implementing it
yet.

caseyflex:
The issue is wanting a unified treatment of PEC across all three cases:
ordinary, anisotropic, and 2D.

caseyflex:
@weiliangjin2021 do you want to change the meshing to use the relevant component even when another component is PEC?

weiliangjin2021:
> @weiliangjin2021 do you want to change the meshing to use the relevant component even when another component is PEC?

Could you edit `is_comp_pec` method in Medium2D, and push the changes directly? @caseyflex

weiliangjin2021:
One caveat in scene visualization is that only an isotropic PEC will be plotted as PEC. An anisotropic medium where some components are PEC is plotted as regular medium. Do you think it's fine @tylerflex ?

caseyflex:
> > @weiliangjin2021 do you want to change the meshing to use the relevant component even when another component is PEC?
> 
> Could you edit `is_comp_pec` method in Medium2D, and push the changes directly? @caseyflex

It seems like `is_comp_pec` should only be a method in AnisotropicMedium, since it doesn't make sense for an arbitrary AbstractMedium. Then, we can have `is_comp_pec` in Medium2D as well, but in addition to taking the component axis, it can take the normal axis, which is known in the mesher. This would be required to map ss, tt to xx, yy, zz.

shashwat-sh (approved):
This makes sense to me. About the plotting, I think the current behaviour of only plotting isotropic PEC as PEC makes sense, since the anisotropic case with some PEC components seems to be a fairly special case, and probably used in a very different context.

caseyflex:
@weiliangjin2021 , I pushed the changes we discussed. Also, I changed it to plot as PEC if any component is PEC, rather than if all components are PEC. Please check if this new handling makes sense.

tidy3d/components/grid/mesher.py
```diff
@@ -373,7 +373,13 @@ def structure_steps(
...
         for structure in structures:
             if isinstance(structure, Structure):
-                if isinstance(structure.medium, (PECMedium, Medium2D)):
+                if (
+                    isinstance(structure.medium, AnisotropicMedium)
+                    and structure.medium.is_comp_pec(axis)
+                ) or isinstance(structure.medium, Medium2D):
+                    # for 2d medium, will always ignore even if not PEC;
+                    # later, this will be handled by _grid_corrections_2dmaterials
+                    # in simulation.py
```
weiliangjin2021, line 382: Looks a missing case when the medium is PECMedium? (Before this PR, it was `if isinstance(structure.medium, (PECMedium, Medium2D)):`)

weiliangjin2021 (approved):
One last comment. Otherwise all look good to me.

tidy3d/components/grid/mesher.py
caseyflex: good catch, I guess I deleted this by accident.

caseyflex:
> One last comment. Otherwise all look good to me.

thanks for the catch. This should be ready, maybe we want to squash the commits though.

weiliangjin2021:
> > One last comment. Otherwise all look good to me.
> 
> thanks for the catch. This should be ready, maybe we want to squash the commits though.

Thanks, I'll take care of that. Usually we squash just before the merge so that changes can be better monitored.

momchil-flex:
So is there anything left to do @weiliangjin2021 ?

weiliangjin2021:
> So is there anything left to do @weiliangjin2021 ?

just start to squash and merge.

PR #1199, opened 2023-10-10
design space exploration plugin

tylerflex:
For a detailed description, see
https://github.com/flexcompute/tidy3d/blob/0297433e864125e0fcfa02af54b488212b032a8e/CHANGELOG.md

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- `ModeData.dispersion` and `ModeSolverData.dispersion` are calculated together with the group index.
 - `tidy3d.plugins.design` tool to explore user-defined design spaces.
+- `ModeData.dispersion` and `ModeSolverData.dispersion` are calculated together with the group index.
 
 ### Changed
 
@@ -79,18 +79,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Can pass `kwargs` to `ComponentModeler.plot_sim()` to use in `Simulation.plot()`.
 - Ensure that mode solver fields are returned in single precision if `ModeSolver.ModeSpec.precision == "single"`.
 - If there are no adjoint sources for a simulation involved in an objective function, make a mock source with zero amplitude and warn user.
-
-## [2.5.0rc1] - 2023-10-10
-
-### Added
-- Time zone in webAPI logging output.
-- Class `Scene` consisting of a background medium and structures for easier drafting and visualization of simulation setups as well as transferring such information between different simulations.
-- Solver for thermal simulation (see `HeatSimulation` and related classes).
-- Specification of material thermal properties in medium classes through an optional field `.heat_spec`.
-
-### Changed
-- Internal refactor of Web API functionality.
-- `Geometry.from_gds` doesn't create unnecessary groups of single elements.
 
 ## [2.4.3] - 2023-10-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `ModeData.dispersion` and `ModeSolverData.dispersion` are calculated together with the group index.
+- `tidy3d.plugins.design` tool to explore user-defined design spaces.
+
+### Changed
+
+### Fixed
+
 
 ## [2.5.0] - 2023-12-13
 
@@ -73,6 +79,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Can pass `kwargs` to `ComponentModeler.plot_sim()` to use in `Simulation.plot()`.
 - Ensure that mode solver fields are returned in single precision if `ModeSolver.ModeSpec.precision == "single"`.
 - If there are no adjoint sources for a simulation involved in an objective function, make a mock source with zero amplitude and warn user.
+
+## [2.5.0rc1] - 2023-10-10
+
+### Added
+- Time zone in webAPI logging output.
+- Class `Scene` consisting of a background medium and structures for easier drafting and visualization of simulation setups as well as transferring such information between different simulations.
+- Solver for thermal simulation (see `HeatSimulation` and related classes).
+- Specification of material thermal properties in medium classes through an optional field `.heat_spec`.
+
+### Changed
+- Internal refactor of Web API functionality.
+- `Geometry.from_gds` doesn't create unnecessary groups of single elements.
 
 ## [2.4.3] - 2023-10-16
 
--- a/docs/api/plugins/design.rst
+++ b/docs/api/plugins/design.rst
@@ -1,0 +1,20 @@
+.. currentmodule:: tidy3d
+
+Design Space Exploration
+------------------------
+
+.. autosummary::
+   :toctree: ../_autosummary/
+   :template: module.rst
+
+   tidy3d.plugins.design.Parameter
+   tidy3d.plugins.design.ParameterFloat
+   tidy3d.plugins.design.ParameterInt
+   tidy3d.plugins.design.ParameterAny
+   tidy3d.plugins.design.Method
+   tidy3d.plugins.design.MethodGrid
+   tidy3d.plugins.design.MethodMonteCarlo
+   tidy3d.plugins.design.MethodRandomCustom
+   tidy3d.plugins.design.MethodRandom
+   tidy3d.plugins.design.DesignSpace
+   tidy3d.plugins.design.Results
--- a/docs/api/plugins/index.rst
+++ b/docs/api/plugins/index.rst
@@ -11,6 +11,7 @@ Plugins
     resonance
     adjoint
     waveguide
+    design
 
 
 .. include:: /api/plugins/mode_solver.rst
@@ -20,3 +21,4 @@ Plugins
 .. include:: /api/plugins/resonance.rst
 .. include:: /api/plugins/adjoint.rst
 .. include:: /api/plugins/waveguide.rst
+.. include:: /api/plugins/design.rst
--- a/tests/test_plugins/test_design.py
+++ b/tests/test_plugins/test_design.py
@@ -1,0 +1,305 @@
+"""Test the parameter sweep plugin."""
+import pytest
+import numpy as np
+import tidy3d as td
+import matplotlib.pyplot as plt
+import scipy.stats.qmc as qmc
+
+import tidy3d.web as web
+from tidy3d.components.base import Tidy3dBaseModel
+
+from tidy3d.plugins import design as tdd
+from tidy3d.plugins.design.method import MethodIndependent
+
+from ..utils import run_emulated, log_capture, assert_log_level
+
+
+SWEEP_METHODS = dict(
+    grid=tdd.MethodGrid(),
+    monte_carlo=tdd.MethodMonteCarlo(num_points=5),
+    custom=tdd.MethodRandomCustom(num_points=5, sampler=qmc.Halton(d=3)),
+    random=tdd.MethodRandom(num_points=5),  # TODO: remove this if not used
+)
+
+
+@pytest.mark.parametrize("sweep_method", SWEEP_METHODS.values(), ids=SWEEP_METHODS.keys())
+def test_sweep(sweep_method, monkeypatch, ids=[]):
+    # Problem, simulate scattering cross section of sphere ensemble
+    # 	simulation consists of `num_spheres` spheres of radius `radius`.
+    #   use defines `scs` function to set up and run simulation as function of inputs.
+    #   then postprocesses the data to give the SCS.
+
+    monkeypatch.setattr(web, "run", run_emulated)
+
+    def emulated_batch_run(self, simulations, path_dir: str = None, **kwargs):
+        data_dict = {task_name: run_emulated(sim) for task_name, sim in simulations.items()}
+        task_ids = dict(zip(simulations.keys(), data_dict.keys()))
+
+        class BatchDataEmulated(Tidy3dBaseModel):
+            """Emulated BatchData object that just returns stored emulated data."""
+
+            data_dict: dict
+            task_ids: dict
+
+            def items(self):
+                for task_name, sim_data in self.data_dict.items():
+                    yield task_name, sim_data
+
+            def __getitem__(self, task_name):
+                return self.data_dict[task_name]
+
+        return BatchDataEmulated(data_dict=data_dict, task_ids=task_ids)
+
+    monkeypatch.setattr(MethodIndependent, "_run_batch", emulated_batch_run)
+
+    # STEP1: define your design function (inputs and outputs)
+
+    def scs_pre(radius: float, num_spheres: int, tag: str) -> td.Simulation:
+        """Preprocessing function (make simulation)"""
+
+        # set up simulation
+        spheres = []
+
+        for i in range(int(num_spheres)):
+            spheres.append(
+                td.Structure(
+                    geometry=td.Sphere(radius=radius),
+                    medium=td.PEC,
+                )
+            )
+
+        mnt = td.FieldMonitor(
+            size=(0, 0, 0),
+            center=(0, 0, 0),
+            freqs=[2e14],
+            name="field",
+        )
+
+        return td.Simulation(
+            size=(1, 1, 1),
+            structures=spheres,
+            grid_spec=td.GridSpec.auto(wavelength=1.0),
+            run_time=1e-12,
+            monitors=[mnt],
+        )
+
+    def scs_post(sim_data: td.SimulationData) -> float:
+        """Postprocessing function (analyze simulation data)"""
+
+        mnt_data = sim_data["field"]
+        ex_values = mnt_data.Ex.values
+
+        # generate a random number to add some variance to data
+        np.random.seed(hash(sim_data) % 1000)
+
+        return np.sum(np.square(np.abs(ex_values))) + np.random.random()
+
+    def scs_pre_multi(*args, **kwargs):
+        sim = scs_pre(*args, **kwargs)
+
+        return [sim, sim, sim]
+
+    def scs_post_multi(*sim_datas):
+        vals = [scs_post(sim_data) for sim_data in sim_datas]
+        return np.mean(vals)
+
+    def scs_pre_dict(*args, **kwargs):
+        sims = scs_pre_multi(*args, **kwargs)
+        keys = "abc"
+        return dict(zip(keys, sims))
+
+    def scs_post_dict(a=None, b=None, c=None):
+        sims = [a, b, c]
+        return scs_post_multi(*sims)
+
+    def scs(radius: float, num_spheres: int, tag: str) -> float:
+        """End to end function."""
+
+        sim = scs_pre(radius=radius, num_spheres=num_spheres, tag=tag)
+
+        # run simulation
+        sim_data = run_emulated(sim, task_name=f"SWEEP_{tag}")
+
+        # postprocess
+        return scs_post(sim_data=sim_data)
+
+    # STEP2: define your design problem
+
+    radius_variable = tdd.ParameterFloat(
+        name="radius",
+        span=(0, 1.5),
+        num_points=5,  # note: only used for MethodGrid
+    )
+
+    num_spheres_variable = tdd.ParameterInt(
+        name="num_spheres",
+        span=(0, 3),
+    )
+
+    tag_variable = tdd.ParameterAny(name="tag", allowed_values=("tag1", "tag2", "tag3"))
+
+    design_space = tdd.DesignSpace(
+        parameters=[radius_variable, num_spheres_variable, tag_variable],
+        method=sweep_method,
+        name="sphere CS",
+    )
+
+    # STEP3: Run your design problem
+
+    # either supply generic function and run one by one
+    sweep_results = design_space.run(scs)
+
+    # or supply function factored into pre and post and run in batch
+    sweep_results2 = design_space.run_batch(scs_pre, scs_post)
+
+    sweep_results3 = design_space.run_batch(scs_pre_multi, scs_post_multi)
+
+    sweep_results4 = design_space.run_batch(scs_pre_dict, scs_post_dict)
+
+    sel_kwargs_0 = dict(zip(sweep_results.dims, sweep_results.coords[0]))
+    sweep_results.sel(**sel_kwargs_0)
+
+    print(sweep_results.to_dataframe().head(10))
+
+    im = sweep_results.to_dataframe().plot.hexbin(x="num_spheres", y="radius", C="output")
+    im = sweep_results.to_dataframe().plot.scatter(x="num_spheres", y="radius", c="output")
+    plt.close()
+
+    design_space2 = tdd.DesignSpace(
+        parameters=[radius_variable, num_spheres_variable, tag_variable],
+        method=tdd.MethodMonteCarlo(num_points=3),
+        name="sphere CS",
+    )
+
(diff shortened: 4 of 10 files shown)


tylerflex:
Hey @lucas-flexcompute could you take a look at this prototype for the sweep plugin. Feel free to make even major suggestions to change the API / naming etc. This is mostly a working draft of the tool and there might be a rough edge or two in the API so feedback is welcome.

There's an accompanying tutorial notebook and implementation in another notebook in the [docs PR](https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/326) if you want a higher level picture. I have some changes /polishing to the docs from today that aren't committed yet since I'm on a plane and it won't let me push, but the tutorial notebook there should be all set (in terms of code). I'll let you know when that's committed (in 3 hours or so).

Feel free to take your time on this whenever is convenient. I will probably start working on other things for a bit and come back to it early next week. Thanks!

tylerflex:
notebooks are ready btw, just pushed them.  still think I might make a flow chart to explain some things but might work on that later.

tylerflex:
Here is a diagram that I put into the tutorial to explain the workflow 

image

tidy3d/plugins/sweep/result.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def add(self, fn_args: Dict[str, float], value: Any) -> Result:
+        """Add a specific argument and value the result."""
+
+        new_values = list(self.values) + [value]
+        new_coords = list(self.coords) + [tuple(fn_args[dim] for dim in self.dims)]
+
+        return self.updated_copy(values=new_values, coords=new_coords)
+
+    def hexbin(self, x: str, y: str, C: str, **hexbin_kwargs):
```
lucas-flexcompute, line 305: I'm not sure this is needed. We should support one straightforward plotting method to keep in line with our API and leave the rest for pandas. I worry if users will start asking for other methods from pandas and we'd end up wrapping all of their plotting interfaces.

tidy3d/plugins/sweep/result.py
```diff
@@ -0,0 +1,322 @@
...
+
+        df = self.data_frame
+
+        if "gridsize" not in hexbin_kwargs:
+            hexbin_kwargs["gridsize"] = len(np.unique(df[x]))
+
+        im = df.plot.hexbin(x=x, y=y, C=C, **hexbin_kwargs)
+        return im
+
+    def scatter(self, x: str, y: str, s: str = None, c: str = None, **scatter_kwargs):
```
lucas-flexcompute, line 318: Maybe rename it to `plot` to be consistent with the rest of our API?

tidy3d/plugins/sweep/method.py
```diff
@@ -0,0 +1,212 @@
...
+                task_ids.append([task_id_dict[task_name] for task_name in task_names_i])
+                args_list = (batch_data[task_name] for task_name in task_names_i)
+                val = fn_post(*args_list)
+
+            result.append(val)
+
+        return fn_args, result, task_ids
+
+
+class MethodRandom(MethodIndependent):
```
lucas-flexcompute, line 162: Sampling the variable independently usually leads to very poor MonteCarlo performance, which I guess would be one of the main uses for this plugin. I'd much prefer if we could use something like Latin Hypercube or orthogonal sampling, maybe through (https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.qmc.LatinHypercube.html or some other implementation).

lucas-flexcompute (commented):
General comments:
The figure might indicate that the results influence the sampling method (which would lead to an optimization, not a simple sweep) because of the lines connecting the results to the variables and method. I think I understand that you're aiming to indicate that the method and variables determine the sampling space, but it does look like a closed loop. I took the liberty of making another version just to compare (I've also encapsulated the `run`/`run_batch` processes to indicate that there's a loop/batch inside those calls). Feel free to ignore it, though!

drawing

I tried:
```python
def post(data: td.SimulationData) -> float:
    """Post-processing function, which processes the tidy3d simulation data to return the function output."""
    return {"transmission": np.sum(data['flux'].flux.values)}
```
but it did not work as expected. It used the whole dictionary as value.

tylerflex:
Thanks @lucas-flexcompute for taking a thoughtful look at this.

I think your diagram is more clear, I'll either use that or change my diagram to be laid out more clearly, remove the dashed lines, and make the fns more "contained" so it's easier to understand.

> I tried:
> 
> ```python
> def post(data: td.SimulationData) -> float:
>     """Post-processing function, which processes the tidy3d simulation data to return the function output."""
>     return {"transmission": np.sum(data['flux'].flux.values)}
> ```
> 
> but it did not work as expected. It used the whole dictionary as value.

I also noticed this after running, I fixed it in a commit this morning and will add more tests.

tidy3d/plugins/sweep/design.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # TODO: make sure coords are hashable BEFORE running parameter sweep
+
+    def run(self, function: Callable) -> Result:
+        """Run the design problem on a user defined function of the design variables."""
+
+        # run the function from the method
+        fn_args, fn_values = self.method.run(design_variables=self.design_variables, fn=function)
+
+        fn_source = inspect.getsource(function)
```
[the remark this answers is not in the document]
tylerflex: ok, in the interest of not caching a general Exception, maybe I will catch a `TypeError` and return None like you suggest? or is this not general enough?

tidy3d/plugins/sweep/result.py
```diff
@@ -0,0 +1,322 @@
...
+    @cached_property
+    def _output_names(self) -> Tuple[str]:
+        """Names assigned to the outputs."""
+        if self.is_empty:
+            raise ValueError("Can't get output names as there is no data.")
+        if self.output_names:
+            return self.output_names
+        return tuple(self.value_as_dict(self.values[0]))
+
+    @cached_property
```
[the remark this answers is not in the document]
tylerflex: good point, I will just return a deep copy.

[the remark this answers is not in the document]
tylerflex: actually on second thought, making a regular property makes more sense I think. that way nothing is cached.

tidy3d/plugins/sweep/result.py
tylerflex: yea I worried about this too, I added it to make things easier for the users but I feel like it just confuses and makes for more code maintenance. I will remove this method and update the notebooks.

tidy3d/plugins/sweep/result.py
tylerflex: Same issue as the hexbin? I think I will also remove `scatter` and more generally just recommend people use `DataFrame` for all plotting needs.

tidy3d/plugins/sweep/method.py
tylerflex: I'm in favor of this, but am not sure about how to implement it for discrete sets (like integers or VariableAny). Any idea? 

Do you think this warrants its own Method? eg. MethodMonteCarlo?

lucas-flexcompute: According to the docs in the scipy implementation, it's safe to scale and take the floor of the random number generated in the [0, 1) range (w.r.t. maintaining distribution characteristics).

lucas-flexcompute: > 1. Should `MethodRandom` just use this `scipy.qmc.LatinHypercube` approach? Or would it make sense to make a separate class for Monte Carlo Methods?

Honestly, I don't see a reason to have just independently random picks. I'd just use LH sampling and note it in the documentation. The average user most likely doesn't care and is not aware of the difference. The user who understands it is probably knowledgeable enough to insert their own variable transformation if they want to reshape the distribution or generate the samples themselves and use `MethodIndependent` (can it be made a concrete class that just runs the given values?)

> 2. Would it make sense to try to generalize the monte carlo methods, for example to use any of the `spicy.qmc` engines [here](https://docs.scipy.org/doc/scipy/reference/stats.qmc.html)? Or other samplers. To what extent do you think we should generalize these methods? If I make this one `MethodRandom`, it might limit somewhat what we can do later.

This might actually be a better approach: having `MethodRandom` accept a scipy.stats.qmc.QMCEngine with LH as the default. This would enable advanced users to to whatever they wish, but a good default for everyone else.

lucas-flexcompute: In fact, it doesn't *have* to be a `QMCEngine` object, but any instance that provides a `.random` method with the desired signature.

tylerflex: sounds good. So for integers, I think scaling the distribution and binning makes sense, but for selecting non-numeric objects (like VariablAny) think it's fine to use the same approach? Like does the Monte Carlo algorithm here apply?

tylerflex:
@lucas-flexcompute I implemented an approach to supporting any `scipy` quasi Monte Carlo method in [this commit.](https://github.com/flexcompute/tidy3d/commit/2fc920e0476247ab336e618c4a77d0090cc0929e)

For right now I implemented in a new class (`MethodMonteCarlo`) just to try it out before renaming the random one.

Basic idea:
* Add a `sampler : scipy.stats.qmc.QMCEngine = None` field to `MethodRandom`.
* When we want to sample this method, we call `Method.get_sampler(design_variables)` which returns the sampler. Basically passing the number of dimensions and handling defaults and edge cases.
    * If `self.sampler` is none use `DEFAULT_MONTE_CARLO_SAMPLER` to load and construct the default `QMCEngine` instance.
    * otherwise, use the provided one and check that the dimensions match the design variables.
* I also added a method to all `Variables` that basically takes a set of points between (0,1) and then returns a list of the variables sampled with this distribution.
    * For floats, we just scale and shift.
    * For ints, we scale, shift, and floor.
    * For "any", we scale, floor (to get integer indices) and use these indices to index into `allowed_values`.

When you get time, could you check over my implementation and make sure it makes sense to you before I proceed? If you have any better ideas for API let me know. I also considered doing it so that `Method.sampler` took a key into a dict of supported `QMCEngine` types, but it gets a little more messy and I figured if a user really wanted to customize this, they could simply pass their own instance instead as that's more general anyway.

tidy3d/plugins/sweep/method.py
lucas-flexcompute: Yes, same thing. I don't think there is any "succession" property that would make categorical values (`VariableAny`) different from an integer set.

lucas-flexcompute (approved):
This version looks pretty good to me!

tidy3d/plugins/sweep/variable.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def sample_random(self, num_samples: int) -> List[int]:
+        """Sample this design variable randomly 'num_samples' times."""
+        low, high = self.span
+        return np.random.randint(low=low, high=high, size=num_samples).tolist()
+
+    def _sample_grid(self) -> List[float]:
+        """Sample this design variable on a grid."""
+        low, high = self.span
+        return np.arange(low, high).tolist()
```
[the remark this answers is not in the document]
tylerflex: Good point, but do you think we should include `high` in the possible values?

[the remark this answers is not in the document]
lucas-flexcompute: It's more usual in Python to have an open upper bound. I think it's more pythoninc this way.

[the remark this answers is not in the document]
tylerflex: ok I changed it to open upper bound and clarified the docstring.
```py

    span: Tuple[int, int] = pd.Field(
        ...,
        title="Span",
        description="``(min, max)`` range within which the variable should be swept. "
        "The ``min`` value is inclusive and the ``max`` value is exclusive. In other words, "
        "a grid search over this variable will iterate over ``np.arange(min, max)``."
    )
```

lucas-flexcompute:
> * Renamed `MethodRandom` to `MethodFullyRandom`. Maybe while this is a pre-release feature, we can keep this (and just not document it?
> * Renamed `MethodMonteCarlo` to `MethodRandom`.
> * Tested the Monte Carlo in the notebook and fixed a bug in the number of points setting.

Sounds good. Maybe @tomflexcompute or @e-g-melo have other use cases where the independent random sampling is better suited than the MonteCarlo approach and I junst don't know it!


> * made the span inclusive for integers. I feel this is maybe more intuitive?
I see you went back with the more common approach from python. I think it's the better choice…

I'll take another look at the whole implementation again tomorrow and play a little with it to see how it feels now!

tylerflex:
> Sounds good. Maybe @tomflexcompute or @e-g-melo have other use cases where the independent random sampling is better suited than the MonteCarlo approach and I just don't know it!

Not sure if there is a good reason to do it. I just wonder if some users would be more comfortable with fully random than with monte carlo if they dont understand what it's doing. Although I'm fine removing it too.

> I see you went back with the more common approach from python. I think it's the better choice…

Yea, I could see this going either way, but best to be consistent with python I think unless there's a good reason not to.

> I'll take another look at the whole implementation again tomorrow and play a little with it to see how it feels now!

Sounds good! it's pretty much the same, but the notebooks are a bit updated and the `MethodRandom` does Monte Carlo now, which seems to be working as expected. I guess just ping me when you've had a chance to play with it and we can decide on some next steps (more changes, ask someone else to review, etc).

tomflexcompute:
Sorry I didn't closely follow the development here. The random sampling feature is for doing things like fabrication reliability analysis or some sensitivity analysis?

lucas-flexcompute:
It can be used for both reliability (after design) or to explore the design space (before optimization).

tylerflex:
Basically this is a tool to help users define and sample some parameter space.  Right now they can do it through just a grid search or a random sampling of the parameter space. We were discussing whether the random sampling should be done with Monte Carlo or if we want to provide an option to do truly random sampling. Monte Carlo is a better choice but i wasn't sure if people wanted to do pure random

tylerflex:
Hey @tomflexcompute and @shashwat-sh could you take a look at this PR? And the associated notebook PR in the docs especially?  I want to get another couple pairs of eyes on this tool to see how it might feel to use it. feel free to be critical or suggest different API. We'll try to get this into 2.6 so not a huge rush but want to make sure it's well polished before releasing publically. Lucas already did a lot of work reviewing the internals so that's less important now. Any feedback would help! The docs PR sweep.ipynb is the tutorial and would be a good place to start. Thanks a lot in advance.

tylerflex:
@lucas-flexcompute I added the `MethodRandomCustom` to do what we talked about. It fixes the schema issue as well.
Please take a look when you get a chance just to make sure it looks ok to you, thanks.
https://github.com/flexcompute/tidy3d/commit/2b8e6bb81a5b3ffccc625cc8138d4ab95bdbf63a

shashwat-sh (commented):
This review is for the notebook sweep.ipynb only - I'll take a look at the code next if you'd like. Overall, the feature is awesome! I think integrating the results with pandas is a great idea - not least because some intrepid user could directly take that database to train a hypothetical ChatGPTidy3D in the future :p
- The word "design" is repeated twice [here](https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/9f44d28b35cc297a533ad4c6050b44409fd04cf1/docs/source/notebooks/Sweep.ipynb#L14) and [here](https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/9f44d28b35cc297a533ad4c6050b44409fd04cf1/docs/source/notebooks/Sweep.ipynb#L54)
- Just curious: why not directly define the three separate functions for `pre`, `post`, and `transmission` in cell 3? Not sure why we need cell 2? In that case, maybe you could say in the text that "the three functions above can be combined into a single function, but we have separated them for convenience to allow plotting the simulation before running it, etc."
- Cell 5: For `VariableInt`, can `num_points` or similar be defined in case the user wants to skip integers, e.g. 0:2:6? I guess in that case they should just use `VariableAny` and specify whatever they want?
- Cell 8: How does the `num_points` in the `MethodRandom` interact with `num_points` of `VariableFloat`? Does the former override the latter? (Or maybe I misunderstood what the `num_points` means here?)
- Cell 9 enumerated list: I wonder if it makes sense to use the specific function named we defined above (`transmission`, `pre`, `post` in the inline example code, just to make it super clear what the function arguments are in our specific example.
- Cell 9: regarding the note underneath, is this validated? Also regarding the note that the variable names when defining the `VariableInt` etc. should match the actual variable names - is that validated too? I haven't looked at the code yet, just the notebook, but it seems there is a lot of freedom for users to go awry in defining these objects, so just wondering to what extent we can validate.
- Cell 10: might be useful to remind the user here that they can control the verbosity of the output in case of huge batches.
- Cell 11: so is the output order arbitrary based on how quickly each simulation finishes? Would the results be in a fixed order in the case of `Design.run()`?
- Cell 12: the pandas integration and visualization is very cool!
- Above cell 13: the combining feature is a great idea!

tylerflex:
Thanks @shashwat-sh . dont worry about reviewing the code, I think we have that covered

> The word "design" is repeated twice here and here

Thanks, not sure how those snuck in there.

> Just curious: why not directly define the three separate functions for pre, post, and transmission in cell 3? Not sure why we need cell 2? In that case, maybe you could say in the text that "the three functions above can be combined into a single function, but we have separated them for convenience to allow plotting the simulation before running it, etc."

So, it’s a little strange the way it’s set up, but basically I wrote a single transmission function (kind of like how a user might do it) in cell [2] and then in cell [3] I show how to write the same function using these `pre` and `post` functions as well as `web.run()`, which come in handy for running batches. I included an explicit mention of why we split it up to clarify.

> Cell 5: For VariableInt, can num_points or similar be defined in case the user wants to skip integers, e.g. 0:2:6? I guess in that case they should just use VariableAny and specify whatever they want?

Yea, originally I had this, but it got pretty messy. If a user has a specific set of integers, they’re better off using `VariableAny`. I added a note explaining this use case.

> Cell 8: How does the num_points in the MethodRandom interact with num_points of VariableFloat? Does the former override the latter? (Or maybe I misunderstood what the num_points means here?)

`MethodRandom.num_points` refers to the total number of sampling points in the design space. `VariableFloat.num_points` refers to how many points the float variable is discretized into when doing a grid scan. I added a note explaining this.

> Cell 9 enumerated list: I wonder if it makes sense to use the specific function named we defined above (transmission, pre, post in the inline example code, just to make it super clear what the function arguments are in our specific example.

Took this suggestion.

> Cell 9: regarding the note underneath, is this validated? Also regarding the note that the variable names when defining the VariableInt etc. should match the actual variable names - is that validated too? I haven't looked at the code yet, just the notebook, but it seems there is a lot of freedom for users to go awry in defining these objects, so just wondering to what extent we can validate.

  It can’t be validated in pydantic since we pass the functions as arguments to the methods. Perhaps I could add some checks to these methods though that inspect the signatures of the passed functions, but I can’t really see what, eg `pre` returns without running it.

> Cell 10: might be useful to remind the user here that they can control the verbosity of the output in case of huge batches.

  took this suggestion

> Cell 11: so is the output order arbitrary based on how quickly each simulation finishes? Would the results be in a fixed order in the case of Design.run()?

  No, the output order is determined by how the sampler `Method` is defined.  Then there’s a step to generate a list of input arguments and run them either sequentially or in a batch. Then the results are combined them in the order they were assembled.

> Cell 12: the pandas integration and visualization is very cool!
> Above cell 13: the combining feature is a great idea!

  Thanks!

lucas-flexcompute:
> @lucas-flexcompute I added the `MethodRandomCustom` to do what we talked about. It fixes the schema issue as well. Please take a look when you get a chance just to make sure it looks ok to you, thanks. [2b8e6bb](https://github.com/flexcompute/tidy3d/commit/2b8e6bb81a5b3ffccc625cc8138d4ab95bdbf63a)

Yes, looks better this way: it's more well organized. I'm not sure about the names now… `MethodRandom` vs `MethodFullyRandom` is not very informative, and now with `MethodRandomCustom` it possibly becomes more confusing. Maybe rename the 2 firsts (back?) to `MethodMonteCarlo` and `MethodRandom`.

tylerflex:
Thanks, I renamed those two and did a bunch of other cleanup:
* profiling tests and removed some unused methods.
* adding more tests.
* added a warning to MethodRandom suggesting use of MethodMonteCarlo. This warning can be disabled with a bool field `warn_monte_carlo` passed to `MethodRandom`.
* Implemented `MethodRandom` as a subclass of `AbstractMethodRandom` by passing my own sampler that returns `random((d,n))` to unify everything.

Also updated the docs with the new naming scheme.

tomflexcompute (commented):
Thanks @tylerflex for the implementation. This is a great addition to our current way of doing parameter sweep. I've only looked at the sweep.ipynb too and have some minor comments:

1. There is the method `to_dataframe()` for `td.ModeSolverData`. The method name here is `data_frame`. Maybe we should consider having a consistent method name here so users don't have to remember which one is for which? 

2. Relevant to the previous point but not related to this PR. There are a few other inconsistent names for arguments/coordinates throughout Tidy3D. One example is frequency `freqs` (in monitors for example). In the monitor data, it's `f`, which is slightly inconvenient. There are a few other examples but I don't remember on top of my head now. Maybe it's too late to change this already but just want to mention it. 

3. The eps plot has something strange (looks like a PML strip?) at y=0. Not sure why. 

4. Some typos:
"slighly" -> "slightly"
"procssing" -> "processing"
"btween" -> "between"
"cerate" -> "create"
"Now we can look inspect the new results" -> "Now we can inspect the new results"

5. In the notebook you mentioned Bayesian optimization has not been implemented yet. This is actually a very exciting feature! Looking forward to seeing it in the future.

tylerflex:
> 1. There is the method `to_dataframe()` for `td.ModeSolverData`. The method name here is `data_frame`. Maybe we should consider having a consistent method name here so users don't have to remember which one is for which?

Changed to `to_dataframe()` and `from_dataframe()` for consistency.  Changed the notebooks and the diagram.

> 2. Relevant to the previous point but not related to this PR. There are a few other inconsistent names for arguments/coordinates throughout Tidy3D. One example is frequency `freqs` (in monitors for example). In the monitor data, it's `f`, which is slightly inconvenient. There are a few other examples but I don't remember on top of my head now. Maybe it's too late to change this already but just want to mention it.

Yea, I think this is too late to change, but maybe there's a way to smooth this process (like in `SimulationData.plot_field`, we could add special handling if `freq` is supplied (warn and rename it as `f`)? Maybe we can think about this separately.

> 3. The eps plot has something strange (looks like a PML strip?) at y=0. Not sure why.

It's because it's a 1D simulation and I manually changed the xlims of the plot. The strip is where the original borders are or something. We should consider a separate PR to fix this plotting issue unless you know a better way to plot 1D sims.

> 4. Some typos:
>    "slighly" -> "slightly"
>    "procssing" -> "processing"
>    "btween" -> "between"
>    "cerate" -> "create"
>    "Now we can look inspect the new results" -> "Now we can inspect the new results"

Thanks, my spell checking plugin for jupyter keeps breaking every week it seems..

> 5. In the notebook you mentioned Bayesian optimization has not been implemented yet. This is actually a very exciting feature! Looking forward to seeing it in the future.

Yea! shouldn't be too hard to implement but will take a bit of thought so wanted to get the basics done first.

daquinteroflex:
Documentation structure looks good! Want me to build it to see it or did you?

The other thought I had was maybe we can link the README.md into the built rst documentation (maybe even for the other READMEs) if we want. I need to sort this out for the CHANGELOG.md so we can also include this if we wanted.

daquinteroflex (approved):

tylerflex:
Thanks @daquinteroflex! that's a good idea about the README.md. I had a thought yesterday when going through the docs, especially the API reference, that it might be useful to have a way for users browsing the features (especially the plugins) to quickly click a link to go to a higher level explanation of the plugin, which could probably be defined in the READMEs. Or at least we could link to the corresponding tutorial notebook.

Maybe we can put this off until a later PR though.

daquinteroflex:
Sounds good I'll put it my list when I sort out the CHANGELOG.md in that PR

tylerflex:
one possible place to stick the README: either directly above or below the API reference for that plugin / category
image

PR #1201, opened 2023-10-11
Test initialization speed of a metalens with 10_000 structures

momchil-flex:
I was playing with a large metalens file and wanted to profile the initialization time (~2.5 minutes). One thing I noticed is that the new `_validate_2d_geometry_has_2d_medium` validator was taking a noticeable chunk of the time (~50s), essentially because it was working with `geometry.bounding_box` instead of `geometry.bounds`. The former constructs a `Box` object which adds a lot of overhead when dealing with hundreds of thousands of geometries.

This PR does a few things:
- Adds a test for generating a metalens simulation with 10k structures, and the test will fail if this doesn't work within 3s. Currently, it succeeds in about 2.1s, so if something heavy gets added inadvertently, this should alert us.
- Introduces a `Profile` context in `tests/utils.py` and uses that in the test to display output (if run with `pytest -rA`) that can help us figure out what can be improved.
- Fixes the inefficiency of the 2d geometry validator.

Here are the first lines of profiling at the current state. The main thing that we may want to look at at some point is `correct_shape`. This is specific to the usage of PolySlab. I also looked at just cylinders, and there the pydantic built-ins were dominating, like this `validate_singleton`. I don't know what this is about.

```
10026/10005    0.027    0.000    2.119    0.000 /home/momchil/Drive/flexcompute/tidy3d-core/tidy3d_frontend/tidy3d/components/base.py:90(__init__)
10028/10015    0.019    0.000    2.063    0.000 /home/momchil/miniconda3/envs/flex/lib/python3.9/site-packages/pydantic/v1/main.py:332(__init__)
10028/10015    0.154    0.000    2.018    0.000 /home/momchil/miniconda3/envs/flex/lib/python3.9/site-packages/pydantic/v1/main.py:1032(validate_model)
110223/70072    0.082    0.000    1.801    0.000 /home/momchil/miniconda3/envs/flex/lib/python3.9/site-packages/pydantic/v1/fields.py:852(validate)
110202/110116    0.077    0.000    1.582    0.000 /home/momchil/miniconda3/envs/flex/lib/python3.9/site-packages/pydantic/v1/fields.py:1152(_apply_validators)
    20028    0.012    0.000    0.982    0.000 /home/momchil/miniconda3/envs/flex/lib/python3.9/site-packages/pydantic/v1/class_validators.py:304(<lambda>)
[...]

PR #1203, opened 2023-10-11
Add support for geometry transformations

lucas-flexcompute:
Support is added through a new `Geometry` subclass: `Transformed` and a few convenience functions to instance it from other geometries.

Related issues:
- https://github.com/flexcompute/tidy3d-core/issues/25
- https://github.com/flexcompute/tidy3d-core/issues/339


tests/test_components/test_geometry.py
tylerflex, line 1: could we also add a transformed geometry to the `SIM_FULL` in `tests/utils.py`? this will put it through all of the file IO and other tests.

tidy3d/__init__.py
```diff
@@ -137,6 +137,7 @@ def set_logging_level(level: str) -> None:
 
 log.info(f"Using client version: {__version__}")
 
+Transformed.update_forward_refs()
```
tylerflex, line 140: can you remind me again what these calls do?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Class `Scene` consisting of a background medium and structures for easier drafting and visualization of simulation setups as well as transferring such information between different simulations.
 - Solver for thermal simulation (see `HeatSimulation` and related classes).
 - Specification of material thermal properties in medium classes through an optional field `.heat_spec`.
+- Support for geometry transformations via the `Transformed` class and the helper functions `Geometry.translated`, `Geometry.scaled`, and `Geometry.rotated`.
```
tylerflex, line 13: might want to also add a separate  line mentioning the `intersections_tilted_plane` or other public-facing methods introduced by this.

tidy3d/components/geometry/base.py
```diff
@@ -15,16 +15,38 @@
...
 POLY_GRID_SIZE = 1e-12
 
+try:
+    trimesh_available = True
+    import trimesh
+except ImportError:
+    trimesh_available = False
+
+
+def _check_trimesh():
```
tylerflex, line 42: could consider making this a decorator

```py
def requires_trimesh(fn):
    @functools.wraps(fn)
    def _fn(*args, **kwargs):
        if not trimesh_available:
            raise Tidy3dImportError(
                 "The package 'trimesh' is required for this operation, but it was not found. "
                 "Please install the 'trimesh' dependencies using, for example, "
                 "'pip install -r requirements/trimesh.txt'."
             )
        return fn(*args, **kwargs)
    return _fn

...

@requires_trimesh
def intersections_tilted_plane(...)
```

tidy3d/components/geometry/base.py
```diff
@@ -1795,6 +1952,320 @@ def _surface_area(self, bounds: Bound) -> float:
...
+                    (x0, y1, z1, 1.0),
+                    (x1, y0, z0, 1.0),
+                    (x1, y0, z1, 1.0),
+                    (x1, y1, z0, 1.0),
+                    (x1, y1, z1, 1.0),
+                ]
+            ),
+            self.inverse.T,
+        )
+        # NOTE (Lucas): Bounds are overestimated.
```
tylerflex, line 2080: in #1201 for example, we have some code checking if objects are 2D by comparing their bound mins and maxs, will overestimating the bounds be a problem for this?

tests/test_components/test_geometry.py
tylerflex, line 1: Probably will need to include many more tests (eg that various transformations give expected results).

tylerflex (commented):
@lucas-flexcompute this is really comprehensive and will be great to have. Thanks a lot. I jumped in to take a look before noticing this was a draft, so sorry if any of the comments were things that you just hadn't got to yet. Just a few initial thoughts as I read through things.

tests/test_components/test_geometry.py
lucas-flexcompute: Absolutely! I've also added `ClipOperation`, which was missing.

lucas-flexcompute: Working on those!

tidy3d/components/geometry/base.py
lucas-flexcompute: If the transformation doesn't involve the axis along which the geometry is 2D, it should not matter. For example: a 2D box with dimension 0 along z that is translated or scaled anywhere, or rotated around z will still have 0 bounds along z. If it is rotated around x or y, then that won't be true anymore.

weiliangjin2021:
Very cool feature! Overall looks good.

shashwat-sh (commented):
I took a quick look - looks good to me! I didn't fully understand all the intersection operation details - will go over it again when you finalize the PR.

tidy3d/components/geometry/base.py
```diff
@@ -1795,6 +1952,320 @@ def _surface_area(self, bounds: Bound) -> float:
...
 
+class Transformed(Geometry):
+    """Class representing a transformed geometry."""
+
+    geometry: annotate_type(GeometryType) = pydantic.Field(
+        ..., title="Geometry", description="Base geometry to be transformed."
+    )
+
+    transform: MatrixReal4x4 = pydantic.Field(
+        np.array(((1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1)), dtype=float),
```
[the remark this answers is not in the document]
lucas-flexcompute: Yes. [facepalm]

tidy3d/components/geometry/base.py
tylerflex: ok sounds good, thanks.

tidy3d/components/geometry/base.py
```diff
@@ -1795,6 +1951,307 @@ def _surface_area(self, bounds: Bound) -> float:
...
+                (x, 0.0, 0.0, 0.0),
+                (0.0, y, 0.0, 0.0),
+                (0.0, 0.0, z, 0.0),
+                (0.0, 0.0, 0.0, 1.0),
+            ],
+            dtype=float,
+        )
+
+    @staticmethod
+    def rotation(angle: float, axis: Axis) -> MatrixReal4x4:
```
weiliangjin2021, line 2143: add a staticmethod for rotation around any axis given by a vector?

lucas-flexcompute: I didn't add it because of our weird axis convention: if the user specifies the y-axis using the general rotation vs the current `axis=1` option, they end results will be rotation on opposite directions. Is that ok?

tidy3d/components/geometry/base.py
```diff
@@ -44,6 +40,26 @@
...
 POLY_GRID_SIZE = 1e-12
 
+try:
+    trimesh_available = True
+    import trimesh
+except ImportError:
+    trimesh_available = False
+
+
+def requires_trimesh(fn):
```
tylerflex, line 50: add docstrings? "when decorating a method, requires that trimesh is importable" or something.

tidy3d/components/geometry/base.py
```diff
@@ -158,10 +174,32 @@ def inside_meshgrid(
...
+    ) -> List[Shapely]:
+        """Return a list of shapely geometries at the plane specified by normal and origin.
+
+        Parameters
+        ----------
+        normal : Coordinate
+            Vector defining the normal direction to the plane.
+        origin : Coordinate
+            Vector defining the plane origin.
+        to_2D : MatrixReal4x4
```
tylerflex, line 188: quick question: would it make sense to leave this optional? for example, if not specified, don't perform any transformation?

tidy3d/components/geometry/base.py
```diff
@@ -2011,6 +2161,344 @@ def _surface_area(self, bounds: Bound) -> float:
...
+        """Returns object's volume within given bounds."""
+        # NOTE (Lucas): Bounds are overestimated.
+        vertices = np.dot(self.inverse, self._vertices_from_bounds(bounds))[:3]
+        inverse_bounds = (tuple(vertices.min(axis=1)), tuple(vertices.max(axis=1)))
+        return abs(np.linalg.det(self.transform)) * self.geometry.volume(inverse_bounds)
+
+    def _surface_area(self, bounds: Bound) -> float:
+        """Returns object's surface area within given bounds."""
+        log.warning("Surface area of transformed elements cannot be calculated.")
+        return 0.0
```
tylerflex, line 2295: Would it make sense to `return None` here? in case someone misses the warning, this will throw a wrench in their operations using the surface area and make it more explicit. I see value of returning 0.0 too but could you explain the reasoning?

tidy3d/components/geometry/base.py
```diff
@@ -2011,6 +2161,344 @@ def _surface_area(self, bounds: Bound) -> float:
...
+            Scaling factor along y. If ``None``, use the value for x.
+        z : float = None
+            Scaling factor along z. If ``None``, use the value for x.
+
+        Returns
+        -------
+        numpy.ndarray
+            Transform matrix with shape (4, 4).
+        """
+        if y is None:
```
tylerflex, line 2343: what worries me about this is that people might try
```
scaling(x=2)
```
and not realize that this is automatically setting y and z.  
```
scaling(y=2)
```
on the other hand will fail.

It's probably ok, but there's a bit of asymmetry here that I am not sure about, maybe if it were me I'd say all to be optional and = 1.0 by default? what is the reasoning here? is scaling all x, y, z far more common? and expecting the user to do it via positional argument rather than keyword argument?

tidy3d/components/geometry/base.py
```diff
@@ -2011,6 +2161,344 @@ def _surface_area(self, bounds: Bound) -> float:
...
+            Axis of rotation: 0, 1, or 2 for x, y, and z, respectively, or a 3D vector.
+
+        Returns
+        -------
+        numpy.ndarray
+            Transform matrix with shape (4, 4).
+        """
+        sin = np.sin(angle)
+        cos = np.cos(angle)
+        if axis == 0:
```
tylerflex, line 2379: this might give an error if axis is a numpy array I think. Maybe we should handle this case by converting axis to np.ndarray and then `tolist()` or something like this? 

image

tidy3d/components/geometry/base.py
```diff
@@ -2011,6 +2161,344 @@ def _surface_area(self, bounds: Bound) -> float:
...
+                [
+                    (cos, -sin, 0.0, 0.0),
+                    (sin, cos, 0.0, 0.0),
+                    (0.0, 0.0, 1.0, 0.0),
+                    (0.0, 0.0, 0.0, 1.0),
+                ]
+            )
+
+        try:
+            x, y, z = axis
```
tylerflex, line 2411: related to above comment, I guess we should also try to ensure that the shape of axis is (3,) or I guess (N,3) or (3, N) if we want to support vectorizing. I feel there could be some possible edge case errors that might come  in here that we can nip in the bud.

tidy3d/components/geometry/base.py
```diff
@@ -2011,6 +2161,344 @@ def _surface_area(self, bounds: Bound) -> float:
...
+        except Exception:
+            raise Tidy3dError(
+                "Argument 'axis' must be one of 0, 1, or 2, (indicating axis x, y, or z), or an "
+                "array of 3 values specifying the axis direction."
+            )
+
+        norm = (x**2 + y**2 + z**2) ** 0.5
```
tylerflex, line 2418: np.linalg.norm(axis) might be slightly simpler

tylerflex: and then
```py
x, y, z = axis / np.linalg.norm(axis)
```

tidy3d/components/geometry/base.py
```diff
@@ -2056,10 +2544,52 @@ def to_polygon_list(base_geometry: Shapely) -> List[Shapely]:
...
+            For more details refer to
+            `Shapely's Documentaton <https://shapely.readthedocs.io/en/stable/project.html>`_.
+        """
+        geom_a = Geometry.evaluate_inf_shape(
+            shapely.unary_union(self.geometry_a.intersections_tilted_plane(normal, origin, to_2D))
+        )
+        geom_b = Geometry.evaluate_inf_shape(
+            shapely.unary_union(self.geometry_b.intersections_tilted_plane(normal, origin, to_2D))
+        )
+        if self.operation == "union":
```
tylerflex, line 2574: This is just organizational, but it might help to have a dictionary somewhere defining the mapping of `ClipOperation.operation` to the shapely operation.

```py
operation_mapping = dict(
    union=shapely.union,
    intersection=shapely.intersection,
    difference=shapely.difference,
    symmetric_difference=shapely.symmetric_difference,
)
```
and then this block of code beomes
```py
operation_fn = operation_mapping[self.operation]
operation_result = operation_fn(geom_a, geom_b)
return ClipOperation.to_polygon_list(operation_result)
```

maybe this mapping could be useful in other places as well.

tidy3d/components/geometry/polyslab.py
```diff
@@ -11,12 +11,18 @@
...
-from ..types import Axis, Bound, PlanePosition, ArrayFloat2D
+from ..types import Axis, Bound, PlanePosition, ArrayFloat2D, Coordinate, MatrixReal4x4, Shapely
 from ...log import log
 from ...exceptions import SetupError, ValidationError
 from ...constants import MICROMETER, fp_eps
 
 from . import base
+from . import triangulation
+
+try:
```
tylerflex, line 22: is there a way to unify these attempted trimesh imports? they seem to pop up in many different files, maybe just put them in a single function if it makes sense to?

tylerflex: or try to import trimesh in one module, which sets the `trimesh_available` boolean. if `trimesh` is not found, it also defines a global variable `trimesh=None`, and then other modules import trimesh from that module?

tidy3d/components/geometry/polyslab.py
```diff
@@ -519,6 +525,59 @@ def _move_axis_reverse(arr):
...
+            Transformation matrix to apply to resulting shapes.
+
+        Returns
+        -------
+        List[shapely.geometry.base.BaseGeometry]
+            List of 2D shapes that intersect plane.
+            For more details refer to
+            `Shapely's Documentaton <https://shapely.readthedocs.io/en/stable/project.html>`_.
+        """
+        if len(self.base_polygon) > 500:
```
tylerflex, line 550: let's define this 500 at the top of the module as an all caps constant.

tidy3d/components/geometry/triangulation.py
```diff
@@ -0,0 +1,134 @@
+from typing import List, Tuple
+
+import numpy as np
+import shapely
+
+from ..types import ArrayFloat1D, ArrayFloat2D
+
+
+class Vertex:
```
tylerflex, line 9: why not make this a Tidy3DBaseModel? just for consistency and documentation?

or even just a dataclass if that's easier.

tylerflex (approved):
Thanks @lucas-flexcompute, I just have a set of minor comments / suggestions. Everything related to the functionality looks really good to me! I'm approving for when the comments are addressed / discussed.

tidy3d/components/geometry/base.py
lucas-flexcompute: We have to, somehow, map the 3D coordinates to 2D. We could leave an identity there as default, which means using (x,y) from the output directly.

tidy3d/components/geometry/base.py
lucas-flexcompute: Zero was just arbitrary. I agree with returning `None` to make it even more explicit.

tidy3d/components/geometry/base.py
lucas-flexcompute: Correct. I believe scaling is usually applied homogeneously in all dimensions, but I see the problem. I think having all arguments default to 1 makes more sense.

tidy3d/components/geometry/base.py
tylerflex: ok I see now. let's leave it as is I think.

tidy3d/components/geometry/base.py
lucas-flexcompute: Performing the parameter validation before the comparison seems to work too, and avoids the type juggling. What do you think?

tidy3d/components/geometry/base.py
lucas-flexcompute: I see. I've included a type "normalization" before any comparisons, which should take care of any valid inputs. I'm not allowing vectorization at this point, so I'm treating that case as an error.

tidy3d/components/geometry/base.py
lucas-flexcompute: I've also added an equivalent for bit-type operations.

tidy3d/components/geometry/base.py
tylerflex: we have this, just FYI, not sure if it helps.
Seems some of the rotation bits added to help support anisotropic medium could be useful in here? but not sure.

https://github.com/flexcompute/tidy3d/blob/f60bf47f84264bc622e548fa41dfc344fb85ceaf/tidy3d/components/geometry/base.py#L542-L555

tidy3d/components/geometry/polyslab.py
lucas-flexcompute: I've created a trimesh in types that can be used elsewhere, otherwise, because of circular imports, we'd have to use it as `base.trimesh` or similar. As a bonus, we get to define `TrimeshType` in the correct place now.

tidy3d/components/geometry/triangulation.py
lucas-flexcompute: I guess this is supposed to be internal-use only. We should be able to replace it with something more efficient as soon as we find something without worrying about backwards compatibility of the interface. I've replaced with dataclass just in case.

lucas-flexcompute:
Because of the discussion about the back-end processing, transformed slanted polyslabs should not be allowed, but if we add the validator in `Transformed` we might reject geometries too eagerly. If the user is applying several cascaded transforms that result in a valid end result, I think we should allow it. That's why I've added the validator to `Structure`, where the geometry is frozen.

tidy3d/components/geometry/base.py
lucas-flexcompute: I did see that, but here I need a 4×4 matrix for homogeneous coordinates (and originally only rotations arouns the 3 main axes). We can merge both by adding, for example, another parameter to `RotationAroundAxis(..., homogeneous_coordinates: bool)`. Then, create 2 more classes for `Translation` and `Scaling`.
TBH I don't see much value in creating classes fro those, there's nothing special there, just matrices, but let me know if you prefer it.

lucas-flexcompute: I've pushed a commit where the `rotation` method uses the `RotationAroundAxis` class internally. The question of how to handle homogeneous coordinates and to replace all transforms with classes remains.

tylerflex: I'm less concerned about creating classes for these, but more just suggesting we have a consistent implementation of rotation if possible and therefore don't have to maintain several versions. I think using RotationAroundAxis internally is good for me on this, thanks!

tidy3d/components/geometry/base.py
tylerflex: sounds good.

tidy3d/components/geometry/polyslab.py
tylerflex: great!

tidy3d/components/geometry/base.py
```diff
@@ -2056,10 +2453,61 @@ def to_polygon_list(base_geometry: Shapely) -> List[Shapely]:
...
+        if not result:
+            raise ValueError(
+                "'operation' must be one of 'union', 'intersection', 'difference', or "
+                "'symmetric_difference'."
+            )
+        return result
+
+    @property
+    def _bit_operation(self) -> Callable[[Any, Any], Any]:
+        """Return a function equivalent to this operation unsing bit operators."""
```
shashwat-sh, line 2469: small typo in `using`

shashwat-sh (approved):
Looks good to me, just a few minor comments.

tidy3d/components/geometry/base.py
```diff
@@ -682,6 +739,61 @@ def surface_area(self, bounds: Bound = None):
...
+            Translation along z.
+
+        Returns
+        -------
+        :class:`Geometry`
+            Translated copy of this geometry.
+        """
+        return Transformed(geometry=self, transform=Transformed.translation(x, y, z))
+
+    def scaled(self, x: float, y: float = None, z: float = None) -> Geometry:
```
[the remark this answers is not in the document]
lucas-flexcompute: Good catch, I'd forgotten to change those!

tidy3d/components/geometry/base.py
```diff
@@ -2011,6 +2172,242 @@ def _surface_area(self, bounds: Bound) -> float:
...
+        description="Transform matrix applied to the base geometry.",
+    )
+
+    @pydantic.validator("transform")
+    def _transform_is_invertible(cls, val):
+        # If the transform is not invertible, this will raise an error
+        _ = np.linalg.inv(val)
+        return val
+
+    @pydantic.root_validator(skip_on_failure=True)
```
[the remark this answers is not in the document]
lucas-flexcompute: I doesn't make much sense to use this validator if `geometry` or `transform` aren't really one of the expected types or the transform is singular. It'll just raise another error. Isn't that what `skip_on_failure` means?

weiliangjin2021 (approved):

tidy3d/components/structure.py
```diff
@@ -42,6 +44,20 @@ class AbstractStructure(Tidy3dBaseModel):
 
     _name_validator = validate_name_str()
 
+    @pydantic.validator("geometry")
+    def _transformed_slanted_polyslabs_not_allowed(cls, val):
+        """Prevents the creation of slanted polyslabs rotated out of plane."""
+        if (
+            isinstance(val, Transformed)
+            and isinstance(val.geometry, PolySlab)
```
weiliangjin2021, line 52: what if a geometrygroup or clipper operator is transformed, and it contains a slanted polyslab? is it included here already?

lucas-flexcompute: Good catch, I guess not. We'll need a recursive check in this case. Let me think it through and change it.

weiliangjin2021 (approved):

PR #1204, opened 2023-10-12
Take make_json_compatible out of _json_string as a standalone method

momchil-flex:


tylerflex (approved):
just saw this and am giving my pre-approval :)

momchil-flex:
Thanks - I actually tagged you on the [backend PR](https://github.com/flexcompute/tidy3d-core/pull/431) that necessitated it. :)

tidy3d/components/base.py
```diff
@@ -727,12 +741,14 @@ def make_json_compatible(json_string: str) -> str:
             json_string = json_string.replace("Infinity", '"Infinity"')
             return json_string.replace(tmp_string, '"-Infinity"')
 
-        json_string = self.json(indent=INDENT, exclude_unset=False)
+        if "exclude_unset" not in kwargs:
```
tylerflex, line 744: maybe in looking at this again, it could be better to just set these as kwargs in the function signature, like
```py
def _json_string(self) -> str:
    return self._json()

def _json(indent=INDENT, exclude_unset=False, **kwargs):
    ...
   
    json_string = self.json(indent=indent, exclude_unset=exclude_unset, **kwargs)
```
I think that will work and avoids unnecessary dict annoyances

momchil-flex: Didn't we discuss something like this once about some other method, and find that if e.g. `indent` is also in `kwargs`, it errors?

momchil-flex: Oh sorry this is different.

tylerflex: yea there was something weird about this, but I think it works in this case. I tried it in a notebook. good to double check though.

momchil-flex: Updated as suggested.

tylerflex: ok, you might need to pass indent and the other kwarg in the backend call to `_json()` too just FYI

momchil-flex: Well I'm fine with the defaults so no, right?

tylerflex: oh right, it should be good yea

PR #1207, opened 2023-10-16
Adding verbose argument to estimate_cost and real_cost

momchil-flex:
more detailed messages

cleaning up some docstrings


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Support for multiple frequencies in `output_monitors` in `adjoint` plugin.
 - GDSII export functions to `Simulation`, `Structure`, and `Geometry`.
+- 'verbose' argument to `estimate_cost` and `real_cost` functions such that the cost is logged if `verbose==True` (default). Additional helpful messages may also be logged.
```
tylerflex, line 11: backslash quotes?

tidy3d/web/api/container.py
```diff
@@ -649,32 +639,57 @@ def delete(self) -> None:
...
         """
-        return sum(job.estimate_cost() for _, job in self.jobs.items())
+        batch_cost = sum(job.estimate_cost(verbose=False) for _, job in self.jobs.items())
+
+        if verbose:
+            console = get_logging_console()
+            if batch_cost is not None and batch_cost > 0:
+                console.log(f"Maximum FlexCredit cost: {batch_cost:1.3f} for the whole batch.")
+            else:
+                console.log("Could not get estimated batch cost!")
```
tylerflex, line 693: maybe say why (eg it might not have finished computing, wait and try again?)

tylerflex (approved):

tomflexcompute (approved):
Looks good. I suppose we are keeping the original naming of "flex unit" instead of "flex credit"？

momchil-flex:
> Looks good. I suppose we are keeping the original naming of "flex unit" instead of "flex credit"？

I assume you mean for the variable names? Yeah this is a bit annoying to change, it involves both changing our python code and (mostly) the api endpoints and the server code...

tidy3d/web/api/container.py
momchil-flex: Actually when it comes to estimated cost, it should always work now. This is probably legacy code. This statement should not be reachable since if something goes wrong, the estimate cost should at least error. So yeah I don't really know of a case when this can be reached and so don't know how to improve the message - but may just leave it just in case?

dbochkov-flexcompute (approved):
Looks good. I just was confused what "Note: the task triggered a minimum flex credit cost due to fast early shutoff." means exactly: 0.025 min cost or 1/10 of estimated cost. Maybe this can be clarified somehow?

momchil-flex:
> Looks good. I just was confused what "Note: the task triggered a minimum flex credit cost due to fast early shutoff." means exactly: 0.025 min cost or 1/10 of estimated cost. Maybe this can be clarified somehow?

It refers to the 1/10 of estimated cost. I don't want to hard-code that as it is complicated though. For example, we only pro-rate the part that is due to time stepping; not the part that is due to e.g. field projections, or in the future, mode solver. Furthermore, the 1/10 factor may change, and the message will become stale if it does. I updated the message to

```
                    "Note: the task cost pro-rated due to early shutoff was below the minimum "
                    "threshold, due to fast shutoff. Decreasing the simulation 'run_time' should "
                    "decrease the estimated, and correspondingly the billed cost of such tasks."
```

and then also updated our how are simulations billed FAQ that we can refer people to for more details. There, I put the current 1/10 factor since we can change the docs if that changes:
https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/306/files#diff-399cec32862278b1cf76d3019a66416c9c5c00e55a067554a69ebebba7684ef3

@tomflexcompute will that change automatically propagate to the knowledge base, or does it have to happen manually?

momchil-flex:
@tomflexcompute any comments on this or my FAQ update before I merge (also see question above on propagating FAQ to knowledge base)?

tomflexcompute:
> @tomflexcompute any comments on this or my FAQ update before I merge (also see question above on propagating FAQ to knowledge base)?

Looks good to me. The FAQs will be propagating to the [Learning Center](https://www.flexcompute.com/tidy3d/learning-center/) and Emerson is working on it. Eventually we will combine the FAQs in our docs and knowledge base and open a FAQs section in the Learning Center. Then the FAQs in both places can be removed or at least synced with the Learning Center.

Issue #1217, opened 2023-10-25
Support zero gradients in adjoint plugin


yaugenst:
## Problem description

Currently, if a monitor records amplitudes that are (close to) zero, the adjoint plugin does not place sources in the adjoint simulation (presumably due to [this line](https://github.com/flexcompute/tidy3d/blob/f60bf47f84264bc622e548fa41dfc344fb85ceaf/tidy3d/plugins/adjoint/components/data/monitor_data.py#L82)), leading to the following exception:
```
jax._src.source_info_util.JaxStackTraceBeforeTransformation: pydantic.v1.error_wrappers.ValidationError: 1 validation error for AdjointBatch
jobs -> __root__
  No sources in simulation. (type=value_error.setup)
```

This precludes the use of functions with "binary" gradients such as `jnp.min()`, `jnp.max()` etc. in adjoint-based inverse design with tidy3d, for example when optimizing the minimum transmission of a mode monitor over a batch of simulations. The current workaround is to just not let adjoint source amplitudes go to zero, either by implementing `custom_vjp`'s (and "faking it") or avoiding the use of those functions altogether.

## Proposed solution

It would be nice if tidy3d could support such "zero gradients" natively. My idea would be to piggyback on the existing check whether or not sources are placed in a simulation, and if no sources are placed in an adjoint simulation the simulation should not even attempt to run but instead return zeroed "dummy" monitors/fields. This will automatically lead to the correct (zero) gradients downstream for those monitors, since the gradient is proportional to $E_{\text{fwd}} \cdot E_{\text{adj}}$.

momchil-flex:
I don't remember anymore why we introduced this, but I remember there was a good reason, and I think we did want to consider a better solution (e.g. along the lines of what you propose) but for some reason it wasn't super straightforward? But yeah we hadn't anticipated this specific issue.

tylerflex:
yea I've come across this from time to time and just haven't fixed it yet. The reason we exclude sources with zero amplitudes is because if we added all of the sources, we'd have one for every element in the simulation data, which could easily be far too many. I think the proposed solution is the way to go. I just am not sure how to provide dummy fields in tidy3d. I guess one approach is to copy over the forward simulation data and just multiply all of the field data by 0 one by one.

tylerflex:
[...]

PR #1218, opened 2023-10-26
handle case with no adjoint sources

tylerflex:
This is a quick and dirty way of fixing things, but it should work. If no adjoint sources detected:
* make a point dipole at the center of the simulation with amplitude=0
* set the run time to something around 1/fwidth to avoid unnecessary long runtime.

This assumes that the simulation will still run if the only source has an amplitude of 0. @momchil-flex do you know? If there was some auto-detection of this case (or when no sources are present) that returned zero'd out data that could be another way to handle things.

Main downside of this is that it charges a useless simulation to the user, but I think to do this properly would require a lot of construction, so I wanted to throw this out there to discuss as a medium-short term fix to this relatively uncommon edge case.

I also am not sure how to set a reliably small run_time. Maybe something >0 but small like 1e-20 would be another approach? 

#1217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove warning that monitors now have `colocate=True` by default.
 
 ### Fixed
+- If no adjoint sources for one simulation in an objective function, make a mock source with zero amplitude and warn user.
 
 ## [2.5.0rc1] - 2023-10-10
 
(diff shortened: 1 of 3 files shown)


momchil-flex:
> This assumes that the simulation will still run if the only source has an amplitude of 0. @momchil-flex do you know? If there was some auto-detection of this case (or when no sources are present) that returned zero'd out data that could be another way to handle things.

Yeah it will run, and it won't shortcut in any way, in fact it will always run for the full `run_time` because the auto shutoff will not detect field decay compared to previous field values (which just always stay at 0).

> Main downside of this is that it charges a useless simulation to the user, but I think to do this properly would require a lot of construction, so I wanted to throw this out there to discuss as a medium-short term fix to this relatively uncommon edge case.
>
> I also am not sure how to set a reliably small run_time. Maybe something >0 but small like 1e-20 would be another approach?

Yeah I'm not sure if there could be some issue with a very tiny `run_time`, maybe just setting it as 1/fwdith is fine. The other issue is that not only will it charge the user, it will also still go through all of the overhead in pre and post processing including the adjoint processing, so it will cost us server resources without the user being charged much (esp. with very low run_time). Ideally it would be great if it can be just skipped... but I can imagine it's not trivial so this is fine.

momchil-flex (approved):

PR #1225, opened 2023-10-30
Avoid duplicate log messages in Jupyter

lucas-flexcompute:
When a rich console is created within a jupyter notebook, it autodetects the environment and outputs log messages by default, which ends up duplicating our messages, because we already have a standard output by default. A [flag](https://rich.readthedocs.io/en/stable/reference/console.html#rich.console.Console) can be set to false to inhbit the autodetection.


lucas-flexcompute:
To test this out, this simple script running as a jupyter cell, ipython session, or a standalone python file should demonstrate the issue:
```python
import tidy3d as td
td.set_logging_file("logfile.log")
td.log.warning("Issue some warning")
```

tylerflex:
Thanks Lucas, I'm not sure I could detect the issue when I tried it in ipython. Could you explain? and also add this to the changelog as well? (just updated on pre/2.5, so you'll want to pull)

lucas-flexcompute:
The issue only affects the jupyter interface. The other runners are not
affected, that's part of the problem. I'm not at home now, but I can fix
the PR and changelog tomorrow.

lucas-flexcompute:
So here is the issue better explained. In the python/ipython terminal:


And in Jupyter:


The second message is a "helper" function of the rich console, which also prints log file messages when it detects the jupyter environment. This is fixed by setting the flag `force_jupyter=False` when creating a rich console for file output only.

PR #1227, opened 2023-10-31
Truncate circular filter in edge case

tylerflex:
If the filter radius was larger than the input array (in pixels) along any dimension, scipy raised an error.

Now, we log a warning and truncate the kernel along that dimension to match the input shape (note that the kernel is odd, so it might be 1 smaller if the input shape is even)

I also added a switch if we want to instead raise, maybe this is unnecessary.


e-g-melo (approved):
Thanks Tyler!

The warning message is pretty clear and will let the user decide to go ahead or change the parameters.

PR #1228, opened 2023-10-31
Removing indent in string stored in hdf5

momchil-flex:


tidy3d/plugins/adjoint/components/base.py
```diff
@@ -93,11 +93,10 @@ def from_tidy3d(cls, tidy3d_obj: Tidy3dBaseModel) -> JaxObject:
 
     """ IO """
 
-    @cached_property
-    def _json_string(self) -> str:
```
tylerflex, line 97: any side effects of removing this `_json_string` property?

tylerflex (approved):

tidy3d/components/base.py
```diff
@@ -563,7 +563,7 @@ def to_hdf5(self, fname: str, custom_encoders: List[Callable] = None) -> None:
 
         with h5py.File(fname, "w") as f_handle:
 
-            f_handle[JSON_TAG] = self._json_string
+            f_handle[JSON_TAG] = self._json()
```
tylerflex, line 566: does this `indent=None` by default? maybe it's more clear to pass it explicitly?

momchil-flex:
I've updated this PR to use `indent=None` internally, and I've also introduced the splitting of the string in the hdf5 file. The one thing I'm not sure is what to do with `web/core`, where I find separate functions for reading the string from the hdf5: https://github.com/flexcompute/tidy3d/blob/07a27f243fe644b345c790ad7edd5c8e16567807/tidy3d/web/core/task_core.py#L29-L34 , and there's one more instance.

I think the logic is that I shouldn't be importing anything from tidy3d/components here, so I can't use something like `Tidy3dBaseModel._json_string_from_hdf5`? Then options seem to be:
- Repeat the code
- Introduce a function in something like tidy3d/utils.py that gets imported both in components/base.py and in web/core.py. But does that also go against the future possibility of completely separating web/core.py?

tidy3d/components/base.py
```diff
@@ -563,7 +574,12 @@ def to_hdf5(self, fname: str, custom_encoders: List[Callable] = None) -> None:
 
         with h5py.File(fname, "w") as f_handle:
 
-            f_handle[JSON_TAG] = self._json_string
+            json_str = self._json_string
+            for ind in range(ceil(len(json_str) / MAX_STRING_LENGTH)):
+                key = f"{JSON_TAG}_{ind}" if ind > 0 else JSON_TAG
```
tylerflex, line 579: Maybe we can write a static method or module level function for this to avoid duplicating the key as a function of index, incase we change this or need to use it elsewhere.

tylerflex, line 579: So this creates `JSON_STRING, JSON_STRING_1, ...`?
I'm curious if it will make the logic simpler to just not perform any special handling for `ind=0`, like `JSON_STRING_0, JSON_STRING_1, ...`. If the string is short enough to fit in `JSON_STRING_0`, we just leave it named like that. Will it break backwards compatibility with script loading maybe?

tylerflex: in Tidy3DBaseModel
```py
@staticmethod
def _json_string_key(index):
    if index:
        return f"{JSON_TAG}_{ind}"
    return JSON_TAG
```

`to_hdf5`
```py
json_str = self._json_string
num_strings = ceil(len(json_str) / MAX_STRING_LENGTH)
for ind in range(num_strings):
    ind_start = int(ind * MAX_STRING_LENGTH)
    ind_stop = min(int(ind + 1) * MAX_STRING_LENGTH, len(json_str))
    key = self. _json_string_key(ind)
    f_handle[key] = json_str[ind_start:ind_stop]
```

```py
@classmethod
def _json_string_from_hdf5(cls, fname: str) -> str:
         """Load the model json string from an hdf5 file."""
         with h5py.File(fname, "r") as f_handle:
               num_string_parts = len([key for key in f_handle.keys() if JSON_TAG in key])
               json_string = ""
               for ind in range(num_string_parts):
                     key = _json_string_key(ind)
                     json_string += f_handle[key][()]
           return json_string
```

tylerflex: anyway, see comment and suggestion above, I think it makes things a bit simpler to keep the handling how it is but just have a single function to get the key, treating ind=0 differently.

tests/test_components/test_IO.py
```diff
@@ -21,6 +21,12 @@
 SIM_DIR = "tests/sims"
 
 
+@pytest.fixture
+def split_string(monkeypatch):
```
tylerflex, line 25: nice :)

tylerflex:
Regarding `web/core`, that's MC-controlled and I think we're duplicating code in there until we can unify it at a later date (@dbochkov-flexcompute?) so I'd say maybe we can just duplicate but add a `#TODO:` to unify with this method later.

dbochkov-flexcompute:
> I've updated this PR to use `indent=None` internally, and I've also introduced the splitting of the string in the hdf5 file. The one thing I'm not sure is what to do with `web/core`, where I find separate functions for reading the string from the hdf5:
> 
> https://github.com/flexcompute/tidy3d/blob/07a27f243fe644b345c790ad7edd5c8e16567807/tidy3d/web/core/task_core.py#L29-L34
> 
> , and there's one more instance.
> I think the logic is that I shouldn't be importing anything from tidy3d/components here, so I can't use something like `Tidy3dBaseModel._json_string_from_hdf5`? Then options seem to be:
> 
> * Repeat the code
> * Introduce a function in something like tidy3d/utils.py that gets imported both in components/base.py and in web/core.py. But does that also go against the future possibility of completely separating web/core.py?

> Regarding `web/core`, that's MC-controlled and I think we're duplicating code in there until we can unify it at a later date (@dbochkov-flexcompute?) so I'd say maybe we can just duplicate but add a `#TODO:` to unify with this method later.

So far, the way we have been dealing with importing tidy3d things into web api is through passing those things during initialization of web api here https://github.com/flexcompute/tidy3d/blob/07a27f243fe644b345c790ad7edd5c8e16567807/tidy3d/web/__init__.py#L32
For example, currently we pass `log`, logging console (should change that to logging console getter eventually), and tidy3d version. So, perhaps `_json_string_from_hdf5()` could also be set in web api from tidy3d in the same way? That is:
```
core_config.set_config(log, get_logging_console(), __version__, Tidy3dBaseModel._json_string_from_hdf5)
```

tylerflex:
yea perhaps this is a way to do it, although it seems a bit strange potentially.

Maybe an alternative approach is to handle this in the SimulationTask, grab the class of the component for the task and then call `cls._json_string_from_hdf5`?

dbochkov-flexcompute:
> yea perhaps this is a way to do it, although it seems a bit strange potentially.
> 
> Maybe an alternative approach is to handle this in the SimulationTask, grab the class of the component for the task and then call `cls._json_string_from_hdf5`? Just not sure how to grab the class without downloading the whole object.

Yeah, I think it would work there. However, the other place where json string is extracted from hdf5 is here https://github.com/flexcompute/tidy3d/blob/07a27f243fe644b345c790ad7edd5c8e16567807/tidy3d/web/api/tidy3d_stub.py#L152-L169 and the purpose there is to actually figure our what simulation class it is

tylerflex:
I see, that's a bit messy, maybe the easiest approach is to just duplicate but make a note and get qingeng to help unify things, but passing in the config also works (even though it seems like not the long term solution to me).

tidy3d/components/base.py
momchil-flex: Sounds good, yeah not appending `ind=0` is for backwards compatibility.

tests/test_components/test_IO.py
momchil-flex: I had something similar worked out in a backend test already :)

momchil-flex:
Updated, with code duplicated in `web.core.py`.

tidy3d/components/base.py
```diff
@@ -22,9 +23,12 @@
 from ..exceptions import FileError
 from ..log import log
 
-# default indentation (# spaces) in files
-INDENT = 4
+
+INDENT_JSON_FILE = 4  # default indentation of json string in json files
```
tylerflex, line 27: we could consider setting this to 2 to reduce the size of the file itself, but no preference either way.

tidy3d/components/base.py
```diff
@@ -430,6 +434,23 @@ def get_sub_model(cls, group_path: str, model_dict: dict | list) -> dict:
                     model_dict = model_dict[key]
         return model_dict
 
+    @staticmethod
+    def _json_string_key(index):
```
tylerflex, line 438: maybe add some type annotations just while we're at this
```py
def _json_string_key(index: int) -> str:
```

tylerflex (approved):
looks good, thank you! just a couple minor notes

tidy3d/components/base.py
tylerflex: note: I'm not sure this will actually change the file size. depends on how they json handles indents I guess.

momchil-flex: Tested. It does change the size. I guess let's do 2 yeah? But this shouldn't matter for upload/download from the server because all this is done through hdf5 files now. So not as important potentially, could be left 4 since it's prettier?

tylerflex: fine by me. I prefer 4 if the file size is not an issue.

Issue #1231, opened 2023-11-06
Merge `docs` and `source` repositories


daquinteroflex:
We'd like to merge the documentation with the source code to have a more maintainable documentation environment and configuration. However, we don't want to give `rtd` organization access. One way to solve this is to implement `rtd` web hooks as described [on these instructions](https://docs.readthedocs.io/en/stable/guides/setup/git-repo-manual.html) The strategy to perform the actual repository merge could follow something like [this blog post](https://gfscott.com/blog/merge-git-repos-and-keep-commit-history/)

This issue is to track progress integrating the docs and source code, and whatever arises from doing that.

daquinteroflex:
A few things to note, each of those pull requests contains the non-conflicting repository status of the docs and source code, **in my own personal forks**. The idea is that by changing the remote now of the docs repo (which again the remotes are my own personal forks), it should be possible to rebase all the git history of the docs into the main repository. 

It is important to mention we should be able to support progressive integration from the docs to the source until whenever we want, as long as all the work is contained within the `docs/` directory.

Whenever we want to integrate with the `develop` branch, then we just should need to merge https://github.com/flexcompute/tidy3d/pull/1234

tylerflex:
>  Are we happy to restructure the docs/ folder so that it only contains documentation inside there?

Yea I think that should be fine.

>  Are we happy with this merge approach in order to maintain git history?
>  Are we happy to stick to the source code tidy3d license?

Maybe @momchil-flex has some thoughts on license and git history.

momchil-flex:
> > Are we happy with this merge approach in order to maintain git history?
> > Are we happy to stick to the source code tidy3d license?
> 
> Maybe @momchil-flex has some thoughts on license and git history.

License sounds fine. Re the merge approach, is that explained in detail somewhere? It is not clear to me from just the messages here.

daquinteroflex:
Sounds good! So the merge approach is in development still. The idea
currently is to move all example notebooks to a separate repo, and merge
the docs/source code to improve shared packaging and development. This
should be nearly tiny impact on the current web examples deployment
approach, and should streamline the documentation builds.

momchil-flex:
The part I'm still not sure about is that you seem to be saying this can preserve the docs history? Would that mean all of the docs commits applied on top of the current tidy3d frontend? But somehow with notebooks removed?

daquinteroflex:
There’s a few ways to do this. Check out this pull request
[...]

PR #1234, opened 2023-11-07
Merge `tidy3d` and `docs` with history

daquinteroflex:
Related to https://github.com/flexcompute/tidy3d/issues/1231 and  https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/321

Note that this is from a branch that carries on some of the commits of the poetry upgrade, but they are different branches so it's easier to upgrade or work in one or another direction accordingly


daquinteroflex:
Good news! Successful merge of the `docs` and `source` repositories, and it's progressively integreable. We should be able to rebase this PR and then integrate into the standard `develop branch accordingly.`

daquinteroflex:
Things that need to be done after the repos are merged:

- [ ] Update the `.readthedocs.yaml` to install based on poetrry
- [ ] Decide a date where the other repo won't be supported anymore

daquinteroflex:
We don't want to maintain history, as it will be maintained by the docs repository. See this new PR https://github.com/flexcompute/tidy3d/pull/1238.

PR #1236, opened 2023-11-07
use mode solver fields in conversion methods

### LINKED ISSUE #1232: Use ModeSolver fields for to_monitor() and to_source()
Currently, for ModeSolver there are fields `freqs` in `to_monitor()` and `direction` in `to_source()`, both fields exist in ModeSolver. These could have been optional and use the ModeSolver fields if not specified.

tylerflex:
Fixes #1232


hirako22:
Looks good to me!

PR #1241, opened 2023-11-09
optimize the performance of web.delete_old().

magiWei:


tylerflex (commented):
Looks fine to me, thanks @magiWei 

Can you also add in the `CHANGELOG.md` under "fixed"
```
Optimized performance of `web.delete_old()`.
```

tidy3d/web/api/webapi.py
```diff
@@ -640,15 +638,7 @@ def delete_old(
...
-    tasks = folder.list_tasks()
-    if not tasks:
-        return 0
-    tasks = list(
-        filter(lambda t: t.created_at < datetime.now(pytz.utc) - timedelta(days=days_old), tasks)
-    )
-    for task in tasks:
-        task.delete()
-    return len(tasks)
+    return folder.delete_old(days_old)
```
lei-flex, line 641: if the data is large, for example, 10K tasks.  will the webapi get time out with new code?
I would like to see the test result to confirm it.

lei-flex: another question is how we tell user the progress during the deletion.
use 30K task as examples since it is current max task count right now.

tylerflex:
Hi @magiWei were you able to address @lei-flex 's comments? We'd like to try to get this PR into 2.6.0 if possible. Thanks!

magiWei:
> Optimized performance of `web.delete_old()`.



> Hi @magiWei were you able to address @lei-flex 's comments? We'd like to try to get this PR into 2.6.0 if possible. Thanks!

Yeah,  you get this PR into 2.6.0

tylerflex:
@lei-flex do you approve the PR now?

lei-flex:
> @lei-flex do you approve the PR now?

I didn't see any code change on it. 
@magiWei , please provide more response about how did u resolve the comments. thanks.

tylerflex:
pinging @magiWei here, could you respond to lei's question when you get a chance? trying to figure out what to do with this PR before it gets stale thanks

magiWei:
> pinging @magiWei here, could you respond to lei's question when you get a chance? trying to figure out what to do with this PR before it gets stale thanks

1. The server will set the deleted data not to operate for 100 days to ensure that the amount of deleted data is not too large.
2. We have tested deleting 30,000 pieces of data within 5 seconds.
3. The interface will return the total number of successfully deleted data.
4. If there is too much data for 100 days and the deletion interface times out, the corresponding error will be thrown and the user can re-operate the failed data.

lei-flex:
> > pinging @magiWei here, could you respond to lei's question when you get a chance? trying to figure out what to do with this PR before it gets stale thanks
> 
>     1. The server will set the deleted data not to operate for 100 days to ensure that the amount of deleted data is not too large.
[lei] that's wrong.  we should allow user to delete any data before 100 days.  only allow them to delete the tasks in 100 days didn't make sense to me. 
> 
>     2. We have tested deleting 30,000 pieces of data within 5 seconds.
[lei] could you send me the script about the test? 
I thought the aws dynamdb has the limit of 1K data update per second for one user, I would try your scripts to see if any thing changed on aws dynamodb. 
> 
>     3. The interface will return the total number of successfully deleted data.
[lei] we should not do it.  return  30K data in one time is not good design. 
> 
>     4. If there is too much data for 100 days and the deletion interface times out, the corresponding error will be thrown and the user can re-operate the failed data.
[lei] the python code should retry itself and we should not ask user write their own to handle it. 


> 
> 
> So ,I think it can be merged to 2.6.0, how do you think? @lei-flex


Overall, the real case:  the user will use this api when they actaully have tons of data to delete and save the storage cost. 
so the deleted tasks must be a lot. 
1. compare the efficiency,   stability and progress update is more important. 
2. I would like to see: 
    a.  chunk the tasks (100 or 1000) that  user need to delete.
    b. commit the chunk to make sure all get done (db and s3). 
    c. update the progress in python code.
    d. process the next chunk until all done.
    e. also the code should at least retry once if something is wrong to bring smooth usage of user side. 

@magiWei @tylerflex sorry, I can't approve the code right now, if some users required to batch delete for urgent, please submit the ticket and I ask @yuweiflexcompute to do it as admin  rather than we expose the non-mature api to user with rush

magiWei:
The current implementation of python client：
1. obtain all tasks, if users have tons of data, it also will timeout or OOM;
2. filter out the tasks at the specified time;
3. traverse and delete them. it will cost too much network request overhead;

>[lei] that's wrong. we should allow user to delete any data before 100 days. only allow them to delete the tasks in 100 days didn't make sense to me.

[qingeng] I don’t think it is very reasonable to delete all the data at once, deleting in batches is a better way.
1. It increases the pressure on the server. 
2. Deleting the synchronization interface will take a long time to wait and increase the probability of timeout.

> [lei] could you send me the script about the test?
I thought the aws dynamdb has the limit of 1K data update per second for one user, I would try your scripts to see if any thing changed on aws dynamodb.

[qingeng] Because aws dynamdb has the limit of 1K data，so we need to control the number of deletions, for example 1K data per request. otherwise, we may have to use a scan with poorer performance.

> update the progress in python code.
It will be a big change, need more discuss.

magiWei:
@tylerflex We won’t merge to 2.6.0 yet, more discussion is needed about the implementation, thank you.

PR #1246, opened 2023-11-11
better error message when user try to download data that doesn't exist.

magiWei:


tidy3d/web/api/mode.py
```diff
@@ -423,13 +423,18 @@ def get_result(
...
+            download_file(
+                self.solver_id,
+                MODESOLVER_RESULT,
+                to_file=to_file,
+                verbose=verbose,
+                progress_callback=progress_callback,
+            )
+        except Exception:
+            raise WebError(
+                "Fail to download file from server. Please confirm if the task status = success."
```
tylerflex, line 436: Please change to
```
f"Failed to download file '{MODESOLVER_RESULT}' from server. Please confirm that the task was successful."
```

tidy3d/web/core/task_core.py
```diff
@@ -472,13 +472,18 @@ def get_sim_data_hdf5(
...
-            self.task_id,
-            SIMULATION_DATA_HDF5,
-            to_file=to_file,
-            verbose=verbose,
-            progress_callback=progress_callback,
-        )
+        try:
+            return download_file(
+                self.task_id,
+                SIMULATION_DATA_HDF5,
```
tylerflex, line 478: Please change to
```
f"Failed to download file '{SIMULATION_DATA_HDF5}' from server. Please confirm that the task was successful."
```

tylerflex (changes requested):
Thanks @magiWei. Please also include a changing item under "fixed"
```
Improved error handling if file can not be downloaded from server.
```
Our practice is to include a changelog item for any non-trivial change to our code so we have a clear record of the timeline for various change. thank you!

magiWei:
> Thanks @magiWei. Please also include a changing item under "fixed"
> 
> ```
> Improved error handling if file can not be downloaded from server.
> ```
> 
> Our practice is to include a changelog item for any non-trivial change to our code so we have a clear record of the timeline for various change. thank you!

added.

PR #1249, opened 2023-11-13
adds `SimulationData.mnt_data_from_file()` method to load individual `MonitorData`

tylerflex:
Syntax:
```py
mnt_data = SimulationData.mnt_data_from_file(fname, mnt_name, **kwargs)
```
loads a monitor data from a subgroup in the `SimulationData` file.

1. grabs the `JSON_STRING` to get the monitor type info.
2. Looks up the proper `MonitorDataType`
3. Iterates through the hdf5 file until a match is found on monitor name
4. Uses `MonitorData.from_file()` with a constructed group path to do the loading.

Some error checking thrown in, only works with `.hdf5` files at the moment.

### DIFF
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -2,9 +2,12 @@
 from __future__ import annotations
 from typing import Callable, Tuple
 
+import pathlib
 import xarray as xr
 import pydantic.v1 as pd
 import numpy as np
+import h5py
+import json
 
 from .monitor_data import MonitorDataTypes, MonitorDataType, AbstractFieldData, FieldTimeData
 from ..simulation import Simulation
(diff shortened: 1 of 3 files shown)


tidy3d/components/data/sim_data.py
```diff
@@ -369,6 +376,65 @@ def get_intensity(self, field_monitor_name: str) -> xr.DataArray:
             field_monitor_name=field_monitor_name, field_name="E", val="abs^2"
         )
 
+    @classmethod
+    def mnt_data_from_file(cls, fname: str, mnt_name: str, **parse_obj_kwargs) -> MonitorDataType:
+        """Loads a monitor data from .hdf5, or .hdf5.gz file.
```
shashwat-sh, line 381: The check below (line 404) does not seem to allow `.hdf5.gz`. Does `pathlib.Path(fname)` internally extract `.gz` files?

shashwat-sh (approved):
Looks good! Just one small clarification below.

tidy3d/components/data/sim_data.py
tylerflex: oh right, no the pathlib.path just is to check the extension but I'll make the docstring clearer that it only works for hdf5.

momchil-flex: I guess if we start gzipping simulation data files, we should still unzip them right after download.

PR #1255, opened 2023-11-16
Some private helper properties in ModeSolver

momchil-flex:


dbochkov-flexcompute (approved):

momchil-flex:
@dbochkov-flexcompute could you take another look? I introduced a new commit to avoid erroring out if `ModeSolver` is imported but scipy is not installed. This is so that we can e.g. run metadata on ModeSolver objects using our web services while keeping those small.

dbochkov-flexcompute:
> @dbochkov-flexcompute could you take another look? I introduced a new commit to avoid erroring out if `ModeSolver` is imported but scipy is not installed. This is so that we can e.g. run metadata on ModeSolver objects using our web services while keeping those small.

yeah, all makes sense to me

PR #1263, opened 2023-11-21
Allow inhomogeneous substrate and superstrate for 2d materials

caseyflex:


tylerflex:
Is this PR going to be replaced by #1395 ? @dmarek-flex @caseyflex ? should we close?

Issue #564, opened 2022-11-03
Use enums instead of literal


tylerflex:
A lot of times we have

```py
from typing import Literal

TypeOfSomething = Literal['opt1', 'opt2']

class M:
    something : TypeOfSomethng = pd.Field('opt1')

    def do_something(self):
        if self.something == 'opt1':
            print('hi')
```

which has a few problems:
* if eg `"opt1"` changes to something else, we need to fix all occurrences of the string.
* IDEs can't use this type hint to help users know what the options are.

Better / more standard approach is to use enums

```py
from enum import Enum

class TypeOfSomething(str, Enum)
    opt1 = 'opt1'
    opt2 = 'opt2'

class M:
    something : TypeOfSomething = pd.Field(TypeOfSomething.opt1)

    def do_something(self):
        if self.something == TypeOfSomething.opt1:
            print('hi')
```

See this for more info: https://pydantic-docs.helpmanual.io/usage/types/#enums-and-choices

instances where this occurs (might be missing some)
- [ ] Logging level
- [ ] Symmetry
- [ ] Direction
- [ ] fitter pol (TE / TM)
- [ ] precision (single / double)
- [ ] near2far exclude surfaces
- [ ] sim_data plotting val (real, imag, int)
- [ ] dispersion constraints (hard, soft)
- [ ] dispersion config environment
- [ ] webapi get tasks order
- [ ] polyslab plane position
- [ ] dipole source polarization
- [ ] field type in monitors

yaugenst-flex:
Closing as stale — no meaningful activity in 1.5+ years. Please reopen or file a new issue if still relevant.

PR #1266, opened 2023-11-21
Ensuring mode solver fields are returned in requested precision

momchil-flex:
This was motivated by two things when using the default single precision:
- Smaller file size for download
- Smaller chance of OOM which can sometimes happen for very large mode solves in the post-processing (colocating, normalizing, sorting modes).


tylerflex (approved):
Looks good to me. I feel like eventually we can move towards using `Enum` for things like this, but for now it's ok. Maybe along with #564 we can try to use enums internally as well throughout the codebase. @daquinteroflex

dbochkov-flexcompute:
Looks good to me as well. When slicing data for GUI I convert everything into single precision no matter what's the input to make storage and downloading more efficient, do you think that's ok to keep it that way? (cc: @xin-flex)

momchil-flex:
Conversely this makes me wonder if we should just always store the mode fields in single precision? The precision argument was added because in some cases it may be important for the mode solve; but casting the final results seems pretty innocuous?

momchil-flex:
I think using the "least amount of work" razor we can assume both are fine as they are. :D

Issue #1261, opened 2023-11-20
Add a phase parameter to frequency-domain monitor data.


e-g-melo:
**Is your feature request related to a problem? Please describe.**
In our meetings with professors interested in the teaching program, they frequently ask how to create field animations using Tidy3d. 

**Describe the solution you'd like**
Currently, we have the FieldTimeMonitor, which allows us to create time-domain animations easily. However, animating the frequency-domain fields would be nice and less memory-intensive for teaching purposes. For example, to show how the waves propagate from a point-dipole source, we could animate the field phase, like below:

![dip_anim](https://github.com/flexcompute/tidy3d/assets/125516370/867310bf-3d85-4e8e-8a3c-4a2413955064)

**Describe alternatives you've considered**
It is possible to get the field components from the FieldData object, change the phase, and create an animation. However, a built-in feature would help Tidy3D beginners.

**Additional context**
Other field animation example from mode data:
![mode_anim](https://github.com/flexcompute/tidy3d/assets/125516370/d82a04e4-58a5-4338-bf93-de0e424ea850)

momchil-flex:
I am a bit confused by the title of this issue vs. what it's asking for. Are you suggesting a convenience method to create the entire animation, like a `FieldData.animate(...)`? Or what do you mean by "add a phase parameter"?

momchil-flex:
Assigning @tylerflex so that someone is on this, but feel free to reassign if you want.

tylerflex:
@e-g-melo could you explain what the API should look like for adding a phase parameter? like 

```py
field_data = field_data.apply_phase(phase)
```
or something directly stored in the field data? like `FieldData(..., phase=1.0)` that we can modify with `updated_copy()`?

e-g-melo:
We could include a phase parameter in the `plot_field` function for both FDTD and ModeSolver data, like below:

```
sim_data.plot_field("field", "Ex", "real", y=0.0, f=200e12, eps_alpha=0.2, phase=phase)
```

```
mode_solver.plot_field("Ex", "real", mode_index=0, f=freq0, ax=ax, phase=phase)
```

And include a function like the `apply_phase(phase)` as well to allow for custom plots or whatever the user wants to do.

```
mode_data_1 = mode_data.apply_phase(phase)
mode_data_1.Ex.isel(mode_index=0, f=freq0)).real.plot(x="x", y="z", ax=ax, cmap="RdBu")
```

```
field_data_1 = field_data.apply_phase(phase)
[...]

PR #1271, opened 2023-11-22
apply phase to `FieldData` and `SimulationData.plot_field`

tylerflex:
#1261 

Adds
*  `FieldData.apply_phase(phase :float) -> FieldData`
* `phase` argument to `SimulationData.plot_field` that either
     * applies a phase to the field data being plotted (if complex-valued)
     * warns the user that it will be ignored (if real-valued and phase is non-zero)

### DIFF
--- a/tidy3d/components/data/dataset.py
+++ b/tidy3d/components/data/dataset.py
@@ -32,6 +32,17 @@ class AbstractFieldDataset(Dataset, ABC):
     def field_components(self) -> Dict[str, DataArray]:
         """Maps the field components to thier associated data."""
 
+    def apply_phase(self, phase: float) -> AbstractFieldDataset:
+        """Create a copy where all elements are phase-shifted by a value (in radians)."""
+        if phase == 0.0:
+            return self
+        phasor = np.exp(1j * phase)
+        field_components_shifted = {}
+        for fld_name, fld_cmp in self.field_components.items():
+            fld_cmp_shifted = phasor * fld_cmp
+            field_components_shifted[fld_name] = fld_cmp_shifted
+        return self.updated_copy(**field_components_shifted)
+
     @property
     @abstractmethod
     def grid_locations(self) -> Dict[str, str]:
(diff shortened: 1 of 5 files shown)


e-g-melo:
@tylerflex, I am using the code below to generate a sequence of field images from a `FieldMonitor` and I am getting that warning about real-valued data. What should I do to get complex-valued data?

```
phi_range = np.linspace(0, 2*np.pi, 36)
for phi, i in zip(phi_range, range(0,len(phi_range))):  
    fig, ax = plt.subplots(1,1, figsize=(5,5), tight_layout=True)

    sim_data.plot_field("field_xy", "Ey", f=freq_c, ax=ax, phase=phi)
    
    ax.set_aspect('equal')       
    plt.savefig(f'DipAnim/dip_{i:03d}.png', bbox_inches='tight')
    plt.close()
```

image

tylerflex:
My bad, there was a bug, It should be fixed now.

e-g-melo (approved):
@tylerflex, thank you for this feature! It will help a lot in promoting Tidy3D within the teaching program context.

Both `apply_field()` and `plot_field` functions are working. However, the warning message still shows up when using `plot_field`. 

image

e-g-melo:
dip_anim_plot_field

tylerflex:
great!

tidy3d/components/data/dataset.py
```diff
@@ -32,6 +32,15 @@ class AbstractFieldDataset(Dataset, ABC):
     def field_components(self) -> Dict[str, DataArray]:
         """Maps the field components to thier associated data."""
 
+    def apply_phase(self, phase: float) -> AbstractFieldDataset:
+        """Create a copy of this dataset where all elements are phase shifted by value (radians)."""
```
lucas-flexcompute, line 36: "… phase-shifted by a value (in radians)"?

tidy3d/components/data/dataset.py
```diff
@@ -32,6 +32,15 @@ class AbstractFieldDataset(Dataset, ABC):
...
+        phasor = np.exp(1j * phase)
+        field_components_shifted = {}
+        for fld_name, fld_cmp in self.field_components.items():
+            fld_cmp_shifted = phasor * fld_cmp
+            field_components_shifted[fld_name] = fld_cmp_shifted
+        return self.updated_copy(**field_components_shifted)
```
lucas-flexcompute, line 42: Should we avoid creating a copy here if `phase == 0.0`? This will be run for every plot, most times with the default value, and I wander if it would have noticeable impact for large datasets.

tylerflex: sure, I can just return. Note that in the `SimulationData.plot_field()`, I have a [check](https://github.com/flexcompute/tidy3d/blob/tyler/phase/tidy3d/components/data/sim_data.py#L447-L455) that phase != 0.0 before calling `apply_phase`, so this shouldn't be called in the plotting unless the phase is non-zero. However, I'll add an extra check here just in case.

tylerflex:
made all recommend changes

PR #1273, opened 2023-11-22
Enforcing limits on the internal memory needed by monitors

momchil-flex:
Currently, monitors that produce a small amount of data but may need a large amount of memory internally are not really validated to prevent OOM on our servers. I was thinking on imposing some limits like max number of freqs, or product of num_freqs x num_points x num_modes for mode monitors, etc. but I realized what probably makes the most sense is restricting the data size of the associated *field* monitor, which is always needed internally.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -72,6 +72,7 @@ WARN_TIME_STEPS = 1e6
 MAX_GRID_CELLS = 20e9
 MAX_CELLS_TIMES_STEPS = 1e16
 WARN_MONITOR_DATA_SIZE_GB = 10
+MAX_MONITOR_INTERNAL_DATA_SIZE_GB = 50
 MAX_SIMULATION_DATA_SIZE_GB = 50
 WARN_MODE_NUM_CELLS = 1e5
 
@@ -992,7 +993,7 @@ class Simulation(AbstractSimulation):
         with log as consolidated_logger:
             datas = self.monitors_data_size
             for monitor_ind, (monitor_name, monitor_size) in enumerate(datas.items()):
-                monitor_size_gb = monitor_size / 2**30
+                monitor_size_gb = monitor_size / 1e9
                 if monitor_size_gb > WARN_MONITOR_DATA_SIZE_GB:
                     consolidated_logger.warning(
                         f"Monitor '{monitor_name}' estimated storage is {monitor_size_gb:1.2f}GB. "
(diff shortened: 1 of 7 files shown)


tidy3d/components/simulation.py
```diff
@@ -1049,19 +1082,20 @@ def warn_mode_size(monitor: AbstractModeMonitor, msg_header: str, custom_loc: Li
...
-            # take monitor downsampling into account
-            num_cells = monitor.downsampled_num_cells(num_cells)
-            num_cells = np.prod(num_cells)
-            monitor_size = monitor.storage_size(num_cells=num_cells, tmesh=tmesh)
-            data_size[name] = float(monitor_size)
+            data_size[monitor.name] = float(self._monitor_data_size(monitor))
 
         return data_size
 
+    def _monitor_data_size(self, monitor):
```
tylerflex, line 1091: Can you add type annotations?

tidy3d/components/simulation.py
```diff
@@ -1009,6 +1010,38 @@ def _validate_monitor_size(self) -> None:
                 f"a maximum of {MAX_SIMULATION_DATA_SIZE_GB:.2f}GB are allowed."
             )
 
+        # Some monitors store much less data than what is needed internally. Make sure that the
+        # internal storage also does not exceed the limit.
+        check_monitor_types = [AbstractModeMonitor, SurfaceIntegrationMonitor, DiffractionMonitor]
+        for monitor in self.monitors:
+            if not isinstance(monitor, tuple(check_monitor_types)):
```
tylerflex, line 1017: Not necessary to convert to tuple here i think

tidy3d/components/simulation.py
```diff
@@ -1009,6 +1010,38 @@ def _validate_monitor_size(self) -> None:
...
+                continue
+
+            mnt_kwargs = {
```
tylerflex, line 1020: I think a cleaner way to do this could be to add a property for monitors that returns the equivalent of the backend field monitor that would be used for that type of monitor (before processing) and then compute the storage size of that monitor here in the validator. What do you think?  This might also be useful on the backend potentially or for debugging purposes

tylerflex (changes requested):
Couple minor comments and suggestions but overall looks good

momchil-flex:
> I think a cleaner way to do this could be to add a property for monitors that returns the equivalent of the backend field monitor that would be used for that type of monitor (before processing) and then compute the storage size of that monitor here in the validator. What do you think? This might also be useful on the backend potentially or for debugging purposes

So I was thinking about doing it like this initially but it doesn't make that much sense actually; it doesn't work exactly like that on the backend. For example, surface integration monitors do not actually add a FieldMonitor that is then post-processed; on the other hand, ModeMonitor-s do. Also even disregarding that it wasn't easy for me to come up with a way to add this in a nicely unified way that would not need this property to be overwritten in multiple classes.

I've modified things to just introduce a server storage method. What do you think about this? I also added a limit on the max data that a mode solver call can produce, because more than 20GB is likely to error currently. Finally, I also realized that the previous implementation was not right for TimeMonitor-s (essentially FluxTimeMonitor is the only one affected) because it was making a FieldTimeMonitor with the same amount of time steps; instead, the fields at a single time step are stored internally at any given time (this also goes to the above point that there's no direct correspondence to a FieldMonitor or FieldTimeMonitor).

tomflexcompute:
Can the current limit ensure OOM never occurs?

momchil-flex:
> Can the current limit ensure OOM never occurs?

Unfortunately no. There's too many ways in which it can happen and we need more work on server-side resource allocation too. But this will prevent some cases which would otherwise almost certainly have gone oom.

tidy3d/components/simulation.py
```diff
@@ -1009,6 +1010,22 @@ def _validate_monitor_size(self) -> None:
                 f"a maximum of {MAX_SIMULATION_DATA_SIZE_GB:.2f}GB are allowed."
             )
 
+        # Some monitors store much less data than what is needed internally. Make sure that the
+        # internal storage also does not exceed the limit.
+        for monitor in self.monitors:
+            num_cells = self._monitor_num_cells(monitor)
+            # intermediate storage needed
+            internal_data_gb = monitor._storage_size_solver(num_cells=num_cells, tmesh=self.tmesh)
```
tylerflex, line 1018: would be a bit easier to understand if we named the variables like
```py
internal_data_bytes = monitor._storage_size_solver(num_cells=num_cells, tmesh=self.tmesh)
internal_data_gb = internal_data_bytes / 2**30
```
also in the other instance where the 2 ** 30 is divided out.

tylerflex (approved):
one minor comment but otherwise looks good.

tomflexcompute (approved):

Issue #1148, opened 2023-09-08
improve docs


tylerflex:
Some things we could do to make this better:

### Landing page
- [x] Toggle between command line and python instructions for installation and configure.
- [x] (possibly?) Toggle between linux / mac and windows instructions
- [ ] Add link to colab notebook with the minimal example and API key loaded as environment variable.
- [X] Add nice banners

### Aesthetics
- [x] Improve overall theme.
- [X] Make the font better
- [X] Improve general page styling so it's more readable
- [x] Add emojis, generally make the docs look nicer.
- [X] wider columns for titles and code.
- [x] make table of contents display more logical
- [x] Find a dark theme logo that works - I think important because I and I assume others, exclusively work in dark themes. Currently chatting with Qing. (I can modify the existing one easily if approved)
- [ ] Cooler intro page and pictures
- [x] Make toctrees nicer

### API Documentation Improvements
From @e-g-melo As an user of documentation, I think that it would help a lot if 
- [ ] The user can find more context when reading though the class definitions 
- [ ] Including equations when needed 
- [ ] Schematics to show geometric parameters
- [ ] An example on how to use that object in a simulation.
- [ ] Link to examples and tutorials
- [x] We need to sort out the way the attributes are propagated.
- [ ] Cleaner API viewer
- [X] Solve the hierarchical class method search issue https://github.com/flexcompute/tidy3d/issues/1197
- [X] Hide irrelevant private methods on API
- [X] ~Sort out the material library toctree situation~ TODO autosummary generate restructure in this case re fundamental hierarchy
- [X] Improve the API tab search and hierarchy
- [ ] Make nice illustration of medium property types, ie anisotropic isotropic for the uninitated
- [ ] Make nice illustrations of all the sources and geometrically how they are represented
- [ ] Standalone examples for most common class methods etc.
- [x] Solve the class page index issue on the right, especially for Simulation that it became quite big.
- [X] Any other way to solve the inheritance method propagation issue apart from the manual autosummary? Maybe there's no alternative and this is it. Completeness by documenting all base classes on absract/ done
- [x] Fix the m2r2 situation. This may require a patch of that repository.
- [x] Backup a main and develop branch of the `tidy3d-docs` just in case which might get overwritten
- [ ] Potentially work on the adjoint docstrings depending on what Tyler wants re his restructure
- [x] Fix all the broken links from the 2.5 structure

### Functionality
- [x] how to build locally.
- [ ] remove all warnings.
[...]

PR #1278, opened 2023-12-01
The big documentation and development upgrade

daquinteroflex:
This is the one we will merge whenever we want to. I am only perfecting things at this point.

Supersedes all others as they are so coupled changes and are hard to break down into smaller bits eg. https://github.com/flexcompute/tidy3d/pull/1238

First stage implementation of https://github.com/flexcompute/tidy3d/issues/1148

Could I propose maybe we do this merge independently of 2.6 but before 2.6? Or the documentation upgrade becomes 2.6 and the further upgrades become 2.7?

I just realised I should open this only when pre/2.6 branch is open and maybe only when 2.5 is fully done? Seems like an easy base change whenever we decide.


tylerflex:
Thanks @daquintero. What I was imagining was something like this. Let me know if it makes sense to you or it complicates things. I dont have a strong opinion, just was under this impression:

* We just released 2.5.0rc3.
* Pretty soon, we will merge `pre/2.6` into `develop` and release `2.6.0`.
* Then we merge this PR into develop (which will look a lot like current `pre/2.6`)
* Then we can release 2.6.0rc1 and start testing with the new docs synced to `tidy3d-docs-demo`?

Or is this different from the plan you had in mind? If we go this route, the best will be to rebase against the pre/2.6 branch I think.

tylerflex:
Sorry i meant pre/2.5

daquinteroflex:
As of the current setup it gets live updated from this branch on this webside: https://flexcompute-tidy3d-docs-demo.readthedocs-hosted.com/en/repo_merge_no_history/

daquinteroflex:
Because rebasing was very painful upgrading from here https://github.com/flexcompute/tidy3d/pull/1319

PR #1279, opened 2023-12-01
Change definition of FluxTime and DFT fields with complex fields

caseyflex:
This PR:

- Changes FluxTime and DFT calculation to only use real part of fields
- Updates sim.complex_fields based on nonlinear models


momchil-flex:
So in the current state, the following would be happening:

- FieldTimeMonitor returns complex fields just in case, but we make it clear that the real fields would typically be the ones of interest, and the imaginary fields are not used in many derived quantities.
- FieldMonitor returns fields that are the DFT of Real(E(t))
- FluxTimeMonitor returns flux based on Real(E(t)) only; FieldTimeMonitor.flux is also correspondingly modified.
- Time and frequency domain fluxes are then consistent with one another and the total power is independent of whether real or complex fields are used.

tidy3d/components/time.py
```diff
@@ -63,7 +63,9 @@ def spectrum(
             Time step to weight FT integral with.
             If array, use to weigh each of the time intervals in ``times``.
         complex_fields : bool
-            Whether time domain fields are complex, e.g., for Bloch boundaries
+            Whether time domain fields are complex, e.g., for Bloch boundaries.
+            Note: this parameter is no longer used, as the DFT definition for complex fields
+            has changed to only use the real part.
```
momchil-flex, line 68: When removing this, let's add a note that the real part of the time signal is used only.

PR #1284, opened 2023-12-04
Error if any frequencies are smaller than 1e5 Hz

momchil-flex:
The basic change is very simple: previously, we only validated that `ModeSolver.freqs` are larger than 1e5. Now we also do it for monitor frequencies, and source central frequency.

However, this required quite a bit of tests changes since we were using frequencies of order 1 often. It's probably anyway better to fix this in tests too though...

I also removed all 1.x simulations that the updater was checking for. Some of the earlier ones had freqs of order 1 too. And in general, I don't think we still need to check compatibility with Tidy3D v1.


tylerflex (approved):
looks good to me, thanks

Issue #1287, opened 2023-12-05
Enabling medium index option for field projection monitors


tomflexcompute:
Current field projection assumes the medium of the projected region has the same index as that at the location of the monitor. Some users would like to have the option to manually define the refractive index for the projected region. This is not physically rigorous but can be convenient to have as a good approximation in some cases.

momchil-flex:
How would this work practically? For example, after computing the far field in the homogeneous medium, we can apply a single interface scattering matrix (i.e. disregarding back-reflections that could in practice happen)? This can even be done by the user already fairly straightforwardly I think, if they need it right now?

momchil-flex:
FYI: https://optics.ansys.com/hc/en-us/articles/360034395194

tomflexcompute:
> How would this work practically? For example, after computing the far field in the homogeneous medium, we can apply a single interface scattering matrix (i.e. disregarding back-reflections that could in practice happen)? This can even be done by the user already fairly straightforwardly I think, if they need it right now?

Something like this is even better. I was just thinking about something more rudimentary, which is what the user was asking for. See [here](https://optics.ansys.com/hc/en-us/articles/360034410214-farfieldexact-Script-command).

momchil-flex:
Actually the link I sent above discusses the difference between the two approaches mentioned here. :)

tomflexcompute:
Artificially setting the index to a different value certainly will cause some artificial results. In some cases however, it can still be a good approximation. For example, imagine a large grating coupler with a few microns of oxide cladding. The simulation domain can only contain oxide to save cost but when projecting the near field field to far away, say hundreds of microns, setting the `background_index=1` will be much more accurate then `background_index=n_oxide`. Sure there will be some small offset due to the finite oxide thickness and some intensity difference due to reflection, but for a quick estimation for the beam spot position and size it's actually pretty good.

momchil-flex:
Yeah, so let's just do that change which should be pretty small and not worry about the "Fresnel correction" for now (which can be a post-processing script eventually).

tomflexcompute:
> Yeah, so let's just do that change which should be pretty small and not worry about the "Fresnel correction" for now (which can be a post-processing script eventually).

Sounds good. One user is waiting to use it so would be great if we can get it in a near future release. Thanks @shashwat-sh in advance!

PR #1291, opened 2023-12-06
Allow field projections through a manually-specified medium

shashwat-sh:
Backend PR: https://github.com/flexcompute/tidy3d-core/pull/474
Addresses [this](https://github.com/flexcompute/tidy3d/issues/1287) issue.

### DIFF
--- a/tidy3d/components/monitor.py
+++ b/tidy3d/components/monitor.py
@@ -11,6 +11,7 @@ from .validators import assert_plane, validate_freqs_not_empty, validate_freqs_m
 from .base import cached_property, Tidy3dBaseModel
 from .mode import ModeSpec
 from .apodization import ApodizationSpec
+from .medium import MediumType
 from .viz import ARROW_COLOR_MONITOR, ARROW_ALPHA
 from ..constants import HERTZ, SECOND, MICROMETER, RADIAN, inf
 from ..exceptions import SetupError, ValidationError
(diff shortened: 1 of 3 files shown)


tidy3d/components/monitor.py
```diff
@@ -11,6 +11,7 @@
 from .base import cached_property, Tidy3dBaseModel
 from .mode import ModeSpec
 from .apodization import ApodizationSpec
+from .medium import MediumType
```
momchil-flex, line 14: @tylerflex do you think importing from medium here could be problematic in the future? I'm not sure anymore if we ever defined the heirarchy of what's allowed to import from what.

momchil-flex (approved):

tidy3d/components/monitor.py
tylerflex: Shoot, i don't remember now but this is giving me Deja vu remembering some annoying refactorings. So I'm not sure. My feeling is we probably shouldn't import it here if possible. is there a way around it? Im on mobile and can't look now in detail

tomflexcompute (approved):
Thanks for the quick implementation!

tidy3d/components/monitor.py
momchil-flex: Bump

shashwat-sh: One idea, to allow for frequency dependence, is a `projection_index` field that accepts either a scalar, or an array whose size is validated to match the size of `monitor.freqs`, so that the user can provide a constant or one value per frequency?

shashwat-sh: That only enables this feature for client-side projections - I think for server side, we need to modify `AbstractFieldProjectionMonitor` anyway, because the solver is unaware of `FieldProjector`

shashwat-sh: At one point we had discussed getting rid of client-side projections altogether, especially if the plan is to provide a standalone server-side projector. Is this still the the goal? If so, maybe the clunkiness is tolerable since it's mostly `field_projection.py` that's clunky now?

momchil-flex: It also seems more annoying for the user if it's not taking medium directly. Let's go with medium.

momchil-flex:
Squash and ship as far as I'm concerned. Could you also `git checkout pre/2.5 -- tests/sims` before your final commit? I already updated the version to 2.5.0 there so the new files you're commiting as 2_5_0rc3 actually aren't right.

PR #1297, opened 2023-12-06
Return 0 spectrum if no times are above the dft cutoff

momchil-flex:
A few times we've had simulations error out because the `times_cut` array is empty. The only way I can see this happening is if the source amps are just 0 (and for some reason the 0/0 returns 0 I guess?). Anyway, I think this should at least prevent the simulations from erroring even if the results would be meaningless  - or all 0s.


caseyflex (changes requested):
Lines 81-83 were supposed to accomplish that, but looking at them again, using np.allclose is suspect -- it will be using atol=1e-8, which is quite arbitrary. Maybe replace this with np.equal, in addition to adding your lines just in case.

caseyflex:
```assert not np.all(np.equal(td.GaussianPulse(amplitude=1e-10, freq0=1, fwidth=.1).spectrum(times=np.linspace(0, 10, 100), freqs=np.linspace(0, 10, 100), dt=.1), 0))```

this currently asserts False, but replacing np.allclose with np.equal should change it

PR #1300, opened 2023-12-06
On the fly vtk import

dbochkov-flexcompute:
For some reason couldn't make it work with plain global variables, so ended up using a dict that contains relevant info. One issue with it (and probably in general using this on the fly approach) is that we can't use imported modules members in definition of function signatures. For this particular case, it is not critical, but could be for others.

### DIFF
--- a/tests/test_data/_test_datasets_no_vtk.py
+++ b/tests/test_data/_test_datasets_no_vtk.py
@@ -24,7 +24,7 @@ def test_triangular_dataset_no_vtk(tmp_path):
     # double check that vtk was not imported
     from tidy3d.components.types import vtk
 
-    assert vtk is None
+    assert vtk["mod"] is None
 
 
 @pytest.mark.usefixtures("hide_vtk")
@@ -34,4 +34,4 @@ def test_tetrahedral_dataset_no_vtk(tmp_path):
     # double check that vtk was not imported
     from tidy3d.components.types import vtk
 
-    assert vtk is None
+    assert vtk["mod"] is None
--- a/tests/test_data/test_datasets.py
+++ b/tests/test_data/test_datasets.py
@@ -124,7 +124,7 @@ def test_triangular_dataset(tmp_path, ds_name):
     assert tri_grid.bounds == ((0.0, 0.0, 0.0), (1.0, 0.0, 1.0))
     assert np.all(tri_grid._vtk_offsets == np.array([0, 3, 6]))
 
-    if vtk is None:
+    if vtk["mod"] is None:
         with pytest.raises(Tidy3dImportError):
             _ = tri_grid._vtk_cells
         with pytest.raises(Tidy3dImportError):
@@ -137,7 +137,7 @@ def test_triangular_dataset(tmp_path, ds_name):
         _ = tri_grid._vtk_obj
 
     # plane slicing
-    if vtk is None:
+    if vtk["mod"] is None:
         with pytest.raises(Tidy3dImportError):
             _ = tri_grid.plane_slice(axis=2, pos=0.5)
     else:
@@ -154,7 +154,7 @@ def test_triangular_dataset(tmp_path, ds_name):
             _ = tri_grid.plane_slice(axis=0, pos=2)
 
     # clipping by a box
-    if vtk is None:
+    if vtk["mod"] is None:
         with pytest.raises(Tidy3dImportError):
             _ = tri_grid.box_clip([[0.1, -0.2, 0.1], [0.2, 0.2, 0.9]])
     else:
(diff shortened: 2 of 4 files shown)


daquinteroflex:
I was looking again into this with more detail. For the sake of brainstorming, what if we had this instead?

rather than the wrapper
```
def verify_time_import():
    try:
        import time
    except ImportError:
        raise Warning("That didn't work")
        

def test_this_works():
    verify_time_import()
    import time # Here we import and write code as usual?
    print(time.time())
    
test_this_works()
```

dbochkov-flexcompute:
@daquinteroflex thanks for looking into timings so carefully. When I run these scripts on my side locally I get the following results:

For `python -X importtime import_tidy3d_script.py`:

1. Before this PR:
```
import time:       259 |     218696 |         trimesh
...
import time:      2427 |     245991 |         vtk
import time:      2284 |     671884 |       tidy3d.components.types
...
import time:      2302 |    1411318 | tidy3d
```

2. After this PR:
```
import time:       258 |     222313 |         trimesh
import time:      2345 |     433283 |       tidy3d.components.types
...
import time:      2420 |    1224365 | tidy3d
```

For `python time_import_2.py`:
1. Before this PR: `average tidy3d load time = 1.571178252696991`
2. After this PR: `average tidy3d load time = 1.2565033292770387`

So, it seems these results are consistent with each other. That is, importing `vtk` adds about 0.25-0.3s on my machine. However, it differs significantly with your results, which basically says that there is no difference. Can you double check that your environment does have `vtk` installed? When I do `pip uninstall vtk` and try `python -X importtime import_tidy3d_script.py` I get the following
```
import time:        81 |         81 |         vtk
```
which is similar to what you get.

Anyway, 0.25-0.3s still seems quite far from from what @Momchil observed

daquinteroflex:
Hi Daniil,

Thanks for this, you were right, I found there was something wrong on my setup. I was being stupid. I didn't do `pip install -r requirements/dev.txt` and did what I've been used to more recently which was `pip install .[dev]`. Turns out for some reason it didn't install `vtk` and I should have checked this. For reference: 


These were the commands I ran on both this PR and pre-PR which have worked on my poetry setup and expected it to be the same on the older version.
```
git checkout daniil/onthefly-vtk-import
micromamba create -n tidy3d_pre2.5 python=3.10 -c conda-forge
micromamba activate tidy3d_pre2.5
python3 -m pip install -e .[dev]
```
and
```
micromamba create -n tidy3d_onthefly python=3.10 -c conda-forge
micromamba activate tidy3d_onthefly
python3 -m pip install -e .[dev]
python time_import.py
```

However, to my surprise, because I checked the `requirements/dev.txt` installation also included `vtk.txt`. It turns out when I ran the above `vtk` didn't get installed on either virtual environment, despite them being globally "imported" when I ran those import timing decomposition tests as above.

Once I properly did the installation running `pip install .[vtk]` then I reproduced the same timing as you mention above:
```
(tidy3d_pre2) daquintero@dxps:~/flexcompute/debug/tidy3d_pre25$ python time_import_2.py 
average tidy3d load time = 1.3655261087417603
(tidy3d_onthefly) daquintero@dxps:~/flexcompute/debug/tidy3d_onthefly$ python time_import_2.py 
average tidy3d load time = 1.0845233035087585
```

daquinteroflex:
Sorry I'm being stupid again on my function wrapper idea - for a moment I thought it'd be different but the sequence is the same. 

What I've proposed is the same as the wrapper, the only difference is that rather than calling the dictionary, you call the import directly within the function. 

So you could keep everything potentially in the wrapper or not within the function and simply use the import vtk within it too without the memory access cost of a new variable maybe?

daquinteroflex:
Apologies for the earlier edits, I think I needed a bit of a break after being connected to the computer for a bit without thinking too clearly.

So I've done a bit of testing, mainly I was curious about what was the time cost of a function call with the import. The memory cost is minimal if we assume the test setup to be valid.

In a directory, create `get_vtk_module.py`:

```
def get_vtk():
    try:
        vtk = __import__("vtk")
        return vtk
    except ImportError:
        raise Warning("That didn't work")
```

This gets imported as a function call in the `call_get_vtk.py`:

```
from get_vtk_module import get_vtk

get_vtk()
```

and then the timing test setup over 100 runs:
```
import subprocess
import time

n = 100
get_vtk_call_time = 0
vtk_import_time = 0

for i in range(n):
    # Timing get_vtk function call
    start_time = time.time()
    subprocess.call(["python", "call_get_vtk.py"])
    get_vtk_call_time += time.time() - start_time

    # Timing simple import vtk
    start_time = time.time()
    subprocess.call(["python", "-c", "import vtk"])
    vtk_import_time += time.time() - start_time

print("Average call time for get_vtk: {}".format(get_vtk_call_time / n))
print("Average import time for vtk: {}".format(vtk_import_time / n))
```

when I run it:
```
(tidy3d_onthefly) daquintero@dxps:~/flexcompute/debug/import_time_test$ python test_compare.py 
Average call time for get_vtk: 0.18276306867599487
Average import time for vtk: 0.18201354026794433
```
So it's a minimal cost, but it maybe reimporting again is fine within the actual functions rather than variables?

Apologies for the mess of messages earlier! I'm going to call it a day as I think I need to unplug from the terminal.

momchil-flex:
> In case we decide to stick with the wrapper approach, we could maybe slightly improve that by replacing vtk dictionary with a vtk config similar to tidy3d config https://github.com/flexcompute/tidy3d/blob/pre/2.5/tidy3d/config.py.

All things considered maybe yeah the current approach with this improvement is good enough? I'm even fine with merging as is for now - I just want to get this fix in, even if it's not the prettiest under the hood.

Issue #1301, opened 2023-12-07
pip install tidy3d failing on M1 Mac


joamatab:
due to gdstk not working on M1 or M2, tidy3d does not work on Mac M1 or M2

https://github.com/heitzmann/gdstk/issues/199

@tylerflex 
@lucas-flexcompute

tylerflex:
For reference, I am using `gdstk == 0.9.37` on a Mac M2 and it works for me. Any version after that seems to break. Also `tidy3d` will generally work as long as you aren't using any methods that require `gdstk`.

Perhaps we should pin a `gdstk <= 0.9.37` requirement for Mac users until it gets sorted out. @daquinteroflex what do you think?

joamatab:
I think pinning an old version of a tool is bad practice :) For example, in gdsfactory we use some of the new gdstk features that were released post 0.9.37, such as remapping layers.

can you guys try this solution on gdstk and see if we can make it work on the latest versions?

 https://blog.tim-smith.us/2015/09/python-extension-modules-os-x/

tylerflex:
@joamatab just to clarify, `pip install tidy3d` should work on M2, just not `pip install tidy3d[gdstk]`, correct?

tylerflex:
Fixed in gdstk 0.9.49

tylerflex:
@daquinteroflex let's pin that version or above in our upcoming 2.6 releases? thanks!

PR #1303, opened 2023-12-07
FIX: Quick patch for gdstk on mac

daquinteroflex:
Re https://github.com/flexcompute/tidy3d/issues/1301 if we want it. I haven't been able to test it as I don't have a Mac, but the syntax is correct I think

### DIFF
--- a/requirements/gdstk.txt
+++ b/requirements/gdstk.txt
@@ -1,3 +1,5 @@
 # gdstk
 
-gdstk
+gdstk<=0.9.37; platform_system != "Darwin"
+gdstk; platform_system != "Darwin"
+


joamatab:
I think pinning an old version of a tool is bad practice :) For example, in gdsfactory we use some of the new gdstk features that were released post 0.9.37, such as remapping layers.

can you guys try the solution Lucas proposed on gdstk and see if we can make it work on the latest versions?

 https://blog.tim-smith.us/2015/09/python-extension-modules-os-x/

tylerflex:
Thanks for the PR @daquinteroflex.  @lucas-flexcompute will try to debug this next week (with my M2) so maybe we'll hold off on merging this PR until then. 

Apparently the installation works if one uses conda to install gdstk. I wonder if another approach could be to simply

```py
try:
    import gdstk
except ImportError:
```
and write a detailed message suggesting trying to install gdstk with conda if someone is using apple silicon?

tylerflex:
However, perhaps this handling is a bit better suited to be in the `gdstk` package where the error occurs, not sure.

daquinteroflex:
Yeah I agree too!

PR #1307, opened 2023-12-10
Allow bend_radius to be negative, but not close to zero

momchil-flex:
We actually allow for negative bend radius, so we need to validate the magnitude.

### DIFF
--- a/tidy3d/components/mode.py
+++ b/tidy3d/components/mode.py
@@ -1,6 +1,7 @@
 """Defines specification for mode solver."""
 
 from typing import Tuple, Union
+from math import isclose
 
 import pydantic.v1 as pd
 import numpy as np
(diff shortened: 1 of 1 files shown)


tidy3d/components/mode.py
```diff
@@ -117,7 +118,14 @@ class ModeSpec(Tidy3dBaseModel):
     def bend_axis_given(cls, val, values):
         """check that ``bend_axis`` is provided if ``bend_radius`` is not ``None``"""
         if val is None and values.get("bend_radius") is not None:
-            raise SetupError("bend_axis must also be defined if bend_radius is defined.")
+            raise SetupError("'bend_axis' must also be defined if 'bend_radius' is defined.")
+        return val
+
+    @pd.validator("bend_radius", always=True)
+    def bend_radius_not_zero(cls, val, values):
+        """check that ``bend_axis`` is provided if ``bend_radius`` is not ``None``"""
```
weiliangjin2021, line 126: Fix docstring here?

weiliangjin2021 (approved):
Forgot too last time

tidy3d/components/mode.py
momchil-flex: Ah yes thanks.

PR #1308, opened 2023-12-11
Add the calculation of the dispersion parameter for group velocity

### LINKED ISSUE #1294: Mode solver data support
## Current
Our mode solver only support the following data:
- Mode field
- Effective index
- Group index
## Expect
We hope the solver can offer the following data:
- Group velocity [from Lukas book]:
  - $v_p(\lambda)=\frac{c}{n_{eff}}$, $v_g(\lambda)=\frac{c}{n_g}$
- Group velocity dispersion parameter $D$ [from Lukas book]:
  - $D(\lambda)=-\frac{\lambda}{c}\frac{d^2n_{eff}}{d\lambda^2}$
- Group velocity dispersion parameter $\beta_2$ [from Agrawal book]:
  - $\beta_2=\frac{1}{c}\left(2\frac{dn}{d\omega}+\omega\frac{d^2n}{d \omega^2} \right)$
  - $D=\frac{\beta_1}{\lambda}=-\frac{2\pi c}{\lambda^2}\beta_2=-\frac{\lambda}{c}\frac{d^2n}{d\lambda^2}$

lucas-flexcompute:
This fixes #1294.


tidy3d/components/data/monitor_data.py
```diff
@@ -1392,13 +1413,21 @@ class ModeData(MonitorData):
         description="Complex-valued effective propagation constants associated with the mode.",
     )
 
-    n_group_raw: ModeIndexDataArray = pd.Field(
+    n_group_raw: GroupIndexDataArray = pd.Field(
```
tylerflex, line 1416: are there any potential side effects from changing this type? For example, in the json files, I think we store a tag with the DataArray type name, I wonder if we need any changes to the updater to convert these types if coming from older versions. On the other hand, if our IO tests pass, then it's probably working properly? just something that caught my attention as a possible issue.

tylerflex (commented):
Thanks @lucas-flexcompute just a couple minor comments:
1. whether we need the `alias=description`
2. whether changing the data array type could cause any issues with loading from jsons from older versions.

tidy3d/components/data/dataset.py
```diff
@@ -354,13 +355,21 @@ class ModeSolverDataset(ElectromagneticFieldDataset):
...
+    n_group_raw: GroupIndexDataArray = pd.Field(
         None,
         alias="n_group",
         title="Group Index",
         description="Index associated with group velocity of the mode.",
     )
 
+    dispersion_raw: ModeDispersionDataArray = pd.Field(
+        None,
+        alias="dispersion",
```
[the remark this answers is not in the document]
lucas-flexcompute: The alias was introduced because there was an original `n_group` field, which we replaced by `n_group_raw` so that the user would get a warning when accessing a missing `n_group` attribute. By then, the alias was required to load old json/hdf5 data with the `n_group` tag.

I kept the same logic for dispersion: issuing a warning when accessing the possibly empty attribute and storing the actual data in a `_raw` field. I guess the alias is not really necessary here, that's an oversight.

tidy3d/components/data/monitor_data.py
lucas-flexcompute: This should not be a problem. Older files have `n_group` as a `ModeIndexDataArray` but they can be loaded as `GroupIndexDataArray` because the only difference is in the description. I've just tested with a file saved in the base branch and it loaded correctly as a `GroupIndexDataArray` even though the hdf5 data has `ModeIndexDataArray` as the type from `n_group_raw`.

tylerflex: sounds good

tylerflex (approved):

tidy3d/components/data/dataset.py
```diff
@@ -388,6 +396,20 @@ def n_group(self):
             )
         return self.n_group_raw
 
+    @property
+    def dispersion(self):
```
tylerflex, line 400: can you add a type hint for return?

tidy3d/components/data/monitor_data.py
```diff
@@ -1392,13 +1413,20 @@ class ModeData(MonitorData):
...
 
+    dispersion_raw: ModeDispersionDataArray = pd.Field(
+        None,
+        title="Dispersion",
+        description="Dispersion parameter for the mode.",
+        units=PICOSECOND_PER_NANOMETER_PER_KILOMETER,
+    )
+
     @property
     def n_eff(self):
```
tylerflex, line 1431: maybe add a type hint here too while you're at it, thanks

lucas-flexcompute: Should this be `ModeIndexDataArray` even though it's not complex? Otherwise we need a new data type or an overly-generic `xr.DataArray`.

tylerflex: oh wait, I might be a bit confused. maybe just `xr.DataArray` is ok. or `ModeIndexDataArray` is probably ok.. up to you.

tidy3d/components/data/dataset.py
```diff
@@ -354,31 +355,38 @@ class ModeSolverDataset(ElectromagneticFieldDataset):
...
+        alias="n_group",  # This is for backwards compatibility only when loading old data
         title="Group Index",
         description="Index associated with group velocity of the mode.",
     )
 
+    dispersion_raw: ModeDispersionDataArray = pd.Field(
+        None,
+        title="Dispersion",
+        description="Dispersion parameter for the mode.",
+        units=PICOSECOND_PER_NANOMETER_PER_KILOMETER,
```
momchil-flex, line 369: This is a crazy unit lol, is this standard?

tidy3d/components/data/dataset.py
```diff
@@ -388,6 +396,20 @@ def n_group(self):
             )
         return self.n_group_raw
 
+    @property
+    def dispersion(self) -> ModeDispersionDataArray:
+        """Dispersion parameter.
+
+        D = -lda / C_0 d2n/d(lda)2
```
momchil-flex, line 403: I think @daquinteroflex should make a note to update this with a proper equation in the style of his api docs update? Or just tell us how to do it already directly here? We already have equations like 

```
    .. math::

        \\epsilon(f) = \\epsilon_\\infty + \\sum_i
        \\frac{\\Delta\\epsilon_i f_i^2}{f_i^2 - 2jf\\delta_i - f^2}
```

is this how this should be reformatted?

momchil-flex (approved):
Thanks, approving in general, just small comments - also could you rebase to develop branch now and update the CHANGELOG accordingly to put the item in Unreleased?

tidy3d/components/data/dataset.py
daquinteroflex: Yep adding it this way is correct! Here you have some examples: this line https://github.com/flexcompute/tidy3d/blob/2f147a3e28195097203746ff8621913906a586bb/tidy3d/components/medium.py#L2042 rendered [here](https://flexcompute-tidy3d-docs-demo.readthedocs-hosted.com/en/repo_merge_no_history/api/_autosummary/tidy3d.PoleResidue.html#tidy3d.PoleResidue)

daquinteroflex: I'll update it on my branch in any case!

lucas-flexcompute: Sounds good. I was not sure if we wanted to favor the web rendering or the terminal help. Changed!

tidy3d/components/data/dataset.py
lucas-flexcompute: Yes... https://www.rp-photonics.com/group_velocity_dispersion.html

momchil-flex: Oh I see I guess it makes sense if you think about long-range fiber optics.

tidy3d/components/data/dataset.py
momchil-flex: Hmm what is the `r` for in front of the `"""` of the docstring?

lucas-flexcompute: That's for a raw string, so that we don't have to escape the backslashes.

momchil-flex:
Alright, let's ship it, just fix the changelog and merge frontend and backend, thanks!

PR #1309, opened 2023-12-12
Print cost breakdown in estimate_cost if there is mode or post-procesing charge

momchil-flex:
Note: post-processing charge is currently only incurred by field projection monitors.


tylerflex (approved):

tylerflex:
should test this in notebooks before release?

momchil-flex:
Yeah I already did manually but I'm also planning to rerun the ones that have `estimate_cost` in them after this is merged.

PR #1312, opened 2023-12-12
Mode solver eps_complex check at central frequency only

momchil-flex:
Whether `eps_model` works with an array of frequencies is not consistent among the various media. Earlier I had an issue with fully anisotropic which I fixed, but now I see that the same is true for (some?) custom media. Since fully clarifying that may be a mess, and for this particular check looking at the central freqeuncy is probably fine, I propose this change for now.

Of course I could also iterate over `freqs`, but I want this check to be very fast even in cases with e.g. large custom media / many freqs, so I chose not to.

### DIFF
--- a/tidy3d/plugins/mode/mode_solver.py
+++ b/tidy3d/plugins/mode/mode_solver.py
@@ -4,6 +4,7 @@ invariance along a given propagation axis.
 
 from __future__ import annotations
 from typing import List, Tuple, Dict
+from math import isclose
 
 import numpy as np
 import pydantic.v1 as pydantic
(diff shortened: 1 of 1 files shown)


tidy3d/plugins/mode/mode_solver.py
```diff
@@ -643,11 +643,12 @@ def _has_fully_anisotropic_media(self) -> bool:
...
         """Check if there are media with a complex-valued epsilon in the plane of the mode at the
-        mode solver freqs. A separate check is done inside the solver, which looks at the actual
+        central frequency. A separate check is done inside the solver, which looks at the actual
         eps and mu and uses a tolerance to determine whether to use real or complex fields, so
         the actual behavior may differ from what's predicted by this property."""
+        freq0 = self.freqs[self.freqs.size // 2]
         for int_mat in self._intersecting_media:
-            max_imag_eps = np.amax(np.abs(np.imag(int_mat.eps_model(np.array(self.freqs)))))
+            max_imag_eps = np.amax(np.abs(np.imag(int_mat.eps_model(freq0))))
             if not np.isclose(max_imag_eps, 0):
```
weiliangjin2021, line 652: `np.isclose` to `math.isclose`?

weiliangjin2021 (commented):
Could you remind me again what is the issue with fully anisotropic medium?

To be safe, we might as well check 3 frequenices: min, mean, and max?

momchil-flex:
> Could you remind me again what is the issue with fully anisotropic medium?

I fixed that here: https://github.com/flexcompute/tidy3d/pull/1290

> To be safe, we might as well check 3 frequenices: min, mean, and max?

Updated and also using math.isclose.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -642,14 +643,16 @@ def _has_fully_anisotropic_media(self) -> bool:
 
     @cached_property
     def _has_complex_eps(self) -> bool:
-        """Check if there are media with a complex-valued epsilon in the plane of the mode at the
-        mode solver freqs. A separate check is done inside the solver, which looks at the actual
+        """Check if there are media with a complex-valued epsilon in the plane of the mode.
+        A separate check is done inside the solver, which looks at the actual
         eps and mu and uses a tolerance to determine whether to use real or complex fields, so
         the actual behavior may differ from what's predicted by this property."""
+        check_freqs = np.unique([np.amin(self.freqs), np.amax(self.freqs), np.mean(self.freqs)])
```
weiliangjin2021, line 650: Just one last minor points: if len(self.freqs) is 1 or 2, check_freqs can be just 1 or 2 as well.

momchil-flex: There's a `np.unique` that should handle that?

weiliangjin2021: Ah yeah

weiliangjin2021 (approved):

PR #1315, opened 2023-12-13
2.6 branch

tylerflex:


daquinteroflex:
PRs that need to be merged only when the release is happening (in order):

1. Notebooks update to "latest" https://github.com/flexcompute/tidy3d-notebooks/pull/34/files
2. FAQ update to "latest" https://github.com/flexcompute/tidy3d-faq/pull/12
3. Need to update submodules now on tidy3d frontend, make sure docs build.

daquinteroflex:
So the problem is that the FAQs develop branch is linked to the helpcenter, which would break all links. For notebooks, actually true, I think I had the intention to check that any new notebooks had their links updated before the release. But we could do that in two PRs

tylerflex:
ok, so this will only be an issue for 2.6? (not 2.7+?)

daquinteroflex:
Yeah exactly, that's why thought to help out organizing the PR order

Issue #1191, opened 2023-10-03
Reduce simulation file size for mode solver


dbochkov-flexcompute:
In case of large simulations with custom media the total size of simulation file can be quite big. When calling server side mode solver, the entire simulation size is uploaded, however, typically only a small subset of data is required. Maybe we could introduce a way that uploads only the required data.

tylerflex:
I wonder if (in the spirit of the `.to_gds` stuff) there could be some way to "export" part of a simulation. For example:
```
sim.subsection(region: Box) -> Simulation.   # generate a simulation containing only the `region`.
```

do you have something similar in heat solver?

momchil-flex:
Assigning @dbochkov-flexcompute so that someone is responsible for this; it's a long-term thing though. I put it for 2.6 for now.

dbochkov-flexcompute:
I have a working prototype and now thinking how to split it between `Simulation` and mode solver plugin
- First option is to just keep everything in mode solver plugin. In this case, since it is used for a very specific purpose, I would not actually care to change simulation dimensions, just remove irrelevant structures/custom medium points and ensure that the same discretization grid is used. This makes it easy to deal with symmetry options, since we are not changing simulation `center`/`size`.
- Second option is to implement a standalone `Simulation.subsection(region: Box) -> Simulation` per Tyler suggestion. This could be potentially useful elsewhere eventually, but requires ironing out a couple of things. First, do we just transfer boundary conditions directly? Perhaps, one way could be to have an optional argument to specify new boundary conditions if needed. Second, have to be careful while dealing with symmetry. The approach that I could see there is the following. If subsection box doesn't go inside symmetry regions, just return a new simulation without a symmetry and center/size taken from the subsection box. If the subsection box does go into symmetry region, automatically expand it symmetrically so that new simulation center coincides with the old simulation center.

Any thoughts @tylerflex @momchil-flex ?

tylerflex:
I'd say overall just depends on how useful this Simulation.subsection would be and if it's worth your time.

Regarding the specific implementation of that:
> First, do we just transfer boundary conditions directly? Perhaps, one way could be to have an optional argument to specify new boundary conditions if needed.

I'd say transfer BCs with option to overwrite them yea.

> Second, have to be careful while dealing with symmetry. The approach that I could see there is the following. If subsection box doesn't go inside symmetry regions, just return a new simulation without a symmetry and center/size taken from the subsection box. 

That sounds reasonable on me.

[...]

PR #1316, opened 2023-12-15
Reduced sim mode solver

dbochkov-flexcompute:
Implements https://github.com/flexcompute/tidy3d/issues/1191. That is, there is a new simulation method `.subsection(region: Box, ...)` that can be used to create a smaller simulation based on original one. 
- Most of simulation fields are inherited, but could be optionally overridden
- Grid can be requested to be the same as in the parent simulation (`grid_spec="identical"`). In this case it's converted into a custom grid and the requested region is expanded to match parent grid lines. One small issue here is that if the parent grid already matches the requested region, than due to numerical precision one additional cell might be included. But that doesn't seem cause any real issues.
- Similarly, due to numerical precision issues, even when child grid is essentially the same as the parent grid, some of structures might become slightly misaligned with grid lines. Most prominently it shows when calling `ModeSolver._solver_eps()` of local mode solver. That is, effective boundaries of structures might differ by one pixel between parent and child simulations.
- Regarding inheriting symmetry: we discussed in https://github.com/flexcompute/tidy3d/issues/1191 that probably we can just not allow selecting a not compatible region which is not centered around symmetry plane. But I realized this could happen in the background of a remote mode solver call and that would cause an error. So, I implemented it in a way so that an incompatible region is automatically expanded to be centered around a symmetry plane.
- `.subsection(region: Box, ...)` is used internally in mode solver web api call and whether it is used or not is controlled by flag `reduce_simulation`. I set it by default to `True`, that is, all mode solver calls will use a reduced simulation. We can change it either to `False`, or make it so that only simulation containing custom mediums are reduced and also display a warning/info notifying that this is done. What do you think?
- This does significantly help to reduce data transfer for remote mode solver. For example, in this notebook https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/096b84d99d301de4e3690fb6b4aa373064a2bb91/docs/source/notebooks/ThermallyTunedRingResonator.ipynb I couldn't really use remote mode solver because each solve would require around 500mb of upload, now it's less than 1mb. I also expect it to be useful for regular optic simulations in the context of heat solver. I do use it in the same notebook to remove irrelevant structures from optic simulation, but I anticipate more impactful usage eventually. (in this notebook, we are already restricted to a single component. Imagine performing heat simulation on a large scale circuit and then use `.subsection()` to  cut out individual components for optic simulations)
- On basic level, this feature extends method `.sel_inside()` from `SpatialDataArray` to other spatial data arrays (`ScalarFieldDataArray`, `ScalarFieldTimeDataArray`), custom medium classes and space time modulation classes. I realize that the name I used `.sel_inside(bounds: Bound)` might be a bit misleading, because we extract data not just inside but that covers `bounds`. An alternative could be `.reduce_to_cover(bounds: Bound)`, `.sel_to_cover(bounds: Bound)`. Any other suggestions?

I put three of you as reviewers, but I don't think all of you need to review the entire thing. Maybe @weiliangjin2021 can looks through changes to custom medium classes and modulation classes, while @tylerflex and/or @momchil-flex can comment on overall implementation?

### DIFF
--- a/tests/test_plugins/test_mode_solver.py
+++ b/tests/test_plugins/test_mode_solver.py
@@ -214,6 +214,18 @@ def verify_dtype(ms):
         assert dtype == field.dtype
 
 
+def check_ms_reduction(ms):
+
+    ms_red = ms.reduced_simulation_copy
+    grids_1d = ms._solver_grid.boundaries
+    grids_1d_red = ms_red._solver_grid.boundaries
+    assert np.allclose(grids_1d.x, grids_1d_red.x)
+    assert np.allclose(grids_1d.y, grids_1d_red.y)
+    assert np.allclose(grids_1d.z, grids_1d_red.z)
+    modes_red = ms.solve()
+    assert np.allclose(ms.data.n_eff.values, modes_red.n_eff.values)
+
+
 def test_mode_solver_validation():
     """Test invalidate mode solver setups."""
 
@@ -319,6 +331,7 @@ def test_mode_solver_simple(mock_remote_api, local):
         verify_pol_fraction(ms)
         verify_dtype(ms)
         dataframe = ms.data.to_dataframe()
+        check_ms_reduction(ms)
 
     else:
         _ = msweb.run(ms)
@@ -378,6 +391,9 @@ def test_mode_solver_custom_medium(mock_remote_api, local, tmp_path):
         )
         modes = ms.solve() if local else msweb.run(ms)
         n_eff.append(modes.n_eff.values)
+
+        if local:
+            check_ms_reduction(ms)
 
         fname = str(tmp_path / "ms_custom_medium.hdf5")
         ms.to_file(fname)
@@ -445,6 +461,9 @@ def test_mode_solver_straight_vs_angled():
         direction="-",
     )
 
+    check_ms_reduction(ms)
+    check_ms_reduction(ms_angled)
+
     for key, val in ms.data.modes_info.items():
         tol = 1e-2
         if key == "mode area":
@@ -483,6 +502,7 @@ def test_mode_solver_angle_bend():
     verify_pol_fraction(ms)
     verify_dtype(ms)
     dataframe = ms.data.to_dataframe()
+    check_ms_reduction(ms)
 
     # Plot field
     _, ax = plt.subplots(1)
@@ -519,6 +539,7 @@ def test_mode_solver_2D():
     verify_pol_fraction(ms)
     verify_dtype(ms)
     dataframe = ms.data.to_dataframe()
+    check_ms_reduction(ms)
 
     mode_spec = td.ModeSpec(
         num_modes=3,
@@ -540,6 +561,7 @@ def test_mode_solver_2D():
     compare_colocation(ms)
     # verify_pol_fraction(ms)
     dataframe = ms.data.to_dataframe()
+    check_ms_reduction(ms)
 
     # The simulation and the mode plane are both 0D along the same dimension
     simulation = td.Simulation(
@@ -552,6 +574,7 @@ def test_mode_solver_2D():
     ms = ModeSolver(simulation=simulation, plane=PLANE, mode_spec=mode_spec, freqs=[td.C_0 / 1.0])
     compare_colocation(ms)
     verify_pol_fraction(ms)
+    check_ms_reduction(ms)
 
 
 @pytest.mark.parametrize("local", [True, False])
@@ -605,6 +628,7 @@ def test_group_index(mock_remote_api, log_capture, local):
         assert len(log_capture) == 2
         _ = modes.dispersion
         assert len(log_capture) == 2
+        check_ms_reduction(ms)
 
     # Group index calculated
     ms = ModeSolver(
@@ -623,6 +647,7 @@ def test_group_index(mock_remote_api, log_capture, local):
         assert (modes.dispersion.sel(mode_index=0).values < 1500).all()
         assert (modes.dispersion.sel(mode_index=1).values > -16500).all()
         assert (modes.dispersion.sel(mode_index=1).values < -15000).all()
+        check_ms_reduction(ms)
 
 
 def test_pml_params():
@@ -685,6 +710,7 @@ def test_mode_solver_nan_pol_fraction():
     )
 
     md = ms.solve()
+    check_ms_reduction(ms)
 
     assert list(np.where(np.isnan(md.pol_fraction.te))[1]) == [8, 9]
 
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -816,6 +816,28 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         """Whether the medium is a PEC."""
         return False
 
+    def sel_inside(self, bounds: Bound) -> AbstractMedium:
+        """Return a new medium that contains the minimal amount data necessary to cover
+        a spatial region defined by ``bounds``.
+
+
+        Parameters
+        ----------
+        bounds : Tuple[float, float, float], Tuple[float, float float]
+            Min and max bounds packaged as ``(minx, miny, minz), (maxx, maxy, maxz)``.
+
+        Returns
+        -------
+        AbstractMedium
+            Medium with reduced data.
+        """
+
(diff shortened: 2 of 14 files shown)


tidy3d/plugins/mode/mode_solver.py
```diff
@@ -926,3 +951,48 @@ def _validate_modes_size(self):
 
     def validate_pre_upload(self, source_required: bool = True):
         self._validate_modes_size()
+
+    @cached_property
+    def reduced_simulation_copy(self):
+        """Strip non-related structures, mediums, monitors, and sources from simulation object.
```
tylerflex, line 957: instead of "non-related" maybe something about "objects not used by the mode solver"? it wasn't immediately clear what `non-related` meant until I dug into things a little.

tidy3d/web/api/mode.py
```diff
@@ -64,6 +65,9 @@ def run(
         Optional callback function called when uploading file with ``bytes_in_chunk`` as argument.
     progress_callback_download : Callable[[float], None] = None
         Optional callback function called when downloading file with ``bytes_in_chunk`` as argument.
+    reduce_simulation : bool = True
```
tylerflex, line 68: probably a good idea to do it this way.

tidy3d/components/simulation.py
```diff
@@ -2984,3 +2986,198 @@ def from_scene(cls, scene: Scene, **kwargs) -> Simulation:
...
+
+        if boundary_spec is None:
+            boundary_spec = self.boundary_spec
+
+        # reduction of custom medium data
+        new_sim_medium = self.medium
+        if remove_outside_custom_mediums:
+
+            # check for special treatment in case of PML
+            if isinstance(boundary_spec, (PML, StablePML, Absorber)):
```
tylerflex, line 3126: I think these three are all instances of `AbsorberSpec`, which itself is a `BoundaryEdge`, not a `BoundarySpec`. something seems a little strange to me, would a `simulation.boundary_spec` ever be an instance of `PML`, etc?

tests/test_plugins/test_mode_solver.py
```diff
@@ -213,6 +213,22 @@ def verify_dtype(ms):
         assert dtype == field.dtype
 
 
+def check_ms_reduction(ms):
+
+    ms_red = ms.reduced_simulation_copy
+    print(ms_red.simulation.symmetry)
```
tylerflex, line 219: do we need to print here? it's fine but our test outputs are already quite long.

tylerflex (commented):
Thanks @dbochkov-flexcompute I think the implementation makes a lot of sense. I just had a few minor comments about some of the conversions. This will be a nice feature to have in ways we haven't anticipated yet, I'm sure!

tidy3d/components/simulation.py
dbochkov-flexcompute: oh, good catch, this does seem wrong and I didn't have a test for that. Will fix it, thanks!

tidy3d/components/medium.py
```diff
@@ -2486,6 +2615,37 @@ def loss_upper_bound(self) -> float:
...
+        bounds : Tuple[float, float, float], Tuple[float, float float]
+            Min and max bounds packaged as ``(minx, miny, minz), (maxx, maxy, maxz)``.
+
+        Returns
+        -------
+        CustomPoleResidue
+            CustomPoleResidue with reduced data.
+        """
+        if not self.eps_inf.does_cover(bounds=bounds):
+            log.warning("Eps inf spatial data array does not fully cover the requested region.")
```
weiliangjin2021, line 2634: Eps inf -> 'eps_inf'

tidy3d/components/medium.py
```diff
@@ -2486,6 +2615,37 @@ def loss_upper_bound(self) -> float:
...
+        eps_inf_reduced = self.eps_inf.sel_inside(bounds=bounds)
+        poles_reduced = []
+        for pole, residue in self.poles:
+
+            if not pole.does_cover(bounds=bounds):
+                log.warning("Pole spatial data array does not fully cover the requested region.")
```
weiliangjin2021, line 2640: I wonder if it is necessary to warn at all. It is expected that some structure will be outside the requested region. It might raise tons of warning from a dispersive custom medium.

tidy3d/components/medium.py
```diff
@@ -3843,6 +4167,9 @@ def eps_dataarray_freq(
             for ind, mat_component in enumerate(self.components.values())
         )
 
+    def _sel_custom_data_inside(self, bounds: Bound):
+        return self
+
```
weiliangjin2021, line 4172: The reduction doesn't seem to apply to anistoropic custom medium?

tidy3d/components/medium.py
dbochkov-flexcompute: actually, here `bounds` would represent intersection of region from `.subsection(region=region, ...)` and the bounding box of underlying structure or simulation itself (if background medium). So, if the original structure was put together correctly (that is, custom medium either covers it entirely or only one point is provided along a given dimension for constant extrapolation), then it won't produce any warnings

tidy3d/components/medium.py
dbochkov-flexcompute: It does apply. It's just a proper `sel_inside()` is already inherited from `AnisotropicMedium` and we just need to neutralize abstract `_sel_custom_data_inside` inherited from `AbstractCustomMedium`

dbochkov-flexcompute:
addressed all comments, plus:
- shortened error/warning message here https://github.com/flexcompute/tidy3d/pull/1316/files#diff-8592632a38ecdcce2cea8cafbc92bd6e4314181e5e15bf07c57741ad4cd11f57R174 when an object is completely outside of simulation. Otherwise objects with custom medium produce lots of printed output.
- there was an error in docstring https://github.com/flexcompute/tidy3d/pull/1316/files#diff-227742ff635e6a1930bb06970d64a6f1f6cb94df41d53545959e00e16e5e7adeR986. Do we test now docstring code for correctness?
- after adding more tests ran into an issue that diffraction monitor validation would fail with an ambiguous message here https://github.com/flexcompute/tidy3d/pull/1316/files#diff-7360ac4bbe331087dbf8aa41d7163c3d1f0e95986f0488ce103cf24e4520049bR1238. I think this is related to the fact that `eps_model(frequency=...)` only works for a single frequency in case of custom mediums, so I changed validation in that case to only "central" frequency for now. But are diffraction monitors even allowed for custom medium?

momchil-flex:
> * after adding more tests ran into an issue that diffraction monitor validation would fail with an ambiguous message here https://github.com/flexcompute/tidy3d/pull/1316/files#diff-7360ac4bbe331087dbf8aa41d7163c3d1f0e95986f0488ce103cf24e4520049bR1238. I think this is related to the fact that `eps_model(frequency=...)` only works for a single frequency in case of custom mediums, so I changed validation in that case to only "central" frequency for now. But are diffraction monitors even allowed for custom medium?

This is a good question. I think there are two issues, both of which probably lie outside the scope of this PR.

- `eps_model` works for a list of frequencies in some cases, and not in others. I encountered this before as well, in fact I think I fixed it for anisotropic media, but it's still an issue for custom media. The method docstring actually suggests that only a `float` should be passed, but I believe there are many instances where we pass an array instead. Ideally we should make this consistent across all media, or it will keep popping up.
- We have a [validator](https://github.com/flexcompute/tidy3d/blob/2d6bd787686bd25fd371143f176ec5a0baa35b8d/tidy3d/components/simulation.py#L1073-L1081) that is *meant* to validate that a diffraction monitor is in a homogeneous medium. However, what it actually does is compute the list of mediums that the plane intersects, and make sure that there's only one element in that list. This does **not** work for custom medium, which is a single medium which may nevertheless still be inhomogeneous physically. This same issue applies to the validator for plane wave/gaussian beam sources, as well as, in some sense, to the [TFSF validator](def _validate_tfsf_structure_intersections) about structure intersections. I am not sure what's a good way to fix this: it seems like we should either prohibit custom medium, or, if custom medium found, actually sample the permittivity and check if it is homogeneous?

weiliangjin2021 (approved):

tylerflex (approved):

PR #1324, opened 2023-12-18
Determine whether repeated tasks will be run.

magiWei:
When user run a task from python client, API Server will check whether there is a task which run successfully and contains the same simulation file by file md5. if exist repeated task, API server will  delete the new task and s3 file and throw specified error and python client needs to deal the error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Determine whether repeated tasks will be run.
 ## [2.5.0] - 2023-12-13
 
 ### Added
--- a/tidy3d/web/api/webapi.py
+++ b/tidy3d/web/api/webapi.py
@@ -17,6 +17,7 @@ from .connect_util import (
 )
 from ..core.environment import Env
 from ..core.constants import SIM_FILE_HDF5, TaskId
+from ..core.exceptions import TaskRepeatedError
 from ..core.task_core import SimulationTask, Folder
 from ..core.task_info import TaskInfo, ChargeType
 from ...components.types import Literal
@@ -88,15 +89,35 @@ def run(
         verbose=verbose,
         progress_callback=progress_callback_upload,
     )
-    start(
-        task_id,
-        solver_version=solver_version,
-        worker_group=worker_group,
-    )
-    monitor(task_id, verbose=verbose)
-    return load(
-        task_id=task_id, path=path, verbose=verbose, progress_callback=progress_callback_download
-    )
+    try:
+        start(
+            task_id,
+            solver_version=solver_version,
+            worker_group=worker_group,
+        )
+        monitor(task_id, verbose=verbose)
+        return load(
+            task_id=task_id,
+            path=path,
+            verbose=verbose,
+            progress_callback=progress_callback_download,
+        )
+    except TaskRepeatedError as e:
+        new_task_id = e.args[0]
+        if verbose:
+            console = get_logging_console()
+            new_url = _get_url(new_task_id)
+            url = _get_url(task_id)
+            console.log(
+                f"There is a task that contains the same simulation file with '{task_id}', "
+                f"View task using web UI at [link={new_url}]'{new_url}'[/link], the original url [link={url}]'{url}'[/link] will be disabled."
+            )
+        return load(
+            task_id=new_task_id,
+            path=path,
+            verbose=verbose,
+            progress_callback=progress_callback_download,
+        )
 
 
 @wait_for_connection
--- a/tidy3d/web/core/exceptions.py
+++ b/tidy3d/web/core/exceptions.py
@@ -10,3 +10,13 @@ class WebError(Exception):
         log = get_logger()
         super().__init__(message)
         log.error(message)
+
+
+class TaskRepeatedError(Exception):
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Determine whether repeated tasks will be run.
```
tylerflex, line 9: I think this should be more specific, perhaps
```
- If a simulation is detected as being run previously, downloads the results from the original simulation instead.
```

tidy3d/web/core/exceptions.py
```diff
@@ -10,3 +10,13 @@ def __init__(self, message: str = None):
...
         log.error(message)
+
+
+class TaskRepeatedError(Exception):
+    """Repeated task in tidy3d"""
+
+    def __init__(self, message: str = None):
+        """Log just the error message and then raise the Exception."""
+        # log = get_logger()
+        super().__init__(message)
```
tylerflex, line 21: I dont think this line of code does anything different from the original init. I think we can delete this whole method.

tidy3d/web/core/http_util.py
```diff
@@ -129,6 +131,10 @@ def wrapper(*args, **kwargs):
         if resp.status_code != ResponseCodes.OK.value:
             if resp.status_code == ResponseCodes.NOT_FOUND.value:
                 return None
+            if resp.status_code == ResponseCodes.UNPROCESSABLE.value and resp.text:
+                result = resp.json()
+                if str(ResponseCodes.TASK_REPEATED.value) == result["code"]:
```
tylerflex, line 136: what's the reason for `str()` casting here and not in the other `if` statements above?

tylerflex, line 136: I'm worried about obscure errors happening if `"code"` or `"detail"` are not found in `result`. Is there a way to handle this more gracefully?  maybe
```py
if any(key not in result for key in ("code", "detail")):
    # do task repeated logic
else:
    # not sure, either raise a new WebError about missing keys or just continue?
```

```

File: tidy3d/web/api/webapi.py [hunk 4]
```diff
@@ -88,15 +89,35 @@ def run(
...
+            worker_group=worker_group,
+        )
+        monitor(task_id, verbose=verbose)
+        return load(
+            task_id=task_id,
+            path=path,
+            verbose=verbose,
+            progress_callback=progress_callback_download,
+        )
+    except TaskRepeatedError as e:
```
tylerflex, line 105: when does this get raised? in `start`? if so, maybe we should put this `except` handling there?  otherwise it will only have this feature if user uses `web.run()`,  not if they do `web.start()` nor if they use `Job.run()`, `Batch.run()`. I think it would be better if we could catch it at all times?

Maybe the issue is that we need a new task_id to continue the loading? if that's the case, maybe a cleaner approach is to allow `web.start` to return an optional `task_id` object and then slightly refactor the webapi internals to use this returned task_id in the subsequent code?

Basically in it's current stage, I think it needs to be refactored to future-proof and make this generally work every time `web.start()` is called.

tylerflex (changes requested):
Thanks @magiWei for the PR.

I think there's a bit of changes required, we should also probably get @lei-flex to look at it. My feeling is that we can refactor this handling a bit better inside of the `start()` function for more generality + some other minor comments, mainly about making sure the code is robust if this feature fails for some reason.

My more general concern is with the design of the caching. So in the future, there will never be duplicate tasks stored on our side? Basically every time a new task is made that is a duplicate of another task, it is immediately deleted? This might be a big change to a user's workflow if they are doing something like several parameter sweeps (like in different folders) with duplicated data.

A few discussion points / suggestions:
1. Is there some way to easily make this an optional feature? 
2. An architecture that would make sense to me is to separate the simulation from the data. So we could have duplicate tasks with the same simulation, but only the data for unique simulations would be stored. So if a user submits task1 and task2 with the same simulation. We keep both tasks in the UI but just do a mapping from simulation -> data and grab the common data. To me this seems like an architecture that should be considered, do you have thoughts on this design?

magiWei:
> Thanks @magiWei for the PR.
> 
> I think there's a bit of changes required, we should also probably get @lei-flex to look at it. My feeling is that we can refactor this handling a bit better inside of the `start()` function for more generality + some other minor comments, mainly about making sure the code is robust if this feature fails for some reason.
> 
> My more general concern is with the design of the caching. So in the future, there will never be duplicate tasks stored on our side? Basically every time a new task is made that is a duplicate of another task, it is immediately deleted? This might be a big change to a user's workflow if they are doing something like several parameter sweeps (like in different folders) with duplicated data.
> 
> A few discussion points / suggestions:
> 
> 1. Is there some way to easily make this an optional feature?
> 2. An architecture that would make sense to me is to separate the simulation from the data. So we could have duplicate tasks with the same simulation, but only the data for unique simulations would be stored. So if a user submits task1 and task2 with the same simulation. We keep both tasks in the UI but just do a mapping from simulation -> data and grab the common data. To me this seems like an architecture that should be considered, do you have thoughts on this design?

About #2, We have considered keeping task1 and task2 in DB, and task1 and task2 link to the same s3 file. If so, all operations involving tasks need to determine whether the task links the same s3 file, including: query, delete, modify, copy functions, etc.  the complexity of the API Server system will be greatly increased. So API server throws an error when checking one repeated task to python client.

PR #1325, opened 2023-12-18
Fix to mesher error in very specific cases

momchil-flex:
Warn when structures are too small compared to the generated mesh.

Also realized that e.g. placing a thin `MeshOverrideStructure` of size dl and mesh step dl will *not* place boundaries on both sides of that structure (i.e. a single mesh step) because of the way the `isclose` check is done. Changed tolerances a bit to fix this as it should be expected behavior.

I also added more docstrings explaining how that part of the mesher works. I must admit it's still a bit of a mess, but I don't have time for a complete overhaul (and it's not clear what that might look like).

### DIFF
--- a/tidy3d/components/grid/mesher.py
+++ b/tidy3d/components/grid/mesher.py
@@ -19,8 +19,14 @@ from ..structure import Structure, MeshOverrideStructure, StructureType
 from ..medium import AnisotropicMedium, Medium2D, PECMedium
 from ...exceptions import SetupError, ValidationError
 from ...constants import C_0, fp_eps
+from ...log import log
 
 _ROOTS_TOL = 1e-10
+
+# Shrink min_step a little so that if e.g. a structure has target dl = 0.1 and a width of 0.1,
+# a grid point will be added on both sides of the structure. Without this factor, the mesher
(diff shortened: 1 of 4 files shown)


tidy3d/components/grid/mesher.py
```diff
@@ -135,40 +141,74 @@ def parse_structures(
...
+        bounding boxes of all structures in the list (some filtering is done to exclude points that
+        will be too close together compared to the absolute lower desired step). At every point, the
+        ``"structs"`` list has length one lower than the ``"coords"`` list, and each element is
+        another list of all structure indexes that are found in the interval formed by ``coords[i]``
+        and ``coords[i + 1]``. This only includes structures that have a physical presence in the
+        interval, i.e. it excludes structures that are completely covered by higher-up ones.
+
+        To build this, we iterate in reverse order as latter structures override earlier ones.
+        We also handle containment in the following way (note - we work with bounding boxes only):
+         1. If a structure that is lower in the list than the current structure  is found to be
```
tomflexcompute, line 154: There is a duplicated space here.

weiliangjin2021 (approved):

PR #1326, opened 2023-12-19
bump `jax` and `jaxlib` versions to `0.4.*`

tylerflex:
@daquinteroflex note: when we merge this, we should also update it in the docs refractor.


daquinteroflex:
Sounds good! Just for reference on the caret spec, it's not implemented with pip, but the equivalent definition is [here](https://python-poetry.org/docs/dependency-specification/#caret-requirements) if it helps

daquinteroflex (approved):
Looks reasonable! Never done a "proper" review so let me know any guidance, other than a "generally looks reasonable".

tylerflex:
for this one, more of just a sanity check yea. But in the future, we will go through the changed files, make comments and suggestions, and then suggest a few changes (if necessary). similar to what I did reviewing your docs PR.

PR #1327, opened 2023-12-19
add string matching to `assert_log_level`

tylerflex:


momchil-flex:
I think what would also be good to have is a context manager that avoids us doing `log_capture.clear()` explicitly. Something like:

```
with AssertLogLeveL(log_capture, log_level_expected, capture_str):
    do_something()
```

and just use this everywhere instead of the current `assert_log_level`. What do you think?

tylerflex:
I think that could be nice, yea.  I can work on that and add a commit to this branch.

tylerflex (commented):
@momchil-flex I added the context manager, I do think this improves the workflow a lot
```py
with AssertLogLevel(log_capture, "WARNING"):
    # do stuff
```

also updated all tests to use it

momchil-flex:
That's nice! Only thing I'm wondering about is that the context manager affects the passed `results`. In *principle* it may be better practice to not `clear` but just to only examine the results that are added during the execution, e.g. if we have a list of outputs, save the length upon entering, and then iterate from that length to the new list length when exiting (also not clearing there). In practice, this is good for our purposes though. But for example, if you don't clear the log, and you run the pytest with `-rA`, at the end it will show you everything that was captured along the way in `stderr output`. Now, it will show empty. What do you think?

tylerflex:
yea I agree, this is probably better. I added a [commit](https://github.com/flexcompute/tidy3d/pull/1327/commits/1053863cf43e7eaa17ca1c6648b1f1d6481346d8) that only clears a `deepcopy()` of the original `log_capture`. that way the original is not affected.

This made a couple tests fail in `test_medium.py` that I'm coordinating with @weiliangjin2021 about. Basically these two (original) lines

```py
    # anisotropic medium, warn allow_gain is ignored
    _ = td.AnisotropicMedium(xx=td.Medium(), yy=mL, zz=mS, allow_gain=True)
    assert_log_level(log_capture, "WARNING")

    _ = td.AnisotropicMedium(xx=td.Medium(), yy=mL, zz=mS, allow_gain=False)
    assert_log_level(log_capture, "WARNING")
```

perhaps were just capturing the warnings from the previous lines. I'm not sure if there's a bug in `AnisotropicMedium` or we dont expect any warnings since `allow_gain` is specified and the log level should be `None`.

tylerflex:
note @momchil-flex I just force pushed to resolve conflicts with develop.

tylerflex:
Ok actually this uncovered a bug in my own implementation. I adjusted the logic and it should work now. Basically how it works is that the context manager never clears the log_capture, but it only looks at the records introduced in the context (between `__enter__` and `__exit__`).

momchil-flex:
Great. Seems like `deepcopy` is unused now. We can merge afterwards.

PR #1328, opened 2023-12-19
from_nk to support arbitrary nk values

### LINKED ISSUE #1321: from_nk to support arbitrary nk values
The current nondispersive medium requires permittivity>1. Some users would like to do single frequency or very narrow band simulation with medium of negative permittivity. Can we expand the capability of `from_nk` such that it returns a pole residue model if given k>n at the particular wavelength?

weiliangjin2021:
Resolves #1321

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -1138,6 +1138,12 @@ class Medium(AbstractMedium):
             medium containing the corresponding ``permittivity`` and ``conductivity``.
         """
         eps, sigma = AbstractMedium.nk_to_eps_sigma(n, k, freq)
+        if eps < 1:
+            raise ValidationError(
+                "Dispersiveless medium must have 'permittivity>=1`. "
+                "Please use 'Lorentz.from_nk()' to covert to a Lorentz medium, or the utility "
+                "function 'td.medium_from_nk()' to automatically return the proper medium type."
+            )
         return cls(permittivity=eps, conductivity=sigma, **kwargs)
 
 
@@ -2998,6 +3004,59 @@ class Lorentz(DispersiveMedium):
             return np.all(coeff_a.values > coeff_b.values)
         return coeff_a > coeff_b
 
+    @classmethod
+    def from_nk(cls, n: float, k: float, freq: float, **kwargs):
+        """Convert ``n`` and ``k`` values at frequency ``freq`` to a single-pole Lorentz
+        medium.
+
+        Parameters
+        ----------
+        n : float
+            Real part of refractive index.
+        k : float = 0
+            Imaginary part of refrative index.
+        freq : float
+            Frequency to evaluate permittivity at (Hz).
+
+        Returns
+        -------
+        :class:`Lorentz`
+            Lorentz medium having refractive index n+ik at frequency ``freq``.
+        """
+        eps_complex = AbstractMedium.nk_to_eps_complex(n, k)
+        eps_r, eps_i = eps_complex.real, eps_complex.imag
+        if eps_r >= 1:
+            log.warning(
+                "For 'permittivity>=1', it is more computationally efficient to "
+                "use a dispersiveless medium constructed from 'Medium.from_nk()'."
+            )
+        # first, lossless medium
+        if isclose(eps_i, 0):
+            if eps_r < 1:
+                fp = np.sqrt((eps_r - 1) / (eps_r - 2)) * freq
+                return cls(
+                    eps_inf=1,
+                    coeffs=[
+                        (1, fp, 0),
+                    ],
+                )
+            return cls(
+                eps_inf=1,
+                coeffs=[
+                    ((eps_r - 1) / 2, np.sqrt(2) * freq, 0),
+                ],
+            )
+        # lossy medium
+        alpha = (eps_r - 1) / eps_i
+        delta_p = freq / 2 / (alpha**2 - alpha + 1)
+        fp = np.sqrt((alpha**2 + 1) / (alpha**2 - alpha + 1)) * freq
+        return cls(
+            eps_inf=1,
+            coeffs=[
+                (eps_i, fp, delta_p),
+            ],
+        )
+
 
 class CustomLorentz(CustomDispersiveMedium, Lorentz):
     """A spatially varying dispersive medium described by the Lorentz model.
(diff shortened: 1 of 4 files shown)


tomflexcompute (approved):
Nice! Thanks for the implementation.

tidy3d/components/medium.py
```diff
@@ -471,6 +471,30 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         description="Modulation spec applied on top of the base medium properties.",
     )
 
+    @classmethod
+    def from_general_nk(cls, n: float, k: float, freq: float, **kwargs):
```
tylerflex, line 475: please add type annotations for the return `Union[Medium, Lorentz]`

tidy3d/components/medium.py
```diff
@@ -1107,6 +1131,12 @@ def from_nk(cls, n: float, k: float, freq: float, **kwargs):
             medium containing the corresponding ``permittivity`` and ``conductivity``.
         """
         eps, sigma = AbstractMedium.nk_to_eps_sigma(n, k, freq)
+        if eps < 1:
+            raise ValidationError(
+                "Dispersiveless medium must have 'permittivity>=1`. "
+                "Please use 'Lorentz.from_nk()' to covert to a Lorentz medium, "
+                "or 'AbstractMedium.from_general_nk() to automate the process."
```
tylerflex, line 1138: they might not know what `AbstractMedium` is, you could also phrase this as `or use the '.from_general_nk()' method in all mediums.`.

tidy3d/components/medium.py
tylerflex: this class is a bit recursive I guess.. the `Medium` inherits from `AbstractMedium` but then `AbstractMedium.from_general_nk` returns a `Medium`. I'm not sure we even want to include this method unless it's very useful because it introduces a circular dependency and I think would break if we ever moved `Medium` and `AbstractMedium` to separate files.. is there a better way? or can we just leave out?

tidy3d/components/medium.py
```diff
@@ -2840,6 +2870,53 @@ def _all_larger(coeff_a, coeff_b) -> bool:
             return np.all(coeff_a.values > coeff_b.values)
         return coeff_a > coeff_b
 
+    @classmethod
+    def from_nk(cls, n: float, k: float, freq: float, **kwargs):
+        """Convert ``n`` and ``k`` values at frequency ``freq`` to :class:`Lorentz`.
```
tylerflex, line 2875: can you add some description of how this is being done? it's hard to tell from the path? for example that there is a single pole and how its computed?

tidy3d/components/medium.py
tylerflex: Another comment, if we keep it, I feel having `from_nk_general` is better than `from_general_nk` only because then the `from_nx` methods can be next to each other alphabetically and easier to compare in IDE and in the API reference.

tylerflex (changes requested):
implementation looks good. The `AbstractMethod.from_general_nk -> Medium` scares me a bit because it introduces circular dependencies. all else equal I would remove it and I think the Lorenz version with the warning in Medium is enough. Maybe @tomflexcompute feels more strongly?

tomflexcompute:
Indeed I would also assume users rarely use `AbstractMedium`. It should be fine to leave it out. Maybe have an abstract method `from_nk` in `AbstractMedium` and implement it differently in `Medium` and `Lorentz`?

tylerflex:
the fundamental problem here I think is that we need the general `from_nk()` method to be aware of `Medium` and `Lorentz`.  But if we put it in `AbstractMedium` (which both of those inherit from) there's a circular dependency. It's also not clear where it would get called from. I think the "safest" if we really want this is to add maybe a utility function in `medium.py` that does this, which would be called something like

```py
td.medium_from_nk(n, k, freq)
```

but it would have to be detached from our regular classes.

tomflexcompute:
I was thinking more like an empty abstract method in the abstract class. Users don't really use it but that from the subclasses (Medium.from_nk or Lorentz.from_nk). However I just realized that in Python the abstract method **has** to be implemented by subclasses so it's actually not suitable since other subclasses like Drude don't have this method. In this case it should be fine to have just `from_nk` in Medium and `Lorentz` with error/warning messages mentioning each other.

tylerflex:
> I was thinking more like an empty abstract method in the abstract class. Users don't really use it but that from the subclasses (Medium.from_nk or Lorentz.from_nk).

This would just set the template for how the subclasses implement `from_nk`. But what we want is a way to switch between the subclass `from_nk` implementations depending on the inputs. Just making an abstractmethod wouldn't be enough to do this. What weiliang implemented would be the approach if we wanted it to be callable from any medium, but just introduces this circular dependency.

tylerflex:
@weiliangjin2021 I forgot, but could you also add a changelog item?

tomflexcompute:
Correct. Probably there is no good way to have it automatically determine what type of medium to return. It should be fine as long as the error/warning messages are clear. 

Maybe in Lorentz.from_nk we also want to check if permittivity>1? If so, we should warn and encourage users to use Medium.from_nk?

weiliangjin2021:
> Probably there is no good way to have it automatically determine what type of medium to return. It should be fine as long as the error/warning messages are clear.

I think dealing with it automatically will make user's life much easier. Consider that you want to do a sweep, but then with lots of debugging, you finally find out that you need to write an additional branching to have `from_nk` working for that sweep parameter.

> Maybe in Lorentz.from_nk we also want to check if permittivity>1? If so, we should warn and encourage users to use Medium.from_nk?

Lorentz works for any permittivity.

> I think the "safest" if we really want this is to add maybe a utility function in medium.py that does this, which would be called something like
> 
> td.medium_from_nk(n, k, freq)

Yeah, it seems like the only feasible approach so far.

tylerflex:
this `medium_from_nk` function is almost related to the topic discussed Monday about making some utility functions / scripts for users. for now I think in `medium.py` is ok though.

tomflexcompute:
> > Maybe in Lorentz.from_nk we also want to check if permittivity>1? If so, we should warn and encourage users to use Medium.from_nk?
> 
> Lorentz works for any permittivity.

In case permittivity>1, using Medium.from_nk is more computationally effective?

weiliangjin2021:
> > > Maybe in Lorentz.from_nk we also want to check if permittivity>1? If so, we should warn and encourage users to use Medium.from_nk?
> > 
> > 
> > Lorentz works for any permittivity.
> 
> In case permittivity>1, using Medium.from_nk is more computationally effective?

Yeah, right.

tidy3d/components/medium.py
weiliangjin2021: It's a single-pole Lorentz model. How it is computed is obvious from the code? I have a PDF note on this, but I think the code is simple enough?

tylerflex: yea it's obvious from the code, I just mean that someone looking at the API reference might just be curious, so a little info could be useful (even if just mentioning that it is a **single** pole Lorentz model).

weiliangjin2021: i see, that has been added in the revision.

tylerflex:
@weiliangjin2021 is this ready to review again?

weiliangjin2021:
> @weiliangjin2021 is this ready to review again?

Yes

tidy3d/components/medium.py
```diff
@@ -1107,6 +1107,12 @@ def from_nk(cls, n: float, k: float, freq: float, **kwargs):
             medium containing the corresponding ``permittivity`` and ``conductivity``.
         """
         eps, sigma = AbstractMedium.nk_to_eps_sigma(n, k, freq)
+        if eps < 1:
+            raise ValidationError(
+                "Dispersiveless medium must have 'permittivity>=1`. "
+                "Please use 'Lorentz.from_nk()' to covert to a Lorentz medium, "
+                "or the utility function 'td.medium_from_nk() to automate the process."
```
tylerflex, line 1114: missing quotes on the end of `'td.medium_from_nk() `

tylerflex: `automate the process` -> `automatically return the proper medium type.` maybe

tylerflex (commented):
looks good, thanks @weiliangjin2021 . just a couple minor comments and then you could merge to pre/2.6 when fixed.

PR #1329, opened 2023-12-20
Added free-carrier absorption (FCA) and free-carrier plasma dispersion (FCPD)

caseyflex:
Adds FCA and FCPD inside TwoPhotonAbsorption class.


tidy3d/components/medium.py
```diff
@@ -229,13 +263,20 @@ def complex_fields(self) -> bool:
 class TwoPhotonAbsorption(NonlinearModel):
     """Model for two-photon absorption (TPA) nonlinearity which gives an intensity-dependent
     absorption of the form :math:`\\alpha = \\alpha_0 + \\beta I`.
+    Also includes free-carrier absorption (FCA) and free-carrier plasma (FCP) effects.
```
momchil-flex, line 266: Is FCP the same as free-carrier dispersion (i.e. modulation of the real part of the permittivity)? I've heard the latter but not the former.

tests/test_components/test_medium.py
```diff
@@ -624,7 +623,7 @@ def test_nonlinear_medium(log_capture):
     # active materials
     with pytest.raises(ValidationError):
         med = td.Medium(
-            nonlinear_spec=td.NonlinearSpec(models=[td.TwoPhotonAbsorption(beta=-1 + 1j, n0=1)])
+            nonlinear_spec=td.NonlinearSpec(models=[td.TwoPhotonAbsorption(beta=-1, n0=1)])
```
momchil-flex, line 626: What exactly is happening with `beta`? It used to be possible to set it as complex, where the imaginary part means modulation of the *real* part of the permittivity? But TPA is purely absorption, so now instead we want beta to be purely real, while any dispersion would be coming from FCP?

Is the medium still unstable if the nonlinear effect is purely lossy like TPA?

momchil-flex (commented):
Some initial comments

tidy3d/components/medium.py
caseyflex: yeah, if that is more familiar I can change the terminology

tests/test_components/test_medium.py
caseyflex: `beta` used to be complex, although typically a real beta would be used. I think you would use KerrNonlinearity instead if you wanted to modify the real part of the permittivity directly, or as you said you can have FCP dispersion coming from the same origin as the TPA.

Complex `beta` is implemented according to the docstring equation for TPA nonlinear polarization, which involves both the complex `n0` and `Re(n0)`. This equation is derived assuming real `beta`. For complex `beta`, we can use the same equation / implementation, but the physics may be less clear, and I cannot think of a use case. So, I deprecated complex `beta`.

TPA is lossy when `beta` is positive -- a negative value of `beta` would instead add energy in a nonlinear fashion, which may actually be used in some cases, although is an active medium and can lead to instability. The more general condition for stability involves the actual nonlinear polarization (and is what we use in the validator), and reduces to this when `n0` is real.

tests/sims/simulation_2_6_0rc1.json
```diff
@@ -816,11 +816,23 @@
...
+                            "beta": 1.0,
+                            "tau": 1.0,
+                            "sigma": 1.0,
+                            "e_e": 1.0,
+                            "e_h": 0.8,
+                            "c_e": 1.0,
+                            "c_h": 1.0,
+                            "n0": null,
+                            "freq0": null
+>>>>>>> 26576d4d (Added free-carrier absorption (FCA) and free-carrier plasma dispersion (FCPD) nonlinearities.)
```
momchil-flex, line 835: Not sure why there's a conflict here. Could you rerun test_local maybe and see if that would fix it (it generates this file). Also, `git checkout develop -- tests/sims/simulation_2_5_0.json`.

momchil-flex (approved):
This looks good, I'll look at the backend changes again too. Is this waiting for the notebook example to be done / work correctly though?

tests/sims/simulation_2_6_0rc1.json
caseyflex: I don't think it needs to wait for the notebook example...

The two simulation json files are fixed now.

PR #1331, opened 2023-12-20
FIX: Try to relink the git index

daquinteroflex:


daquinteroflex:
Not on the `tidy3d` side

PR #1333, opened 2023-12-21
adjoint support for PoleResidue

tylerflex:
Still a work in progress.

Can get grad w.r.t. eps_inf, but poles is a bit weird, mostly due to the type annotation and how that works with jax plugin.


tylerflex:
Closing this to pick up later

PR #1335, opened 2023-12-24
adjoint refactor

tylerflex:
this one works. Still improving things though and fixing a couple tests.

## Main changes:
- [x] jax tracer for field `x` moved to a new field `x_jax`. 
    - [ ] When values passed to `x`, the traced version is copied to `x_jax` and an untraced copy is saved in `x`.
    - [ ] No need to patch or remove validators or existing methods (`.bounds`) as they used the untraced copies.
- [ ] `jax_leaf:bool` flag added in the `pd.Field` to specify that this field is specifically a leaf node (contains tracers). `jax_field` only means that the field contains either a `JaxObject` or `tuple` of them.
- [x] `.to_tidy3d()` and `.from_tidy3d()` conversions generalized to base class. 
    - [ ] `to_simulation()` conversion still requires a bit of manual work for now.
- [ ] Made `JaxDataArray` a `JaxObject` and simplified the internals a lot as it uses existing methods now.
- [ ] ran black with the updated version, which created a lot of changes.

## Status
- [ ] Pipeline tests mostly passing where it counts.
- [ ] Notebooks 2 and 3 run successfully.
- [ ] Some adjoint tests fail.

## To do later
- [ ] Fix a couple tests that are failing because of serializing `JaxDataArray`

## To do in another stage / PR of the refactor? 
- [ ] move from defining jax fields in `pd.Field` to using class attrs.
- [ ] simplify `to_simulation()` by adding "transformations" to `to_tidy3d()` and `from_tidy3d()`.
- [ ] have web API accept JaxSimulation directly? no need for conversion?

### DIFF
--- a/tests/test_plugins/test_adjoint.py
+++ b/tests/test_plugins/test_adjoint.py
@@ -492,38 +492,38 @@ def test_adjoint_pipeline_tyler(local, use_emulated_run, tmp_path):
     # no gradients close to zero
     assert not all(np.all(np.isclose(x, 0)) for x in (df_deps, df_dsize, df_dvertices, d_eps_base)), "REALLY BAD"
     assert not any(np.any(np.isclose(x, 0)) for x in (df_deps, df_dsize, df_dvertices, d_eps_base))
-
+    assert np.isclose(df_deps, 1278130200000000.0)
     print("gradient: ", df_deps, df_dsize, df_dvertices, d_eps_base)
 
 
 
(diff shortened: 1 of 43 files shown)


tylerflex:
I got ahead of myself with this one, the data array updates opened a can of worms with file IO. To revisit in stage 2.

For now, the simpler version (to merge) is #1337 . Still cleaning that up

PR #1337, opened 2023-12-24
adjoint refactor: stage 1

tylerflex:
This will be stage 1 of some future adjoint refactors. This one finally works though, and successfully separates the jax tracers from the original tidy3d fields. The main advantage of this is that we dont have to worry about the jax tracers interactively badly with the existing tidy3d validators and methods (eg. `Box.bounds()`). 

## Summary of changes:
- [x] jax tracer for field `x` moved to a new field `x_jax`. 
    - [x] When values passed to `x`, the traced version is copied to `x_jax` and an untraced copy is saved in `x`.
    - [x] No need to patch or remove validators or existing methods (`.bounds`) as they used the untraced copies.
- [x] `stores_jax_for:str` flag added in the `pd.Field` to specify which field this field stores jax for. `jax_field=True` now only means that the field contains either a `JaxObject` or `tuple` of them.
- [x] `.to_tidy3d()` and `.from_tidy3d()` conversions generalized to base class. 
    - [x] `to_simulation()` conversion still requires a bit of manual work for now.
- [x] Simplified handlings generally throughout the code
    - [x] JaxDataArray
    - [x] Converting numpy to list to not break jax serialization.
    - [x] Saving to json.
- [x] More gracefully handle if the `JVPTracer` import fails in future jax versions (which could break the plugin if it changes).

## Tests
- [x] Adjoint tests all pass, give same gradients as pre/2.6
- [x] Notebooks 2, 3 and 5 give same results when running locally.

## To do in another stage
- [ ] handle logic for mixed JaxObject and tidy3d object in a `jax_field`. 
- [ ] simplify `to_simulation()` by adding "transformations" to `to_tidy3d()` and `from_tidy3d()`.
- [ ] Simplify JaxDataArray internals.
- [ ] have web API accept JaxSimulation directly? no need for conversion?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - All solver output is now compressed. However, it is automatically unpacked to the same `simulation_data.hdf5` by default when loading simulation data from the server.
+- Internal refactor of `adjoint` plugin to separate `jax`-traced fields from regular `tidy3d` fields.
 
 ### Fixed
 
--- a/tests/test_plugins/test_adjoint.py
+++ b/tests/test_plugins/test_adjoint.py
@@ -1,6 +1,7 @@
 """Tests adjoint plugin."""
 
 from typing import Tuple, Dict, List
+import builtins
 
 import pytest
 import pydantic.v1 as pydantic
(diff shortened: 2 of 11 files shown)


momchil-flex:
This actually seems pretty clever! I don't have any comments, we can merge as long as everything works. To this end I can build a solver version with this changes and we should run all notebooks. However, some backend tests currently fail on the level of `JaxSimulationData` IO. I'll send you a file which errors under `JaxSimulationData.from_file("jax_sim_data_fwd.hdf5")`.

tylerflex:
Thanks, I added a fix in the latest commit. Basically when we separate the passed value into the "normal" field and the Jax-traced field, we apply `jax.lax.stop_gradient(v)` to the value to "un-trace" it with jax.  This operation was erroring if `v` was a type that jax doesn't know, such as `["Infinity", "Infinity", "Infinity"]`. I changed it now such that the validator will `try` to perform this un-tracing, but if it fails due to a `TypeError` like this, it just put the raw, passed value in the field instead. Fixes the issue and I added a test to confirm (which failed before this commit).

tylerflex:
I'll squash all the commits into one now and if things go as plan can merge this afternoon.

PR #1338, opened 2023-12-25
run updated black

tylerflex:
Also upgrades black version to 23.12.1

### DIFF
--- a/docs/generate_doc.py
+++ b/docs/generate_doc.py
@@ -9,7 +9,6 @@ LOW_LOSS_THRESHOLD = 2e-5
 
 
 def generate_material_library_doc():
-
     # Display unit for "Valid range" in table, select from 'eV', 'um', 'THz'
     unit = "um"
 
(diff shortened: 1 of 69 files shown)


tylerflex:
We can worry about this after the break

daquinteroflex:
What do you think was wrong with it? Seems like it ran fine?

tylerflex:
So I
* upgraded my black to `23.12.1`
* ran black
* updated black version in pyproject.toml
* ran poetry lock
but then I think the tests failed. Maybe this was because the tests run from pre/2.6?

tylerflex:
Huh, now that I look at it thought, the tests passed. Not sure what was going on, then. I might have misunderstood something.

daquinteroflex:
Ahh that test that didn't pass seems like it was a black reformatting test you did in the working commit (maybe through the precommit?), and I suspect maybe after you ran again it worked?

PR #1347, opened 2024-01-03
Bringing back basic requirements file

momchil-flex:
We need a set of basic requirements to be able to build minimal webservices used for validation. I brought this file back from the grave it was sent to in 2.6. I am not sure if this is the best place for it, and also I guess it would be good if we can test somehow that these reqs are sufficient. I guess we could have a separate github action that only installs those and tries to run some of the tests that should work?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,12 +34,14 @@ pydantic = "^2.0"
 PyYAML = "*"
 dask = "*"
 toml = "*"
+### NOT CORE
 scipy = "*"
 boto3 = "1.23.1"
 requests = "*"
 pyjwt = "*"
 click = "8.0.3"
 responses = "*"
+### END NOT CORE
 
 ### Optional dependencies ### 
 # development core


daquinteroflex:
I think sounds good! 

If there's no compatibility issues for the backend, how about we just move it into `scripts/basic_requirements.txt` as it's kind of a development item would be the only thing I'd propose. 

I think we'll have to begin increasing the complexity of our test environments based on what you're suggesting and also for example on Daniil's `no-vtk` installation tests too, and maybe we want to do this in a way to scale for the future. I'll have a think how to best integrate it with the existing packaging and testing flow too for another PR.

momchil-flex:
I wanted to avoid maintaining two separate files and was thinking of automatically scraping the pyproject.toml isntead. A related suggestion came from Juncheng. With the current modification to the `pyproject.toml` file, the following command
```
sed -i.bak '/### NOT CORE/,/### END NOT CORE/s/^/#/' pyproject.toml
```
can be used to comment out the dependencies that are between the "NOT CORE" comments. This can then easily be incorporated before installing tidy3d if only the basic version would be needed.

daquinteroflex:
Yeah, makes sense to compose it out of the pyproject.toml. Because there are a lot of dependencies under `[tool.poetry.dependencies]` that are just separated with comments, like the list is pretty big under there and many are written as optional dependencies, will this script include all of them too? If it works well then great, but wonder if it'd be easier something like this to parse?

Happy to go with whatever works too! Just maybe thinking this is easiest unless the extraction functionality is already implemented elsewhere?

```
[tool.poetry.dependencies]
### START CORE ###
python = ">=3.9,<3.12"
pyroots = ">=0.5.0"
xarray = ">=0.16.2"
importlib-metadata = "^6.0.0"
h5netcdf = "1.0.2"
h5py = "^3.0.0"
rich = "<12.6.0"
matplotlib = "*"
shapely = "^2.0"
pydantic = "^2.0"
PyYAML = "*"
dask = "*"
toml = "*"
### END CORE ###
scipy = "*"
boto3 = "1.23.1"
requests = "*"
pyjwt = "*"
click = "8.0.3"
responses = "*"

### Optional dependencies ### 
# development core
pre-commit = {version="*", optional = true}
black = {version="22.3.0", optional = true}

... there are quite a lot of dependencies here under [tool.poetry.dependencies] I'm concerned how it's going to match.
```

momchil-flex:
So my version assumes that everything below the "core" block is optional. I think this is a fair assumption to make? The nice thing about it is that it's harder to come up with an sed command that can correctly manipulate your version (although your version was indeed what I initially suggested).

momchil-flex:
I'll just merge this as is because it works, but we can reconsider in the future. :)

PR #1348, opened 2024-01-03
Unified validation check for dependency fields

dbochkov-flexcompute:
After playing a little bit with it, managed to implement what we wanted as a convenient decorator. The usage as follows: instead of doing something like 
```
@pydantic.validator("monitors", always=True)
def _warn_monitor_simulation_frequency_range(cls, val, values):
    """Warn if any DFT monitors have frequencies outside of the simulation frequency range."""

    if "sources" not in values:
        raise ValidationError(
            "could not validate `_warn_monitor_simulation_frequency_range` "
            "as `sources` failed validation"
        )

    ...
```
we can just do 
```
@pydantic.validator("monitors", always=True)
@check_previous_fields_validation(["sources"])
def _warn_monitor_simulation_frequency_range(cls, val, values):
    """Warn if any DFT monitors have frequencies outside of the simulation frequency range."""
    ...
```
this check whether dependency fields have passed validation, and, if not, it skips (not errors) the current validator and prints something like
```
WARNING: Could not execute validator                               
'_warn_monitor_simulation_frequency_range' because field 'sources' 
failed validation.
```

### DIFF
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -86,6 +86,28 @@ def _get_valid_extension(fname: str) -> str:
     )
 
 
+def skip_if_fields_missing(fields: List[str]):
+    """Decorate ``validator`` to check that other fields have passed validation."""
+
+    def actual_decorator(validator):
+        @wraps(validator)
+        def _validator(cls, val, values):
+            """New validator function."""
+            for field in fields:
+                if field not in values:
+                    log.warning(
+                        f"Could not execute validator '{validator.__name__}' because field "
+                        f"'{field}' failed validation."
+                    )
+                    return val
+
+            return validator(cls, val, values)
+
+        return _validator
+
+    return actual_decorator
+
+
 class Tidy3dBaseModel(pydantic.BaseModel):
     """Base pydantic model that all Tidy3d components inherit from.
     Defines configuration for handling data structures
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -9,6 +9,7 @@ import xarray as xr
 import matplotlib as mpl
 
 from .base import cached_property
+from .base import skip_if_fields_missing
 from .validators import assert_objects_in_sim_bounds
 from .validators import validate_mode_objects_symmetry
 from .geometry.base import Geometry, Box
@@ -224,6 +225,7 @@ class Simulation(AbstractSimulation):
         return updater.update_to_current()
 
     @pydantic.validator("grid_spec", always=True)
+    @skip_if_fields_missing(["sources"])
     def _validate_auto_grid_wavelength(cls, val, values):
         """Check that wavelength can be defined if there is auto grid spec."""
         if val.wavelength is None and val.auto_grid_used:
@@ -242,6 +244,7 @@ class Simulation(AbstractSimulation):
     # _plane_waves_in_homo = validate_plane_wave_intersections()
 
     @pydantic.validator("boundary_spec", always=True)
+    @skip_if_fields_missing(["symmetry"])
     def bloch_with_symmetry(cls, val, values):
         """Error if a Bloch boundary is applied with symmetry"""
         boundaries = val.to_list
@@ -255,6 +258,7 @@ class Simulation(AbstractSimulation):
         return val
 
     @pydantic.validator("boundary_spec", always=True)
+    @skip_if_fields_missing(["medium", "size", "structures", "sources"])
     def plane_wave_boundaries(cls, val, values):
         """Error if there are plane wave sources incompatible with boundary conditions."""
         boundaries = val.to_list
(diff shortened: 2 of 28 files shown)


momchil-flex:
Why would we only warn instead of error? It seems to me that if previous fields failed validation we should error with the validation error of that field.

momchil-flex:
Also I think this PR should go to develop so that we patch this asap.

tidy3d/components/base.py
```diff
@@ -86,6 +86,28 @@ def _get_valid_extension(fname: str) -> str:
...
 
+def check_previous_fields_validation(required_fields):
+    """Decorate ``validator`` to check that other fields have passed validation."""
+
+    def actual_decorator(validator):
+        @wraps(validator)
+        def _validator(cls, val, values):
+            """New validator function."""
+            for field in required_fields:
+                if field not in values:
```
tylerflex, line 97: just to confirm, if `field` fails validation, it's just not put in `values`? (as opposed to put into values with `None`?)

tidy3d/components/base.py
```diff
@@ -86,6 +86,28 @@ def _get_valid_extension(fname: str) -> str:
     )
 
 
+def check_previous_fields_validation(required_fields):
```
tylerflex, line 89: can you type annotate `required_fields : List[str]`?

tidy3d/components/simulation.py
tylerflex, line 1: there are probably a lot of cases in `Simulation` that could use this. For [example](https://github.com/flexcompute/tidy3d/blob/c0adac91959671fb07c9b41fb798e5e4a8de8272/tidy3d/components/simulation.py#L702-L705)

tylerflex (changes requested):
Thanks a lot @dbochkov-flexcompute, I think this will be great to have. A couple minor comments and also there are probably many instances here and there where we can implement it already, especially in `Simulation`. For example these [lines](https://github.com/flexcompute/tidy3d/blob/c0adac91959671fb07c9b41fb798e5e4a8de8272/tidy3d/components/simulation.py#L702-L705).

I wonder if, more generally, we can assume that if any `values.get(key)` or `values[key]` is present in a field validator, that we should probably `@check_previous_fields_validation` for that `key`.

Another pretty minor comment, perhaps the name of the decorator could be more specific, like
```
skip_if_validation_failed(fields: List[str])
```
because `check` is maybe a bit ambiguous as to what the decorator is doing. (Just a minor suggestion).

dbochkov-flexcompute:
> Why would we only warn instead of error? It seems to me that if previous fields failed validation we should error with the validation error of that field.

Just to be clear, the failed field still throws its own error. But I'm ok either way, the difference is very subtle. If we error, then pydantic will collect this additional error and overall we get the following output:
```
09:47:51 CST ERROR: Automatic grid generation requires the input of 'wavelength'
             or sources.
             ERROR: Could not execute validator 'plane_in_sim_bounds' because   
             field 'simulation' failed validation.
...
pydantic.v1.error_wrappers.ValidationError: 2 validation errors for ModeSolver
simulation -> grid_spec
  Automatic grid generation requires the input of 'wavelength' or sources. (type=value_error.setup)
plane
  Could not execute validator 'plane_in_sim_bounds' because field 'simulation' failed validation. (type=value_error)
```
Note that pydantic error list contains two errors (`pydantic.v1.error_wrappers.ValidationError: 2 validation errors for ModeSolver`).

If we just pop a warning, we get
```
16:58:19 CST ERROR: Automatic grid generation requires the input of 'wavelength'
             or sources.                                                        
             WARNING: Could not execute validator 'plane_in_sim_bounds' because 
             field 'simulation' failed validation.                              
...
pydantic.v1.error_wrappers.ValidationError: 1 validation error for ModeSolver
simulation -> grid_spec
  Automatic grid generation requires the input of 'wavelength' or sources. (type=value_error.setup)
```
Note that in this case pydantic error list contains just one error (`pydantic.v1.error_wrappers.ValidationError: 1 validation error for ModeSolver`).

This doesn't really make much difference, but I think when we explicitly error a field ("A") that was not validated just because a previous field ("B") failed validation, this might make an impression that something is wrong with setting up field "A", while in reality it could be that just resolving issues with filed "B" is enough. It might be more relevant for GUI were failed fields are marked visually.

momchil-flex:
Oh ok that makes sense thanks!

tidy3d/components/base.py
dbochkov-flexcompute: yeah, just double checked with pydantic documentation https://docs.pydantic.dev/1.10/usage/validators/:
> If validation fails on another field (or that field is missing) it will not be included in values, hence if 'password1' in values and ... in this example.

dbochkov-flexcompute:
revised this PR, bigger changes:
- rebased to develop
- per Tyler suggestion to make it more specific renamed the decorator to `skip_if_fields_missing`
- added it wherever it seemed to make sense

momchil-flex (approved):
Looks great to me!

tylerflex (approved):

tylerflex:
thanks @dbochkov-flexcompute

PR #1353, opened 2024-01-05
FEAT: Generalisation of large imports version 1 

daquinteroflex:
Just focused on restructuring to on-the-fly imports in a general way. I think it'd be good to benchmark to make sure it's any better than before, but will add the a benchmarking utility as well on pre/2.6 as no develop CLI on this

### DIFF
--- a/tests/test_components/test_packaging.py
+++ b/tests/test_components/test_packaging.py
@@ -1,0 +1,58 @@
+"""
+Thank you, ChatGPT, for the core of the following test code.
+"""
+
+import pytest
+from tidy3d.packaging import check_import, verify_packages_import, Tidy3dImportError
+
+assert check_import("tidy3d") == True
+
+
+# Mock module import function to simulate availability
+def mock_check_import(module_name):
+    if module_name == "tidy3d":
+        return True
+    return False
+
+
+def test_verify_packages_import_all_required():
+    @verify_packages_import(["tidy3d", "module2"], required="all")
+    def my_function():
+        pass
+
+    with pytest.raises(Tidy3dImportError) as exc_info:
+        my_function()
+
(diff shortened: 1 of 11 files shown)


daquinteroflex:
A few notes:

- I think we should think more about how we decouple dependencies as we add more functionality. Eg. currently `TrimeshType` is in types.py which is a core package. The issue here as well is that trimesh is not a core dependency, yet it runs in a core script so it will inherently be unable to be decoupled. We should have types for the corresponding extra functionality we want to implement.

daquinteroflex:
If we observe our import time distribution, in terms of external libraries you could argue it is dominated by:

- xarray (includes pandas)
- trimesh
- numpy

The rest seems to be internal. However, we have a situation here where our basic installation only requires, of these packages:

- xarray

and turns out we don't explicitly require `numpy` but we use it directly, we need to change this.

image

daquinteroflex:
Like I think we have a fundamental architectural domain issue where we need to decouple the domains properly if we really want to get this type of performance improvements where I can functionally perform an import depending on the functional domain, currently they're too intertwined where it requires importing everything or domain interaction with other packages that are not necessarily core.

daquinteroflex:
I have tried to do what I've been asked to in terms of on-the-fly importing, but I think we need to restructure between domains and operations if we really want to get performance improvements. In any case, with this commit you can see now trimesh does not get imported on the core import which maybe frees up resources (again you have to time imports on average as it's hardware dependent, but at least the reduced operation is now there)

image

daquinteroflex:
Rebasing into my own branch so we can properly benchmark into 2.6. But this branch is available for 2.5.3 patch if we wanted.

PR #1357, opened 2024-01-05
Momchil/merge develop in 26

momchil-flex:


tylerflex:
Hm, so we're going to basically apply black via this PR? if so should we also update the black version in pyproject.toml?

CHANGELOG.md
```diff
@@ -1063,7 +1074,8 @@ which fields are to be projected is now determined automatically based on the me
 - Job and Batch classes for better simulation handling (eventually to fully replace webapi functions).
 - A large number of small improvements and bug fixes.
 
-[Unreleased]: https://github.com/flexcompute/tidy3d/compare/v2.5.0...develop
+[Unreleased]: https://github.com/flexcompute/tidy3d/compare/v2.5.1...develop
```
tylerflex, line 1077: hm, probably this should be `v2.5.1...pre/2.6` (although I dont recall normally changing this might be good to)

tylerflex (approved):
looks good to me, one super minor comment about changelog link

CHANGELOG.md
momchil-flex: This just comes from develop now. I think it's fine yeah we don't normally put pre/ link there.

tylerflex (approved):
k should hopefully be all set now

CHANGELOG.md
momchil-flex: Actually I will make this update but only when finalizing the CHANGELOG for 2.6.0rc1, not now.

tylerflex: yea I tried out the Link and it seemed to work fine. we can do it then.

PR #1358, opened 2024-01-08
FEAT: Automatic release from branch v*

daquinteroflex:
Hi Momchil,

I've improved the release process. 

So now:
-  You make a branch in `tidy3d/` that starts with the letter "v" like this example branch
- When you commit to this branch, the `sync-to-readthedocs` action will automatically make a branch in `tidy3d-docs` with the same name
- It will push the contents into that branch. 

You can see this worked well:
* https://github.com/flexcompute/tidy3d/actions/runs/7448894014/job/20264267298
* https://github.com/flexcompute-readthedocs/tidy3d-docs/tree/v_test_release

Then you can easily just build the documentation for that branch in the readthedocs admin page.

(I am unsure what happens when it is overwritten)

### DIFF
--- a/.github/workflows/sync-to-readthedocs-repo.yaml
+++ b/.github/workflows/sync-to-readthedocs-repo.yaml
@@ -7,6 +7,7 @@ on:
       - main
       - latest
       - 'pre/*'
+      - 'v*'
   workflow_dispatch:
     inputs:
       env:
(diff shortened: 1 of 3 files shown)


momchil-flex:
What is your motivation for this? I thought it was already good enough in that I both develop and pre/ branches are synced, so I can just build the docs from the top level of those. I guess the idea of this addition is that I could try any earlier state too, right?

daquinteroflex:
I realized that it was going to be a pain to do what I did before.

 Before, you had to manually create a branch on two repositories, but to do that you had to push manually to the tidy3d-docs repo. So this action didn't really make sense until now. Now, you just have to make a branch that starts with a v, and it does the rest for you in terms of pushing and I think syncing.

momchil-flex:
Ok I see what you mean, this PR does *more* than just add the `v_` capability. It also makes sure that a new branch is created in the tidy3d-docs repo.

daquinteroflex:
Yeah that's the idea. However, I've now updated the github action for it to be any branch.

 Now, I'm doing two things here you might need for the release so you might need to merge this PR for the full release to work:
 * I am updating the linked submodule state of tidy3d/docs to the latest commit history in pre/2.6 I just refactored
 * I am updating the deployment action for the branch thing.
 * Docs are building locally as we speak

daquinteroflex:
Local docs are looking good. @momchil-flex feel free to review. It's ready to merge imo.

The release flow is as follows:
* The release flow just involves making a branch now either pre/  or v/ in tidy3d frontend. This branch will propagate automatically, and you should be able to just click build documentation branch from the readthedocs control panel.

PR #1360, opened 2024-01-08
release workflow auto generates release notes

tylerflex:
We used to have to manually generate release notes on GitHub through clicking

image

This workflow change makes this happen automatically:

## IMPORTANT:
from my testing on a dummy repo, I needed       
```
permissions:
  contents: write
```
to get the release action to run at all, but maybe this is not related to the release notes and should be removed.


daquinteroflex (approved):
Looked through the documentation and looks good!

PR #1362, opened 2024-01-09
FIX: Sort out sync issue, :wave: submodule, :+1: subtree

daquinteroflex:
Attempt to sort out the sync issue via git subtree rather than submodules.


daquinteroflex:
Closing because unlikely we'll want to store all the history also in this repo, gonna get messy? Alternatively good way to track multiple project states.

PR #1363, opened 2024-01-09
FIX: sync issue, :wave: submodules, :1st_place_medal: subtrees

daquinteroflex:
Submodules are painful and complicated to keep in sync.

Hence, git developers came up with the next great thing: [subtrees](https://www.atlassian.com/git/tutorials/git-subtree).

Approximately works about the same, but is treated more like a directory within another directory in practice.

This branch is a test that it works properly.


daquinteroflex:
Subtrees will cause other issues and worsen our decoupling. Need to sort this out with submodules

Issue #1274, opened 2023-11-29
Add mode information in ModeMonitorData


tomflexcompute:
Current mode monitor result only stores the effective indices and group indices of the modes. Since we have added polarization fractions and mode area information in the mode solver result, one user is wondering if we can add the same info to the mode monitor result as well for easier mode sorting.

lucas-flexcompute:
One thing we have to consider is that polarization fraction and mode area are quantities calculated locally based on the available field distributions in `ModeSolverData` (because it derives from `ElectromagneticFieldData`). `ModeData` does not include the mode fields, so it's not possible to do the same thing. I see alternatives if we really want to implement this feature:

1. Calculate on the backend and add it directly to the results. This will increase our post-processing calculations for every mode monitor, which I'm not sure is a good idea.
2. Return the modal field information for `ModeMonitor`. If we go this route, it seems to me that `ModeMonitor` and `ModeSolverMonitor` could be merged, but I haven't checked any implementation details and differences between them. @momchil-flex would probably be the first person to chime in here, I guess. This is the solution I'd prefer, as it would decrease our API complexity in the end.

momchil-flex:
> One thing we have to consider is that polarization fraction and mode area are quantities calculated locally based on the available field distributions in `ModeSolverData` (because it derives from `ElectromagneticFieldData`). `ModeData` does not include the mode fields, so it's not possible to do the same thing. I see 2 alternatives if we really want to implement this feature:
> 
> 1. Calculate on the backend and add it directly to the results. This will increase our post-processing calculations for every mode monitor, which I'm not sure is a good idea.

I would think this would never be comparable to the dominating computation time (computing the modes, and also other postprocessing that we already do anyway, like colocating, normalizing, and reordering), but do you have a particular worry in mind?

> 2. Return the modal field information for `ModeMonitor`. If we go this route, it seems to me that `ModeMonitor` and `ModeSolverMonitor` could be merged, but I haven't checked any implementation details and differences between them. @momchil-flex would probably be the first person to chime in here, I guess. This is the solution I'd prefer, as it would decrease our API complexity in the end.

[...]

PR #1367, opened 2024-01-09
Add field data to ModeMonitor (#1274)

lucas-flexcompute:
The implementation proposed here merges `ModeSolverData` and into `ModeData`. It also deprecates the use of `ModeSolverMonitor` and changes the implementation on `ModeSolverData` to be derived from `ModeData` (to avoid most code duplication and make deprecation later easier). There are 2 issues at this point:
1. `ModeSolverData` ended up inheriting the `amps` field, which will be unused.
2. This warning: https://github.com/flexcompute/tidy3d/blob/fa402883fafa3de37fbe0b22cbfb3b04e2263552/tidy3d/components/data/monitor_data.py#L105-L112 is being emitted (unnecessarily?) and getting caught in [an adjoint test](https://github.com/flexcompute/tidy3d/blob/fa402883fafa3de37fbe0b22cbfb3b04e2263552/tests/test_plugins/test_adjoint.py#L1618).

Thoughts, @tylerflex and @momchil-flex?

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -12,7 +12,7 @@ from pandas import DataFrame
 
 from .data_array import FluxTimeDataArray, FluxDataArray
 from .data_array import MixedModeDataArray, ModeAmpsDataArray
-from .data_array import ModeIndexDataArray, GroupIndexDataArray, ModeDispersionDataArray
+from .data_array import GroupIndexDataArray, ModeDispersionDataArray
 from .data_array import FieldProjectionAngleDataArray, FieldProjectionCartesianDataArray
 from .data_array import FieldProjectionKSpaceDataArray
 from .data_array import DataArray, DiffractionDataArray
@@ -33,7 +33,7 @@ from ..monitor import DiffractionMonitor
 from ..source import SourceTimeType, CustomFieldSource
 from ..medium import Medium, MediumType
 from ...exceptions import SetupError, DataError, Tidy3dNotImplementedError, ValidationError
-from ...constants import ETA_0, C_0, MICROMETER, PICOSECOND_PER_NANOMETER_PER_KILOMETER
+from ...constants import ETA_0, C_0, MICROMETER
 from ...log import log
 
 from ..base_sim.data.monitor_data import AbstractMonitorData
@@ -80,7 +80,7 @@ class MonitorData(AbstractMonitorData, ABC):
 class AbstractFieldData(MonitorData, AbstractFieldDataset, ABC):
     """Collection of scalar fields with some symmetry properties."""
 
-    monitor: Union[FieldMonitor, FieldTimeMonitor, PermittivityMonitor, ModeSolverMonitor]
+    monitor: Union[FieldMonitor, FieldTimeMonitor, PermittivityMonitor, ModeMonitor]
 
     symmetry: Tuple[Symmetry, Symmetry, Symmetry] = pd.Field(
         (0, 0, 0),
@@ -519,9 +519,11 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
 
     @cached_property
     def mode_area(self) -> FreqModeDataArray:
-        """Effective mode area corresponding to a 2D monitor.
+        r"""Effective mode area corresponding to a 2D monitor.
 
-        Effective mode area is calculated as: (∫|E|²dA)² / (∫|E|⁴dA)
+        .. math:
+
+           \frac{\left(\int |E|^2 \, {\rm d}S\right)^2}{\int |E|^4 \, {\rm d}S}
         """
         intensity = self.intensity
         # integrate over the plane
@@ -536,25 +538,30 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
         return FreqModeDataArray(area)
 
     def dot(
-        self, field_data: Union[FieldData, ModeSolverData], conjugate: bool = True
+        self, field_data: Union[FieldData, ModeData, ModeSolverData], conjugate: bool = True
     ) -> ModeAmpsDataArray:
-        """Dot product (modal overlap) with another :class:`.FieldData` object. Both datasets have
+        r"""Dot product (modal overlap) with another :class:`.FieldData` object. Both datasets have
         to be frequency-domain data associated with a 2D monitor. Along the tangential directions,
         the datasets have to have the same discretization. Along the normal direction, the monitor
         position may differ and is ignored. Other coordinates (``frequency``, ``mode_index``) have
         to be either identical or broadcastable. Broadcasting is also supported in the case in
-        which the other ``field_data`` has a dimension of size ``1`` whose coordinate is not in the
-        list of coordinates in the ``self`` dataset along the corresponding dimension. In that case,
-        the coordinates of the ``self`` dataset are used in the output.
+        which the other ``field_data`` has a dimension of size 1 whose coordinate is not in the list
+        of coordinates in the ``self`` dataset along the corresponding dimension. In that case, the
+        coordinates of the ``self`` dataset are used in the output.
+
+        The dot product is defined as:
+
+        .. math:
+
+           \frac{1}{4} \int \left( E_0 \times H_1^* + H_0^* \times E_1 \) \, {\rm d}S
 
         Parameters
         ----------
         field_data : :class:`ElectromagneticFieldData`
             A data instance to compute the dot product with.
         conjugate : bool, optional
-            If ``True`` (default), the dot product is defined as ``1 / 4`` times the integral of
-            ``E_self* x H_other - H_self* x E_other``, where ``x`` is the cross product and ``*`` is
-            complex conjugation. If ``False``, the complex conjugation is skipped.
+            If ``True`` (default), the dot product is defined as above. If ``False``, the definition
+            is similar, but without the complex conjugation of the $H$ fields.
 
         Note
         ----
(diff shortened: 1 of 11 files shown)


tylerflex:
hm, let me look into this.

tylerflex:
see https://github.com/flexcompute/tidy3d/pull/1368/files

lucas-flexcompute:
> Alright, I see. Basically the adjoint test is using the emulated run function in `tests/utils.py` to get the `ModeMonitorData` for the test. In that emulated run function, the mode data constructor was missing a `grid_expanded`. I pass that in and the test works, I also formatted the test a bit better to use our new `AssertLogLevel` context manager and do string matching.
> 
> I'll make a PR to your branch

Awesome, thanks! How do you feel about the overall implementation here, though? Should I keep going this direction and start on the backend changes?

tidy3d/components/monitor.py
```diff
@@ -638,10 +645,23 @@ class ModeMonitor(AbstractModeMonitor):
...
+    direction: Direction = pydantic.Field(
+        "+",
+        title="Propagation Direction",
+        description="Direction of propagation along the normal axis of the monitor used for the "
+        "computed mode field profiles.",
+    )
+
     def storage_size(self, num_cells: int, tmesh: int) -> int:
         """Size of monitor storage given the number of points after discretization."""
         # stores 3 complex numbers per frequency, per mode.
```
tylerflex, line 657: comment is deprecated, or maybe add a similar comment about the fields too.

tidy3d/components/monitor.py
```diff
@@ -673,6 +693,17 @@ class ModeSolverMonitor(AbstractModeMonitor):
         "primal grid nodes).",
     )
 
+    @pydantic.root_validator(skip_on_failure=True)
+    def deprecation_warning(cls, values):
+        """Emit a deprecation warning."""
+        log.warning(
+            "ModeSolverMonitor will be deprecated in the future and can now be replaced by a "
```
tylerflex, line 700: put the class names in single quotes

tidy3d/components/monitor.py
```diff
@@ -673,6 +693,17 @@ class ModeSolverMonitor(AbstractModeMonitor):
         "primal grid nodes).",
     )
 
+    @pydantic.root_validator(skip_on_failure=True)
+    def deprecation_warning(cls, values):
```
tylerflex, line 697: we should ensure that this change is reflected in our notebooks after this?

tidy3d/components/data/monitor_data.py
```diff
@@ -536,25 +538,30 @@ def mode_area(self) -> FreqModeDataArray:
         return FreqModeDataArray(area)
 
     def dot(
-        self, field_data: Union[FieldData, ModeSolverData], conjugate: bool = True
+        self, field_data: Union[FieldData, ModeData], conjugate: bool = True
```
tylerflex, line 541: for the meantime should this accept either FieldData, ModeSolverData, OR ModeData?

tidy3d/components/data/monitor_data.py
```diff
@@ -992,6 +1040,12 @@ def overlap_sort(
             modes is less than this threshold, a warning about a possible discontinuity is
             displayed.
         """
+        if len(self.field_components) == 0:
+            # raise DataError(
```
tylerflex, line 1044: resolve comment? (remove or discuss?)

tidy3d/components/data/monitor_data.py
```diff
@@ -615,24 +622,29 @@ def _interpolated_tangential_fields(self, coords: ArrayFloat2D) -> Dict[str, Dat
         return fields
 
     def outer_dot(
```
tylerflex, line 624: if `field_data` is a `ModeData` then it's field components can now be `None`. do we explicitly make sure this is not the case anywhere when implementing these `dot` functions? (or other functions that use the fields?)

tylerflex (commented):
looks good to me! just a few comments.

tidy3d/components/data/monitor_data.py
momchil-flex: I *think* given that `field_components` still returns `return {field_name: field for field_name, field in fields.items() if field is not None}`, there should be no change in behavior

tidy3d/components/monitor.py
momchil-flex: Yeah, but I think there's only one notebook that uses a ModeSolverMonitor so should be easy.

tidy3d/components/data/monitor_data.py
tylerflex: yea, I'm more worried about if the dot functions try to grab one of those components but it is missing, whether that is handled gracefully or you just get an internal KeyError

tidy3d/components/monitor.py
```diff
@@ -638,10 +645,23 @@ class ModeMonitor(AbstractModeMonitor):
         "and is hard-coded for other monitors depending on their specific function.",
     )
 
+    direction: Direction = pydantic.Field(
+        "+",
+        title="Propagation Direction",
+        description="Direction of propagation along the normal axis of the monitor used for the "
+        "computed mode field profiles.",
+    )
```
momchil-flex, line 653: I understand why you've done this but I think it may be confusing to have this argument but still return mode amps in both directions. Like, someone reading the api docs may think that they have to set to "-" if they're doing mode decomposition in the backward direction. It's not terrible but it's a bit annoying to have this inconsistency (`direction` only affecting the mode fields but not the mode amps). However, I do agree that the other options that I see are also not great:
- Do not provide `direction` and store fields in both directions, just like amps
- Store amps only for the specified `direction`

So I don't know, maybe at least we should call it `store_fields_direction`? Or how about make `store_fields` a spec, `store_fields = StoreFieldsSpec(direction="+")`? We could imagine extending this in the future to e.g. only request a subset of frequencies, like just the central freq, etc.

tidy3d/components/data/monitor_data.py
momchil-flex: Yeah that's always been a possibility in `FieldData`, so it's been handled, you should just hit this: https://github.com/flexcompute/tidy3d/blob/1f22335e92d92cf875a1f70e4832acd9515a4c8b/tidy3d/components/data/monitor_data.py#L394C1-L395

tidy3d/components/monitor.py
lucas-flexcompute: Comments don't age well and can't be tested, so I removed it because I think the code is simple enough to document itself here.

tidy3d/components/data/monitor_data.py
lucas-flexcompute: Yes, I feel the error would be more in line to other functions (like trying to calculate intensity without the fields present), but because we only return a copy of the original data with modes reordered, it makes sense to order an empty set of modes and not bother the user with checking if the fields exist before calling the method. So I prefer to return a copy, but I left the option to raise an error if that's preferable.

tidy3d/components/data/monitor_data.py
lucas-flexcompute: That's right. All user-facing functions should handle missing components gracefully. Internal methods *do not* have those checks, for example: `ModeData._isel`, `ModeData._find_ordering_one_freq`, etc., but those are use-at-your-own-risk methods, so I guess the user should be responsible for the check.

tidy3d/components/monitor.py
lucas-flexcompute: Yes, I considered the same options when implementing this way but storing fields in both directions would be a major change and make `ModeData` and `ModeSolverData` interfaces incompatible.

I think your idea of using `store_fields` with a spec might be the better solution here. I'll go with that for now.

lucas-flexcompute: After implementing the `StoreFieldsSpec` I think we should reconsider. This looks a lot like over-engineering to me. Maybe using `store_fields_direction` makes more sense. I reverted the changes to this simpler implementation but left both in the commit history just in case.

Frequency and multiple direction selection are rare use cases, and can be accomplished by setting up multiple monitors. That way we keep the usual use case straightforward, but the advanced uses still possible.

momchil-flex: Yeah, I agree with this evaluation.

momchil-flex (approved):
Looks good to me now!

tylerflex (approved):
looks good to me, sorry if you were waiting on me.

momchil-flex:
So I looked through this again to see about any remaining parts of the code that may need to change. I think on the backend nothing much needs to change apart from replacing some ModeSolverMonitor instances in the tests with a ModeMonitor that stores modes. The only other thing I see is not in the backend per se but in the denormalizer: ModeData now has to optionally denormalize the fields as well, if present (I'll probably ask @dbochkov-flexcompute to take a look at that), and MC has to incorporate it in GUI eventually.

Well, that brings me to my main comment actually. ModeSolverData *does* serve a purpose that currently cannot be replaced (or doesn't make sense to be replaced) by a ModeData: namely, it is the return data structure of the ModeSolver. It doesn't make sense to be replaced by ModeData because that one currently always carries decomposition amplitudes, and optionally mode fields. Well, so there's technically two options:
- Make `amps` in the `ModeData` optional, and make the `ModeSolver` return `ModeData`. I feel like this is confusing though.
- Probably my preference: just don't deprecate `ModeSolverMonitor`. Keep it for its purpose as a monitor corresponding to a `ModeSolverData`, and maybe just print a warning in a `Simulation` if a `ModeSolverMonitor` is added that says that a `ModeMonitor` can also store fields?

lucas-flexcompute:
Yes, I agree with the second option as well. I don't even think we need the warning. Just updating the tutorials and making it clear in the docs that a ModeSolverMonitor is no longer required should be enough to (slowly) move users to using a single monitor.

I'll finish any pending rebase issues tomorrow morning and commit a final version for merging front and backe-ends.

lucas-flexcompute:
Since there's no need to remove ModeSolverDate, backend should be done. I've rebased and removed the deprecation warning.

PR #1369, opened 2024-01-09
Fix to internal storage estimate for surface integration monitors

momchil-flex:


tests/sims/simulation_2_5_1.h5
tylerflex, line 1: We probably dont want to track these, maybe we need `*.h5` in the `gitignore`? see [this](https://github.com/flexcompute/tidy3d/blob/develop/.gitignore#L2).  I think I mentioned this on pre/2.6 but maybe we need it in develop too. (how was this file generated? is it from the new zipping logic? alternatively, we want to make sure we unzip as `.hdf5` extension?)

@daquinteroflex

tidy3d/components/simulation.py
```diff
@@ -1196,11 +1196,19 @@ def monitors_data_size(self) -> Dict[str, float]:
...
+        """Total number of cells included in monitor based on simulation grid."""
+        if isinstance(monitor, SurfaceIntegrationMonitor):
+            monitors = monitor.integration_surfaces
+        else:
+            monitors = [monitor]
+
+        num_cells_total = 0
+        for mnt in monitors:
+            num_cells = self.discretize_monitor(mnt).num_cells
+            # take monitor downsampling into account
```
tylerflex, line 1208: Maybe I'd consider separating the logic for computing the number of cells for a regular monitor from the logic for computing the number of cells for a `SurfaceIntegrationMonitor`.

```py
def _num_cells_vol_monitor(self, monitor: Monitor) -> int:
    """Get the number of measurement cells in a monitor given the simulation grid and downsampling."""
    sub_grid = self.discretize_monitor(monitor)
    num_cells = sub_grid.num_cells
    num_cells_downsampled = monitor.downsampled_num_cells(num_cells)
    return np.prod(np.array(num_cells, dtype=np.int64))

def _num_cells_monitor(self, monitor: Monitor) -> int:
    if isinstance(monitor, SurfaceIntegrationMonitor):
        return sum([self._num_cells_vol_monitor(mnt) for mnt in monitor.integration_surfaces])
    return self._num_cells_vol_monitor(monitor)
```

I think it's a bit better to separate the logic out like this, what do you think?

tylerflex (changes requested):

tests/sims/simulation_2_5_1.h5
daquinteroflex: I think maybe when I started off the tidy3d/docs restructure, I was editing the .gitignore to include sphinx files. I may have accidentally removed this. Making a tiny PR for this now

tylerflex: Yea it's already in pre/2.6. So at least we'll have a minor merge conflict

momchil-flex: Actually there is nothing to change here, this is intended. We purposefully gitignore hdf5 files and not h5 files so we can actually track h5 files when needed. These h5 files were added in order to test updater for h5. There was an issue at some point in which the updater tests were fine, but loading a simulation with custom data from hdf5 was failing because something about the data arrays wasn't being updated correctly or something.

tylerflex: ok, that's good to know. I guess that means we should remove `.h5` from the gitignore in pre/2.6 then?

daquinteroflex: Just removed  it in my pre/2.6 PR to eventually propagate

tidy3d/components/simulation.py
momchil-flex: How's this, I left the function as internal to `_num_cells_monitor`?

By the way merging is now blocked if changes are requested, so if things look good you need to somehow mark that they've been resolved. (I can override, but better to do it properly)

tylerflex: that sounds fine (if you think it wouldn't be useful elsewhere) thanks, resolved

momchil-flex: I still get this

image

tylerflex (approved):

PR #1371, opened 2024-01-10
Use simulation reduction automatically only in case of custom mediums

dbochkov-flexcompute:
- shrink bounds when performing `grid.discretize_inds` on new region to avoid including possible spurious pixel
- switch simulation reduction in mode solver to be automatic only for simulations with custom medium + display warning if that happens

### DIFF
--- a/tidy3d/plugins/mode/mode_solver.py
+++ b/tidy3d/plugins/mode/mode_solver.py
@@ -973,7 +973,8 @@ class ModeSolver(Tidy3dBaseModel):
         This might significantly reduce upload time in the presence of custom mediums.
         """
 
-        # we preserve extracells along the normal direction to ensure
+        # we preserve extracells along the normal direction to ensure there is enough data for
+        # subpixel
         extended_grid = self._get_solver_grid(preserve_layer_behind=True, truncate_symmetry=False)
         grids_1d = extended_grid.boundaries
         new_sim_box = Box.from_bounds(
--- a/tidy3d/plugins/mode/mode_solver.py
+++ b/tidy3d/plugins/mode/mode_solver.py
@@ -973,7 +973,7 @@ class ModeSolver(Tidy3dBaseModel):
         This might significantly reduce upload time in the presence of custom mediums.
         """
 
-        # we preserve extracells along the normal direction to ensure there is enough data for
+        # we preserve extra cells along the normal direction to ensure there is enough data for
         # subpixel
         extended_grid = self._get_solver_grid(preserve_layer_behind=True, truncate_symmetry=False)
         grids_1d = extended_grid.boundaries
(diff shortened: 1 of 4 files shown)


tidy3d/web/api/mode.py
```diff
@@ -83,6 +85,21 @@ def run(
     if verbose:
         console = get_logging_console()
 
+    if reduce_simulation == "auto":
+        sim = mode_solver.simulation
+        contains_custom = any(isinstance(s.medium, AbstractCustomMedium) for s in sim.structures)
+        contains_custom = contains_custom or isinstance(sim.medium, AbstractCustomMedium)
```
momchil-flex, line 91: We do something like that in a bunch of places but yeah it seems like there's still no in-built way to get "all media, including the simulation one" and "all structures, including the simulation as a structure"...

tidy3d/web/api/mode.py
```diff
@@ -83,6 +85,21 @@ def run(
...
+
+        if reduce_simulation:
+            log.warning(
+                "The associated `Simulation` object contains custom mediums. It will be "
+                "automatically restricted to the mode solver plane to reduce data for uploading. "
+                "To force uploading the original `Simulation` object use `reduce_simulation=False`."
+                " Setting `reduce_simulation=True` will force simulation reduction in all cases and"
+                " silence this warning."
+            )
+
```
momchil-flex, line 102: I think it looks good, probably we should update the notebooks then to avoid the warning?

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -973,7 +973,8 @@ def reduced_simulation_copy(self):
         This might significantly reduce upload time in the presence of custom mediums.
         """
 
-        # we preserve extracells along the normal direction to ensure
+        # we preserve extracells along the normal direction to ensure there is enough data for
```
momchil-flex, line 976: extra cells

momchil-flex (approved):

tidy3d/web/api/mode.py
dbochkov-flexcompute: Well, I guess I could've just used `Simulation.mediums`

tidy3d/web/api/mode.py
dbochkov-flexcompute: was there any other besides BullseyeCavityPSO?

momchil-flex: BullseyeCavityPSO will not have a warning anymore as it doesn't use custom medium, I was thinking of the heat notebooks actually.

tidy3d/web/api/mode.py
momchil-flex: Ah yeah so there is something like that!

dbochkov-flexcompute:
addressed the comments, also updated heat notebooks https://github.com/flexcompute/tidy3d-notebooks/pull/24

note, there is a change in web api container.py, I noticed that it was causing spurious errors due to pydantic trying to initialize the field `Batch.simulations` with every possible simulation class

PR #1381, opened 2024-01-12
Adding validate_pre_upload to AbstractSimulation

momchil-flex:
This is just for convenience so we can call it regardless of the simulation type.


tylerflex (approved):

tylerflex:
should we merge this into 2.6?

momchil-flex:
Either way works but yeah I can put it in 2.6, I'll only need it there.

PR #1382, opened 2024-01-12
FEAT: Sorted out the packaging of m2r3 :runner: 

daquinteroflex:
Released `m2r3` to PYPI. Think it should work fine now


tylerflex:
Great, thanks. I'm not sure what we do now, maybe we need to merge this PR, delete the v2.6.0rc1 tag, and then push it again? @momchil-flex ?

PR #1384, opened 2024-01-15
FIX: Upgrade FAQ to tidy3d-faq submodule :mag_right: 

daquinteroflex:
I had to build on top of this https://github.com/flexcompute/tidy3d/pull/1375 for this, so maybe we should review them separately as they're different features (and a bit painful to drop out the commits)


tylerflex:
This looks good, maybe change the bass of the PR to #1375? unless you'd rather just merge this into 2.6 after? but I approve the merging of this into 1375 if that's easier.

tylerflex (approved):

PR #1385, opened 2024-01-15
DataArray accepts file path and handle

tylerflex:
Formerly, the docstring made it look like to save a `DataArray`, one should pass a `fname: str`. However, the function internals accept an h5py file handle and this method is called as such, internally.

This PR makes `DataArray.to_hdf5()` work either when passed a file name or a file handle.


e-g-melo (approved):
Thanks @tylerflex!

Both options are working very well in the code below now.

```py
Nx = 1
Ny = 10
Nz = 11
x = [0]
y = np.linspace(-Ly / 2, Ly / 2, Ny)
z = [0]
coords = dict(x=x, y=y, z=z)
xgrid, ygrid, zgrid = np.meshgrid(x, y, z, indexing="ij")

phase = td.SpatialDataArray(2 * np.pi * ygrid / Ly, coords=coords)

phase.to_hdf5("phase_file.hdf5", group_path="phase")

with h5.File("phase_handle.hdf5", "w") as f:
    phase.to_hdf5(f, group_path="phase")
```

tylerflex:
I dont think it's urgent. let's just do 2.6 for simplicity.

PR #1387, opened 2024-01-18
FEAT: On-the-fly benchmarking + faster imports 🕐 v2

daquinteroflex:
Continuing on the phases of the potential refactor:

* Full refactor of vtk


daquinteroflex:
We're going a different approach here

PR #1389, opened 2024-01-22
Allow users to override the global cache setting for individual task.

magiWei:


momchil-flex:
Thanks @magiWei. I think doing this through a `run` argument is actually a bit less convenient than having a configuration-level parameter like we currently set environment to run on. Do you think we could set it in that way, maybe even add it to the `Env` class? Like,

```
from tidy3d.web.core.environment import Env
Env.set_caching("off")
```

This would make it more straightforward to e.g. run scripts or notebooks with caching off without having to add the parameter in every `run` call. Furthermore in your current PR, we would also need to propagate this parameter to `Job.run` and `Batch.run`.

magiWei:
> Thanks @magiWei. I think doing this through a `run` argument is actually a bit less convenient than having a configuration-level parameter like we currently set environment to run on. Do you think we could set it in that way, maybe even add it to the `Env` class? Like,
> 
> ```
> from tidy3d.web.core.environment import Env
> Env.set_caching("off")
> ```
> 
> This would make it more straightforward to e.g. run scripts or notebooks with caching off without having to add the parameter in every `run` call. Furthermore in your current handling, we would also need to propagate this parameter to `Job.run` and `Batch.run`.


I think this is also a better method. It will not destroy the current method. What may need to be added is to guide users to set configuration-level parameters. If we reach an agreement, I will modify the code.

tylerflex:
I agree that a global configuration approach would be better than using kwargs.

One idea is to try doing it like we have `td.config`. The approach here is basically to have an instance of a mutable pydantic datstructure that other parts of the code import and use internally.

https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/config.py#L8-L33

```py
td.config.logging_level = "ERROR"
```

perhaps we can have a new config in web based on this approach?

magiWei:
> I agree that a global configuration approach would be better than using kwargs.
> 
> One idea is to try doing it like we have `td.config`. The approach here is basically to have an instance of a mutable pydantic datstructure that other parts of the code import and use internally.
> 
> https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/config.py#L8-L33
> 
> ```python
> td.config.logging_level = "ERROR"
> ```
> 
> perhaps we can have a new config in web based on this approach?

I modified the code with global configuration approach, so, what's the next job?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `DataArray.to_hdf5()` accepts both file handles and file paths.
 
 ### Fixed
+- Add enable_caching parameter for `run()`, allow users to override the global cache setting for individual task submissions.
```
tylerflex, line 14: please update the changelog item so users can see how to set this. For example
- Users can toggle task caching through `from tidy3d.web.core.environment import Env`. `Env.set_caching(enableCaching: bool = True)`.

Also, this should probably be in the `### Added` section.

tests/test_web/test_tidy3d_task.py
tylerflex, line 1: could we add a test to ensure that this `Env.set_caching(...)` works as expected? Basically just to confirm that if the cache is turned off, the `enableCaching` is properly set in the task and mode solver post request, and that if cache is turned back on, the request body is also updated as expected? Thanks

tylerflex (commented):
Thanks @magiWei, I didn't see that. I had a couple comments, but overall looks good. I think this approach through `Env` is fine for now, although I dont think we have much documentation for the user so far about how to use it, we can fix that later.

CHANGELOG.md
magiWei: modified

tests/test_web/test_tidy3d_task.py
magiWei: I modified the unit test code, and QA also tested it in the dev environment. I'm not sure exactly how you want to test it. Maybe you can give me an example. Thanks.

tylerflex: So, these are all good, thanks. What we need though is a front end unit test that verifies that when `Env.set_caching(val)` is called, then the `enableCaching` is properly set to `val` inside of the requests. 

For onoe example, a unit test like
```py
Env.set_caching(True)
assert Env.current.enable_caching == True

Env.set_caching(False)
assert Env.current.enable_caching == False
```
would be a good start just to validate that the value being passed into the http requests is indeed being affected by a change in `set_caching()`

tidy3d/web/core/environment.py
```diff
@@ -164,5 +165,16 @@ def set_current(self, config: EnvironmentConfig) -> None:
         """
         self._current = config
 
+    def enable_caching(self, enable_caching: bool = True) -> None:
+        """Set the current environment.
```
tylerflex, line 169: this should probably be "Set the environment configuration setting with regards to caching simulation results."

tests/test_web/test_tidy3d_task.py
magiWei: add one test file named test_env.py

tylerflex (approved):

tylerflex:
Looks good to me, thanks @magiWei 
@momchil-flex I'll leave it to you to take a final look and do the merging.

magiWei:
> Looks good to me, thanks @magiWei @momchil-flex I'll leave it to you to take a final look and do the merging.

Thanks @tylerflex @momchil-flex , The webAPI associated with this pr will be deployed to prod on 2024-02-04 beijing time.

tylerflex:
> Thanks @tylerflex @momchil-flex , The webAPI associated with this pr will be deployed to prod on 2024-02-04 beijing time.

@magiWei does that mean we should avoid releasing this until Feb 4? Or is it safe to release now?

magiWei:
> > Thanks @tylerflex @momchil-flex , The webAPI associated with this pr will be deployed to prod on 2024-02-04 beijing time.
> 
> @magiWei does that mean we should avoid releasing this until Feb 4? Or is it safe to release now?

Good question, if it is released before Feb 4, the `Env.enable_caching(False)` will not work because webapi is not deployed. So I recommend releasing it after 2.4 Feb 4.

magiWei:
@tylerflex Webapi has been deployed to prod env.

PR #1392, opened 2024-01-24
Add missing dispersion information to dataframe conversion

lucas-flexcompute:
Minor fix I found when creating a notebook for the workshop.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `DataArray.to_hdf5()` accepts both file handles and file paths.
 
 ### Fixed
+- Add dispersion information to dataframe output when available from moed solver.
```
tylerflex, line 14: moed -> mode. Maybe also mention "dispersion (ps/(nm km))" in here so users know where to look?

tidy3d/components/data/monitor_data.py
```diff
@@ -1363,6 +1363,7 @@ def modes_info(self) -> xr.Dataset:
             "wg TM fraction": self.pol_fraction_waveguide["tm"],
             "mode area": self.mode_area,
             "group index": self.n_group_raw,  # Use raw field to avoid issuing a warning
+            "dispersion (ps/(nm km))": self.dispersion_raw,  # Use raw field to avoid issuing a warning
```
tylerflex, line 1366: if we're including units here, maybe put in the other ones? or does this break backwards compatibility?

tylerflex (changes requested):

tidy3d/components/data/monitor_data.py
lucas-flexcompute: Rigorously, it does break backwards compatibility, but I wouldn't think this is critical in any way. It is still quite new and probably used only in post-processing.

My guess for the original lack of units is that we only added units to fields that don't follow Tidy3D's standard units (µm, s), so mode area is implied to be in μm² and wavelength in μm. Although I'd prefer to have units everywhere, I don't really like the breaking change.

What do you think?

tylerflex: that's fine with me.

PR #1394, opened 2024-01-24
Allow importing specific field from vtk

dbochkov-flexcompute:
A small PR that came out of working on the charge example https://github.com/flexcompute/tidy3d-notebooks/pull/25 and it does two things:
- allows importing data fields from `.vtu` files with specific names if such files contain multiple fields (e.g. "Electrons", "Holes", etc)
- removes spurious warnings when reloading simulations containing `PerturbationMedium` with `CustomChargePerturbation`/`CustomHeatPerturbation`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remote mode solver web api automatically reduces the associated `Simulation` object to the mode solver plane before uploading it to server.
 - All solver output is now compressed. However, it is automatically unpacked to the same `simulation_data.hdf5` by default when loading simulation data from the server.
 - Internal refactor of `adjoint` plugin to separate `jax`-traced fields from regular `tidy3d` fields.
-- Added argument `field` in class method `.from_vtu()` of `TriangularGridDataset` and `TetrahedralGridDataset` for specifying the name of data field to load.
+- Added an optional argument `field` in class method `.from_vtu()` of `TriangularGridDataset` and `TetrahedralGridDataset` for specifying the name of data field to load.
 
 ### Fixed
 - Removed spurious warnings realted to reloading simulation containing `PerturbationMedium` with `CustomChargePerturbation`/`CustomHeatPerturbation`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,8 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remote mode solver web api automatically reduces the associated `Simulation` object to the mode solver plane before uploading it to server.
 - All solver output is now compressed. However, it is automatically unpacked to the same `simulation_data.hdf5` by default when loading simulation data from the server.
 - Internal refactor of `adjoint` plugin to separate `jax`-traced fields from regular `tidy3d` fields.
+- Added argument `field` in class method `.from_vtu()` of `TriangularGridDataset` and `TetrahedralGridDataset` for specifying the name of data field to load.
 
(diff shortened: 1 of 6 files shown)


tests/test_data/test_datasets.py
```diff
@@ -352,6 +374,14 @@ def test_tetrahedral_dataset(tmp_path, ds_name):
     assert np.all(tet_grid._vtk_offsets == np.array([0, 4, 8]))
     assert tet_grid.name == ds_name
 
+    # since vtk is attempted to be imported during a dependent function call
+    # let's call one of those so that vtk dict is properly populated
+    try:
```
tylerflex, line 379: what is the reason behind using a `try / except` here? (when would vtk not be imported in the test?) maybe it could make sense to make this a `pytest.mark.parameterized()` where we run the function once with vtk and once without? maybe I'm just a bit confused about what this block of code is doing here.

tidy3d/components/data/dataset.py
```diff
@@ -777,26 +777,28 @@ def _read_vtkUnstructuredGrid(fname: str):
...
     @requires_vtk
-    def from_vtu(cls, file: str) -> UnstructuredGridDataset:
+    def from_vtu(cls, file: str, field: str = None) -> UnstructuredGridDataset:
         """Load unstructured data from a vtu file.
 
         Parameters
         ----------
         fname : str
             Full path to the .vtu file to load the unstructured data from.
+        field : str
```
tylerflex, line 792: add a ` = None` to ensure users know it's optional

CHANGELOG.md
```diff
@@ -29,8 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remote mode solver web api automatically reduces the associated `Simulation` object to the mode solver plane before uploading it to server.
 - All solver output is now compressed. However, it is automatically unpacked to the same `simulation_data.hdf5` by default when loading simulation data from the server.
 - Internal refactor of `adjoint` plugin to separate `jax`-traced fields from regular `tidy3d` fields.
+- Added argument `field` in class method `.from_vtu()` of `TriangularGridDataset` and `TetrahedralGridDataset` for specifying the name of data field to load.
```
tylerflex, line 32: maybe add word "optional" (Added optional argument...)

tylerflex (commented):
Thanks @dbochkov-flexcompute just a couple minor comments.

tests/test_data/test_datasets.py
dbochkov-flexcompute: Revised this a bit. So, currently it works the following way:
- This test is designed to work for both when `vtk` is present and when it is not. 
- Which branch to check is determined by parameter `no_vtk`.
- I tried using `pytest.mark.parameterized()` to conveniently test both cases, but I couldn't make it work. it seems to be tricky to properly unload already loaded module. So, instead I just re-run the same tests in a separate pytest call [here](https://github.com/flexcompute/tidy3d/pull/1394/files#diff-002670f0bd6def8a37a0f88151b0e337d4d89b38dd18c46b32ea8bda1431c905R9) with a monkeypatch to skip loading `vtk` and `no_vtk=True`

tylerflex (approved):

tests/test_data/test_datasets.py
tylerflex: I see, I think the new implementation is fine.

btw I ran into something similar with wanting to test `adjoint` with and without jax installed, here's what I came up with in the end.

https://github.com/flexcompute/tidy3d/blob/pre/2.6/tests/test_plugins/test_adjoint.py#L1711-L1741

(I guess I also didn't use parameterize, maybe for similar reasons)

PR #1395, opened 2024-01-25
Allow inhomogeneous substrate and superstrate for 2d materials with more complex intersections

dmarek-flex:
Followed the main ideas in @caseyflex's [original PR](https://github.com/flexcompute/tidy3d/pull/1263) and some modifications added by in [Shashwat's PR](https://github.com/flexcompute/tidy3d/pull/1345) but made finding the correct adjacent media more robust by tracking extra information during the subdivision process. 

Limitations 
- intersections of 2D materials with 2D PECs are not reliable
- snapping 2D materials to the grid means that the precise plane where intersections are computed and the final position of the 2D material are slightly different

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Automatic subdivision of 2D materials with inhomogeneous substrate/superstrate.
 - Mode field profiles can be stored directly from a `ModeMonitor` by setting `store_fields_direction`.
 
 ### Changed
--- a/tests/test_components/test_grid_spec.py
+++ b/tests/test_components/test_grid_spec.py
@@ -114,9 +114,11 @@ def test_autogrid_2dmaterials():
         grid_spec=td.GridSpec.auto(),
         run_time=1e-12,
     )
-    assert np.isclose(sim.volumetric_structures[0].geometry.center[2], 1, rtol=RTOL)
+    assert np.isclose(sim.volumetric_structures[0].geometry.bounding_box.center[2], 1, rtol=RTOL)
     grid_dl = sim.discretize(box.geometry).sizes.z[0]
-    assert np.isclose(sim.volumetric_structures[0].geometry.size[2], grid_dl, rtol=RTOL)
+    assert np.isclose(
+        sim.volumetric_structures[0].geometry.bounding_box.size[2], grid_dl, rtol=RTOL
+    )
 
     # now if we increase conductivity, the in-plane grid size should decrease
     sigma2 = 4.5
--- a/tests/test_components/test_simulation.py
+++ b/tests/test_components/test_simulation.py
@@ -1897,7 +1897,9 @@ def test_sim_volumetric_structures(log_capture, tmp_path):  # noqa F811
             run_time=1e-12,
         )
         if isinstance(struct.geometry, td.Box):
-            assert np.isclose(sim.volumetric_structures[0].geometry.size[2], grid_dl, rtol=RTOL)
+            assert np.isclose(
+                sim.volumetric_structures[0].geometry.bounding_box.size[2], grid_dl, rtol=RTOL
+            )
         else:
             assert np.isclose(sim.volumetric_structures[0].geometry.length_axis, grid_dl, rtol=RTOL)
         assert np.isclose(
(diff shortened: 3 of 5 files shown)


tylerflex (commented):
Thanks @dmarek-flex. I did a read through and it looks good. I'll leave the more detailed checking of the logical operations to casey and weiliang though. 

One comment I have though is that this `simulation.py` file is getting extremely long and a lot of this 2D material stuff probably should live in its own file somewhere. I'm wondering whether you or @caseyflex see any way to separate it out a bit from `Simulation` so we can have the gory details of this stuff perhaps in a separate place? or does that seem to difficult given the way it's organized?

caseyflex (approved):
Thanks @dmarek-flex -- this looks really good, I'm glad you were able to get it working. A couple comments:

1. In the case where no subdivision is needed, can we keep the original geometry? Is it needed to convert it to a polyslab?
2. Along the same lines, I was looking again at how this changes treatment of the previously supported case of homogeneous substrate and superstrate. Previously, we snapped to grid, then look up and down one grid length to find neighbors. Now, we find neighbors first, and then snap to grid. I think the new treatment is better for these applications we've been discussing, but I just wanted to make sure I understand the changes.
3. Regarding @tylerflex's point that simulation.py is getting long -- maybe the purely geometric subdivision operations should be moved to some file in the geometry directory.
4. Previously, when the substrate was not inhomogeneous, we gave an error that this was not supported and had to be subdivided manually. Now, if subdivision fails, the error tells the user to create an issue on github. Are there known situations where the new subdivision fails? I guess creating an issue on github should be reserved for unexpected behavior, not just unsupported behavior.
5. Regarding intersections with 2D PEC -- could you just elaborate on that a bit more? I know there's been discussion but I'm not quite clear on it. How is it not reliable? (what happens if you try it)
6. We should check that the example notebooks which use 2d materials still work under the new treatment.

Thanks a lot for this!

dmarek-flex:
Some responses to @caseyflex 

1. Done
2. Yes, we look for neighbors using a small fraction of the grid spacing (`DIST_NEIGHBOR_REL_2D_MED`), then one of the final steps is snapping to the grid. I found that approach to be more consistent.
3. Would geometry/utils.py be a good place? @tylerflex 
4. I removed that error reporting, of course there might be bugs in the future. But the idea should work for most cases. Maybe a warning for 2D materials intersecting would be appropriate.
5. I tried to get a better idea of this issue this morning. I also get some unexpected results even when using  2D finite conductive materials, so I don't think it is just an issue with PECs. There seems to be two main problems: 1) snapping to the grid makes it possible to miss the exact intersection point between two 2D materials when they meet at 90 degrees, 2) when the [volumetric_equivalents](https://github.com/flexcompute/tidy3d/blob/2187624c59a0884bee45624b652ec8920589287e/tidy3d/components/medium.py#L4315) are produced, the normal component of the superstrate material overwrites the normal component of the materials below the 2D surface. This effect is noticeable when trying to connect two conductive materials at 90 degrees, when looking at some of the components of the permittivity I could see an air gap form. But I think these issues are outside the scope of this PR, and the other steps produce the correct polygons and find the correct adjacent media.
6. Is there a list of these notebooks or way to quickly test?

tylerflex:
> Would geometry/utils.py be a good place? @tylerflex

maybe we can move everything into a new file `geometry/utils_2d.py`

weiliangjin2021 (commented):
Nice work! Two questions:
1. `get_neighbors` and `subdivide` seem to call each other recursively. Will there be many recursions during the subdivision?
2. What's the behavior of subdivision with another 2d material? Will be good to add a test to clarify that.

dmarek-flex:
> Nice work! Two questions:
> 
>     1. `get_neighbors` and `subdivide` seem to call each other recursively. Will there be many recursions during the subdivision?
> 
>     2. What's the behavior of subdivision with another 2d material? Will be good to add a test to clarify that.

1) It used to be more like that, but in my changes I don't see any recursion. After each iteration of the loop the newly created subdivided structures have to be added to the background list for later checking, but other than that I think it is pretty straight forward.

2) At this point 2D materials will become an equivalent volumetric structure first before they enter in as a possible superstrate/substrate, so they end up being treated like any other 3D shape. As I mentioned earlier though, I think there is an issue in the logic where the volumetric equivalents are created when considering 2D materials. I am still trying to understand this issue though.

weiliangjin2021 (approved):
> At this point 2D materials will become an equivalent volumetric structure first before they enter in as a possible superstrate/substrate, so they end up being treated like any other 3D shape. As I mentioned earlier though, I think there is an issue in the logic where the volumetric equivalents are created when considering 2D materials. I am still trying to understand this issue though.

That makes sense. It can be revisited in the future.

tylerflex (approved):
Looks good @dmarek-flex. Last thing is that I'm not sure if we need all 5 commits. especially if a few are just formatting. Can we squash to 1-3 depending on how many logical "chunks" of changes are in there before the "rebase and merge"?

PR #1423, opened 2024-01-31
truly asyncronous task running via `web.run_fn_async`

tylerflex:
Kind of an experimental feature, basically adds a `web.run()` that is compatible with [asyncio](https://realpython.com/async-io-python/). So users can write their own asynchronous functions wrapping this function and tasks will run concurrently.

The initial idea of this was to provide users of the Design plugin a way to define their measurement functions without needing pre/post, while allowing for concurrent task running. It solves that problem but still has a learning curve.

### main things I'm still unsure about
1. Naming: unfortunately `run_async` is already taken by essentially the `Batch` wrapper, so I went with `async_run`.
2. Will anyone actually use this? it seems powerful but `asyncio` is pretty difficult to use properly. Wouldn't hurt to have but a bit worried about it being hard to set up. Tried my best to explain stuff in the [notebooks](https://github.com/flexcompute/tidy3d-notebooks/pull/35).
3. Is there a better way to do what I want that completely hides asyncio? Couldn't come up with anything but I am sort of new at this tool.

Any other comments appreciated

### Basic Explanation

When used in an `async` function, the `web.async_run` will continue execution while waiting for `web.monitor()`. This means submission of several tasks can happen while waiting for previously submitted tasks to be monitored. Enables a batch-like functionality without needing to explicitly package `simulations`.

```py
    async def run_task(sim):
        """Async version of running a task."""
        return await async_run(sim, task_name="blah")

    async def main():
        """Call the `run_task` on a collection of simulations and gather results."""
        return await asyncio.gather(*(run_task(sim) for sim in sims))

    # will submit all sims first until they start finishing, then downloads as they become ready.
    asyncio.run(main())
```

```
11:54:50 CST INFO: Started asynchronous task with name 'async_1e-12'.                 
11:54:53 CST INFO: Started asynchronous task with name 'async_2e-12'.                 
11:54:55 CST INFO: Started asynchronous task with name 'async_3e-12'.                 
11:55:16 CST INFO: Finished asynchronous task with name 'async_1e-12'.                
11:55:17 CST INFO: Finished asynchronous task with name 'async_2e-12'.                
11:55:19 CST INFO: Finished asynchronous task with name 'async_3e-12'. 
```

This was also integrated into the `design` plugin so that one can define an asynchronous function `f` involving a `web.run_fn_async()` and pass it to `tdd.Design.run_async(f)` to run asynchronously without having to explicitly batch or write the function in terms of pre and post processing steps.

```py
async def f_async(num: int, t: float):
    sim = pre(num=num, t=t)
    data = await td.web.async_run(sim, task_name=f"num={num}_t={t}", verbose=False)
    return post(data)

result_async = await design_space.run_async(f_async)
```

Notebook / tutorial changes:
https://github.com/flexcompute/tidy3d-notebooks/pull/35

### DIFF
--- a/tidy3d/web/__init__.py
+++ b/tidy3d/web/__init__.py
@@ -27,6 +27,8 @@ from .core import core_config
 from ..log import log, get_logging_console
 from ..version import __version__
 
+from .api.webapi_async import async_run
+
 migrate()
 
 core_config.set_config(log, get_logging_console(), __version__)
@@ -55,4 +57,6 @@ __all__ = [
     "configure",
     "run_async",
     "test",
+    "async_run",
+    "monitor_fn_async",
 ]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Automatic subdivision of 2D materials with inhomogeneous substrate/superstrate.
 - Mode field profiles can be stored directly from a `ModeMonitor` by setting `store_fields_direction`.
+- `tidy3d.web.webapi_async.async_run` function uses `asyncio` to define an asynchronous run function that continues function execution while `await`-ing for tasks being monitored.
+- `tidy3d.plugins.design.Design.run_async(f)` that runs asynchronously over an asynchronous function `f` (using `td.web.async_run` internally).
 
 ### Changed
 - `DataArray.to_hdf5()` accepts both file handles and file paths.
--- a/tests/test_plugins/test_design.py
+++ b/tests/test_plugins/test_design.py
@@ -4,8 +4,10 @@ import numpy as np
 import tidy3d as td
 import matplotlib.pyplot as plt
 import scipy.stats.qmc as qmc
+import asyncio
 
 import tidy3d.web as web
+from tidy3d.web.api.webapi_async import to_async
 from tidy3d.components.base import Tidy3dBaseModel
 
 from tidy3d.plugins import design as tdd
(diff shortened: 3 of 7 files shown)


lucas-flexcompute (commented):
Instead of `run_fn_async`, how about `async_run`? `run_fn_async` does not sound intuitive to me at all…

And I'm not sure this implementation really uses `asyncio` properly, because `monitor` uses `time.sleep` instead of `asyncio.sleep`. As I understand, we'd have to rewrite the web API to use non-blocking async calls to make them really async. I think that if we send many tasks at once, we might actually become slower with this implementation, because each task will block independently.

lucas-flexcompute:
I was reading a little more and the `run_in_executor` does exactly what we need, although `time.sleep` blocks, it continues running other tasks in parallel, so I think this works as a first implementation.

In an ideal one, we'd use HTTP requests wit asyncio support, but that would be a more deep change to the web API because I don't think `requests` support it..

lucas-flexcompute (approved):

tylerflex:
> I was reading a little more and the run_in_executor does exactly what we need, although time.sleep blocks, it continues running other tasks in parallel, so I think this works as a first implementation.

Yea, it's a bit of a quick and dirty implementation, I did try it with 40 cases and it worked as expected but I dont know if it would cause issues for many more..

> In an ideal one, we'd use HTTP requests wit asyncio support, but that would be a more deep change to the web API because I don't think requests support it..

yea I agree, this would be the ideal but seems like it would be a huge overhaul..

tidy3d/web/__init__.py
```diff
@@ -55,4 +57,6 @@
     "configure",
     "run_async",
     "test",
+    "async_run",
+    "monitor_fn_async",
```
dbochkov-flexcompute, line 61: is `monitor_fn_async` imported from anywhere?

dbochkov-flexcompute (commented):
Looks great to me and I think this is very useful! A few non-critical comments/questions:
- would it be better to rename functions that run using `asyncio` capabilities, say, as `run_with_asyncio()`? I just think having both `async_run` and `run_async` can be confusing.
- I see that asyncio version of run functions duplicate the original ones significantly. I wonder if there is a good way to remove that code duplication somehow? Maybe along the lines of
```
def _run(..., asyncio: bool):
    ...

def run(...):
    return _run(..., asyncio=False)

async def async_run(...):
    return _run(..., asyncio=True)
```
I realize this is definitely not the most elegant solution. Also, all of these function are pretty short anyway, so maybe it's not worth worrying about it.
- Can we use `to_async()` on user defined non-asyncio functions to quickly convert them? Just thinking, when putting together a study the user might want to create a start with a usual function for easier debugging, would be nice to be able to automatically convert it into asyncio one

tidy3d/web/__init__.py
tylerflex: no, that got left in by accident, will remove it, thanks!

tylerflex:
Thanks @dbochkov-flexcompute.

> Can we use to_async() on user defined non-asyncio functions to quickly convert them? Just thinking, when putting together a study the user might want to create a start with a usual function for easier debugging, would be nice to be able to automatically convert it into asyncio one

I dont know exactly how this would work. Let's say the user has a function
```py
def f(*args):
    sim = make_sim(*args)
    data = td.web.run(sim)
    return postprocess(data)
```
the version using async run would look like
```py
async def f(*args):
    sim = make_sim(*args)
    data = await td.web.run_asyncio(sim)
    return postprocess(data)
```
but if the user just applied `f = to_async(f)`, I think it might not have the desired effect of awaiting each individual task. It's an interesting thought though, I will try it out.

tylerflex:
closing until we decide to take this up again

PR #1443, opened 2024-02-05
added lumped ports to smatrix plugin

dmarek-flex:
Implements the original frontend changes by Shashwat in this [PR](https://github.com/flexcompute/tidy3d/pull/1341)

A couple of additions:

- Checks and possibly forces the lumped port to snap to the grid, since the sheet resistance is also automatically snapped to the grid
- Computes the current at each port using a complete contour, which alleviates some of the need for port refinement
- Fixes for some unit tests

Known limitations:
Not directly related to this PR, but 2D material intersections are a known issue at the moment. The lumped ports are 2D materials, which means they must connect to fully 3D materials. 

Supersedes [PR 1341](https://github.com/flexcompute/tidy3d/pull/1341)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `Simulation` now accepts `LumpedElementType`, which currently only supports the `LumpedResistor` type. `LumpedPort` together with `LumpedResistor` make up the new `TerminalComponentModeler` in the `smatrix` plugin.
 - Uniaxial medium Lithium niobate to material library.
 - Added support for conformal mesh methods near PEC structures that can be specified through the field `pec_conformal_mesh_spec` in the `Simulation` class.
 
--- a/tests/test_plugins/test_terminal_component_modeler.py
+++ b/tests/test_plugins/test_terminal_component_modeler.py
@@ -1,0 +1,140 @@
+import pytest
+import numpy as np
+import pydantic.v1 as pydantic
+import matplotlib.pyplot as plt
+
+import tidy3d as td
+from tidy3d.plugins.smatrix import (
+    AbstractComponentModeler,
+    LumpedPortDataArray,
+    TerminalComponentModeler,
+)
+from tidy3d.exceptions import Tidy3dKeyError
+from ..utils import run_emulated
+from .terminal_component_modeler_def import make_component_modeler
+
+
+def run_component_modeler(monkeypatch, modeler: TerminalComponentModeler):
+    sim_dict = modeler.sim_dict
+    batch_data = {task_name: run_emulated(sim) for task_name, sim in sim_dict.items()}
+    monkeypatch.setattr(TerminalComponentModeler, "_run_sims", lambda self, path_dir: batch_data)
+    # for the random data, the power wave matrix might be singular, leading to an error
+    # during inversion, so monkeypatch the inv method so that it operates on a dummy
+    # identity matrix
+    monkeypatch.setattr(AbstractComponentModeler, "inv", lambda matrix: np.eye(len(modeler.ports)))
+    s_matrix = modeler.run(path_dir=modeler.path_dir)
+    return s_matrix
+
+
+def test_validate_no_sources(tmp_path):
+    modeler = make_component_modeler(planar_pec=True, path_dir=str(tmp_path))
+    source = td.PointDipole(
+        source_time=td.GaussianPulse(freq0=2e14, fwidth=1e14), polarization="Ex"
+    )
+    sim_w_source = modeler.simulation.copy(update=dict(sources=(source,)))
+    with pytest.raises(pydantic.ValidationError):
+        _ = modeler.copy(update=dict(simulation=sim_w_source))
+
+
+def test_no_port(tmp_path):
+    modeler = make_component_modeler(planar_pec=True, path_dir=str(tmp_path))
+    _ = modeler.ports
+    with pytest.raises(Tidy3dKeyError):
+        modeler.get_port_by_name(port_name="NOT_A_PORT")
+
+
+def test_plot_sim(tmp_path):
+    modeler = make_component_modeler(planar_pec=False, path_dir=str(tmp_path))
+    modeler.plot_sim(z=0)
+    plt.close()
+
+
+def test_plot_sim_eps(tmp_path):
+    modeler = make_component_modeler(planar_pec=False, path_dir=str(tmp_path))
+    modeler.plot_sim_eps(z=0)
+    plt.close()
+
+
+def test_make_component_modeler(tmp_path):
+    _ = make_component_modeler(planar_pec=False, path_dir=str(tmp_path))
+
+
+def test_run(monkeypatch, tmp_path):
+    modeler = make_component_modeler(planar_pec=True, path_dir=str(tmp_path))
+    monkeypatch.setattr(TerminalComponentModeler, "run", lambda self, path_dir: None)
+    modeler.run(path_dir=str(tmp_path))
+
+
+def test_run_component_modeler(monkeypatch, tmp_path):
+    modeler = make_component_modeler(planar_pec=True, path_dir=str(tmp_path))
+    s_matrix = run_component_modeler(monkeypatch, modeler)
+
+    for port_in in modeler.ports:
+        for port_out in modeler.ports:
+            coords_in = dict(port_in=port_in.name)
+            coords_out = dict(port_out=port_out.name)
+
+            assert np.all(s_matrix.sel(**coords_in) != 0), "source index not present in S matrix"
+            assert np.all(
+                s_matrix.sel(**coords_in).sel(**coords_out) != 0
+            ), "monitor index not present in S matrix"
+
+
+def test_s_to_z_component_modeler():
+    # Test case is 2 port T network with reference impedance of 50 Ohm
+    A = 20 + 30j
+    B = 50 - 15j
+    C = 60
+
+    Z11 = A + C
+    Z21 = C
+    Z12 = C
+    Z22 = B + C
+
+    Z0 = 50.0
+    # Manual creation of S parameters Pozar Table 4.2
+    deltaZ = (Z11 + Z0) * (Z22 + Z0) - Z12 * Z21
+    S11 = ((Z11 - Z0) * (Z22 + Z0) - Z12 * Z21) / deltaZ
+    S12 = (2 * Z12 * Z0) / deltaZ
+    S21 = (2 * Z21 * Z0) / deltaZ
+    S22 = ((Z11 + Z0) * (Z22 - Z0) - Z12 * Z21) / deltaZ
+
+    port_names = ["lumped_port_1", "lumped_port_2"]
+    freqs = [1e8]
+
+    values = np.array(
+        [[[S11, S12], [S21, S22]]],
+        dtype=complex,
+    )
+    # Put coords in opposite order to check reordering
+    coords = dict(
+        f=np.array(freqs),
+        port_out=port_names,
+        port_in=port_names,
+    )
+
+    s_matrix = LumpedPortDataArray(data=values, coords=coords)
+    z_matrix = AbstractComponentModeler.s_to_z(s_matrix, reference=Z0)
+    z_matrix_at_f = z_matrix.sel(f=1e8)
+    assert np.isclose(z_matrix_at_f[0, 0], Z11)
+    assert np.isclose(z_matrix_at_f[0, 1], Z12)
+    assert np.isclose(z_matrix_at_f[1, 0], Z21)
+    assert np.isclose(z_matrix_at_f[1, 1], Z22)
+
+
+def test_ab_to_s_component_modeler():
+    coords = dict(
+        f=np.array([1e8]),
+        port_out=["lumped_port_1", "lumped_port_2"],
+        port_in=["lumped_port_1", "lumped_port_2"],
+    )
+    # Common case is reference impedance matched to loads, which means ideally
+    # the a matrix would be an identity matrix, and as a result the s matrix will be
+    # given directly by the b_matrix
+    a_values = np.eye(2, 2)
+    a_values = np.reshape(a_values, (1, 2, 2))
+    b_values = (1 + 1j) * np.random.random((1, 2, 2))
+    a_matrix = LumpedPortDataArray(data=a_values, coords=coords)
+    b_matrix = LumpedPortDataArray(data=b_values, coords=coords)
+    S_matrix = AbstractComponentModeler.ab_to_s(a_matrix, b_matrix)
+    assert np.isclose(S_matrix, b_matrix).all()
--- a/tidy3d/components/lumped_element.py
+++ b/tidy3d/components/lumped_element.py
@@ -1,0 +1,130 @@
+"""Defines lumped elements that should be included in the simulation."""
+from __future__ import annotations
+from abc import ABC
+from typing import Union, List, Optional
+
+import pydantic.v1 as pydantic
+
+from .base import cached_property, skip_if_fields_missing
+
+from .geometry.base import Box
+from ..components.structure import MeshOverrideStructure
+from .types import Axis
+from .viz import PlotParams, plot_params_lumped_element
+from ..constants import OHM
+from ..components.validators import validate_name_str, assert_plane
+from ..exceptions import ValidationError
+
+DEFAULT_LUMPED_ELEMENT_NUM_CELLS = 3
+
+
+class LumpedElement(Box, ABC):
+    """Base class describing lumped elements."""
+
+    name: str = pydantic.Field(
+        ...,
+        title="Name",
+        description="Unique name for the lumped element.",
+        min_length=1,
+    )
+
+    voltage_axis: Axis = pydantic.Field(
+        ...,
+        title="Voltage Drop Axis",
+        description="Specifies the axis along which the component is oriented and along which the "
+        "associated voltage drop will occur. Must be in the plane of the element.",
+    )
+
+    num_grid_cells: Optional[pydantic.PositiveInt] = pydantic.Field(
+        DEFAULT_LUMPED_ELEMENT_NUM_CELLS,
+        title="Lumped element grid cells",
+        description="Number of mesh grid cells associated with the lumped element along each direction. "
+        "Used in generating the suggested list of :class:`MeshOverrideStructure` objects."
+        "A value of ``None`` will turn off mesh refinement suggestions.",
+    )
+
+    _name_validator = validate_name_str()
+    _plane_validator = assert_plane()
+
+    @cached_property
+    def normal_axis(self):
+        """Normal axis of the lumped element."""
+        return self.size.index(0.0)
+
+    @cached_property
+    def plot_params(self) -> PlotParams:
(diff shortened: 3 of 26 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added `LumpedResistor` and`LumpedPort`, which make up the new `TerminalComponentModeler` in the `smatrix` plugin
```
tylerflex, line 9: space after `and`. Also, I think some mention of the fact that `Simulation` accepts `lumped_elements` could be useful.

tests/test_plugins/test_terminal_component_modeler.py
tylerflex, line 1: I think we need some tests regarding if lumped elements are added to `Simulation` directly. What is the expected behavior there?

tidy3d/components/lumped_element.py
```diff
@@ -0,0 +1,81 @@
...
+        if size and size.count(0.0) == 1 and size.index(0.0) == val:
+            # if not planar, then a separate validator should be triggered, not this one
+            raise ValidationError(
+                f"'voltage_axis' must be in the plane of lumped element '{name}'."
+            )
+        return val
+
+
+class LumpedResistor(LumpedElement):
+    """Base class describing lumped elements."""
```
tylerflex, line 64: Update this description. Also, perhaps add a bit more information about what a LumpedResistor is and how it is added to simulations etc.

tidy3d/components/lumped_element.py
```diff
@@ -0,0 +1,81 @@
...
+        ...,
+        title="Resistance",
+        description="Resistance value in ohms.",
+        unit=OHM,
+    )
+
+    @cached_property
+    def sheet_conductance(self):
+        """Effective sheet conductance."""
+        lateral_axis = 3 - self.voltage_axis - self.normal_axis
```
tylerflex, line 76: huh. I guess this works? :D

tidy3d/components/simulation.py
```diff
@@ -345,6 +352,13 @@ class Simulation(AbstractSimulation):
         All the monitor implementations.
     """
 
+    lumped_elements: Tuple[LumpedElementType, ...] = pydantic.Field(
+        (),
+        title="Lumped Elements",
+        description="Tuple of lumped elements in the simulation. "
+        "Only ``LumpedResistor`` is supported currently.",
```
tylerflex, line 359: make a class link through
```
:class:`LumpedResistor`
```

tidy3d/components/simulation.py
```diff
@@ -345,6 +352,13 @@ class Simulation(AbstractSimulation):
         All the monitor implementations.
     """
 
+    lumped_elements: Tuple[LumpedElementType, ...] = pydantic.Field(
```
tylerflex, line 355: so what happens if this is added to the simulation? Does it act as a source? as a medium? I think we need more detailed explanation.

tidy3d/components/simulation.py
```diff
@@ -3255,17 +3329,44 @@ def get_dls(geom: Geometry, axis: Axis, num_dls: int) -> List[float]:
...
+                if tan_dirs[0] == lumped_element.voltage_axis:
+                    medium_dict = {
+                        "ss": Medium(conductivity=conductivity),
+                        # "tt": Medium(conductivity=conductivity),
+                        "tt": self.medium,
+                    }
+                else:
+                    medium_dict = {
+                        "tt": Medium(conductivity=conductivity),
+                        # "ss": Medium(conductivity=conductivity),
```
tylerflex, line 3353: remove comments?

tidy3d/components/viz.py
tylerflex, line 1: can you provide a screenshot showing how this looks?

tidy3d/plugins/smatrix/smatrix.py
```diff
@@ -69,50 +137,270 @@ class SMatrixDataArray(DataArray):
...
+    path_dir: str = pd.Field(
+        DEFAULT_DATA_DIR,
+        title="Directory Path",
+        description="Base directory where data and batch will be downloaded.",
+    )
+
+    solver_version: str = pd.Field(
+        None,
+        title="Solver Version",
+        description_str="Custom solver version to use, "
```
tylerflex, line 232: instead of `, otherwise`. Let's just do `. If not supplied, uses default ...`

tidy3d/plugins/smatrix/smatrix.py
```diff
@@ -69,50 +137,270 @@ class SMatrixDataArray(DataArray):
...
+
+    def load(self, path_dir: str = DEFAULT_DATA_DIR) -> DataArray:
+        """Load a scattering matrix from saved `BatchData` object."""
+        path_dir = self.get_path_dir(path_dir)
+
+        batch_data = BatchData.load(path_dir=path_dir)
+        return self._construct_smatrix(batch_data=batch_data)
+
+    @staticmethod
+    def s_to_z(s_matrix: DataArray, reference: Complex) -> DataArray:
```
tylerflex, line 343: let's make sure these operations are tested in the unit tests (if not already). Like that they give expected matrices given some example inputs.

tidy3d/plugins/smatrix/smatrix.py
tylerflex, line 1: let's take this opportunity to split this file up into several files (eg. port.py, lumped_port.py, component_modeler.py, etc.). It already should have been broken up before now but I think now is the best time to do it.

tylerflex (changes requested):
Thanks @dmarek-flex ! I took a first pass at sort of a high level. Will be really nice to have this. Just have a few main questions.

1. What happens if LumpedElements are added to a regular Simulation right now? Does it affect the FDTD simulation? When would it be used?
2. We should either modify our existing SMatrix.ipynb example to add the new component modeler or even better make an entirely new notebook to showcase this feature.
3. We should also make a notebook (or modify notebook) showing the lumped elements added directly to the simulation in whatever use case there is for that. Do you have something we could modify (eg a test) for that?

tidy3d/components/source.py
```diff
@@ -466,6 +466,14 @@ class ReverseInterpolatedSource(Source):
         "placement at the specified location using linear interpolation.",
     )
 
+    confine_to_bounds: bool = pydantic.Field(
+        False,
+        title="Confine to Analytical Bounds",
+        description="If ``True``, any source amplitudes which, after discretization, fall beyond "
+        "the analytical bounds of the source are zeroed out, but only along directions where "
+        "the source has a non-zero extent.",
```
weiliangjin2021, line 474: Could you elaborate in the description on when this should be `True`.

tidy3d/plugins/smatrix/smatrix.py
```diff
@@ -48,12 +57,71 @@ class Port(Box):
...
+        description="Number of mesh grid cells associated with the port along each direction. "
+        "This is used only if ``refine_mesh`` is ``True``.",
+    )
+
+    _plane_validator = assert_plane()
+
+    @cached_property
+    def injection_axis(self):
+        """Injection axis of the port."""
+        return self.size.index(0.0)
```
weiliangjin2021, line 103: what about a 2D simulation where the lumped port becomes 1D?

weiliangjin2021 (commented):
Nice work! One main question I have is that your smatrix plugin provides lots of convenience in generating refined mesh, snapping, etc.. But if someone defines lumped elements in simulation objects directly, they'll need to repeat the code in smatrix plugin to define grids, etc.

My suggestion is to implement mesh refinement and snapping as a simulation method, similar to volumetric_equivalence. The users can generate recommended grids from the method, and another method to snap the lumped elements. They are optional features in the simulation class, but applied automatically in smatrix plugin. What do you think?

tidy3d/components/simulation.py
```diff
@@ -3297,8 +3398,9 @@ def snap_to_grid(geom: Geometry, axis: Axis) -> Geometry:
                 temp_structure = structure.updated_copy(geometry=temp_geometry, medium=new_medium)
 
                 if structure.medium.is_pec:
-                    pec_delta = fp_eps * max(np.abs(snapped_center), 1.0)
-                    new_bounds = (snapped_center - pec_delta, snapped_center + pec_delta)
+                    pec_plus = np.nextafter(snapped_center, +np.inf, dtype=np.float32)
+                    pec_minus = np.nextafter(snapped_center, -np.inf, dtype=np.float32)
+                    new_bounds = (pec_minus, pec_plus)
```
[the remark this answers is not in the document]
dmarek-flex: It's not equivalent, but it is a more robust method to add a small increment to a floating point number, which I think is the goal of this step. Simply adding fp_eps is not correct, since fp_eps is the machine precision with respect to 1.0. Adding fp_eps to a 32bit float larger than 1 will not change it.

Maybe I misinterpreted the purpose of this step, but if you want the smallest increment for a given float size then the new changes are preferred.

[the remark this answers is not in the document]
dmarek-flex: I saw this usage in a couple of different places and I had an initial quick fix, but I later found the np.nextafter function which is better.

tidy3d/components/lumped_element.py
dmarek-flex: Yea it does work :), and disclaimer that most of the code is originally Shashwat's so I can't take credit for the cleverness!

tidy3d/components/lumped_element.py
```diff
@@ -0,0 +1,81 @@
...
+        return self.size[self.voltage_axis] / self.size[lateral_axis] / self.resistance
```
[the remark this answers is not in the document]
dmarek-flex: The resistance is now enforced to be a positive float.

tidy3d/components/viz.py
dmarek-flex: Screenshot from 2024-02-08 13-52-46

Blue is lumped resistor, yellow indicates conductors, and light blue is a dielectric medium

dmarek-flex: `plot_3d()` doesn't seem to work at the moment. The lumped resistor is not showing up.

tylerflex: it probably isn't registered with plot_3d yet (this function is actually implemented in the web GUI side and we download the html), will need to coordinate with xin to get support for that.

tidy3d/plugins/smatrix/smatrix.py
dmarek-flex: Yea, didn't consider that. Not sure how to best handle this. I guess a 2D simulation might make sense as long as the injection axis and voltage axis are the dimensions that are part of the simulation.

dmarek-flex: I added a validator to `TerminalComponentModeler` to ensure that the underlying simulation is 3D. I can't think of an application at the moment where it would be useful to include 2D simulations.

dmarek-flex:
> Nice work! One main question I have is that your smatrix plugin provides lots of convenience in generating refined mesh, snapping, etc.. But if someone defines lumped elements in simulation objects directly, they'll need to repeat the code in smatrix plugin to define grids, etc.
> 
> My suggestion is to implement mesh refinement and snapping as a simulation method, similar to volumetric_equivalence. The users can generate recommended grids from the method, and another method to snap the lumped elements. They are optional features in the simulation class, but applied automatically in smatrix plugin. What do you think?

Let me know if the way I exposed the mesh override is to your liking @weiliangjin2021. Snapping to the grid will automatically happen for any 2D material (which LumpedElements are converted to). The reason why I need it in the TerminalComponentModeler is so that the current source and monitor are exactly coincident with the LumpedResistor when it is ultimately snapped to the grid. Might not be useful in other cases.

tylerflex (approved):
Thanks for taking care of these comments @dmarek-flex ! Looks good!

The only thing I think we should do is to add module level docstrings at the top of the new `smatrix/` files, just describing what the module is defining. I think ruff doesn't check this but we try to do it for all modules.

Other than that, I approve

tidy3d/components/simulation.py
```diff
@@ -69,6 +71,9 @@
 # maximum number of sources
 MAX_NUM_SOURCES = 1000
 
+# maximum number of lumped elements
+MAX_NUM_LUMPED_ELEMENTS = 65530
```
weiliangjin2021, line 75: I wonder if it's necessary to bound the maximal number of elements. I think this number 65530 is coming from the upper limit of the number of materials in the simulation, and is validated [here](https://github.com/flexcompute/tidy3d/blob/722ebe641a0faaff6cd6f56cb2349a3d5adc7430/tidy3d/components/scene.py#L78-L92). Or do I miss anything? If so, I'm not sure if we should move the `lumped_elements` field to scene.py or add a validator in simulation to count the total number of medium for structures + lumped_elements. The latter is easier I think.

tidy3d/components/lumped_element.py
```diff
@@ -0,0 +1,140 @@
...
+        description="Specifies the axis along which the component is oriented and along which the "
+        "associated voltage drop will occur. Must be in the plane of the element.",
+    )
+
+    # refine_mesh: bool = pydantic.Field(
+    #     True,
+    #     title="Lumped element mesh refinement",
+    #     description="If ``True``, enables mesh refinement in the lumped element region to ensure that "
+    #     "a port voltage and current can be computed accurately. May also be used without ports.",
+    # )
```
weiliangjin2021, line 45: To be removed?

tidy3d/components/simulation.py
```diff
@@ -3670,3 +3823,15 @@ def subsection(
         )
 
         return new_sim
+
+    def suggest_mesh_overrides(self) -> List[MeshOverrideStructure]:
```
weiliangjin2021, line 3827: In the future, we might want to suggest for other structures, e.g. bulk PEC. so perhaps we should change the function signature to `(self, **kwargs)` to allow for extension?

weiliangjin2021 (commented):
All look good. Thanks for the enormous efforts (4000 lines!). Just a few minor comments.

tidy3d/components/source.py
```diff
@@ -466,6 +466,15 @@ class ReverseInterpolatedSource(Source):
         "placement at the specified location using linear interpolation.",
     )
 
+    confine_to_bounds: bool = pydantic.Field(
+        False,
+        title="Confine to Analytical Bounds",
+        description="If ``True``, any source amplitudes which, after discretization, fall beyond "
+        "the analytical bounds of the source are zeroed out, but only along directions where "
```
weiliangjin2021, line 473: How about changing the word "bounds" to "bounding box", as "bounds" can refer to multiple things? Also mention that the bounding box is inclusive. So perhaps "fall outside the bounding box of the source are zeroed out, but... The bounding box is inclusive."

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -0,0 +1,425 @@
...
+        plot_sources = []
+        for port_source in self.ports:
+            source_0 = self.to_source(port=port_source, snap_center=None)
+            plot_sources.append(source_0)
+        sim_plot = self.simulation.copy(update=dict(sources=plot_sources))
+        return sim_plot.plot_eps(x=x, y=y, z=z, ax=ax, **kwargs)
+
+    def to_source(self, port: LumpedPort, snap_center: float) -> UniformCurrentSource:
+        """Create a current source from a given lumped port."""
+        # The source is expanded slightly to make sure that, when discretized, it covers the
```
weiliangjin2021, line 68: Could you remind me where the source is slightly expanded, I don't find that in this function?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -0,0 +1,425 @@
...
+                    len(e_coords[port.voltage_axis]),
+                )
+            }
+            e_field = e_field.interp(**e_coords_interp)
+            voltage = -e_field.integrate(coord=e_component).squeeze(drop=True)
+            # Return data array of voltage with coordinates of frequency
+            return voltage
+
+        def port_current(port: LumpedPort, sim_data: SimulationData) -> xr.DataArray:
+            """Helper to compute current flowing through the port."""
```
weiliangjin2021, line 315: Would be helpful to add more explanation here on how current is computed. `port_voltage` is quite clear, but `port_current` seems quite complicated.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -0,0 +1,425 @@
...
+            # Integration that corresponds with the contribution to current from cap contours
+            hcap_current = (
+                ((hcap_plus - hcap_minus) * dcap).squeeze(drop=True).reset_coords(drop=True)
+            )
+            current = current + hcap_current
+
+            if port.current_axis != (port.voltage_axis + 1) % 3:
+                current *= -1
+            # Return data array of current with coordinates of frequency
+            return current
```
weiliangjin2021, line 383: If we are to compute impedance from the mode solver result, is the typical computation following the same approach here: compute voltage, current, and then division?

weiliangjin2021 (commented):
A few more comments, but mostly for clarification.

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: I think that is an old comment, I don't see any place where that occurs.

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: You only need the voltage and current to compute the scattering parameters. Computing the impedance would be a separate operation that could be useful though. Using a mode solver result the voltage and current would be computed differently using 2D integrals on the transverse plane.

dmarek-flex: You might need to compute impedance to choose a proper load and/or reference impedance.

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: I added a small diagram, which should help

weiliangjin2021 (commented):
Two comments I have (discussed with @dmarek-flex briefly earlier) are:

- In terminal.py, move `to_source` , `to_current_monitor`, `to_voltage_monitor`, and `to_load` methods to `LumpedPort` class, as they are generated from the lumped port, quite isolated from the rest.
- Remove `refine_mesh` field in lumped_port, but use `num_grid_cells=None` to indicate no mesh refinement on the ports. Or add `refine_mesh` field to lumped_resistor. Basically, lumped_port and lumped_resistor should have the same fields regarding the mesh.

What do you think @tylerflex?

tylerflex:
sounds good to me

tidy3d/plugins/smatrix/ports/lumped.py
```diff
@@ -0,0 +1,173 @@
...
+    )
+
+    impedance: Complex = pd.Field(
+        50,
+        title="Reference impedance",
+        description="Reference port impedance for scattering parameter computation.",
+        units=OHM,
+    )
+
+    num_grid_cells: Union[pd.PositiveInt, None] = pd.Field(
```
weiliangjin2021, line 65: I think you can do `Optional[pd.PostiveInt]` where `Optional` is imported from `typing`.

weiliangjin2021 (approved):

PR #1447, opened 2024-02-06
fix obscure error if val not supported in SimData.plot_field

tylerflex:
If `SimulationData.plot_field(..., val="int")`, it would through an obscure unbound local error. Now it throws a KeyError telling the user to change `val` to a supported value.


lucas-flexcompute (approved):
Great! I missed that when proposed these changes!

tylerflex:
no worries, I only just stumbled on it since I always forget the syntax for plot_field

PR #1449, opened 2024-02-06
Warn if periodic boundaries are used with an angled source

### LINKED ISSUE #1444: Validator for angled plane wave with periodic boundaries?
So far a few users have reported simulation with angled plane wave and the result is not changing with angle. It turns out that they started with normal incidence and periodic boundary condition. When changing the incident angle, they didn't switch to Bloch boundaries. Can we add a warning or even validation error for periodic boundary with angled plane wave? Obviously need to be a bit careful here as only the direction where the wavevector component is nonzero needs to be Bloch. @momchil-flex feel free to assign to anyone you see fit.

momchil-flex:
Fixes #1444 

Interestingly, we already had such a validator. We warn in various cases, e.g. if the bloch vector of the boundaries does not match the angle of the source. However we were missing the case of periodic boundaries, which are a separate class from Bloch.

It would perhaps make sense to change the warning to error. I am not sure if there's any use case in which we want the simulation to run with a mismatched source angle. But, in order to avoid anything unexpected, I think keeping it a warning is also fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add dispersion information to dataframe output when available from mode solver under the column "dispersion (ps/(nm km))".
 - Skip adjoint source for diffraction amplitudes of NaN.
 - Helpful error message if `val` supplied to `SimulationData.plot_field` not supported.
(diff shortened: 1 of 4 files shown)


momchil-flex:
You're right, I meant to say "Fixed validator that warns if angled plane wave does not match bloch vector". But I'll rephrase to

"Fixed validator that warns if angled plane wave does not match the simulation boundaries, which was not warning for periodic boundaries."

PR #1453, opened 2024-02-07
Added EME solver

caseyflex:
I left some comments below about some key aspects of the organization.

Todo:

- [x] Fix derived fields plotting
- [x] Disallow fully anisotropic materials. Warn about other advanced materials.
- [x] Unit tests
- [x] Subgrid visualization
- [x] plot_grid
- [x] precision in `storage_size`
- [x] plot port modes and mode solver modes and coeffs
- [x] mode_indices -> num_modes in monitors
- [x] finalize or get rid of methods for converting to / from fdtd sim
- [x] Create `AbstractYeeGridSimulation` as abstract superclass of `EMESimulation` and `Simulation`, and similarly for `sim_data`.


tidy3d/components/data/data_array.py
```diff
@@ -650,6 +653,47 @@ class HeatDataArray(DataArray):
...
+    >>> port_index = [0, 1]
+    >>> coords = dict(x=x, y=y, z=z, f=f, mode_index=mode_index, port_index=port_index)
+    >>> fd = ScalarModeFieldDataArray((1+1j) * np.random.random((2,3,4,2,5,2)), coords=coords)
+    """
+
+    __slots__ = ()
+    _dims = ("x", "y", "z", "f", "mode_index", "port_index")
+
+
+class EMESMatrixDataArray(DataArray):
```
caseyflex, line 676: This is different than the smatrix plugin data array, which allows more than two ports. This, together with the EMESMatrixDataset which has S11, S12, S21, S22, is better for the two-port case. It doesn't have the problem of possible different num_modes at the two ports. Also, we can't import smatrix plugin from components.

tidy3d/components/eme/data/dataset.py
```diff
@@ -0,0 +1,95 @@
...
+        description="S matrix relating output modes at port 2 to input modes at port 1.",
+    )
+    S22: EMESMatrixDataArray = pd.Field(
+        ...,
+        title="S22 matrix",
+        description="S matrix relating output modes at port 2 to input modes at port 2.",
+    )
+
+
+class EMECoefficientDataset(Dataset):
```
caseyflex, line 36: Same data structure as EMESMatrixDataset, but the components have different meaning, so better to have it as a different class.

tidy3d/components/eme/data/sim_data.py
```diff
@@ -0,0 +1,437 @@
...
+            if np.any(np.iscomplex(data.values)):
+                data *= np.exp(1j * phase)
+            else:
+                log.warning(
+                    f"Non-zero phase of {phase} specified but the data being plotted is "
+                    "real-valued. The phase will be ignored in the plot."
+                )
+        return data
+
+    def plot_field(
```
caseyflex, line 54: Unfortunately, I had to copy this from `components/sim_data.py`. Would it make sense to move some of this field plotting functionality into base_sim?

tidy3d/components/eme/simulation.py
```diff
@@ -0,0 +1,335 @@
...
+    )
+
+    monitors: Tuple[EMEMonitorType, ...] = pd.Field(
+        (),
+        title="Monitors",
+        description="Tuple of monitors in the simulation. "
+        "Note: monitor names are used to access data after simulation is run.",
+    )
+
+    boundary_spec: BoundarySpec = pd.Field(
```
caseyflex, line 100: I will probably disallow boundary_spec and sources. Boundary conditions are incorporated instead in ModeSpec, and sources that aren't in the EME modes will be handled in the smatrix plugin.

tidy3d/components/eme/simulation.py
```diff
@@ -0,0 +1,335 @@
...
+            structures=self.structures,
+            symmetry=self.symmetry,
+            grid_spec=self.grid_spec,
+            boundary_spec=self.boundary_spec,
+            version=self.version,
+            **kwargs,
+        )
+
+    @cached_property
+    def grid(self) -> Grid:
```
caseyflex, line 326: Currently, the grid is generated by the grid_spec using the ordinary grid generation code, by converting the sim to an fdtd sim first. This is probably the best way if you just care about the EME simulation. However, it is possible that you have an FDTD simulation with a pre-existing grid, and you want to make sure the EME simulation is using exactly the same grid. In this case, the grid generation here won't generate exactly the same grid. So I might add another field for if the user wants to specify an actual grid to override the grid_spec.

tidy3d/components/eme/data/sim_data.py
momchil-flex: I think what would rather make sense is to have some intermediate level abstract class from which both EME and FDTD simulations inherit (and probably mode in the future, as `ModeSolver` has a `plot_field` method too), but not Heat and e.g. charge in the future. It is either that or decoupling the plot function completely in 3.0, something like `plot_field(sim_data, ...)` instead of `sim_data.plot_field(...)`. But for the time being an `AbstractElectromagneticSimulationData` with a `plot_field` method that both `Simulation` and `EMESimultion` inherit maybe makes most sense?

momchil-flex: @tylerflex @daquintero

tidy3d/components/data/data_array.py
```diff
@@ -650,6 +653,47 @@ class HeatDataArray(DataArray):
     _dims = "T"
 
 
+class EMEScalarFieldDataArray(AbstractSpatialDataArray):
```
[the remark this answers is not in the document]
caseyflex: For the smatrix plugin, I'm referring to this:
https://github.com/flexcompute/tidy3d/blob/e5c1439074d56b76c6ae15c54c46471204ac0a01/tidy3d/plugins/smatrix/component_modelers/modal.py#L302
Each port may have a different number of modes, so it just uses the max when setting up the DataArray coords.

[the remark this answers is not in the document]
caseyflex: yeah, I actually initially had it behave something like that. The difficulty I had was that I wanted `EMEFieldData` to inherit from `ElectromagneticFieldData`. In this setup, `EMEPortFieldData` could inherit from `ElectromagneticFieldData`, except that there is supposed to be a one-to-one correspondence between `Monitor`s and `MonitorData`s.

tidy3d/components/data/dataset.py
```diff
@@ -198,6 +199,32 @@ def symmetry_eigenvalues(self) -> Dict[str, Callable[[Axis], float]]:
             Hz=lambda dim: +1 if (dim == 2) else -1,
         )
 
+    def _restrict_to_box(self, box: Box):
```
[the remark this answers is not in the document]
caseyflex: in the backend... I'll move it there...

tidy3d/components/eme/data/sim_data.py
caseyflex: that sounds reasonable to me

tidy3d/components/data/dataset.py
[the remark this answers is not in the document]
caseyflex: it's used for `EMEModeSolverMonitor`. Currently, the behavior is to take the EME modes (which span the entire simulation in the tangential directions) and restrict them to the `EMEModeSolverMonitor` geometry.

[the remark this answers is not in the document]
caseyflex: The EMEModeSolverData has a list of ModeSolverData, each of which is associated to a ModeSolverMonitor. These monitors aren't in the list of simulation monitors, but rather they are just used internally while computing modes. But they are tied to the ModeSolverData that is returned.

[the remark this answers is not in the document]
caseyflex: I think I'll change this behavior though, to just store the full `ModeSolverData` on any mode plane intersecting the monitor geometry.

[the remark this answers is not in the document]
caseyflex: because otherwise I'd probably want to restrict the `ModeSolverMonitor` as well, but those are required to be 2D and could often become 1D after restriction

tidy3d/components/data/data_array.py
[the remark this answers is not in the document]
caseyflex: `port_index` in `EMEFieldData` labels the actual device port. `EMEFieldData` has nothing to do with the EME grid, it is rather defined on the FDTD grid.

tidy3d/components/eme/data/sim_data.py
tylerflex: Interestingly, this is the kind of problem we're trying to fix for 3.0 :) Now that we have all these types of simulations, it gets really annoying to have to use inheritance to specify how all of these methods work.

I think the approach momchil suggested probably makes some sense for now.

However, how I'm thinking about it for 3.0 is that the simulation has some properties or methods that simply separate the optical parts out, and then we have some `plot_optical` or `plot_thermal` etc. that filter by these properties and plot them in a generic way.

Related, but I'm thinking that the Simulation and SimulationData might want to be unified in a single object, so `SimulationData.plot_field(mnt_name)` would just work as is now and just continue to error if the `mnt_name` was not a `FieldMonitor`.

tylerflex: sorry let me clarify, I dont mean that in 3.0 Simualtion gets unified with SimulationData. I mean that maybe we want to throw EME, Heat, FDTD, etc all into a single Simulation object (and same with their data) basically not distinguishing between the solvers besides the various monitors and sources that are contained within.

caseyflex:
@tylerflex  I added `AbstractYeeGridSimulation` and `AbstractYeeGridSimulationData` as discussed -- it saved a lot of code duplication. I generally tried to move everything related to epsilon, grids, and EM fields into these classes, while keeping things only relevant for time-domain in `Simulation` and things only relevant for EME in `EMESimulation`.

tylerflex:
That's great! So overall it refactored pretty nicely? eg. without a lot of rough edges? If so, that's a good sign for the future. Next I'll implement the mode solver as an `AbsttractYeeGridSimulation`, probably it will reuse a lot of the same internals.

caseyflex:
> That's great! So overall it refactored pretty nicely? eg. without a lot of rough edges? If so, that's a good sign for the future. Next I'll implement the mode solver as an `AbsttractYeeGridSimulation`, probably it will reuse a lot of the same internals.

Yeah, I think it refactored nicely. I might still change a bit before EME pre-release (e.g. consolidating size validators).

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 - 2D heat simulations are now fully supported. 
 
+### Added
+- EME solver through `EMESimulation` class.
```
momchil-flex, line 15: 7000 lines PR, 5 words changelog :D 

(I think that's fine, just funny)

tidy3d/components/data/sim_data.py
```diff
@@ -619,7 +476,11 @@ def plot_field(
 
         # select the extra coordinates out of the data from user-specified kwargs
         for coord_name, coord_val in sel_kwargs.items():
-            if field_data.coords[coord_name].size <= 1:
+            if (
+                field_data.coords[coord_name].size <= 1
+                or coord_name == "eme_port_index"
+                or coord_name == "mode_index"
```
momchil-flex, line 482: I *think* I see why but out of curiosity what is the problem if this goes to `interp` given that the values should always be integer and effectively no interpolation should happen?

tidy3d/components/simulation.py
```diff
@@ -89,117 +90,28 @@
 PML_HEIGHT_FOR_0_DIMS = 0.02
 
 
-class Simulation(AbstractSimulation):
+class AbstractYeeGridSimulation(AbstractSimulation, ABC):
+    """
+    Abstract class for a simulation involving electromagnetic fields defined on a Yee grid.
     """
-    Custom implementation of Maxwell’s equations which represents the physical model to be solved using the FDTD
```
momchil-flex, line 97: This seems like quite a mess in the diff here but in practice it should be pretty small changes I assume? Could you summarize what happened?

tidy3d/components/eme/simulation.py
```diff
@@ -0,0 +1,853 @@
...
+# maximum numbers of simulation parameters
+MAX_GRID_CELLS = 20e9
+WARN_MONITOR_DATA_SIZE_GB = 10
+MAX_MONITOR_INTERNAL_DATA_SIZE_GB = 50
+MAX_SIMULATION_DATA_SIZE_GB = 50
+WARN_MODE_NUM_CELLS = 1e5
+
+
+# eme specific simulation parameters
+MAX_NUM_FREQS = 100
```
momchil-flex: This seems unnecessarily limiting? But I might see your reasoning since the computation can become quite heavy for a large number of freqs.

momchil-flex (approved):

lucas-flexcompute (approved):
I'm guessing the EME sweep helpers will be worked on in another PR, right? In that case, this is looking good to me.

docs/api/index.rst
daquinteroflex, line 1: Docs structures look good so far!

tidy3d/components/data/sim_data.py
```diff
@@ -779,3 +640,162 @@ def plot_scalar_array(
         ax.set_ylim(min(y_coord_values), max(y_coord_values))
 
         return ax
+
+
+class SimulationData(AbstractYeeGridSimulationData):
```
daquinteroflex, line 645: Nice, interesting, I agree. It's kind of creating a separation between the solvers whilst still sharing core aspects to some level. I"m wondering how this works in the context of EME specific extensions? Would you edit AbstractYeeGridSimulationData directly, and do these extensions have to couple in to FDTDSimulationData inherently/will they be supported?

tidy3d/components/eme/data/sim_data.py
```diff
@@ -0,0 +1,302 @@
...
+from .dataset import EMESMatrixDataset
+from ...data.data_array import EMESMatrixDataArray, EMEScalarFieldDataArray
+from ...data.sim_data import AbstractYeeGridSimulationData
+
+from ...types import annotate_type
+from ....exceptions import SetupError
+from ...data.monitor_data import FieldData, ModeData, ModeSolverData
+
+
+class EMESimulationData(AbstractYeeGridSimulationData):
```
daquinteroflex, line 20: This just answered my previous question haha

CHANGELOG.md
caseyflex: :D

tidy3d/components/eme/grid.py
```diff
@@ -0,0 +1,543 @@
...
+        axis = self.axis
+        size[axis] = 0
+        boundary_planes = []
+        for cell_boundary in self.boundaries:
+            center[axis] = cell_boundary
+            boundary_planes.append(Box(center=center, size=size))
+        return boundary_planes
+
+    @property
+    def cells(self) -> List[Box]:
```
daquinteroflex, line 510: Maybe naive question. This is interesting, just out of curiousity, is this how it gets constructed in the backend too in the sense that if a user makes  a custom grid fucntion based on box cells it would also work?

tidy3d/components/eme/simulation.py
daquinteroflex: Hmm this is also pretty interesting in the context of grid translations between other solvers, I wonder to what level do we want other solvers to have this functionality too

tidy3d/components/simulation.py
daquinteroflex: Hmm from what I can tell it's just organising AbstractYeeGridSimulation and Simulation in the same file. I'm wondering (although maybe a bit painful) if it makes sense to move this to the `base_sim/abstract_yee_grid` maybe thingking that at some point 3.0 it might also make sense to rename our current simulation class `FDTDSimulation` although we can also move this later too! Maybe @tylerflex has some thoughts too

tidy3d/components/data/sim_data.py
caseyflex:
```
da = xr.DataArray(
    data=[[0, np.nan, np.nan], [1, np.nan, np.nan]],
    dims=("x", "y"),
    coords={"x": [0, 1], "y": [0, 1, 2]},
)
print(da.interp(y=0))
print(da.sel(y=0))
```

```
<xarray.DataArray (x: 2)> Size: 16B
array([nan, nan])
Coordinates:
  * x        (x) int64 16B 0 1
    y        int64 8B 0
<xarray.DataArray (x: 2)> Size: 16B
array([0., 1.])
Coordinates:
  * x        (x) int64 16B 0 1
    y        int64 8B 0
```

Seems to be a weird issue with xarray multidimensional interpolation when nans are present?

tidy3d/web/api/asynchronous.py
```diff
@@ -16,14 +16,14 @@ def run_async(
     simulation_type: str = "tidy3d",
     parent_tasks: Dict[str, List[str]] = None,
 ) -> BatchData:
-    """Submits a set of Union[:class:`.Simulation`, :class:`.HeatSimulation`] objects to server,
+    """Submits a set of Union[:class:`.Simulation`, :class:`.HeatSimulation`, :class:`.EMESimulation`] objects to server,
```
daquinteroflex, line 19: This is maybe something we should add later, but I believe we would benefit from having a `SolverSimulationType` definition, also in the context of 3.0, so that all these methods can just use that type both at the docstring and parameter typing check level @tylerflex

daquinteroflex (approved):
11/10 job! Congratulations, so much effort here, and it's really good!

I made a few tiny comments that maybe we can also address in the 3.0 refactor really, but it's great

tidy3d/components/simulation.py
caseyflex: I moved all of the fields and methods that weren't FDTD-specific but involve EM fields on Yee grid to `AbstractYeeGridSimulation`. I left duplicates of some methods and fields in `Simulation` when their docstrings had information specific to FDTD simulations.

tidy3d/components/eme/simulation.py
caseyflex: yeah, when we use a lot of frequencies we can run into performance issues and memory issues in the mode solver I believe

tidy3d/components/eme/grid.py
caseyflex: It's not far off. In principle the EME cells are just a list of cells, yes, so in that sense it can be decoupled from the actual simulation geometry a bit. This could be useful for periodic repetition of cells or other reuse of mode solving in the future

tidy3d/components/eme/simulation.py
caseyflex: btw, now the grid generation is handled in AbstractYeeGridSimulation instead of via an intermediate fdtd simulation.

tidy3d/components/simulation.py
daquinteroflex: (Not related to this review specifically)

You know I was thinking about this fields/methods code duplicates issue recently. I believe we won't be able to escape it fundamentally in an inheritance-based data scheme (which is good and bad for different reasons), but the alternative which is strongly-typed methods that operate on data container classes have their own set of problems too I guess more on the "user-friendly linking methods to data classes" aspect. This might just be fundamental to having multiple classes have "slightly different variations of the same method" be it in docstrings or implementation. Problems can become more prevalent if we have classmethods using otherclassmethods in a breaking manner that I think we generally try to avoid by writing pure class functions. It is very strongly coupled to the way we choose to represent our data really and have users interact with it, and any approach involves considerable work. There might be a conceptual point of having to draw the line of how exactly we want to interact with our data, and in a way maybe these quirks could be part of it. The real question is the effect this has on scalability and integration with other methods, and how much this affects our development understanding and maintenance of the code. These are not easy questions but I think @tylerflex was thinking about this too in his concept

momchil-flex: > Hmm from what I can tell it's just organising AbstractYeeGridSimulation and Simulation in the same file. I'm wondering (although maybe a bit painful) if it makes sense to move this to the `base_sim/abstract_yee_grid` maybe thingking that at some point 3.0 it might also make sense to rename our current simulation class `FDTDSimulation` although we can also move this later too! Maybe @tylerflex has some thoughts too

I think this makes sense, but it can be made in a separate PR. We could also go more closely over whether some of the code duplication may be avoided. I'm also down to rename it to `FDTDSimulation`. We may still do something like both `import FDTDSimulation` and `import FDTDSimulation as Simulation` in tidy3d init if we want to keep some backwards compatibility.

tidy3d/components/data/sim_data.py
momchil-flex: Oh ok so I didn't actually see. :) Makes sense.

tidy3d/components/eme/simulation.py
```diff
@@ -0,0 +1,859 @@
...
+        freqs = self.freqs
+        grid_spec = self.grid_spec
+        sources = []
+        if grid_spec.auto_grid_used and grid_spec.wavelength is None:
+            if not np.all(np.isclose(freqs, freqs[0])):
+                raise SetupError(
+                    "Multiple 'sim.freqs' are supplied. Please supply "
+                    "a 'wavelength' value for 'grid_spec' to control automatic "
+                    "grid generation."
+                )
```
dbochkov-flexcompute, line 869: Can you define an EMESimulation and submit it without ever invoking `.grid`? I'm just thinking whether it could happen that the user provides a faulty grid spec setup, but can still send it to server for solving, where it will just error. If that can happen, then it would probably makes sense to implement this grid check as a validator.

tidy3d/components/eme/simulation.py
```diff
@@ -0,0 +1,859 @@
...
+    sweep_spec: Optional[EMESweepSpecType] = pd.Field(
+        None,
+        title="EME Sweep Specification",
+        description="Specification for a parameter sweep to be performed during the EME "
+        "propagation step. The results are stored "
+        "in 'sim_data.smatrix'. Other simulation monitor data is not included in the sweep.",
+    )
+
+    constraint: Literal["passive", "unitary", "none"] = pd.Field(
+        "none",
```
momchil-flex: Usually, we wouldn't use `"none"` but rather just `None`, i.e. make the constraint optional. If there's no good reason to keep the string I would suggest refactoring in that way.

tidy3d/components/eme/simulation.py
caseyflex: added to `sim._post_init_validators`, thanks

tidy3d/components/eme/simulation.py
caseyflex: done

PR #1455, opened 2024-02-08
adds multiprocessing to design plugin

tylerflex:
will replace #1423 

Adds a way to run design problems with an optional `map_fn` argument to `Design.run`, which maps the user defined function to the `input_kwargs`.

By default, `Design.run()` uses builtin `map`, which runs sequentially.

Also adds an implementation of this using `multiprocessing.Pool.map`, which can be called as `Design.run_multiprocess`.

```py
# original (sequential, using built-in map
sweep_results = design_space.run(scs)

# multiprocessing (built in)
sweep_results = design_space.run_multiprocess(scs, num_processes=2)

# multiprocessing (user supplied map function)
from multiprocessing import Pool

with Pool() as p:
    sweep_results = design_space.run(scs, map_fn=p.map)

```

NOTE: before merging, I need to make sure I adjust the submodule commit. I'm running something now with a different state but will remember to reset it later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Users can toggle https ssl version through `from tidy3d.web.core.environment import Env` and `Env.set_ssl_version(ssl_version: ssl.TLSVersion)`
 - Free-carrier absorption (FCA) and free-carrier plasma dispersion (FCPD) nonlinearities inside `TwoPhotonAbsorption` class.
 - `log_path` argument in `set_logging_file`, set to `False` by default.
-- `design` plugin includes `Design.run_custom_map` and `Design.run_multiprocess` to enable running single measurement functions in a custom or concurrent fashion.
+- `design` plugin includes `Design.run_multiprocess` to enable running single measurement functions in a custom or concurrent fashion. Alternatively, `Design.run` accepts an optional `map_fn` kwarg to allow user-defined `map` functions.
 
 ### Changed
 - `DataArray.to_hdf5()` accepts both file handles and file paths.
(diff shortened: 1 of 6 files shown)


dbochkov-flexcompute:
Looks like a much more lightweight alternative to using asyncio. My initial concern though (actually, maybe it is applicable to asyncio version too): does running multiple `web.run()` in parallel not cause overwriting of simulation results? Is there something that tells web api to write data from different runs into different files?

tylerflex:
> does running multiple web.run() in parallel not cause overwriting of simulation results? Is there something that tells web api to write data from different runs into different files?

Yea this is a valid concern, I've been thinking about it too. So there's a period of time between when the results are downloaded and then loaded into `SimulationData` where things could (in principle) go wrong. It's possible also that if the there could be issues if the download takes a while and we try downloading two files at the same time.

I feel like a couple possible solutions are:
1. Be really clear to users that if they go this route, they specify a unique filename for their data (`path`). But even this might not be sufficient if the data is always downloaded as `monitor_data.hdf5` or whatever it is..
2. Change the default data filename (on the server too) to include the task_id.

Wonder if @momchil-flex has any thoughts?

tylerflex:
or @dbochkov-flexcompute if you can think of a way to mitigate this?

tylerflex:
Also, worst case scenario, we can get rid of the builtin `Design.run_multiprocess` (and the `multiprocess` dependency) and just include a section in the tutorial notebook showing how to do this with the `map_fn` kwarg AND adding a note to be sure to use different data filenames.

dbochkov-flexcompute:
First idea that comes to mind is to change working directory for each thread, so that every web.run() can write using the same file names, just in different directories. Maybe with `os.chdir`? However, I haven't tried that myself, so not sure if this fully feasible

lucas-flexcompute:
Would it be possible to check the data path for each task in the pool and issue a warning/error if any two map to the same place? Somehow I don't think changing directory is a good idea… could there be OS-specific issues with this related to permissions or some other particularity we're not foreseeing?

tylerflex:
> Would it be possible to check the data path for each task in the pool and issue a warning/error if any two map to the same place?

Would be nice, but I dont have direct access to the `web.run()` calls from the user supplied function, so I'm not sure how to check this.

dbochkov-flexcompute:
another hacky option: add an option to web api not to overwrite file if already exists and use a random name instead. Could be controlled by a global variable which is set and unset before and after `run`/`run_multiprocess`

tylerflex:
closing until we decide to take this up again

PR #1457, opened 2024-02-09
Updates to help unify backend logging

momchil-flex:


daquinteroflex:
Trying to fix the action situation fyi

lucas-flexcompute (approved):
I understand the changes required because of the backend log, so this looks fine. But thinking about the frontend only, is there any reason to use "USER" level instead of "INFO"? Or "SUPPORT" instead of "DEBUG"? I wonder if we should add a general comment for the future instructing frontend devs to only issue log messages using the standard levels, because that's what users would normally expect.

daquinteroflex:
Many apologies for the notifications, I thought it was going to be an easy fix.

 I'm going to branch out and test in another PR, as I did try to reproduce this error in another personal PR and didn't get it. Will just clean up this PR and drop Momchil's specific commit on the other one that needs to be merged into pre/2.6

daquinteroflex:
Oh my god! I'm so sorry! I don't know what I did

PR #1458, opened 2024-02-09
Add missing Transformed geometry class to docs

lucas-flexcompute:


daquinteroflex (approved):
Sounds good! 

Happy to build the docs to visualize how it all looks like if you want a hand checking that. Maybe you might be interested to see how the notebooks and the API are coreferenced like for your example through the "See Also" docstring links eg. https://github.com/flexcompute/tidy3d/blob/0251fedcd5c262404f0604b4c71606f42ebd04bb/tidy3d/components/simulation.py#L314-L330 ?) You can build how it looks like with the  `tidy3d develop build-docs` if interested too!

tylerflex (approved):

PR #1463, opened 2024-02-12
Improved efficiency of outer_dot

caseyflex:
Older version of this new PR: https://github.com/flexcompute/tidy3d/pull/1464
This older PR reduces overhead and vectorizes. Vectorizing increases memory usage though, which may be unacceptable. The new PR reduces overhead but does not vectorize.

This seems to greatly improve the efficiency of `outer_dot`. 

The efficiency of mode sorting could possibly be improved as well by switching from `dot` to `outer_dot`, but that is not yet included in this PR.

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -677,61 +677,31 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
         h_1 = "H" + dim1
         h_2 = "H" + dim2
 
-        # Prepare array with proper dimensions for the dot product data
-        arrays = (fields_self[e_1], fields_other[e_1])
-        coords = (arrays[0].coords, arrays[1].coords)
+        keys = (e_1, e_2, h_1, h_2)
 
-        # Common frequencies to both data arrays
-        f = np.array(sorted(set(coords[0]["f"].values).intersection(coords[1]["f"].values)))
+        def get_field_data(fields: dict, key: str, is_self: bool):
+            """Grab a field from a dictionary of field components, renaming `mode_index` coord if present."""
+            arr = fields[key]
+            if "mode_index" in arr.coords:
+                new_key = "mode_index_0" if is_self else "mode_index_1"
+                arr = arr.rename(mode_index=new_key)
+            return arr
 
-        # Mode indices, if available
-        modes_in_self = "mode_index" in coords[0]
-        mode_index_0 = coords[0]["mode_index"].values if modes_in_self else np.zeros(1, dtype=int)
-        modes_in_other = "mode_index" in coords[1]
-        mode_index_1 = coords[1]["mode_index"].values if modes_in_other else np.zeros(1, dtype=int)
+        e_self_1, e_self_2, h_self_1, h_self_2 = (
+            get_field_data(fields_self, key, is_self=True) for key in keys
+        )
+        e_other_1, e_other_2, h_other_1, h_other_2 = (
+            get_field_data(fields_other, key, is_self=False) for key in keys
+        )
 
-        dtype = np.promote_types(arrays[0].dtype, arrays[1].dtype)
-        dot = np.empty((f.size, mode_index_0.size, mode_index_1.size), dtype=dtype)
+        # Cross products of fields
+        e_self_x_h_other = e_self_1 * h_other_2 - e_self_2 * h_other_1
+        h_self_x_e_other = h_self_1 * e_other_2 - h_self_2 * e_other_1
 
-        # Calculate overlap for each common frequency and each mode pair
-        for i, freq in enumerate(f):
-            indexer_self = {"f": freq}
-            indexer_other = {"f": freq}
-            for mi0 in mode_index_0:
-                if modes_in_self:
-                    indexer_self["mode_index"] = mi0
-                e_self_1 = fields_self[e_1].sel(indexer_self, drop=True)
-                e_self_2 = fields_self[e_2].sel(indexer_self, drop=True)
-                h_self_1 = fields_self[h_1].sel(indexer_self, drop=True)
-                h_self_2 = fields_self[h_2].sel(indexer_self, drop=True)
-
-                for mi1 in mode_index_1:
-                    if modes_in_other:
-                        indexer_other["mode_index"] = mi1
-                    e_other_1 = fields_other[e_1].sel(indexer_other, drop=True)
-                    e_other_2 = fields_other[e_2].sel(indexer_other, drop=True)
-                    h_other_1 = fields_other[h_1].sel(indexer_other, drop=True)
-                    h_other_2 = fields_other[h_2].sel(indexer_other, drop=True)
-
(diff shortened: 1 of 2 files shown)


tidy3d/components/data/monitor_data.py
```diff
@@ -677,61 +677,33 @@ def outer_dot(
...
-        result = xr.DataArray(dot, coords=coords)
-
-        # Remove mode index coordinate if the input did not have it
-        if not modes_in_self:
-            result = result.isel(mode_index_0=0, drop=True)
-        if not modes_in_other:
-            result = result.isel(mode_index_1=0, drop=True)
-
-        return result
+        e_self_1 = fields_self[e_1]
```
tylerflex, line 680: looks good overall, might be a bit more compact to reframe the field unpacking like
```py
keys = (e_1, e_2, h_1, h_2)

def get_field_data(fields: dict, key: str, is_self: bool):
    """Grab a field from a dictionary of field components, renaming `mode_index` coord if present."""
    arr = fields[key]
    if "mode_index" in arr.coords:
        new_key = "mode_index_0" if is_self else "mode_index_1"
        arr = arr.rename(mode_index=new_key)
    return arr

e_self_1, e_self_2, h_self_1, h_self_2 = [get_field_data(fields_self, key, is_self=True) for key in keys)
e_other_1, e_other_2, h_other_1, h_other_2 = [get_field_data(fields_other, key, is_self=False) for key in keys)
```

momchil-flex (commented):
It's nice that this is faster but it does not seem like it is equivalent to the previous handling? Things that appear to be missing are the check for common frequencies (the outer dot only at the set of common freqs is computed), and dropping out the mode index dimension if missing (this would be the case when outer dot is invoked between mode data and field data or two field data objects, which is also allowed).

I think it should just be a matter of adding some xarray handling though while preserving the efficiency gain which effectively should be coming from the fact that you call vectorize over everything, as opposed to collecting the product in a loop.

Unless I'm missing something, this should have been caught by tests. Since it wasn't, I'd ask you to add some as part of this PR - specifically that outer dot works with mode data and field data, and with objects that share only some frequencies. I would *think* both of those cases would error with your current changes. See if there are already some outer dot tests that can be expanded.

I am not sure about speeding up the overlap sort, although you could certainly try! The way it is currently set up actually is that it does not compute the full outer product because we thought this would be too expensive. Instead it adds some more complicated iterations to only compute some limited products if it finds that a mode is out of place. If you ever get to try you should specifically test what happens with 50, or 100 modes, in which case even the vectorized outer dot may become expensive. I'm trying to think if there are also some memory considerations... Do you have an idea of what type of memory overhead one gets for outer dot of some mode data with itself, as a factor of the memory occupied by the memory data on its own?

caseyflex:
> It's nice that this is faster but it does not seem like it is equivalent to the previous handling? Things that appear to be missing are the check for common frequencies (the outer dot only at the set of common freqs is computed), and dropping out the mode index dimension if missing (this would be the case when outer dot is invoked between mode data and field data or two field data objects, which is also allowed).
> 
> I think it should just be a matter of adding some xarray handling though while preserving the efficiency gain which effectively should be coming from the fact that you call vectorize over everything, as opposed to collecting the product in a loop.
> 
> Unless I'm missing something, this should have been caught by tests. Since it wasn't, I'd ask you to add some as part of this PR - specifically that outer dot works with mode data and field data, and with objects that share only some frequencies. I would _think_ both of those cases would error with your current changes. See if there are already some outer dot tests that can be expanded.
> 
> I am not sure about speeding up the overlap sort, although you could certainly try! The way it is currently set up actually is that it does not compute the full outer product because we thought this would be too expensive. Instead it adds some more complicated iterations to only compute some limited products if it finds that a mode is out of place. If you ever get to try you should specifically test what happens with 50, or 100 modes, in which case even the vectorized outer dot may become expensive. I'm trying to think if there are also some memory considerations... Do you have an idea of what type of memory overhead one gets for outer dot of some mode data with itself, as a factor of the memory occupied by the memory data on its own?

Actually, both the check for common frequencies and dropping out the mode index dimension are already handled correctly. I added explicit tests for both. Using only common frequencies is the default xarray binary op behavior. Regarding mode_index -- I actually don't introduce it in the first place unless it is already there, and then I just rename it appropriately.

Regarding overlap sort -- I've tried the vectorized dot with 100 modes, and in the old version it takes over a minute, and the new version takes less than a second. So I think the vectorized dot speed would not be an issue. Regarding memory overhead -- it does seem like the new version uses more memory. If there are N modes in each dataset, then each is broadcast to N x N in order to compute the cross products. I don't see a way around this, it seems like a tradeoff between time and memory. We could add a parameter to the function to choose whether to use the slow version (default) or the fast version?

caseyflex:
> > It's nice that this is faster but it does not seem like it is equivalent to the previous handling? Things that appear to be missing are the check for common frequencies (the outer dot only at the set of common freqs is computed), and dropping out the mode index dimension if missing (this would be the case when outer dot is invoked between mode data and field data or two field data objects, which is also allowed).
> > I think it should just be a matter of adding some xarray handling though while preserving the efficiency gain which effectively should be coming from the fact that you call vectorize over everything, as opposed to collecting the product in a loop.
> > Unless I'm missing something, this should have been caught by tests. Since it wasn't, I'd ask you to add some as part of this PR - specifically that outer dot works with mode data and field data, and with objects that share only some frequencies. I would _think_ both of those cases would error with your current changes. See if there are already some outer dot tests that can be expanded.
> > I am not sure about speeding up the overlap sort, although you could certainly try! The way it is currently set up actually is that it does not compute the full outer product because we thought this would be too expensive. Instead it adds some more complicated iterations to only compute some limited products if it finds that a mode is out of place. If you ever get to try you should specifically test what happens with 50, or 100 modes, in which case even the vectorized outer dot may become expensive. I'm trying to think if there are also some memory considerations... Do you have an idea of what type of memory overhead one gets for outer dot of some mode data with itself, as a factor of the memory occupied by the memory data on its own?
> 
> Actually, both the check for common frequencies and dropping out the mode index dimension are already handled correctly. I added explicit tests for both. Using only common frequencies is the default xarray binary op behavior. Regarding mode_index -- I actually don't introduce it in the first place unless it is already there, and then I just rename it appropriately.
> 
> Regarding overlap sort -- I've tried the vectorized dot with 100 modes, and in the old version it takes over a minute, and the new version takes less than a second. So I think the vectorized dot speed would not be an issue. Regarding memory overhead -- it does seem like the new version uses more memory. If there are N modes in each dataset, then each is broadcast to N x N in order to compute the cross products. I don't see a way around this, it seems like a tradeoff between time and memory. We could add a parameter to the function to choose whether to use the slow version (default) or the fast version?

Actually, let me correct this. From profiling, it seems that the main speedup here is not from vectorization but from minimizing overhead from xarray calls. I made an alternative PR in light of this, that still removes this overhead but is no longer vectorized. This should alleviate the memory concern. https://github.com/flexcompute/tidy3d/pull/1464

momchil-flex:
Closing in favor of #1464

PR #1465, opened 2024-02-13
Improved efficiency of overlap_sort

caseyflex:
Uses the faster outer_dot to improve efficiency of overlap_sort. Sometimes, sorting modes can be a bottleneck in mode solving, and this should prevent that from occuring.

Based off of https://github.com/flexcompute/tidy3d/pull/1464

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -1146,6 +1146,8 @@ class ModeData(ModeSolverDataset, ElectromagneticFieldData):
             for freq_id in range(f0_ind + step, last_ind, step):
                 # Get next frequency to sort
                 data_to_sort = self._isel(f=[freq_id])
+                # Assign to the base frequency so that outer_dot will compare them
+                data_to_sort = data_to_sort._assign_coords(f=[self.monitor.freqs[f0_ind]])
 
                 # Compute "sorting w.r.t. to neighbor" and overlap values
 
@@ -1191,6 +1193,18 @@ class ModeData(ModeSolverDataset, ElectromagneticFieldData):
         update_dict = {key: field.isel(**isel_kwargs) for key, field in update_dict.items()}
         return self._updated(update=update_dict)
 
+    def _assign_coords(self, **assign_coords_kwargs):
+        """Wraps ``xarray.DataArray.assign_coords`` for all data fields that are defined over frequency and
+        mode index. Used in ``overlap_sort`` but not officially supported since for example
+        ``self.monitor.mode_spec`` and ``self.monitor.freqs`` will no longer be matching the
+        newly created data."""
+
+        update_dict = dict(self._grid_correction_dict, **self.field_components)
+        update_dict = {
+            key: field.assign_coords(**assign_coords_kwargs) for key, field in update_dict.items()
+        }
+        return self._updated(update=update_dict)
+
     def _find_ordering_one_freq(
         self,
         data_to_sort: ModeData,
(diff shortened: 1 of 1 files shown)


caseyflex:
> @caseyflex should we close this (or make a draft) in view of #1464 ?

no sorry -- we should close this one: https://github.com/flexcompute/tidy3d/pull/1463

This PR uses the revised outer_dot PR https://github.com/flexcompute/tidy3d/pull/1464 to improve overlap_sort.

tylerflex:
ok I see, sorry for the confusion

tidy3d/components/data/monitor_data.py
```diff
@@ -1126,6 +1127,18 @@ def _isel(self, **isel_kwargs):
         update_dict = {key: field.isel(**isel_kwargs) for key, field in update_dict.items()}
         return self._updated(update=update_dict)
 
+    def _assign_coords(self, **isel_kwargs):
+        """Wraps ``xarray.DataArray.assign_coords`` for all data fields that are defined over frequency and
+        mode index. Used in ``overlap_sort`` but not officially supported since for example
+        ``self.monitor.mode_spec`` and ``self.monitor.freqs`` will no longer be matching the
+        newly created data."""
+
+        update_dict = dict(self._grid_correction_dict, **self.field_components)
```
momchil-flex, line 1136: _grid_correction_dict is a property so it shouldn't be added here at all?

tidy3d/components/data/monitor_data.py
```diff
@@ -1081,6 +1081,7 @@ def overlap_sort(
             for freq_id in range(f0_ind + step, last_ind, step):
                 # Get next frequency to sort
                 data_to_sort = self._isel(f=[freq_id])
+                data_to_sort = data_to_sort._assign_coords(f=[self.monitor.freqs[f0_ind]])
```
momchil-flex, line 1084: I am confused, why is this needed? For a regular DataArray, isel-ing already redefines the coordinates, too. Is something not working right in our `_isel` wrapper?

image

tidy3d/components/data/monitor_data.py
```diff
@@ -1147,19 +1160,20 @@ def _find_ordering_one_freq(
...
             # Get one mode from data_to_sort
 
-            one_mode = data_to_sort._isel(mode_index=[mode_index])
+            # one_mode = data_to_sort._isel(mode_index=[mode_index])
 
             # Project to all modes of interest from data_template
-            amps_reduced[:, i] = data_template_reduced.dot(one_mode).data.ravel()
+            # amps_reduced[:, i] = data_template_reduced.dot(one_mode).data.ravel()
             if self.monitor.store_fields_direction == "-":
                 amps_reduced[:, i] *= -1
```
momchil-flex, line 1178: Seems like the whole loop can be removed then since the `-1` factor can be applied to all of `amps_reduced`.

tidy3d/components/data/monitor_data.py
caseyflex: I guess this is adding `grid_primal_correction` and `grid_dual_correction` to the update list. I copied this from `_isel`, but it is probably not needed as these fields keep their old values...

caseyflex: oh actually, it does do something -- it updates the coordinates on those two fields as well as the field_components. This may actually be needed so best to leave it.

tidy3d/components/data/monitor_data.py
caseyflex: It's a bit of a hack -- I want to isel `f=[freq_id]`, which is the next frequency to sort. But, I want to compute outer_dot against the base frequency `f0`. But because of how outer_dot works, it only keeps frequencies which are common between the two datasets. So, here, I isel `f=[freq_id]`, then I assign coords to `f0`, so that outer_dot will think they are at the same frequency and know to compare them.

tidy3d/components/data/monitor_data.py
momchil-flex: Ah, got it.

tidy3d/components/data/monitor_data.py
momchil-flex: Ohhhhh. Funny. One thing we could do also do is to allow `outer_dot` to work without a frequency dimension, and just pop it? But, eh, I think this is fine.

momchil-flex (approved):

momchil-flex:
I don't know if @weiliangjin2021 still wants to look at this, but it should be good to merge after a rebase. The conflicts are probably in the changelog. You could squash the commits too.

caseyflex:
> I don't know if @weiliangjin2021 still wants to look at this, but it should be good to merge after a rebase. The conflicts are probably in the changelog. You could squash the commits too.

Rebased and squashed. Merge when you're ready. thanks!

weiliangjin2021 (approved):
I didn't take a very careful look. From what I understand, it all look good.

PR #1467, opened 2024-02-13
adjoint util erosion-dilation penalty

tylerflex:
notebook PR: https://github.com/flexcompute/tidy3d-notebooks/pull/46

### DIFF
--- a/tidy3d/plugins/adjoint/utils/penalty.py
+++ b/tidy3d/plugins/adjoint/utils/penalty.py
@@ -8,6 +8,8 @@ from ....components.base import Tidy3dBaseModel
 from ....components.types import ArrayFloat2D
 from ....constants import MICROMETER
 
+from .filter import ConicFilter, BinaryProjector
+
 # Radius of Curvature Calculation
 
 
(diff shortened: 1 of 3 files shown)


tidy3d/plugins/adjoint/utils/penalty.py
```diff
@@ -134,3 +136,119 @@ def penalty_fn(radius):
...
+    for Very Large-Scale Integrated-Photonics Inverse Design" (2022).
+    """
+
+    length_scale: pd.NonNegativeFloat = pd.Field(
+        ...,
+        title="Length Scale",
+        description="Length scale of erosion and dilation."
+        "Corresponds to ``radius`` in the :class:`ConicFilter` used for filtering."
+        "The parameter array is dilated and eroded by half of this value with each operation."
+        "Roughly corresponds to the desired minimum feature size and radius of curvature",
```
momchil-flex, line 162: Missing period

tidy3d/plugins/adjoint/utils/penalty.py
```diff
@@ -134,3 +136,119 @@ def penalty_fn(radius):
...
+        description="Strength of the ``tanh`` projection."
+        "Corresponds to ``beta`` in the :class:`BinaryProjector`."
+        "Higher values correspond to stronger discretization.",
+    )
+
+    penalty_strength = pd.Field(
+        1.0,
+        title="Penalty Strength",
+        description="Optional value to scale the returned penalty by.",
+    )
```
momchil-flex, line 186: I don't think we need a `penalty_strength` Field? It can be trivially applied after the penalty evaluation.

tidy3d/plugins/adjoint/utils/penalty.py
```diff
@@ -134,3 +136,119 @@ def penalty_fn(radius):
...
+        "Corresponds to ``eta`` in the :class:`BinaryProjector`.",
+    )
+
+    delta_eta: pd.PositiveFloat = pd.Field(
+        0.01,
+        title="Delta Eta Cutoff",
+        description="The binarization threshold for erosion and dilation operations "
+        "The thresholds are 0 + `delta_eta` on the low end and `1 - delta_eta` on the high end. "
+        "This value must be tuned precisely to balance binarization with differentiability. "
+        "We strongly suggest leaving to the default value.",
```
momchil-flex, line 202: "leaving to" -> "using"

If it must be tuned precisely, do we know if this default is good for a broad range of settings?

tidy3d/plugins/adjoint/utils/penalty.py
tylerflex: yea I agree, wasn't sure so decided to put it in originally but will remove.

tidy3d/plugins/adjoint/utils/penalty.py
tylerflex: I haven't tested a lot of values, so I don't know. Maybe I'll remove this "tuned precisely" comment and instead just say that "this value works well".

Will do a fabrication constraint notebook where I'll go into more details on each of these penalties so perhaps I'll discuss it there a bit more. As we talked about before, if this value is too high (eg. 0.5) then the erosion/dilation effect is reduced or eliminated. But if it's too low (eg. 0.0) then it won't handle intermediate values the way we want.

tylerflex: Changed the last two sentences to "The default value balances binarization with differentiability so we strongly suggest using it unless there is a good reason to set it differently."

tylerflex:
resolved the docstring comments and fixed submodule

tylerflex:
I also set the default `beta = 100` for the penalty since that's what it was set in all of the notebooks. Note this is different from the β value that sets the ε. This β value just controls the projection of the erosion / dilation, which we generally want to be high enough to lead to basically discretized values anyway.

momchil-flex (approved):
Yeah this was clear to me after having worked with that penalty a bit already; hopefully will not be confusing to users.

twhughes:
In my notebook PR I removed `beta=100` from the `ErosionDilationPenalty` initialization to avoid confusion.

PR #1470, opened 2024-02-14
Updates to help unify backend logging (Reset back to original #1457)

daquinteroflex:
I am extremely sorry @momchil-flex ! I accidentally deleted your branch whilst working on top of it to fix the Github actions issue. I will not do something like that again, and have learnt my lesson.

The old relevant PR is https://github.com/flexcompute/tidy3d/pull/1457 and I have just reset back to your commit as you can see. I will sort out the github actions issue in another branch. I am extremely sorry as I thought it was just a minor script update, but was not.

Relevant comments from Lucas were:
> I understand the changes required because of the backend log, so this looks fine. But thinking about the frontend only, is there any reason to use "USER" level instead of "INFO"? Or "SUPPORT" instead of "DEBUG"? I wonder if we should add a general comment for the future instructing frontend devs to only issue log messages using the standard levels, because that's what users would normally expect.

What I should have done was to branch out and make another PR to fix the github action separately - which is what I've done now. The weird thing is that we can see I haven't changed anything, and yet the github action is working fine. But I think I found out why the error was happening which is related to the VENV caching and am working to fix it from happening again. This is why your commit doesn't error in the new PR I've made (and you can see it's the same commit state). Bug fixed in https://github.com/flexcompute/tidy3d/pull/1469

### DIFF
--- a/tidy3d/log.py
+++ b/tidy3d/log.py
@@ -3,18 +3,21 @@
 import inspect
 from datetime import datetime
 
-from typing import Union, List
+from typing import Union, List, Callable
 from typing_extensions import Literal
 
 from rich.console import Console
 
-
-LogLevel = Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
+# Note: "SUPPORT" and "USER" levels are meant for backend runs only.
+# Logging in frontend code should just use the standard debug/info/warning/error/critical.
+LogLevel = Literal["DEBUG", "SUPPORT", "USER", "INFO", "WARNING", "ERROR", "CRITICAL"]
 LogValue = Union[int, LogLevel]
 
 # Logging levels compatible with logging module
 _level_value = {
     "DEBUG": 10,
+    "SUPPORT": 12,
+    "USER": 15,
     "INFO": 20,
     "WARNING": 30,
     "ERROR": 40,
@@ -27,6 +30,8 @@ DEFAULT_LEVEL = "WARNING"
 
 DEFAULT_LOG_STYLES = {
     "DEBUG": None,
+    "SUPPORT": None,
+    "USER": None,
     "INFO": None,
     "WARNING": "red",
     "ERROR": "red bold",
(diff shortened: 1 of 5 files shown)


tidy3d/log.py
```diff
@@ -374,12 +398,16 @@ def set_logging_file(
...
     if "file" in log.handlers:
         try:
-            log.handlers["file"].file.close()
+            log.handlers["file"].console.file.close()
         except Exception:  # TODO: catch specific exception
             log.warning("Log file could not be closed")
         finally:
             del log.handlers["file"]
 
+    if fname == "":
```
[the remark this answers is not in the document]
momchil-flex: That should work too, will change.

momchil-flex:
> I understand the changes required because of the backend log, so this looks fine. But thinking about the frontend only, is there any reason to use "USER" level instead of "INFO"? Or "SUPPORT" instead of "DEBUG"? I wonder if we should add a general comment for the future instructing frontend devs to only issue log messages using the standard levels, because that's what users would normally expect.

@lucas-flexcompute that makes sense, will add a comment around the levels definitions.

tidy3d/log.py
```diff
@@ -27,6 +29,8 @@
 
 DEFAULT_LOG_STYLES = {
     "DEBUG": None,
+    "SUPPORT": None,
```
tylerflex, line 32: is the user able to set these logging levels? Would there be a use case for that? I'm wondering if it makes sense to hide them somehow to avoid possible confusion, or if we should just document them along with the other log levels

momchil-flex: They can set them in their local session, but as per Lucas' comment, there wouldn't really be any frontend messages printed at support or user level anyway.

tylerflex (approved):

tidy3d/log.py
tylerflex: I see. I wonder if there's some way to document these two new levels, maybe as a short note in the config.logging_level description?

momchil-flex: So, yeah, I'm not sure what's best. I just pushed a commit with some more comments about the two new levels, what do you think?

momchil-flex:
Now `test_develop_cli` is failing @daquinteroflex

tidy3d/log.py
tylerflex: Looks good

daquinteroflex:
> Now `test_develop_cli` is failing @daquinteroflex

Yeah sorry just a rebase to the latest pre/2.6 I think sorted it!

PR #1472, opened 2024-02-15
PATCH: Poetry dev github actions :pill: 

daquinteroflex:
Fixes the test-develop-cli action that has been failing for most people


daquinteroflex:
@tylerflex this will fix your failing actions and because it's a patch I'll just rebase in

Issue #1403, opened 2024-01-26
Set adjoint `run_time` to forward source `fwidth` (if same `freq0`?)


tylerflex:
the current code for setting (single frequency) adjoint fwidth and run_time is below (paraphrased for simplicity)

```py
    @cached_property
    def _fwidth_adjoint(self) -> float:
        """Frequency width to use for adjoint source, user-defined or the average of the sources."""

        # if user-specified, use that
        if self.fwidth_adjoint is not None:
            return self.fwidth_adjoint

        freqs_adjoint = self.freqs_adjoint

        # multiple output frequency case
        if self._is_multi_freq:
            return FWIDTH_FACTOR_MULTIFREQ * self._min_delta_freq

        # otherwise, grab from sources and output monitors
        num_sources = len(self.sources)  # should be 0 for adjoint already but worth checking

        # if no sources, just use a constant factor times the mean adjoint frequency
        if num_sources == 0:
            return FWIDTH_FACTOR * np.mean(freqs_adjoint)

        # if more than one forward source, use their maximum
        if num_sources > 1:
            log.warning(f"{num_sources} sources, using their maximum 'fwidth' for adjoint source.")

        fwidths = [src.source_time.fwidth for src in self.sources]
        return np.max(fwidths)

```

```py
    @cached_property
    def _run_time_adjoint(self: float) -> float:
        """Return the run time of the adjoint simulation as a function of its fwidth."""

        if self.run_time_adjoint is not None:
            return self.run_time_adjoint

        return 100 / self._fwidth_adjoint
```

A problem is if the user's forward run_time is short for some reason (like 10/fwidth), then the adjoint one can be set to 100/fwidth, which was not ideal for a user who was running low on credits.

Perhaps to handle this we should add a new condition in the `_run_time_adjoint`

```py
# if both forward and adjoint are single frequency (with the same frequency) use forward simulation run_time.
if len(self.sources) == len(self.freqs_adjoint) == 1:
    if np.isclose(self.sources[0].source_time.freq0 == self.freqs_adjoint[0]):
        return self.run_time

```
? 
@momchil-flex  @tomflexcompute

tylerflex:
Optionally, we could just reduce the `RUN_TIME_FACTOR=100` (`run_time_adj = RUN_TIME_FACTOR/fwidth`) to a more reasonable number between 10-100?

momchil-flex:
[...]

PR #1473, opened 2024-02-15
more accurate adjoint run_time

tylerflex:
Previously it was
```py
run_time_adj = 100 / fwidth_adj
```
Now it is
```py
run_time_adj = run_time_fwd + 10 / fwidth_adj
```

#1403

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Uniaxial medium Lithium niobate to material library.
 
 ### Changed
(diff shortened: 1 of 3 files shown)


tylerflex:
@momchil-flex this is rebased against pre/2.7 and I think ready to merge if you approve of the change.

momchil-flex (approved):
Approved, up to you if you want to put it in a 2.6 patch though, 2.7 will surely take another couple months.

tylerflex:
eh, let's just do 2.7. I'm slightly worried eg this run time ends up being too long or too short for some notebooks, so I'd like to fully test before it gets released and I think it's not very urgent.

PR #1474, opened 2024-02-15
Validate that incompatible advanced materials don't intersect

caseyflex:
Certain advanced materials (custom, nonlinear, time modulated, fully anistropic) are not allowed to intersect, but this was not validated on the frontend.

@tylerflex test_IO fails because the old tests/sims/simulation_* files are based on utils.py that actually had such materials intersecting. I revised utils.py to fix this, but test_IO still fails on the old sim files -- how do I address this?


tylerflex:
> @tylerflex test_IO fails because the old tests/sims/simulation_* files are based on utils.py that actually had such materials intersecting. I revised utils.py to fix this, but test_IO still fails on the old sim files -- how do I address this?

Hm, might need to try manually fixing the oldest sim file? I'm not sure I see another way around it. 

If we want to escape the error, we could try calling `pytest.raises` and passing either a `match` or `escape` for that specific validation message, but I think it would be better to just fix the file if it hopefully isn't too much of a pain. Looks like it's just adjusting the size/center of a couple structures?

caseyflex:
> > @tylerflex test_IO fails because the old tests/sims/simulation_* files are based on utils.py that actually had such materials intersecting. I revised utils.py to fix this, but test_IO still fails on the old sim files -- how do I address this?
> 
> Hm, might need to try manually fixing the oldest sim file? I'm not sure I see another way around it.
> 
> If we want to escape the error, we could try calling `pytest.raises` and passing either a `match` or `escape` for that specific validation message, but I think it would be better to just fix the file if it hopefully isn't too much of a pain. Looks like it's just adjusting the size/center of a couple structures?

yeah, I can do - it's all the old sim files from 2.4 - 2.6

tidy3d/components/simulation.py
```diff
@@ -899,6 +899,70 @@ def _validate_2d_geometry_has_2d_medium(cls, val, values):
...
+        conflict_inds = []
+        conflict_types = []
+        for i in range(len(advanced_indices)):
+            ind1 = advanced_indices[i]
+            type1 = advanced_types[i]
+            for j in range(i + 1, len(advanced_indices)):
+                ind2 = advanced_indices[j]
+                type2 = advanced_types[j]
+                # before checking intersection, check if types would conflict
+                types = {type1, type2}
```
tylerflex, line 934: I think just `if type1 == type2` should be fine

tidy3d/components/simulation.py
```diff
@@ -899,6 +899,70 @@ def _validate_2d_geometry_has_2d_medium(cls, val, values):
...
+                advanced_types.append("nonlinear")
+            elif structure.medium.time_modulated:
+                advanced_indices.append(i)
+                advanced_types.append("time_modulated")
+
+        # check all pairs for possible conflicts
+        # (media that aren't allowed to intersect)
+        conflict_inds = []
+        conflict_types = []
+        for i in range(len(advanced_indices)):
```
tylerflex, line 927: probably easier to just iterate over
```py
for ind1, type1 in zip(advanced_indices, advanced_types):
```
and same below

tylerflex:
> yeah, I can do - it's all the old sim files from 2.4 - 2.6
I think if you delete all but 2.4 and re-run the tests, it should auto generate the other ones. but maybe try in a backup branch so it doesn't mess things up..

caseyflex:
> > yeah, I can do - it's all the old sim files from 2.4 - 2.6
> 
> I think if you delete all but 2.4 and re-run the tests, it should auto generate the other ones. but maybe try in a backup branch so it doesn't mess things up..

Thanks I'll try that -- should be better, I could fix the jsons by hand but not sure how to manually edit the h5 files.

caseyflex:
> > > yeah, I can do - it's all the old sim files from 2.4 - 2.6
> > 
> > 
> > I think if you delete all but 2.4 and re-run the tests, it should auto generate the other ones. but maybe try in a backup branch so it doesn't mess things up..
> 
> Thanks I'll try that -- should be better, I could fix the jsons by hand but not sure how to manually edit the h5 files.

It didn't auto-generate the old ones -- it just auto-generated 2.6.0.h5 and 2.6.0.json. The tests then passed, but the old sim h5 and json files are gone.

tylerflex (changes requested):
So this might be a good time to revisit this question we had a few months ago about how to best define incompatibilities between different parts of the spec. While I think the code here is ok, I'm wondering if we can define the incompatibilities a bit more as a data structure or property/method so we can generalize this in the future. Some idea:

```py
class AbstractMedium:

    @property
    def _is_advanced(self) -> bool:
        """Whether a given medium is 'advanced' <- might need a more specific word for it."""

    def _compatible_with(other: AbstractMediumMedium) -> bool:
        """Whether these two mediums are compatible if in structures that intersect."""
```

```py
class Structure:

    def _compatible_with(other: Structure) -> bool:
        """Whether these two structures are compatible."""
        if not (self.medium._is_advanced and other.medium.is_advanced):
            return True
        
       # note, if 1st condition fails, the 2nd won't get triggered.
        if not self.medium._is_compatible(other.medium) and self.intersects(other):
            return False
```

And then I think the validator becomes much simpler to implement?

```py
def _validate_advanced_material_intersections(cls, val, values):
    for i, structure_i in enumerate(structures):
        for j, structure_j in enumerate(structures[i+1:]:
             if not structure_i.is_compatible(structure_j):
                 raise ...
```

What do you think?

tylerflex:
> It didn't auto-generate the old ones -- it just auto-generated 2.6.0.h5 and 2.6.0.json. The tests then passed, but the old sim h5 and json files are gone.

Shoot.. hopefully there's a way to find and replace all of these otherwise I don't know exactly how else to handle it. (I think we need the intermediate files for the updater tests)

tidy3d/components/simulation.py
caseyflex: I think the reason I did it this way is to start the inner for loop at j = i+1, so I don't check each pair twice

tylerflex:
> Thanks I'll try that -- should be better, I could fix the jsons by hand but not sure how to manually edit the h5 files.

hm, I was going to say to load a simulation from json and then export an h5 `to_file` but that would require checking out each version and sounds like a huge pain too.

Just FYI the hdf5 file has a dataset at the top level called "JSON_STR" or something which just contains the json string. Maybe that could be directly edited as a last resort.

tidy3d/components/simulation.py
caseyflex: If we go with your suggestion for adding the medium.is_compatible(other_medium) function, then that breaks the symmetry a bit and we might end up checking each pair twice. Which is ok, just some redundancy which could actually help prevent errors

caseyflex:
> > Thanks I'll try that -- should be better, I could fix the jsons by hand but not sure how to manually edit the h5 files.
> 
> hm, I was going to say to load a simulation from json and then export an h5 `to_file` but that would require checking out each version and sounds like a huge pain too.
> 
> Just FYI the hdf5 file has a dataset at the top level called "JSON_STR" or something which just contains the json string. Maybe that could be directly edited as a last resort.

yeah both a bit unfortunate, but I can try...

tidy3d/components/simulation.py
tylerflex: yea, in my code snippet for the new suggestion, I also start the loop at i+1 in the validator , so that should resolve it?

caseyflex: ah, right.

caseyflex:
> So this might be a good time to revisit this question we had a few months ago about how to best define incompatibilities between different parts of the spec. While I think the code here is ok, I'm wondering if we can define the incompatibilities a bit more as a data structure or property/method so we can generalize this in the future. Some idea:
> 
> ```python
> class AbstractMedium:
> 
>     @property
>     def _is_advanced(self) -> bool:
>         """Whether a given medium is 'advanced' <- might need a more specific word for it."""
> 
>     def _compatible_with(other: AbstractMediumMedium) -> bool:
>         """Whether these two mediums are compatible if in structures that intersect."""
> ```
> 
> ```python
> class Structure:
> 
>     def _compatible_with(other: Structure) -> bool:
>         """Whether these two structures are compatible."""
>         if not (self.medium._is_advanced and other.medium.is_advanced):
>             return True
> 
>        if self.medium.type == other.medium.type:
>             return True
>         
>        # note, if 1st condition fails, the 2nd won't get triggered.
>         if not self.medium._is_compatible(other.medium) and self.intersects(other):
>             return False
> ```
> 
> And then I think the validator becomes much simpler to implement?
> 
> ```python
> def _validate_advanced_material_intersections(cls, val, values):
>     for i, structure_i in enumerate(structures):
>         for j, structure_j in enumerate(structures[i+1:]:
>              if not structure_i.is_compatible(structure_j):
>                  raise ...
> ```
> 
> What do you think?

I restructured according to your suggestion and I think it made the logic much cleaner. I also got rid of the notion of the "type" of an advanced medium, since in principle it could have multiple.

tidy3d/components/medium.py
```diff
@@ -667,6 +667,48 @@ def time_modulated(self) -> bool:
         """Whether any component of the medium is time modulated."""
         return self.modulation_spec is not None and self.modulation_spec.applied_modulation
 
+    @cached_property
+    def nonlinear(self) -> bool:
```
tylerflex, line 671: I'd prefer `is_nonlinear`, `is_custom`, etc. Minor point bakes it more clear these are `bool` and also leaves these names available in case we need them later.

tidy3d/components/medium.py
```diff
@@ -667,6 +667,48 @@ def time_modulated(self) -> bool:
...
+    def fully_anisotropic(self) -> bool:
+        """Whether the medium is fully anisotropic."""
+        return isinstance(self, FullyAnisotropicMedium)
+
+    @property
+    def _is_advanced(self) -> bool:
+        """Whether the medium is 'advanced'. Advanced media can have additional restrictions
+        compared to ordinary media; for example, some pairs of advanced medium types
+        are not allowed to intersect in a simulation."""
+        if self.nonlinear or self.time_modulated or self.custom or self.fully_anisotropic:
```
tylerflex, line 690: I think we can replace these 3 lines with just 
```py
return self.nonlinear or self.time_modulated or self.custom or self.fully_anisotropic
```

tidy3d/components/medium.py
```diff
@@ -667,6 +667,48 @@ def time_modulated(self) -> bool:
...
+    def custom(self) -> bool:
+        """Whether the medium is custom."""
+        return isinstance(self, AbstractCustomMedium)
+
+    @cached_property
+    def fully_anisotropic(self) -> bool:
+        """Whether the medium is fully anisotropic."""
+        return isinstance(self, FullyAnisotropicMedium)
+
+    @property
```
tylerflex, line 685: cached property? or maybe it's not such a big deal here.

tidy3d/components/simulation.py
```diff
@@ -899,6 +899,30 @@ def _validate_2d_geometry_has_2d_medium(cls, val, values):
...
+            if structure.medium._is_advanced:
+                advanced_indices.append(i)
+                advanced_structures.append(structure)
+        for i, (ind1, structure_ind1) in enumerate(zip(advanced_indices, advanced_structures)):
+            for ind2, structure_ind2 in zip(
+                advanced_indices[i + 1 :], advanced_structures[i + 1 :]
+            ):
+                if not structure_ind1._compatible_with(structure_ind2):
+                    raise ValidationError(
+                        f"The structure at structures[{ind1}] and the structure at structures[{ind2}] "
```
tylerflex, line 920: let's put the code bits `(structures[{ind1}])` in single quotes `'` just to make the terminal output nicer.

tidy3d/components/simulation.py
```diff
@@ -899,6 +899,30 @@ def _validate_2d_geometry_has_2d_medium(cls, val, values):
...
+                        "have incompatible advanced medium types, and so are not allowed to intersect. "
```
tylerflex, line 921: let's insert the medium types while we're at it.
```py
... medium types of '{structure_ind1.medium.type}' and '{structure_ind2.medium.type}', respectively.
```

tylerflex (commented):
much improved, thanks Casey! couple minor comments.

Also, I assume once #1478 is merged the tests might pass?

tidy3d/components/medium.py
```diff
@@ -667,6 +667,48 @@ def time_modulated(self) -> bool:
...
+        if self.nonlinear or self.time_modulated or self.custom or self.fully_anisotropic:
+            return True
+        return False
+
+    def _compatible_with(self, other: AbstractMedium) -> bool:
+        """Whether these two media are compatible if in structures that intersect."""
+        # only need to check compatibility if both media are advanced
+        if not (self._is_advanced and other._is_advanced):
+            return True
+        for (med1, med2) in [(self, other), (other, self)]:
```
tylerflex, line 699: perhaps could be simplified / refactored a bit more

```py
@staticmethod
def _check_compatibility(med1, med2) -> bool:
    
    if self.custom:
        # custom and fully_anisotropic is OK
        if other.nonlinear or other.time_modulated:
            return False
    if self.fully_anisotropic:
        if other.nonlinear or other.time_modulated:
            return False
    if self.nonlinear:
        if other.time_modulated:
            return False    
    return True

def _compatible_with(self, other):
    if not (self._is_advanced and other._is_advanced):
        return True
    return self._check_compatibility(self, other) and self._check_compatibility(other, self)
```

tidy3d/components/medium.py
caseyflex: I might leave time_modulated instead of changing it to is_time_modulated, just to avoid the need for a backend PR.

tidy3d/components/medium.py
```diff
@@ -667,6 +667,48 @@ def time_modulated(self) -> bool:
...
+        """Whether the medium is custom."""
+        return isinstance(self, AbstractCustomMedium)
+
+    @cached_property
+    def fully_anisotropic(self) -> bool:
+        """Whether the medium is fully anisotropic."""
+        return isinstance(self, FullyAnisotropicMedium)
+
+    @property
+    def _is_advanced(self) -> bool:
```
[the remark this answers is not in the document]
caseyflex: yeah, agreed that is nicer

caseyflex:
> much improved, thanks Casey! couple minor comments.
> 
> Also, I assume once #1478 is merged the tests might pass?

I had to delete the json and h5 files for `2.5.1`. I wasn't quite sure if we are keeping those, or if so what the fix was to update SIM_FULL in this old file?

tidy3d/components/medium.py
tylerflex: ok that's fine

tylerflex:
> I had to delete the json and h5 files for 2.5.1. I wasn't quite sure if we are keeping those, or if so what the fix was to update SIM_FULL in this old file?

I'm not sure I understand, but after that PR, I think we are just maintaining a `simulation_sample.json` now (no explicit version numbers) and that file corresponds to the last minor version. I dont think there should have been a 2.5.1 so maybe that file got either generated by a test or wasn't deleted. Maybe @momchil-flex knows

tidy3d/components/medium.py
tylerflex: hm, might be confusing if they follow different conventions?

momchil-flex:
Hm yeah I don't know why 2.5.1 is still there but it can be deleted.

caseyflex:
I renamed time_modulated to is_time_modulated and made a corresponding backend PR. Also, I added validators and tests to forbid nonlinear_spec or modulation_spec on components of a `Medium2D`.

weiliangjin2021 (approved):
If I understand correctly, the incompatibility is checked not against actual structure intersections, but bounding box? It sounds too strict, but fast to do.

caseyflex:
> If I understand correctly, the incompatibility is checked not against actual structure intersections, but bounding box? It sounds too strict, but fast to do.

In GUI, if I understand correctly, they don't check intersections at all, just disallow conflicting types in the entire simulation.

e-g-melo:
> > If I understand correctly, the incompatibility is checked not against actual structure intersections, but bounding box? It sounds too strict, but fast to do.
> 
> In GUI, if I understand correctly, they don't check intersections at all, just disallow conflicting types in the entire simulation.
> 
> I guess we can relax this in the future if required. But it would be better to make the advanced material properties compatible in the solver. This is meant to be temporary until we can support that functionality. I don't see a fast or easy way to actually check intersections on frontend. Do you?

That is correct. It was a temporary workaround in GUI, but indeed, we should allow the user to include different advanced material properties.

tylerflex:
Looks good to me @caseyflex . Just a reminder about this [comment](https://github.com/flexcompute/tidy3d/pull/1474#discussion_r1492689266). I think we should change mentions of `advanced material` to mentions of `has compatibility`  in the code variables, tests, comments, changlog etc. Just for specificity and to avoid a future where we might need to introduce "really advanced materials" etc.  :) 

https://github.com/flexcompute/tidy3d/pull/1474#discussion_r1492689266

caseyflex:
> Looks good to me @caseyflex . Just a reminder about this [comment](https://github.com/flexcompute/tidy3d/pull/1474#discussion_r1492689266). I think we should change mentions of `advanced material` to mentions of `has incompatibilities` in the code variables, tests, comments, changlog etc. Just for specificity and to avoid a future where we might need to introduce "really advanced materials" etc. :)
> 
> [#1474 (comment)](https://github.com/flexcompute/tidy3d/pull/1474#discussion_r1492689266)

done, should be fine to merge

tylerflex:
Perhaps let's merge this in for 2.7? to avoid any potential issues? I'll leave it up to @momchil-flex

e-g-melo (approved):
Hi @caseyflex! Thanks for including this validation. I've just tried it using a simple setup and it worked fine.

I have only a comment, not related to this issue. Why do we show all these warnings after the validation error? 
image

weiliangjin2021:
> > If I understand correctly, the incompatibility is checked not against actual structure intersections, but bounding box? It sounds too strict, but fast to do.
> 
> In GUI, if I understand correctly, they don't check intersections at all, just disallow conflicting types in the entire simulation.
> 
> I guess we can relax this in the future if required. But it would be better to make the advanced material properties compatible in the solver. This is meant to be temporary until we can support that functionality. I don't see a fast or easy way to actually check intersections on frontend. Do you?

I don't see an easy and efficient way either.

momchil-flex:
> Hi @caseyflex! Thanks for including this validation. I've just tried it using a simple setup and it worked fine.
> 
> I have only a comment, not related to this issue. Why do we show all these warnings after the validation error? image

@tylerflex @lucas-flexcompute @dbochkov any idea if these warnings can be avoided e.g. with some pydantic settings?

weiliangjin2021:
should this be merged to pre/2.7?

weiliangjin2021:
> Is there a problem with it being in the next 2.6 patch?

In our previous workflow, the patch will be merged to develop. Has this been changed?

momchil-flex:
Ok I see your point I thought this was merged in develop and you were just asking if it should go in 2.7 instead.

I'll fix tomorrow

lucas-flexcompute:
> > Hi @caseyflex! Thanks for including this validation. I've just tried it using a simple setup and it worked fine.
> > I have only a comment, not related to this issue. Why do we show all these warnings after the validation error? image
> 
> @tylerflex @lucas-flexcompute @dbochkov any idea if these warnings can be avoided e.g. with some pydantic settings?

My understanding is that field validators are still executed if other validators have failed unless we set `skip_on_failure=True` on them, so that could be a solution.

PR #1478, opened 2024-02-16
Store a simulation sample file every time schema.py is called

momchil-flex:


tests/sims/simulation_sample.json
tylerflex, line 1: So just so I understand, basically when we release version x.y.z, `simulation_sample` will be overwritten by SIM_FULL from version `x.y.z`? or from the **previous** version?

So after the release, we update `__version__` and start working on a branch for, say `x.y.{z+1}`, `simulation_sample` will then be the old version and the IO tests will load it into `x.y.{z+1}`?

tylerflex: I guess a problem is if we use this release script for a patch, it will not really test the updating from `x.y.z` to `x.{y+1}.z`?

momchil-flex: Yes, that's my thinking.

momchil-flex: How's this?

tylerflex: looks good to me!

tylerflex (approved):

PR #1480, opened 2024-02-19
PATCH: Simulation API docs improvements

daquinteroflex:
Some of my `Simulation` API docs docs got moved into `AbstractSimulation` during the git resolution between the two branches. This means that some docs went into `AbstractSimulation` when they were written for FDTD. So this corrects part of this on the `symmetry` parameter which Emerson pointed out. However, it also means that in order to have proper documentation for each of these parameters for the FDTD simulation, it is necessary to redeclare them within the FDTD `Simulation` class. I expect this to be the case for any other specific `AbstractSimulation` subclasses.


tylerflex (approved):
Looks fine to me but let's get @dbochkov-flexcompute to take a quick look too since he wrote the `AbstractSimulation`. I'm not 100% sure about the fields.

daquinteroflex:
I have added a few patches from my attempt at a super thorough docs review btw, haven't rebased them so easier to review Tyler. Haven't found any other updates so far so maybe good enough to merge this PR  for the release when Daniil checks.

tidy3d/components/base_sim/simulation.py
```diff
@@ -37,28 +37,6 @@ class AbstractSimulation(Box, ABC):
...
-
-    * `Modeling dispersive material in FDTD <https://www.flexcompute.com/fdtd101/Lecture-5-Modeling-dispersive-material-in-FDTD/>`_
-
-    **GUI:**
-
-    * `Mediums <https://www.flexcompute.com/tidy3d/learning-center/tidy3d-gui/Lecture-2-Mediums/>`_
-
     """
 
     structures: Tuple[Structure, ...] = pd.Field(
```
dbochkov-flexcompute, line 42: should docstrings for structures be also cleared from FDTD related info? I.e., starting from `**Usage Caveats**`

daquinteroflex:
I believe tests will pass this time so feel free to merge in whenever in case building the 2.6 pre release

tylerflex (approved):
looks good to me! Thank you!

PR #1486, opened 2024-02-22
RELEASE: Upgrade release flow and tiny fix :zap:

daquinteroflex:
This upgrades:

- [X] So that `sync-readthedocs-repo` gets live updated from develop
- [X] A tiny aesthetic fix on the first docs page.
- [x] When a release is tagged, it merges develop to latest branch. This will trigger the documentation sync flow on the main website.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@master
+    - name: Exit if any RC release
+      if: contains(github.ref, 'rc') == false
+      uses: everlytic/branch-merge@1.1.2
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        source_ref: "develop"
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -13,6 +13,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@master
+    - uses: everlytic/branch-merge@1.1.2
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        source_ref: "develop"
+        target_branch: "latest"
+        commit_message_template: 'RELEASE: :shipit: :boom: :tada: Merged {source_ref} into target {target_branch}'
```
momchil-flex, line 21: We shouldn't do that if the release was not tagged on develop. Namely I'm thinking of pre-releases.

One option is to check if `rc` is in the tag?

daquinteroflex: Ooh yeah that makes sense

PR #1490, opened 2024-02-22
Fix to monitor discretize inds if there's an empty set along at least…

momchil-flex:
… one dimension


tylerflex (approved):
I'm not sure I understand all of the ramifications of this but I trust you!

tylerflex:
Maybe add a changelog Fixed if you think it warrants it?

Issue #1352, opened 2024-01-05
Add metadata to tidy3d components


tomflexcompute:
The following is coming from a user's feedback:

When doing a batch for parameter sweep, users need to actively keep track of the parameter values used for each simulation. If we reload the BatchData in the future, there is no record of the parameter values used, which leads to difficulties in data analysis. 

Would be good to add a `comment` in td.Simulation such that users can store reminder information when creating a simulation. This information can be carried into simulation data and extracted when doing data analysis to remind themselves. For example,

sim = td.Simulation(center=xxx,..., comment="a=3, b=4")
sim_data = web.run(xxx)
print(sim_data.comment)

There are many alternatives but the general idea is to store some user-specified information somewhere that can be retrieved later on as a reminder.

tylerflex:
We discussed an idea like this regarding the parameter sweep tool. I think it could be useful to allow users to append some arbitrary metadata to a `Simulation` (or any tidy3d component for that matter). In xarray/pandas (probably others), this typically is called `.attrs` and is a `dict`. https://docs.xarray.dev/en/latest/generated/xarray.DataArray.attrs.html

Maybe we could follow that convention too? I guess some besides naming, we should also consider whether there would be any security issues with allowing this data to be uploaded to our servers? Should we validate it in any way? Also, should we retain `attrs` if the object is modified (eg `updated_copy()`)?

@tomflexcompute once the `design` sweep tool is related in 2.6.0rc1 you can tell the user that they can use it. It basically does exactly what they want (the inputs are associated with the outputs and there are references to the simulation data, it basically does all of this accounting under the hood).

tomflexcompute:
> We discussed an idea like this regarding the parameter sweep tool. I think it could be useful to allow users to append some arbitrary metadata to a `Simulation` (or any tidy3d component for that matter). In xarray/pandas (probably others), this typically is called `.attrs` and is a `dict`. https://docs.xarray.dev/en/latest/generated/xarray.DataArray.attrs.html
> 

Thanks for sharing your thoughts @tylerflex . That sounds good to me. 

> Maybe we could follow that convention too? I guess some besides naming, we should also consider whether there would be any security issues with allowing this data to be uploaded to our servers? Should we validate it in any way? Also, should we retain `attrs` if the object is modified (eg `updated_copy()`)?
> 

Why are there security concerns? Intuitively I think if a user copies a simulation, the comment should be copied as well. They need to update it if they want.

[...]

PR #1493, opened 2024-02-23
Add metadata to Tidy3D components

tylerflex:
Addresses #1352 

How do you all feel about this?

1. Used the name `.attrs`, because it follows the same [convention](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.attrs.html) as `xarray`, `pandas`, maybe others.
2. `attrs` are dictionaries (mutable) so they can be set on the fly, `sim.attrs['foo'] = bar`.
3. we still can't set the attrs directly since tidy3d objects dont support item assignment, eg.`sim.attrs = {}` doesn't work.
4. in `xarray`, attrs remain when the objects are copied and updated. but are removed when the objects are modified. We only need to handle the copy part since we can't modify tidy3d objects. therefore if `sim.copy()`, then the new object's attrs are the same.
5. Important point, since `attrs` are mutable, we want to ensure our internal code doesn't depend on attrs in any way. Otherwise we introduce state. So as a rule, we don't use `attrs` for any calculation? That being said, maybe we can set attrs internally if it makes sense to, but never code like `if obj.attrs["something"] == something: ...`

All of the logic is spelled out in the [test](https://github.com/flexcompute/tidy3d/pull/1493/files#diff-d8a2875091a9aff89fbeb80d849d5ee59a5f67a08a88919e0a63036f4126cdb1).

Also please check the [description](https://github.com/flexcompute/tidy3d/pull/1493/files#diff-d331aef8c3290adc3733d4ce375c5a4b139ad41bd5634113ca404cc827e3d949) and see if it is clear.

Thanks

### DIFF
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -192,6 +192,19 @@ class Tidy3dBaseModel(pydantic.BaseModel):
                     )
         return values
 
+    attrs: dict = pydantic.Field(
+        {},
+        title="Attributes",
+        description="Dictionary storing arbitrary metadata for a Tidy3D object. "
+        "This dictionary can be freely used by the user for storing data without affecting the "
+        "operation of Tidy3D as it is not used internally. "
+        "Note that, unlike regular Tidy3D fields, ``attrs`` are mutable. "
+        "For example, the following is allowed for setting an ``attr`` ``obj.attrs['foo'] = bar``. "
+        "Also note that `Tidy3D`` will raise a ``TypeError`` if ``attrs`` contain objects "
+        "that can not be serialized. One can check if ``attrs`` are serializable "
+        "by calling ``obj.json()``.",
+    )
+
     def copy(self, **kwargs) -> Tidy3dBaseModel:
         """Copy a Tidy3dBaseModel.  With ``deep=True`` as default."""
         if "deep" in kwargs and kwargs["deep"] is False:
--- a/tidy3d/components/grid/grid.py
+++ b/tidy3d/components/grid/grid.py
@@ -8,7 +8,7 @@ import pydantic.v1 as pd
 from ..base import Tidy3dBaseModel
 from ..data.data_array import DataArray, SpatialDataArray, ScalarFieldDataArray
 from ..data.dataset import UnstructuredGridDatasetType, UnstructuredGridDataset
-from ..types import ArrayFloat1D, Axis, TYPE_TAG_STR, InterpMethod, Literal
+from ..types import ArrayFloat1D, Axis, InterpMethod, Literal
 from ..geometry.base import Box
 
 from ...exceptions import SetupError
@@ -43,7 +43,7 @@ class Coords(Tidy3dBaseModel):
     @property
     def to_dict(self):
         """Return a dict of the three Coord1D objects as numpy arrays."""
-        return {key: np.array(value) for key, value in self.dict(exclude={TYPE_TAG_STR}).items()}
+        return {key: self.dict()[key] for key in "xyz"}
 
     @property
     def to_list(self):
@@ -386,9 +386,7 @@ class Grid(Tidy3dBaseModel):
         >>> grid = Grid(boundaries=coords)
         >>> Nx, Ny, Nz = grid.num_cells
         """
-        return [
-            len(coords1d) - 1 for coords1d in self.boundaries.dict(exclude={TYPE_TAG_STR}).values()
-        ]
+        return [len(self.boundaries.dict()[dim]) - 1 for dim in "xyz"]
 
     @property
     def _primal_steps(self) -> Coords:
@@ -412,7 +410,7 @@ class Grid(Tidy3dBaseModel):
             applied.
         """
 
-        primal_steps = self._primal_steps.dict(exclude={TYPE_TAG_STR})
+        primal_steps = {dim: self._primal_steps.dict()[dim] for dim in "xyz"}
         dsteps = {key: (psteps + np.roll(psteps, 1)) / 2 for (key, psteps) in primal_steps.items()}
 
         return Coords(**dsteps)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Support for `.gz` files in `Simulation` version updater.
 - Warning if a nonuniform custom medium is intersecting `PlaneWave`, `GaussianBeam`, `AstigmaticGaussianBeam`, `FieldProjectionCartesianMonitor`, `FieldProjectionAngleMonitor`, `FieldProjectionKSpaceMonitor`, and `DiffractionMonitor`.
+- Tidy3D objects may store arbitrary metadata in an `.attrs` dictionary.
 
 ### Changed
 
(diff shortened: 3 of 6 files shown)


lucas-flexcompute (approved):

lucas-flexcompute (approved):
Just one question: it is possible to get an error if the user adds non-serializable data to the dictionary, because it will be included in the json representation of the model, right?

I think that's the correct behavior here, but we might also want to add a note about that in the description (and maybe try to catch it early and warn the user when creating the json string).

tylerflex:
> Just one question: it is possible to get an error if the user adds non-serializable data to the dictionary, because it will be included in the json representation of the model, right?

Yea it should be, good point. I'll add a test and also see what I can do about catching it.

tidy3d/components/base.py
```diff
@@ -180,6 +180,20 @@ class Config:
 
     _cached_properties = pydantic.PrivateAttr({})
 
+    attrs: dict = pydantic.Field(
+        {},
+        title="Attributes",
+        description="Dictionary storing arbitrary metadata for a Tidy3D object. "
```
tylerflex, line 186: @lucas-flexcompute can you check the new description? any comments?

momchil-flex:
> 5. Important point, since `attrs` are mutable, we want to ensure our internal code doesn't depend on attrs in any way. Otherwise we introduce state. So as a rule, we don't use `attrs` for any calculation? That being said, maybe we can set attrs internally if it makes sense to, but never code like `if obj.attrs["something"] == something: ...`

This point sounds like going against the idea of using these for the `jax_info`. Or would that be considered internal usage? Basically when submitting a JaxSimulation to the server, we split it into a Simulation and JaxInfo, and add the latter as attrs to the former.

One more question to bring up (@xin-flex) if two simulations only differ in their `attrs` should they be considered the same or different in terms of caching?

tylerflex:
> This point sounds like going against the idea of using these for the jax_info. Or would that be considered internal usage? Basically when submitting a JaxSimulation to the server, we split it into a Simulation and JaxInfo, and add the latter as attrs to the former.

I think we could consider setting `jax_info` as a `pd.Field` in the `JaxSimulation` instead of using `attrs` maybe?

tylerflex:
Hm, but then how do we upload this information? maybe that doesn't work after all.

tylerflex:
@momchil-flex could you take a look into this PR and let me know if you need any other changes?  I think I'll be doing the jax refactor stuff we discussed without attrs, although I'm still working that out.

tidy3d/components/base.py
```diff
@@ -180,6 +180,20 @@ class Config:
...
+        "This dictionary can be freely used by the user for storing data without affecting the "
+        "operation of Tidy3D as it is not used internally. "
+        "Note that, unlike regular Tidy3D fields, ``attrs`` are mutable. "
+        "For example, the following is allowed for setting an ``attr`` ``obj.attrs['foo'] = bar``. "
+        "Also note that if objects that are not serializable (eg types) are added to "
```
momchil-flex: what do you mean by "eg types"? Like if I have something like `[Callable]` in my attrs?

tidy3d/components/base.py
```diff
@@ -180,6 +180,20 @@ class Config:
...
+        "``attrs``, ``Tidy3D`` will raise a ``TypeError`` whenever that instance is serialized, "
+        "eg. when writing to file. Therefore, users need to ensure that their ``attrs`` are "
```
momchil-flex: e.g.

tidy3d/components/grid/grid.py
```diff
@@ -15,12 +15,15 @@
...
+# keys to exclude from dict in grid operations
+DICT_EXCLUDE_KEYS = {TYPE_TAG_STR, "attrs"}
+
 
 class Coords(Tidy3dBaseModel):
     """Holds data about a set of x,y,z positions on a grid.
 
     Example
-    -------
+    -------tests/test_web/test_webapi_heat.py
```
momchil-flex: this seems weird?

tidy3d/components/grid/grid.py
```diff
@@ -15,12 +15,15 @@
 # data type of one dimensional coordinate array.
 Coords1D = ArrayFloat1D
 
+# keys to exclude from dict in grid operations
+DICT_EXCLUDE_KEYS = {TYPE_TAG_STR, "attrs"}
```
momchil-flex: Why was this necessary here and might it be necessary in other places too (in the future)?

momchil-flex (commented):
The way this is currently, identical simulations with different `attrs` will not go through the caching mechanism, right? I think this may be fine, but I do remember maybe at some point discussing having the opposite behavior? It's probably harder to implement though. Any thoughts?

tidy3d/components/grid/grid.py
tylerflex: Basically, `attrs` gets written into the `dict()` and the `grid` uses `self.dict(exclude={"type"})` to get the xyz items. If we don't exclude the `attrs`, it will get included in the dictionary and will break the grid.

Options moving forward are to keep adding to this, or try to change the grid iteration to just iterate through `'xyz'` directly. The latter sounds a bit more robust so I'll try that.

tylerflex: yea this worked better, I just changed the `grid` code to iterate through `xyz` and that way if we change the contents of `dict()` we don't need to worry about this.

tidy3d/components/grid/grid.py
tylerflex: yea, what? :D fixed.

tidy3d/components/base.py
tylerflex: fixed

tidy3d/components/base.py
tylerflex: I mean that you can't put `int` in your attrs, e.g. `attrs = {'test': int}`

tylerflex: I just removed this `eg types` comment because it might just be confusing

tylerflex:
```
Also note that `Tidy3D`` will raise a ``TypeError`` if ``attrs`` contain objects that can not be serialized.
```

tylerflex:
> The way this is currently, identical simulations with different attrs will not go through the caching mechanism, right? I think this may be fine, but I do remember maybe at some point discussing having the opposite behavior? It's probably harder to implement though. Any thoughts?

I'm not excactly sure how the caching functions. Do you know what it compares based of?  In my mind, it probably makes most sense to ignore attrs when deciding whether to load a cached dataset but I don't know

momchil-flex:
> > The way this is currently, identical simulations with different attrs will not go through the caching mechanism, right? I think this may be fine, but I do remember maybe at some point discussing having the opposite behavior? It's probably harder to implement though. Any thoughts?
> 
> I'm not excactly sure how the caching functions. Do you know what it compares based of? In my mind, it probably makes most sense to ignore attrs when deciding whether to load a cached dataset but I don't know

I believe it compares the hdf5 file byte by byte, so it will take some engineering to modify this to exclude attrs.

But, in some use cases, that's probably safer. For example, the `SimulationData` will contain the `Simulation` that contains the `attrs`. If caching kicks in, the returned `SimulationData` will have different `attrs` than the `Simulation` that was submitted. To avoid this it would take even more engineering...

tylerflex:
k, I feel comfortable taking the most conservative route and just caching based on attrs for now.. or until someone complains :)

PR #1498, opened 2024-02-24
Unstructured custom medium

dbochkov-flexcompute:
This PR allows to use unstructured grid datasets (`TriangularGridDataset` and `TetrahedralGridDataset`) in unstructured medium classes. All necessary associated functionality is extended and tested, which includes:
- unstructured grid dataset interpolation improvement:
-- new vectorized interpolation
-- option `extrapolate` by nearest neighbor values
-- nearest neighbor interpolation
- loading `DataArray`  without coordinates (for example, when `arr = td.PointDataArray(np.ones((3, 2)), dims=["x", "y"])`, that is `dims` provided but not `coords`)
- `sel_inside` and `does_cover` for unstructured grid datasets
- `Coords.spatial_interp` is extended to work both with `SpatialDataArray` and `UnstructuredGridDataset`

There are quite a lot of changes, so maybe what would make sense is:
- for @tylerflex to review on the high level
- for @weiliangjin2021 to review more in detail parts related to custom medium classes and their testing
- for @caseyflex to review new interpolation method for unstructured grid datasets, that is, file [tidy3d/components/data/dataset.py](https://github.com/flexcompute/tidy3d/pull/1498/files#diff-3d6bc96fe49f31f2d21f3f369c8e3262ea0264f353d8c092ef5065c9e8db9d04) 

what do you think?

Example notebooks on heat and charge will be converted to use only unstructured data after this PR is merged. Previews:
- https://github.com/flexcompute/tidy3d-notebooks/blob/daniil/unstructured-medium-heat/ThermallyTunedRingResonator.ipynb
- https://github.com/flexcompute/tidy3d-notebooks/blob/daniil/unstructured-medium-heat/MetalHeaterPhaseShifter.ipynb
- https://github.com/flexcompute/tidy3d-notebooks/blob/daniil/unstructured-medium-heat/MachZehnderModulator.ipynb

### DIFF
--- a/tidy3d/components/data/dataset.py
+++ b/tidy3d/components/data/dataset.py
@@ -2,16 +2,17 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Union, Dict, Callable, Any
+from typing import Union, Dict, Callable, Any, Tuple
 
 import xarray as xr
 import numpy as np
+from scipy.interpolate import NearestNDInterpolator
 import pydantic.v1 as pd
 from matplotlib.tri import Triangulation
 from matplotlib import pyplot as plt
 import numbers
 
-from .data_array import DataArray
+from .data_array import DataArray, DATA_ARRAY_MAP
 from .data_array import ScalarFieldDataArray, ScalarFieldTimeDataArray, ScalarModeFieldDataArray
 from .data_array import ModeIndexDataArray, GroupIndexDataArray, ModeDispersionDataArray
 from .data_array import TriangleMeshDataArray
@@ -21,11 +22,15 @@ from .data_array import PointDataArray, IndexedDataArray, CellDataArray, Spatial
 from ..viz import equal_aspect, add_ax_if_none, plot_params_grid
 from ..base import Tidy3dBaseModel, cached_property
 from ..base import skip_if_fields_missing
-from ..types import Axis, Bound, ArrayLike, Ax, Coordinate, Literal
+from ..types import Axis, Bound, ArrayLike, Ax, Coordinate, Literal, annotate_type
 from ...packaging import vtk, requires_vtk
 from ...exceptions import DataError, ValidationError, Tidy3dNotImplementedError
-from ...constants import PICOSECOND_PER_NANOMETER_PER_KILOMETER
+from ...constants import PICOSECOND_PER_NANOMETER_PER_KILOMETER, inf
 from ...log import log
+
+
+DEFAULT_MAX_SAMPLES_PER_STEP = 1e4
+DEFAULT_MAX_CELLS_PER_STEP = 1e4
 
 
 class Dataset(Tidy3dBaseModel, ABC):
@@ -528,6 +533,11 @@ class UnstructuredGridDataset(Dataset, np.lib.mixins.NDArrayOperatorsMixin, ABC)
         # we redirect name to values.name
         return self.values.name
 
+    @property
+    def is_complex(self) -> bool:
+        """Data type."""
+        return np.iscomplexobj(self.values)
+
     @pd.validator("cells", always=True)
     def match_cells_to_vtk_type(cls, val):
         """Check that cell connections does not have duplicate points."""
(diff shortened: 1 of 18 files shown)


tylerflex (commented):
Thanks @dbochkov-flexcompute this is really impressive. I took a skim through the code and also the notebooks. I just had a general and very basic question for now about how to interpret the values in the custom dataset and also how we interpolate these datasets onto the yee lattice.

If we have a triangular grid dataset of permittivity values. Do the "values" determine the permittivities in the 'center' of each triangle? (average of all vertices?) and then when we are evaluating this dataset at a point (x,y,z) how does it work? For example do we find the elements surrounding the point and then linearly interpolate from the values on each surface? 

Similar question for tetrahedral. I assume here we take the values to mean the permittivity in the volume defined by the tetrahedral, so does that mean to evaluate at (x,y,z) we would check which tetrahedral it is enclosed by and simply interpolate the value?

Might be something else entirely but I'm just trying to understand a bit better how the values are defined and how they get interpolated / evaluated. It will be great to highlight this also in the docstrings and maybe notebooks if not already.

dbochkov-flexcompute:
@tylerflex in our unstructured datasets values are assigned to points rather than cells. And this makes interpolation fairly straightforward: to get a value at an arbitrary point we just need to find which cell (triangle/tetrahedron) contains that point, and then use standard barycentric (linear) interpolation from values defined at the vertices of that cell. I will mentioned that in docstrings for `Triangular`/`TetrahedralGridDataset`

tidy3d/components/data/dataset.py
```diff
@@ -924,13 +976,16 @@ def box_clip(self, bounds: Bound) -> UnstructuredGridDataset:
...
         return self._from_vtk_obj(clean_clip)
 
-    @requires_vtk
     def interp(
         self,
         x: Union[float, ArrayLike],
         y: Union[float, ArrayLike],
         z: Union[float, ArrayLike],
-        fill_value: float = 0,
+        fill_value: Union[float, Literal["extrapolate"]] = "extrapolate",
```
caseyflex, line 984: Perhaps we should warn that the default behavior is changing?

tidy3d/components/data/dataset.py
```diff
@@ -942,8 +997,17 @@ def interp(
             y-coordinates of sampling points.
         z : Union[float, ArrayLike]
             z-coordinates of sampling points.
-        fill_value : float = 0
-            Value to use when filling points without interpolated values.
+        fill_value : Union[float, Literal["extrapolate"]] = "extrapolate"
+            Value to use when filling points without interpolated values. If ``extrapolate`` then
+            nearest values are used.
+        use_vtk : bool = False
```
caseyflex, line 1003: Can you clarify here what are the differences between vtk's interpolation and our own interpolation? In terms of result or performance

tidy3d/components/data/dataset.py
```diff
@@ -942,8 +997,17 @@ def interp(
...
+            Use vtk's interpolationfunctionality or Tidy3D's own implementation.
+        method: Literal["linear", "nearest"] = "linear"
+            Interpolation method to use.
+        max_samples_per_step : int = 1e4
```
caseyflex, line 1007: Some of these parameters are rather technical and used only in our own interpolation. Maybe @tylerflex has a better suggestion, but one possibility is to make these parameters kwargs here, and make the `interp_vtk` and `interp_py` functions public so that the `interp_py` docstring would have the details of these parameters

tidy3d/components/data/dataset.py
caseyflex: Also, `use_vtk` is only used if `method="linear"` I guess. My understanding is that `use_vtk` would only be used if the default behavior had performance issues? Maybe it makes sense to get rid of `use_vtk` and just have this interp function use our own interpolation (either interp_nearest or interp_py). Then, we could make interp_vtk a public function for the case when one does need to use vtk interpolation.

tidy3d/components/data/dataset.py
```diff
@@ -955,6 +1019,131 @@ def interp(
         x = np.atleast_1d(x)
         y = np.atleast_1d(y)
         z = np.atleast_1d(z)
+
+        if method == "nearest":
+            interpolated_values = self._interp_nearest(x=x, y=y, z=z)
+        else:
+            if fill_value == "extrapolate" and method != "nearest":
```
caseyflex, line 1026: don't need `method != nearest` in this check or below, since we are in `else`

tidy3d/components/data/dataset.py
```diff
@@ -985,7 +1174,404 @@ def interp(
...
+        fill_value : float
+            Value to use when filling points without interpolated values.
+        max_samples_per_step : int = 1e4
+            Max number of points to interpolate at per iteration.
+        max_cells_per_step : int = 1e4
+            Max number of cells to interpolate from per iteration.
+        rel_tol : float = 1e-10
+            Relative tolerance when comparing coordinates.
+        axis_ignore : Union[Axis, None] = None
+            For interpolati
```
caseyflex, line 1209: comment got cutoff. I guess axis_ignore is None for Tetrahedral and normal axis for Triangular?

tidy3d/components/data/dataset.py
```diff
@@ -1345,34 +2025,100 @@ def interp(
...
+        fill_value : Union[float, Literal["extrapolate"]] = "extrapolate"
+            Value to use when filling points without interpolated values. If ``extrapolate`` then
+            nearest values are used.
+        use_vtk : bool = False
+            Use vtk's interpolationfunctionality or Tidy3D's own implementation.
+        method: Literal["linear", "nearest"] = "linear"
+            Interpolation method to use.
         ignore_normal_pos : bool = True
-            Assume data is invariant along the normal direction to the grid plane.
+            (Depreciated) Assume data is invariant along the normal direction to the grid plane.
```
caseyflex, line 2036: Sometimes when we deprecate something we continue to support the old behavior (and maintain same defaults if possible). But I guess this would not really be used?

tidy3d/components/data/dataset.py
```diff
@@ -1678,3 +2474,69 @@ def sel(
...
+) -> Union[ArrayLike, xr.DataArray, UnstructuredGridDataset]:
+    """Get a unity replica of dataarray/dataset."""
+    if isinstance(data_array, UnstructuredGridDataset):
+        return data_array.updated_copy(values=xr.ones_like(data_array.values))
+    if isinstance(data_array, xr.DataArray):
+        return xr.ones_like(data_array)
+    return np.ones_like(data_array)
+
+
+def _check_same_coordinates(
```
caseyflex, line 2514: Are we consistent in the tolerances, like using `rel_tol=1e-10` above but the `np.allclose` defaults here?

caseyflex (approved):
Thanks Daniil, looks good. My comments were mostly about the API. I just skimmed through the new interpolation logic -- let me know if there's anything there you want me to look at more closely.

tidy3d/components/data/dataset.py
dbochkov-flexcompute: True. Actually, after thinking about it more I decided to stick with our protocol here (no api breaks) even though `fill_value="extrapolate"` typically makes more sense and these classes were introduced just recently. I did add a warning though to let user know that this behavior will change. To do that, I assigned default value to `None` which is converted to 0 internally. This allows to understand whether `fill_value` was set explicitly or the default value is used.

tidy3d/components/data/dataset.py
dbochkov-flexcompute: fixed

tidy3d/components/data/dataset.py
dbochkov-flexcompute: that's exactly right, fixed.

dbochkov-flexcompute: also, to make code a bit more clear, I renamed the main internal interpolation function into `_interp_py_general` to highlight it is implemented for both 2D and 3D cells simulateneously. Then `TriangularGridDataset` and `TetrahedralGridDataset` each have their own `_interp_py` that wraps `_interp_py_general`.

tidy3d/components/data/dataset.py
dbochkov-flexcompute: yeah, introduced this for no clear reason and cannot think of when it would be helpful. I think it's ok to stop supporting it since it was introduced just a couple versions ago and it wasn't the default value.

tidy3d/components/data/dataset.py
```diff
@@ -1676,5 +2576,122 @@ def sel(
...
+            rel_tol=rel_tol,
+            axis_ignore=None,
+        )
+
 
 UnstructuredGridDatasetType = Union[TriangularGridDataset, TetrahedralGridDataset]
+CustomSpatialDataType = Union[SpatialDataArray, TriangularGridDataset, TetrahedralGridDataset]
+CustomSpatialDataTypeAnnotated = Union[
+    SpatialDataArray, annotate_type(Union[TriangularGridDataset, TetrahedralGridDataset])
+]
```
weiliangjin2021, line 2635: I wonder why not 
```
CustomSpatialDataType = Union[SpatialDataArray, UnstructuredGridDatasetType]
CustomSpatialDataTypeAnnotated = Union[SpatialDataArray, annotate_type(UnstructuredGridDatasetType)]
```

weiliangjin2021 (approved):
Quite intensive work as I glance through medium.py. All look good to me.

tidy3d/components/data/dataset.py
dbochkov-flexcompute: fixed as suggested

dbochkov-flexcompute:
Rebased. If no other comments, I will go ahead and merge this

PR #1502, opened 2024-02-26
fix pec freq arg

dmarek-flex:
Small bug when trying to plot epsilon with coord_key not equals to `centers` and `freq=None`. Other media assume an infinite frequency but PEC results in TypeError at multiplication

`eps_sample = sim.epsilon(box=box, coord_key="Ex")`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix to 3D surface integration monitors with some surfaces completely outside of the simulation domain which would sometimes still record fields.
 - Better error handling if remote `ModeSolver` creation gives response of `None`.
 - Validates that certain incompatible material types do not have intersecting bounds.
+- Fixed handling of the `frequency` argument in PEC medium.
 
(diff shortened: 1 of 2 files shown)


tylerflex:
Thanks @dmarek-flex, appreciate you submitting the fix along with the bug report.

Maybe let's try to merge this into `develop` just to get it out faster.
Also, can you add a line to the CHANGELOG.md under ### fixed?

Thanks!

weiliangjin2021:
I wonder why not add the decorator `@ensure_freq_in_range` as in other materials?

weiliangjin2021:
> Maybe let's try to merge this into `develop` just to get it out faster. Also, can you add a line to the CHANGELOG.md under ### fixed?

If merged to develop, we'll need to merge develop to pre/2.7 since the development for RF is on pre/2.7 branch?

tylerflex:
we could also just apply this commit to both branches potentially. Or it's ok if we keep merging this into pre/2.7 for simplicity.

dmarek-flex:
> I wonder why not add the decorator `@ensure_freq_in_range` as in other materials?

Yes, I can do that and it would fix it as well.

PR #1503, opened 2024-02-27
FEAT: Custom documentation and linked github actions

daquinteroflex:
- [X] Ghastly implementation for custom method/attribute documentation on API pages given a lack of alternatives.
- [X] Fixes the favicon issue
- [X] Updates the faq submodule to the correct pages

Based on [ this explanation](https://github.com/orgs/community/discussions/25288#discussioncomment-3247293), we need to merge this so that the linked actions get interconnected.


daquinteroflex:
Hello! @tylerflex  and @tomflexcompute ,

So you're going to hate this, possibly as much as I do, and I'm sorry but this is the best I could come up with. Alas, I've tried everything I could throw at it given the constraints.

**The situation is:** 
We want to provide users with custom docstring references. However, sphinx autosummary doesn't really work well with deeply nested methods and classes. Before, we used to have everything kind of being listed. Now, we have only specific attributes and methods specific to the class.

**How this patch works? Note it is not a solution**
- We have a custom `_custom_autosummary` directory that is version controlled. In it, we have a toctree like the example I provide in which we link to methods in parent classes that the child class can use. The link is created in the toctree, and will lead to the docstring reference of the parent class method.

**What I have tried?**
Everything I could think of internally and extenal of sphinx. Considered an extension again, customized autosummary as far as I could. The alternative is manually tracking and editing the _autosummary generated files which are not pretty. The problem is stub docstring generation in deep inheritance is very tricky.

What happens next?
- The issue is that when we migrate fully to pydantic v2, we will build upon it to have much nicer documentation. I would rather build upon that for our custom docs than patch this already existing patch haha.
- Also I would like to approach architecture 3.0 with less hierarchy which will solve many of these issues, and others that have not yet come up.

Further solutions if we really don't like this:
- Improve the source code of the autosummary extension so it can better handle deeply nested documentation, but this is very involved.
- Write our own extension, but also very involved.

image
image

tylerflex:
Thanks @daquinteroflex I think this solution seems reasonable. I guess we should talk a bit about how this would work and the use case though. I think there were varying opinions about how many of the parent attributes to document. Displaying them all was cluttering up the docs (especially with inherited pydantic or xarray attributes) but too few and we lose the important methods, such as `to_file` etc.

So in this solution, it seems like we only display the attributes and methods defined explicitly within the class? and ignore any of the inherited attributes and methods unless they are defined in the `_custom_autosummary`?

What I worry about is that we'd need to write in all of the methods (eg. `to_file`) for each class if we want them to show up?

In my personal opinion, the ideal solution would be something like like the following (sorry for not touching base on this earlier, just throwing this idea out there).

1. We should display all inherited attributes and methods defined in Tidy3DBaseModel or after. So anything that flexcompute has defined.
2. We should display a subset of inherited attributes and methods from pydantic / xarray that we can configure. For example, if they match a set of names that we pre-define (eg. "json()", "copy()", "interp()", "sel()"). 

The 1st ensures that the user knows of all methods and attributes that we intended them to be able to use.
The 2nd ensures that we don't show the user anything more than we think they need to know from the classes that came before Tidy3DBaseModel.

Hopefully this would balance us showing the user what they need to see without cluttering the docs, while making it pretty simple to configure without needing to copy everything over?

@momchil-flex might want to chime in too.

daquinteroflex:
It's quite a tricky one really. sphinx doesn't really support the type our deeply nested class API like we want it, so this implementation is really just patching up temporarily until we can migrate to `autodoc_pydantic` fully which will be easier to customize any type of behavior we want I think. It is really inflexible for what we want to do, even the above, because it can't distinguish levels of methods inheritance from the built-in functionality from what I can tell. 

I don't really like my implementation as I have done it, but without writing a custom extension for this use case, or modifying an existing one like `autodoc_pydantic`, I do not believe it is possible to achieve that specific requirement using standard sphinx. Also, all solutions are relatively involved, and if this documentation style is going to last until we do the full pydantic v2 migration - would it be worth doing that custom implementation now? It kind of feels like a catch22 in all cases.

What I did notice earlier is that it would be possible, for example, to create a table of contents with just the child class attributes/methods, but in the actual docstring representation, it can generate all the `inherited_methods` from the parents class. The idea is that we could provide a link to class-specific methods/attributes, but also show all the other methods the classes can access without cluttering the table of class API contents. Would we prefer this?

tylerflex:
> It is really inflexible for what we want to do, even the above, because it can't distinguish levels of methods inheritance from the built-in functionality from what I can tell.

I see. What about a list of "globally excluded" methods and attributes so we can just put anything we dont need from xarray or pydantic in here and just ignore them? And simply display everything else as before?

> API class page table of contents with just the child class attributes/methods, but in the actual rest of the page that contains all the methods, it can generate all the inherited_methods from the parents class without indexing them at the top which makes it more navigateable. The idea is that we could provide a link to child-class-specific methods/attributes, but also show all the other parent methods the child-classes can access without cluttering the table of child-class API contents. Would we prefer this?

I think I would personally prefer this, but I tend to lie more on the "show everything, even if not relevant" side of the spectrum here. Will let others chime in.

daquinteroflex:
> I see. What about a list of "globally excluded" methods and attributes so we can just put anything we dont need from xarray or pydantic in here and just ignore them? And simply display everything else as before?

I have just tested this and it works to remove specific methods using the :exclude-members: directive in the rst template file. I would just have to make the big list of things we don't want to represent, but it appears feasible to me. Maybe we can combine this with the index of child-specific attributes/methods and this may overall be helpful for people navigating possibly? I'll also wait for any further feedback too in terms of documentation preferences and explore how feasible what we want to do is within the existing configuration.

tylerflex:
For DataArray, we'd have to decide which of these we would want to exclude. I guess it is quite a lot :/ but we could leave some of the more useful ones in.  https://docs.flexcompute.com/projects/tidy3d/en/v2.5.2/_autosummary/tidy3d.ScalarFieldDataArray.html?highlight=ScalarFieldDataArray

daquinteroflex:
> It would be probably a long list but it would just need to be defined once right? (not for for each tidy3d component?). We can just look at some of our existing API references and decide what we want to remove, eg.

So I've just implemented something like this. What we can do is merge it into develop so it can be seen, and I can do any updates before merging into latest.

> For DataArray, we'd have to decide which of these we would want to exclud

I've removed all the private methods, and we can select the non private ones we want to exclude. I'm inclined to emrge into develop and update from there if this is the approach we want to go for.

tylerflex:
ok we can take a look, thanks dario.

daquinteroflex:
Unfortunately, I believe there is a bug between pydantic classes interacting with config annotations, and this sphinx issue that explores something similar https://github.com/sphinx-doc/sphinx/issues/11387 . I think this is the reason why the `:exclude-members:` directive is not fully excluding the list of inherited special attributes. 

I could explore automating the `_custom_autosummary` generation actually from parsing the inherited-attribute files manually, and then creating the autosummary from that - but the caveat is creating a proper link, as always, to the parent class that declared it. I feel because all the parent classes should be documented, standard python users should be able to know they can use these methods in child classes, and in a way they get familiar with relevant conceptual definitions and relationships. What we can do, in any case, is add a few more common usage examples within the class which use these inherited methods and illustrate these relationships - which I wanted to do too.

I know this solution is not great, but do think that it is "good enough" given the circumstances/constraints - and I hate using that term - so that maybe I can focus more on the pydantic v2 migration/docs direction when possible?

tylerflex:
I'm fine with this, thanks for looking into it.

daquinteroflex:
Available here https://docs.flexcompute.com/projects/tidy3d/en/develop/api/_autosummary/tidy3d.SimulationData.html#tidy3d.SimulationData

PR #1510, opened 2024-03-01
Directing users towards log if warnings found or simulation diverged or errored

### LINKED ISSUE #1483: Direct user towards `tidy3d.log` if there are warnings or errors
In 2.6 `tidy3d.log` has been refactored to include all warnings raised by simulation validators, as well as extra backend warnings if applicable, e.g. in the case of diverged simulations. We should:

- [ ] After simulation run, display a warning to the user to check the log if any WARNING messages are found in the log.
- [ ] Make sure this also works for a diverged case. Consider directly displaying the log if diverged.
- [ ] Download and display the log (or ask user to have a look at it) in the case of an erroring simulation. In some cases there could be an error that the user could act on.
- [ ] Add tidy3d.log to the divergence email @tomflexcompute

momchil-flex:
Fixes #1483, except the last point (add the tidy3d.log file to the divergence email).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `tidy3d.plugins.design.Results` store the `BatchData` for batch runs in the `.batch_data` field.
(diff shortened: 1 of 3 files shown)


tomflexcompute (approved):
Thanks. It's very nice.

tomflexcompute:
I'll communicate with MC to add the divergence info into automatic emails.

PR #1511, opened 2024-03-01
fix adjoint bug for point monitor data from jax update

tylerflex:
When jax is upgraded, the adjoint quickstart fails when constructing adjoint source because `complex()` doesn't work for multi-dimensional jax arrays (even if just one data point).

```py
>>> import jax.numpy as jnp
>>> a = jnp.array([[1]])

# jax == 0.4.4
>>> complex(a) 
(1+0j)

# jax == 0.4.24
>>> complex(a)
TypeError: Only scalar arrays can be converted to Python scalars; got arr.ndim=2

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Validates that certain incompatible material types do not have intersecting bounds.
 - Fixed handling of the `frequency` argument in PEC medium.
 - Corrected plotting cmap if `val='re'` passed to `SimulationData.plot_field`.
+- Bug when converting point `FieldMonitor` data to scalar amplitude for adjoint source in later jax versions.
 
 ## [2.6.0] - 2024-01-21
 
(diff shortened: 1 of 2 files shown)


tylerflex:
This might need to also be merged into 2.7

momchil-flex (approved):
Looks good, but why do we need to explicitly convert, do you know?

Up to you re 2.6/2.7

tylerflex:
we need to convert because we pass this value to our `GaussianPulse.amplitude`, I dont think it can handle an array.

I'll merge into develop, but maybe we should do a merge from develop -> pre/2.7 before we do any 2.7.0rc1 as there are a few patches already.

tylerflex:
sorry, I guess I misunderstood. we don't pass this value directly as amplitude is real. I'm not sure why we cast, but I'm sure there was a good reason for it otherwise I would have probably neglected to :)

tylerflex:
yea actually I think it's in fact just that we need to pass a scalar to the source time fields (phase and amplitude) and I didn't want to cast it to float of course. that must be why

PR #1517, opened 2024-03-05
Small reorganization of web.monitor logging

momchil-flex:
Tom pointed out that if a task is queued for a bit, some users don't realize what is happening, and terminate the notebook execution (while the task still runs eventually and users get charged). This is an attempt to improve our messaging. Here's what it looks like when a task is in queue now:

image


tomflexcompute (approved):
Thanks it's very helpful.

tylerflex (approved):
looks good, thanks

PR #1519, opened 2024-03-06
Improve voltage computation for lumped ports

dmarek-flex:
The E field is discontinuous at the port boundaries where conductors are present. In this picture the ground plane is at z=0 and the strip is at z=1000. Interpolation is inaccurate there, and even though only the end points are affected, it has a noticeable effect on accuracy.

image

Now, all field points outside the port bounds are ignored and linear extrapolation is used to estimate the E field on the exact port bounds.

### DIFF
--- a/tidy3d/plugins/smatrix/component_modelers/terminal.py
+++ b/tidy3d/plugins/smatrix/component_modelers/terminal.py
@@ -1,4 +1,5 @@
 """Tool for generating an S matrix automatically from a Tidy3d simulation and lumped port definitions."""
+
 from __future__ import annotations
 
 from typing import Tuple, Dict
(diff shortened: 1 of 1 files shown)


tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -193,16 +194,25 @@ def port_voltage(port: LumpedPort, sim_data: SimulationData) -> xr.DataArray:
             e_component = "xyz"[port.voltage_axis]
             field_data = sim_data[f"{port.name}_E{e_component}"]
             e_field = field_data.field_components[f"E{e_component}"]
+
+            # Remove E field outside the port region
+            e_field = e_field.sel(
+                {
+                    e_component: slice(
+                        port.bounds[0][port.voltage_axis], port.bounds[1][port.voltage_axis]
+                    )
```
weiliangjin2021, line 203: When you use `slice` here, is it exclusive on both start and end? My impression is that it is inclusive in the beginning.

tylerflex (approved):
Looks good to me, was going to say to include a changelog item but it probably isn't necessary since the original version of this hasn't been released yet, right?

dmarek-flex:
> Looks good to me, was going to say to include a changelog item but it probably isn't necessary since the original version of this hasn't been released yet, right?

Yea I was about to add something to the changelog and saw the lumped ports stuff right above, so I decided not to change anything.

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: Now I understand your point, I added which makes the excludes the boundaries. I didn't see a better way so I used `drop_sel`

weiliangjin2021 (approved):

PR #1520, opened 2024-03-06
Bugfixes related to simulation.subsection

dbochkov-flexcompute:
- Bug in `ModeSolver.reduced_simulation_copy` that was causing mode profiles to be all `NaN`.
- Bug in `Simulation.subsection()` that was causing zero-size dimensions not to be preserved.
- Behavior of `UniformGrid` is made consistent with `AutoGrid` in case of zero-size dimensions. That is, grid is created as `[center - dl / 2, center + dl / 2]` instead of `[center, center + dl]`

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -1563,6 +1563,18 @@ class Simulation(AbstractSimulation):
 
         return val
 
+    @pydantic.validator("grid_spec", always=True)
+    @skip_if_fields_missing(["size"])
+    def _no_custom_grid_along_zero_dims(cls, val, values):
+        """Check that wavelength can be defined if there is auto grid spec."""
+
+        size = values["size"]
+
+        for s, grid_spec_axis in zip(size, [val.grid_x, val.grid_y, val.grid_z]):
+            if s == 0 and isinstance(grid_spec_axis, CustomGrid):
+                raise SetupError("'CustomGrid' is not supported along zero-size dimensions.")
+        return val
+
     """ Post-init validators """
 
     def _post_init_validators(self) -> None:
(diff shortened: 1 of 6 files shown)


tidy3d/components/simulation.py
```diff
@@ -1563,6 +1563,18 @@ def _check_normalize_index(cls, val, values):
 
         return val
 
+    @pydantic.validator("grid_spec", always=True)
+    @skip_if_fields_missing(["size"])
+    def _no_custom_grid_along_zero_dims(cls, val, values):
+        """Check that wavelength can be defined if there is auto grid spec."""
```
momchil-flex, line 1569: Docstring needs and update

momchil-flex (approved):
Looks good, just one small comment, but to be safe I'll also run the backend tests again after merging.

dbochkov-flexcompute:
fixed the docstrings + also realized that when requesting a subsection with zero-size dimensions (i.e. get a 2d simulation out of 3d sim) we should make sure that along zero-size dimensions:
- the resulting subsection does not get accidentally expanded
- boundary conditions are automatically set to periodic

tidy3d/components/simulation.py
```diff
@@ -3787,6 +3782,18 @@ def subsection(
...
+        for dim in range(3):
+            if new_bounds[0][dim] == new_bounds[1][dim] and not isinstance(
+                boundary_spec.to_list[dim][0], Periodic
+            ):
+                axis_name = "xyz"[dim]
+                log.warning(
+                    f"The resulting simulation subsection has size zero along axis '{axis_name}'. "
+                    "Periodic boundary conditions are automatically set along this dimension."
+                )
+                boundary_spec = boundary_spec.updated_copy(**{"xyz"[dim]: Boundary.periodic()})
```
momchil-flex, line 3795: When can this happen?

momchil-flex: Ok never mind just saw your comment. :)

PR #1525, opened 2024-03-07
Creating parent directory for 'to_file' when downloading a gz file

momchil-flex:
In 2.6 solver output is gziped, and it turns out that if we have something like `to_file="data/simulation_data.hdf5"` in web.run and the `data` folder does not exist, this would error. This had already been taken care of in the regular file download, now also fixing it for the gziped one which uses a tempfile when calling `download_file`, so the parent directory is not created there.


tylerflex (approved):
changelog? I think this looks good but am not 100% sure, are there any potential issues? eg if `pathlib` version is out of date? if you feel pretty confident feel free to merge.

momchil-flex:
> changelog? I think this looks good but am not 100% sure, are there any potential issues? eg if `pathlib` version is out of date? if you feel pretty confident feel free to merge.

Not sure what you mean by that, but I just copied how we've been doing it so far in `download_file`, so I'm not really introducing anything new.

@lucas-flexcompute could you take a quick look?

lucas-flexcompute:
I see I'm late to the party, but it looks correct to me.

PR #1532, opened 2024-03-09
[draft for discussion] 2dmaterial

weiliangjin2021:


tidy3d/components/simulation.py
```diff
@@ -3638,10 +3640,12 @@ def snap_to_grid(geom: Geometry, axis: Axis) -> Geometry:
...
-                if structure.medium.is_pec:
-                    pec_plus = increment_float(snapped_center, 1.0)
-                    pec_minus = increment_float(snapped_center, -1.0)
-                    new_bounds = (pec_minus, pec_plus)
+                # new_bounds = ((1.0+100*fp_eps)*snapped_center - dls[0] / 2, (1.0-100*fp_eps)*snapped_center + dls[1] / 2)
+                # if structure.medium.is_pec:
+                # pec_plus = increment_float(snapped_center, 1.0)
+                # pec_minus = increment_float(snapped_center, -1.0)
+                # new_bounds = (pec_minus, pec_plus)
+                new_bounds = (snapped_center, snapped_center)
```
weiliangjin2021, line 3648: @dmarek-flex two questions:

- All 2d materials are treated as 0 thickness now?
- What's the difference does it make by turning off slight expansion?

dmarek-flex:
I made another attempt after trying a few different ideas. I found that the major issue is how the 2d volumetric equivalent material tends to overwrite the underlying background material's normal component. For conductor intersections this becomes a problem because the conductor that is added later will either extend the existing conductor or prematurely truncate it.

- I changed the behaviour of volumetric_equivalent to average the normal component as well, instead of simply choosing the positive side material's normal component.

- Now if a conductor is found on either side of the 2d material, the 2d material is given a 0 thickness.

tidy3d/components/simulation.py
```diff
@@ -3569,6 +3572,41 @@ def snap_to_grid(geom: Geometry, axis: Axis) -> Geometry:
             snapped_center = snap_coordinate_to_grid(self.grid, center, axis)
             return set_bounds(geom, (snapped_center, snapped_center), axis)
 
+        def is_normal_volumetric_equivalent_accurate(
```
dmarek-flex, line 3575: This was just a quick placeholder for checking if the materials are conductors which is based on some given value for conductivity.

tidy3d/components/medium.py
```diff
@@ -5162,8 +5168,8 @@ def get_background(comp: Axis) -> PoleResidue:
             meds = [get_component(med=med, comp=comp) for med in adjacent_media]
             # the Yee site for the E field in the normal direction is fully contained
             # in the medium on the + side
-            if comp == axis:
-                return meds[1]
+            # if comp == axis:
```
dmarek-flex, line 5171: By commented this out the normal component will also be computed as a weighted average

tidy3d/components/medium.py
```diff
@@ -5112,7 +5118,7 @@ def _weighted_avg(
                 return med
             else:
                 raise ValidationError("Invalid medium type for the components of 'Medium2D'.")
-            poles += [(a, weight * c) for (a, c) in pole_res.poles]
+            poles += [(a, weight * c) for (a, c) in pole_res.poles if c != 0.0]
```
dmarek-flex, line 5121: I also added this filtering behavior for the poles. Otherwise, a lot of poles are appended with c=0 get added which I think does nothing

PR #1534, opened 2024-03-12
FIX: Testing configuration :test_tube:

daquinteroflex:
Per @momchil-flex feedback on `test_local`


pytest.ini
```diff
@@ -1,7 +1,7 @@
 [pytest]
 addopts = --doctest-modules
 doctest_optionflags = NORMALIZE_WHITESPACE ELLIPSIS
-norecursedirs = tests/_test_local tests/test_cli tests/_test_data tests/full_test_notebooks tidy3d/web
+norecursedirs = tests/_test_local tests/test_cli tests/_test_data tests/_test_notebooks tidy3d/web
```
momchil-flex, line 4: I thought if the folder starts with `_` it will already ignore it?

daquinteroflex: So I modified the default behavior via the `pytest.ini`, so whenever the pytest is run from the frontend it will use this configuration to perform even the docstring tests which was reported not to properly work before this check.

daquinteroflex:
Since this is affecting PRs into develop would it be ok to merge? Maybe @tylerflex can approve as possibly awake at this time?

tylerflex:
I'm awake and it looks ok to me, but maybe let's hold off until momchil has a final look just to be sure. I don't use `test_local.sh` so I'm not sure how it's supposed to work these days.

momchil-flex:
Yeah looks good I'm merging.

PR #1537, opened 2024-03-12
FEAT: Urgent update ruff version fix :running:

daquinteroflex:
I was helping out @marc-flex and realised that the `ruff` version was not fixed. This caused a bit of mayhem in the linting, and is relatively urgent. I believe progressively everyone migrating to this version until we decide to change it will occur file by file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
       name: black <<< NOTE:, if you get a message stating that "All done! n file(s) reformatted", you must git add and commit the changes again to complete the commit. >>>
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.1.5"
+    rev: "v0.3.2"
     hooks:
      - id: ruff
        args: [ --fix ]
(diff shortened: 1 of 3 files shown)


tylerflex (approved):
Looks good. but just wondering if we need to also apply `ruff` to the code after this? I suppose that will cause a lot of merge conflicts potentially?

tylerflex:
At some point, we should apply to the entire codebase. I guess the question is when is the 'best' time to do that? maybe right before we do an official 2.7.0 release could be one option? Another would be to just do it now and ask everyone to update their ruff.  What do you think is best?

daquinteroflex:
It's kind of like when do we want to deal with this pain. It is most likely not at the end of a release, and I feel it'd be better towards the start of the release cycle. But I don't intentionally want to add extra suffering to people migrating really

tylerflex:
maybe we could apply it to develop immediately after `2.7.0` release? and before making `pre/2.8`? And also ask everyone to upgrade then? In the meantime just do the minimal changes (affecting the pre-commit)

daquinteroflex:
That makes sense! In any case, the pre-commit should make this progressive whenever people are using it.

daquinteroflex:
Will merge this in then, tomorrow morning

PR #1542, opened 2024-03-13
FEAT: Compute characteristic impedance from fields

dmarek-flex:
Addresses  [Issue 517](https://github.com/flexcompute/tidy3d-core/issues/517). Most of computations are done, now it is just a question of the best way to expose this functionality to users. I wonder if @weiliangjin2021 or @tylerflex have any input on whether we should start a "microwave" plugin that includes functionality that might not be commonly used in photonics. I mention that because it would be very easy to add time domain Voltage and Current monitors now.

Some tasks remaining:
- [x] Update smatrix plugin to reuse path integrals
- [x] Add tests

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -10,6 +10,7 @@ import xarray as xr
 from tidy3d.log import _get_level_int
 from tidy3d.web import BatchData
 from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d import ModeIndexDataArray
 
 """ utilities shared between all tests """
 np.random.seed(4)
@@ -78,7 +79,7 @@ def cartesian_to_unstructured(
     xyz = [array.x, array.y, array.z]
     lens = [len(coord) for coord in xyz]
 
-    num_len_zero = sum(l == 1 for l in lens)
+    num_len_zero = sum(length == 1 for length in lens)
 
     if num_len_zero == 1:
         normal_axis = lens.index(1)
@@ -245,7 +246,7 @@ def cartesian_to_unstructured(
 def make_spatial_data(
     size,
     bounds,
-    lims=[0, 1],
+    lims=(0, 1),
     seed_data=None,
     unstructured=False,
     perturbation=0.1,
@@ -875,6 +876,72 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
             **field_cmps,
         )
 
+    def make_field_time_data(monitor: td.FieldTimeMonitor) -> td.FieldTimeData:
+        """make a random FieldTimeData from a FieldTimeMonitor."""
+        field_cmps = {}
+        coords = {}
+        grid = simulation.discretize_monitor(monitor)
+        tmesh = simulation.tmesh
+        for field_name in monitor.fields:
+            spatial_coords_dict = grid[field_name].dict()
+
+            for axis, dim in enumerate("xyz"):
+                if monitor.size[axis] == 0:
+                    coords[dim] = [monitor.center[axis]]
+                else:
+                    coords[dim] = np.array(spatial_coords_dict[dim])
+
+            (idx_begin, idx_end) = monitor.time_inds(tmesh)
+            tcoords = tmesh[idx_begin:idx_end]
+            coords["t"] = tcoords
+            field_cmps[field_name] = make_data(
+                coords=coords, data_array_type=td.ScalarFieldTimeDataArray, is_complex=False
+            )
+
+        return td.FieldTimeData(
+            monitor=monitor,
+            symmetry=(0, 0, 0),
+            symmetry_center=simulation.center,
+            grid_expanded=grid,
+            **field_cmps,
+        )
+
+    def make_mode_solver_data(monitor: td.ModeSolverMonitor) -> td.ModeSolverData:
+        """make a random ModeSolverData from a ModeSolverMonitor."""
+        field_cmps = {}
+        coords = {}
+        grid = simulation.discretize_monitor(monitor)
+        index_coords = {}
+        index_coords["f"] = list(monitor.freqs)
+        index_coords["mode_index"] = np.arange(monitor.mode_spec.num_modes)
+        index_data_shape = (len(index_coords["f"]), len(index_coords["mode_index"]))
+        index_data = ModeIndexDataArray(
+            (1 + 1j) * np.random.random(index_data_shape), coords=index_coords
+        )
+        for field_name in ["Ex", "Ey", "Ez", "Hx", "Hy", "Hz"]:
+            spatial_coords_dict = grid[field_name].dict()
+
+            for axis, dim in enumerate("xyz"):
+                if monitor.size[axis] == 0:
+                    coords[dim] = [monitor.center[axis]]
+                else:
+                    coords[dim] = np.array(spatial_coords_dict[dim])
+
+            coords["f"] = list(monitor.freqs)
+            coords["mode_index"] = index_coords["mode_index"]
+            field_cmps[field_name] = make_data(
+                coords=coords, data_array_type=td.ScalarModeFieldDataArray, is_complex=True
+            )
+
+        return td.ModeSolverData(
+            monitor=monitor,
+            symmetry=(0, 0, 0),
+            symmetry_center=simulation.center,
+            grid_expanded=grid,
+            n_complex=index_data,
+            **field_cmps,
+        )
+
     def make_eps_data(monitor: td.PermittivityMonitor) -> td.PermittivityData:
         """make a random PermittivityData from a PermittivityMonitor."""
         field_mnt = td.FieldMonitor(**monitor.dict(exclude={"type", "fields"}))
@@ -920,6 +987,8 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
 
     MONITOR_MAKER_MAP = {
         td.FieldMonitor: make_field_data,
+        td.FieldTimeMonitor: make_field_time_data,
+        td.ModeSolverMonitor: make_mode_solver_data,
         td.ModeMonitor: make_mode_data,
         td.PermittivityMonitor: make_eps_data,
         td.DiffractionMonitor: make_diff_data,
--- a/tidy3d/plugins/microwave/impedance_calculator.py
+++ b/tidy3d/plugins/microwave/impedance_calculator.py
@@ -1,0 +1,75 @@
+"""Class for computing characteristic impedance of transmission lines."""
+
+from __future__ import annotations
+
+
+import pydantic.v1 as pd
+import numpy as np
+from typing import Optional
+from ...components.data.monitor_data import FieldData, FieldTimeData, ModeSolverData
+
+from ...components.base import Tidy3dBaseModel
+from ...exceptions import DataError, ValidationError
+
+from .path_integrals import VoltageIntegralAxisAligned, CurrentIntegralAxisAligned
+from .path_integrals import MonitorDataTypes, IntegralResultTypes
+
+
+class ImpedanceCalculator(Tidy3dBaseModel):
+    """Tool for computing the characteristic impedance of a transmission line."""
+
+    voltage_integral: Optional[VoltageIntegralAxisAligned] = pd.Field(
+        ...,
+        title="Voltage Integral",
+        description="Definition of path integral for computing voltage.",
+    )
+
+    current_integral: Optional[CurrentIntegralAxisAligned] = pd.Field(
+        ...,
+        title="Current Integral",
+        description="Definition of contour integral for computing current.",
+    )
+
+    def compute_impedance(self, em_field: MonitorDataTypes) -> IntegralResultTypes:
+        """Compute impedance for the supplied ``em_field`` using ``voltage_integral`` and
+        ``current_integral``. If only a single integral has been defined, impedance is
+        computed using the total flux in ``em_field``."""
+        if not isinstance(em_field, (FieldData, FieldTimeData, ModeSolverData)):
+            raise DataError("'em_field' type not supported by impedance calculator.")
+
+        # If both voltage and current integrals have been defined then impedance is computed directly
+        if self.voltage_integral:
+            voltage = self.voltage_integral.compute_voltage(em_field)
+        if self.current_integral:
+            current = self.current_integral.compute_current(em_field)
+
+        # If only one of the integrals has been provided then fall back to using total power (flux)
+        # with Ohm's law. The input field should cover an area large enough to render the flux
+        # computation accurate. If the input field is a time signal, then it is real and flux
+        # corresponds to the instantaneous power. Otherwise the input field is in frequency domain,
+        # where flux indicates the time-averaged power 0.5*Re(V*conj(I))
+        if not self.voltage_integral:
+            flux = em_field.flux
+            if isinstance(em_field, FieldTimeData):
+                voltage = flux / current
+            else:
+                voltage = 2 * flux / np.conj(current)
+        if not self.current_integral:
+            flux = em_field.flux
+            if isinstance(em_field, FieldTimeData):
+                current = flux / voltage
+            else:
+                current = np.conj(2 * flux / voltage)
+
+        impedance = voltage / current
+        return impedance
+
+    @pd.validator("current_integral", always=True)
+    def check_voltage_or_current(cls, val, values):
+        """Raise validation error if both ``voltage_integral`` and ``current_integral``
+        were not provided."""
+        if not values.get("voltage_integral") and not val:
+            raise ValidationError(
+                "Atleast one of 'voltage_integral' or 'current_integral' must be provided."
+            )
+        return val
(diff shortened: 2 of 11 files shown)


weiliangjin2021:
Agree that we should have a RF plugin. The other related issue is that we can have a wave port (which I think is the same as the current mode source, but slight difference in s-matrix parameter extraction?)

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,290 @@
...
+    )
+
+    snap_path_to_grid: bool = pd.Field(
+        False,
+        title="Snap path to grid",
+        description="It might be desireable to integrate exactly along the Yee grid associated with "
+        "a field. If enabled, the integration path will be snapped to the grid.",
+    )
+
+    def compute_integral(self, scalar_field):
```
weiliangjin2021, line 40: missing docstring

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,290 @@
...
+        coords_interp = np.concatenate((coords_interp, coordinates))
+        coords_interp = np.concatenate((coords_interp, [max_bound]))
+        coords_interp = {coord: coords_interp}
+        # Use extrapolation for the 2 additional endpoints
+        scalar_field = scalar_field.interp(
+            coords_interp, method="linear", kwargs={"fill_value": "extrapolate"}
+        )
+        return scalar_field.integrate(coord=coord)
+
+    def _get_field_along_path(
```
weiliangjin2021, line 71: missing docstring

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,290 @@
...
+
+class CurrentIntegralAA(Box):
+    """Class for computing conduction current via Ampere's Circuital Law on an axis-aligned loop."""
+
+    _plane_validator = assert_plane()
+
+    sign: Direction = pd.Field(
+        "+",
+        title="Direction of contour integral.",
+        description="Positive indicates V=Vb-Va where position b has a larger coordinate along the axis of integration.",
```
weiliangjin2021, line 159: Should the sign here for current instead?

tidy3d/plugins/microwave/impedance_calculator.py
```diff
@@ -0,0 +1,63 @@
...
+            current = self.current_integral.compute_current(em_field.field_components)
+
+        # If only one of the integrals has been provided then fall back to using total power (flux)
+        # with Ohm's law. The input field should cover an area large enough to render the flux computation accurate.
+        if not self.voltage_integral:
+            flux = em_field.flux
+            if isinstance(em_field, FieldTimeData):
+                voltage = flux / current
+            else:
+                voltage = 2 * flux / np.conj(current)
```
weiliangjin2021, line 46: is the difference of field being FieldTimeData or not coming from whether the field is real/complex?

weiliangjin2021 (commented):
Looks much cleaner now! In "path_integrals.py", there are duplicate codes in "normal_axis" and "remaining_axes". I wonder if you can define an abstract class that has those two methods, and then `AxisAlignedPathIntegral` and `CurrentIntegralAA` subclass from this class.

tidy3d/plugins/microwave/impedance_calculator.py
dmarek-flex: That seems like the same condition to me. The difference between the computations comes from either considering the instantaneous power (time-domain/real) vs  time-averaged power (freq domain/complex).

weiliangjin2021: I see. Will be good to add a explanation here.

weiliangjin2021 (approved):

dmarek-flex:
I updated the changelog for this PR, but I was a little confused by the state of the versioning. It seems like LumpedElements have been added in 2.6, but that is not the case.

weiliangjin2021:
> I updated the changelog for this PR, but I was a little confused by the state of the versioning. It seems like LumpedElements have been added in 2.6, but that is not the case.

Yeah, it's messed up. I suppose it'll be cleaned up upon release. For now, you can edit near the part that should go pre/2.7. E.g. the Added section should go pre/2.7; There are 2 Changed section, the former one should go pre/2.7.

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+from ...components.data.dataset import FieldDataset, FieldTimeDataset, ModeSolverDataset
+from ...components.data.data_array import AbstractSpatialDataArray
+from ...components.base import cached_property
+from ...components.types import Axis, Direction
+from ...components.geometry.base import Box
+from ...components.validators import assert_line, assert_plane
+from ...exceptions import DataError
+
+
+class AbstractAxesRH:
```
tylerflex: the convenetion is to inherit abstract classes from `abc.ABC`. I think basically this just makes it not possible to initialize this class, with a useful error message.

Also, mightt want to consider making it inherit from `Tidy3dBaseModel` as well?

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+    """Represents an axis-aligned right-handed coordinate system with one axis preferred."""
+
+    @cached_property
+    def main_axis(self) -> Axis:
```
tylerflex, line 33: make this an `@abc.abstractmethod` too, to enforce that subclasses implement this method.

Then, the `raise NotImplementedError` is not needed,

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+class AxisAlignedPathIntegral(AbstractAxesRH, Box):
+    """Class for defining the simplest type of path integral which is aligned with Cartesian axes."""
+
+    _line_validator = assert_line()
+
+    extrapolate_to_endpoints: bool = pd.Field(
+        False,
+        title="Extrapolate to endpoints",
```
tylerflex: use title case for `title`, here and in other `Field`s

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+        description="If the endpoints of the path integral terminate at or near a material interface, "
+        "the field is likely discontinuous. This option ignores fields outside and on the bounds "
+        "of the integral. Should be turned on when computing voltage between two conductors. ",
```
tylerflex: remove trailing space.

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+class AxisAlignedPathIntegral(AbstractAxesRH, Box):
+    """Class for defining the simplest type of path integral which is aligned with Cartesian axes."""
+
+    _line_validator = assert_line()
+
+    extrapolate_to_endpoints: bool = pd.Field(
+        False,
+        title="Extrapolate to endpoints",
+        description="If the endpoints of the path integral terminate at or near a material interface, "
+        "the field is likely discontinuous. This option ignores fields outside and on the bounds "
```
tylerflex: Maybe instead of "This option", something like "When this field is ``True``" is more explicit

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+
+    snap_path_to_grid: bool = pd.Field(
+        False,
+        title="Snap path to grid",
+        description="It might be desireable to integrate exactly along the Yee grid associated with "
+        "a field. If enabled, the integration path will be snapped to the grid.",
+    )
+
+    def compute_integral(self, scalar_field: AbstractSpatialDataArray):
+        """Computes the defined integral given the input `scalar_field`."""
```
tylerflex: for docstrings, we need to use double backslash quotes for variables. Because it gets interpreted as restructured text (RST) in the docs. Be sure to fix if it comes up in other spots too

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+        if not scalar_field.does_cover(self.bounds):
+            raise DataError("scalar field does not cover the integration domain")
```
tylerflex: Capitalize first letter and add perioid

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+                coord1: self.center[axis1],
+                coord2: self.center[axis2],
+            }
+            # Select field nearest to center of integration line
+            scalar_field = scalar_field.sel(
+                remaining_coords,
+                method="nearest",
+                drop=False,
+            )
+        else:
```
tylerflex, line 130: I've noticed a lot of DataArray wrangling in your PRs :). just waned to ask if any of these operations can be captured in methods of our `DataArray` class to make life easier? Or if they are sufficiently specific to these operations that they make more sense just written out here like this?

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+                scalar_field = scalar_field.interp(
+                    coord2dict, method="linear", kwargs={"bounds_error": True}
+                )
+
+        return scalar_field
+
+    @cached_property
+    def main_axis(self) -> Axis:
+        """Axis for performing integration."""
+        val = next((index for index, value in enumerate(self.size) if value != 0), None)
```
tylerflex: what does this do exactly? looks like it maybe takes the first index where size != 0 and if it doesn't find any it returns None?

if so, might be more readable to do something like
```py
for index, value in enumerate(self.size):
    if value != 0:
        return index
return None

```

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+        return val
+
+
+class VoltageIntegralAA(AxisAlignedPathIntegral):
```
tylerflex: I'm not 100% sold on `AA` in the class name. Do you think it would be very clear to the user that this means `AxisAligned`? Would `VoltageIntegralAxisAligned` be too long of a class name?

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+    """Class for computing the voltage between two points defined by an axis-aligned line."""
+
+    sign: Direction = pd.Field(
+        "+",
+        title="Direction of path integral.",
+        description="Positive indicates V=Vb-Va where position b has a larger coordinate along the axis of integration.",
+    )
+
+    def compute_voltage(self, em_field: FieldDataset | ModeSolverDataset | FieldTimeDataset):
```
tylerflex: I might be wrong but `|` might not work in some versions of python, so might want to double check the tests and `ruff`. It's also possible this doesn't render well in our docs, so just be aware of that. 

One other option is to define
```
EMFieldTypes = Union[FieldDataset | ModeSolverDataset | FieldTimeDataset]
```
at the top of the file and use it for type annotations.

tylerflex: what is the return type here? (and add in other places too)

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+        """Axis for performing integration."""
+        val = next((index for index, value in enumerate(self.size) if value != 0), None)
+        return val
+
+
+class VoltageIntegralAA(AxisAlignedPathIntegral):
+    """Class for computing the voltage between two points defined by an axis-aligned line."""
+
+    sign: Direction = pd.Field(
+        "+",
```
tylerflex: do we want this to be "+" by default? or could this lead to unexpected errors if users don't realize they might need to set it?

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+        title="Direction of path integral.",
```
tylerflex: no periods in the titles. I should edit the style guide so it's more clear, my bad.

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+        description="Positive indicates V=Vb-Va where position b has a larger coordinate along the axis of integration.",
+    )
+
+    def compute_voltage(self, em_field: FieldDataset | ModeSolverDataset | FieldTimeDataset):
+        """Compute voltage along path defined by a line."""
+        e_component = "xyz"[self.main_axis]
+        field_name = f"E{e_component}"
+        # Validate that the field is present
+        if field_name not in em_field:
+            raise DataError(f"field_name '{field_name}' not found")
```
tylerflex: `field_name` in the string is a variable so it renders better in single quotes.
```py
f"'field_name' '{field_name}' not found"
```

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+        e_field = em_field[f"E{e_component}"]
```
tylerflex: can use `field_name` to index this right?

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+        integral = self.compute_integral(e_field)
+
+        voltage = -integral
```
tylerflex: kind of nit picking here but why not juts replacing these with
```py
if self.sign == "+":
    voltage *= -1
```
instead of this double negative? or am I missing something?

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+    sign: Direction = pd.Field(
+        "+",
+        title="Direction of contour integral",
+        description="Positive indicates current flowing in the positive normal axis direction.",
+    )
+
+    extrapolate_to_endpoints: bool = pd.Field(
+        False,
+        title="Extrapolate to endpoints",
+        description="This parameter is passed to `AxisAlignedPathIntegral` objects when computing the contour integral.",
```
tylerflex: double check the double backslash for RST (same comment)

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -0,0 +1,299 @@
...
+        current = current.drop_vars((h_component, v_component))
+        return current
+
+    @cached_property
+    def main_axis(self) -> Axis:
+        """Axis normal to loop"""
+        val = next((index for index, value in enumerate(self.size) if value == 0), None)
+        return val
+
+    def _to_path_integrals(self, h_horizontal, h_vertical) -> Tuple[AxisAlignedPathIntegral, ...]:
```
tylerflex, line 251: add docstring

tylerflex (changes requested):
Thanks @dmarek-flex ! Looks really good overall. This will be a great addition.

My comments are pretty minor but apply a bit globally at times so just make sure you address them not only where I commented. eg. the quotes, capitalization, and other formatting. 

I do wonder if we need to build some more advanced `DataArray` handling for your applications into our main package, let me know if you have any thoughts on that.

tylerflex:
@dmarek-flex will there be a notebook tutorial to go along with this?

dmarek-flex:
> @dmarek-flex will there be a notebook tutorial to go along with this?

Yes, that would be a good idea. I have added an [issue](https://github.com/flexcompute/tidy3d-notebooks/issues/96) for the planned microwave notebooks to help with tracking.

tidy3d/plugins/microwave/path_integrals.py
dmarek-flex: I agree, I have switched to the full expansion.

tidy3d/plugins/microwave/path_integrals.py
dmarek-flex: Yea I have become much better at understanding xarray! I think the operations that I have done here and in the lumped ports module are fairly unique. They are needed since usually there are only a handful of grid points used for ports, and there are some corner cases when there is only a single grid point.

At the moment I can't think of other use cases that would benefit from these operations being accessible directly from `DataArray`. I will keep it in mind though as we add more functionality.

tests/utils.py
```diff
@@ -875,6 +876,72 @@ def make_field_data(monitor: td.FieldMonitor) -> td.FieldData:
             **field_cmps,
         )
 
+    def make_field_time_data(monitor: td.FieldTimeMonitor) -> td.FieldTimeData:
```
dmarek-flex, line 879: @tylerflex Make sure to check these changes in case I missed something. I used these extensions to tests the different types of input data to the impedance calculator.

tylerflex: just looked it over and seems good to me.

tylerflex (approved):
Re: notebook will be good to have that as otherwise we've noticed users will seldom find these features on their own. Otherwise looks good to me!

PR #1545, opened 2024-03-14
Starting point for `SimulationSpec`

tylerflex:
Defines a possible starting point for how to implement the processing of a `Simulation` into a well-defined specification that can be used by the FDTD solver.


tylerflex:
Would it be helpful to keep this open or should we close it just to keep the PR list less cluttered? @daquinteroflex

tylerflex:
If you think it's useful to keep open, feel free to re-open, closing for now.

daquinteroflex:
Yep makes sense to close for now till we work out the plan in more detail

PR #1550, opened 2024-03-15
add tracking of `simulation_type` for smatrix and design plugins

tylerflex:
@momchil-flex @lei-flex 

I need to get statistics about how many tasks are run from `adjoint`, `design` and `smatrix` plugins.

One simple way is to pass the plugin name through `simulation_type`, similar to how we do it for `adjoint`.

I'm proposing a change here that passes `simulation_type` for batches run in `design` and `smatrix`. and then in the backend we can treat these as regular `"tidy3d"` simulation types?

Let me know if you have a better solution. 

One thing that could make this easier is to do eg. `"tidy3d_design"` `"tidy3d_smatrix"` and then in the backend, we do the normal pipeline `if "tidy3d" in simulation_type`

let me know what you think

### DIFF
--- a/tidy3d/plugins/design/method.py
+++ b/tidy3d/plugins/design/method.py
@@ -87,7 +87,7 @@ class MethodIndependent(Method, ABC):
         self, simulations: Dict[str, Simulation], path_dir: str = None, **kwargs
     ) -> BatchData:
         """Create a batch of simulations and run it. Mainly separated out for ease of testing."""
-        batch = web.Batch(simulations=simulations, **kwargs)
+        batch = web.Batch(simulations=simulations, simulation_type="tidy3d_design", **kwargs)
         if path_dir:
             kwargs["path_dir"] = path_dir
         return batch.run(**kwargs)
--- a/tidy3d/plugins/smatrix/smatrix.py
+++ b/tidy3d/plugins/smatrix/smatrix.py
@@ -350,6 +350,7 @@ class ComponentModeler(Tidy3dBaseModel):
             folder_name=self.folder_name,
             callback_url=self.callback_url,
             verbose=self.verbose,
+            simulation_type="tidy3d_smatrix",
         )
 
     @cached_property
--- a/tidy3d/web/api/webapi.py
+++ b/tidy3d/web/api/webapi.py
@@ -47,6 +47,7 @@ def run(
     progress_callback_download: Callable[[float], None] = None,
     solver_version: str = None,
     worker_group: str = None,
+    simulation_type: str = "tidy3d",
 ) -> SimulationDataType:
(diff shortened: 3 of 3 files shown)


momchil-flex:
> I need to get statistics about how many tasks are run from `adjoint`, `design` and `smatrix` plugins.
> 
> One simple way is to pass the plugin name through `simulation_type`, similar to how we do it for `adjoint`.
> 
> I'm proposing a change here that passes `simulation_type` for batches run in `design` and `smatrix`. and then in the backend we can treat these as regular `"tidy3d"` simulation types?

This will already work right off the bat on the backend. Basically everywhere we do something special, we only do it `if "adjoint" in simulation_type`.

The main thing I'm wondering is if this is the way we want to go. At some point I was thinking about moving *away* from `simulation_type` as a database variable - I was thinking that when the metadata PR #1352 is in, we could refactor things to put both the jax_vjp and the adjoint simulation type there. However, there is a merit to writing something directly to the database. One question is, do we still want this to be a kwarg eveywhere though (now also added to `run` and `upload`), or is it better to indeed transition to putting this in the simulation metadata, and only making the backend write that value to the database?

tylerflex:
> This will already work right off the bat on the backend. Basically everywhere we do something special, we only do it if "adjoint" in simulation_type.

Yea I ran a quick test with `simulation_type="test"` and it worked fine.

> The main thing I'm wondering is if this is the way we want to go. At some point I was thinking about moving away from simulation_type as a database variable - I was thinking that when the metadata PR https://github.com/flexcompute/tidy3d/issues/1352 is in, we could refactor things to put both the jax_vjp and the adjoint simulation type there. 

I'd like to avoid using the metadata for this. my thinking was I'd refactor #1551  to upload the `JaxSimulation` directly. And we can just use the `type` field to tag whether it's adjoint or not?

> However, there is a merit to writing something directly to the database. One question is, do we still want this to be a kwarg eveywhere though (now also added to run and upload), or is it better to indeed transition to putting this in the simulation metadata, and only making the backend write that value to the database?

I don't mind modifying this PR to not add the option in the `run` call. It was already a kwarg in `upload` I think. For this PR I only need to add it to the `Batch` constructor, where it's already accepted.


My thinking would be if this is functionality is already supported through `simulation_task` and doesn't break anything, maybe we can just merge this for now and then separately think about how we want to track simulation metadata in a more comprehensive way in another PR?

momchil-flex:
Sure, sounds good. Could you set the `docs/notebooks` submodule to the latest `develop` state if it's not? (currently conflicting).

tylerflex:
should be good to go.

momchil-flex:
Seems like a black issue now

Issue #1549, opened 2024-03-15
Adjoint Refactor Phase 2


tylerflex:
## Why?
- [ ] Fix issues where `JaxX` methods inherited from `td.X` components ignore `input_structures`, `output_monitors` eg. #1459 
- [ ] Simplify internal handling of Jax-traced structures and monitors
    - [ ] simpler accounting for how many of each type of monitor, structure.
    - [ ] simplify conversion from `JaxSimulation` to `Simulation`

## Goals
- [ ] `JaxSimulation` methods inherited from `Simulation` work without patching (eg `.plot()`)
- [ ] make web API support `JaxSimulation` upload directly.
- [ ] eliminate need for separate `JaxInfo` file upload.

## Strategy
- [ ] Validate `input_structures` to make a `td.Structure` copy and append it to `JaxSimulation.structures`. 
- [ ] Validate `output_monitors` to make a `td.Monitor` copy and append it to `JaxSimulation.monitors`. 
- [ ] Similarly, add `MonitorData` from `output_monitors` to the `JaxSimulationData.data` tuple so it's accessible there.
- [ ] Remove patching for `JaxSimulationMethods`
- [ ] Simplify `JaxSimulation.to_simulation()` (just remove extraneous fields eg. `output_monitors` and `input_structures` and type-cast).
- [ ] Simplify conversion from `JaxSimulationData` to adjoint `JaxSimulation`.
- [ ] Make `JaxSimulation` upload able through web API, simplify the `tda.web` internals.
- [ ] Remove `JaxInfo` from code, if possible.

## If possible.
- [ ] Remove need for separate `JaxStructureStaticX`, allow mixing of jax and regular types.

tylerflex:
Fixed by #1697 . we'll be using native autograd support now.

PR #1551, opened 2024-03-15
adjoint refactor 2.7

tylerflex:
#1549


momchil-flex:
I think this is unlikely to be 2.7 right?

tylerflex:
Pretty unlikely. I keep having ideas and trying different things and nothing is working quite well yet. I think if something does happen it will probably be in a different PR since this idea kind of fizzled out as I got into the details.

PR #1555, opened 2024-03-15
Zero thickness for volumetric equivalence of 2D material

weiliangjin2021:
Different subclass for volumeric equivalence of Medium2D

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -4477,6 +4477,12 @@ class AnisotropicMedium(AbstractMedium):
         return self.updated_copy(**dict(zip(["xx", "yy", "zz"], new_comps)))
 
 
+class AnisotropicMediumFromMedium2D(AnisotropicMedium):
+    """The same as ``AnisotropicMedium``, but converted from Medium2D.
+    (This class is for internal use only)
+    """
+
+
 class FullyAnisotropicMedium(AbstractMedium):
     """Fully anisotropic medium including all 9 components of the permittivity and conductivity
     tensors.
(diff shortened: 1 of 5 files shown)


tidy3d/components/medium.py
```diff
@@ -4287,6 +4287,12 @@ def sel_inside(self, bounds: Bound):
         return self.updated_copy(**dict(zip(["xx", "yy", "zz"], new_comps)))
 
 
+class AnisotropicMediumFromMedium2D(AnisotropicMedium):
+    """The same as ``AnisotropoicMeidum``, but converted from Medium2D.
```
dmarek-flex: typo "Anisotropic Medium"

dmarek-flex:
Looks good, but I think I will wait to approve until I try it with the backend changes.

tidy3d/components/simulation.py
```diff
@@ -3638,10 +3638,7 @@ def snap_to_grid(geom: Geometry, axis: Axis) -> Geometry:
                 temp_geometry = set_bounds(snapped_geometry, bounds=new_bounds, axis=axis)
```
weiliangjin2021, line 3638: I'm not sure about this line (and the line above):
```
new_bounds = (snapped_center - dls[0] / 2, snapped_center + dls[1] / 2)
temp_geometry = set_bounds(snapped_geometry, bounds=new_bounds, axis=axis)
```
The bound here is still of dl-thickness. Is that the right approach @dmarek-flex ?

dmarek-flex: Good point.
Now that we have special treatment for 2d materials I don't think we need to include the volumetric equivalent in the list of structures that is passed to the `subdivide` function. So we don't need the `temp_geometry` or `background_structures_temp`. new_bounds should start as (center,center)

tylerflex (approved):
Looks good. just a few comments
* Do we want to allow `AnisotropicMediumFromMedium2D` to be imported form `tidy3d`? if so, import it into `__init__.py`.
* Please add changelog line, if you think it warrants it.
Thanks!

dmarek-flex (approved):

caseyflex (approved):

PR #1558, opened 2024-03-18
Fix: port computations for coarse grids

dmarek-flex:
While testing a new example I found some issues with the way LumpedPorts are handled.

- Small bug in port snapping that was missed when changing interface, now have improved test coverage.
- For grids that do not line up exactly with LumpedPort bounds, the computation for current might choose the wrong bounding point.
- For very coarse grids there might not be a single data point within the port bounds, so I added a method that checks this condition before the simulations are launched.

I did not modify the changelog, since these are all changes that could be considered within the original LumpedPorts PR.

### DIFF
--- a/tidy3d/plugins/smatrix/component_modelers/terminal.py
+++ b/tidy3d/plugins/smatrix/component_modelers/terminal.py
@@ -9,6 +9,7 @@ import numpy as np
 import xarray as xr
 
 from ....constants import C_0, fp_eps
+from ....exceptions import SetupError
 from ....components.simulation import Simulation
 from ....components.geometry.utils_2d import snap_coordinate_to_grid
 from ....components.data.sim_data import SimulationData
@@ -148,7 +149,7 @@ class TerminalComponentModeler(AbstractComponentModeler):
                 list(self.simulation.monitors) + port_voltage_monitors + port_current_monitors
             )
             for port, center in zip(self.ports, snap_centers):
-                port_source = self.to_source(self._source_time, port=port, snap_center=center)
+                port_source = port.to_source(self._source_time, snap_center=center)
                 update_dict = dict(
                     sources=[port_source],
                     monitors=new_mnts,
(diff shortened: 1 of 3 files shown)


tylerflex (approved):

dmarek-flex:
BTW, keep in mind that the current and voltage computations are about to be changed in #1542, so they will be much simpler.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -263,13 +276,21 @@ def port_current(port: LumpedPort, sim_data: SimulationData) -> xr.DataArray:
             h1_field = h_field.isel({orth_component: orth_index - 1})
             h2_field = h_field.isel({orth_component: orth_index})
             h_field = h1_field - h2_field
+
+            # Find exact bounds of port taking into consideration the Yee grid
```
weiliangjin2021, line 280: a bit more explanation?

weiliangjin2021 (approved):

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: I need to modify this computation anyways in #1542, and it should be much simpler.

weiliangjin2021: Is this PR closed in favor of #1542. It seems the same change happening there already?

dmarek-flex: No I was planning on merging this PR soon, because it fixes some issue with coarse grids. Part of the issue was in this current computation. There is just a bit of overlap with  #1542. I think the code in #1542 will be much more clear with the usage of the path integrals.

weiliangjin2021: sounds good. feel free to merge.

PR #1560, opened 2024-03-19
Inverse Design Plugin

tylerflex:


lucas-flexcompute (approved):
I don't see anything that looks out of place here. The user interface is pretty great! Definitely easier to use than going trough adjoint directly.

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -0,0 +1,175 @@
...
+            # update optimizer and parameters
+            updates, opt_state = optax_optimizer.update(-grad, opt_state, params)
+            params = optax.apply_updates(params, updates)
+
+            # cap the parameters
+            params = jnp.minimum(params, 1.0)
+            params = jnp.maximum(params, 0.0)
+
+            history["params"].append(params)
+            history["opt_state"].append(opt_state)
```
momchil-flex: I always found this tricky previously when we wrote similar optimization history scripts. Here it seems like the params that you write do not match all the other values that you have written in history, i.e. the grad, objective function, etc. were computed for the params *before* the update, while here you write them to history *after* the update. You can certainly move `history["params"].append(params)` to be above with the other history updates but I think there *was* some reason to put it later, maybe related to what happens when you continue an optimization? I remember I thought about this once and it was not as straightforward as one would expect.

momchil-flex (commented):
Overall looks great, should be very useful!

I see you've added the invdes.png image here and also in the notebooks PR, and it may not be used here?

One thing I noticed when I worked with the similar history handling that we have in our notebooks is that in some cases (esp. if I e.g. push things to large design regions like the 12x14um grating coupler I optimized) the memory needed to store the history can be quite large. This can make both the file large so it is inconvenient to handle (e.g. if you want to keep a bunch of optimization files they can take up a lot of space), and there's even a risk of running our of cpu memory completely since you store the whole history specifically for `grad`, `params`, and also the `opt_state` which itself is of a similar or larger size to `params`. Along those lines I feel like some more control over what exactly is stored, and some good default settings, may provide for a better experience?

For example, I probably always want to store the history of the `objective_fn_val`, `penalty`, and `post_process_val`. In most cases I probably just want the last `params` and `opt_state` so that I can examine my last design and continue an optimization. I probably almost never actually need to look at the `grad`. Now, I realize that using these as defaults (do not store history for `params`, `grad`, `opt_state`) limits functionality a bit (e.g. if I want to make a gif of the params evolution, or if I want to start from a state that's not the latest one). But if we allow the option to enable those, do you think it may be better?

tidy3d/plugins/invdes/optimizer.py
tylerflex: So what I did is when the optimizer is run from an empty state `Optimizer.run(params)`, it will store the initial `params` and `opt_state` before running this loop. 

https://github.com/flexcompute/tidy3d/blob/18449686c9fe520138170792b2b108684d4eb9d8/tidy3d/plugins/invdes/optimizer.py#L76

So actually for `num_steps` steps, we end up having `num_steps+1` values for `params` and `opt_state` in the history but only `num_steps` values for the other stuff.

It's not always clear whether people want the final state / parameters to be **after** the final update or what. But I decided to include it here just for completeness. And also so that if we continue a run it will not waste that gradient calculation.

tylerflex:
> I see you've added the invdes.png image here and also in the notebooks PR, and it may not be used here?

One is used for the notebook itself, and the other is used in the README.md for `invdes`, but I suppose I could try making them point to the same place.

> One thing I noticed when I worked with the similar history handling that we have in our notebooks is that in some cases (esp. if I e.g. push things to large design regions like the 12x14um grating coupler I optimized) the memory needed to store the history can be quite large. This can make both the file large so it is inconvenient to handle (e.g. if you want to keep a bunch of optimization files they can take up a lot of space), and there's even a risk of running our of cpu memory completely since you store the whole history specifically for grad, params, and also the opt_state which itself is of a similar or larger size to params. 

A few things:
1. The history saving is optional (in case that wasn't apparent). so I suppose users could turn this off if they don't want it.
2. Without the `opt_state` or `params`, the history will not be able to be continued, so I feel it should probably be included, but I was considering adding a few options to further customize the optimization loop, which I'll explain below.
3. We still restrict the number of parameters in the custom medium to 500,000 I think, so my feeling is it shouldn't be too big of an issue? Correct me if I'm wrong though. Maybe the issue iwas more storing he whole `SimulationData` object? which we did in some notebooks but I decided not to store here.

> Along those lines I feel like some more control over what exactly is stored, and some good default settings, may provide for a better experience?

I think changing the fields in the `InverseDesignResults` will lead to a bunch of annoying edge cases internally, but I'm considering the following two things:
1. add a `callback` function option to the `Optimizer`, so the user could in principle do something like
```py
history = dict(params=[], ...)
def callback(current_results, step_index: int) -> None:
    history['params'].append(params)

optimizer = Optimizer(..., callback=callback)
optimizer.run(params0)

history['params'] 
```
2. Allowing the user to specify a `post_process_fn` that returns `aux_data`, so they could even store the eg. `SimulationData` as 
```py
def postprocess(sim_data, ...):
    ...
    return objective, dict(data=sim_data)

history = dict(data=[])
def callback(..., aux_data):
    history['data'].append(aux_data['data'])
```



> For example, I probably always want to store the history of the objective_fn_val, penalty, and post_process_val. In most cases I probably just want the last params and opt_state so that I can examine my last design and continue an optimization. I probably almost never actually need to look at the grad. Now, I realize that using these as defaults (do not store history for params, grad, opt_state) limits functionality a bit (e.g. if I want to make a gif of the params evolution, or if I want to start from a state that's not the latest one). But if we allow the option to enable those, do you think it may be better?

I think the callback / aux_data could be good enough to support this, but the learning curve is a bit high. What do you think?

tidy3d/plugins/invdes/optimizer.py
momchil-flex: Yeah I think maybe not wasting the gradient computation is what was tripping me up before too. But if I understand correctly this can definitely be confusing now? First of all having `history` contain lists of different lengths is already not great. But probably the biggest problem is that if someone does `result.get_last("objective_fn_val")`, this value will not correspond to the simulation I get from `result.get_sim()`, if I then do a manual run on this and compute the objective. This seems like something that would happen often.

momchil-flex:
> > One thing I noticed when I worked with the similar history handling that we have in our notebooks is that in some cases (esp. if I e.g. push things to large design regions like the 12x14um grating coupler I optimized) the memory needed to store the history can be quite large. This can make both the file large so it is inconvenient to handle (e.g. if you want to keep a bunch of optimization files they can take up a lot of space), and there's even a risk of running our of cpu memory completely since you store the whole history specifically for grad, params, and also the opt_state which itself is of a similar or larger size to params.
> 
> A few things:
> 
> 1. The history saving is optional (in case that wasn't apparent). so I suppose users could turn this off if they don't want it.

You mean the history saving to file? That's true, but it's nice to have. And apart from that it's always stored in memory currently.

> 2. Without the `opt_state` or `params`, the history will not be able to be continued, so I feel it should probably be included, but I was considering adding a few options to further customize the optimization loop, which I'll explain below.

Yeah but you only need the last value, not the whole history. My suggestion was to store only the last value by default.

> 3. We still restrict the number of parameters in the custom medium to 500,000 I think, so my feeling is it shouldn't be too big of an issue? Correct me if I'm wrong though. Maybe the issue iwas more storing he whole `SimulationData` object? which we did in some notebooks but I decided not to store here.

Storing `SimulationData` could also be an issue, but that's not what I was thinking. But yeah let's try to estimate this. Let's say you have 500k params and those, and you have those in params, grad, and I think my observation was that opt_state contains at least another 2x. So say you have 2M double-precision floats at every iteration, and you do 100 iterations. That's 1.6GB. Not terrible, but also not that great.

> > Along those lines I feel like some more control over what exactly is stored, and some good default settings, may provide for a better experience?
> 
> I think changing the fields in the `InverseDesignResults` will lead to a bunch of annoying edge cases internally, but I'm considering the following two things:
> 
> 1. add a `callback` function option to the `Optimizer`, so the user could in principle do something like
> 
> ```python
> history = dict(params=[], ...)
> def callback(current_results, step_index: int) -> None:
>     history['params'].append(params)
> 
> optimizer = Optimizer(..., callback=callback)
> optimizer.run(params0)
> 
> history['params'] 
> ```
> 
> 2. Allowing the user to specify a `post_process_fn` that returns `aux_data`, so they could even store the eg. `SimulationData` as
> 
> ```python
> def postprocess(sim_data, ...):
>     ...
>     return objective, dict(data=sim_data)
> 
> history = dict(data=[])
> def callback(..., aux_data):
>     history['data'].append(aux_data['data'])
> ```
> 
> > For example, I probably always want to store the history of the objective_fn_val, penalty, and post_process_val. In most cases I probably just want the last params and opt_state so that I can examine my last design and continue an optimization. I probably almost never actually need to look at the grad. Now, I realize that using these as defaults (do not store history for params, grad, opt_state) limits functionality a bit (e.g. if I want to make a gif of the params evolution, or if I want to start from a state that's not the latest one). But if we allow the option to enable those, do you think it may be better?
> 
> I think the callback / aux_data could be good enough to support this, but the learning curve is a bit high. What do you think?

This does get more complicated. Maybe for starters we leave as is. But I was more driving towards something like an e.g. `store_complete_history` flat set to `False` by default, in which case store only the history of the objective function and penalty, while only the last values for grad, params, and opt_state. Or it could even be `True` by default. Not sure how much internal complexity this adds though.

tylerflex:
@momchil-flex sounds good for all points, I'll play around with a `store_full_history : bool = True` field and see how it works internally. 

I think more complicated callbacks can be added later, including for more customized history storage, display, and parameter scheduling.

tylerflex:
@momchil-flex I implemented something in this [commit](https://github.com/flexcompute/tidy3d/pull/1560/commits/634c5cf91037e9e3dbb894c7e95a9fffa52f08c2). Could you take a look when convenient? thanks.

tidy3d/plugins/invdes/README.md
```diff
@@ -0,0 +1,545 @@
...
+
+The `InverseDesign` object can be exported to a `td.Simulation` given some parameters using the `to_simulation(params)` method.
+
+Let's do this with a set of parameters all at 0.5 as it's a better starting point for this specific problem.
+
+```py
+
+params0 = 0.5 * np.ones_like(params0)
+```
+It can be useful to run the initial problem to ensure it looks correct before optimizing. For conveniencen, the `to_simulation_data()` method generates the simulation and runs it through `web.run()` to return the `SimulationData`, which can be visualized.```
```
yaugenst: conveniencen -> convenience

tidy3d/plugins/invdes/README.md
```diff
@@ -0,0 +1,545 @@
...
+]
+
+f, axes = plt.subplots(1,3,figsize=(10,4), tight_layout=True)
+for ax, sim in zip(axes, simulations):
+    sim.plot_eps(z=0, ax=ax)
+plt.show()
+
+```
+
+We also need to construct a post-processing function that will tell the optmizer how to post process the data from each of the `simulations` into a single objective.
```
yaugenst: optmizer -> optimizer

tidy3d/plugins/invdes/base.py
```diff
@@ -0,0 +1,86 @@
...
+    def _make_np_array(arr: typing.Any) -> np.ndarray:
+        """Turn any array into a ``numpy`` array."""
+        jax_arr = jnp.array(arr)
+        untraced_arr = jax.lax.stop_gradient(jax_arr)
+        return np.array(untraced_arr)
+
+    @staticmethod
+    def _check_fname(fname: str) -> None:
+        """Warn if fname isn't pickle."""
+        if all(ext not in fname for ext in ("pkl", "pickle", "dill")):
```
yaugenst: I guess this might lead to unintentional errors if any part of the extension is contained within the filename (and the file does not have the correct format) or if the extension is capitalized (which happens on Windows way more often than you might expect). So perhaps `if not any(os.path.splitext(fname).lower == ext for ext in ("pkl", "pickle", "dill"))`, or alternatively using pathlib's `Path.suffix()`.

tidy3d/plugins/invdes/design.py
```diff
@@ -0,0 +1,314 @@
...
+        title="Task Verbosity",
+        description="If ``True``, will print the regular output from ``web`` functions.",
+    )
+
+    @property
+    @abc.abstractmethod
+    def objective_fn(self) -> typing.Callable[[jnp.ndarray], float]:
+        """construct the objective function for this ``InverseDesign`` object."""
+
+    def _add_verbosity_to_kwargs(self, **kwargs) -> dict:
```
yaugenst: Nitpick but this can be done via `kwargs.setdefault("verbose", self.verbose)`. Would probably eliminate the need for this function entirely.

tidy3d/plugins/invdes/design.py
yaugenst, line 1: Looks great overall and definitely makes the adjoint plugin _much_ more user-friendly. I would have two high-level comments here:

 1. Penalties seem to be geared towards being... well, penalties, i.e., having a term in the objective function. It would be nice to have built-in support for constraints, as these are very commonly used (with MMA, for example). To have minimal support for this, the objective function's `aux_data` could return both the penalty value as well as its gradient, and the user could then plug these into their constrained optimizer of choice. This would mean that one needs access to all penalties separately (currently they are being summed over).
 2. It's not obvious how one would do robust optimization. The penalties on the design region work well for imposing certain feature sizes, but for robust optimization one typically wants to evaluate the objective for multiple design variations. I think this is possible using `InverseDesignMulti`, but might not be obvious to users, especially seeing that the documentation makes it seem like this is intended for running optimizations with multiple sources. It might be enough to clarify this in the documentation (and perhaps some examples in a notebook?).

Both of these features can be added later, though, and don't require a fundamental change to the plugin, so they're probably fine for now.

tidy3d/plugins/invdes/function.py
```diff
@@ -0,0 +1,78 @@
...
+
+import typing
+
+import tidy3d as td
+import tidy3d.plugins.adjoint as tda
+import jax.numpy as jnp
+
+
+class Utilities:
+    """Funcional utilities that help define postprocessing functions more simply in ``invdes``."""
```
yaugenst: Funcional -> Functional

tidy3d/plugins/invdes/function.py
```diff
@@ -0,0 +1,78 @@
+# post processing function utilities
+
+import typing
+
+import tidy3d as td
+import tidy3d.plugins.adjoint as tda
+import jax.numpy as jnp
+
+
+class Utilities:
```
yaugenst: Is there a reason why "functional" utilities are contained within a class? As there is no state, having these as functions in a module might be cleaner perhaps?

tidy3d/plugins/invdes/function.py
```diff
@@ -0,0 +1,78 @@
...
+        return jnp.array(arr)
+
+    @staticmethod
+    def get_amps(sim_data: tda.JaxSimulationData, monitor_name: str, **sel_kwargs) -> jnp.ndarray:
+        """Grab amplitudes from a ``ModeMonitorData`` and select out values."""
+
+        monitor_data = sim_data[monitor_name]
+
+        if not isinstance(monitor_data, td.ModeData):
+            raise ValueError("'grab_amps' only works with data from 'ModeMonitor's.")
```
yaugenst: Function is called `get_amps`, I guess the error message should also refer to `get_amps` (not `grab_amps`).

tidy3d/plugins/invdes/function.py
```diff
@@ -0,0 +1,78 @@
...
+        monitor_name: str,
+        field_component: td.components.types.EMField,
+        **sel_kwargs
+    ) -> jnp.ndarray:
+        """Grab field component from a ``FieldMonitorData`` and select out values."""
+
+        monitor_data = sim_data[monitor_name]
+
+        if not isinstance(monitor_data, td.FieldData):
+            raise ValueError("'grab_field_component' only works with data from 'FieldMonitor's.")
```
yaugenst: `grab_field_component` -> `get_field_component`?

tidy3d/plugins/invdes/function.py
```diff
@@ -0,0 +1,78 @@
...
+        """Sum values in the ``tda.JaxDataArray``."""
+
+        arr = cls.make_array(arr)
+        return jnp.sum(arr)
+
+    @classmethod
+    def sum_abs_squared(cls, arr: tda.JaxDataArray) -> float:
+        """Sum the absolute value squared of a ``tda.JaxDataArray``."""
+        arr = cls.make_array(arr)
+        arr_abs_squared = abs(arr) ** 2
```
yaugenst: Purely nitpicky opinion but I prefer `jnp.abs()` as it's more explicit. Completely identical though so feel free to ignore.

tidy3d/plugins/invdes/optimizer.py
yaugenst, line 1: Fine as-is, I'm just wondering, since `AdamOptimizer` is really just `optax.adam`, maybe there is a way to just support all `optax` optimizers out of the box without specifically enumerating all of them? I guess the `optax` people have already spent a lot of effort figuring out the optimizer API so besides some bookkeeping tasks here might as well reuse that? Perhaps a generic `OptaxOptimizer` class where one can just name the specific optimizer to use?
I'm mostly thinking about this in terms of future optimizer support for libraries other than Optax. Should tidy3d support only specific named optimizers or interface directly with a few optimization libraries and re-use their API?

tidy3d/plugins/invdes/region.py
```diff
@@ -0,0 +1,264 @@
...
+        # rmin, rmax = np.array(self.geometry.bounds)
+        # lengths = rmax - rmin
+        side_lengths = np.array(self.size)
+        num_pixels = np.ceil(side_lengths / self.pixel_size)
+        # TODO: if the structure is infinite but the simulation is finite, need reduced bounds
+        num_pixels[np.isinf(num_pixels)] = 1
+        return tuple(int(n) for n in num_pixels)
+
+    @property
+    def params_random(self) -> np.ndarray:
```
yaugenst, line 173: I think it's good to have default initialization helpers. However, by far the most common (and also usually best) way to initialize design regions is uniform 0.5 (due to the projection), and this is also used in the tutorial notebook. Since the invdes plugin is geared toward ease of use, I think constructions like `.params_ones / 2` should be avoided as this requires the user to already have that knowledge (which then renders pre-made initialization a bit pointless). Not really sure how to solve this elegantly though. Maybe a `.params_default` property? And for uniform initialization without noise, maybe instead of a property something like `.params_uniform(x)` where `x` is the value that the array should be initialized to?

tidy3d/plugins/invdes/transformation.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class FilterProject(InvdesBaseModel):
+    """Transformation involving convolution by a conic filter followed by a ``tanh`` projection.
+
+    .. image:: ../../_static/img/filter_project.png
+
+    """
+
+    radius: float = pd.Field(
```
yaugenst, line 26: Wouldn't it be more intuitive to have the user provide the physical dimensions and have the parametrization take care of the `sqrt(3)` factor internally?

tidy3d/plugins/invdes/transformation.py
```diff
@@ -0,0 +1,73 @@
...
+        """Get the conic filter associated with this transformation (after supplying grid size)."""
+        return tda_filter.ConicFilter(radius=self.radius, design_region_dl=design_region_dl)
+
+    def to_projector(self) -> tda_filter.BinaryProjector:
+        """Get the binary projector associated with this transformation."""
+        return tda_filter.BinaryProjector(
+            beta=self.beta, eta=self.eta, strict_binarize=self.strict_binarize, vmin=0, vmax=1
+        )
+
+    def evaluate(self, spatial_data: jnp.ndarray, design_region_dl: float) -> jnp.ndarray:
```
yaugenst, line 67: Fine as-is, but I'm wondering if it would make sense to have the `evaluate` function output in the range [0, 1]. This would make it possible to concatenate a bunch of different parameterizations (or transformations as they are called here). So, having an arbitrary series of transformations and only converting to physical materials when they go into the simulation.

tidy3d/plugins/invdes/base.py
tylerflex: Good idea, changed to
```py
suffix = pathlib.Path(fname).suffix
if not any(suffix.lower == ext for ext in ("pkl", "pickle", "dill")):
```

tidy3d/plugins/invdes/design.py
tylerflex: No, that's a good idea. I wasn't aware of this, but that's much better. (might be some other parts of the code that could use it too). Changed it.

tidy3d/plugins/invdes/design.py
tylerflex: I added these two points to the "`invdes` enhancements" [issue](https://github.com/flexcompute/tidy3d/issues/1581) for now. We'll be rolling this out in 2.7 prerelease so there might be some time to add them either now or in future releases. I'll gives some thought to them before it is released. And maybe when you join, you could take a stab at some of these improvements too.

tidy3d/plugins/invdes/function.py
tylerflex: Yea so originally I had these as individual functions, but it becomes a bit annoying to manage:
* The documentation needs to add all of the functions one by one, they aren't really organized.
* They aren't inside of a contained namespace (like `invdes.Utils`) so it kind of clutters the `invdes` namespace.

It seemed nicer to just contain them inside of a class, but I'm not sure, maybe we can just use the module itself as the namespace and just do `from invdes.functions import _`.

Do you have any strong opinions either way? I'm not sold on a particular approach, but do think grouping them in a namespace is generally nicer than free floating functions.

tylerflex: I don't really know what's better between doing it through class methods or just through a module. Note a similar discussion #1616 @dmarek-flex where I recommended class methods. Maybe we can think / discuss pros and cons a bit here

tidy3d/plugins/invdes/function.py
tylerflex: Ah, good catch, thanks. Fixed this and the field component one.

tidy3d/plugins/invdes/function.py
tylerflex: That's fine with me.

tidy3d/plugins/invdes/optimizer.py
tylerflex: Yea this is definitely a good thing to discuss. I would love to just be able to wrap any generic `optax` optimizer as that would be a more general and cleaner solution.

However, we ultimately want to be able to upload the entire `Optimizer` and `Design` to our web API to do this computation server side, so it needs to be serialized. Also, we can't really pickle things because it's not safe to unpickle data server side. So in some sense, I was stuck with defining a pydantic model that can generate the proper `optax` optimizer.

We also want to allow users to eventually set up these optimizers on the GUI, so we need some Pydantic-based interface to optax there I think.

Maybe we can have some kind of hybrid solution where you can use our `AdamOptimizer` or the base `Optimizer`, which can be initialized `.from_optax(optax_optimizer)` but only used client-side?

Any ideas? I'll also put this on the "future enhancements" list just to keep track of it.

tidy3d/plugins/invdes/region.py
tylerflex: Hm, the issue with uniform 0.5 params without any randomness added in is that sometimes the symmetry in the design region can lead to a gradient of exactly 0. This is true for the mode converter example, for example. There's essentially 0 conversion from fundamental to higher order mode due to the structure symmetry along y, so you need to break that symmetry initially to get any optimization rolling.

Besides this concern, I'm kind of ok with setting a default param if we can find a universally good one, but I do worry about this edge case as it comes up frequently enough to cause problems.

(I think if we do a default, w just have a property `. params_default` like you suggest and make `params=None` an optional kwarg to `Optimizer.run()` should work fine).

tidy3d/plugins/invdes/transformation.py
tylerflex: Yea in retrospect, this might have been better. I think we discussed it at some point internally.

This transformation is strongly related to the built-in [adjoint one](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/plugins/adjoint/utils/filter.py#L22-L33). So some of the reasoning was to not create two projection functions with different conventions. We can't really change that one due to backwards compatibility, so I wanted to just make them consistent.

That being said, it turns out (in my observation) that the feature sizes are much more influenced by the `length_scale` of the dilation-erosion penalty. The radius of the filter is almost more to "help" move the optimization towards larger feature sizes, but the value itself doesn't have much of an effect on things. Would be interesting to get a better perspective on this though if you have any thoughts.



So one line of

tidy3d/plugins/invdes/transformation.py
tylerflex: As far as I'm aware, these `evaluate()` functions (for transformations) all output in the range (0,1) (assuming the domain is in (0,1)). are you suggesting making this explicit? ie capping the functions using `jnp.clip`?

yaugenst: Uhm yeah, I think I got confused somewhere, maybe looking at the wrong file, but this is fine.

tidy3d/plugins/invdes/function.py
yaugenst: I can't really comment regarding the documentation since I'm not too familiar with how that is handled here.
From a purely prgramming-related perspective though, a class (to me) is something that bundles code and state/data, where ideally the code doesn't "make sense" without the data. So the purist in me would say that even `staticmethod`s are a code smell (not to say they aren't useful sometimes).
If it's just about namespacing, then I think modules work fine? That depends mostly on how it's handled in `__init__.py`. I see that currently the whole module is imported into the `invdes` namespace via `from .function import Utilities`. If this is changed to just `from . import function`, then the functions in that module would be callable (after importing `invdes`) as, e.g., `function.make_array()`, so there would still be a separation of namespaces if so desired. Should rename `functions.py` to `utilities.py` then, probably.
Not a strong opinion, but an opinion 😅

tidy3d/plugins/invdes/optimizer.py
yaugenst: > However, we ultimately want to be able to upload the entire Optimizer and Design to our web API to do this computation server side, so it needs to be serialized. Also, we can't really pickle things because it's not safe to unpickle data server side. So in some sense, I was stuck with defining a pydantic model that can generate the proper optax optimizer.

Ok I see, I wasn't thinking about that and that makes sense. Server-side optimization is great, and will save a lot of time (besides being the only option for GUI). Means that (at least a minimal version of) their API needs to be replicated though.

I think it might still be possible by using strings as the optimizer names (can be extracted automatically) and supplying the proper kwargs, kind of like scipy's minimize `method` and `options` parameters. These things should all be serializable. To get "nicer" code completion one could also use an enum instead of strings. The only issue I see with that approach is that it might become a bit tricky if there are certain optimizers that require special handling of the state dict (for restarting). Simplest initial solution could be to just exclude those though.

tidy3d/plugins/invdes/region.py
yaugenst: Right, that's a thing that happens (I haven't had to deal with it too much). In that case, though, I think the randomness should be as small as possible so there is just enough scattering to lead to a measurable gradient. It shouldn't be in [0, 1] since you run into the problem of having values close to 0 or 1 that are almost like dead pixels in the optimization (depending on the strength of the projection), which then also leads to suboptimal results.

As I'm writing this, I'm wondering if it would make sense to add a check that warns the user if the norm of the gradient is very close to zero after the first adjoint run, and suggest adding some randomness to the initialization? The problem is that this is an edge case that does happen, and it is not intuitive as a beginner to realize what's going on. But adding noise as a default is also not great because it leads to worse performing devices on average vs. uniform initialization (I tested this some time ago).

tidy3d/plugins/invdes/transformation.py
yaugenst: > So one line of

Guess something is missing here? 😄 

> The radius of the filter is almost more to "help" move the optimization towards larger feature sizes, but the value itself doesn't have much of an effect on things. Would be interesting to get a better perspective on this though if you have any thoughts.

Yeah that's true, although it also depends on the specific filter. Gaussian and cone definitely, but a uniform kernel is somewhat stricter (although the factor doesn't apply there so maybe that's a moot point). In general one can also apply erosion/dilation and derived morphological operations directly in the parametrization (not as a penalty), which can also lead to strict feature sizes. It's a bit tricky to decide which method to pick as a default, as it always depends on the user's requirements, so perhaps it is best to stick to what's found most commonly in literature for now, as that will be what people are familiar with.

Regarding the `radius` argument though, perhaps it would be best to be explicit about this in the documentation, that it doesn't really enforce feature sizes? As you said, it does to some extent, but in my opinion it really is just a regularization technique for the optimization (although people don't usually frame it that way).

tidy3d/plugins/invdes/function.py
dmarek-flex: In my case, I have a lot of supporting functions that users should not need to use, so I would like to hide them somewhat. I guess that could be accomplished by modifying my `__init__.py` to be more selective. In C++, I would have probably used a namespace instead of class as well, so maybe that is a clue!

dmarek-flex: I think I am tilting towards the module/namespace approach now. I wasn't aware of how to do that in Python, but it seems pretty straightforward.

tylerflex: If we went with module, we could just keep the private functions with a leading underscore and not imported in `__init__.py`. I also am starting to come around to this idea. my main concern is just wondering if it will be straightforward for the user to explore which functions are available in the module. I guess if, eg. `import invdes.utils as utils`, they could call `dir(utils)`, which doesn't look good, but apparently `help(utils)` might be formatted quite well for us. Perhaps this is fine. Maybe I'll give it a try.

tidy3d/plugins/invdes/optimizer.py
tylerflex: > Means that (at least a minimal version of) their API needs to be replicated though.

Right, that's basically the idea behind implementing the Adam one here as a pydantic model.

> I think it might still be possible by using strings as the optimizer names (can be extracted automatically) and supplying the proper kwargs, kind of like scipy's minimize method and options parameters. These things should all be serializable. To get "nicer" code completion one could also use an enum instead of strings. 

Hm, this is an interesting idea too. So something like `Optimizer(type="Adam", beta1=0.99)`?  I think the issue is still potentially how to serialize this. Right now for us to conveniently handle data through the web, it makes most sense to have it implemented as a `Tidy3DBaseModel` subclass, with well defined fields. I'm not quite sure how to handle these extra kwargs dynamically. There are some ways to generate pydantic models dynamically, but it seems maybe more complicated than I'd like for this design 🤔

tidy3d/plugins/invdes/region.py
tylerflex: >  I think the randomness should be as small as possible so there is just enough scattering to lead to a measurable gradient

This could work, but the magnitude of the random variation is too small, (in my experience) the gradient can be extremely close to 0 and the optimizer will have trouble starting.

> I'm wondering if it would make sense to add a check that warns the user if the norm of the gradient is very close to zero after the first adjoint run

I think this could be a good feature to add, yea. Maybe I'll give this a try with initial parameters of 0.5 and then we can set the initial parameters with random variation in our notebook examples, when needed.

tidy3d/plugins/invdes/transformation.py
tylerflex: > In general one can also apply erosion/dilation and derived morphological operations directly in the parametrization (not as a penalty), which can also lead to strict feature sizes.

Yes, we use the filter for our erosion and dilation operations actually.

>  perhaps it is best to stick to what's found most commonly in literature for now, as that will be what people are familiar with.

The literature (at least O Sigmund papers) tend to use a radius R for the "filter" radius I believe. So that's the same as the `radius` field. At least last I remember.

> Regarding the radius argument though, perhaps it would be best to be explicit about this in the documentation, that it doesn't really enforce feature sizes? As you said, it does to some extent, but in my opinion it really is just a regularization technique for the optimization (although people don't usually frame it that way).

Agree, probably we can be more descriptive here. I'll add that.

tidy3d/plugins/invdes/optimizer.py
yaugenst: I was thinking something along the lines of:
```python
Optimizer(type="Adam", options: None | dict[str, float | bool | <whatever types need to be supported>] = {"learning_rate": 1e-3, "b1": 0.1, ...etc...})
```

The downside of this is that the `options` dict is somewhat opaque, since what arguments it accepts depends on the optimizers, and one would need to refer to the Optax documentation to get that information. However, if this is documented properly, it can be a nice way of interfacing to all optimizers. One could also think about extracting e.g. the `learning_rate` explicitly like
```python
Optimizer(type="Adam", learning_rate=1e-3, extra_options = {"b1": 0.1, ...etc...})
```
since this is shared across optimizers, and the other parameters don't necessarily need to be tweaked (particularly for Adam).

If all arguments for the optimizers should be made explicit then yeah I agree, I can't really think of approaches that are not either pretty hacky or involve spelling everything out. Well I guess one other approach could be to support all arguments for all Optax optimizers explicitly (there aren't actually that many different ones) and have the unused ones not do anything (depending on the optimizer). But that also doesn't feel very clean...

tidy3d/plugins/invdes/function.py
tylerflex: Went with a module approach. Seems fine to me, will keep it. Thanks for the suggestions @yaugenst.

tidy3d/plugins/invdes/optimizer.py
tylerflex: hm, I'm inclined to punt this towards "future enhancements" and wait, maybe until someone actually requests other optimizers. LBFGS is one that we might want to include eventually but I haven't found one that works well with our `adjoint` plugin. Like you said in another thread, we should consider adding constraints as well, perhaps we can look into `nlopt` integration.

tidy3d/plugins/invdes/region.py
tylerflex: I changed thins to make `params0` optional in `Optimizer.run()` and use `0.5 * np.ones(...)` as a default.

Also added a warning in the optimizer if the gradient is ever `np.allclose(gradient, 0.0)`

```
                  "All elements of the gradient are almost zero. This likely indicates "
                    "a problem with the optimization set up. This can occur if the symmetry of the "
                    "simulation and design region are preventing any data to be recorded in the "
                    "'output_monitors'. In this case, we recommend creating the initial parameters "
                    " as 'params0 = DesignRegion.params_random' and passing this to "
                    "'Optimizer.run()' to break the symmetry in the design region. "
                    "This zero gradient can also occur if the objective function return value does "
                    "not have a contribution from the input arguments. We recommend carefully "
                    "inspecting your objective function to ensure that the variables passed to the "
                    "function are contributing to the return value."
```

PR #1565, opened 2024-03-21
fix tool.ruff.lint warnings and add numpy2 check

tylerflex:


daquinteroflex (approved):
Nice, yeah this was needed!

PR #1570, opened 2024-03-26
warn instead of error if task already submitted

tylerflex:
seems to work fine this way, and running `web.start(x)` twice doesn't error anymore.

There is some output (eg the error message) that I'm not sure how to clear though.

### DIFF
--- a/tidy3d/web/api/webapi.py
+++ b/tidy3d/web/api/webapi.py
@@ -19,6 +19,7 @@ from ..core.environment import Env
 from ..core.constants import SIM_FILE_HDF5, TaskId
 from ..core.task_core import SimulationTask, Folder
 from ..core.task_info import TaskInfo, ChargeType
+from ..core.exceptions import WebError as WebErrorCore
 from ...components.types import Literal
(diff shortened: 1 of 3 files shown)


tidy3d/web/api/webapi.py
```diff
@@ -273,10 +274,15 @@ def start(
...
-    )
+
+    # note: this try block breaks if task has already been submitted, bypasses error.
+    try:
+        task.submit(
+            solver_version=solver_version,
+            worker_group=worker_group,
+        )
+    except WebErrorCore:
+        log.warning("Task has already been started, skipping.")
```
lei-flex, line 285: I think it is not correct fix. 
the code need check the return code and then propose the correct message to final user.
Let's talk to @QimingFlex

lei-flex (changes requested):

tidy3d/web/api/webapi.py
tylerflex: agree

tylerflex:
Fixed by MC in backend

PR #1571, opened 2024-03-27
add unstructured datasets in docs

dbochkov-flexcompute:
also reformatted similarly to the analogous fdtd page


daquinteroflex (approved):
Looks good! Just needs a submodule update by the looks of it.

PR #1573, opened 2024-03-27
fix logger assignment in web core

dbochkov-flexcompute:
From documentation of `logging.basicConfig()` it looks like it gets automatically invoked whenever `logging` is used first time in the code. After that calling `logging.basicConfig()` will not have effect unless flag `force=True` is used.

Apparently, in our `web.core` code we substitute `logging` with tidy3d's `log` a bit too late, some of imports already call it. Rearranging order of imports and logger assignment resolves the issue.

### DIFF
--- a/tidy3d/web/__init__.py
+++ b/tidy3d/web/__init__.py
@@ -1,8 +1,15 @@
 """ imports interfaces for interacting with server """
-from .api.container import Job, Batch, BatchData
-from .cli.migrate import migrate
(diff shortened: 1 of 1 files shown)


tomflexcompute (approved):
Thanks @dbochkov-flexcompute for the fix!

lucas-flexcompute (approved):
Thanks @dbochkov-flexcompute !

PR #1577, opened 2024-04-01
adjoint run_local disables restrictive validators

### LINKED ISSUE #1575: Disable some `adjoint` validators when using `tda.web.run_local`.
## Problem: 

we have some validators (max number structures, max number pixels) that are there because the post processing time is very long and it burns server time. However, some users want to be able to go above these limits either way and are probably ok with the higher data download and run time requirements.

## Solution:

Adjust these validators to only happen at `tda.web.run()` time. The error message can instruct users to use `tda.web.run_local()` to run the gradient computation locally, but to expect more data download and a longer processing time (a few minutes or more.)

tylerflex:
Moves some validators to methods implemented in `JaxObject._validate_web_adjoint`, which only get called in `tda.web.run()` and `tda.web.run_async` calls. Error message includes instruction to try `tda.web.run_local` or `tda.web.run_async_local`.

Fixes #1574 and #1575

### DIFF
--- a/tidy3d/plugins/adjoint/components/simulation.py
+++ b/tidy3d/plugins/adjoint/components/simulation.py
@@ -20,7 +20,7 @@ from ....components.types import Ax, annotate_type
 from ....constants import HERTZ, SECOND
 from ....exceptions import AdjointError
 
-from .base import JaxObject
+from .base import JaxObject, WEB_ADJOINT_MESSAGE
 from .structure import (
     JaxStructure,
     JaxStructureType,
@@ -177,18 +177,6 @@ class JaxSimulation(Simulation, JaxObject):
         return val
 
     @pd.validator("input_structures", always=True)
-    def _restrict_input_structures(cls, val):
-        """Restrict number of input structures."""
-        num_input_structures = len(val)
-        if num_input_structures > MAX_NUM_INPUT_STRUCTURES:
-            raise AdjointError(
-                "For performance, adjoint plugin restricts the number of input structures to "
-                f"{MAX_NUM_INPUT_STRUCTURES}. Found {num_input_structures}."
-            )
-
-        return val
-
-    @pd.validator("input_structures", always=True)
     @skip_if_fields_missing(["structures"])
     def _warn_overlap(cls, val, values):
         """Print appropriate warning if structures intersect in ways that cause gradient error."""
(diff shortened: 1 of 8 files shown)


tylerflex:
@tomflexcompute (since you get a lot of adjoint requests..) just so you know, this PR will lift some of the size restrictions for the users if they use `adjoint.web.run_local` instead of `adjoint.web.run()`. This does the post processing locally, and at their own "expense" and potential risk of the processing taking a long time (several minutes or more).

tomflexcompute:
> @tomflexcompute (since you get a lot of adjoint requests..) just so you know, this PR will lift some of the size restrictions for the users if they use `adjoint.web.run_local` instead of `adjoint.web.run()`. This does the post processing locally, and at their own "expense" and potential risk of the processing taking a long time (several minutes or more).

That's great to know! Thanks.

tidy3d/plugins/adjoint/components/simulation.py
```diff
@@ -281,6 +269,30 @@ def _warn_nonlinear_input_structure(cls, val):
...
+                f"{MAX_NUM_INPUT_STRUCTURES}. Found {num_input_structures}. "
+                "You can still run this simulation through "
+                "'tidy3d.plugins.adjoint.web.run_local' or 'tidy3d.plugins.adjoint.web.run_local' "
+                ", which are similar to 'run' / 'run_async', but "
+                "perform the gradient postprocessing calculation locally after the simulation runs."
+                "Note that the postprocessing time can become "
+                "quite long (several minutes or more) if these restrictions are exceeded."
+                "Furthermore, the local versions of 'adjoint' require downloading field data "
+                "inside of the 'input_structures', which can greatly increase the size of data "
+                "needing to be downloaded."
```
momchil-flex: Put the part of this message which is common to all validators in a `RUN_LOCAL_MESSAGE` somewhere and then do `AdjointError("...case specific message..." + RUN_LOCAL_MESSAGE)`.

momchil-flex (commented):
Maybe worth to think already about what we should also do with these validators if we decide to loosen the restrictions but charge more. Should we warn about the extra charge? And should we lay the groundwork for this here already?

tylerflex:
> Maybe worth to think already about what we should also do with these validators if we decide to loosen the restrictions but charge more. Should we warn about the extra charge? And should we lay the groundwork for this here already?

I think, when the time comes, we can just change these exceptions to warnings and edit the message to just generally describe how much extra we'll charge and that the user can use `run_local` to avoid this if they are ok with the extra data download.

tidy3d/plugins/adjoint/components/simulation.py
tylerflex: sounds good, just implemented this in latest push.

momchil-flex (approved):

PR #1579, opened 2024-04-01
add new validators for heat simulation

dbochkov-flexcompute:
Added three new validators for heat simulations:
- do not let define a heat source with empty list of structures
- do not let define a heat simulation without containing at least one `SolidSpec` material (note, this is not full-proof validation, since we can have another `FluidSpec` structure to completely overlap any `SolidSpec` material)
- do not let define a heat simulation with only Neumann BC. This leads to an undefined solutions.

### DIFF
--- a/tests/test_components/test_heat.py
+++ b/tests/test_components/test_heat.py
@@ -226,7 +226,9 @@ def make_heat_source():
 
 
 def test_heat_source():
-    _ = make_heat_source()
+    source = make_heat_source()
+    with pytest.raises(pd.ValidationError):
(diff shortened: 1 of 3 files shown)


dbochkov-flexcompute:
fyi @marc-flex

momchil-flex (approved):

e-g-melo (approved):
Thanks @dbochkov-flexcompute! 

I didn't find other situations to validate so far.

dbochkov-flexcompute:
Just realized, we should also prohibit including 2D and 1D structures. Gonna push one more commit here

PR #1580, opened 2024-04-01
Improve mode solver PML

momchil-flex:


tidy3d/plugins/mode/derivatives.py
```diff
@@ -120,7 +126,30 @@ def create_s_matrices(omega, shape, npml, dlf, dlb, dmin_pml=(True, True)):
...
+
+        avg_xminus = np.mean(tensor_diag[:, : npml[0], :])
+        avg_xplus = np.mean(tensor_diag[:, Nx - npml[0] + 1 :, :])
+        avg_yminus = np.mean(tensor_diag[:, :, : npml[1]])
+        avg_yplus = np.mean(tensor_diag[:, :, Ny - npml[1] + 1 :])
+        return np.array([avg_xminus, avg_xplus, avg_yminus, avg_yplus])
+
+    eps_avg = pml_average_allsides(eps_tensor)
+    mu_avg = pml_average_allsides(mu_tensor)
+    return np.sqrt(eps_avg * mu_avg)
```
weiliangjin2021: I thought wave impedance is defined as sqrt(mu/eps)?

momchil-flex: Ah, good point, I think I was mislead by the fact that this factor multiplies `ETA_0`. I will rethink/rename things, thanks.

weiliangjin2021 (approved):

PR #1586, opened 2024-04-08
fix(): SCEM-5311 - Make Tidy3D plot_3d compatible with script limits

tim-goddard-flexcompute:
The Tidy3D notebooks environment is about to introduce a number of limits on script content in cell output as a security precaution. This conflicts with the way the `plot_3d` function operates, which is dependent on the ability to output JavaScript.

This change modifies the injected JavaScript to function whether the script is evaluated inline, or a similar copy is included in the environment. Our notebooks environment will add this script at the same time as introducing the change to strip script output from cells, preserving the ability to render 3d plots.

The script has been included inline, rather than from an external source, to most closely mimic the existing behavior, avoid adding an external script inclusion to customer environments.

### DIFF
--- a/tidy3d/components/viz.py
+++ b/tidy3d/components/viz.py
@@ -3,8 +3,7 @@ from __future__ import annotations
 
 from typing import Any
 from functools import wraps
-import random
-import time
+from html import escape
 
 import matplotlib.pyplot as plt
 from matplotlib.patches import PathPatch, ArrowStyle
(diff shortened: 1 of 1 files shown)


majinge7846 (approved):

tylerflex:
Thanks @tim-goddard-flexcompute, looks good to me. I'm just trying to figure out whether to release this as a patch 2.6.4 or whether it can wait a few weeks to release in our 2.7.0. Is it an urgent fix?

tylerflex (approved):

tylerflex:
Also, just wondering whether this has been tested in eg. a notebook or if that's still needed?

tylerflex:
FYI just tested on `StartHere.ipynb` and works as expected

image

Issue #1373, opened 2024-01-10
Homogeneous medium validators with custom medium


momchil-flex:
We have a [validator](https://github.com/flexcompute/tidy3d/blob/2d6bd787686bd25fd371143f176ec5a0baa35b8d/tidy3d/components/simulation.py#L1073-L1081) that is meant to validate that a diffraction monitor is in a homogeneous medium. However, what it actually does is compute the list of mediums that the plane intersects, and make sure that there's only one element in that list. This does not work for custom medium, which is a single medium which may nevertheless still be inhomogeneous physically. This same issue applies to the validator for plane wave/gaussian beam sources, as well as, in some sense, to the [TFSF validator](def _validate_tfsf_structure_intersections) about structure intersections. I am not sure what's a good way to fix this:

- Error if custom medium found
- Just warn for the user to make sure that it's homogeneous
- Actually sample the permittivity and check if it is homogeneous

tylerflex:
I think the warning approach might make the most sense to me, in the sense that it's the least restrictive and also fastest.
Maybe we just exclude any custom media from the list and warn?

momchil-flex:
In fact looking at the solver, TFSF where the source surface crosses a custom medium would straight up not work, or the results may be wrong, even if the custom medium is uniform on those surfaces. We should just validate that out.

For plane wave, diffraction monitor, we could just warn if in custom medium.

weiliangjin2021:
What about checking if custom medium is uniform, if not, warn that it should be uniform on the plane?

> In fact looking at the solver, TFSF where the source surface crosses a custom medium would straight up not work, or the results may be wrong, even if the custom medium is uniform on those surfaces.

@momchil-flex could you explain a bit more why it doesn't work even for a uniform custom medium?

momchil-flex:
> What about checking if custom medium is uniform, if not, warn that it should be uniform on the plane?

If we actually do the check, we should error if it's not uniform, not warn. But I wonder if the check may be expensive for e.g. a large plane, e.g. a large-area metalens?

> > In fact looking at the solver, TFSF where the source surface crosses a custom medium would straight up not work, or the results may be wrong, even if the custom medium is uniform on those surfaces.
> 
> @momchil-flex could you explain a bit more why it doesn't work even for a uniform custom medium?

[...]

PR #1588, opened 2024-04-08
Warning if a nonuniform custom medium is intersecting certain sources and monitors

weiliangjin2021:
Resolves https://github.com/flexcompute/tidy3d/issues/1373
- Add is_spatially_uniform property to materials
- Add is_uniform property to data array and unstructured dataset

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Support for `.gz` files in `Simulation` version updater.
+- Warning if a nonuniform custom medium is intersecting `PlaneWave`, `GaussianBeam`, `AstigmaticGaussianBeam`, `FieldProjectionCartesianMonitor`, `FieldProjectionAngleMonitor`, `FieldProjectionKSpaceMonitor`, and `DiffractionMonitor`.
 
 ### Changed
 
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -681,6 +681,11 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
     _name_validator = validate_name_str()
 
     @cached_property
+    def is_spatially_uniform(self) -> bool:
+        """Whether the medium is spatially uniform."""
+        return True
+
+    @cached_property
     def is_time_modulated(self) -> bool:
         """Whether any component of the medium is time modulated."""
         return self.modulation_spec is not None and self.modulation_spec.applied_modulation
@@ -1485,6 +1490,13 @@ class CustomIsotropicMedium(AbstractCustomMedium, Medium):
         return val
 
     @cached_property
+    def is_spatially_uniform(self) -> bool:
+        """Whether the medium is spatially uniform."""
+        if self.conductivity is None:
+            return self.permittivity.is_uniform
+        return self.permittivity.is_uniform and self.conductivity.is_uniform
+
+    @cached_property
     def n_cfl(self):
         """This property computes the index of refraction related to CFL condition, so that
         the FDTD with this medium is stable when the time step size that doesn't take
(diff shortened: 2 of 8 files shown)


tidy3d/components/medium.py
```diff
@@ -1450,7 +1455,7 @@ def _eps_inf_greater_no_less_than_one(cls, val):
         if not CustomIsotropicMedium._validate_isreal_dataarray(val):
             raise SetupError("'permittivity' must be real.")
 
-        if np.any(_get_numpy_array(val) < 1):
+        if np.any(_get_numpy_array(val) < 1 - fp_eps):
```
weiliangjin2021: Note that in our previous validation `if np.any(_get_numpy_array(val) < 1):`, it can error in a uniform custom medium of eps_inf =1 from unstructured dataset.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Uniaxial medium Lithium Niobate to material library.
 - Added support for conformal mesh methods near PEC structures that can be specified through the field `pec_conformal_mesh_spec` in the `Simulation` class.
 - EME solver through `EMESimulation` class.
+- Warning if a nonuniform custom medium is intersecting field projection monitor, diffraction monitor, planewave source, Gaussian beam, and AstigmaticGaussianBeam.
```
tylerflex: please put the single backslash quotes for variable / class names. And maybe we can use the class names for all of them for consistency?

tylerflex (approved):
looks good to me, thanks @weiliangjin2021 !

CHANGELOG.md
tylerflex: another minor comment is that we recently introduced restrictions / warnings on certain medium types intersecting. And now this PR does something similar. I wonder if we need to start thinking about a way to organize classes of components that don't work together, either in a single simulation or in an overlapping geometry. note: @caseyflex

caseyflex: yeah, might be good. One thing is that this is dependent on which solver / simulation type we are performing (not just a property of the components themselves)

tidy3d/components/medium.py
dbochkov-flexcompute: I don't think we need this change. I think the error comes from how the test is set up, not due to data being an unstructured dataset. See https://github.com/flexcompute/tidy3d/pull/1592

dbochkov-flexcompute (approved):

tidy3d/components/simulation.py
```diff
@@ -2494,15 +2494,24 @@ def _projection_monitors_homogeneous(cls, val, values):
...
+            for monitor_ind, monitor in enumerate(val):
+                if isinstance(monitor, (AbstractFieldProjectionMonitor, DiffractionMonitor)):
+                    mediums = Scene.intersecting_media(monitor, total_structures)
+                    # make sure there is no more than one medium in the returned list
+                    if len(mediums) > 1:
+                        raise SetupError(
+                            f"{len(mediums)} different mediums detected on plane "
+                            f"intersecting a {monitor.type}. Plane must be homogeneous."
+                        )
+                    # if <=1 medium, check if the medium is spatially uniform
```
momchil-flex: Is it possible to have 0 media? That kinda feels like something is wrong too, unless we're not taking the simulation background into account, but I think we are?

In other words not sure if we can remove the `for mat in mediums` and / or raise a different error if `mediums` is empty.

tidy3d/components/simulation.py
```diff
@@ -2755,23 +2764,33 @@ def _source_homogeneous_isotropic(cls, val, values):
...
+                    if len(mediums) == 1 and isinstance(
+                        list(mediums)[0], (AnisotropicMedium, FullyAnisotropicMedium)
+                    ):
+                        raise SetupError(
+                            f"An anisotropic medium is detected on plane intersecting a {source.type} "
+                            f"source. Injection of {source.type} into anisotropic media currently is "
+                            "not supported."
+                        )
+
+                    # if <=1 medium, check if the medium is spatially uniform
```
momchil-flex: Same comment as above.

momchil-flex (approved):
Hey, sorry, just now had a look at this, I have a small comment and then if you could please rebase to fix conflicts.

tidy3d/components/simulation.py
weiliangjin2021: It shouldn't be possible. An error has been raised now if 0 media.

PR #1589, opened 2024-04-08
mkdirs in to_gds_file calls if they dont exist

tylerflex:
Fixes (5) in https://flow360.atlassian.net/browse/SCEM-5706


lucas-flexcompute (approved):

lucas-flexcompute:
The tests are failing because you want mkdir, not makedirs: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

tylerflex:
Ah thanks, I forgot to fix that last one. Should be good now.

PR #1592, opened 2024-04-09
don't interpolate when setting up tests for uniform unstructured medium

dbochkov-flexcompute:


tests/test_components/test_custom.py
```diff
@@ -72,7 +72,8 @@ def make_spatial_data(value=0, dx=0, unstructured=False, seed=None, uniform=Fals
         data = np.random.random((Nx, Ny, Nz)) + value
     arr = td.SpatialDataArray(data, coords=dict(x=X + dx, y=Y, z=Z))
     if unstructured:
-        return cartesian_to_unstructured(arr, seed=seed)
+        method = "direct" if uniform else "linear"
```
weiliangjin2021, line 75: Why will `linear` slightly decrease eps_inf?

dbochkov-flexcompute: not sure, probably numerical round off errors. This keyword is only passed to xarray.interp for transferring data from an xarray to an unstructured dataset

dbochkov-flexcompute: https://github.com/flexcompute/tidy3d/blob/pre/2.7/tests/utils.py#L225-L234

weiliangjin2021: I see. It's possible that users might generate data after going through lots of steps (e.g. interpolation)  and then come across this eps_inf<1 error? I guess it'll be confusing for users to figure it out.

On the other hand, I think we need to update backend too if we only warn for eps_inf < 1- fp_eps. So probably good to merge yours for now?

dbochkov-flexcompute: Yeah, I imagine that could happen. It seems like we haven't had issues like that with cartesian custom medium, so it's probably ok to keep this check strict

Issue #1568, opened 2024-03-22
Multiple small issues of new flow


daquinteroflex:
- [x] sync to read the docs on the release is happening on the main branch, not the latest branch
- [x] Improve the way the latest branch and develop branch get updated so they don't get too far apart
- [ ] Have more clarity on the actual tidy3d-notebooks deployment, review Shaungs documentation
- [ ] Improve inherited methods display on documentation
- [ ] Pytest options for core, notebooks, etc
- [ ] Big ruff update on the repository
- [ ] Fix the notebooks testing script
- [ ] Notebooks visualisation docstings 
- [ ] Improve explanations on the new developer flow per Alec feedback
- [ ] Finish polishing off the poetry run test <multiple_options>

Pending from docs restructure:
- [ ] Add link to colab notebook with the minimal example and API key loaded as environment variable.
- [ ] Cooler intro page and pictures
- [ ]  The user can find more context when reading though the class definitions
- [ ]  Including equations when needed
- [ ]  Schematics to show geometric parameters
- [ ]  An example on how to use that object in a simulation.
- [ ]  Link to examples and tutorials
- [ ]  Cleaner API viewer
- [ ] Make nice illustration of medium property types, ie anisotropic isotropic for the uninitated
- [ ] Make nice illustrations of all the sources and geometrically how they are represented
- [ ] Standalone examples for most common class methods etc.
- [ ] Update README.md for notebooks
- [ ] Upgrade the jax requirements to caret and test across all platforms
- [x] #1604
- [ ] CLI test single notebook via arguments
- [ ] Update the development docs to the latest flow.
- [ ]See if we can remove the extra API classes nder the table
- [ ] Include the relevant markdown files inot their corresponding API sections.
- [ ] https://github.com/orgs/flexcompute/projects/2/views/1?pane=issue&itemId=49735837
- [ ] Sort out the old 404s from old links

yaugenst-flex:
Closing as stale — this tracking issue has had no activity since May 2024. Only 2 of ~15 checklist items were completed. Remaining items can be filed as individual issues if still relevant.

PR #1593, opened 2024-04-10
bump Jax-related versions

tylerflex:
@daquinteroflex 

I need to sort out some python installation stuff on my side. `poetry` is not working for me right now so I didnt' generate the new `.lock` file yet.

Just wondering if you think this versioning bump looks ok from your side. It's related to https://flow360.atlassian.net/browse/SCEM-5706


daquinteroflex:
I had in my list in https://github.com/flexcompute/tidy3d/issues/1568 to "upgrade the jax requirements to caret and test across all platforms", maybe this is a good time if this is what we want to do, whilst we mantain the version limiting. That way we can consolidate it in a single requirement platform independent possibly. Happy to implement that since I had it pending and run the tests in poetry if you like, or feel free to do it if you prefer to instead?

tylerflex:
Yea I think this is a good time, please go ahead. I tried generating the lock file but it was taking > 10 minutes. Is this normal? Could be that my hotel's internet is really bad too :D because it was taking a while to download all the requirements.

daquinteroflex:
It did take some time to make the poetry lock. I think jaxlib is pretty big now. Made a PR to this PR https://github.com/flexcompute/tidy3d/pull/1597

daquinteroflex (approved):

tylerflex:
I'm thinking maybe let's wait until 2.7.0rc1 is released before merging this, just in case.

FYI
@momchil-flex

momchil-flex:
Done in #1622

PR #1595, opened 2024-04-10
smatrix backwards compatible import

tylerflex:
smatrix notebooks had
```
from tidy3d.plugins.smatrix.smatrix import Port
```

this makes that still possible for backwards compatibility.


dmarek-flex (approved):
Thanks for fixing this!

Issue #642, opened 2023-01-01
upgrade to shapely>= 2.0.0


joamatab:
How could we make tidy3d compatible with the latest version of shapely?

https://github.com/gdsfactory/gdsfactory/pull/996

momchil-flex:
I've opened #643 that fixes this. There's only one change needed in the code.

However, when it comes to building it, shapely 2.0 failed to build on one of our workers that uses python 3.7. That is why I have kept the required version to 1.8.0 for python 3.7 and 2.0 for higher python. There were some weird issues where different patches of 1.8 would error for different python versions, which is why ideally I want to just have >= 2.0 eventually. Is this what you're planning for gdsfactory too?

joamatab:
Hi Momchil,

we test gdsfactory plugins including tidy3d with python3.7 and python3.10
https://github.com/gdsfactory/gdsfactory/actions/runs/3840773093

how about we unpin shapely so each python version gets the latest shapely version?

joamatab:
see the error of tidy3d with shapely 2.0.0 on python3.10

https://github.com/gdsfactory/gdsfactory/actions/runs/3871932000/jobs/6600214073

momchil-flex:
Yeah, this is fixed in #643 , but we haven't merged this in yet as we want to make sure setting the version to 2.0 will not cause other issues. The shapely R tree which we are using in the mesher had some version issues, like it would produce an error on some versions of shapely and python...

@tylerflex if we're building a 1.8.2 patch maybe we could just merge the change to `mesher.py` from that PR. It just ensures that the mesher works with both shapely 1.x and 2.x.

joamatab:
how about not pin the version? 

or if you switch to poetry you can specify 
pytest = "^2.0.0"

momchil-flex:
We are working towards switching to poetry

tylerflex:
@momchil-flex I'm fine with #643 going in 1.8.2. I'm a bit unsure about the shapely / python specifics but if that PR fixes things feel free to merge it into develop.

tylerflex:
For reference: our poetry PR #624

momchil-flex:
I'm reluctant to not pin the version for now since I've experienced issues on some combinations of python + shapely and want to avoid those. @joamatab if we are to push a patch (before we switch to poetry), would you rather the pinning stays with 1.8 as it is now, or 2.0?

joamatab:
Pinning to the most current version would be best in my opinion 


[...]

PR #1598, opened 2024-04-10
:wrench: FIX: v2.5.2 requirements included

daquinteroflex:
> Ahhhhhhhhh this was because I was working back them from the 2.4 defacto develop when I started rather than the pre/2.5 I think since I branched out incorrectly and migrated from there. This is why there is the mismatch in the translation of the requirements to pyproject.toml . I'm just judging by the  https://github.com/flexcompute/tidy3d/blob/v2.5.2/requirements/web.txt vs https://github.com/flexcompute/tidy3d/blob/v2.6.0/pyproject.toml

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ documentation = "https://docs.flexcompute.com/projects/tidy3d/en/latest/"
 python = ">=3.9,<3.12"
 pyroots = ">=0.5.0"
 xarray = ">=0.16.2"
-importlib-metadata = "^6.0.0"
+importlib-metadata = ">=6.0.0"
 h5netcdf = "1.0.2"
 h5py = "^3.0.0"
 rich = "<12.6.0"
@@ -36,14 +36,14 @@ dask = "*"
 toml = "*"
 ### NOT CORE
 scipy = "*"
-boto3 = "1.23.1"
-requests = "*"
+boto3 = "1.28.*"
+requests = "2.31.*"
 pyjwt = "*"
-click = "8.0.3"
+click = "8.1.*"
 responses = "*"
 ### END NOT CORE
 
-### Optional dependencies ### 
+### Optional dependencies ###
 # development core
 bump-my-version =  {version="*", optional = true}
 black = {version="23.12.1", optional = true}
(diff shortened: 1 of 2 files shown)


daquinteroflex:
Note that this PR is correct, it errored because pandas just released a new version 20 mins ago and the new poetry serveer has not fully reconciled with pypi I think. Should involve just running the aciton again.

EDIT: I've only updated the poetry lock up to the required dependencies now, whilst they sort it out.

pyproject.toml
```diff
@@ -23,27 +23,27 @@ documentation = "https://docs.flexcompute.com/projects/tidy3d/en/latest/"
...
 xarray = ">=0.16.2"
-importlib-metadata = "^6.0.0"
+importlib-metadata = ">=6.0.0"
 h5netcdf = "1.0.2"
 h5py = "^3.0.0"
 rich = "<12.6.0"
 numpy = "<2"
 matplotlib = "*"
-shapely = "^2.0"
+shapely = ">=2.0"
```
momchil-flex: What is the difference between "^" and ">="?

daquinteroflex: It actually has different meanings according to how it is used: https://python-poetry.org/docs/dependency-specification/#caret-requirements 

It means the higher minor version for a given major version in this context >=2, <3. Whereas >= just means the highest version from this point >=2 including 3.

daquinteroflex: Note that I changed it to match this: https://github.com/flexcompute/tidy3d/blob/v2.5.2/requirements/basic.txt#L10C1-L10C13

since there were no further PRs on that I found and https://github.com/flexcompute/tidy3d/issues/642

momchil-flex: Ok, actually `^` does make a lot of sense, I think we maybe just didn't know about it (or it is not supported in `requirements.txt` file?). Back when shapely introduced v2, we had something like >= 1.8; <2.0 in the file. So maybe `^` is better now actually? :) This can be done either here or in the 2.7 change...

tylerflex: yea probably we should keep the `^`, and even apply it to things  like `numpy`, `shapely`, `importlib-metadata`.

Maybe best for the 2.7 change (with the less restrictive versions) though?

tylerflex:
Note: for 2.7 we should also probably relax the requirements for `rich`, in addition to these ones.

pyproject.toml
daquinteroflex: Yep that sounds good, will make another PR for 2.7 on this

PR #1605, opened 2024-04-11
fix interpolation for unstructured data

dbochkov-flexcompute:
- skipping degenerate cells (when they are generated for example from slicing unstructured grid)
- also improved accuracy of interpolation. Previously interpolating a constant single precision field would introduce ~1e-6 errors

### DIFF
--- a/tidy3d/components/data/dataset.py
+++ b/tidy3d/components/data/dataset.py
@@ -546,6 +546,11 @@ class UnstructuredGridDataset(Dataset, np.lib.mixins.NDArrayOperatorsMixin, ABC)
         """Data type."""
         return np.iscomplexobj(self.values)
 
+    @property
+    def _double_type(self):
+        """Corresponding double data type."""
+        return np.complex128 if self.is_complex else np.float64
+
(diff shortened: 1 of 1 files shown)


dbochkov-flexcompute:
for some reason checks failed initially https://github.com/flexcompute/tidy3d/actions/runs/8653808570/attempts/1, but rerunning them went fine

tylerflex:
```
FAILED tests/test_plugins/test_design.py::test_sweep[monte_carlo] - tidy3d.exceptions.SetupError: AutoGrid coordinates along axis 0 do not match the simulation domain, indicating an unexpected error in the meshing. Please create a github issue so that the problem can be investigated. In the meantime, switch to uniform or custom grid along 0.
```
this is weird. I think we should investigate. Maybe some of the random points in the Monte Carlo sampling triggered the auto grid to fail at an edge case?

@momchil-flex @weiliangjin2021

momchil-flex:
We can't really do much about this unless we have the actual simulation... If we see this happen again (or even if we don't..) we could update this test to write the simulation to file if an error happens.

PR #1606, opened 2024-04-12
fix field data flux

tylerflex:
The `adjoint` plugin `FieldData.flux` previously returned a `float` that summed over all of the flux values.

For multi-frequency adjoint, this is not quite what we want, rather we want a `JaxDataArray` with frequency coordinate.

(I think this was added before multi-frequency adjoint and just wasn't fixed when that was introduced).

This PR properly returns a `JaxDataArray` (`FluxDataArray`).

A potential issue for backwards compatibility if users have `.flux` in their notebooks. I'm not sure how to resolve this. Options are

1. Just change it to the "expected" behavior (return data array) and explain if users are confused.
2. Do some automatic `sum()` if certain conditions are met (like a single frequency coordinate)
3. Add a `sum: bool=True` kwarg so we can instruct users that want all the flux values to set this to `False`

fyi @tomflexcompute

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bug in PolySlab intersection if slab bounds are `inf` on one side.
 - Better error message when trying to transform a geometry with infinite bounds.
 - `JaxSimulation.epsilon` properly handles `input_structures`.
+- `FieldData.flux` in adjoint plugin properly returns `JaxDataArray` containing frequency coordinate `f` instead of summing over values.
 
 ## [2.6.3] - 2024-04-02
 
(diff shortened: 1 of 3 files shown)


momchil-flex:
Ah. Kinda bad. I kinda prefer 1 (not sure how 2 differs, if this is an issue only with multi-freq). It's probably rare to want to sum the values, and even if you do, you can still do it very easily in your own function. Not sure how many people we'll have to explain this to but it feels like we should just pull the band-aid...

tylerflex:
Option 2 would keep the current handling (return `float`) for single frequency, but return a `DataArray` for multi-frequency.

tylerflex:
I just revised this PR to try the handling in 2. see the new code here: 
https://github.com/flexcompute/tidy3d/blob/1b39cc2daa1fba35d236f6961f267a913ef0a073/tidy3d/plugins/adjoint/components/data/monitor_data.py#L163-L174

also added a more proper test of both cases.

Maybe I'd prefer this for now?

momchil-flex:
I think this makes sense. Do we really want to put it in a patch though? Since we're planning the 2.7.0rc1 for Monday is it better to just put it there?

tylerflex:
Fine with me. I'll try to change it to pre/2.7 before Monday

tylerflex:
@momchil-flex rebased against 2.7

PR #1616, opened 2024-04-15
models for microstrips

dmarek-flex:
I am working on a notebook which required some of the functionality in this PR, and I thought it would be nice to include it in the microwave plugin for quick back of the envelope calculations users might want to do.

These are basically a collection of free functions which I have packaged into classes as static methods. Other than better documentation,  do you have any advice on what the best way to expose this functionality?


tylerflex:
> These are basically a collection of free functions which I have packaged into classes as static methods. Other than better documentation, do you have any advice on what the best way to expose this functionality?

I think this is a good way to do it. I have a similar functionality in `invdes` plugin I'm working with: a few of the helper functions for post processing are contained as static methods and class methods in an `invdes.Utils` class. It seems nicer to just do `help(Utils)` instead of trying to look through all of the isolated functions.

tylerflex (approved):

weiliangjin2021:
Looks good to me. Just need more docstrings, e.g. return types etc.

dmarek-flex:
I change the API to use a module/namespace approach since there is no data associated with the classes. Let me know if anything needs to be improved.

weiliangjin2021:
> I change the API to use a module/namespace approach since there is no data associated with the classes. Let me know if anything needs to be improved.

That looks fine. One thing to consider is whether we might need to override some private functions in the future, e.g. a slightly different formulation, or geometry. In that case, even if there is no data, it's more convenient to use class inheritance to avoid rewriting everything. This was the case for mode solver.

dmarek-flex:
> > I change the API to use a module/namespace approach since there is no data associated with the classes. Let me know if anything needs to be improved.
> 
> That looks fine. One thing to consider is whether we might need to override some private functions in the future, e.g. a slightly different formulation, or geometry. In that case, even if there is no data, it's more convenient to use class inheritance to avoid rewriting everything. This was the case for mode solver.

That is a good point. I think in this case though the functions are very specific, and as a result it is fairly unlikely they will be useful for other geometries or use cases.

weiliangjin2021 (approved):

PR #1619, opened 2024-04-16
fix source discretization

dmarek-flex:
related to the backend [PR](https://github.com/flexcompute/tidy3d-core/pull/594).

Essentially unifies some code that optionally slightly enlarges the box when using `grid.discretize_inds`

### DIFF
--- a/tidy3d/components/geometry/utils_2d.py
+++ b/tidy3d/components/geometry/utils_2d.py
@@ -4,11 +4,11 @@ import shapely
 from typing import Tuple, List
 
 from ..types import Axis
-from ...constants import inf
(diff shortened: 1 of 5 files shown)


tidy3d/components/geometry/utils_2d.py
```diff
@@ -18,24 +18,6 @@
 DIST_NEIGHBOR_REL_2D_MED = 1e-5
 
 
-def increment_float(val: np.float32, sign) -> np.float32:
```
dmarek-flex, line 21: moved `increment_float` to `utils.py `to avoid a circular dependency

PR #1620, opened 2024-04-16
Lifting python < 3.12 requirement

momchil-flex:


momchil-flex:
Closing in favor of #1622

Issue #1317, opened 2023-12-15
Future `python==3.12` compatibility.


daquinteroflex:
Just thought to write a list of depreciated commands we need to update throughout the frontend

- [X] All the `typing.List`, `typing.Tuple`, etc. need to be updated to standard `list`, `tuple` type declarations. UPDATE: turns out they're only depreciating them from development, but still mantain backward compatibility. I thought I had read they were being depreciated, but I guess I misunderstood this as not supported. I think they will support it for backward compatibility until they choose t[o remove them ](https://docs.python.org/3/library/typing.html#deprecation-timeline-of-major-features).
- [X] jax updates
- [x] `gdstk` update for 3.12 windows support? re https://github.com/heitzmann/gdstk/issues/228. Actions issue can be partly linked to [this](https://github.com/python-poetry/poetry/issues/5599). Possibly we need [this](https://github.com/python-poetry/poetry/pull/9150) to be propagated along in our poetry config to force a specific installation.
- [x] We know `trimesh` ->  `rtree` are [not bundling unlike previous versions](https://github.com/geometries/trimesh_upstream?tab=readme-ov-file#current-status) `libspatialindex`. Need to decide what to do about this. rtree files are [here](https://pypi.org/project/Rtree/#files) and [libspatialindex](https://libspatialindex.org/en/latest/index.html) is here. Fixed by updating rtree in the dependencies.

We can work around:
- [x] poetry issues on pandas 3.9 wheels on their latest release https://github.com/pandas-dev/pandas/issues/58289

Poetry issues we might have:
- Related to [binary installs](https://github.com/python-poetry/poetry/issues/6798). Sorted by manually installing onto the venv via pip poetry 1.9 releases only-binary

tylerflex:
hm, so I guess python 3.12 is already out since October? are we behind in not supporting it yet?

If we change `typing.Tuple` -> `tuple` will that affect our package when running older versions of python?

momchil-flex:
I'm pretty sure it's OK already in 3.9 but I don't remember if 3.8 accepts those, maybe not.

We *could* drop 3.8 soon although probably a few users would prefer not yet. :)

qt-thomas:
I'd be happy if this was done sooner rather than later. My environment is at 3.12 and pip installs 2.5.2 by default, instead of 2.6.

momchil-flex:
[...]

PR #1622, opened 2024-04-17
:package: BUILD: Python 3.12 compatibility

daquinteroflex:
Related to:

https://github.com/flexcompute/tidy3d/pull/1620
https://github.com/flexcompute/tidy3d/issues/1317
https://github.com/flexcompute/tidy3d/pull/1593


momchil-flex:
So the only remaining issue is mac with python 3.12 and gdstk? Probably @lucas-flexcompute needs to build a new wheel or something?

lucas-flexcompute:
I did not notice macOS was missing python 3.11 and 3.12! I think it's just a matter of updating the GH actions for CIBW. I'm checking now: https://github.com/heitzmann/gdstk/actions/runs/8730179893

lucas-flexcompute:
The 3.12 versions are now on PyPI.

momchil-flex:
Seems like now xarray errors...

daquinteroflex:
Thanks both for your help and advice! Think it's passing now and seems reasonable imo. However, maybe we want to merge https://github.com/flexcompute/tidy3d/pull/1593 before this one as it depends on those updates too.

pyproject.toml
```diff
@@ -35,6 +36,7 @@ PyYAML = "*"
 dask = "*"
 toml = "*"
 ### NOT CORE
+pandas = "<=2.2.1"
```
momchil-flex: Since pandas is a requirement of xarray, and this is presumably done to make sure xarray installs well, I think it should go in CORE

pyproject.toml
```diff
@@ -66,21 +68,13 @@ gdspy = {version="*", optional = true}
...
-]
+jaxlib = {version = "^0.4.13,<=0.4.26", source="jaxsource", optional = true}
+jax = {version = "^0.4.13,<=0.4.26", extras = ["cpu"], source="jaxsource", optional = true}
 
 # trimesh
 networkx = {version = "^2.6.3", optional = true}
-rtree = {version = "1.0.1", optional = true}
-trimesh = {version = ">=4,<4.2", optional = true}
+rtree = {version = "1.2.0", optional = true}
+trimesh = {version = "^4", optional = true}
```
momchil-flex, line 77: Are you sure about this? Who set `trimesh < 4.2` before, @caseyflex ?

pyproject.toml
```diff
@@ -100,10 +94,11 @@ sphinx-notfound-page = {version="*", optional = true}
 sphinx-tabs = {version="*", optional = true}
 nbdime = {version="*", optional = true}
 myst-parser = {version="*", optional = true}
-optax = {version="*", optional = true}
+optax = {version=">=0.2.2", optional = true}
 signac = {version="*", optional = true}
+flax = {version=">=0.8.2", optional = true}
 sax = {version="*", optional = true}
-vtk = {version="<=9.2.6", optional = true}
+vtk = {version=">=9.2.6", optional = true}
```
momchil-flex, line 101: Similarly, are you sure about this? @dbochkov any thoughts?

tests/test_timing/import_tidy3d.py
```diff
@@ -1,17 +1,17 @@
...
+# for _ in range(n):
+#     s = time.time()
+#     subprocess.call(["python", "-c", "import tidy3d"])
+#     tidy3d_load_time += time.time() - s
+#
+#     s = time.time()
+#     subprocess.call(["python", "-c", "pass"])
+#     python_load_time += time.time() - s
+#
+# print(f"average tidy3d load time = {(tidy3d_load_time - python_load_time) / n}")
```
momchil-flex: Why?

momchil-flex (commented):
Not sure whether to put this in rc1 or not. I guess it may be nice to test right away...

pyproject.toml
caseyflex: I set it that way to match gdsfactory, which has since updated to `>=4, <4.4`. I think `^4` is better so we don't have to keep updating it. I don't know if the rtree version matters much as long as it is `>= 1.0.1`.

momchil-flex:
AHHH the dangers of force pushing!!! Yesterday I had done `git push -f origin dario/python_3_12` to remove the commit that excluded the Mac OS text. Today I just completely not on purpose ran that command again (I was looking to repeat a different command) and it overwrote your branch!

I assume you have a copy locally so hopefully no big deal... Sorry!

pyproject.toml
dbochkov-flexcompute: Should be ok starting with 2.7

pyproject.toml
daquinteroflex: Great!

tests/test_timing/import_tidy3d.py
daquinteroflex: Whoops that was temporary to speed up testing, but now have brought it back.

daquinteroflex:
No worries haha! All feedback implemented.

I'm ok if we want to add it to rc1. I believe the main doubt were the jax upgrade questions @tylerflex fyi

momchil-flex:
Well, better to get feedback right away than when we do the official release :)

PR #1623, opened 2024-04-18
warning and changelog for `JaxPolySlab.vertices` no longer tracing derivatives

tylerflex:
see https://github.com/flexcompute/tidy3d-notebooks/pull/103 for notebook change


tidy3d/plugins/adjoint/utils/penalty.py
```diff
@@ -70,6 +79,17 @@ def evaluate(self, points: ArrayFloat2D) -> float:
         desired minimum value. If ``wrap``, it is assumed that the points wrap around to form a
         closed geometry instead of an isolated line segment."""
 
+        if not is_jax_object(points):
+            log.warning(
+                "The points passed to 'RadiusPenalty.evaluate()' are not a 'jax' array. "
+                "If you are passing the 'JaxPolySlab.vertices' field directly, note that the "
+                "derivative information for this field is no longer traced by as of 'tidy3d 2.7'. "
```
momchil-flex: - I always find using "you" in these messages a bit strange. Here it can be easily changed, `If you are passing` -> `If passing`. In other places it may be more awkward to rephrase, so not sure about a general policy. But I feel like usually packages are not talking to "you", i.e. I feel like there is some implicit standard about this.

- ~~by~~ as of

- `'tidy3d 2.7'` seems a bit strange formatting? Do we put `''` around package names (also `'jax'` here)? And even if so maybe `'tidy3d' v2.7`?

tidy3d/plugins/adjoint/utils/penalty.py
```diff
@@ -70,6 +79,17 @@ def evaluate(self, points: ArrayFloat2D) -> float:
...
+                "The derivative information is contained in 'JaxPolySlab.vertices_jax'. "
+                " Therefore, we recommend changing your code to either pass that field or pass "
```
momchil-flex: Double space at the end of previous line and at the beginning of this one.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ability to add `path` to `updated_copy()` method to recursively update sub-components of a tidy3d model. For example `sim2 = sim.updated_copy(size=new_size, path="structures/0/geometry")` creates a recursively updated copy of `sim` where `sim.structures[0].geometry` is updated with `size=new_size`.
 
 ### Changed
+- IMPORTANT NOTE: differentiable fields in the `adjoint` plugin (`JaxBox.size`, `JaxBox.center`, `JaxPolySlab.vertices`) no longer store the derivative information after the object is initialized. For example, an objective function involving `polyslab.vertices` will not include the vertices in the gradient function. Instead, this information is now stored in a field of the same name with `_jax` suffix, eg. `JaxPolySlab.vertices_jax`. For some applications, such as evaluating penalty values, you will need to change to `radius_penalty.evluate(polyslab.vertices_jax)` or use the vertices as generated by your parameterization functions (`make_vertices(params)`).
```
momchil-flex: - Maybe "For example, if using `JaxPolySlab.vertices` directly in an objective fucntion, the vertices will have no effect on the gradient." ?
- Same comment about using "you" as in my other comment

momchil-flex (approved):
Looks good thanks! Some (nit) formatting / phrasing suggestions.

tidy3d/plugins/adjoint/utils/penalty.py
tylerflex: changed to 

> "The points passed to 'RadiusPenalty.evaluate()' are not a 'jax' array. If passing the 'JaxPolySlab.vertices' field directly, note that the derivative information for this field is no longer traced by jax as of version '2.7'. The derivative information is contained in 'JaxPolySlab.vertices_jax'.  Therefore, we recommend changing your code to either pass that field or pass the output of the parameterization functions directly, eg. 'penalty.evaluate(make_vertices(params))'."

tidy3d/plugins/adjoint/utils/penalty.py
tylerflex: fixed

CHANGELOG.md
tylerflex: fixed

PR #1630, opened 2024-04-22
batch refactoring / multi-threading

tylerflex:
## Goals:
1. Simplifies how `Batch` and `Job` work internally.
2. Improves the visual output from `Batch`.
3. Implements multi-threading for batch operations (upload, download, etc).

image

## Refactor

Previously, `Job` objects were defined like this

```py
class Job:
    ...
    task_id = None

    @pd.root_validator()
    def _upload(cls, values) -> None:
          # upload task
          # set task id internally
```

This meant that when you created a `Job`, the task is immediately uploaded.
Also when you save a `Job.to_file()` the task info is contained.

This design made it a bit annoying to do normal multi-threading over job upload.

In the refactored version, I set it up to have `upload` in its own method.

```py
class Job:
    ...
    task_id_cached = None

    @cached_property
    def task_id(self):
        # return self.task_id_cached or call `self.upload()` 

    def upload(self):
         ...

    def json(self):
          # extend json to add `cached_properties["task_id"] to json file for backwards compatibility when loading from file

```

Now a similar refactoring was done for `Batch` to move the `upload()` bits to their own method.

Before, the `Batch` created all `Job` objects and uploaded them at initialization time in a validator.

```py
class Batch:
    ...
    jobs : Dict[str, Job] = None

    @pd.root_validator
    def _upload():
        # for each simulation in the batch, make a job (which automatically uploads it).

```

instead, now that we have the `Job` with its own `upload` method, I simplified this to.

```py
class Batch:
    ...
    jobs : Dict[str, Job] = None

    @cached_property
    def jobs():
        # construct all of the Jobs for this Batch

    def upload(self):
        # upload each Job in a for loop
```

## Multithreading

Implemented basic multi-threading for `Batch.upload()` and `Batch.download()`. Works as expected on the Design plugin notebook.

Also made it so that `Batch.run()` will download by default.

## Output / Logging

- The batch upload, download, and BatchData loading all just log a single output instead of one line for every file.

### DIFF
--- a/tidy3d/plugins/adjoint/web.py
+++ b/tidy3d/plugins/adjoint/web.py
@@ -241,7 +241,8 @@ class AdjointJob(Job):
         ----
         To monitor progress of the :class:`Job`, call :meth:`Job.monitor` after started.
         """
-        upload_jax_info(task_id=self.task_id, jax_info=self.jax_info, verbose=self.verbose)
+        if self.jax_info is not None:
+            upload_jax_info(task_id=self.task_id, jax_info=self.jax_info, verbose=self.verbose)
         super().start()
 
 
@@ -254,44 +255,25 @@ class AdjointBatch(Batch):
         description="Type of simulation, used internally only.",
     )
 
-    jobs: Dict[str, AdjointJob] = pd.Field(
-        None,
-        title="Simulations",
-        description="Mapping of task names to individual AdjointJob object for each task "
-        "in the batch. Set by ``AdjointBatch.upload``, leave as None.",
-    )
-
     jax_infos: Dict[str, JaxInfo] = pd.Field(
         ...,
         title="Jax Info Dict",
         description="Containers of information needed to reconstruct JaxSimulation for each item.",
     )
 
-    @pd.root_validator()
-    def _add_jax_infos(cls, values) -> None:
-        """Add jax_info fields to the uploaded jobs."""
-        jax_infos = values.get("jax_infos")
-        jobs = values.get("jobs")
-
-        if jobs is None:
-            return values
-
-        for task_name, job in jobs.items():
-            jax_info = jax_infos[task_name]
-            values["jobs"][task_name] = job.updated_copy(jax_info=jax_info)
-
-        return values
+    _job_type = AdjointJob
 
     def start(self) -> None:
-        """Start running all tasks in the :class:`Batch`.
+        """Start running a :class:`AdjointBatch`. after uploading jax info for each job.
 
         Note
         ----
-        To monitor the running simulations, can call :meth:`Batch.monitor`.
+        To monitor progress of the :class:`Batch`, call :meth:`Batch.monitor` after started.
         """
-        for _, job in self.jobs.items():
-            upload_jax_info(task_id=job.task_id, jax_info=job.jax_info, verbose=self.verbose)
-            job.start()
+        for task_name, job in self.jobs.items():
+            jax_info = self.jax_infos.get(task_name)
+            upload_jax_info(task_id=job.task_id, jax_info=jax_info, verbose=job.verbose)
+        super().start()
 
 
 def webapi_run_adjoint_fwd(
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -3,9 +3,11 @@ from __future__ import annotations
 
 import os
 from abc import ABC
-from typing import Dict, Tuple
+from typing import Dict, Tuple, Optional
 import time
 import json
+from concurrent.futures import ThreadPoolExecutor
+import concurrent
 
 from rich.progress import Progress
 import pydantic.v1 as pd
@@ -20,8 +22,11 @@ from ...log import log, get_logging_console
 
 from ...exceptions import DataError
 
+# Max # of workers for parallel upload / download: above 10, performance is same but with warnings
+DEFAULT_NUM_WORKERS = 10
 DEFAULT_DATA_PATH = "simulation_data.hdf5"
 DEFAULT_DATA_DIR = "."
+BATCH_MONITOR_PROGRESS_REFRESH_TIME = 0.02
 
 
 class WebContainer(Tidy3dBaseModel, ABC):
@@ -147,7 +152,14 @@ class Job(WebContainer):
         None, title="Parent Tasks", description="Tuple of parent task ids, used internally only."
     )
 
-    task_id: TaskId = None
+    task_id_cached: TaskId = pd.Field(
+        None,
+        title="Task ID (Cached)",
+        description="Optional field to specify ``task_id``. Only used as a workaround internally "
+        "so that ``task_id`` is written when ``Job.to_file()`` and then the proper task is loaded "
+        "from ``Job.from_file()``. We recommend leaving unset as setting this field along with "
+        "fields that were not used to create the task will cause errors.",
+    )
 
     _upload_fields = (
         "simulation",
@@ -163,12 +175,14 @@ class Job(WebContainer):
         """Save ``Job`` to dictionary. Add the `task_id` if it has been cached."""
 
         self_json = super().json(**kwargs)
+
+        task_id = self._cached_properties.get("task_id")
+
+        if not task_id:
+            return self_json
+
         self_dict = json.loads(self_json)
-
-        task_id = self._cached_properties.get("_task_id")
-        if task_id:
-            self_dict["task_id"] = task_id
-
+        self_dict["task_id_cached"] = task_id
         return json.dumps(self_dict)
 
     def run(self, path: str = DEFAULT_DATA_PATH) -> SimulationDataType:
@@ -191,14 +205,16 @@ class Job(WebContainer):
         return self.load(path=path)
 
     @cached_property
-    def _task_id(self):
-        if self.task_id:
-            return self.task_id
+    def task_id(self) -> TaskId:
+        """The task ID for this ``Job``. Uploads the ``Job`` if it hasn't already been uploaded."""
+        if self.task_id_cached:
+            return self.task_id_cached
         return self.upload()
 
     def upload(self) -> TaskId:
         # upload kwargs with all fields except task_id
-        upload_kwargs = {key: getattr(self, key) for key in self._upload_fields}
+        self_dict = self.dict()
+        upload_kwargs = {key: self_dict.get(key) for key in self._upload_fields}
         return web.upload(**upload_kwargs)
 
     def get_info(self) -> TaskInfo:
@@ -210,7 +226,7 @@ class Job(WebContainer):
             :class:`TaskInfo` object containing info about status, size, credits of task and others.
         """
 
-        return web.get_info(task_id=self._task_id)
+        return web.get_info(task_id=self.task_id)
 
     @property
     def status(self):
@@ -224,7 +240,7 @@ class Job(WebContainer):
         ----
         To monitor progress of the :class:`Job`, call :meth:`Job.monitor` after started.
         """
-        web.start(self._task_id, solver_version=self.solver_version)
+        web.start(self.task_id, solver_version=self.solver_version)
 
     def get_run_info(self) -> RunInfo:
         """Return information about the running :class:`Job`.
@@ -234,7 +250,7 @@ class Job(WebContainer):
         :class:`RunInfo`
             Task run information.
         """
-        return web.get_run_info(task_id=self._task_id)
+        return web.get_run_info(task_id=self.task_id)
 
     def monitor(self) -> None:
         """Monitor progress of running :class:`Job`.
@@ -244,7 +260,7 @@ class Job(WebContainer):
         To load the output of completed simulation into :class:`.SimulationData` objects,
         call :meth:`Job.load`.
         """
-        web.monitor(self._task_id, verbose=self.verbose)
+        web.monitor(self.task_id, verbose=self.verbose)
 
     def download(self, path: str = DEFAULT_DATA_PATH) -> None:
         """Download results of simulation.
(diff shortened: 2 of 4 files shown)


tidy3d/web/api/container.py
```diff
@@ -162,6 +162,20 @@ class Job(WebContainer):
         "parent_tasks",
     )
 
+    def json(self, **kwargs):
+        """Save ``Job`` to dictionary. Add the `task_id` if it has been cached."""
```
lei-flex, line 175: we don't need cache here. all these codes will be very fast since in memory. 
the cache won't bring perf improvement significally compared to async call.

tidy3d/web/api/container.py
```diff
@@ -176,25 +170,19 @@ def run(self, path: str = DEFAULT_DATA_PATH) -> SimulationDataType:
...
-    def _upload(cls, values) -> None:
-        """Upload simulation to server without running."""
-
-        # task_id already present, don't re-upload
-        if values.get("task_id") is not None:
-            return values
+    @cached_property
+    def task_id(self):
+        upload_kwargs = {key: getattr(self, key) for key in self._upload_fields}
+        return web.upload(**upload_kwargs)
```
lei-flex: why don't we use **self._upload_fields directly?

tidy3d/web/api/container.py
```diff
@@ -176,25 +170,19 @@ def run(self, path: str = DEFAULT_DATA_PATH) -> SimulationDataType:
...
 
+    def upload(self):
         # upload kwargs with all fields except task_id
-        upload_kwargs = {key: values.get(key) for key in cls._upload_fields}
-        task_id = web.upload(**upload_kwargs)
-
-        # then set the task_id and return
-        values["task_id"] = task_id
-        return values
+        self.task_id
```
lei-flex: the function with upload but just directly return sefl.task_id will confuse the reader very much.

tidy3d/web/api/container.py
lei-flex: the disk IO at least have 200MB per second.
but if we keep 200M data in memroy, the python could be out of memory very soon

tidy3d/web/api/container.py
```diff
@@ -191,9 +196,9 @@ def run(self, path: str = DEFAULT_DATA_PATH) -> SimulationDataType:
         return self.load(path=path)
 
     @cached_property
-    def _task_id(self):
-        if self.task_id:
-            return self.task_id
+    def task_id(self):
```
lei-flex: add the return for this since it changed to return self.upload()

tidy3d/web/api/container.py
```diff
@@ -719,14 +731,18 @@ def download(self, path_dir: str = DEFAULT_DATA_DIR) -> None:
...
-            if "error" in job.status:
-                log.warning(f"Not downloading '{task_name}' as the task errored.")
-                continue
+                if "error" in job.status:
+                    log.warning(f"Not downloading '{task_name}' as the task errored.")
+                    continue
 
-            job.download(path=job_path)
+                print(f"downloading '{task_name}'")
+                future = executor.submit(lambda: job.download(path=job_path))
```
lei-flex: how to wait for all files get download successfully?

tidy3d/web/api/container.py
tylerflex: @lei-flex this function just adds the `task_id` to the `.json()` file, if it has already been computed.

The reason I include it is not for performance, but rather because in the current `tidy3d` we support saving and loading a `Job` to file with the `task_id` information stored.

After the refactor, the `task_id` was no longer written to file, so this is the workaround for backwards compatibility.

tidy3d/web/api/container.py
tylerflex: `_upload_fields` is a tuple of strings referring to the keys. So we need to use `getattr()` to get the dictionary corresponding to the upload kwargs.

I agree this part is a bit ugly. I might rework the implementation a bit. The point, though, is that the `_upload_fields` class attribute defines what arguments must be passed to `web.upload()`.

 It was necessary to define this for adjoint, but now I'm trying to make use of it to generalize things a bit.

tidy3d/web/api/container.py
tylerflex: ok how about I change so that

```py
    @cached_property
    def task_id(self)  -> TaskID:
        return self.upload()

    def upload(self) -> TaskID:
        # upload kwargs with all fields except task_id
        upload_kwargs = {key: getattr(self, key) for key in self._upload_fields}
        return web.upload(**upload_kwargs)
```

the idea is that when `.task_id` is asked for, if it hasn't been computed, it will `upload()` the task.

tidy3d/web/api/container.py
tylerflex: ah yes, I forgot (I haven't done a full lint check yet, just basics). Thanks

tidy3d/web/api/container.py
lei-flex: will it lead to memory issue if we keep loading more and more data into cache?

tidy3d/web/api/container.py
tylerflex: Good point, so I want to basically block the return of this function until all files have downloaded / futures have executed? I think through `ThreadPoolExecutor` I can check the status of the queue and maybe return if it is empty. I'll look into this.

tidy3d/web/api/container.py
tylerflex: It's just loading the task id string. I think it will be fine

tidy3d/web/api/container.py
lei-flex: did you try with instance_attrs = vars(obj)    or obj.__dict__  ?

tylerflex: I think using pydantic `obj.dict().get(key)` will work. I'll try that.

tylerflex: seemed to work fine with this approach

tidy3d/web/api/container.py
tylerflex: It looks like the `ThreadPoolExecutor` context manager already waits to continue until all `futures` have been completed. So it should be fine written as is.

tidy3d/web/api/container.py
```diff
@@ -13,12 +15,14 @@
...
 from ..core.constants import TaskId, TaskName
-from ...components.base import Tidy3dBaseModel
+from ...components.base import Tidy3dBaseModel, cached_property
 from ...components.types import annotate_type
 from ...log import log, get_logging_console
 
 from ...exceptions import DataError
 
+# Max # of workers for parallel upload / download: above 10, performance is same but with warnings
+DEFAULT_NUM_WORKERS = 10
```
lucas-flexcompute, line 26: Is this dependent on the user connection or our servers? This is more out of curiosity; 10 seems like a reasonable default to me.

tidy3d/web/api/container.py
```diff
@@ -354,15 +366,17 @@ def load_sim_data(self, task_name: str) -> SimulationDataType:
         task_data_path = self.task_paths[task_name]
         task_id = self.task_ids[task_name]
         web.get_info(task_id)
+
         return web.load(
             task_id=task_id,
             path=task_data_path,
             replace_existing=False,
-            verbose=self.verbose,
+            verbose=False,
```
lucas-flexcompute, line 383: I don't understand why this must be false now… does it interact with the multi-threaded processes somehow?

tidy3d/web/api/container.py
```diff
@@ -504,41 +522,62 @@ def run(self, path_dir: str = DEFAULT_DATA_DIR) -> BatchData:
...
+    def upload(self) -> None:
+        """Upload a series of tasks associated with this ``Batch`` using multi-threading."""
+
+        if self.verbose:
+            console = get_logging_console()
+            console.log(f"Uploading data for {self.num_jobs} tasks.")
+
+        with ThreadPoolExecutor(max_workers=self.num_workers) as executor:
+            for task_name, job in self.jobs.items():
+                future = executor.submit(job.upload)
```
lucas-flexcompute: We could use [`concurrent.futures.as_completed`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait) to update a progress bar here only based on the number of uploads completed. At least is better feedback than nothing. The same could be used for downloads. There is an [example](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor-example) o fhow to use it to wait for completed futures.

lucas-flexcompute (approved):

tidy3d/web/api/container.py
tylerflex: I don't know actually. @lei-flex ?

tylerflex: @lucas-flexcompute it might be our severs, because I tried several things to change the configuration to allow more workers but it still gave me a warning above 10.

tidy3d/web/api/container.py
tylerflex: When we loaded `N` tasks, it used to print out `N` separate messages automatically. I think it cluttered the output significantly, so now it just logs that it's uploading `N` tasks and then logs that it's downloading `N` tasks.

tylerflex: maybe I should add a log here, though?
```py
if self.verbose:
    console.log("Loading {N} tasks from files.")
```

tidy3d/web/api/container.py
tylerflex: Thanks, this worked like a charm. Now I have a single progress bar for download and upload that updates based on the number of tasks uploaded / downloaded.

tidy3d/web/api/container.py
tylerflex: I tried this out and it was a bit clunky maybe. Do we want to log something every time a user iterates through batch data? Not sure what the user gains from this. what do you think?

```py
for task_name, sim_data in batch_data.items():  # <- would log
    ...
```

daquinteroflex: When I did the similar multi threading script, I think that the original ogging caused some bugs on either downloading all tasks or sometimes just a section of them. Removing the original logging verbosity made it work as expected. However, maybe this implementation sorts that issue under the hood properly.

tidy3d/web/api/container.py
```diff
@@ -462,13 +477,17 @@ class Batch(WebContainer):
...
-        title="Simulations",
-        description="Mapping of task names to individual Job object for each task in the batch. "
-        "Set by ``Batch.upload``, leave as None.",
+    num_workers: Optional[pd.PositiveInt] = pd.Field(
+        DEFAULT_NUM_WORKERS,
+        title="Number of Workers",
+        description="Number of workers for multi-threading upload and download of batch. "
+        "Corresponds to ``max_workers`` argument passed to "
+        "``concurrent.futures.ThreadPoolExecutor``. When left ``None``, will pass the maximum "
+        "number of threads available on the system.",
```
daquinteroflex, line 494: Hmm I was reading about this as it was interesting:

https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor
> Changed in version 3.8: Default value of max_workers is changed to min(32, os.cpu_count() + 4). This default value preserves at least 5 workers for I/O bound tasks. It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL. And it avoids using very large resources implicitly on many-core machines. ThreadPoolExecutor now reuses idle worker threads before starting max_workers worker threads too.

Sounds like the default when it is None is limited to avoid it consuming all the resources and maybe it changed since they wrote the original docs, so maybe we want to reference this as maybe this affects the super power users who want to max out their CPUs

daquinteroflex (approved):
Nice! Looks great, excited to test this!

tidy3d/web/api/container.py
tylerflex: hm, for me `os.cpu_count() + 4` is 12. 🤔

tidy3d/web/api/container.py
lucas-flexcompute: Yes, I see the how it can be annoying. I think I agree with leaving it `False` as it is.

tylerflex:
Thanks @lei-flex @daquinteroflex @lucas-flexcompute for the comments. 
I think this is ready to go. Did a bit of testing with WebAPI.ipynb, `Design.ipynb` and the adjoint batching tutorial. All looks good. Seems faster than before and also the output is much more reasonable. 

Let me know if you think this needs more review or work on any particular aspects, thanks!

tidy3d/plugins/adjoint/web.py
```diff
@@ -254,44 +255,25 @@ class AdjointBatch(Batch):
...
         ----
-        To monitor the running simulations, can call :meth:`Batch.monitor`.
+        To monitor progress of the :class:`Batch`, call :meth:`Batch.monitor` after started.
         """
-        for _, job in self.jobs.items():
-            upload_jax_info(task_id=job.task_id, jax_info=job.jax_info, verbose=self.verbose)
-            job.start()
+        for task_name, job in self.jobs.items():
+            jax_info = self.jax_infos.get(task_name)
+            upload_jax_info(task_id=job.task_id, jax_info=jax_info, verbose=job.verbose)
```
lei-flex, line 275: the multiple thread is requried here as well to upload jax_info as well.

lei-flex: 2. s3_util should define a interfact to put the memory content to s3 directly without write to disk

tidy3d/web/api/container.py
```diff
@@ -176,25 +199,23 @@ def run(self, path: str = DEFAULT_DATA_PATH) -> SimulationDataType:
             Object containing simulation results.
         """
 
+        self.upload()
         self.start()
         self.monitor()
         return self.load(path=path)
```
lei-flex, line 205: https://github.com/flexcompute/tidy3d/pull/1647
see the PR above about my comments. 
we could start the download early once the job is done without waiting for all tasks get done. 
the same ideas to upload the tasks as well.  we don't need wait for all files get upload then start run. 
1. the benefit is to reduce the peak concurrency of batch jobs.
2. and reduce the overall time for the whole batch.

tidy3d/plugins/adjoint/web.py
tylerflex: What do you mean by #2? can you provide a code change in your PR?

lei-flex (approved):
ship it.

PR #1632, opened 2024-04-23
:wrench: FIX: bumpmyversion pyproject.toml match

daquinteroflex:
I've tested it and it seems to work as I expect


momchil-flex:
It seems like you are updating the notebooks submodule to a 2.7 state here (there's an EME notebook), but for some reason the test-latest-submodules pass even though you're asking to merge this in develop?

daquinteroflex:
It was a nunanced implementation reason why it was not failing, but it's failing correctly now https://github.com/flexcompute/tidy3d/actions/runs/8817620572/job/24204477926

PR #1634, opened 2024-04-24
Show cost for EME, HEAT, and MODE. Remove URL from EME webapi.

caseyflex:
Tested for HEAT and EME.

### DIFF
--- a/tidy3d/web/api/webapi.py
+++ b/tidy3d/web/api/webapi.py
@@ -29,6 +29,15 @@ RUN_REFRESH_TIME = 1.0
 # file names when uploading to S3
 SIM_FILE_JSON = "simulation.json"
 
+# not all solvers are supported yet in GUI
+GUI_SUPPORTED_TASK_TYPES = ["FDTD", "MODE_SOLVER", "HEAT"]
+
+# if a solver is in beta stage, cost is subject to change
+BETA_TASK_TYPES = ["HEAT", "EME"]
+
+# map task_type to solver name for display
+SOLVER_NAME = {"FDTD": "FDTD", "HEAT": "Heat", "MODE_SOLVER": "Mode", "EME": "EME"}
+
 
 def _get_url(task_id: str) -> str:
     """Get the URL for a task on our server."""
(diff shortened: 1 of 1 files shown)


tidy3d/web/api/webapi.py
```diff
@@ -222,8 +222,9 @@ def upload(
                 "Tidy3D's EME solver is currently in the beta stage. Cost of EME simulations "
                 "is subject to change in the future."
             )
-        url = _get_url(task.task_id)
-        console.log(f"View task using web UI at [link={url}]'{url}'[/link].")
+        if task_type == "FDTD" or task_type == "MODE_SOLVER" or task_type == "HEAT":
```
momchil-flex: How about we define `GUI_SUPPORTED_TASK_TYPES = ["FDTD", "MODE_SOLVER", "HEAT"]` on top and then here do `if task_type in GUI_SUPPORTED_TASK_TYPES`?

tidy3d/web/api/webapi.py
```diff
@@ -366,6 +367,19 @@ def monitor(task_id: TaskId, verbose: bool = True) -> None:
...
+                "or abort/delete the task in the web "
+                "UI. Terminating the Python script will not stop the job running on the cloud."
+            )
+            est_flex_unit = task_info.estFlexUnit
+            if est_flex_unit is not None and est_flex_unit > 0:
+                console.log(
+                    f"Maximum FlexCredit cost: {est_flex_unit:1.3f}. Use 'web.real_cost(task_id)'"
+                    f" to get the billed FlexCredit cost after a simulation run."
+                )
+
```
momchil-flex, line 393: Is this exactly the same code as in `FDTD`? If so let's put it in a separate function to call in both places.

momchil-flex (commented):
Thanks! I've merged your webapi PR and redeployed the service - give it a try now!

However, what exactly was happening before? In principle if the webapi errors, a metadata pipeline is meant to be kicked off such that the estimate should always eventually appear. Was that not what you were seeing?

My other comments are a bit nit-picky, but it would be great to avoid some code duplication.

How about we define a `SOLVER_NAME` dict that links `task_type` to the name we want printed in log messages. Then we can have a single message [here](https://github.com/flexcompute/tidy3d/blob/3339b3923b4fea6192027804ee50b7574a6f7259/tidy3d/web/api/webapi.py#L215-L224) and also replace [this](https://github.com/flexcompute/tidy3d/blob/3339b3923b4fea6192027804ee50b7574a6f7259/tidy3d/web/api/webapi.py#L360-L365) with just `solver_name = SOLVER_NAME[task_type]`?

caseyflex:
> Thanks! I've merged your webapi PR and redeployed the service - give it a try now!
> 
> However, what exactly was happening before? In principle if the webapi errors, a metadata pipeline is meant to be kicked off such that the estimate should always eventually appear. Was that not what you were seeing?
> 
> My other comments are a bit nit-picky, but it would be great to avoid some code duplication.
> 
> How about we define a `SOLVER_NAME` dict that links `task_type` to the name we want printed in log messages. Then we can have a single message [here](https://github.com/flexcompute/tidy3d/blob/3339b3923b4fea6192027804ee50b7574a6f7259/tidy3d/web/api/webapi.py#L215-L224) and also replace [this](https://github.com/flexcompute/tidy3d/blob/3339b3923b4fea6192027804ee50b7574a6f7259/tidy3d/web/api/webapi.py#L360-L365) with just `solver_name = SOLVER_NAME[task_type]`?



I see. When I called `web.estimate_cost`, I got an estimated cost of 0.0 and `metadataStatus='success'`, even though the metadata pipeline is giving the correct estimated cost. I don't know if this is using the webapi or some other web service. This is actually still the case, so I'm not sure why the estimated cost is 0. It doesn't appear to be running the metadata pipeline in the `estimate_cost` function.

I see that now when I call `web.run` with `verbose=True`, it shows the correct behavior. estimated cost is 0 and `metadataStatus` is processing, then after a while estimated cost is correct and `metadataStatus` is `processed`. This must be because the metadataPipeline is working correctly.

I converted this to a draft, I need to test a bit more.

caseyflex:
@momchil-flex 
I unified the treatment a bit between the solvers. It looked like the non-FDTD solver status loop was just a subset of the FDTD solver status loop. The unified logic seems to work well, and makes it easier to implement stuff like EME progress bar in the future. Could you check if this looks ok? If so, the PR should be good to merge.

momchil-flex:
Looks good, played around a bit and it worked well. Hopefully no major issues will come up. :)

PR #1636, opened 2024-04-24
Various improvements to EME solver.

caseyflex:
Summary of changes:

- Adds `EMEFreqSweep` with `freq_scale_factors`
- Adds mode solver relative to existing basis via `_data_on_yee_grid_relative`
- Changes mode solver eigenvalue formula in `eigs_to_effective_index` to avoid small `keff` approximation
- Adds `sweep_index` to more EME data structures. Rearranges some of the other dims to be more consistent and user-friendly.
- Handling of `sweep_index` in `smatrix_in_basis` and `field_in_basis`
- Reorganizes some of the plotting functionality in `sim_data.py` to avoid code duplication. The main thing here was wanting a function `plot_field_monitor_data` that takes the monitor data directly as an argument instead of taking the monitor name.
- Adds `normalize` field to `EMESimulation` which specifies whether to normalize the port modes to unity flux (thereby normalizing the scattering matrix and expansion coefficients). `True` by default, but can be set to `False` in certain cases involving evanescent port modes.
- Other minor changes based on feedback: improved monitor storage size estimation, better handling of lists / np arrays, warning for large number of modes with constraint, convenience methods for port modes, cell-dependent `EMELengthSweep`

Todo:

- [x] improve test coverage
- [x] replace numpy operations with xarray operations in smatrix_in_basis and field_in_basis

Later:

- further improve code quality in smatrix_in_basis / field_in_basis

### DIFF
--- a/tests/test_components/test_eme.py
+++ b/tests/test_components/test_eme.py
@@ -21,7 +21,7 @@ def make_eme_sim():
     freqs = [freq0]
     sim_size = 3 * lambda0, 3 * lambda0, 3 * lambda0
     waveguide_size = (lambda0 / 2, lambda0, td.inf)
-    min_steps_per_wvl = 20
+    min_steps_per_wvl = 10
 
     # EME parameters
     monitor_size = (2 * lambda0, 2 * lambda0, 0.1 * lambda0)
@@ -302,11 +302,11 @@ def test_eme_simulation(log_capture):  # noqa: F811
     with AssertLogLevel(log_capture, "INFO"):
         sim = sim.updated_copy(grid_spec=grid_spec)
     # multiple freqs are ok, but not for autogrid
-    _ = sim.updated_copy(grid_spec=td.GridSpec.uniform(dl=1), freqs=[1e10] + sim.freqs)
+    _ = sim.updated_copy(grid_spec=td.GridSpec.uniform(dl=0.2), freqs=[1e10] + list(sim.freqs))
     with pytest.raises(SetupError):
         _ = td.EMESimulation(
             size=sim.size,
-            freqs=sim.freqs + [1e10],
+            freqs=list(sim.freqs) + [1e10],
             monitors=sim.monitors,
             structures=sim.structures,
             grid_spec=grid_spec,
@@ -315,12 +315,15 @@ def test_eme_simulation(log_capture):  # noqa: F811
         )
 
     # test port offsets
-    with pytest.raises(pd.ValidationError):
+    with pytest.raises(SetupError):
         _ = sim.updated_copy(port_offsets=[sim.size[sim.axis] * 2 / 3, sim.size[sim.axis] * 2 / 3])
 
     # test duplicate freqs
     with pytest.raises(pd.ValidationError):
-        _ = sim.updated_copy(freqs=sim.freqs + sim.freqs)
+        _ = sim.updated_copy(freqs=list(sim.freqs) + list(sim.freqs))
+    # test empty freqs
+    with pytest.raises(pd.ValidationError):
+        _ = sim.updated_copy(freqs=[])
 
     # test unsupported media
     # fully anisotropic
@@ -407,27 +410,85 @@ def test_eme_simulation(log_capture):  # noqa: F811
     with pytest.raises(SetupError):
         _ = sim.updated_copy(size=(1000, 1000, 1000))
     with pytest.raises(SetupError):
+        _ = sim.updated_copy(size=(1000, 500, 3), monitors=[], store_port_modes=True)
+    with pytest.raises(SetupError):
+        _ = sim.updated_copy(size=(1000, 500, 3), monitors=[], store_port_modes=False)
+    with AssertLogLevel(log_capture, "WARNING", "slow-down"):
+        _ = sim.updated_copy(size=(500, 500, 3), monitors=[])
+
+    with pytest.raises(SetupError):
         _ = sim.updated_copy(
-            freqs=sim.freqs + list(1e14 * np.linspace(1, 2, 1000)),
+            freqs=list(sim.freqs) + list(1e14 * np.linspace(1, 2, 1000)),
             grid_spec=sim.grid_spec.updated_copy(wavelength=1),
         )
     large_monitor = sim.monitors[2].updated_copy(size=(td.inf, td.inf, td.inf))
+    _ = sim.updated_copy(
+        size=(10, 10, 10),
+        monitors=[large_monitor],
+        freqs=list(1e14 * np.linspace(1, 2, 1)),
+        grid_spec=sim.grid_spec.updated_copy(wavelength=1),
+    )
+    with AssertLogLevel(log_capture, "WARNING", contains_str="estimated storage"):
+        _ = sim.updated_copy(
+            size=(10, 10, 10),
+            monitors=[large_monitor],
+            freqs=list(1e14 * np.linspace(1, 2, 5)),
+            grid_spec=sim.grid_spec.updated_copy(wavelength=1),
+        )
     with pytest.raises(SetupError):
         _ = sim.updated_copy(
             size=(10, 10, 10),
             monitors=[large_monitor],
-            freqs=list(1e14 * np.linspace(1, 2, 50)),
+            freqs=list(1e14 * np.linspace(1, 2, 20)),
             grid_spec=sim.grid_spec.updated_copy(wavelength=1),
         )
+    with pytest.raises(SetupError):
+        _ = sim.updated_copy(
+            size=(10, 10, 10),
+            monitors=[large_monitor, large_monitor.updated_copy(name="lmon2")],
+            freqs=list(1e14 * np.linspace(1, 2, 5)),
+            grid_spec=sim.grid_spec.updated_copy(wavelength=1),
+        )
+
+    # test monitor that does not intersect any EME cells
+    mode_monitor = td.EMEModeSolverMonitor(
+        size=(0.1, 0.1, 0.1),
+        center=(0, 0, -1.5),
+        name="modes",
+    )
+    with pytest.raises(SetupError):
+        _ = sim.updated_copy(monitors=[mode_monitor], port_offsets=(0.5, 0.5))
+    # test eme cell interval space
+    mode_monitor = mode_monitor.updated_copy(
+        size=(td.inf, td.inf, td.inf), eme_cell_interval_space=8
+    )
+    sim2 = sim.updated_copy(monitors=[mode_monitor])
+    assert sim2._monitor_num_eme_cells(monitor=mode_monitor) == 2
 
     # test monitor num modes
     sim_tmp = sim.updated_copy(monitors=[sim.monitors[0].updated_copy(num_modes=1)])
     assert sim_tmp._monitor_num_modes_cell(monitor=sim_tmp.monitors[0], cell_index=0) == 1
 
+    # test monitor num freqs
+    sim_tmp = sim.updated_copy(monitors=[sim.monitors[0].updated_copy(freqs=[sim.freqs[0]])])
+    assert sim_tmp._monitor_num_freqs(monitor=sim_tmp.monitors[0]) == 1
+
     # test sweep
     _ = sim.updated_copy(sweep_spec=td.EMELengthSweep(scale_factors=list(np.linspace(1, 2, 10))))
+    _ = sim.updated_copy(
+        sweep_spec=td.EMELengthSweep(
+            scale_factors=np.stack((np.linspace(1, 2, 7), np.linspace(1, 2, 7)))
+        )
+    )
+    with pytest.raises(SetupError):
+        _ = sim.updated_copy(sweep_spec=td.EMELengthSweep(scale_factors=[]))
+    # second shape of length sweep must equal number of cells
+    with pytest.raises(SetupError):
+        _ = sim.updated_copy(sweep_spec=td.EMELengthSweep(scale_factors=np.array([[1, 2], [3, 4]])))
     _ = sim.updated_copy(sweep_spec=td.EMEModeSweep(num_modes=list(np.arange(1, 5))))
     # test sweep size limit
+    with pytest.raises(SetupError):
+        _ = sim.updated_copy(sweep_spec=td.EMELengthSweep(scale_factors=[]))
     with pytest.raises(SetupError):
         _ = sim.updated_copy(
             sweep_spec=td.EMELengthSweep(scale_factors=list(np.linspace(1, 2, 200)))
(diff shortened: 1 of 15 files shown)


tests/test_plugins/test_mode_solver.py
```diff
@@ -812,7 +812,7 @@ def test_mode_solver_nan_pol_fraction():
     md = ms.solve()
     check_ms_reduction(ms)
 
-    assert list(np.where(np.isnan(md.pol_fraction.te))[1]) == [8, 9]
+    assert list(np.where(np.isnan(md.pol_fraction.te))[1]) == [9]
```
caseyflex, line 815: @dbochkov-flexcompute I see you added this test. I changed the formula in `eigs_to_effective_index`, and as a result sometimes evanescent modes which were previously `nan` are no longer `nan`. As a result I had to modify this test. Could you take a look and see if this makes sense?

caseyflex:
@tylerflex the test `tests/test_plugins/test_adjoint.py::test_no_poynting` is failing in this branch -- somehow the `NotImplementedError` in `JaxSimulationData.get_poynting_vector` is not being raised. Could you take a look? I tried to debug but I'm not too familiar with the adjoint tests.

tidy3d/components/data/sim_data.py
```diff
@@ -191,15 +250,15 @@ def _get_scalar_field(
         """
 
         if field_name[0] == "S":
-            dataset = self.get_poynting_vector(field_monitor_name)
+            dataset = self._get_poynting_vector(field_monitor_data)
```
tylerflex, line 263: @caseyflex the test fails because it asserts that `_get_scalar_field()` calls `get_poynting_vector` and therefore errors because this method raises a  `NotImplementedError` in `JaxSimulationData`.

You can safely comment out the last two lines in the adjoint test

```py
    with pytest.raises(NotImplementedError):
        sim_data._get_scalar_field(mnt_name_differentiable, "S", "abs")
```

I think they are a bit unnecessary.

tylerflex:
@caseyflex due to some subtleties with how `JaxSimulationData` raises `NotImplementedError` when calling `get_poynting_vector` and now a lot of simulation data stuff calls `_get_poynting_vector` internally. 

See the fix here:
https://github.com/flexcompute/tidy3d/pull/1636/files#r1619410128

tylerflex:
by the way, should I just review once the to-do items are done and it's in it's final state?

caseyflex:
You could wait on smatrix_in_basis and field_in_basis. Everything else
should be pretty final already.

tylerflex:
ok I guess ping me when it's done, thanks

tidy3d/components/data/sim_data.py
caseyflex: thanks!

caseyflex:
> ok I guess ping me when it's done, thanks

I think it is all ready for review now

tests/test_components/test_eme.py
```diff
@@ -431,36 +444,100 @@ def test_eme_simulation(log_capture):  # noqa: F811
     with pytest.raises(SetupError):
         _ = sim.updated_copy(sweep_spec=td.EMEModeSweep(num_modes=list(np.arange(150, 200))))
 
+    # warn about num modes with constraint
+    with AssertLogLevel(log_capture, "INFO"):
+        _ = sim.updated_copy(
+            constraint="passive",
+            eme_grid_spec=td.EMEUniformGrid(num_cells=1, mode_spec=td.EMEModeSpec(num_modes=40)),
+        )
+        _ = sim.updated_copy(
```
tylerflex, line 507: which command is supposed to error? it would be better to isolate one (or both) in their own AssertLogLevel context.

tests/test_components/test_eme.py
```diff
@@ -431,36 +444,100 @@ def test_eme_simulation(log_capture):  # noqa: F811
     with pytest.raises(SetupError):
         _ = sim.updated_copy(sweep_spec=td.EMEModeSweep(num_modes=list(np.arange(150, 200))))
 
+    # warn about num modes with constraint
```
tylerflex: says warn but log capture is "INFO"? so this test might not be catching what you expect. Might be useful to add `contains_str="string match"` to make sure it catches the warning message you want.

tests/test_components/test_eme.py
tylerflex, line 1: I ran a code coverage test. Overall it looks great, but it does seem a lot of edge cases are not tested. Attaching one page from the report (couldn't upload html). It might be worth testing these edge cases to avoid issues in the future.

Not there are also some edge cases untested in sim_data.py and sweep.py but most in simulation.py

[coverage_eme.pdf](https://github.com/user-attachments/files/15514529/coverage_eme.pdf)

tidy3d/components/eme/data/sim_data.py
```diff
@@ -52,15 +53,78 @@ def _extract_mode_solver_data(
...
         )
         grid_expanded = self.simulation.discretize_monitor(monitor=monitor)
         return ModeSolverData(**update_dict, monitor=monitor, grid_expanded=grid_expanded)
 
+    @cached_property
+    def port_modes_tuple(self) -> List[Tuple[ModeSolverData, ModeSolverData]]:
+        """Port modes as a tuple ``(port_modes_1, port_modes_2)``."""
+        if self.port_modes is None:
+            raise SetupError(
+                "The field 'port_modes' is 'None'. Please set 'store_port_modes' "
```
tylerflex, line 85: maybe specify where to set this `store_port_modes`. Do you need to update it in the `EMESimulation` and re-run or just make an updated copy of the `EMESimulationData`?

same comment below

tidy3d/components/eme/data/sim_data.py
```diff
@@ -89,109 +153,107 @@ def smatrix_in_basis(
...
+
+            S11 = S11.sel(mode_index_in=keep_mode_inds1, mode_index_out=keep_mode_inds1).to_numpy()
+            S12 = S12.sel(mode_index_in=keep_mode_inds2, mode_index_out=keep_mode_inds1).to_numpy()
+            S21 = S21.sel(mode_index_in=keep_mode_inds1, mode_index_out=keep_mode_inds2).to_numpy()
+            S22 = S22.sel(mode_index_in=keep_mode_inds2, mode_index_out=keep_mode_inds2).to_numpy()
+
+            if self.simulation._sweep_modes:
+                port_modes1, port_modes2 = self.port_modes_list[sweep_index]
+
+            if modes1_provided:
```
tylerflex, line 228: This bit is a bit hard to follow.  I wonder specifically if the [xarray.DataArray.dot](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.dot.html) could be used to help with the calculations below, rather than doing it in numpy and having to keep track of axes.

I think, in general, trying to stay in xarray land throughout this whole computation would be easier to follow and less error prone.

tidy3d/components/eme/data/sim_data.py
```diff
@@ -241,64 +303,92 @@ def field_in_basis(
...
                 eme_port_index=[port_index],
+                mode_index=mode_index,
+            )
+
+        for sweep_index in sweep_indices:
+            if self.simulation._sweep_modes:
+                port_modes = self.port_modes_list[sweep_index][port_index]
+            if modes_provided:
+                overlaps = modes.outer_dot(port_modes, conjugate=False).to_numpy()
+                if not modes_present:
```
tylerflex, line 386: similar comment, can we do this within the xarray API to make this independent of axes ordering?

tidy3d/components/eme/monitor.py
```diff
@@ -59,14 +71,23 @@ class EMEMonitor(AbstractMonitor, ABC):
 
     @abstractmethod
     def storage_size(
-        self, num_cells: int, num_eme_cells: int, num_freqs: int, num_modes: int
+        self,
+        num_cells: int,
+        num_transverse_cells: int,
+        num_eme_cells: int,
```
tylerflex, line 77: the original `Monitor.storage_size` has a different call signature right? I wonder if we should add a generic `**kwargs` to the call signature there so that technically one could call the original Monitor.storage_size with these extra kwargs (If that situation arises) and not get errors.

tidy3d/components/eme/simulation.py
```diff
@@ -208,28 +220,20 @@ class EMESimulation(AbstractYeeGridSimulation):
...
-            )
-        return val
-
     @pd.validator("freqs", always=True)
     def _validate_freqs(cls, val):
         """Freqs cannot contain duplicates."""
+        if len(val) == 0:
+            raise SetupError("Simulation 'freqs' cannot be empty.")
         if len(set(val)) != len(val):
             raise SetupError(f"Simulation 'freqs={val}' cannot " "contain duplicate frequencies.")
```
tylerflex: remove `" "`

tylerflex: `'Simulation'` in quotes

tidy3d/components/eme/simulation.py
```diff
@@ -542,9 +553,34 @@ def _post_init_validators(self) -> None:
...
         # self._warn_monitor_interval()
         log.end_capture(self)
 
+    def _validate_constraint(self):
+        """Constraint can be slow with too many modes. Warn in this case."""
+        constraint = self.constraint
+        max_num_modes = self.max_num_modes
+        if constraint is not None and max_num_modes > WARN_CONSTRAINT_NUM_MODES:
+            log.warning(
+                f"The simulation has 'constraint='{constraint}'', and the maximum "
```
tylerflex: I think the additional single quotes `'` around `{constraint}` will mean this will not render properly. You probably need to escape the single quotes. `\'` but double check this.

tidy3d/components/eme/simulation.py
```diff
@@ -542,9 +553,34 @@ def _post_init_validators(self) -> None:
         self._validate_modes_size()
         self._validate_sweep_spec()
         self._validate_symmetry()
+        self._validate_constraint()
         # self._warn_monitor_interval()
         log.end_capture(self)
 
+    def _validate_constraint(self):
```
tylerflex, line 588: curious why this needs to be a post init validator not a regular one?

tidy3d/components/eme/simulation.py
```diff
@@ -542,9 +553,34 @@ def _post_init_validators(self) -> None:
...
+            log.warning(
+                f"The simulation has 'constraint='{constraint}'', and the maximum "
+                f"number of EME modes in the simulation is '{max_num_modes}'. "
+                f"Using more than '{WARN_CONSTRAINT_NUM_MODES}' modes together with "
+                "a constraint can significantly slow down the simulation. Consider "
+                "reducing the number of modes or setting 'constraint=None'."
+            )
+
+    def _validate_port_offsets(self):
+        """Port offsets cannot jointly exceed simulation length."""
```
tylerflex, line 602: does this need to take `self.center` into account?

Maybe I dont quite understand, is this kind because the ports are guaranteed to be outside of the simulation if their offset is greater than the simulation size?

tidy3d/components/eme/simulation.py
```diff
@@ -582,13 +620,26 @@ def _validate_sweep_spec(self):
...
+                        "Simulation 'sweep_spec.scale_factors' has shape "
+                        f"'{scale_factors_shape}'. The size of the second dimension "
+                        "must equal the number of EME cells in the simulation, which is "
+                        f"'{self.eme_grid.num_cells}'."
+                    )
+        elif isinstance(self.sweep_spec, EMEFreqSweep):
+            for i, scale_factor in enumerate(self.sweep_spec.freq_scale_factors):
+                scaled_freqs = np.array(self.freqs) * scale_factor
+                if np.min(scaled_freqs) < MIN_FREQUENCY:
+                    raise ValidationError(
```
tylerflex: reason for not using `SetupError`? If I recall, there might be some issues with using `ValidationError` handling with our new log / error handling.

tidy3d/components/eme/simulation.py
```diff
@@ -520,17 +524,24 @@ def mode_solver_monitors(self) -> List[ModeSolverMonitor]:
     @property
     def port_modes_monitor(self) -> EMEModeSolverMonitor:
         """EME Mode solver monitor for only the port modes."""
+        if self._sweep_modes:
```
tylerflex: how does this fit into `self._num_sweep`? seems like you can just use it directly?

tidy3d/components/eme/simulation.py
```diff
@@ -887,3 +981,24 @@ def grid(self) -> Grid:
...
+                # Monitor is outside of simulation domain; can happen e.g. for integration surfaces
+                return 0
+            num_cells = self.discretize_monitor(monitor).num_cells
+            # take monitor downsampling into account
+            num_cells = list(monitor.downsampled_num_cells(num_cells))
+            # pop propagation axis
+            num_cells.pop(self.axis)
+            return np.prod(np.array(num_cells, dtype=np.int64))
+
+        # if isinstance(monitor, SurfaceIntegrationMonitor):
```
tylerflex: remove comments?

tidy3d/components/eme/simulation.py
```diff
@@ -887,3 +981,24 @@ def grid(self) -> Grid:
...
+            monitor given the simulation grid and downsampling."""
+            if not self.intersects(monitor):
+                # Monitor is outside of simulation domain; can happen e.g. for integration surfaces
+                return 0
+            num_cells = self.discretize_monitor(monitor).num_cells
+            # take monitor downsampling into account
+            num_cells = list(monitor.downsampled_num_cells(num_cells))
+            # pop propagation axis
+            num_cells.pop(self.axis)
+            return np.prod(np.array(num_cells, dtype=np.int64))
```
tylerflex, line 997: maybe
```py
np.prod(np.array(num_cells)).astype(int)
```
unless you want `np.int64` specifically?

tidy3d/plugins/mode/solver.py
```diff
@@ -212,6 +216,26 @@ def compute_modes(
...
             target_neff_p *= 1 + TARGET_SHIFT
 
+        # preprocess solver_basis_fields
+        basis_E = None
+        if solver_basis_fields is not None:
+            basis_E = solver_basis_fields[:3, ...]
+            try:
+                basis_E = basis_E.reshape((3, Nx * Ny, num_modes))
+            except ValueError as e:
+                log.error(
```
tylerflex: let's raise an exception (maybe another ValueError) instead of `log.error()`.

tylerflex (changes requested):
Thanks @caseyflex . I think this looks good overall. I left some comments. 

In general it seems there are a ton of edge cases being handled here. While I'm sure it's correct, it was a bit difficult to follow the complexity in the code when reading through. More importantly, though, I am a little concerned about making sure the edge cases are tested properly to avoid unexpected issues in use. I think adding some unit tests for these various edge would go a long way towards the stability of the code.

The command I use to test coverage is:

`pytest -rA tests --cov-report=html --cov=tidy3d && open htmlcov/index.html`

must also `pip install pytest-cov`

Another general comment is that I would suggest trying to use `xarray` instead of `to_numpy()` and using axis manipulation. One can imagine in the future if the order of some of the data dimensions gets changed, or we add another dimension, this can quickly become unsustainable. So if there's any way to rewrite the array manipulation bits using the named axes of xarray, that would also really improve the stability and readability of the code.

Thanks a lot for these improvements to the EME solver and all of the hard work to get them out!

momchil-flex:
When I run through the web, it seems that the task monitoring stays stuck at "starting up solver" until it eventually goes to "success". This may not be something introduced by this PR but worth noting to fix.

image

momchil-flex:
So currently EMESimulation supports both `freqs` and a `sweep_spec` that can be a frequency sweep, right? Providing multiple `freqs` will compute the modes exactly, while providing a frequency sweep spec will use the mode computation in the reduced basis? What happens if both are provided?

caseyflex:
> So currently EMESimulation supports both `freqs` and a `sweep_spec` that can be a frequency sweep, right? Providing multiple `freqs` will compute the modes exactly, while providing a frequency sweep spec will use the mode computation in the reduced basis? What happens if both are provided?

Yeah, that's right -- I should clarify in the docstrings. When both are provided, it kind of works as expected -- performs exact mode solves at all the `freqs`, and then applies the `freq_scale_factors` to each exact solution. Although this would definitely be an obscure use case.

tidy3d/components/eme/monitor.py
caseyflex: I'm not sure, it might be better to get an error if you pass something like `num_eme_cells` to a monitor which is not an `EMEMonitor`. Likewise for all these other parameters. And `tmesh` doesn't make sense as a parameter for an `EMEMonitor`. So I think it's better to have different function signatures.

tidy3d/components/eme/simulation.py
caseyflex: It needs the `sim.eme_grid` to already be initialized, in order to check the max number of modes

tidy3d/components/eme/simulation.py
caseyflex: The `port_offsets` defines the amount by which the ports (the beginning and end of the EME grid) are offset inward from the simulation boundaries. If the sum of the offsets is larger than the simulation size, the EME grid will have negative extent.

tidy3d/components/eme/simulation.py
caseyflex: this was copied from `Simulation`. I'm not sure why int64 was used there, but it is probably safer in case the simulation is very large.

momchil-flex: Yeah, there was a reason for that. Some machines (I think I've only seen it on Windows) seem to run 32-bit numpy and `np.prod(np.array(num_cells))` can overflow.

yaugenst-flex:
@caseyflex fyi, autograd tests are failing because of #1734 and will pass once #1736 is merged

tests/test_plugins/test_mode_solver.py
dbochkov-flexcompute: yeah, it seems fine. This test just checks that if there are modes with nan's in pol_fraction, they are put at the end of list

tylerflex (approved):
As far as I'm concerned, this is good to go assuming whatever comments I had earlier are either resolved or marked somewhere for revisiting whenever possible. Thanks @caseyflex

PR #1638, opened 2024-04-25
Heat solver improvements

dbochkov-flexcompute:
changes associated with backed https://github.com/flexcompute/tidy3d-core/pull/616 and additional ones:
- allow any geometry type in heat simulations
- `relative_min_dl` parameter to restrict minimal mesh size
- missing pydantic discriminators causing false warnings
- functions to clean unstructured data from degenerate cells and unused points
- safeguards in spatial data arrays against unsorted data
- validators to check that monitor cross solid medium
- correct way to handle monitors overlapping symmetry regions
- more careful slicing of unstructured datasets along edges

fyi @e-g-melo

### DIFF
--- a/tests/test_components/test_heat.py
+++ b/tests/test_components/test_heat.py
@@ -27,7 +27,13 @@ from tidy3d import TemperatureMonitor
 from tidy3d import TemperatureData
 from tidy3d.exceptions import DataError
 
-from ..utils import STL_GEO, assert_log_level, log_capture
+from ..utils import (
+    STL_GEO,
+    assert_log_level,
+    log_capture,
+    AssertLogLevel,
+    cartesian_to_unstructured,
+)
 
 
 def make_heat_mediums():
@@ -105,16 +111,22 @@ def test_heat_bcs():
 
 
 def make_heat_mnts():
-    temp_mnt1 = TemperatureMonitor(size=(1, 2, 3), name="test")
-    temp_mnt2 = TemperatureMonitor(size=(1, 2, 3), name="tet", unstructured=True)
-    temp_mnt3 = TemperatureMonitor(size=(1, 0, 3), name="tri", unstructured=True, conformal=True)
-    temp_mnt4 = TemperatureMonitor(size=(1, 0, 3), name="empty", unstructured=True, conformal=False)
+    temp_mnt1 = TemperatureMonitor(size=(1.6, 2, 3), name="test")
+    temp_mnt2 = TemperatureMonitor(size=(1.6, 2, 3), name="tet", unstructured=True)
+    temp_mnt3 = TemperatureMonitor(
+        center=(0, 1, 0), size=(1.6, 0, 3), name="tri", unstructured=True, conformal=True
+    )
+    temp_mnt4 = TemperatureMonitor(
+        center=(0, 1, 0), size=(1.6, 0, 3), name="empty", unstructured=True, conformal=False
+    )
+    temp_mnt5 = TemperatureMonitor(center=(0, 0.7, 0.8), size=(3, 0, 0), name="line")
+    temp_mnt6 = TemperatureMonitor(center=(0.7, 0.6, 0.8), size=(0, 0, 0), name="point")
 
-    return (temp_mnt1, temp_mnt2, temp_mnt3, temp_mnt4)
+    return (temp_mnt1, temp_mnt2, temp_mnt3, temp_mnt4, temp_mnt5, temp_mnt6)
 
 
 def test_heat_mnt():
-    temp_mnt, _, _, _ = make_heat_mnts()
+    temp_mnt, _, _, _, _, _ = make_heat_mnts()
 
     with pytest.raises(pd.ValidationError):
         _ = temp_mnt.updated_copy(name=None)
@@ -124,7 +136,7 @@ def test_heat_mnt():
 
 
 def make_heat_mnt_data():
-    temp_mnt1, temp_mnt2, temp_mnt3, temp_mnt4 = make_heat_mnts()
+    temp_mnt1, temp_mnt2, temp_mnt3, temp_mnt4, temp_mnt5, temp_mnt6 = make_heat_mnts()
 
     nx, ny, nz = 9, 6, 5
     x = np.linspace(0, 1, nx)
@@ -188,7 +200,27 @@ def make_heat_mnt_data():
 
     mnt_data4 = TemperatureData(monitor=temp_mnt4, temperature=None)
 
-    return (mnt_data1, mnt_data2, mnt_data3, mnt_data4)
+    nx, ny, nz = 9, 1, 1
+    x = np.linspace(0, 1, nx)
+    y = np.linspace(0, 2, ny)
+    z = np.linspace(0, 3, nz)
+    T = np.random.default_rng().uniform(300, 350, (nx, ny, nz))
+    coords = dict(x=x, y=y, z=z)
+    temperature_field = td.SpatialDataArray(T, coords=coords)
+
+    mnt_data5 = TemperatureData(monitor=temp_mnt5, temperature=temperature_field)
+
+    nx, ny, nz = 1, 1, 1
+    x = np.linspace(0, 1, nx)
+    y = np.linspace(0, 2, ny)
+    z = np.linspace(0, 3, nz)
+    T = np.random.default_rng().uniform(300, 350, (nx, ny, nz))
+    coords = dict(x=x, y=y, z=z)
+    temperature_field = td.SpatialDataArray(T, coords=coords)
+
+    mnt_data6 = TemperatureData(monitor=temp_mnt6, temperature=temperature_field)
+
+    return (mnt_data1, mnt_data2, mnt_data3, mnt_data4, mnt_data5, mnt_data6)
 
 
 def test_heat_mnt_data():
@@ -196,12 +228,14 @@ def test_heat_mnt_data():
 
 
 def make_uniform_grid_spec():
-    return UniformUnstructuredGrid(dl=0.1, min_edges_per_circumference=5, min_edges_per_side=3)
+    return UniformUnstructuredGrid(
+        dl=0.1, min_edges_per_circumference=5, min_edges_per_side=3, relative_min_dl=1e-3
+    )
 
 
 def make_distance_grid_spec():
     return DistanceUnstructuredGrid(
-        dl_interface=0.1, dl_bulk=1, distance_interface=1, distance_bulk=2
+        dl_interface=0.1, dl_bulk=1, distance_interface=1, distance_bulk=2, relative_min_dl=1e-5
     )
 
 
@@ -213,12 +247,15 @@ def test_grid_spec():
         _ = grid_spec.updated_copy(min_edges_per_circumference=-1)
     with pytest.raises(pd.ValidationError):
         _ = grid_spec.updated_copy(min_edges_per_side=-1)
+    with pytest.raises(pd.ValidationError):
+        _ = grid_spec.updated_copy(relative_min_dl=-1e-4)
 
     grid_spec = make_distance_grid_spec()
+    _ = grid_spec.updated_copy(relative_min_dl=0)
     with pytest.raises(pd.ValidationError):
-        grid_spec.updated_copy(dl_interface=-1)
+        _ = grid_spec.updated_copy(dl_interface=-1)
     with pytest.raises(pd.ValidationError):
-        grid_spec.updated_copy(distance_interface=2, distance_bulk=1)
+        _ = grid_spec.updated_copy(distance_interface=2, distance_bulk=1)
 
 
 def make_heat_source():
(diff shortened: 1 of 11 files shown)


tidy3d/components/data/data_array.py
```diff
@@ -265,9 +265,23 @@ class AbstractSpatialDataArray(DataArray, ABC):
     _dims = ("x", "y", "z")
     _data_attrs = {"long_name": "field value"}
 
+    @property
+    def _spatially_sorted(self) -> SpatialDataArray:
+        """Check whether sorted and sort if not."""
+        needs_sorting = []
+        for axis in "xyz":
+            if np.any(self.coords[axis].values[1:] < self.coords[axis].values[:-1]):
```
tylerflex: does this fail if the array has 1 data point along an axis?

tidy3d/components/data/data_array.py
```diff
@@ -280,10 +294,17 @@ def sel_inside(self, bounds: Bound) -> SpatialDataArray:
         SpatialDataArray
             Extracted spatial data array.
         """
+        if any(bmin > bmax for bmin, bmax in zip(*bounds)):
+            raise DataError(
+                "Min and max bounds must be packaged as ``(minx, miny, minz), (maxx, maxy, maxz)``."
```
tylerflex: put the code in normal single quotes for exceptions `'` as it gets printed to the terminal.

Same comment for similar errors below

tidy3d/components/data/data_array.py
```diff
@@ -356,33 +381,47 @@ class SpatialDataArray(AbstractSpatialDataArray):
...
+        data = np.array(sorted_self.data)
+
+        data_left_bound = coords[axis].data[0]
 
-        if np.isclose(center, coords[axis].data[0]):
+        if np.isclose(center, data_left_bound):
             num_duplicates = 1
-        elif center > coords[axis].data[0]:
+        elif center > data_left_bound:
             raise DataError("Reflection center must be outside and on the left of the data region.")
```
tylerflex: "on the left" -> "to the left" sounds more natural. (maybe let's just change it here while we're already doing some construction on DataArray)

tidy3d/components/data/dataset.py
```diff
@@ -550,14 +550,41 @@ def _double_type(self):
...
-        # using val.astype(np.int32/64) directly causes issues when dataarray are later checked ==
-        return CellDataArray(val.data.astype(vtk["id_type"], copy=False), coords=val.coords)
+    @pd.validator("points", always=True)
+    def points_right_indexing(cls, val):
+        """Check that points are indexed corrrectly."""
+        indices_expected = np.arange(len(val.data))
+        indices_given = val.index.data
+        if np.any(indices_expected != indices_given):
+            raise ValidationError(
+                "Points array is expected to be indexed in a simple fashion (0, 1, 2, ...)."
```
tylerflex: Suggest adding a little bit more explanation in case it's not clear how to fix this validator.
```
"'points.index' expects a simple array of indices for each data, eg. '(0, 1, 2, ...)' or 'np.arange(len(points.data))'. "
```

Same comments in the validator below

tylerflex: Also, I'm slightly confused why we need this `index` as a field at all if it seems it only accepts one possible value?

tidy3d/components/data/dataset.py
```diff
@@ -612,24 +637,93 @@ def check_cell_vertex_range(cls, val, values):
...
+            return CellDataArray(
+                data=data, coords=dict(cell_index=cell_index, vertex_index=cells.vertex_index)
+            )
+        return cells
+
+    @classmethod
+    def _remove_unused_points(
+        cls, points: PointDataArray, values: IndexedDataArray, cells: CellDataArray
+    ):
+        """Remove explicitly degenerate cells if any.
```
tylerflex: seems like same docstring as above, maybe `degenerate` should become `unused`?

tidy3d/components/data/dataset.py
```diff
@@ -612,24 +637,93 @@ def check_cell_vertex_range(cls, val, values):
...
+
+    @pd.validator("cells", always=True)
+    def warn_degenerate_cells(cls, val):
+        """Check that cell connections does not have duplicate points."""
+        degenerate_cells = cls._find_degenerate_cells(val)
+        num_degenerate_cells = len(degenerate_cells)
+        if num_degenerate_cells > 0:
+            log.warning(
+                f"Unstructured grid contains {num_degenerate_cells} degenerate cell(s). "
+                "Consider applying 'clean()'."
```
tylerflex: can you elaborate here? where is `clean()` and how would the user specifically call it? (at least add a code snippet, like `dataset = dataset.clean()` or something?

Same comment below

tidy3d/components/data/dataset.py
```diff
@@ -2349,6 +2506,10 @@ def sel_inside(self, bounds: Bound) -> TriangularGridDataset:
         TriangularGridDataset
             Extracted spatial data array.
         """
+        if any(bmin > bmax for bmin, bmax in zip(*bounds)):
+            raise DataError(
+                "Min and max bounds must be packaged as ``(minx, miny, minz), (maxx, maxy, maxz)``."
```
tylerflex, line 2517: single, normal quotes

tidy3d/components/heat/data/monitor_data.py
```diff
@@ -89,23 +90,70 @@ def warn_no_data(cls, val, values):
 
         return val
 
-    @property
+    @cached_property
     def symmetry_expanded_copy(self) -> TemperatureData:
```
tylerflex, line 94: consider breaking into smaller functions if it makes sense to.

tidy3d/components/heat/grid.py
```diff
@@ -9,8 +9,18 @@
 from ...exceptions import ValidationError
 
 
-class UniformUnstructuredGrid(Tidy3dBaseModel):
+class UnstructuredGrid(Tidy3dBaseModel):
+    """Abstract unstructured grid."""
```
tylerflex, line 14: should we make this inherit from `abc.ABC` if it's abstract?

tidy3d/components/heat/grid.py
```diff
@@ -9,8 +9,18 @@
...
+
+    relative_min_dl: pd.NonNegativeFloat = pd.Field(
+        1e-3,
+        title="Relative Mesh Size Limit",
+        description="The minimal allowed mesh size relative to the largest dimension of the simulation domain."
+        "Use 'relative_min_dl=0' to remove this constraint.",
```
tylerflex: these ones use double backslash quotes for the docs RST 😭  it's a bit confusing just catching where I find this.

tidy3d/components/heat/simulation.py
```diff
@@ -116,24 +111,66 @@ def check_unsupported_geometries(cls, val):
...
+    def _check_cross_solids(objs, values):
+        """Given model dictionary ``values``, check whether objects in list ``objs`` cross
+        a ``SolidSpec`` medium.
+        """
 
-            if isinstance(structure.geometry, GeometryGroup):
-                geometries = structure.geometry.geometries
+        # list of structures including background as a Box()
+        structure_bg = Structure(
+            geometry=Box(
```
tylerflex, line 137: just wondering if we have any sort of check in case these are not in `values`. If so whether we should error gracefully. (maybe the validator / methods that call this already handle it?)

tidy3d/components/heat/simulation.py
```diff
@@ -116,24 +111,66 @@ def check_unsupported_geometries(cls, val):
...
+        if val is None:
+            return val
+
+        failed_mnt_inds = cls._check_cross_solids(val, values)
+
+        if len(failed_mnt_inds) > 0:
+            monitor_names = ", ".join(f"'{val[ind].name}'" for ind in failed_mnt_inds)
+            raise SetupError(
+                f"Monitor(s) {monitor_names} do(es) not cross any solid materials ('heat_spec=SolidSpec(...)'). "
+                "Heat equation is solved only inside solid materials. "
```
tylerflex: The heat equation

tidy3d/components/heat/simulation.py
```diff
@@ -116,24 +111,66 @@ def check_unsupported_geometries(cls, val):
...
+
+        if val is None:
+            return val
+
+        failed_mnt_inds = cls._check_cross_solids(val, values)
+
+        if len(failed_mnt_inds) > 0:
+            monitor_names = ", ".join(f"'{val[ind].name}'" for ind in failed_mnt_inds)
+            raise SetupError(
+                f"Monitor(s) {monitor_names} do(es) not cross any solid materials ('heat_spec=SolidSpec(...)'). "
```
tylerflex: really minor, but I think it would be fine to just do something like
```py
monitor_names = [f"'{val[ind].name}'" for ind in failed_mnt_inds)]
raise SetupError(
                f"Monitors {monitor_names} do not cross any solid materials ('heat_spec=SolidSpec(...)'). "
)
```

And just keep everything in plural

tidy3d/components/heat/simulation.py
```diff
@@ -116,24 +111,66 @@ def check_unsupported_geometries(cls, val):
...
+                "Heat equation is solved only inside solid materials. "
+                "Thus, no information will be recorded in this monitor."
```
tylerflex: these monitors? or if you keep it hybrid singular / plural, maybe "this(these) monitor(s)"?

tylerflex (approved):
Thanks @dbochkov-flexcompute . The changes look very good. I can't comment on all of the technical aspects, but had a few minor comments throughout.

tidy3d/components/data/data_array.py
dbochkov-flexcompute: yeah, I think it should, thanks for catching that

tidy3d/components/heat/grid.py
dbochkov-flexcompute: yeah, even though I do remember which quotes to use where, these kind of things still slip in :'(

tidy3d/components/data/data_array.py
dbochkov-flexcompute: huh, surprisingly, it doesn't fail. For a single-element array `[1:]` and `[:-1]` both return empty array `[]`, which I guess makes sense. And comparison between two empty arrays is another empty array. Moreover:
```
np.any([]) -> False
np.all([]) -> True
```

I will still add an explicit check for the length of the array so not to worry about these details

tidy3d/components/data/dataset.py
dbochkov-flexcompute: `index` here is one of the xarray coordinates of `PointDataArray`. But, yeah, we don't need this capability of xarray's to provide flexible coordinate values, it was primarily used to be able to easily write this data in hdf5 using our existing tools. This check is more to prevent users from providing more complicated `index` values and expecting that they would be taken into account in our unstructured datasets. If there is a way to force `PointDataArray` always have certain coordinates, that would be easier. When defining, the easiest way to provide what's needed is to use `PointDataArray(data, dims=["index", "axis"])`

dbochkov-flexcompute: we could consider extending our unstructured datasets to take into account any `index` values, but that probably will not add any useful functionality

tests/test_components/test_heat.py
```diff
@@ -296,23 +333,12 @@ def test_heat_sim():
...
-    vertices = np.array([(0, 0), (1, 0), (1, 1)])
-    p = td.PolySlab(vertices=vertices, axis=2, slab_bounds=(-1, 1))
-    _, structure = make_heat_structures()
-    structure = structure.updated_copy(geometry=p, name="polyslab")
-    _ = heat_sim.updated_copy(structures=list(heat_sim.structures) + [structure])
-
-    # stl support
-    structure = structure.updated_copy(geometry=STL_GEO, name="stl")
-    _ = heat_sim.updated_copy(structures=list(heat_sim.structures) + [structure])
-
```
marc-flex, line 309: Do we not support polyslab and STL any more? Or have I missed something?

tidy3d/components/heat/boundary.py
```diff
@@ -85,9 +86,11 @@ class HeatBoundarySpec(Tidy3dBaseModel):
...
         title="Boundary Conditions Placement",
         description="Location to apply boundary conditions.",
+        discriminator=TYPE_TAG_STR,
     )
 
     condition: HeatBoundaryConditionType = pd.Field(
         title="Boundary Conditions",
         description="Boundary conditions to apply at the selected location.",
+        discriminator=TYPE_TAG_STR,
     )
```
marc-flex, line 96: what was this for?

tidy3d/components/heat/grid.py
```diff
@@ -9,8 +10,18 @@
...
-class UniformUnstructuredGrid(Tidy3dBaseModel):
+class UnstructuredGrid(Tidy3dBaseModel, ABC):
+    """Abstract unstructured grid."""
 
+    relative_min_dl: pd.NonNegativeFloat = pd.Field(
+        1e-3,
+        title="Relative Mesh Size Limit",
+        description="The minimal allowed mesh size relative to the largest dimension of the simulation domain."
+        "Use ``relative_min_dl=0`` to remove this constraint.",
+    )
```
marc-flex, line 21: What happens if the smallest feature is smaller than this relative mesh size? Maybe this would be an edge case?

tidy3d/components/heat/simulation.py
```diff
@@ -222,6 +271,26 @@ def names_exist_grid_spec(cls, val, values):
...
+            desired_min_dl = val.dl
+        if isinstance(val, DistanceUnstructuredGrid):
+            desired_min_dl = min(val.dl_interface, val.dl_bulk)
+
+        if desired_min_dl < min_dl:
+            log.warning(
+                f"The resulting limit for minimal mesh size from parameter 'relative_min_dl={val.relative_min_dl}' is {min_dl}, while provided mesh size in 'grid_spec' is {desired_min_dl}. "
+                "Consider lowering parameter 'relative_min_dl' if a finer grid is required."
+            )
+
```
marc-flex, line 291: I think this answers my previous question.

marc-flex (approved):
Some of these changes (like the symmetry expanded copy and related tests) will need to be extended on the device simulation branch when we merge :'( so I guess we'll have another round later

tidy3d/components/heat/boundary.py
tylerflex: unfortunately, pydantic needs a bit of help to figure out what type to load the field as. This `discriminator` just tells it to load based on the `.type` field, which we automatically add to all tidy3d components.

tests/test_components/test_heat.py
dbochkov-flexcompute: No, still supporting all that. Just now we support all geometry types, so there is no validator that checks type of geometry anymore, and these tests don't really check for anything now

tidy3d/components/heat/simulation.py
dbochkov-flexcompute: It would still be nice to compare this against the smallest geometry feature, but there is no sure way yet on the frontend to determine that as that might come from intersection of geometry objects. We should try to fix this eventually

PR #1639, opened 2024-04-25
Fix near2far for 2d simulations

QimingFlex:
Implemented a preliminary fix for the 'near2far' in 2D. The key differences between 2D and 3D of 'near2far' in math are as follows:

1. Field Components - In 3D, they include Er, Ephi, Etheta, Hr, Hphi, Htheta; in 2D, for TMz mode, they are Ez and Hphi, and for TEz mode, Ephi and Hz.
2. Parameters - In 3D, these are theta, phi, and f; in 2D, phi and f.
3. Propagation phase.
4. RCS coefficient.

Given these differences, my initial idea was to develop a 2D variant of the FieldProjectionAngleMonitor, omitting the theta parameter and inheriting from the AbstractFieldProjectionMonitor. However, since the AbstractFieldProjectionMonitor is heavily designed for 3D features and depends on theta, it seems a 2D monitor would require a ground-up redesign. My current approach involves a rudimentary adaptation where the 3D monitor is used as 2D by setting theta = pi/2, and I added a 2D version of the far-field computation. This mirrors the 3D code with adjustments for the propagation phase and the RCS coefficient. I am relatively new to contributing to a large project, any comments and suggestions on how to more efficiently integrate these changes would be greatly appreciated.

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -2040,6 +2040,12 @@ class AbstractFieldProjectionData(MonitorData):
         discriminator=TYPE_TAG_STR,
     )
 
+    is_2d_simulation: bool = pd.Field(
+        False,
+        title="2D Simulation",
+        description="Indicates whether the monitor data is for a 2D simulation.",
+    )
+
     @property
     def field_components(self) -> Dict[str, DataArray]:
         """Maps the field components to their associated data."""
@@ -2149,10 +2155,14 @@ class AbstractFieldProjectionData(MonitorData):
         return ETA_0 / np.sqrt(eps_complex)
 
     @staticmethod
-    def propagation_phase(dist: Union[float, None], k: complex) -> complex:
-        """Phase associated with propagation of a distance with a given wavenumber."""
+    def propagation_factor(dist: Union[float, None], k: complex, is_2d_simulation: bool) -> complex:
+        """A normalization factor that includes both phase and amplitude decay associated with propagation over a distance with a given wavenumber."""
         if dist is None:
             return 1.0
+
+        if is_2d_simulation:
+            return -np.exp(1j * k * dist) * np.sqrt(1j * k / (8 * np.pi * dist))
+
         return -1j * k * np.exp(1j * k * dist) / (4 * np.pi * dist)
 
     @property
@@ -2232,14 +2242,19 @@ class AbstractFieldProjectionData(MonitorData):
         k = self.k[None, None, None, ...]
         eta = self.eta[None, None, None, ...]
 
-        constant = k**2 / (8 * np.pi * eta)
+        if self.is_2d_simulation:
+            constant = k**2 / (16 * np.pi * eta)
+        else:
+            constant = k**2 / (8 * np.pi * eta)
 
         # normalize fields by the distance-based phase factor
         coords_sph = self.coords_spherical
         if coords_sph["r"] is None:
             phase = 1.0
         else:
-            phase = self.propagation_phase(dist=coords_sph["r"][..., None], k=k)
+            phase = self.propagation_factor(
+                dist=coords_sph["r"][..., None], k=k, is_2d_simulation=self.is_2d_simulation
+            )
         Etheta = self.Etheta.values / phase
         Ephi = self.Ephi.values / phase
         rcs_data = constant * (np.abs(Etheta) ** 2 + np.abs(Ephi) ** 2)
@@ -2351,10 +2366,14 @@ class FieldProjectionAngleData(AbstractFieldProjectionData):
 
         # the phase factor associated with the old distance must be removed
         r = self.coords_spherical["r"][..., None]
-        old_phase = self.propagation_phase(dist=r, k=self.k[None, None, None, :])
+        old_phase = self.propagation_factor(
+            dist=r, k=self.k[None, None, None, :], is_2d_simulation=self.is_2d_simulation
+        )
 
         # the phase factor associated with the new distance must be applied
-        new_phase = self.propagation_phase(dist=proj_distance, k=self.k)
+        new_phase = self.propagation_factor(
+            dist=proj_distance, k=self.k, is_2d_simulation=self.is_2d_simulation
+        )
 
         # net phase
         phase = new_phase[None, None, None, :] / old_phase
@@ -2470,7 +2489,7 @@ class FieldProjectionCartesianData(AbstractFieldProjectionData):
         # the phase factor associated with the old distance must be removed
         k = self.k[None, None, None, :]
         r = self.coords_spherical["r"][..., None]
-        old_phase = self.propagation_phase(dist=r, k=k)
+        old_phase = self.propagation_factor(dist=r, k=k, is_2d_simulation=self.is_2d_simulation)
 
         # update the field components' projection distance
         norm_dir, _ = self.monitor.pop_axis(["x", "y", "z"], axis=self.monitor.proj_axis)
@@ -2479,7 +2498,7 @@ class FieldProjectionCartesianData(AbstractFieldProjectionData):
 
         # the phase factor associated with the new distance must be applied
         r = self.coords_spherical["r"][..., None]
-        new_phase = self.propagation_phase(dist=r, k=k)
+        new_phase = self.propagation_factor(dist=r, k=k, is_2d_simulation=self.is_2d_simulation)
 
         # net phase
         phase = new_phase / old_phase
@@ -2593,10 +2612,14 @@ class FieldProjectionKSpaceData(AbstractFieldProjectionData):
 
         # the phase factor associated with the old distance must be removed
         r = self.coords_spherical["r"][..., None]
-        old_phase = self.propagation_phase(dist=r, k=self.k[None, None, None, :])
+        old_phase = self.propagation_factor(
+            dist=r, k=self.k[None, None, None, :], is_2d_simulation=self.is_2d_simulation
+        )
 
         # the phase factor associated with the new distance must be applied
-        new_phase = self.propagation_phase(dist=proj_distance, k=self.k)
+        new_phase = self.propagation_factor(
+            dist=proj_distance, k=self.k, is_2d_simulation=self.is_2d_simulation
+        )
 
         # net phase
         phase = new_phase[None, None, None, :] / old_phase
--- a/tidy3d/components/field_projection.py
+++ b/tidy3d/components/field_projection.py
@@ -98,6 +98,11 @@ class FieldProjector(Tidy3dBaseModel):
         "surface current densities.",
     )
 
+    @cached_property
+    def is_2d_simulation(self) -> bool:
+        non_zero_dims = sum(1 for size in self.sim_data.simulation.size if size != 0)
+        return non_zero_dims == 2
+
     @pydantic.validator("origin", always=True)
     @skip_if_fields_missing(["surfaces"])
     def set_origin(cls, val, values):
@@ -344,11 +349,20 @@ class FieldProjector(Tidy3dBaseModel):
                 colocation_points[idx] = points
 
         for idx, points in enumerate(colocation_points):
-            if np.array(points).size == 1:
+            if np.array(points).size <= 1:
                 colocation_points[idx] = None
 
         currents = currents.colocate(*colocation_points)
         return currents
+
+    def integrate_1d(
+        self,
+        function: np.ndarray,
+        phase: np.ndarray,
+        pts_u: np.ndarray,
+    ):
+        """Trapezoidal integration in two dimensions."""
+        return np.trapz(np.squeeze(function) * np.squeeze(phase), pts_u, axis=0)
 
     def integrate_2d(
         self,
@@ -405,6 +419,12 @@ class FieldProjector(Tidy3dBaseModel):
         idx_w, idx_uv = surface.monitor.pop_axis((0, 1, 2), axis=surface.axis)
         _, source_names = surface.monitor.pop_axis(("x", "y", "z"), axis=surface.axis)
 
+        # integration dimension for 2d far field projection
+        zero_dim = (dim for dim, size in enumerate(self.sim_data.simulation.size) if size == 0)
+        if self.is_2d_simulation:
+            integration_axis = {0, 1, 2} - {zero_dim, surface.axis}
+            idx_int_1d = integration_axis.pop()  # Get the remaining axis as an integer
+
         idx_u, idx_v = idx_uv
         cmp_1, cmp_2 = source_names
 
@@ -434,21 +454,38 @@ class FieldProjector(Tidy3dBaseModel):
 
                 phase_ij = phase[idx_u][:, None] * phase[idx_v][None, :] * phase[idx_w]
 
-                J[idx_u, i_th, j_ph] = self.integrate_2d(
-                    currents_f[f"E{cmp_1}"].values, phase_ij, pts[idx_u], pts[idx_v]
-                )
+                if self.is_2d_simulation:
+                    J[idx_u, i_th, j_ph] = self.integrate_1d(
+                        currents_f[f"E{cmp_1}"].values, phase_ij, pts[idx_int_1d]
+                    )
 
-                J[idx_v, i_th, j_ph] = self.integrate_2d(
-                    currents_f[f"E{cmp_2}"].values, phase_ij, pts[idx_u], pts[idx_v]
-                )
+                    J[idx_v, i_th, j_ph] = self.integrate_1d(
+                        currents_f[f"E{cmp_2}"].values, phase_ij, pts[idx_int_1d]
+                    )
 
-                M[idx_u, i_th, j_ph] = self.integrate_2d(
-                    currents_f[f"H{cmp_1}"].values, phase_ij, pts[idx_u], pts[idx_v]
-                )
+                    M[idx_u, i_th, j_ph] = self.integrate_1d(
+                        currents_f[f"H{cmp_1}"].values, phase_ij, pts[idx_int_1d]
+                    )
 
-                M[idx_v, i_th, j_ph] = self.integrate_2d(
-                    currents_f[f"H{cmp_2}"].values, phase_ij, pts[idx_u], pts[idx_v]
-                )
+                    M[idx_v, i_th, j_ph] = self.integrate_1d(
+                        currents_f[f"H{cmp_2}"].values, phase_ij, pts[idx_int_1d]
+                    )
+                else:
+                    J[idx_u, i_th, j_ph] = self.integrate_2d(
+                        currents_f[f"E{cmp_1}"].values, phase_ij, pts[idx_u], pts[idx_v]
+                    )
+
+                    J[idx_v, i_th, j_ph] = self.integrate_2d(
+                        currents_f[f"E{cmp_2}"].values, phase_ij, pts[idx_u], pts[idx_v]
+                    )
+
+                    M[idx_u, i_th, j_ph] = self.integrate_2d(
+                        currents_f[f"H{cmp_1}"].values, phase_ij, pts[idx_u], pts[idx_v]
+                    )
+
+                    M[idx_v, i_th, j_ph] = self.integrate_2d(
+                        currents_f[f"H{cmp_2}"].values, phase_ij, pts[idx_u], pts[idx_v]
+                    )
 
         if len(theta) < 2:
             integrate_for_one_theta(0)
@@ -576,7 +613,9 @@ class FieldProjector(Tidy3dBaseModel):
         medium = monitor.medium if monitor.medium else self.medium
         k = AbstractFieldProjectionData.wavenumber(medium=medium, frequency=freqs)
         phase = np.atleast_1d(
-            AbstractFieldProjectionData.propagation_phase(dist=monitor.proj_distance, k=k)
+            AbstractFieldProjectionData.propagation_factor(
+                dist=monitor.proj_distance, k=k, is_2d_simulation=self.is_2d_simulation
+            )
         )
 
         for surface in self.surfaces:
@@ -619,7 +658,11 @@ class FieldProjector(Tidy3dBaseModel):
             for name, field in zip(field_names, fields)
         }
         return FieldProjectionAngleData(
-            monitor=monitor, projection_surfaces=self.surfaces, medium=medium, **fields
+            monitor=monitor,
+            projection_surfaces=self.surfaces,
+            medium=medium,
+            is_2d_simulation=self.is_2d_simulation,
+            **fields,
         )
 
     def _project_fields_cartesian(
@@ -666,7 +709,9 @@ class FieldProjector(Tidy3dBaseModel):
         ):
             r, theta, phi = monitor.car_2_sph(_x, _y, _z)
             phase = np.atleast_1d(
-                AbstractFieldProjectionData.propagation_phase(dist=r, k=wavenumber)
+                AbstractFieldProjectionData.propagation_factor(
+                    dist=r, k=wavenumber, is_2d_simulation=self.is_2d_simulation
+                )
             )
 
             for surface in self.surfaces:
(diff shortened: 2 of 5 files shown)


momchil-flex:
Thanks @QimingFlex!

I think it should be possible to reuse our monitors as they are. I think you haven't yet familiarized yourself with pydantic validators, which is what we use ubiquitously in the code to ensure the user can only set things the way we want them to. I think that's what we can use here and you should start looking through some examples of those in our code (@dmarek-flex could help if needed).

I think what you describe, setting theta = pi/2, works specifically because the simulation 0-dim is set to `z`. If we were to instead rotate everything such that e.g. the 0-dim is `x`, then we would do something like phi = 0 and scan theta from 0 to 2*pi right? So I think what we can do is
1. Use a validator to enforce that no `theta` is provided to the user if a 2D simulation. Explain in the docstring that in 2D simulations, only phi is needed, and is always in the plane of the simulation.
2. Under the hood, convert the phi-s provided by the user, with the definition above, to a (theta, phi) to be passed to a regular 3D monitor.
3. Finally reframe the data back to that reference frame.

I think this can be done with the field data too. We don't necessarily want to introduce whole new objects to store the 2D data. We can use the same, and Etheta and Htheta will just always be zeros. We also don't currently explicitly split 2D sims into TE or TM. Depending on the exciting source, those could even be mixed. So it would just be again a matter of rotating everything from the global reference frame where `z` is the polar axis, to the 2D refarence frame that only has r and phi. Does that make sense?

QimingFlex:
Thanks for the detailed explanation and guidance @momchil-flex. I'll start with the pydantic validators and consult Damian. Your approach to simulations and data transformations is clear and helpful!

momchil-flex:
Actually, on second thought I feel like everything should be handled internally. That is to say, there will be no difference in the reference frame used from the user perspective depending on whether the simulation is 2D or 3D. The reason why I'm thinking this may make the most sense is so that users easily get consistent results between 2D and 3D sim setups. For example [here](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/GratingCoupler.html#Far-Field-Projection) I want to be able to just change the simulation size along y to 0, nothing else, and have everything working. That means that in the 2D sim I will still get the far field w.r.t. theta, and theta is defined w.r.t. the z axis.

I feel like from a usability perspective this is maybe better than requiring the user to rotate the axes based on the simulation 0-sized-dim? @tomflexcompute any input here?

tomflexcompute:
> Actually, on second thought I feel like everything should be handled internally. That is to say, there will be no difference in the reference frame used from the user perspective depending on whether the simulation is 2D or 3D. The reason why I'm thinking this may make the most sense is so that users easily get consistent results between 2D and 3D sim setups. For example [here](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/GratingCoupler.html#Far-Field-Projection) I want to be able to just change the simulation size along y to 0, nothing else, and have everything working. That means that in the 2D sim I will still get the far field w.r.t. theta, and theta is defined w.r.t. the z axis.
> 
> I feel like from a usability perspective this is maybe better than requiring the user to rotate the axes based on the simulation 0-sized-dim? @tomflexcompute any input here?

Yeah totally agree.

momchil-flex:
On the flip side though if someone is doing scattering off a Cylinder they would have to set different settings in the projection monitor (different angles, specifically) depending on which axis is the zero-size dim. Still, I feel like this is the better way.

QimingFlex:
Hi @momchil-flex , I just updated the revision. After some thought, I believe it's better to conduct near-to-far in the x-y plane for the following reasons: 
1. This conforms to the standards of cylindrical coordinates where phi starts from the x-axis and the z-component is along the infinite direction. I guess it's less common to consider a case where phi starts from the y-axis with the x-component along the infinite direction?
2. Such a revision would require much more effort to me at this moment as I still need some more time to get familiar with the code. Perhaps the directivity monitor will help me understand the code more so this won't be a problem later.

In this revision, the simulation size along the z-direction determines whether it's a 2D or 3D simulation. If a 2D simulation is determined, theta is set to pi/2, and the propagation constant is related to the Hankel function. These computations are handled internally. However, I encountered an issue with computing the RCS. It seems related to the monitor, which is a 3D monitor and cannot determine if it's a 2D or 3D simulation. To address this, I introduced radar_cross_section_2d.

tidy3d/components/field_projection.py
```diff
@@ -534,6 +559,10 @@ def project_fields(
             return self._project_fields_cartesian(proj_monitor)
         return self._project_fields_kspace(proj_monitor)
 
+    def is_3d_simulation(self) -> bool:
+        """Determine if the simulation is 2D or 3D based on the size attribute."""
+        return self.sim_data.simulation.size[2] != 0.01
```
momchil-flex: I assume this was set for your specific test? Normally we check if `sim.size[index] > 0`. We could also consider `if sim.grid.num_cells[index] > 1` to capture more general cases with a single pixel along a direction.

momchil-flex (commented):
Unfortunately I don't think this is a good approach @QimingFlex . An immediate case that comes to mind that is not handled well is a grating coupler. If you look at our grating coupler examples, the natural thing to do would be to compare 2d and 3d simulations where the 0D in the 2d sim is along y.

We could consider enabling this for now just so 2D projections are supported *in some way*, and make sure we validate things well and explain to users how to use the monitor. But I would prefer if we just handle the general case directly...

tidy3d/components/field_projection.py
QimingFlex: Thank you for the comments, @momchil-flex ! To achieve that goal for a general case, I'm considering how to modify the monitor if we allow 0-dim for every dimension. For example, Ez/Hphi and Hz/Ephi are generated when z is 0-dim, Ey/Hphi and Hy/Ephi are generated when y is 0-dim, and the same applies for x being 0-dim. Therefore, we would need to introduce new fields in the field monitor to consider all these three cases. Is this correct?

QimingFlex: The condition `if sim.grid.num_cells[index] > 1` seems to fit well!

momchil-flex:
> Thank you for the comments, @momchil-flex ! To achieve that goal for a general case, I'm considering how to modify the monitor if we allow 0-dim for every dimension. For example, Ez/Hphi and Hz/Ephi are generated when z is 0-dim, Ey/Hphi and Hy/Ephi are generated when y is 0-dim, and the same applies for x being 0-dim. Therefore, we would need to introduce new fields in the field monitor to consider all these three cases. Is this correct?

You should use the exact same data structures as for the 3D monitors, i.e. Er, Ephi, Etheta, with the polar coordinates defined w.r.t. the global z axis. For e.g. a 2D sim in the xy-plane then, you would have theta = pi / 2 and what you call Ez and Hphi for a TM sim would actually be Etheta and Hphi. All the other 3D components will be zero. If the plane is oriented along yz, then phi is set to 0, and what would be now Ex/Hphi in a purely 2D case correspond to Ephi/Htheta of the 3D data structure. For an xz plane, phi = pi/2 and Ey/Hphi again correspond to Ephi/Htheta. Try to visualize this and let me know if it makes sense to you.

I mean, as I'm writing this I do realize that it sounds convoluted. But like I mentioned above, the nice thing would be that I can just swap 2D and 3D sims without needing to change anything else in my scripts, i.e. take a 3D sim and choose to make one of the axes 0D and compare the projection results directly. I am not sure how much this complicates everything internally. Like, I feel like it should be possible to write the 2D projection as a special case of the 3D projection, and not have to do too much going back and forth between different coordinates. Conceptually, it should be possible to write the Green's function for a 2D sim with arbitrary normal axis as a 3D Green's function in the polar coordinates with a global z axis, i.e. in principle you can still write `G(x, y)` as a `G(r, theta, phi)`, assuming `G(x, y)` is just constant with `z`. Now, practically there are a couple extra steps to actually put this in the code, but I still think eventually producing results in the same coordinate system as the 3D monitors would be best.

tylerflex:
What's the status of this PR? is it for 2.7?
@QimingFlex @weiliangjin2021

QimingFlex:
> > Thank you for the comments, @momchil-flex ! To achieve that goal for a general case, I'm considering how to modify the monitor if we allow 0-dim for every dimension. For example, Ez/Hphi and Hz/Ephi are generated when z is 0-dim, Ey/Hphi and Hy/Ephi are generated when y is 0-dim, and the same applies for x being 0-dim. Therefore, we would need to introduce new fields in the field monitor to consider all these three cases. Is this correct?
> 
> You should use the exact same data structures as for the 3D monitors, i.e. Er, Ephi, Etheta, with the polar coordinates defined w.r.t. the global z axis. For e.g. a 2D sim in the xy-plane then, you would have theta = pi / 2 and what you call Ez and Hphi for a TM sim would actually be Etheta and Hphi. All the other 3D components will be zero. If the plane is oriented along yz, then phi is set to 0, and what would be now Ex/Hphi in a purely 2D case correspond to Ephi/Htheta of the 3D data structure. For an xz plane, phi = pi/2 and Ey/Hphi again correspond to Ephi/Htheta. Try to visualize this and let me know if it makes sense to you.
> 
> I mean, as I'm writing this I do realize that it sounds convoluted. But like I mentioned above, the nice thing would be that I can just swap 2D and 3D sims without needing to change anything else in my scripts, i.e. take a 3D sim and choose to make one of the axes 0D and compare the projection results directly. I am not sure how much this complicates everything internally. Like, I feel like it should be possible to write the 2D projection as a special case of the 3D projection, and not have to do too much going back and forth between different coordinates. Conceptually, it should be possible to write the Green's function for a 2D sim with arbitrary normal axis as a 3D Green's function in the polar coordinates with a global z axis, i.e. in principle you can still write `G(x, y)` as a `G(r, theta, phi)`, assuming `G(x, y)` is just constant with `z`. Now, practically there are a couple extra steps to actually put this in the code, but I still think eventually producing results in the same coordinate system as the 3D monitors would be best.

I just went through the code for both Monitors and MonitorData today and got a better understanding of the monitor classes. What you posted makes sense to me. I will test all three 2D scenarios to ensure we get consistent results.

QimingFlex:
It still needs some time to be finished. I'm not sure if it is for 2.7, any ideas? @tomflexcompute  @momchil-flex

QimingFlex:
> What's the status of this PR? is it for 2.7? @QimingFlex @weiliangjin2021

It still needs some time to be finished. I'm not sure if it is for 2.7, any ideas? @tomflexcompute @momchil-flex

tomflexcompute:
> > What's the status of this PR? is it for 2.7? @QimingFlex @weiliangjin2021
> 
> It still needs some time to be finished. I'm not sure if it is for 2.7, any ideas? @tomflexcompute @momchil-flex

If it's too much of a rush, I think we can certainly leave it for a later version release or pre-release.

tylerflex:
Ok, for now I'll tag it for 2.8 and mark as a draft just to focus on the PRs that are in their final stages.

QimingFlex:
@momchil-flex I've fixed the code for a general 2D farfield projection and tested it with RCS for all three scenarios in 2d. For a 0-dim along the x-direction, users should set φ = π/2  in the farfield monitor to get the desired farfield. For a 0-dim along the y-direction, users should set φ = 0 . For a 0-dim along the z-direction, users should set θ = π/2. This assumes users are familiar with the standard definition of spherical coordinates.

I have one question regarding the RCS calculation, which is a class method of the AbstractFieldProjectionData. In this case, I haven't found a way to determine if it is a 2D or 3D simulation based solely on the data class. Do you have any ideas on this?

momchil-flex:
> @momchil-flex I've fixed the code for a general 2D farfield projection and tested it with RCS for all three scenarios in 2d. For a 0-dim along the x-direction, users should set φ = π/2 in the farfield monitor to get the desired farfield. For a 0-dim along the y-direction, users should set φ = 0 . For a 0-dim along the z-direction, users should set θ = π/2. This assumes users are familiar with the standard definition of spherical coordinates.

Nice! We should have a Simulation level validator that checks this and explains it to the user if it is not set correctly. Namely, if simulation is 2D, iterate over monitors, and for every field projection monitor make sure that the angles are set correctly. Checking if the simulation is 2D is a bit tricky on the validator level actually as you don't have access to the grid as the Simulation is not yet fully initialized. So it could be just a size == 0 along some direction check. If you want to use `sim.grid.num_cells == 0` then you have to have it as a pre-upload validator, which is also possible, but not sure if needed. Generally, I think we prefer 2D sims to be set with a size strictly 0. But that may not be what users always do... and I think in the case of Bloch boundaries along the 0 dim, we actually do currently need a non-zero size to be set.

> I have one question regarding the RCS calculation, which is a class method of the AbstractFieldProjectionData. In this case, I haven't found a way to determine if it is a 2D or 3D simulation based solely on the data class. Do you have any ideas on this?

You could add a `2d_normal_axis: Axis = None` Field to the data structure to store whether this was associated with a 2D sim, and which axis was normal to it?

momchil-flex:
I feel like to avoid confusion we shouldn't merge this before we also fix the server-side projections. Those are much more commonly used anyway. @dmarek-flex now has very good understanding of how they work, maybe you can work together to see what will need to change, at least to scope it to decide whether it will be a lot of effort or not, and we can decide what to do.

Note that both the exact and the approximate projections will need to be fixed there.

QimingFlex:
> I feel like to avoid confusion we shouldn't merge this before we also fix the server-side projections. Those are much more commonly used anyway. @dmarek-flex now has very good understanding of how they work, maybe you can work together to see what will need to change, at least to scope it to decide whether it will be a lot of effort or not, and we can decide what to do.
> 
> Note that both the exact and the approximate projections will need to be fixed there.
 
I agree. I will prioritize the server-side computation.

QimingFlex:
Hi @momchil-flex , what is the equivalent of  `if sim.size[index] > 0` on the backend?

QimingFlex:
@momchil-flex I added validators in the frontend. With this fix, users can compute far field components Ephi, Etheta, Hphi, and Htheta in 2D. There are two remaining tasks for a full fix:

1. Frontend computation.
2. Update the monitor data class to indicate a 2D simulation for the correct constant for RCS computation in 2d.

tidy3d/components/simulation.py
```diff
@@ -2657,6 +2657,70 @@ def _projection_monitors_distance(cls, val, values):
                         )
         return val
 
+    @pydantic.validator("monitors", always=True)
+    @skip_if_fields_missing(["size"])
```
momchil-flex, line 2662: In what cases is this needed? Do we really want to skip this?

tidy3d/components/simulation.py
```diff
@@ -2657,6 +2657,70 @@ def _projection_monitors_distance(cls, val, values):
...
+            phi_value = 0
+        elif sim_size[2] == 0:
+            plane = "x-y"
+            theta_value = np.pi / 2
+
+        for idx, monitor in enumerate(val):
+            if isinstance(monitor, AbstractFieldProjectionMonitor):
+                if not monitor.far_field_approx:
+                    raise SetupError(
+                        "Exact far field projection is coming soon."
```
momchil-flex: Exact far field projection in 2D simulations is coming soon

tidy3d/components/simulation.py
```diff
@@ -2657,6 +2657,70 @@ def _projection_monitors_distance(cls, val, values):
...
+        """
+
+        if val is None:
+            return val
+
+        sim_size = values.get("size")
+
+        # Check if 3D simulation
+        if all(size != 0 for size in sim_size):
+            return val
```
momchil-flex, line 2684: Technically 1D wouldn't be correct either so I think we should explicitly check if 2D.

tidy3d/components/simulation.py
```diff
@@ -2657,6 +2657,70 @@ def _projection_monitors_distance(cls, val, values):
...
+            if isinstance(monitor, AbstractFieldProjectionMonitor):
+                if not monitor.far_field_approx:
+                    raise SetupError(
+                        "Exact far field projection is coming soon."
+                        "Please set `far_field_approx = True`."
+                    )
+                if plane == "y-z" and (len(monitor.phi) != 1 or monitor.phi[0] != phi_value):
+                    raise SetupError(
+                        "For a 2D simulation in the y-z plane, the observation angle `phi` "
+                        f"of the monitor [{idx}] should be set to `pi/2`."
```
momchil-flex: `f"of monitor '{monitor.name}' should be set to 'np.pi / 2'."`

In warning/error messages, we do not use markdown syntax, but instead just single quotes `' '` around variables, etc.

I think it's better to name the monitor rather than use the index - so `idx` can probably be removed from the loop.

Please fix in other messages too, including the `' '` syntax around `phi` and `theta`.

momchil-flex (commented):
You can edit `SIM_FULL` as needed to pass the validation. It was created fairly arbitrarily to have all sorts of components included in it.

tidy3d/components/simulation.py
QimingFlex: I thought we need a validated simulation size to determine if the simulation is 2d?

tidy3d/components/simulation.py
```diff
@@ -2657,6 +2657,71 @@ def _projection_monitors_distance(cls, val, values):
...
+        """
+
+        if val is None:
+            return val
+
+        sim_size = values.get("size")
+
+        # Skip validation if 1D or 3D simulation
+        non_zero_dims = sum(1 for size in sim_size if size != 0)
+        if non_zero_dims == 1 or non_zero_dims == 3:
```
weiliangjin2021: Does ProjectionMonitor work with 1D simulation? If not, is it validated somewhere already?

QimingFlex: I haven't hear 1D simulation for far field projection, maybe we can just determine if it is for 2d like this?
    `non_zero_dims = sum(1 for size in sim.size if size != 0)
    if non_zero_dims == 2:`
          ...

weiliangjin2021: i mean if it doesn't work with 1D simulation, we can throw an error.

QimingFlex: Yeah, sure. I will raise an error if this isn't validated somewhere else.

tidy3d/components/simulation.py
momchil-flex: Oh ok I forgot what the purpose of `skip_if_fields_missing` is, sounds good.

tidy3d/components/simulation.py
momchil-flex: Yeah, I don't think it's validated anywhere else.

Note also that ideally we should validate out the user-side projection if simulation is < 3D, until we fix that and remove the validation for 2D.

docs/notebooks
momchil-flex, line 1: So actually now the two submodule states here have gone stale. I know you were told that you should do this update, but you should revert it now. The original problem was with the `develop` branch, which wasn't up to date. Now it is, and the submodules have progressed further, so if we keep this here your PR will *revert* them. Generally, we should be fixing the `develop` branch when this happens, rather than you updating your submodules in the PR. If only the submodules test breaks but you have not made any changes to the submodules, ignore it. 

Let's figure out the revert once everything else is ready, I can help you with that.

tidy3d/components/data/monitor_data.py
```diff
@@ -2149,11 +2159,15 @@ def eta(self) -> complex:
         return ETA_0 / np.sqrt(eps_complex)
 
     @staticmethod
-    def propagation_phase(dist: Union[float, None], k: complex) -> complex:
+    def propagation_phase(dist: Union[float, None], k: complex, is_2d_simulation: bool) -> complex:
         """Phase associated with propagation of a distance with a given wavenumber."""
```
momchil-flex: This comment is not related to your changes, but `propagation_phase` and the docstring seems misleading since this is not just a phase factor. It is an overall normalization factor that contains both phase and amplitude decay due to the propagation. Maybe good to fix?

QimingFlex:
Hi @momchil-flex So far, we have fixed the FieldAngleProjectionMonitor for both frontend and backend with far-field approximation. For the remaining work, I propose the following:
1. Updates to Cartesian space and k-space (any references on k-space would be helpful). 
2. Solve for the exact projection.

momchil-flex:
> Hi @momchil-flex So far, we have fixed the FieldAngleProjectionMonitor for both frontend and backend with far-field approximation. For the remaining work, I propose the following:
> 
> 1. Updates to Cartesian space and k-space (any references on k-space would be helpful).
> 2. Solve for the exact projection.

Sounds good. We already validate 2. for an angle projection monitor. We should validate out 1. if it's not yet supported, i.e. in your validator you should add a check if k-space or cartesian monitor + 2D sim and error out.

Regarding k-space, it's conceptually exactly the same as angle, except the angles are defined not on a uniform grid in (theta, phi), but through a uniform grid of (kx, ky). The main difference is that in the k-space definition the angles are frequency dependent as they are basically determined through kx, ky and |k| = omega * n / c through simple trigonometry. Does that clear it up for you?

momchil-flex:
Having said that it seems like making the 2D KSpace monitors work should be quite simple, but see if you want to try it right away or keep it for later (and validate them out then).

QimingFlex:
> Regarding k-space, it's conceptually exactly the same as angle, except the angles are defined not on a uniform grid in (theta, phi), but through a uniform grid of (kx, ky). The main difference is that in the k-space definition the angles are frequency dependent as they are basically determined through kx, ky and |k| = omega * n / c through simple trigonometry. Does that clear it up for you?

Thanks. I have one more question regarding the proj_axis for both Cartesian space and k-space. The docstring says it is normal to the monitor. In the example in the docstrings, which uses a box monitor with proj_axis = 2, how does this proj_axis relate to the monitor? Since the monitor has 6 faces, only 2 of them have normals in the z-direction?

QimingFlex:
> Having said that it seems like making the 2D KSpace monitors work should be quite simple, but see if you want to try it right away or keep it for later (and validate them out then).

Let's keep that for later? For now, let's check if k-space or Cartesian monitor + 2D sims is used, and handle any errors accordingly.

tidy3d/components/simulation.py
```diff
@@ -2657,6 +2658,85 @@ def _projection_monitors_distance(cls, val, values):
...
+                    raise SetupError(
+                        f"Monitor '{monitor.name}' is not supported in 1D simulations."
+                    )
+
+                if isinstance(
+                    monitor, (FieldProjectionCartesianMonitor, FieldProjectionKSpaceMonitor)
+                ):
+                    raise SetupError(
+                        f"Monitor '{monitor.name}' in 2D simulations is coming soon. "
+                        "Please use FieldProjectionAngleMonitor."
```
momchil-flex: "Please use 'FieldProjectionAngleMonitor' instead."

momchil-flex (approved):
OK I think this is ready now with one tiny message string change.

tidy3d/components/data/monitor_data.py
```diff
@@ -2040,6 +2040,12 @@ class AbstractFieldProjectionData(MonitorData):
         discriminator=TYPE_TAG_STR,
     )
 
+    is_2d_simulation: bool = pd.Field(
+        False,
+        title="2D Simulation",
+        description=("Indicates whether the monitor data is for a 2D simulation."),
```
weiliangjin2021: The parentheses "()" here can be removed?

tidy3d/components/data/monitor_data.py
```diff
@@ -2149,11 +2155,15 @@ def eta(self) -> complex:
...
-        """Phase associated with propagation of a distance with a given wavenumber."""
+    def propagation_factor(dist: Union[float, None], k: complex, is_2d_simulation: bool) -> complex:
+        """A normalization factor that includes both phase and amplitude decay associated with propagation over a distance with a given wavenumber."""
         if dist is None:
             return 1.0
-        return -1j * k * np.exp(1j * k * dist) / (4 * np.pi * dist)
+
+        if is_2d_simulation:
+            return -np.exp(1j * k * dist) * np.sqrt(1j * k / (8 * np.pi * dist))
+        else:
```
weiliangjin2021: no need for "else"

tidy3d/components/simulation.py
weiliangjin2021: Where is 1D checked now?

QimingFlex:
@weiliangjin2021 It's validated [here](https://github.com/flexcompute/tidy3d/blob/c23823f3be43cb8ffe6efbb168bbdd1a9f3cf5b7/tidy3d/components/simulation.py#L2702) within the loop for monitors.

weiliangjin2021 (approved):

PR #1642, opened 2024-04-26
move `adjoint` into main components

tylerflex:
## Goal

instead of needing to import and use `tda.Jax_` components and add jax objects to `JaxSimulation.input_structures`, this PR attempts to add `jax` differentiability support to regular `td.` components. So you could in principle write an objective function completely in the `td.` namespace and then differentiate it with `jax`.

## Intro

Note for @yaugenst. I gave this a try, which would enable running regular `td.Simulation` with adjoint support. It was going ok but then I ran into an issue with defining a test `data = run(sim)`. `jax` does not want to call my `run_bwd` function for my `@custom_vjp` when computing the derivative of this step. I think it doesn't detect that the data is flowing through `data = run(sim)` as the gradient is just 0.0.

I ran into this same issue in the 2.7 refactor #1551 and don't remember exactly how I resolved it in the end..

Don't feel like you need to look into this, especially not until you join officially, but eventually I might need a 2nd pair of eyes on it eventually, so thought I'd publish as a draft PR.

## The Strategy

1. Add a new field to `Tidy3DBaseModel` called `jax_info : dict = {}`.  This dictionary should basically store all `jax`-traceable data for an instance, keeping it separate from the original fields, which might contain other types.
2. When we initialize a `Tidy3DBaseModel`any `pd.Field()` annotated as `jax_field=True` will store the passed data into `jax_info`. The passed data is also sanitized with `jax.lax.stop_gradient()` and passed to the original field.
3. Define `Simulation` and `SimulationData` as `@register_pytree_node_class` classes and define their tree flatten/unflatten operations. Basically the children are just the `jax_field` and everything else goes into the `aux_data`.
#
## A different strategy
Another approach could be to simply try to get all existing fields to work with `jax` types, as needed. The main challenge is the data objects I think. These are based on `xarray`, which seems fundamentally incompatible with `jax` unfortunately.

## Priority
This would definitely improve the usability of tidy3d for adjoint, but there's a ton of other stuff to do within the existing framework. So we need to balance the priority and not sink too much time into this. I'd say it's worth considering the feasibility and seeing if we can get this simple test working with an emulated `run()` function.  if it shows promise, we might want to build off of this new approach going forward and deprecate the `adjoint` plugin (still support it but not add any new features)

### DIFF
--- a/tests/test_components/test_adjoint.py
+++ b/tests/test_components/test_adjoint.py
@@ -4,20 +4,75 @@ import tidy3d as td
 import jax
 import jax.numpy as jnp
 
+
+def make_box(x: float) -> td.Box:
+    """Make a box storing (x,x,x) in its size."""
+    return td.Box(
+        center=(0, 0, 0),
+        size=(x, x, x),
+    )
+
+
+def make_structure(x: float) -> td.Structure:
+    """Make a structure with a geometry of a Box storing (x,x,x) in its size."""
+    b = make_box(x)
+    return td.Structure(geometry=b, medium=td.Medium(permittivity=2.0))
+
+
 def test_jax_field():
+    """Test storage of jax fields in tidy3d objects."""
+
+    def f(x):
+        b = make_box(x)
+        return jnp.sum(jnp.array(b.jax_info["size"]))
+
+    val = f(1.0)
+    grad = jax.grad(f)(1.0)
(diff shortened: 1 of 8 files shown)


tylerflex:
closing this but we can look at it later for reference

PR #1643, opened 2024-04-26
fix plotting of 2d objects

dmarek-flex:
Found and issue with plotting `Transformed` objects that happen to have 0 thickness. The problem was that no intersections were found when using Trimesh to compute intersections. This is because all of the triangles were directly coincident on the intersection plane. This will not happen for fully 3D geometries.

I fixed it by reverting to simple intersection computations when the plane happens to be aligned with an axis. Another option would have been to give the object a very small thickness before meshing.

With this fix, I can correctly display geometries like:

image

### DIFF
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -1456,7 +1456,76 @@ class Centered(Geometry, ABC):
         return val
 
 
-class Planar(Geometry, ABC):
+class SimplePlaneIntersection(Geometry, ABC):
+    """A geometry where intersections with an axis aligned plane may be computed efficiently."""
+
+    def intersections_tilted_plane(
+        self, normal: Coordinate, origin: Coordinate, to_2D: MatrixReal4x4
+    ) -> List[Shapely]:
+        """Return a list of shapely geometries at the plane specified by normal and origin.
+        Checks special cases before relying on the complete computation.
+
+        Parameters
+        ----------
+        normal : Coordinate
+            Vector defining the normal direction to the plane.
+        origin : Coordinate
+            Vector defining the plane origin.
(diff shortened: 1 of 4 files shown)


lucas-flexcompute (approved):
Looks good!

tidy3d/components/geometry/base.py
```diff
@@ -1811,7 +1880,7 @@ def surfaces_with_exclusion(cls, size: Size, center: Coordinate, **kwargs):
         return surfaces
 
     @verify_packages_import(["trimesh"])
-    def intersections_tilted_plane(
+    def _do_intersections_tilted_plane(
```
tylerflex, line 1883: does this break anything eg. in the backend? I'm a little worried about removing a "public" (eg no leading `_`) method. Would it make sense to leave this the same name?

tylerflex (commented):
Implementation looks good. A question about changing the `intersections_tilted_plane` method name. Do we need to do this? I'm worried it will cause complications but am not sure about specifically where this would come up.

If we do change it, we might want to deprecate that method first (eg. call the `_do_intersections_tilted_plane`, but log a warning?

tylerflex:
Can you also add a changelog item? thanks

tidy3d/components/geometry/base.py
dmarek-flex: I think it should be fine, since the original `intersections_tilted_plane` is defined in the base class. The interface for these geometries is unchanged. Now, there is just a quick check to see if a simpler method can be used to compute the intersection, before falling back to the method involving generating a mesh.

tylerflex: ok I see. thanks! should be fine then

tylerflex (approved):

PR #1644, opened 2024-04-26
extra geometries for 2d material

dmarek-flex:
Added the ability for Transformed/GeometryGroup/ClipOperation geometries to be composed of 2D materials.

Solves [issue 576](https://github.com/flexcompute/tidy3d-core/issues/576)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Convenience method `constant_loss_tangent_model` in `FastDispersionFitter` to fit constant loss tangent material model.
 - Classmethods in `DispersionFitter` to load complex-valued permittivity or loss tangent data.
 - Pre-upload validator to check that mode sources overlap with more than 2 grid cells.
+- Support `2DMedium` for `Transformed`/`GeometryGroup`/`ClipOperation` geometries.
 
 ### Changed
 - `tidy3d convert` from `.lsf` files to tidy3d scripts is moved to another repository at `https://github.com/hirako22/Lumerical-to-Tidy3D-Converter`.
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -368,7 +368,14 @@ class Geometry(Tidy3dBaseModel, ABC):
     @cached_property
     def _normal_2dmaterial(self) -> Axis:
         """Get the normal to the given geometry, checking that it is a 2D geometry."""
-        raise ValidationError("'Medium2D' is not conpatible with this geometry class.")
+        raise ValidationError("'Medium2D' is not compatible with this geometry class.")
+
+    def _update_from_bounds(self, bounds: Tuple[float, float], axis: Axis) -> Geometry:
+        """Returns an updated geometry which has been transformed to fit within ``bounds``
+        along the ``axis`` direction."""
+        raise NotImplementedError(
+            "'_update_from_bounds' is not compatible with this geometry class."
+        )
 
     @equal_aspect
     @add_ax_if_none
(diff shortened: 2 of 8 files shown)


tidy3d/components/geometry/utils_2d.py
```diff
@@ -63,17 +59,37 @@ def set_bounds(geom: Geometry, bounds: Tuple[float, float], axis: Axis) -> Geome
         new_center[axis] = (bounds[0] + bounds[1]) / 2
         new_length = bounds[1] - bounds[0]
         return geom.updated_copy(center=new_center, length=new_length)
+    if isinstance(geom, Transformed):
+        min_bound = np.array([0, 0, 0, 1.0])
```
tylerflex: out of curiosity, why is this a z unit vector? should it be set to the unit normal vector for the geometry?

tylerflex (approved):
Looks good! Thanks @dmarek-flex. One minor question

tidy3d/components/geometry/utils_2d.py
dmarek-flex: That is a good question. I think I got the idea from a different [section](https://github.com/flexcompute/tidy3d/blob/844b6bab56232190e531cadec36de22d156cf5ca/tidy3d/components/geometry/base.py#L2297-L2320).

They arrays are actually made to be size 4, so that the Transform can be applied. I think I need to double-check this part though.

dmarek-flex: Yea I think the 1 is needed to apply translation transformations, but maybe @lucas-flexcompute can just confirm that it is ok?

lucas-flexcompute: Yes, the 1 at the 4th position is to set up [homogeneous coordinates](https://en.wikipedia.org/wiki/Homogeneous_coordinates) for 3D transformations.

weiliangjin2021 (approved):
Nice work! All look good.

dmarek-flex:
I made the modifications that @tylerflex suggested in the [coaxial ports PR](https://github.com/flexcompute/tidy3d/pull/1669)

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Convenience method `constant_loss_tangent_model` in `FastDispersionFitter` to fit constant loss tangent material model.
 - Classmethods in `DispersionFitter` to load complex-valued permittivity or loss tangent data.
 - Pre-upload validator to check that mode sources overlap with more than 2 grid cells.
+- Support `2DMedium` for `Transformed`/`GeometryGroup`/`ClipOperation` geometries
```
tylerflex: period

tidy3d/components/geometry/base.py
```diff
@@ -368,7 +368,11 @@ def _get_center(pt_min: float, pt_max: float) -> float:
     @cached_property
     def _normal_2dmaterial(self) -> Axis:
         """Get the normal to the given geometry, checking that it is a 2D geometry."""
-        raise ValidationError("'Medium2D' is not conpatible with this geometry class.")
+        raise ValidationError("'Medium2D' is not compatible with this geometry class.")
+
+    def _update_from_bounds(self, bounds: Tuple[float, float], axis: Axis) -> Geometry:
+        """Set the bounds of a geometry in the axis direction."""
+        raise ValidationError("'_update_from_bounds' is not compatible with this geometry class.")
```
tylerflex: I think this should be a `NotImplementedError`

tylerflex: Also, perhaps the docstring could be more clear. So this method basically makes an updated copy of the geometry that's transformed to fit within some bounds in the axis direction? from the current docstring it sound like it just sets `geometry.bounds`

tidy3d/components/geometry/polyslab.py
```diff
@@ -431,6 +431,15 @@ def _normal_2dmaterial(self) -> Axis:
             raise ValidationError("'Medium2D' requires the 'PolySlab' bounds to be equal.")
         return self.axis
 
+    def _update_from_bounds(self, bounds: Tuple[float, float], axis: Axis) -> PolySlab:
+        """Set the bounds of a geometry in the axis direction."""
+        if axis != self.axis:
+            raise ValidationError(
```
tylerflex: probably `ValueError` makes more sense. these `ValidtionErrors` are used mainly for when a `pydantic.validator` fails when initializing an object

tylerflex (approved):
Looks much better I think! Thanks @dmarek-flex. Couple minor comments

dmarek-flex:
Thanks @tylerflex for finding those lingering issues. I'll merge this when all tests pass.

PR #1647, opened 2024-04-29
feat: optimize the job process to start download data early

lei-flex:


tylerflex (commented):
I think the problem with this is that we lose the nice progress bars and display for `batch.run()`. because it calls all of the `job.` methods directly..

tylerflex:
We can re-open this later when we want to re-visit and optimize the batch processing further.

PR #1648, opened 2024-04-29
try to make Batch.run fully multi-process

tylerflex:


tylerflex:
This didnt work on windows. Will re-open if we want to try again later.

PR #1650, opened 2024-04-30
:package: BUILD: Update for MacOS 14 compatibility

daquinteroflex:
Ok so just to update and for people to know if a user asks this.
The latest 2.6.4 tidy3d development installation will not properly pip install in the just-updated macos-latest MacOS 14 platform (we've tested up to MacOS 12 and other latest platforms). I think the updates we're doing in 2.7 should sort this out, unless we want to do a 2.6.5 release. Making a PR to fix related issues too.
You can spot the difference between [this](https://github.com/flexcompute/tidy3d/actions/runs/8849420754) and [this](https://github.com/flexcompute/tidy3d/actions/runs/8852714664/job/24373302494) action on the same commit if you're curious why.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -11,6 +11,7 @@ on:
       - 'pre/*'
 
 jobs:
+
   pre-commit:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


momchil-flex:
Great, thanks!

PR #1651, opened 2024-04-30
:package: BUILD: Fix the trigger develop cli action to sync branches

daquinteroflex:
Hello! I think we need to merge this as soon as we can in order for the develop actions to begin running from the branch that triggered the build.

Also I found a bug on tidy3d-develop-cli that was testing the develop installation only on the develop  branch, but not on the actual branches where the commit was created in which I'm fixing.

Note that the develop action won't be fixed until we merge this into develop

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -112,22 +112,11 @@ jobs:
     #  -----  install & configure poetry  -----
     #----------------------------------------------
     - name: Install Poetry
-      uses: abatilo/actions-poetry@v2
+      uses: snok/install-poetry@v1
       with:
-        version: 1.7.1
-
-    # After installing Poetry add to PATH
-    - name: Add Poetry's bin directory to PATH
-      run: |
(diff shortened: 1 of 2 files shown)


momchil-flex:
Since we merged #1650 to pre/2.7, this here is only relevant if we want to make a 2.6.5 right?

daquinteroflex:
Unfortunately, this PR is relevant if we want all develop-cli tests to pass on all branches including pre-/2.7. This is becuase of the way that the "trigger" action occurs when the "run tests" workflow has completed. It can only trigger if the action is in the main branch. However, our action in the main develop branch had a bug that it was only checking out the develop branch, rather than the branch that triggered the action on the tidy3d-develop-cli action. This PR fixes that.

momchil-flex:
Ah ok.

momchil-flex:
@tylerflex you're marked as reviewer here

tylerflex (approved):

PR #1656, opened 2024-05-01
disable failing default log level test

tylerflex:
This test fails sometimes on  GitHub actions (2.7) 

I'm not able to reproduce it locally though and nothing in the logs indicates why this would happen. Just putting this PR out there in case we want to just disable it and figure out later. 

Do either of you know what could be going on? maybe it has to do with some 2.7 changes to logging?

### DIFF
--- a/tests/test_package/test_config.py
+++ b/tests/test_package/test_config.py
@@ -11,9 +11,10 @@ def test_logging_level():
     """Make sure setting the logging level in config affects the log.level"""
 
     # default level
-    assert td.log.handlers["console"].level == _level_value[DEFAULT_LEVEL]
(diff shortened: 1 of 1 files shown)


lucas-flexcompute (approved):
Could it be that other tests that change the the log level from the default are being run before this? As a general rule I think that if any test changes the default, it should reset it. Either way, if this is the problem, this test doesn't really matter.

tylerflex:
I might have found where this happened. There was a call to `td.config.logging_level = "ERROR"` in the test module for the new `invdes` tool. I removed that and will see if that fixes things.

tylerflex:
Yea actually that seemed to resolve this error. I will close this PR

PR #1660, opened 2024-05-02
fix: GDS export wrappers in `JaxSimulation` (#1334)

### LINKED ISSUE #1334: GDS export wrappers in `JaxSimulation`
Enable `JaxSimulation` to export simulation directly rather than making the user convert to a regular `Simulation` first.

```py
# Convert the final adjoint simulation to a regular one.
sim_final = sim_adj.to_simulation()[0]

# Export the final simulation to GDS.
sim_final.to_gds_file(fname="inverse_design.gds", z=0, permittivity_threshold=6, frequency=200e14)
```
becomes
```py
# Export the final simulation to GDS.
sim_adj.to_gds_file(fname="inverse_design.gds", z=0, permittivity_threshold=6, frequency=200e14)
```

yaugenst-flex:
Fixes #1334 by overriding `to_gds()` in `JaxSimulation` to call `self.to_simulation()` first. Addresses both `to_gds()` and `to_gds_file()` this way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for `.gz` files in `Simulation` version updater.
 - Warning if a nonuniform custom medium is intersecting `PlaneWave`, `GaussianBeam`, `AstigmaticGaussianBeam`, `FieldProjectionCartesianMonitor`, `FieldProjectionAngleMonitor`, `FieldProjectionKSpaceMonitor`, and `DiffractionMonitor`.
 - Tidy3D objects may store arbitrary metadata in an `.attrs` dictionary.
+- `JaxSimulation` now supports the following GDS export methods: `to_gds()`, `to_gds_file()`, `to_gdspy()`, and `to_gdstk()`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Bug in plotting and computing tilted plane intersections of transformed 0 thickness geometries.
+- `Simulation.to_gdspy()` and `Simulation.to_gdstk()` now place polygons in GDS layer `(0, 0)` when no `gds_layer_dtype_map` is provided instead of erroring.
 
 ## [2.7.0rc1] - 2024-04-22
 
--- a/tidy3d/plugins/adjoint/components/simulation.py
+++ b/tidy3d/plugins/adjoint/components/simulation.py
@@ -1,4 +1,5 @@
 """Defines a jax-compatible simulation."""
+
 from __future__ import annotations
 
 from typing import Tuple, Union, List, Dict, Literal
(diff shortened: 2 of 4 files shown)


tylerflex (commented):
just took a quick look, could we also add
1. the other gds conversion methods (I think there's also `to_gds_file()`, might be one or two more but that's the main one people will use I think).
2. Just copy over the docstring from the original methods so people can see the parameters (I realize this is kind of ugly but it's what we have until we can better refactor things).
3. Add a line in the changelog under "added" in "unreleased"

Thanks!

yaugenst-flex:
Hey @tylerflex, `to_gds_file()` works and is actually what's being tested, since it calls `to_gds()` internally. I'll add the docstring and changelog notes.

yaugenst-flex:
Ah, but `to_gdstk()` and `to_gdspy()` probably won't work like this, you're right.

tylerflex:
oh I see. hm, maybe we dont need the specific gdstk and gdspy ones but if you're already working on them, it wouldn't hurt. Good to know that the to_gds_file will work automatically, I guess just mention all of them that will work from JaxSimulation now in the changelog and should be good.

tidy3d/plugins/adjoint/components/simulation.py
```diff
@@ -433,6 +435,123 @@ def to_simulation(self) -> Tuple[Simulation, JaxInfo]:
...
+        ----------
+        cell : ``gdstk.Cell`` or ``gdspy.Cell``
+            Cell object to which the generated polygons are added.
+        x : float = None
+            Position of plane in x direction, only one of x,y,z can be specified to define plane.
+        y : float = None
+            Position of plane in y direction, only one of x,y,z can be specified to define plane.
+        z : float = None
+            Position of plane in z direction, only one of x,y,z can be specified to define plane.
+        permittivity_threshold : float = 1.001
```
tylerflex: weird that it's 1.001 in the docstring but 1 in the default, do we know what's up with that?

CHANGELOG.md
```diff
@@ -9,11 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for `.gz` files in `Simulation` version updater.
 - Warning if a nonuniform custom medium is intersecting `PlaneWave`, `GaussianBeam`, `AstigmaticGaussianBeam`, `FieldProjectionCartesianMonitor`, `FieldProjectionAngleMonitor`, `FieldProjectionKSpaceMonitor`, and `DiffractionMonitor`.
 - Tidy3D objects may store arbitrary metadata in an `.attrs` dictionary.
+- `JaxSimulation` now supports all GDS export methods.
```
tylerflex: could you list them here? e.g. `to_gds()`, `to_gds_file()`, `to_gdspy()`, and `to_gdstk()`.? Just so users immediately will know what they are called if they see this line, thanks!

tylerflex (commented):
Looks good to me, thanks @yaugenst-flex , just a couple minor comments and also please squash the commits to 1 (or 2 maybe with one of them to fix the gds{py, tk} methods) if possible). Thanks!

tidy3d/plugins/adjoint/components/simulation.py
yaugenst-flex: Huh interesting, I missed that. But it has been around at least since eea873a. I'll change the docstring.

yaugenst-flex:
> please squash the commits to 1 (or 2 maybe with one of them to fix the gds{py, tk} methods) if possible)

Done!

tylerflex (approved):
Looks great! thanks @yaugenst-flex . I will merge once the tests pass!

PR #1664, opened 2024-05-02
remove multi-threading from adjoint plugin

tylerflex:
May require a backend PR.

Turns out the multi-threading logic was causing the seg faulting in invdes plugin on ubuntu. I don't understand why.

Making this PR in case we need it later but will continue looking into it


tylerflex:
We dont need this because #1665 seems to fix the seg fault issues

PR #1665, opened 2024-05-02
use joblib in adjoint multithreading

tylerflex:
I wasn't exactly sure how to specify the joblib dependency, just put it in the core dependencies for now.

### DIFF
--- a/tidy3d/plugins/adjoint/components/geometry.py
+++ b/tidy3d/plugins/adjoint/components/geometry.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 
 from abc import ABC
 from typing import Tuple, Union, Dict, List
-from multiprocessing import Pool
+from joblib import Parallel, delayed
 
 import pydantic.v1 as pd
 import numpy as np
@@ -540,12 +540,14 @@ class JaxPolySlab(JaxGeometry, PolySlab, JaxObject):
         """Generate arguments for ``vertex_vjp``."""
 
         num_verts = len(self.vertices)
-        args = [range(num_verts)]
 
-        # append all of the arguments that are the same for each call
-        constant_args = [e_mult_xyz, d_mult_xyz, sim_bounds, wvl_mat, eps_out, eps_in]
-        args += [[arg] * num_verts for arg in constant_args]
-        return args
+        arg_list = []
+
+        for i in range(num_verts):
+            args_i = [i] + [e_mult_xyz, d_mult_xyz, sim_bounds, wvl_mat, eps_out, eps_in]
+            arg_list.append(args_i)
+
+        return arg_list
 
     def store_vjp_sequential(
         self,
@@ -559,8 +561,10 @@ class JaxPolySlab(JaxGeometry, PolySlab, JaxObject):
         """Stores the gradient of the vertices given forward and adjoint field data."""
         # Construct arguments to pass to the parallel vertices_vjp computation
 
-        args = self._make_vertex_args(e_mult_xyz, d_mult_xyz, sim_bounds, wvl_mat, eps_out, eps_in)
-        vertices_vjp = tuple(map(self.vertex_vjp, *args))
+        arg_list = self._make_vertex_args(
+            e_mult_xyz, d_mult_xyz, sim_bounds, wvl_mat, eps_out, eps_in
+        )
+        vertices_vjp = tuple(self.vertex_vjp(*args) for args in arg_list)
         vertices_vjp = tuple(tuple(x) for x in vertices_vjp)
 
         return self.updated_copy(vertices_jax=vertices_vjp)
@@ -578,8 +582,9 @@ class JaxPolySlab(JaxGeometry, PolySlab, JaxObject):
         """Stores the gradient of the vertices given forward and adjoint field data."""
 
         args = self._make_vertex_args(e_mult_xyz, d_mult_xyz, sim_bounds, wvl_mat, eps_out, eps_in)
-        with Pool(num_proc) as pool:
-            vertices_vjp = pool.starmap(self.vertex_vjp, zip(*args))
+
+        vertices_vjp = Parallel(n_jobs=num_proc)(delayed(self.vertex_vjp)(*arg) for arg in args)
+
         vertices_vjp = tuple(tuple(x) for x in vertices_vjp)
         return self.updated_copy(vertices_jax=vertices_vjp)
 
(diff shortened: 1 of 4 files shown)


momchil-flex (approved):
Looks good to me, except it would be good to test if the speed up is still there. Do we have something like that in the tests?

tylerflex:
Not explicitly, but I can test it locally

prashkh:
> Looks good to me, except it would be good to test if the speed up is still there. Do we have something like that in the tests?

We have tests for run_batch. I can update that to also run mode solvers serially and check to make sure that the speed up is there.

momchil-flex:
> > Looks good to me, except it would be good to test if the speed up is still there. Do we have something like that in the tests?
> 
> We have tests for run_batch. I can update that to also run mode solvers serially and check to make sure that the speed up is there.

Here we're replacing multiprocessing with joblib in the adjoint processing @prashkh , not related to the mode solver.

tylerflex:
@momchil-flex weird results.. I used `time` (Mac OS) to time  the adjoint pipeline test `test_adjoint_pipeline` which runs through both with local and client-side logic. I changed the `num_proc` from 1-4 for both pathways and these were the results.

#### pre/2.7
1 -> 11.68s user 0.58s system 149% cpu 8.206 total
2 -> 44.56s user 8.45s system 313% cpu 16.915 total
3 -> 53.17s user 10.02s system 335% cpu 18.818 total
4 -> 55.86s user 10.41s system 339% cpu 19.501 total

#### Tyler/adjoint/joblib
1 -> 11.66s user 0.66s system 144% cpu 8.537 total
2 -> 16.93s user 1.84s system 185% cpu 10.132 total
3 -> 18.05s user 1.50s system 177% cpu 10.996 total
4 -> 19.76s user 1.55s system 184% cpu 11.532 total

It would seem to me that `multiprocess` made it WORSE and using `joblib` just makes it about the same amount of time. Not sure what to make of this..

tidy3d/plugins/adjoint/components/geometry.py
```diff
@@ -578,8 +582,9 @@ def store_vjp_parallel(
         """Stores the gradient of the vertices given forward and adjoint field data."""
 
         args = self._make_vertex_args(e_mult_xyz, d_mult_xyz, sim_bounds, wvl_mat, eps_out, eps_in)
-        with Pool(num_proc) as pool:
-            vertices_vjp = pool.starmap(self.vertex_vjp, zip(*args))
+
+        vertices_vjp = Parallel(n_jobs=num_proc)(delayed(self.vertex_vjp)(*arg) for arg in args)
```
prashkh, line 586: It might be worth testing which backend results in the highest speed up. By default you are using 'loky'. Might be worth testing 'multiprocessing' backend as well by changing backend using parallel_config context manager.

https://joblib.readthedocs.io/en/latest/generated/joblib.parallel_backend.html

prashkh (approved):
Looks great! My only comment would be to see if `loky`, which is the default backend choice, is better or worse compared to `multiprocessing` backend for this task.

tylerflex:
When I bump the number of vertices in the polyslab up to 600, I start to see a benefit of more processors ... not sure if it's worth the reduced performance when not limited by this though. At least it's better than before.

```
1. 37.14s user 0.67s system 107% cpu 35.036 total
2. 42.07s user 1.12s system 188% cpu 22.967 total
3. 42.32s user 1.07s system 207% cpu 20.906 total
4. 45.41s user 1.59s system 226% cpu 20.790 total
```

tylerflex:
@prashkh thanks for the tip, the `backend="multiprocessing"` unfortunately seems to make things perform worse, so sticking with default I think for now

1. 35.69s user 0.52s system 111% cpu 32.569 total
2. 57.60s user 5.04s system 209% cpu 29.910 total
3. 60.16s user 7.72s system 215% cpu 31.564 total
4. 67.56s user 7.46s system 228% cpu 32.821 total

momchil-flex:
Interesting. I remember testing on the server and for the particular job we were using there was definitely a big speedup. I don't remember anymore if we tested locally and/or if we tested potential overhead in the case of small number of vertices. I guess in all cases in your tests the joblib change here looks better than the current state, so probably ok to merge...

For what it's worth, on my machine I don't see a slow down if I increase the num_proc in that test. Can you give me the 600 vertices to try?

tylerflex:
Here's how I construct the `VERTICES` that goes into the test
```
n = 301
xs0 = np.linspace(-1, 0, n)
ys0 = np.ones_like(xs0)

xs1 = xs0[::-1]
ys1 = 0* ys0

xs = np.concatenate((xs0, xs1))
ys = np.concatenate((ys0, ys1))
VERTICES = np.stack((xs, ys)).T.tolist()
```

and then what I did was change the test to only do `local=True` and then adjust the `NUM_PROC_LOCAL` in `simulation.py` (of adjoint)

tylerflex:
this works pretty well too.. changed to it since testing
```py
n = 600
thetas = 2 * np.pi * np.linspace(0, 1, n + 1)[:n]
xs = np.cos(thetas)
ys = np.sin(thetas)

VERTICES = np.stack((xs, ys), axis=-1).tolist()

```

momchil-flex:
Yeah I don't see a slowdown for 3 vertices and it gets faster for 600 - both multiprocessing and joblib, but joblib does seem to do better. I'm just looking at the pytest timing for that test.

momchil-flex:
Ok, actually, weird, for me the multiprocessing version has no overhead for the small case. The joblib one does. goes from 12 to 14 to 16s with 1-2-4 num_proc respectively...

But for the large number of vertices, joblib does better than multiprocessing at 4 cores...

tylerflex:
ok I'll go ahead and merge this, then. maybe it will take some of the pressure off the large adjoint tasks coming in lately.

tylerflex:
Well, turns out when I rebase `invdes` against this after merging, I get the following warning when running the tests
```
RuntimeWarning: os.fork() was called. os.fork() is incompatible with multithreaded code, and JAX is multithreaded, so this will likely lead to a deadlock.
```

something must be going on in `invdes` that's messing with the multi-threaded code, again even if it shouldn't be being called at all.. Will need to dig into this again unfortunately. At least `joblib` seems to pass the tests and maybe performs better for now.

tylerflex:
correction, the multi-threaded pipeline is indeed called with `invdes` (because it's using an emulation of `tda.web.run()` which calls the `webapi_run_fwd/adj` functions. But I just don't understand why this would error for `invdes` and not regular adjoint. The only difference as far as I can tell is that in `inves`, we have a method of `InverseDesign` that returns the objective function, rather than it being defined directly in the script..

PR #1669, opened 2024-05-02
support coaxial lumped ports

dmarek-flex:
Related to [issue 523](https://github.com/flexcompute/tidy3d-core/issues/523).

Adding support for coaxial ports. This requires adding support for lumped elements with coaxial geometry as well. The basic idea is the same as for the original rectangular lumped port except now:

- Current source is radially symmetric and implemented as a CustomCurrentSource
- Path integral for current computation has to use the core conductor of the port, rather than the lumped resistor itself. This means the port definition requires a direction parameter, which tells the port on which side of the port plane to compute the path integral for current.
- The geometry related to the port is an annulus instead of a rectangle


dmarek-flex:
Ok this PR is ready for review now! A couple of things that I was not confident about:

- I had to use this "noqa" commands to get rid of ruff errors in `test_medium.py`
- Also in docstrings I had to use the format ":class:`.Simulation`" to get the proper linking in the docs, but I think it is usually suggested to use double back ticks. I don't know if this matters at all.

tidy3d/components/geometry/base.py
```diff
@@ -2881,5 +2916,24 @@ def _surface_area(self, bounds: Bound) -> float:
...
+        """Get the normal to the given geometry, checking that it is a 2D geometry."""
+
+        normals = {geom._normal_2dmaterial for geom in self.geometries}
+
+        if len(normals) != 1:
+            raise ValidationError(
+                "'Medium2D' requires all geometries in the 'GeometryGroup' to "
+                "share exactly one dimension with zero size."
+            )
+        normal = next(iter(normals))
```
tylerflex: what is this line doing? can we make it more straightforward?

tylerflex: normals is a `set` of `Axis` corresponding to the axes of all of the 2D geometries? 
so we're first checking that they are all aligned on the same axis and then that all of the bounds are at the same normal position?

so is `next(iter(normals))` just taking the (only) element of the set?

tidy3d/components/geometry/utils_2d.py
```diff
@@ -63,17 +59,37 @@ def set_bounds(geom: Geometry, bounds: Tuple[float, float], axis: Axis) -> Geome
         new_center[axis] = (bounds[0] + bounds[1]) / 2
         new_length = bounds[1] - bounds[0]
         return geom.updated_copy(center=new_center, length=new_length)
+    if isinstance(geom, Transformed):
```
tylerflex: when I see these chains of `if isinstance()`, it tends to suggest to me that these should be written as methods of the class, and then just called once, passing the instance in with whatever arguments are needed. for example, would it make sense for something like a `Geometry.update_from_bounds(self, bounds: Bounds) -> Geometry` method implemented in each of these classes? In this example, perhaps we could just replace all of this code with `geom.update_from_bounds(new_bounds)`?

tests/test_components/test_medium.py
tylerflex, line 1: It would be helpful to do a code coverage check for all of these new 2D material / microwave / RF changes. just to see how good the coverage is.

After pip installing `pytest-cov` you can run


```
pytest -rA tests  --cov-report=html --cov=tidy3d && open htmlcov/index.html
```

could you try this and take a look? just to see how good the coverage is? Ideally we should approach 100%, but as a practical rule of thumb, anything above our current coverage of ~92% is probably sufficient. Important thing Is that we're testing edge cases before they can cause bugs

tidy3d/components/lumped_element.py
```diff
@@ -107,24 +140,148 @@ def _voltage_axis_in_plane(cls, val, values):
...
+        ...,
+        title="Normal Axis",
+        description="Specifies the normal axis, which defines"
+        "the orientation of the circles making up the coaxial lumped element.",
     )
 
+    def to_mesh_overrides(self) -> List[MeshOverrideStructure]:
+        """Creates a suggested :class:`MeshOverrideStructure` list that could be added to the
+        :class:`Simulation`"""
+        # Create a mesh override when refinement is needed.
```
tylerflex: should this just be included in the method docstring?

tidy3d/components/lumped_element.py
```diff
@@ -107,24 +140,148 @@ def _voltage_axis_in_plane(cls, val, values):
...
+        # So in the normal direction, we'll make sure there are at least
+        # 2 cell layers above and below whose size is the same as the in-plane
+        # cell size in the override region. Also, to ensure that the port itself
+        # is aligned with a grid boundary in the normal direction, two separate
+        # override regions are defined, one above and one below the analytical
+        # port region.
+        mesh_overrides = []
+        size = 3 * [self.outer_diameter]
+        size[self.normal_axis] = 0
+        if self.num_grid_cells:
```
tylerflex, line 223: It seems like a lot of this logic could be simplified with the `pop_axis()` method, or at least assigning some variables up front. For example `override_dl[self.normal_axis]` is referenced a lot. 

```py
size = unpop_axis(0, (self.outer_diameter, self.outer_diameter), axis=self.normal_axis)
```

```py
dl = self.outer_diameter / self.num_grid_cells
override_dl = pop_axis(dl/2, (dl, dl), axis=self.normal_axis)
```
etc.

Up to you whether you think this makes things more complex, but in my read, it might be more declarative and easier to read.

tidy3d/components/lumped_element.py
```diff
@@ -107,24 +140,148 @@ def _voltage_axis_in_plane(cls, val, values):
...
+                    dl=override_dl,
+                )
+            )
+        return mesh_overrides
+
+    @pydantic.validator("center", always=True)
+    def _center_not_inf(cls, val):
+        """Make sure center is not infinitiy."""
+        if any(np.isinf(v) for v in val):
+            raise ValidationError("center can not contain td.inf terms.")
```
tylerflex: put 'td.inf' in single quotes

tylerflex: and `center`

tidy3d/components/lumped_element.py
```diff
@@ -107,24 +140,148 @@ def _voltage_axis_in_plane(cls, val, values):
...
+        return val
+
+    @pydantic.validator("inner_diameter", always=True)
+    @skip_if_fields_missing(["outer_diameter"])
+    def _ensure_inner_diameter_is_smaller(cls, val, values):
+        """Ensures that the inner diameter is smaller than the outer diameter, so that the final shape is an annulus."""
+        outer_diameter = values.get("outer_diameter")
+        if val >= outer_diameter:
+            raise ValidationError(
+                f"The inner diameter {val} of a coaxial lumped element must be less than the outer diameter {outer_diameter}."
```
tylerflex: maybe let's use the variable names `inner_diameter` and `outer_diameter` (in single quotes)

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -0,0 +1,163 @@
...
+        2, title="Axis", description="Specifies dimension of the planar axis (0,1,2) -> (x,y,z)."
+    )
+
+    position: float = pd.Field(
+        ...,
+        title="Position",
+        description="Position of the plane along the ``axis``.",
+    )
+
+    vertices: ArrayFloat2D = pd.Field(
```
tylerflex, line 47: I think we have a `Vertices` type in types.py, which is used for PolySlab. would it be useful here? or maybe better to keep things separate?

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -0,0 +1,163 @@
...
+        "If you wish to indicate a closed contour, the final vertex should be made "
+        "equal to the first vertex. ``vertices[-1]==vertices[0]``",
+        units=MICROMETER,
+    )
+
+    def compute_integral(
+        self, field: FieldParameter, em_field: MonitorDataTypes
+    ) -> IntegralResultTypes:
+        """Computes the path integral defined by ``vertices`` given the input ``em_field``.
+        The ``FieldParameter`` determines whether the electric or magnetic field is integrated."""
```
tylerflex: class link?

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -0,0 +1,163 @@
...
+        # Determine the differential length elements along the path
+        dl_x = self._compute_dl_component(x_path, self.is_closed_contour)
+        dl_y = self._compute_dl_component(y_path, self.is_closed_contour)
+        dl_x = xr.DataArray(dl_x, dims="s")
+        dl_y = xr.DataArray(dl_y, dims="s")
+
+        # Compute the dot product between differential length element and vector field
+        integrand = field1_interp * dl_x + field2_interp * dl_y
+        # Integrate along the path
+        result = integrand.integrate(coord="s")
```
tylerflex, line 120: nice 👍 so this DataArray indexing seems to work pretty well?

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -0,0 +1,163 @@
...
+        result = result.reset_coords(drop=True)
+
+        if isinstance(em_field, FieldData):
+            return FreqDataArray(data=result.data, coords=result.coords)
+        elif isinstance(em_field, FieldTimeData):
+            return TimeDataArray(data=result.data, coords=result.coords)
+        else:
+            assert isinstance(em_field, ModeSolverData)
+            return FreqModeDataArray(data=result.data, coords=result.coords)
```
tylerflex, line 129: it's sort of unfortunate in hindsight that we made all of these DataArrays different classes, instead of just something like
```py
DataArray[('x', 'y', 'z', 'f')](data=data, coords=coords)
```
but I guess we need to live with that for now.

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @staticmethod
+    def _compute_dl_component(coord_array: xr.DataArray, closed_contour=False) -> np.array:
+        """Computes the differential length element along the integration path."""
+        dl = np.gradient(coord_array)
+        if closed_contour:
+            end_points = (coord_array[1] - coord_array[-2]).values / 2
+            dl[0] = end_points
+            dl[-1] = end_points
```
tylerflex: explain these 3 Ines? a little confused here.

tylerflex: for example, why is `coord_array[0]` and `coord_array[-1]` ignored?

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -0,0 +1,163 @@
...
+        # overall shape of vertices
+        if val.shape[1] != 2:
+            raise SetupError(
+                "'CustomPathIntegral2D.vertices' must be a 2 dimensional array shaped (N, 2). "
+                f"Given array with shape of '{val.shape}'."
+            )
+        return val
+
+
+class CustomVoltageIntegral2D(CustomPathIntegral2D):
```
tylerflex, line 168: I think we went with the module approach (vs the class method approach) to these sorts of things? I'm not opposed to the class approach, but what's the reasoning? do we plan to add fields to the CustomPathIntegral classes down the line?

tidy3d/plugins/smatrix/ports/base_lumped.py
```diff
@@ -0,0 +1,101 @@
...
+    -------
+    >>> port_in = ['port1', 'port2']
+    >>> port_out = ['port1', 'port2']
+    >>> f = [2e14]
+    >>> coords = dict(
+    ...     port_in=ports_in,
+    ...     port_out=ports_out,
+    ...     f=f
+    ... )
+    >>> fd = LumpedPortDataArray((1 + 1j) * np.random.random((2, 2, 1)), coords=coords)
```
tylerflex, line 37: just FYI you can test the docstring examples with

```
pytest --doctest-modules tidy3d
```

tidy3d/plugins/smatrix/ports/base_lumped.py
```diff
@@ -0,0 +1,101 @@
...
+    __slots__ = ()
+    _dims = ("port_out", "port_in", "f")
+    _data_attrs = {"long_name": "lumped port matrix element"}
+
+
+class AbstractLumpedPort(Tidy3dBaseModel, ABC):
+    """Class representing a single lumped port"""
+
+    name: str = pd.Field(
+        ...,
```
tylerflex, line 49: might be other name validations added recently, eg in #1543 . Should we make sure these are applied here?

tidy3d/plugins/smatrix/ports/base_lumped.py
```diff
@@ -0,0 +1,101 @@
...
+        title="Name",
+        description="Unique name for the port.",
+        min_length=1,
+    )
+
+    impedance: Complex = pd.Field(
+        50,
```
tylerflex: might be good to make this a module constant, eg `DEFAULT_REFERENCE_IMPEDANCE = 50` after the imports.

This helps a bit with testing and documentation.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -0,0 +1,381 @@
...
+    @pd.validator("inner_diameter", always=True)
+    @skip_if_fields_missing(["outer_diameter"])
+    def _ensure_inner_diameter_is_smaller(cls, val, values):
+        """Ensures that the inner diameter is smaller than the outer diameter, so that the final
+        shape is an annulus."""
+        outer_diameter = values.get("outer_diameter")
+        if val >= outer_diameter:
+            raise ValidationError(
+                f"The inner diameter {val} of a coaxial lumped element must be less than the "
+                "outer diameter {outer_diameter}."
```
tylerflex: missing `f` for the fstring

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -0,0 +1,381 @@
...
+    @cached_property
+    def injection_axis(self):
+        """Injection axis of the port."""
+        return self.normal_axis
+
+    @pd.validator("center", always=True)
+    def _center_not_inf(cls, val):
+        """Make sure center is not infinity."""
+        if any(np.isinf(v) for v in val):
+            raise ValidationError("Center can not contain td.inf terms.")
```
tylerflex: single quotes for `td.inf` and probably make `center` lowercase and in single quotes as a field name.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -0,0 +1,381 @@
...
+        self, source_time: GaussianPulse, snap_center: float, grid: Grid
+    ) -> CustomCurrentSource:
+        """Create a current source from the lumped port."""
+        # Discretized source amps are manually zeroed out later if they
+        # fall on Yee grid locations outside the analytical source region.
+
+        # Get transverse axes
+        trans_axes = self.remaining_axes
+
+        coord1 = "xyz"[trans_axes[0]]
```
tylerflex: possible applications of `pop_axis` here and below, eg.
```py
coord1, (coord2, coord3) = pop_axis('xyz', axis=self.normal_axis)
```

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -0,0 +1,381 @@
...
+        # the current flowing in the core conductor in the positive direction.
+        if self.direction == "+":
+            current *= -1.0
+
+        return current
+
+    @cached_property
+    def remaining_axes(self) -> Tuple[Axis, Axis]:
+        """Get in-plane axes, ordered to maintain a right-handed coordinate system."""
+        axes: List[Axis] = [0, 1, 2]
```
tylerflex: I might be wrong, but I think for python 12 support, we need to use `list[Axis]` now (not typing.List), although this might not apply to method annotations.

tidy3d/plugins/smatrix/ports/rectangular_lumped.py
```diff
@@ -0,0 +1,265 @@
...
+        voltage_mon_size = list(self.size)
+        voltage_mon_size[self.injection_axis] = 0.0
+        voltage_mon_size[self.current_axis] = 0.0
+        # Create a voltage monitor
+        return FieldMonitor(
+            center=center,
+            size=voltage_mon_size,
+            freqs=freqs,
+            fields=[f"E{e_component}"],
+            name=f"{self.name}_voltage",
```
tylerflex: would be nice for this to be a method 

```py
def _voltage_monitor_name(self) -> str:
    return f"{self.name}_voltage"
```

so we could use it here and below. and then if it changes, we just need to change it once.

same for current

tidy3d/plugins/smatrix/ports/rectangular_lumped.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def compute_current(self, sim_data: SimulationData) -> FreqDataArray:
+        """Helper to compute current flowing through the port."""
+        # Diagram of contour integral, dashed line indicates location of sheet resistance
+        # and electric field used for voltage computation. Voltage axis is out-of-page.
+        #
+        #                                    current_axis = ->
+        #                                    injection_axis = ^
+        #
+        #                  |                   h2_field ->             |
```
tylerflex, line 170: should we include this in the docstring?

tylerflex (changes requested):
Thanks for this massive contribution @dmarek-flex 

I had a lot of small comments, but I wasn't able to fully go through everything in total detail.

As you can probably gather, I'm a fan of trying to use the more declarative helpers (eg pop_axis) whenever possible. There's a lot of necessary lower-level data manipulation in these new features. Whenever possible, I think it helps a lot to try to separate this from the higher level, more declarative stuff. For example, try to hide the procedural code into methods (like pop axis) so it can be reused as much as possible.

I gave some suggestions where I saw things, but there were many places where the same operations pop up.

As the RF /microwave codebase grows in complexity, eventually you'll probably hit a point where you'll want as much of this abstracted away as possible, so I'd just consider that moving forward.

Anyway, thanks a lot for the great work! I'm going to add @QimingFlex as a reviewer too to get another pair of eyes on it, specifically the physics part. and so he gets some review experience

tidy3d/components/geometry/utils_2d.py
dmarek-flex: Yea I agree! One complication is that this change is from a commit introduced in a [different PR](https://github.com/flexcompute/tidy3d/pull/1644). I need to use the changes in that PR or these coaxial ports are not possible. So right now, I have rebased onto that branch.

I'll address this in the other PR, unless you have a different suggestion.

tidy3d/components/geometry/base.py
dmarek-flex: Yes, I use a set to see if all geometries have identical normals, then I need to actually get the value of it for the next step. A clearer option would be perhaps `list(normals)[0]`

tidy3d/components/geometry/utils_2d.py
tylerflex: ah I see, I think what you suggest works.

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: I'll update the code with comments, but if you are interested it is because I would like to use a central difference scheme to approximate the derivative at the first and last point when the contour is closed. (The first and last point have identical positions for my definition of a closed contour). So I need the second point and the penultimate point to compute the values for the zeroth point(which is the same as the last point).

tylerflex: ok I see. I guess it might be worth noting that np.gradient probably uses 3 points in it's calculation (https://numpy.org/doc/stable/reference/generated/numpy.gradient.html) so is it worth doing this for the end point too? It seems this approximation kind of ignores `coord_array[0]`, which could make it less accurate?

tylerflex: Maybe a quick and easy fix could be something like
```py
if closed_contour:
    grad_end = np.gradient([coord_array[-2], coord_array[0],  coord_array[1]])
    dl[0] = dl[-1] = grad_end

```

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Convenience method `constant_loss_tangent_model` in `FastDispersionFitter` to fit constant loss tangent material model.
 - Classmethods in `DispersionFitter` to load complex-valued permittivity or loss tangent data.
 - Pre-upload validator to check that mode sources overlap with more than 2 grid cells.
+- Support `2DMedium` for `Transformed`/`GeometryGroup`/`ClipOperation` geometries
```
tylerflex: missing period

tests/test_components/test_medium.py
dmarek-flex: I have been aiming to keep it high, but I noticed I lack tests for the microwave models I added. I'll fix that in a separate [issue](https://github.com/flexcompute/tidy3d/issues/1692). Everything else is above 95%, except for the base class in 

> tidy3d/plugins/smatrix/component_modelers/base.py

, which involves some Batch type function calls that I am not exactly sure how to test.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
dmarek-flex: Should I also extend that change to things like `tuple`?

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
dmarek-flex: The only thing that prevents me from using pop_axis here is I find it easier to always remain in a 2D coordinate system that is equivalent to "xy". If "y" is removed I prefer "zx" as the order of coordinates so that I don't need to introduce any minus signs when performing contour integrals. That is why I have that logic in `self.remaining_axes`.

But I can add a method that will act like `pop_axis` but returns the order that I find convenient.

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: I copied the definition from Polyslab where `ArrayFloat2D` is used. However, maybe the fact that I expect vertices in a different order would be confusing. You'll notice that I ask users to supply vertices as (z,x). which is different from `PolySlab`. Do you think I should change this to follow `PolySlab`?

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: Yea, its pretty simple!

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: It has its pros and cons. The name of the class removes some burden from the user having to think about coordinates. But yeah that logic was the best I could think of that would return expected types.

tidy3d/plugins/smatrix/ports/base_lumped.py
dmarek-flex: Haha, that just saved me so much time!

tidy3d/plugins/smatrix/ports/base_lumped.py
dmarek-flex: I don't think I understand. `AbstractLumpedPort` inherits from `Tidy3dBaseModel` so it should include that validator right?

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: Well both methods give the same result (the last equation of the note that you linked is what I was emulating). However, I will concede that your fix is nicer looking :smile: !

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -0,0 +1,379 @@
...
+        trans_axes = self.remaining_axes
+        coord1 = "xyz"[trans_axes[0]]
+        coord2 = "xyz"[trans_axes[1]]
+
+        H1 = "H" + coord1
+        H2 = "H" + coord2
+
+        # Size of current monitor needs to encompass the current carrying 2D sheet
+        # Needs to have a nonzero thickness so a closed loop of gridpoints around the 2D sheet can be formed
+        dl = 2 * (increment_float(center[self.normal_axis], 1.0) - center[self.normal_axis])
```
weiliangjin2021: How is the subnormal float handled here? np.nextafter 0 is about 1e-45, while the minimal positive normal number is 1e-38.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -0,0 +1,379 @@
...
+        coord2 = "xyz"[trans_axes[1]]
+
+        E1 = "E" + coord1
+        E2 = "E" + coord2
+
+        # Size of voltage monitor can essentially be 1D from ground to signal conductor
+        # We arbitrarily choose the positive first axis as the location for the path integral
+        # Any of the four possible choices should give the same result.
+        voltage_mon_size = [0] * 3
+        voltage_mon_size[trans_axes[0]] = self.outer_diameter
```
weiliangjin2021: Can size and center be the same as here https://github.com/flexcompute/tidy3d/blob/a385e15718ffc8a337c2b535f99fdbbbb0d50bae/tidy3d/plugins/smatrix/ports/coaxial_lumped.py#L273-L277?

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -0,0 +1,379 @@
...
+        coord3 = "xyz"[self.normal_axis]
+
+        # Just need a rough estimate for the number of cells
+        field_coords = field_data.field_components["H" + coord1].coords
+
+        # Estimate number of points needed along path integral
+        num_coords_1 = len(field_coords[coord1].values)
+        num_coords_2 = len(field_coords[coord2].values)
+        num_coords = max(num_coords_1, num_coords_2)
+        num_path_coords = round(np.pi * num_coords / 4) * 4 + 1
```
weiliangjin2021, line 311: add some explanations on division by 4 and then multiply back?

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -0,0 +1,379 @@
...
+
+        # These x,y coordinates are relate to the local coordinate frame
+        xt, yt = generate_circle_coordinates(
+            (self.outer_diameter + self.inner_diameter) / 4, num_path_coords
```
weiliangjin2021, line 315: So the loop is in the middle of the 2 conductors. Will be good to add that to the docstring of `compute_current`.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -0,0 +1,379 @@
...
+    normal_axis: Axis = pd.Field(
+        ...,
+        title="Normal Axis",
+        description="Specifies the axis which is normal to the concentric circles.",
+    )
+
+    direction: Direction = pd.Field(
+        ...,
+        title="Direction",
+        description="Specifies which side of the port the inner conductor is extending from. "
```
weiliangjin2021: Is this equivalent to the injection direction in the mode source? If so, it's easier to understand in that way.

weiliangjin2021 (commented):
This is tons of work! First pass except for custom path integral part. All look solid and easy to follow. One question when you compute the current by using a loop in the middle of the two conductors: I assume the current computation will depend on the loop size if the dielectric material between the two conductors is lossy. How is lossy dielectric usually handled in lumped port?

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: The reasoning here is that although these classes are basically just renaming the base class they do have data associated with them, so overall I think they make sense as a class. Originally I did have some additional members but I realized that they were more confusing than helpful. (e.g. I had a sign parameter, but users can just define their path in the opposite direction if they wanted).

dmarek-flex: Also to mirror how the AxisAligned (in path_integrals.py) versions of these integrals are used.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
dmarek-flex: I fixed this in `increment_float` itself now. The result of `increment_float` could still be a subnormal number, but it will differ by the original value by at least 1e-38.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
dmarek-flex: Yes, they can. I moved those computations into a cached property.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
dmarek-flex: It is sort of equivalent, but not exactly. The direction parameter has no effect on the fields produced by the underlying current source. LumpedPorts cannot provide that type of directional behavior, since they only contain electrical currents. 

This direction parameter is needed solely for positioning the current path integral, which needs to be placed in such a way that it surrounds a conduction current, which means it needs to surround some type of conductor or else the displacement current will make the computation less accurate. 

For the coaxial port the best method I could come up with is using the inner conductor that should be present in the `Simulation`. However, I need the user to tell me which side of the port the inner conductor is positioned. I am open to suggestions though.

dmarek-flex:
> This is tons of work! First pass except for custom path integral part. All look solid and easy to follow. One question when you compute the current by using a loop in the middle of the two conductors: I assume the current computation will depend on the loop size if the dielectric material between the two conductors is lossy. How is lossy dielectric usually handled in lumped port?

At the end of the day the computation for current is an approximation. It should be more accurate at low frequencies, when fields are closer to TEM, when path is close to the surfaces of conductors.  As long as the lossy material does not impact the field distribution a lot, the computation should still be accurate. 

We could allow for an optional parameter for the user to place the contour where ever they want. The reason I chose the midpoint is that usually there are only a handful of grid cells in the port and I want to make sure that the path is far enough away from the conductors, so that staircasing and conformal mesh affects to the fields are avoided.

tidy3d/plugins/smatrix/ports/rectangular_lumped.py
dmarek-flex: Is there a way to keep the white space formatting? It doesn't really look nice

tidy3d/components/lumped_element.py
```diff
@@ -107,24 +140,148 @@ def _voltage_axis_in_plane(cls, val, values):
...
+        # The port is a flat surface, but when computing the port current,
+        # we'll eventually integrate the magnetic field just above and below
+        # this surface, so the mesh override needs to ensure that the mesh
+        # is fine enough not only in plane, but also in the normal direction.
+        # So in the normal direction, we'll make sure there are at least
+        # 2 cell layers above and below whose size is the same as the in-plane
+        # cell size in the override region. Also, to ensure that the port itself
+        # is aligned with a grid boundary in the normal direction, two separate
+        # override regions are defined, one above and one below the analytical
+        # port region.
```
QimingFlex: This refinement along the normal axis ensures that the H field is as close as possible to the sheet, thereby allowing for a more accurate approximation of the current? How does this affect accuracy?

dmarek-flex: That is actually a great question :smile:! I tried a simulation of a microstrip with regular Rectangular lumped ports. Here are three sets of results where I have computed the root mean square error of the 4 S parameters:

```
# Refine Ports Off
# RMSE td:  0.11258103731991326
# RMSE td:  0.16233394675835855
# RMSE td:  0.1623539382006575
# RMSE td:  0.11795092811347076

# Refine Ports On with Normal Refinement (10 grid cells across port)
# RMSE td:  0.09013413380040426
# RMSE td:  0.11568887419087617
# RMSE td:  0.11568856206400055
# RMSE td:  0.09021164537686488

# Refine Ports On without Normal Refinement (10 grid cells across port)
# RMSE td:  0.10821387714324963
# RMSE td:  0.09544047779802187
# RMSE td:  0.09544046629736976
# RMSE td:  0.10825511559197067
```

So definitely port refinement can improve accuracy, but I think when refining along the normal axis it may be detrimental, or at least it does not improve the accuracy much. I think this is due to the self-inductance, which is related to the size and distribution of the current source and smaller size along the normal axis increases the self-inductance. It's affect on S parameters is also more pronounced at higher frequencies.

Shashwat originally implemented this functionality, which also helped in situations where the original grid did not exactly line up with the port center along the normal axis. Since then I improved how path integrals are performed and how ports are snapped to grid cell boundaries, so it might be less necessary.

But I think modifying this part of the code should be left for a different PR, what do you think?

QimingFlex: Thanks, @dmarek-flex! I raised up this question because the refinement along the normal direction introduces some computational costs, including an increased number of cells and a restricted timestep. It appears that this refinement doesn’t lead to improved accuracy, so could we consider reducing this refinement in the future? I have no concerns at this moment, all of these setup are solid. I just want to confirm that the final setup will not require intensive mesh refinement along the normal direction.

QimingFlex (approved):

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
weiliangjin2021: Probably clearer to say "the direction of the signal travelling in the transmission line"? Sometimes I found it hard to visualize "the side of the port the inner conductor is extending from", because I usually picture that the conductor is starting from the port.

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @staticmethod
+    def _compute_dl_component(coord_array: xr.DataArray, closed_contour=False) -> np.array:
+        """Computes the differential length element along the integration path."""
+        dl = np.gradient(coord_array)
+        if closed_contour:
+            # If the contour is closed, we can use central different on the starting/end point
+            # which will be more accurate than the default forward/backward choice in np.gradient
+            grad_end = np.gradient([coord_array[-2], coord_array[0], coord_array[1]])
+            dl[0] = dl[-1] = grad_end[1]
```
weiliangjin2021, line 139: This is equivalent to `dl[0] = dl[-1] = (coord_array[1]-coord_array[-2])/2`? If so, this is clearer and more efficient than `np.gradient`?

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -0,0 +1,163 @@
...
+        return dl
+
+    @cached_property
+    def is_closed_contour(self) -> bool:
+        return (self.vertices[0, :] == self.vertices[-1, :]).all()
```
weiliangjin2021: Will exact equal potentially return false negative?

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -0,0 +1,163 @@
...
+from .path_integrals import AbstractAxesRH
+
+MonitorDataTypes = Union[FieldData, FieldTimeData, ModeSolverData]
+EMScalarFieldType = Union[ScalarFieldDataArray, ScalarFieldTimeDataArray, ScalarModeFieldDataArray]
+IntegralResultTypes = Union[FreqDataArray, FreqModeDataArray, TimeDataArray]
+FieldParameter = Literal["E", "H"]
+
+
+class CustomPathIntegral2D(AbstractAxesRH):
+    """Class for defining a custom path integral defined as a curve on an axis-aligned plane."""
```
weiliangjin2021: Will be helpful to provide a formula here explaining how path integral is performed; I think it's based on trapezoidal rule, and thus 2nd order convergence?

weiliangjin2021 (approved):
A few minor comments. Otherwise looks good.

tidy3d/components/lumped_element.py
dmarek-flex: Yes, agreed. I can't think of a simple way to set the normal axis grid cell size right now that would make more sense than the way it is presently done. We should revisit this choice in the future I think. I just added an [issue](https://github.com/flexcompute/tidy3d/issues/1702) to track this.

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: @tylerflex made the opposite point :smiley:  I think reusing the gradient function is also pretty clear, it shows that we are applying the same operation but just in a periodic fashion. Your suggestion is probably more efficient, but maybe not observable.

tylerflex: I dont think it's equivalent @weiliangjin2021 . The `np.gradient` I think uses a 2nd order derivative, [link](https://numpy.org/doc/stable/reference/generated/numpy.gradient.html). and since the rest of this function uses `np.gradient()` I assumed it might make sense to use the same scheme over the whole array.

dmarek-flex: In the case where the spacing is uniform it should be equivalent, which should correspond to my case. See the last equation in that link.

tidy3d/plugins/smatrix/ports/rectangular_lumped.py
tylerflex: hm, yea not sure. I guess we can just leave it as a code comment.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
tylerflex: oh right, yea `tuple[Axis, Axis]`

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
tylerflex: I see. Hm, if you find the other method convenient and can re-use it in other places, feel free. otherwise feel free to ignore.

tidy3d/plugins/microwave/custom_path_integrals.py
tylerflex: I do actually think a consistent convention throughout would be nice... we decided to use the dimension "popping" convention long ago.. but it was a bit of an arbitrary choice that we're stuck with in a lot of places.

@momchil-flex thoughts?

tidy3d/plugins/microwave/custom_path_integrals.py
tylerflex: interesting. Are there future plans to make the spacing non-uniform? I guess I don't see any issue with being consistent about how the derivative is computed.

dmarek-flex: The geometric spacing between vertices can already be non-uniform. The numpy gradient function is just a more generalized function where you could give it coordinates to perform the gradient over. We are simply using it to perform central differences, where the implied spacing is 1.

tylerflex: ooh, I see.

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: Here is added explanation rendered in the docs:

image

tidy3d/plugins/microwave/custom_path_integrals.py
tylerflex: another reason why I like np.gradient: I think I mentioned in another comment, that I prefer the more declarative, high level stuff (like `np.gradient()`, `pop_axis()` all else equal. And if we do have to perform mathematical operations, such as `(f1-f2)/2.0`, ideally that stuff should be abstracted away in its own function. I think this helps with readability.

tidy3d/plugins/microwave/custom_path_integrals.py
weiliangjin2021: nice

tidy3d/plugins/microwave/custom_path_integrals.py
weiliangjin2021: I forgot about nonuniform vertices. Then np.gradient is a good choice here.

dmarek-flex: After playing around with a notebook, using some of these computations more, I think I will change the behaviour of custom path integral. Think about a rectangular path where the user has defined path `vertices` as the corners of a rectangle. The computed `dl` is not accurate at the corner. It is probably better to interpolate the fields at the midpoints of each line segment, where the `dl` can be computed accurately.

dmarek-flex: Now, after sleeping on it. I decided to keep it the way it is. As long as enough vertices are used the result is accurate, and what I mentioned above would be much more complicated to implement.

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: It was a simple fix to revert to the polyslab ordering of vertex coordinates.

tylerflex (approved):
@dmarek-flex looks good to me, barring a few changes to get the tests working, I think this should be basically good to go?

dmarek-flex:
> @dmarek-flex looks good to me, barring a few changes to get the tests working, I think this should be basically good to go?

Yep, think I just need to rebase and the tests will pass again.

tylerflex:
looks like it just needs submodule update?

dmarek-flex:
> looks like it just needs submodule update?

Yea, but I am also trying to diagnose a regression on the backend tests that I just noticed before merging. 

Ok fixed, it was something that I was testing that somehow snuck into the commits

PR #1671, opened 2024-05-03
feat: Support selection by nearest value in `JaxDataArray`

### LINKED ISSUE #1617: `method="nearest"` support for `JaxDataArray`
Add support for `method="nearest"` in `JaxDataArray.sel`.

For example, calling `arr.sel(x=0, method="nearest")` will select the data at the `x` nearest to `0`.

See https://docs.xarray.dev/en/latest/generated/xarray.DataArray.sel.html

Implementation could look something like this:

```py
def sel_single(self, key, value, method="nearest"):
    coords = self.coords[key]
    argmin = np.argmin(np.abs(coords-value))
    return self.isel_single(key, argmin)
```

yaugenst-flex:
Closes #1617

Also took the liberty to rework the mechanics of the old selection by vectorizing everything using numpy.
Note that there is a slight difference between xarray's exact selection and ours now, because I base selections off of `np.isclose` instead of doing exact float comparisions (`x == y`). I think this is more intuitive, as otherwise something like `x == 0.2` fails if `x` is a numpy float (`0.20000000000000018`).
I can change this  back to match xarray's behavior exactly if that's what we want to do.

### DIFF
--- a/tests/test_plugins/test_adjoint.py
+++ b/tests/test_plugins/test_adjoint.py
@@ -7,6 +7,8 @@ import pytest
 import pydantic.v1 as pydantic
 import jax.numpy as jnp
 import numpy as np
+from numpy.testing import assert_allclose
+from xarray import DataArray
 from jax import grad
 import jax
 import time
@@ -836,6 +838,76 @@ def test_jax_data_array():
     assert da3.shape == (3, 2, 2)
 
     assert da2 == da3
+
+    n = 11
+    cs = list(range(n))
+    vals = np.random.uniform(0, 1, (n, n, 1, 1, 2))
+    coords = dict(x=cs, y=cs, z=[0], f=[0], direction=["+", "-"])
+
+    jda = JaxDataArray(values=vals, coords=coords)
+    xda = DataArray(data=vals, coords=coords)
+
+    # test exact, single value, non-numeric
+    xda_vals = xda.sel(direction="+", method=None).values
+    jda_vals = jda.sel(direction="+", method=None).values
+    assert_allclose(xda_vals, jda_vals)
+
+    # test exact, single value, 1d coordinate
+    xda_vals = xda.sel(f=0, method=None).values
+    jda_vals = jda.sel(f=0, method=None).values
+    assert_allclose(xda_vals, jda_vals)
+
+    # test exact, single value, 1d coordinate as list
+    xda_vals = xda.sel(f=[0], method=None).values
+    jda_vals = jda.sel(f=[0], method=None).values
+    assert_allclose(xda_vals, jda_vals)
+
+    # # disabled because it fails on xarray's side
+    # # test nearest, single value, non-numeric
+    # xda_vals = xda.sel(direction="+", method="nearest").values
+    # jda_vals = jda.sel(direction="+", method="nearest").values
+    # assert_allclose(xda_vals, jda_vals)
+
+    # test exact, multiple values, non-numeric
+    xda_vals = xda.sel(direction=["+", "-"], method=None).values
+    jda_vals = jda.sel(direction=["+", "-"], method=None).values
+    assert_allclose(xda_vals, jda_vals)
+
+    # test nearest, single value
+    xda_vals = xda.sel(x=0.1, method="nearest").values
+    jda_vals = jda.sel(x=0.1, method="nearest").values
+    assert_allclose(xda_vals, jda_vals)
+
+    # test nearest, single value, 1d coordinate
+    xda_vals = xda.sel(f=0.1, method="nearest").values
+    jda_vals = jda.sel(f=0.1, method="nearest").values
+    assert_allclose(xda_vals, jda_vals)
+
+    # test nearest, single value, 1d coordinate as list
+    xda_vals = xda.sel(f=[0.1], method="nearest").values
+    jda_vals = jda.sel(f=[0.1], method="nearest").values
+    assert_allclose(xda_vals, jda_vals)
+
+    # test nearest, multiple values
+    xda_vals = xda.sel(x=[0.0, 0.1, 0.3], method="nearest").values
+    jda_vals = jda.sel(x=[0.0, 0.1, 0.3], method="nearest").values
+    assert_allclose(xda_vals, jda_vals)
+
+    # test nearest, multiple coords, single value
+    xda_vals = xda.sel(x=0.1, y=-0.3, method="nearest").values
+    jda_vals = jda.sel(x=0.1, y=-0.3, method="nearest").values
+    assert_allclose(xda_vals, jda_vals)
+
+    # test nearest, multiple coords, multiple values
+    xda_vals = xda.sel(x=[-1.5, 0.1], y=[-0.3, 0.0, 0.9], method="nearest").values
+    jda_vals = jda.sel(x=[-1.5, 0.1], y=[-0.3, 0.0, 0.9], method="nearest").values
+    assert_allclose(xda_vals, jda_vals)
+
+    # test nearest, multiple coords, duplicate values
+    xda_vals = xda.sel(x=[-1.5, 0.2, 0.1, -1.5, 0.2], y=[-0.3, 0.0, 0.9], method="nearest")
+    jda_vals = jda.sel(x=[-1.5, 0.2, 0.1, -1.5, 0.2], y=[-0.3, 0.0, 0.9], method="nearest")
+    assert_allclose(xda_vals.values, jda_vals.values)
+    assert_allclose(xda_vals.coords["x"], jda_vals.coords["x"])
 
 
 def test_jax_sim_data(use_emulated_run):
--- a/tidy3d/plugins/adjoint/components/data/data_array.py
+++ b/tidy3d/plugins/adjoint/components/data/data_array.py
@@ -1,7 +1,8 @@
 """Defines jax-compatible DataArrays."""
+
 from __future__ import annotations
 
-from typing import Tuple, Any, Dict, List
+from typing import Tuple, Any, Dict, List, Literal, Sequence, Union
 
 import h5py
 import pydantic.v1 as pd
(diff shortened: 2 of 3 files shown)


tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,28 +313,37 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
 
         return self_sel
 
-    def sel(self, indexers: dict = None, method: str = "nearest", **sel_kwargs) -> JaxDataArray:
+    def sel(self, indexers: dict = None, method: str | None = None, **sel_kwargs) -> JaxDataArray:
```
tylerflex: does this `|` fail in earlier versions of python? we actually recently stopped supporting 3.8 so maybe it's ok now

tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,28 +313,37 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
...
         """Select a value from the :class:`.JaxDataArray` by indexing into coordinate values."""
+        if method is None:
+            return self._sel_exact(sel_kwargs)
+        elif method == "nearest":
+            return self._sel_nearest(sel_kwargs)
+        else:
+            raise NotImplementedError(f"Unkown selection method: {method}.")
```
tylerflex, line 338: `ValueError` might be more appropriate but I'm not totally sure about that.

tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,28 +313,37 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
 
         return self_sel
 
-    def sel(self, indexers: dict = None, method: str = "nearest", **sel_kwargs) -> JaxDataArray:
+    def sel(self, indexers: dict = None, method: str | None = None, **sel_kwargs) -> JaxDataArray:
         """Select a value from the :class:`.JaxDataArray` by indexing into coordinate values."""
```
tylerflex: could you add a short mention of `method` option?

EDIT: actually let's just add the full `Parameters` and `Returns` here. I should have done that earlier..

tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,28 +313,37 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
...
+        if method is None:
+            return self._sel_exact(sel_kwargs)
+        elif method == "nearest":
+            return self._sel_nearest(sel_kwargs)
+        else:
+            raise NotImplementedError(f"Unkown selection method: {method}.")
+
+    def _sel_exact(self, sel_kwargs: dict) -> JaxDataArray:
```
tylerflex: could you add a short docstring (and to `_sel_nearest`)

tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,28 +313,37 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
...
-                    raise DataError(
-                        f"Could not select '{coord_name}={sel_kwarg}', value not found."
-                    )
-                coord_index = coord_list.index(sel_kwarg)
-                isel_kwargs[coord_name] = coord_index
+        for coord_name, vals in sel_kwargs.items():
+            coord_list = np.asarray(self.get_coord_list(coord_name))
+            vals = np.atleast_1d(vals)
+            dist = np.abs(coord_list[None] - vals[:, None])
+            indices = np.where(np.isclose(dist, 0))[1]
```
tylerflex: Just so I understand, the only difference between _sel_exact and _sel_nearest seems to be how the indices are computed given the `coord_list` and `vals`

```py
# exact
vals = np.atleast_1d(vals)
dist = np.abs(coord_list[None] - vals[:, None])
indices = np.where(np.isclose(dist, 0))[1]
if indices.size == 0:
    raise DataError(f"Could not select '{coord_name}={vals}', some values were not found.")
```

```py
# nearest
vals = np.asarray(vals)
dist = np.abs(coord_list[:, None] - vals[None])
indices = np.argmin(dist, axis=0)
```

Is my understanding correct?

Furthermore, it could probably be written to just compute indices given some `dist` array.

Would it make sense then to refactor a bit to either just stick these in the same method (with an if / else) or just localize the `_get_indices(dist) -> incides` logic?

tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,28 +313,37 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
...
+                )
+            isel_kwargs[coord_name] = np.squeeze(indices)
+        return self.isel(**isel_kwargs)
+
+    def _sel_nearest(self, sel_kwargs: dict) -> JaxDataArray:
+        isel_kwargs = {}
+        for coord_name, vals in sel_kwargs.items():
+            coord_list = np.asarray(self.get_coord_list(coord_name))
+            vals = np.asarray(vals)
+            dist = np.abs(coord_list[:, None] - vals[None])
```
tylerflex: can you explain the indexing here? so it seems to me that we add a dimension to `vals` so it would be shaped `(N,) -> (1, N)` and then `coord_list` would be `(M,) -> (M, 1)`. And then we have a distance `(M, N)`. We take argmin with respect to the `coord_list` axis, giving `(N,)` values corresponding to the smallest distance between each value to all of the coords?

tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,28 +313,37 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
...
-                if sel_kwarg not in coord_list:
-                    raise DataError(
-                        f"Could not select '{coord_name}={sel_kwarg}', value not found."
-                    )
-                coord_index = coord_list.index(sel_kwarg)
-                isel_kwargs[coord_name] = coord_index
+        for coord_name, vals in sel_kwargs.items():
+            coord_list = np.asarray(self.get_coord_list(coord_name))
+            vals = np.atleast_1d(vals)
+            dist = np.abs(coord_list[None] - vals[:, None])
```
tylerflex: Similar comment below but trying to understand the idea here, so `coord_list[None]` gives an array shaped `(1, M)` and then `vals[:, None]` gives `(1, N)`, we take the difference and get an array shaped `(N, M)`.

The next line, by applying `np.where(...)[1]`, we select out all of the `N` indices into `coord_list`?

What's a little confusing is that in the nearest method, it seems the `dist` array is transposed compared to this one, it might be better to either keep them consistent, or probably just to refactor so that we compute this `dist` method for both `nearest` and `exact` and then have a separate logic for computing `indices` given the `dist` (for nearest and exact)?

tests/test_plugins/test_adjoint.py
tylerflex, line 1: this is great 👍

tylerflex (commented):
just some preliminary comments, looking good though

tidy3d/plugins/adjoint/components/data/data_array.py
tylerflex: Also, maybe we want it to be something like `method: typing.Literal[None, "nearest"]`?

yaugenst-flex: > does this | fail in earlier versions of python? we actually recently stopped supporting 3.8 so maybe it's ok now

Yes absolutely! Just checked, it was introduced in 3.10, so I'll have to change that. I'd _really_ like to use this & `match` statements though...

> Also, maybe we want it to be something like method: typing.Literal[None, "nearest"]?

Yes I was thinking about the `Literal` too, I'll add it. Not sure why I didn't go with it in the end. Probably it seemed redundant given the `if/else` check.

tidy3d/plugins/adjoint/components/data/data_array.py
yaugenst-flex: Was considering that. My thinking was that there are in principle more methods (supported by xarray), but we don't implement them. `ValueError` also makes sense, I think it's fine either way.

tidy3d/plugins/adjoint/components/data/data_array.py
yaugenst-flex: Yes the calculation of the distance matrix should probably be refactored into its own method, you're right.

tylerflex: well, after looking at it (see comments below) probably we can just keep the same distance method calculation, but we just need to refactor so that we compute the indices differently given the distance and some `method`?

tidy3d/plugins/adjoint/components/data/data_array.py
yaugenst-flex: Yes exactly, which is then an index array of shape `(M,)` that can be used to index into `coord_list`. I came across a bug though, in that this does not handle non-numeric types such as coords of the form `["+", "-"]`. Have to come up with a fix for that.

tidy3d/plugins/adjoint/components/data/data_array.py
tylerflex: ok, I'm fine leaving as is

tidy3d/plugins/adjoint/components/data/data_array.py
yaugenst-flex: Yes, the reason for doing this was that it preserves the ordering of the coordinates this way. I'll have another look if I can do that without swapping the indices.

tidy3d/plugins/adjoint/components/data/data_array.py
tylerflex: my main concern was the docs, but actually I realized methods get documented using their docstrings (not the type annotations). We probably want to just add a proper `Parameters` and  `Returns` format in the docstring, mentioning `method`.  Could you add that while you're at it?

tidy3d/plugins/adjoint/components/data/data_array.py
tylerflex: ah.. maybe in that case we should just revert to the exact selection? not sure how xarray handles this.

tests/test_plugins/test_adjoint.py
tylerflex: let's add a test for the `direction` coords too

tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,30 +313,76 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
...
-    def sel(self, indexers: dict = None, method: str = "nearest", **sel_kwargs) -> JaxDataArray:
-        """Select a value from the :class:`.JaxDataArray` by indexing into coordinate values."""
+    def sel(
+        self, indexers: dict = None, method: Literal[None, "nearest"] = None, **sel_kwargs
+    ) -> JaxDataArray:
+        """Select a value from the :class:`.JaxDataArray` by indexing into coordinates by value.
+
+        Parameters
+        ----------
+        method : Literal[None, "nearest"] = None
```
tylerflex, line 326: we'll want to add `sel_kwargs` I think here (probably first parameter, just for the docs display). Just so users understand that they need to pass key: value pairs corresponding to the keys and values of the `.coords` they want to select.

tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,30 +313,76 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
...
-                coord_index = coord_list.index(sel_kwarg)
-                isel_kwargs[coord_name] = coord_index
+
+            try:  # handle non-numeric types (e.g. str)
+                coord_list = jnp.asarray(coord_list)
+            except TypeError:
+                isel_kwargs[coord_name] = self._indices_literal(coord_list, vals)
+                continue
+
+            vals = jnp.atleast_1d(vals)
```
tylerflex: nice 👍

tidy3d/plugins/adjoint/components/data/data_array.py
```diff
@@ -312,30 +313,76 @@ def isel(self, **isel_kwargs) -> JaxDataArray:
...
+                indices = jnp.where(jnp.isclose(dist, 0))[1]
+            elif method == "nearest":
+                indices = jnp.argmin(dist, axis=1)
+
+            if indices.size == 0:
+                raise DataError(
+                    f"Could not select '{coord_name}={vals}', some values were not found."
+                )
+
+            isel_kwargs[coord_name] = jnp.squeeze(indices)
```
tylerflex: just want to make sure: this `squeeze` doesn't have any unexpected effects if say my values array has a dimension of size 1? I want to make sure we dont "squeeze" out any dimensions in the values that still have a coordinate in `coords`. Can we add a test for this edge case?

tylerflex (commented):
Looking almost ready, just a couple comments.

tidy3d/plugins/adjoint/components/data/data_array.py
yaugenst-flex: added

tylerflex: sorry, nitpick-y comment, but I think saying that it's a dictionary might be confusing. Could we just change "Dictionary" to "Keyword arguments"?

tylerflex (commented):
looks good, just a comment on the docstring and then after [this](https://github.com/flexcompute/tidy3d/pull/1671/files#r1589388198) is addressed we can squash and merge.

tidy3d/plugins/adjoint/components/data/data_array.py
yaugenst-flex: Oh nice yea good catch, there is a difference between `.sel(x=0)` and `.sel(x=[0])`. Fixed with tests, matches xarray behavior.

tylerflex (approved):
Thanks @yaugenst I'll squash and merge this once the tests pass!

PR #1672, opened 2024-05-03
Fast Dispersion Fitter: keep poles slightly away from input freqs

### LINKED ISSUE #1631: Division by 0 error in FastFitter
The code to reproduce the error:
```
import numpy as np
from tidy3d.plugins.dispersion import FastDispersionFitter

N = 2
wvl_um = np.linspace(0.5, 0.6, N)
n_data = np.ones(N) * 2
k_data = np.ones(N) * 0.5

fitter = FastDispersionFitter(wvl_um = wvl_um, n_data = n_data, k_data = k_data)
medium, rms_error = fitter.fit(max_num_poles=2, tolerance_rms=1e-3)
```

Error message:

> tidy3d_frontend/tidy3d/plugins/dispersion/fit_fast.py:360: RuntimeWarning: divide by zero encountered in divide
>   pole_matrix[:, offset + 2 * i] = 1 / (1j * omega + pole) + 1 / (
> File "tidy3d_frontend/tidy3d/plugins/dispersion/fit_fast.py", line 498, in fit_residues
>     x_vector = scipy.optimize.lsq_linear(a_matrix_real, b_vector_real).x
>                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "python3.11/site-packages/scipy/optimize/_lsq/lsq_linear.py", line 316, in lsq_linear
>     unbd_lsq = np.linalg.lstsq(A, b, rcond=-1)
>                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "python3.11/site-packages/numpy/linalg/linalg.py", line 2326, in lstsq
>     x, resids, rank, s = gufunc(a, b, rcond, signature=signature, extobj=extobj)
>                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "python3.11/site-packages/numpy/linalg/linalg.py", line 124, in _raise_linalgerror_lstsq
>     raise LinAlgError("SVD did not converge in Linear Least Squares")
> numpy.linalg.LinAlgError: SVD did not converge in Linear Least Squares

caseyflex:
Sometimes in the fitter, the poles move close to the input freqs. When they move too close, the response function is invalid and the fitter errors. This PR keeps them a certain distance (1e-10 rad/s) away.

This PR should fix this issue:
https://github.com/flexcompute/tidy3d/issues/1631

### DIFF
--- a/tidy3d/plugins/dispersion/fit_fast.py
+++ b/tidy3d/plugins/dispersion/fit_fast.py
@@ -33,6 +33,9 @@ DEFAULT_TOLERANCE_RMS = 1e-5
 # this avoids divide by zero errors with lossless poles
 SCALE_FACTOR = 1.01
 
+# when poles are close to omega, can cause invalid response function, and we reject model
+OMEGA_POLE_CLOSE_ATOL = 1e-10
+
 
 class AdvancedFastFitterParam(Tidy3dBaseModel):
     """Advanced fast fitter parameters."""
@@ -520,6 +523,16 @@ class FastFitterData(AdvancedFastFitterParam):
         """
 
         model = self.iterate_poles()
+
+        # if any of the poles align with the test frequencies exactly, we reject the new model
+        if any(
+            np.isclose(omega, 1j * pole, rtol=0, atol=OMEGA_POLE_CLOSE_ATOL)
+            or np.isclose(omega, -1j * pole, rtol=0, atol=OMEGA_POLE_CLOSE_ATOL)
+            for omega in self.omega
+            for pole in self.complex_poles
+        ):
+            return self
+
         model = model.fit_residues()
 
         return model
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bug in plotting and computing tilted plane intersections of transformed 0 thickness geometries.
 - `Simulation.to_gdspy()` and `Simulation.to_gdstk()` now place polygons in GDS layer `(0, 0)` when no `gds_layer_dtype_map` is provided instead of erroring.
 - `task_id` now properly stored in `JaxSimulationData`.
+- Bug in `FastDispersionFitter` when poles move close to input frequencies.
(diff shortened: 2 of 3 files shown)


tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -519,6 +522,16 @@ def iterate_fit(self) -> FastFitterData:
...
         model = self.iterate_poles()
+
+        # if any of the poles align with the test frequencies exactly, we reject the new model
+        if any(
+            np.isclose(omega, 1j * pole, rtol=0, atol=OMEGA_POLE_CLOSE_ATOL)
+            or np.isclose(omega, -1j * pole, rtol=0, atol=OMEGA_POLE_CLOSE_ATOL)
+            for omega in self.omega
+            for pole in self.complex_poles
+        ):
+            return self
```
tylerflex, line 534: do we want to warn or anything here?

tylerflex (approved):
Looks good, just wondering if we should warn when this gets triggered

tidy3d/plugins/dispersion/fit_fast.py
caseyflex: thanks. @weiliangjin2021 , do you think we should warn here? I actually think the resulting model in these cases is as expected / as desired, so no need to warn the user. The fix is just preventing the fitter from trying to further improve a model which is already quite good, when the improvement would limit in this error. But I'm not sure exactly about your use case.

caseyflex: I could also use fp_eps instead of 1e-10. Might be more robust that way.

weiliangjin2021: Yeah, since it's accurate enough, I think no need to warn.

weiliangjin2021 (approved):

momchil-flex:
This can be merged right?

caseyflex:
Yes!

PR #1673, opened 2024-05-03
handle edge case when `estimate_cost()` returns None for jobs in `Batch`

tylerflex:
not sure a changelog item is needed because this never came up before I refactored in #1630 

Comes up now when some of the tasks error.

### DIFF
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -884,7 +884,11 @@ class Batch(WebContainer):
         float
             Estimated total cost of the tasks in FlexCredits.
         """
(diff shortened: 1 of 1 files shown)


lucas-flexcompute (approved):
Should this issue a warning?

tylerflex:
> Should this issue a warning?

Setting `batch_cost` to `None` will trigger a log explaining that the batch could not be calculated.
https://github.com/flexcompute/tidy3d/blob/7084793cd6f918dcab5b815ee6f13c4eb7cf0f03/tidy3d/web/api/container.py#L898

Think this is sufficient?

PR #1674, opened 2024-05-03
bugfix: add task_id to JaxSimulationData

tylerflex:
`task_id` was not being added to `JaxSimulationData`. Fixed here

See #1670

### DIFF
--- a/tidy3d/plugins/adjoint/web.py
+++ b/tidy3d/plugins/adjoint/web.py
@@ -123,6 +123,7 @@ def run(
         callback_url=callback_url,
         verbose=verbose,
     )
+    # TODO: add task_id
     return JaxSimulationData.from_sim_data(sim_data, jax_info)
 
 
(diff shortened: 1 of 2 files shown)


tidy3d/plugins/adjoint/web.py
```diff
@@ -123,6 +123,7 @@ def run(
         callback_url=callback_url,
         verbose=verbose,
     )
+    # TODO: add task_id
```
yaugenst-flex, line 126: I guess these TODO's are here because the PR only explicitly adds to `JaxSimulationData`? I think I'm missing some context here...

tylerflex: the PR only adds to the JaxSimulationData when we have access to the `task_id` because we can't grab the task_id from a completed `web.run()` call, I have no way of getting it in these places unless we figure out some way around it. I think for the issue brought up by the user, it should be sufficient

yaugenst-flex (approved):

PR #1676, opened 2024-05-03
validate single pixel mode source

### LINKED ISSUE #1667: Validate out single-pixel mode source or monitor in a non-0D sim
When a simulation is not 0D along a given dimension, but a mode source or mode monitor only has a single pixel along that dimension, something is wrong with the setup - mode objects generally assume that the modes should decay by the mode plane boundaries, which cannot be the case with a single pixel. Most likely, the size of the mode plane was too small along that dimension (but not 0, because we do validate that only one dimension of the mode object size is 0). We should error in such cases. This probably needs to be checked in  `validate_pre_upload`, because it's only then that we have access to grid discretization so we can compute the number of pixels in the object.

dmarek-flex:
Validate that mode sources have at least 3 cells (even with size 0 it is possible to pick up 2 cells, so I chose 2 as the limit).

Also included are some changes to make ruff happy and fix some other tests that were affected by this new validation step.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -35,7 +35,7 @@ from .structure import Structure, MeshOverrideStructure
 from .source import SourceType, PlaneWave, GaussianBeam, AstigmaticGaussianBeam, CustomFieldSource
 from .source import CustomCurrentSource, CustomSourceTime, ContinuousWave
 from .source import TFSF, Source, ModeSource
-from .monitor import MonitorType, Monitor, FreqMonitor, SurfaceIntegrationMonitor
+from .monitor import ModeMonitor, MonitorType, Monitor, FreqMonitor, SurfaceIntegrationMonitor
 from .monitor import AbstractModeMonitor, FieldMonitor, TimeMonitor, FieldTimeMonitor
 from .monitor import PermittivityMonitor, DiffractionMonitor, AbstractFieldProjectionMonitor
 from .monitor import FieldProjectionAngleMonitor, FieldProjectionKSpaceMonitor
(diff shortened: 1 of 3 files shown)


tidy3d/components/simulation.py
```diff
@@ -3054,6 +3055,24 @@ def warn_mode_size(monitor: AbstractModeMonitor, msg_header: str, custom_loc: Li
                     custom_loc = ["monitors", mnt_ind]
                     warn_mode_size(monitor=monitor, msg_header=msg_header, custom_loc=custom_loc)
 
+    def _validate_num_cells_in_mode_source(self) -> None:
+        """Warn if mode sources intersect with a very small number of grid cells."""
+        for source in self.sources:
+            if isinstance(source, ModeSource):
+                disc_grid = self.discretize(source)
+                check_dims = [0, 1, 2]
+                check_dims.pop(source.injection_axis)
```
tylerflex: we have a utility method `Geometry.pop_axis` that can do this for us. I think it's safer to use in case we ever want to change conventions (although that's unlikely now)

```py
_, check_dims = ModeSource.pop_axis([0,1,2], axis=source.injection_axis)
```

Also probably would make sense to switch the variable names a bit, we tend to use the convention of `axis` being an `int` in `[0,1,2]` and `dims` being a `str` in `xyz`

tylerflex (approved):
Looks good, just a pretty minor comment in the validator. thanks @dmarek-flex for taking care of this so quickly!

momchil-flex (approved):
Looks good, only it should apply to mode monitors too.

PR #1679, opened 2024-05-03
Relax smallest triangle area in TriangleMesh to 1e-36

weiliangjin2021:
It's relaxed to the minimal positive normal 32-float number

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `tidy3d convert` from `.lsf` files to tidy3d scripts is moved to another repository at `https://github.com/hirako22/Lumerical-to-Tidy3D-Converter`.
+- Relax validation of smallest triangles area in `TriangleMesh`.
 
 ### Fixed
 - Bug in plotting and computing tilted plane intersections of transformed 0 thickness geometries.
--- a/tidy3d/components/geometry/mesh.py
+++ b/tidy3d/components/geometry/mesh.py
@@ -12,13 +12,15 @@ from ..types import Ax, Bound, Coordinate, MatrixReal4x4, Shapely
 from ..viz import add_ax_if_none, equal_aspect
 from ...log import log
 from ...exceptions import ValidationError, DataError
-from ...constants import fp_eps, inf
+from ...constants import inf
 from ..data.dataset import TriangleMeshDataset
 from ..data.data_array import TriangleMeshDataArray, DATA_ARRAY_MAP
 from ..data.validators import validate_no_nans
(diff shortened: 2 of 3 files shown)


weiliangjin2021:
One question @caseyflex : do we really need to worry about zero area at all? If the area is too small, it can be just left out.

weiliangjin2021:
> One question @caseyflex : do we really need to worry about zero area at all? If the area is too small, it can be just left out.

I probably find the answer myself: if the entire surface is very small so that almost every triangle is ignored, it'll be severely off.

caseyflex:
> > One question @caseyflex : do we really need to worry about zero area at all? If the area is too small, it can be just left out.
> 
> I probably find the answer myself: if the entire surface is very small so that almost every triangle is ignored, it'll be severely off.

I see, I had wondered about this earlier too, but yeah that case could definitely occur.

tylerflex (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `tidy3d convert` from `.lsf` files to tidy3d scripts is moved to another repository at `https://github.com/hirako22/Lumerical-to-Tidy3D-Converter`.
+- Relax validation of smallest triangles area in `TriangleMesh` from 1e-7 to 1e-36.
```
tylerflex: minor comment: consider leaving out the specific numbers or putting in single backslash quotes. (also I guess units of um^2?)

tidy3d/components/geometry/mesh.py
```diff
@@ -60,7 +62,7 @@ def _check_mesh(cls, val: TriangleMeshDataset) -> TriangleMeshDataset:
         if val is None:
             return None
         mesh = cls._triangles_to_trimesh(val.surface_mesh)
-        if not all(np.array(mesh.area_faces) > fp_eps):
+        if not all(np.array(mesh.area_faces) > AREA_SIZE_THRESHOLD):
             raise ValidationError(
                 "The provided mesh has triangles with zero area. "
```
tylerflex: should we modify to "near zero area" and include the threshold we're using in the message?

weiliangjin2021: Good point. I also wonder if we should only error if there are triangles of area `(0, AREA_SIZE_THRESHOLD]`. If the area is exactly 0 in double precision, it can be safely dropped. @caseyflex

caseyflex: I haven't seen this come up. On the one hand, it seems like it can be safely dropped. On the other hand, it seems like an actual problem with the mesh in this case, so maybe we want to keep the error for now (dropping the triangle on backend, but keeping validation error on frontend for 0 area triangles)

Issue #1668, opened 2024-05-02
Validate that custom datasets can interpolate


momchil-flex:
If an array with repeated coords is provided, `interp` produces an error. There may also be other cases in which `interp` doesn't work. I think we should make sure that interpolation works both for CustomFieldSource and for CustomMedium datasets as it is assumed to be working in the solver. There may be others too. So one thing that comes to mind is to have a validator that tests an interpolation to some arbitrary coordinates and raises an error if that fails for some reason, like say `f"Interpolation of custom data provided in {custom_data_object} fails with {str(e)}"`.

PR #1681, opened 2024-05-06
Handle all remaining lint errors in tests

yaugenst-flex:
Tests were still throwing a lot of lint errors, which made writing new ones a bit of a pain due to our pre-commit config.

Before (181 errors total):

image

After:

image

PR #1684, opened 2024-05-07
Validate that custom datasets can interpolate

marc-flex:
This PR addresses issue #1668

### DIFF
--- a/tests/test_components/test_medium.py
+++ b/tests/test_components/test_medium.py
@@ -833,3 +833,39 @@ def test_coaxial_lumped_resistor():
             normal_axis=1,
             name="R",
         )
+
+
+def test_custom_medium(log_capture):  # noqa: F811
+    Nx, Ny, Nz, Nf = 4, 3, 1, 1
+    X = np.linspace(-1, 1, Nx)
+    Y = np.linspace(-1, 1, Ny)
+    Z = [0]
+    freqs = [2e14]
+    n_data = np.ones((Nx, Ny, Nz, Nf))
+    n_dataset = td.ScalarFieldDataArray(n_data, coords=dict(x=X, y=Y, z=Z, f=freqs))
+
+    def create_mediums(n_dataset):
+        ## Three equivalent ways of defining custom medium for the lens
+
+        # define custom medium with n/k data
+        _ = td.CustomMedium.from_nk(n_dataset, interp_method="nearest")
+
+        # define custom medium with permittivity data
+        eps_dataset = td.ScalarFieldDataArray(n_dataset**2, coords=dict(x=X, y=Y, z=Z, f=freqs))
+        _ = td.CustomMedium.from_eps_raw(eps_dataset, interp_method="nearest")
+
+        # define each component of permittivity via "PermittivityDataset"
+        eps_xyz_dataset = td.PermittivityDataset(
+            eps_xx=eps_dataset, eps_yy=eps_dataset, eps_zz=eps_dataset
+        )
+        _ = td.CustomMedium(eps_dataset=eps_xyz_dataset, interp_method="nearest")
+
+    create_mediums(n_dataset=n_dataset)
+    assert_log_level(log_capture, None)
+
+    with pytest.raises(pydantic.ValidationError):
+        # repeat some entries so data cannot be interpolated
+        X2 = [X[0]] + list(X)
+        n_data2 = np.vstack((n_data[0, :, :, :].reshape(1, Ny, Nz, Nf), n_data))
+        n_dataset2 = td.ScalarFieldDataArray(n_data2, coords=dict(x=X2, y=Y, z=Z, f=freqs))
+        create_mediums(n_dataset=n_dataset2)
--- a/tidy3d/components/data/data_array.py
+++ b/tidy3d/components/data/data_array.py
@@ -5,6 +5,7 @@ from abc import ABC
 
 import xarray as xr
 import numpy as np
+import pandas
 import dask
 import h5py
 
@@ -92,6 +93,48 @@ class DataArray(xr.DataArray):
         for attr_name, attr in cls._data_attrs.items():
             val.attrs[attr_name] = attr
         return val
+
+    def _interp_validator(self, field_name: str = None) -> None:
+        """Make sure we can interp()/sel() the data."""
+        # NOTE: this does not check every 'DataArray' by default. Instead, when required, this check can be
+        # called from a validator, as is the case with 'CustomMedium' and 'CustomFieldSource'.
+
+        if field_name is None:
+            field_name = "DataArray"
+
+        dims = self.coords.dims
+
+        for dim in dims:
+            # in case we encounter some /0 or /NaN we'll ignore the warnings here
+            with np.errstate(divide="ignore", invalid="ignore"):
+                # check that we can interpolate
+                try:
+                    x0 = np.array(self.coords[dim][0])
+                    self.interp({dim: x0}, method="linear")
+                    self.interp({dim: x0}, method="nearest")
+                    # self.interp_like(self.isel({self.dim: 0}))
+                except pandas.errors.InvalidIndexError as e:
+                    raise DataError(
+                        f"'{field_name}.interp()' fails to interpolate along {dim} which is used by the solver. "
+                        "This may be caused, for instance, by duplicated data "
+                        f"in this dimension (you can verify this by running "
+                        f"'{field_name}={field_name}.drop_duplicates(dim=\"{dim}\")' "
+                        f"and interpolate with the new '{field_name}'). "
+                        "Plase make sure data can be interpolated."
+                    ) from e
+                # in case it can interpolate, try also to sel
+                try:
+                    x0 = np.array(self.coords[dim][0])
+                    self.sel({dim: x0}, method="nearest")
+                except pandas.errors.InvalidIndexError as e:
+                    raise DataError(
+                        f"'{field_name}.sel()' fails to select along {dim} which is used by the solver. "
+                        "This may be caused, for instance, by duplicated data "
+                        f"in this dimension (you can verify this by running "
+                        f"'{field_name}={field_name}.drop_duplicates(dim=\"{dim}\")' "
+                        f"and run 'sel()' with the new '{field_name}'). "
+                        "Plase make sure 'sel()' can be used on the 'DataArray'."
+                    ) from e
 
     @classmethod
     def assign_coord_attrs(cls, val):
--- a/tidy3d/components/source.py
+++ b/tidy3d/components/source.py
@@ -16,7 +16,7 @@ from .types import Coordinate, Direction, Polarization, Ax, FreqBound
 from .types import ArrayFloat1D, Axis, PlotVal, ArrayComplex1D, TYPE_TAG_STR
 from .validators import assert_plane, assert_volumetric
 from .validators import warn_if_dataset_none, assert_single_freq_in_range, _assert_min_freq
-from .data.dataset import FieldDataset, TimeDataset
+from .data.dataset import FieldDataset, TimeDataset, ScalarFieldDataArray
 from .data.validators import validate_no_nans
 from .data.data_array import TimeDataArray
 from .geometry.base import Box
@@ -791,6 +791,15 @@ class CustomFieldSource(FieldSource, PlanarSource):
                 if tangential_field in val.field_components:
                     return val
         raise SetupError("No tangential field found in the suppled 'field_dataset'.")
+
+    @pydantic.validator("field_dataset", always=True)
+    def _check_fields_interpolate(cls, val: FieldDataset) -> FieldDataset:
+        """Checks whether the filds in 'field_dataset' can be interpolated."""
+        if isinstance(val, FieldDataset):
+            for name, data in val.field_components.items():
+                if isinstance(data, ScalarFieldDataArray):
+                    data._interp_validator(name)
+        return val
 
 
 """ Source current profiles defined by (1) angle or (2) desired mode. Sets theta and phi angles."""
--- a/tests/test_components/test_source.py
+++ b/tests/test_components/test_source.py
@@ -323,3 +323,31 @@ def test_custom_source_time(log_capture):  # noqa: F811
         dataset = td.components.data.dataset.TimeDataset(values=vals)
         cst = td.CustomSourceTime(source_time_dataset=dataset, freq0=freq0, fwidth=0.1e12)
         assert np.allclose(cst.amp_time([0]), [1], rtol=0, atol=ATOL)
+
+
+def test_custom_field_source(log_capture):
+    Nx, Ny, Nz, Nf = 4, 3, 1, 1
+    X = np.linspace(-1, 1, Nx)
+    Y = np.linspace(-1, 1, Ny)
+    Z = [0]
+    freqs = [2e14]
+    n_data = np.ones((Nx, Ny, Nz, Nf))
+    n_dataset = td.ScalarFieldDataArray(n_data, coords=dict(x=X, y=Y, z=Z, f=freqs))
+
+    def make_custom_field_source(field_ds):
+        custom_source = td.CustomFieldSource(
+            center=(1, 1, 1), size=(2, 2, 0), source_time=ST, field_dataset=field_ds
+        )
+        return custom_source
+
+    field_dataset = td.FieldDataset(Ex=n_dataset, Hy=n_dataset)
+    make_custom_field_source(field_dataset)
+    assert_log_level(log_capture, None)
+
+    with pytest.raises(pydantic.ValidationError):
+        # repeat some entries so data cannot be interpolated
+        X2 = [X[0]] + list(X)
+        n_data2 = np.vstack((n_data[0, :, :, :].reshape(1, Ny, Nz, Nf), n_data))
+        n_dataset2 = td.ScalarFieldDataArray(n_data2, coords=dict(x=X2, y=Y, z=Z, f=freqs))
+        field_dataset = td.FieldDataset(Ex=n_dataset, Hy=n_dataset2)
+        make_custom_field_source(field_dataset)
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -1896,6 +1896,15 @@ class CustomMedium(AbstractCustomMedium):
             )
         return val
 
+    @pd.validator("permittivity", "conductivity", always=True)
+    def _check_permittivity_conductivity_interpolate(cls, val, values, field):
+        """Check that the custom medium 'SpatialDataArrays' can be interpolated."""
+
+        if isinstance(val, SpatialDataArray):
+            val._interp_validator(field.name)
+
+        return val
+
     @cached_property
     def is_isotropic(self) -> bool:
         """Check if the medium is isotropic or anisotropic."""


marc-flex:
All I've done so far is to introduce a validator that checks that data can be interpolated in each of the dimensions.

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,23 @@ def assign_data_attrs(cls, val):
...
 
+    @classmethod
+    def interp_validator(cls, val):
+        """Make sure we can interpolate the data."""
+
+        dims = val.coords.dims
+        for dim in dims:
+            try:
+                val.interp({dim: 0})
+            except:
```
tylerflex: you'll want to `except` a specific `Exception` here (whatever `xarray` raises when `interp()` fails).  One convenient way to do this while still retaining the original exception traceback (which can be useful for debugging) is the following pattern

```py
try:
    ...
except ExceptionType as e:
    raise MyException(...) from e
```

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,23 @@ def assign_data_attrs(cls, val):
             val.attrs[attr_name] = attr
         return val
 
+    @classmethod
+    def interp_validator(cls, val):
```
tylerflex: I suppose this checks all `DataArray` objects, which maybe isn't so bad? but I guess this is mainly an issue for `ModeSource` and `CustomFieldSource` objects.  I can't think of an instance where we want to allow a DataArray with repeating coords but it would be interesting to see if the tests pass.

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,23 @@ def assign_data_attrs(cls, val):
...
+        """Make sure we can interpolate the data."""
+
+        dims = val.coords.dims
+        for dim in dims:
+            try:
+                val.interp({dim: 0})
```
tylerflex: I think this might fail if the data is not numeric. For example, the `ModeSource.direction` can be `("-", "+")`. It might be more robust to try an approach where we sel the first coordinate and try to interp into it.

```py
val.interp_like(val.isel({val.dim: 0}))
```

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,23 @@ def assign_data_attrs(cls, val):
...
+            except:
+                raise DataError(
+                    f"DataArray.interp() fails to interpolate to {dim}=0. This may be caused by duplicated data "
```
tylerflex: please put any code stuff in single (regular) quotes `'`. It will render nicer in the terminal output, thanks!

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,23 @@ def assign_data_attrs(cls, val):
...
+                    f'in this dimension (you can verify this with DataArray.drop_duplicates(dim="{dim}")). '
```
tylerflex: how specifically can they verify with this command?

tylerflex (changes requested):
Thanks @marc-flex I think this is looking pretty good, just a few minor tweaks and also we should see if we want to apply this validator to all `DataArray` objects (as written currently) or rather as a `Simulation` post-init validator that loops over sources and does the interp check. I'm curious to see if the front end tests pass with these changes or if there are instances where we explicitly allow duplicate coords in some data array objects.

tylerflex:
Also, could you please add a few tests to `tests/test_data/test_data_array.py`? Basically try to replicate a few scenarios and edge cases for this behavior and ensure that the proper thing happens. Thanks

marc-flex:
> Thanks @marc-flex I think this is looking pretty good, just a few minor tweaks and also we should see if we want to apply this validator to all `DataArray` objects (as written currently) or rather as a `Simulation` post-init validator that loops over sources and does the interp check. I'm curious to see if the front end tests pass with these changes or if there are instances where we explicitly allow duplicate coords in some data array objects.

You're right. This doesn't pass front-end test. I'll give it another go with all your comments and restricting the check to custom fields and sources.

tylerflex:
It might just be failing because of `direction` coords. but the fix I suggested about `isel`-ing the first coordinate might fix them, so perhaps try that first,

tylerflex:
Or it could just be that some of the front end test data arrays have extra coordinates (by accident)

momchil-flex:
> > Thanks @marc-flex I think this is looking pretty good, just a few minor tweaks and also we should see if we want to apply this validator to all `DataArray` objects (as written currently) or rather as a `Simulation` post-init validator that loops over sources and does the interp check. I'm curious to see if the front end tests pass with these changes or if there are instances where we explicitly allow duplicate coords in some data array objects.
> 
> You're right. This doesn't pass front-end test. I'll give it another go with all your comments and restricting the check to custom fields and sources.

I guess also CustomMedium types with `interp_method == "linear"` (as opposed to the default `"nearest"`). Actually I don't know if `"nearest"` also fails in some cases, i.e. maybe the data array `sel` method also fails if there are repeated coordinates?

marc-flex:
@tylerflex I have tried several things. With the solution you mentioned (`val.interp_like(val.isel({val.dim: 0}))`) tests kept failing. So I have modified it to use an array of values from within the coordinates for each dimension (not sure that's generic enough).

Also, since this made checks fail, I'm only checking for custom source and mediums. Basically, I call a check function (defined at the `DataArray` level) within the validators. 

I have created tests for custom medium and custom field source.

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,43 @@ def assign_data_attrs(cls, val):
...
+                # in case it can interpolate, try also to sel
+                try:
+                    x0 = np.array(self.coords[dim][0])
+                    self.sel({dim: x0}, method="nearest")
+                except pandas.errors.InvalidIndexError:
+                    raise DataError(
+                        f"'DataArray.sel()' fails to select along {dim} which is used in the back-end. "
+                        "This may be caused, for instance, by duplicated data "
+                        f"in this dimension (you can verify this by running 'DataArray.drop_duplicates(dim=\"{dim}\")' "
+                        "and run sel() with the new 'DataArray')."
```
tylerflex: please put `sel` in single quotes as well, thanks

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,43 @@ def assign_data_attrs(cls, val):
...
+                        f"in this dimension (you can verify this by running 'DataArray.drop_duplicates(dim=\"{dim}\")' "
+                        "and interpolate with the new DataArray)."
+                        "Plase make sure you can interpolate the data."
+                    )
+                # in case it can interpolate, try also to sel
+                try:
+                    x0 = np.array(self.coords[dim][0])
+                    self.sel({dim: x0}, method="nearest")
+                except pandas.errors.InvalidIndexError:
+                    raise DataError(
```
tylerflex, line 130: please ensure there are spaces between the periods at end of sentences and then the next sentence (if you don't add, manually, eg 

```
"end of sentence. "
"Blah"
````
they will get combined without the space

File: tidy3d/components/data/data_array.py [hunk 8]
```diff
@@ -93,6 +94,43 @@ def assign_data_attrs(cls, val):
             val.attrs[attr_name] = attr
         return val
 
+    def _interp_validator(self):
```
tylerflex: let's add a return type annotation `def _interp_validator(self) -> None:

tidy3d/components/data/data_array.py
tylerflex: and double check the other `raise` messages eg above

tidy3d/components/source.py
```diff
@@ -792,6 +792,16 @@ def _tangential_component_defined(cls, val: FieldDataset, values: dict) -> Field
                     return val
         raise SetupError("No tangential field found in the suppled 'field_dataset'.")
 
+    @pydantic.validator("field_dataset", always=True)
+    def _check_fields_interpolate(cls, val: FieldDataset) -> FieldDataset:
+        """Checks whether the filds in 'field_dataset' can be interpolated."""
+        if isinstance(val, FieldDataset):
+            fields = ["Ex", "Ey", "Ez", "Hx", "Hy", "Hz"]
```
tylerflex: the `FieldDataset` should have a `@property` called `field_components : dict`, so you can simply iterate through that more easily

```py
for field, data in val.field_components.items():
    if isinstance(data, SpatialDataArray):
        # do the check on data
```

tests/test_components/test_medium.py
```diff
@@ -5,7 +5,7 @@
 import matplotlib.pyplot as plt
 import tidy3d as td
 from tidy3d.exceptions import ValidationError, SetupError
-from ..utils import assert_log_level, log_capture, AssertLogLevel
```
tylerflex: I think the code formatter (ruff) is removing `log_captiure` since it's not used in the test directly.

However, we do add it as a `pytest.fixture`, passing it into eg. `def test_custom_medium(log_capture):` so this is a mistake from ruff that might be causing test failures.

You can resolve it by adding `log_capture` back into the imports and then ignoring this linting rule with a comment, for example:
https://github.com/flexcompute/tidy3d/blob/3ee34904eb6687a86a5fb3f4ed6d3295c228cd83/tests/test_data/test_monitor_data.py#L27

```
from ..utils import log_capture  # noqa: F401 
```

tylerflex:
Thanks @marc-flex just FYI, this PR #1681 changed some of the tests in pre/2.7 so it looks like you'll need to rebase against that branch and fix any conflicts that might come up.

I'm not sure which tests are failing but we can look into it more. One that probably fails is related to `log_capture` but you can see my comment above.

marc-flex:
> Thanks @marc-flex just FYI, this PR #1681 changed some of the tests in pre/2.7 so it looks like you'll need to rebase against that branch and fix any conflicts that might come up.

I have just rebased against pre/2.7 and force-pushed it

tylerflex:
@marc-flex @momchil-flex is this something we want to get into 2.7.0rc2? and is it ready or still needing changes?

tylerflex (commented):
Sorry @marc-flex I realized that this review was left as "pending" for the past week. here are some additional comments. 

When you think it's getting ready, feel free to un-mark as a draft.

tidy3d/components/data/data_array.py
```diff
@@ -115,7 +115,7 @@ def _interp_validator(self):
                         f"'DataArray.interp()' fails to interpolate along {dim} which is used in the back-end. "
                         "This may be caused, for instance, by duplicated data "
                         f"in this dimension (you can verify this by running 'DataArray.drop_duplicates(dim=\"{dim}\")' "
-                        "and interpolate with the new DataArray)."
+                        "and interpolate with the new DataArray). "
```
tylerflex: data array in single quotes here

tidy3d/components/source.py
```diff
@@ -796,10 +796,9 @@ def _tangential_component_defined(cls, val: FieldDataset, values: dict) -> Field
     def _check_fields_interpolate(cls, val: FieldDataset) -> FieldDataset:
         """Checks whether the filds in 'field_dataset' can be interpolated."""
         if isinstance(val, FieldDataset):
-            fields = ["Ex", "Ey", "Ez", "Hx", "Hy", "Hz"]
-            for field in fields:
-                if isinstance(val.__getattribute__(field), ScalarFieldDataArray):
-                    val.__getattribute__(field)._interp_validator()
+            for _, data in val.field_components.items():
+                if isinstance(data, ScalarFieldDataArray):
```
tylerflex, line 800: Per original issue, I think we want to perform this for `CustomMedium` as well.

Will let @momchil-flex confirm what he wants when he's back online.

Otherwise looks good! Thanks @marc-flex

marc-flex: Isn't [this](https://github.com/flexcompute/tidy3d/blob/b57594ffa802cd879cb7cd923701331ba3b59310/tidy3d/components/medium.py#L1900) doing it for `CustomMedium`?

tylerflex: huh, for some reason I didnt see that at all. Yea that should do it!

tylerflex: only thing I would consider adding to make it just a bit more user friendly would be an optional keyword argument to `interp_validator(self, field_name: str=None)`

and then we can pass `permittivity` or `Ex`, etc and the error message can include it.

This might help, eg. users identify which field is causing issues.

If this overly complicates things, feel free to ignore.

marc-flex: > If this overly complicates things, feel free to ignore.

Not at all! I think it'll be much nicer

weiliangjin2021 (approved):
Just one comment: custom datasets are validated in `CustomMedium`. Shall we validate all other types of custom materials as well, e.g. CustomPoleResidue, CustomLorentz, etc.?

tidy3d/components/source.py
marc-flex: @tylerflex I have added field name. Let me know what you think. Thanks!

I was also looking that `eps_dataset` within `CustomMedium` has fields that derive from `DataArray`. Should we include this into the validator for `CustomMedium`?  I guess this is along what @weiliangjin2021 is commenting below.

marc-flex:
> Just one comment: custom datasets are validated in `CustomMedium`. Shall we validate all other types of custom materials as well, e.g. CustomPoleResidue, CustomLorentz, etc.?

You're opening Pandora's box here. Maybe we should do this for all custom-defined `DataArrays` fields? Or should we limit this PR to the fields defined in issue #1668 ? What do you think @momchil-flex @tylerflex ? I'm guessing this wouldn't just be the validator, it'd also be implementing a corresponding test

momchil-flex:
> > Just one comment: custom datasets are validated in `CustomMedium`. Shall we validate all other types of custom materials as well, e.g. CustomPoleResidue, CustomLorentz, etc.?
> 
> You're opening Pandora's box here. Maybe we should do this for all custom-defined `DataArrays` fields? Or should we limit this PR to the fields defined in issue #1668 ? What do you think @momchil-flex @tylerflex ? I'm guessing this wouldn't just be the validator, it'd also be implementing a corresponding test

To me it kinda makes sense to test all custom-defined data array fields, and I can't think of currently existing objects where we wouldn't want that. However, a) I don't know if I'm forgetting something already and b) I'm not sure if in the future something will come up. But maybe let's leave b) for future devs to worry about if validating all custom data arrays passes all tests (so hopefully a) is ok)?

marc-flex:
> > > Just one comment: custom datasets are validated in `CustomMedium`. Shall we validate all other types of custom materials as well, e.g. CustomPoleResidue, CustomLorentz, etc.?
> > 
> > 
> > You're opening Pandora's box here. Maybe we should do this for all custom-defined `DataArrays` fields? Or should we limit this PR to the fields defined in issue #1668 ? What do you think @momchil-flex @tylerflex ? I'm guessing this wouldn't just be the validator, it'd also be implementing a corresponding test
> 
> To me it kinda makes sense to test all custom-defined data array fields, and I can't think of currently existing objects where we wouldn't want that. However, a) I don't know if I'm forgetting something already and b) I'm not sure if in the future something will come up. But maybe let's leave b) for future devs to worry about if validating all custom data arrays passes all tests (so hopefully a) is ok)?

I can remove the specific tests for custom medium/source and have the validator at the `DataArray` level. The only issue is that it wouldn't then say what property made the validator fail which might be frustrating for the users. So I guess I'd suggest to keep the scope of this PR as is, and add another issue where we collect any other fields we'd like to validate/test?

tylerflex:
I agree with @marc-flex . Maybe it's best to just limit the scope of these validators to the components that are causing issues as per #1684 . because I additionally worry that adding too strict validation DataArray may lead to some unintended consequences.

momchil-flex:
Ok yeah sounds good to me.

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,48 @@ def assign_data_attrs(cls, val):
...
+                try:
+                    x0 = np.array(self.coords[dim][0])
+                    self.sel({dim: x0}, method="nearest")
+                except pandas.errors.InvalidIndexError:
+                    raise DataError(
+                        f"'{field_name}.sel()' fails to select along {dim} which is used in the back-end. "
+                        "This may be caused, for instance, by duplicated data "
+                        f"in this dimension (you can verify this by running "
+                        f"'{field_name}={field_name}.drop_duplicates(dim=\"{dim}\")' "
+                        "and run 'sel()' with the new '{field_name}'). "
```
tylerflex: add `f` for the {field_name}

tidy3d/components/source.py
tylerflex: Looks good! minus a missing `f` in the fstring ;). Thanks!

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,48 @@ def assign_data_attrs(cls, val):
...
+                        "Plase make sure you can use 'sel()' on the data."
+                    )
```
tylerflex: for the `pandas.errors.InvalidIndexError` handling, consider doing the following

```py
except pandas.errors.InvalidIndexError as e:
   raise DataError(...) from e
```

What this will do is raise your DataError and also display the original pandas Index error stack trace. It could be helpful to see what exactly is causing it, and for users to report us the stack trace so we can debug.

I might recommend first verifying in the test output that this doesn't look absolutely terrible though :D

tylerflex (changes requested):

tidy3d/components/data/data_array.py
tylerflex: if you want to read all about it.. https://peps.python.org/pep-3134/#:~:text=Explicit%20Exception%20Chaining,-The%20__cause&text=If%20the%20call%20to%20open,IOError%20as%20the%20original%20cause.

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,48 @@ def assign_data_attrs(cls, val):
             val.attrs[attr_name] = attr
         return val
 
+    def _interp_validator(self, field_name: str = None) -> None:
+        """Make sure we can interp()/sel() the data."""
+        # NOTE: this does not check every 'DataArray' bu default. Instead, when required, this check can be
```
momchil-flex: bu -> by

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,48 @@ def assign_data_attrs(cls, val):
...
+            with np.errstate(divide="ignore", invalid="ignore"):
+                # check that we can interpolate
+                try:
+                    x0 = np.array(self.coords[dim][0])
+                    self.interp({dim: x0}, method="linear")
+                    self.interp({dim: x0}, method="nearest")
+                    # self.interp_like(self.isel({self.dim: 0}))
+                except pandas.errors.InvalidIndexError:
+                    raise DataError(
+                        f"'{field_name}.interp()' fails to interpolate along {dim} which is used in the back-end. "
```
momchil-flex: very nit but I think we don't usually talk about backend but just "solver"

tidy3d/components/data/data_array.py
```diff
@@ -93,6 +94,48 @@ def assign_data_attrs(cls, val):
...
+                        "This may be caused, for instance, by duplicated data "
+                        f"in this dimension (you can verify this by running "
+                        f"'{field_name}={field_name}.drop_duplicates(dim=\"{dim}\")' "
+                        f"and interpolate with the new '{field_name}'). "
+                        "Plase make sure you can interpolate the data."
```
momchil-flex: I had a comment on a Tyler PR recently wondering if we want to use "you" directions in warnings/errors. I feel like it's not something I usually see in python packages. E.g. rephrasing to "Make sure data can be interpolated". But I am not sure if we should be enforcing this in the future or not.

momchil-flex (approved):

tidy3d/components/data/data_array.py
marc-flex: I just tried and the error doesn't seem to change. Not sure if I should be seeing any change, though

tidy3d/components/data/data_array.py
marc-flex: I changed it to what you suggested. This is something to keep in mind and, actually, something that maybe we should enforce even if only to keep consistent?

tylerflex:
@marc-flex I think this is basically ready to go. Last thing you'll need to do is update the docs/notebooks submodule to the latest version (pre/2.7 branch).

After you make that commit, we'll want to rebase this against tidy3d `pre/2.7` and squash all the commits into one.

marc-flex:
@tylerflex @momchil-flex  I have rebased and I thought everything would be OK from my side but it fails the "latest" test. Do let me know how to fix this.

momchil-flex:
Actually the latest test is fine now, it's the formatting test that fails. :) You need to run `bash scripts/test_local.sh` (and you can abort after the formatting step), or 

```
black tidy3d/
black tests/
```

tylerflex:
(make sure your black is the same as the version defined in pyproject.toml, or it will format most of the files in tidy3d)

marc-flex:
Not sure why poetry didn't complain about that when I committed the squash+rebase. I think it's all good now

PR #1686, opened 2024-05-07
Adding num_proc arg to adjoint run_local, and properly avoiding validators

momchil-flex:


tests/test_plugins/test_adjoint.py
```diff
@@ -1563,27 +1563,41 @@ def make_custom_medium(num_cells: int) -> JaxCustomMedium:
...
     sim_data = run_local(sim, task_name="test", path=str(tmp_path / RUN_FILE))
 
+    # no error when calling it inside a gradient computation
+    def f(permittivity):
+        sim = make_sim_(permittivity=permittivity, num_input_structures=test_max_num_structs + 1)
+        sim_data = run_local(sim, task_name="test", path=str(tmp_path / RUN_FILE))
+        return objective(extract_amp(sim_data))
+
+    _ = grad(f)(EPS)
+
```
tylerflex, line 1610: let's throw in a test asserting that AdjointError is raised if `grad ` is called on a function where `run` is used? for good measure? unless this seems unnecessary

tidy3d/plugins/adjoint/web.py
```diff
@@ -583,14 +605,15 @@ def webapi_run_async_adjoint_bwd(
...
-@partial(custom_vjp, nondiff_argnums=tuple(range(1, 6)))
+@partial(custom_vjp, nondiff_argnums=tuple(range(1, 7)))
 def run_local(
     simulation: JaxSimulation,
     task_name: str,
     folder_name: str = "default",
     path: str = "simulation_data.hdf5",
     callback_url: str = None,
     verbose: bool = True,
+    num_proc: int = 1,
```
tylerflex: let's use the default defined in `simulation.py` `NUM_PROC_LOCAL` here

tylerflex (changes requested):
Couple minor things but overall looks good. Think it requires a changelog fixed item?

tylerflex (approved):
looks good, will leave it up to you regarding changelog. maybe since the skipped validators were introduced in 2.7 rc1 (I think?) we can just leave it out as it will be fixed in 2.7

tylerflex:
@prashkh It seems (based on the tests) that on python3.9, all the results are `None`. Would you mind looking into this? Im not really sure how to proceed. Perhaps there is some error that is getting caught by the `try/except` and setting the results to `None`? or potentially the mode solvers are never getting run?

tylerflex:
it's possible when I tried to recover your changes, I missed some detail?

prashkh:
@tylerflex This PR is for the test_adjoint right? or are you talking about run_batch addition for the web mode solver?

tylerflex:
Yea sorry, wrong PR.. supposed to be commenting on this one #1685 :D

PR #1688, opened 2024-05-08
TEST: fix medium test comparison

dmarek-flex:
Resolved some issues with test that used exact comparisons. The default tolerances in np.isclose are suitable.

### DIFF
--- a/tests/test_components/test_custom.py
+++ b/tests/test_components/test_custom.py
@@ -31,6 +31,8 @@ freqs = [2e14]
 ST = td.GaussianPulse(freq0=np.mean(freqs), fwidth=np.mean(freqs) / 5)
 SIZE = (2, 0, 2)
 
+RTOL = td.constants.fp_eps
+
 
 def make_scalar_data():
     """Makes a scalar field data array."""
@@ -312,7 +314,7 @@ def test_medium_raw():
     med = CustomMedium.from_eps_raw(eps_raw)
     for field in [eps_raw_s, eps_raw_u]:
         meds = CustomMedium.from_eps_raw(field)
-        assert med.eps_model(1e14) == meds.eps_model(1e14)
+        assert np.isclose(med.eps_model(1e14), meds.eps_model(1e14), rtol=RTOL)
 
     # lossy
     data = np.random.random((Nx, Ny, Nz, 1)) + 1 + 1e-2 * 1j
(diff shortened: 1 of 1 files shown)


weiliangjin2021:
Thanks for fixing another float number issue! There are a few more exact comparisons in this test file. Change them all in this PR?

tests/test_components/test_custom.py
```diff
@@ -465,7 +465,7 @@ def test_medium_nk(unstructured):
         meds = CustomMedium.from_nk(n=ns, k=-ks, freq=freqs[0])
     med = CustomMedium.from_nk(n=n, k=-k, allow_gain=True)
     meds = CustomMedium.from_nk(n=ns, k=-ks, freq=freqs[0], allow_gain=True)
-    assert med.eps_model(1e14) == meds.eps_model(1e14)
+    assert np.isclose(med.eps_model(1e14), meds.eps_model(1e14))
```
marc-flex: Shouldn't we set some tolerance? Are we using single or double for this?

marc-flex (approved):
Thanks @dmarek-flex  for fixing this! 
I guess another question would be why it started complaining about this now. Like, what changed?

tests/test_components/test_custom.py
marc-flex: Another thing to consider (taken from the documentation):
> The default value is appropriate for expected values of order unity: if the expected values are significantly smaller than one, it can result in false positives. atol should be carefully selected for the use case at hand. A zero value for atol will result in False if either a or b is zero.

yaugenst-flex: You could make the eps dynamic, i.e. something like
```python
a = med.eps_model(1e14)
b = meds.eps_model(1e14)
assert np.isclose(a, b, atol=np.spacing(a))
```
Guess that should work for all cases? Might need a factor of two in there, not sure.

dmarek-flex: I will use eps for doubles for the relative tolerance. I didn't think we needed to be very strict here, but why not!

The absolute tolerance is needed only if values are close to 0. The default used in np.isclose is 1e-8, which is sufficient for these values which are random numbers with magnitudes in the range of 1-2.

dmarek-flex:
> I guess another question would be why it started complaining about this now. Like, what changed?

Well part of the test relies on a random number generation, which maybe is why it happens from time to time. Although there seems to be a constant seed used. Over the day I have been getting both passes and fails. I don't know why that would not show up on the github tests though. And then on one of the versions of the test some interpolation is used, which could result in some introduced error I guess.

dmarek-flex:
I ended up using a relative tolerance which is equal to single precision epsilon. Other tests would fail if I tried to be very strict with double precision epsilon. In lieu of choosing a relative tolerance for each comparison, I just chose to relax the tolerance.

dmarek-flex:
@marc-flex It would be helpful if you could checkout this branch yourself and see if any test fails.

marc-flex:
> @marc-flex It would be helpful if you could checkout this branch yourself and see if any test fails.

Of course! :) I just checked and all test passed

weiliangjin2021 (approved):

PR #1689, opened 2024-05-08
Introduce SubpixelSpec to allow for selecting subpixel averaging methods on different materials

weiliangjin2021:
### New API
```
subpixel = td.SubpixelSpec(
    dielectric = True,
    metal = td.Staircasing()/td.VolumetricAveraging(staircase_normal_component=True),
    pec = td.Staircasing()/td.HeuristicPECStaircasing()/td.PECConformal(timestep_reduction=0.1),
)
```
Metal is defined as `Re[eps]<1` at the central frequency.

You can still set `subpixel=True/False`:
- `True` using the default `SubpixelSpec()`: dielectric subpixel averaging on dielectric interface, staircasing on metal interface, and conformal (previously called Benkler) scheme on PEC;
- `False`: staircasing on any material interface.

### TODO
- [x] Fix failed adjoint tests, test_IO, etc.
- [x] Update API in doc
- [x] Various places validating subpixel field

### Note
- The default subpixel method for PEC is now Benkler (renamed to MetalConformal). The default setting will decrease the time step size by 30%. But this deduction will only take effect if we find any material is PEC in `scene.mediums`.
- In Volumetric Averaging, it has a field `staircase_normal_component`. When `True`, apply staircasing if the field component is mostly normal to the interface, since volumetric averaging won't help there.

### Open question:
- For dielectric materials, it's always worse to apply volume averaging. So currently, dielectric still only take boolean values. I'm not sure if it's better to be consistent with other materials (pec, metal) for dielectric to also take Staircasing and detailed subpixel method name.


tylerflex (commented):
I think the API looks pretty good. Nothing stands out to me as an issue.

One thing is that we should choose the class names carefully. Specifically,
```
BenklerSubpixel
HeuristicSubpixel
```
are both not clear from the title what exactly they are and when to use them. I'm not sure about the specific details, eg maybe these are commonly used names in the field, but if there are any more intuitive names for these, we should consider using them? Or even making up our own.

weiliangjin2021:
> One thing is that we should choose the class names carefully. Specifically,
> 
> ```
> BenklerSubpixel
> HeuristicSubpixel
> ```
> 
> are both not clear from the title what exactly they are and when to use them. I'm not sure about the specific details, eg maybe these are commonly used names in the field, but if there are any more intuitive names for these, we should consider using them? Or even making up our own.

Good point. I have concerns about the names as well, but forgot to bring it up. E.g. Benkler is too specific. How about a general name `ConformalSubpixel`, since people usually refers to it as conformal mesh in RF?

Regarding Heuristic, how about HeuristicStaircasingSubpixel? @momchil-flex

momchil-flex:
Yeah, I agree with both of those suggestions. `Conformal` is commonly used. `Heuristic` is just something Victor had implemented a long time ago that we are keeping for backwards compatibility. :D It is not in any way commonly known in the literature.

weiliangjin2021:
One more question on names: is it necessary for each of them ending in`Subpixel`? `StaircasingSubpixel` and `ConformalSubpixel` sounds strange, because the two words in the former are contradictory; while the two in the latter mean the same.

momchil-flex:
> One more question on names: is it necessary for each of them ending in`Subpixel`? `StaircasingSubpixel` and `ConformalSubpixel` sounds strange, because the two words in the former are contradictory; while the two in the latter mean the same.

Yeah, I think dropping Subpixel makes sense.

> For dielectric materials, it's always worse to apply volume averaging. So currently, dielectric still only take boolean values. I'm not sure if it's better to be consistent with other materials (pec, metal) for dielectric to also take Staircasing and detailed subpixel method name.

I think it's probably better in the long run yeah. You don't have to allow volumetric to be an actual option if it requires more effort on the backend. What should we call the dielectric subpixel?

weiliangjin2021:
Still haven't figured out a good name for dielectric subpixel averaging, so it remains as boolean for now. Docstrings have been updated. Now the new API looks like this
```
subpixel = td.SubpixelSpec(
    dielectric = True,
    metal = td.Staircasing()/td.VolumetricAveraging(),
    pec = td.Staircasing()/td.HeuristicPECStaircasing()/td.MetalConformal(timestep_reduction=0.1),
)
```

@tylerflex how do I fix the failing test_IO.py that loads an incompatible `tests/sims/simulation_sample.h5` and `tests/sims/simulation_sample.json` from 2.7.0rc1 and rc2?

tylerflex:
hm, try running this script?
https://github.com/flexcompute/tidy3d/blob/pre/2.7/scripts/sample.py

weiliangjin2021:
> hm, try running this script? https://github.com/flexcompute/tidy3d/blob/pre/2.7/scripts/sample.py

Thanks, that fixes it.

weiliangjin2021:
Sorry another question @tylerflex: it seems that allowing a simulation field to be the union of a basic type (e.g. bool in `subpixel` and float in `run_time`) and a class is causing issues in test_log [here](https://github.com/flexcompute/tidy3d/blob/9dd89f6afdc9ca74d12074c28669b72a1ee954d0/tests/test_package/test_log.py#L219): the warning list is empty when `subpixel` or `runt_time` take an object although I can see the warning on screen. It's not empty if the two fields take bool/float.

tylerflex:
> Sorry another question @tylerflex: it seems that allowing a simulation field to be the union of a basic type (e.g. bool in subpixel and float in run_time) and a class is causing issues in test_log [here](https://github.com/flexcompute/tidy3d/blob/9dd89f6afdc9ca74d12074c28669b72a1ee954d0/tests/test_package/test_log.py#L219): the warning list is empty when subpixel or runt_time take an object although I can see the warning on screen. It's not empty if the two fields take bool/float.

Hm, not sure about this, sorry. Maybe @dbochkov-flexcompute has some idea?

dbochkov-flexcompute:
> Sorry another question @tylerflex: it seems that allowing a simulation field to be the union of a basic type (e.g. bool in `subpixel` and float in `run_time`) and a class is causing issues in test_log [here](https://github.com/flexcompute/tidy3d/blob/9dd89f6afdc9ca74d12074c28669b72a1ee954d0/tests/test_package/test_log.py#L219): the warning list is empty when `subpixel` or `runt_time` take an object although I can see the warning on screen. It's not empty if the two fields take bool/float.

It seems to be related to the fact that pydantic tries to initialize a field with all possible options, if no guidance for it is given. That is, adding `discriminator=TYPE_TAG_STR` into `metal` and `pec` fields of `SubpixelSpec` resolves the issue

weiliangjin2021:
> > Sorry another question @tylerflex: it seems that allowing a simulation field to be the union of a basic type (e.g. bool in `subpixel` and float in `run_time`) and a class is causing issues in test_log [here](https://github.com/flexcompute/tidy3d/blob/9dd89f6afdc9ca74d12074c28669b72a1ee954d0/tests/test_package/test_log.py#L219): the warning list is empty when `subpixel` or `runt_time` take an object although I can see the warning on screen. It's not empty if the two fields take bool/float.
> 
> It seems to be related to the fact that pydantic tries to initialize a field with all possible options, if no guidance for it is given. That is, adding `discriminator=TYPE_TAG_STR` into `metal` and `pec` fields of `SubpixelSpec` resolves the issue

That is working! thanks

tidy3d/components/eme/simulation.py
```diff
@@ -175,8 +176,8 @@ class EMESimulation(AbstractYeeGridSimulation):
         validate_default=True,
     )
 
-    subpixel: bool = pd.Field(
-        True,
+    subpixel: Union[bool, SubpixelSpec] = pd.Field(
+        SubpixelSpec(),
         title="Subpixel Averaging",
         description="If ``True``, uses subpixel averaging of the permittivity "
```
dbochkov-flexcompute: probably need to update `description` to include info about `SubpixelSpec` as in regular `Simulation`?

dbochkov-flexcompute (commented):
looks good to me! just one thing: would it make sense to rename `MetalConformal` to `PECConformal` since it goes into `pec: ` but not `metal: ` field?

weiliangjin2021:
> looks good to me! just one thing: would it make sense to rename `MetalConformal` to `PECConformal` since it goes into `pec: ` but not `metal: ` field?

We plan to add good conductor handling in the near quarter (a new field `good_conductor` in `SubpixelSpec`). I hope I can reuse `MetalConformal` class in the field.

tidy3d/components/subpixel_spec.py
```diff
@@ -0,0 +1,129 @@
...
+    For PEC interface, the algorithm is based on:
+
+        A. Taflove and S. C. Hagness, "Computational electromagnetics: the
+        finite-difference time-domain method", Chapter 10.3 (2005).
+    """
+
+
+class VolumetricAveraging(AbstractSubpixelAveragingMethod):
+    """Apply volumetric averaging scheme to material properties of Yee grids on structure boundaries.
+    The material property is averaged in the volume surrounding the Yee grid.
```
dmarek-flex, line 55: Does volumetric averaging simply average the permittivity between a metal and dielectric material with weights based on their relative volumes?

tidy3d/components/subpixel_spec.py
```diff
@@ -0,0 +1,129 @@
...
+    """Defines specification for subpixel averaging schemes when added to ``Simulation.subpixel``."""
+
+    dielectric: bool = pd.Field(
+        True,
+        title="Subpixel averaging method on dielectric material interface",
+        description="If ``True``, applies subpixel averaging of the permittivity"
+        "on dielectric material interfaces; otherwise, uses staircasing.",
+    )
+
+    metal: MetalSubpixelType = pd.Field(
```
dmarek-flex, line 126: This is used in the case of interface between dielectric and metal right? Not metal and metal?

tidy3d/components/simulation.py
```diff
@@ -252,6 +255,19 @@ def num_cells_in_monitor(monitor: Monitor) -> int:
             return sum(num_cells_in_monitor(mnt) for mnt in monitor.integration_surfaces)
         return num_cells_in_monitor(monitor)
 
+    @cached_property
+    def _subpixel(self) -> SubpixelSpec():
+        """Subpixel averaging method evaludated based on self.subpixel."""
```
dmarek-flex: `evaluated`

tidy3d/components/subpixel_spec.py
weiliangjin2021: That's right

tidy3d/components/subpixel_spec.py
weiliangjin2021: Between metal + any material other than PEC

weiliangjin2021:
> > looks good to me! just one thing: would it make sense to rename `MetalConformal` to `PECConformal` since it goes into `pec: ` but not `metal: ` field?
> 
> We plan to add good conductor handling in the near quarter (a new field `good_conductor` in `SubpixelSpec`). I hope I can reuse `MetalConformal` class in the field.

Now as I read more about good conductor subpixel, it's likely to have quite different fields compared to PEC-type conformal. In that regard, there is no need to share this class between PEC and metal, so it's a good idea to rename it to `PECConformal`.

tidy3d/components/eme/simulation.py
```diff
@@ -175,11 +176,13 @@ class EMESimulation(AbstractYeeGridSimulation):
...
-        "based on structure definition, resulting in much higher accuracy for a given grid size.",
+        "based on structure definition, resulting in much higher accuracy for a given grid size. "
```
momchil-flex: I think we need to flip this message around and say what SubpixelSpec does; then say what setting `subpixel=True` corresponds to in terms of a SubpixelSpec (dielectric subpixel, everything else staircased).

tidy3d/components/simulation.py
```diff
@@ -150,15 +150,20 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
...
+    subpixel: Union[bool, SubpixelSpec] = pydantic.Field(
+        SubpixelSpec(),
         title="Subpixel Averaging",
-        description="If ``True``, uses subpixel averaging of the permittivity "
-        "based on structure definition, resulting in much higher accuracy for a given grid size.",
+        description="If ``False``, uses staircasing on structure interfaces; if ``True``, uses "
+        "default subpixel averaging methods of the permittivity based on structure definition "
+        "resulting in much higher accuracy for a given grid size. "
+        "Alternatively, user may supply a :class:`SubpixelSpec` to this field to select subpixel "
+        "averaging methods separately for dieletric, metal, and PEC materials.",
```
momchil-flex: Same comment as for eme sim. Side note - why is the `subpixel` field overwritten in the EME simulation currently @caseyflex ? I don't see any difference right now. I guess the thought is that in the future you may include definitions/toggles for some eme-specific subpixel in the normal direction there?

tidy3d/components/subpixel_spec.py
```diff
@@ -0,0 +1,138 @@
...
+class VolumetricAveraging(AbstractSubpixelAveragingMethod):
+    """Apply volumetric averaging scheme to material properties of Yee grids on structure boundaries.
+    The material property is averaged in the volume surrounding the Yee grid.
+    """
+
+    staircase_normal_component: bool = pd.Field(
+        True,
+        title="Staircasing for field components substantially normal to interface",
+        description="Volumetric averaging works accurately if the electric field component "
+        "is substantially tangential to the interface. If ``True``, apply staircasing instead "
```
momchil-flex: instead -> only

tidy3d/components/subpixel_spec.py
```diff
@@ -0,0 +1,138 @@
...
+        stability criterion",
+        IEEE Transactions on Antennas and Propagation, 54(6), 1843 (2006).
+    """
+
+    timestep_reduction: float = pd.Field(
+        DEFAULT_COURANT_REDUCTION_PEC_CONFORMAL,
+        title="Time Step Size Reduction Rate",
+        description="Reduction factor between 0 and 1 such that the simulation's time step size "
+        "will be ``1 - timestep_reduction`` times its default value. "
+        "Accuracy can be improved with a smaller time step size; but simulation time increased as well.",
```
momchil-flex: ; -> ,

momchil-flex (commented):
Few minor comments about the current code. Otherwise, two more questions to discuss further:
- for `dielectric`, should we avoid using boolean but define an explicitly named spec instead (also in view of defining different ones in the future)
- are we eventually going to have a `good_conductor` kind of field in `SubpixelSpec` too, and how is that going to interact with `metal`?

tidy3d/components/simulation.py
caseyflex: I think it can be removed from `EMESimulation` in the same way it was removed from `Simulation`, and just left in `AbstractYeeGridSimulation`.

tidy3d/components/eme/simulation.py
weiliangjin2021: when `subpixel=True`, now conformal is also applied to PEC, as it is in general much more accurate than staircasing.

weiliangjin2021:
> * are we eventually going to have a `good_conductor` kind of field in `SubpixelSpec` too, and how is that going to interact with `metal`?

In the 1st release, we might just limit `good_conductor` to `Medium` type, as surface-impedance model is usually applied to approximate the permittivity as purely imaginary coming from DC conductivity. In the next step, maybe we can add a field to medium to indicate whether it's allowed to be handled as surface-impedance model. If allowed and the material has Im[ep]/Re[ep] larger than a threshold, it'll be approximated as `Medium` type at the central frequency.

momchil-flex (approved):

dmarek-flex (approved):

tidy3d/components/simulation.py
```diff
@@ -252,6 +257,19 @@ def num_cells_in_monitor(monitor: Monitor) -> int:
             return sum(num_cells_in_monitor(mnt) for mnt in monitor.integration_surfaces)
         return num_cells_in_monitor(monitor)
 
+    @cached_property
+    def _subpixel(self) -> SubpixelSpec():
```
tylerflex: remove `()`

tidy3d/plugins/adjoint/components/simulation.py
```diff
@@ -177,9 +178,11 @@ def _output_monitors_colocate_false(cls, val):
 
     @pd.validator("subpixel", always=True)
     def _subpixel_is_on(cls, val):
-        """Assert subpixel is on."""
-        if not val:
-            raise AdjointError("'JaxSimulation.subpixel' must be 'True' to use adjoint plugin.")
+        """Assert dielectric subpixel is on."""
+        if (isinstance(val, SubpixelSpec) and isinstance(val.dielectric, Staircasing)) or not val:
+            raise AdjointError(
+                "'JaxSimulation.subpixel.dielectric' must not be 'Staircasing' to use adjoint plugin."
```
tylerflex: people might be confused if `subpixel=False`. Maybe we can just say that 
```
'JaxSimulation.subpixel' must be 'True' or a specific 'SubpixelSpec' to use adjoint plugin.
```
?

tylerflex (approved):
Looks good @weiliangjin2021 !

tidy3d/plugins/adjoint/components/simulation.py
tylerflex: actually maybe what I wrote is also not specific enough, `a specific 'SubpixelSpec'` -> `a 'SubpixelSpec' with no dielectric staircasing`?

PR #1697, opened 2024-05-10
🗻 autograd tidy3d (tidygrad?)

tylerflex:
basic framework for autogradding tidy3d.

1. Adds `autograd.ArrayBox` type annotation in traced fields. For now just `Medium.permittivity`.
2. Adds `json_encoder` for this type, to write it to json as an untraced `float`.
3. Adds a `autograd.extend.primitive` version of `web.run()` that tells autograd how to connect the traced-fields to the xarray values in `SimulationData.data` through an adjoint simulation. Only works for `AmpsDataArray`. Only uses emulated run.

## Todo Done
- [x] Runs with emulated test
- [x] Runs with web API
- [x] Runs with extraneous monitors, structures
- [x] Grad roughly matches numerical.
- [x] Do all existing tidy3d tests pass with the changes so far?
- [x] Test data array selection, interpolation in objective function.
- [x] Where will all of the code live?
- [x] Put floating helper methods as class methods.
- [x] Auto handle traced vs untraced simulation fields.
- [x] Auto handle traced vs untraced data fields.
- [x] Handle integration with box bounds.
- [x] Implement PIC structure fields
   - [x] `Medium`
   - [x] `CustomMedium`
   - [x] `PolySlab`
   - [x] `Box`
 - [x] Implement PIC adjoint sources
    - [x] `DiffractionMonitor`
    - [x] `ModeMonitor`
- [x] `run_batch` support
- [x] Replace existing web API
- [x] Autograd dependency

## **Todo Before 2.7.0**
- [x] Invdes
    - [x] penalty and transformation classes (ya) <-
    - [x] make sure we like the final API
- [x] Autograd plugin at `tidy3d/plugins/autograd`
    - [x] Where do we import `autograd` from?
    - [x] Functional autograd utilities (ya)
- [ ] Polishing
    - [x] Warn or disallow autograd usage for some contains (eg. num structures). (th)
    - [ ] `yield` instead of return `dict` in `run_async` and `invdes`
    - [x] why is the bend notebook slow?
- [ ] Testing
    - [x] Test speed of many structures. [think 400 max is a safe limit (2 min)]
    - [x] Test validators and edge cases.
    - [ ] Test numerical accuracy.
- [x] Reviews
    - [x] Only use `autograd.numpy` where we really need it.
    - [x] Thorough code reviews.
- [x] Documentation
    - [x] Update all Autograd compatible notebooks.
    - [x] Write readme, including limitations and supported features.
    - [x] Deprecate adjoint / jax stuff.
    - [x] Document limitations of autograd.

## Todo for 2.8.0
- [ ] Multi-frequency -> new notebook
- [ ] Web API
    - [ ] Server-side processing.
    - [ ] Selective ignoring of non-traced batch elements (if necessary)
- [ ] Feature support
    - [ ] GeometryGroup -> metalens
    - [ ] ComplexPolyslab
    - [ ] AnisotropicMedium
    - [ ] FieldMonitor -> metalens
    - [ ] `.flux`
    - [ ] `get_intensity()`
    - [ ] `get_poynting()`
- [ ] Invdes
    - [ ] Optimizer extensions
        - [ ] scipy / LBFGS
        - [ ] nlopt
        - [ ] pytorch
        - [ ] optax
     - [ ] Postprocess function definitions
 - [ ] Jax, PyTorch, tensorflow converters. 

## Todo eventually
- [ ] `DataArray` improvements
   - [ ] differentiable interpolation with autograd `ArrayBox` dtypes.
   - [ ] Be able to return just the single element `DataArray`, right now need to grab `values` for it to register.
- [ ] Invdes 
    - [ ] GUI integration
    - [ ] web API support
- [ ] Feature support
    - [ ] utility to look at gradient at some location? gradient monitor?


tests/test_components/test_base.py
tylerflex, line 1: note: default behavior of copy changed so now you CAN set `deep=False` (although `True` is default).

tidy3d/web/api/webapi.py
tylerflex, line 1: I will need to be able to pass the parent tasks to `run()` for the server-side implementation.

tidy3d/components/types.py
tylerflex, line 1: note: just making things autograd compatible

poetry.lock
yaugenst-flex, line 1: Lots of updates/changes here, but probably not really required for the PR?

results.prof
yaugenst-flex, line 1: Remove?

tests/test_plugins/test_invdes.py
```diff
@@ -1,31 +1,28 @@
...
-import tidy3d.plugins.adjoint as tda
 import tidy3d.plugins.invdes as tdi
 
 # use single threading pipeline
-from . import test_adjoint as ta
 
 from .test_adjoint import use_emulated_run, use_emulated_run_async  # noqa: F401
 
-from ..utils import run_emulated, assert_log_level, AssertLogLevel
+from ..utils import run_emulated, run_async_emulated, AssertLogLevel
```
yaugenst-flex, line 16: Also a general comment, but if these fixtures are put into `conftest.py` instead of `utils`, then they would be available to pytest without importing. Would fix all of the `F401` warnings. We should probably restructure all of the tests a bit.

tidy3d/components/autograd.py
```diff
@@ -0,0 +1,67 @@
...
+        bmin = get_static(bmin)
+        bmax = get_static(bmax)
+
+        coord_values = _arr.coords[dim].values
+
+        # reset all coordinates outside of bounds to the bounds, so that dL = 0 in integral
+        coord_values[coord_values < bmin] = bmin
+        coord_values[coord_values > bmax] = bmax
+
+        # TODO: this can be a little bit slow, might be a faster implementation
```
yaugenst-flex: Not sure, but I think this might be faster if the assignment is outside of the loop, i.e. build the coordinates in the loop, and then only do one assignment outside. Something like:
```python
all_coords = {}
for dim, (bmin, max) in zip(dims, bounds):
    ...  # lines 44:52
    all_coords[dim] = coord_values
_arr = _arr.assign_coords(**all_coords)
```

tidy3d/components/base.py
```diff
@@ -205,15 +210,12 @@ def _special_characters_not_in_name(cls, values):
         "by calling ``obj.json()``.",
     )
 
-    def copy(self, **kwargs) -> Tidy3dBaseModel:
+    def copy(self, deep: bool = True, **kwargs) -> Tidy3dBaseModel:
         """Copy a Tidy3dBaseModel.  With ``deep=True`` as default."""
-        if "deep" in kwargs and kwargs["deep"] is False:
```
yaugenst-flex, line 210: `if not kwargs.get("deep", True)`?

tidy3d/components/base.py
```diff
@@ -205,15 +210,12 @@ def _special_characters_not_in_name(cls, values):
...
-            raise ValueError("Can't do shallow copy of component, set `deep=True` in copy().")
-        kwargs.update(dict(deep=True))
```
yaugenst-flex, line 212: `kwargs.update(deep=True)`

tidy3d/components/base.py
```diff
@@ -922,6 +927,133 @@ def make_json_compatible(json_string: str) -> str:
...
+
+                # or just grab the static value out of the values
+                elif include_untraced_data_arrays:
+                    field_mapping[path] = get_static(x.values)
+
+            # for sequences, add (i,) to the path and handle each value individually
+            elif isinstance(x, (list, tuple)):
+                for i, val in enumerate(x):
+                    sub_paths = tuple(list(path) + [i])
+                    handle_value(val, path=sub_paths)
```
yaugenst-flex: Tiny performance boost, just do `handle_value(val, path=path + (i,))` or `handle_value(val, path=(*path, i))`

tidy3d/components/base.py
```diff
@@ -922,6 +927,133 @@ def make_json_compatible(json_string: str) -> str:
...
+
+            # for dictionaries, add the (key,) to the path and handle each value individually
+            elif isinstance(x, dict):
+                for key, val in x.items():
+                    sub_paths = tuple(list(path) + [key])
+                    handle_value(val, path=sub_paths)
```
yaugenst-flex: Same here: `handle_value(val, path=path + (key,))`

tidy3d/components/base.py
```diff
@@ -922,6 +927,133 @@ def make_json_compatible(json_string: str) -> str:
...
+        Returns
+        -------
+        dict
+            mapping of traced fields used by ``autograd``
+
+        """
+
+        field_mapping = {}
+
+        def handle_value(x: Any, path: tuple[str, ...]) -> None:
```
yaugenst-flex, line 953: If this is potentially slow, there is also the option of avoiding the recursion altogether by using a stack and at least avoid those recursive function calls, e.g.:
```python
stack = [(self.dict(), starting_path)]

while stack:
    x, path = stack.pop()
    ...  # pretty much the same logic as  before, except for replacing handle_value() calls with stack.append()
```
This might be a bit faster for deeply nested structures. Probably won't be very significant though.

tidy3d/components/base.py
```diff
@@ -922,6 +927,133 @@ def make_json_compatible(json_string: str) -> str:
...
+        if starting_path:
+            for key in starting_path:
+                self_dict = self_dict[key]
+
+        handle_value(self_dict, path=starting_path)
+
+        # convert the resulting field_mapping to an autograd-traced dictionary
+        return dict_ag(field_mapping)
+
+    def insert_traced_fields(self, field_mapping: AutogradFieldMap) -> Tidy3dBaseModel:
```
yaugenst-flex, line 995: Similar comment to `strip_traced_fields` here too, this can be faster without recursion, the following code is roughly 1.6x faster:
```python
def insert_traced_fields2(self, field_mapping):
    self_dict = self.dict()

    def insert_value(x, path: tuple[str, ...], sub_dict: dict):
        """Insert a value into the path into a dictionary."""
        current_dict = sub_dict
        for key in path[:-1]:
            if isinstance(current_dict[key], tuple):
                current_dict[key] = list(current_dict[key])
            current_dict = current_dict[key]

        final_key = path[-1]
        if isinstance(current_dict[final_key], tuple):
            current_dict[final_key] = list(current_dict[final_key])

            if "AUTOGRAD" in sub_element.attrs:
                current_dict[final_key].attrs["AUTOGRAD"] = x
        else:
            current_dict[final_key] = x

    for path, value in field_mapping.items():
        insert_value(value, path=path, sub_dict=self_dict)

    return self.parse_obj(self_dict)
```

I only tested this on some mock data though so there is a chance it breaks things :grimacing:

tidy3d/components/data/monitor_data.py
```diff
@@ -77,6 +78,36 @@ def _updated(self, update: Dict) -> MonitorData:
...
+        """Generate adjoint sources for this ``MonitorData`` instance."""
+
+        # TODO: if there's data in the MonitorData, but no adjoint source, then
+        # user is trying to differentiate something that is un-supported by us
+        # warn?
+
+        return []
+
+    @staticmethod
+    def flip_direction(direction: str) -> str:
```
yaugenst-flex: I guess the type hint here should be `direction: Union[str, DataArray]`?

tidy3d/components/data/monitor_data.py
```diff
@@ -1561,6 +1592,78 @@ def to_dataframe(self) -> DataFrame:
...
+        return adjoint_sources
+
+    def make_adjoint_sources_amps(self) -> list[ModeSource]:
+        """Generate adjoint sources for ``ModeMonitorData.amps``."""
+
+        coords = self.amps.coords
+
+        adjoint_sources = []
+
+        for freq in coords["f"]:
```
yaugenst-flex, line 1621: Maybe a note for later, but I think this can be sped up significantly using xarray's `apply_ufunc` instead of those nested loops.

tidy3d/components/data/monitor_data.py
```diff
@@ -2581,6 +2684,100 @@ def _make_dataset(self, fields: Tuple[np.ndarray, ...], keys: Tuple[str, ...]) -
...
+    def make_adjoint_sources_amps(self) -> list[PlaneWave]:
+        """Make adjoint sources for outputs that depend on DiffractionData.`amps`."""
+
+        amps = self.amps
+        coords = amps.coords
+
+        adjoint_sources = []
+
+        # loop over all coordinates in the diffraction amplitudes
+        for freq in coords["f"]:
```
yaugenst-flex, line 2706: Same here, I think this can be done more efficiently using xarray utilities using filtering and mapping, something along the lines of `valid_amps = amps.where((amps != 0.0) & (~np.isnan(amps)), drop=True)` and then doing something like `valid_amps.map(self.adjoint_source_amp)`. It's fine as-is but maybe we can invest some time later to optimize these things.

tidy3d/components/base_sim/simulation.py
```diff
@@ -134,7 +135,7 @@ def _structures_not_at_edges(cls, val, values):
                 struct_bounds = list(struct_bound_min) + list(struct_bound_max)
 
                 for sim_val, struct_val in zip(sim_bounds, struct_bounds):
-                    if isclose(sim_val, struct_val):
+                    if anp.isclose(sim_val, struct_val):
```
momchil-flex, line 138: This is not as innocent a change as it may look. There are two main differences.

- The default values for the tolerance are different, probably not that important here
- The numpy version converts values to arrays. This has proven slow before in some cases, like I actually remember changing a `numpy.isclose` to a `math.isclose` in order to fix a performance issue somewhere, in the case of too many structures. I think this was in the mesher, where the scaling may have been N_structures ** 2, but it's definitely something to keep in mind. In this case though I see no performance difference in the backend `test_simulation.py` between the two functions, so it seems like it's fine.

tidy3d/components/data/sim_data.py
```diff
@@ -804,3 +813,60 @@ def source_spectrum_fn(freqs):
...
+        # fields to update the 'fwd' simulation with to make it 'adj'
+        sim_adj_update_dict = dict(
+            sources=sources_adj,
+            boundary_spec=bc_adj,
+            monitors=adjoint_monitors,
+        )
+
+        # set the ADJ grid spec wavelength to the original wavelength (for same meshing)
+        grid_spec_original = sim_original.grid_spec
+        if len(sim_original.sources) and grid_spec_original.wavelength is None:
```
yaugenst-flex: `if len(sim_original.sources)` --> `if sim_original.sources`

tidy3d/components/data/sim_data.py
```diff
@@ -804,3 +813,60 @@ def source_spectrum_fn(freqs):
...
+
+        return sim_original.updated_copy(**sim_adj_update_dict)
+
+    def make_adjoint_sources(self, data_vjp_paths: set[tuple]) -> list[Source]:
+        """Generate all of the non-zero sources for the adjoint simulation given the VJP data."""
+
+        # TODO: determine if we can do multi-frequency sources
+
+        # map of index into 'self.data' to the list of datasets we need adjoint sources for
+        adj_src_map = {}
```
yaugenst-flex: This can be simplified somewhat using `collections.defaultdict`:
```python
adj_src_map = defaultdict(list)
for _, index, dataset_name in data_vjp_paths:
    adj_src_map[index].append(dataset_name)
```

tidy3d/components/geometry/base.py
```diff
@@ -96,6 +98,26 @@ def _ensure_equal_shape(*arrays):
         if len(shapes) > 1:
             raise ValueError("All coordinate inputs (x, y, z) must have the same shape.")
 
+    @staticmethod
+    def make_shapely_box(minx: float, miny: float, maxx: float, maxy: float) -> shapely.box:
```
yaugenst-flex, line 102: Ok this might go into golfing territory but the whole function could be written as `shapely.geometry.box(*map(get_static, (minx, miny, maxx, maxy)))`

tidy3d/components/geometry/base.py
```diff
@@ -96,6 +98,26 @@ def _ensure_equal_shape(*arrays):
...
+
+        minx = get_static(minx)
+        miny = get_static(miny)
+        maxx = get_static(maxx)
+        maxy = get_static(maxy)
+
+        return shapely.box(minx, miny, maxx, maxy)
+
+    @staticmethod
+    def make_shapely_point(minx: float, miny: float) -> shapely.Point:
```
yaugenst-flex, line 113: Same here.

tidy3d/components/geometry/base.py
```diff
@@ -2263,6 +2301,200 @@ def _surface_area(self, bounds: Bound) -> float:
...
+        )
+
+        # post-process these values to give the gradients w.r.t. center and size
+        vjps_center_size = self.derivatives_center_size(vjps_faces=vjps_faces)
+
+        # store only the gradients asked for in 'field_paths'
+        derivative_map = {}
+        for field_path in field_paths:
+            field_name, *index = field_path
+            field_name = field_name
```
yaugenst-flex: Remove `field_name = field_name`

tidy3d/components/geometry/polyslab.py
```diff
@@ -1344,6 +1361,206 @@ def _surface_area(self, bounds: Bound) -> float:
...
+
+        interp_kwargs = {}
+        for dim, centers_dim in zip("xyz", edge_centers.T):
+            # only include dims where the data has more than 1 coord, to avoid warnings and errors
+            coords_data = der_map.field_components[f"E{dim}"].coords
+            if np.array(coords_data).size > 1:
+                interp_kwargs[dim] = xr.DataArray(centers_dim, dims=edge_index_dim)
+
+        components = {}
+        for fld_name, arr in der_map.field_components.items():
```
yaugenst-flex, line 1491: Nitpick but this really looks more like a dict comprehension to me...
```python
components = {
        name: arr.interp(**interp_kwargs).sum("f")
        for name, arr in der_map.field_components.items()
}
```

tidy3d/components/simulation.py
```diff
@@ -3195,6 +3203,52 @@ def _warn_time_monitors_outside_run_time(self) -> None:
...
+
+            adjoint_monitors_fld.append(mnt_fld)
+            adjoint_monitors_eps.append(mnt_eps)
+
+        return adjoint_monitors_fld, adjoint_monitors_eps
+
+    @property
+    def freqs_adjoint(self) -> list[float]:
+        """Unique list of all frequencies. For now should be only one."""
+        freqs = []
```
yaugenst-flex: This should be a `set`:
```python
freqs = set()
for mnt in self.monitors:
    if isinstance(mnt, FreqMonitor):
        freqs.update(mnt.freqs)
freqs = sorted(freqs)
```

tidy3d/components/simulation.py
```diff
@@ -3227,7 +3281,7 @@ def _run_time(self) -> float:
 
         # get the maximum refractive index evaluated over each of all the source central frequencies
         all_ref_inds = [self.get_refractive_indices(src.source_time.freq0) for src in self.sources]
-        avg_ref_inds = [np.mean(n) for n in all_ref_inds]
+        avg_ref_inds = [np.mean(np.array(n)) for n in all_ref_inds]
```
yaugenst-flex, line 3288: Is the `np.array(n)` really necessary? I would have thought that `np.mean()` already takes care of that?

tidy3d/components/types.py
```diff
@@ -69,9 +69,10 @@ def load_complex(cls, val):
     def convert_to_numpy(cls, val):
         """Convert the value to np.ndarray and provide some casting."""
         arr_numpy = np.array(val, ndmin=1, dtype=cls.dtype, copy=True)
-        arr_tidy3d = np.ndarray(shape=arr_numpy.shape, dtype=arr_numpy.dtype)
-        arr_tidy3d[:] = arr_numpy
-        return arr_tidy3d
+        try:
+            return arr_numpy.astype(arr_numpy.dtype)
```
yaugenst-flex: This is being casted to its own `dtype`?

tidy3d/web/api/autograd/autograd.py
```diff
@@ -0,0 +1,698 @@
...
+    if num_traced_structures > MAX_NUM_TRACED_STRUCTURES:
+        raise ValueError(
+            f"Autograd support is currently limited to {MAX_NUM_TRACED_STRUCTURES} structures with "
+            f"traced fields. Found {num_traced_structures} structures with traced fields."
+        )
+
+    return True
+
+
+def is_valid_for_autograd_async(simulations: dict[str, td.Simulation]) -> bool:
```
yaugenst-flex, line 73: I'd suggest to just return the result of those checks:
```python
def is_valid_for_autograd_async(simulations: dict[str, td.Simulation]) -> bool:
    return isinstance(simulations, dict) and all(
        is_valid_for_autograd(sim) for sim in simulations.values()
    )
```

yaugenst-flex (commented):
What a PR :smile:
Looks good, and obviously we've been playing around with it and it works. There are some areas where performance could be improved but I don't think there is really any blocking issue.

tests/test_plugins/test_invdes.py
tylerflex: hm, ok. Maybe we can do this as part of #1662 ? or maybe a separate test reorg?

tidy3d/components/autograd.py
tylerflex: yea good idea, thanks. implemented

tests/test_plugins/test_invdes.py
yaugenst-flex: If it's only this change then #1662 probably makes sense, I'll look into it. At a later stage we'll need to look at the tests a bit more closely too though.

tidy3d/components/base.py
tylerflex: thanks, this line was removed but I did forget to update the kwargs in `copy()`. Fixed.

tidy3d/components/base.py
tylerflex: thanks, I actually didnt know until now that you can add tuples like this 🤦

tidy3d/components/base.py
tylerflex: woah, nice. we could use that 1.6x. Changed

tidy3d/components/base.py
tylerflex: this line was removed anyway

tidy3d/components/base.py
tylerflex: I don't understand actually, sorry but could you write the function for me if you think it will improve speed? (we can also leave for later if you want)

tidy3d/components/data/monitor_data.py
tylerflex: fixed

tidy3d/components/data/monitor_data.py
tylerflex: hm, yea will note. not sure if this is too slow yet, though.

tidy3d/components/data/monitor_data.py
tylerflex: ok I'll make a note TODO for later

tidy3d/components/data/sim_data.py
tylerflex: thanks, I was afraid of using defaultdict but this does make it better. I think it pops up in other places too, will change

tidy3d/components/geometry/base.py
tylerflex: I might just leave it for clarity, but yea this would work too..

tylerflex: unless you have strong objections

tidy3d/components/simulation.py
tylerflex: I think it is.. this was some bug I caught in the testing. It's something weird related to the data structure of `n`. It's like a tuple of tuples or something.

tidy3d/components/types.py
tylerflex: hm. yea maybe this whole try/except block is unnecessary. Will remove and test.

tidy3d/web/api/autograd/autograd.py
tylerflex: yea I refactored this earlier today and it does this and is a bit cleaner overall.

tidy3d/components/geometry/polyslab.py
tylerflex: I'm being old fashioned, but I find this less readable :|

tylerflex:
Ok addressed everything, minus a few places where I like the boring syntax :P

poetry.lock
tylerflex: I need to run `poetry lock` for tests to pass :/

results.prof
tylerflex: oops, thanks

tidy3d/components/geometry/polyslab.py
yaugenst-flex: Haha sure 😄 doesn't make much of a difference either way.

tidy3d/components/geometry/base.py
yaugenst-flex: No I'm a bit sceptical myself haha

tidy3d/components/base.py
yaugenst-flex: Sure yeah, like this:
```python
field_mapping = {}
self_dict = self.dict()

# Navigate to the correct subset of self.dict() if starting_path is provided
if starting_path:
    for key in starting_path:
        self_dict = self_dict[key]

stack = [(self_dict, starting_path)]
while stack:
    x, path = stack.pop()

    if isbox(x):
        field_mapping[path] = x
    elif isinstance(x, DataArray):
        if AUTOGRAD_KEY in x.attrs:
            field_mapping[path] = x.attrs[AUTOGRAD_KEY]
        elif include_untraced_data_arrays:
            field_mapping[path] = get_static(x.values)
    elif isinstance(x, (list, tuple)):
        for i, val in enumerate(x):
            sub_paths = tuple(list(path) + [i])
            stack.append((val, sub_paths))
    elif isinstance(x, dict):
        for key, val in x.items():
            sub_paths = tuple(list(path) + [key])
            stack.append((val, sub_paths))

return dict_ag(field_mapping)
```

But note I haven't actually tested this so it might not work as-is, but I hope it gets the idea across 😅

tidy3d/components/simulation.py
```diff
@@ -3235,15 +3235,14 @@ def make_adjoint_monitors(self, structure_indices: set[int]) -> tuple[list, list
         return adjoint_monitors_fld, adjoint_monitors_eps
 
     @property
-    def freqs_adjoint(self) -> list[float]:
+    def freqs_adjoint(self) -> set[float]:
```
yaugenst-flex: The return type here is still a list, `sorted(set)` converts the set to a sorted list.

tylerflex: ah right, fixed. also used a `ValueError` instead of `assert`

tidy3d/components/base.py
tylerflex: it worked! but was about 5% slower than the original :/

yaugenst-flex: oh damn :(

tidy3d/components/grid/mesher.py
```diff
@@ -223,12 +228,12 @@ def parse_structures(
...
             if max(intervals["structs"][coord_ind]) >= num_unenforced:
                 max_step = structure_steps[max(intervals["structs"][coord_ind])]
-                max_steps.append(float(max_step))
+                max_steps.append(max_step)
             # otherwise, define the max step as the minimum over all medium steps
             # of media in this interval
             else:
                 max_step = np.amin(structure_steps[intervals["structs"][coord_ind]])
-                max_steps.append(float(max_step))
+                max_steps.append(max_step)
```
momchil-flex, line 236: I'm not really sure why `float` was needed there and hence whether it's safe to be removed. Maybe it was to convert a size 1 array to a float? I guess if the tests are passing, we should hope it's fine.

It seems like some of the changes in this file are cosmetic rather than necessiated by autograd?

momchil-flex (approved):
I took a look around (and Yannick had already commented yesterday on a couple things that I also had in mind). I think generally it's exciting to have this more native way to do adjoint. I am still a little apprehensive about the introduction of autograd on such a deep level, but I guess let's see how it goes.

One thing that would be nice to clarify though (for future development) is what exactly do we need to keep in mind to avoid issues when devleoping things that may have nothing to do with adjoint/autograd. Obviously some of the changes in the core code were made because of autograd (e.g. using `autograd.numpy` instead of `numpy`, `anp.real(x)` instead of `x.real`, etc.). Say I just want to add a new validator on Simulation.structures. Anything specific that I should keep in mind?

yaugenst-flex (approved):

tylerflex:
> One thing that would be nice to clarify though (for future development) is what exactly do we need to keep in mind to avoid issues when devleoping things that may have nothing to do with adjoint/autograd. Obviously some of the changes in the core code were made because of autograd (e.g. using `autograd.numpy` instead of `numpy`, `anp.real(x)` instead of `x.real`, etc.). Say I just want to add a new validator on Simulation.structures. Anything specific that I should keep in mind?

hm, yea so I will clarify these in a section of the [README](https://github.com/flexcompute/tidy3d/blob/09d311ca2749360b77e61ca09e77e87834de73e9/tidy3d/plugins/autograd/README.md) too, but I think the main ones are:
* `numpy` -> `autograd.numpy`
* `.real` -> `np.real()`
* `float()` is not supported as far as I can tell.
* `isclose()` -> `np.isclose()`
* `array[i] = something` needs a different approach (happens in mesher a lot)
* whenever we pass things to other modules, like `shapely` especially, we need to be careful that they are untraced.
* I just made structures static before any meshing, as a cutoff point. So if we add a new `make_grid()` call somewhere, eg in a validator, just need to be aware.

That probably covers it.

tidy3d/components/base_sim/simulation.py
tylerflex: yea I definitely did this with some hesitation, but the `math.isclose()` function broke with autograd so I had to change it. Good thing it doesn't break any tests or lead to performance changes. I'll assume it's ok then?

tidy3d/components/grid/mesher.py
tylerflex: yea I wasn't sure either, but similar to `math.isclose()`, it was breaking autograd and when I removed it everything w worked fine.

And yea the other changes are cosmetic, but actually the `make_shapely_box()` was a remnant from when I was trying to make this autograd compatible. If it doesn't bother you, maybe we can just leave it in case we revisit this?

tidy3d/components/base_sim/simulation.py
momchil-flex: Yeah, let's assume that until proven wrong.

tidy3d/components/grid/mesher.py
momchil-flex: Yeah that's fine.

momchil-flex:
> > One thing that would be nice to clarify though (for future development) is what exactly do we need to keep in mind to avoid issues when devleoping things that may have nothing to do with adjoint/autograd. Obviously some of the changes in the core code were made because of autograd (e.g. using `autograd.numpy` instead of `numpy`, `anp.real(x)` instead of `x.real`, etc.). Say I just want to add a new validator on Simulation.structures. Anything specific that I should keep in mind?
> 
> hm, yea so I will clarify these in a section of the [README](https://github.com/flexcompute/tidy3d/blob/09d311ca2749360b77e61ca09e77e87834de73e9/tidy3d/plugins/autograd/README.md) too, but I think the main ones are:
> 
> * `numpy` -> `autograd.numpy`
> * `.real` -> `np.real()`
> * `float()` is not supported as far as I can tell.
> * `isclose()` -> `np.isclose()`
> * `array[i] = something` needs a different approach (happens in mesher a lot)
> * whenever we pass things to other modules, like `shapely` especially, we need to be careful that they are untraced.
> * I just made structures static before any meshing, as a cutoff point. So if we add a new `make_grid()` call somewhere, eg in a validator, just need to be aware.
> 
> That probably covers it.

I think also even before those autograd specific issues the question is when can we convert to static internally and when we should not. Ideally never? But in practice we may have to do that sometimes either for coding convenience or for performance?

I'm genuinely both excited and scared of this change. :) The dream was always to have a tidy3d simulation that works exactly as the usual one, but everything is differentiable. This is a big step towards that dream. But I'm super anxious knowing how much we've struggled with things like performance in the case of like 1M structures previously. Just hoping this will not bring some of this back.

tylerflex:
> I think also even before those autograd specific issues the question is when can we convert to static internally and when we should not. Ideally never? But in practice we may have to do that sometimes either for coding convenience or for performance?

I think only when we send these structures to the mesher do we need to make them static.

> I'm genuinely both excited and scared of this change. :) The dream was always to have a tidy3d simulation that works exactly as the usual one, but everything is differentiable. This is a big step towards that dream. But I'm super anxious knowing how much we've struggled with things like performance in the case of like 1M structures previously. Just hoping this will not bring some of this back.

Yea, it's a bit of a risk but I feel pretty optimistic about it and I think the functionality it provides will be worth it.

If it makes you feel better, there are a few fail safes in place in case things don't go as expected. Also, in the future we can consider ways to optimize further, including conditions where we can skip the static structures. For example, maybe in some conditions (many structures) we can just error right away if tracers are present. See here: https://github.com/flexcompute/tidy3d/blob/tyler/autograd/tidy3d/web/api/autograd/autograd.py#L47-L70

Or additionally, tag the simulation with some flag if we know for sure it doesn't have tracers.


Assuming this is ok to merge once I squash commits? I just added the changelog. of course I forgot it before :D

momchil-flex:
>> Or additionally, tag the simulation with some flag if we know for sure it doesn't have tracers.

> I was thinking of suggesting something like this during our `test_simulation.py` time discussion, but I felt like there's no way to have such a tag without iterating through the structures anyway. Maybe there is some way to do it dynamically but not obvious. Like, it would have to happen during the pydantic model construction? Start with `False`, set to `True` as soon as something with a tracer is added, then if `True` skip that check altogether for subsequent objects?

Yea may be no way to avoid iteration, but your idea could work.

Maybe one way that would be faster than currently is this:

instead of
```py
def static_structures(self):
    return [s.to_static() for s in s.structures]
```
we could just do
```py
def static_structures(self):
    return self.to_static().structures
```

and that could maybe cut down on just calling `to_static()` too many times?

tylerflex:
or we could use `@cached_property` to cache a `static_self`?

tylerflex:
actually I just tried both of these ideas and they basically froze my computer :D.

I'll merge and then note some improvements we can consider next.

Thanks for the review

PR #1698, opened 2024-05-10
`JaxComplexPolySlab` support

### LINKED ISSUE #1675: `JaxComplexPolySlab` support
Hey @yaugenst-flex can you take on this project after `slab_bounds`? I think it will be pretty natural after that issue and will get you more exposure to the wild world of tidy3d `adjoint`. Also was requested by a customer so will be a high impact feature. Thanks!

Basically the user may have some `vertices` that define a `JaxPolySlab` but become crossing under erosion / dilation / or sidewall extrusion. We currently use the `ComplexPolySlab` plugin to handle such cases, which lets the user split these crossing poly slabs into several individual ones. But this is not compatible with `JaxPolySlab`. 

I think a simple test case can be a triangle `PolySlab` with several vertices and a `sidewall_angle > 0`. By adjusting the number of vertices and the sidewall_angle, we can trigger this validator to fail. A new `JaxComplexPolySlab` wrapper should allow us to split it into `JaxPolySlab` and let the user select one (or a few) to model in their `JaxStructure`.

Here's a notebook showing how complex polyslab works: https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.plugins.polyslab.ComplexPolySlab.html

After this one, I think we can start digging into a bit more long term projects, and architecture work with some minor improvements sprinkled in. 

Thanks a lot!

yaugenst-flex:
Closes #1675 

Initial version of `JaxComplexPolySlab`. Works in principle, but still needs tests. Supports differentiation w.r.t. both `vertices` and `slab_bounds`.

### DIFF
--- a/tests/test_plugins/test_adjoint.py
+++ b/tests/test_plugins/test_adjoint.py
@@ -11,6 +11,7 @@ from numpy.testing import assert_allclose
 from xarray import DataArray
 from jax import grad
 import jax
+from jax.test_util import check_grads
 import time
 import matplotlib.pyplot as plt
 import h5py
@@ -20,7 +21,13 @@ import gdstk
 import tidy3d as td
 
 from tidy3d.exceptions import DataError, Tidy3dKeyError, AdjointError
-from tidy3d.plugins.adjoint.components.geometry import JaxBox, JaxPolySlab, MAX_NUM_VERTICES
+from tidy3d.plugins.polyslab import ComplexPolySlab
+from tidy3d.plugins.adjoint.components.geometry import (
+    JaxBox,
+    JaxPolySlab,
+    JaxComplexPolySlab,
+    MAX_NUM_VERTICES,
+)
 from tidy3d.plugins.adjoint.components.geometry import JaxGeometryGroup
 from tidy3d.plugins.adjoint.components.medium import JaxMedium, JaxAnisotropicMedium
 from tidy3d.plugins.adjoint.components.medium import JaxCustomMedium, MAX_NUM_CELLS_CUSTOM_MEDIUM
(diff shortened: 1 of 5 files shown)


tidy3d/plugins/adjoint/components/geometry.py
```diff
@@ -279,6 +288,15 @@ class JaxPolySlab(JaxGeometry, PolySlab, JaxObject):
         stores_jax_for="vertices",
     )
 
+    slab_bounds_jax: Tuple[JaxFloat, JaxFloat] = pd.Field(
```
tylerflex, line 291: did this sneak in? although I think it's fine as we'll just do it in the slab_bounds PR

tylerflex (commented):
Wow, this is looking really good! Thanks @yaugenst-flex . I think this turned out to be a lot more changes than I anticipated. So this pretty much makes JaxPolySlab dilation and sidewall operations traceable?

tidy3d/plugins/adjoint/components/geometry.py
yaugenst: I think I thought that it was required at some point because it's used in the polygon splitting [here](https://github.com/flexcompute/tidy3d/blob/06472617a646e6379b55c51b8afa2e626f36f9f3/tidy3d/plugins/adjoint/components/geometry.py#L850), although the PR would probably work fine without it. But this way, the `JaxComplexPolySlab` supports differentiating w.r.t. `slab_bounds` out of the box (tested) and the `slab_bounds` PR only needs to add the VJP.

yaugenst-flex:
> Wow, this is looking really good! Thanks @yaugenst-flex . I think this turned out to be a lot more changes than I anticipated. So this pretty much makes JaxPolySlab dilation and sidewall operations traceable?

Well, at least those that are used in the polyslab splitting routine, yeah. Although `check_grads` still fails sometimes, so there is _something_ suspicious going on that I haven't quite figured out yet..

tests/test_plugins/test_adjoint.py
```diff
@@ -1913,6 +1930,10 @@ def test_sidewall_angle_validator(log_capture, sidewall_angle, log_expected):  #
 
     jax_polyslab1 = JaxPolySlab(axis=POLYSLAB_AXIS, vertices=VERTICES, slab_bounds=(-1, 1))
 
+    msg = "'JaxPolySlab' does not yet perform the full adjoint gradient treatment "
+    for entry in td.log._static_cache:
```
tylerflex, line 1934: we have an `assert_log_level` and `AssertLogLevel` (context manager version) in `tests/utils`. But this is probably fine.

tylerflex (approved):

PR #1699, opened 2024-05-12
fix smatrix component task_ids to/from file 

tylerflex:
No changelog needed I think because the PR that broke this hasn't been released.

 I refactored Batch and Job recently #1630 . Before, the Job had task_id as a field that was created in a validator by uploading the task. Similarly, Batch.jobs was a field that was created in a validator by uploading all of the tasks. This meant that these fields were saved to file automatically. In the new refactor, these fields are just properties that are created when the Job/Batch is .upload() ed for the first time. So they aren’t saved to file natively, I needed to patch that into the .json() methods. Turned out I had done it so that a Batch worked properly when saved to file, but I needed to handle a component modeler too.

### DIFF
--- a/tidy3d/plugins/smatrix/component_modelers/base.py
+++ b/tidy3d/plugins/smatrix/component_modelers/base.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import Tuple, Dict, Union
 import os
 from abc import ABC, abstractmethod
+import json
 
 import pydantic.v1 as pd
 import numpy as np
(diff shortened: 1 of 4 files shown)


tylerflex:
Tested with `Design.ipynb`, `Smatrix.ipynb` and `WebAPI.ipynb` and all run fine.

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -109,12 +119,30 @@ def _task_name(port: Port, mode_index: int = None) -> str:
     def sim_dict(self) -> Dict[str, Simulation]:
         """Generate all the :class:`Simulation` objects for the S matrix calculation."""
 
+    def json(self, **kwargs):
+        """Save ``Job`` to dictionary. Add the `task_id` if it has been cached."""
```
dmarek-flex: Check that this docstring is correct.

dmarek-flex (approved):
Thanks for fixing this!

tidy3d/plugins/smatrix/component_modelers/base.py
tylerflex: Thanks, fixed it

PR #1700, opened 2024-05-13
fix xarray data access in SpatialDataArray reflection

dbochkov-flexcompute:
thanks for spotting one too many xarray -> numpy data views. It seems like for newer versions of xarray (2024.3.0) this somehow was ok, but not for older ones (like 2023.2.0)


tylerflex (approved):

tylerflex:
@dbochkov-flexcompute No problem. can you add a fixed in changelog? and then I'll merge this.

tylerflex:
actually I think we dont need changelog if this was only for old versions of xarray. I'll just go ahead and merge it

tylerflex (approved):

PR #1707, opened 2024-05-15
fix ruff tests since last check

tylerflex:
since #1681


tests/test_plugins/test_component_modeler.py
```diff
@@ -402,7 +390,7 @@ def test_import_smatrix_smatrix():
 
 def test_to_from_file_batch(monkeypatch, tmp_path):
     modeler = make_component_modeler(path_dir=str(tmp_path))
-    s_matrix = run_component_modeler(monkeypatch, modeler)
+    _ = run_component_modeler(monkeypatch, modeler)
```
yaugenst-flex, line 393: Maybe no assignment at all?

tests/test_plugins/test_invdes.py
```diff
@@ -72,10 +74,10 @@ def test_region_params():
 
     design_region = make_design_region()
 
-    PARAMS_0 = np.random.random(design_region.params_shape)
-    PARAMS_0 = design_region.params_random
-    PARAMS_0 = design_region.params_ones
-    PARAMS_0 = design_region.params_zeros
+    _ = np.random.random(design_region.params_shape)
```
yaugenst-flex, line 77: Same here, since it's just testing whether the properties error.

yaugenst-flex (commented):
Nice thanks. Guess the only question is whether things that are not used should be assigned at all. I went with no assignment in my last fixes, here it's `_ =`.

tests/test_plugins/test_component_modeler.py
tylerflex: hm, doesn't really bother me. If we want to extend the tests later, it's kind of nice to know that *something* is returned there. In fact, I dont mind even leaving it as a variable..

yaugenst-flex: OK let's do it like this then. I think we can merge this now and figure out a linting config that makes sense for the tests in #1680

Issue #1628, opened 2024-04-20
Multilayer waveguide plugin


smgitrep:
Tidy3d has strip and rib waveguide plugins that can be used with the modesolver, but are somewhat limited in scope.

Could these plugins be modified to include air and substrate media above and below the cladding and box, respectively?

A plugin that supports multilayer waveguides with the ability to overlay the mode plofile with the layer stack including lines indicating the interfaces and sidewalls would be great too.

momchil-flex:
Since Lucas already has a lot on his plate, maybe extending our waveguide plugin is a good starter project for one of the new people joining us soon? I mean, I know that there are some projects already planned for them, but this should be not too hard to do and should also help familiarize with our API.

tylerflex:
I have some other projects in mind for the people joining in the near future (Yannick, Matt) that I think will target more directly what they'll be working on. But feel free to reassign to someone else.

lucas-flexcompute:
If this is not high priority, I can implement it relatively quickly probably by next week.

smgitrep:
Just wondering:
1) when the geometry edges over field plots will be released
2) if the multilayer waveguide option will be implemented

lucas-flexcompute:
Hi @smgitrep, both features should be out in our next release.

Issue #1599, opened 2024-04-10
improve mode_solver


joamatab:
At the moment the mode solver `RectangularDielectric` in tidy3d is somewhat to strip/rib

How can we use tidy3d mode solver to ?

- [ ] solve waveguides with varying refractive index
- [ ] solve multi-layer waveguides
- [ ] compute mode leakage to substrate
- [ ] plot lines corresponding to the interfaces between layers and sidewalls, so the mode profile is overlaid with the layer stack and geometry

momchil-flex:
Related to #1628 

Generally all of those are possible now but not with the `RectangularDielectric` plugin. Not sure how much bandwidth Lucas will have but he's promised to implement some of these in the plugin.

lucas-flexcompute:
The only request not covered by the PRs is mode leakage calculation, because that seems to be problem-dependent. I don't think we can provide a general and **robust** solution covering all cases for layered media…

PR #1711, opened 2024-05-21
Add support for layered cladding in waveguide plugin (#1628, #1599)

lucas-flexcompute:
Add support for layered cladding above and below waveguide core.

### DIFF
--- a/tidy3d/plugins/waveguide/rectangular_dielectric.py
+++ b/tidy3d/plugins/waveguide/rectangular_dielectric.py
@@ -1,6 +1,7 @@
 """Rectangular dielectric waveguide utilities."""
 
-from typing import List, Any
+from typing import List, Any, Union
+from typing_extensions import Annotated
 
 import numpy
 from matplotlib import pyplot
@@ -19,13 +20,15 @@ from ...components.viz import add_ax_if_none
 
 from ...components.source import ModeSource, GaussianPulse
 from ...components.structure import Structure
-from ...components.types import ArrayFloat1D, Ax, Axis, Coordinate, Literal, Size1D, Union
+from ...components.types import ArrayFloat1D, Ax, Axis, Coordinate, Literal, Size1D
 from ...components.types import TYPE_TAG_STR
 from ...constants import C_0, inf, MICROMETER, RADIAN
 from ...exceptions import Tidy3dError, ValidationError
 from ...log import log
 
 from ..mode.mode_solver import ModeSolver
+
+AnnotatedMedium = Annotated[MediumType, pydantic.Field(discriminator=TYPE_TAG_STR)]
 
 
 class RectangularDielectric(Tidy3dBaseModel):
@@ -49,7 +52,7 @@ class RectangularDielectric(Tidy3dBaseModel):
     core_width: Union[Size1D, ArrayFloat1D] = pydantic.Field(
         ...,
         title="Core width",
-        description="Core width at the top of the waveguide.  If set to an array, defines "
+        description="Core width at the top of the waveguide. If set to an array, defines "
         "the widths of adjacent waveguides.",
         units=MICROMETER,
     )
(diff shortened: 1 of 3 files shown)


tidy3d/plugins/waveguide/rectangular_dielectric.py
```diff
@@ -68,18 +71,18 @@ class RectangularDielectric(Tidy3dBaseModel):
         discriminator=TYPE_TAG_STR,
     )
 
-    clad_medium: MediumType = pydantic.Field(
+    clad_medium: Union[AnnotatedMedium, List[AnnotatedMedium]] = pydantic.Field(
```
tylerflex: The convention we've been using is to use tuples so that things are enforced to be immutable. but I dont know if we need to be as strict about this in plugins (and maybe generally it would have been ok to keep as lists as long as we don't internally mutate things)

tidy3d/plugins/waveguide/rectangular_dielectric.py
```diff
@@ -89,17 +92,19 @@ class RectangularDielectric(Tidy3dBaseModel):
         units=MICROMETER,
     )
 
-    clad_thickness: Size1D = pydantic.Field(
+    clad_thickness: Union[Size1D, ArrayFloat1D] = pydantic.Field(
```
tylerflex, line 98: might need to validate that the array is all non-negative as I'm not sure it will check this if an array is supplied.

tylerflex: Not sure about performance, and if it becomes annoying in the code, but could also just make it a `Tuple[Size1D, ...]`

tidy3d/plugins/waveguide/rectangular_dielectric.py
```diff
@@ -210,41 +215,51 @@ def _set_array(cls, val):
             raise ValidationError("Values may not be negative.")
         return result
 
+    @pydantic.validator("clad_medium", always=True)
+    def _set_medium_list(cls, val):
```
tylerflex: might be better as a `@property` because this could be confusing if the user supplies a single medium. Also could break existing scripts.
```py
def _medium_list(self) -> list[MediumType]:
    if isinstance(val, MediumType):
        return [self.clad_medium]
    return self.clad_medium
````

File: tidy3d/plugins/waveguide/rectangular_dielectric.py [hunk 4]
```diff
@@ -210,41 +215,51 @@ def _set_array(cls, val):
...
         """Set default cladding thickness based on the max wavelength in the cladding medium."""
         if val is None:
             wavelength = values["wavelength"]
-            medium = values["clad_medium"]
+            medium = values["clad_medium"][0]
             n = numpy.array([medium.nk_model(f)[0] for f in C_0 / wavelength])
             lda = wavelength / n
-            return 1.5 * lda.max()
-        return val
+            return numpy.array([1.5 * lda.max()])
```
tylerflex: consider setting 1.5 as a constant somewhere, or a field, if it would make sense to.

tylerflex (changes requested):
Thanks @lucas-flexcompute. Just a few suggestions. I would suggest adding a few tests though to make sure that the array-like mediums work as expected.

tidy3d/plugins/waveguide/rectangular_dielectric.py
lucas-flexcompute: Tuple is fine. I had it as tuple before and changed to list because I was having trouble with the `AnnotatedMedium` and wanted to check. Then I forgot to change back.

tidy3d/plugins/waveguide/rectangular_dielectric.py
lucas-flexcompute: I've added a few tests for this and it seems to work as intended.

tidy3d/plugins/waveguide/rectangular_dielectric.py
lucas-flexcompute: The breaking problem would also happen for `clad_thickness` and `box_thickness` too. I'll set them up as properties as well.

tylerflex (approved):

tylerflex:
@lucas-flexcompute the tests were failing from this PR last I checked. I should have mentioned it instead of approving right away. Seems like these might have been merged into pre/2.7. 

When you get a chance, we will want to fix these up.

PR #1712, opened 2024-05-23
Fix missing check in validator

lucas-flexcompute:


tylerflex:
I wonder if the Union fields need
`discriminator=TYPE_TAG_STR`

?

lucas-flexcompute:
I'm not sure. It seems to happen only on Windows.

lucas-flexcompute:
> I wonder if the Union fields need `discriminator=TYPE_TAG_STR`

I think the unions cannot have a discriminator, because a list won't have a `type` attribute. That's the reason why I created the annotated medium type, so it could be used within the list.

lucas-flexcompute:
The only test that is failing is the submodule. I've already updated it locally, but it doesn't seem to make any difference. @daquinteroflex @tylerflex any idea how to fix?

tylerflex:
I think possibly your first Multilayer waveguide PR was failing submodules too, so pre/2.7 might be already out of date. Let me push a submodule update to pre/2.7 and see if it fixes things.

tylerflex:
shoot, that didnt fix it maybe @daquinteroflex can take a look when he has time

tylerflex:
before merging (once the submodule test is fixed) just a reminder to squash all of the commits into one if you dont mind.

tylerflex (approved):
Approve pending test passing and squashed commits.

Also curious what the issue was?

lucas-flexcompute:
> Approve pending test passing and squashed commits.
> 
> Also curious what the issue was?

The problem was with `isinstance(..., MediumType)`. For some reason that doesn't work on Windows. So I replaced all checks with `isinstance(..., "some concrete type")`. I've also changed all instance validators to being root validators because I think that was also triggering the same issue internally.

daquinteroflex:
Hello! Sorry about that, that's weird. I think I've fixed it now and for reference these are the commands I run to update it:

```raw
 1004  git checkout lucas/waveguide_plugin/test_fix
 1005  cd docs/
 1006  cd notebooks/
 1007  git pull origin pre/2.7 
 1008  cd ..
 1009  git add .
 1010  git commit -am ":wrench: Update submodule?"
 1011  git log
 1012  git push origin lucas/waveguide_plugin/test_fix 
 1013  history
```

PR #1715, opened 2024-05-23
Validate Medium2D components to be either both PEC or non-PEC

weiliangjin2021:
In principle, we should warn in AnisotropicMedium if the 3 components are not simultaneously PEC or non-PEC. This is because the solver will just convert the anisotropic medium to PECMedium if any of its component is PEC. But since hopefully it is unlikely to be used directly in AnistropicMedium, let's just add a validator in Medium2D.

### DIFF
--- a/tests/test_components/test_medium.py
+++ b/tests/test_components/test_medium.py
@@ -433,6 +433,9 @@ def test_medium2d(log_capture):  # noqa: F811
     plt.close()
     assert_log_level(log_capture, "WARNING")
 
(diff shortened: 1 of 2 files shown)


tylerflex (approved):
some issue with the submodule checks see #1712 

once it's resolved feel free to merge.

tylerflex:
@weiliangjin2021 submodule needs updating I think and then we can merge

dmarek-flex (approved):

tylerflex:
Just make sure to squash commits too, thanks!

PR #1719, opened 2024-05-24
Enable snapping_points in grid generation

weiliangjin2021:
TODO:
- [x] Add tests and update CHANGELOG
- [x] Stress tests and check edge cases

## Feature Description
Enforce placing grid points at certain locations. This is helpful in improving simulation accuracy by enforcing grid points at certain geometry singularities. Around the snapping point, the maximal step size is unmodified.

**Implementation**:
- Start from intervals returned by `parse_structures`. If any interval contains a snapping point, split the interval. The maximal grid step size will be the same at the two splitted intervals.
- We don't split if the insertion results in very small interval. Currently I set the threshold as `np.amin(max_dl_list) * 0.5`, but it's an open question and worth more investigation.

## Example
Here is an example comparing the grids after the insertion of snapping points:
```
medium = td.Medium(permittivity = 16)

box = td.Structure(
    geometry=td.Box(size=[0.1,]*3, center = (-0.12, 0, 0)), 
    medium=medium,
)

point1 = [0.123, 0.145, 0]
point2 = [-0.148, 0.1, 0]

point_outside = [1, 0.13, 0]

grid_spec=td.GridSpec.auto(
    min_steps_per_wvl=10,
    wavelength=1,
)

grid_spec_snapping = grid_spec.updated_copy(snapping_points=[point1, point2, point_outside])

sim = td.Simulation(
    size=(0.6,) * 3,
    grid_spec=grid_spec,
    run_time=1e-13,
    structures=[box],
    boundary_spec = td.BoundarySpec.all_sides(boundary=td.PECBoundary()),
)
sim_snapping = sim.updated_copy(grid_spec = grid_spec_snapping)

fig, ax = plt.subplots(1, 2, figsize=(8, 5), tight_layout=True)
sim.plot(z=0, ax=ax[0])
sim.plot_grid(z=0, ax=ax[0], colors='red', linewidth=1)

sim_snapping.plot(z=0, ax=ax[1])
sim_snapping.plot_grid(z=0, ax=ax[1], colors='red', linewidth=1)
```
image


tests/test_components/test_grid_spec.py
tylerflex, line 1: should a test involve some grid specs with fixpoints?

tidy3d/components/grid/grid_spec.py
```diff
@@ -572,6 +585,15 @@ class GridSpec(Tidy3dBaseModel):
         "uses :class:`.AutoGrid`.",
     )
 
+    fixpoints: Tuple[Point, ...] = pd.Field(
```
tylerflex: I'm wondering if these should be `fixed_points` rather than `fixpoints`. Is "fixpoints" commonly used?

tidy3d/components/geometry/primitives.py
```diff
@@ -166,6 +166,19 @@ def _surface_area(self, bounds: Bound) -> float:
         return area
 
 
+class Point(Sphere):
```
tylerflex: Does this need to inherit from `Sphere`? or can we just define fixed points as `tuple[Coordinate, ...]` to keep it simple? `Point` would add another class for users to learn.

tidy3d/components/grid/grid_spec.py
```diff
@@ -423,6 +428,7 @@ def _make_coords_initial(
         wavelength: float,
         symmetry: Symmetry,
         is_periodic: bool,
+        fixpoints: List[Point],
```
tylerflex: might be safest to convert to list internally and pass tuples?

tidy3d/components/geometry/primitives.py
tylerflex: If we keep `Point`, be sure to add to the API reference `api.rst`. but I'd probably rather use `Coordinate` tuples

tylerflex (commented):
Thanks @weiliangjin2021 see my comments.

tidy3d/components/grid/grid_spec.py
weiliangjin2021: It's used in one software. In fact, I'm not so in favor of this name. How about other names such as `snapping_points`, `enforced_grid_points`? @tylerflex @QimingFlex

tidy3d/components/geometry/primitives.py
weiliangjin2021: Good point, I'll change to use Coordinate

tidy3d/components/grid/grid_spec.py
weiliangjin2021: Rename to `snapping_points` now.

tylerflex: Sounds good to me

QimingFlex (commented):
Thanks, @weiliangjin2021 ! Finally, the snapping points should replace the diagonally placed meshoverride boxes to handle edges. Regarding the open question about the threshold, from my understanding, it is currently determined by the largest allowable step for each structure interval. Adding two more parameters—one for the smallest allowable step and another for the enforced step at the point—might be useful. Perhaps these two parameters could be combined into one.

weiliangjin2021:
> it is currently determined by the largest allowable step for each structure interval.

Not really. Along a direction, the simulation domain is divided into multiple intervals. Each interval has its own maximal allowed step size. The threshold value is half of the minimal of those steps sizes.

> Adding two more parameters—one for the smallest allowable step and another for the enforced step at the point

Could you explain a bit more?

QimingFlex:
I see. I misunderstood it as the smallest allowed step. So currently, the minimum step is given by 'min_steps_per_wvl' and the largest time step is given by these intervals. Is this correct?

weiliangjin2021:
> I see. I misunderstood it as the smallest allowed step. So currently, the minimum step is given by 'min_steps_per_wvl' and the largest time step is given by these intervals. Is this correct?

Is there a typo in your sentence (how does time step enter here)? 'min_steps_per_wvl' determines the maximal allowed step size in each interval.

QimingFlex:
Sorry it is a typo, ‘space’ step.

tests/test_components/test_grid_spec.py
weiliangjin2021: done

weiliangjin2021:
> Sorry it is a typo, ‘space’ step.

Well, then I still don't get your question. I guess it should be clear that the threshold is `np.amin(max_dl_list) * 0.5`?

tidy3d/components/grid/grid_spec.py
```diff
@@ -31,6 +31,7 @@ def make_coords(
         periodic: bool,
         wavelength: pd.PositiveFloat,
         num_pml_layers: Tuple[pd.NonNegativeInt, pd.NonNegativeInt],
+        snapping_points: Tuple[Coordinate, ...],
```
tylerflex, line 34: consider making this optional `= ()` if it would make sense to minimize changes needed, same with `_make_coords_initial` and the other methods

tylerflex (approved):

tidy3d/components/grid/grid_spec.py
weiliangjin2021: I actually prefer not using optional arguments here: 1) this part is internal, so it doesn't affect user API; 2) when using optional arguments, one often forgets updating codes to pass the new arguments, which can sometimes take a while to debug.

tylerflex: sounds fine to me 👍

QimingFlex:
> > Sorry it is a typo, ‘space’ step.
> 
> Well, then I still don't get your question. I guess it should be clear that the threshold is `np.amin(max_dl_list) * 0.5`?

I tested with several examples and understand the threshold. The parameter looks good to me!

QimingFlex (approved):

tidy3d/components/grid/mesher.py
```diff
@@ -63,6 +73,63 @@ class GradedMesher(Mesher):
...
+                continue
+            # search insertion location
+            ind = np.searchsorted(interval_coords, new_coord, side="left")
+            # Skip snapping_points if the distance to the existing interval boundarires are
+            # smaller than half of the minimal maximal step size
+            if abs(new_coord - interval_coords[ind]) < min_step:
+                continue
+            if abs(new_coord - interval_coords[ind - 1]) < min_step:
+                continue
+            # insertion
```
momchil-flex, line 128: The comment says "half the minimal max step size" but you check `min_step` not `min_step / 2`. I think the code is correct, not the comment.

Another approach could be to first add the coordinates and only then call filter min steps. Currently, the [coordinates from the structures are first filtered](https://github.com/flexcompute/tidy3d/blob/392f4bb98352f06b9dd54895152d7074ff242c21/tidy3d/components/grid/mesher.py#L305-L307), and then the snapping coords are added here. This would effectively always prioritize structure coords over snapping coords, when those are close. Not sure what's better?

momchil-flex (approved):
Looks good, main comment from usability perspective is that in some cases you might want to set the point along e.g. one of the axes only, but I think it's fine, we don't need to complicate things so much for now.

tidy3d/components/grid/mesher.py
weiliangjin2021: yeah, I tend to favor prioritizing structure to avoid shifting interval boundaries when adding a snapping point very close to the structure boundary. In that case, if a snapping point is added inappropriately, at least it doesn't mess up numeric dispersion accuracy.

PR #1721, opened 2024-05-28
Autograd utils

yaugenst-flex:
This has everything (for now).

### DIFF
--- a/tidy3d/plugins/autograd/functions.py
+++ b/tidy3d/plugins/autograd/functions.py
@@ -2,7 +2,7 @@ from typing import Iterable, Union, Tuple, Literal, List
 
 import autograd.numpy as np
 from autograd.scipy.signal import convolve as convolve_ag
-from .types import _pad_modes
+from tidy3d.plugins.autograd.types import PaddingType
 
 
 def _make_slices(rule: Union[int, slice], ndim: int, axis: int) -> Tuple[slice, ...]:
@@ -161,7 +161,7 @@ def pad(
     array: np.ndarray,
     pad_width: Union[int, Tuple[int, int]],
     *,
-    mode: _pad_modes = "constant",
+    mode: PaddingType = "constant",
     axis: Union[int, Iterable[int], None] = None,
     constant_value: float = 0.0,
 ) -> np.ndarray:
@@ -259,7 +259,7 @@ def convolve(
     array: np.ndarray,
     kernel: np.ndarray,
     *,
-    padding: _pad_modes = "constant",
+    padding: PaddingType = "constant",
     axes: Union[Tuple[List[int], List[int]], None] = None,
     mode: Literal["full", "valid", "same"] = "same",
 ) -> np.ndarray:
@@ -312,7 +312,7 @@ def grey_dilation(
     size: Union[Union[int, Tuple[int, int]], None] = None,
     structure: Union[np.ndarray, None] = None,
     *,
-    mode: _pad_modes = "reflect",
+    mode: PaddingType = "reflect",
     maxval: float = 1e4,
 ) -> np.ndarray:
     """Perform grey dilation on an array.
@@ -367,7 +367,7 @@ def grey_erosion(
     size: Union[Union[int, Tuple[int, int]], None] = None,
     structure: Union[np.ndarray, None] = None,
     *,
-    mode: _pad_modes = "reflect",
+    mode: PaddingType = "reflect",
     maxval: float = 1e4,
 ) -> np.ndarray:
     """Perform grey erosion on an array.
@@ -422,7 +422,7 @@ def grey_opening(
     size: Union[Union[int, Tuple[int, int]], None] = None,
     structure: Union[np.ndarray, None] = None,
     *,
-    mode: _pad_modes = "reflect",
+    mode: PaddingType = "reflect",
     maxval: float = 1e4,
 ) -> np.ndarray:
     """Perform grey opening on an array.
@@ -456,7 +456,7 @@ def grey_closing(
     size: Union[Union[int, Tuple[int, int]], None] = None,
     structure: Union[np.ndarray, None] = None,
     *,
-    mode: _pad_modes = "reflect",
+    mode: PaddingType = "reflect",
     maxval: float = 1e4,
 ) -> np.ndarray:
     """Perform grey closing on an array.
@@ -490,7 +490,7 @@ def morphological_gradient(
     size: Union[Union[int, Tuple[int, int]], None] = None,
     structure: Union[np.ndarray, None] = None,
     *,
-    mode: _pad_modes = "reflect",
+    mode: PaddingType = "reflect",
     maxval: float = 1e4,
 ) -> np.ndarray:
     """Compute the morphological gradient of an array.
@@ -524,7 +524,7 @@ def morphological_gradient_internal(
     size: Union[Union[int, Tuple[int, int]], None] = None,
     structure: Union[np.ndarray, None] = None,
     *,
-    mode: _pad_modes = "reflect",
+    mode: PaddingType = "reflect",
     maxval: float = 1e4,
 ) -> np.ndarray:
     """Compute the internal morphological gradient of an array.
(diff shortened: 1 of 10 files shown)


tidy3d/plugins/autograd/functions.py
```diff
@@ -161,7 +162,7 @@ def pad(
     array: np.ndarray,
     pad_width: Union[int, Tuple[int, int]],
     *,
-    mode: _pad_modes = "constant",
+    mode: PaddingType = "constant",
```
tylerflex, line 165: maybe we should start using Enums for these kinds of things. There's an open issue for it but nobody's found a great solution yet.

https://github.com/flexcompute/tidy3d/issues/564

tidy3d/plugins/autograd/functions.py
```diff
@@ -581,3 +582,57 @@ def morphological_gradient_external(
...
+
+    Parameters
+    ----------
+    array : np.ndarray
+        The input array to be rescaled.
+    out_min : float
+        The minimum value of the output range.
+    out_max : float
+        The maximum value of the output range.
+    in_min : float, optional
```
tylerflex, line 601: our convention (which might not be standard) is to use `: type = default value` for optional types, eg `float = 0.0`.

tidy3d/plugins/autograd/functions.py
```diff
@@ -581,3 +582,57 @@ def morphological_gradient_external(
...
+    -------
+    np.ndarray
+        The rescaled array.
+    """
+    scaled = (array - in_min) / (in_max - in_min)
+    return scaled * (out_max - out_min) + out_min
+
+
+def threshold(
+    array: np.ndarray, vmin: float = 0.0, vmax: float = 1.0, level: Union[float, None] = None
```
tylerflex, line 616: I think while this is more explicit, just `float = None` would have been ok too for us, just FYI

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -0,0 +1,66 @@
...
+
+from ..functions import convolve
+from ..types import KernelType, PaddingType
+from ..utilities import make_kernel
+
+
+def make_filter(
+    filter_type: KernelType,
+    size: Union[int, Tuple[int, ...]],
+    *,
```
tylerflex, line 14: what is this?

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -0,0 +1,66 @@
...
+    normalize: bool = True,
+    padding: PaddingType = "reflect",
+):
+    """Create a filter function based on the specified kernel type and size.
+
+    Parameters
+    ----------
+    filter_type : KernelType
+        The type of kernel to create (`circular` or `conic`).
```
tylerflex, line 23: circular filter kind of sucks, so wouldn't be opposed to just removing it..

tylerflex: also what about gaussian filter? is that separate?

tidy3d/plugins/autograd/invdes/misc.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Returns
+    -------
+    float
+        The calculated grey indicator.
+    """
+    return np.mean(4 * array * (1 - array))
+
+
+def get_kernel_size_px(radius: float, dl: float) -> float:
```
tylerflex, line 25: return of type `int`?

tidy3d/plugins/autograd/invdes/penalties.py
```diff
@@ -0,0 +1,190 @@
...
+    -----
+    The curvature can be positive or negative, indicating the direction of curvature.
+    The radius of curvature is defined as 1 / |κ|, where κ is the curvature.
+    """
+    num = dp[0] * ddp[1] - dp[1] * ddp[0]
+    den = np.power(dp[0] ** 2 + dp[1] ** 2, 1.5)
+    return num / den
+
+
+def bezier_with_grads(
```
tylerflex, line 96: we can consider splitting the penalties into topology-based and shape-based modules?

tylerflex (approved):
Some comments, but we can change these later. Overall looks very good. I think the main usability issue I am uncertain about is whether it is tricker to deal with pixels, but I'll try this out in the notebooks and have a better opinion later.
Thanks @yaugenst-flex

tidy3d/plugins/autograd/functions.py
yaugenst-flex: Agreed. So something like
```python
class PaddingType(StrEnum):
    CONIC = "conic"

def f(p: PaddingType):
    ...

f(PaddingType.CONIC)
```

tidy3d/plugins/autograd/functions.py
yaugenst-flex: Ah ok. I assumed this is numpydoc style, where the convention is to either do `x : type, optional` or `x : type, default=value`, where `optional` is preferred if the default would not be used as a value, like in instances where the default is `None`. So according to that I actually did it wrong :joy:
Anyways, I'll stick to our convention then.

tidy3d/plugins/autograd/functions.py
yaugenst-flex: Ideally we jump to Python >= 3.10 and can write `float | None` instead :grin:

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: Keyword-only argument, i.e. anything behind `*,` needs to be passed as a keyword argument. This and positional-only arguments `(a, /, b, c, *, d=d, e=e)` should probably be used more...
Here's a pretty nice writeup why: https://github.com/pola-rs/polars/issues/6993

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: Yeah the Gaussian filter is implemented differently because it uses `scipy.ndimage.gaussian_filter` directly. I should wrap it though. I'm actually for keeping the circular filter and eventually adding more, like mean filters, median filters, etc... the whole bunch. They might not be very good for the parametrization, but can be useful for other things like implementing penalties / different fabrication constraints.

tidy3d/plugins/autograd/invdes/misc.py
yaugenst-flex: oopsie

tidy3d/plugins/autograd/invdes/penalties.py
yaugenst-flex: Yeah that makes sense, just to keep the size of these modules manageable over time. I think everything should still be available from the `tidy3d.plugins.autograd.invdes` namespace though (you might want to mix shape-based and topology).

tidy3d/plugins/autograd/invdes/filters.py
tylerflex: I bug people a lot to use keyword arguments as much as possible. Didn't know you could enforce it like this, interesting. 🤔

tidy3d/plugins/autograd/functions.py
tylerflex: Maybe in late 2025 when python 3.9 is no longer supported? Unless there's a good argument for only supporting newer versions of python? I guess the `|` syntax would be quite nice 🤔 but probably will make some users unhappy to not support 3.8 or 3.9

tidy3d/plugins/autograd/functions.py
tylerflex: yea, we kind of broke the convention a while ago, there's a function that parses these in the base model for our docs API and so it expects the new convention. I'm not totally opposed to just sticking with standard numpydoc  but someone's gotta go through and do the conversions..

tylerflex: https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/base.py#L875-L939

tylerflex: oh wait, ignore that, it's for pydantic models.. I'm actually not sure anymore how these method docstrings get rendered for our docs..

tidy3d/plugins/autograd/functions.py
yaugenst: Yeah it's on the "at some point" list... I just really miss `|` for unions and `match`...

PR #1722, opened 2024-05-28
Updated comment section for CustomFieldSource to explain how to norma…

prashkh:
…lize the results properly


tomflexcompute (approved):
Thanks Prash. There seems to be an extra space between `grid` and `is` in line 742.

tidy3d/components/source.py
```diff
@@ -737,6 +737,12 @@ class CustomFieldSource(FieldSource, PlanarSource):
         the directionality of the source will likely not be perfect, even if both the ``E`` and ``H`` fields are
         provided. An empty normalizing run may be needed to accurately normalize results.
 
+        To create this empty simulation it is recommended that users create a simulation with no structures but just a flux
+        monitor (``tidy3D.FluxMonitor``) next to the custom source, ensuring that the flux monitor is at least one grid cell
+        away from the source. Moreover, for accurate normalization, users must ensure that the same grid  is used to run
+        the original simulation as well as the empty simulation. The total flux calculated at the flux monitor of the empty
```
momchil-flex, line 763: I think adding a comment about how to achieve the same grid when using auto grid will also be useful. Namely what we did yesterday: define the same structures you used to have in your original Simulation in the GridSpec `override_structures` of the empty sim.

tylerflex (approved):

tidy3d/components/source.py
prashkh: Thanks for the feedback! Added a short explanation on how to create empty simulation with the grid from original simulation.

momchil-flex (approved):
Looks good, could you just squash the three commits into 1 and then force push?

prashkh:
> Looks good, could you just squash the three commits into 1 and then force push?

@momchil-flex  I squashed the commits and force pushed it. Let me know if it is all good.

momchil-flex:
For some reason it is still 3 commits @prashkh - but I'll just do it myself.

prashkh:
@momchil-flex thanks so much! not sure what happened but I thought I squashed and force pushed it

PR #1724, opened 2024-05-29
Introduce the DirectivityMonitor and DirectivityData classes

QimingFlex:
Added the 'DirectivityData' class, introduced a new 'gain' property, and overwrote the 'normalize' function to include flux normalization.

### DIFF
--- a/tidy3d/components/data/data_array.py
+++ b/tidy3d/components/data/data_array.py
@@ -700,6 +700,27 @@ class FieldProjectionAngleDataArray(DataArray):
     _data_attrs = {"long_name": "radiation vectors"}
 
 
+class DirectivityDataArray(DataArray):
+    """Directivity in the frequency domain as a function of angles theta and phi. Directivity is a
+    dimensionless quantity defined as the ratio of the radiation intensity in a given direction
+    to the average radiation intensity over all directions.
+
+    Example
+    -------
+    >>> f = np.linspace(1e14, 2e14, 10)
+    >>> r = np.atleast_1d(5)
+    >>> theta = np.linspace(0, np.pi, 10)
+    >>> phi = np.linspace(0, 2*np.pi, 20)
+    >>> coords = dict(r=r, theta=theta, phi=phi, f=f)
+    >>> values = np.random.random((len(r), len(theta), len(phi), len(f)))
+    >>> data = DirectivityDataArray(values, coords=coords)
+    """
+
+    __slots__ = ()
+    _dims = ("r", "theta", "phi", "f")
+    _data_attrs = {"long_name": "radiation intensity"}
+
+
 class FieldProjectionCartesianDataArray(DataArray):
     """Far fields in frequency domain as a function of local x and y coordinates.
 
@@ -997,6 +1018,7 @@ DATA_ARRAY_TYPES = [
     FieldProjectionCartesianDataArray,
     FieldProjectionKSpaceDataArray,
     DiffractionDataArray,
+    DirectivityDataArray,
     FreqModeDataArray,
     FreqDataArray,
     TimeDataArray,
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -21,6 +21,7 @@ from ..grid.grid import Coords, Grid
 from ..medium import Medium, MediumType
 from ..monitor import (
     DiffractionMonitor,
+    DirectivityMonitor,
     FieldMonitor,
     FieldProjectionAngleMonitor,
     FieldProjectionCartesianMonitor,
@@ -59,6 +60,7 @@ from ..validators import enforce_monitor_fields_present, required_if_symmetry_pr
 from .data_array import (
     DataArray,
     DiffractionDataArray,
+    DirectivityDataArray,
     EMEFreqModeDataArray,
     FieldProjectionAngleDataArray,
     FieldProjectionCartesianDataArray,
@@ -1972,6 +1974,34 @@ class FluxTimeData(MonitorData):
     )
 
 
+class DirectivityData(MonitorData):
+    """
+    Data associated with a :class:`.DirectivityMonitor`.
+
+    Example
+    -------
+    >>> from tidy3d import DirectivityDataArray
+    >>> f = np.linspace(1e14, 2e14, 10)
+    >>> r = np.atleast_1d(5)
+    >>> theta = np.linspace(0, np.pi, 10)
+    >>> phi = np.linspace(0, 2*np.pi, 20)
+    >>> coords = dict(r=r, theta=theta, phi=phi, f=f)
+    >>> scalar_field = DirectivityDataArray(farfield_values, coords=coords)
+    >>> monitor = DirectivityMonitor(center=(1,2,3), size=(2,2,2), freqs=f, name='n2f_monitor', phi=phi, theta=theta)
+    >>> data = DirectivityData(monitor=monitor,Directivity=scalar_field)
+    """
+
+    monitor: DirectivityMonitor = pd.Field(
+        ...,
+        title="Directivity monitor",
+        description="Monitor describing the angle-based projection grid on which to measure directivity data.",
+    )
+
+    directivity: DirectivityDataArray = pd.Field(
+        ..., title="Directivity", description="Directivity with an angle-based projection grid."
+    )
+
+
 ProjFieldType = Union[
     FieldProjectionAngleDataArray,
     FieldProjectionCartesianDataArray,
@@ -2957,6 +2987,7 @@ MonitorDataTypes = (
     FieldProjectionCartesianData,
     FieldProjectionAngleData,
     DiffractionData,
+    DirectivityData,
 )
 
 MonitorDataType = Union[MonitorDataTypes]
--- a/tidy3d/components/monitor.py
+++ b/tidy3d/components/monitor.py
@@ -1087,6 +1087,21 @@ class FieldProjectionAngleMonitor(AbstractFieldProjectionMonitor):
         return BYTES_COMPLEX * len(self.theta) * len(self.phi) * len(self.freqs) * 6
 
 
+class DirectivityMonitor(FieldProjectionAngleMonitor, FluxMonitor):
+    """:class:`Monitor` that records the directivity of antennas in the frequency domain
+    at specified observation angles.
+    """
+
+    def storage_size(self, num_cells: int, tmesh: ArrayFloat1D) -> int:
+        """Size of monitor storage given the number of points after discretization."""
+        # stores 1 complex number per pair of angles, per frequency,
+        # for Er, Etheta, Ephi, Hr, Htheta, and Hphi (6 components)
+        # stores 1 real number per frequency for flux
+        return BYTES_COMPLEX * len(self.theta) * len(self.phi) * len(
+            self.freqs
+        ) * 6 + BYTES_REAL * len(self.freqs)
+
+
 class FieldProjectionCartesianMonitor(AbstractFieldProjectionMonitor):
     """:class:`Monitor` that samples electromagnetic near fields in the frequency domain
     and projects them on a Cartesian observation plane.
@@ -1420,4 +1435,5 @@ MonitorType = Union[
     FieldProjectionCartesianMonitor,
     FieldProjectionKSpaceMonitor,
     DiffractionMonitor,
+    DirectivityMonitor,
 ]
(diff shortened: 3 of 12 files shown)


tidy3d/components/data/monitor_data.py
```diff
@@ -2098,6 +2099,78 @@ def renormalize_fields(self, proj_distance: float) -> FieldProjectionAngleData:
         # compute updated fields and their coordinates
         return self.make_renormalized_data(phase, proj_distance)
 
+class DirectivityData(FieldProjectionAngleData):
```
weiliangjin2021: Expose this to the tidy3d namespace [here](https://github.com/flexcompute/tidy3d/blob/7f69690abd87b9cf6e76f5b5d0e341ad7b47ba05/tidy3d/__init__.py#L69) and [here](https://github.com/flexcompute/tidy3d/blob/7f69690abd87b9cf6e76f5b5d0e341ad7b47ba05/tidy3d/__init__.py#L254)

weiliangjin2021 (commented):
I think the fields in DirectivityData should be the angular power distribution (directivity). It shouldn't contain flux, or unnormalized E/H. The normalization should be performed in the backend already.

tidy3d/components/data/monitor_data.py
```diff
@@ -2098,6 +2099,78 @@ def renormalize_fields(self, proj_distance: float) -> FieldProjectionAngleData:
...
+        description="DirectivityMonitor with an angle-based projection grid.",
+    )
+
+    flux: FluxDataArray = pd.Field(
+        ..., title="Flux", description="Flux values in the frequency-domain."
+    )
+
+
+    @property
+    def gain(self) -> xr.DataArray:
```
weiliangjin2021: It's not really `gain`, right? Gain has a few additional effects to be considered.

weiliangjin2021 (changes requested):
A few other TODOs:

- [x] Edit [CHANGELOG](https://github.com/flexcompute/tidy3d/blob/develop/CHANGELOG.md)
- [x] Add tests to monitor [here](https://github.com/flexcompute/tidy3d/blob/develop/tests/test_components/test_monitor.py) and [monitor data](https://github.com/flexcompute/tidy3d/blob/develop/tests/test_data/test_monitor_data.py)

QimingFlex:
> I think the fields in DirectivityData should be the angular power distribution (directivity). It shouldn't contain flux, or unnormalized E/H. The normalization should be performed in the backend already.

Yes, the normalization is performed in the backend  (https://github.com/flexcompute/tidy3d-core/blob/f11d1618d4d90dc4400f8df4e0e766c42ac1bbdf/tidy3d_backend/run.py#L318). But it seems normalize is a class function of the MonitorData objects. So, initially I think we need to load the unnormalized data from monitor raw into DirectivityData first, and then normalize it.

weiliangjin2021:
> Yes, the normalization is performed in the backend (https://github.com/flexcompute/tidy3d-core/blob/f11d1618d4d90dc4400f8df4e0e766c42ac1bbdf/tidy3d_backend/run.py#L318). But it seems normalize is a class function of the MonitorData objects. So, initially I think we need to load the unnormalized data from monitor raw into DirectivityData first, and then normalize it.

That's a different normalization. For far field projection monitor or flux monitor, that normalization makes the field in the monitor_data relevant. E.g. in the flux case, the normalized flux will be 1 in vacuum. For DirectivityMonitor, the relevant quantity is the directivity: power divided by flux.

tidy3d/components/data/data_array.py
```diff
@@ -693,6 +693,24 @@ class FieldProjectionAngleDataArray(DataArray):
     _data_attrs = {"long_name": "radiation vectors"}
 
 
+class DirectivityDataArray(DataArray):
+    """Far fields in frequency domain as a function of angles theta and phi.
```
weiliangjin2021: "Far fields" --> "Directivity"

tidy3d/components/monitor.py
```diff
@@ -1074,6 +1074,21 @@ def storage_size(self, num_cells: int, tmesh: ArrayFloat1D) -> int:
         return BYTES_COMPLEX * len(self.theta) * len(self.phi) * len(self.freqs) * 6
 
 
+class DirectivityMonitor(FieldProjectionAngleMonitor, FluxMonitor):
+    """:class:`Monitor` that samples electromagnetic near fields in the frequency domain,
+    projects them at given observation angles,  and computes directivity. [TODO]
```
weiliangjin2021: Update description to describe directivity here? and remove [TODO] and extra space before "and"

weiliangjin2021 (commented):
Thanks Qiming, all look nice! A few minor comments below

weiliangjin2021:
And don't forget updating those doc files [here](https://github.com/flexcompute/tidy3d/tree/develop/docs/api):
- monitors.rst
- output_data.rst

You can preview the doc by building the doc page locally with `tidy3d develop build-docs` by following instructions here: https://docs.flexcompute.com/projects/tidy3d/en/latest/development/index.html

tylerflex (changes requested):
Looks good to me, thanks @QimingFlex !

A few comments for better testing:

1. Could you add the new monitor data to the `MONITOR_DATA` and `MONITOR_DATA_SYM` in `test_sim_data.py`? that way we can make sure simulation data works properly with these objects in them.
2. Could you add the `DirectivityMonitor` to the `SIM_FULL` in `tests/utils.py`? That way we can also test the various things like saving and loading to file.

By the way, if you run
```py
 pytest -rA tests --cov-report=html --cov=tidy3d && open htmlcov/index.html
```

you can see what lines of code have been run through by the tests. It's not as important for this PR since most of the functionality is inherited, but for example, could you ensure that `DirectivityMonitor.storage_size` has been called?

Thanks!

weiliangjin2021 (approved):

QimingFlex:
Thanks Tyler!

I've added the new monitor to the test files and testing with it.

> ```python
>  pytest -rA tests --cov-report=html --cov=tidy3d && open htmlcov/index.html
> ```
I encountered an error when running this:
`ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --cov-report=html --cov=tidy3d
  inifile: /home/flexqiming/tidy3d-core/tidy3d_frontend/pytest.ini
  rootdir: /home/flexqiming/tidy3d-core/tidy3d_frontend`
 Could you help me fix this?

tylerflex:
Hi @QimingFlex you need to `pip install pytest-cov` that should do it.

QimingFlex:
Thanks @tylerflex , it works!

Regarding these:

> Could you add the new monitor data to the MONITOR_DATA and MONITOR_DATA_SYM in test_sim_data.py? that way we can make sure simulation data works properly with these objects in them.
    Could you add the DirectivityMonitor to the SIM_FULL in tests/utils.py? That way we can also test the various things like saving and loading to file.

It passed with macos but failed on ubantu with these errors:
` XFAIL tests/test_plugins/test_adjoint.py::test_adjoint_pipeline[True] - Gradients changed due to new input structures in pre/2.7
XFAIL tests/test_plugins/test_adjoint.py::test_adjoint_pipeline[False] - Gradients changed due to new input structures in pre/2.7
FAILED tests/test_plugins/test_design.py::test_sweep[sweep_method1] - tidy3d.exceptions.SetupError: AutoGrid coordinates along axis 0 do not match the simulation domain, indicating an unexpected error in the meshing. Please create a github issue so that the problem can be investigated. In the meantime, switch to uniform or custom grid along 0.`

I explored a bit but have no clue where the problem might be. Do you have any thoughts?

tylerflex:
> tidy3d.exceptions.SetupError: AutoGrid coordinates along axis 0 do not match the simulation domain, indicating an unexpected error in the meshing. Please create a github issue so that the problem can be investigated. In the meantime, switch to uniform or custom grid along 0.

@weiliangjin2021 might know about this one. The other error about adjoint should not fail the test (it's just disabled).

weiliangjin2021:
> > tidy3d.exceptions.SetupError: AutoGrid coordinates along axis 0 do not match the simulation domain, indicating an unexpected error in the meshing. Please create a github issue so that the problem can be investigated. In the meantime, switch to uniform or custom grid along 0.
> 
> @weiliangjin2021 might know about this one. The other error about adjoint should not fail the test (it's just disabled).

This was brought up before, but we didn't continue investigation because it can hardly be reproduced. This seems to be the case here, too?

QimingFlex:
The error happens by: 

> add the DirectivityMonitor to the SIM_FULL in tests/utils.py? That way we can also test the various things like saving and loading to file.

weiliangjin2021:
> The error happens by:
> 
> > add the DirectivityMonitor to the SIM_FULL in tests/utils.py? That way we can also test the various things like saving and loading to file.

I just ran it locally, and all tests pass.

QimingFlex:
> > The error happens by:
> > > add the DirectivityMonitor to the SIM_FULL in tests/utils.py? That way we can also test the various things like saving and loading to file.
> 
> I just ran it locally, and all tests pass.

Can you try again with the latest push? I commented out the DirectivityMonitor in utils.py in the last push, which might be the reason it passed.

tylerflex:
@QimingFlex I think there's some randomness in this test, so from time to time it just fails. You might want to just re-run to see if it resolved itself.

QimingFlex:
> @QimingFlex I think there's some randomness in this test, so from time to time it just fails. You might want to just re-run to see if it resolved itself.

Oh, I see. It just passed!

QimingFlex:
> could you ensure that `DirectivityMonitor.storage_size` has been called?

Confirmed it has been called.

Screenshot from 2024-06-06 14-03-05

tidy3d/components/data/data_array.py
```diff
@@ -700,6 +700,24 @@ class FieldProjectionAngleDataArray(DataArray):
     _data_attrs = {"long_name": "radiation vectors"}
 
 
+class DirectivityDataArray(DataArray):
+    """Directivity in frequency domain as a function of angles theta and phi.
+    Example
```
tylerflex, line 708: might need a newline above Example or this might not render (just FYI)

tidy3d/components/data/data_array.py
```diff
@@ -700,6 +700,24 @@ class FieldProjectionAngleDataArray(DataArray):
     _data_attrs = {"long_name": "radiation vectors"}
 
 
+class DirectivityDataArray(DataArray):
+    """Directivity in frequency domain as a function of angles theta and phi.
```
tylerflex: What are the units?

tidy3d/components/data/monitor_data.py
```diff
@@ -1972,6 +1974,32 @@ class FluxTimeData(MonitorData):
     )
 
 
+class DirectivityData(MonitorData):
+    """Data associated with a :class:`.DirectivityMonitor`.
+    Example
```
tylerflex, line 1981: might need newline

tidy3d/components/data/monitor_data.py
```diff
@@ -1972,6 +1974,32 @@ class FluxTimeData(MonitorData):
...
+    >>> phi = np.linspace(0, 2*np.pi, 20)
+    >>> coords = dict(r=r, theta=theta, phi=phi, f=f)
+    >>> scalar_field = DirectivityDataArray(farfield_values, coords=coords)
+    >>> monitor = DirectivityMonitor(center=(1,2,3), size=(2,2,2), freqs=f, name='n2f_monitor', phi=phi, theta=theta)
+    >>> data = DirectivityData(monitor=monitor,Directivity=scalar_field)
+    """
+
+    monitor: DirectivityMonitor = pd.Field(
+        ...,
+        title="DirectivityMonitor",
```
tylerflex: for variable / class names, put inside of double backslash quotes inside of the Fields.

However, here I think we want this to be read as plain text, so maybe
```
title="Directivity Monitor",
description="Monitor describing the angle-based projection grid on which to measure directivity data."
```

tylerflex (approved):
Thanks @QimingFlex, just a couple minor things to polish it up 100%. Otherwise looks good.

tidy3d/components/data/data_array.py
QimingFlex: Directivity is a dimensionless quantity defined as the ratio of the radiation intensity in a given direction to the average radiation intensity over all directions.

tidy3d/components/data/monitor_data.py
QimingFlex: added

tidy3d/components/data/data_array.py
tylerflex: Sounds good, could you copy that sentence into the docstring? will be helpful in case it is not clear for someone.

tidy3d/components/data/monitor_data.py
QimingFlex: Corrected. I used lowercase for 'monitor' to maintain consistency with other monitors.
`title="Directivity monitor",`

PR #1725, opened 2024-05-30
Moving scipy into core deps

momchil-flex:
Some of our validators now require scipy for STL structures. Since our validation/metadata lambdas have been increased in maximum allowed package size, we can move scipy into the core deps.


weiliangjin2021:
This doesn't affect `pip install`, but something on MC side?

momchil-flex:
Correct, it only affects various lambda builds.

weiliangjin2021 (approved):

PR #1726, opened 2024-05-30
more robust `Batch.monitor` if run status missing 

tylerflex:


momchil-flex (approved):
Looks good, only changelog needs a rebase after I merged the other fix.

Issue #1666, opened 2024-05-02
Warning if sources/monitors exactly at simulation domain edge


momchil-flex:
When a source or monitor is 0D along some dimension and is exactly at the simulation domain edge (or within some small tolerance) along that dimension, the injected/recorded fields can be unphysical in various ways depending on the boundary conditions. Best to warn that such objects should be moved away from the simulation domain boundary.

momchil-flex:
Fixed by #1733

PR #1733, opened 2024-06-03
Validate sources and monitors that are exactly at the simulation domain bounds

### LINKED ISSUE #1732: EME Monitor Validation
If a monitor is located at the boundary of an EME simulation, the simulation will run without an error and crash.

Add a validator checking if a monitor is at the boundary of an EME simulation. Attached is an example notebook where this occurs.
[EME_validation.ipynb.zip](https://github.com/user-attachments/files/15519481/EME_validation.ipynb.zip)

momchil-flex:
Fixes #1732, #1666 (fyi @QimingFlex)

We do strict validation because such simulations error in some cases, and it's hard to fix everything internally. It may also nudge people not to put their sources and monitors right where they will be affected by the PML and produce slightly unphysical results.

@dbochkov since this is a modification on the base `Simulation` class, some heat tests are failing. Do you think this change makes sense for heat still, i.e. I could just modify the tests - or do you think there are cases in which we actually don't want to error?

### DIFF
--- a/tidy3d/components/validators.py
+++ b/tidy3d/components/validators.py
@@ -162,7 +162,9 @@ def assert_unique_names(field_name: str):
     return field_has_unique_names
 
 
-def assert_objects_in_sim_bounds(field_name: str, error: bool = True):
+def assert_objects_in_sim_bounds(
+    field_name: str, error: bool = True, strict_inequality: bool = False
+):
     """Makes sure all objects in field are at least partially inside of simulation bounds."""
 
     @pydantic.validator(field_name, allow_reuse=True, always=True)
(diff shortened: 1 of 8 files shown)


tidy3d/components/validators.py
```diff
@@ -173,11 +175,14 @@ def objects_in_sim_bounds(cls, val, values):
...
+        strict_ineq = [size != 0 and strict_inequality for size in sim_size]
+
         for position_index, geometric_object in enumerate(val):
-            if not sim_box.intersects(geometric_object.geometry):
+            if not sim_box.intersects(geometric_object.geometry, strict_inequality=strict_ineq):
                 message = (
-                    f"'simulation.{field_name}[{position_index}]'"
-                    "is completely outside of simulation domain."
+                    f"'simulation.{field_name}[{position_index}]' "
+                    "is outside of the simulation domain."
```
caseyflex, line 185: maybe "is completely outside of the interior of the simulation domain" just so people don't get confused

caseyflex (approved):

dbochkov-flexcompute:
> @dbochkov since this is a modification on the base Simulation class, some heat tests are failing. Do you think this change makes sense for heat still, i.e. I could just modify the tests - or do you think there are cases in which we actually don't want to error?

Yeah, I think it's ok to error for heat simulations as well to be safe. Just wanted to double check, if monitor is exactly on symmetry plane, this will not error, right? (it's ok to put monitor right exactly on symmetry plane in heat)

momchil-flex:
Yeah, simulation domain boundaries do not truncate to the symmetry plane (this is only done on the backend eventually when getting the actual computational domain).

QimingFlex:
@momchil-flex This is helpful! I'm working on #1666 and will finish it ASAP.

momchil-flex:
@QimingFlex sorry I didn't close this but I thought my PR actually fixes it. What did you have in mind to still work on?

QimingFlex:
Oh, never mind. I thought issue #1666 wasn't fixed.

PR #1736, opened 2024-06-03
Skip interpolation validator if tracing autograd boxes

### LINKED ISSUE #1734: #1684 broke autograd tests
The new interp validator in #1684 broke autograd tests, as arrays of dtype object (during bw pass?) can not be interpolated. @marc-flex

yaugenst-flex:
Fixes #1734

Note: Checking for `ArrayBox` specifically because the validator should still fail in other instances where `self.values` might have the wrong `dtype`.

### DIFF
--- a/tidy3d/components/data/data_array.py
+++ b/tidy3d/components/data/data_array.py
@@ -1,4 +1,5 @@
 """Storing tidy3d data at it's most fundamental level as xr.DataArray objects"""
+
 from __future__ import annotations
 from typing import Dict, List, Union
 from abc import ABC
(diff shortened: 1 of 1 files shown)


tidy3d/components/data/data_array.py
```diff
@@ -113,9 +114,17 @@ def assign_data_attrs(cls, val):
...
+        """Make sure we can interp()/sel() the data.
+
+        NOTE
+        ----
+        This does not check every 'DataArray' by default. Instead, when required, this check can be
+        called from a validator, as is the case with 'CustomMedium' and 'CustomFieldSource'.
+        """
+        # skip this validator if currently tracing for autograd because
+        # self.values will be dtype('object') and not interpolatable
+        if isbox(self.values.ravel()[0]):
```
tylerflex: Thanks @yaugenst-flex

My only concern is whether `.ravel()` will be slow, but I suppose interpolation already in the validator will be a bit slow too. There's also `self.values.flat[0]`, I'm not sure the difference but this might be a little faster as it returns an iterator rather than a new array.

Another option is that I think when there are tracers in the `values`, then the full traced ArrayBox will be in  `self.attrs[AUTOGRAD_KEY]`, so we could check `if AUTOGRAD_KEY in self.attrs:`, although it's probably less robust.

tylerflex: another option could be to skip if `self.values.dtype == object`

yaugenst-flex: >  There's also self.values.flat[0], I'm not sure the difference but this might be a little faster as it returns an iterator rather than a new array.

`ravel()` only returns a copy if necessary (i.e., the array is not contiguous), but yeah `.flat[0]` is probably better.

> another option could be to skip if self.values.dtype == object

wanted to avoid this since I think it's too general, the validator should generally error on `dtype=object`

tylerflex: ok sounds good to me (re flat vs ravel vs dtype).

tylerflex (approved):

momchil-flex (approved):
Looks good except latest submodules is failing (which I guess means it is on pre/2.7), not sure if you want to update it together with this PR?

yaugenst-flex:
@momchil-flex fixed. this is ready to merge then i think

Issue #1529, opened 2024-03-08
Extremely large mode loss 


pascualmunoz:
**Describe the bug**
Calculating the modes of a silicon or silicon nitride core, and silica cladded, waveguides (typical single mode dimensions in the C-band), provides k_eff values of order 1e-5 for fundamental mode (mostly confined in the core). In the case e.g. of Si the k value is 1.34e-13. 

For information, the same cross-section but with loss-less materials provides k_eff = 0.0.

**Expected behavior**
A much lower value of k_eff, similar to material loss.

pascualmunoz:
```python
# standard python imports
import numpy as np
import matplotlib.pylab as plt

# tidy3D import
import tidy3d as td
from tidy3d.constants import C_0
import tidy3d.web as web
from tidy3d.plugins.mode import ModeSolver

# central frequency
wvl_um = 1.45
freq0 = C_0 / wvl_um

mat = td.material_library['cSi']['Green2008']
eps = mat.eps_model(freq0)

eps

n = np.sqrt( (np.abs(eps)+np.real(eps)) / 2.0 )

k = np.sqrt( (np.abs(eps)-np.real(eps)) / 2.0 )

print(n,k)

print(np.sqrt(eps))

3.478220107067083 0.0008218546723818859
(3.478220107067083+0.000821854672654681j)

```

while in refractiveindex.info we have:

https://refractiveindex.info/?shelf=main&book=Si&page=Green-2008

wvl = 1.45 µm
n = 3.4580
k = 1.3846e-13

weiliangjin2021:
@pascualmunoz it'll be very helpful if we can have your geometry and mode solver parameters to reproduce this issue.

momchil-flex:
@weiliangjin2021 actually I think the issue is simply in our Green2008 material fit. What Pascal is showing is that it has much higher loss at that wavelength than the reference.

pascualmunoz:
Yes exactly. And the same happens with silica models. Perhaps worth to revise the material library.

momchil-flex:
@weiliangjin2021 in 2.7, apart from still creating the two separate Green2008 models, let's plan to switch the default materials to "Palik_Lossless" for both cSi and SiO2. This should be done with a warning that the default has changed, and the warning can be suppressed if the user explicitly puts `material_library['cSi']['Palik_Lossless']` instead of `material_library['cSi']`.

pascualmunoz:
[...]

PR #1738, opened 2024-06-03
Add lossless Green2008 cSi and modify default SiO2 in material_library

weiliangjin2021:
Address https://github.com/flexcompute/tidy3d/issues/1529:

- Switch the default variant from `Horiba` to `Palik_Lossess` for SiO2;
- Add Lossless Green2008 for cSi. The default remains unchanged as only Green2008 covers visible spectrum

### DIFF
--- a/tidy3d/material_library/material_library.py
+++ b/tidy3d/material_library/material_library.py
@@ -6,6 +6,7 @@ import pydantic.v1 as pd
 from ..components.medium import PoleResidue, Medium2D, AnisotropicMedium, Sellmeier
 from ..components.base import Tidy3dBaseModel
 from ..components.types import Axis
+from ..log import log
 from ..exceptions import SetupError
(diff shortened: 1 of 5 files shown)


tylerflex (approved):

tidy3d/material_library/material_library.py
```diff
@@ -99,6 +100,11 @@ def __getitem__(self, variant_name):
     @property
     def medium(self):
         """The default medium."""
+        if self.name == "Silicon Dioxide":
+            log.warning(
+                "Since Tidy3d 2.7, the default variant for silicon dioxide has been switched from "
```
momchil-flex: Let's try to always use `Tidy3D` instead of `Tidy3d`.

momchil-flex (approved):
Thanks, just a tiny change and then merge.

PR #1739, opened 2024-06-04
More tests and safety checks for autograd module

yaugenst-flex:


tylerflex (approved):
Looks good, only thing is the string formatting consistency with our existing code.

Variable names in stuff to STDOUT (like the "size" in `size must be an iterable`) should be in single quotes `'`. 

Also we tend to use sentence case (?) in our exceptions and error messages, so please add period at the end.

Otherwise looks good! Thanks

PR #1740, opened 2024-06-05
Convenient perturbations

dbochkov-flexcompute:
This PR introduces some convenience functions to define medium perturbations. Currently, to define a medium with a perturbation model, one must provide perturbation model for each parameter of the medium (`permittivity` and `conductivity` for dieletrics, `eps_inf` and `poles` for dispersive). While this is fairly flexible way, it is not so straightforward to setup, especially in gui. This PR adds two general perturbation models `PermittivityPerturbation` and `IndexPerturbation`. 
- They are general in the sense that they can be applied equally to both dielectric and dispersive materials.
- `PermittivityPerturbation` is defined as perturbations to permittivity and conductivity of medium, while `IndexPerturbation` is defined as perturbations to n and k of medium at a specified frequency. 
- `PermittivityPerturbation` is exact for all frequencies, in the sense that permittivity and conductivity changes are independent of frequency. `IndexPerturbation` is exact only at a specified frequency, that is, specified dn and dk will be exactly imposed at a provided frequency and will only be approximately imposed away from it.
- Under the hood, for dielectric materials `PermittivityPerturbation` is directly translated to perturbations of permittivity and conductivity, while `IndexPerturbation` is first converted to effective permittivity and conductivity changes according to 
```
deps = 2 * n * dn + dn ** 2 - 2 * k * dk - dk ** 2
dsigma = 2 * omega * (k * dn + n * dk + dk * dn) * EPSILON_0
```
- For dispersive materials, `PermittivityPerturbation` is translated to perturbations of `eps_inf` and an additional Pole-residue pair with `c = deps / 2 / EPSILON_0` and `a = 0`. Similarly to dielectric case, the effect of `IndexPerturbation` is first converted to changes in permittivity and conductivity, and then applied using the same idea as for `PermittivityPerturbation`

Basic examples:
```
med1 = td.PerturbationMedium(
    permittivity=8,
    conductivity=0.01,
    perturbation_spec=td.IndexPerturbation(
        dn=td.ParameterPerturbation(
            heat=td.LinearHeatPerturbation(coeff=0.001, temperature_ref=300)
        ),
        dk=td.ParameterPerturbation(
            heat=td.LinearHeatPerturbation(coeff=0.0002, temperature_ref=300)
        ),
        freq=td.C_0 / 1.55,
    ),
)
```
```
si = td.material_library["aSi"]["Horiba"]
med1 = td.PerturbationPoleResidue.from_unperturbed(
    medium=si,
    perturbation_spec=td.IndexPerturbation(
        dn=td.ParameterPerturbation(
            heat=td.LinearHeatPerturbation(coeff=0.001, temperature_ref=300)
        ),
        dk=td.ParameterPerturbation(
            heat=td.LinearHeatPerturbation(coeff=0.0002, temperature_ref=300)
        ),
        freq=td.C_0 / 1.55,
    ),
)
```

I am not sure if we have enough time to put this into 2.7. If not, maybe we can just do a 2.7.1 whenever it's ready, so that it unblocks some GUI development

fyi @e-g-melo


tidy3d/components/parameter_perturbation.py
```diff
@@ -1101,3 +1116,239 @@ def is_complex(self) -> bool:
...
+    ...     electron_coeff=0.0001,
+    ...     hole_ref=0,
+    ...     hole_coeff=0.0002,
+    ... )
+    >>> deps = ParameterPerturbation(heat=heat_perturb)
+    >>> dsigma = ParameterPerturbation(charge=charge_perturb)
+    >>> permittivity_pb = PermittivityPerturbation(deps=deps, dsigma=dsigma)
+    """
+
+    deps: Optional[ParameterPerturbation] = pd.Field(
```
momchil-flex: Maybe `delta_eps`, etc. is better than just `d`. I forget if we have any convention about this anywhere already. We use `dn_dwvl` [here](https://github.com/flexcompute/tidy3d/blob/7f69690abd87b9cf6e76f5b5d0e341ad7b47ba05/tidy3d/components/medium.py#L3065-L3089) but this is an actual derivative..

momchil-flex (approved):
Looks good as far as I can tell, maybe @marc-flex should take a look if he hasn't already?

tidy3d/components/parameter_perturbation.py
dbochkov-flexcompute: yeah, I wasn't sure about this naming too. When we refer to refraction index we typically just use `n` and `k`, so having `dn` and `dk` kind of made sense, short and clear. And then I just extended it to eps and sigma. However, for permittivity and conductivity we use actual `permittivity` and `conductivity` as class field names. That is why when perturbation classes were introduced initially, I followed the convention of adding `_perturbation`, that is, `permittivity_perturbation`, `conductivity_perturbation`, `eps_inf_perturbation`, `poles_perturbation`. Also because things like `dpoles` or `delta_poles` seemed worse.

Having said that, different options I see to use in `PermittivityPerturbation` and `IndexPerturbation`:
- Stick with the `_perturbation` convection. That is, use either 
`n_perturbation`, `k_perturbation`, `permittivity_perturbation`, `conductivity_perturbation`
or `index_perturbation`, `extinction_perturbation`, `permittivity_perturbation`, `conductivity_perturbation`. In this case, defining perturbations directly and through `perturbation_spec` would be uniform, but the downside it's quite a mouthful:
```
med_direct = td.PerturbationMedium(
    permittivity=5,
    permittivity_perturbation=td.ParameterPerturbation(...),
)

med_spec_equivalent = td.PerturbationMedium(
    permittivity=5,
    perturbation_spec=td.PermittivityPerturbation(
        permittivity_perturbation=td.ParameterPerturbation(...),
    ),
)

med_spec_via_n = td.PerturbationMedium(
    permittivity=5,
    perturbation_spec=td.IndexPerturbation(
        n_perturbation=td.ParameterPerturbation(...),
    ),
)
```

- Use shortened field names similar to what I have now `dn`, `dk`, `deps`, `dsigma`, but something more explicit like `delta_n`, `delta_k`, `delta_eps`, `delta_sigma` or, on the contrary, just plain `n`, `k`, `eps`, `sigma` since they will be inside `SomethingPerturbation` already. This somewhat breaks uniformity, but maybe it can be considered as highlighting the difference between defining perturbation models directly on medium fields or as `perturbation_spec`
```
med_direct = td.PerturbationMedium(
    permittivity=5,
    permittivity_perturbation=td.ParameterPerturbation(...),
)

med_spec_equivalent = td.PerturbationMedium(
    permittivity=5,
    perturbation_spec=td.PermittivityPerturbation(
        delta_eps=td.ParameterPerturbation(...),
    ),
)

med_spec_via_n = td.PerturbationMedium(
    permittivity=5,
    perturbation_spec=td.IndexPerturbation(
        delta_n=td.ParameterPerturbation(...),
    ),
)

**Comment by e-g-melo** [comment]
Thanks @dbochkov-flexcompute! That will help us a lot in GUI.

File: tidy3d/components/parameter_perturbation.py (hunk 1 above)
Comment (dbochkov-flexcompute) [reply tidy3d/components/parameter_perturbation.py]: also, open to discuss naming of new specs as well. For example:
`IndexPerturbation` -> `NKPerturbation`
`PermittivityPerturbation` -> `DieletricPerturbation`

Comment (tylerflex) [reply tidy3d/components/parameter_perturbation.py]: I think I prefer the original `IndexPerturbation`, `PermittivityPerturbation`

File: tidy3d/components/medium.py [hunk 2]
```diff
@@ -5471,13 +5623,64 @@ class PerturbationPoleResidue(PoleResidue, AbstractPerturbationMedium):
...
+        if isinstance(p_spec, IndexPerturbation):
+            eps_complex = PoleResidue._eps_model(
+                eps_inf=eps_inf, poles=poles, frequency=p_spec.freq
+            )
+            n, k = Medium.eps_complex_to_nk(eps_c=eps_complex)
+            deps_range, _ = p_spec._deps_dsigma_ranges(n, k)
+        elif isinstance(p_spec, PermittivityPerturbation):
+            deps_range, _ = p_spec._deps_dsigma_ranges()
+        else:
+            raise SetupError("Unknown type of 'perturbation_spec'.")
```
tylerflex, line 5709: consider adding the `type(p_spec)` in the message

tylerflex (approved):
Looks good to me @dbochkov-flexcompute , thanks.

tidy3d/components/parameter_perturbation.py
momchil-flex: Agree re IndexPerturbation, PermittivityPerturbation.

About the rest, I am not sure. `_perturbation` has its merits but is indeed quite verbose. Regarding your comment that it could just be `n` instead of `dn` or `delta_n`, could it also just be `delta` since the `SomethingPerturbation` defines what exactly is being perturbed?

```
med_spec_equivalent = td.PerturbationMedium(
    permittivity=5,
    perturbation_spec=td.PermittivityPerturbation(
        delta=td.ParameterPerturbation(...),
    ),
) 
```

If not then I do prefer `delta_n`, etc. to `dn`. Or at least `d_n`. Otherwise you get awkward to parse kwarg names like `deps` or `dpoles`.

dbochkov-flexcompute: yeah, just both `IndexPerturbation` and `PermittivityPerturbation` represent perturbation of two parameters each, n and k for `IndexPerturbation`, eps and sigma for `PermittivityPerturbation`. Like this:
```
med_1 = td.PerturbationMedium(
    permittivity=5,
    perturbation_spec=td.PermittivityPerturbation(
        delta_eps=td.ParameterPerturbation(...),
        delta_sigma=td.ParameterPerturbation(...),
    ),
) 

med_2 = td.PerturbationMedium(
    permittivity=5,
    perturbation_spec=td.IndexPerturbation(
        delta_n=td.ParameterPerturbation(...),
        delta_k=td.ParameterPerturbation(...),
        freq=freq0.
    ),
) 
```

dbochkov-flexcompute: btw, any preference between `freq` and `frequency`? It looks like we use both options in our code, for example, `eps_model(frequency=...)` but `from_nk(n=..., k=..., freq=...)`

tylerflex: we unfortunately use `f` in the DataArrays 😢 . (and also mixed notation `freq0` and `f_width` in the gaussian pulse... ) what do you think? I'm maybe inclined towards `freq` to be consistent with `from_nk`

dbochkov-flexcompute:
Changed namings to
```
PermeabilityPerturbation(
    delta_eps=...,
    delta_sigma=...,
)

IndexPerturbation(
    delta_n=...,
    delta_k=...,
    freq=...,
)
```

dbochkov-flexcompute:
also rebased to the latest develop. If it is not going to throw a wrench into the works, it would be good to have this in 2.7.0

PR #1741, opened 2024-06-05
add WavePorts

dmarek-flex:
add a new type of port to the TerminalComponentModeler. WavePorts are  the optimal port for exciting transmission lines, and they function like a hybrid between LumpedPorts and the modal-based Port.

Still a work in progress. 

Here are some thoughts:
- Not sure the best way to make the computed port impedances available to users. Computing the port impedances requires mode solving which is launched through `ModeSolverMonitor`s along with the tidy3d simulations. However, some users might want to know the impedances beforehand or inspect modes before launching the simulations.
- I have changed some of the naming to reflect that the ports of the `TerminalComponentModeler` are based on pairs of terminals ( a positive and negative connection or node in the circuit). `LumpedPort`s and `WavePort` are then both `TerminalPort`s
- I realized it is helpful to add `name`s and possibly `attrs` to `DataArray`s that I am creating. Maybe I should put all of these `DataArray`s into the microwave plugin or the Tidy3D files.

### DIFF
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -735,6 +735,29 @@ class Geometry(Tidy3dBaseModel, ABC):
         return axis, position
 
     @staticmethod
+    def parse_two_xyz_kwargs(**xyz) -> List[Tuple[Axis, float]]:
+        """Turns x,y,z kwargs into indices of axes and the position along each axis.
+
+        Parameters
+        ----------
+        x : float = None
+            Position in x direction, only two of x,y,z can be specified to define line.
+        y : float = None
+            Position in y direction, only two of x,y,z can be specified to define line.
+        z : float = None
+            Position in z direction, only two of x,y,z can be specified to define line.
+
+        Returns
+        -------
+        [(int, float), (int, float)]
+            Index into xyz axis (0,1,2) and position along that axis.
+        """
+        xyz_filtered = {k: v for k, v in xyz.items() if v is not None}
+        assert len(xyz_filtered) == 2, "exactly two kwarg in [x,y,z] must be specified."
+        xyz_list = list(xyz_filtered.items())
+        return [("xyz".index(axis_label), position) for axis_label, position in xyz_list]
+
+    @staticmethod
     def rotate_points(points: ArrayFloat3D, axis: Coordinate, angle: float) -> ArrayFloat3D:
         """Rotate a set of points in 3D.
 
--- a/tidy3d/components/viz.py
+++ b/tidy3d/components/viz.py
@@ -75,18 +75,15 @@ def equal_aspect(plot):
 """ plot parameters """
 
 
-class PlotParams(Tidy3dBaseModel):
-    """Stores plotting parameters / specifications for a given model."""
+class AbstractPlotParams(Tidy3dBaseModel):
+    """Abstract class for storing plotting parameters.
+    Corresponds with select properties of ``matplotlib.artist.Artist``.
+    """
 
     alpha: Any = pd.Field(1.0, title="Opacity")
-    edgecolor: Any = pd.Field(None, title="Edge Color", alias="ec")
-    facecolor: Any = pd.Field(None, title="Face Color", alias="fc")
-    fill: bool = pd.Field(True, title="Is Filled")
-    hatch: str = pd.Field(None, title="Hatch Style")
     zorder: float = pd.Field(None, title="Display Order")
-    linewidth: pd.NonNegativeFloat = pd.Field(1, title="Line Width", alias="lw")
 
-    def include_kwargs(self, **kwargs) -> PlotParams:
+    def include_kwargs(self, **kwargs) -> AbstractPlotParams:
         """Update the plot params with supplied kwargs."""
         update_dict = {
             key: value
@@ -101,6 +98,32 @@ class PlotParams(Tidy3dBaseModel):
         for ignore_key in ("type", "attrs"):
             kwarg_dict.pop(ignore_key)
         return kwarg_dict
+
+
+class PathPlotParams(AbstractPlotParams):
+    """Stores plotting parameters / specifications for a path.
+    Corresponds with select properties of ``matplotlib.lines.Line2D``.
+    """
+
+    color: Any = pd.Field(None, title="Color", alias="c")
+    linewidth: pd.NonNegativeFloat = pd.Field(2, title="Line Width", alias="lw")
+    linestyle: str = pd.Field("--", title="Line Style", alias="ls")
+    marker: Any = pd.Field("o", title="Marker Style")
+    markeredgecolor: Any = pd.Field(None, title="Marker Edge Color", alias="mec")
+    markerfacecolor: Any = pd.Field(None, title="Marker Face Color", alias="mfc")
+    markersize: pd.NonNegativeFloat = pd.Field(10, title="Marker Size", alias="ms")
+
+
+class PlotParams(AbstractPlotParams):
+    """Stores plotting parameters / specifications for a given model.
+    Corresponds with select properties of ``matplotlib.patches.Patch``.
+    """
+
+    edgecolor: Any = pd.Field(None, title="Edge Color", alias="ec")
+    facecolor: Any = pd.Field(None, title="Face Color", alias="fc")
+    fill: bool = pd.Field(True, title="Is Filled")
+    hatch: str = pd.Field(None, title="Hatch Style")
+    linewidth: pd.NonNegativeFloat = pd.Field(1, title="Line Width", alias="lw")
 
 
 # defaults for different tidy3d objects
--- a/tidy3d/plugins/microwave/custom_path_integrals.py
+++ b/tidy3d/plugins/microwave/custom_path_integrals.py
@@ -11,10 +11,26 @@ import xarray as xr
 from ...components.base import cached_property
 from ...components.data.data_array import FreqDataArray, FreqModeDataArray, TimeDataArray
 from ...components.data.monitor_data import FieldData, FieldTimeData, ModeSolverData
-from ...components.types import ArrayFloat2D, Axis
+from ...components.geometry.base import Geometry
+from ...components.types import ArrayFloat2D, Ax, Axis, Bound, Coordinate
+from ...components.viz import add_ax_if_none
 from ...constants import MICROMETER, fp_eps
 from ...exceptions import DataError, SetupError
-from .path_integrals import AbstractAxesRH, IntegralResultTypes, MonitorDataTypes
+from .path_integrals import (
+    AbstractAxesRH,
+    CurrentIntegralAxisAligned,
+    IntegralResultTypes,
+    MonitorDataTypes,
+    VoltageIntegralAxisAligned,
+    _check_em_field_supported,
+)
+from .viz import (
+    ARROW_CURRENT,
+    plot_params_current_path,
+    plot_params_voltage_minus,
+    plot_params_voltage_path,
+    plot_params_voltage_plus,
+)
 
 FieldParameter = Literal["E", "H"]
 
@@ -72,8 +88,6 @@ class CustomPathIntegral2D(AbstractAxesRH):
         :class:`.IntegralResultTypes`
             Result of integral over remaining dimensions (frequency, time, mode indices).
         """
-        if not isinstance(em_field, (FieldData, FieldTimeData, ModeSolverData)):
-            raise DataError("'em_field' type not supported.")
 
         (dim1, dim2, dim3) = self.local_dims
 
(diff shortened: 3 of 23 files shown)


dmarek-flex:
@weiliangjin2021 @QimingFlex Still a bit rough at the moment in terms of docstrings and other details. But if you want to take a look over the next week while I am validating the accuracy, I am ready for feedback.

tidy3d/plugins/microwave/impedance_calculator.py
```diff
@@ -82,6 +82,7 @@ def compute_impedance(self, em_field: MonitorDataTypes) -> IntegralResultTypes:
                 current = np.conj(2 * flux / voltage)
 
         impedance = voltage / current
+        impedance.name = "impedance (Ohm)"
```
weiliangjin2021: Add a `unit` field to avoid hard-coding "Ohm"? Similarly to voltage and current.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
```diff
@@ -122,9 +122,13 @@ def to_source(
...
-        inds = grid.discretize_inds(box=bounding_box)
-        num1 = inds[trans_axes[0]][1] - inds[trans_axes[0]][0]
-        num2 = inds[trans_axes[1]][1] - inds[trans_axes[1]][0]
+        if grid:
+            inds = grid.discretize_inds(box=bounding_box)
+            num1 = inds[trans_axes[0]][1] - inds[trans_axes[0]][0]
+            num2 = inds[trans_axes[1]][1] - inds[trans_axes[1]][0]
+        else:
+            num1 = 11
+            num2 = 11
```
weiliangjin2021: this is for debug purpose?

weiliangjin2021 (commented):
1st pass through the code. I'll definitely need to take several more rounds to have a better understanding. Very cool to see all components stick together! One main question is if we should support multiple ways of computing impedance: using any two quantities of V, I, P.

dmarek-flex:
> One main question is if we should support multiple ways of computing impedance: using any two quantities of V, I, P.

Yeah, it is all hidden away in the `ImpedanceCalculator`. User can provide any definition of voltage they want (by defining the associated voltage path integral). Same goes for current computation. If both integrals are provided, the impedance is computed as expected V/I. If only one is provided, the  `ImpedanceCalculator` will r[evert to using the total flux the indirectly compute the other quantity](https://github.com/flexcompute/tidy3d/blob/612f55164fcacdc531fdaa0b6397e0a1e27b2bab/tidy3d/plugins/microwave/impedance_calculator.py#L71-L85). If none are provided, a ValidationError gets raised.

tidy3d/plugins/smatrix/ports/coaxial_lumped.py
dmarek-flex: Hmm no, but it is convenient to be able to create the source without knowing the grid information, so I put a default value for the number of points. But maybe I should make that a variable somewhere like `DEFAULT_GRID_POINTS` or alternative.

tidy3d/plugins/microwave/impedance_calculator.py
dmarek-flex: I added some methods to set these types of metadata. Is what I did acceptable? Another way I was thinking of doing this was to create DataClasses with this information, but then I would need one DataClass for each type in `IntegralResultTypes`, which would be overly complicated.

dmarek-flex:
> 1st pass through the code. I'll definitely need to take several more rounds to have a better understanding. Very cool to see all components stick together! One main question is if we should support multiple ways of computing impedance: using any two quantities of V, I, P.

Might help for your review to check out this [notebook](https://drive.google.com/file/d/1LHHCua8wty-gdqsH4zVV9GbJLrdjPmhD/view?usp=drive_link).

tidy3d/plugins/microwave/impedance_calculator.py
weiliangjin2021: Current version looks good to me. I thought that everything is immutable in Tidy3D, but data array is an exception? I don't know if `data_array.rename("Z0")` is better than `data_array.name="Z0`? @tylerflex

tylerflex: DataArrays (like numpy arrays) are mutable. If the `name` is being used just for cosmetic purposes, eg plotting, I think setting it directly is fine. but if it's being used for computation in any way (not recommended), we should probably make a copy, just to ensure that the functions have no side effects.

tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -0,0 +1,184 @@
...
+            freqs=freqs,
+            name=self._field_monitor_name,
+            colocate=False,
+        )
+        return [field_mon]
+
+    def to_mode_solver_monitor(self, freqs: FreqArray) -> ModeSolverMonitor:
+        """Mode solver monitor to compute modes that will be used to
+        compute characteristic impedances."""
+        mode_mon = ModeSolverMonitor(
```
weiliangjin2021, line 106: mode_solver class has a method `to_mode_solver_monitor` [here](https://github.com/flexcompute/tidy3d/blob/bb075743059441258c25043341fb1091097d8723/tidy3d/plugins/mode/mode_solver.py#L963)

tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -0,0 +1,184 @@
...
+        impedance_calc = ImpedanceCalculator(
+            voltage_integral=self.voltage_integral, current_integral=self.current_integral
+        )
+        if isinstance(sim_mode_data, SimulationData):
+            mode_solver_data = sim_mode_data[self._mode_monitor_name]
+        else:
+            mode_solver_data = sim_mode_data
+
+        impedance_array = impedance_calc.compute_impedance(mode_solver_data)
+        impedance_array = impedance_array.sel(mode_index=self.mode_index)
```
weiliangjin2021: `sel` before computing impedance to skip unnecessary computing?

weiliangjin2021 (commented):
Went through another pass. All look quite nice to me! I'm still wondering if we have a way to compute currents that doesn't need contour path integration. One way that I can think of is as follows (but don't know if it works well):

- On the modal plane (let's assume it to be xy-plane), compute **J** at every grid point through **J** = Curl **H**. Next we integrate **J**z over the entire modal plane. During the integration, we integrate only positive or negatie **J**z, as otherwise the current cancels out. So it's _int dS Heaviside(J_Z) J_z_. Does it agree with the contour path approach for TEM mode?

Some minor comments:

- It'll be very helpful to also visualize voltage/current integration path in the plot.
- For voltage integration path, in many cases, it is an straight line. When it's axis-aligned, it'll be more convenient to allow the definition through 2 endpoints?

dmarek-flex:
> * On the modal plane (let's assume it to be xy-plane), compute **J** at every grid point through **J** = Curl **H**. Next we integrate **J**z over the entire modal plane. During the integration, we integrate only positive or negatie **J**z, as otherwise the current cancels out. So it's _int dS Heaviside(J_Z) J_z_. Does it agree with the contour path approach for TEM mode?

That could work, and it seems like a discrete version of Stoke's theorem, so should be equivalent. It is also probably an easier way to implement the computation on the Yee grid when the conductors are complicated shapes. I don't think that I could use that method for computing currents at the ports though, because we do not know whether the current is flowing into or out of the port before computing. Could be used for compute Z0 when we can assume the propagation direction.

However, since it seems possible to apply this to more complex shapes more easily, I think I will try. We could still make the user define a path around the conductor of interest. Instead of computing the line integral though, we compute the current inside the contour using curl H.

> * It'll be very helpful to also visualize voltage/current integration path in the plot.

Yes, good idea! I'll experiment.

> * For voltage integration path, in many cases, it is an straight line. When it's axis-aligned, it'll be more convenient to allow the definition through 2 endpoints?

So do you mean an option to create path integrals given P1(x1,y1,z1) and P2(x2,y2,z2) and just ensure that that it is axis aligned with a validator? If so I can do that!

weiliangjin2021:
> So do you mean an option to create path integrals given P1(x1,y1,z1) and P2(x2,y2,z2) and just ensure that that it is axis aligned with a validator? If so I can do that!

Hopefully we can also support axis-unaligned straight path. But you mentioned there is complication for that

dmarek-flex:
> > So do you mean an option to create path integrals given P1(x1,y1,z1) and P2(x2,y2,z2) and just ensure that that it is axis aligned with a validator? If so I can do that!
> 
> Hopefully we can also support axis-unaligned straight path. But you mentioned there is complication for that

Agreed, at least we can easily swap out implementations for voltage and current integrals easily. But that is an improvement that is needed.

tidy3d/plugins/smatrix/ports/wave.py
dmarek-flex: After checking, it won't be convenient to use that functionality in the WavePort. First, I would need to call the `to_mode_solver` method on the port. But that method is only a convenience method for checking modes before simulation, plus it requires `Simulation`.

tidy3d/plugins/smatrix/ports/wave.py
dmarek-flex: had to make small change to monitor data, but was able to do this

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -138,6 +145,58 @@ def _compute_dl_component(coord_array: xr.DataArray, closed_contour=False) -> np
...
+            When ``True``, the points will be ordered clockwise with respect to the positive
+            direction of the ``normal_axis``.
+
+        Returns
+        -------
+        :class:`.CustomPathIntegral2D`
+            A path integral defined on a circular path.
+        """
+
+        # Helper for generating x,y vertices around a circle in the local coordinate frame
```
tylerflex: please add type annotations and add as function docstring `"""`

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -191,14 +192,21 @@ def compute_voltage(self, em_field: MonitorDataTypes) -> IntegralResultTypes:
...
         if self.sign == "+":
             voltage *= -1
+
+        voltage = VoltageIntegralAxisAligned._set_data_array_attributes(voltage)
         # Return data array of voltage while keeping coordinates of frequency|time|mode index
         return voltage
 
+    @staticmethod
+    def _set_data_array_attributes(data_array: IntegralResultTypes) -> IntegralResultTypes:
+        data_array.name = "V"
```
tylerflex, line 236: docstring (pops up in a few other places too)

tidy3d/plugins/microwave/impedance_calculator.py
```diff
@@ -82,14 +84,21 @@ def compute_impedance(self, em_field: MonitorDataTypes) -> IntegralResultTypes:
...
-                "Atleast one of 'voltage_integral' or 'current_integral' must be provided."
+                "At least one of 'voltage_integral' or 'current_integral' must be provided."
             )
         return val
+
+    @staticmethod
+    def _set_data_array_attributes(data_array: IntegralResultTypes) -> IntegralResultTypes:
+        """Helper to set additional metadata for ``IntegralResultTypes``."""
+        data_array.name = "Z0"
+        return data_array.assign_attrs(units=OHM, long_name="characteristic impedance")
```
tylerflex, line 103: is it possible to set this as `data_attrs` in `IntegralResultType`?

tylerflex: if so, we could even just set the `data_array.name` without calling this helper.

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -35,7 +36,7 @@ class AbstractComponentModeler(ABC, Tidy3dBaseModel):
         description="Simulation describing the device without any sources present.",
     )
 
-    ports: Tuple[Union[Port, LumpedPort], ...] = pd.Field(
+    ports: Tuple[Union[Port, AbstractTerminalPort], ...] = pd.Field(
```
tylerflex: typically it's safer to annotate with a `Union[]` of all possible types, instead of the abstract type. So there would be like a `TerminalPortType = Union[...]` and then `Union[Port, TerminalPortType]`

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -214,43 +215,46 @@ def load(self, path_dir: str = DEFAULT_DATA_DIR) -> DataArray:
...
+        port_pos_gt_grid_bounds = np.argwhere(port_position > grid_boundaries)
+
+        # no port index can be determined
+        if len(port_pos_gt_grid_bounds) == 0:
+            raise SetupError(f"Port position '{port_position}' outside of simulation bounds.")
+        port_index = port_pos_gt_grid_bounds[-1]
+
+        # shift the port to the left
+        if port.direction == "+":
+            shifted_index = port_index - 2
```
tylerflex, line 256: where does this `2` come from again? and does it factor in anywhere else? it's been a while but I vaguely remember needing to correct the results to account for it?

tylerflex (approved):
thanks @dmarek-flex . Looks overall good to me. Although I can't follow many of the specific details.

tidy3d/plugins/smatrix/component_modelers/base.py
dmarek-flex: I'm not sure, I just moved `_shift_value_signed` into the base class so that I could use it too. It could be because the tangential H field is offset from E field by half a step. I'm not sure how the direction of the offset is set for the source and monitor, but maybe depending on the sign they could be opposite? So you need 2 cells to make sure they don't match

dmarek-flex: e.g. the H field of the source offsets half cell to the right, but the field monitor records H field offset to the left half cell.

tidy3d/plugins/microwave/impedance_calculator.py
dmarek-flex: `IntegralResultTypes = Union[FreqDataArray, FreqModeDataArray, TimeDataArray]`

I wanted to avoid making unique classes for each possible output type, so I don't think that is possible. The name and attribute are simply for clarity when plotting or printing.

dmarek-flex:
> It'll be very helpful to also visualize voltage/current integration path in the plot.

Just adding some tests, but path integrals can be plotted now. 
Screenshot from 2024-07-04 15-38-47

weiliangjin2021:
> > It'll be very helpful to also visualize voltage/current integration path in the plot.
> 
> Just adding some tests, but path integrals can be plotted now. Screenshot from 2024-07-04 15-38-47

Nice figure!

tidy3d/components/viz.py
```diff
@@ -73,18 +73,15 @@ def _plot(*args, **kwargs) -> Ax:
 """ plot parameters """
 
 
-class PlotParams(Tidy3dBaseModel):
-    """Stores plotting parameters / specifications for a given model."""
+class AbstractPlotParams(Tidy3dBaseModel):
```
dmarek-flex, line 78: @tylerflex Might want to check my changes here which I used for plotting the path integrals

tidy3d/components/geometry/base.py
```diff
@@ -684,6 +684,29 @@ def parse_xyz_kwargs(**xyz) -> Tuple[Axis, float]:
         axis = "xyz".index(axis_label)
         return axis, position
 
+    @staticmethod
+    def parse_two_xyz_kwargs(**xyz) -> List[Tuple[Axis, float]]:
```
dmarek-flex, line 738: @tylerflex also here which might affect others

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -18,16 +19,34 @@
...
+    plot_params_voltage_path,
+    plot_params_voltage_plus,
+)
 
 MonitorDataTypes = Union[FieldData, FieldTimeData, ModeSolverData]
 EMScalarFieldType = Union[ScalarFieldDataArray, ScalarFieldTimeDataArray, ScalarModeFieldDataArray]
 IntegralResultTypes = Union[FreqDataArray, FreqModeDataArray, TimeDataArray]
 
 
+def _check_em_field_supported(em_field: Any):
```
dmarek-flex, line 41: @weiliangjin2021 @tylerflex I wonder if you have any ideas for checking class method parameters. I was experimenting with decorators and looked into pydantic support. decorators mess up the docs and pydantic support does not work in our version, I think in V2 it will be [supported](https://docs.pydantic.dev/latest/concepts/validation_decorator/).

tylerflex: as far as I understand, we use decorators for this kind of thing in a few places, but agree it's not great for the docs.
https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/medium.py#L95

dmarek-flex:
@weiliangjin2021 I am ready to merge this as soon as you confirm. Tests are passing after rebase and my testing notebook is giving good results. I just changed the CHANGELOG to reflect the changes in this PR.

weiliangjin2021 (approved):
Please go ahead

PR #1742, opened 2024-06-05
Relax stability condition for simulations smaller then 3D by filtering cell sizes based on dimension

QimingFlex:
@momchil-flex I tested this with a 2D simulation. Is there any file where I should add a test function?


momchil-flex:
I think it's fine as long as none of our 2D sims do not diverge. :)

momchil-flex:
Ah I forgot that there should be a changelog entry. I'll make a note to add one when updating the changelog for the release.

QimingFlex:
Noted. I'll add a changelog entry in the next update.

PR #1743, opened 2024-06-06
support for differentiation w.r.t. `CustomMedium.eps_dataset` components

tylerflex:
sadly doesn't fix our topology optimization problem but at least we can support this now.

Was also super minimal changes to the code (compared to adjoint plugin), which is a hopeful sign for our future.


yaugenst-flex (approved):
Cool!

Issue #1708, opened 2024-05-17
Strange PML plotting in 2D


tylerflex:
<img width="307" alt="image" src="https://github.com/flexcompute/tidy3d/assets/92756888/4ac0dca2-fa1d-4a95-ab77-9fcc7fbec4e4">

for more details refer to https://flow360.atlassian.net/browse/SCEM-6289

momchil-flex:
Off the top of my head this is probably something to do with 2D simulations still being assigned a non-zero `dl` along the 0-dim, while PMLs are just plotted with effectively 0 width.

Frank-DingYC:
Yes, the problem is exactly what you describe. However, the structures are with 0-dim too while they still have some width. Could we just make the PML and structures look the same?

Issue #1416, opened 2024-01-26
Convert SimulationData to .mat file


tylerflex:
SimulationData objects are often saved to `.hdf5` file. But some users have requested an easy way to save to `.mat` file for use in Matlab. 

Let's create a new method of `SimulationData` that saves the data to a `.mat` file specified by the user. 

For the internals, I think something like
```py
scipy.io.savemat("data.mat", sim_data.dict())
```
might be how it can be done, but needs some testing.

PR #1744, opened 2024-06-07
Issue #1416 and #1708 changes

m-bone:
Modified simulation.py so 2D simulation plots have correctly drawn boundaries as in Issue #1708 .
Before  and after example plots
<p float="left">
  
   
</p>
Added test for 2D boundary width in test_viz.py which checks the width of plot objects is significantly larger than the plot width

Added ability to output `SimulationData` to .mat file as requested in Issue #1416
Implemented as `to_mat` method in sim_data.py with formatting functions included in mat.py
Test for formatting added to test_sim_data.py and utils.py

These are small standalone changes so they can likely be added to 2.7.0 without much issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Users can now export `SimulationData` to MATLAB `.mat` files with the `to_mat_file` method.
+
+### Fixed
+- Bug where boundary layers would be plotted too small in 2D simulations.
+
 ## [2.7.0] - 2024-06-17
 
 ### Added
--- a/tests/test_components/test_viz.py
+++ b/tests/test_components/test_viz.py
@@ -4,6 +4,7 @@ import matplotlib.pyplot as plt
 import pytest
 import tidy3d as td
 from tidy3d.components.viz import Polygon
+from tidy3d.constants import inf
 
 
 def test_make_polygon_dict():
@@ -36,3 +37,45 @@ def test_0d_plot(center_z, len_collections):
     assert len(ax.collections) == len_collections
 
     plt.close()
+
+
+def test_2d_boundary_plot():
+    """
+    Test that boundary box structures are drawn to full size for 2D plots where the simulation size is 0
+    """
+
+    # Dummy objects to pad the simulation
+    freq0 = td.C_0 / 0.75
+
+    # create source
+    source = td.PointDipole(
+        center=(0, 0, 0),
+        source_time=td.GaussianPulse(freq0=freq0, fwidth=freq0 / 10.0),
+        polarization="Ez",
+    )
+
+    # Simulation details
+    per_boundary = td.Boundary.periodic()
+    pml_boundary = td.Boundary.pml(num_layers=2)
+
+    sim = td.Simulation(
+        size=(0, 1, 1),
+        grid_spec=td.GridSpec.auto(min_steps_per_wvl=25),
+        structures=[],
+        sources=[source],
+        monitors=[],
+        run_time=120 / freq0,
+        boundary_spec=td.BoundarySpec(x=per_boundary, y=pml_boundary, z=pml_boundary),
+    )
+
+    pml_box = sim._make_pml_box(pml_axis=1, pml_height=1, sign=1)
+
+    # Should have infinite size in x
+    assert pml_box.size[0] == inf
+
+    # Create new 3D simulation
+    sim = sim.updated_copy(size=(1, 1, 1))
+    pml_box = sim._make_pml_box(pml_axis=1, pml_height=1, sign=1)
+
+    # should have a non-infinite size as x is specified
+    assert pml_box.size[0] != inf
--- a/tests/test_data/test_sim_data.py
+++ b/tests/test_data/test_sim_data.py
@@ -8,9 +8,11 @@ import tidy3d as td
 from tidy3d.components.data.data_array import ScalarFieldTimeDataArray
 from tidy3d.components.data.monitor_data import FieldTimeData
 from tidy3d.components.data.sim_data import SimulationData
+from tidy3d.components.file_util import replace_values
 from tidy3d.components.monitor import FieldMonitor, FieldTimeMonitor, ModeMonitor
 from tidy3d.exceptions import DataError, Tidy3dKeyError
 
+from ..utils import get_nested_shape
 from .test_data_arrays import FIELD_MONITOR, SIM, SIM_SYM
 from .test_monitor_data import (
     make_diffraction_data,
@@ -414,3 +416,57 @@ def test_loading_non_field_data():
     sim_data = make_sim_data()
     with pytest.raises(DataError):
         sim_data.load_field_monitor("flux")
+
+
+def test_replace_values_dict():
+    """
+    Test that replacing values in initial dict hasn't changed the shape and that values have been correctly replaced.
+    Used in simData.to_mat() method.
+    """
+    # Create data for test
+    test_data = make_sim_data()
+    test_data_dict = test_data.dict()
+    test_data_dict["none_test"] = None  # Check that replace works at top level of nested dict
+
+    # Get the original shape of nested dict
+    original_shape = get_nested_shape(test_data_dict)
+
+    # Modify dict and get shape of dict
+    new_data = replace_values(test_data_dict, None, [])
+    new_shape = get_nested_shape(new_data)
+
+    assert (
+        original_shape == new_shape
+        and new_data["simulation"]["medium"]["name"] == []
+        and new_data["none_test"] == []
+    )
+
+
+def test_replace_values_list():
+    """
+    Test that replacing values in initial list hasn't changed the shape and that values have been correctly replaced.
+    """
+    # Create data for test
+    test_list = [
+        {"nest1": {"nest2": {"nest3": None}}},
+        None,
+        [5, "asdf", 12.34, None, False, [3.14, None]],
+        3.14,
+        (None, None, "notNone"),
+    ]
+    original_shape = get_nested_shape(test_list)
+
+    # Modify list and get shape
+    new_list = replace_values(test_list, 3.14, [])
+    new_shape = get_nested_shape(new_list)
+
+    assert original_shape == new_shape and new_list[3] == [] and new_list[2][5][0] == []
+
+
+def test_to_mat_file(tmp_path):
+    """
+    Test output of ``.mat`` file completes without error.
+    """
+    sim_data = make_sim_data()
+    path = str(tmp_path / "test.mat")
+    sim_data.to_mat_file(path)
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1158,3 +1158,39 @@ def get_test_root_dir():
     """return the root folder of test code"""
 
     return Path(__file__).parent
+
+
+def get_nested_shape(nested_obj: Any) -> Any:
+    """
+    Recursively iterate through a nested object replacing values with None.
+    Empty list/tuple/dict are replaced with None.
+    Builds the structure for comparison to other nested objects.
+    Used to check structure hasn't changed when nested_obj data has been altered.
+    Similar concept to .shape method for numpy arrays.
+
+    Parameters
+    ----------
+    nested_obj : Any
+        A nested object to be reduced to its shape.
+
+    Returns
+    -------
+    Any
+        The nested object with values replaced with None whilst keeping the same nested structure.
+    """
+    if isinstance(nested_obj, dict):
+        if len(nested_obj) == 0:
+            return None
+        else:
+            return {key: get_nested_shape(nested_obj[key]) for key in nested_obj}
+
+    # Tuple of dicts, enter and continue iteration
+    elif isinstance(nested_obj, (tuple, list)):
+        if len(nested_obj) == 0:
+            return None
+        else:
+            return type(nested_obj)(get_nested_shape(val) for val in nested_obj)
+
+    # Replace everything else with None
+    else:
+        return None
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -12,11 +12,13 @@ import h5py
 import numpy as np
 import pydantic.v1 as pd
 import xarray as xr
+from scipy.io import savemat
 
-from ...exceptions import DataError, Tidy3dKeyError
+from ...exceptions import DataError, FileError, Tidy3dKeyError
 from ...log import log
 from ..base import JSON_TAG
 from ..base_sim.data.sim_data import AbstractSimulationData
+from ..file_util import replace_values
 from ..monitor import Monitor
 from ..simulation import Simulation
 from ..source import Source
@@ -1003,3 +1005,44 @@ class SimulationData(AbstractYeeGridSimulationData):
 
         monitor_name = Structure.get_monitor_name(index=structure_index, data_type=data_type)
         return self[monitor_name]
+
+    def to_mat_file(self, fname: str, **kwargs):
+        """Output the ``SimulationData`` object as ``.mat`` MATLAB file.
+
+        Parameters
+        ----------
+        fname : str
+            Full path to the output file. Should include ``.mat`` file extension.
+        **kwargs : dict, optional
+            Extra arguments to ``scipy.io.savemat``: see ``scipy`` documentation for more detail.
+
+        Example
+        -------
+        >>> simData.to_mat_file('/path/to/file/data.mat') # doctest: +SKIP
+        """
+        # Check .mat file extension is given
+        extension = pathlib.Path(fname).suffixes[0].lower()
+        if len(extension) == 0:
+            raise FileError(f"File '{fname}' missing extension.")
+        if extension != ".mat":
+            raise FileError(f"File '{fname}' should have a .mat extension.")
+
+        # Handle m_dict in kwargs
+        if "m_dict" in kwargs:
+            raise ValueError(
+                "'m_dict' is automatically determined by 'to_mat_file', can't pass to 'savemat'."
+            )
+
+        # Get SimData object as dictionary
+        sim_dict = self.dict()
+
+        # Remove NoneType values from dict
+        # Built from theory discussed in https://github.com/scipy/scipy/issues/3488
+        modified_sim_dict = replace_values(sim_dict, None, [])
+
+        try:
+            savemat(fname, modified_sim_dict, **kwargs)
+        except Exception as e:
+            raise ValueError(
+                "Could not save supplied 'SimulationData' to file. As this is an experimental feature, we may not be able to support the contents of your dataset. If you receive this error, please feel free to raise an issue on our front end repository so we can investigate."
+            ) from e
--- a/tidy3d/components/file_util.py
+++ b/tidy3d/components/file_util.py
@@ -2,6 +2,9 @@
 
 import gzip
 import shutil
+from typing import Any
+
+import numpy as np
 
 
 def compress_file_to_gzip(input_file, output_gz_file):
@@ -28,3 +31,37 @@ def extract_gzip_file(input_gz_file, output_file):
     with gzip.open(input_gz_file, "rb") as file_in:
         with open(output_file, "wb") as file_out:
             shutil.copyfileobj(file_in, file_out)
+
+
+def replace_values(values: Any, search_value: Any, replace_value: Any) -> Any:
+    """
+    Create a copy of ``values`` where any elements equal to ``search_value`` are replaced by ``replace_value``.
+
+    Parameters
+    ----------
+    values : Any
+        The input object to iterate through.
+    search_value : Any
+        An object to match for in ``values``.
+    replace_value : Any
+        A replacement object for the matched value in ``values``.
+
+    Returns
+    -------
+    Any
+        values type object with ``search_value`` terms replaced by ``replace_value``.
+    """
+    # np.all allows for arrays to be evaluated
+    if np.all(values == search_value):
+        return replace_value
+    if isinstance(values, dict):
+        return {
+            key: replace_values(val, search_value, replace_value) for key, val in values.items()
+        }
+    elif isinstance(
+        values, (tuple, list)
+    ):  # Parts of the nested dict structure include tuples with more dicts
+        return type(values)(replace_values(val, search_value, replace_value) for val in values)
+
+    # Used to maintain values that are not search_value or containers
+    return values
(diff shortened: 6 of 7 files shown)


CHANGELOG.md
```diff
@@ -23,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ModeSolver.plot_field` correctly returning the plot axes.
 - Avoid error if non-positive refractive index used for integration resolution in adjoint.
 - Make `Batch.monitor` robust if the run status is not found.
+- Bug where boundary layers would be plot too small in 2D simulations
```
tylerflex: plot -> plotted. Add period

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `RectangularWaveguide` supports layered cladding above and below core.
 - `SubpixelSpec` accepted by `Simulation.subpixel` to select subpixel averaging methods separately for dielectric, metal, and PEC materials. Specifically, added support for conformal mesh methods near PEC structures that can be specified through the field `pec` in the `SubpixelSpec` class. Note: previously, `subpixel=False` was implementing staircasing for every material except PEC. Now, `subpixel=False` implements direct staircasing for all materials. For PEC, the behavior of `subpixel=False` in Tidy3D < 2.7 is now achieved through `subpixel=SubpixelSpec(pec=HeuristicPECStaircasing())`, while `subpixel=True` in Tidy3D < 2.7 is now achieved through `subpixel=SubpixelSpec(pec=Staircasing())`. The default is `subpixel=SubpixelSpec(pec=PECConformal())` for more accurate PEC modelling.
 - Lossless `Green2008` variant for crystalline silicon added to material library.
+- Users can now export `SimulationData` to MATLAB .mat files with the `to_mat` method
```
tylerflex: add period, consider `.mat` in backslash quotes

tests/test_components/test_viz.py
```diff
@@ -36,3 +36,48 @@ def test_0d_plot(center_z, len_collections):
...
+
+    # create source
+    source = td.PointDipole(
+        center=(0, 0, 0),
+        source_time=td.GaussianPulse(freq0=freq0, fwidth=freq0 / 10.0),
+        polarization="Ez",
+    )
+
+    # Simulation details
+    perBoundary = td.Boundary.periodic()
```
tylerflex: just FYI we use "snake case" for variable names in python in our code style, eg `per_boundary`.

tests/test_components/test_viz.py
```diff
@@ -36,3 +36,48 @@ def test_0d_plot(center_z, len_collections):
...
+        size=(0, 1, 1),
+        grid_spec=td.GridSpec.auto(min_steps_per_wvl=25),
+        structures=[],
+        sources=[source],
+        monitors=[],  # [monitor1],
+        run_time=120 / freq0,
+        boundary_spec=td.BoundarySpec(x=perBoundary, y=pmlBoundary, z=pmlBoundary),
+    )
+
+    # Create plot
```
tylerflex: this works, but I'd consider something even more simple to test that doesn't depend on matplotlib. For example (and double check my specifics)

```py
sim = Simulation(size=(0,1,1), ...)

# make PML box along the y axis
pml_box = sim._make_pml_box(pml_axis=1, pml_height=1, sign=1)

# should have infinite size in x
assert pml_box.size[0] == td.inf

# the same PML box when the size is non zero
sim = sim.updated_copy(size=(1,1,1))
pml_box = sim._make_pml_box(pml_axis=1, pml_height=1, sign=1)

# should have a non-infinite size
assert pml_box.size[0] != td.inf

```

this also checks both conditions

tests/test_data/test_sim_data.py
```diff
@@ -411,3 +412,22 @@ def test_loading_non_field_data():
...
+    testDataDict = testData.dict()
+
+    # Get the original shape of nested dict
+    originShape = get_dict_shape(testDataDict)
+
+    # Modify dict and get shape of dict
+    newData = replace_values_dict(testDataDict, None, [])
+    newShape = get_dict_shape(newData)
+
+    assert originShape == newShape
```
tylerflex: can we also check whether a None in the `test_data_dict` is properly replaced by `[]`?

tests/utils.py
```diff
@@ -1181,3 +1181,35 @@ def get_test_root_dir():
     """return the root folder of test code"""
 
     return Path(__file__).parent
+
+
+def get_dict_shape(d):
```
tylerflex: would prefer not using single character variable names (unless in an integrable, like `for i in range(...)`. Maybe let's rename `d` to something like `values`?)

tylerflex: please add type annotations for functions, in this instance would be something like
```py
def get_dict_shape(d: dict) -> dict:
```

tidy3d/components/data/sim_data.py
```diff
@@ -792,6 +792,28 @@ def plot_scalar_array(
...
+        """Output the SimulationData object as .mat MATLAB file
+
+        Parameters
+        ----------
+        fname : str
+            Full path to the output file. Should include .mat file extension
+
+        Example
+        -------
+        >>> simData.to_mat('/path/to/file/data.mat')
```
tylerflex: We test these example strings from time to time, and this one won't run as is because simData is not defined. So we can exclude it by adding
```
>>> simData.to_mat('/path/to/file/data.mat')  # doctest: +SKIP
```

tidy3d/components/mat.py
tylerflex, line 1: I'm thinking, since this file actually has nothing Matlab specific, maybe it's best somewhere else. I'm thinking we should move the contents to the `tidy3d/components/file_util.py`.

tidy3d/components/mat.py
```diff
@@ -0,0 +1,63 @@
...
+def replace_none_in_tuple_dict(t, searchValue, replaceValue):
+    """
+    Check if a tuple has a dictionary inside it.
+    If it does, carry on recursing through that dict replacing None values.
+
+    Parameters
+    ----------
+    t : tuple
+        The input tuple.
+    searchValue :
```
tylerflex: add types here too

tidy3d/components/mat.py
```diff
@@ -0,0 +1,63 @@
...
+        A string/int/object to look for in the dict. Included here to be passed forward to the recurive func call
+    replaceValue :
+        A string/int/object to replace the value in the dict. Included here to be passed forward to the recurive func call
+
+    Returns
```
tylerflex: not sure these returns are formatted properly. They're usually like

```
        Returns
        -------
        type
            description
```

(note indentation, also it will just be the type, not the variable name)

tidy3d/components/mat.py
```diff
@@ -0,0 +1,63 @@
...
+        if isinstance(value, dict):
+            replace_values_dict(value, searchValue, replaceValue)
+            newTup.append(value)
+        else:
+            return
+
+    return tuple(newTup)
+
+
+def replace_values_dict(d, searchValue, replaceValue):
```
tylerflex: It will be worth considering if any of these options would save us some code / complexity:

https://stackoverflow.com/questions/45335445/how-to-recursively-replace-dictionary-values-with-a-matching-key

but I think the implementation here is reasonable.

tidy3d/components/mat.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Returns
+    -------
+    dict: The dictionary with None values replaced by empty lists.
+    """
+    for key, value in d.items():
+        if isinstance(value, dict):
+            replace_values_dict(value, searchValue, replaceValue)
+        if isinstance(
+            value, tuple
```
tylerflex: might be worth checking `if isinstance(value, (tuple, list))` to handle the list case as well, if it ever pops up. (we usually have tuples only just to make things immutable but I think at times a list could sneak in)

tidy3d/components/mat.py
```diff
@@ -0,0 +1,63 @@
...
+    -------
+    newTup: A remade tuple with dicts that have had None values replaced by empty lists.
+
+    """
+    newTup = []
+    for value in t:
+        if isinstance(value, dict):
+            replace_values_dict(value, searchValue, replaceValue)
+            newTup.append(value)
+        else:
```
tylerflex: what happens if `t` is a tuple of tuples?

tidy3d/components/mat.py
tylerflex: One approach to consider is just making this generic `replace_values(values)`

if `values` is a dict, we just loop through the `items()` and recurse.

if `values` is a list or tuple, we just loop through it and recurse.

We can just pass `self.dict()` in originally.

Some pseudo-code.

```py
def replace_values(values, search_value, replace_value):

    if val == search_value:
        return replace_value

    elif isinstance(values, dict):
        return {key: replace_values(val, search_value, replace_value) for key, val in values.items()}

    elif isinstance(values, (tuple, list)):
        return type(values)((replace_values(val, search_value, replace_value) for val in val))

    return val

```

tidy3d/components/mat.py
```diff
@@ -0,0 +1,63 @@
...
+    dict: The dictionary with None values replaced by empty lists.
+    """
+    for key, value in d.items():
+        if isinstance(value, dict):
+            replace_values_dict(value, searchValue, replaceValue)
+        if isinstance(
+            value, tuple
+        ):  # Parts of the nested dict structure include tuples with more dicts
+            replace_none_in_tuple_dict(value, searchValue, replaceValue)
+        elif value is searchValue:
```
tylerflex: this works for `None` as `None is None` but I think for more generic find replace we want `==`. In python `is` checks if these are literally the same objects. `None` are all the same object technically. but say if 

```py
a = 1.0
b = 1.0
a is b
```
will give `False`

tests/utils.py
```diff
@@ -1181,3 +1181,35 @@ def get_test_root_dir():
...
+    if isinstance(d, dict):
+        return {k: get_dict_shape(d[k]) for k in d}
+
+    # Tuple of dicts, enter and continue iteration
+    elif isinstance(d, tuple):
+        for val in d:
+            if isinstance(val, dict):
+                return {k: get_dict_shape(val[k]) for k in val}
+            else:
+                # If val is just a regular tuple then replace with None
```
tylerflex: same comment about what if it's a nested tuple? might need to restructure this similar to the other function

tidy3d/components/mat.py
```diff
@@ -0,0 +1,63 @@
...
+            d[key] = replaceValue
+
+    return d
```
tylerflex: it seems this function modifies `d` in place already. do we need to return it? or would it be better to return a copy and leave the passed value unmodified?

tylerflex (changes requested):
Thanks @m-bone. This is looking very good overall. I had a few comments to address. A few general things to keep in mind:

* In the future, it's a bit easier to do a separate PR for every new feature, so we can easily isolate changes. for now it's ok.
* Please use snake case eg `this_is_my_variable` for variables instead of `camelCase`.
* please add type annotations to the call signatures for the functions arguments and return types.
* This PR will be merged into `pre/2.7` branch. (or actually maybe actually develop after we release 2.7.0). We can walk through it later once all of the code is finalized but there'll be some changes we'll need to do to make it compatible and get all of the code formatting tests to pass.
* There is a bit of code in our Tidy3dBaseModel._json_string method that looks for any mentions of `Infinity` and replaces with `"Infinity"` (in quotes). I wonder if it would make sense to try using this new find replace feature to do this more efficiently. eg 

```py
def _json(self, indent=INDENT, exclude_unset=False, **kwargs) -> str:
    self_dict = self.dict(indent=indent, exclude_unset=exclude_unset, **kwargs)
    self_dict = replace_values(self_dict, td.inf, "Infinity")
    self_dict = replace_values(self_dict, -td.inf, "-Infinity")
    return json.dumps(self_dict)
```

this function is used to insert all of the simulation info into files, so we'd need to make sure it has better performance than the existing implementation, but if so it could be useful to have.

tidy3d/components/data/sim_data.py
m-bone: Oh interesting! Added now

tidy3d/components/mat.py
m-bone: Good spot, yeah that makes more sense from a generic use point of view

tidy3d/components/mat.py
m-bone: You're right it's ambiguous. I was thinking people may want to still use an unmodified version, but if this primarily used to "correct" a variable or prep for output then they won't be bothered. Right now it's overwriting the original so it's neither neat code or a useful feature. What do you think is most useful for a generic func?

tylerflex: I'd say making a copy is probably the safest. I think my pseudo implementation does this.

tidy3d/components/file_util.py
```diff
@@ -28,3 +28,40 @@ def extract_gzip_file(input_gz_file, output_file):
     with gzip.open(input_gz_file, "rb") as file_in:
         with open(output_file, "wb") as file_out:
             shutil.copyfileobj(file_in, file_out)
+
+
+def replace_values(
+    values: list | dict | tuple, search_value: str | int | object, replace_value: str | int | object
```
tylerflex: we can just use `typing.Any` for these type annotations I think. since technically in this recursive definition they can be anything.

tidy3d/components/file_util.py
```diff
@@ -28,3 +28,40 @@ def extract_gzip_file(input_gz_file, output_file):
...
+) -> list | dict | tuple:
+    """
+    Recursively replace all None values with an empty list in a nested dictionary.
```
tylerflex: update docstring to reflect that this is generic (not just None -> []). Probably can delete the 2nd two sentences so it would just be something like
```
Create a copy of ``values`` where any elements equal to ``search_value`` are replaced by ``replace_value``.
```

m-bone:
Not sure how best to describe multiple types in documentation. I'm sort of using two different formats for Parameters and Returns but I couldn't find an existing example for multiple types.

    """
    Recursively replace all None values with an empty list in a nested dictionary.
    Can enter tuples and continue iterating through dicts stored inside.
    Could be modified to be a generic search and replace for nested dicts.

    Parameters
    ----------
    values :
        The input list/dict/tuple object to iterate through.
    searchValue :
        A string/int/object to look for in values.
    replaceValue :
        A string/int/object to replace the value in values.

    Returns
    -------
    list/dict/tuple
        values type object with searchValue terms replaced by replaceValue.
    """

I've redesigned the dict shape check as a generic `get_nested_shape` based on the change to `replace_values`.
Will experiment with _json next.

yaugenst-flex:
Hey @m-bone, we loosely follow [numpydoc style](https://numpydoc.readthedocs.io/en/latest/format.html). As for multiple types, use [Unions](https://docs.python.org/3/library/typing.html#typing.Union). There are examples of this scattered throughout the code base, here is one: https://github.com/flexcompute/tidy3d/blob/0571af5f390994b0efe92a4dca838db8c42c1650/tidy3d/plugins/autograd/functions.py#L175

tidy3d/components/file_util.py
```diff
@@ -28,3 +28,40 @@ def extract_gzip_file(input_gz_file, output_file):
...
+
+    Returns
+    -------
+    list/dict/tuple
+        values type object with searchValue terms replaced by replaceValue.
+    """
+    try:
+        if values == search_value:
+            return replace_value
+    except ValueError:  # Used to catch arrays which cannot be evaluated with ==
```
tylerflex: python is complaining because it wants `np.all()` or `np.any()` here? I think it wouldn't be unreasonable to just throw a `np.all(values == search_value)` in here. and not try/except. We probably dont want this failing silently.

tidy3d/components/file_util.py
tylerflex: Makes it easier for the docstrings and API reference too.

tests/utils.py
```diff
@@ -1181,3 +1180,39 @@ def get_test_root_dir():
     """return the root folder of test code"""
 
     return Path(__file__).parent
+
+
+def get_nested_shape(nested_obj: list | dict | tuple) -> list | dict | tuple:
```
tylerflex: we can just use typing.Any for the types here.

tests/test_data/test_sim_data.py
tylerflex, line 1: let's also test the `to_mat` function directly. using a `SimulationData` loaded from `.from_file()` from the `tests/data` perhaps?

tidy3d/components/data/sim_data.py
```diff
@@ -792,6 +792,28 @@ def plot_scalar_array(
...
+        Parameters
+        ----------
+        fname : str
+            Full path to the output file. Should include .mat file extension
+
+        Example
+        -------
+        >>> simData.to_mat('/path/to/file/data.mat') # doctest: +SKIP
+        """
+        # Get SimData object as dictionary
```
tylerflex: let's throw a check here to make sure the extension of `fname` is `.mat`. Otherwise the user probably doesn't want this method and they should use `to_file()`.

tidy3d/components/data/sim_data.py
```diff
@@ -792,6 +792,28 @@ def plot_scalar_array(
 
         return ax
 
+    def to_mat(self, fname):
```
tylerflex: type annotation and could you put the variable stuff in the docstring (eg `SimulationData`, `.mat`) in double backslash quotes? (side note: I forget this all the time, it's why we have a 2nd or 3rd pair of eyes on these PRs).

tylerflex: Also, after reading through, I wonder if we should rename this `to_mat_file()` just to be a bit more explicit. Sorry to switch it up last minute, but what do you think?

tidy3d/components/data/sim_data.py
```diff
@@ -792,6 +792,28 @@ def plot_scalar_array(
...
+        """
+        # Get SimData object as dictionary
+        simDict = self.dict()
+
+        # Remove NoneType values from dict
+        # Built from theory discussed in https://github.com/scipy/scipy/issues/3488
+        modifiedSimDict = replace_values(simDict, None, [])
+
+        # Savemat hasn't actually implemented long_field_names, but set as true in case someone notices the bug
+        savemat(fname, modifiedSimDict, long_field_names=True)
```
tylerflex: because this is somewhat of an experimental feature, I would say we might want to do something like
```py
try:
    savemat(...)
except Exception as e:
    raise ValueError(f"Could not save supplied 'SimulationData' to file. As this is an experimental feature, we may not be able to support the contents of your dataset. If you receive this error, please feel free to raise an issue on our front end repository so we can investigate.") from e
```

which will give a nicer error and then provide the original stack exchange (because of that `from e` bit).

tylerflex (commented):
Hey @m-bone Thanks for the revisions. Had a few other comments for now. we can discuss more on Monday and then get this PR into shape to merge once everything is addressed. Thanks and enjoy your weekend!

tidy3d/components/file_util.py
yaugenst-flex: Just a heads up but the `type1 | type2 | type3` type hint syntax is only supported in Python >=3.10, and we have to support 3.9, so you'll have to use a `Union` instead.

tests/test_data/test_sim_data.py
m-bone: Sure, is it enough to just check that `to_mat` doesn't throw an error? Scipy appears to create a `.mat` file even when it fails, and these files are binaries we can't read back in easily, so I'm not sure what's sensible to assert

tylerflex: yea I think just running through is probably fine. Would check at the very least that `scipy` didnt change anything regarding `savemat` that causes immediate failure. 

When it fails, does the file have any data in it?

tidy3d/components/data/sim_data.py
tylerflex: We could also consider making this `savemat`, just like `scipy`.

tylerflex: speaking of this, it could be useful to just accept also some `**kwargs` in the function, which can then be passed to `scipy.io.savemat` if the user wants. For example

```py
def savemat(file_name, **kwargs):
    
     if "m_dict" in kwargs:
        raise ValueError("m_dict" is automatically determined by the object, can't pass to `savemat`.)

     m_dict = ...

     return scipy.io.savemat(file_name, m_dict, **kwargs)

```

CHANGELOG.md
tylerflex, line 1: seems like a bunch of changelog items got duplicated here, probably in the fix of the merge conflicts?

tests/data/test_mat_file.mat
tylerflex, line 1: this file was in the gitignore but got included still. probably we can just remove it.

It occurred to me, we could try using  `tmp_path` in the test. See this [example](https://github.com/flexcompute/tidy3d/blob/develop/tests/test_components/test_base.py#L36-L47)

we can make a temporary file (with a `'.mat'` extension added) and do the tests, and the file will not be saved. This way saves us from having to deal with the file both here and in the gitignore. Let's try that.

tests/test_data/test_sim_data.py
```diff
@@ -414,3 +416,56 @@ def test_loading_non_field_data():
...
+    original_shape = get_nested_shape(test_list)
+
+    # Modify list and get shape
+    new_list = replace_values(test_list, 3.14, [])
+    new_shape = get_nested_shape(new_list)
+
+    assert original_shape == new_shape and new_list[3] == [] and new_list[2][5][0] == []
+
+
+def test_to_mat_file():
```
tylerflex: for reference, would look like this with temporary file
```
def test_to_mat_file(tmp_path):
    ....
    path = str(tmp_path / "test.mat")
    sim_data.to_mat_file(path)
```

tidy3d/components/data/sim_data.py
```diff
@@ -1003,3 +1005,44 @@ def get_adjoint_data(self, structure_index: int, data_type: str) -> MonitorDataT
 
         monitor_name = Structure.get_monitor_name(index=structure_index, data_type=data_type)
         return self[monitor_name]
+
+    def to_mat_file(self, fname: str, **kwargs):
+        """Output the ``SimulationData`` object as ``.mat`` MATLAB file
```
tylerflex: period at end of sentence. and in the parameter descriptions too.

.gitignore
tylerflex, line 1: for reference, probably can remove this change if we go with temporary file.

tylerflex (changes requested):
Thanks @m-bone I just took another read through and had a few minor comments. But I think hopefully it's clear how to rebase after these are fixed up and should be ready to go very soon!

CHANGELOG.md
m-bone: Is this the duped "Changed" tags? I've updated where it said `to_mat`

tylerflex (approved):

tylerflex (approved):

PR #1747, opened 2024-06-07
fix heat plotting, missing changelog items

dbochkov-flexcompute:
- fix error in heat plotting due to some matplotlib deprecations
- add missing changelog items


tidy3d/components/heat/simulation.py
```diff
@@ -892,7 +892,7 @@ def _get_structure_source_plot_params(
             delta_rate = rate - source_min
             delta_rate_max = source_max - source_min + 1e-5
             rate_fraction = delta_rate / delta_rate_max
-            cmap = cm.get_cmap(HEAT_SOURCE_CMAP)
+            cmap = colormaps[HEAT_SOURCE_CMAP]
```
momchil-flex: Does this work with both old and new versions of matplotlib (i.e. all that we officially support in our deps)?

dbochkov-flexcompute: it's supported from matplotlib==3.5.0. We actually had this requirement for some other reasons https://github.com/flexcompute/tidy3d/commit/1b5b46c7a7f97c37dc3f71455a8fef9c293e140d, but not anymore I think https://github.com/flexcompute/tidy3d/blob/e77b8bc46917de50c595d3e017b5eb8468c98967/pyproject.toml#L32

PR #1749, opened 2024-06-07
support for FieldData differentiation

tylerflex:
WIP, intensity and flux dont work yet as they need `interp`

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -6,16 +6,17 @@ import warnings
 from abc import ABC
 from typing import Any, Callable, Dict, List, Tuple, Union
 
-import numpy as np
+import autograd.numpy as np
 import pydantic.v1 as pd
 import xarray as xr
 from pandas import DataFrame
 
-from ...constants import C_0, ETA_0, MICROMETER
+from ...constants import C_0, ETA_0, MICROMETER, MU_0
 from ...exceptions import DataError, SetupError, Tidy3dNotImplementedError, ValidationError
 from ...log import log
 from ..base import TYPE_TAG_STR, cached_property, skip_if_fields_missing
 from ..base_sim.data.monitor_data import AbstractMonitorData
+from ..geometry.base import Box
 from ..grid.grid import Coords, Grid
 from ..medium import Medium, MediumType
 from ..monitor import (
@@ -33,7 +34,16 @@ from ..monitor import (
     MonitorType,
     PermittivityMonitor,
 )
-from ..source import CustomFieldSource, GaussianPulse, ModeSource, PlaneWave, Source, SourceTimeType
+from ..source import (
+    CustomCurrentSource,
+    CustomFieldSource,
+    GaussianPulse,
+    ModeSource,
+    PlaneWave,
+    PointDipole,
+    Source,
+    SourceTimeType,
+)
 from ..types import (
     ArrayFloat1D,
     ArrayFloat2D,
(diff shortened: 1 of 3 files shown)


tylerflex:
This works when just messing with raw field components though. See https://github.com/flexcompute/tidy3d-notebooks/pull/118

tidy3d/components/data/monitor_data.py
```diff
@@ -996,6 +1009,111 @@ def to_source(
             field_dataset=dataset, source_time=source_time, center=center, size=size, **kwargs
         )
 
+    def make_adjoint_sources(
+        self, dataset_names: list[str]
+    ) -> List[Union[CustomCurrentSource, PointDipole]]:
+        """Converts a :class:`.FieldData` to a list of adjoint current or point sources."""
+
+        if np.allclose(np.array(self.monitor.size), np.zeros(3)):
```
yaugenst-flex: This works without `np.zeros()` and casting, i.e. `np.allclose(self.monitor.size, 0)`

tidy3d/components/data/monitor_data.py
```diff
@@ -996,6 +1009,111 @@ def to_source(
...
+
+        for polarization, field_component in self.field_components.items():
+            if field_component is None:
+                continue
+
+            for freq0 in field_component.coords["f"]:
+                omega0 = 2 * np.pi * freq0
+                scaling_factor = 1 / (MU_0 * omega0)
+
+                forward_amp = complex(np.squeeze(field_component.sel(f=freq0).values))
```
yaugenst-flex: Seems like `complex()` casting here wouldn't be necessary? It goes directly into `np.angle`, i.e. if these values were real before the cast the phase would still be zero, and `np.angle` accepts real arguments.

tidy3d/components/data/monitor_data.py
```diff
@@ -996,6 +1009,111 @@ def to_source(
...
+
+        sources = []
+
+        # Define source geometry based on coordinates in the data
+        data_mins = []
+        data_maxs = []
+
+        def shift_value(coords) -> float:
+            """How much to shift the geometry by along a dimension (only if > 1D)."""
+            return 1e-5 if len(coords) > 1 else 0
```
yaugenst-flex: Should this `1e-5` be set somewhere else maybe as a global constant?

yaugenst-flex (approved):
LGTM, some minor comments but they don't really need fixing, up to you.

tidy3d/components/data/monitor_data.py
tylerflex: yea maybe.. I dont remember where this came from exactly. @momchil-flex do you remember? I guess we just wanted to make sure that the source was always to the same side of the coordinate? any updates since then in the meshing so I dont need to worry about this as much?

tylerflex: but I can move it to a global constant either way.

tidy3d/components/data/monitor_data.py
tylerflex: I think it was partially to cast it out of a `np.ndarray`. 

How about I use `self.get_amplitude(field_component.sel(f=freq0))`, which is making use of the helper function I have in the other 2 adjoint source methods for consistency?

tidy3d/components/data/monitor_data.py
tylerflex: thanks

tidy3d/components/data/monitor_data.py
momchil-flex: I don't really remember either... We should have left more comments :)

tidy3d/components/data/monitor_data.py
yaugenst-flex: Ah I see, yeah then I guess that would make sense :+1:

PR #1751, opened 2024-06-08
fix deepcopy bug in autograd

tylerflex:
Long story short, seems when we `deepcopy()` a `td.DataArray` containing autograd tracers, bad things can happen:
1. The `run_async` tracing ignores anything but the last `SimulationData` in the batch.
2. The gradient information can get corrupted..

This PR does two things:
1. Changes default behavior of `deepcopy(td.DataArray)` to do a shallow copy if there are tracers present.
2. Properly passes the `deep` kwarg in `Tidy3dBaseModel.copy()` and related methods.


yaugenst-flex (approved):
Amazing...

PR #1752, opened 2024-06-10
Avoiding double upload when using Job.run

momchil-flex:
It seems that some of the container reorganizations made the `web.Job` workflow upload a task twice. The underlying reason is that `Job.task_id_cached` is not set internally when the first upload happens - it is currently only used for file operations. However, `Job.task_id` assumes that `task_id_cached` *had* actually been set, and if not, uploads the task again. So in `Job.run` first `web.upload` is called, then `web.run` with `self.task_id`, which does a second upload.

This fix works but is probably not how we want to handle it? I'm leaving it to @tylerflex though as you should know better what the internals here are.

### DIFF
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -215,7 +215,13 @@ class Job(WebContainer):
         # upload kwargs with all fields except task_id
         self_dict = self.dict()
         upload_kwargs = {key: self_dict.get(key) for key in self._upload_fields}
-        return web.upload(**upload_kwargs)
+        # set the cached task_id
+        self.Config.frozen = False
+        self.Config.allow_mutation = True
+        self.task_id_cached = web.upload(**upload_kwargs)
+        self.Config.allow_mutation = False
+        self.Config.frozen = True
+        return self.task_id
 
     def get_info(self) -> TaskInfo:
         """Return information about a :class:`Job`.
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -209,19 +209,19 @@ class Job(WebContainer):
         """The task ID for this ``Job``. Uploads the ``Job`` if it hasn't already been uploaded."""
         if self.task_id_cached:
             return self.task_id_cached
-        return self.upload()
+        return self._upload()
 
-    def upload(self) -> TaskId:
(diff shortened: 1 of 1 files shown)


tylerflex:
@momchil-flex I just pushed a commit that I think handles it. Can you double check if this makes sense to you?

tidy3d/web/api/container.py
```diff
@@ -209,13 +209,19 @@ def task_id(self) -> TaskId:
...
+    def _upload(self) -> TaskId:
+        """Upload this job and return the task ID for handling."""
         # upload kwargs with all fields except task_id
         self_dict = self.dict()
         upload_kwargs = {key: self_dict.get(key) for key in self._upload_fields}
-        return web.upload(**upload_kwargs)
+        task_id = web.upload(**upload_kwargs)
+        return task_id
+
+    def upload(self) -> None:
```
momchil-flex, line 222: It should probably still return the task_id?

momchil-flex (commented):
I think this works yeah. I can't approve because it's my own PR. :)

tidy3d/web/api/container.py
tylerflex: I dont think it ever does for `Job.upload()`.

tylerflex: the user can always grab `job.task_id`, but the task id isn't returned. at least that's how it was before and I think generally makes sense?

tylerflex: oh I guess there never was a `Job.upload` before 2.7.. and the old version I had implemented did return task_id. But I think in general it should not. since it's already in the job?

momchil-flex: No, previously it would return it, and I think that makes sense since `web.upload` returns it.

https://github.com/flexcompute/tidy3d/blob/4918797e504db4677c064f682a9335c48a538471/tidy3d/web/api/container.py#L218

tylerflex: yea I think this old version was not working quite properly though. so this was a bit of a hack. The new approach seems reasonable to me and we can choose to return it or not. If you want to return task_id, it shouldn't affect  anything and I don't care strongly one way or the other, feel free to do whatever before squashing and merging.

PR #1756, opened 2024-06-10
Introduced RF material library to microwave plugins

QimingFlex:
Hey @weiliangjin2021 , I mirrored the optical material library to create an RF microwave material library and included three materials as examples. Could you please provide some feedback so I can refine the format of the RF material library?

Example:
`from tidy3d import rf_material_library
FR4 = rf_material_library['FR4']
print(FR4.medium)`


tidy3d/__init__.py
```diff
@@ -256,6 +256,7 @@
 # get material `mat` and variant `var` as `material_library[mat][var]`
 from .material_library.material_library import material_library
 from .material_library.parametric_materials import Graphene
+from .plugins.microwave.rf_material_library import rf_material_library
```
weiliangjin2021: no need to expose it to tidy3d namespace. Instead, add it to `plugins/microwave/__init__.py`

weiliangjin2021 (commented):
Thanks Qiming, this looks great! Two main comments:

- A material can contain several variants. E.g. there can be several difference types of FR4 depending on the data source. Will be good to still follow this format?
- It would be better to define those Abstract classes in tidy3d rather than in plugin. In that case, both RF material library and photonics material library items can subclass from them. This helps reduce redundant codes.

QimingFlex:
>     * A material can contain several variants. E.g. there can be several difference types of FR4 depending on the data source. Will be good to still follow this format?
Exactly, I will follow the optical library to include these variants, including the "design" and "process" variants we've discussed for the Rogers laminates.
>     * It would be better to define those Abstract classes in tidy3d rather than in plugin. In that case, both RF material library and photonics material library items can subclass from them. This helps reduce redundant codes.
Sounds good, I will handle it properly.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -0,0 +1,68 @@
...
+    Rogers3010=MaterialItem(
+        medium=PoleResidue(
+            eps_inf=2.080628548409516,
+            poles=[
+                (
+                    (-14424652487575.385 + 0j),
+                    (40799026234878.12 - 0j),
+                ),
+            ],
+            frequency_range=(10e9, 10e9),
```
weiliangjin2021: this looks off: empty frequency_range?

QimingFlex: The parameters are fitted at a single frequency of 10 GHz. Shall we set it as `frequency = 10e9`?

weiliangjin2021: well, we shouldn't include material in material library that is only valid at a single frequency, as most simulation is broadband in FDTD. And you don't really need fitting for single frequency. If the reference doesn't provide a frequency range, try to come up with a frequency range.

QimingFlex: Yeah, I just did some search, for microwave circuits the frequency typically ranges from 1GHz to 30GHz, I believe this could be a good range to extract parameters.

weiliangjin2021: sounds good!

QimingFlex: The range of 1-30 GHz is typical for most Rogers laminates. For other materials targeted at lower frequencies, like FR-4 in the 1-2 GHz range, we can handle them accordingly.

weiliangjin2021: in that case, you can just use frequency ranges for those different materials. For the same material, we can also provide different variants targeting at different frequency range.

QimingFlex:
I introduced variants to the materials. Now, the RF materials follow a similar structure to the optical ones.
Example:
```python
from tidy3d import rf_material_library

Rogers3010 = rf_material_library['Rogers3010']['design']

print(Rogers3010)

File: tidy3d/material_library/material_library.py [hunk 3]
```diff
@@ -10,6 +10,7 @@
 from ..components.types import Axis
 from ..exceptions import SetupError
 from ..log import log
+from ..plugins.microwave.rf_material_reference import RF_ReferenceData
```
weiliangjin2021: plugin is not guaranteed to be installed, so the main shouldn't import anything from plugins. Maybe we don't need `RF_ReferenceData`, but just move all its field to `ReferenceData`? All fields are optional in `Referencedata`, so we just fill the relevant fields. The RF reference data can stay in `rf_material_reference.py`.

weiliangjin2021 (commented):
There are lots of formatting going on. Which `ruff` version are you using?

weiliangjin2021:
(also temporarily change the merge branch to `develop`, until we have `pre/2.8` branch)

weiliangjin2021:
The formatting here makes no different to `ruff`. I guess @QimingFlex  format it with `black`? How does it work now with this new formatting setting @tylerflex ?

tylerflex:
@yaugenst recently updated the notion guide here for that. 
https://www.notion.so/flexcompute/Tidy3D-Developer-Guide-60d5df306c2f4f29bea8c3b401ba6382?pvs=4#1737de671e154f96ae55690d2c96147b

I think just `ruff format .` should work

tylerflex:
Also, recommend installing the pre-commit hooks to get this to happen automatically. I only recently installed it and it saves me a lot of time now `pre-commit install` from `tidy3d/`

weiliangjin2021:
> @yaugenst recently updated the notion guide here for that. https://www.notion.so/flexcompute/Tidy3D-Developer-Guide-60d5df306c2f4f29bea8c3b401ba6382?pvs=4#1737de671e154f96ae55690d2c96147b
> 
> I think just `ruff format .` should work

When running `ruff format .` on develop branch or this branch, it doesn't modify any files, while `black` will format one of the branch.

tidy3d/material_library/material_library.py
QimingFlex: Sounds good!

QimingFlex:
Yeah, I used both `ruff --fix .` and `ruff format .`. My ruff is version 0.3.5.

tylerflex:
hm, I'm sorry, I actually haven't 100% figured out the best option. For me the pre-commit has been working well, without GitHub CI complaining, so maybe try installing that?

weiliangjin2021:
> hm, I'm sorry, I actually haven't 100% figured out the best option. For me the pre-commit has been working well, without GitHub CI complaining, so maybe try installing that?

My understanding is that `ruff` is not as strict as `black`, so it cannot discern the formatting difference between this and develop branch. In that case, pre-commit shouldn't help either, and GitHub CI will not complain about the formatting occurs in this branch, as both of them are based on `ruff`.

I guess just revert all formatting in this PR?

yaugenst-flex:
Hey all! The correct notion page is this one: https://www.notion.so/flexcompute/Tidy3D-Developer-Guide-23ceee49660e42fca06484bfcaa96b5c?pvs=4

I think everyone has access to that.

Regarding ruff: Everything on develop was formatted using ruff 0.4.8 - this is also pinned as a dependency, and is what the pre-commit hook uses. If you use anything else, i.e. older ruff versions or black, the formatting will be different and it will _not_ pass in CI.

In general it's a good idea to keep a dev environment for each branch, and keep it up to date with the poetry lock file on that branch. I think @daquinteroflex set up the poetry packaging and wrote a guide on how to set up the env [here](https://docs.flexcompute.com/projects/tidy3d/en/stable/development/index.html).

Currently it's probably a bit confusing still how to set everything up, we should write a simple guide for this and communicate it to everyone...

daquinteroflex:
Hello! Yeah, agree with Yannick on what he mentions.

 We should make a simpler version that is more straightforward and the one you've written on notion looks great!  So yes, the idea of using the poetry environments guarantees we'll all have the same dependencies across our computers between PRs. Because it changed in the latest ruff update, that's why this formatting in this PR is failing as it was an older version. The `pre-commit` should streamline all of this for you without having to think about it after `poetry run pre-commit install`. 
 
 Happy to help you @QimingFlex if you need a hand too

weiliangjin2021:
All look good! One remaining item is to add it to the doc. The photonics material library [doc page](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/material_library.html) is generated [here](https://github.com/flexcompute/tidy3d/blob/develop/docs/generate_doc.py).

QimingFlex:
> All look good! One remaining item is to add it to the doc. The photonics material library [doc page](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/material_library.html) is generated [here](https://github.com/flexcompute/tidy3d/blob/develop/docs/generate_doc.py).

Sure, I'm working on this!

QimingFlex:
@weiliangjin2021 generated the documentation and placed it under the folder `tidy3d/api`. Am I doing this correctly? How is the RST file transformed into a website?

weiliangjin2021:
> @weiliangjin2021 generated the documentation and placed it under the folder `tidy3d/api`. Am I doing this correctly? How is the RST file transformed into a website?

Also need to include this RST in index.rst etc. E.g. photonics material library appears also [here](https://github.com/flexcompute/tidy3d/blob/develop/docs/api/mediums.rst) and [here](https://github.com/flexcompute/tidy3d/blob/develop/docs/api/index.rst). How to build it into a website locally can be found [here](https://docs.flexcompute.com/projects/tidy3d/en/latest/development/documentation.html).

weiliangjin2021 (approved):
All look good to me. Just rebase to pre/2.8.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Users can now export `SimulationData` to MATLAB `.mat` files with the `to_mat_file` method.
+- Introduce RF material library including several commonly used RF PCB materials.
```
tylerflex: would be nice to mention where users can import this from.

docs/generate_doc.py
```diff
@@ -227,8 +228,177 @@ def num2str(num):
                     f.write("\n")
 
 
+def generate_rf_material_library_doc():
```
tylerflex, line 231: is this mostly copied from the `genereate_material_library_doc`? can we write in a way that we dont need to re-implement this in the future?

tylerflex (approved):
Thanks @QimingFlex ! Just a couple comments from me. Overall looks good!

CHANGELOG.md
QimingFlex: Updated.

docs/generate_doc.py
QimingFlex: Yeah, mostly from generate_material_library_doc. I'll keep this in mind and reduce the redundancy later.

PR #1759, opened 2024-06-12
shallow copy all autograd Box used in tidy3d

tylerflex:
This is a more elegant and general solution to fixing the deepcopy bug in autograd.

It overrides the behavior of `deepcopy(autograd.Box)` in tidy3d so that it returns a shallow copy, tested on the notebooks with previous issues and it mitigates them. Also handles `updated_copy()` in simulation / pre-processing.

Still a bit of black magic, but is nice because it avoids
* A new autograd fork
* Shallow copies everywhere in tidy3d

Don't mind waiting for 2.7.1, only caveat with not having this is that an updated copy of a tidy3d component, eg `polyslab.updated_copy(slab_bounds=(1,1))` will ruin the gradient, so it could be an annoying bug.

Note: it doesn't solve the autograd bug generally, so something like
```
def objective(params):
    p = deepcopy(params)
```
outside of tidy3d will still cause issues, but this is sort of expected.

### DIFF
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -1,5 +1,6 @@
 # test autograd integration into tidy3d
 
+import copy
 import cProfile
 import typing
 from importlib import reload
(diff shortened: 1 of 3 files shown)


yaugenst-flex (approved):
Yeah this is a pretty neat fix, if not a bit magic :smile:

This "obviously" works fine for our use case, I just wonder if in principle the more correct way to do this would be by overriding both `__copy__` and `__deepcopy__` and make both of those use autograd primitives (that we are actually free to  define here), so something like:
```python
copy = primitive(copy.copy)
deepcopy = primitive(copy.deepcopy)

defvjp(copy, lambda ans, x: lambda g: copy(g))
defvjp(deepcopy, lambda ans, x, memo: lambda g: deepcopy(g, memo))

Box.__copy__ = lambda v: copy(v)
Box.__deepcopy__ = lambda v, memo: deepcopy(v, memo)
```

This way we get correct copying behavior and also preserve the graph.

tylerflex:
@yaugenst-flex I think this might be un-necessary. The graph is computed by running the tracers through the function. if something is shallow copied, the `_node.parents` info is unchanged, so I dont think we need to define vjps.  I'm also not 100% sure if the vjp of deepcopy should be a deepcopy, that might be unnecessary vs just returning self. I can test this though

tylerflex:
These changes were  included in [ba30233](https://github.com/flexcompute/tidy3d/pull/1759/commits/ba302339065b3f94db6a9129938ec3f0d87975b4) and testing on the notebooks seems to work just fine. I'm ok with this as it preserves the intended behavior of deepcopy.

The only questions in my mind are:
1. Do we need to `copy(g)` / `deepcopy(g)` in the vjp? I'm not sure it matters either way, so for performance reasons we could just `return g`?
2. Do we need to define a `vjp` for `copy`? it would work without it.

PR #1760, opened 2024-06-12
🗻 autograd: server-side gradient processing 

tylerflex:
currently a WIP, needs backend bits and pieces.

For now, organizes the autograd pipeline into a local and a remote pathway, depending on a kwarg passed to `run()`.

The local one operates the same as before, the remote one has the basics of what we need to implement server side. Specifically, I would need the following Web API implemented.

```py
sim_data = web.run(sim_combined, simulation_type="autograd_fwd")
```
on the `sim_combined` (sim with original + grad monitors) should split the data into `original` and `fwd` and store the `fwd` associated with this task ID. Then return just the original `SimulationData`.

```py
sim_data = web.run(sim_adj, simulation_type="autograd_bwd", parent_tasks=[task_id_fwd])
```

should run the adjoint simulation `sim_adj` and compute the gradient information combined with the cached forward `SimulationData` associated with `task_id_fwd `

```py
fields_vjp = download_vjp_fields_autograd(task_id_adj)
```

should download the result of the adjoint post processing associated with the task ID from the `autograd_bwd` call above.

I will work to make exactly what Neds to happen in each of these tasks explicit. it's just reorganizing bits of the existing front end code.

### DIFF
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -11,10 +11,7 @@ import matplotlib.pylab as plt
 import numpy as np
 import pytest
 import tidy3d as td
-from tidy3d.web import run_async
-from tidy3d.plugins.polyslab import ComplexPolySlab
-from tidy3d.web import Job, run_async
-from tidy3d.web.api.autograd.autograd import run
+from tidy3d.web.api.autograd.autograd import run, run_async
 
 from ..utils import SIM_FULL, AssertLogLevel, run_emulated
 
@@ -116,6 +113,8 @@ _run_was_emulated = [False]
 def use_emulated_run(monkeypatch):
     """If this fixture is used, the `tests.utils.run_emulated` function is used for simulation."""
 
+    from tidy3d.web.api.container import Job
+
     if TEST_MODE in ("pipeline", "speed"):
         VJP = "VJP"
         task_id_fwd = "task_fwd"
@@ -198,7 +197,9 @@ def use_emulated_run(monkeypatch):
         monkeypatch.setattr(Job, "run", emulated_job_run)
         monkeypatch.setattr(Job, "task_id", task_id_fwd)
 
-        reload(webapi)
+        import tidy3d
+
+        reload(tidy3d.web.api.autograd.autograd)
 
         _run_was_emulated[0] = True
 
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -113,24 +113,19 @@ _run_was_emulated = [False]
 def use_emulated_run(monkeypatch):
     """If this fixture is used, the `tests.utils.run_emulated` function is used for simulation."""
 
+    import tidy3d
     from tidy3d.web.api.container import Job
 
     if TEST_MODE in ("pipeline", "speed"):
         VJP = "VJP"
         task_id_fwd = "task_fwd"
         task_id_bwd = "task_bwd"
+
         cache = {}
 
-        monkeypatch.setattr(
-            "tidy3d.web.api.autograd.autograd.get_fwd_sim_data",
-            lambda task_id: cache[task_id_fwd][AUX_KEY_SIM_DATA_FWD],
-        )
-        monkeypatch.setattr(
-            "tidy3d.web.api.autograd.autograd.get_vjp_traced_fields",
-            lambda task_id: cache[task_id_bwd][VJP],
-        )
-
         import tidy3d.web.api.webapi as webapi
+
+        # reload(tidy3d.web.api.autograd.autograd)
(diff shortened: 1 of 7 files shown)


tidy3d/web/api/autograd/autograd.py
```diff
@@ -27,6 +29,9 @@
 
 MAX_NUM_TRACED_STRUCTURES = 500
 
+# default value for whether to do local gradient calculation (True) or server side (False)
+LOCAL_GRADIENT = True
```
momchil-flex, line 37: We will set it to False eventually?

tests/test_components/test_autograd.py
```diff
@@ -114,15 +115,84 @@ def use_emulated_run(monkeypatch):
...
+
+                aux_data = {}
+                _ = postprocess_fwd(
+                    sim_data_combined=sim_data_combined,
+                    sim_original=sim_original,
+                    aux_data=aux_data,
+                )
+
+                sim_data_orig = aux_data[AUX_KEY_SIM_DATA_ORIGINAL]
+                sim_data_fwd = aux_data[AUX_KEY_SIM_DATA_FWD]
```
momchil-flex: In adjoint, there's a call for this like

```
        sim_data, sim_data_store = JaxSimulationData.split_fwd_sim_data(
            sim_data=sim_data, jax_info=jax_info
        )
```

momchil-flex (commented):
[This](https://github.com/flexcompute/tidy3d-core/blob/75631b7cfea0e0451920fb804d5f2478af2f98cd/tidy3d_backend/run.py#L340-L363) is the gist of the adjoint handling on the backend, apart from some things like ensuring the right files are downloaded which happen elsewhere. It's a bit hard for me to orient myself here yet but in the end it should reduce to something similar.

tidy3d/web/api/autograd/autograd.py
tylerflex: Yea, setting to True until this works :)

tests/test_components/test_autograd.py
tylerflex: yea, I guess I could refactor this into it's own method of `Simulation` if it would make things easier.

tylerflex:
> [This](https://github.com/flexcompute/tidy3d-core/blob/75631b7cfea0e0451920fb804d5f2478af2f98cd/tidy3d_backend/run.py#L340-L363) is the gist of the adjoint handling on the backend, apart from some things like ensuring the right files are downloaded which happen elsewhere. It's a bit hard for me to orient myself here yet but in the end it should reduce to something similar.

yea I tried to emulate that a bit in this part of the emulated `Job.run()`

https://github.com/flexcompute/tidy3d/blob/81c554465c4a32b1d625b865ecbb1d51e3838911/tests/test_components/test_autograd.py#L145-L187

tylerflex:
@momchil-flex I want to make sure this doesn't fall through the cracks for 2.8 since probably it will be highly requested.. do you think you'll find time to work on the backend bits? is there anything I can do to make it easier?

momchil-flex:
> @momchil-flex I want to make sure this doesn't fall through the cracks for 2.8 since probably it will be highly requested.. do you think you'll find time to work on the backend bits? is there anything I can do to make it easier?

I'll give it a try tomorrow. Meanwhile do you want to try to introduce the `split_fwd_sim_data` method from the comment above and remove the "bit of a hack" comment in the code about this?

tylerflex:
> Meanwhile do you want to try to introduce the split_fwd_sim_data method from the comment above and remove the "bit of a hack" comment in the code about this?

Thanks. I won't be able to get to this until tomorrow eastern US time. I guess just let me know if you've already started so we don't conflict.

tylerflex:
@momchil-flex I pushed this [commit](https://github.com/flexcompute/tidy3d/pull/1760/commits/ca401adebfa3b381d10c32d25eb3e6921d0d538e) that reorganizes things a bit.

Adds:
```py
sim_data_original, sim_data_fwd = sim_data_combined.split_original_fwd(num_mnts_original=num_mnts_original)
```
although I'm not sure you'll need this directly as it's done in `postprocess_fwd` ([code](https://github.com/flexcompute/tidy3d/blob/tyler/autograd_/serverside/tidy3d/web/api/autograd/autograd.py#L466-L487)).

Removes the "kind of a hack" comment, basically needed a way to figure out what were original vs gradient monitors. But I restructured stuff so now one has access to the original sim directly.

A few other small improvements and cleanup.

Should be set. Note that the async bits are not done in this yet, maybe we can start with regular `web.run()`

tylerflex:
@momchil-flex I reworked this according to our comments and also rebased against develop. could you revisit when you get a chance? thanks

tylerflex:
done in #1824

PR #1765, opened 2024-06-14
:wrench: Include invdes module in API docs

daquinteroflex:
Pretty urgent pre-2.7 release patch to include the `invdes` module on the API docs. Possibly my bad in not catching this in review sorry!

Doing a thorough upgrade on docs https://github.com/flexcompute/tidy3d/pull/1745 and caught this in time

### DIFF
--- a/docs/api/plugins/index.rst
+++ b/docs/api/plugins/index.rst
@@ -11,8 +11,9 @@ Plugins
     resonance
     autograd
     adjoint
-    waveguide
     design
+    invdes
+    waveguide
     microwave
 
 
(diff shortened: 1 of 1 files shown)


momchil-flex (approved):
Looks good to me. For docs, things are not that urgent I guess since even if we had released, we could always update or `latest` build with this fix.

Note that you should remove the submodule update.

daquinteroflex:
Sounds good! I've just updated the submodule state to the last `tidy3d-notebooks/develop` for the tests to pass too

PR #1768, opened 2024-06-17
🗻 autograd: geometry group

tylerflex:
See example 
https://github.com/flexcompute/tidy3d-notebooks/blob/b33c5304deec54cbcf68a6ada8b6ca489e9a6123/Autograd7Metalens.ipynb


Untitled.ipynb
yaugenst-flex, line 1: This probably slipped in?

tidy3d/web/api/autograd/autograd.py
```diff
@@ -58,7 +58,7 @@ def is_valid_for_autograd(simulation: td.Simulation) -> bool:
         return False
 
     # if too many structures, raise an error
-    structure_indices = [i for key, i, *_ in traced_fields.keys() if key == "structures"]
+    structure_indices = {i for key, i, *_ in traced_fields.keys() if key == "structures"}
```
yaugenst-flex, line 63: Probably I don't understand this - so indices will always refer to the same object, even if the index appears multiple times?

yaugenst-flex (commented):
Looks good!

tylerflex:
hey @yaugenst-flex when you get a chance, could you look again with the new added changes to add a `components/autograd/` directory and also passing things in the `DerivativeInfo`. Thanks

Untitled.ipynb
tylerflex: yea, removed

tidy3d/web/api/autograd/autograd.py
tylerflex: yea, so the field paths can look something like
```
('structures', 0, 'medium', 'permittivity')
('structures', 0, 'geometry', 'size', 1)
('structures', 1, 'geometry', 'vertices')
('structures', 1, 'geometry', 'geometries', 3, 'center', 0)

```

we have a limit only on the number of unique structures, not on the number of traced fields. So now I just count every structure that has traced fields in this check. Hence the set instead of list.

tidy3d/components/autograd/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+    AutogradTraced,
+    TracedCoordinate,
+    TracedFloat,
+    TracedSize,
+    TracedSize1D,
+    TracedVertices,
+)
+from .utils import get_static
+
+# from .derivative_utils import DerivativeInfo, integrate_within_bounds
```
yaugenst-flex: Any reason why this is commented out?

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -0,0 +1,124 @@
...
+from ..base import Tidy3dBaseModel
+from ..data.data_array import ScalarFieldDataArray
+from ..types import Bound, tidycomplex
+from .types import PathType
+from .utils import get_static
+
+# class passed to gradient computations
+# defined here to make things more compact, extendable, and better documented
+
+# we do this because importing these creates circular imports
```
yaugenst-flex, line 14: I guess this is fine for now but I think we should figure out how to do this more robustly... either reorganizing the types so that this doesn't happen or possibly use [forward references](https://peps.python.org/pep-0484/#forward-references).
Or maybe the best way would actually to `from __future__ import annotations` 🤔

tidy3d/components/autograd/utils.py
```diff
@@ -0,0 +1,15 @@
+# utilities for working with autograd
+
+import typing
+
+from autograd.tracer import getval
+
+
+def get_static(x: typing.Any) -> typing.Any:
```
yaugenst-flex, line 8: Is this really necessary? Why don't we just use `getval` directly? If it's because of naming, we could also `from autograd.tracer import getval as get_static` here?

yaugenst-flex (commented):
Nice I think the `DerivativeInfo` change is great. Also opens up the possibility for some derivative-specific helpers down the line.
Having a `utils.py` and a `derivative_utils.py` feels a bit weird to me, but not sure if it needs changing. I guess we'll merge this and I'll add #1769's `interpn` to a `functions.py` here?

tidy3d/components/autograd/__init__.py
tylerflex: Because it causes circular imports when this directory is imported into eg base.py. I'll remove the comment though

tidy3d/components/autograd/derivative_utils.py
tylerflex: The issue is that MonitorData such as FieldData can't be imported into eg geometry and medium, where these derivates are computed. 

One option might be to extract the derivative computation to somewhere in web/api/autograd instead of being methods of the classes. However, i do kind of like having them defined in the classes as it keeps things organized. 

I think annotations will not really help here but could be wrong. Also I'm wary about messing with forward refs unless it's necessary.

tidy3d/components/autograd/utils.py
tylerflex: Yea so originally this had some special handling for other types before i learned about getval. I'd say for now let's leave it in case we need to modify things? But probably in the future we can just use getval everywhere we use getstatic

tylerflex:
> Nice I think the `DerivativeInfo` change is great. Also opens up the possibility for some derivative-specific helpers down the line.

Yea it cleans things up for sure and now i feel less resistance to adding many more options eg for resolution for the eventual cylinder integrals or other things. 

> Having a `utils.py` and a `derivative_utils.py` feels a bit weird to me, but not sure if it needs changing. I guess we'll merge this and I'll add #1769's `interpn` to a `functions.py` here?

Yea it seems a bit weird to me too. Wasn't sure what to rename it to. If you add interpn maybe we can move these to functions.py. Or maybe there's a more descriptive name? I'm fine with it for now

tidy3d/components/autograd/__init__.py
tylerflex: Actually i need to double check this. It might be fixed now that i use dict of scalar field data array

tidy3d/components/autograd/derivative_utils.py
yaugenst-flex: No this should be the exact use case for `from __future__ import annotations` then 😄 Postponed evaluation of annotations, i.e. you don't need to import the type you're using as a type hint. And it's essentially a forward ref under the hood.

tylerflex: Yea maybe. I need to read up on forward refs a bit more.  Seems like a code smell to me though?

tylerflex: Practically speaking there was no issue with just passing the field data as generic dicts of data arrays.

PR #1769, opened 2024-06-17
Differentiable n-dimensional `nearest` and `linear` interpolation for `DataArray`

yaugenst-flex:
Implements differentiable interpolation methods for `DataArray`.

- [x] Implement interpolation function in `plugins/autograd/functions.py`.
- [x] Write tests & documentation for above.
- [x] Integrate into `td.DataArray`.
- [x] Test `DataArray` implementation.

### DIFF
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -344,7 +344,7 @@ def make_monitors() -> dict[str, tuple[td.Monitor, typing.Callable[[td.Simulatio
             value += abs(anp.sum(val.values))
         intensity = anp.nan_to_num(anp.sum(sim_data.get_intensity(mnt_data.monitor.name).values))
         value += intensity
-        # value += anp.sum(mnt_data.flux.values) # not yet supported
+        value += anp.sum(mnt_data.flux.values)
         return value
 
     field_point = td.FieldMonitor(
@@ -745,6 +745,35 @@ def test_too_many_traced_structures(monkeypatch, log_capture, use_emulated_run):
 
     with pytest.raises(ValueError):
         ag.grad(objective)(params0)
+
+
+@pytest.mark.parametrize("colocate", [True, False])
+@pytest.mark.parametrize("objtype", ["flux", "intensity"])
+def test_interp_objectives(use_emulated_run, colocate, objtype):
+    monitor = td.FieldMonitor(
+        center=(0, 0, 0),
+        size=(td.inf, td.inf, 0),
+        freqs=[FREQ0],
+        name="monitor",
+        colocate=colocate,
+    )
+
+    def objective(args):
+        structures_traced_dict = make_structures(args)
+        structures = list(SIM_BASE.structures)
+        for structure_key in structure_keys_:
+            structures.append(structures_traced_dict[structure_key])
+
+        sim = SIM_BASE.updated_copy(monitors=[monitor], structures=structures)
+        data = run(sim, task_name="autograd_test", verbose=False)
+
+        if objtype == "flux":
+            return anp.sum(data[monitor.name].flux.values)
+        elif objtype == "intensity":
+            return anp.sum(data.get_intensity(monitor.name).values)
+
+    grads = ag.grad(objective)(params0)
+    assert np.any(grads > 0)
 
 
 def test_autograd_deepcopy():
--- a/tests/test_plugins/autograd/test_functions.py
+++ b/tests/test_plugins/autograd/test_functions.py
@@ -1,6 +1,7 @@
 import numpy as np
 import numpy.testing as npt
 import pytest
+import scipy.interpolate
 import scipy.ndimage
 from autograd.test_util import check_grads
 from scipy.signal import convolve as convolve_sp
@@ -10,6 +11,7 @@ from tidy3d.plugins.autograd.functions import (
     grey_dilation,
     grey_erosion,
     grey_opening,
+    interpn,
     morphological_gradient,
     morphological_gradient_external,
     morphological_gradient_internal,
@@ -300,3 +302,38 @@ def test_threshold_exceptions(array, vmin, vmax, level, expected_message):
     """Test threshold function for expected exceptions."""
     with pytest.raises(ValueError, match=expected_message):
         threshold(array, vmin, vmax, level)
+
+
+@pytest.mark.parametrize("dim", [1, 2, 3, 4])
+@pytest.mark.parametrize("method", ["linear", "nearest"])
+class TestInterpn:
+    @staticmethod
+    def generate_points_values_xi(rng, dim):
+        points = tuple(np.linspace(0, 1, 10) for _ in range(dim))
+        values = rng.random([p.size for p in points])
+        xi = tuple(np.linspace(0, 1, 5) for _ in range(dim))
+        return points, values, xi
+
+    def test_interpn_val(self, rng, dim, method):
+        points, values, xi = self.generate_points_values_xi(rng, dim)
+        xi_grid = np.meshgrid(*xi, indexing="ij")
+
+        result_custom = interpn(points, values, xi, method=method)
+        result_scipy = scipy.interpolate.interpn(points, values, tuple(xi_grid), method=method)
+        npt.assert_allclose(result_custom, result_scipy)
+
+    @pytest.mark.parametrize("order", [1, 2])
+    @pytest.mark.parametrize("mode", ["fwd", "rev"])
+    def test_interpn_values_grad(self, rng, dim, method, order, mode):
+        points, values, xi = self.generate_points_values_xi(rng, dim)
+        check_grads(lambda v: interpn(points, v, xi, method=method), modes=[mode], order=order)(
+            values
+        )
+
+
+class TestInterpnExceptions:
+    def test_invalid_method(self, rng):
+        """Test that an exception is raised for an invalid interpolation method."""
+        points, values, xi = TestInterpn.generate_points_values_xi(rng, 2)
+        with pytest.raises(ValueError, match="interpolation method"):
+            interpn(points, values, xi, method="invalid_method")
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -393,7 +393,7 @@ custom_sellmeier_u = td.CustomSellmeier(
 # Make a few autograd ArrayBoxes for testing
 start_node = VJPNode.new_root()
 tracer = new_box(1.0, 0, start_node)
-tracer_arr = new_box([[[1.0]]], 0, start_node)
+tracer_arr = new_box(np.array([[[1.0]]]), 0, start_node)
 
 SIM_FULL = td.Simulation(
     size=(8.0, 8.0, 8.0),
@@ -852,6 +852,24 @@ SIM_FULL = td.Simulation(
 )
 
 
+def get_spatial_coords_dict(simulation: td.Simulation, monitor: td.Monitor, field_name: str):
+    """Returns MonitorData coordinates associated with a Monitor object"""
+    grid = simulation.discretize_monitor(monitor)
+    spatial_coords = grid.boundaries if monitor.colocate else grid[field_name]
+    spatial_coords_dict = spatial_coords.dict()
+
+    coords = {}
+    for axis, dim in enumerate("xyz"):
+        if monitor.size[axis] == 0:
+            coords[dim] = [monitor.center[axis]]
+        elif monitor.colocate:
+            coords[dim] = spatial_coords_dict[dim][:-1]
+        else:
+            coords[dim] = spatial_coords_dict[dim]
+
+    return coords
+
+
 def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.SimulationData:
     """Emulates a simulation run."""
     from scipy.ndimage.filters import gaussian_filter
@@ -872,19 +890,11 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
     def make_field_data(monitor: td.FieldMonitor) -> td.FieldData:
         """make a random FieldData from a FieldMonitor."""
         field_cmps = {}
-        coords = {}
         grid = simulation.discretize_monitor(monitor)
-
         for field_name in monitor.fields:
-            spatial_coords_dict = grid[field_name].dict()
-
-            for axis, dim in enumerate("xyz"):
-                if monitor.size[axis] == 0:
-                    coords[dim] = [monitor.center[axis]]
-                else:
-                    coords[dim] = np.array(spatial_coords_dict[dim])
-
+            coords = get_spatial_coords_dict(simulation, monitor, field_name)
             coords["f"] = list(monitor.freqs)
+
             field_cmps[field_name] = make_data(
                 coords=coords, data_array_type=td.ScalarFieldDataArray, is_complex=True
             )
@@ -900,17 +910,10 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
     def make_field_time_data(monitor: td.FieldTimeMonitor) -> td.FieldTimeData:
         """make a random FieldTimeData from a FieldTimeMonitor."""
         field_cmps = {}
-        coords = {}
         grid = simulation.discretize_monitor(monitor)
         tmesh = simulation.tmesh
         for field_name in monitor.fields:
-            spatial_coords_dict = grid[field_name].dict()
-
-            for axis, dim in enumerate("xyz"):
-                if monitor.size[axis] == 0:
-                    coords[dim] = [monitor.center[axis]]
-                else:
-                    coords[dim] = np.array(spatial_coords_dict[dim])
+            coords = get_spatial_coords_dict(simulation, monitor, field_name)
 
             (idx_begin, idx_end) = monitor.time_inds(tmesh)
             tcoords = tmesh[idx_begin:idx_end]
@@ -930,7 +933,6 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
     def make_mode_solver_data(monitor: td.ModeSolverMonitor) -> td.ModeSolverData:
         """make a random ModeSolverData from a ModeSolverMonitor."""
         field_cmps = {}
-        coords = {}
         grid = simulation.discretize_monitor(monitor)
         index_coords = {}
         index_coords["f"] = list(monitor.freqs)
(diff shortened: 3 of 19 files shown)


tests/test_components/test_autograd.py
```diff
@@ -332,7 +332,7 @@ def field_vol_postprocess_fn(sim_data, mnt_data):
             value += abs(anp.sum(val.values))
         intensity = anp.nan_to_num(anp.sum(sim_data.get_intensity(mnt_data.monitor.name).values))
         value += intensity
-        # value += anp.sum(mnt_data.flux.values) # not yet supported
+        value += anp.sum(anp.abs(mnt_data.flux.values))
         return value
```
tylerflex, line 348: I think we will also need an isolated test for a set of objective functions each with **only** one of flux, intensity, and poynting output. Just to ensure that the gradients are non-zero.

tests/utils.py
```diff
@@ -865,6 +865,16 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
     """Emulates a simulation run."""
     from scipy.ndimage.filters import gaussian_filter
 
+    def get_spatial_coords(grid, monitor, field_name):
+        spatial_coords = grid.boundaries if monitor.colocate else grid[field_name]
```
tylerflex: wonder if this should be built in as a `FieldMonitor` property? probably fine here though.

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -15,6 +15,11 @@
 FieldData = dict[str, ScalarFieldDataArray]
 PermittivityData = dict[str, ScalarFieldDataArray]
 
+__all__ = [
```
tylerflex: just curious, why is this up here instead of end of file?

tidy3d/components/autograd/functions.py
```diff
@@ -0,0 +1,91 @@
...
+
+import autograd.numpy as anp
+import numpy as np
+from numpy.typing import NDArray
+from scipy.interpolate import RegularGridInterpolator
+
+from .types import InterpolationType
+
+
+def _evaluate_nearest(
```
tylerflex, line 11: could you add some docstrings to explain what these are doing?

tidy3d/components/autograd/functions.py
```diff
@@ -0,0 +1,91 @@
+import itertools
+
+import autograd.numpy as anp
+import numpy as np
+from numpy.typing import NDArray
```
tylerflex, line 5: interesting, wasn't aware of these. wonder if we should use them more in the future

tidy3d/components/autograd/functions.py
```diff
@@ -0,0 +1,91 @@
...
+    ix = zip(indices, (1 - yi for yi in norm_distances))
+    iy = zip((i + 1 for i in indices), norm_distances)
+
+    value = anp.zeros(1)
+    for h in itertools.product(*zip(ix, iy)):
+        edge_indices, weights = zip(*h)
+        weight = anp.ones(1)
+        for w in weights:
+            weight = weight * w
+        # term = anp.array(values[edge_indices]) * weight[_slice]
```
tylerflex: remove comments?

tidy3d/components/autograd/functions.py
```diff
@@ -0,0 +1,91 @@
...
+    idx_res = tuple(anp.where(yi <= 0.5, i, i + 1) for i, yi in zip(indices, norm_distances))
+    return values[idx_res]
+
+
+def _evaluate_linear(
+    indices: NDArray[np.int64], norm_distances: NDArray[np.float64], values: NDArray[np.float64]
+) -> NDArray[np.float64]:
+    _slice = (slice(None),) + (None,) * (values.ndim - len(indices))
+
+    ix = zip(indices, (1 - yi for yi in norm_distances))
```
tylerflex, line 69: quick explanation of this?

tidy3d/components/autograd/functions.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Parameters
+    ----------
+    points : tuple[NDArray[np.float64], ...]
+        The points defining the rectilinear grid in n dimensions.
+    values : NDArray[np.float64]
+        The data values on the rectilinear grid.
+    xi : tuple[NDArray[np.float64], ...]
+        The coordinates to sample the gridded data at.
+    method : InterpolationType, optional
```
tylerflex: `= "linear"` I guess for consistency ¯\_(ツ)_/¯

tidy3d/components/autograd/functions.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Raises
+    ------
+    ValueError
+        If the interpolation method is not supported.
+
+    See Also
+    --------
+    scipy.interpolate.interpn : Interpolation on a rectilinear grid in arbitrary dimensions.
+        https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interpn.html
```
tylerflex: might need to do a special formatting for RST links 
```
`Link text <link URL>`_
```

tidy3d/components/autograd/functions.py
```diff
@@ -0,0 +1,91 @@
...
+    else:
+        raise ValueError(f"Unsupported interpolation method: {method}")
+
+    itrp = RegularGridInterpolator(points, values, method=method)
+    grid = anp.meshgrid(*xi, indexing="ij")
+    grid, shape, ndim, nans, _ = itrp._prepare_xi(tuple(grid))
+    indices, norm_distances = itrp._find_indices(grid.T)
+
+    result = interp_fn(indices, norm_distances, values)
+    # nans = anp.reshape(nans, (-1,) + (1,) * (result.ndim - 1))
```
tylerflex: remove comments?

tidy3d/components/base.py
```diff
@@ -1014,7 +1014,7 @@ def insert_value(x, path: tuple[str, ...], sub_dict: dict):
             sub_element = current_dict[final_key]
             if isinstance(sub_element, DataArray):
                 current_dict[final_key] = sub_element.copy(deep=False, data=x)
-                if "AUTOGRAD" in sub_element.attrs:
+                if isbox(x):
```
tylerflex: hm, could you explain this change? 

I think this logic makes a bit more sense but want to understand it a bit better.

tidy3d/components/data/data_array.py
```diff
@@ -70,21 +73,19 @@ def __init__(self, data, *args, **kwargs):
         """Initialize ``DataArray``."""
 
         # initialize with untraced data
-        data_untraced = getval(data)
-        super().__init__(data_untraced, *args, **kwargs)
-
-        # if the passed data has tracers, store them in attrs dict
         if isbox(data):
+            super().__init__(data._value, *args, **kwargs)
```
tylerflex: `._value -> getval()`? nitpick but seems a bit cleaner to me

tidy3d/components/data/data_array.py
```diff
@@ -70,21 +73,19 @@ def __init__(self, data, *args, **kwargs):
...
             self.attrs[AUTOGRAD_KEY] = data
-            # NOTE: this is done because if we pass the traced array directly, it will create a
-            # numpy array of `ArrayBox`, which is extremely slow
+        else:
```
tylerflex: could also just do
```py
super().__init__(getval(data), ...)
if isbox(data):
    # set attrs
```

tidy3d/components/data/data_array.py
```diff
@@ -283,6 +284,37 @@ def multiply_at(self, value: complex, coord_name: str, indices: List[int]) -> Da
...
 
+    def interp(
+        self,
+        coords: Mapping[Any, Any] | None = None,
+        method: InterpOptions = "linear",
+        assume_sorted: bool = False,
+        kwargs: Mapping[str, Any] | None = None,
+        **coords_kwargs: Any,
+    ):
+        if self.has_tracers:
```
tylerflex: docstring?

tidy3d/components/data/data_array.py
```diff
@@ -283,6 +284,37 @@ def multiply_at(self, value: complex, coord_name: str, indices: List[int]) -> Da
         self_mult[{coord_name: indices}] *= value
         return self_mult
 
+    def interp(
+        self,
+        coords: Mapping[Any, Any] | None = None,
+        method: InterpOptions = "linear",
+        assume_sorted: bool = False,
+        kwargs: Mapping[str, Any] | None = None,
+        **coords_kwargs: Any,
```
tylerflex, line 294: are we safe if the xarray API changes in the future? (or if it's different in the past). or will it break this?

tidy3d/components/data/monitor_data.py
```diff
@@ -589,8 +589,13 @@ def poynting(self) -> ScalarFieldDataArray:
...
         dim1, dim2 = self._tangential_dims
 
-        e1_h2 = tan_fields["E" + dim1] * tan_fields["H" + dim2].conj()
-        e2_h1 = tan_fields["E" + dim2] * tan_fields["H" + dim1].conj()
+        e1 = tan_fields["E" + dim1]
+        e2 = tan_fields["E" + dim2]
+        h1 = tan_fields["H" + dim1]
+        h2 = tan_fields["H" + dim2]
+
+        e1_h2 = e1 * (np.real(h2) - 1j * np.imag(h2))
```
tylerflex: should we overwrite the `DataArray.conj()` method to do this? or is it better to just leave it here for now?

tylerflex (changes requested):
Thanks @yaugenst-flex , let's discuss these last few things but overall looks great

tidy3d/components/autograd/derivative_utils.py
yaugenst-flex: I think I was shuffling things around and might have moved this out of habit. The convention I'm used to is to have this defined near the top of the file so it's easy to see which functions a module is supposed to export. I can change it back though since we usually have it at the bottom it seems.

tidy3d/components/autograd/functions.py
yaugenst-flex: Yeah there are a bunch of useful types in numpy: https://numpy.org/doc/stable/reference/typing.html

`NDArray` can be further typed to specify its dtype (`NDArray[float]` for example), which might be handy sometimes, otherwise it's just `Any`

tidy3d/components/base.py
yaugenst-flex: I think it was because `sub_element` might not have `.attrs[AUTOGRAD_KEY]` if somehow only the `.values` are being traced (and the `__init__` that does the `.values`/`.attrs` splitting never gets called). So `isbox(x)` will make sure that the `.attrs` will always be populated with tracers if `x` is being traced.

tidy3d/components/data/data_array.py
yaugenst-flex: Thought it was redundant because `getval()` only checks whether `data` is `Box` and then returns its `._value`, so we are checking `isbox()` twice basically.

tidy3d/components/data/data_array.py
yaugenst-flex: Yup sounds good. I think this is left over from when I was playing around with doing different things in the init.

tidy3d/components/data/data_array.py
yaugenst-flex: I copied these arguments directly from xarray. If they change the API then this will likely break, but then those xarray version wouldn't be compatible anyways?

tidy3d/components/data/monitor_data.py
yaugenst-flex: Yeah we probably should, good point.

tests/utils.py
```diff
@@ -865,6 +865,16 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
...
+        spatial_coords_dict = spatial_coords.dict()
+
+        return {
+            dim: (coords[:-1] if len(coords) > 1 else coords)
```
dmarek-flex: I think the issue causing `test_run_component_modeler` to fail is here. In my test, it is missing one coordinate because here it is dropped. I am sure you had a reason to drop it, but maybe the fix is to drop it if `colocate=True`, but keep it when `colocate=False`.

tests/utils.py
```diff
@@ -881,19 +891,11 @@ def make_data(
...
         """make a random FieldData from a FieldMonitor."""
         field_cmps = {}
-        coords = {}
         grid = simulation.discretize_monitor(monitor)
-
         for field_name in monitor.fields:
-            spatial_coords_dict = grid[field_name].dict()
-
-            for axis, dim in enumerate("xyz"):
-                if monitor.size[axis] == 0:
```
dmarek-flex, line 882: This logic seems to be missing from your changes. Just to be careful I would try to include this behavior.

`if monitor.size[axis] == 0:`
`    coords[dim] = [monitor.center[axis]]`
`else:	`
`    coords[dim] = np.array(spatial_coords_dict[dim])`

tests/utils.py
yaugenst-flex: Oooooh good catch!! Yeah the last coordinate needs to be dropped if the fields are colocated to the boundaries, but otherwise they should not be modified. This fixes all the tests actually, thanks a ton!

tests/utils.py
yaugenst-flex: This should be accounted for via the `len(coords) > 1` check I think

dmarek-flex: The `monitor.center` is set by the user, so it might not match the cell locations. When the size is 0 along one dimension the monitor will default to interpolating along that dimension onto the point specified by the user.

yaugenst-flex: I see, ok makes sense. I restored the old behavior.

tidy3d/components/data/monitor_data.py
yaugenst-flex: Actually I think it might be better to keep it like this for this PR, otherwise we run into the `.attrs` vs `.values` handling again.

tests/utils.py
yaugenst-flex: It's a function in the test utils now, I think it probably makes most sense there. There is also a microwave test that makes use of it.

tests/test_components/test_autograd.py
yaugenst-flex: Added

tidy3d/components/autograd/functions.py
yaugenst-flex: These actually shouldn't be commented out, good catch.

tidy3d/components/base.py
yaugenst-flex: I reverted this

tidy3d/components/data/monitor_data.py
yaugenst-flex: Nevermind, implemented it.

tidy3d/components/data/data_array.py
tylerflex: yea I think what worries me is that the underscore `_value` makes this seem like something we need to be careful about using, in case it changes for whatever reason. the `getval()` is public-facing API so it might be safer, even if it calls `isbox()` twice?

tests/test_components/test_autograd.py
```diff
@@ -703,6 +703,35 @@ def objective(*args):
...
+            structures.append(structures_traced_dict[structure_key])
+
+        sim = SIM_BASE.updated_copy(monitors=[monitor], structures=structures)
+        data = run(sim, task_name="autograd_test", verbose=True)
+
+        if objtype == "flux":
+            return anp.sum(data[monitor.name].flux.values)
+        elif objtype == "intensity":
+            return anp.sum(data.get_intensity(monitor.name).values)
+
```
tylerflex, line 774: in view of the notebook failing, maybe we need to add a `parameterize` for the size of the monitor (planar vs point?)

tests/test_plugins/autograd/test_functions.py
```diff
@@ -300,3 +302,38 @@ def test_threshold_exceptions(array, vmin, vmax, level, expected_message):
     """Test threshold function for expected exceptions."""
     with pytest.raises(ValueError, match=expected_message):
         threshold(array, vmin, vmax, level)
+
+
+@pytest.mark.parametrize("dim", [1, 2, 3, 4])
+@pytest.mark.parametrize("method", ["linear", "nearest"])
+class TestInterpn:
```
tylerflex, line 309: What do these test classes do? want to learn 😆

tidy3d/components/base.py
```diff
@@ -1014,8 +1014,8 @@ def insert_value(x, path: tuple[str, ...], sub_dict: dict):
             sub_element = current_dict[final_key]
             if isinstance(sub_element, DataArray):
                 current_dict[final_key] = sub_element.copy(deep=False, data=x)
-                if "AUTOGRAD" in sub_element.attrs:
-                    current_dict[final_key].attrs["AUTOGRAD"] = x
+                if AUTOGRAD_KEY in sub_element.attrs:
```
tylerflex, line 1017: yikes, good catch

tidy3d/components/data/data_array.py
```diff
@@ -283,6 +285,79 @@ def multiply_at(self, value: complex, coord_name: str, indices: List[int]) -> Da
...
+                raise KeyError(f"Cannot interpolate: {missing_keys} not in coords.")
+
+            obj = self if assume_sorted else self.sortby(list(coords.keys()))
+
+            points = tuple(dict(obj.coords).values())
+            xi = tuple(coords.get(k, obj.coords[k]) for k in obj.dims)
+            vals = interpn(points, obj.tracers, xi, method=method)
+
+            da = DataArray(vals, dict(obj.coords) | coords)  # tracers go into .attrs
+            if isbox(self.values.flat[0]):  # if tracing .values instead of .attrs
```
tylerflex, line 335: or `da.contains_tracers`? although this seems to no longer exist?

tylerflex: maybe it's fine for now until we standardize all of this.

tylerflex (approved):
Looking good! 👍 feel like this is ready once the notebook issue is figured out. thanks Yannick!

tidy3d/components/data/data_array.py
yaugenst: Yup you're right. I changed the init according to your other suggestions so it's back to the old behavior now anyways :)

yaugenst-flex: Yup you're right. I changed the init according to your other suggestions so it's back to the old behavior now anyways :)

tests/test_components/test_autograd.py
yaugenst-flex: Will do. The test doesn't fail for point monitors though 😂

tests/test_plugins/autograd/test_functions.py
yaugenst-flex: It's really just a way to group related tests so that the tests contained in this class have access to some common methods (like the array construction here) and possibly fixtures that are only defined within the scope of one test set. So here I'm grouping the tests for interp that check equivalence to scipy and those that check grads because they can both reuse the same array setup and parametrization.
The whole class is parametrized such that all methods contained within that test class will be checked according to that common parametrization. In this case that's dimensionality and interpolation method.
The autograd test is then further sub-parametrized to check gradient order and mode.

tidy3d/components/data/data_array.py
yaugenst-flex: Yeah I can also reintroduce that property, but I think the whole handling needs some work so maybe we leave it as is for the patch.

tests/test_plugins/autograd/test_functions.py
tylerflex: ah, that's interesting. I could probably use that in the main autograd tests, maybe I'll give it a try.

tidy3d/components/data/data_array.py
tylerflex: ok

yaugenst-flex:
Metalens notebook is passing, so this should be ready to go @tylerflex @momchil-flex

tylerflex:
Looks good to me! Feel free to merge when everyone approves. Could you also add a PR for the autograd metalens notebook?

tylerflex:
and perhaps change existing autograd notebooks to use `get_intensity` where applicable.

(quickstart, metallic antenna).

tylerflex:
@dmarek-flex any last comments / suggestions before we merge this?

dmarek-flex (approved):
Looks good

PR #1772, opened 2024-06-18
Add progress bar for EME

caseyflex:
Together with the backend PR, this adds a progress bar for EME simulations.

### DIFF
--- a/tidy3d/web/api/webapi.py
+++ b/tidy3d/web/api/webapi.py
@@ -436,32 +436,47 @@ def monitor(task_id: TaskId, verbose: bool = True) -> None:
         time.sleep(REFRESH_TIME)
 
     # while running but percentage done is available
-    if verbose and task_type == "FDTD":
+    if verbose:
         # verbose case, update progressbar
         console.log("running solver")
-        with Progress(console=console) as progress:
-            pbar_pd = progress.add_task("% done", total=100)
-            perc_done, _ = get_run_info(task_id)
+        if task_type == "FDTD":
+            with Progress(console=console) as progress:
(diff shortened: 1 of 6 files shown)


tidy3d/web/api/webapi.py
```diff
@@ -436,32 +436,49 @@ def monitor_preprocess() -> None:
...
+                ):
+                    perc_done, _ = get_run_info(task_id)
+                    new_description = "solver progress"
+                    progress.update(pbar_pd, completed=perc_done, description=new_description)
+                    time.sleep(RUN_REFRESH_TIME)
+
+                perc_done, _ = get_run_info(task_id)
+                new_description = "solver progress"
+                progress.update(
+                    pbar_pd, completed=perc_done, refresh=True, description=new_description
```
momchil-flex: In the FDTD case, we set `completed=100` here to force the progress bar to look "done" once the solver is done. Any reason you set `perc_done` instead? Which I think should have no effect then?

momchil-flex (approved):
Thanks, this is great! In my tests it looked a little uneven in the progress (I mean a bit uneven time between updates) but it's nice to have *something* happening. :)

tidy3d/web/api/webapi.py
caseyflex: thanks for the catch!

caseyflex:
> Thanks, this is great! In my tests it looked a little uneven in the progress (I mean a bit uneven time between updates) but it's nice to have _something_ happening. :)

Yeah, it's definitely not entirely even, just based on counting steps, but should be better than nothing. We can update it to use the time estimation later.

PR #1773, opened 2024-06-18
Move dispersion fitter from plugins to tidy3d

weiliangjin2021:
In surface impedance treatment, the impedance is dispersive. Previously Debye model has been used to fit the impedance. CCPR model should require fewer number of poles. The plan is to add the fitter as a method (or cached property) to the good conductor material class, so that before submitting the simulation, one can examine if the fitted impedance looks good with the default fitting parameters.

### DIFF
--- a/tidy3d/components/fitter/fit.py
+++ b/tidy3d/components/fitter/fit.py
@@ -12,15 +12,13 @@ import scipy.optimize as opt
 from pydantic.v1 import Field, validator
 from rich.progress import Progress
 
-from tidy3d.web.core.environment import Env
-
-from ...components.base import Tidy3dBaseModel, cached_property, skip_if_fields_missing
-from ...components.medium import AbstractMedium, PoleResidue
-from ...components.types import ArrayFloat1D, Ax
-from ...components.viz import add_ax_if_none
 from ...constants import C_0, HBAR, MICROMETER
 from ...exceptions import SetupError, ValidationError, WebError
 from ...log import get_logging_console, log
+from ..base import Tidy3dBaseModel, cached_property, skip_if_fields_missing
+from ..medium import AbstractMedium, PoleResidue
+from ..types import ArrayFloat1D, Ax
+from ..viz import add_ax_if_none
 
 
 class DispersionFitter(Tidy3dBaseModel):
@@ -614,7 +612,7 @@ class DispersionFitter(Tidy3dBaseModel):
         :class:`DispersionFitter`
             A :class:`DispersionFitter` instance.
         """
-        resp = requests.get(url_file, verify=Env.current.ssl_verify)
+        resp = requests.get(url_file)
 
         try:
             resp.raise_for_status()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Bug where boundary layers would be plotted too small in 2D simulations.
 
+### Changed
+- Dispersive medium fitter in `plugins.dispersion` has been moved to Tidy3D module.
+
 ## [2.7.0] - 2024-06-17
 
 ### Added
--- a/docs/api/mediums.rst
+++ b/docs/api/mediums.rst
@@ -25,8 +25,8 @@ Spatially varying
 
    tidy3d.CustomMedium
 
-Dispersive Mediums
-------------------
+Dispersive Medium
+-----------------
 
 Spatially uniform
 ^^^^^^^^^^^^^^^^^
@@ -52,6 +52,16 @@ Spatially varying
    tidy3d.CustomDrude
    tidy3d.CustomDebye
 
+Dispersive Medium Fitter
+------------------------
+
+.. autosummary::
+   :toctree: _autosummary/
+   :template: module.rst
+
+   tidy3d.FastDispersionFitter
+   tidy3d.AdvancedFastFitterParam
+   tidy3d.DispersionFitter
 
 Medium Perturbations
 --------------------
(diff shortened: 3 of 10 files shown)


tidy3d/components/fitter/fit.py
```diff
@@ -614,7 +612,7 @@ def from_url(
         :class:`DispersionFitter`
             A :class:`DispersionFitter` instance.
         """
-        resp = requests.get(url_file, verify=Env.current.ssl_verify)
+        resp = requests.get(url_file)
```
weiliangjin2021, line 615: I'm not sure about this change. It works fine without verify. From git history, it was added by MC. Can we still call `Env` in the main module?

caseyflex (approved):
Sounds good to me. In the future, it could be interesting to compare the results of this fitter to the current fitting results for Graphene. The fitting method for Graphene is less general and robust than this one -- this might work for a broader range of parameters (T, mu, gamma). We would want the resulting poles and residues to be similar though.

tylerflex (commented):
Hi @weiliangjin2021 , the implementation looks good.

I'm also ok with moving this into the core components, but it does potentially cause some circular import danger. Especially as we add more and more features to mediums.. so I'm slightly worried. I guess you already mentioned that you'll need to address by splitting up medium.py?

Have you considered moving this impedance / conductance work to its own plugin? potentially even an `rf` plugin if it goes there eventually? maybe this should be considered at this stage but since I dont understand all of the details, I'll defer to you as to what's best.

weiliangjin2021:
> I'm also ok with moving this into the core components, but it does potentially cause some circular import danger. Especially as we add more and more features to mediums.. so I'm slightly worried. I guess you already mentioned that you'll need to address by splitting up medium.py?

The changes in this PR itself won't cause complications in circular import?

tylerflex:
> The changes in this PR itself won't cause complications in circular import?

No I don't think so. But we just need to be careful about it. 

@caseyflex and I discussed possibly moving some more general features (fitter, harmonic inversion) to their own separate repositories to be included as dependencies and pip installed. Seems like a similar idea to this PR, but then users can use the fitters more generally? Perhaps we can try this at some point. Casey, feel free to elaborate on your thoughts about the future of the fitter.

weiliangjin2021:
> possibly moving some more general features (fitter, harmonic inversion) to their own separate repositories to be included as dependencies and pip installed. Seems like a similar idea to this PR, but then users can use the fitters more generally?

Good idea. In that case, the fitter can be generalized to be applied to any passive causal response, not just permittivity. Or, the fitter won't import Tidy3D Medium.

caseyflex:
> > The changes in this PR itself won't cause complications in circular import?
> 
> No I don't think so. But we just need to be careful about it.
> 
> @caseyflex and I discussed possibly moving some more general features (fitter, harmonic inversion) to their own separate repositories to be included as dependencies and pip installed. Seems like a similar idea to this PR, but then users can use the fitters more generally? Perhaps we can try this at some point. Casey, feel free to elaborate on your thoughts about the future of the fitter.

Yeah, I think moving the dispersion fitter and the resonance finder to separate repositories would be a nice way to simplify the code base a bit. I don't love the idea of moving the fitter to components, since it isn't one of the core components of a simulation but rather a tool used to generate a medium for a simulation.

I can look into the move if people think it makes sense. It would require eliminating dependency on tidy3d classes like Medium, but I don't think that would be very difficult.

tylerflex:
Sounds like a plan to me. Maybe we can start with fitter?

weiliangjin2021:
> > > The changes in this PR itself won't cause complications in circular import?
> > 
> > 
> > No I don't think so. But we just need to be careful about it.
> > @caseyflex and I discussed possibly moving some more general features (fitter, harmonic inversion) to their own separate repositories to be included as dependencies and pip installed. Seems like a similar idea to this PR, but then users can use the fitters more generally? Perhaps we can try this at some point. Casey, feel free to elaborate on your thoughts about the future of the fitter.
> 
> Yeah, I think moving the dispersion fitter and the resonance finder to separate repositories would be a nice way to simplify the code base a bit. I don't love the idea of moving the fitter to components, since it isn't one of the core components of a simulation but rather a tool used to generate a medium for a simulation.
> 
> I can look into the move if people think it makes sense. It would require eliminating dependency on tidy3d classes like Medium, but I don't think that would be very difficult.

Sounds good. I'll then focus on the solver side of SIBC for now. I think I'll still need 2 or 3 weeks for that.

After the isolation, the component will still import the fitter library, right?

caseyflex:
> > > > The changes in this PR itself won't cause complications in circular import?
> > > 
> > > 
> > > No I don't think so. But we just need to be careful about it.
> > > @caseyflex and I discussed possibly moving some more general features (fitter, harmonic inversion) to their own separate repositories to be included as dependencies and pip installed. Seems like a similar idea to this PR, but then users can use the fitters more generally? Perhaps we can try this at some point. Casey, feel free to elaborate on your thoughts about the future of the fitter.
> > 
> > 
> > Yeah, I think moving the dispersion fitter and the resonance finder to separate repositories would be a nice way to simplify the code base a bit. I don't love the idea of moving the fitter to components, since it isn't one of the core components of a simulation but rather a tool used to generate a medium for a simulation.
> > I can look into the move if people think it makes sense. It would require eliminating dependency on tidy3d classes like Medium, but I don't think that would be very difficult.
> 
> Sounds good. I'll then focus on the solver side of SIBC for now. I think I'll still need 2 or 3 weeks for that.
> 
> After the isolation, the component will still import the fitter library, right?

Yeah, we can have the fitter library as a core dependency of tidy3d, right @tylerflex?

tylerflex:
> Yeah, we can have the fitter library as a core dependency of tidy3d, right @tylerflex?

Yea I think that would make sense. We can just have a lightweight, generic fitter repo and then add it as a core dependency.

weiliangjin2021:
> > Yeah, we can have the fitter library as a core dependency of tidy3d, right @tylerflex?
> 
> Yea I think that would make sense. We can just have a lightweight, generic fitter repo and then add it as a core dependency.

Sounds a good plan. Let me close this PR, and handle it over to @caseyflex

caseyflex:
Great, yeah I'll take a stab at this next week.

PR #1774, opened 2024-06-19
🗻 autograd: broadband

tylerflex:
no longer WIP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Broadband adjoint support in autograd for adjoint sources with the same spatial dependence.
+
 ## [2.7.1]
 
 ### Added
@@ -13,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ModeSolver` methods to plot the mode plane simulation components, including `.plot()`, `.plot_eps()`, `.plot_structures_eps()`, `.plot_grid()`, and `.plot_pml()`.
 - Support for differentiation with respect to monitor attributes that require interpolation, such as flux and intensity.
 - Support for automatic differentiation with respect to `.eps_inf` and `.poles` contained in dispersive mediums `td.PoleResidue` and `td.CustomPoleResidue`.
+- Support for automatic differentiation with respect to `.eps_inf` and `.poles` contained in dispersive mediums `td.PoleResidue` and `td.CustomPoleResidue`.
+
+### Changed
 
 ### Fixed
 - Bug where boundary layers would be plotted too small in 2D simulations.
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -16,6 +16,8 @@ from tidy3d.web import BatchData
 """ utilities shared between all tests """
 np.random.seed(4)
 
+# function used to generate the data for emulated runs
+DATA_GEN_FN = np.random.random
 
 FREQS = np.array([1.90, 2.01, 2.2]) * 1e12
 SIM_MONITORS = td.Simulation(
@@ -880,7 +882,7 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
         """make a random DataArray out of supplied coordinates and data_type."""
         data_shape = [len(coords[k]) for k in data_array_type._dims]
         np.random.seed(1)
-        data = np.random.random(data_shape)
+        data = DATA_GEN_FN(data_shape)
 
         data = (1 + 0.5j) * data if is_complex else data
         data = gaussian_filter(data, sigma=1.0)  # smooth out the data a little so it isnt random
@@ -939,7 +941,7 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
         index_coords["mode_index"] = np.arange(monitor.mode_spec.num_modes)
         index_data_shape = (len(index_coords["f"]), len(index_coords["mode_index"]))
         index_data = ModeIndexDataArray(
-            (1 + 1j) * np.random.random(index_data_shape), coords=index_coords
+            (1 + 1j) * DATA_GEN_FN(index_data_shape), coords=index_coords
         )
         for field_name in ["Ex", "Ey", "Ez", "Hx", "Hy", "Hz"]:
             coords = get_spatial_coords_dict(simulation, monitor, field_name)
@@ -977,7 +979,7 @@ def run_emulated(simulation: td.Simulation, path=None, **kwargs) -> td.Simulatio
         orders_x = np.linspace(-1, 1, 3)
         orders_y = np.linspace(-2, 2, 5)
         coords = dict(orders_x=orders_x, orders_y=orders_y, f=f)
-        values = np.random.random((len(orders_x), len(orders_y), len(f)))
+        values = DATA_GEN_FN((len(orders_x), len(orders_y), len(f)))
         data = td.DiffractionDataArray(values, coords=coords)
         field_data = {field: data for field in ("Er", "Etheta", "Ephi", "Hr", "Htheta", "Hphi")}
         return td.DiffractionData(monitor=monitor, sim_size=(1, 1), bloch_vecs=(0, 0), **field_data)
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -20,7 +20,7 @@ from ..base_sim.data.sim_data import AbstractSimulationData
 from ..file_util import replace_values
 from ..monitor import Monitor
 from ..simulation import Simulation
-from ..source import Source
+from ..source import ModeSource, Source
 from ..structure import Structure
 from ..types import Ax, Axis, ColormapType, FieldVal, PlotScale, annotate_type
 from ..viz import add_ax_if_none, equal_aspect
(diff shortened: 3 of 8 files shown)


tylerflex:
@yaugenst @momchil-flex if we want to try to get this in for 2.7.1, it works. Except only custom medium gradients are fully tested for accuracy.

tests/utils.py
```diff
@@ -16,6 +16,8 @@
 """ utilities shared between all tests """
 np.random.seed(4)
 
+# function used to generate the data for emulated runs
+DATA_GEN_FN = np.random.random
```
yaugenst-flex, line 20: Are we planning to use something else here? I think in general it might be a good idea to introduce a fixture that returns a generator, instead of relying on a global `np.random.seed` (considered legacy and might be deprecated at some point). I have something like that for the [autograd plugin tests](https://github.com/flexcompute/tidy3d/blob/develop/tests/test_plugins/autograd/conftest.py), but we might want to tweak and reuse this globally.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Users can now export `SimulationData` to MATLAB `.mat` files with the `to_mat_file` method.
 - `ModeSolver` methods to plot the mode plane simulation components, including `.plot()`, `.plot_eps()`, `.plot_structures_eps()`, `.plot_grid()`, and `.plot_pml()`.
 - Support for differentiation with respect to monitor attributes that require interpolation, such as flux and intensity.
+- Broadband adjoint support in autograd for adjoint sources with the same spatial dependence.
+- Support for automatic differentiation with respect to `.eps_inf` and `.poles` contained in dispersive mediums `td.PoleResidue` and `td.CustomPoleResidue`.
```
yaugenst-flex, line 19: Dispersive probably shouldn't be in here?

tidy3d/components/data/sim_data.py
```diff
@@ -978,9 +982,158 @@ def make_adjoint_sim(
...
+        log.info(
+            "Several adjoint sources, from one monitor. "
+            "Only difference between them is the source time. "
+            "Constructing broadband adjoint source and performing post-run normalization "
+            f"of fields with {len(post_norm_amps)} frequencies."
+        )
+
+        return [src_broadband], post_norm_amps
+
+    def _make_broadband_source(self, adj_srcs: List[Source], num_fwidth: float = 0.5) -> Source:
```
yaugenst-flex: Do we use `list` or `typing.List`? Collections like `list` can be used as generic types since Python 3.9, so I guess we should probably switch to using that everywhere since we don't support <3.9? I'll also have to change all of my type hints...

tidy3d/components/data/sim_data.py
```diff
@@ -978,9 +982,158 @@ def make_adjoint_sim(
...
+        # no adjoint sources, gradient is 0
+        if num_adj_src_monitors == 0:
+            log.warning(
+                "Something unexpected happened. There are No adjoint sources, "
+                " yet the adjoint pipeline was triggered. No gradient will be computed "
+                "with respect to simulation data output. If you receive this, please file a bug "
+                "report on the tidy3d github repository with as much information as "
+                "you can provide about your setup."
+            )
+            return [], None
```
yaugenst-flex: When would this happen? Can this happen when then gradient w.r.t. a specific monitor is exactly zero, and would this always be a considered a bug?

tidy3d/components/data/sim_data.py
```diff
@@ -978,9 +982,158 @@ def make_adjoint_sim(
...
+        all_freqs = set()
+        for mnt_name, freqs in sources_unique_freqs.items():
+            freq = tuple(freqs)[0]
+            all_freqs.add(freq)
+            if len(all_freqs) > 1:
+                raise ValueError(
+                    error_msg_pre(mnt_name)
+                    + error_msg_post
+                    + " Detected different frequencies in different monitors."
+                )
```
yaugenst-flex: There is probably no need to loop over `sources_unique_freqs` twice here?

tidy3d/components/data/sim_data.py
```diff
@@ -978,9 +982,158 @@ def make_adjoint_sim(
...
+
+        log.info("Several adjoint sources from different monitors, all with same single frequency.")
+
+        # passed validation, put all sources into a single list and set None for post-normalize amps
+        adj_srcs = []
+        for srcs in sources_adj_dict.values():
+            adj_srcs += srcs
```
yaugenst-flex: I would write `adj_srcs = sum(sources_adj_dict.values(), [])`.
Not sure if it's maybe less readable.

tidy3d/components/medium.py
```diff
@@ -2551,8 +2551,9 @@ def _derivative_field_cmp(
         eps_data: PermittivityDataset,
         dim: str,
     ) -> np.ndarray:
-        coords_interp = {key: val for key, val in eps_data.coords.items() if len(val) > 1}
-        dims_sum = {dim for dim in eps_data.coords.keys() if dim not in coords_interp}
+        coords_interp = {key: eps_data.coords[key] for key in "xyz"}
+        coords_interp = {key: val for key, val in coords_interp.items() if len(val) > 1}
```
yaugenst-flex, line 2557: Maybe `coords_interp = {key: val for key, val in eps_data.coords.items() if key in "xyz" and len(val) > 1}`?

tidy3d/components/medium.py
```diff
@@ -2581,8 +2582,11 @@ def _derivative_field_cmp(
 
         # TODO: probably this could be more robust. eg if the DataArray has weird edge cases
         E_der_dim = E_der_map[f"E{dim}"]
-        E_der_dim_interp = E_der_dim.interp(**coords_interp).fillna(0.0).sum(dims_sum)
-        vjp_array = np.array(E_der_dim_interp.values).astype(complex)
+        E_der_dim_interp = E_der_dim.interp(**coords_interp).fillna(0.0).sum(dims_sum).real
+        E_der_dim_interp = E_der_dim_interp.sum("f")
+
+        vjp_array = np.array(E_der_dim_interp.values).astype(float)
```
yaugenst-flex, line 2590: Was the `.astype(complex)` here wrong? Wondering if the different return type here affects other things.

yaugenst-flex (approved):
Sweet! Only some minor comments.

yaugenst-flex:
Tests are failing? @tylerflex

tests/utils.py
tylerflex: I agree. and no plans (yours sounds good). I just left it here since it makes it easier for testing. Should I change it? or is it fine to leave like this for now?

CHANGELOG.md
tylerflex: Yea, it turned out that dispersive was in the wrong place, I'll fix when I rebase.

tidy3d/components/data/sim_data.py
tylerflex: yea good point, I'll change

tidy3d/components/data/sim_data.py
tylerflex: hmmm. yea I dont know. so if we do something like `y = 0*x`, I'm not sure if autograd still traces `y`? in any case, it's unexpected, but I dont think we need to error? but the user might want to know what's going on so I felt a warning would be sufficient to catch this weird edge case.

tidy3d/components/data/sim_data.py
tylerflex: no, I actually tried it together, but these are just two separate checks so it was a lot simpler to first be able to ensure that all `freqs` were length 1 and then do the 2nd check.

tidy3d/components/data/sim_data.py
tylerflex: learn something new every day :D. let's try this.

tidy3d/components/medium.py
tylerflex: I actually like mine better since this one breaks over two lines.. sometimes I need to modify these comprehensions and have trouble following when they split over several lines.

tidy3d/components/medium.py
tylerflex: not sure it was wrong, I just did the real part later. basically we just need to make sure the vjp dtype matches the `data` dtype in the corresponding DataArray. For this component I think it's a float. (So yea it might have been wrong before?)

tests/utils.py
yaugenst: Was just a general comment, lef‘s leave it for now. Would make sense to add this in one swoop together with the rest of the tests.

tidy3d/components/data/sim_data.py
yaugenst: Yeah no I agree with not erroring, I was just confused because this tells the user to submit a bug report 😄 Maybe the text should be changed to something like „If this is unexpected …“?

tidy3d/components/data/sim_data.py
yaugenst: but wouldnt this work exactly the same if line 1065 was simply removed? and line 1064 moved to the top? either way not a big deal, was just wondering

tidy3d/components/medium.py
yaugenst: yeah fair

tidy3d/components/data/sim_data.py
tylerflex: yea actually that's probably correct

tidy3d/components/data/sim_data.py
tylerflex: good point, I'll update it

tylerflex:
Looks like the broadband approach with fitting is working. so let's not merge this yet. I will combine all into one PR covering each of the cases.

tylerflex:
I put everything into one unified approach #1830

PR #1776, opened 2024-06-19
Rework make_filter interface to accept either size_px or radius and d…

### LINKED ISSUE #1775: `autograd` utilities interface changes
Just making a note of this so we have a record. Some things I think will improve usability:

1. Everything that we want users to use could be importable from `tidy3d.plugins.autograd`. I found it a bit confusing to know which subdirectory to find various things, eg `invdes`, `functions`.  Still fine (if not advisable) to have things organized into these sub-folders, though, just import them into __init__
2. Bypass need to get pixel size first and then make filter in two separate calls. 
```py
px = get_pixel_size(lx, radius)
filter = make_filter(px)

# becomes
filter = make_filter(lx, radius)
```
I'd consider either `make_filter()` taking `radius` and `length_scale` directly, and then a separate `make_filter_from_pixels()` (or similar), or overload the kwargs as we discussed to accept all three and do some error checking.

yaugenst-flex:
…l arguments, import all user-facing functions into the plugins/autograd namespace, and add tests

Closes #1775


tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -5,28 +5,69 @@
...
+    ValueError
+        If neither 'size_px' nor both 'radius' and 'dl' are provided.
+    """
+    if size_px is not None:
+        return (size_px,) if np.isscalar(size_px) else tuple(size_px)
+    elif radius is not None and dl is not None:
+        kernel_size = get_kernel_size_px(radius=radius, dl=dl)
+        return (kernel_size,) if np.isscalar(kernel_size) else tuple(kernel_size)
+    else:
+        raise ValueError("Either 'size_px' or both 'radius' and 'dl' must be provided.")
```
tylerflex, line 163: to remove any ambiguity, it might be best to just only handle specific cases?
```
    if size_px is not None and (radius is None and dl is None):
        return (size_px,) if np.isscalar(size_px) else tuple(size_px)
    elif radius is not None and dl is not None and size_px is None:
        kernel_size = get_kernel_size_px(radius=radius, dl=dl)
        return (kernel_size,) if np.isscalar(kernel_size) else tuple(kernel_size)
    else:
        raise ValueError("Either 'size_px' or both 'radius' and 'dl' must be provided.")
```

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -5,28 +5,69 @@
...
 ):
     """Create a filter function based on the specified kernel type and size.
 
     Parameters
     ----------
-    filter_type : KernelType
-        The type of kernel to create (`circular` or `conic`).
-    size : Union[int, Tuple[int, ...]]
-        The size of the kernel in pixels for each dimension. Can be a scalar or a tuple.
+    radius : Union[float, Tuple[float, ...]], optional
```
tylerflex: guess we can test whether this will render well in our docs, it might be inconsistent throughout but we generally have been doing `= None`

tidy3d/plugins/autograd/utilities.py
```diff
@@ -77,6 +77,23 @@ def make_kernel(kernel_type: KernelType, size: Iterable[int], normalize: bool =
     return kernel
 
 
+def get_kernel_size_px(radius=None, dl=None):
+    if radius is None or dl is None:
```
tylerflex, line 103: docstring? return type `int`?

tidy3d/plugins/autograd/utilities.py
```diff
@@ -77,6 +77,23 @@ def make_kernel(kernel_type: KernelType, size: Iterable[int], normalize: bool =
...
+        raise ValueError("Either 'size_px' or both 'radius' and 'dl' must be provided.")
```
tylerflex, line 104: If we are restrictive in the kwargs we accept earlier on, probably this will never need to be checked?

tidy3d/plugins/autograd/__init__.py
tylerflex, line 1: module level docstring? thought ruff would check this. maybe we disabled it?

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- All filter functions in `plugins/autograd` now accept either an absolute size in pixels or a `radius` and `dl` argument.
+
```
tylerflex: mention that the previous pixel size argument `size` is now `size_px`

tylerflex (commented):
Thanks @yaugenst . Looks good overall, a few details.
The main thing to discuss is the fact that this breaks backwards compabibility with the `size` -> `size_px` change. What are your thoughts on that? is `autograd` so experimental that we can do this safely?

tidy3d/plugins/autograd/utilities.py
yaugenst-flex: I though this here would be the common denominator for all functions that use `get_kernel_size_px`, so it makes to check this here? Otherwise we'd need to add this check to every functions that handles both `radius` and `dl`?

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: Ah, my thinking was that if a kernel size in pixels is given, then that would take precedence over radius and dl. Maybe it makes sense to do this and warn the user? Or just error?

yaugenst-flex:
> The main thing to discuss is the fact that this breaks backwards compabibility with the size -> size_px change. What are your thoughts on that? is autograd so experimental that we can do this safely?

Yeah this is pretty unfortunate, but `size` is I think too ambiguous... I think (I hope) it's fine? I'm sure there will be other breaking changes too...

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: Ah damn yeah, can we keep it in this PR and if it's a problem I'll change it everywhere in another PR?

tylerflex:
Let's change target of this to pre/2.8?

yaugenst-flex:
Oh yeah was just in the process of doing that :)

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: Ok, this is done for everything within the autograd plugin at least

tylerflex: cool thanks

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+### Changed
```
tylerflex: duplicate changelog?

Also maybe add a line saying that all user-facing features can be imported now directly from `tidy3d.plugins.autograd`?

docs/faq
tylerflex, line 1: will need to update this submodule to latest state of develop

tidy3d/plugins/autograd/README.md
```diff
@@ -152,6 +152,8 @@ Next on our roadmap (targeting 2.8 and 2.9, summer 2024) is to support:
 - `PoleResidue` and other dispersive models.
 - custom (spatially-dependent) dispersive models, allowing topology optimization with metals.
 
+- `ComplexPolySlab`
```
tylerflex: remove this change? also, wondering if you ever updated this readme to reflect the field data operations (put them in the "supported" section?) if not, maybe we can do that in this PR? 

idea is to have this README contain the latest state of released, in progress, and todo.

tidy3d/plugins/autograd/functions.py
tylerflex, line 1: these changes exist already in develop from the interp PR? should we merge `develop` into `pre/2.8` maybe?

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+from tidy3d.components.base import Tidy3dBaseModel
 
 from ..functions import convolve
 from ..types import KernelType, PaddingType
-from ..utilities import make_kernel
+from ..utilities import get_kernel_size_px, make_kernel
+
+
+class AbstractFilter(Tidy3dBaseModel, abc.ABC):
+    """A filter class for creating and applying convolution filters.
```
tylerflex: maybe just `An abstract class` since `filter` is mentioned later?

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+
+    Parameters
+    ----------
+    kernel_size : Tuple[pd.PositiveInt, ...]
```
tylerflex: docstring should reflect single int case

tylerflex: actually, we dont need this parameters list in the docstring do we?

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+    )
+    normalize: bool = pd.Field(
+        True, description="Whether to normalize the kernel so that it sums to 1."
+    )
+    padding: PaddingType = pd.Field("reflect", description="The padding mode to use.")
+
+    @classmethod
+    def from_radius_dl(
+        cls, radius: Union[float, Tuple[float, ...]], dl: Union[float, Tuple[float, ...]], **kwargs
+    ) -> "AbstractFilter":
```
tylerflex: use `from __future__ import annotations` and then regular `AbstractFilter`

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+            Size of the kernel in pixels for each dimension.
+        normalize : bool
+            Whether to normalize the kernel so that it sums to 1.
+
+        Returns
+        -------
+        NDArray
+            The kernel.
+        """
+        ...
```
tylerflex: what does this do? like `pass`?

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+
+    def __call__(self, array: NDArray) -> NDArray:
+        """Apply the filter to an input array.
+
+        Parameters
+        ----------
+        array : NDArray
```
tylerflex: does `NDArray` maybe need to be in double backslash? wonder if the users will know what this is and if we should annotate with `np.ndarray` for docstrings?

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+
+        Returns
+        -------
+        NDArray
+            The filtered array.
+        """
+        original_shape = array.shape
+        squeezed_array = np.squeeze(array)
+        size_px = tuple(np.atleast_1d(self.kernel_size))
+        print(size_px)
```
tylerflex: remove

tylerflex: or make `log.info` might be better with a short message?

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+        convolved_array = convolve(squeezed_array, kernel, padding=self.padding)
+        return np.reshape(convolved_array, original_shape)
+
+
+class ConicFilter(AbstractFilter):
+    """A conic filter for creating and applying convolution filters."""
+
+    @staticmethod
+    @lru_cache(maxsize=1)
+    def get_kernel(size_px: Iterable[int], normalize: bool) -> NDArray:
```
tylerflex, line 102: can this be a `@cached_property`?

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+
+    Returns
+    -------
+    Tuple[int, ...]
+        The size of the kernel in pixels for each dimension.
+
+    Raises
+    ------
+    ValueError
+        If neither 'size_px' nor both 'radius' and 'dl' are provided.
```
tylerflex: double backslash for `RST`

tidy3d/plugins/autograd/invdes/filters.py
tylerflex: up to you. but just need to clearly define what happens if all 3 are provided perhaps. I might suggest  error or at least warn? seems like untented use.

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+    kernel_size = _get_kernel_size(radius, dl, size_px)
 
-        convolved_array = convolve(squeezed_array, _kernel[squeezed_array.ndim], padding=padding)
-        return np.reshape(convolved_array, original_shape)
+    if filter_type == "conic":
+        filter_class = ConicFilter
+    elif filter_type == "circular":
+        filter_class = CircularFilter
+    else:
+        raise ValueError(f"Unsupported filter_type: {filter_type}")
```
tylerflex: "must be either 'conic' or 'circular'"?

tidy3d/plugins/autograd/invdes/filters.py
```diff
@@ -1,71 +1,226 @@
...
+    return filter_instance
 
 
 make_conic_filter = partial(make_filter, filter_type="conic")
 make_conic_filter.__doc__ = """make_filter() with a default filter_type value of `conic`.
 
 See Also
 --------
-make_filter : Function to create a filter based on the specified kernel type and size.
+:func:`~filters.make_filter` : Function to create a filter based on the specified kernel type and size.
```
tylerflex, line 218: what does tilde do? learning :D

tidy3d/plugins/autograd/invdes/parametrizations.py
```diff
@@ -1,46 +1,124 @@
...
+    filter : AbstractFilter
+        The filter to apply.
+    beta : float = 1.0
+        The beta parameter for the tanh projection.
+    eta : float = 0.5
+        The eta parameter for the tanh projection.
+    """
+
+    filter: AbstractFilter
+    beta: pd.NonNegativeFloat = 1.0
```
tylerflex: sort of confused why the filter is passed here but not the projector? (do we have a projector?) also would be good to include the full `pd.Field()` with the descriptions and titles there, instead of in docstring.

tidy3d/plugins/autograd/invdes/parametrizations.py
```diff
@@ -1,46 +1,124 @@
...
+        radius: Union[float, Tuple[float, ...]] = None,
+        dl: Union[float, Tuple[float, ...]] = None,
+        *,
+        size_px: Union[int, Tuple[int, ...]] = None,
+        beta: float = 1.0,
+        eta: float = 0.5,
+        filter_type: KernelType = "conic",
+        padding: PaddingType = "reflect",
+    ) -> FilterAndProject:
+        """Create a FilterAndProject instance from parameters.
```
tylerflex: backslash for `FilterAndProject`. 

What parameters? maybe clarify "from parameters for both ``make_filter`` and ``tanh_projection``"

tidy3d/plugins/autograd/invdes/parametrizations.py
```diff
@@ -1,46 +1,124 @@
...
+
+    @classmethod
+    def from_params(
+        cls,
+        radius: Union[float, Tuple[float, ...]] = None,
+        dl: Union[float, Tuple[float, ...]] = None,
+        *,
+        size_px: Union[int, Tuple[int, ...]] = None,
+        beta: float = 1.0,
+        eta: float = 0.5,
```
tylerflex: should the defaults be defined as module level constants perhaps? in case want to change them later more simply?

tidy3d/plugins/autograd/invdes/parametrizations.py
```diff
@@ -1,46 +1,124 @@
...
+            The type of filter kernel to use.
+        padding : PaddingType = "reflect"
+            The padding type to use for the filter.
+
+        Returns
+        -------
+        FilterAndProject
+            An instance of FilterAndProject.
+        """
+        if filter_type == "conic":
```
tylerflex: seems like duplicated code? maybe this should get written as a function or class method somewhere?

tidy3d/plugins/autograd/invdes/parametrizations.py
```diff
@@ -1,46 +1,124 @@
...
+            filter_class = ConicFilter
+        elif filter_type == "circular":
+            filter_class = CircularFilter
+        else:
+            raise ValueError(f"Unsupported filter_type: {filter_type}")
+
+        if size_px is not None:
```
tylerflex: I thought we had something that handles this for us? yea basically maybe all of this could be `make_filter(...)`?

tidy3d/plugins/autograd/invdes/penalties.py
```diff
@@ -1,86 +1,140 @@
...
+    delta_eta : float = 0.01
+        The binarization threshold for erosion and dilation operations.
     """
-    filtproj = make_filter_and_project(filter_size, beta, eta, padding=padding)
-    eta_dilate = 0.0 + delta_eta
-    eta_eroded = 1.0 - delta_eta
 
-    def _dilate(array: np.ndarray, beta: float):
-        return filtproj(array, beta=beta, eta=eta_dilate)
+    filtproj: FilterAndProject
```
tylerflex: use pd.Field instead of docstrings

tidy3d/plugins/autograd/invdes/penalties.py
```diff
@@ -1,86 +1,140 @@
...
-    delta_eta : float, optional
-        The binarization threshold for erosion and dilation operations. Default is 0.01.
-    padding : PaddingType, optional
-        The padding type to use for the filter. Default is "reflect".
-
-    Returns
-    -------
-    Callable
-        A function that computes the erosion/dilation penalty for a given array.
+    filtproj : FilterAndProject
```
tylerflex: hm, this is an interesting approach.. but maybe we need a good default value for this based on the default penalty params?

tidy3d/plugins/autograd/invdes/penalties.py
```diff
@@ -1,86 +1,140 @@
...
+        cls,
+        radius: Union[float, Tuple[float, ...]] = None,
+        dl: Union[float, Tuple[float, ...]] = None,
+        *,
+        size_px: Union[int, Tuple[int, ...]] = None,
+        beta: float = 100.0,
+        eta: float = 0.5,
+        delta_eta: float = 0.01,
+        padding: PaddingType = "reflect",
+    ) -> "ErosionDilationPenalty":
```
tylerflex: use future annotations

tidy3d/plugins/autograd/invdes/penalties.py
```diff
@@ -1,86 +1,140 @@
...
 
-    return _erosion_dilation_penalty
+    @classmethod
+    def from_params(
+        cls,
+        radius: Union[float, Tuple[float, ...]] = None,
+        dl: Union[float, Tuple[float, ...]] = None,
+        *,
+        size_px: Union[int, Tuple[int, ...]] = None,
+        beta: float = 100.0,
```
tylerflex: in my recent experience, this is probably too high. I might recommend like 20.. not sure if this breaks backwards compatibility technically

tylerflex (commented):
Looks pretty good. 

I'm having trouble visualizing the API / if there are any changes? especially to invdes? does it break backwards compatibility?

I also think it would make some sense to define the default parameters, especially for beta, as constants somewhere so we can tweak them more easily.

Thanks @yaugenst-flex

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: It is clearly defined, no? If `size_px` is given then that takes precedence over everything else. The reason being that if you supply an explicit size in pixels for the filter, then you probably do not want to calculate a different size from other arguments? Maybe it's enough to add this to the docstring?
I thought this was a relatively common pattern, see for example the `period` argument in [np.interp](https://numpy.org/doc/stable/reference/generated/numpy.interp.html).

yaugenst-flex: It is clearly defined, no? If `size_px` is given then that takes precedence over everything else. The reason being that if you supply an explicit size in pixels for the filter, then you probably do not want to calculate a different size from other arguments? Maybe it's enough to add this to the docstring?
I thought this was a relatively common pattern, see for example the `period` argument in [np.interp](https://numpy.org/doc/stable/reference/generated/numpy.interp.html).

tidy3d/plugins/autograd/functions.py
yaugenst-flex: Oh yeah I guess they do. Does it maybe make sense to keep working on `pre/2.8` for the features and then cherry-pick things for patches to `develop`? That would prevent the branches from diverging too much.

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: ooh yeah i see. i'll remove those then!

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: Yeah it's functionally identical. I like using `pass` to convey that something is not supposed to do anything and an ellipsis if something is to be implemented (like a todo in a sense). Hence `...` in method stubs for abstract classes. But I can change it if you want.

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: Yeah good point...

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: Yeah no this is not supposed to be here 😅

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: No because the kernel dimensionality is decided at runtime from the array that is passed in, so it needs to be a method. And a method with `self` as an argument shouldn't be cached because the cache will keep a reference to the object, meaning it can't be garbage collected, leading to memory leaks. Hence why it is a `staticmethod` (no reference to `self`).

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: I can change it, but what happens when we have 10 filters at some point 😅

tidy3d/plugins/autograd/invdes/filters.py
yaugenst-flex: > If you prefix the content with ~, the link text will only be the last component of the target. For example, :py:meth:`~Queue.Queue.get` will refer to Queue.Queue.get but only display `get` as the link text.

from [here](https://www.sphinx-doc.org/en/master/usage/referencing.html#cross-referencing-syntax)

I just googled how to reference class methods...

tidy3d/plugins/autograd/invdes/parametrizations.py
yaugenst-flex: Because we don't have a projector class 😆
I can make one, but right now I only made classes for stateful objects, i.e. the ones that had `make_`-functions associated with them. Since the projections are pure functions, I didn't do that for those. Do we want projector classes too?

tidy3d/plugins/autograd/invdes/parametrizations.py
yaugenst-flex: I'm pretty unhappy with the parameters and the `from_params()` constructor actually...
The problem is that in principle, the `from_params()` will just need to contain all arguments that are necessary to construct the classes that are used within the parametrization. Same goes for the penalty class. So it's a sort of catch-all method that really doesn't make it much simpler just constructing a Filter yourself and passing it to the default constructor. Not sure how to best solve that tbh :/

tidy3d/plugins/autograd/invdes/parametrizations.py
yaugenst-flex: Sounds good, but let's maybe put them in `plugins/autograd/constants.py`?

tidy3d/plugins/autograd/invdes/parametrizations.py
yaugenst-flex: The reason for this is that I thought that we might be getting rid of the functional interface, i.e. `make_filter()`. If you think we should keep it then I will reuse it here.

tidy3d/plugins/autograd/invdes/penalties.py
yaugenst-flex: Ah you mean not having the parametrization here be a customizable thing but built into the penalty? We can do that, wasn't sure which way is best.

tidy3d/plugins/autograd/invdes/penalties.py
yaugenst-flex: It would, because that default is used in a bunch of notebooks 😀 should we change it anyways?

yaugenst-flex:
> I'm having trouble visualizing the API / if there are any changes? especially to invdes? does it break backwards compatibility?

It doesn't break backwards compatibility, no, only if we decide to get rid of the functional interface altogether.

tidy3d/plugins/autograd/functions.py
tylerflex: I'm not sure, guess first are we committed to merging this into 2.8 or should we just get it into develop? no real preference here.

Then, I think we generally merge develop into the pre branches before release, so it will happen eventually.

tidy3d/plugins/autograd/invdes/filters.py
tylerflex: no strong preference, mostly curious

tidy3d/plugins/autograd/invdes/filters.py
tylerflex: we have a FilterType? can print that perhaps

tidy3d/plugins/autograd/invdes/filters.py
tylerflex: ooh, nice

tidy3d/plugins/autograd/invdes/parametrizations.py
tylerflex: no it's fine, just wondering.

tidy3d/plugins/autograd/invdes/parametrizations.py
tylerflex: not sure if this makes sense, but could try `FilterAndProject` just taking all the parameters as fields, and then having a `filter` property?  I might be confused though. (and/or a `from_filter` class method)

tidy3d/plugins/autograd/invdes/parametrizations.py
tylerflex: sure

tidy3d/plugins/autograd/invdes/parametrizations.py
tylerflex: if it doesn't re implement the same code, I'm fine keeping it.

tidy3d/plugins/autograd/invdes/penalties.py
tylerflex: yea or (similar comment above) we can make the parameters (beta, eta, etc.) be the pd.Fields here.

tylerflex: with good defaults

tidy3d/plugins/autograd/invdes/penalties.py
tylerflex: probably... but perhaps a goal before 2.8 should be to figure out very good defaults for these things.

tylerflex:
let's leave the functional interface but introduce these new classes and maybe change notebooks to use the classes?

tidy3d/plugins/autograd/invdes/filters.py
tylerflex: ok that's fine with me. I tend to be a bit conservative about things like this (tend towards erroring), if only to make things simpler, but if it's ok for numpy then I'm fine with it.

tylerflex:
@yaugenst-flex just checking, where did we leave off with this PR? are you waiting for my review?

yaugenst-flex:
I thin it was most of the way there, some unresolved things regarding the class constructors for the filters, but I'll finish it up and you can review again, this should be done for 2.8

yaugenst-flex:
@momchil-flex ~this is actually backwards compatible and does not require changes to the notebooks I'll run some to be sure but In principle the notebooks can be updated later~ -> nope

see https://github.com/flexcompute/tidy3d-notebooks/pull/174

PR #1780, opened 2024-06-20
fix finite precision bug when placing path integrals in CoaxialLumpedPorts

dmarek-flex:
Original implementation was not robust when the port center did not exactly equal the coordinate values.


weiliangjin2021:
Could you add a test?

tylerflex (approved):

weiliangjin2021 (approved):

Issue #1746, opened 2024-06-07
convenience function to compute frequency from wavelength


tylerflex:
would be nice to provide a shortcut to compute frequency from wavelength as we use frequency generally internally and lots of users work with wavelengths. In our examples and internally, we often have `freq = td.C_0 / wavelength`, but it would be nice for users to do this conversion without needing to know the formula.

Add a new utility function `td.frequency(wavelength)` that computes the frequency from a wavelength in `um`.

Perhaps we can define it in `tidy3d/constants.py`.

Then we should probably replace all instances of `td.C_0 / wavelength` with this function in both the source code and also the docs.

yaugenst-flex:
Closing — superseded by the FreqRange utility class added in PR #2542 (Jul 2025), which provides frequency/wavelength conversion and assignment.

PR #1783, opened 2024-06-21
Frequency conversion utility function

AndrewSykes1:
Adds feature described #1746


tylerflex:
Thanks @AndrewSykes1 ! We appreciate the contribution. After some internal discussion (which we should have updated in the original issue..). We think it would be better to have this function named `wvl_to_freq` so it's more explicit and there's less chance of confusion as to what it does.

Also, adding a changelog item under `added` would be useful to alert users to this (eg. `
```
- Convenience function for converting from wavelength (um) to frequency (Hz) using `td.wvl_to_freq(wvl)`
```

If you want, I can pick up this PR and make the changes, or you could modify it if you have time. Up to you.

Thanks again!

AndrewSykes1:
Sounds good, I'll have time to make the changes so should be no problem. Thanks for the correction!

PR #1784, opened 2024-06-22
Removed no sources in simulation log info.

e-g-melo:
The "No sources in simulation" log info was commented out to avoid showing this message in GUI.

image


e-g-melo:
Actually, I will resubmit this following these instructions (https://docs.flexcompute.com/projects/tidy3d/en/latest/development/index.html).

e-g-melo:
Hi! I am getting this error after commenting out that info log:
image

Should I remove this?
image

tests/data/invdes_history.json
momchil-flex, line 1: @tylerflex should we add this to gitignore?

momchil-flex (approved):
I think this is good, we can just squash and merge but I think we should add the `invdes_history.json` file to gitignore and remove from this PR. Tyler and I could do it once he confirms.

tests/data/invdes_history.json
tylerflex: sounds good to me

tylerflex (approved):
Thanks @e-g-melo The PR looks ok to me. Could you file an issue (if not already) so we can make sure to fix this in the long term? Not sure who to assign at this point.

PR #1789, opened 2024-06-27
🗻 autograd specify background permittivity of structure

tylerflex:
adds `Structure.autograd_background_permittivity : float = None` that can be used for permittivity outside.

Useful for structures that are holes in another material as the default permittivity outside is the simulation.medium

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for differentiation with respect to `ComplexPolySlab.vertices`.
 - Users can now export `SimulationData` to MATLAB `.mat` files with the `to_mat_file` method.
 - Introduce RF material library. Users can now export `rf_material_library` from `tidy3d.plugins.microwave`.
+- Users can specify the background medium for a structure in automatic differentiation by supplying `Structure.autograd_background_permittivity`.
 
 ### Changed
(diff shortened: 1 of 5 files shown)


yaugenst-flex (approved):
LGTM!

Would we eventually want to grab the permittivities automatically in the future? E.g., if there are different materials on the boundaries of the structure?

tylerflex:
> Would we eventually want to grab the permittivities automatically in the future? E.g., if there are different materials on the boundaries of the structure?

Yes, this would be ideal. The main issue is figuring out where to sample the permittivity from. We need to grab the permittivity far enough away from the subpixeled region because we're interested in the background permittivity, but we also dont want to go too far away from the boundary that we either run into another structure or go out of bounds of the structure. For `Box`, this is pretty easy, since the edges are along the cartesian dimensions, so we can just go eg 2 grid cells over from the edge. But for PolySlab and other geometries, we can either:
1. Run `Simulation.epsilon()` with and without the structure, which can be slow.
2. Pick a distance from each edge and scan along it to interpolate the permittivity. But then we need to figure out what distance to choose and this can be quite slow as well.

Not sure if you have other ideas here?

yaugenst-flex:
@tylerflex I think this needs a schema update? Getting an extra fields not permitted...

tylerflex:
I think the solver needs to be rebuilt with this front end.

PR #1790, opened 2024-06-27
🗻 complex polyslab autograd support

tylerflex:
quite a bit easier now :D


yaugenst-flex (approved):
Wow what the .. 😂
Really cool that it works like this though!

PR #1792, opened 2024-06-29
fix `PolySlab` in `autograd` gradient when `sidewall_angle` or `dilation` used

tylerflex:


yaugenst-flex (approved):
Does this mean differentiation w.r.t. `sidewall_angle` and `dilation` works or is this fix only for the `vertices`?

tylerflex:
> Does this mean differentiation w.r.t. sidewall_angle and dilation works or is this fix only for the vertices?

No, these don't work yet. In fact, I'm not super sure how to define them (at least how to define gradient w.r.t sidewall_angle).

Before this fix, basically everything worked fine, but I never tested with a non-zero dilation or sidewall angle. It turned out things broke. I will update the changelog to be more specific, I see how this is a bit confusing.

Issue #1397, opened 2024-01-25
add `ModeSolver.plot`


tylerflex:
Would be nice to have a simple method to visualize the `ModeSolver` plane within the simulation, perhaps with a different color than a regular monitor or source.

momchil-flex:
Also to visualize the plane cross-section (media/epsilon)

momchil-flex:
@e-g-melo this is related to you working on a 2D plot of the mode solver that also shows the PML right?

e-g-melo:
Hi @momchil-flex! Yes, it is.

tylerflex:
@e-g-melo did your PR fix this issue? just checking in

e-g-melo:
That PR allows you to plot the mode plane defined on `ModeSolver`. However, it won't include plotting the mode plane within the `Simulation` domain.

e-g-melo:
Not sure if that is necessary, right?

tylerflex:
I think it's ok and we can reopen if it's requested.

PR #1793, opened 2024-07-01
Added: mode plane plotting functions (#1397).

e-g-melo:
Added mode plane plotting functions: `plot`, `plot_eps`, `plot_structures_eps`, `plot_grid`, and `plot_pml`. #1397 

image

### DIFF
--- a/tidy3d/plugins/mode/mode_solver.py
+++ b/tidy3d/plugins/mode/mode_solver.py
@@ -10,6 +10,8 @@ from typing import Dict, List, Tuple
 import numpy as np
 import pydantic.v1 as pydantic
 import xarray as xr
+from matplotlib.collections import PatchCollection
+from matplotlib.patches import Rectangle
 
 from ...components.base import Tidy3dBaseModel, cached_property, skip_if_fields_missing
 from ...components.boundary import PML, Absorber, Boundary, BoundarySpec, PECBoundary, StablePML
@@ -41,6 +43,7 @@ from ...components.types import (
     Symmetry,
 )
 from ...components.validators import validate_freqs_min, validate_freqs_not_empty
+from ...components.viz import plot_params_pml
 from ...constants import C_0
 from ...exceptions import SetupError, ValidationError
 from ...log import log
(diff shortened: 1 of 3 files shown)


tidy3d/plugins/mode/mode_solver.py
```diff
@@ -1141,6 +1143,324 @@ def plot_field(
...
+                monitor_alpha=0,
+                ax=ax,
+            )
+
+        # Mode plane grid.
+        plane_grid = self.grid_snapped.centers
+        coord_0 = getattr(plane_grid, coord[0])[1:-1]
+        coord_1 = getattr(plane_grid, coord[1])[1:-1]
+
+        # Number of PML layers in ModeSpec.
```
tylerflex, line 1364: there is a `Simulation.pml_thicknesses` method. Could we use this?  or the `Simulation.plot_pml` to avoid duplicating work?

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -1141,6 +1143,324 @@ def plot_field(
             **sel_kwargs,
         )
 
+    def plot(
+        self,
+        ax: Ax = None,
+        **patch_kwargs,
+    ) -> Ax:
+        """Plot each mode plane simulation's components.
```
tylerflex: "each mode plane"? is this correct because it seems to only plot the single mode plane?

tylerflex (commented):
Thanks @e-g-melo, I put a few comments to suggest reorganizing so that we can re-use the logic.

Otherwise, there are a couple other things we should also add:
1. Please put a line in the `CHANGELOG.md` under ## added briefly describing the new functions.
2. Please add some test, probably in `tests/test_plugins/test_mode_solver.py` that just call each of these methods with a `ModeSolver` isntance. This way we are notified right away if anything breaks in the future.

Appreciate the contribution!

tidy3d/plugins/mode/mode_solver.py
e-g-melo: The `Simulation.pml_thicknesses` stores only the FDTD boundaries. The mode plane PML thickness is defined by `num_pml` in `ModeSpec`, which actually defines only the number of cells in each plane direction.

tidy3d/plugins/mode/mode_solver.py
e-g-melo: Thanks, it's a typo. I will remove it.

e-g-melo:
Thanks @tylerflex! I will implement the changes.

tidy3d/plugins/mode/mode_solver.py
tylerflex: I think one simple solution would be to make an `updated_copy()` of the simulation, and add some PML layers (just for plotting) corresponding to `num_pml`.  Then you should be able to plot PMLs (and the other methods) normally. Does that make sense / does it seem like it should work?

e-g-melo: It should work for mode planes spanning into the simulation domain boundaries, but it will fail for mode planes smaller than the simulation domain, right?

tylerflex: might be possible to use this method, or the `subsection` method
https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/plugins/mode/mode_solver.py#L1162-L1205

e-g-melo: And @momchil-flex commented that for the mode solver the PML is included within the mode plane limits.

e-g-melo:
Hi @tylerflex! I've addressed your comments.

@momchil-flex, please confirm if the mode solver PML lies within or outside the mode plane limits. If outside, I will make the changes suggested by @tylerflex.

Thanks!!

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -1141,6 +1143,324 @@ def plot_field(
...
+
+        Parameters
+        ----------
+        ax : matplotlib.axes._subplots.Axes = None
+            Matplotlib axes to plot on, if not specified, one is created.
```
momchil-flex, line 1157: Should probably also have `patch_kwargs` here.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -1141,6 +1143,324 @@ def plot_field(
...
+        axes.pop(n_axis)
+        a_center[n_axis] = self.plane.center[n_axis]
+        h_lim = [
+            a_center[n_axis] - self.plane.size[axes[0]] / 2,
+            a_center[n_axis] + self.plane.size[axes[0]] / 2,
+        ]
+        v_lim = [
+            a_center[n_axis] - self.plane.size[axes[1]] / 2,
+            a_center[n_axis] + self.plane.size[axes[1]] / 2,
+        ]
```
momchil-flex: As far as I can tell this same code block is used in 4 different places (and with a small modification in the fifth). It's always better to make a small helper function to avoid code duplication (otherwise it's always possible that someone modifies something here but not in the other instance of the same code).

For the axes, I think you could do a one line
```
n_axis, t_axes = self.plane.pop_axis([0, 1, 2], self.normal_axis) # I renamed axes to t_axes for "tangential"
```

Then you can also introduce a helper function like `self._center_and_lims(n_axis, t_axes)` which returns `a_center, h_lim, v_lim` and call that in all plotting functions here.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -1141,6 +1143,324 @@ def plot_field(
...
+            if pml_thick_1_minus > 0:
+                pml_rect.append(Rectangle((coord_0[0], coord_1[0]), mp_w, pml_thick_1_minus))
+            if pml_thick_1_plus > 0:
+                pml_rect.append(
+                    Rectangle((coord_0[0], coord_1[-num_pml_1 - 1]), mp_w, pml_thick_1_plus)
+                )
+
+            pc = PatchCollection(
+                pml_rect, alpha=0.7, facecolor="gray", edgecolor="gray", hatch="x", zorder=np.inf
+            )
```
momchil-flex, line 1408: Probably better to import `plot_params_pml` from [here](https://github.com/flexcompute/tidy3d/blob/af371ad9e59d3038c8a0166f9d74abfca749d17f/tidy3d/components/viz.py#L109) and then use `PatchCollection(pml_rect, **plot_params_pml.to_kwargs())`

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -1141,6 +1143,324 @@ def plot_field(
...
+                vlim=v_lim,
+                source_alpha=0,
+                monitor_alpha=0,
+                ax=ax,
+            )
+
+        # Mode plane grid.
+        plane_grid = self.grid_snapped.centers
+        coord_0 = getattr(plane_grid, coord[0])[1:-1]
+        coord_1 = getattr(plane_grid, coord[1])[1:-1]
```
momchil-flex: A bit nit-picky but getattr is often considered not good practice. In this particular case you don't need to define `coord` at all I believe. You can do

```
plane_grid = self.grid_snapped.centers.to_list
coord_0 = plane_grid[axes[0]]
coord_1 = plane_grid[axes[1]]
```

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -1141,6 +1143,324 @@ def plot_field(
...
+            lumped_element_alpha=0,
+            ax=ax,
+            **patch_kwargs,
+        )
+
+    def plot_eps(
+        self,
+        freq: float = None,
+        ax: Ax = None,
+    ) -> Ax:
```
momchil-flex, line 1192: Let's add an `alpha` parameter like in `Simulation.plot_eps`, it will be handy if we want to plot the fields and overlay the epsilon. https://github.com/flexcompute/tidy3d/blob/af371ad9e59d3038c8a0166f9d74abfca749d17f/tidy3d/components/simulation.py#L391

momchil-flex (commented):
Thanks @e-g-melo, this would be a very convenient addition. I have a few comments about the code organization. Apart from that, yeah, unfortunately the mode solver PML eat into the simulation rather than going out of it. This is a bit unfortunate but also a bit tricky to change as it affects all sorts of internals and just needs to be done carefully... But yeah probably makes it not very straightforward to follow Tyler's suggestion to use the Simulation pml plotting.

On the other hand could you address Tyler's suggestion to add a test for these new functions?

tidy3d/plugins/mode/mode_solver.py
e-g-melo: Hi! `patch_kwargs` is unused in `Simulation.plot()`, and there are no comments about it. I will include the comments in `ModeSolver.plot()` as soon as we understand that.

tidy3d/plugins/mode/mode_solver.py
e-g-melo: Thanks @momchil-flex! The code is much better now.

tidy3d/plugins/mode/mode_solver.py
e-g-melo: I've introduced `plot_params_pml`.

tidy3d/plugins/mode/mode_solver.py
e-g-melo: Thanks! Code changed

tidy3d/plugins/mode/mode_solver.py
e-g-melo: Done!

PR #1795, opened 2024-07-03
Support older version of xarray in EME smatrix_in_basis

caseyflex:
Several people have reported that the EME function `smatrix_in_basis`, for example in the MMI demo notebook, fails when using an old version of xarray (which we still want to support). The reason for the issue is the renaming `dims` -> `dim` in the arguments to `xr.DataArray.dot`; the new `dim` doesn't work under the old version of xarray.

This PR introduces a bit of error handling to support both versions of xarray. I would like to get it in as a patch since it affects the demo notebook.


momchil-flex:
Sounds good. Could you just add a CHANGELOG item to Fixed and then you can merge.

momchil-flex (approved):

PR #1797, opened 2024-07-03
Fixed docstrings for directivitydata and custom source

QimingFlex:
Fixed docstrings for both directivitydata and the custom source. Could you @momchil-flex double check if the custom source example make scene?


yaugenst-flex (approved):
Thanks for the fix! Looks good to me. We can maybe consider for some examples to include `# doctest: +SKIP` to just skip the test if we want to avoid setting up simulations and such in examples, that might make sense in cases where the setup is very cumbersome. Here it still seems fine though.

momchil-flex:
Yeah, looks good!

Issue #1633, opened 2024-04-24
Potential improvement items for EME solver


tomflexcompute:
Been testing the EME solver today and it's no doubt a great addition to the FDTD solver. Thanks @caseyflex for the great efforts in the past few months. This is highly requested by a number of users and will be used by many people in the future. Here I list a few potential improvement items from my testing so far. Maybe some of them are already in the roadmap or proposed by someone so feel free to ignore them in that case:

- [x] Currently there is no cost estimate before and after the simulation. Since we only charge the mode calculation, in principle the cost can be computed similarly to a regular mode solver task. Would be helpful to show the cost at least after the simulation. If the same estimate_cost can work for EMESimulation that would be even better. 
- [x] When the task is submitted, a message with the link to GUI is provided. Clicking the link leads to a message saying "EME task is not supported on the web currently". Since EME support on the web will take a good amount of time to be implemented, maybe we can remove that message for now.
- [x] When the task is running, there is no progress bar like the FDTD simulation. Would it be possible to provide some sort of progress bar or status indicators so users get a rough sense of the progress? When doing a length sweep of tens of simulations, the simulation time can be quite substantial. Would be good to indicate the task is running correctly. 
- [x] In the tutorial it mentions that the S parameter can be projected into a different basis. For simulating a 1x2 MMI for example, this is needed to compute the transmission to the two waveguides. Based on the API reference, doing so requires users to supply a mode field from a mode solver data. However getting this data is not super convenient at the moment. Initially I thought EMEModeSolverMonitor was for this purpose but it turned out not to be the case. Maybe we should provide a way so users can set up a mode solver directly in a EMESimulation? Let me know if I'm missing something and there is already a way to do it. 
<img width="843" alt="image" src="https://github.com/flexcompute/tidy3d/assets/116006359/1b85f7d6-3128-49f7-9073-58b88f1a982b">

- [x] `freqs` in `EMESimulation` only accepts lists. Regular `Simulation` also accepts numpy array for example. Would be nice to have the same behavior here for `EMESimualtion`. Same for `scale_factors` in `EMELengthSweep`.

- [x] A few minor issues on the EME tutorial. These are pretty small so no hurry in fixing them. I can fix them later.
1. In the notebook title we only capitalize the first word, i.e. "EME solver".  
[...]

PR #1802, opened 2024-07-05
Added support for EMESimulation in ModeSolver plugin

caseyflex:
Previously, we could add a `ModeSolverMonitor` to an `EMESimulation`, or we could convert the scene to a `Simulation` to use the `ModeSolver` plugin. Now, we can directly pass an `EMESimulation` to the `ModeSolver` plugin.

The change was suggested by @tomflexcompute here: https://github.com/flexcompute/tidy3d/issues/1633

It will also facilitate the planned `ModeSimulation` classes.

### DIFF
--- a/tidy3d/components/eme/data/sim_data.py
+++ b/tidy3d/components/eme/data/sim_data.py
@@ -32,12 +32,12 @@ class EMESimulationData(AbstractYeeGridSimulationData):
         "associated with the monitors of the original :class:`.EMESimulation`.",
     )
 
-    smatrix: EMESMatrixDataset = pd.Field(
-        ..., title="S Matrix", description="Scattering matrix of the EME simulation."
+    smatrix: Optional[EMESMatrixDataset] = pd.Field(
+        None, title="S Matrix", description="Scattering matrix of the EME simulation."
     )
 
     port_modes: Optional[EMEModeSolverData] = pd.Field(
-        ...,
+        None,
         title="Port Modes",
         description="Modes associated with the two ports of the EME device. "
         "The scattering matrix is expressed in this basis.",
--- a/tidy3d/plugins/mode/mode_solver.py
+++ b/tidy3d/plugins/mode/mode_solver.py
@@ -4,8 +4,9 @@ invariance along a given propagation axis.
 
 from __future__ import annotations
 
+from functools import wraps
 from math import isclose
-from typing import Dict, List, Tuple
+from typing import Dict, List, Tuple, Union
 
 import numpy as np
 import pydantic.v1 as pydantic
@@ -22,6 +23,8 @@ from ...components.data.data_array import (
 )
 from ...components.data.monitor_data import ModeSolverData
 from ...components.data.sim_data import SimulationData
+from ...components.eme.data.sim_data import EMESimulationData
+from ...components.eme.simulation import EMESimulation
 from ...components.geometry.base import Box
 from ...components.grid.grid import Grid
 from ...components.medium import FullyAnisotropicMedium
@@ -69,6 +72,25 @@ FIELD_DECAY_CUTOFF = 1e-2
 # Maximum allowed size of the field data produced by the mode solver
 MAX_MODES_DATA_SIZE_GB = 20
 
+MODE_SIMULATION_TYPE = Union[Simulation, EMESimulation]
+MODE_SIMULATION_DATA_TYPE = Union[SimulationData, EMESimulationData]
+
+
+def require_fdtd_simulation(fn):
+    """Decorate a function to check that ``simulation`` is an FDTD ``Simulation``."""
+
+    @wraps(fn)
+    def _fn(self, **kwargs):
+        """New decorated function."""
+        if not isinstance(self.simulation, Simulation):
+            raise SetupError(
+                f"The function '{fn.__name__}' is only supported "
+                "for 'simulation' of type FDTD 'Simulation'."
+            )
+        return fn(self, **kwargs)
+
+    return _fn
+
 
 class ModeSolver(Tidy3dBaseModel):
     """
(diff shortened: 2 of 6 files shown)


tidy3d/components/eme/data/sim_data.py
```diff
@@ -32,7 +32,7 @@ class EMESimulationData(AbstractYeeGridSimulationData):
         "associated with the monitors of the original :class:`.EMESimulation`.",
     )
 
-    smatrix: EMESMatrixDataset = pd.Field(
+    smatrix: Optional[EMESMatrixDataset] = pd.Field(
         ..., title="S Matrix", description="Scattering matrix of the EME simulation."
```
tylerflex: if it's optional, do we want it to have a default value? say `None`?

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -89,8 +111,11 @@ class ModeSolver(Tidy3dBaseModel):
         * `Prelude to Integrated Photonics Simulation: Mode Injection <https://www.flexcompute.com/fdtd101/Lecture-4-Prelude-to-Integrated-Photonics-Simulation-Mode-Injection/>`_
     """
 
-    simulation: Simulation = pydantic.Field(
-        ..., title="Simulation", description="Simulation defining all structures and mediums."
+    simulation: MODE_SIMULATION_TYPE = pydantic.Field(
+        ...,
+        title="Simulation",
+        description="Simulation defining all structures and mediums.",
```
tylerflex: `Simulation or EMESimulation` unless it's clear already from the type hint?

tidy3d/web/api/mode.py
```diff
@@ -319,7 +320,20 @@ def create(
...
-        mode_solver.simulation.validate_pre_upload(source_required=False)
+        if isinstance(mode_solver.simulation, Simulation):
+            mode_solver.simulation.validate_pre_upload(source_required=False)
+        elif isinstance(mode_solver.simulation, EMESimulation):
+            # TODO: support this web api in the future
+            raise SetupError(
+                "'EMESimulation' is not yet supported in the "
+                "remote mode solver web api. Please construct a 'Simulation' "
+                "with the same 'medium', 'structures', 'grid_spec', 'symmetry', etc. as the "
+                "'EMESimulation' in order to use the remote mode solver. "
```
tylerflex: should we provide a method to do this simply? and then we can support it?

tidy3d/web/api/mode.py
```diff
@@ -319,7 +320,20 @@ def create(
...
+                "Alternatively, you can add a 'ModeSolverMonitor' to the 'EMESimulation'."
+            )
+            # mode_solver.simulation.validate_pre_upload()
+        else:
+            raise SetupError("Simulation type not supported in the remote mode solver web api.")
```
tylerflex, line 337: this would never be raised based on the type of `simulation`? but I guess can;'t hurt to have? same comment in the sim_data one.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -993,6 +1028,7 @@ def to_mode_solver_monitor(self, name: str, colocate: bool = None) -> ModeSolver
             name=name,
         )
 
+    @require_fdtd_simulation
     def sim_with_source(
```
tylerflex, line 1032: is it possible to modify these to work with EME simulations? in principle they just create a modal source / monitors and update the simulation?

tylerflex (changes requested):

tidy3d/web/api/mode.py
caseyflex: I previously had a method `EMESimulation.to_fdtd_sim`, but we decided not to keep it because the `Simulation` was not physically equivalent to the `EMESimulation`. However, it does make sense to have something like this just for this temporary purpose of enabling remote mode solving, until it is natively supported. So I'll add `EMESimulation._to_fdtd_sim` as a private method, and suggest using it in this error message.

tidy3d/web/api/mode.py
caseyflex: Yeah, this is mostly for down the line when we add mode types, in particular `ModeSimulation`, just to make sure nothing slips through the cracks.

tidy3d/plugins/mode/mode_solver.py
caseyflex: Not really, because `EMESimulation` does not currently support any sources, or `ModeMonitor`. For `ModeSolverMonitor`, it is actually supported, so I'll get rid of the requirement there.

tidy3d/web/api/mode.py
tylerflex: ok. although I'd say it's probably ok to make it public (probably strange to suggest using a private method). but maybe with a disclaimer in the docstring?

tomflexcompute (approved):
That's a great addition. Thanks @caseyflex . I wonder if one uses the local mode solver to solve for the modes and dose `smatrix_in_basis` for example with the mode solving result, would the accuracy be noticeably worse?

tylerflex (approved):
make sure to update the changelog to put this in the Added  section of Unreleased instead of 2.7.1

caseyflex:
> That's a great addition. Thanks @caseyflex . I wonder if one uses the local mode solver to solve for the modes and dose `smatrix_in_basis` for example with the mode solving result, would the accuracy be noticeably worse?

I actually imagine it wouldn't matter so much. The mode profile should be similar, only the details and the `n_eff` would change a bit. And `n_eff` is very important for propagating EME modes but perhaps not so much for port modes.

momchil-flex (approved):
Looks good. Does it really facilitate the `ModeSimulation` refactor?

caseyflex:
> Looks good. Does it really facilitate the `ModeSimulation` refactor?

We would like to pass a `ModeSimulation` to the `ModeSolver` as well, without trying to come up with a way to convert it to an FDTD simulation.

caseyflex:
> > Looks good. Does it really facilitate the `ModeSimulation` refactor?
> 
> We would like to pass a `ModeSimulation` to the `ModeSolver` as well, without trying to come up with a way to convert it to an FDTD simulation.

(even if this isn't the primary web api, it will help internally to support it, and it may be helpful for users as well)

PR #1803, opened 2024-07-05
:cherries: pick bugfix from develop

dmarek-flex:
I merged this two-part bugfix into develop, but I also want it in pre/2.8 for testing new changes.


weiliangjin2021 (approved):
I'm not sure when we'll perform develop merging to pre/2.8. If not so soon, fine merging this first.

tylerflex (approved):

PR #1804, opened 2024-07-06
Wvltofreq feature

AndrewSykes1:
Adds feature described and changelog item associated in #1746

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Support for differentiation with respect to `GeometryGroup.geometries` elements.
 - Users can now export `SimulationData` to MATLAB `.mat` files with the `to_mat_file` method.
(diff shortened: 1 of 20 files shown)


momchil-flex:
Thanks for the contribution. Before we review this, it seems that right now there are merge conflicts. Could you try to rebase this against the latest `develop` branch and squash your commits into a single one? Let me know if you need help with this.

tylerflex:
Also @AndrewSykes1 I'm sorry to switch things up again but after internal discussion with @yaugenst, we think it would be best to go with `wvl2freq` because it's consistent with the numpy syntax (like [deg2rad](https://numpy.org/doc/stable/reference/generated/numpy.deg2rad.html) for instance). 

@momchil-flex @daquinteroflex do you have any objectives to this? or preferences?

daquinteroflex:
Hello, thanks for your contribution Andrew, this is great. Yeah the numpy function conversion makes sense as it's used in other python packages too so maybe that makes it more intuitive to users. `wvl2freq` sounds good to me . `lambda2freq` does sound nicer to me but less physically representative, so in this sense `wvl2freq` is more on point with this naming flow.

PR #1807, opened 2024-07-08
Lazily load `scipy.io` in `SimulationData.to_mat_file()`

yaugenst-flex:


momchil-flex (approved):
Thanks. No need for the changelog item since this feature has not yet been released, it is making its debut in 2.7.1.

tylerflex:
@m-bone FYI. I didnt catch this, apparently scipy.io has a long import time so we're doing it only when the function is called now.

yaugenst-flex:
image-mh

for reference

m-bone:
Ah nice find that's a ridiculous load time just for this

PR #1808, opened 2024-07-09
dummy version updater for base simulation

dbochkov-flexcompute:
A quick and dummy default updater in base simulation class so that version number is updated for heat and eme

### DIFF
--- a/tests/test_components/test_eme.py
+++ b/tests/test_components/test_eme.py
@@ -86,6 +86,17 @@ def make_eme_sim():
     return sim
 
 
+def test_sim_version_update(log_capture):
+    sim = make_eme_sim()
+    sim_dict = sim.dict()
+    sim_dict["version"] = "ancient_version"
+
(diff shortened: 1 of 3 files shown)


tylerflex (approved):
thanks @dbochkov-flexcompute. Seems good to me

Does this mean that the version will now be updated for `Simulation` before the `Updater` has a chance to update it? are there any possible unintended consequences?

dbochkov-flexcompute:
> Does this mean that the version will now be updated for Simulation before the Updater has a chance to update it? are there any possible unintended consequences?

no, I implemented this using the same name as existing function in `Simulation` https://github.com/flexcompute/tidy3d/blob/daniil/update-sim/tidy3d/components/simulation.py#L2020. So that existing function overrides the introduced one in case of `Simulation`. When we introduce breaking changes for heat or eme, then those classes should implement their own `_update_simulation` as necessary.

PR #1811, opened 2024-07-09
added plot_units to simulation and scene classes

dmarek-flex:
While working in the RF world I have been manually modifying plots so that the axes look good. But it is probably best to add that ability directly to tidy3d. I added an optional parameter to the Simulation and Scene classes that will automatically set the axes labels and title, along with a proper scaling of the tick labels. 

Here is an example without setting plot_units:

Screenshot from 2024-07-09 14-28-45

And here when using "mm":

Screenshot from 2024-07-09 14-28-27

### DIFF
--- a/tidy3d/components/base_sim/simulation.py
+++ b/tidy3d/components/base_sim/simulation.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Tuple
+from typing import Optional, Tuple
 
 import autograd.numpy as anp
 import pydantic.v1 as pd
@@ -16,9 +16,14 @@ from ..geometry.base import Box
 from ..medium import Medium, MediumType3D
 from ..scene import Scene
 from ..structure import Structure
-from ..types import TYPE_TAG_STR, Ax, Axis, Bound, Symmetry
+from ..types import TYPE_TAG_STR, Ax, Axis, Bound, LengthUnit, Symmetry
 from ..validators import assert_objects_in_sim_bounds, assert_unique_names
-from ..viz import PlotParams, add_ax_if_none, equal_aspect, plot_params_symmetry
+from ..viz import (
+    PlotParams,
+    add_ax_if_none,
+    equal_aspect,
+    plot_params_symmetry,
+)
 from .monitor import AbstractMonitor
 
 
@@ -102,6 +107,14 @@ class AbstractSimulation(Box, ABC):
         description="String specifying the front end version number.",
     )
 
+    plot_length_units: Optional[LengthUnit] = pd.Field(
+        "μm",
+        title="Plot Units",
+        description="When set to a supported ``LengthUnit``, "
+        "plots will be produced with proper scaling of axes and "
+        "include the desired unit specifier in labels.",
+    )
+
     """ Validating setup """
 
     # make sure all names are unique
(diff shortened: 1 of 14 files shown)


tidy3d/components/base_sim/simulation.py
```diff
@@ -102,6 +108,14 @@ class AbstractSimulation(Box, ABC):
         description="String specifying the front end version number.",
     )
 
+    plot_units: Optional[LengthUnit] = pd.Field(
+        None,
```
tylerflex: should it be default of `"um"`?

tidy3d/components/simulation.py
```diff
@@ -901,6 +934,12 @@ def set_plot_params(boundary_edge, lim, side, thickness):
         ax.set_xlim([ulim_minus, ulim_plus])
         ax.set_ylim([vlim_minus, vlim_plus])
 
+        # clean up the axis display
+        axis, position = Box.parse_xyz_kwargs(x=x, y=y, z=z)
```
tylerflex: worth combining these three lines in a single `Box` method? Seems they're almost always called together.

tidy3d/components/viz.py
```diff
@@ -343,3 +345,28 @@ class SimulationViewerInjector {
...
+    When the ``plot_units`` are specified, the plot axes are scaled, and
+    the title and axis labels include the desired units.
+    """
+    xlabel = axis_labels[0]
+    ylabel = axis_labels[1]
+    if plot_units is not None:
+        ax.set_xlabel(f"{xlabel} ({plot_units})")
+        ax.set_ylabel(f"{ylabel} ({plot_units})")
+        # Formatter to help plot in arbitrary units
+        scale_factor = UnitScaling[plot_units]
```
tylerflex: should we raise a custom `KeyError` if `plot_units` doesn't exist in UnitScaling?

tylerflex (changes requested):
Thanks @dmarek-flex! This will definitely be nice to have

tidy3d/components/base_sim/simulation.py
dmarek-flex: I could do that, but it will change the plot labels from previous versions. It would now include the "(um)" in axis labels, which was not added in previous tidy3d. When it is set to "None", the behavior should be exactly the same as previous tidy3d versions.

Let me know what you prefer

tylerflex: hm, I feel like probably they should always say "um"? so maybe I'd be in favor of simply including them from 2.8 onwards. we'll see what others say.

tylerflex: to clarify, always say "um" by default..

tomflexcompute: I think it is a good idea to always have the unit label to avoid confusion especially if we are going to support other units. Perhaps instead of "um", it can be the more technically correct "μm".

tomflexcompute (approved):

dmarek-flex:
@tylerflex any idea why that set of tests failed?

tylerflex:
Probably something doesn't like the μ.

tylerflex:
potentially related to the regex? https://github.com/flexcompute/tidy3d/blob/develop/scripts/make_script.py#L69

dmarek-flex:
@yaugenst-flex the more the merrier. I thought maybe you had changed make_script.py recently, but I think that was just the formatting commit.

tidy3d/components/base_sim/simulation.py
```diff
@@ -102,6 +107,14 @@ class AbstractSimulation(Box, ABC):
         description="String specifying the front end version number.",
     )
 
+    plot_units: Optional[LengthUnit] = pd.Field(
```
weiliangjin2021: Probably `plot_length_units` is clearer in case we have other units in the plot, e.g. time.

weiliangjin2021 (approved):
This is very helpful for now. Hopefully later we can set global units for time, length, etc.

yaugenst-flex (approved):
Looks great! Yeah `make_script.py` got formatted and I did some minor refactoring, didn't touch the actual functionality. There is currently a problem with generating scripts from simulations that contain `DataArray`s, but that's a bit tricky to sort out.

tylerflex (approved):

PR #1813, opened 2024-07-10
Local caching of simulations

yaugenst-flex:
Introduces caching of local simulations. Checks whether `SimulationData.simulation` from a local simulation data file has the same sha256 hash as the submitted simulation and returns the local copy if true.

The check is currently performed by hashing both the local and the submitted simulation, which is potentially a bit slow. We could also add the simulation hash as an attribute to the simulation data as suggested by @tylerflex so we can directly compare that, but that would need to be implemented on the backend?

Do we also want to support async/Batch?

### DIFF
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -134,11 +134,17 @@ class Tidy3dBaseModel(pydantic.BaseModel):
         except TypeError:
             return hash(self.json())
 
-    def _hash_self(self) -> str:
+    @cached_property  # cached because model is immutable
+    def sha256(self) -> str:
         """Hash this component with ``hashlib`` in a way that is the same every session."""
         bf = io.BytesIO()
         self.to_hdf5(bf)
         return hashlib.sha256(bf.getvalue()).hexdigest()
+
+    def _hash_self(self) -> str:
+        """Deprecated method. Use the `sha256` property instead."""
+        log.warning("The '_hash_self' method is deprecated. Use the 'sha256' property instead.")
+        return self.sha256
 
     def __init__(self, **kwargs):
         """Init method, includes post-init validators."""
--- a/tidy3d/web/api/webapi.py
+++ b/tidy3d/web/api/webapi.py
@@ -4,13 +4,12 @@ import os
 import time
 from datetime import datetime, timedelta
 from pathlib import Path
-from typing import Callable, Dict, List
+from typing import Callable, Dict, List, Union
 
 import pytz
 from requests import HTTPError
 from rich.progress import Progress
 
-from ...components.data.sim_data import SimulationData
 from ...components.types import Literal
 from ...exceptions import WebError
 from ...log import get_logging_console, log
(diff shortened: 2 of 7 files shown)


tidy3d/components/base.py
```diff
@@ -128,12 +128,18 @@ def __hash__(self) -> int:
         except TypeError:
             return hash(self.json())
```
yaugenst-flex, line 135: Note that at least since the introduction of `Tidy3dBaseModel.attrs`, `__hash__` will _always_ hash the json string (because dicts are not hashable), so the `try/except` here is somewhat superfluous. Should we do something about that @tylerflex?

tylerflex (commented):
Looks good! thanks @yaugenst-flex 

Regarding the individual discussions:

Regarding attrs: I forget what we decided but we should consider removing the attrs from the hash calculation.

I think we should also add this to `Job.run()`, `run_async`, and `Batch.run()`.

Not sure if we need backend changes to implement storing the hash in the SimulationData hdf5. might be possible to stick it in at the `web.download()` stage? or the `to_hdf5`?

What was the reason for renaming `hash_self` to `sha256`? just curious.

tidy3d/web/api/webapi.py
```diff
@@ -134,6 +139,18 @@ def run(
     :meth:`tidy3d.web.api.container.Batch.monitor`
         Monitor progress of each of the running tasks.
     """
+    if use_local_cache and Path(path).is_file():
+        local_sim = SimulationData.from_file(path)
+        if local_sim.simulation.sha256 == simulation.sha256:
```
momchil-flex: Our `sha256` is just the hdf5 bytes stream, so it *could* be possible to get those without loading the SimulationData, getting the simulation, and calling `sha256` on it. The `simulation` group in the `SimulationData` hdf5 should just be the same bytes as a `Simulation.to_hdf5()` would produce?

tidy3d/web/api/webapi.py
```diff
@@ -134,6 +139,18 @@ def run(
     :meth:`tidy3d.web.api.container.Batch.monitor`
         Monitor progress of each of the running tasks.
     """
+    if use_local_cache and Path(path).is_file():
+        local_sim = SimulationData.from_file(path)
```
momchil-flex: Maybe better to put this whole thing in a try-except and continue with the run if something goes wrong.

yaugenst-flex:
> I think we should also add this to `Job.run()`, `run_async`, and `Batch.run()`.

Ok I'll try to do that.

> Not sure if we need backend changes to implement storing the hash in the SimulationData hdf5. might be possible to stick it in at the `web.download()` stage? or the `to_hdf5`?

Ok yeah that sounds good, will look into it.

> What was the reason for renaming `hash_self` to `sha256`? just curious.

Couple of reasons:
 1. There is not really a good reason to hide this from the user.
 2. `_hash_self` clashes a bit with `__hash__(self)` in naming, even though they do completely different things.
 3. Related to 2 - it is (almost) self-explanatory what `.sha256` is.
 4. Not that it's likely that anyone is ever going to ask for that, but it establishes a convention for different unique identifier methods, e.g. maybe someone wants a `md5` checksum.

tidy3d/web/api/webapi.py
yaugenst-flex: That's true, but it would require a bunch more logic for getting the hdf5 stream and calculating the hash directly on that, basically reimplementing what's already in the base model? If we instead add the calculated hash to the simulation data directly (e.g. in `download()`), that might not be necessary?

momchil-flex: Yes, not sure what you mean in `download()` though?

Also it would still be nice to load the hash without doing `SimulationData.from_file`, but yeah it should be possible to read directly from the hdf5.

yaugenst-flex: The idea was to "intercept" the simulation data when it is being downloaded, hashing the simulation, and putting the hash in `.attrs` of the `SimulationData` so that it's available without needing to recalculate the hash. But I don't think that's feasible because there is no client-side assembly of the simulation data happening at all, so if we went that route, then that would need to be implemented on the backend. I can do the bytes thing though.

momchil-flex: I see, but maybe it makes more sense in `load` rather than in `download`, because `download` should purely download the file without actually loading the simulation data.

momchil-flex: Sorry, that's kind of like what you are saying in the second part of your comment. :) But yeah it could be done in the `load` function... but I don't think I like this route because I don't like the idea of the user-side and server-side files differing. If we want to do it we should probably just do it upon simulation init, or at least upload.

yaugenst-flex: This is proving to be a bit more tricky than I thought. The problem is that for single simulations, I can hash the byte stream just fine since it just means reading the file. However, for simulation data files, the file is split into the groups `data` and `simulation`, the latter being whose byte stream I need to hash. However, I haven't found a way to load a single hdf5 group as a byte stream from disk. I don't think this is even possible because hdf5 does not store groups contiguously (only datasets). So the only way I can see right now to hash the simulation correctly is to extract the bytestream as it is written _to_ disk, which does involve reconstructing the objects. Any ideas? Otherwise I'd perhaps just leave it like this for now, it's still way faster than going through the web pipeline.

yaugenst-flex:
Superseded by https://github.com/flexcompute/tidy3d/pull/2871

PR #1814, opened 2024-07-10
Improved mesh repair options in TriangleMesh

caseyflex:
Changes:

- Automatically process and validate mesh. In particular, this removes faces with area less than threshold, and merges close vertices. This was risky previously when the threshold was large, but now that the area threshold is 1e-36, it is hard for me to imagine a scenario where this causes difficulty. And automatic processing should make mesh import much more convenient.
- Adds functions `fix_winding` and `fill_holes` to make it easier to repair the mesh. I still didn't want to call these by default, as they can greatly change the mesh and may not always be desired. But they can make it easier to turn a non-watertight mesh into a watertight mesh with consistent winding.
- If `intersections_plane` fails, give an appropriate warning and use the bounding box instead. This makes certain validators more robust. Although I didn't have an example of a mesh that triggered the error, so I wasn't able to test this thoroughly.

### DIFF
--- a/tidy3d/components/geometry/mesh.py
+++ b/tidy3d/components/geometry/mesh.py
@@ -57,34 +57,71 @@ class TriangleMesh(base.Geometry, ABC):
         return val
 
     @pydantic.validator("mesh_dataset", always=True)
+    @verify_packages_import(["trimesh"])
     def _check_mesh(cls, val: TriangleMeshDataset) -> TriangleMeshDataset:
         """Check that the mesh is valid."""
         if val is None:
             return None
+
+        import trimesh
+
         mesh = cls._triangles_to_trimesh(val.surface_mesh)
         if not all(np.array(mesh.area_faces) > AREA_SIZE_THRESHOLD):
-            raise ValidationError(
+            old_tol = trimesh.tol.merge
+            trimesh.tol.merge = np.sqrt(2 * AREA_SIZE_THRESHOLD)
+            new_mesh = mesh.process(validate=True)
+            trimesh.tol.merge = old_tol
+            val = TriangleMesh.from_trimesh(new_mesh).mesh_dataset
+            log.warning(
                 f"The provided mesh has triangles with near zero area < {AREA_SIZE_THRESHOLD}. "
-                "Consider using numpy-stl's 'from_file' import with 'remove_empty_areas' set "
-                "to True and a suitable 'AREA_SIZE_THRESHOLD' to remove them."
+                "Triangles which have one edge of their 2D oriented bounding box shorter than "
+                f"'sqrt(2*{AREA_SIZE_THRESHOLD}) are being automatically removed.'"
+            )
+            if not all(np.array(new_mesh.area_faces) > AREA_SIZE_THRESHOLD):
+                raise ValidationError(
+                    f"The provided mesh has triangles with near zero area < {AREA_SIZE_THRESHOLD}. "
+                    "The automatic removal of these triangles has failed. You can try "
+                    "using numpy-stl's 'from_file' import with 'remove_empty_areas' set "
+                    "to True and a suitable 'AREA_SIZE_THRESHOLD' to remove them."
+                )
+        if not mesh.is_watertight:
+            log.warning(
+                "The provided mesh is not watertight. "
+                "This can lead to incorrect permittivity distributions, "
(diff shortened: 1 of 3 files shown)


lucas-flexcompute (approved):
Looks good. The only comment I have is regarding the added import of trimesh in `_check_mesh`. Would this have an impact in our loading time? I remember @daquintero worked on the import machinery to improve that.

caseyflex:
> Looks good. The only comment I have is regarding the added import of trimesh in `_check_mesh`. Would this have an impact in our loading time? I remember @daquintero worked on the import machinery to improve that.

I'm not too familiar with these changes. It looks like the method was already calling `_triangles_to_trimesh` which has an `import trimesh` inside it, so I don't think this PR actually changes anything on that front.

daquinteroflex:
Hello! Just have been going through the PR.  A lot of the discussion of the imports was in this PR if interested: https://github.com/flexcompute/tidy3d/pull/1300 

In any case, I think the approach we were taking was to try to reduce the top-level imports for non core packages like trimesh. (I personally think we should consider these non-core dependency imports in the 3.0 resutrucutre where we separate this into different modules separate from the top level tidy3d module, also in the same lines of @yaugenst-flex  dependency analysis recently)

The implementation in this PR  like [this](https://github.com/flexcompute/tidy3d/pull/1814/files#diff-3fac2c9dd48d82d516d02821a07c62192a303026921dc8b401c49e7fbd1f8d01R60-R66) seems in line to what we discussed previously when doing the import requirements checks, so seems good to me for now!

If we wanted to do the timing import check we can always just run the [ benchmarking import command from the docs](https://docs.flexcompute.com/projects/tidy3d/en/latest/development/recommendations.html#benchmarking-package-import) if interested:

```
poetry run tidy3d develop benchmark-timing-operations
```

tidy3d/components/geometry/mesh.py
```diff
@@ -57,35 +57,71 @@ def _warn_if_none(cls, val: TriangleMeshDataset) -> TriangleMeshDataset:
...
+            old_tol = trimesh.tol.merge
+            trimesh.tol.merge = np.sqrt(2 * AREA_SIZE_THRESHOLD)
+            new_mesh = mesh.process(validate=True)
+            trimesh.tol.merge = old_tol
+            val = TriangleMesh.from_trimesh(new_mesh).mesh_dataset
+            log.warning(
                 f"The provided mesh has triangles with near zero area < {AREA_SIZE_THRESHOLD}. "
-                "Consider using numpy-stl's 'from_file' import with 'remove_empty_areas' set "
-                "to True and a suitable 'AREA_SIZE_THRESHOLD' to remove them."
+                "These triangles are being automatically removed."
```
weiliangjin2021: Perhaps more precise warning: "Triangles which have one edge of their 2D oriented bounding box shorter than sqrt(2*threshold) are being automatically removed"

weiliangjin2021 (approved):

tidy3d/components/geometry/mesh.py
```diff
@@ -347,25 +384,42 @@ def intersections_plane(
...
+                    "This may be overly strict; consider using 'TriangleMesh.fill_holes' "
+                    "to repair the non-watertight mesh."
+                )
+            else:
+                log.warning(
+                    "Unable to compute 'TriangleMesh.intersections_plane'. "
+                    "Using bounding box instead."
+                )
+            log.warning(e)
+            return self.bounding_box.intersections_plane(x=x, y=y, z=z)
```
momchil-flex: Did you manage to test this part of the code with a simulation that hits this error? And shouldn't there be a unit test for it? Also maybe instead of `log.warning(e)`, `log.warning(f"Error encountered: {e}")` (I think this works but maybe `str(e)` is needed?)

caseyflex: Unfortunately, no -- even after playing around a bunch, I wasn't able to encounter this error. I think the trimesh library is also designed not to give this error, even for a wide range of meshes (e.g. even when not watertight, and even when this intersects the plane). I did improve the other tests a bit though.

On the bright side, although this isn't tested, it won't make the handling worse.

momchil-flex:
I can merge this now right @caseyflex ?

caseyflex:
> I can merge this now right @caseyflex ?

Yeah, should be fine to merge

PR #1818, opened 2024-07-11
Fixing DataArray.tracers for zero-sized data

momchil-flex:


tidy3d/components/data/data_array.py
```diff
@@ -81,7 +81,10 @@ def __init__(self, data, *args, **kwargs):
 
     @property
     def tracers(self) -> Box:
-        if AUTOGRAD_KEY not in self.attrs and not isbox(self.data.flat[0]):  # no tracers
+        if self.data.size == 0:
+            return None
+        elif AUTOGRAD_KEY not in self.attrs and not isbox(self.data.flat[0]):
```
tylerflex, line 86: might be slightly faster to assign `self.data.flat[0]` to a variable to avoid doing this twice. but I dont know how expensive this operation really is.

tylerflex (approved):
Looks good but would like @yaugenst-flex to take a look since he wrote this part.

yaugenst-flex (approved):
LGTM, I'm just wondering where this happened to cause a bug? I guess zero-sized data will always lead to an exception somewhere, not necessarily autograd-related?

tidy3d/components/data/data_array.py
yaugenst-flex: This should be very fast because it's just a view into the object I believe? But actually yeah this might be a good case for the [walrus operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions). Was avoiding it because I thought it was introduced way after 3.8... time flies.

tylerflex: not sure if we need to involve the walrus if we can just assign a variable?

yaugenst-flex: This seems pretty clean to me?
```python
@property
def tracers(self) -> Box:
    if self.data.size == 0:
        return None
    elif not isbox(flat := self.data.flat[0]) and AUTOGRAD_KEY not in self.attrs:  # switched order of evaluation so flat is always assigned
        return None
    elif isbox(flat):
        return anp.array(self.values.tolist())
    else:
        return self.attrs[AUTOGRAD_KEY]
```
Compared to
```python
@property
def tracers(self) -> Box:
    if self.data.size == 0:
        return None

    flat = self.data.flat[0]
    if AUTOGRAD_KEY not in self.attrs and not isbox(flat):
        return None
    elif isbox(flat):
        return anp.array(self.values.tolist())
    else:
        return self.attrs[AUTOGRAD_KEY]
```
~Note that we need to add this level of nesting since `self.data.flat[0]` might error when `self.data.size == 0`.~ (edited because stupid)

momchil-flex: You don't need `else` after `return`?

yaugenst-flex: oh man 😂  yeah

momchil-flex:
> LGTM, I'm just wondering where this happened to cause a bug? I guess zero-sized data will always lead to an exception somewhere, not necessarily autograd-related?

There's a backend test that tests some edge cases. I think one way this could happen is if you have a TimeMonitor with a start time that is after the time at which the simulation reaches early shutoff. Then the simulation data contains an array with size on the `t` coordinate of 0. Before autograd, this used to pass. :)

momchil-flex:
To recap should I be changing anything here or not? :)

yaugenst-flex:
I think it's fine as-is. Don't think the flatiter will be a bottleneck 😄

tylerflex:
agree

PR #1822, opened 2024-07-11
Differentiable local field projection

### LINKED ISSUE #1114: adjoint Field projection monitor support

yaugenst-flex:


tylerflex:
👀

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - More convenient mesh importing from another simulation through `grid_spec = GridSpec.from_grid(sim.grid)`.
 - `autograd` gradient calculations can be performed on the server by passing `local_gradient = False` into `web.run()` or `web.run_async()`.
 - Automatic differentiation with `autograd` supports multiple frequencies through single, broadband adjoint simulation when the objective function can be formulated as depending on a single dataset in the output `SimulationData` with frequency dependence only.
+- Automatic differentiation support for local field projection.
```
tylerflex: could you provide more detail on the specific call(s)? for example, include the name of the method(s) that are differentiable now? `FieldProjector(fld_mnt, ...).project(...)`

tidy3d/components/autograd/utils.py
```diff
@@ -2,12 +2,59 @@
...
+from autograd.tracer import getval, isbox
+
+from .constants import AUTOGRAD_KEY
+
+
+def _is_traced_objary(x: typing.Any) -> bool:
+    return isinstance(x, np.ndarray) and x.size > 0 and isbox(x.flat[0])
+
+
+def get_data(x: typing.Any) -> typing.Any:
```
tylerflex: docstrings here (and other functions)

tidy3d/components/base.py
```diff
@@ -956,8 +955,8 @@ def handle_value(x: Any, path: tuple[str, ...]) -> None:
             """recursively update ``field_mapping`` with path to the autograd data."""
 
             # this is a leaf node that we want to trace, add this path and data to the mapping
-            if isbox(x):
-                field_mapping[path] = x
+            if is_traced(x):
```
tylerflex: hm.. so just curious. if x is a DataArray, will this already just trigger this condition and then assign the `field_mapping[path]`? I'm worried that this logic will never reach `elif isinstance(x, DataArray)` when x is traced.

tylerflex (commented):
a few comments. I'm mostly a little confused about the base.py stuff?

tidy3d/components/autograd/utils.py
```diff
@@ -2,14 +2,37 @@
...
+    """Check whether x is traced (contains an autograd Box)"""
+    return isbox(x) or _is_traced_objary(x)
 
 
 def get_static(x: typing.Any) -> typing.Any:
     """Get the 'static' (untraced) version of some value."""
     return getval(x)
 
 
+def get_box(x: typing.Any) -> typing.Any:
```
tylerflex: `is_box` and `is_traced` look better to me here. Basically we're just adding a special case to handle `np.ndarray` with dtype=object and `ArrayBox` elements?

tidy3d/components/base.py
```diff
@@ -1013,13 +1012,10 @@ def insert_value(x, path: tuple[str, ...], sub_dict: dict):
...
             if isinstance(sub_element, DataArray):
-                current_dict[final_key] = sub_element.copy(deep=False, data=x)
-                if isbox(x):
-                    current_dict[final_key].attrs[AUTOGRAD_KEY] = x
-
-                elif AUTOGRAD_KEY in current_dict[final_key].attrs:
-                    current_dict[final_key].attrs.pop(AUTOGRAD_KEY)
+                current_dict[final_key] = sub_element.insert_tracers(x)
 
+                # if not is_traced(x) and AUTOGRAD_KEY in current_dict[final_key].attrs:
```
tylerflex: remove comment, handled in `insert_tracers`

tidy3d/components/data/data_array.py
```diff
@@ -24,7 +24,10 @@
     WATT,
 )
 from ...exceptions import DataError, FileError
+from ..autograd.constants import AUTOGRAD_KEY
 from ..autograd.functions import interpn
+
+# from ..autograd.utils import get_box, get_static, is_traced
```
tylerflex: remove comment

tidy3d/components/data/data_array.py
```diff
@@ -71,16 +71,23 @@ class DataArray(xr.DataArray):
     _data_attrs: Dict[str, str] = {}
 
     def __init__(self, data, *args, **kwargs):
-        """Initialize ``DataArray``."""
-
-        # initialize with untraced data
-        super().__init__(getval(data), *args, **kwargs)
-        # and put tracers in .attrs
+        super().__init__(data, *args, **kwargs)
```
tylerflex: note: this inserts the dtype=object into the `values`, as we discussed. we should revisit this choice later.

tidy3d/components/data/data_array.py
```diff
@@ -71,16 +71,23 @@ class DataArray(xr.DataArray):
...
         if isbox(data):
             self.attrs[AUTOGRAD_KEY] = data
```
tylerflex: note: we could consider just adding this to a `self.autograd_tracers` attribute instead of attrs. not sure why I used attrs originally.

tidy3d/components/autograd/utils.py
tylerflex: should we consider just calling this `get_tracer`? to be consistent with our naming?

tylerflex: or not sure if there's a better word for these things, maybe just use `autograd_box` or `box` everywhere?

tidy3d/components/data/data_array.py
```diff
@@ -71,16 +71,23 @@ class DataArray(xr.DataArray):
     _data_attrs: Dict[str, str] = {}
 
     def __init__(self, data, *args, **kwargs):
-        """Initialize ``DataArray``."""
```
tylerflex: put docstring back?

tidy3d/components/data/data_array.py
```diff
@@ -71,16 +71,23 @@ class DataArray(xr.DataArray):
...
         if isbox(data):
             self.attrs[AUTOGRAD_KEY] = data
 
     @property
-    def tracers(self) -> Box:
+    def has_tracers(self) -> bool:
+        return self.tracers is not None
+
+    @property
+    def static(self):
```
tylerflex: docstrings and type annotations for these properties.

tidy3d/components/data/data_array.py
```diff
@@ -71,16 +71,23 @@ class DataArray(xr.DataArray):
...
+        vals = self.values if self.values.size == 0 else np.vectorize(getval)(self.values)
```
tylerflex: does `getval` need a wrapper in `utils.py`? I'm not 100% sure about what happens here if `self.values` is a `np.ndarray` of `dtype=object`. my impression is that getval will not work here unless we apply it to `np.ndarray(self.values.tolist())` but could be wrong.

tidy3d/components/data/data_array.py
```diff
@@ -348,17 +360,23 @@ def interp(
             **coords_kwargs,
         )
 
+    def insert_tracers(self, tracers):
```
tylerflex: might be better to move right under init to keep it more organized

tidy3d/components/data/data_array.py
```diff
@@ -71,16 +71,23 @@ class DataArray(xr.DataArray):
...
 
     @property
-    def tracers(self) -> Box:
+    def has_tracers(self) -> bool:
+        return self.tracers is not None
+
+    @property
+    def static(self):
+        vals = self.values if self.values.size == 0 else np.vectorize(getval)(self.values)
+        static_self = self.copy(deep=False, data=vals)
```
tylerflex: I think we can use `insert_tracers` here instead. And then modify that method to handle the `attrs.pop(AUTOGRAD_KEY, None)`

We might also want to change the name of that method to reflect that we are not necessarily inserting tracers only. Maybe a more accurate name for it is `.update_with_data()`

tidy3d/components/data/data_array.py
tylerflex: we do have a `to_static()` in the `Tidy3dBaseModel`. Maybe one or both of those should change to be consistent with each other?

tidy3d/components/data/data_array.py
```diff
@@ -332,12 +339,17 @@ def interp(
 
             points = tuple(obj.coords[k] for k in obj.dims)
             xi = tuple(coords.get(k, obj.coords[k]) for k in obj.dims)
-            vals = interpn(points, obj.tracers, xi, method=method)
+            interp_vals = obj.values
+
+            if interp_vals.dtype == np.dtype("object"):  # dtype object
```
tylerflex: I feel like this could be its own method, or maybe it should use the `getbox` util?

tidy3d/components/data/data_array.py
```diff
@@ -332,12 +339,17 @@ def interp(
...
+                interp_vals = anp.array(interp_vals.tolist())
+
+            vals = interpn(points, interp_vals, xi, method=method)
 
             da = DataArray(vals, dict(obj.coords) | coords)  # tracers go into .attrs
-            if isbox(self.values.flat[0]):  # if tracing .values instead of .attrs
+            if isbox(vals):  # if tracing .values instead of .attrs
```
tylerflex: I think here we can use `self.insert_tracers(vals)`?

tidy3d/components/data/data_array.py
```diff
@@ -348,17 +360,23 @@ def interp(
...
+        traced_self = self.copy(deep=False, data=tracers)
+        if isbox(tracers):
+            traced_self.attrs[AUTOGRAD_KEY] = tracers
+        return traced_self
+
     def conj(self, *args: Any, **kwargs: Any):
         """Return the complex conjugate of this DataArray."""
-        if self.tracers is not None:
-            return self.__array_wrap__(anp.conj(self.tracers))
+        if isbox(self.values.flat[0]) or AUTOGRAD_KEY in self.attrs:
```
tylerflex: probably we can refactor this. or at least use `get_box` from utils? and for real as well

tylerflex (commented):
Some comments.. Mostly things I thought could be refactored, renamed, simplified. 

Otherwise I think it's consistent / correct. Only thing I'm not 100% sure on is in the `DataArray.static` property.

```py
vals = self.values if self.values.size == 0 else np.vectorize(getval)(self.values)

```
will `getval` work if `self.values` is a np.ndarray of dtype=object?

otherwise feeling pretty good about it in its current state

tidy3d/components/autograd/utils.py
yaugenst-flex: Yeah exactly, this got refactored to the point where it probably doesn't make sense to have anymore :laughing:
If we don't do it here, we'll still need to handle object arrays somewhere else though.

tidy3d/components/data/data_array.py
tylerflex: ah, maybe the vectorize already takes care of this.

tidy3d/components/autograd/utils.py
yaugenst-flex: I just used box because that's the autograd terminology. Either is fine. Tracer is probably less confusing though if we stick with one name.

tidy3d/components/data/data_array.py
yaugenst-flex: Oops. Although, to be fair, this is like
```python
# initialize x
x = 0
```

tidy3d/components/data/data_array.py
yaugenst-flex: yeah the vectorize is there exactly for that purpose. `np.array(self.values.tolist())` doesn't work actually because `getval` will not strip tracers from nested lists, i.e. if the array is >1d then the `tolist` approach will not work

tidy3d/components/data/data_array.py
yaugenst-flex: let's pick `to_static` then

tidy3d/components/autograd/utils.py
tylerflex: yea it makes sense, but maybe we can just use one word for this
* `tracer` = `ArrayBox / Box`
* `traced` = contains a `tracer`
* `get_tracer` = grab the `tracer` from something
* `untraced` = copy of something without the tracer
* `insert_tracer` = copy of something with the tracer inserted.
?

tidy3d/components/data/data_array.py
tylerflex: ah, good to know. Does this pop up anywhere else? (ie where the `anp.array(...tolist())` was called previously?) if so, maybe we should refactor it into one helper function

tylerflex: sorry, got a bit confused.. never mind

tidy3d/components/autograd/utils.py
tylerflex: object arrays only pop up in the DataArray I think, so maybe we can just handle it there? unless it makes it easier to do it here.

yaugenst-flex:
@tylerflex There are some changes to `DataArray` properties here. The naming and mechanics of these are probably not ideal, but since it's not user facing let's address them in the `DataArray` refactor? We'll end up changing most of these things anyways.
Also, #1939 lists this for `2.8rc1`, should I rebase on `pre/2.8`? Although I think it makes sense to include this only with the gradient fixes in `develop`. So either as part of `2.7.3` or rebase to `pre/2.8` only after `develop` is merged.

tylerflex:
> Also, https://github.com/flexcompute/tidy3d/issues/1939 lists this for 2.8rc1, should I rebase on pre/2.8? Although I think it makes sense to include this only with the gradient fixes in develop. So either as part of 2.7.3 or rebase to pre/2.8 only after develop is merged.

@momchil-flex what do you think? basically there were some changes in my (recently merged) #1923 that Yannick used in this PR. So ideally we would do another develop -> pre/2.8 merge before Yannick rebases against 2.8?

momchil-flex:
@yaugenst-flex what do you think will be simplest? I'm already running notebook tests for 2.7.3 and since it's hard to tell if this PR may not have various small effects, I'd rather not include it for now. However we could merge it to develop after 2.7.3 and then eventually into pre/2.8, or we can merge develop into pre/2.8 again and then you rebase the PR. Either way it will likely only come out in 2.8.0rc1 since we may not do any other 2.7. patches, but if we did the first approach would put it in there too.

yaugenst-flex:
@momchil-flex I think it makes sense either way to merge develop into pre/2.8 after the 2.7.3 release so we keep them somewhat in sync. I'll rebase to pre/2.8 after that, shouldn't be a problem.

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dask = "*"
 toml = "*"
 autograd = "1.6.2"
 scipy = "*"
+opt-einsum = "^3.3.0"
```
tylerflex: we need to be careful about this due to the danger of bloating our core dependencies. It became an issue recently and we agreed to let MC know if we add any dependencies. Is this absolutely necessary you think?

tests/ruff.toml
```diff
@@ -10,4 +10,5 @@ extend-ignore = [
     "E731", # lambda assignment
     "F841", # unused local variable
     "S101", # asserts allowed in tests
+    "NPY201", # numpy 2.* compatibility check
```
tylerflex, line 13: snuck in from another PR?

tidy3d/components/autograd/utils.py
```diff
@@ -2,14 +2,37 @@
...
     """Get the 'static' (untraced) version of some value."""
     return getval(x)
 
 
+def get_tracer(x: typing.Any) -> typing.Any:
+    """Get the 'boxed' (traced) version of some value."""
+    if not is_traced(x):
+        return None
+
+    if not _is_traced_objary(x):
```
tylerflex: pretty minor but slighly cleaner might be
```py
if _is_traced_objary(x):
     return anp.array(x.tolist())

return x
```
?

tidy3d/components/data/data_array.py
```diff
@@ -133,9 +161,8 @@ def _interp_validator(self, field_name: str = None) -> None:
         This does not check every 'DataArray' by default. Instead, when required, this check can be
         called from a validator, as is the case with 'CustomMedium' and 'CustomFieldSource'.
         """
-        # skip this validator if currently tracing for autograd because
-        # self.values will be dtype('object') and not interpolatable
-        if self.tracers is not None:
+        # skip this validator if currently tracing for autograd
+        if self.has_tracers:
```
tylerflex: why skip exactly? can't we interpolate tracer. containing DataArrays now?

tidy3d/components/data/data_array.py
```diff
@@ -352,15 +380,15 @@ def interp(
...
+        if isbox(self.values.flat[0]) or AUTOGRAD_KEY in self.attrs:
+            return self.insert_tracers(anp.conj(anp.array(self.values.tolist())))
         return super().conj(*args, **kwargs)
 
     @property
     def real(self):
         """Return the real part of this DataArray."""
-        if self.tracers is not None:
-            return self.__array_wrap__(anp.real(self.tracers))
+        if isbox(self.values.flat[0]) or AUTOGRAD_KEY in self.attrs:
```
tylerflex: perhaps we can refactor this logic out (eg. the one in real, conj, etc).

```py
def apply_traced(self, fn):

    if isbox(self.values.flat[0]) or AUTOGRAD_KEY in self.attrs:
        tracers = anp.array(self.values.tolist()))
        new_values = fn(tracers)
        return self.insert_tracers(new_values)

    return fn(self)
    
@property
def real(self):
    return self.apply_traced(anp.real)
```

tylerflex: also it seems the `anp.array(self.values.tolist())` and `if` statement both could use the convenience methods / functions used to check these things?

tidy3d/components/data/data_array.py
tylerflex, line 1: I am blanking on a lot of the discussions we had regarding the data array handing, but will mostly assume that this logic follows what we already talked about before?

Is there anything you think I should take a more detailed look at?

tidy3d/components/data/monitor_data.py
```diff
@@ -2232,8 +2231,10 @@ def power(self) -> xr.DataArray:
         ``xarray.DataArray``
             Power at points relative to the local origin.
         """
-        power_theta = 0.5 * np.real(self.Etheta.values * np.conj(self.Hphi.values))
-        power_phi = 0.5 * np.real(-self.Ephi.values * np.conj(self.Htheta.values))
+        conj = np.vectorize(np.conj)
+        real = np.vectorize(np.real)
+        power_theta = 0.5 * real(self.Etheta.values * conj(self.Hphi.values))
```
tylerflex: do we need to do this treatment in other parts of this class? I think this kind of operation does pop up from time to time?

tidy3d/plugins/autograd/README.md
```diff
@@ -144,7 +143,12 @@ The following components are traceable as outputs of the `td.SimulationData`
   - `SimulationData.get_intensity`
   - `SimulationData.get_poynting`
 
+- Local field projections using `td.FieldProjector.project_fields()` for:
```
tylerflex: was this mentioned as a "to do" item? if so would be good to remove it.

tylerflex (commented):
Looks really good! thanks @yaugenst-flex . I dont know if you want @QimingFlex to look at the field projection bits? I couldn't understand most of the physics of the operations, but did follow through the code and thought it looked good.

pyproject.toml
yaugenst-flex: `opt-einsum` is a pretty tiny dependency (~200kb) and itself only depends on numpy, and it's already a dependency of jax (although that's not core). Otherwise I can also rewrite that part using regular numpy einsum, not a big deal.

tests/ruff.toml
yaugenst-flex: Actually no, this got flagged (`np.trapz` usage) and I disabled it. We should enable it again once we support numpy 2.

tidy3d/components/autograd/utils.py
yaugenst-flex: good point

tidy3d/components/data/data_array.py
yaugenst-flex: Yeah this probably should have been removed in the interp PR...

tidy3d/components/data/monitor_data.py
yaugenst-flex: If we want to differentiate w.r.t. those methods then yeah this needs to be done. Although I'm hoping that we can get rid of this "hack" with a DataArray refactor...

tidy3d/plugins/autograd/README.md
yaugenst-flex: not seeing that anywhere in this PR, but I know I've seen it somewhere...

tidy3d/components/data/data_array.py
yaugenst-flex: Not really sure how to handle the `super()` call this way? But yeah I can at least partially refactor this..

tidy3d/components/data/data_array.py
yaugenst-flex: Yeah I think it's fine for the scope of this PR, so no need to dive deep for now, I vote that we put together a proper DataArray refactor soonish though.

yaugenst-flex:
Hey @QimingFlex, would be great if you could have a quick look at the field projection bits in this PR, I reworked some of it to more easily accomodate autograd. I did try to make sure that all of the field projection values that come out of this are still exactly the same as before, but it wouldn't hurt to have another pair of eyes on it.

tidy3d/components/data/data_array.py
tylerflex: do we need super or can `np.real(self)` suffice?

pyproject.toml
tylerflex: @lei-flex what do you think about adding this dependency?

tidy3d/components/data/data_array.py
yaugenst-flex: wouldn't that be infinite recursion? pretty sure `np.real(self)` will call `self.real`

tidy3d/components/field_projection.py
```diff
@@ -429,96 +425,68 @@ def _far_fields_for_surface(
...
+            constants=(0, 1, 2),
+            optimize="optimal",
+        )
 
-                    M[idx_u, i_th, j_ph] = self.integrate_2d(
-                        currents_f[f"H{cmp_1}"].values, phase_ij, pts[idx_u], pts[idx_v]
-                    )
+        jm = []
+        for eh_key in (f"E{cmp_1}", f"E{cmp_2}", f"H{cmp_1}", f"H{cmp_2}"):
+            vals = currents_f[eh_key].attrs.get(AUTOGRAD_KEY, currents_f[eh_key].values)
```
QimingFlex: How does autograd interact with the current computation here?

QimingFlex:
Hi @yaugenst-flex ,looks good to me, just a few questions regarding the surface current computation.

tidy3d/components/field_projection.py
```diff
@@ -429,96 +425,68 @@ def _far_fields_for_surface(
...
-
-                    J[idx_v, i_th, j_ph] = self.integrate_2d(
-                        currents_f[f"E{cmp_2}"].values, phase_ij, pts[idx_u], pts[idx_v]
-                    )
+        st = np.sin(theta)
+        ct = np.cos(theta)
+        sp = np.sin(phi)
+        cp = np.cos(phi)
+
+        expr = oe.contract_expression(
```
QimingFlex: I'm finding this part challenging to understand. The original code followed Taflove's textbook, which was more comprehensible. Would it be beneficial to add some explanatory comments here to clarify the logic and purpose of these operations?

tidy3d/components/field_projection.py
yaugenst-flex: this is related to how the traced values are stored. essentially, they can either be in `AUTOGRAD_KEY` in the `.attrs` dictionary, or they might be stored in the `DataArray`'s `.values` directly. So this line just extracts those. When no autograd is involved, this is equivalent to just calling `currents[eh_key].values`.

tidy3d/components/field_projection.py
yaugenst-flex: hey @QimingFlex sorry for the late reply on this. yeah I understand this is quite a big difference to how it was implemented previously. the calculations are the same, so it's still following taflove, but the implementation is obviously quite different.
this is because the calculation is now vectorized over theta and phi. the reason for this is mainly that this is much more efficient during the backward pass, otherwise the memory requirements can go through the roof when looping over each array element.
without gradients, this implementation is around as fast as the original implementation for a small number of theta, and faster for a large number of theta.

because of this vectorization, the einsum makes it much simpler to keep track of the dimensions in those arrays (i.e., it is clearly labeled which dimensions remain, which ones disappear / get summed over). but I understand your concern that this is less readable in terms of just following the computation.

do you think we should make this compromise or should I try to refactor it a bit?

QimingFlex: Thanks, @yaugenst-flex. I've reviewed the operations, and they look good to me! It's great to see some acceleration achieved for the frontend Near2far. I do have one suggestion: it might be better to keep with the original abbreviations, such as `phase` instead of `ehp`, as the former is more intuitive. The same applies to terms like `st` and `ct`. Although the original names are longer, they are generally easier to follow.

pyproject.toml
yaugenst-flex: I removed the opt-einsum dependency, using numpy instead

tidy3d/components/field_projection.py
yaugenst-flex: thanks! yeah that's a good point, I implemented your suggestions as well as some other readability improvements as well

QimingFlex (approved):

tylerflex (approved):

PR #1826, opened 2024-07-11
expose power wave amplitude computations

dmarek-flex:
I found that it might be necessary to get absolute values from an `smatrix` batch run. In my use case, it was to compute the radiation efficiency of an antenna. You need to compare the total radiated power (FieldProjectionMonitor) to the total power delivered to the antenna (coming from the port). Might be useful in the future to have access to these values as well for other computations.

I don't love the fact that I have to carry around the `SimulationData` and `BatchData` after running the `ComponentModeler` and `TerminalComponentModeler`. Maybe in the future we could create some kind of `ComponentModelerData`, which holds the `BatchData` along with post-processing results, as they are requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Users can specify the background medium for a structure in automatic differentiation by supplying `Structure.autograd_background_permittivity`.
 - `DirectivityMonitor` to compute antenna directivity.
 - Added `plot_length_units` to `Simulation` and `Scene` to allow for specifying units, which improves axis labels and scaling when plotting.
+- Added the helper function `compute_power_delivered_by_port`  to `TerminalComponentModeler` which computes power delivered to a microwave network from a port.
 
 ### Changed
 
--- a/tidy3d/plugins/smatrix/component_modelers/terminal.py
+++ b/tidy3d/plugins/smatrix/component_modelers/terminal.py
@@ -8,6 +8,7 @@ import numpy as np
 import pydantic.v1 as pd
 
 from ....components.base import cached_property
+from ....components.data.data_array import FreqDataArray
 from ....components.data.sim_data import SimulationData
 from ....components.geometry.utils_2d import snap_coordinate_to_grid
 from ....components.simulation import Simulation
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Users can specify the background medium for a structure in automatic differentiation by supplying `Structure.autograd_background_permittivity`.
 - `DirectivityMonitor` to compute antenna directivity.
 - Added `plot_length_units` to `Simulation` and `Scene` to allow for specifying units, which improves axis labels and scaling when plotting.
+- Added helper function to `TerminalComponentModeler` which computes power delivered to a microwave network from a port.
```
tylerflex: please mention the function name so users know what to call.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -249,3 +241,30 @@ def _check_grid_size_at_ports(
...
+        port: Union[LumpedPort, CoaxialLumpedPort], sim_data: SimulationData
+    ) -> FreqDataArray:
+        """Helper to compute the total power delivered to the network by a port for a given
+        simulation result. Units of power are Watts.
+        """
+        power_a, power_b = TerminalComponentModeler.compute_power_waves(
+            sim_data=sim_data, port=port
+        )
+        # Power delivered is the incident power minus the reflected power
+        return 0.5 * (np.abs(power_a) ** 2 - np.abs(power_b) ** 2)
```
tylerflex: just wondering, if these are power already, why is it taking absolute value squared?  or is this another notion of power?

tylerflex (approved):
Looks good to me! just question about definition of power and changelog.

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: that is a fair point. The equation is correct, but technically `power_a` is the amplitude of a "power wave". I'll choose a better name. FYI, the amplitudes are proportional to voltage and current:

image

weiliangjin2021 (approved):

PR #1829, opened 2024-07-12
🗻 autograd: broadband fit

tylerflex:


tylerflex:
I put everything into one unified approach #1830

PR #1830, opened 2024-07-12
🗻 autograd: broadband objectives with single adjoint simulation

tylerflex:


tylerflex:
@momchil-flex @yaugenst-flex sorry for the "bait and switch".  I just realized ultimately it makes more sense to unify the multi-frequency under one single approach rather than have various paths. So what I'm doing now is this:

The vip-containing `SimulationData` gives us a list of a bunch of adjoint sources. with `SourceTime` amplitudes corresponding to the amplitudes needed (assuming that the adjoint `Simulation` is normalized by that source time).

I sort all of the adjoint sources by source spatial dependence (eg. `Source` without the time dependence). Since the fitting approach only needs to be done source by source.

I then compute the coupling through `source_time.spectrum()` between all of the adjoint frequencies. This tells me how much injecting unit amplitude at frequency `i` affects frequency `j`.

Then I solve a linear system with least squares to find a new set of amplitudes to inject with each source. I also turn off `normalize_index` in the adjoint simulation, since this is taken care of in these amplitudes already`.

To avoid numerical issues, after collecting all of the adjoint sources, I normalize them with the `np.linalg.norm()` of the combined. This way, we always inject power with unit of amplitude 1 into the adjoint simulation and avoid divergence or really low field values (that potentially underflow?)

After the adjoint sim is run, the adjoint fields are multiplied by this `norm()` to get the expected values again.

Note: I have NOT implemented yet:
1. automatic `fwidth` reduction, if the fit is bad.
2. automatic `run_time` setting in adjoint source if (1) or if the adjoint source spectra combine to create something that lasts a long time.

I'll continue testing next week, and when it seems ready will release it for review, thanks!

tylerflex:
OK This PR is ready for review again.

Here's how it handles multi-frequency adjoint problems:

if the number of frequencies in the objective function is 1, it just does what we did before.

if the number of "ports" in the objective function (to put another way, if we can formulate a broadband adjoint source with one source). We do broadband adjoint with a single source and then post-normalize the adjoint fields using the vjp values.

Otherwise, we would use other more complicated approaches, in this case, I raised a NotImplementedError.

I'm still writing the two demos
1. WDM
2. bandpass filter.

Should be done tomorrow or thursday.

tidy3d/components/data/sim_data.py
```diff
@@ -36,6 +36,9 @@
 # maps monitor type (string) to the class of the corresponding data
 DATA_TYPE_NAME_MAP = {val.__fields__["monitor"].type_.__name__: val for val in MonitorDataTypes}
 
+# residuals below this are considered good fits for broadband adjoint source creation
+RESIDUAL_CUTOFF_ADJOINT = 0.5
```
yaugenst-flex: This seems really large? Or what's the typical scale here?

tidy3d/components/data/sim_data.py
```diff
@@ -953,13 +956,18 @@ def source_spectrum_fn(freqs):
 
     def make_adjoint_sim(
         self, data_vjp_paths: set[tuple], adjoint_monitors: list[Monitor]
-    ) -> Simulation:
+    ) -> tuple[Simulation, float]:
```
yaugenst-flex: Wondering if it would make sense to introduce some info object here similar to the `DerivativeInfo` that we have already instead of carrying around (a potentially increasing number of) outputs.

tidy3d/components/data/sim_data.py
```diff
@@ -971,34 +979,244 @@ def make_adjoint_sim(
...
 
+    @staticmethod
+    def get_amp(src_time: GaussianPulse) -> complex:
+        """grab the complex amplitude from a ``SourceTime``."""
+        mag = src_time.amplitude
+        phase = np.exp(1j * src_time.phase)
+        return mag * phase
+
+    @staticmethod
+    def set_amp(src_time: GaussianPulse, amp: complex) -> GaussianPulse:
```
yaugenst-flex: Bit nitpicky but this is not really `set`ting anything. It's more like a `from_amp`?

tidy3d/components/data/sim_data.py
```diff
@@ -971,34 +979,244 @@ def make_adjoint_sim(
...
+        log.info(
+            "Several adjoint sources, from one monitor. "
+            "Only difference between them is the source time. "
+            "Constructing broadband adjoint source and performing post-run normalization "
+            f"of fields with {len(post_norm_amps)} frequencies."
+        )
+
+        return [src_broadband], post_norm_amps
+
+    def _make_broadband_source(self, adj_srcs: list[Source], num_fwidth: float = 0.5) -> Source:
```
yaugenst-flex: `num_fwidth` not used?

tidy3d/components/medium.py
```diff
@@ -2551,8 +2551,11 @@ def _derivative_field_cmp(
         eps_data: PermittivityDataset,
         dim: str,
     ) -> np.ndarray:
-        coords_interp = {key: val for key, val in eps_data.coords.items() if len(val) > 1}
-        dims_sum = {dim for dim in eps_data.coords.keys() if dim not in coords_interp}
+        """Compute derivative with respect to the ``dim`` components within the custom medium."""
+
+        coords_interp = {key: eps_data.coords[key] for key in "xyz"}
+        coords_interp = {key: val for key, val in coords_interp.items() if len(val) > 1}
+        dims_sum = [dim for dim in "xyz" if dim not in coords_interp]
```
yaugenst-flex, line 2558: Not related to this PR but I think we are using this sort of code quite often, maybe we should have some utilities for this.

tidy3d/components/medium.py
```diff
@@ -2581,8 +2584,11 @@ def _derivative_field_cmp(
 
         # TODO: probably this could be more robust. eg if the DataArray has weird edge cases
         E_der_dim = E_der_map[f"E{dim}"]
-        E_der_dim_interp = E_der_dim.interp(**coords_interp).fillna(0.0).sum(dims_sum)
-        vjp_array = np.array(E_der_dim_interp.values).astype(complex)
+        E_der_dim_interp = E_der_dim.interp(**coords_interp).fillna(0.0).sum(dims_sum).real
+        E_der_dim_interp = E_der_dim_interp.sum("f")
+
+        vjp_array = np.array(E_der_dim_interp.values).astype(float)
```
yaugenst-flex: `np.array(..., dtype=float)`? Avoids potentially unnecessary casting.

tidy3d/components/data/sim_data.py
yaugenst-flex: After reading through the whole thing, maybe it would make sense to return, instead of a simulation, some `AdjointData` type similar to `SimulationData`, that containts the simulation as well as other info we might need? Not sure if it's a good idea but maybe worth considering.

yaugenst-flex (approved):
Cool not much to say other than the comments that I left!

tidy3d/components/data/sim_data.py
tylerflex: This is a relic of my testing :D but it's not used. I'd say a reasonable number is maybe like 1e-6 but it's not 100% clear.

tidy3d/components/data/sim_data.py
tylerflex: yea that might be a good idea (similar to `DerivativeInfo`?). I could do that to clean things up.

tidy3d/components/data/sim_data.py
tylerflex: yea, I even considered putting these methods inside of `GuassianPulse` directly.. maybe I should

tidy3d/components/data/sim_data.py
tylerflex: ah true, thanks.

tidy3d/components/medium.py
tylerflex: hm, you mean like an interp that ignores dimensions with 1 coordinate?

tidy3d/components/data/sim_data.py
tylerflex: Added `GaussianPulse.amp_complex` property as well as `GaussianPulse.from_amp_complex()` class method constructor and modified the code to use these.

tidy3d/components/data/sim_data.py
tylerflex: alright, I made am `AdjointSourceInfo` in `sim_data.py` to store everything. seems cleaner and I can extend it easily later.

```py
class AdjointSourceInfo(Tidy3dBaseModel):
    """Stores information about the adjoint sources to pass to autograd pipeline."""

    sources: tuple[Source, ...] = pd.Field(
        ...,
        title="Adjoint Sources",
        description="Set of processed sources to include in the adjoint simulation."
    )

    post_norm: Union[float, DataArray] = pd.Field(
        ...,
        title="Post Normalization Values",
        description="Factor to multiply the adjoint fields by after running "
        "given the adjoint source pipeline used."
    )

    normalize_sim: bool = pd.Field(
        ...,
        title="Normalize Adjoint Simulation",
        description="Whether the adjoint simulation needs to be normalized "
        "given the adjoint source pipeline used."
    )

```

tidy3d/components/data/sim_data.py
```diff
@@ -953,52 +980,256 @@ def source_spectrum_fn(freqs):
...
+            src_spatial_json = src.json(exclude={"source_time"})
+            json_to_sources[src_spatial_json] = src
+            spatial_to_src_times[src_spatial_json].append(src.source_time)
+
+        num_ports = len(spatial_to_src_times)
+        num_unique_freqs = len({src.source_time.freq0 for src in adj_srcs})
+
+        # next, figure out which treatment / normalization to apply
+        if num_unique_freqs == 1:
+            log.info("adjoint source creation: one unique frequency, no normalization")
```
momchil-flex: Capital letter, period in the end? :) Also for the two other log.info-s below.

I was also thinking of asking if at least some of those `log.info`-s you've put around in the autograd handling shouldn't be `log.debug`. I mean, not very important, but maybe we should have some policy?

tidy3d/components/data/sim_data.py
```diff
@@ -953,52 +980,256 @@ def source_spectrum_fn(freqs):
...
+
+        # dictionary mapping unique spatial dependence of each Source to list of time-dependencies
+        json_to_sources = defaultdict(None)
+        spatial_to_src_times = defaultdict(list)
+        for src in adj_srcs:
+            src_spatial_json = src.json(exclude={"source_time"})
+            json_to_sources[src_spatial_json] = src
+            spatial_to_src_times[src_spatial_json].append(src.source_time)
+
+        num_ports = len(spatial_to_src_times)
```
momchil-flex: So you are excluding the `SourceTime` of all adjoint sources, and counting their unique definitions after that exclusion?

In some cases, like when differentiating through FieldData.flux, the adjoint source would be a CustomFieldSource, right? Does this work for those too?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Mode solver plugin now supports 'EMESimulation'.
 - `TriangleMesh` class: automatic removal of zero-area faces, and functions `fill_holes` and `fix_winding` to attempt mesh repair.
+- Automatic differentiation with `autograd` supports multiple frequencies through single, broadband adjoint simulation when the objective depends on a single port.
```
momchil-flex: I'm afraid "a single port" is not specific enough, like people may think of a physical port rather than one mode of the physical port, and even for the same mode if we have both forward and backward direction dependence this still will not be applied?

tidy3d/components/medium.py
momchil-flex: Maybe we already have it [here](https://github.com/flexcompute/tidy3d/blob/85bc87284aca992c63e2e21bcd008bfcfd5cb349/tidy3d/components/grid/grid.py#L168) and the corresponding [internal](https://github.com/flexcompute/tidy3d/blob/85bc87284aca992c63e2e21bcd008bfcfd5cb349/tidy3d/components/grid/grid.py#L54) that is called for xarray interpolation. One of the purposes of this special interpolation function is to disregard a dimension with a single coordinate along a given dimension, and just `isel` that instead.

tidy3d/components/source.py
```diff
@@ -201,6 +201,20 @@ def end_time(self) -> float | None:
 
         return self.offset * self.twidth + END_TIME_FACTOR_GAUSSIAN * self.twidth
 
+    @property
+    def amp_complex(self) -> complex:
+        """grab the complex amplitude from a ``SourceTime``."""
+        mag = self.amplitude
```
momchil-flex: Nit but why define this at all, just use `self.amplitude` in the return.

Also, currently this is not a private property, it will appear in the documentation, so let's have a properly formatted docstring starting with a capital letter? And in `from_amp_complex` we should probably have the arguments descriptions. A bit annoying, I know, and maybe even not worth the documentation clutter, so we could make them private?

tidy3d/web/api/autograd/autograd.py
```diff
@@ -485,28 +487,13 @@ def _run_bwd(
...
-                "There is likely zero output in the data, or you have no traceable monitors. "
-                "As a result, the 'SimulationData' returned has no contribution to the gradient. "
-                "Skipping the adjoint simulation. "
-                "If this is unexpected, please double check the post-processing function to ensure "
-                "there is a path from the 'SimulationData' to the objective function return value."
-            )
-
-            # TODO: add a test for this
-            # construct a VJP of all zeros for all tracers in the original simulation
-            return {path: 0 * value for path, value in sim_fields_original.items()}
```
momchil-flex, line 589: What was the reason to add and now remove this? Like are you sure it cannot happen anymore?

tidy3d/components/data/sim_data.py
tylerflex: > Use log.info to log general information about the application's execution, such as key events and major steps, which are useful for end users or system administrators. Use log.debug for detailed diagnostic information useful for developers, such as internal states and variable values during execution, to troubleshoot specific issues.

from chat GPT, to me it seems like this falls under info? because it's more generally something that the user might want to see and also it basically just tells us what major steps are happening internally?

tidy3d/components/data/sim_data.py
tylerflex: > So you are excluding the SourceTime of all adjoint sources, and counting their unique definitions after that exclusion?

That's right.

> In some cases, like when differentiating through FieldData.flux, the adjoint source would be a CustomFieldSource, right? Does this work for those too?

Good question. I'm not really sure to be honest. Right now it would treat two otherwise identical CustomFieldSource with different data as separate "ports".  I think this is the proper handling though, since the data in the source defines the spatial dependence of `J(r, ω)` and we're trying to do here is essentially separate `J(r)` and `J(ω)`, so two different datasets should be treated as different `J(r)` I think.

The only case I think this could become slightly ambiguous for the user is if they had two identical `FieldMonitor` "`a`, and `b` and then for some reason `a` and `b` had different contributions to the objective function. In that case, we could come up with a scenario where the adjoint sources for `a` and `b` could theoretically be combined into one (ie by just summing the datasets with the corresponding source time amplitudes). But this seems like such an edge case that it's probably not worrying about to me.

tylerflex: just to clarify, in the case described above, we would simply require a separate adjoint simulation for `a` and `b`, but it wouldn't error in an unexpected way.

CHANGELOG.md
tylerflex: I'm using the same nomenclature that's in the smatrix plugin and I think photonforge as well. but I think if it clarifies thing, I can just replace with `when the objective function can be formulated as depending on a single dataset in the output 'SimulationData' that has frequency dependence only`.

how does that sound?

tidy3d/components/source.py
tylerflex: sounds good to all, I'll just add the arguments

tidy3d/web/api/autograd/autograd.py
tylerflex: that's right.

tidy3d/components/medium.py
tylerflex: this looks almost like what I need, but I still need to pass the coordinates of the points. it seems to me that spatial interp doesn't accept any coordinates?

momchil-flex: It's a method of our own `Coords` class. It does feel a bit weird that we can only use it for spatial interp though. But yeah the usage is kind of reverse, instead of `array.interp(coords)` we do something like `coords.spatial_interp(array)`. I don't remember why it came about like this...

tidy3d/components/data/sim_data.py
momchil-flex: My worry was rather that I'm still not sure what happens if you look at the json only, i.e. if the actual data gets compared or just a dummy placeholder tag or something that would be identical for all custom sources. If it works the way you describe it sounds good.

tidy3d/components/data/sim_data.py
momchil-flex: Fair

CHANGELOG.md
momchil-flex: Yep!

tidy3d/components/data/sim_data.py
tylerflex: Hm.. actually the way it is set up, I suppose the datasets would be excluded from the jsons. So two identical FieldMonitors a and b would actually be considered the same spatial dependent AdjointSource, and therefore would potentially confuse the broadband adjoint. I'll probably need to look into testing this as an edge case. but for otherwise different FieldMonitor definitions, it will have no problem.

tidy3d/components/medium.py
tylerflex: I see. for now I'll probably just leave it the way it is. but Yannick and I have planned a more systematic overhaul of the data array operations, so we can include refactoring the interp a bit in that?

momchil-flex: sounds good

tidy3d/components/data/sim_data.py
momchil-flex: Yeah, that's the edge case I had in mind. Although... technically if they're identical in their definitions, the recorded datasets should also be the same?

Ah but something weird may happen where the two are used in different ways in the objective function and so the adjoint source currents are different?

Very edgy.

tylerflex: well the recorded datasets would be the same, but what we get back is the VJP not the recorded datasets. so yea if the data is used in two different ways, it wouldn't show up.

I think one way to distinguish spatial dependence this could be to instead make a copy of all of the adjoint sources with the **same** source time, and then hash these. But then I dont know exactly if the datasets are included in the hash or not..

tylerflex: yea so I gave this a stab (hashing temporary copies of the sources with identical source_time) and it seemed to work well

https://github.com/flexcompute/tidy3d/commit/93e3b31223e0678462fb76578133df1c95fbdae1

This should be good enough.

PR #1831, opened 2024-07-12
prioritize snapping points in AutoGrid

dmarek-flex:
I found the snapping_points idea very useful, but also that the default value for the `min_step` was too restrictive. Here I propose some changes that give priority to the snapping points over the computed interval boundaries.

1. I changed the behavior so that when `dl_min` is defined, it becomes the size used as the `min_step`. This is needed for Simulations with PECs since the calculated `max_dl` is much too large to be useful.
2. If the snapping_point is close to an existing boundary it will replace it. The logic I used should still result in intervals that satisfy the `min_step` requirement though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added the helper function `compute_power_delivered_by_port`  to `TerminalComponentModeler` which computes power delivered to a microwave network from a port.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


weiliangjin2021 (commented):
This is great, thanks! Just add a test and CHANGELOG and good to merge.

dmarek-flex:
> This is great, thanks! Just add a test and CHANGELOG and good to merge.

Done

dmarek-flex:
@tylerflex @weiliangjin2021 Any last comments?

tylerflex (approved):

weiliangjin2021 (approved):

PR #1834, opened 2024-07-15
fix: remove `autograd.py`

yaugenst-flex:
The old `autograd.py` did not get deleted when we introduced `autograd/`, which currently breaks at least doctest collection.
Also skips one doctest.

Note: I think this happend in #1769


tylerflex (approved):

tylerflex:
Do you have merge privileges? if so, go for it (rebase and merge option)

image

PR #1836, opened 2024-07-15
simulation.subsection to include PML

weiliangjin2021:
~~ElectromagneticFieldData.dot to reindex coords for aligning almost identical coordinates~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mode solver plugin now supports 'EMESimulation'.
 - `TriangleMesh` class: automatic removal of zero-area faces, and functions `fill_holes` and `fix_winding` to attempt mesh repair.
 - Added progress bar for EME simulations.
+- Support for grid specifications from grid cell boundary coordinates via `CustomGridBoundaries` that subclasses from `GridSpec1d`.
+- More convenient mesh importing from another simulation through `grid_spec = GridSpec.from_grid(sim.grid)`.
 
 ### Changed
 - Error if field projection monitors found in 2D simulations, except `FieldProjectionAngleMonitor` with `far_field_approx = True`. Support for other monitors and for exact field projection will be coming in a subsequent Tidy3D version.
--- a/tidy3d/components/grid/grid_spec.py
+++ b/tidy3d/components/grid/grid_spec.py
@@ -8,7 +8,7 @@ from typing import List, Tuple, Union
 import numpy as np
 import pydantic.v1 as pd
 
-from ...constants import C_0, MICROMETER, fp_eps
+from ...constants import C_0, MICROMETER, inf
 from ...exceptions import SetupError
 from ...log import log
 from ..base import Tidy3dBaseModel
@@ -136,6 +136,78 @@ class GridSpec1d(Tidy3dBaseModel, ABC):
         add_right = bounds[-1] + last_step * np.arange(1, num_layers[1] + 1)
         return np.concatenate((add_left, bounds, add_right))
 
+    @staticmethod
+    def _postprocess_unaligned_grid(
+        axis: Axis,
+        simulation_box: Box,
+        machine_error_relaxation: bool,
+        bound_coords: Coords1D,
+    ) -> Coords1D:
+        """Postprocess grids whose two ends  might be aligned with simulation boundaries.
+        This is to be used in `_make_coords_initial`.
+
+        Parameters
+        ----------
+        axis : Axis
+            Axis of this direction.
+        structures : List[StructureType]
+            List of structures present in simulation, the first one being the simulation domain.
+        machine_error_relaxation : bool
+            When operations such as translation are applied to the 1d grids, fix the bounds
+            were numerically within the simulation bounds but were still chopped off.
+        bound_coords : Coord1D
+            1D grids potentially unaligned with the simulation boundary
+
+        Returns
+        -------
+        :class:`.Coords1D`:
+            1D coords to be used as grid boundaries.
+
+        """
+        center, size = simulation_box.center[axis], simulation_box.size[axis]
+        # chop off any coords outside of simulation bounds, beyond some buffer region
+        # to take numerical effects into account
+        bound_min = np.nextafter(center - size / 2, -inf, dtype=np.float32)
+        bound_max = np.nextafter(center + size / 2, inf, dtype=np.float32)
+
+        if bound_max < bound_coords[0] or bound_min > bound_coords[-1]:
+            axis_name = "xyz"[axis]
+            raise SetupError(
+                f"Simulation domain does not overlap with the provided grid in '{axis_name}' direction."
+            )
+
+        if size == 0:
+            # in case of zero-size dimension return the boundaries between which simulation falls
+            ind = np.searchsorted(bound_coords, center, side="right")
+
+            # in case when the center coincides with the right most boundary
+            if ind >= len(bound_coords):
+                ind = len(bound_coords) - 1
+
+            return bound_coords[ind - 1 : ind + 1]
+
+        else:
+            bound_coords = bound_coords[bound_coords <= bound_max]
+            bound_coords = bound_coords[bound_coords >= bound_min]
+
+            # if not extending to simulation bounds, repeat beginning and end
+            dl_min = bound_coords[1] - bound_coords[0]
+            dl_max = bound_coords[-1] - bound_coords[-2]
+            while bound_coords[0] - dl_min >= bound_min:
+                bound_coords = np.insert(bound_coords, 0, bound_coords[0] - dl_min)
+            while bound_coords[-1] + dl_max <= bound_max:
+                bound_coords = np.append(bound_coords, bound_coords[-1] + dl_max)
+
+            # in case operations are applied to coords, it's possible the bounds were numerically within
+            # the simulation bounds but were still chopped off, which is fixed here
+            if machine_error_relaxation:
+                if np.isclose(bound_coords[0] - dl_min, bound_min):
+                    bound_coords = np.insert(bound_coords, 0, bound_coords[0] - dl_min)
+                if np.isclose(bound_coords[-1] + dl_max, bound_max):
+                    bound_coords = np.append(bound_coords, bound_coords[-1] + dl_max)
+
+            return bound_coords
+
 
 class UniformGrid(GridSpec1d):
     """Uniform 1D grid. The most standard way to define a simulation is to use a constant grid size in each of the three directions.
@@ -176,8 +248,6 @@ class UniformGrid(GridSpec1d):
             Axis of this direction.
         structures : List[StructureType]
             List of structures present in simulation, the first one being the simulation domain.
-        **kwargs:
-            Other arguments all go here.
 
         Returns
         -------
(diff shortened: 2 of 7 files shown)


weiliangjin2021:
One test still errored. So I think this implementation is not robust. I'll try the custom grid idea.

weiliangjin2021:
Please check out the new implementation. `CustomGridBoundaries` is risky as we don't validate if the coordinate covers exactly the entire simulation domain. So for now I plan not to expose it to external users: it will not be added to doc API, and not added to tidy3d namespace. It can only be used in the `imported` classmethod.

tidy3d/components/grid/grid_spec.py
```diff
@@ -657,6 +698,14 @@ def make_grid(
         coords = Coords(**coords_dict)
         return Grid(boundaries=coords)
 
+    @classmethod
+    def imported(cls, grid: Grid) -> GridSpec:
```
tylerflex: maybe `from_grid` is more in line with our naming?

tidy3d/components/grid/grid_spec.py
```diff
@@ -199,6 +199,47 @@ def _make_coords_initial(
...
+    exactly the simulation domain size along the dimension, including PML layers, symmetric regions.
+
+    Example
+    -------
+    >>> grid_1d = CustomGridCoords(boundaries=[-0.2, 0.0, 0.2, 0.4, 0.5, 0.6, 0.7])
+    """
+
+    coords: Coords1D = pd.Field(
+        ...,
+        title="Grid boundary coordinates",
```
tylerflex: Title case

tidy3d/components/grid/grid_spec.py
```diff
@@ -199,6 +199,47 @@ def _make_coords_initial(
...
+
+    def _make_coords_initial(
+        self,
+        axis: Axis,
+        structures: List[StructureType],
+        **kwargs,
+    ) -> Coords1D:
+        """Unused. Override ``make_coords`` directly as the
+        grid coordiate for the complete simulation domain is provided already.
+        """
```
tylerflex, line 306: remove this? what's it for?

tylerflex: just initializing a subclass?

tidy3d/components/grid/grid_spec.py
```diff
@@ -657,6 +698,14 @@ def make_grid(
         coords = Coords(**coords_dict)
         return Grid(boundaries=coords)
 
+    @classmethod
+    def imported(cls, grid: Grid) -> GridSpec:
+        """Import from another simulation's grids."""
```
tylerflex: might be useful to include the syntax to get the grid, ie `Simulation.grid`

tylerflex (changes requested):
This will make things so much easier! Thanks @weiliangjin2021 

Dont forget changelog

tidy3d/components/grid/grid_spec.py
weiliangjin2021: yeah, since it's an abstract method in the parent class.

weiliangjin2021:
> Please check out the new implementation. `CustomGridBoundaries` is risky as we don't validate if the coordinate covers exactly the entire simulation domain. So for now I plan not to expose it to external users: it will not be added to doc API, and not added to tidy3d namespace. It can only be used in the `imported` classmethod.

Maybe I can validate it in simulation's root validator: sim.simulation_bounds align with the grid coords?

weiliangjin2021:
@dbochkov-flexcompute now that I added the validator to validate the simulation domain boundaries match the grid boundaries, `mode_solver.reduced_simulation_copy` will fail the validation because it removes PML layers. Should I add it back?

dbochkov-flexcompute:
> @dbochkov-flexcompute now that I added the validator to validate the simulation domain boundaries match the grid boundaries, `mode_solver.reduced_simulation_copy` will fail the validation because it removes PML layers. Should I add it back?

hmm, not sure if that would resolve the issue. I think we should make `CustomGridBoundaries` just to match only original sim bounds, without pml, so that pml layers are constructed as usual. That would also avoid an issue when subsection is taken in region of non-uniform grid and that wouldn't have uniform spacing near boundaries like pml. Btw, are coordinates outside of target region discarded? Can't find the place where it is done.

tidy3d/components/simulation.py
```diff
@@ -4331,7 +4361,7 @@ def subsection(
             raise SetupError("Requested region does not intersect simulation domain")
 
         # restrict to the original simulation domain
-        new_bounds = Box.bounds_intersection(self.bounds, region.bounds)
+        new_bounds = Box.bounds_intersection(self.simulation_bounds, region.bounds)
```
dbochkov-flexcompute: let's do:
```
if include_pml_cells:
    new_bounds = Box.bounds_intersection(self.simulation_bounds, region.bounds)
else:
    new_bounds = Box.bounds_intersection(self.bounds, region.bounds)
```
where `include_pml_cells: bool = False` is a new parameter. Then in mode solver we just call it with `include_pml_cells=True`

weiliangjin2021:
> hmm, not sure if that would resolve the issue. I think we should make `CustomGridBoundaries` just to match only original sim bounds, without pml, so that pml layers are constructed as usual.

Well, that can be achieved with `CustomGrid`. In my test, the grid in original sim bounds with `CustomGrid` matches exactly. However, once we start to construct pml layers, there is machine precision issue (the grid differ by around 1e-15). That is reason for this `CustomGridBoundaries` to include the complete grids without any further constructions.

> Btw, are coordinates outside of target region discarded? Can't find the place where it is done.

No postprocessing is currently performed. The coordinate must span exactly the simulation domain size.

tidy3d/components/simulation.py
weiliangjin2021: What are the use cases that pml cells need to discarded? I think it is not necessary for mode solver?

weiliangjin2021:
> > hmm, not sure if that would resolve the issue. I think we should make `CustomGridBoundaries` just to match only original sim bounds, without pml, so that pml layers are constructed as usual.
> 
> Well, that can be achieved with `CustomGrid`. In my test, the grid in original sim bounds with `CustomGrid` matches exactly. However, once we start to construct pml layers, there is machine precision issue (the grid differ by around 1e-15). That is reason for this `CustomGridBoundaries` to include the complete grids without any further constructions.

You are right. There is something wrong in my test setup. I'll refactor it to optionally construct pml layers.

tidy3d/components/simulation.py
dbochkov-flexcompute: when I was initially designing this function the idea was that it should always return a subset of existing simulation. That is, in the limiting case of `new_sim = sim.subsection(region=td.Box(size=[inf, inf, inf]))` we get essentially the same simulation. A more practical example: when you have a large waveguide circuit and want to focus on one particular component, i.e., you can use something like `new_sim = sim.subsection(region=td.Box(size=(Lx, Ly, inf), center(xc, yc, 0))` without worrying that size in the z direction is accidentally increased

weiliangjin2021:
> > @dbochkov-flexcompute now that I added the validator to validate the simulation domain boundaries match the grid boundaries, `mode_solver.reduced_simulation_copy` will fail the validation because it removes PML layers. Should I add it back?
> 
> hmm, not sure if that would resolve the issue. I think we should make `CustomGridBoundaries` just to match only original sim bounds, without pml, so that pml layers are constructed as usual. That would also avoid an issue when subsection is taken in region of non-uniform grid and that wouldn't have uniform spacing near boundaries like pml. Btw, are coordinates outside of target region discarded? Can't find the place where it is done.

Now `CustomGridBoundaries` works almost identical to `CustomGrid`, except that it takes grid boundaries rather than grid size.

tidy3d/components/grid/grid_spec.py
```diff
@@ -136,6 +136,78 @@ def _add_pml_to_bounds(num_layers: Tuple[int, int], bounds: Coords1D) -> Coords1
         add_right = bounds[-1] + last_step * np.arange(1, num_layers[1] + 1)
         return np.concatenate((add_left, bounds, add_right))
 
+    @staticmethod
+    def _postprocess_unaligned_grid(
+        axis: Axis,
+        structures: List[StructureType],
```
dbochkov-flexcompute: it looks like only `structures[0].geometry` is needed in this function, is that correct? I think it would make sense to replace `structures: List[StructureType]` with `simulation_box: Box`

tidy3d/components/grid/grid_spec.py
```diff
@@ -199,6 +271,47 @@ def _make_coords_initial(
...
+        *kwargs
+            Other arguments all go here.
+
+        Returns
+        -------
+        :class:`.Coords1D`:
+            1D coords to be used as grid boundaries.
+        """
+
+        return self._postprocess_unaligned_grid(axis, structures, False, self.coords)
```
dbochkov-flexcompute: To increase readability `return self._postprocess_unaligned_grid(axis=axis, structures=structures, machine_error_relaxation=False, bound_coords=self.coords)`

tidy3d/components/grid/grid_spec.py
```diff
@@ -263,49 +376,9 @@ def _make_coords_initial(
...
-            # the simulation bounds but were still chopped off, which is fixed here
-            if self.custom_offset is not None:
-                if np.isclose(bound_coords[0] - dl_min, bound_min):
-                    bound_coords = np.insert(bound_coords, 0, bound_coords[0] - dl_min)
-                if np.isclose(bound_coords[-1] + dl_max, bound_max):
-                    bound_coords = np.append(bound_coords, bound_coords[-1] + dl_max)
-
-            return bound_coords
+        return self._postprocess_unaligned_grid(
+            axis, structures, self.custom_offset is not None, bound_coords
```
dbochkov-flexcompute: To increase readability `return self._postprocess_unaligned_grid(axis=axis, structures=structures, machine_error_relaxation=self.custom_offset is not None, bound_coords=bound_coords)`

tidy3d/components/simulation.py
```diff
@@ -4322,6 +4323,8 @@ def subsection(
             Remove structures outside of the new simulation domain.
         remove_outside_custom_mediums : bool = True
             Remove custom medium data outside of the new simulation domain.
+        include_pml_cells : bool = False
+            Keep PML cells in simulation boundaries.
```
dbochkov-flexcompute: we can add something like "Note that retained PML cells will be converted to regular cells, and the simulation domain boundary will be moved accordingly."

dbochkov-flexcompute (approved):
Looks great! Thanks for fixing this and making it more robust. Just a few minor comments below

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Mode solver plugin now supports 'EMESimulation'.
 - `TriangleMesh` class: automatic removal of zero-area faces, and functions `fill_holes` and `fix_winding` to attempt mesh repair.
+- Support for grid specifications from grid cell boundary coordinates via `CustomGridBoundaries` that subclasses from `GridSpec1d`.
+- More convenient mesh importing from another simulation through `grid_spec = GridSpec.from_grid(sim.grid)``.
```
tylerflex: fix backslash quote (extra one at end)

tidy3d/components/grid/grid_spec.py
```diff
@@ -199,6 +271,52 @@ def _make_coords_initial(
...
+
+    Example
+    -------
+    >>> grid_1d = CustomGridCoords(boundaries=[-0.2, 0.0, 0.2, 0.4, 0.5, 0.6, 0.7])
+    """
+
+    coords: Coords1D = pd.Field(
+        ...,
+        title="Grid Boundary Coordinates",
+        description="An array of grid boundariy coordinates.",
```
tylerflex: boundary

tidy3d/components/grid/grid_spec.py
```diff
@@ -199,6 +271,52 @@ def _make_coords_initial(
...
+    ) -> Coords1D:
+        """Customized 1D coords to be used as grid boundaries.
+
+        Parameters
+        ----------
+        axis : Axis
+            Axis of this direction.
+        structures : List[StructureType]
+            List of structures present in simulation, the first one being the simulation domain.
+        *kwargs
```
tylerflex: these kwargs aren't used?

tylerflex (approved):
couple minor things.

tidy3d/components/grid/grid_spec.py
weiliangjin2021: It seems a historic thing (the same in UniformGrid). I think we kept it to make function signature consistent across subclasses, otherwise previously pylint will complain.

tylerflex: if so maybe we should remove from the docstring then

weiliangjin2021: OK, I'll also remove it from UniformGrid

PR #1838, opened 2024-07-17
Add ModeSimulation classes

### LINKED ISSUE #1414: Improve mode solver? maybe rewrite it (plugins.mode_solver2?)
Feel like our mode solver might need a bit of a makeover. Perhaps we could try a new plugin to replace the existing one? but still support existing for backwards compatibility?

caseyflex:
This PR introduces `ModeSimulation` classes as a new interface for the mode solver. The `ModeSimulation` classes more closely parallel `Simulation` and `EMESimulation` (and inherit a lot of the functionality from `AbstractYeeGridSimulation`). This makes a lot of user-facing operations more convenient.

The `ModeSimulation` can be 2D or 3D; if it is 3D, the mode plane must be specified as well. There are functions to convert to and from other `AbstractYeeGridSimulation` types. The existing `ModeSolver` plugin still works, but is just a wrapper around the `ModeSimulation` classes.

Todo:

- [x] Move `solve` and data operations to `ModeSimulation` or `ModeSimulationData` classes.
- [ ] improve test coverage a bit
- [ ] Check consistency in plotting functions
- [ ] Check that memory usage and performance is mostly unchanged


caseyflex:
@tylerflex (and others): Currently, `ModeSimulation` does not have sources or monitors. It just treats the entire `ModeSimulation` as the plane for mode solving, and saves the full `ModeSolverData` in `ModeSimulationData.modes`.

I will add `interval_space` and `fields` parameters to the `ModeSimulation` class to allow for downsampling here.

tidy3d/components/mode/data/sim_data.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Returns
+        -------
+        ModeSolverData
+            :class:`.ModeSolverData` object containing the effective index and mode fields.
+        """
+        mode_solver_data = self.data_raw
+        return mode_solver_data.symmetry_expanded_copy
+
+    def plot_modes(
```
tylerflex, line 61: should we also provide a `SimulationData.plot_fields()` wrapper for this? in case people want to use the same api call?

tidy3d/components/mode/data/sim_data.py
```diff
@@ -0,0 +1,119 @@
...
+        ..., title="Mode simulation", description="Mode simulation associated with this data."
+    )
+
+    data_raw: ModeSolverData = pd.Field(
+        ...,
+        title="Raw Data",
+        description="Raw mode data containing the field and effective index " "on unexpanded grid.",
+    )
+
+    data: Tuple[Literal[None], ...] = pd.Field(
```
tylerflex: is this needed?

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,337 @@
...
+
+    boundary_spec: BoundarySpec = pd.Field(
+        BoundarySpec.all_sides(PECBoundary()),
+        title="Boundaries",
+        description="Specification of boundary conditions along each dimension. If ``None``, "
+        "PML boundary conditions are applied on all sides. NOTE: for Mode simulations, this "
+        "is required to be PECBoundary on all sides. The 'ModeSpec' can be used to apply PML.",
+    )
+
+    monitors: Tuple[None, ...] = pd.Field(
```
tylerflex, line 97: should we consider instead removing Monitors and sources from the abstract case and then adding them back into the Simulation and EMESimulation? (same with `.data`)

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,337 @@
...
+        (),
+        title="Monitors",
+        description="Tuple of monitors in the simulation. "
+        "Note: monitor names are used to access data after simulation is run. "
+        "NOTE: monitors are not supported in a 'ModeSimulation'.",
+    )
+
+    sources: Tuple[None, ...] = pd.Field(
```
tylerflex, line 104: could also just have the type be `Literal[()]`? (same for others)

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,337 @@
...
+        title="Grid Specification",
+        description="Specifications for the simulation grid along each of the three directions.",
+        validate_default=True,
+    )
+
+    @pd.validator("size", always=True)
+    def is_plane(cls, val):
+        """Raise validation error if not planar."""
+        if val.count(0.0) != 1:
+            raise ValidationError(f"ModeSimulation must be planar, given size={val}")
```
tylerflex: I think we already have this validator, perhaps in `validators.py`

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,337 @@
...
+            center=self.center,
+            size=self.size,
+            medium=self.medium,
+            structures=self.structures,
+            symmetry=self.symmetry,
+            grid_spec=self.grid_spec,
+            boundary_spec=self.boundary_spec,
+            version=self.version,
+            subpixel=self.subpixel,
+            run_time=RUN_TIME,
```
tylerflex, line 242: consider exposing this (and some other things)  in `**kwargs` so the user can customize it.

For example, what about the source time `fwidth`, and the `direction`?

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,337 @@
...
+        """Convert :class:`.ModeSimulation` to :class:`.Simulation`.
+        This should only be used to obtain the same material properties
+        for mode solving or related purposes; the sources and monitors of the
+        resulting simulation are not meaningful."""
+
+        # source to silence warnings
+        plane = self.geometry
+        freq0 = self.freqs[0]
+        source_time = GaussianPulse(freq0=freq0, fwidth=0.1 * freq0)
+        source = ModeSource(
```
tylerflex: should we reuse `to_source`?

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,337 @@
...
+            medium=self.medium,
+            structures=self.structures,
+            symmetry=self.symmetry,
+            grid_spec=self.grid_spec,
+            boundary_spec=self.boundary_spec,
+            version=self.version,
+            subpixel=self.subpixel,
+            run_time=RUN_TIME,
+            sources=[source],
+            monitors=[],
```
tylerflex, line 244: could we add some way to switch whether this makes a monitor or a source?

tidy3d/plugins/mode/mode_solver.py
tylerflex, line 1: my initial thinking was that a lot of the mode solver math bits would be moved into components and then the plugin version would just wrap `ModeSimulation`. 

At the moment it seems for example that the `ModeSimulation` doesn't support a local solve option?

tylerflex (commented):
Thanks @caseyflex , sorry it took so long to get to this. Looks overall good.

One question is whether we want the ModeSimulation to support local solves (using the mode solver internals moved into components). and then the plugin would be a light wrapper on this?

How it's set up now also works, but just want to throw that out there.

tidy3d/components/mode/data/sim_data.py
tylerflex: sorry, revisiting this after reading more, I think we could consider just `data = Literal[()]`?

tidy3d/components/mode/simulation.py
caseyflex: yeah, I think that makes sense

caseyflex: they have to be individually defined for each simulation anyway, since each type of simulation accepts different types of sources and monitors

caseyflex: well, but they all do have a lot in common, in terms of the geometric properties and basic api for monitors, so maybe we should leave it for the time being

tidy3d/components/mode/data/sim_data.py
caseyflex: yeah. the only difference is that since there aren't any monitors, you don't pass a monitor name

tidy3d/components/mode/simulation.py
caseyflex: this function is private, and not meant to be used by the user. It is only used internally in the ModeSolver plugin to convert `ModeSimulation` to `Simulation`, as a workaround until `ModeSimulation` is supported natively by the web api. The user shouldn't turn a mode simulation into an FDTD simulation, as this would rarely make any sense, since a mode simulation is 2D transverse to the propagation direction. It's just used to get material properties on the grid

tidy3d/components/mode/simulation.py
caseyflex: good call

tidy3d/components/mode/simulation.py
caseyflex: Same comment as above. This function isn't meant for the user, it's just for internal use for generating material properties

caseyflex:
> Thanks @caseyflex , sorry it took so long to get to this. Looks overall good.
> 
> One question is whether we want the ModeSimulation to support local solves (using the mode solver internals moved into components). and then the plugin would be a light wrapper on this?
> 
> How it's set up now also works, but just want to throw that out there.

Yeah, I think that's a good question, whether we want ModeSimulation to support local solves. Currently, you would have to create a ModeSolver and set `simulation` to the `ModeSimulation` -- you no longer need an FDTD simulation to do mode solves, but you would still have to use the ModeSolver plugin.

We certainly could move all the mode solver internals to the ModeSimulation directory / class, and make the plugin just a thin wrapper around that. I have one major concern, relating to the autogrid. If you make a ModeSimulation with autogrid, it will generate its own grid which can be used for mode solving just fine. However, another use case is an FDTD simulation, and you want to mode solve in just a planar region of the simulation. In this case, you want to use the grid coming from the entire FDTD simulation, and not just allow the mode solver region to generate its own grid. In spite of that, we could move the core mode solver code from `plugins` to `components`, inside the `mode` directory, and make it closely integrated with `ModeSimulation`, which could then have an easier local solve api; however, we would still want the mode solver internals to operate on a simulation which could be a `Simulation`, `EMESimulation`, or `ModeSimulation`, and not restrict it to operating on a `ModeSimulation`. Although it is possible to restructure in this way, this might be an argument against. Another argument is that it's unfortunate to keep moving stuff to components (should the mode solver be a separate repo?). A final argument is that we greatly prefer the remote mode solver for accuracy reasons; but the local one is definitely a huge convenience.

tidy3d/components/mode/data/sim_data.py
caseyflex: for some reason this doesn't work:
```E           pydantic.v1.error_wrappers.ValidationError: 2 validation errors for ModeSimulation
E           sources
E             unexpected value; permitted:  (type=value_error.const; given=(); permitted=())
E           monitors
E             unexpected value; permitted:  (type=value_error.const; given=(); permitted=())

File: tidy3d/components/mode/simulation.py (hunk 3 above)
Comment (tylerflex) [reply tidy3d/components/mode/simulation.py:97]: I see, I'll leave it up to you, but we could consider making the base classes use `Literal[()]` for these fields perhaps? and then if the subclasses add sources / monitors, we can define them as `Tuple[SpecificTypes, ...]`? just a thought.

File: tidy3d/components/mode/simulation.py (hunk 6 above)
Comment (tylerflex) [reply tidy3d/components/mode/simulation.py:242]: I see. that makes sense I think. 

Would some kind of convenience method to eg. export the ModeSimulation to a source / monitor and add to existing simulation be useful? Or just unnecessary?

I'm just trying to get a sense for how a user might have these ModeSimulations interact with Simulations in most practical cases

```py
fdtd_sim = Simulation(...)

mode_sim = ModeSimulation.from_sim(fdtd_sim)

# pseudocode warning
fdtd_sim = mode_sim.to_simulation(type='monitor', sim= fdtd_sim)

```

tylerflex:
> Yeah, I think that's a good question, whether we want ModeSimulation to support local solves. Currently, you would have to create a ModeSolver and set `simulation` to the `ModeSimulation` -- you no longer need an FDTD simulation to do mode solves, but you would still have to use the ModeSolver plugin.  We certainly could move all the mode solver internals to the ModeSimulation directory / class, and make the plugin just a thin wrapper around that.

All else equal, being able to do local without needing the plugin sounds best to me.

>  I have one major concern, relating to the autogrid. If you make a ModeSimulation with autogrid, it will generate its own grid which can be used for mode solving just fine. However, another use case is an FDTD simulation, and you want to mode solve in just a planar region of the simulation. In this case, you want to use the grid coming from the entire FDTD simulation, and not just allow the mode solver region to generate its own grid. In spite of that, we could move the core mode solver code from `plugins` to `components`, inside the `mode` directory, and make it closely integrated with `ModeSimulation`, which could then have an easier local solve api; however, we would still want the mode solver internals to operate on a simulation which could be a `Simulation`, `EMESimulation`, or `ModeSimulation`, and not restrict it to operating on a `ModeSimulation`. Although it is possible to restructure in this way, this might be an argument against. 

Hm, maybe I dont fully understand. So you're saying that in some sense maybe it makes sense for the mode solver to be not its own Simulation class, but rather some kind of "run" function that gets applied to other simulation types? For example, have `run(sim: Simulation) -> SimulationData` and `run_mode_solver(sim: Simulation) -> ModeSolverData` ?

And this would I guess cause a pretty big overhaul of the code architecture?

Can some of this be alleviated by importing grids from Simulation or EMESimulation ie with https://github.com/flexcompute/tidy3d/pull/1836, or by providing to_ and from_ methods to allow importing and exporting ModeSimulation from other simulation types (while preserving grids?)

> Another argument is that it's unfortunate to keep moving stuff to components (should the mode solver be a separate repo?). 

I think as long as the mathematical bits are sufficiently abstracted away from the pydantic / api, it should be ok. but yea we've sort of been moving in this integrated components approach for some time now (eg with HeatSimulation) so I feel it's maybe a more general concern? I think we should reduce coupling as much as we can, but at the same time, sometimes having separate plugins creates its own issues all the same.

> A final argument is that we greatly prefer the remote mode solver for accuracy reasons; but the local one is definitely a huge convenience.

I use the local one quite a bit, since a lot of times my remote mode solves (even when small) get queued and it can be a bit annoying to wait for them, especially when I've already run them previously.

momchil-flex:
I haven't digested everything here but real quick comments:

> However, another use case is an FDTD simulation, and you want to mode solve in just a planar region of the simulation. In this case, you want to use the grid coming from the entire FDTD simulation, and not just allow the mode solver region to generate its own grid.

The current `ModeSolver` has a `.reduced_simulation_copy` method which makes a copy of the mode solver that only has a span as big as its `plane`, and throws away structures that are not interesected by that plane, and even throws away custom medium data that is outside of what's needed for the interpolation needed in the solver. There's some tricky details about getting the discretization right to match the simulation grid exactly - in fact @weiliangjin2021 has a [PR](https://github.com/flexcompute/tidy3d/pull/1836/files) that's ready to be merged that improves exactly on that aspect. The original motivation for this was exactly the case where a large amount of custom medium data needs to be uploaded for a mode solver that only requires a small section of it.

The use case of defining a mode solver within an FDTD simulation (on the same grid as the FDTD simulation) is definitely a very common one, including internally for the functioning of our mode sources and monitors. I think the obvious way to handle this in the `ModeSimulation` is to have a similar method to the `.reduced_simulation_copy` (probably reusing a lot of the code), but since `ModeSimulation` no longer stores an FDTD `simulation` inside of it, it would rather be a classmethod, something like `ModeSimulation.from_simulation(simulation, **mode_kwargs)`. This would then produce a `ModeSimulation` instance with the grid exactly matching the one of the input `simulation`. I mean, I'm saying the obvious way, but there will probably be some pitfalls to be wary of.

Regarding local run, yeah, it's definitely important to still support... Maybe we should limit the scope for now though and first work on the API and the web run, while still allowing a local run for debug purposes through a conversion to `ModeSolver`? Then separately work on bringing the plugin solver code into the core tidy3d components. As you refer.

caseyflex:
Thanks for the feedback @tylerflex @momchil-flex . I'm still thinking about whether I should move the core mode solver code into components for this PR or wait until later. Like you said, this would be helpful but it might be better to limit the scope of this PR. However, if it would make the code a lot nicer / avoid code duplication, it may be preferable to do it here.

I looked into the PRs on `subsection` and `reduced_simulation_copy`. I think it makes sense to have something like that, `ModeSimulation.from_simulation`. I see that @weiliangjin2021 's PR matching the grid exactly is very helpful for this. Thinking about these details, I realized another technical concern. I currently made the `ModeSimulation` a 2D simulation, because that's intuitively how the mode solver works. But I see that `reduced_simulation_copy` actually preserves an extra cell on either side of the mode solver plane, so that subpixel can work. The comments say it is for subpixel for custom medium, but I'm not sure I understand exactly what is going on there. I suppose for `ModeSimulation` we could either have a 2D simulation, or a 3D simulation with a specified mode solving plane? This would allow the most flexibility for both use cases.

tidy3d/components/mode/simulation.py
caseyflex: This still doesn't seem to work:

```
E           pydantic.v1.error_wrappers.ValidationError: 2 validation errors for ModeSimulation
E           sources
E             unexpected value; permitted:  (type=value_error.const; given=(); permitted=())
E           monitors
E             unexpected value; permitted:  (type=value_error.const; given=(); permitted=())
```

PR #1840, opened 2024-07-18
Update `pre/2.8` to fix the sync docs action

daquinteroflex:
So the sync docs issue appeared to Marc in one of his pre/2.8 PRs per https://github.com/flexcompute/tidy3d/actions/runs/9990645191

So this is not urgent, but mainly of interest if we want  to see the live 2.8 branch docs at some point and also want that sync action error not to be raised in this branch. It was already fixed in `develop` so just involves propagating along to `pre/2.8`.

I guess most of the conflicts should be resolved in another commit from another PR into develop?


momchil-flex:
Sorry is this basically a merge develop into pre/2.8 PR?

daquinteroflex:
Yep, opened this because of the merge conflicts to have the reference and to reference a conflict resolution PR that fixes this if this is what we want to do?

momchil-flex:
@tylerflex @yaugenst-flex we want to merge develop into pre/2.8 but there are a bunch of conflicts coming from autograd. Could one of you help resolve? It can be on this PR or you can start fresh.

yaugenst-flex:
I think I'll have to do this in another PR since the branch here is just `develop`.

momchil-flex:
Ah, right.

PR #1850, opened 2024-07-22
Updated validator for 2D far field projection to include FieldProjectionCartesianMonitor

QimingFlex:
Thanks to comments from @yaugenst-flex , the verbosity of the code has been greatly reduced. It's now ready for a final review by @momchil-flex .

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -2882,11 +2882,11 @@ class Simulation(AbstractYeeGridSimulation):
     def _projection_monitors_2d(cls, val, values):
         """
         Validate if the field projection monitor is set up for a 2D simulation and
-        ensure the observation angle is configured correctly.
+        ensure the observation parameters are configured correctly.
 
         - For a 2D simulation in the x-y plane, 'theta' should be set to 'pi/2'.
-        - For a 2D simulation in the y-z plane, 'phi' should be set to 'pi/2'.
-        - For a 2D simulation in the x-z plane, 'phi' should be set to '0'.
+        - For a 2D simulation in the y-z plane, 'phi' should be set to 'pi/2' or '2*pi/3'.
+        - For a 2D simulation in the x-z plane, 'phi' should be set to '0' or 'pi'.
 
         Note: Exact far field projection is not available yet. Currently, only
         'far_field_approx = True' is supported.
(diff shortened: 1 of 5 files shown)


tylerflex (approved):
Thanks @QimingFlex

tidy3d/components/simulation.py
```diff
@@ -2716,20 +2716,67 @@ def _projection_monitors_2d(cls, val, values):
...
+                                f"monitor '{monitor.name}' should be set to '0' or '1'."
+                            )
+                        elif monitor.proj_axis in (0, 1) and (
+                            len(monitor.y) != 1 or monitor.y[0] != 0
+                        ):
+                            raise SetupError(
+                                "For a 2D simulation in the x-y plane with "
+                                f"'proj_axis = {monitor.proj_axis}', 'y' of monitor "
+                                f"'{monitor.name}' should be set to 0."
+                            )
```
yaugenst-flex: This seems really verbose, maybe something like this would work too?

```python
config = {
    "y-z": {"index": 0, "coord": "x"},
    "x-z": {"index": 1, "coord": "y"},
    "x-y": {"index": 2, "coord": "z"},
}[plane]

invalid_proj_axis = config["index"]
valid_proj_axes = [i for i in range(3) if i != invalid_proj_axis]

if monitor.proj_axis == invalid_proj_axis:
    raise SetupError(
        f"For a 2D simulation in the {plane} plane, the 'proj_axis' of "
        f"monitor '{monitor.name}' should be set to one of {valid_proj_axes}."
    )

coord = getattr(monitor, config["coord"])
if monitor.proj_axis in valid_proj_axes and (len(coord) != 1 or coord[0] != 0):
    raise SetupError(
        f"For a 2D simulation in the {plane} plane with "
        f"'proj_axis = {monitor.proj_axis}', '{config['coord']}' of monitor "
        f"'{monitor.name}' should be set to 0."
    )
```

yaugenst-flex (approved):
Thanks! Only a minor comment regarding the validation.

tidy3d/components/simulation.py
QimingFlex: Thanks @yaugenst-flex ! This is very helpful, I will revise it.

PR #1858, opened 2024-07-29
Update to ruff@0.5.5

yaugenst-flex:
Updating to latest ruff. Linter catching some additional errors, which I fixed.


tylerflex (approved):
Looks good. I guess people might need to make sure to update their local ruff versions for compatibility?

yaugenst-flex:
I'm assuming people upgrade their poetry venv on branch checkouts :smile: but otherwise yeah, you'll have to update. The formatting is unaffected, but there are some slight linting differences.

I'm trying to do these updates a little more frequently so that we don't end up with a huge chunk of changes at once like last time.

PR #1859, opened 2024-07-29
Validate the use of dielectrics only in RectangularDielectric

lucas-flexcompute:
This will give a better error message if the user tries to set metallic materials when using the `RectangularDielectric` class.


tomflexcompute (approved):
Thanks @lucas-flexcompute !

tylerflex (approved):

PR #1860, opened 2024-07-29
Introduced axial ratio to DirectivityMonitor

QimingFlex:
Introduced field axial_ratio in the DirectivityMonitor.

@dmarek-flex suggested renaming the DirectivityMonitor class to RadiationMonitor as radiation is more general and can include many characteristics such as directivity, gain, axial ratio, etc. I think this is reasonable and better. What do you think, @weiliangjin2021? Do you have other suggestions?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Support for differentiation with respect to `ComplexPolySlab.vertices`.
-- Introduce RF material library. Users can now export `rf_material_library` from `tidy3d.plugins.microwave`.
+- Introduce RF material library. Users can now import `rf_material_library` from `tidy3d.plugins.microwave`.
 - Users can specify the background medium for a structure in automatic differentiation by supplying `Structure.autograd_background_permittivity`.
 - `DirectivityMonitor` to compute antenna directivity.
 - Added `plot_length_units` to `Simulation` and `Scene` to allow for specifying units, which improves axis labels and scaling when plotting.
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `WavePort` to the `TerminalComponentModeler` which is the suggested port type for exciting transmission lines.
 - Added plotting functionality to voltage and current path integrals in the `microwave` plugin.
 - Added convenience functions `from_terminal_positions` and `from_circular_path` to simplify setup of `VoltageIntegralAxisAligned` and `CustomCurrentIntegral2D`, respectively.
+- Added `axial_ratio` to `DirectivityData.axial_ratio` for the `DirectivityMonitor`, defined as the ratio of the major axis to the minor axis of the polarization ellipse. 
 
 ### Changed
 - Priority is given to `snapping_points` in `GridSpec` when close to structure boundaries, which reduces the chance of them being skipped.
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -58,6 +58,7 @@ from ..types import (
 )
 from ..validators import enforce_monitor_fields_present, required_if_symmetry_present
 from .data_array import (
+    AxialRatioDataArray,
     DataArray,
     DiffractionDataArray,
     DirectivityDataArray,
@@ -1989,16 +1990,17 @@ class DirectivityData(MonitorData):
 
     Example
     -------
-    >>> from tidy3d import DirectivityDataArray
+    >>> from tidy3d import DirectivityDataArray, AxialRatioDataArray
     >>> f = np.linspace(1e14, 2e14, 10)
-    >>> r = np.atleast_1d(5)
+    >>> r = np.atleast_1d(1e6)
     >>> theta = np.linspace(0, np.pi, 10)
     >>> phi = np.linspace(0, 2*np.pi, 20)
     >>> coords = dict(r=r, theta=theta, phi=phi, f=f)
     >>> values = np.random.random((len(r), len(theta), len(phi), len(f)))
-    >>> scalar_field = DirectivityDataArray(values, coords=coords)
+    >>> scalar_directivity_field = DirectivityDataArray(values, coords=coords)
+    >>> scalar_axial_ratio_field = AxialRatioDataArray(values, coords=coords)
     >>> monitor = DirectivityMonitor(center=(1,2,3), size=(2,2,2), freqs=f, name='n2f_monitor', phi=phi, theta=theta)
-    >>> data = DirectivityData(monitor=monitor,directivity=scalar_field)
+    >>> data = DirectivityData(monitor=monitor, directivity=scalar_directivity_field, axial_ratio=scalar_axial_ratio_field)
     """
 
     monitor: DirectivityMonitor = pd.Field(
@@ -2009,6 +2011,10 @@ class DirectivityData(MonitorData):
 
     directivity: DirectivityDataArray = pd.Field(
         ..., title="Directivity", description="Directivity with an angle-based projection grid."
+    )
+
+    axial_ratio: AxialRatioDataArray = pd.Field(
+        ..., title="Axial Ratio", description="Axial ratio with an angle-based projection grid."
     )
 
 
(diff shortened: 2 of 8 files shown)


weiliangjin2021:
> Introduced field axial_ratio in the DirectivityMonitor.
> 
> @dmarek-flex suggested renaming the DirectivityMonitor class to RadiationMonitor as radiation is more general and can include many characteristics such as directivity, gain, axial ratio, etc. I think this is reasonable and better. What do you think, @weiliangjin2021? Do you have other suggestions?

I don't think we can rename it now, as it's released already in v2.7.0. But you can add a new monitor called RadiationMonitor, and add a warning in DirectivityMonitor that it'll be deprecated in v3.0.

QimingFlex:
> > Introduced field axial_ratio in the DirectivityMonitor.
> > @dmarek-flex suggested renaming the DirectivityMonitor class to RadiationMonitor as radiation is more general and can include many characteristics such as directivity, gain, axial ratio, etc. I think this is reasonable and better. What do you think, @weiliangjin2021? Do you have other suggestions?
> 
> I don't think we can rename it now, as it's released already in v2.7.0. But you can add a new monitor called RadiationMonitor, and add a warning in DirectivityMonitor that it'll be deprecated in v3.0.

Yeah, sounds good.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Support for differentiation with respect to `ComplexPolySlab.vertices`.
 - Introduce RF material library. Users can now export `rf_material_library` from `tidy3d.plugins.microwave`.
+- Users can now export `SimulationData` to MATLAB `.mat` files with the `to_mat_file` method.
+- Introduce RF material library. Users can now import `rf_material_library` from `tidy3d.plugins.microwave`.
```
dmarek-flex: Duplicate RF material bullet points and possible mistake adding the MATLAB one?

tidy3d/components/data/monitor_data.py
```diff
@@ -2011,6 +2013,10 @@ class DirectivityData(MonitorData):
         ..., title="Directivity", description="Directivity with an angle-based projection grid."
     )
 
+    axial_ratio: AxialRatioDataArray = pd.Field(
+        ..., title="Axial ratio", description="Axial ratio with an angle-based projection grid."
+    )
```
dmarek-flex: I believe our convention is to capitalize each word in the `title`

tidy3d/components/data/monitor_data.py
```diff
@@ -1989,16 +1990,17 @@ class DirectivityData(MonitorData):
 
     Example
     -------
-    >>> from tidy3d import DirectivityDataArray
+    >>> from tidy3d import DirectivityDataArray, AxialRatioDataArray
     >>> f = np.linspace(1e14, 2e14, 10)
     >>> r = np.atleast_1d(5)
```
dmarek-flex: This is pedantic, but the radiation characteristics should be independent of r, since radiation intensity is scaled by r^2. Including r is not required, maybe if we rename later we can remove r, unless I am missing something.

dmarek-flex (approved):
Thanks @QimingFlex, it looks pretty clear to me! I just found a couple of things to nitpick.

CHANGELOG.md
QimingFlex: I changed "import" from "export" in line 14 and messed this up. I will revert these changes.

tidy3d/components/data/monitor_data.py
QimingFlex: Actually this projection distance r is used to compute the propagation factor to get the correct Etheta and Ephi.

weiliangjin2021 (approved):
All look nice!

weiliangjin2021:
Actually one comment: could you add the definition and the reference for the axial ratio? As you mentioned in the solver, the definition is not unique, so it will be good to clarify which definition is used.

tidy3d/components/data/monitor_data.py
dmarek-flex: That's right, but it ultimately isn't necessary for describing the resulting Directivity. Directivity only depends on frequency and angular position, the same as radiation intensity.

QimingFlex: That makes sense.

QimingFlex: I just found that if we give up introducing the coordinate "r" here, we will need to do a corresponding coordinate change in the backend, which will make the code ugly. As this class inherits from FieldProjectionAngleMonitor, which has the coordinate "r", it's better to keep it consistent I guess?  Users typically don't need to specify this value and can use the default value of 1e6.

weiliangjin2021: Looks fine to keep it if the default value works fine in most simulations.

dmarek-flex: Yea, totally fine to keep it. I just wanted to point out that we don't absolutely need it in these DataArrays.

QimingFlex:
> Actually one comment: could you add the definition and the reference for the axial ratio? As you mentioned in the solver, the definition is not unique, so it will be good to clarify which definition is used.

Yeah, I updated the reference to Balanis's equations (2-65) to (2-67). The backend computations are equivalent to these equations. Balanis uses absolute values and phase differences between two components for the computations. Instead, we compute using complex numbers directly. This has also been mentioned in the backend computations.

weiliangjin2021:
> > Actually one comment: could you add the definition and the reference for the axial ratio? As you mentioned in the solver, the definition is not unique, so it will be good to clarify which definition is used.
> 
> Yeah, I updated the reference to Balanis's equations (2-65) to (2-67). The backend computations are equivalent to these equations. Balanis uses absolute values and phase differences between two components for the computations. Instead, we compute using complex numbers directly. This has also been mentioned in the backend computations.

Great! One minor format: the doc can read better if formatted in [this way](https://github.com/flexcompute/tidy3d/blob/b5a9cd1d40023ad6ca7429425052fade5c90e64b/tidy3d/components/subpixel_spec.py#L29-L34).

Also I wonder if the reference should also be cited in the `DirectivityMonitor` docstring, as users might not even read the docstring of `AxialRatioDataArray`

QimingFlex:
> > > Actually one comment: could you add the definition and the reference for the axial ratio? As you mentioned in the solver, the definition is not unique, so it will be good to clarify which definition is used.
> > 
> > 
> > Yeah, I updated the reference to Balanis's equations (2-65) to (2-67). The backend computations are equivalent to these equations. Balanis uses absolute values and phase differences between two components for the computations. Instead, we compute using complex numbers directly. This has also been mentioned in the backend computations.
> 
> Great! One minor format: the doc can read better if formatted in [this way](https://github.com/flexcompute/tidy3d/blob/b5a9cd1d40023ad6ca7429425052fade5c90e64b/tidy3d/components/subpixel_spec.py#L29-L34).
> 
> Also I wonder if the reference should also be cited in the `DirectivityMonitor` docstring, as users might not even read the docstring of `AxialRatioDataArray`

Yeah, both have been updated, also updated drectivity reference in the `DirectivityMonitor`.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `WavePort` to the `TerminalComponentModeler` which is the suggested port type for exciting transmission lines.
 - Added plotting functionality to voltage and current path integrals in the `microwave` plugin.
 - Added convenience functions `from_terminal_positions` and `from_circular_path` to simplify setup of `VoltageIntegralAxisAligned` and `CustomCurrentIntegral2D`, respectively.
+- Added axial ratio computation to `DirectivityMonitor`
```
tylerflex: please mention how to grab this information, for example `DirectivityMonitor.axial_ratio`. Also what is the axial ratio? consider adding `, defined as the ratio of the major axis to the minor axis of the polarization ellipse`.

Adding this info into the changelog reduces the barrier of entry for people who just catch sight of the new feature to start using it.

otherwise PR looks good!

tylerflex (approved):
just one changelog comment.

CHANGELOG.md
tylerflex: also, maybe the more technically correct thing here is to say that it is added to the "`DirectivityData.axial_ratio` corresponding to a `DirectivityMonitor`".?

QimingFlex: Rivised, thanks @tylerflex !

PR #1866, opened 2024-07-31
Fix bug when running the same mode solver both locally and remote

momchil-flex:


weiliangjin2021 (approved):
Confirmed that it's working!

PR #1868, opened 2024-07-31
add units to `IndexPerturbation.frequency`

tylerflex:
https://flow360.atlassian.net/browse/SCEM-7107


daquinteroflex (approved):
Looks good!

Curious as to whether they will run into more issues with any more fields throughout the API classes if they don't have units too.

PR #1871, opened 2024-07-31
:hammer: Update the submodule checks

daquinteroflex:
Hi Momchil,

I think this implements the test-submodule implementation we chatted about. 

Let me know any thoughts.

- [x] Performs all checks on PRs onto latest
- [x] Does a test submodule check before the release process occurs.
- [x] Perform the submodule checks on the `pre/*` and `develop` branches once a day.

I have tested the daily and latest PR workflow run as expected in my fork:
- https://github.com/daquinteroflex/tidy3d/actions/runs/10205603421/job/28236822661
- https://github.com/daquinteroflex/tidy3d/actions/runs/10205209567

I can do full release demoing in my fork too to thoroughy test the flow but logically think it should work as is.


momchil-flex (approved):
Yes! I like this. :) Hopefully the last time we revisit...

Code looks good overall but only way to know if it works as expected is through testing. I know you did some already so let's get this in and then monitor for inconsistencies.

daquinteroflex:
Sounds good! Also to update the daily tests run as expected https://github.com/daquinteroflex/tidy3d/actions/runs/10226877623

daquinteroflex:
Just thought to check if you also get the notifications now that it's running:

image

momchil-flex:
Unfortunately no. I was looking into this for another action yesterday and it seems like it's not easy to control who gets notification. Only the user who created the action does. :(

daquinteroflex:
We could use this solution possibly: https://stackoverflow.com/questions/62304258/github-actions-notifications-on-workflow-failure what do you think?

PR #1873, opened 2024-08-01
Move some EME validators to pre_upload

caseyflex:


momchil-flex (approved):
Looks good, only missing a changelog item

PR #1875, opened 2024-08-01
Move dispersion fitter implementation to components

caseyflex:
This PR is exploring an intermediate approach between two old PRs, neither of which was merged:
https://github.com/flexcompute/tidy3d/pull/1773
https://github.com/flexcompute/tidy3d/pull/1809

The first PR moves the fitter to components. The second PR separates it out to a new package.

In this PR, the core fitting algorithm is moved to components, but the fitter plugin with the quality-of-life features stays in plugins.


tidy3d/components/dispersion_fitter.py
```diff
@@ -0,0 +1,936 @@
...
+OMEGA_POLE_CLOSE_ATOL = 1e-10
+
+# Range for checking upper bound of Im[eps], in addition to extrema method.
+# The range is in unit of eV and it's in log scale.
+LOSS_CHECK_MIN = -10
+LOSS_CHECK_MAX = 4
+LOSS_CHECK_NUM = 1000
+
+
+def imag_ep_extrema(poles: ArrayComplex1D, residues: ArrayComplex1D) -> ArrayFloat1D:
```
tylerflex: what is `ep`? might want to consider giving this function a more intuitive name.

tylerflex (commented):
Looks good to me. I think having it like this (inside of components, but somewhat separated) makes sense for the time being.

tylerflex (approved):

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -698,147 +91,26 @@ def fit(
...
-                "Dispersion fitter cannot have 'max_num_poles' less than 'min_num_poles'."
-            )
-
-        omega = PoleResidue.angular_freq_to_eV(PoleResidue.Hz_to_angular_freq(self.freqs[::-1]))
-        eps = self.eps_data[::-1]
-
-        init_model = FastFitterData.initialize(
-            omega, eps, eps_inf, advanced_param or AdvancedFastFitterParam()
+        omega_data = PoleResidue.Hz_to_angular_freq(self.freqs[::-1])
+        eps_data = self.eps_data[::-1]
```
weiliangjin2021: Unrelated question to this PR: does fitter only work with descending frequency list?

weiliangjin2021 (approved):
Looks nice!

tidy3d/plugins/dispersion/fit_fast.py
caseyflex: Good call, it should work regardless of the order. Let me remove this `[::-1]`. I think this was an artifact from early testing and debugging of the fitter.

caseyflex:
@tylerflex revised. can we merge this?

tylerflex:
yes, but maybe this is best for 2.8?  @momchil-flex

momchil-flex:
Yeah let's do 2.8 please.

momchil-flex:
Also note that it shouldn't be moving the notebooks submodule state unless you explicitly intend to.

PR #1882, opened 2024-08-06
Restricting allowed ModeSolver.plane types

momchil-flex:


tidy3d/plugins/mode/mode_solver.py
```diff
@@ -74,6 +75,7 @@
 
 MODE_SIMULATION_TYPE = Union[Simulation, EMESimulation]
 MODE_SIMULATION_DATA_TYPE = Union[SimulationData, EMESimulationData]
+MODE_PLANE_TYPE = Union[Box, ModeSource, ModeMonitor, ModeSolverMonitor]
```
tylerflex, line 78: why not allow other source and monitor types? or even Mode Simulation types?

tylerflex (approved):
looks good just wondering if we want to allow more box subtypes

tidy3d/plugins/mode/mode_solver.py
momchil-flex: So my thinking is that if I could just do what I want, I would just require this to always be `Box` so the definition is clear and there's no wide range of types that this attribute can take. So I kind of took the route of allowing the minimum extra that should preserve compatibility at least within our examples. Technically you are right though that some people may have used this in other ways too, e.g. passing other monitors. Probably quite an edge case but not impossible. So.. not sure.

PR #1884, opened 2024-08-06
Return modes info table even if fields not available

dbochkov-flexcompute:


momchil-flex (approved):
That's funny, looking at the code it seems like that was what was intended even? Otherwise no reason to have that separate check to populate the field-requiring variables only if fields are present. Not just populate, re-compute!

tylerflex (approved):

PR #1885, opened 2024-08-06
Turning off incidence matrix

weiliangjin2021:
Temporary solution to find correct modes in the presence of metals and conformal subpixel methods. Seems to be robust in my tests. In the future, we can explore other preconditioner methods.

- When incidence matrix is off and conformal PEC is applied, correct modes are only found if we project permeability to the conductive model.


weiliangjin2021:
Some tests in backend test_pec.py are failing with this change. Still need to look into it.

PR #1886, opened 2024-08-06
Add LossyMetalMedium and subpixel options for lossy_metal including SIBC

weiliangjin2021:
## SIBC Applicable Medium

Theoretically we can use class composition to apply SIBC to all material types. However, in practice, it's almost only applied to DC-conductivity type material and `eps_inf` is ignored. To that regard, I subclass `LossyMetalMedium` from `Medium`, and  free parameters are conductivity, frequency_range, and fitting parameters.

Fitting is performed in its frequency_range, so frequency_range is a required field here. In fact, unlike other material types, frequency_range is a very important parameter here: the fitter results are sensitive to frequency range.

```
mat = td.LossyMetalMedium(conductivity=1e2, frequency_range=(2e14, 3e14))
mat.plot()
```
And the result:
image

## Subpixel Spec

`SubpixelSpec` now has four fields: dielectric, metal, PEC, and lossy_metal. When lossy_metal takes SIBC, solver will apply SIBC to structures made of `td.LossyMetalMedium`. We don't validate if the metal is lossy enough, as it heavily depends on the geometries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added new support functions for the Design plugin: automated batching of `Simulation` objects, and summary functions with `DesignSpace.estimate_cost` and `DesignSpace.summarize`.
 - Added validation and repair methods for `TriangleMesh` with inward-facing normals.
 - Added `from_admittance_coeffs` to `PoleResidue`, which allows for directly constructing a `PoleResidue` medium from an admittance function in the Laplace domain.
+- Added material type `LossyMetalMedium` that has high DC-conductivity. Its boundaries can be modeled by a surface impedance boundary condition (SIBC). This is treated as a subpixel method, which can be switched by setting `SubpixelSpec(lossy_metal=method)` where `method` can be `Staircasing()`, `VolumetricAveraging()`, or `SurfaceImpedance()`.
 
 ### Changed
 - Priority is given to `snapping_points` in `GridSpec` when close to structure boundaries, which reduces the chance of them being skipped.
--- a/tidy3d/components/grid/mesher.py
+++ b/tidy3d/components/grid/mesher.py
@@ -17,7 +17,7 @@ from ...constants import C_0, fp_eps
 from ...exceptions import SetupError, ValidationError
 from ...log import log
 from ..base import Tidy3dBaseModel
-from ..medium import AnisotropicMedium, Medium2D, PECMedium
+from ..medium import AnisotropicMedium, LossyMetalMedium, Medium2D, PECMedium
 from ..structure import MeshOverrideStructure, Structure, StructureType
 from ..types import ArrayFloat1D, Axis, Bound, Coordinate
 
@@ -522,7 +522,7 @@ class GradedMesher(Mesher):
         min_steps = []
         for structure in structures:
             if isinstance(structure, Structure):
-                if isinstance(structure.medium, (PECMedium, Medium2D)) or (
+                if isinstance(structure.medium, (PECMedium, Medium2D, LossyMetalMedium)) or (
                     isinstance(structure.medium, AnisotropicMedium)
                     and structure.medium.is_comp_pec(axis)
                 ):
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -54,6 +54,7 @@ from .dispersion_fitter import (
     LOSS_CHECK_MAX,
     LOSS_CHECK_MIN,
     LOSS_CHECK_NUM,
+    fit,
     imag_resp_extrema_locs,
 )
 from .geometry.base import Geometry
@@ -76,6 +77,7 @@ from .types import (
     Complex,
     FreqBound,
     InterpMethod,
+    Literal,
     PoleAndResidue,
     TensorReal,
 )
@@ -91,6 +93,12 @@ FILL_VALUE = "extrapolate"
 # cap on number of nonlinear iterations
 NONLINEAR_MAX_NUM_ITERS = 100
 NONLINEAR_DEFAULT_NUM_ITERS = 5
+
+# Lossy metal
+LOSSY_METAL_DEFAULT_SAMPLING_FREQUENCY = 20
+LOSSY_METAL_SCALED_REAL_PART = 10.0
+LOSSY_METAL_DEFAULT_MAX_POLES = 3
+LOSSY_METAL_DEFAULT_TOLERANCE_RMS = 1e-3
 
 
 def ensure_freq_in_range(eps_model: Callable[[float], complex]) -> Callable[[float], complex]:
@@ -1070,6 +1078,50 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
             Complex-valued relative permittivity.
         """
         return eps_real * (1 + 1j * loss_tangent)
+
+    @staticmethod
+    def eV_to_angular_freq(f_eV: float):
+        """Convert frequency in unit of eV to rad/s.
+
+        Parameters
+        ----------
+        f_eV : float
+            Frequency in unit of eV
+        """
+        return f_eV / HBAR
+
+    @staticmethod
+    def angular_freq_to_eV(f_rad: float):
+        """Convert frequency in unit of rad/s to eV.
+
+        Parameters
+        ----------
+        f_rad : float
+            Frequency in unit of rad/s
+        """
+        return f_rad * HBAR
+
+    @staticmethod
+    def angular_freq_to_Hz(f_rad: float):
+        """Convert frequency in unit of rad/s to Hz.
+
+        Parameters
+        ----------
+        f_rad : float
+            Frequency in unit of rad/s
+        """
+        return f_rad / 2 / np.pi
+
+    @staticmethod
+    def Hz_to_angular_freq(f_hz: float):
+        """Convert frequency in unit of Hz to rad/s.
+
+        Parameters
+        ----------
+        f_hz : float
+            Frequency in unit of Hz
+        """
+        return f_hz * 2 * np.pi
 
     @ensure_freq_in_range
     def sigma_model(self, freq: float) -> complex:
@@ -3052,50 +3104,6 @@ class PoleResidue(DispersiveMedium):
         return PoleResidue(eps_inf=eps_inf, poles=list(zip(a_coeffs, c_coeffs)))
 
     @staticmethod
-    def eV_to_angular_freq(f_eV: float):
-        """Convert frequency in unit of eV to rad/s.
-
-        Parameters
-        ----------
-        f_eV : float
-            Frequency in unit of eV
-        """
-        return f_eV / HBAR
-
-    @staticmethod
-    def angular_freq_to_eV(f_rad: float):
-        """Convert frequency in unit of rad/s to eV.
-
-        Parameters
-        ----------
-        f_rad : float
-            Frequency in unit of rad/s
-        """
-        return f_rad * HBAR
-
-    @staticmethod
-    def angular_freq_to_Hz(f_rad: float):
-        """Convert frequency in unit of rad/s to Hz.
-
-        Parameters
-        ----------
-        f_rad : float
-            Frequency in unit of rad/s
-        """
-        return f_rad / 2 / np.pi
-
-    @staticmethod
-    def Hz_to_angular_freq(f_hz: float):
-        """Convert frequency in unit of Hz to rad/s.
-
-        Parameters
-        ----------
-        f_hz : float
-            Frequency in unit of Hz
-        """
-        return f_hz * 2 * np.pi
-
-    @staticmethod
     def imag_ep_extrema(poles: Tuple[PoleAndResidue, ...]) -> ArrayFloat1D:
         """Extrema of Im[eps] in the same unit as poles.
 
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added plotting functionality to voltage and current path integrals in the `microwave` plugin.
 - Added convenience functions `from_terminal_positions` and `from_circular_path` to simplify setup of `VoltageIntegralAxisAligned` and `CustomCurrentIntegral2D`, respectively.
 - Added `axial_ratio` to `DirectivityData.axial_ratio` for the `DirectivityMonitor`, defined as the ratio of the major axis to the minor axis of the polarization ellipse. 
+- Added material type `LossyMetalMedium` that has high DC-conductivity. Its boundaries can be modeled by surface impedance boundary condition (SIBC). This is treated as a subpixel method, and can be switched by setting `SubpixelSpec(lossy_metal=method)` where `method` can be `Staircasing()`, `VolumetricAveraging()`, or `SurfaceImpedance()`.
```
tylerflex: by **a** surface boundary condition

some slight rearranging of the next sentence

This is treated as a subpixel method, which can be switched by setting `SubpixelSpec(lossy_metal=method)`, where `method` can be `Staircasing()`, `VolumetricAveraging()`, or `SurfaceImpedance()`.

tidy3d/components/medium.py
```diff
@@ -5974,6 +5982,164 @@ def perturbed_copy(
...
 
+class SkinDepthFitterParam(Tidy3dBaseModel):
+    """Advanced parameters for fitting complex-valued skin depth ``2j/k`` over its frequency
+    bandwidth, where k is the complex-valued wavenumber inside the lossy metal. Real part
+    of this quantity corresponds to physical skin depth.
+    """
+
+    max_num_poles: pd.PositiveInt = pd.Field(
+        LOSSY_METAL_DEFAULT_MAX_POLES,
+        title="Maximal number of poles",
```
tylerflex: title case for these

tidy3d/components/medium.py
```diff
@@ -5974,6 +5982,164 @@ def perturbed_copy(
...
+    frequency_sampling_points: pd.PositiveInt = pd.Field(
+        LOSSY_METAL_DEFAULT_SAMPLING_FREQUENCY,
+        title="Number of sampling frequencies",
+        description="Number of sampling frequencies used in fitting.",
+    )
+
+    log_sampling: bool = pd.Field(
+        True,
+        title="Frequencies sampled in log scale",
+        description="Whether to sample frequencies logarithmically (True),  "
```
tylerflex: please put true and false in double backslash

tidy3d/components/medium.py
```diff
@@ -5974,6 +5982,164 @@ def perturbed_copy(
...
+class SkinDepthFitterParam(Tidy3dBaseModel):
+    """Advanced parameters for fitting complex-valued skin depth ``2j/k`` over its frequency
+    bandwidth, where k is the complex-valued wavenumber inside the lossy metal. Real part
+    of this quantity corresponds to physical skin depth.
+    """
+
+    max_num_poles: pd.PositiveInt = pd.Field(
+        LOSSY_METAL_DEFAULT_MAX_POLES,
+        title="Maximal number of poles",
+        description="Maximal number of poles in complex-conjugate poles residue model for "
```
tylerflex: "pole residue model" (singular)

tidy3d/components/medium.py
```diff
@@ -5974,6 +5982,164 @@ def perturbed_copy(
...
+    log_sampling: bool = pd.Field(
+        True,
+        title="Frequencies sampled in log scale",
+        description="Whether to sample frequencies logarithmically (True),  "
+        "or linearly (False).",
+    )
+
+
+class LossyMetalMedium(Medium):
+    """Material with high DC-conductivity that can be modeled with surface
```
tylerflex: with a surface impedance

tidy3d/components/medium.py
```diff
@@ -5974,6 +5982,164 @@ def perturbed_copy(
...
+        title="Frequency Range",
+        description="Frequency range of validity for the medium.",
+        units=(HERTZ, HERTZ),
+    )
+
+    fit_param: SkinDepthFitterParam = pd.Field(
+        SkinDepthFitterParam(),
+        title="Complex-valued skin depth fitting parameters",
+        description="Parameters for fitting complex-valued dispersive skin depth over "
+        "the frequency range by using pole-residue pair model.",
```
tylerflex: could you explain?

` by using pole-residue pair model`

maybe just remove "by"?

tidy3d/components/medium.py
```diff
@@ -5974,6 +5982,164 @@ def perturbed_copy(
...
+        residues /= scaling_factor
+
+        return PoleResidue(eps_inf=res_inf, poles=list(zip(poles, residues)))
+
+    @cached_property
+    def num_poles(self) -> int:
+        """Number of poles in the fitted model."""
+        return len(self.skin_depth_model.poles)
+
+    def complex_skin_depth(self, frequencies):
```
tylerflex: please include type annotations for public methods, like this one `sampling_frequencies` and others (if present)

tidy3d/components/subpixel_spec.py
```diff
@@ -114,6 +117,25 @@ def courant_ratio(self) -> float:
...
+    """Apply 1st order (Leontovich) surface impedance boundary condition to
+    structure made of :class:`.LossyMetalMedium`.
+    """
+
+    timestep_reduction: float = pd.Field(
+        DEFAULT_COURANT_REDUCTION_SIBC_CONFORMAL,
+        title="Time Step Size Reduction Rate",
+        description="Reduction factor between 0 and 1 such that the simulation's time step size "
+        "will be ``1 - timestep_reduction`` times its default value. "
+        "Accuracy can be improved with a smaller time step size, but simulation time increased as well.",
```
tylerflex: but the simulation time will be increased

tidy3d/components/subpixel_spec.py
```diff
@@ -114,6 +117,25 @@ def courant_ratio(self) -> float:
...
+class SurfaceImpedance(PECConformal):
+    """Apply 1st order (Leontovich) surface impedance boundary condition to
+    structure made of :class:`.LossyMetalMedium`.
+    """
+
+    timestep_reduction: float = pd.Field(
+        DEFAULT_COURANT_REDUCTION_SIBC_CONFORMAL,
+        title="Time Step Size Reduction Rate",
+        description="Reduction factor between 0 and 1 such that the simulation's time step size "
+        "will be ``1 - timestep_reduction`` times its default value. "
```
tylerflex: will be -> is

tidy3d/components/subpixel_spec.py
```diff
@@ -140,16 +162,32 @@ class SubpixelSpec(Tidy3dBaseModel):
         discriminator=TYPE_TAG_STR,
     )
 
+    lossy_metal: LossyMetalSubpixelType = pd.Field(
+        SurfaceImpedance(),
+        title="Subpixel averaging method on lossy metal interfaces",
```
tylerflex: Subpixel Averaging Method for Lossy Metal Interfaces

tylerflex (approved):
Implementation looks good, just some comments on the documentation bits, thanks @weiliangjin2021 !

weiliangjin2021:
Thanks Tyler. I'll also add the doc in the revision. `LossyMetalMedium` can be added to the 1st session in https://github.com/flexcompute/tidy3d/blob/develop/docs/api/mediums.rst, but not sure where to add `SkinDepthFitterParam`?

tylerflex:
> but not sure where to add SkinDepthFitterParam?

Yea not sure, maybe a fit parameters section in mediums? how was this handled when the fitter was recently added into core components?

weiliangjin2021:
> > but not sure where to add SkinDepthFitterParam?
> 
> Yea not sure, maybe a fit parameters section in mediums? how was this handled when the fitter was recently added into core components?

A new section sounds good. The API doc doesn't change even though the fitter is migrated to core: https://github.com/flexcompute/tidy3d/blob/pre/2.8/docs/api/plugins/dispersion.rst

tidy3d/components/medium.py
```diff
@@ -5974,6 +5982,172 @@ def perturbed_copy(
...
+    >>> lossy_metal = LossyMetalMedium(conductivity=10, frequency_range=(9e9, 10e9))
+
+    """
+
+    permittivity: Literal[1] = pd.Field(
+        1.0, title="Permittivity", description="Relative permittivity.", units=PERMITTIVITY
+    )
+
+    frequency_range: FreqBound = pd.Field(
+        ...,
```
caseyflex: I wonder if we could come up with a sensible default for frequency_range based on the sources in the simulation. For example, I do something similar for nonlinear materials -- freq0 is a parameter for the nonlinear model, but the default can often be automatically determined from the sources in the simulation. You could probably do something similar with fwidth.

https://github.com/flexcompute/tidy3d/blob/61966a923dc537b9576cfe888b9de5208d9005be/tidy3d/components/medium.py#L162

caseyflex (approved):
Looks good!

And yeah, only the core fitter functionality was moved to core, the user interface was kept in plugins so the API docs didn't need to change.

tidy3d/components/medium.py
weiliangjin2021: That's a good idea. So we get `frequency_range` from source bandwidth if it's `None`, or use it directly if it's not `None`. One concern that I have is that one might be interested in visualizing the fitting result. If taken from the source, that seems only possible until simulation object has been initialized, or one has to manually enter the frequency range in the `plot` method. So not sure which approach is the best.

tidy3d/components/grid/mesher.py
```diff
@@ -522,9 +522,13 @@ def structure_steps(
         min_steps = []
         for structure in structures:
             if isinstance(structure, Structure):
-                if isinstance(structure.medium, (PECMedium, Medium2D)) or (
-                    isinstance(structure.medium, AnisotropicMedium)
-                    and structure.medium.is_comp_pec(axis)
+                if (
+                    isinstance(structure.medium, (PECMedium, Medium2D))
+                    or isinstance(structure.medium, LossyMetalMedium)
```
momchil-flex: this could just be one line `isinstance(structure.medium, (PECMedium, Medium2D, LossyMetalMedium))`

tidy3d/components/medium.py
```diff
@@ -5974,6 +5982,172 @@ def perturbed_copy(
...
+    log_sampling: bool = pd.Field(
+        True,
+        title="Frequencies Sampling In Log Scale",
+        description="Whether to sample frequencies logarithmically (``True``),  "
+        "or linearly (``False``).",
+    )
+
+
+class LossyMetalMedium(Medium):
+    """Lossy metal modeled with a surface impedance boundary condition (SIBC).
```
momchil-flex: Maybe "Lossy metal that can be modeled with..." because it can also be turned off right?

tidy3d/components/medium.py
```diff
@@ -5974,6 +5982,172 @@ def perturbed_copy(
...
+
+    Notes
+    -----
+
+        SIBC is suitable when the skin depth is much smaller than the structure feature size.
```
momchil-flex: Maybe instead of "suitable" -> "most accurate"?

Should we say something like if that is not the case, use a regular medium instead?

momchil-flex (approved):
Overall, this is great, just a couple of minor comments on the code.

There's one big thing we still need to think about though, brought up yesterday by @lucas-flexcompute . It is partly related to this PR but also a broader concern. Namely, in our multiphysics planning, we are imagining that we will have a single "scene" definition with media that have all their optical, electronic, heat, properties etc. that can be reused between various simulations. This would also include doing both an RF simulation and an optical frequency simulation for the same device, e.g. for modulators. The optical properties of the materials will of course be vastly different in the two frequency ranges and we probably need a way to define those separately. Lucas briefly mentioned maybe having a whole separate RF specification of the medium the way we have `heat_spec` and will have `charge_spec`, say `rf_spec`. This technically maybe suggests that we shouldn't have the optical definitions be fundamental either and we should move them to an `optical_spec` or something like that? We should discuss how to handle this well with the future goals in mind. Specifically with respect to this PR, it is probably likely that the LossyMetalMedium will be used for RF materials only. If I define a simulation using such a medium and I want to run both an RF and an optical simulation, things will get messy, and will probably even break because the optical simulation will be completely outside the frequency range. Have you already given all this some thought? Should we schedule a small meeting to brainstorm?

tidy3d/components/medium.py
weiliangjin2021: right

tidy3d/components/medium.py
weiliangjin2021: or set supbixel.lossy_metal to volumetric averaging

weiliangjin2021: It's now edited as

> SIBC is most accurate when the skin depth is much smaller than the structure feature size. If not the case, please use a regular medium instead, or set ``simulation.subpixel.lossy_metal`` to ``td.VolumetricAveraging()`` or ``td.Staircasing()``.

weiliangjin2021:
> Specifically with respect to this PR, it is probably likely that the LossyMetalMedium will be used for RF materials only.

It can be useful in infrared simulations where metal has Im[n] ~ 10^1. But it does seem rarely used. @tomflexcompute have you seen people use SIBC in photonics simulations?

If it's useful, we'll need to support SIBC for other dispersion models because metal has complicated dispersion in infrared region. In that case, it's better to add a field in AbstractMedium classes to indicate whether the material is considered as lossy metal or not. 

> If I define a simulation using such a medium and I want to run both an RF and an optical simulation, things will get messy, and will probably even break because the optical simulation will be completely outside the frequency range. Have you already given all this some thought?

Ah good point. It's not considered yet. One approach is that SIBC is only applied if the medium frequency range overlaps with the source frequency range; switch to volumetric averaging if there is no overlapping. 

> Should we schedule a small meeting to brainstorm?

Sure

tomflexcompute:
> It can be useful in infrared simulations where metal has Im[n] ~ 10^1. But it does seem rarely used. @tomflexcompute have you seen people use SIBC in photonics simulations?
> 

Probably unlikely for visible to mid-IR. In THz sometimes people use impedance boundaries to model metal for metamaterials for example.

tidy3d/version.py
```diff
@@ -1,3 +1,3 @@
 """DO NOT EDIT: Modified automatically with .bump2version.cfg"""
 
-__version__ = "2.8.0rc1"
+__version__ = "2.9.0rc2"
```
momchil-flex: What was this change about? Just so you can map to your solver? :)

That said, we may seriously consider postponing this for 2.9 if we don't settle on a good scalable way to implement this that works for RF + optical simulations.

weiliangjin2021: Yeah, this is for RF folks to run simulations with SIBC. The benchmark is ongoing now.

weiliangjin2021: I'd still aim for 2.8; otherwise there can be lots of conflicts to resolve

PR #1889, opened 2024-08-07
clarifying the documentation for ApodizationSpec

FilipeFcp:
Hi @momchil-flex. I updated the documentation of the ApodizationSpec and added an image.

I could not attach the HTML file here, but it looks like this:

image


momchil-flex:
Thanks, looks good but it seems you need to run `ruff` on your code and make a new commit (or rather squash on this one and force push).

tomflexcompute (approved):
Thanks @FilipeFcp . Looks very good. Just need to pass the tests.

FilipeFcp:
I have reviewed all the development guides, and everything should be correct now. I have also squashed the commits. I believe @momchil-flex is the one who should merge it. Is that correct?

tylerflex:
In general, you can assign 1 or more reviewers who you think will be best suited to handle it. For example, for something like this related to docs, one of Tom, Emerson, or Dario would be good choices.

daquinteroflex:
Hi @FilipeFcp, looks great, thanks!

Maybe we need to update the development guide on making sure the `pre-commit install` is manually run as that will `ruff` automatically your code on every commit, and also to specify that now that the team has grown maybe it makes sense to ask for review people working on the domain rather than just Momchil to not overwhelm him. There's this diagram I made about this back in May we can update with new people https://www.notion.so/flexcompute/Who-to-Ask-in-Tidy3D-May-24-e6739ac25aea493eb4628680ea3ef632

momchil-flex:
Yeah the pre-commit is very nice, we should add it to the development guide if not already there @daquinteroflex 

@FilipeFcp as others point out for a small change like this just tagging a single relevant reviewer is fine. They can also merge this. For large changes having someone senior like me, tyler, or weiliang review and merge is better. In this particular case I requested this so just having me as a reviewer would be the way to go.

daquinteroflex (approved):

momchil-flex (approved):

FilipeFcp:
Thank you @momchil-flex, @daquinteroflex, and @tylerflex for reviewing and taking the time to clarify everything.

PR #1891, opened 2024-08-07
autograd gradient fixes and miscellaneous improvements

tylerflex:


tests/test_components/test_autograd.py
```diff
@@ -58,16 +58,17 @@
...
 FREQ0 = td.C_0 / WVL
 FREQS = [FREQ0]
+FWIDTH = FREQ0 / 10
 
 # sim sizes
 LZ = 7 * WVL
 
 # NOTE: regular stuff is broken in 2D need to change volume and face integration to handle this
-IS_3D = True
+IS_3D = False
```
yaugenst-flex: Is the comment still relevant if 2d is default now?

tests/test_components/test_autograd.py
```diff
@@ -76,19 +77,29 @@
...
-        td.PointDipole(
+        # td.PointDipole(
+        #     center=(0, 0, -LZ / 2 + WVL),
+        #     polarization="Ey",
+        #     source_time=td.GaussianPulse(
+        #         freq0=FREQ0,
+        #         fwidth=FWIDTH,
+        #         amplitude=1.0,
+        #     ),
+        # ),
```
yaugenst-flex: Should this be removed or why is it commented out?

tests/test_components/test_autograd.py
```diff
@@ -238,14 +250,19 @@ def make_structures(params: anp.ndarray) -> dict[str, td.Structure]:
...
+    med = td.Medium(permittivity=3.0)
 
     # Structure with variable .medium
     eps = 1 + anp.abs(vector @ params)
-    conductivity = eps / 10.0
+    sigma = 0.1 * (anp.tanh(vector @ params) + 1)
+
+    permittivity, conductivity = eps, sigma
+    # permittivity, conductivity = eps, 0.0 # only permittivity
+    # permittivity, conductivity = 2.0, sigma # only conductivity
```
yaugenst-flex: Should probably make this a parametrized fixture at some point...

tests/test_components/test_autograd.py
```diff
@@ -401,34 +421,37 @@ def mode_postprocess_fn(sim_data, mnt_data):
...
         value = 0.0
         for _, val in mnt_data.field_components.items():
-            value += abs(anp.sum(val.values))
+            value = value + abs(anp.sum(val.values))
+        # field components numerical is 3x higher
         intensity = anp.nan_to_num(anp.sum(sim_data.get_intensity(mnt_data.monitor.name).values))
         value += intensity
+        # intensity numerical is 4.79x higher
         value += anp.sum(mnt_data.flux.values)
+        # flux is 18.4x lower
```
yaugenst-flex, line 449: Are these the factors for the grads?

tests/test_components/test_autograd.py
```diff
@@ -533,8 +563,103 @@ def postprocess(data: td.SimulationData) -> float:
...
+    assert np.all(_plane_coords == plane_coords)
+
+    vertices_next = np.roll(vertices, axis=0, shift=-1)
+    edges = vertices_next - vertices
+
+    basis_vecs = p.edge_basis_vectors(edges=edges)
+
+
+@pytest.mark.parametrize("structure_key, monitor_key", (("medium", "mode"),))
+def _test_autograd_numerical(structure_key, monitor_key):
```
yaugenst-flex: Is this test not supposed to be run?

tidy3d/components/grid/grid_spec.py
yaugenst-flex, line 1: I wonder if it's maybe a good idea to still at least warn the user that there were tracers present in the grid spec. In general this can be relevant for the gradient value, and if we just silently drop it...

yaugenst-flex (approved):
Lots of fixes! Looks good, only some minor comments.

tests/test_components/test_autograd.py
tylerflex: no, stuff is fixed in 2D, and also I forgot to make 3D default so will change it back, thanks

tests/test_components/test_autograd.py
tylerflex: Plane wave worked better for many numerical tests, and the point dipole created some havoc in the diffraction monitor. so for now I just commented it out and replaced it with the PlaneWave, but In the future I might want to add it back. I'll create a point dipole variable at the module level and remove this comment.

tests/test_components/test_autograd.py
tylerflex: hm, this is just a remnant from my numerical testing. it's all resolved now. I think I'll just remove these two individual cases.

tests/test_components/test_autograd.py
tylerflex: yes.

tests/test_components/test_autograd.py
tylerflex: it needs to call the actual web API to run sims, so yea it's not supposed to be run in the automated tests. In the future I will remove the leading underscore for numerical testing.

Any suggestions on better ways to handle this?

yaugenst: Yeah we should probably introduce pytest marks (slow, web, etc…) to mark test sets and then selectively run them, with some sane default sets

yaugenst: could run the full test suite then only before release for example, or on a big merge

tidy3d/components/grid/grid_spec.py
tylerflex: Good Idea, I went ahead and added this. if any tracers are present in override structures, user will get this warning
```
        if any(s.strip_traced_fields() for s in self.override_structures):
            log.warning(
                "The override structures were detected as having a dependence on the objective "
                "function parameters. This is not supported by our automatic differentiation "
                "framework. The derivative will be un-traced through the override structures. "
                "To make this explicit and remove this warning, use 'y = autograd.tracer.getval(x)'"
                " to remove any derivative information from values being passed to create "
                "override structures. Alternatively, 'obj = obj.to_static()' will create a copy of "
                "an instance without any autograd tracers."
            )
```

and it will continue with the static version of the override structures.

I also modified the test to ensure this warning is raised and the code still runs afterwards

tests/test_components/test_autograd.py
tylerflex: That would be good. Although the numerical testing will still be extremely long if we test every combination of structure and monitor. So I'm not sure we want to do it in an automated way. Instead we could leave it for individual feature testing

yaugenst: not parametrization, just custom marks. they are basically labels for tests, and then you can selectively run specific labels (e.g. all the tests that are labelled as „slow“)
i agree with not running every possible combination

it‘s just an alternative to the „underscore method“ 😄

tylerflex: ah ok, so here's what I did
```
RUN_NUMERICAL = False
...
@pytest.mark.skipif(not RUN_NUMERICAL, reason="Numerical gradient tests runs through web API.")
...
def test_numerical
```

That way it will skip it by default and we can change this behavior by setting `RUN_NUMERICAL`.

I didnt see a way to skip a test by default and trigger it by command line argument. how does this sound?

yaugenst-flex: yeah this is fine for now. what i mean by marks is [this](https://docs.pytest.org/en/stable/example/markers.html), i.e.:
```python
@pytest.mark.slow  # or for example @pytest.mark.numerical
def test_XX(): ...
```
where `slow` or `numerical` are custom marks registered with pytest that we define (in`pytest.ini` or `pyproject.toml`, and then you can run `pytest -v -m slow`, which will run only the tests marked `slow`.
i think this is something to tackle in a larger test refactor though, so i'll just add it to the pile

PR #1894, opened 2024-08-08
Improve passivity enforcement near high-Q poles in FastDispersionFitter

### LINKED ISSUE #1878: Fitting creates active material
**Describe the bug**
When fitting 3 refractive index values, I get a material that has a negative k value, and diverges.

**To Reproduce**
To reproduce, see the attached example notebook: 
image

caseyflex:
The passivity enforcement including `imag_ep_extrema` and `omega_range` was not sufficient when the fit `PoleResidue` model contained high-Q poles. This PR adds extra samples around each pole in order to improve the passivity enforcement. Specifically, the samples are computed using an explicit formula for extrema around a single pole.

The fit for the example data in the issue remains poor -- this is to be expected, since the data exhibits dispersion but has no loss.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added value_and_grad function to the autograd plugin, importable via `from tidy3d.plugins.autograd import value_and_grad`. Supports differentiating functions with auxiliary data (`value_and_grad(f, has_aux=True)`).
 
+### Fixed
+- Improved passivity enforcement near high-Q poles in `FastDispersionFitter`. Failed passivity enforcement could lead to simulation divergences.
+
 ## [2.7.2] - 2024-08-07
 
 ### Added
--- a/tests/test_plugins/test_dispersion_fitter.py
+++ b/tests/test_plugins/test_dispersion_fitter.py
@@ -272,3 +272,16 @@ def test_dispersion_guess(random_data):
     medium, rms = fitter.fit(num_tries=2)
 
     medium_new, rms_new = fitter.fit(num_tries=1, guess=medium)
+
+
+def test_dispersion_loss_samples():
+    wvls = np.array([275e-3, 260e-3, 255e-3])
+    n_nAlGaN = np.array([2.72, 2.68, 2.53])
+
+    nAlGaN_fitter = FastDispersionFitter(wvl_um=wvls, n_data=n_nAlGaN)
+    nAlGaN_mat, _ = nAlGaN_fitter.fit()
+
+    freq_list = nAlGaN_mat.angular_freq_to_Hz(nAlGaN_mat._imag_ep_extrema_with_samples())
+    ep = nAlGaN_mat.eps_model(freq_list)
+    for e in ep:
+        assert e.imag >= 0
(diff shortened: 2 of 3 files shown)


alec-flexcompute (approved):
Looks good if it passes this test! Not sure how the fixes you made exactly work but it looks reasonable :)

caseyflex:
I’m still checking over some things. This may discard too many models if
the criterion is just based on high Q.

caseyflex:
OK, I'm happy with this now. I fixed the formula and am no longer rejecting models on the basis of being high-Q. This PR just adds one or two extra samples around each pole, except for marginally-stable poles, the handling of which is not changed.

It might be worth checking some of the material_library fits to see if anything changed (indicating that previously the fit wasn't actually passive)

alec-flexcompute:
ah, this is a good point. Nice work @caseyflex!

caseyflex:
> ah, this is a good point. Nice work @caseyflex!

thanks for finding the issue!

tylerflex (approved):

weiliangjin2021 (approved):
Nice work! When I tried Alec's medium, the original extrema finder is able to find the frequency that violates the passivity. Is the fix to sample more frequency points around this region?

Also I wonder if it's better to merge to pre/2.8 to avoid resolving lots of conflicts?

caseyflex:
> Nice work! When I tried Alec's medium, the original extrema finder is able to find the frequency that violates the passivity. Is the fix to sample more frequency points around this region?
> 
> Also I wonder if it's better to merge to pre/2.8 to avoid resolving lots of conflicts?

So originally, I tried using logic as in `imag_ep_extrema_with_samples`, using an additional frequency range based on the data frequency range. However, the fitter was able to come up with a model that was passive at these additional frequencies too, while still not being uniformly passive, even in the frequency range of interest. So I concluded that the sampling was not working well because of these high-Q poles. The new samples I added should be tailored to the poles themselves, so as to avoid this pitfall of uniform logspace sampling.

Regarding pre/2.8 or develop: technically this is a bug fix, so typically it would go in develop. I think either way I'll be the one resolving the conflicts, and since I did both the fitter separation PR and this one, I should be able to do it without much difficulty either way.

weiliangjin2021:
> So originally, I tried using logic as in `imag_ep_extrema_with_samples`, using an additional frequency range based on the data frequency range. However, the fitter was able to come up with a model that was passive at these additional frequencies too, while still not being uniformly passive, even in the frequency range of interest. So I concluded that the sampling was not working well because of these high-Q poles. The new samples I added should be tailored to the poles themselves, so as to avoid this pitfall of uniform logspace sampling.

What a smart fitter!

> Regarding pre/2.8 or develop: technically this is a bug fix, so typically it would go in develop. I think either way I'll be the one resolving the conflicts, and since I did both the fitter separation PR and this one, I should be able to do it without much difficulty either way.

Sounds good.

caseyflex:
I checked against most of the material_library. It didn't seem to change most of the fits at all; it changed the following fits in a very small / insignificant way:
```
AlAs/FernOnton1971
Si3N4/Luke2015PMLStable
Si3N4/Luke2015Sellmeier
```

so it's possible that the fitter was previously returning mediums that slightly violated passivity at some frequencies for these materials; but anyway it's ok, we aren't using the fitter result in the material library.

And this confirms that the fitter still works well with this change, so we should be OK to merge this.

PR #1895, opened 2024-08-08
Raise error early when port source is snapped into PML

weiliangjin2021:


QimingFlex:
Thanks, @weiliangjin2021 ! I tested it, and it looks great. I have one suggestion regarding the error message: could we provide more detailed instructions? For example:

`Port {port.name} normal is less than 2 cells to the boundary on the +{'xyz'[normal_axis]} side. Please either increase the mesh resolution near the port or move the port away from the boundary.`

weiliangjin2021:
> Thanks, @weiliangjin2021 ! I tested it, and it looks great. I have one suggestion regarding the error message: could we provide more detailed instructions? For example:
> 
> `Port {port.name} normal is less than 2 cells to the boundary on the +{'xyz'[normal_axis]} side. Please either increase the mesh resolution near the port or move the port away from the boundary.`

Good point!

QimingFlex (approved):

tylerflex (approved):

PR #1896, opened 2024-08-09
removed a redundant change log for 2d projection

QimingFlex:
There is one redundant change log regarding 2D projections by accident, I just deleted the first one.


tomflexcompute (approved):
Thanks @QimingFlex .

PR #1901, opened 2024-08-13
Fix CustomSourceTime with times completely outside envelope range

### LINKED ISSUE #1899: CustomSourceTime with no time values within the simulation time range
One simulation encountered the following error during the solver run

```
  File "/local_data/shared_cache/tidy3d_solver/release-24.2.1/lib/python3.11/site-packages/tidy3d/components/source.py", line 323, in amp_time
    envelope[~mask] = values.interp(t=time_shifted[~mask]).to_numpy()
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_data/shared_cache/tidy3d_solver/release-24.2.1/lib/python3.11/site-packages/xarray/core/dataarray.py", line 2322, in interp
    ds = self._to_temp_dataset().interp(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_data/shared_cache/tidy3d_solver/release-24.2.1/lib/python3.11/site-packages/xarray/core/dataset.py", line 4004, in interp
    obj, newidx = missing._localize(obj, {k: v})
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_data/shared_cache/tidy3d_solver/release-24.2.1/lib/python3.11/site-packages/xarray/core/missing.py", line 557, in _localize
    minval = np.nanmin(new_x_loaded)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_data/shared_cache/tidy3d_solver/release-24.2.1/lib/python3.11/site-packages/numpy/lib/nanfunctions.py", line 343, in nanmin
    res = np.fmin.reduce(a, axis=axis, out=out, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: zero-size array to reduction operation fmin which has no identity
"""
```

It seems like in some cases `~mask` can be empty? We should catch something like this during validation though.

caseyflex:
Fixes this issue:
https://github.com/flexcompute/tidy3d/issues/1899

The behavior of CustomSourceTime is to interpolate for times within the envelope definition time range, and use nearest selection for times outside the envelope definition time range. When evaluating at a list of times which is completely outside the envelope definition time range, xarray.interp fails, because we pass it an empty list. This PR checks separately for this case, when no interpolation is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `DataArray` interpolation failure due to incorrect ordering of coordinates when interpolating with autograd tracers.
+- Error in `CustomSourceTime` when evaluating at a list of times entirely outside of the range of the envelope definition times.
 
 ## [2.7.2] - 2024-08-07
 
--- a/tests/test_components/test_source.py
+++ b/tests/test_components/test_source.py
@@ -307,7 +307,19 @@ def test_custom_source_time(log_capture):
         atol=ATOL,
     )
 
+    # all times out of range
+    _ = cst.amp_time([-1])
+    _ = cst.amp_time(-1)
+    assert np.allclose(cst.amp_time([2]), np.exp(-1j * 2 * np.pi * 2 * freq0), rtol=0, atol=ATOL)
+
     assert_log_level(log_capture, None)
+
+    vals = td.components.data.data_array.TimeDataArray([1, 2], coords=dict(t=[-1, -0.5]))
+    dataset = td.components.data.dataset.TimeDataset(values=vals)
+    cst = td.CustomSourceTime(source_time_dataset=dataset, freq0=freq0, fwidth=0.1e12)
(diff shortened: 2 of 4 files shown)


momchil-flex:
Thanks, the fix looks good, but don't we still want a validator to error or at least warn that (as far as I understand?) this source will basically inject nothing in the simulation?

caseyflex:
> Thanks, the fix looks good, but don't we still want a validator to error or at least warn that (as far as I understand?) this source will basically inject nothing in the simulation?

It will actually inject something -- the usual behavior of CustomSourceTime is to just use the first or last value of the envelope provided for times outside of the envelope definition time range. But yeah, we should probably validate this because this is unlikely to be intentional.

caseyflex:
Added validation (a warning based on simulation tmesh). This caused another error where tmesh couldn't be generated if custom media aren't loaded, so I skip this validation if tmesh can't be computed.

momchil-flex (approved):

PR #1903, opened 2024-08-15
Add Simulation.num_computational_cells property

momchil-flex:


alec-flexcompute (approved):
Looks good! If it passes these tests you wrote I think it should be good to go

e-g-melo (approved):
Thanks @momchil-flex!

I would suggest another name because `num_cells` sounds very similar to `num_computational_cells`. But, you have explained the difference in the documentation, so that is not critical. It is up to you.

momchil-flex:
> I would suggest another name because `num_cells` sounds very similar to `num_computational_cells`. But, you have explained the difference in the documentation, so that is not critical. It is up to you.

I'm open do discussion, I just thought that `num_computational_cells` makes it a bit clearer what the difference from `num_cells` is, as opposed to just `grid_points`. But I could change to `num_computational_grid_points` to be more in line with the log and GUI? It's a bit unfortunate that we use `cells` in some places and `points` in others.

e-g-melo:
> I'm open do discussion, I just thought that `num_computational_cells` makes it a bit clearer what the difference from `num_cells` is, as opposed to just `grid_points`. But I could change to `num_computational_grid_points` to be more in line with the log and GUI? It's a bit unfortunate that we use `cells` in some places and `points` in others.

That is true! We should have used the same name. Optionally, we could change grid points to grid cells in Log, GUI, and the documentation. Do you think that would cause confusion?

momchil-flex:
Yeah I think unfortunately that could be confusing too as it will e.g. appear differently in the log for different solver versions...

momchil-flex:
I guess I'll just go with `num_computational_grid_points` here.

Issue #1735, opened 2024-06-03
Automatic setup of path integrals


dmarek-flex:
As simulations become more complex, setting up the path integrals for each port will become tedious! We should implement a method that automatically creates path integrals from very simple user input. For example, defining a port should only require the user to specify a port plane along with the structures representing the positive terminal and the negative terminal (ground reference). Then, using the geometry of the chosen structures we should be able to automatically define both integrals needed for computing voltage and current.

dmarek-flex:
The contour integrals needed by current computations can also be improved. Ideally we want the path around the conductor to be as close as possible to the surface. This will require access to the Yee grid and possibly depend on the conformal mesh strategy used. @weiliangjin2021 I am curious if the magnetic field of a Yee cell that has been modified by conformal meshing could still be used for these computations, or would a buffer cell be needed?

weiliangjin2021:
> I am curious if the magnetic field of a Yee cell that has been modified by conformal meshing could still be used for these computations, or would a buffer cell be needed?

For those cells, its permeability is also affected (no longer 1). In that regard, best to avoid crossing material interfaces?

dmarek-flex:
Ok, so we need the closest cells that do not intersect with the surfaces of material interfaces.

tylerflex:
@dmarek-flex updates on this? was it solved by #1904 ?

dmarek-flex:
No, not resolved yet

PR #1904, opened 2024-08-15
Fixed invalid s-parameters from wave port.

QimingFlex:
S-matrix is defined by:
CodeCogsEqn

where I is the current flows **into** the port, so I here introduced a invalidator to throw an error if the integral sign of `CurrentIntegralAxisAligned` doesn't match the port's direction.  In the subsequent S-matrix computation, if the voltage path is inversely chosen, it results in a negative Z_R, leading to invalid S-matrix. To address this, we now flip the sign where a negative Z_R occurs, which corresponds to where a negative voltage is given. This yields the correct S-matrix.

### DIFF
--- a/tidy3d/plugins/microwave/custom_path_integrals.py
+++ b/tidy3d/plugins/microwave/custom_path_integrals.py
@@ -6,13 +6,14 @@ from typing import Literal
 
 import numpy as np
 import pydantic.v1 as pd
+import shapely
 import xarray as xr
 
 from ...components.base import cached_property
 from ...components.data.data_array import FreqDataArray, FreqModeDataArray, TimeDataArray
 from ...components.data.monitor_data import FieldData, FieldTimeData, ModeSolverData
 from ...components.geometry.base import Geometry
-from ...components.types import ArrayFloat2D, Ax, Axis, Bound, Coordinate
+from ...components.types import ArrayFloat2D, Ax, Axis, Bound, Coordinate, Direction
 from ...components.viz import add_ax_if_none
 from ...constants import MICROMETER, fp_eps
 from ...exceptions import DataError, SetupError
@@ -396,3 +397,16 @@ class CustomCurrentIntegral2D(CustomPathIntegral2D):
             arrowprops=ARROW_CURRENT,
         )
         return ax
+
+    @cached_property
+    def sign(self) -> Direction:
+        """Uses the ordering of the vertices to determine the direction of the current flow."""
+        linestr = shapely.LineString(coordinates=self.vertices)
+        is_ccw = shapely.is_ccw(linestr)
+        # Invert statement when the vertices are given as (x, z)
+        if self.axis == 1:
+            is_ccw = not is_ccw
+        if is_ccw:
+            return "+"
+        else:
+            return "-"
--- a/tidy3d/plugins/smatrix/component_modelers/terminal.py
+++ b/tidy3d/plugins/smatrix/component_modelers/terminal.py
@@ -16,7 +16,7 @@ from ....components.source import GaussianPulse
 from ....components.types import Ax
 from ....components.viz import add_ax_if_none, equal_aspect
 from ....constants import C_0, OHM
-from ....exceptions import ValidationError
+from ....exceptions import Tidy3dError, ValidationError
 from ....web.api.container import BatchData
 from ..ports.base_lumped import AbstractLumpedPort
 from ..ports.base_terminal import AbstractTerminalPort, TerminalPortDataArray
@@ -174,7 +174,7 @@ class TerminalComponentModeler(AbstractComponentModeler):
 
     @cached_property
     def _source_time(self):
-        """Helper to create a time domain pulse for the frequeny range of interest."""
+        """Helper to create a time domain pulse for the frequency range of interest."""
         freq0 = np.mean(self.freqs)
         fdiff = max(self.freqs) - min(self.freqs)
         fwidth = max(fdiff, freq0 * FWIDTH_FRAC)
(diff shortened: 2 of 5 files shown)


momchil-flex (approved):
I think it makes sense but @weiliangjin2021 or ideally @dmarek-flex should know better.

dmarek-flex:
Thanks, this fix will work and help reverse user error. The only problem I find is that it kind of changes the Port setup behind the scenes. I much prefer the Validator approach that you did for the current integral, then how the sign of the voltage is changed in the port.

I think a better fix for this problem is to add another validator to the port for the voltage integral that ensures the positive terminal (and not the negative terminal) of the voltage integral is inside the current loop. This should be easy for the AxisAligned versions of the integrals. More work for the Custom versions. It would also be nice if you could extend your current integral validator to handle the case of the `CustomCurrentIntegral2D` (you just need to figure out if the vertices are ordered CW or CCW).

Setting up these integrals correctly is a bit of a pain, which is also why we are going to try and [automate these steps](https://github.com/flexcompute/tidy3d/issues/1735).

QimingFlex:
@dmarek-flex I'm thinking the same way but I haven't figured out how we know if the terminal is positive. That's why I try to change it behind the scenes as the port impedance will tell us if the integral is correct or not.

weiliangjin2021:
>  I much prefer the Validator approach that you did for the current integral, then how the sign of the voltage is changed in the port.

>  I'm thinking the same way but I haven't figured out how we know if the terminal is positive. That's why I try to change it behind the scenes as the port impedance will tell us if the integral is correct or not.

If pursuing that approach, I think you can just throw an error where you currently flip the sign.

But as @dmarek-flex also mentioned, before we have the automatic approach, there will be lots of pain setting up those path. I'm thinking if flipping the sign of voltage (and consequently the impedance) is all we need when voltage path is of the wrong direction, we can just silently handle it internally?

QimingFlex:
@weiliangjin2021 Yeah, agree it is a lot of pain setting these paths if seeking automation. I went through the formulations and am pretty sure that flipping the sign is all we need. Throwing an error where I flip the sign wouldn't be ideal I guess as this is in post-processing and would ask users to re-run the simulation.

dmarek-flex:
> @dmarek-flex I'm thinking the same way but I haven't figured out how we know if the terminal is positive. That's why I try to change it behind the scenes as the port impedance will tell us if the integral is correct or not.

Determining which terminal is the positive one is straightforward using the sign parameter. For the `VoltageIntegralAxisAligned`, when sign="+" the positive terminal is the terminal with the larger coordinate (the path is a simple straight line along one dimension). You can see the logic in the plotting function to double check.

dmarek-flex:
> I'm thinking if flipping the sign of voltage (and consequently the impedance) is all we need when voltage path is of the wrong direction, we can just silently handle it internally?

That should work as a temporary fix until we improve other parts.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -236,6 +236,12 @@ def port_VI(port_out: AbstractTerminalPort, sim_data: SimulationData):
         Z_numpy = port_impedances.transpose(*PortDataArray._dims).values.reshape(
             (len(self.freqs), len(port_names), 1)
         )
+
+        # Check for negative real part of port impedance and adjust V and Z accordingly
+        negative_real_Z = np.real(Z_numpy) < 0
+        V_numpy = np.where(negative_real_Z, -V_numpy, V_numpy)
+        Z_numpy = np.where(negative_real_Z, -Z_numpy, Z_numpy)
```
dmarek-flex, line 248: Maybe add a sanity check to make sure that the sign choice is consistent over all frequencies. The impedance should not change sign as the frequency changes (at least for typical transmission lines)

tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -185,3 +190,19 @@ def _check_voltage_or_current(cls, val, values):
...
+        Validate that the sign of current_integral matches the direction.
+        """
+        if val is None or isinstance(val, CustomCurrentIntegral2D):
+            return val
+
+        direction = values.get("direction")
+        name = values.get("name")
+        if val.sign != direction:
+            raise ValidationError(
+                f"'current_integral' sign must match the '{name}' direction ({direction})."
```
dmarek-flex: add missing quotes around direction

tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -185,3 +190,19 @@ def _check_voltage_or_current(cls, val, values):
                 "At least one of 'voltage_integral' or 'current_integral' must be provided."
             )
         return val
+
+    @pd.validator("current_integral", always=True)
+    def validate_current_integral_sign(cls, val, values):
+        """
+        Validate that the sign of current_integral matches the direction.
```
dmarek-flex: add double single quotes around current_integral

dmarek-flex:
After further discussion, let's keep it simple and just silently swap the voltage sign when the real part of the line impedance is found to be negative.

Let's just at least implement the sanity check to make sure the sign is the same for all frequencies.

dmarek-flex:
@QimingFlex please take a look at the changes I made and do a final check.

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -396,3 +397,16 @@ def plot(
             arrowprops=ARROW_CURRENT,
         )
         return ax
+
+    @cached_property
```
QimingFlex, line 401: I think the ordering of vertices alone is not enough to determine the current flow. For example, consider two ports placed in the x-y plane. If we have the port direction as '+', then the ccw orientation represents '+' and there is no problem. While if we have the port direction as '-', then the sign of the current should be '-' as well. But with this approach, it will be incorrectly determined as '+'.

dmarek-flex: This function only determines the sign of the current integral. The port direction and current integral sign are still compared in the WavePort's validator. In your example, the validator will correctly determine that there is a mismatch.

dmarek-flex: I just generalized the code you had  before to include the `CustomCurrentIntegral2D`

QimingFlex: Ah I see, so this is a sign of the integral rather than the sign of current flow. The rest looks good, thanks!

dmarek-flex (approved):

dmarek-flex:
@QimingFlex ok I think you can squash, rebase, and merge now!

PR #1907, opened 2024-08-16
Updated the medium intersection validator to handle 2D projection monitors without raising errors.

QimingFlex:
I updated the `_projection_monitors_homogeneous` validator to handle the error of medium intersections outside the computational domain in 2D simulations. The original validator checks if each face of a projection monitor intersects more than one medium. In 2D simulations, this was generating errors when faces normal to the 0-dim direction were included in the validation, which is not necessary. I removed those unnecessary faces accordingly for 2d simulations to avoid such errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Stopped raising error when field projection monitor intersecting structures outside of the simulation domain.
+
 ## [2.7.3] - 2024-09-12
 
 ### Added
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -136,6 +136,12 @@ NUM_STRUCTURES_WARN_EPSILON = 10_000
 # height of the PML plotting boxes along any dimensions where sim.size[dim] == 0
 PML_HEIGHT_FOR_0_DIMS = inf
 
+# fraction of structure size to monitor along 0 dim in 2D simulations
+FRACTION = 0.9
+
+# Minimum size of monitor
+MIN_MONITOR_SIZE = 1e-12
+
 
 class AbstractYeeGridSimulation(AbstractSimulation, ABC):
     """
@@ -2756,7 +2762,51 @@ class Simulation(AbstractYeeGridSimulation):
         with log as consolidated_logger:
             for monitor_ind, monitor in enumerate(val):
                 if isinstance(monitor, (AbstractFieldProjectionMonitor, DiffractionMonitor)):
-                    mediums = Scene.intersecting_media(monitor, total_structures)
+                    if monitor.volume() != 0.0:
+                        if not (
+                            all(
+                                monitor_min >= sim_min
+                                for monitor_min, sim_min in zip(
+                                    monitor.bounds[0], structure_bg.geometry.bounds[0]
+                                )
+                            )
+                            and all(
+                                monitor_max <= sim_max
+                                for monitor_max, sim_max in zip(
+                                    monitor.bounds[1], structure_bg.geometry.bounds[1]
+                                )
+                            )
+                        ):
+                            exclude_surfaces = []
+                            dimensions = ["x", "y", "z"]
+                            for i, dim in enumerate(dimensions):
+                                if monitor.bounds[0][i] <= structure_bg.geometry.bounds[0][i]:
+                                    exclude_surfaces.append(f"{dim}-")
+                                if monitor.bounds[1][i] >= structure_bg.geometry.bounds[1][i]:
+                                    exclude_surfaces.append(f"{dim}+")
+                            new_min = np.maximum(monitor.bounds[0], structure_bg.geometry.bounds[0])
+                            new_max = np.minimum(monitor.bounds[1], structure_bg.geometry.bounds[1])
+                            new_size = tuple(new_max - new_min)
+                            # Add a minimum value to the monitor along 0-dim in 2D simulations
+                            new_size = tuple(
+                                MIN_MONITOR_SIZE if num == 0 else num for num in new_size
+                            )
+                            new_center = tuple((new_max + new_min) / 2)
+                            monitor_test = monitor.updated_copy(center=new_center, size=new_size)
+                            monitor_dict = monitor_test.dict()
+                            # Append exclude surfaces for 2D simulation to existing ones
+                            existing_exclude_surfaces = monitor_dict.get("exclude_surfaces") or []
+                            updated_exclude_surfaces = list(
+                                set(existing_exclude_surfaces + exclude_surfaces)
+                            )
+                            monitor_dict["exclude_surfaces"] = updated_exclude_surfaces
+                            surfaces = monitor_test.surfaces_with_exclusion(**monitor_dict)
+                            mediums = set()
+                            for surface in surfaces:
+                                _mediums = Scene.intersecting_media(surface, total_structures)
+                                mediums.update(_mediums)
+                    else:
+                        mediums = Scene.intersecting_media(monitor, total_structures)
                     # make sure there is no more than one medium in the returned list
                     if len(mediums) > 1:
                         raise SetupError(
--- a/tests/test_components/test_simulation.py
+++ b/tests/test_components/test_simulation.py
@@ -2977,3 +2977,35 @@ def test_validate_sources_monitors_in_bounds():
             grid_spec=td.GridSpec(wavelength=1.0),
             monitors=[mode_monitor],
         )
+
+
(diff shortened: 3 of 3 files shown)


QimingFlex:
Sorry not ready for review,  just found some errors need to fix.

tidy3d/components/scene.py
```diff
@@ -234,6 +234,8 @@ def intersecting_media(
             Object for which intersecting media are to be detected.
         structures : List[:class:`.AbstractMedium`]
             List of structures whose media will be tested.
+        zero_dim_ind : int
```
tylerflex: technically in the docstrings for optional arguments (with defaults) we'd put the default in here like 

```
zero_dim_ind : int = None
```

tidy3d/components/scene.py
```diff
@@ -246,8 +248,25 @@ def intersecting_media(
...
             return mediums
 
-        # if the test object is a volume, test each surface recursively
-        surfaces = test_object.surfaces_with_exclusion(**test_object.dict())
+        # If the test object is a volume, test each surface recursively
+        # and exclude surfaces for 2D simulations where one dimension is ignored.
+        exclude_surfaces_2d: List[str] = []
+        if zero_dim_ind is not None:
+            exclude_surfaces_2d = {0: ["x-", "x+"], 1: ["y-", "y+"], 2: ["z-", "z+"]}.get(
+                zero_dim_ind, []
```
tylerflex: I suppose this could just be a list? 

```py
[["x-", "x+"], ["y-", "y+"], ["z-", "z+"]][zero_dim_ind]
```

Also if the zero_dim_ind is not found, we probably just want to error here right? instead of returning an empty list?

tidy3d/components/simulation.py
```diff
@@ -2707,7 +2714,10 @@ def _projection_monitors_homogeneous(cls, val, values):
         with log as consolidated_logger:
             for monitor_ind, monitor in enumerate(val):
                 if isinstance(monitor, (AbstractFieldProjectionMonitor, DiffractionMonitor)):
-                    mediums = Scene.intersecting_media(monitor, total_structures)
+                    if isinstance(monitor, AbstractFieldProjectionMonitor):
```
tylerflex: do we need this `if else` what is different about the diffraction monitor?

tylerflex (changes requested):
Thanks @QimingFlex, just a few comments. Might want to get @dbochkov-flexcompute to take a quick look just to make sure this code doesn't have any unintended consequences in the `Scene`.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Error when plotting mode plane PML and the simulation has symmetry.
 - Validators using `TriangleMesh.intersections_plane` will fall back on bounding box in case the method fails for a non-watertight mesh.
 - Bug when running the same `ModeSolver` first locally then remotely, or vice versa, in which case the cached data from the first run is always returned.
+- Error when field projection monitor intersecting structures outside of the simulation domain in 2D simulations.
```
momchil-flex: This should be in Unreleased, not in 2.7.2

momchil-flex (commented):
I think that this PR may only be fixing a subclass of the edge cases that are in principle possible. For example, it doesn't even have to be a 2D simulation for an issue to arise: assume you have a structure entirely outside of the (3D) simulation domain - I think currently we allow this with a warning. A projection monitor could also intersect that and the error will be raised even though actually it's ok. Also this would be true for DiffractionMonitor, which in this PR is not addressed at all.

I am also not a fan of modifying `Scene.intersecting_media` for this edge case (but also if my described edge case above is true, the `zero_dim_ind` modification is not general enough anyway). Maybe a better approach is to make these validators `post_init` so that we can compute `self.simulation_bounds`; make a `Box.from_bounds` using these bounds; and update the monitor we are testing to have center and size that are given by the intersection of its nominal center and size and the simulation Box. Then pass that to `intersecting_media`.

This probably should also be done for other homogeneous media validators like for PlaneWave, GaussianBeam, and maybe tfsf? Could you take a look @QimingFlex ?

QimingFlex:
> Thanks @QimingFlex, just a few comments. Might want to get @dbochkov-flexcompute to take a quick look just to make sure this code doesn't have any unintended consequences in the `Scene`.

Thanks @tylerflex for the comments. Yeah, modifying the `scene` seems likely to cause potential risks, so I'll have to drop this idea. I'll come up with a better approach.

QimingFlex:
> I think that this PR may only be fixing a subclass of the edge cases that are in principle possible. For example, it doesn't even have to be a 2D simulation for an issue to arise: assume you have a structure entirely outside of the (3D) simulation domain - I think currently we allow this with a warning. A projection monitor could also intersect that and the error will be raised even though actually it's ok. Also this would be true for DiffractionMonitor, which in this PR is not addressed at all.
> 
> I am also not a fan of modifying `Scene.intersecting_media` for this edge case (but also if my described edge case above is true, the `zero_dim_ind` modification is not general enough anyway). Maybe a better approach is to make these validators `post_init` so that we can compute `self.simulation_bounds`; make a `Box.from_bounds` using these bounds; and update the monitor we are testing to have center and size that are given by the intersection of its nominal center and size and the simulation Box. Then pass that to `intersecting_media`.
> 
> This probably should also be done for other homogeneous media validators like for PlaneWave, GaussianBeam, and maybe tfsf? Could you take a look @QimingFlex ?

Thanks @momchil-flex . Yeah, modifying the scene is not a good idea for this corner case. Initially, I took the approach of updating the monitor to pass this test, but it failed in some of the diffraction tests. I'll need some time to update this to be more general, including the 3D case and the Diffraction Monitor you mentioned.

QimingFlex:
Hi @momchil-flex , My thoughts on the thorough fix. There are three key elements in this fix: structures, simulation domain, and projection monitors. Regarding the first two elements (structures and simulation domain), for both 2D and 3D simulations, we must allow structures to extend outside the simulation domain in some cases: 
1. A structure defined by finite size but terminated by PML. In this case, we need to reduce the size along the PML directions to allow the structure to extrude into the PMLs. 
2. Defining a structure's size as 0 along the 0-dimension in a 2D simulation would result in a surface rather than a volume, causing the simulation to fail. Therefore, it's better to allow structures to extend outside the simulation domain.
Now, concerning projection monitors, I believe the goal of this fix is to stop throwing intersection medium errors when the projection monitor is outside the simulation domain. To achieve this, we can modify either the structures or the monitor. Modifying structures is much more challenging as it involves various geometries, so we need to focus on modifying the monitor.
My initial approach was to remove the two surface monitors that were outside the simulation domain, which I thought was the safest fix. While this failed in some autograd tests. Another option is to assign a very large or infinite size to the monitor along the direction that extends outside the simulation domain. This has the same effect as the first approach but fails in some other tests. At this point, I think it's better to use an infinite size for the monitor and then investigate the failed tests next. Does this make sense to you?

momchil-flex:
@QimingFlex the approach that I propose was in my previous comment

> Maybe a better approach is to make these validators post_init so that we can compute self.simulation_bounds; make a Box.from_bounds using these bounds; and update the monitor we are testing to have center and size that are given by the intersection of its nominal center and size and the simulation Box. Then pass that to intersecting_media.

Do you see any issue with this proposed method? Maybe this explanation is not very clear, we can discuss further.

QimingFlex:
> @QimingFlex the approach that I propose was in my previous comment
> 
> > Maybe a better approach is to make these validators post_init so that we can compute self.simulation_bounds; make a Box.from_bounds using these bounds; and update the monitor we are testing to have center and size that are given by the intersection of its nominal center and size and the simulation Box. Then pass that to intersecting_media.
> 
> Do you see any issue with this proposed method? Maybe this explanation is not very clear, we can discuss further.

I believe one problematic case could arise in 2D simulations. When calculating the intersection, the updated monitor's size along the zero-dimension would become 0. This would transform the monitor from its intended four-surfaces structure into a single surface out of simulation domain, which is incorrect. And in general 3D simulations it might still throw errors on surfaces at the simulation edge.

QimingFlex:
@momchil-flex This fix considers the case where a box monitor is given. It checks its bounds against the simulation bounds. If any dimension of the monitor exceeds the simulation domain, we do the following:

1. Remove the corresponding surface that is totally out of the simulation domain.
2. Trim the monitor size to be the same as the simulation bounds.

This would generally lead to the problem I mentioned above, where a 2D simulation could reduce a box monitor to a surface monitor. To avoid this, I introduced a very small value along the 0-dimension to keep the monitor as a box. Finally this fix is able to prevent errors from being thrown in both 2D and 3D simulations.

However, the backend needs to be fixed as well. Do the backend perform some medium checks in C++? Currently, I can't get correct solutions which I believe is related to the wrong medium assigned to the monitor.

PR #1909, opened 2024-08-18
Dilation support for `JaxPolySlab`

yaugenst-flex:
Adds basic support for `dilation` argument in `JaxPolySlab`.

This also enables differentiation w.r.t. dilation, but note that this currently only works if no sidewall angle is present (see tests). Actually differentiating w.r.t. everything works (i.e., `dilation`, `sidewall_angle`, and `slab_bounds`) as long as there is no / very little sidewall angle.

### DIFF
--- a/tidy3d/components/geometry/polyslab.py
+++ b/tidy3d/components/geometry/polyslab.py
@@ -10,6 +10,7 @@ import autograd.numpy as np
 import pydantic.v1 as pydantic
 import shapely
 import xarray as xr
+from autograd.tracer import isbox
 from matplotlib import path
 
 from ...constants import LARGE_NUMBER, MICROMETER, fp_eps
@@ -1316,6 +1317,13 @@ class PolySlab(base.Planar):
         shapely_poly = PolySlab.make_shapely_polygon(vertices)
         if shapely_poly.is_valid:
             return vertices
+        elif isbox(vertices):
+            raise NotImplementedError(
+                "The dilation caused damage to the polygon. "
+                "Automatically healing this is currently not supported when "
+                "differentiating w.r.t. the vertices. Try increasing the spacing "
+                "between vertices or reduce the amount of dilation."
+            )
         # perform healing
         poly_heal = shapely.make_valid(shapely_poly)
         return PolySlab._proper_vertices(list(poly_heal.exterior.coords))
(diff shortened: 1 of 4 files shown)


tidy3d/plugins/adjoint/components/geometry.py
```diff
@@ -526,6 +527,19 @@ def _proper_vertices(vertices: ArrayFloat2D) -> jnp.ndarray:
...
+    @staticmethod
+    def _heal_polygon(vertices: jnp.ndarray) -> jnp.ndarray:
+        """heal a self-intersecting polygon."""
+        shapely_poly = PolySlab.make_shapely_polygon(jax.lax.stop_gradient(vertices))
+        if shapely_poly.is_valid:
+            return vertices
+
+        raise NotImplementedError(
+            "It looks like the dilation causes damage to the polygon. Automatically "
+            "healing this is currently not supported for JAX PolySlabs. Try increasing "
```
tylerflex: `'JaxPolySlab' objects`

tidy3d/components/geometry/polyslab.py
```diff
@@ -1316,6 +1317,13 @@ def _heal_polygon(vertices: np.ndarray) -> np.ndarray:
         shapely_poly = PolySlab.make_shapely_polygon(vertices)
         if shapely_poly.is_valid:
             return vertices
+        elif isbox(vertices):
+            raise NotImplementedError(
+                "It looks like the dilation causes damage to the polygon. "
+                "Automatically healing this is currently not supported when "
+                "differentiating w.r.t. the vertices. Try increasing the spacing "
```
tylerflex, line 1324: `'vertices'`

tidy3d/components/geometry/polyslab.py
```diff
@@ -1316,6 +1317,13 @@ def _heal_polygon(vertices: np.ndarray) -> np.ndarray:
         shapely_poly = PolySlab.make_shapely_polygon(vertices)
         if shapely_poly.is_valid:
             return vertices
+        elif isbox(vertices):
+            raise NotImplementedError(
+                "It looks like the dilation causes damage to the polygon. "
```
tylerflex: Let's just say "The dilation caused damage to the polygon. ", same with the error message in JaxPolySlab ('it looks like' sounds slightly strange in an error message to me).

tylerflex (commented):
Thanks @yaugenst-flex ! just a couple minor comments

tylerflex (approved):
Feel free to merge @yaugenst-flex when you're ready. no rush

PR #1911, opened 2024-08-20
Only warn once per structure into PML to avoid duplicate warnings

momchil-flex:
This was brought up by MC, a structure that touches the PML on multiple sides would raise multiple identical warnings. One option would be to include more information in the warning but to avoid verbosity I went the opposite way of just issuing one warning per structure.

### DIFF
--- a/tests/test_package/test_log.py
+++ b/tests/test_package/test_log.py
@@ -129,7 +129,7 @@ def test_logging_warning_capture():
         medium=td.Medium(permittivity=2, frequency_range=[0.5, 1]),
     )
 
-    # 2 warnings: inside pml
+    # 1 warning: inside pml
     box_in_pml = td.Structure(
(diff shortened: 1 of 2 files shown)


tylerflex (changes requested):
Looks fine to me. 
1. Instead of using `warn = False`, couldn't we just `continue` after the logger is called?
2. Double check the submodule commit, not sure if we want that?

tylerflex (approved):

momchil-flex:
> 1. Instead of using `warn = False`, couldn't we just `continue` after the logger is called?

Oh I see. We're trying to continue two nested loops though - is there a way?

> 2. Double check the submodule commit, not sure if we want that?

Ah yeah thanks.

tylerflex:
ah didnt see the 2nd loop, I think it's fine then

PR #1914, opened 2024-08-21
More autograd improvements

tylerflex:
Introduces a set of additional improvements to autograd plugin:
1. PolySlab grad monitor data is only stored on the plane where needed.
2. Validate FWD sim, even when it's a server-side run.
3. Properly handle error if no adjoint sources are present and add a more useful fix.

Let me know if you'd like this merged int https://github.com/flexcompute/tidy3d/pull/1891 for one PR? But would be good to include both in a 2.7.3


yaugenst-flex (approved):
LGTM

(also, I'm in favor of small PR)

PR #1915, opened 2024-08-22
🗻  add cylinder differentiation support

tylerflex:


tidy3d/components/geometry/primitives.py
```diff
@@ -215,6 +222,105 @@ def _only_middle_for_infinite_length_slanted_cylinder(cls, val, values):
             )
         return val
 
+    def to_polyslab(self, num_pts_radius: int = N_PTS_CYLINDER_POLYSLAB, **kwargs) -> PolySlab:
+        """Convert instance of ``Cylinder`` into a discretized version using ``PolySlab``.
+
+        Parameters
+        ----------
+        num_pts_radius : int = 100
+            Number of points in the radius of the discretized polyslab.
```
momchil-flex: The hard-coded default is 51 now.

Also, do you mean number of points in the *circumference*?

tidy3d/components/geometry/primitives.py
```diff
@@ -215,6 +222,105 @@ def _only_middle_for_infinite_length_slanted_cylinder(cls, val, values):
...
+                vjp_xs = np.sum(xs_ * vjps_vertices_xs)
+                vjp_ys = np.sum(ys_ * vjps_vertices_ys)
+
+                vjps[path] = vjp_xs + vjp_ys
+
+            elif "center" in path:
+                _, center_index = path
+                if center_index == self.axis:
+                    raise NotImplementedError(
+                        "Can't current differentiate Cylinder with respect to its 'center' along "
```
momchil-flex: "Currently cannot"

tidy3d/components/geometry/primitives.py
```diff
@@ -215,6 +222,105 @@ def _only_middle_for_infinite_length_slanted_cylinder(cls, val, values):
...
+        """Set of x and y points for the unit circle when discretizing cylinder as a polyslab."""
+        angles = np.linspace(0, 1, num_pts_radius + 1)[:-1]
+        xs = np.cos(angles)
+        ys = np.sin(angles)
+        return np.stack((xs, ys), axis=0)
+
+    def compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradFieldMap:
+        """Compute the adjoint derivatives for this object."""
+
+        polyslab = self.to_polyslab()
```
momchil-flex: For a large cylinder, the default 51 points may not be enough. Should there be a points per wvl kind of thing again or something like that?

tests/test_components/test_autograd.py
```diff
@@ -610,7 +623,7 @@ def objective(*args):
     assert anp.all(grad != 0.0), "some gradients are 0"
 
     # numerical gradients
-    delta = 1e-1
+    delta = 1e-3
```
momchil-flex, line 626: Clever workaround - not sure I fully understand the tests, should I believe that gradients for all supported Cylinder parameters are tested vs. numerical differences?

tidy3d/components/geometry/primitives.py
```diff
@@ -5,23 +5,30 @@
...
+from .polyslab import PolySlab
 
 # for sampling conical frustum in visualization
 _N_SAMPLE_CURVE_SHAPELY = 40
 
 # for shapely circular shapes discretization in visualization
 _N_SHAPELY_QUAD_SEGS = 200
 
+# number of points to discretize polyslab used to represent cylinder in `Cylinder.to_polyslab()`
+N_PTS_CYLINDER_POLYSLAB = 51
```
momchil-flex: Note some of the previous constants are private `_N...`, maybe this one should be too?

tylerflex: sure I can do that

tests/test_components/test_autograd.py
tylerflex: what's the workaround?

the numerical gradient tests are more things that are done individually to test various combinations of structure (medium / geometry) and monitor. In this case I tested the cylinder parameters and they matched numerical yes.

tidy3d/components/geometry/primitives.py
tylerflex: hm, yea probably would be better. just implemented something where it does 10 pts per wavelength in the material to discretize the diameter. Note that I dont think it needs to be very finely discretized based on my numerical tests, this seems sufficient and I'll confirm with a little bit more testing.

tidy3d/components/geometry/primitives.py
tylerflex: fixed this, and yea circumference. will change all mentions of radius to circumference..

tylerflex:
numerical tests worked with the new parameters and I also tested the metalens notebook again and it's working good.

tests/test_components/test_autograd.py
momchil-flex: Oh sorry I just meant the general workaround taken in this PR to discretize the cylinder as a polygon. But come to think of it there's no other way, any handling would effectively be doing that.

tidy3d/components/geometry/primitives.py
```diff
@@ -5,23 +5,33 @@
...
 from . import base
+from .polyslab import PolySlab
 
 # for sampling conical frustum in visualization
 _N_SAMPLE_CURVE_SHAPELY = 40
 
 # for shapely circular shapes discretization in visualization
 _N_SHAPELY_QUAD_SEGS = 200
 
+# number of points to discretize polyslab used to represent cylinder in `Cylinder.to_polyslab()`
```
momchil-flex: `Default number` ?

tidy3d/components/geometry/primitives.py
```diff
@@ -215,6 +225,123 @@ def _only_middle_for_infinite_length_slanted_cylinder(cls, val, values):
...
+
+    def compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradFieldMap:
+        """Compute the adjoint derivatives for this object."""
+
+        # compute number of points in the diameter of the polyslab using resolution info
+        wvl0 = C_0 / derivative_info.frequency
+        wvl_mat = wvl0 / max(1.0, np.sqrt(abs(derivative_info.eps_in)))
+
+        diameter = 2 * np.pi * self.radius
+        wvls_in_diameter = diameter / wvl_mat
```
momchil-flex: Again it seems like you mean circumference and not diameter... diameter is just 2 * radius. So the value is correct since you're computing `num_pts_circumference` but the names should be `circumference `and `wvls_in_circumference`?

momchil-flex (approved):

tidy3d/components/geometry/primitives.py
tylerflex: added

tidy3d/components/geometry/primitives.py
tylerflex: oops, just missed these variables. fixed

tylerflex:
ok, should be all resolved. Just squashed too.

Issue #1526, opened 2024-03-07
Translation of field data for easier overlap computation


momchil-flex:
**Is your feature request related to a problem? Please describe.**
If two different sets of (mode) field data are computed at different locations, it is not currently easy to compute the overlap. We should have a method to translate the coordinates of of the datasets to be centered on the other one's center.

**Describe the solution you'd like**
Something like this should work, but needs to be incorporated properly, and testing.

```
def translate_coords(mode: ModeSolverData, vector: List[float]):
    new_Ex = mode.Ex.assign_coords({
        "x": mode.Ex.x.values + vector[0],
        "y": mode.Ex.y.values + vector[1],
        "z": mode.Ex.z.values + vector[2],
    })
    new_Ey = mode.Ey.assign_coords({
        "x": mode.Ey.x.values + vector[0],
        "y": mode.Ey.y.values + vector[1],
        "z": mode.Ey.z.values + vector[2],
    })
    new_Ez = mode.Ez.assign_coords({
        "x": mode.Ez.x.values + vector[0],
        "y": mode.Ez.y.values + vector[1],
        "z": mode.Ez.z.values + vector[2],
    })
    new_Hx = mode.Hx.assign_coords({
        "x": mode.Hx.x.values + vector[0],
        "y": mode.Hx.y.values + vector[1],
        "z": mode.Hx.z.values + vector[2],
    })
    new_Hy = mode.Hy.assign_coords({
        "x": mode.Hy.x.values + vector[0],
        "y": mode.Hy.y.values + vector[1],
        "z": mode.Hy.z.values + vector[2],
    })
    new_Hz = mode.Hz.assign_coords({
        "x": mode.Hz.x.values + vector[0],
        "y": mode.Hz.y.values + vector[1],
        "z": mode.Hz.z.values + vector[2],
    })
    new_monitor = mode.monitor.updated_copy(
        center=[old_cent + vec for old_cent, vec in zip(mode.monitor.center, vector)]
    )
    return mode.updated_copy(
        monitor=new_monitor,
        Ex=new_Ex,
        Ey=new_Ey,
        Ez=new_Ez,
        Hx=new_Hx,
        Hy=new_Hy,
        Hz=new_Hz,
    )
```

momchil-flex:
Done in #2238

PR #1916, opened 2024-08-22
Added shift_field and outer_dot_shifted function to ElectroMagneticFi…

prashkh:
…eldData Class

@momchil-flex Based on your suggestion a few weeks ago, I started adding functions to the `ElectroMagneticFieldData` class so that we can easily shift fields and do mode overlaps. Let me know if this is what you were thinking of for implementation. I checked the results of the new `outer_dot_shifted` with the results from`outer_dot` when the relative shift is (0,0) and things are consistent.

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -633,6 +633,228 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
             fields[component] = field.interp(kwargs=kwargs, **interp_dict)
 
         return fields
+    
+
+    def _interpolate_and_shift(self, field: xr.DataArray, t1_new: np.ndarray, t2_new: np.ndarray, 
+                               dt1_shift: float, dt2_shift: float, res: float) -> xr.DataArray:
+        r"""Helper method to interpolate and shift a single field component.
+
+        Parameters
+        ----------
+        field : xr.DataArray
+            The field component to be interpolated and shifted
+        t1_new : np.ndarray
+            New coordinates for the first transverse dimension
+        t2_new : np.ndarray
+            New coordinates for the second transverse dimension
+        dt1_shift : float
+            Shift in the first transverse dimension [μm]
+        dt2_shift : float
+            Shift in the second transverse dimension [μm]
+        res : float
+            Resolution of the new grid [μm]
+
+        Returns
+        -------
+        xr.DataArray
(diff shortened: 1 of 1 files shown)


momchil-flex (commented):
Thanks @prashkh, this would certainly be super useful to have.

I had a somewhat different idea about how this can be incorporated in the code, but maybe I didn't convey it too well, or maybe you had your reasons to make some of these design decisions. But here's what I had in mind, let me think if you think this will work:

We can have something like a `ElectromagneticFieldData.translated_copy` method which allows you to translate the data along x, y, z. This does not interpolate - it just shifts all relevant data like the coordinates and the associated monitor position, as illustrated [here](https://github.com/flexcompute/tidy3d/issues/1526). This was written very quickly and would need to be tested, and also I think we need to shift `grid_expanded` (and just overall take a look at what attributes field data contains and shift everything that needs shifting).

This first step is the nontrivial one, but after that doesn't everything just come for free? Our current `outer_dot` already handles data that are not on the same grid, and I *think* it handles even the case where the grid spans are not the same, but if not, it could be fixed to do so. So a user would just do something like

```
outer_dot_shifted = data1.outer_dot(data2.translated_copy(dx=5, dy=0))
```

Basically you won't need to re-interpolate to a uniform grid, you won't need to handle things like not knowing which dimensions exactly are the tangential ones, and you won't need to introduce a lot of duplicated code in your current `outer_dot_shifted` method.

Regarding the point that you don't need to care about which dimensions are the tangential ones, in my example above I assume the *user* knows that and uses the `translated_copy` method properly. That is to say we allow translating along x y and z and it's up to the user to translate along x and y if their normal direction is z, etc.

prashkh:
@momchil-flex thanks for the feedback and the detailed explanation. I think the approach you described makes sense now. Let me try to implement that sometime next week and I'll let you know if I run into any issues.

caseyflex:
Will this resolve this issue?
https://github.com/flexcompute/tidy3d/issues/1526

momchil-flex:
> Will this resolve this issue? #1526

Yeah I reference it in my comment. It just needs to be done a bit more carefully and tested.

caseyflex:
> > Will this resolve this issue? #1526
> 
> Yeah I reference it in my comment. It just needs to be done a bit more carefully and tested.

Maybe the link was wrong in your comment?

momchil-flex:
Ah you are right sorry! Updated it just in case someone clicks it in the future.

PR #1917, opened 2024-08-23
🗻 better error when flux monitor used in autograd

tylerflex:


momchil-flex (approved):
It's OK for now even if it's the case, but does this handling mean that the error will only be raised after the forward simulation is done? If so it may be better to be able to catch errors like these in advance.

tylerflex:
I thought about this, it would be great it we could, but we have no way of knowing if a FluxMonitor is going to be used in an objective function (and therefore triggering this error) without running the forward simulation.

If we want to do this, we'd need some way to more explicitly specify which monitors are going to be used in the objective function, but it makes the API clunkier.

Maybe this is something that can be solved with this approach of running an emulated simulation to figure out (a) what monitors are used and (b) how to set up the adjoint source. but let's leave this for another PR?

momchil-flex:
Yep sounds good.

PR #1919, opened 2024-08-23
added validator to error out Bloch boundaries in along 0-dim in 2D simulations

QimingFlex:
2D simulations now only support PEC, PMC, and periodic boundaries. We raise an error for Bloch boundaries as we don't yet have a solution for them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `PolySlab` now raises error when differentiating and dilation causes damage to the polygon.
+- Validator `boundaries_for_zero_dims` to raise error when Bloch boundaries are used along 0-sized dims.
 
 ### Fixed
 - `DataArray` interpolation failure due to incorrect ordering of coordinates when interpolating with autograd tracers.
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -2410,7 +2410,7 @@ class Simulation(AbstractYeeGridSimulation):
     @pydantic.validator("boundary_spec", always=True)
     @skip_if_fields_missing(["size", "symmetry"])
     def boundaries_for_zero_dims(cls, val, values):
-        """Error if absorbing boundaries, unmatching pec/pmc, or symmetry is used along a zero dimension."""
+        """Error if absorbing boundaries, bloch boundaries, unmatching pec/pmc, or symmetry is used along a zero dimension."""
         boundaries = val.to_list
         size = values.get("size")
         symmetry = values.get("symmetry")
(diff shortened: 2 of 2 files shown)


momchil-flex:
When we first discussed this, I meant that you should just validate out Bloch boundaries in 2D simulations with *field projection monitors*, because we won't handle these correctly. Otherwise, we _sort of_ support Bloch boundaries in 2D simulations. This is illustrated for example [here](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/GratingEfficiency.html#Off-normal-incidence) - however it is true that a non-zero simulation size needs to be used along the Bloch dimension (while still having a single pixel), which is basically due to [our definition](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.BlochBoundary.html#tidy3d.BlochBoundary) of the Bloch vector as `2 * pi / (size along dimension)`. There is an old [issue](https://github.com/flexcompute/tidy3d-core/issues/158) to enable this somehow, but it requires some thought about the architecture.

In view of that, I think we should indeed validate out Bloch boundaries in 0-sized dims, but not the way you currently do this: it should not matter whether the simulation is 2d or 1d or whatever. It just matters whether a given dimension has size 0 and Bloch boundaries in it. So you don't need `is_2d_simulation`, you just need to check if the boundary condition in a `size==0` dim is Bloch, and then modify the error message to something like

```
                    raise SetupError(
                        f"The simulation has zero size along the {axis} axis, "
                        "using a Bloch boundary along such an axis is not supported because of "
                        "the Bloch vector definition in units of '2 * pi / (domain size)'. Use a small "
                        "but nonzero size along the dimension instead."
                    )
```

This still leaves one problem though: when there *are* projection monitors in the simulation, we turn on the 2D treatment when the simulation size is strictly 0. So if a user puts a projection monitor in a Bloch simulation with a single pixel along a dim (but slightly nonzero size), the projection will not be correct. I think maybe we should also validate that out? Namely, we can add a validator for simulation dimensions which have a single pixel in them (this will have to be post-init, since you need to compute the grid). Currently, this validator could just check that if the size of that dimension is not 0, and the boundary spec is not Bloch, it can warn that it is better to explicitly set the size of single-pixel simulations to 0. If Bloch boundaries found instead, we could check if there are projection monitors and error saying that these will produce wrong results with Bloch boundaries. In the future, we can hopefully fix some of the mess by 

1. Making 0-dim sims support Bloch boundaries (will need to think about how to define the Bloch vector then)
2. Making field projection monitors support Bloch boundaries.

What do you think?

QimingFlex:
Yeah, I understand our situation.
First, we need to throw an error if Bloch boundaries are assigned to a 0-dimensional structure. This should be a general error message for all simulations, as it violates our definition of the Bloch vector: 2 * pi / (size along dimension).
Next, if users define a Bloch boundary in a 3D simulation with a single pixel along one dimension and use a projection monitor, we should also throw an error, as this would generate incorrect solutions. After further consideration, I realize this is an issue for general 3D problems. For example, if we define a cylinder with a finite size along one dimension but apply periodic/Bloch boundaries along that dimension, and excite the problem with a normal incident wave, this would be equivalent to a 2D problem we defined for a cylinder. However, it won't give the correct results as 2D treatment is not turned on. 
Given this situation, I think we should throw an error when projection monitor touches periodic boundaries in a 3D simulation. This can be done in the frontend. 
In the future, both 1 and 2 seem promising, need to rethink about it after investigating the problem of  `oblique incidence to infinite long structure`.

momchil-flex:
Yes, that's exactly right. Merging this now and will wait for another PR from you to validate projection monitors touching periodic/Bloch boundaries in any-sized simulation.

PR #1923, opened 2024-08-27
2nd round of autograd fixes

tylerflex:
* some adjoint source fixes for FieldData involving H fields.
* better gradients for Box shifting boundaries based on permittivity.
* re-introduce automatic permittivity grabbing for geometry group shifting boundaries.
* better error and warning handling across the board.
* adds a way to visualize adjoint fields (only for development purposes currently, no public API).

### DIFF
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -37,7 +37,7 @@ TEST_POLYSLAB_SPEED = False
 
 # whether to run numerical gradient tests, off by default because it runs real simulations
 RUN_NUMERICAL = False
-_NUMERICAL_COMBINATION = ("center_list", "mode")
+_NUMERICAL_COMBINATION = ("size_element", "mode")
 
 TEST_MODES = ("pipeline", "adjoint", "speed")
 TEST_MODE = "speed" if TEST_POLYSLAB_SPEED else "pipeline"
@@ -529,7 +529,7 @@ if TEST_POLYSLAB_SPEED:
     args = [("polyslab", "mode")]
 
 
-args = [("size_element", "mode")]
+# args = [("size_element", "mode")]
 
 
 def get_functions(structure_key: str, monitor_key: str) -> typing.Callable:
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -37,6 +37,7 @@ TEST_POLYSLAB_SPEED = False
 
 # whether to run numerical gradient tests, off by default because it runs real simulations
 RUN_NUMERICAL = False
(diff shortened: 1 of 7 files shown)


tests/test_components/test_autograd.py
```diff
@@ -471,7 +471,7 @@ def field_vol_postprocess_fn(sim_data, mnt_data):
     def field_point_postprocess_fn(sim_data, mnt_data):
         value = 0.0
         for _, val in mnt_data.field_components.items():
-            value += abs(anp.sum(val.values))
+            value += abs(anp.sum(abs(val.values)))
```
yaugenst-flex, line 474: double `abs`? not that it matters i guess

tidy3d/web/api/autograd/autograd.py
yaugenst-flex, line 1: Could be nice to have private debug run function, or maybe a `@debug` decorator that could do things like this? Not really important but might be convenient to include in the future.

yaugenst-flex (approved):
This is great. I've been testing box and hfield changes and things seem to work much better now.

tidy3d/components/data/monitor_data.py
```diff
@@ -1069,46 +1068,33 @@ def to_adjoint_field_sources(self, fwidth: float) -> List[CustomCurrentSource]:
...
             for name, field_component in self.field_components.items():
+                # get the VJP values at frequency and apply adjoint phase
                 field_component = field_component.sel(f=freq0)
-                forward_amps = field_component.values
-                values = -1j * forward_amps
+                values = -1j * field_component.values
+
+                # make source go backwards
+                if "H" in name:
+                    values *= -1
```
momchil-flex, line 1083: Is this what we do in the adjoint plugin too?

I am trying to think about whether this is the correct thing to do in all cases. It's probably possible to see it somehow from the adjoint formulation if written for the H field. I think this is correct. I'm just wondering a little because the intuition of this minus sign "making source go backwards" makes sense for a real propagating mode, but it gets a bit murkier for evanescent or lossy modes. The other option is to take np.conj(values). If the E field is real and the mode is lossless and propagating, then the H field is imaginary and the two operations are the same. In general the `np.conj(values)` operation would be like time reversal while the `values *= -1` operation is like reflection of the mode in the source plane, which produce different results in some cases.

I do think the `-1` is correct here though but do you remember if it comes out of the adjoint math?

tidy3d/web/api/autograd/autograd.py
tylerflex: which line? the adjoint field visualization?

yaugenst-flex: yeah exactly, that and the global toggle to trigger that code

PR #1924, opened 2024-08-27
:wrench: Fix 404 mamba install link

daquinteroflex:


alec-flexcompute (approved):
Looks good, hard to think of any problems that can arise from this change ;)

PR #1925, opened 2024-08-28
Bump to xarray>=2023.08.0

yaugenst-flex:
Need to bump xarray because we depend on `xr.DataArray.coords.copy()` quite a bit, which is not implemented for `xarray<2023.08.0`. A user encountered this error because they were using an older xarray version.

This technically is a bug that exists across most previous releases, not sure how this would be handled (i.e., where is this supposed to get merged into?).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2708,8 +2708,8 @@ files = [
 numpy = [
     {version = ">=1.26.0", markers = "python_version >= \"3.12\""},
     {version = ">=1.23.3", markers = "python_version >= \"3.11\" and python_version < \"3.12\""},
-    {version = ">1.20", markers = "python_version < \"3.10\""},
     {version = ">=1.21.2", markers = "python_version >= \"3.10\" and python_version < \"3.11\""},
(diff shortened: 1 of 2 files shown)


weiliangjin2021 (approved):

daquinteroflex:
Good catch. We could technically add a little FAQ on the docs if we believe it to be useful. One other thing I personally find useful, is that people running into a similar error as that user could simply search on google the error first. If we have a traceback on an github issue/PR such as this one, then they should be able to know how to fix it. Re-patching a release wouldn't be feasible, but we can just make sure our latest version is correct onwards so just ask them to update to that. 

This is just what comes to mind, maybe there's a better approach you might have in mind?

yaugenst-flex:
Sounds good to me. I guess the way to handle this through releases is via maintaining older branches and doing maintenance/bugfix releases on those. Eventually we might want to do that, but I agree it's not feasible currently.

momchil-flex (approved):
Yeah just bumping the version if it's problematic seems acceptable to me. We've done it before for other deps too...

daquinteroflex (approved):

PR #1928, opened 2024-08-29
mode solver preconditioner improvement and turning off incidence matrix

weiliangjin2021:
- Disable incidence matrix
- Enable left and right preconditioner to make sure in `Ax= lambda M x`, only eps^-1 and mu^-1 appear in `A` and `M`.

### DIFF
--- a/tidy3d/plugins/mode/solver.py
+++ b/tidy3d/plugins/mode/solver.py
@@ -496,10 +496,6 @@ class EigSolver(Tidy3dBaseModel):
         mat_dtype = cls.matrix_data_type(eps, mu, der_mats, mat_precision, is_tensorial=False)
         mat = cls.type_conversion(mat, mat_dtype)
 
-        # Trim small values in single precision case
-        if mat_precision == "single":
-            cls.trim_small_values(mat, tol=fp_eps)
-
         # Casting starting vector to target data type
         vec_init = cls.type_conversion(vec_init, mat_dtype)
 
@@ -548,6 +544,10 @@ class EigSolver(Tidy3dBaseModel):
             if precon_left is not None:
                 generalized_M = precon_left @ generalized_M
                 mat = precon_left @ mat
+
+        # Trim small values in single precision case
+        if mat_precision == "single":
+            cls.trim_small_values(mat, tol=fp_eps)
 
         if analyze_conditioning:
             aca = mat.conjugate().T * mat
--- a/tidy3d/plugins/mode/solver.py
+++ b/tidy3d/plugins/mode/solver.py
@@ -22,6 +22,8 @@ TOL_EIGS = fp_eps
 TOL_TENSORIAL = 1e-6
 # shift target neff by this value, both rel and abs, whichever results in larger shift
 TARGET_SHIFT = 10 * fp_eps
+# Preconditioner: "Jacobi" or "Material" based
+PRECONDITIONER = "Material"
 
 
 class EigSolver(Tidy3dBaseModel):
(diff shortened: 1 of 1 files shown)


tidy3d/plugins/mode/solver.py
```diff
@@ -508,10 +505,30 @@ def any_pec(eps_vec, threshold=0.9 * np.abs(pec_val)):
...
         if enable_preconditioner:
-            precon = sp.diags(1 / mat.diagonal())
-            mat = mat * precon
+
+            def conditional_inverted_vec(eps_vec, threshold=1):
+                """Returns a diagonal sparse matrix whose i-th element in the diagonal
+                is |eps_i|^-1 if |eps_i|>threshold, and |eps_i| otherwise.
+                """
+                abs_vec = np.abs(eps_vec)
+                return sp.spdiags(np.where(abs_vec > threshold, 1.0 / abs_vec, abs_vec), [0], N, N)
```
momchil-flex: Very clever fix, hope it works! I took a look at your notes and it makes sense, but I do not understand the `conditional_inverted_vec` function here. You take 1 / eps if eps > 1 otherwise you take eps? How does that make sense?

weiliangjin2021: This is to make sure to scale down epsilon when it’s huge, but we don’t want to make the matrix ill conditioned if epsilon is close to 0

momchil-flex: But an epsilon of 2 and of 1/2 will return the same value in this function? Isn't this weird?

momchil-flex: And more importantly, unphysical?

weiliangjin2021: It doesn’t change the eigenvalue problem, as it’s just preconditioner.

momchil-flex: Ah ok that's the part I missed!

weiliangjin2021: Yeah, that’s in the last section of the notes

caseyflex (approved):
The improved preconditioning makes sense to me. Very nice fix!

tidy3d/plugins/mode/solver.py
```diff
@@ -22,6 +22,8 @@
 TOL_TENSORIAL = 1e-6
 # shift target neff by this value, both rel and abs, whichever results in larger shift
 TARGET_SHIFT = 10 * fp_eps
+# Preconditioner: "Jacobi" or "Material" based
+PRECONDITIONER = "Material"
```
weiliangjin2021, line 26: Now I also include the previous preconditioning method (Jacobi) along with the new one. For non-magnetic material (mu=1 everywhere), the two are actually very similar. The new one performs better when both epsilon and mu can be huge.

tidy3d/plugins/mode/solver.py
```diff
@@ -507,18 +506,59 @@ def any_pec(eps_vec, threshold=0.9 * np.abs(pec_val)):
...
             aac = mat * mat.conjugate().T
             diff = aca - aac
-            print(spl.norm(diff, ord=np.inf), spl.norm(aca, ord=np.inf), spl.norm(aac, ord=np.inf))
-            print(spl.norm(diff, ord="fro"), spl.norm(aca, ord="fro"), spl.norm(aac, ord="fro"))
+            print(
+                f"inf-norm: A*A: {spl.norm(aca, ord=np.inf)}, AA*: {spl.norm(aac, ord=np.inf)}, nonnormality: {spl.norm(diff, ord=np.inf)}, relative nonnormality: {spl.norm(diff, ord=np.inf)/spl.norm(aca, ord=np.inf)}"
+            )
+            print(
+                f"fro-norm: A*A: {spl.norm(aca, ord='fro')}, AA*: {spl.norm(aac, ord='fro')}, nonnormality: {spl.norm(diff, ord='fro')}, relative nonnormality: {spl.norm(diff, ord='fro')/spl.norm(aca, ord='fro')}"
+            )
```
weiliangjin2021, line 561: The accuracy of eigenvalue problem is usually measured by how far the matrix is away from normal matrix (normal matrix has conditioning number =1 in solving eigenvalues). But I'm not sure which metric is a good measurement. Probably the relative one makes more sense? @caseyflex

caseyflex: Yeah I agree that relative norm makes more sense here.

PR #1930, opened 2024-08-29
minor fix bloch validator

tylerflex:
This was just something I noticed. The old code led to false positive warnings if the test_val was just a little bit less than an integer.. this is more robust.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -2323,6 +2323,46 @@ class Simulation(AbstractYeeGridSimulation):
                     )
         return val
 
+    @pydantic.validator("monitors", always=True)
+    @skip_if_fields_missing(["boundary_spec", "medium", "size", "structures", "sources"])
+    def bloch_boundaries_diff_mnt(cls, val, values):
+        """Error if there are diffraction monitors incompatible with boundary conditions."""
+
+        monitors = val
+
+        if not val or not any(isinstance(mnt, DiffractionMonitor) for mnt in monitors):
+            return val
+
+        boundaries = values.get("boundary_spec").to_list
+        sources = values.get("sources")
+        size = values.get("size")
+        sim_medium = values.get("medium")
+        structures = values.get("structures")
+        for source_ind, source in enumerate(sources):
+            if not isinstance(source, PlaneWave):
+                continue
+
+            _, tan_dirs = cls.pop_axis([0, 1, 2], axis=source.injection_axis)
+            medium_set = Scene.intersecting_media(source, structures)
+            medium = medium_set.pop() if medium_set else sim_medium
+
+            for tan_dir in tan_dirs:
+                boundary = boundaries[tan_dir]
+
+                # check the Bloch boundary + angled plane wave case
+                num_bloch = sum(isinstance(bnd, (Periodic, BlochBoundary)) for bnd in boundary)
+                if num_bloch > 0:
+                    cls._check_bloch_vec(
+                        source=source,
+                        source_ind=source_ind,
+                        bloch_vec=boundary[0].bloch_vec,
+                        dim=tan_dir,
+                        medium=medium,
+                        domain_size=size[tan_dir],
+                        has_diff_mnt=True,
+                    )
+        return val
+
     @pydantic.validator("boundary_spec", always=True)
     @skip_if_fields_missing(["medium", "center", "size", "structures", "sources"])
     def tfsf_boundaries(cls, val, values):
@@ -3774,6 +3814,7 @@ class Simulation(AbstractYeeGridSimulation):
         dim: Axis,
         medium: MediumType,
         domain_size: float,
+        has_diff_mnt: bool = False,
     ):
         """Helper to check if a given Bloch vector is consistent with a given source."""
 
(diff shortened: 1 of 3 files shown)


tidy3d/components/simulation.py
```diff
@@ -3777,23 +3777,26 @@ def _check_bloch_vec(
         if bloch_vec != expected_bloch_vec:
             test_val = np.real(expected_bloch_vec - bloch_vec)
 
-            if np.isclose(test_val % 1, 0) and not np.isclose(test_val, 0):
+            test_val_is_int = np.isclose(test_val, np.round(test_val))
+            src_name = f" '{source.name}'" if source.name else ""
+
+            if test_val_is_int and not np.isclose(test_val, 0):
                 # the given Bloch vector is offset by an integer
                 log.warning(
```
tylerflex, line 3835: can we just turn off this validator maybe? it is pretty annoying if you are setting things up properly

tidy3d/components/simulation.py
```diff
@@ -3777,23 +3777,26 @@ def _check_bloch_vec(
...
+            #         "of the source. Consider using 'BlochBoundary.from_source()'.",
+            #         custom_loc=["boundary_spec", "xyz"[dim]],
+            #     )
+            if not test_val_is_int:
                 # the given Bloch vector is neither equal to the expected value, nor
                 # off by an integer
                 log.warning(
                     f"The Bloch vector along dimension '{dim}' may be incorrectly "
-                    f"set with respect to the source '{source.name}'. The absolute "
+                    f"set with respect to the source{src_name}. The absolute "
```
momchil-flex, line 3850: Missing space

tidy3d/components/simulation.py
```diff
@@ -3777,23 +3777,26 @@ def _check_bloch_vec(
...
+            #     # the given Bloch vector is offset by an integer
+            #     log.warning(
+            #         f"The wave vector of source{src_name} along dimension "
+            #         f"'{dim}' is equal to the Bloch vector of the simulation "
+            #         "boundaries along that dimension plus an integer reciprocal "
+            #         "lattice vector. If using a 'DiffractionMonitor', diffraction "
+            #         "order 0 will not correspond to the angle of propagation "
+            #         "of the source. Consider using 'BlochBoundary.from_source()'.",
+            #         custom_loc=["boundary_spec", "xyz"[dim]],
+            #     )
```
momchil-flex: The way I read this, this was specifically added to avoid confusion when using such a source + a DiffractionMonitor. How about we only warn in that case then, i.e. check if a DiffractionMonitor is present rather than say "if using a diffraction monitor". This should fix your adjoint-specific annoyance with these warnings.

I remember at least one user was confused about this at the time.

tidy3d/components/simulation.py
tylerflex: The space is actually contained in `src_name`. The reason I made this variable was because otherwise the error message contains "source None" unless the user specifies a source name. So the space was added to handle both cases properly. 

```py
src_name = f" '{source.name}'" if source.name else ""
```

tidy3d/components/simulation.py
tylerflex: So I tried doing this, but the way our validators are set up in simulation, these don't have access to the `monitors` unfortunately. If I try to change anything, it breaks a ton of tests.. not sure if it's better to try to just rearrange things in simulation or simply remove the warning

tylerflex: I came up with a solution that works, just added a separate validator for "monitors" and configured things to only check this condition if DiffractionMonitors are present there. seems to work properly.

docs/notebooks
tylerflex, line 1: develop submodule is out of date..

momchil-flex:
This still doesn't look quite right though. The warning about the source k not being an integer multiple of the Bloch vector should always be there, even without diffraction monitors. This is because this setup seems unphysical (the source does not match the boundary conditions) and is extremely unlikely to be correct. I'm not sure why we left it as warning and not a straight out error but either way it should be there even without DiffractionMonitor.

tylerflex:
> This still doesn't look quite right though. The warning about the source k not being an integer multiple of the Bloch vector should always be there, even without diffraction monitors. This is because this setup seems unphysical (the source does not match the boundary conditions) and is extremely unlikely to be correct. I'm not sure why we left it as warning and not a straight out error but either way it should be there even without DiffractionMonitor.

Unless I'm mistaken, this is how I have it set up. There are two validators, one checks that k is an integer multiple of the Bloch vector and is always triggered. The other checks that the Bloch vector is exactly equal to k, and is only checked now if a diffraction monitor is present. Could you take another look and verify whether it's consistent?

tidy3d/components/simulation.py
```diff
@@ -2323,6 +2323,46 @@ def plane_wave_boundaries(cls, val, values):
                     )
         return val
 
+    @pydantic.validator("monitors", always=True)
+    @skip_if_fields_missing(["boundary_spec", "medium", "size", "structures", "sources"])
+    def bloch_boundaries_diff_mnt(cls, val, values):
+        """Error if there are plane wave sources incompatible with boundary conditions."""
```
momchil-flex: This docstring

momchil-flex (approved):
Ah sorry my bad. Then it's just the docstring that needs to be changed.

PR #1934, opened 2024-08-30
move autograd post norm amps to simulation

tylerflex:
This simplifies the autograd internals and fixes the caching issue.

There's a corresponding backend PR too but I think you might want to take a closer look at that one and do some testing in case I missed something.


momchil-flex:
Closed in favor of #1994

PR #1936, opened 2024-08-30
improving beam_waist documentation

### LINKED ISSUE #1196: Clarify `waist_difference` better in docstring.
the `td.GaussianBeam.waist_difference` [docstring](https://docs.flexcompute.com/projects/tidy3d/en/latest/_autosummary/tidy3d.GaussianBeam.html
) descriptions says the following 

"
Distance from the beam waist along the propagation direction. When direction is + and waist_distance is positive, the waist is on the - side (behind) the source plane. When direction is + and waist_distance``is negative, the waist is on the ``+ side (in front) of the source plane.
"

It wasn't really clear from this how the behavior should look for `direction="-"`, eg in a downward propagating beam incident on a grating coupler. If we want the beam to converge in the propagation direction (so on the "+" side of src for direction="+" and on the "-" side for direction="-"), it seems we always want a negative beam distance? I feel we should clarify this better in the docstring. 

Is my understanding correct and how should we rephrase this?

Also note, there's some misformatting of the quotes in the description as well.

FilipeFcp:
Fixes #1196

It looks like this:

image

@momchil-flex I tested, and for a rotated source the distance is defined in the propagation axis.


daquinteroflex (approved):
Obrigado Filipe! Looks great and well implemented.

PR #1946, opened 2024-09-04
Autograd FAQs

### LINKED ISSUE #1940: FAQ section for autograd readme
Our autograd interface currently has some sharp edges that we need to document, as well as some general things that tend to crop up when it comes to autograd.

Things to address:
- "sharp bits"
- how to use with jax and PyTorch
- other faqs

yaugenst-flex:
Closes #1940


yaugenst-flex:
Will be included in #2025

PR #1948, opened 2024-09-04
fix bug in cylinder VJP related to angles

tylerflex:
was a pretty dumb mistake. miraculously the numerical tests and the cylinder demos worked before I did this fix.. only caught it this morning when trying out something slightly different.. goes to show how tricky this stuff can get.

Also catches an edge case in the PolySlab VJP that would error in some 2D simulations

### DIFF
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -79,7 +79,7 @@ DA_SHAPE_X = 1 if IS_3D else 1
 DA_SHAPE = (DA_SHAPE_X, 1_000, 1_000) if TEST_CUSTOM_MEDIUM_SPEED else (DA_SHAPE_X, 12, 12)
 
 # number of vertices in the polyslab
-NUM_VERTICES = 100_000 if TEST_POLYSLAB_SPEED else 30
+NUM_VERTICES = 100_000 if TEST_POLYSLAB_SPEED else 110
 
 PNT_DIPOLE = td.PointDipole(
     center=(0, 0, -LZ / 2 + WVL),
@@ -394,15 +394,15 @@ def make_structures(params: anp.ndarray) -> dict[str, td.Structure]:
     custom_pole_res = td.Structure(geometry=box, medium=custom_med_pole_res)
 
     radius = 0.4 * (1 + anp.abs(vector @ params))
-    cyl_center_y = vector @ params
-    cyl_center_z = -vector @ params
+    cyl_center_y = 0  # vector @ params
+    cyl_center_z = 0  # -vector @ params
     cylinder_geo = td.Cylinder(
-        radius=radius,
+        radius=radii,
         center=(0, cyl_center_y, cyl_center_z),
         axis=0,
         length=LX / 2 if IS_3D else td.inf,
     )
-    cylinder = td.Structure(geometry=cylinder_geo, medium=med)
+    cylinder = td.Structure(geometry=cylinder_geo, medium=polyslab.medium)
 
     return dict(
         medium=medium,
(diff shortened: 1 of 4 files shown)


tylerflex:
@momchil-flex just fyi. we'll need to merge this before 2.7.3. shouldn't affect notebooks since cylinder is only used in unmerged https://github.com/flexcompute/tidy3d-notebooks/pull/166

tests/test_components/test_autograd.py
yaugenst-flex, line 1: More of a general comment/question: Is there any way to test these VJPs without field data? E.g. for flux we can just have an objective that creates random field data and differentiate through that to assert correctness of the grads of that computation, without having to check the full chain (going through simulation & run etc.).

If we could do something similar here, I think the tests could be 1. much faster and 2. catch these kinds of mistakes (since `check_grads` will just fail if the VJP is implemented incorrectly). I think we'd just need to define a `primitive` in the test that calls `compute_derivatives` in the backward, and then we can test that w.r.t. changes in the geometry, without having to go through fields and simulation?

tidy3d/components/geometry/polyslab.py
```diff
@@ -1516,7 +1516,12 @@ def pop_axis_vect(self, coord: np.ndarray) -> Tuple[np.ndarray, Tuple[np.ndarray
     @staticmethod
     def normalize_vect(arr: np.ndarray) -> np.ndarray:
         """normalize an array shaped (N, d) along the `d` axis and return (N, 1)."""
-        return arr / np.linalg.norm(arr, axis=-1)[..., None]
+        norm = np.linalg.norm(arr, axis=-1)
+
+        if np.all(norm > 0):
+            return arr / np.linalg.norm(arr, axis=-1)[..., None]
+
+        return arr
```
yaugenst-flex: How about:
```python
norm = np.linalg.norm(arr, axis=-1, keepdims=True)
norm = np.where(norm == 0, 1, norm)
return arr / norm
```

tidy3d/components/geometry/primitives.py
```diff
@@ -270,7 +270,7 @@ def _points_unit_circle(
         self, num_pts_circumference: int = _N_PTS_CYLINDER_POLYSLAB
     ) -> np.ndarray:
         """Set of x and y points for the unit circle when discretizing cylinder as a polyslab."""
-        angles = np.linspace(0, 1, num_pts_circumference + 1)[:-1]
+        angles = 2 * np.pi * np.linspace(0, 1, num_pts_circumference + 1)[:-1]
```
yaugenst-flex: :smile:

Maybe `angles = np.linspace(0, 2 * np.pi, num_pts_circumference, endpoint=False)` for clarity? No `+ 1)[:-1]` needed.

yaugenst-flex (approved):

tests/test_components/test_autograd.py
tylerflex: I thought about this. In principle we could, but the real challenge is that the gradient that we want to compute is completely encoded in the fwd and adjoint E fields and we use Maxwells equations to derive it. So I dont know if there's some way to validate whether the gradient is correct using just random FieldData. If we could, it would be super handy though.

In this case, the bug was actually in `to_polyslab()` since it was only creating 1/2pi of the correct polylsab.. so the test that ended up finding the issue was comparing the PolySlab to Cylinder gradients with emulated data.  So perhaps omparing the gradients like this for equivalent structures is one way to test consistency but we still can't know the "true" gradients until we do a numerical test.

tidy3d/components/geometry/primitives.py
tylerflex: wow, took me too long to find out about `endpoint=False`.. thanks.

tidy3d/components/geometry/polyslab.py
tylerflex: hm yea looks good. nice numpy skills :D

tylerflex:
Thanks @yaugenst-flex just made both changes

yaugenst-flex (approved):
Looks great :tada:

tylerflex:
sorry, I misread the reviewers icon and thought I needed to re-request a review :P

PR #1950, opened 2024-09-05
server side inverse design web API (WIP)

### LINKED ISSUE #1942: Web API for `invdes` plugin

tylerflex:
WIP


momchil-flex:
We really should get this in sooner rather than later. I mean a call to the inverse design pipeline, not sure if it will be within this PR.

PR #1953, opened 2024-09-09
fixed bug that replaced all lumped elements in simulations, when using TerminalComponentModeler

dmarek-flex:
Just had to copy any existing lumped elements before adding loads from the ports.


tylerflex (approved):
Ah yea, this can be a tricky one. I often wonder if some kind of `append: bool` argument to `updated_copy()` might be useful for some things. For example, it would be a shortcut to doing `list(self.field) + list(new_things)`. maybe some other time.

QimingFlex (approved):

PR #1954, opened 2024-09-11
Fixed angle simulations

### LINKED ISSUE #1386: Broadband fixed angle source (BFAST) support
The current Bloch boundary causes a wavelength-dependent incident angle, which is not ideal in broadband simulations with oblique incidence. Users usually want to fix the incident angle as it's commonly the case in experiments. Do we have any plans to implement BFAST in the future?

dbochkov-flexcompute:
Summary of changes:
- introduced field `angular_spec` in `PlaneWave`, which can take values `FixedAngleSpec()` and `FixedInPlaneKSpec()`. Those are currently empty specs, but in future we could add parameters to them. 
- automatic time step reduction for `FixedAngleSpec()` due to a new stability criterion
- minor change in diffraction monitor data, so that results can be processed at multiple bloch vectors simultaneously
- btw, do you prefer `FixedAngleSpec()`/`FixedInPlaneKSpec()` or `ConstAngleSpec()`/`ConstInPlaneKSpec()`

example notebook https://github.com/flexcompute/tidy3d-notebooks/blob/daniil/fixed-angle/AngularFrequencyDependence.ipynb

### DIFF
--- a/tests/test_components/test_simulation.py
+++ b/tests/test_components/test_simulation.py
@@ -814,6 +814,7 @@ def test_nyquist():
         frequency_range = (-2, -1)
         monitors = ()
         _cached_properties = {}
+        _fixed_angle_sources = ()
 
     m = MockSim()
     assert td.Simulation.nyquist_step.fget(m) == 1
@@ -3027,3 +3028,96 @@ def test_validate_sources_monitors_in_bounds():
             grid_spec=td.GridSpec(wavelength=1.0),
             monitors=[mode_monitor],
         )
+
+
+def test_fixed_angle_sim():
+    wvl_um = 1.0
+    freq0 = td.C_0 / wvl_um
+    fwidth = freq0 / 5
+
+    freqs = freq0 + 0.5 * fwidth * np.linspace(-1, 1, 11)
+    med = td.Medium(permittivity=5)
+    sphere = td.Structure(
+        geometry=td.Sphere(radius=0.5),
+        medium=med,
+    )
+    flux_r_mnt = td.FluxMonitor(
+        center=(-1, 0, 0), size=(0, td.inf, td.inf), freqs=freqs, name="flux_r"
+    )
+    source = td.PlaneWave(
+        angle_phi=np.pi / 6,
+        angle_theta=np.pi / 5,
+        angular_spec=td.FixedAngleSpec(),
+        direction="+",
+        center=(-0.9, 0, 0),
+        size=(0, td.inf, td.inf),
+        pol_angle=np.pi / 4,
+        source_time=td.GaussianPulse(freq0=freq0, fwidth=fwidth),
+    )
+    sim_size = (2.2, 2.2, 2.2)
+    sim = td.Simulation(
+        structures=[sphere],
+        sources=[source],
+        monitors=[flux_r_mnt],
+        size=sim_size,
+        grid_spec=td.GridSpec.auto(min_steps_per_wvl=15),
+        boundary_spec=td.BoundarySpec(
+            x=td.Boundary.absorber(), y=td.Boundary.periodic(), z=td.Boundary.periodic()
+        ),
+        run_time=10 / fwidth,
+    )
+
+    assert sim._is_fixed_angle
+
+    with pytest.raises(pydantic.ValidationError):
+        _ = sim.updated_copy(
+            boundary_spec=td.BoundarySpec(
+                x=td.Boundary.pml(),
+                y=td.Boundary.bloch_from_source(source=source, axis=1, domain_size=2.2),
+                z=td.Boundary.bloch_from_source(source=source, axis=2, domain_size=2.2),
+            )
+        )
+
+    with pytest.raises(pydantic.ValidationError):
+        _ = sim.updated_copy(med=td.Medium(conductivity=0.001))
+
+    anisotropic_med = td.FullyAnisotropicMedium(permittivity=[[2, 0, 0], [0, 1, 0], [0, 0, 3]])
+    with pytest.raises(pydantic.ValidationError):
+        _ = sim.updated_copy(structures=[sphere.updated_copy(medium=anisotropic_med)])
+
+    with pytest.raises(pydantic.ValidationError):
+        _ = sim.updated_copy(sources=[source, source])
+
+    with pytest.raises(pydantic.ValidationError):
+        _ = sim.updated_copy(
+            structures=[sphere.updated_copy(medium=td.Medium(conductivity=-0.1, allow_gain=True))]
+        )
+
+    with pytest.raises(pydantic.ValidationError):
+        _ = sim.updated_copy(monitors=[td.FieldTimeMonitor(size=[td.inf, td.inf, 0], name="time")])
+
+    with pytest.raises(pydantic.ValidationError):
+        _ = sim.updated_copy(monitors=[td.FluxTimeMonitor(size=[td.inf, td.inf, 0], name="time")])
+
+    nonlinear_med = td.Medium(
+        permittivity=3,
+        nonlinear_spec=td.NonlinearSpec(
+            models=[
+                td.KerrNonlinearity(n2=-1 + 1j, n0=1),
+            ],
+            num_iters=20,
+        ),
+    )
+    with pytest.raises(pydantic.ValidationError):
+        _ = sim.updated_copy(structures=[sphere.updated_copy(medium=nonlinear_med)])
+
+    time_modulated_med = td.Medium(
+        permittivity=2,
+        modulation_spec=td.ModulationSpec(
+            permittivity=td.SpaceTimeModulation(
+                time_modulation=td.ContinuousWaveTimeModulation(freq0=td.C_0, amplitude=1, phase=0),
+            )
+        ),
+    )
+    with pytest.raises(pydantic.ValidationError):
+        _ = sim.updated_copy(structures=[sphere.updated_copy(medium=time_modulated_med)])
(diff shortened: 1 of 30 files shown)


tidy3d/components/simulation.py
```diff
@@ -2417,6 +2433,59 @@ def tfsf_with_symmetry(cls, val, values):
...
+    @staticmethod
+    def _get_fixed_angle_sources(sources: Tuple[SourceType, ...]):
+        """Get list of plane wave sources with ``FixedAngleSpec``."""
+
+        return [
+            source for source in sources if isinstance(source, PlaneWave) and source._is_fixed_angle
+        ]
+
+    @pydantic.root_validator(skip_on_failure=True)
+    def check_fixed_angle_components(cls, values):
```
dmarek-flex, line 2582: I assume `lumped_elements` and other 2D Medium will work properly in these simulations?

tidy3d/components/source.py
```diff
@@ -1120,6 +1135,18 @@ class PlaneWave(AngledFieldSource, PlanarSource):
         * `Using FDTD to Compute a Transmission Spectrum <https://www.flexcompute.com/fdtd101/Lecture-2-Using-FDTD-to-Compute-a-Transmission-Spectrum/>`__
     """
 
+    angular_spec: Union[FixedInPlaneKSpec, FixedAngleSpec] = pydantic.Field(
+        FixedInPlaneKSpec(),
+        title="Angular Dependence Specification",
+        description="Specification of plane wave propagation direction dependence on wavelength.",
```
dmarek-flex, line 491: Could you provide an explanation of when a user should use `FixedInPlaneKSpec` or `FixedAngleSpec`? I guess usually when Bloch boundary conditions are used the source must be `FixedInPlaneKSpec`?

dmarek-flex (approved):
Looks good to me, just a little clarification for beginners like me would be helpful!

tidy3d/components/source.py
dbochkov-flexcompute: added a short note about it. Also see new notebook https://github.com/flexcompute/tidy3d-notebooks/blob/daniil/fixed-angle/AngularFrequencyDependence.ipynb

tidy3d/components/simulation.py
dbochkov-flexcompute: yes, this shouldn't cause any issues

tests/test_components/test_simulation.py
```diff
@@ -3027,3 +3028,96 @@ def test_validate_sources_monitors_in_bounds():
...
+        monitors=[flux_r_mnt],
+        size=sim_size,
+        grid_spec=td.GridSpec.auto(min_steps_per_wvl=15),
+        boundary_spec=td.BoundarySpec(
+            x=td.Boundary.absorber(), y=td.Boundary.periodic(), z=td.Boundary.periodic()
+        ),
+        run_time=10 / fwidth,
+    )
+
+    # assert sim.i
```
momchil-flex: remove

tidy3d/components/simulation.py
```diff
@@ -2420,16 +2424,28 @@ def plane_wave_boundaries(cls, val, values):
...
-                        bloch_vec=boundary[0].bloch_vec,
-                        dim=tan_dir,
-                        medium=medium,
-                        domain_size=size[tan_dir],
-                    )
+                if isinstance(source.angular_spec, FixedAngleSpec):
+                    num_bloch = sum(isinstance(bnd, BlochBoundary) for bnd in boundary)
+                    if num_bloch > 0:
+                        raise SetupError(
+                            "Fixed angle plane wave sources ('FixedAngleSpec' and `angle_theta != 0`) do "
```
momchil-flex: I think we normally don't use markdown in error messages even in cases like this, even if by now I'm so used to markdown I kind of like it. Still, maybe `'angle_theta' != 0`?

tidy3d/components/simulation.py
```diff
@@ -2420,16 +2424,28 @@ def plane_wave_boundaries(cls, val, values):
...
+                            f"not require the Bloch boundary along dimension {tan_dir}. "
+                            "Either set the boundary conditions to 'Periodic' to proceed to simulate a plane "
+                            "wave with frequency-independent propagation direction, or switch to "
+                            "'FixedInPlaneKSpec' specification to simulate a plane wave with "
+                            "frequency-dependent propagation direction."
```
momchil-flex: "with a fixed in-plane Bloch vector (frequency-dependent propagation direction)."

tidy3d/components/simulation.py
```diff
@@ -3860,6 +3938,16 @@ def background_structure(self) -> Structure:
...
         return self.scene.background_structure
 
+    @cached_property
+    def _fixed_angle_sources(self):
+        """List of plane wave sources with ``FixedAngleSpec``."""
+        return self._get_fixed_angle_sources(self.sources)
+
+    @cached_property
+    def _is_fixed_angle(self):
+        """List of plane wave sources with ``FixedAngleSpec``."""
```
momchil-flex: """Whether the simulation contains fixed angle sources."""

tidy3d/components/source.py
```diff
@@ -1095,6 +1095,23 @@ def _bend_axis(self) -> Axis:
...
+    """Plane wave is injected such that its in-plane wavevector is constant. That is,
+    the injected field satisfies Bloch boundary conditions and its propagation direction is
+    frequency dependent.
+    """
+
+
+class FixedAngleSpec(AbstractAngularSpec):
+    """Plane wave is injected such that its propagation direction is frequency independent.
+    When using this option boundary conditions in tangential directions must be set to periodic.
+    """
```
momchil-flex, line 466: I remember that the reason we decided to go with Spec-s instead of just something like `fixed_angle=False`? was that these can contain some settings, right? It seems like for now this was not needed but do you think it's something we should add in the future, like the "alpha" parameter in fixed angle?

tidy3d/components/simulation.py
```diff
@@ -4540,8 +4641,13 @@ def nyquist_step(self) -> int:
...
         freq_max = max(freq_source_max, freq_monitor_max)
 
+        # in fixed angle simulations both E and H are available at full and half steps
+        fixed_angle_factor = 1
+        if len(self._fixed_angle_sources) > 0:
+            fixed_angle_factor = 2
+
         if freq_max > 0:
-            nyquist_step = int(1 / (2 * freq_max) / self.dt) - 1
+            nyquist_step = int(1 / (2 * freq_max) / self.dt * fixed_angle_factor) - 1
```
momchil-flex, line 4648: As usual you've thought of everything :) Did this manifest as an issue that you had to debug to find out or you just thought about it in the process of developing?

momchil-flex (approved):

tidy3d/components/source.py
dbochkov-flexcompute: yes, there could be some additional parameters eventually after we learn more about performance of this in various situations, but not at the present time. I wonder, if it would be better to follow the field `subpixel` journey. Have it initially as bool (`fixed_angle=True`), and later generalize to a spec if needed (`fixed_angle=FixedAngleSpec(...)`). I don't think we will need to add any parameters for regular (bloch) plane waves

tidy3d/components/simulation.py
dbochkov-flexcompute: it was naturally following from backend changes

tidy3d/components/source.py
tylerflex, line 1: might be a good time to split `source` into several modules (like `geometry`) if it isn't too much of a pain.

tidy3d/components/simulation.py
```diff
@@ -3860,6 +3938,16 @@ def background_structure(self) -> Structure:
         )
         return self.scene.background_structure
 
+    @cached_property
+    def _fixed_angle_sources(self):
+        """List of plane wave sources with ``FixedAngleSpec``."""
+        return self._get_fixed_angle_sources(self.sources)
+
+    @cached_property
+    def _is_fixed_angle(self):
```
tylerflex: heads up: a few cached properties in this PR are missing return types

tidy3d/components/simulation.py
```diff
@@ -2556,6 +2572,57 @@ def tfsf_with_symmetry(cls, val, values):
...
 
+    @staticmethod
+    def _get_fixed_angle_sources(sources: Tuple[SourceType, ...]):
+        """Get list of plane wave sources with ``FixedAngleSpec``."""
+
+        return [
+            source for source in sources if isinstance(source, PlaneWave) and source._is_fixed_angle
+        ]
+
+    @pydantic.root_validator(skip_on_failure=True)
```
tylerflex: do you want to add `skip_if_missing` here? (might have the name mixed up, but validator skipping if fields (like `sources`) fail?

tylerflex (approved):
Looks good to me. thanks @dbochkov-flexcompute ! Just a few minor comments but other than that, good to merge.

tidy3d/components/source.py
momchil-flex: So from a developer perspective, my preference would be  to have it as it is now, i.e. I'd rather not mix types. For subpixel it ended up like that for historic and compatibility reasons. From a user perspective I can see how it's slightly more convenient to be able to just say `True`, so I'm also not against that, but all things considered I'm fine the way it is esp. given that we may add arguments to the fixed angle spec (will make users more aware of them).

tidy3d/components/source.py
dbochkov-flexcompute: I am thinking of splitting `source.py` into:
- `source/time.py`: for time dependencies
- `source/current.py`: for current based sources
- `source/field.py`: for field based sources

how does that sound?

tylerflex: sounds good to me, thanks

Issue #1949, opened 2024-09-05
Claryfing units for TFSF source.


FilipeFcp:
The documentation of the TFSF says that the source injects $1W/um^2$ of power, but the units are in Intensity, not power, right? So seems that the intensity is constant, but the Power varies with the source size, am I correct?

I had to get the $|E|^2$ and got a bit confused if I should multiply by the source area or not. Maybe would be interesting to clarify that according to the Poyting theorem. $I = (1/2)c n \epsilon_0 |E_0|^2 = 1$.

momchil-flex:
> The documentation of the TFSF says that the source injects 1 W / u m 2 of power, but the units are in Intensity, not power, right? So seems that the intensity is constant, but the Power varies with the source size, am I correct?

Correct. I guess the proper wording would be something like the source injects 1W of power per square micron of  the source injection area, so the power injected by the source in W is given by the area of the injection plane in units of um**2. Or something like that. You want to have a go at rephrasing it?

FilipeFcp:
I think you are right. I just tested it and this is what I got for a plane wave:

PR #1957, opened 2024-09-13
small change in the documentation

FilipeFcp:
Hi @momchil-flex . Please let me know what you think when you have some time:

image

#1949

### DIFF
--- a/tidy3d/components/source.py
+++ b/tidy3d/components/source.py
@@ -1207,10 +1207,11 @@ class TFSF(AngledFieldSource, VolumeSource):
     Notes
     -----
 
-        The TFSF source injects :math:`\\frac{1 W}{\\mu m^2}` of power along the :attr:`injection_axis`. Note that in the
-        case of angled incidence, :math:`\\frac{1 W}{\\mu m^2}` is still injected along the source's :attr:`injection_axis`,
-        and not the propagation direction, unlike a :class:`PlaneWave` source. This allows computing
-        scattering and absorption cross-sections without the need for additional normalization.
+        The TFSF source injects :math:`1 W` of power per :math:`\\mu m^2` of source area along the :attr:`injection_axis`.
+        Hence, the normalization for the incident field is :math:`|E_0|^2 = \\frac{2}{c\\epsilon_0}`, for any source size.
+        Note that in the case of angled incidence, the same power is injected along the source's :attr:`injection_axis`,
+        and not the propagation direction. This allows computing scattering and absorption cross-sections
+        without the need for additional normalization.
 
         The TFSF source allows specifying a box region into which a plane wave is injected. Fields inside this region
         can be interpreted as the superposition of the incident field and the scattered field due to any scatterers


momchil-flex (approved):
Sounds great! Maybe also update the FAQ section a little?

momchil-flex:
Actually, are you sure about the distinction from PlaneWave? Now that I think about it I'm a little confused about this, because the auxiliary simulation to the TFSF one just uses a PlaneWave. I thought that the difference is only in the area factor.

FilipeFcp:
Hi Momchil. Yes, I understand that the difference is the area factor. A PlaneWave (infinity with periodic boundary) will inject 1 W of power. A TFSF will inject 1 $\frac{W}{\mu^2}$ intensity, or 1/sourceArea power.

Regarding the [FAQ](https://docs.flexcompute.com/projects/tidy3d/en/latest/faq/docs/faq/How-are-results-normalized.html), it is describing a `num_freq` attribute for the `SourceTime` object, that seems deprecated. Is the normalization still an issue for frequencies distant from the central one? Can I remove the reference for the `num_freq` attribute?

momchil-flex:
> Hi Momchil. Yes, I understand that the difference is the area factor. A PlaneWave (infinity with periodic boundary) will inject 1 W of power. A TFSF will inject 1 W μ 2 intensity, or 1/sourceArea power.

What I was questioning was the statement about angled incidence. Is that really true? I thought that the normalization should not depend on the angle and the relationship between TFSF and PlaneWave should be the same at all angles.

> Regarding the [FAQ](https://docs.flexcompute.com/projects/tidy3d/en/latest/faq/docs/faq/How-are-results-normalized.html), it is describing a `num_freq` attribute for the `SourceTime` object, that seems deprecated. Is the normalization still an issue for frequencies distant from the central one? Can I remove the reference for the `num_freq` attribute?

I think this should be just `Source.num_freqs` instead of `SourceTime.num_freqs`. This argument is only available in ModeSource and Gaussian-type sources.

FilipeFcp:
I think you are right. I just tested it and this is what I got for a plane wave:

deg flux
10 1.00007653,
20 1.00046873,
30 1.00205553
40 1.02323759
50 1.07655406

I believe the error for 40 and 50 is due to the fields taking too long to decay.

So I will change this part as well, and open another PR for the FAQ.

momchil-flex:
Ah yeah I see, yeah PML are not great at absorbing oblique angles, you could try changing the normal direction boundaries to Absorber and put like 100 layers, might look better.

FilipeFcp:
Hi @momchil-flex,

I ran some tests to fully understand this.

Yes, both the TFSF and Plane wave sources keep their quantities constant in the propagation direction (TFSF intensity and PW power). That information might be outdated, so I believe it's safe to remove it from the documentation and FAQ?

image



By the way, the absorber with many layers worked well for large incidence angles.

momchil-flex:
Not sure what you mean? To me that information sounds correct still. The power is constant for a plane that is normal to the source injection axis (the axis normal to the source geometry) right? Meaning that the power in the propagation direction (the angled axis) would have an extra factor of 1 / cos(theta)?

FilipeFcp:
Yes, this is correct. What I meant is that the same applies to both TFSF and PW. The TFSF docstring implies it would be different for a PW. I thought the FAQ had the same issue, but it's actually correct there.

momchil-flex:
@FilipeFcp sorry is this ready for merge then?

FilipeFcp:
Hi @momchil-flex. I believe it is.

PR #1958, opened 2024-09-13
Error if projection monitor touches periodic/bloch boundaries in 3D simulations

QimingFlex:
This validator raises an error in 3D simulations when a projection monitor touches periodic/Bloch boundaries. For 2D simulations, we allow periodic boundaries, which won't trigger this error. For 2D with Bloch boundaries, we already error it out in another PR where we just don't allow Bloch boundaries along the 0-dimension.


tidy3d/components/simulation.py
```diff
@@ -2494,6 +2494,47 @@ def boundaries_for_zero_dims(cls, val, values):
 
         return val
 
+    @pydantic.validator("monitors", always=True)
+    @skip_if_fields_missing(["size", "boundary_spec"])
```
momchil-flex, line 2521: I guess also "center"?

momchil-flex (commented):
Now that I think about this, I think field projection monitors will always be wrong if there are periodic or Bloch boundaries, regardless of whether they touch the boundaries or not. Either way, to do a correct projection, a periodic Green's function must be used. So I think the error should just be whenever there's any field projection monitor and any periodic/bloch boundaries together - for a 3D simulation, I cannot come up with a case in which this will work well even if e.g. the monitor is along xy and the boundaries are along z, it's again weird. And the error could suggest to the user to look into DiffractionMonitor, which is probably what they want to use in most cases, unless they have periodicity on one axis and not the other - that's something we don't really support now.

QimingFlex:
> Now that I think about this, I think field projection monitors will always be wrong if there are periodic or Bloch boundaries, regardless of whether they touch the boundaries or not. Either way, to do a correct projection, a periodic Green's function must be used. So I think the error should just be whenever there's any field projection monitor and any periodic/bloch boundaries together - for a 3D simulation, I cannot come up with a case in which this will work well even if e.g. the monitor is along xy and the boundaries are along z, it's again weird. And the error could suggest to the user to look into DiffractionMonitor, which is probably what they want to use in most cases, unless they have periodicity on one axis and not the other - that's something we don't really support now.

Ah, I see. This makes sense. Previously, I thought projection with just one period was physically meaningful, but yeah it's not practically meaningful.

QimingFlex:
@momchil-flex I updated the code to throw an error when a FieldProjectionMonitor is used with periodic/bloch boundaries. But it appears [this](https://github.com/flexcompute/tidy3d/blob/0c5736fa5dd72e8c86b7c73cb74ffa6db05bd4dc/tests/utils.py#L400) sim object is used in many test functions and will throw a bunch of errors (some related to DiffractionMonitor) if I change the boundaries from periodic/bloch to PMLs. Could you give me some suggestions for the fix?

tidy3d/components/simulation.py
```diff
@@ -2517,6 +2517,36 @@ def boundaries_for_zero_dims(cls, val, values):
...
+            axis = axis_names[dim]
+            num_periodic_bloch_boundaries = sum(
+                isinstance(bnd, (Periodic, BlochBoundary)) for bnd in boundary
+            )
+            if num_periodic_bloch_boundaries > 0:
+                for monitor in val:
+                    if isinstance(monitor, AbstractFieldProjectionMonitor) and is_3d_simulation:
+                        raise SetupError(
+                            f"Using 'FieldProjectionMonitor' '{monitor.name}' "
+                            f"with periodic/Bloch boundaries along the {axis} axis would lead to incorrect results. "
```
momchil-flex, line 2543: I think no need to print the axis as using the projection monitor with any periodic/Bloch boundaries is wrong.

`Using 'FieldProjectionMonitor' '{monitor.name}'  with periodic/Bloch boundaries will lead to incorrect results.`

momchil-flex (commented):
Hm, yeah, I see the issue - not sure, will open it up for discussion.

momchil-flex:
Addressed in https://github.com/flexcompute/compute/pull/4259

PR #1959, opened 2024-09-15
Serializable `invdes` postprocessing

### LINKED ISSUE #1943: Integration of serializable expressions as post-processing option for `invdes`

yaugenst-flex:
Is this roughly what you had in mind @tylerflex?

This is branched off of #1848 
Closes #1943

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- Added `tidy3d.plugins.expressions` module for constructing and serializing mathematical expressions and simulation metrics like `ModeAmps` and `ModePower`.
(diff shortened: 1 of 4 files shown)


tylerflex:
Yea this is pretty much exactly what I had in mind.

yaugenst-flex:
ok nice I'll fix the tests and wrap this up then

tylerflex:
How would this work with InverseDesignMulti?

yaugenst-flex:
yeah I have to check that. I _think_ it might just work? will test

tylerflex:
Each simulation will need its own post process function. So i guess maybe a list of them is needed. But seems kind of clunky at first glance.

yaugenst-flex:
how does this currently work? doesn't the `optimizer.run()` take only a single fn?

tylerflex:
It does but the function takes the batch data as arguments (i think, it might also take the individual simulation data's as args too). Check the last couple cells of the main tutorial Invdes.ipynb (on mobile)

PR #1960, opened 2024-09-15
Make `BatchData` a mapping

### LINKED ISSUE #1955: Make `BatchData` more `dict`-like
[`BatchData`](https://docs.flexcompute.com/projects/tidy3d/en/v2.0.2/_autosummary/tidy3d.web.BatchData.html#tidy3d.web.BatchData) currently defines an `.items()` method with which one can iterate (`for k, v in batch_data.items()`). This suggests to the users that `BatchData` behaves like a dictionary (actually, the `.dict_from_xxx` classmethods suggest that too), but the `.keys()` method returns unexpected results (`dict_keys(['attrs', 'type', 'simulations', 'folder_name', ...])` instead of the simulation id's) and there is no `.values()` method.

yaugenst-flex:
Makes `BatchData` behave like a regular Python dict, can call `.items()`, `.values()`, `.keys()`, etc.

Closes #1955

### DIFF
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -6,6 +6,7 @@ import concurrent
 import os
 import time
 from abc import ABC
+from collections.abc import Mapping
 from concurrent.futures import ThreadPoolExecutor
 from typing import Dict, Optional, Tuple
 
@@ -339,7 +340,7 @@ class Job(WebContainer):
         return web.estimate_cost(self.task_id, verbose=verbose, solver_version=self.solver_version)
 
 
-class BatchData(Tidy3dBaseModel):
+class BatchData(Tidy3dBaseModel, Mapping):
     """
     Holds a collection of :class:`.SimulationData` returned by :class:`Batch`.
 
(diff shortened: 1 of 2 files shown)


tidy3d/web/api/container.py
```diff
@@ -391,16 +392,18 @@ def load_sim_data(self, task_name: str) -> SimulationDataType:
             verbose=False,
         )
 
-    def items(self) -> Tuple[TaskName, SimulationDataType]:
```
tylerflex, line 394: not sure if we want to remove this. The goal of `BatchData.items()` is to load each of the `sim_data` objects one by one from file without loading them all in memory. So I think it's possible we need to overwrite the `items()` and the `values()` accordingly?

tylerflex (changes requested):
Nice, this will be an improvement I think.

Let's discuss `items()` and `values()` from file?

Also could we add a changelog and maybe one test to verify that one or more of the new `Mapping` methods are working properly?

tidy3d/web/api/container.py
yaugenst-flex: but i think that‘s exactly what a mapping does? values and items are generators

tylerflex: hmm ok. This might require some testing. 

so my experience originally was that just having a `batch_data : dict[str, SimulationData]` in memory was problematic. So this BatchData basically hid this by making things `dict`-like but the iteration was done by loading the values into memory one by one. I'm not sure if just making the `values()` a generator fixes this? but I could be wrong. is there a good way to test?

yaugenst-flex: this exactly how those views in mappings are implemented: https://github.com/python/cpython/blob/ef530ce7c61dc16387fb68d48aebfbbbfe02adbb/Lib/_collections_abc.py#L897-L899

essentially, `__getitem__` and `__iter__` define everything that is needed for the implementation of `.values()`, `.keys()`, and `.items()`. since both of these methods are defined in terms of single simulations, and the derived methods are generators. loading simulation data from disk is triggered by `__getitem__` for that particular item. a test for this would essentially test whether generators in python work?

the `SimulationData` will only be in memory once you have iterated over those generators and stored the results, i.e. something like `list(batch_data.values()`.

tylerflex: ah ok! I missed the sublety here with the getitem loading the sim data. nice, implementation looks good to me then

tylerflex:
lets' merge into 2.8

yaugenst-flex:
sounds great, rebased and added changelog item

tylerflex (approved):

PR #1962, opened 2024-09-18
Create ParameterSpec

yaugenst-flex:
Adds `InitializationSpec` classes:
 - `UniformInitializationSpec(value=...)` -> uniform initialization with single value
 - `RandomInitializationSpec(min_value=..., max_value=..., seed=None)` -> random initialization between `min_value` and `max_value`, optional `seed`
 - `CustomInitializationSpec(params=...)`, where `params` is `ArrayLike` -> user-defined initial parameters

### Usage:

in `invdes.DesignRegion`:
```python
opt = DesignRegion(initialization_spec=UniformInitializationSpec())
```

more generally:
```python
spec = RandomInitializationSpec(min_value=0, max_value=1)
params = spec.create_parameters(shape=(x, y, ...))
```

The invdes optimizer calls `design_region.initial_parameters` to initialize the optimization.

@momchil-flex

### DIFF
--- a/tidy3d/plugins/invdes/initialization.py
+++ b/tidy3d/plugins/invdes/initialization.py
@@ -1,0 +1,86 @@
+# module providing classes for initializing the parameters in an inverse design problem
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Optional, Union
+
+import numpy as np
+import pydantic.v1 as pd
+
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.components.types import ArrayLike
+
+
+class AbstractInitializationSpec(Tidy3dBaseModel, ABC):
+    """Abstract base class for initialization specifications."""
+
+    @abstractmethod
+    def create_parameters(self, shape: tuple[int, ...]) -> np.ndarray:
+        """Generate the parameter array based on the specification."""
+        pass
+
+
+class RandomInitializationSpec(AbstractInitializationSpec):
+    """Specification for random initial parameters.
+
+    When a seed is provided, a call to `create_parameters` will always return the same array.
+    """
+
+    min_value: float = pd.Field(
+        0.0,
+        title="Minimum Value",
+        description="Minimum value for the random parameters (inclusive).",
+    )
+    max_value: float = pd.Field(
+        1.0,
+        title="Minimum Value",
+        description="Maximum value for the random parameters (exclusive).",
+    )
+    seed: Optional[int] = pd.Field(None, description="Seed for the random number generator.")
+
+    def create_parameters(self, shape: tuple[int, ...]) -> np.ndarray:
+        """Generate the parameter array based on the specification."""
+        rng = np.random.default_rng(self.seed)
+        return rng.uniform(self.min_value, self.max_value, shape)
+
+
+class UniformInitializationSpec(AbstractInitializationSpec):
+    """Specification for uniform initial parameters."""
+
+    value: float = pd.Field(
+        0.5,
+        title="Value",
+        description="Value to use for all elements in the parameter array.",
+    )
+
+    def create_parameters(self, shape: tuple[int, ...]) -> np.ndarray:
+        """Generate the parameter array based on the specification."""
+        return np.full(shape, self.value)
+
+
+class CustomInitializationSpec(AbstractInitializationSpec):
+    """Specification for custom initial parameters provided by the user."""
+
+    params: ArrayLike = pd.Field(
+        ...,
+        title="Parameters",
+        description="Custom parameters provided by the user.",
+    )
+
+    def create_parameters(self, shape: tuple[int, ...]) -> np.ndarray:
+        """Return the custom parameters provided by the user."""
+        params = np.asarray(self.params)
+        if params.shape != shape:
+            raise ValueError(
+                f"Provided 'params.shape' ('{params.shape}') does not match "
+                f"the shape of the custom parameters ('{shape}')."
+            )
+        return params
+
+
+InitializationSpecType = Union[
+    RandomInitializationSpec,
+    UniformInitializationSpec,
+    CustomInitializationSpec,
+]
--- a/tidy3d/plugins/invdes/optimizer.py
+++ b/tidy3d/plugins/invdes/optimizer.py
@@ -77,14 +77,16 @@ class AbstractOptimizer(InvdesBaseModel, abc.ABC):
         print(f"\tpost_process_val = {result.post_process_val[-1]:.3e}")
         print(f"\tpenalty = {result.penalty[-1]:.3e}")
 
-    def _initialize_result(self, params0: anp.ndarray = None) -> InverseDesignResult:
-        """Create an initially empty ``InverseDesignResult`` from the starting parameters."""
+    def initialize_result(self) -> InverseDesignResult:
+        """
+        Create an initially empty `InverseDesignResult` from the starting parameters.
 
-        # initialize optimizer
-        if params0 is None:
-            params0 = self.design.design_region.params_half
-        params0 = anp.array(params0)
-
+        Returns
+        -------
+        InverseDesignResult
+            An instance of `InverseDesignResult` initialized with the starting parameters and state.
+        """
+        params0 = self.design.design_region.initial_parameters
         state = self.initial_state(params0)
 
         # initialize empty result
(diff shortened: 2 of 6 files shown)


e-g-melo:
Thanks @yaugenst-flex! It seems very good.

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -65,6 +66,12 @@ class AbstractOptimizer(InvdesBaseModel, abc.ABC):
         "last computed state of these variables.",
     )
 
+    parameter_spec: typing.Optional[ParameterSpecType] = pd.Field(
+        None,
```
tylerflex: should the default just be a `UniformParameterSpec()`?

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -80,10 +87,19 @@ def display_fn(self, result: InverseDesignResult, step_index: int) -> None:
     def _initialize_result(self, params0: anp.ndarray = None) -> InverseDesignResult:
```
tylerflex: should we just change the API to accept `parameter_spec` as a required argument? and then deprecation warning if `params0` passed?

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -80,10 +87,19 @@ def display_fn(self, result: InverseDesignResult, step_index: int) -> None:
...
+            raise ValueError(
+                "Initial parameters can be provided either through 'parameter_spec' or as 'params0', but not both."
+            )
+
         if params0 is None:
-            params0 = self.design.design_region.params_half
-        params0 = anp.array(params0)
+            shape = self.design.design_region.params_shape
+            if self.parameter_spec is None:
+                params0 = UniformParameterSpec(value=0.5).create_parameters(shape=shape)
```
tylerflex: dont need `value=0.5` right?

tidy3d/plugins/invdes/parameters.py
```diff
@@ -0,0 +1,65 @@
...
+        pass
+
+
+class RandomParameterSpec(AbstractParameterSpec):
+    """Specification for random initial parameters.
+
+    When a seed is provided, a call to `create_parameters` will always return the same array.
+    """
+
+    min_value: float = pd.Field(0.0, description="Minimum value for the random parameters.")
```
tylerflex: inclusive or exclusive?

tylerflex: note that rng.uniform is exclusive of max apparently

tidy3d/plugins/invdes/parameters.py
```diff
@@ -0,0 +1,65 @@
...
+class CustomParameterSpec(AbstractParameterSpec):
+    """Specification for custom initial parameters provided by the user."""
+
+    params: ArrayLike = pd.Field(..., description="Custom parameters provided by the user.")
+
+    def create_parameters(self, shape: tuple[int, ...]) -> np.ndarray:
+        """Return the custom parameters provided by the user."""
+        params = np.asarray(self.params)
+        if params.shape != shape:
+            raise ValueError("Provided shape does not match the shape of the custom parameters.")
```
tylerflex: would be helpful to display the shape
```py
            raise ValueError(f"Provided 'params.shape' ('{params.shape}') does not match the shape of the custom parameters ('{shape}')}.")
```

tylerflex (changes requested):
Overall looks good to me.

We should also discuss whether to just bite the bullet and get rid of the `params` passing in favor of this API, which I'd be ok with. This one seems overall more easily serializable.

tidy3d/plugins/invdes/optimizer.py
yaugenst-flex: if we don't support `params0` then yes, otherwise we'd have to let `params0` take precedence over this if defined

tidy3d/plugins/invdes/optimizer.py
yaugenst-flex: sure

tidy3d/plugins/invdes/optimizer.py
yaugenst-flex: yeah this is just explicit. default is 0.5

tidy3d/plugins/invdes/parameters.py
yaugenst-flex: huh good point, wasn't aware of this. I think it might be better to just document it instead of working around it? I doubt it matters in practice

tidy3d/plugins/invdes/optimizer.py
tylerflex: yea true, so what do you think? should we just transition to this new way of doing things or support both?

yaugenst-flex: no I guess this is fine, especially if we support passing custom parameters via `CustomParameterSpec`

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -65,6 +66,12 @@ class AbstractOptimizer(InvdesBaseModel, abc.ABC):
         "last computed state of these variables.",
     )
 
+    parameter_spec: ParameterSpecType = pd.Field(
+        default_factory=lambda: UniformParameterSpec(value=0.5),
```
tylerflex: Why does it need a factory? And not sure how the docs will render this

tylerflex (commented):
Looks good. Just a minor comment. Also should the parameter spec maybe go into the inverse design? Instead of the optimizer spec? Now I'm not sure. Thoughts?

tidy3d/plugins/invdes/optimizer.py
yaugenst-flex: yeah doesn't need it, for some reason I thought it'd be nicer to not instantiate before calling the class, but it seems a bit pointless. removed it

yaugenst-flex:
> Also should the parameter spec maybe go into the inverse design? Instead of the optimizer spec? Now I'm not sure. Thoughts?

I think it makes more sense to have the parameters be part of the optimizer, at least I would have argued that that's standard practice. it gets a bit weird with topopt-specific initialization, but in general I think parameters belong to the optimizer

yaugenst-flex:
well actually maybe I should take that back, they "belong" to the design region. hmmmmmmm...

tylerflex:
yea probably `DesignRegion` is what I meant. still think there's a good argument for either there or `Optimizer` but should discuss now I guess.

yaugenst-flex:
hmmmm probably we should tell the `DesignRegion` how to initialize parameters. logically that makes sense, since the parameters themselves are closely tied to the design region. so putting them in the optimizer will lead to stronger coupling between design region and optimizer than necessary.
another pretty compelling reason is i think if we introduce multiple design regions down the line, and each of those regions has its own method of initialization (i.e. one region would have a `RandomParameterSpec`, and another a `UniformParameterSpec`).

which means reverting all the changes in this PR :rofl:
so we pass parameters directly to `optimizer.run()` again?

tylerflex:
would we really need to pass parameters to `optimizer.run()`?

yaugenst-flex:
well we'd need to pass them to the optimizer somewhere :D i don't think we should pass a copy of the design region's `ParameterSpec` to the optimizer, especially in light of eventually having multiple parameter specs. so the options i see are either passing the parameters to the optimizer via a field on initialization, or we pass them during `run()`?

tylerflex:
couldn't the optimizer grab them from `self.design.design_region.param_spec`? maybe as a property `initial_parameters` of the `ParameterSpec`?

https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/plugins/invdes/optimizer.py#L80-L91

tylerflex:
also kind of annoying but we might need a separate optimizer for every design region, if we want it to be really general. For example, I had to do this for the GC + taper..

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -77,14 +77,11 @@ def display_fn(self, result: InverseDesignResult, step_index: int) -> None:
...
+    def _initialize_result(self) -> InverseDesignResult:
         """Create an initially empty ``InverseDesignResult`` from the starting parameters."""
 
-        # initialize optimizer
-        if params0 is None:
-            params0 = self.design.design_region.params_half
-        params0 = anp.array(params0)
-
+        params0 = self.design.design_region.initial_parameters
+        self.design.design_region._check_params(params0)
```
tylerflex: could we just check this in the `initial_parameters`?

tidy3d/plugins/invdes/region.py
```diff
@@ -160,30 +172,6 @@ def params_shape(self) -> typing.Tuple[int, int, int]:
         num_pixels[np.isinf(num_pixels)] = 1
         return tuple(int(n) for n in num_pixels)
 
-    def params_uniform(self, value: float) -> np.ndarray:
```
tylerflex, line 163: is it worth just keeping these but throwing deprecation warnings? guess for this demo it's not necessary..

tidy3d/plugins/invdes/initialization.py
tylerflex, line 1: `pd.Field`s are missing `title`

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -93,9 +90,21 @@ def _initialize_result(self, params0: anp.ndarray = None) -> InverseDesignResult
...
-        self.design.design_region._check_params(params0)
-        starting_result = self._initialize_result(params0)
+        """Run this inverse design problem from an optional initial set of parameters.
+
+        Parameters
+        ----------
+        post_process_fn : typing.Optional[typing.Callable] = None
+            Optional post-processing function to apply during optimization.
+        params0 : anp.ndarray = None
+            Deprecated. Initial set of parameters. This will be ignored. Use ``parameter_spec`` instead.
```
tylerflex: maybe ``TopologyDesignRegion.intialization_spec``

tylerflex (commented):
looks good , just a few minor comments

tidy3d/plugins/invdes/region.py
yaugenst-flex: do we replace them with dummy methods? because we can't really sample anything other than the parameter spec that was provided. or do keep the full methods around?

yaugenst-flex: ok keeping the methods around and warning users for now that we'll remove it

PR #1964, opened 2024-09-19
implementing interpolation option for `plot_field` method

### LINKED ISSUE #1952: Add interpolation option for `SimulationData.plot_field` method.
I would suggest implementing a variable for interpolation in the `SimulationData.plot_field` method, so the user has an easy way to plot a smooth field profile.

Without interpolation:

image

With interpolation:

image

In the graph above, the `.interp` method for the Xarray field data object was used to increase the density of points from the original grid points to 1000.

small priority

FilipeFcp:
Hi @tylerflex,

I’ve implemented the interpolation option. Here's how I approached it:

The default is set to False, so it generates a regular plot.
If set to `True`, it interpolates with 200 x 200 points.
If a `tuple (a, b)` is provided, it interpolates with a x b points.
If an `int a` is provided, it interpolates with a x a points.
I added an assertion to prevent users from interpolating more than 1000 points in any dimension to avoid potential notebook crashes due to memory limitations.

Let me know if this approach is too complex; I can simplify it if needed, like just True or False.

Fixes #1952

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Users can pass the `shading` argument in the `SimulationData.plot_field` to `Xarray.plot` method. When `shading='gouraud'`, the image is interpolated, producing a smoother visualization.
 - Users can manually specify the background medium for a structure to be used for geometry gradient calculations by supplying `Structure.background_permittivity`. This is useful when there are overlapping structures or structures embedded in other mediums.
 - Autograd functions can now be called directly on `DataArray` (e.g., `np.sum(data_array)`) in objective functions.
 
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -447,6 +447,7 @@ class AbstractYeeGridSimulationData(AbstractSimulationData, ABC):
         vmin: float = None,
         vmax: float = None,
         ax: Ax = None,
+        shading: str = "flat",
         **sel_kwargs,
     ) -> Ax:
         """Plot the field data for a monitor with simulation plot overlaid.
@@ -481,6 +482,8 @@ class AbstractYeeGridSimulationData(AbstractSimulationData, ABC):
             inferred from the data and other keyword arguments.
         ax : matplotlib.axes._subplots.Axes = None
             matplotlib axes to plot on, if not specified, one is created.
+        shading: str = 'flat'
+            Shading argument for Xarray plot method ('flat','nearest','goraud')
         sel_kwargs : keyword arguments used to perform ``.sel()`` selection in the monitor data.
             These kwargs can select over the spatial dimensions (``x``, ``y``, ``z``),
             frequency or time dimensions (``f``, ``t``) or ``mode_index``, if applicable.
@@ -493,7 +496,6 @@ class AbstractYeeGridSimulationData(AbstractSimulationData, ABC):
         matplotlib.axes._subplots.Axes
             The supplied or created matplotlib axes.
         """
-
         # get the DataArray corresponding to the monitor_name and field_name
         # deprecated intensity
         if field_name == "int":
@@ -640,6 +642,8 @@ class AbstractYeeGridSimulationData(AbstractSimulationData, ABC):
             vmax=vmax,
             cmap_type=cmap_type,
             ax=ax,
+            shading=shading,
+            infer_intervals=True if shading == "flat" else False,
         )
 
     def plot_field(
@@ -654,6 +658,7 @@ class AbstractYeeGridSimulationData(AbstractSimulationData, ABC):
         vmin: float = None,
         vmax: float = None,
         ax: Ax = None,
+        shading: str = "flat",
         **sel_kwargs,
     ) -> Ax:
         """Plot the field data for a monitor with simulation plot overlaid.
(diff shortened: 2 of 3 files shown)


tylerflex:
Hi @FilipeFcp Thanks for this!
The interpolation approach makes some sense, but also introduces these new parameters (like number of points). I wonder if instead we should include a way to downsample the fields?

In xarray there is a convenient way to grab every `n` data point by using `slice`
https://docs.xarray.dev/en/latest/user-guide/indexing.html

```py
>>> import xarray as xr
>>> da = xr.DataArray([1, 2, 3], [("x", [0, 1, 2])])
>>> da.isel(x=slice(None, None, 2))
<xarray.DataArray (x: 2)> Size: 16B
array([1, 3])
Coordinates:
  * x        (x) int64 16B 0 2
```

perhaps we could instead allow the user to input the `n` for `slice(None, None, n)` (with default of 1)? and use that to downsample in all directions?

Just a thought let me know how you think that would work or if you prefer the interpolation approach

FilipeFcp:
What I had in mind was a way to add more data points, making it easier to create smoother visualizations. For example, moving from this:

image

To this:

image

I understand that the slice option is useful for quickly plotting a large dataset, but could it be used to generate a smoother image? However, if users are most likely looking to downsample datasets before plotting rather than generating more points for visualization purposes, the slice option would make more sense.

tylerflex:
ah ok, I misunderstood. Maybe in the same sprit, it could be nice if the `interpolation` argument was a `float = 1` that just specified the ratio of points in the interpolation coords to the number of points in the raw data? for example `nterpolation=2` would mean we sample the points at 2x resolution of the supplied data. Would this make sense? 

The only thing about the current approach I think is not ideal is that it requires the hardcoding of these values (200x200) whereas the data might have a different aspect ratio, so it could lead to weird results. 

Are there other examples of interpolation in plotting and what do they do?

Also, could you include a changelog item under Added for the final PR? thanks

FilipeFcp:
Hi @tylerflex, your float approach is actually way better. 

Ideally, the user could be able to pass the `shading` argument for plotting, but it seems to be bugged at the moment: https://github.com/pydata/xarray/issues/3002.

tylerflex:
looks like they could also try this?

https://github.com/pydata/xarray/issues/3002#issuecomment-735419937

FilipeFcp:
Oh... I saw this comment, but I tried with infer_interval = True not False. Yes, it works with False.

I think we can pass the shading argument through - plot_field -> plot_field_monitor_data -> plot_scalar_array -> field_data.plot. Or create a **args in plot_scalar_array and input it into field_data.plot.

tylerflex:
> Oh... I saw this comment, but I tried with infer_interval = True not False. Yes, it works with False.

Great

> I think we can pass the shading argument through - plot_field -> plot_field_monitor_data -> plot_scalar_array -> field_data.plot. Or create a **args in plot_scalar_array and input it into field_data.plot.


yea I think either a shading argument or just exploring all of the **kwargs for xarray would work. I dont have a strong preference. what do you think is better? will this be used often you think?

tylerflex:
Hi @FilipeFcp sorry I let this slide, what's the current status of it?

I would say that the current implementation looks pretty good, but we should also add one test, just to make sure calling this with a value of, say 2, will work? and also adding a CHANGLOG item mentioning the new feature

Thanks again!

FilipeFcp:
Hi @tylerflex,

It was my bad. I had a lot of meetings and other tasks when Tom was in China, and I forgot about this for a while.

If you don't mind, I’ve pushed a different version that passes the shading argument. It seems simpler, though we lost the option to reduce the data for plotting. It was ready a month ago, just pending a review and push.

Let me know which implementation you think is better, and I’ll add it to the changelog.

Thank you.

tidy3d/components/data/sim_data.py
```diff
@@ -819,7 +831,7 @@ def plot_scalar_array(
         y_coord_values = field_data.coords[y_coord_label]
         ax.set_xlim(min(x_coord_values), max(x_coord_values))
         ax.set_ylim(min(y_coord_values), max(y_coord_values))
-
+        self.__dict__["field_data_"] = field_data
```
tylerflex: what does this do? was it left in by mistake?

tylerflex:
> It was my bad. I had a lot of meetings and other tasks when Tom was in China, and I forgot about this for a while.

No worries at all. I also forgeot.

> If you don't mind, I’ve pushed a different version that passes the shading argument. It seems simpler, though we lost the option to reduce the data for plotting. It was ready a month ago, just pending a review and push.

Yea this seems great I like it better since it's more simple and adds less complication.

> Let me know which implementation you think is better, and I’ll add it to the changelog.

That would be great when you get around to it, thanks!

tidy3d/components/data/sim_data.py
FilipeFcp: At some point, I wanted to access the field data, so I did this ugly workaround. I forgot to remove it.

FilipeFcp:
Ok, I think is done.
Let me know if there is anything to improve.

Thank you!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Users can pass the `shading` argument in the `SimulationData.plot_field` to `Xarray.plot` method. When shading='gouraud', the image is interpolated, producing a smoother visualization.
```
tylerflex: could you put `shading='gouraud'` into the backslash quotes?

tylerflex (changes requested):
Please modify a test showing the new shading option passed to `plot_field`. You can see some examples here

https://github.com/flexcompute/tidy3d/blob/develop/tests/test_data/test_sim_data.py#L139-L145

tylerflex (approved):

tylerflex:
Thanks @FilipeFcp !

PR #1969, opened 2024-09-23
Make serializable invdes backwards-compatible

yaugenst-flex:
I think this is all that's required to make the new invdes backward-compatible?


tylerflex (approved):

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -86,7 +88,14 @@ def initialize_result(self) -> InverseDesignResult:
         InverseDesignResult
             An instance of `InverseDesignResult` initialized with the starting parameters and state.
         """
-        params0 = self.design.design_region.initial_parameters
+        if params0 is not None:
+            td.log.warning(
+                "The 'params0' argument is deprecated and will be removed in the "
+                "future. Please use an 'initialization_spec' in the design region "
```
tylerflex: pretty minor but I might say `'DesignRegion.initialization_spec'` just to be more explicit.

PR #1970, opened 2024-09-23
String to Expression parser for invdes

yaugenst-flex:
Still need to add metrics, but basic functionality is there.

Usage:
```python
from tidy3d.plugins.expressions import Parser

parser = Parser()

expressions = [
    "abs(sin(x/2))",
    "x^2 + sin(y)",
    "2 (sin(x)/cos(x))**2",
    "2x + |x|^2",
    "2(x + y)",
    "2*(x+2y)",
    "sqrt(2)*(x+2y)",
]

for expr_str in expressions:
    expr = parser.parse(expr_str)
    print(f"Expression: {expr_str}")
    print(f"Parsed: {expr}\n")
```

Output:
```
Expression: abs(sin(x/2))
Parsed: abs(Sin((x / 2)))

Expression: x^2 + sin(y)
Parsed: ((x ** 2) + Sin(y))

Expression: 2 (sin(x)/cos(x))**2
Parsed: (2 * ((Sin(x) / Cos(x)) ** 2))

Expression: 2x + |x|^2
Parsed: ((2 * x) + (abs(x) ** 2))

Expression: 2(x + y)
Parsed: (2 * (x + y))

Expression: 2*(x+2y)
Parsed: (2 * (x + (2 * y)))

Expression: sqrt(2)*(x+2y)
Parsed: ((2 ** 0.5) * (x + (2 * y)))
```

### DIFF
--- a/tidy3d/plugins/expressions/parser.py
+++ b/tidy3d/plugins/expressions/parser.py
@@ -1,14 +1,36 @@
+import importlib
+import inspect
 import re
+from enum import Enum, auto
 from typing import List, Tuple
 
-from .functions import Cos, Exp, Log, Log10, Sin, Tan
+from .functions import Function
 from .operators import Abs
 from .types import ExpressionType
 from .variables import Constant, Variable
 
+# Dynamically collect all function classes
+function_classes = {}
+functions_module = importlib.import_module(".functions", package=__package__)
+for name, obj in inspect.getmembers(functions_module):
+    if inspect.isclass(obj) and issubclass(obj, Function) and obj is not Function:
+        function_classes[name.lower()] = obj
+
+function_names_pattern = r"|".join(map(re.escape, function_classes.keys()))
+
 
 class ParserError(Exception):
     """Custom exception class for parser errors."""
+
+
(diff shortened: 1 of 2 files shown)


tidy3d/plugins/expressions/parser.py
tylerflex, line 1: impressive :)

tylerflex:
On first glance, looks good. Just wondering whether there were any existing tools to do this parsing? slightly worried about maintainability

yaugenst-flex:
There's a bunch of refactoring to do still, I think I can make it pretty easy to add new functions/operators that we implement to the parser basically just via a decorator, something like

```python
# functions.py

@register_function
class Sqrt:
    ...
```

and then the parsing would work.

But yeah this is pretty prone to edge cases and I'm sure there are a lot of strings that might be failing. One aspect is just documenting it well and making it as intuitive as possible, the other aspect of course is extensive testing, but tests for this are also a bit tricky to write. I can check if I can find any existing libraries (of course tons of parsing libraries exist), the more tricky part is making it work for our use case.

PR #1972, opened 2024-09-23
`Structure.from_permittivity_array`

### LINKED ISSUE #1971: Convenience method for creating CustomMedium
kind of a shower thought, but I think the usability for creating CustomMedium would be improved if we had some kind of method that automatically created the coordinates and DataArray bits given some geometry. For example:

```py
s = td.Structure.from_data(
    geometry=box,
    eps_data=np.random.random((Nx, Ny, Nz))
)
```

and this would just use the `geometry.bounds` to create some evenly spaced `coords` inside and set up the `CustomMedium`.

tylerflex:
Fixes #1971 

Open to changing the name `Structure.from_custom_data` since it technically just creates structure from custom permittivity data? maybe it could be generalized?

But basically I think this will simplify some adjoint demos so would be good to have in 2.7.4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Convenience method `Structure.from_permittivity_array(geometry, eps_data)`, which creates structure containing `CustomMedium` with `eps_data` array sampled within `geometry.bounds`.
+
 ### Fixed
 - Ensure `path` argument in `run()` function is respected when running under autograd or the adjoint plugin.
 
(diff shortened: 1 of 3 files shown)


yaugenst-flex (approved):
Oh nice this is really useful. I think the name is fine in principle. To be a bit more explicit, maybe something like `.from_permittivity_array()`?

tylerflex:
> `from_permittivity_array`

Sounds good.

BTW implemented in an example here 
https://github.com/flexcompute/tidy3d-notebooks/pull/173

tylerflex:
@yaugenst-flex  made name change here and in notebook

momchil-flex (approved):

momchil-flex:
Approved, ideally the procedure should be rebase and merge notebook PR -> update notebook submodule state here to `develop` -> rebase and merge this one.

tylerflex:
>  rebase and merge notebook PR -> update notebook submodule state here to develop -> rebase and merge this one.

done, I'll merge when the tests pass again

PR #1973, opened 2024-09-23
`invdes` plugin GUI support

### LINKED ISSUE #1943: Integration of serializable expressions as post-processing option for `invdes`

### LINKED ISSUE #1944: Serializable functions/expressions

yaugenst-flex:
This is a rebased & cleaned-up version of #1848, with changes properly grouped. Got a bit messy with rebasing those commits (lots of overlapping changes introduced at different points of development), so figured this was cleaner.

Didn't want to force-push there in case we want to keep a reference to the original commits around.

### DIFF
--- a/tidy3d/plugins/expressions/README.md
+++ b/tidy3d/plugins/expressions/README.md
@@ -1,0 +1,303 @@
+# Expressions module
+
+The `expressions` module provides a way to construct and serialize complex mathematical expressions involving simulation metrics in Tidy3D.
+It allows users to define objective functions for optimization tasks, for example in conjunction with the `invdes` plugin.
+This module is essential for creating expressions that can be easily saved, loaded, and passed between different components of a simulation workflow.
+
+## Introduction
+
+The `expressions` module is designed to facilitate the construction of serializable mathematical expressions involving simulation metrics like mode power, mode coefficients, field intensity, and more.
+These expressions can be combined using standard arithmetic operators.
+This functionality is important when defining objective functions for optimization routines, especially in inverse design tasks where the objective functions may need to be transmitted or stored.
+
+### Creating metrics
+
+To start using the `expressions` module, you can create metrics using predefined metric classes such as `ModeAmp` and `ModePower`.
+
+Metrics are a special instance of a `Variable`, which itself is a subclass of `Expression`, that take a `SimulationData` as input and are intended to return a single scalar.
+
+```python
+from tidy3d.plugins.expressions import ModeAmp, ModePower
+
+# Create a ModeAmp metric
+mode_coeff = ModeAmp(monitor_name="monitor1", freqs=[1.0])
+
+# Create a ModePower metric
+mode_power = ModePower(monitor_name="monitor2", freqs=[1.0])
+```
+
+### Combining metrics with operators
+
+Metrics can be combined using standard arithmetic operators like `+`, `-`, `*`, `/`, `**`, and functions like `abs()`.
+Together, they will form an `Expression`.
+
+```python
+# Define an objective function using metrics
+f = abs(mode_coeff) - mode_power / 2
+```
+
+### Functions
+
+The `expressions` module also provides a set of mathematical functions that can be used to create more complex expressions.
+
+```python
+from tidy3d.plugins.expressions import Sin, Cos
+
+f = Sin(mode_coeff) + Cos(mode_power)
+```
+
+### Evaluating expressions
+
+Once you have a metric, you can evaluate it using simulation data.
+
+```python
+# Assume "data" is a SimulationData object obtained from a simulation
+result = f.evaluate(data)
+
+# ...or just
+result = f(data)
+```
+
+### Serializing and deserializing expressions
+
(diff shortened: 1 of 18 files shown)


tylerflex (approved):
Looks great!

tidy3d/plugins/expressions/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+#### Important notes on variable evaluation
+
+- **Unnamed Variables**: If you have an unnamed variable (`Variable()`), you must provide exactly one positional argument during evaluation.
+Providing multiple positional arguments will raise a `ValueError`.
+
+- **Named Variables**: For each named variable, you must provide a corresponding keyword argument.
+If a named variable is missing during evaluation, a `ValueError` will be raised.
+
+- **Multiple Unnamed Variables**: If your expression contains multiple unnamed variables, you cannot provide multiple positional arguments.
```
momchil-flex, line 231: This detailed readme is great, thanks for putting it together. One comment here is just that this is making me wonder why bother support unnamed variables at all?

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -77,43 +77,98 @@ def display_fn(self, result: InverseDesignResult, step_index: int) -> None:
...
+        ----------
+        post_process_fn : Optional[Callable] = None
+            Function to apply on the simulation data results to produce the final objective function
+            value. If not provided, then ``Optimizer.design.metric`` must be defined.
+        callback : Optional[Callable] = None
+            Callback function to apply at every iteration step for extra functionality. Does not
+            need to be differentiable. This takes the optimizer ``result`` as a positional argument
+            and the ``step_index`` and ``aux_data`` as optional arguments.
+        params0 : anp.ndarray = None
+            Deprecated. Initial set of parameters. This will be ignored. Use ``TopologyDesignRegion.intialization_spec`` instead.
```
momchil-flex: Now it won't be ignored.

tidy3d/plugins/invdes/optimizer.py
```diff
@@ -77,43 +77,98 @@ def display_fn(self, result: InverseDesignResult, step_index: int) -> None:
...
+        Returns
+        -------
+        InverseDesignResult
+            An instance of `InverseDesignResult` initialized with the starting parameters and state.
+        """
+        if params0 is not None:
+            td.log.warning(
+                "The 'params0' argument is deprecated and will be removed in the future. "
+                "Please use a 'DesignRegion.initialization_spec' in the design region "
+                "to specify initial parameters instead."
```
momchil-flex: Either in this warning or in another one we should clarify that `'params0'` still takes precedence if both that and `initializatin_spec` are provided.

momchil-flex (approved):
Just a couple small notes.

tidy3d/plugins/expressions/README.md
yaugenst-flex: it's not strictly necessary, but I thought it would be a nice convenience to not always have to provide keyword arguments, i.e. you might just want to do:
```python
a = Variable()
f = 2 * a**2
f(2)  # = 8
```

if there were no unnamed variables, you'd have to do
```python
a = Variable("a")
f = 2 * a**2
f(a=2)  # = 8
```

which is ever so slightly more cumbersome...

tidy3d/plugins/invdes/optimizer.py
yaugenst-flex: done

tidy3d/plugins/invdes/optimizer.py
yaugenst-flex: done

tidy3d/plugins/expressions/README.md
momchil-flex: Right, only now you have to remember the extra rule that you can only have one of those and stumble upon errors if you end up used to mixing them.. So in some sense just teaching people to always use the slightly more cumbersome version seems like less potential annoyances down the road.

yaugenst-flex: yeah fair enough, we can make the name required

yaugenst-flex: @momchil-flex this is actually a bit more involved than I thought initially, because we'd also require the metrics to be named (since they are also variables). and then there needs to be logic in the `InverseDesign` while constructing the objective function that extracts the names from potentially nested expression objects to pass the simulation data. like, I can add it, but it'll be a relatively big change to introduce now. I'm leaning towards leaving it as-is for now and implement that change if it turns out that the current approach is causing confusion?
it's a relatively small change w.r.t. what's user-facing.

momchil-flex: Yeah, ok, if it's that involved it's fine by me (even to potentially keep as is forever, it's a nit-y thing).

yaugenst-flex:
ok, well I guess we're probably ready to go here then? 😬

tylerflex:
yea I think so! can be in a separate PR, but check out on slack the error I got using invdes today. might be related to an edge case in 2D sims that we should probably handle before we officially release this

Issue #1479, opened 2024-02-17
Native support for Cauchy model


tomflexcompute:
Cauchy model for the refractive index is quite commonly used for semiconductors and glass materials, similar to the Sellmeier model. Currently users can define a Cauchy model and use our fitter to fit it, which works well in most cases. Is it possible to convert a Cauchy model to a pole-residue model directly without fitting? If so, we can support this model and avoid asking users to manually do the fitting. 

Feel free to close this issue if such a conversion is not possible and fitting is necessary.

weiliangjin2021:
Is the two-term form sufficient?

tomflexcompute:
It looks like people often use 3 or 4 terms. Two term form would just be equivalent to Sellmeier?

tylerflex:
can we close this because of #1974?

yaugenst-flex:
Closing — this is covered by #1974 (construct medium from a transfer function, merged Oct 2024), which provides a general mechanism that subsumes the Cauchy model request.

PR #1974, opened 2024-09-23
Construct medium from a transfer function

dmarek-flex:
Not sure if this is of interest to others, but when working on lumped elements I found that I needed to construct a `PoleResidue` medium given an admittance function in the Laplace domain, which is most commonly written as a rational expression. This might be useful in other cases where you would like to model a material that is not easily converted to `PoleResidue` format, but the differential equation governing the polarization current is available.

More precisely, given $\epsilon_\infty$ and an admittance relationship between electric field and a polarization current density

$$ J_p(s) = Y(s)E(s)$$

where

$$Y(s) = \frac{a_0 + a_1 s + \dots + a_M s^M}{b_0 + b_1 s + \dots + b_N s^N}$$

we can find an equivalent complex permittivity (assuming $Y(s)$ is causal, substitute $s = j \omega$) 

$$\epsilon(\omega) = \epsilon_\infty - \frac{1}{\\epsilon_0 j \omega}Y(j \omega).$$

Finally, we compute the poles and residues of $\frac{1}{\\epsilon_0 j \omega}Y(j \omega)$ to convert it into a `PoleResidue`  medium.

At this point I am just curious if this would be useful outside of lumped elements.

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -15,6 +15,7 @@ import autograd.numpy as np
 import numpy as npo
 import pydantic.v1 as pd
 import xarray as xr
+from scipy import signal
 
 from ..constants import (
     C_0,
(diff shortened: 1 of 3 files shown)


tidy3d/components/medium.py
```diff
@@ -15,6 +15,7 @@
 import numpy as npo
 import pydantic.v1 as pd
 import xarray as xr
+from scipy import signal
```
tylerflex, line 18: we need to double check with MC because this might add scipy as a "core" dependency, which can slow down the backend. Is it possible to instead do a lazy import of this? see `SimulationData.to_mat_file`

https://github.com/flexcompute/tidy3d/blob/b068ed8d7107fe4381d3013e42d72e74b5a7c2e7/tidy3d/components/data/sim_data.py#L1322-L1329

tylerflex (commented):
Just a comment about scipy. otherwise looks good!

tylerflex:
oh yea, and reminder to add changelog item for it

tidy3d/components/medium.py
tylerflex: not sure if we want to do exactly this. but maybe double check where we import scipy in other parts of `components/` to see if we have some standard way of doing it. Thanks

tylerflex: actually I just did a quick check and it does seem like we use it in a couple places imported from top level, so it's probably fine and we can address all of these when it comes up again.

caseyflex (changes requested):
Nice! My one suggestion would be that the name `from_transfer_function` is a bit ambiguous, I might think that `a` and `b` are just the numerator and denominator coeffs of `eps(s)` unless I read carefully. Maybe `from_admittance_coeffs` or something to make it clear that there is an additional step before partial fraction decomposition?

Actually, along those lines, maybe we can also split this function into two functions. One `from_admittance_coeffs` or something, and the other one can just be `_real_partial_fraction_decomposition` which does partial fraction decomposition using scipy and then takes only one representative of each complex conjugate pair.

tidy3d/components/medium.py
```diff
@@ -3188,6 +3189,103 @@ def compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradFieldM
...
+            An equivalent :class:`.PoleResidue` medium is constructed using an equivalent frequency-dependent
+            complex permittivity defined as
+
+            .. math::
+                \\epsilon(s) = \\epsilon_\\infty - \\frac{1}{\\epsilon_0 s}
+                \\frac{a_0 + a_1 s + \\dots + a_M s^M}{b_0 + b_1 s + \\dots + b_N s^N}.
+        """
+
+        if a.ndim != 1 or np.any(np.iscomplex(a)):
+            raise ValidationError("Coefficients must be a one-dimensional array of real numbers.")
```
weiliangjin2021: "Numerator coefficients ..."

tidy3d/components/medium.py
```diff
@@ -3188,6 +3189,103 @@ def compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradFieldM
...
+        if b.ndim != 1 or np.any(np.iscomplex(b)):
+            raise ValidationError("Coefficients must be a one-dimensional array of real numbers.")
```
weiliangjin2021: "Denominator coefficients ..."

tidy3d/components/medium.py
```diff
@@ -3188,6 +3189,103 @@ def compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradFieldM
...
+        assert b.ndim == 1
```
weiliangjin2021: redundant?

tidy3d/components/medium.py
```diff
@@ -3188,6 +3189,103 @@ def compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradFieldM
...
+
+        # Modify the transfer function defining a complex conductivity to match the complex
+        # frequency-dependent portion of the pole residue model
+        # Divide by -j*omega*epsilon (s*epsilon)
+        b = np.append(b * EPSILON_0, 0)
+        # Compute residues and poles using scipy
+        # TODO can we find a better tolerance?
+        # Frequency range is usually quite large GHz-THz, so tolerance of 1 is
+        # relatively accurate.
+        pole_tol = 1
```
weiliangjin2021: how about using `rel_tol * min(frequency_range, min_frequency)`, and pass a default value of `rel_tol`?

tidy3d/components/medium.py
```diff
@@ -3188,6 +3189,103 @@ def compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradFieldM
...
+            # in simplest form. So skip this pole.
+            if res == 0:
+                continue
+            # Causal and stability check
+            if np.real(pole) > 0:
+                raise ValidationError("Transfer function is invalid. It is non-causal.")
+            # Check for higher order pole
+            if len(np.argwhere(np.array(p_filtered) == pole)) != 0:
+                raise ValidationError(
+                    "Transfer function is invalid. A higher order pole was detected."
```
weiliangjin2021: add a hint that this can be resolved by using a smaller tolerance?

weiliangjin2021:
Nice work!
 
> Actually, along those lines, maybe we can also split this function into two functions. One `from_admittance_coeffs` or something, and the other one can just be `_real_partial_fraction_decomposition` which does partial fraction decomposition using scipy and then takes only one representative of each complex conjugate pair.

I also feel it's better to split this function, so that one can apply it directly in other cases when it's not from a transfer function. E.g. it can be applied https://github.com/flexcompute/tidy3d/issues/1479 and partially (though not the best practice) https://github.com/flexcompute/tidy3d/pull/1240

caseyflex (approved):

tidy3d/components/medium.py
dmarek-flex: The frequency range is not known in this function. I added it as a function parameter with a reasonable default.

weiliangjin2021 (approved):

caseyflex:
@weiliangjin2021 @dmarek-flex 
In principle the partial fraction decomposition is uniquely defined, so if we start from a transfer function which is stable and passive, we obtain a pole residue model which is stable and passive. I wonder if this is numerically robust here? If we start with a pole on the frequency axis, could we end up finding a pole slightly below it? If we start with Im(eps) = 0, could we end up with Im(eps) < 0?

dmarek-flex:
> @weiliangjin2021 @dmarek-flex In principle the partial fraction decomposition is uniquely defined, so if we start from a transfer function which is stable and passive, we obtain a pole residue model which is stable and passive. I wonder if this is numerically robust here? If we start with a pole on the frequency axis, could we end up finding a pole slightly below it? If we start with Im(eps) = 0, could we end up with Im(eps) < 0?

I assume that is a possibility, I was going to try and give a better answer, but could not find one! Instead I at least added more tests and validation to this convenience function to ensure that the supplied admittance transfer function is realizable from passive components. But it doesn't guarantee that the final computed PoleResidue medium will be passive over all frequencies if there is some kind of numerical error when computing residues and poles.

caseyflex:
> > @weiliangjin2021 @dmarek-flex In principle the partial fraction decomposition is uniquely defined, so if we start from a transfer function which is stable and passive, we obtain a pole residue model which is stable and passive. I wonder if this is numerically robust here? If we start with a pole on the frequency axis, could we end up finding a pole slightly below it? If we start with Im(eps) = 0, could we end up with Im(eps) < 0?
> 
> I assume that is a possibility, I was going to try and give a better answer, but could not find one! Instead I at least added more tests and validation to this convenience function to ensure that the supplied admittance transfer function is realizable from passive components. But it doesn't guarantee that the final computed PoleResidue medium will be passive over all frequencies if there is some kind of numerical error when computing residues and poles.

Maybe you can check PoleResidue._imag_eps_extrema_with_samples to warn the user if the resulting model is not passive?

dmarek-flex:
@caseyflex Good idea, I added the test you suggested and at very high frequencies one of the admittance functions I was testing does show slight gain, although it is very small ~-1e-22. In fact, in this case I think it is numerical error when computing the complex permittivity that is causing a small negative part to appear. So I added a warning to the log when a gain term larger than machine precision is detected.

caseyflex:
Sounds good, thanks!

dmarek-flex:
I'm gonna merge this unless you have any final comments @tylerflex

tylerflex (approved):
go for it!

PR #1977, opened 2024-09-24
Fix bug in subsection with nonlinear materials source detection

caseyflex:
Subsection removes sources not intersecting the region, but some nonlinear materials rely on these sources to obtain n0 and freq0.

This PR updates all such nonlinear materials during subsection to obtain and hardcode n0 and freq0 so they no longer require the sources.


FilipeFcp:
Hi @caseyflex.
It worked here. Thank you!

FilipeFcp (approved):

weiliangjin2021 (approved):

PR #1980, opened 2024-09-25
expanded lumped elements

dmarek-flex:
Adding support for simple RLC circuits as lumped elements, as well as more generalized networks that are described by an  admittance transfer function in the Laplace domain. These elements get directly translated into equivalent `PoleResidue` medium given their admittance and geometric dimensions.

Implementation follows the `LumpedResistor` as far as how the lumped elements are added to the simulation, but allows for arbitrary networks. The formulation for the equivalent medium properties starts from the desired VI admittance $Y(s)$ relationship in the Laplace domain. 

$$I(s) = Y(s)V(s)$$

Assume we connect the lumped element, so that the voltage drop is along the $z$ axis in a grid cell with $\Delta x$, $\Delta y$, $\Delta z$. Then, a current density associated with the lumped element $J_L$ and the electric field in the cell $$E(s)$$ can be related through $Y(s)$ as

$$J_{L}(s) = \frac{\Delta z}{\Delta x \Delta y} Y(s)E(s)$$

Assuming $Y(s)$ is causal, we replace $s$ with $j \omega$, which allows $J_L$ to be added to the FDTD update equations just like the polarization current densities used to model dispersive media. So all we need to do is put $\frac{\Delta z}{\Delta x \Delta y} Y(s)$ into the same format as the `PoleResidue` medium, which starts with converting the complex conductivity $\frac{\Delta z}{\Delta x \Delta y} Y(j \omega)$ into an equivalent complex permittivity. Then, we compute a partial fraction expansion to retrieve the poles and residues.

A couple of details:
- The lumped element can be distributed over multiple grid cells by scaling its admittance by the dimensions of the Box representing the lumped element. (Assuming E is constant within the box).
- The lumped element is represented by a Medium2D, so the scaling along the normal axis is factored in when the Medium2D is added to the simulation.
- A pure inductance either by itself or in parallel with the rest of a circuit requires modelling a lossless Drude medium with a second order pole at $\omega=0$. Currently, I add a bit of loss, so that it can be represented by a PoleResidue medium.
- Interestingly, a pure capacitance results in a non-dispersive medium which augments the background material by adding to $\varepsilon_\infty$

~~The first commit in this PR is from another [PR](https://github.com/flexcompute/tidy3d/pull/1974) being merged.~~


dmarek-flex:
Maybe lets start with feedback from @QimingFlex and @weiliangjin2021 before we get @tylerflex to review.

tidy3d/components/lumped_element.py
```diff
@@ -281,19 +335,461 @@ def to_structure(self) -> Structure:
...
+        np.ndarray
+            The resulting complex conductivity.
+        """
+
+        if dt is None:
+            # If the FDTD time step is not provided, use the limiting case of dt -> 0.
+            # In many cases, this approximation is quite accurate.
+            K_tan = -1j * 2 * np.pi * freqs
+        else:
+            K_tan = -1j * (2 / dt) * np.tan(2 * np.pi * freqs * dt / 2)
```
weiliangjin2021: It looks ADE implementation of CCPR model is assuming dt->0. I wonder when will the numerical conductivity here be useful.

tidy3d/components/lumped_element.py
```diff
@@ -281,19 +335,461 @@ def to_structure(self) -> Structure:
...
+        L = self.inductance
+        if self._number_circuit_elements == 1:
+            # Lumped element is simply a resistor, inductor, or capacitor.
+            if R:
+                return ((1, 0), (R, 0))
+            if C:
+                return ((0, C), (1, 0))
+            if L:
+                return ((1, 0), (0, L))
+        elif self.circuit_topology == "series":
```
weiliangjin2021: in old days when we were still using pylint, it will complain that one should use "if" rather than "elif" here, as the function returns in the previous branching

tidy3d/components/lumped_element.py
```diff
@@ -281,19 +335,461 @@ def to_structure(self) -> Structure:
...
+            return LumpedRLC.to_series_circuit_transfer_function(R, L, C)
+        else:
```
weiliangjin2021: Similarly, no need to use "else"

weiliangjin2021 (commented):
First pass, all look quite good! Any notebook examples?

tidy3d/components/lumped_element.py
```diff
@@ -146,32 +176,56 @@ def _voltage_axis_in_plane(cls, val, values):
...
+    )
+
     @cached_property
     def _sheet_conductance(self):
         """Effective sheet conductance."""
-        lateral_axis = 3 - self.voltage_axis - self.normal_axis
-        return self.size[self.voltage_axis] / self.size[lateral_axis] / self.resistance
+        size_voltage = self.size[self.voltage_axis]
+        size_lateral = self.size[self.lateral_axis]
+        return size_voltage / size_lateral / self.resistance
```
QimingFlex, line 278: Thanks @dmarek-flex ! One question regarding general lumped elements, do we treat lumped elements as 2D sheet rather than a 3D structure from Taflove? As the conductance here is defined by dz/dx/R rather than dz/(dx*dy)/R.

QimingFlex: Looking forward to the notebook so I can play with examples.

dmarek-flex:
> First pass, all look quite good! Any notebook examples?

Forthcoming ... :smile:

tidy3d/components/lumped_element.py
dmarek-flex: This is left over from my initial implementation and I kept it for easy comparison to the reference paper and possibly future testing in case it is needed. But, it is probably not needed.

tidy3d/components/lumped_element.py
dmarek-flex: This part is not very clear, but I based my new implementation on the old `LumpedResistor` implementation. We treat lumped elements as 2D sheets currently that end up becoming volumetric when the Medium2D is added to the simulation. The final scaling along the normal axis of the 2D sheet is applied there and the resulting thickness is equivalent to the grid spacing.

dmarek-flex:
@weiliangjin2021 @QimingFlex I have been considering modifying lumped elements so that they are a similar composition as `Structure`. Meaning that they are composed of a geometry with some additional details like `voltage_axis` and a circuit (instead of medium). Although right now the choice of geometry is limited either box or annulus in the special case of CoaxialLumpedResistor, it still might be nice for users to reuse the definition of a circuit in multiple places in the `Simulation`.

weiliangjin2021 (approved):
All look in good shape to me!

tidy3d/components/geometry/utils.py
```diff
@@ -2,12 +2,18 @@
 
 from __future__ import annotations
 
+from enum import Enum
 from math import isclose
 from typing import Tuple, Union
 
 import numpy as np
+import pydantic as pydantic
```
QimingFlex, line 10: It should be equivalent to just 'import pydantic' ?

tidy3d/components/lumped_element.py
```diff
@@ -81,22 +84,52 @@ class LumpedResistor(AbstractLumpedResistor, Box):
...
+        """Normal axis of the lumped element, which is the axis where the element has zero size."""
         return self.size.index(0.0)
 
+    @cached_property
+    def lateral_axis(self):
+        """Lateral axis of the lumped element."""
+        return 3 - self.voltage_axis - self.normal_axis
+
+    @cached_property
+    def _voltage_axis_2d(self) -> Axis:
```
QimingFlex: We have ‘Axis2D’

tidy3d/components/lumped_element.py
```diff
@@ -81,22 +84,52 @@ class LumpedResistor(AbstractLumpedResistor, Box):
...
+    @cached_property
+    def _snapping_spec(self) -> SnappingSpec:
+        """Returns the snapping behavior for each dimension of the lumped element."""
+        snap_location = [SnapLocation.Boundary] * 3
+        snap_behavior = [SnapBehavior.Closest] * 3
+        snap_location[self.lateral_axis] = SnapLocation.Center
+        snap_behavior[self.lateral_axis] = SnapBehavior.Expand
+        return SnappingSpec(location=snap_location, behavior=snap_behavior)
+
+    _plane_validator = assert_plane()
```
QimingFlex, line 122: What's the usage of this?

tidy3d/components/lumped_element.py
```diff
@@ -281,19 +365,464 @@ def to_structure(self) -> Structure:
...
+            if R:
+                return ((1, 0), (R, 0))
+            if C:
+                return ((0, C), (1, 0))
+            if L:
+                return ((1, 0), (0, L))
+        if self.circuit_topology == "series":
+            return RLCNetwork._to_series_circuit_transfer_function(R, L, C)
+        return RLCNetwork._to_parallel_circuit_transfer_function(R, L, C)
+
```
QimingFlex, line 550: They both should return the same value, right?

QimingFlex:
@dmarek-flex It looks good! I haven't had the chance to examine the RLC parameters as shown in the notebook in detail, but I'll do that if I have the time. One more question about _admittance_transfer_function_scaling: could we suggest a default box or handle this internally instead of asking users to define one? From the user's perspective, the lumped element should be dimensionless.

tidy3d/components/geometry/utils.py
dmarek-flex: The difference is that to access symbols from pydantic you have to use `pydantic.symbol` instead of just `symbol`. Same idea as the numpy import above, but just using the entire word.

dmarek-flex:
> One more question about _admittance_transfer_function_scaling: could we suggest a default box or handle this internally instead of asking users to define one?

Users do not need to call this method directly, I need to have a `Box` parameter to handle the case when snapping is used and the precise geometry of the lumped element changes slightly.

I think that was your question :smile: . If you mean do we need users to define the size of the box of the lumped element to begin with, then the answer is yes. In the future we could have a better method that places a lumped element between two defined edges, but right now we need the user to size the `Box` correctly depending on how it is connected in the structure.

QimingFlex:
> > One more question about _admittance_transfer_function_scaling: could we suggest a default box or handle this internally instead of asking users to define one?
> 
> Users do not need to call this method directly, I need to have a `Box` parameter to handle the case when snapping is used and the precise geometry of the lumped element changes slightly.
> 
> I think that was your question 😄 . If you mean do we need users to define the size of the box of the lumped element to begin with, then the answer is yes. In the future we could have a better method that places a lumped element between two defined edges, but right now we need the user to size the `Box` correctly depending on how it is connected in the structure.

Yeah, I mean the latter. I'm curious about how the box size would affect the final results. We can determine the optimal size internally in the future, so users won't need to define a box.

dmarek-flex:
> Yeah, I mean the latter. I'm curious about how the box size would affect the final results. We can determine the optimal size internally in the future, so users won't need to define a box.

So there are really 2 sizes that need to be determined 1) the size along the voltage axis, and 2) the lateral size, which is referred to as the current axis in the code. 1) is easy, it must be long enough to connect the two edges (e.g. the signal and ground of a microstrip). 2) is less obvious, but I have found that to get accurate results you need the element to be wide enough so that the current flows evenly from signal to ground, so you want the element to be wide enough so that it coincides with the region where the electric field is mostly uniform (e.g. make it as wide as the signal strip). 

A more realistic choice is to size the box to roughly the size of the physical lumped element you would like to model by looking up its physical dimensions.

tidy3d/components/geometry/utils.py
weiliangjin2021: simpler with `import pydantic as pd` as in other places of tidy3d?

tidy3d/components/lumped_element.py
```diff
@@ -281,19 +397,464 @@ def to_structure(self) -> Structure:
...
+    inductance: Optional[pydantic.PositiveFloat] = pydantic.Field(
+        None,
+        title="Inductance",
+        description="Inductance value in henrys.",
+        unit=HENRY,
+    )
+
+    circuit_topology: Optional[Literal["series", "parallel"]] = pydantic.Field(
+        "series",
```
weiliangjin2021, line 561: `Optional` here will allow the field taking the value `None`, which I believe is not your intention? You can remove `Optional`, and the field will take default value `series` if not defined.

tidy3d/components/geometry/utils.py
```diff
@@ -206,3 +212,128 @@ def validate_no_transformed_polyslabs(geometry: GeometryType, transform: MatrixR
...
+
+def snap_box_to_grid(grid: Grid, box: Box, snap_spec: SnappingSpec, rtol=fp_eps) -> Box:
+    """Snaps a :class:`.Box` to the grid, so that the boundaries of the box are aligned with grid centers or boundaries.
+    The way in which each dimension of the `box` is snapped to the grid is controlled by ``snap_spec``.
+    """
+
+    def closest_value(test: float, first: float, second: float) -> float:
+        """Helper to choose the closest value to a given test value."""
+        dfirst = np.abs(test - first)
+        dsecond = np.abs(test - second)
```
weiliangjin2021: `abs` instead of `np.abs` for performance gain: `np.abs` will construct a numpy array.

tidy3d/components/geometry/utils.py
```diff
@@ -206,3 +212,128 @@ def validate_no_transformed_polyslabs(geometry: GeometryType, transform: MatrixR
...
+            if min_left_idx > 0:
+                min_snap = closest_value(min, coords[min_left_idx - 1], coords[min_left_idx])
+            else:
+                min_snap = coords[min_left_idx]
+            if max_left_idx > 0:
+                max_snap = closest_value(max, coords[max_left_idx - 1], coords[max_left_idx])
+            else:
+                max_snap = coords[max_left_idx]
+        elif snap_type == SnapBehavior.Expand:
+            if min_left_idx == 0 or np.isclose(coords[min_left_idx], min, rtol=rtol):
```
weiliangjin2021: similarly, more efficient to use `math.isclose`

tidy3d/components/geometry/utils.py
```diff
@@ -206,3 +212,128 @@ def validate_no_transformed_polyslabs(geometry: GeometryType, transform: MatrixR
...
+
+    def closest_value(test: float, first: float, second: float) -> float:
+        """Helper to choose the closest value to a given test value."""
+        dfirst = np.abs(test - first)
+        dsecond = np.abs(test - second)
+        return first if dfirst < dsecond else second
+
+    def find_snapping_locations(
+        min: float, max: float, coords: np.ndarray, snap_type: SnapBehavior
+    ) -> tuple[float, float]:
```
weiliangjin2021, line 319: Maybe define a helper function that finds the closest value in the array to the left or right. Then you don't need to repeat those lines in 3 snap types x 2 test values.

Also, probably not a good idea to use `min` and `max` as variable names to override python internal functions, but be more specific, e.g. `interval_min`.

tidy3d/components/geometry/utils.py
```diff
@@ -206,3 +212,128 @@ def validate_no_transformed_polyslabs(geometry: GeometryType, transform: MatrixR
...
+                min_snap = coords[min_left_idx]
+            if max_left_idx == 0 or np.isclose(coords[max_left_idx], max, rtol=rtol):
+                max_snap = coords[max_left_idx]
+            else:
+                max_snap = coords[max_left_idx - 1]
+        return (min_snap, max_snap)
+
+    # Iterate over each axis and apply the specified snapping behavior.
+    min_b = list(box.bounds[0])
+    max_b = list(box.bounds[1])
```
weiliangjin2021: one line: `min_b, max_b = (list(f) for f in box.bounds)`

weiliangjin2021 (commented):
Half way through the new implementation. Quite exciting new automation features!

tidy3d/components/geometry/utils.py
```diff
@@ -206,3 +212,144 @@ def validate_no_transformed_polyslabs(geometry: GeometryType, transform: MatrixR
...
+
+
+def snap_box_to_grid(grid: Grid, box: Box, snap_spec: SnappingSpec, rtol=fp_eps) -> Box:
+    """Snaps a :class:`.Box` to the grid, so that the boundaries of the box are aligned with grid centers or boundaries.
+    The way in which each dimension of the `box` is snapped to the grid is controlled by ``snap_spec``.
+    """
+
+    def get_closest_value(test: float, coords: np.ArrayLike, upper_bound_idx: int) -> float:
+        """Helper to choose the closest value in an array to a given test value,
+        using the index of the upper bound."""
```
weiliangjin2021: will be helpful to explain `upper_bound_idx`, e.g. " ... closest value in coords[upper_bound_idx-1, upper_bound_idx] to a given test value ..., padded with edge values ...".

tidy3d/components/geometry/utils.py
```diff
@@ -206,3 +212,144 @@ def validate_no_transformed_polyslabs(geometry: GeometryType, transform: MatrixR
...
+        upper_bound = coords[upper_bound_idx]
+        dlower = abs(test - lower_bound)
+        dupper = abs(test - upper_bound)
+        return lower_bound if dlower < dupper else upper_bound
+
+    def get_lower_bound(
+        test: float, coords: np.ArrayLike, upper_bound_idx: int, rel_tol: float
+    ) -> float:
+        """Helper to choose the lower bound in an array for a given test value,
+        using the index of the upper bound. If the test value is close to the upper
```
weiliangjin2021, line 295: "the index of the lower bound"

tidy3d/components/lumped_element.py
```diff
@@ -81,22 +86,52 @@ class LumpedResistor(AbstractLumpedResistor, Box):
...
+        else:
+            return self.voltage_axis - 1
+
+    @cached_property
+    def _snapping_spec(self) -> SnappingSpec:
+        """Returns the snapping behavior for each dimension of the lumped element."""
+        snap_location = [SnapLocation.Boundary] * 3
+        snap_behavior = [SnapBehavior.Closest] * 3
+        snap_location[self.lateral_axis] = SnapLocation.Center
+        snap_behavior[self.lateral_axis] = SnapBehavior.Expand
```
weiliangjin2021, line 137: Curious about the choice of `center` and `expand`

tidy3d/components/lumped_element.py
```diff
@@ -133,6 +168,61 @@ def to_mesh_overrides(self) -> list[MeshOverrideStructure]:
...
 
+    def to_geometry(self, grid: Grid = None) -> Box:
+        """Converts the :class:`RectangularLumpedElement` object to a :class:`Box`."""
+        box = Box(size=self.size, center=self.center)
+        if grid and self.snap_perimeter_to_grid:
+            snap_location = [SnapLocation.Boundary] * 3
+            snap_behavior = [SnapBehavior.Closest] * 3
+            snap_location[self.lateral_axis] = SnapLocation.Center
+            snap_behavior[self.lateral_axis] = SnapBehavior.Expand
+            snap_spec = SnappingSpec(location=snap_location, behavior=snap_behavior)
```
weiliangjin2021: this is the same as `self._snapping_spec`?

tidy3d/components/lumped_element.py
```diff
@@ -133,6 +168,61 @@ def to_mesh_overrides(self) -> list[MeshOverrideStructure]:
...
+        that can be used to later compute voltage and current flowing through the element.
+        """
+
+        center = list(self.center)
+        # Size of monitor needs to be nonzero along the normal axis so that the magnetic field on
+        # both sides of the sheet will be available
+        mon_size = list(self.size)
+        mon_size[self.normal_axis] = 2 * (
+            increment_float(center[self.normal_axis], 1.0) - center[self.normal_axis]
+        )
```
weiliangjin2021, line 230: should the monitor be based on snapped geometry that might have difference `center` and `size`?

tidy3d/components/lumped_element.py
```diff
@@ -146,32 +236,58 @@ def _voltage_axis_in_plane(cls, val, values):
...
+    )
+
+    def _sheet_conductance(self, box: Box = None):
+        """Effective sheet conductance."""
+        size = self.size
+        if box:
+            size = box.size
+        size_voltage = size[self.voltage_axis]
+        size_lateral = size[self.lateral_axis]
+        return size_voltage / size_lateral / self.resistance
```
weiliangjin2021: just return `self._admittance_transfer_function_scaling(box) / self.resistance`

tidy3d/components/lumped_element.py
```diff
@@ -281,19 +397,464 @@ def to_structure(self) -> Structure:
...
+        .. [1]  J. A. Pereda, F. Alimenti, P. Mezzanotte, L. Roselli and R. Sorrentino, "A new algorithm
+                for the incorporation of arbitrary linear lumped networks into FDTD simulators," IEEE
+                Trans. Microw. Theory Tech., vol. 47, no. 6, pp. 943-949, Jun. 1999.
+    """
+
+    network: Union[RLCNetwork, AdmittanceNetwork] = pydantic.Field(
+        None,
+        title="Network",
+        description="The linear element produces an equivalent medium that emulates the "
+        "voltage-current relationship described by the ``network`` field.",
```
weiliangjin2021, line 883: it's likely that you'll need `discriminator=TYPE_TAG_STR` here.

weiliangjin2021 (commented):
Just went through the rest. Since we had `LumpedResistor` already, I wonder if we shall have `LumpedCapacitor` and `LumpedInductor`, even though they are just special cases of RLCNetwork? I guess it depends on how often capacitor and inductor are used by itself.

tidy3d/components/geometry/utils.py
dmarek-flex: In this case, it is actually trying to find the value itself, not the index

weiliangjin2021: ah yeah, right

tidy3d/components/lumped_element.py
dmarek-flex: This was an option that improved the accuracy quite a bit. Generating the equivalent medium is quite sensitive to the exact size of the Box, and so I need to make sure that the "simulated" box is the same size as the box used to generated the equivalent medium. Basically, this snapping enforces that the component of electric field parallel to the current is fully enclosed by the equivalent media, so that staircasing will not change the dimensions of the box.

dmarek-flex:
> Since we had `LumpedResistor` already, I wonder if we shall have `LumpedCapacitor` and `LumpedInductor`, even though they are just special cases of RLCNetwork?

The only reason I kept `LumpedResistor` was for backwards compatibility, I think `RLCNetwork` is a pretty simple interface that handles these cases, so I wasn't planning on add `LumpedCapacitor` and `LumpedInductor` .

tidy3d/components/lumped_element.py
dmarek-flex: Yea, I was relying the fact that usually the monitor will include additional cells. But to be sure, I will need to snap to the load position.

dmarek-flex:
I think the PR is close to its final state, so I invite ~@tylerflex~ @caseyflex  to take a look too now. If it helps, you can take a look at my work in progress [tutorial notebook ](https://github.com/flexcompute/tidy3d-notebooks/tree/dmarek/lumped_element_tutorial).

tidy3d/components/lumped_element.py
```diff
@@ -52,51 +64,74 @@ def to_mesh_overrides(self) -> list[MeshOverrideStructure]:
...
         title="Voltage Drop Axis",
         description="Specifies the axis along which the component is oriented and along which the "
         "associated voltage drop will occur. Must be in the plane of the element.",
     )
 
-    _plane_validator = assert_plane()
+    snap_perimeter_to_grid: bool = pd.Field(
+        True,
+        title="Snap Perimeter to Grid",
+        description="When enabled, the perimeter of the lumped element is snapped to the simulation grid, "
```
caseyflex, line 92: Can you say here which part of the grid the perimeter is snapped to?

tidy3d/components/lumped_element.py
```diff
@@ -52,51 +64,74 @@ def to_mesh_overrides(self) -> list[MeshOverrideStructure]:
...
+        else:
+            return self.voltage_axis - 1
+
+    @cached_property
+    def _snapping_spec(self) -> SnappingSpec:
+        """Returns the snapping behavior for each dimension of the lumped element."""
+        snap_location = [SnapLocation.Boundary] * 3
+        snap_behavior = [SnapBehavior.Closest] * 3
+        snap_location[self.lateral_axis] = SnapLocation.Center
+        snap_behavior[self.lateral_axis] = SnapBehavior.Expand
```
caseyflex, line 137: The choices of snap location and snap behavior here seem quite important, but they are hard for someone unfamiliar like me to understand. Could you leave a comment in the code explaining them?

tidy3d/components/lumped_element.py
```diff
@@ -133,7 +168,57 @@ def to_mesh_overrides(self) -> list[MeshOverrideStructure]:
...
+        """The admittance transfer function of the network needs to be scaled depending on the dimensions of the lumped element.
+        The scaling emulates adding networks with equal admittances in series and parallel, and is needed when distributing the
+        network over a finite volume.
+        """
+        size = self.size
+        if box:
+            size = box.size
+        size_voltage = size[self.voltage_axis]
+        size_lateral = size[self.lateral_axis]
+        # The final scaling along the normal axis is applied when the resulting 2D medium is averaged with the background media.
```
caseyflex, line 216: Can you elaborate in a comment here a bit more why we scale by size_voltage / size_lateral?

tidy3d/components/lumped_element.py
```diff
@@ -133,7 +168,57 @@ def to_mesh_overrides(self) -> list[MeshOverrideStructure]:
             )
         return mesh_overrides
 
-    @pydantic.validator("voltage_axis", always=True)
+    def to_geometry(self, grid: Grid = None) -> Box:
+        """Converts the :class:`RectangularLumpedElement` object to a :class:`Box`."""
+        box = Box(size=self.size, center=self.center)
+        if grid and self.snap_perimeter_to_grid:
+            return snap_box_to_grid(grid, box, self._snapping_spec)
```
caseyflex, line 191: I am curious about snapping to the grid in the normal direction. Since the structure is realized with a 2d material, it will be snapped to a primal grid boundary in volumetric_equivalent:

https://github.com/flexcompute/tidy3d/blob/2c9ae235a137ae94cdff72f29e9ae608f2ccf0b8/tidy3d/components/simulation.py#L1321

I suppose the behavior here is consistent with that and it is just snapped to closest boundary twice -- once when turning the lumped element into a 2d structure, and once turning that into volumetric equivalent.

tests/test_components/test_lumped_element.py
```diff
@@ -0,0 +1,205 @@
...
+        b=b,
+    )
+
+    med_network = network._to_medium(sf)
+    # Check conversion to geometry and to structure
+    linear_element = linear_element.updated_copy(
+        network=network,
+    )
+    _ = linear_element.to_geometry
+    _ = linear_element.to_structure
```
caseyflex: Careful - I think the new methods are functions, not cached properties, so these lines won't be doing anything. Can you run the test coverage (https://docs.flexcompute.com/projects/tidy3d/en/latest/development/usage.html) and make sure it's very close to 100%?

Also, I didn't see if you added a backend solver test, but that would be important here too.

tidy3d/components/lumped_element.py
```diff
@@ -272,28 +378,504 @@ def _sheet_conductance(self):
...
+
+    network: Union[RLCNetwork, AdmittanceNetwork] = pd.Field(
+        ...,
+        title="Network",
+        description="The linear element produces an equivalent medium that emulates the "
+        "voltage-current relationship described by the ``network`` field.",
+        discriminator=TYPE_TAG_STR,
+    )
+
+    def to_structure(self, grid: Grid = None) -> Structure:
```
caseyflex, line 887: These methods have a bit of code duplication, like calling _admittance_transfer_function_scaling. I wonder if they could be restructured a bit to be more unified in the abstract parent class.

tidy3d/components/lumped_element.py
```diff
@@ -272,28 +378,504 @@ def _sheet_conductance(self):
...
         )
 
+    def admittance(self, freqs: np.ndarray) -> np.ndarray:
+        """Returns the admittance of this lumped element at the frequencies specified by ``freqs``.
+
+        Note
+        ----
+
+        Admittance is returned using the physics convention for time-harmonic fields
+        :math:`\\exp{-j \\omega t}`, so the imaginary part of the admittance will have
```
caseyflex, line 912: This is the same convention used throughout Tidy3d, in principle

tidy3d/components/lumped_element.py
```diff
@@ -272,28 +378,504 @@ def _sheet_conductance(self):
...
+    )
+
+    inductance: Optional[pd.PositiveFloat] = pd.Field(
+        None,
+        title="Inductance",
+        description="Inductance value in henrys.",
+        unit=HENRY,
+    )
+
+    network_topology: Literal["series", "parallel"] = pd.Field(
```
caseyflex, line 560: There could be other mixed network topologies as well?

tidy3d/components/lumped_element.py
```diff
@@ -272,28 +378,504 @@ def _sheet_conductance(self):
...
+        return RLCNetwork._to_parallel_network_transfer_function(R, L, C)
+
+    @staticmethod
+    def _to_series_network_transfer_function(
+        R: float, L: float, C: float
+    ) -> tuple[tuple[float, ...], tuple[float, ...]]:
+        """Converts the RLC series network to a rational expression representing the
+        admittance of the network in the Laplace domain.
+        """
+        if R and C and L:
```
caseyflex, line 604: I wonder if we could be a bit more general and unified here if we define RLCNetwork recursively, as either a single R, L, or C element, or as a parallel or series connection of RLCNetworks.

tidy3d/components/lumped_element.py
```diff
@@ -272,28 +378,504 @@ def _sheet_conductance(self):
...
+            # Lumped element is simply a resistor, inductor, or capacitor.
+            if R:
+                # Technically zeroth order network
+                med = Medium(permittivity=1, conductivity=(scaling_factor / R))
+                return PoleResidue.from_medium(med)
+            if C:
+                med = Medium(permittivity=1 + scaling_factor * C / EPSILON_0, conductivity=0)
+                return PoleResidue.from_medium(med)
+            if L:
+                # TODO would be nice to be able to set the damping term exactly to 0
```
caseyflex, line 668: why do we need some nonzero loss here?

tidy3d/plugins/microwave/auto_path_integrals.py
```diff
@@ -0,0 +1,76 @@
...
+        sign=polarity,
+        extrapolate_to_endpoints=True,
+        snap_path_to_grid=True,
+    )
+
+    # Snap the current integral to a box that encloses the element along the lateral and normal axes
+    # using the closest positions of the magnetic field
+    snap_location = [SnapLocation.Center] * 3
+    snap_behavior = [SnapBehavior.Expand] * 3
+    snap_behavior[V_axis] = SnapBehavior.Off
```
caseyflex, line 60: Can you comment why no snapping in voltage axis?

tests/test_components/test_lumped_element.py
caseyflex: Generally I think we would use to_geometry for a method name and as_geometry for a property name to avoid confusion

tidy3d/components/lumped_element.py
```diff
@@ -146,54 +231,75 @@ def _voltage_axis_in_plane(cls, val, values):
             )
         return val
 
-    @cached_property
-    def _sheet_conductance(self):
+
+class LumpedResistor(RectangularLumpedElement):
+    """Class representing a rectangular lumped resistor. Lumped resistors are appended to the list
```
caseyflex, line 264: Can you include working examples in the docstrings?

I think it would be helpful to even have complete examples showing the process from defining the lumped element through getting the complete simulation.

caseyflex (changes requested):
Hi @dmarek-flex ,

This is incredible, really exciting work. Here's some comment and suggestions...

tidy3d/components/lumped_element.py
dmarek-flex: You might remember me asking questions a while ago about supporting Drude medium with zero loss, or equivalently supporting a PoleResidue medium with second order pole at omega=0. At the moment, it is not possible, so I have to add a bit of loss.

tidy3d/components/lumped_element.py
dmarek-flex: No, for the `RLCNetwork` it is either 1-3 elements in series or 1-3 elements in parallel.

caseyflex: hmm, that's not what I see on wikipedia:
https://en.wikipedia.org/wiki/RLC_circuit#Other_configurations

dmarek-flex: Oh, I meant that is all we support when using the `RLCNetwork`. I made the `AdmittanceNetwork` to support all other types of networks.

tidy3d/plugins/microwave/auto_path_integrals.py
```diff
@@ -0,0 +1,76 @@
...
+from ...components.grid.grid import Grid
+from ...components.lumped_element import LinearLumpedElement
+from ...components.types import Direction
+from .path_integrals import (
+    CurrentIntegralAxisAligned,
+    VoltageIntegralAxisAligned,
+)
+
+
+def path_integrals_from_lumped_element(
```
caseyflex, line 14: Is the “path integral” terminology standard here? I would call this a line integral

tidy3d/components/lumped_element.py
dmarek-flex: Starting think about how I might implement this, which I think is definitely possible but not exactly trivial. Basically, I would just need to apply rules for adding admittances in series or parallel, which would require performing operations on polynomials, using `numpy` or maybe `sympy` if I want to do it symbolically. I used `lcapy` to help me with this in the notebook I linked, but I could do something similar in Tidy3D.

If you agree, I could make this functionality into a separate issue to be implemented at a future time. The reason being this PR enables the simulation of any linear network, so this would be more of an interface improvement.

caseyflex: I agree this is a low priority. You can make an issue if you want. Thanks!

tidy3d/components/lumped_element.py
caseyflex: Gotcha. Sounds good then

tidy3d/components/lumped_element.py
caseyflex: Not sure if I missed this earlier, but also, are lumped elements handled differently than 2D materials? Why can’t we directly include them in the structure list and then afterwards construct 2d equivalent and finally volumetric equivalent in simulation postprocessing?

dmarek-flex: Yes, in effect the lumped element will be snapped twice to the grid. I need to perform it earlier, so that I can correctly setup ports in the `TerminalComponentModeler`. There are a lot of other things that depend on the precise location of the lumped element (source/monitors/path integrals).

Also yes, I believe the lumped elements behave exactly like a structure composed of a 2D medium, but just includes some additional snapping behavior right before being converted to a `Structure`. That being said, I think it is possible that the behavior of lumped elements and 2D medium might differ in the future. Also in order to maintain backwards compatibility, we can't really change it at this time.

tidy3d/components/lumped_element.py
weiliangjin2021: It is good to still emphasize, since most people in RF use the opposite convention.

tidy3d/plugins/microwave/auto_path_integrals.py
dmarek-flex: If I could go back in time, I might have called these tools `LineIntegral` s, but I think those terms are interchangeable anyways. I like path because it indicates that it could be a line, curve, or closed loop.

dmarek-flex: I think the only negative is that there might be a slight confusion with path integrals from quantum mechanics.

tidy3d/components/lumped_element.py
dmarek-flex: I'm not sure I see exactly what you mean here. I had to split up the methods into `to_geometry`, `to_structure`, and `to_medium`, because I wanted to try and make the network definition and geometry definition be independent of each other. I also need the separate `to_geometry` method for the purpose of plotting.

caseyflex: I see, that makes sense then.

tests/test_components/test_lumped_element.py
dmarek-flex: Good catch, I improved testing. lumped_elements is at 100% coverage and I also improved coverage for the other files I changed.

caseyflex: thanks!

tidy3d/components/lumped_element.py
dmarek-flex: Just let me know if the new explanation is clear enough.

tidy3d/components/lumped_element.py
dmarek-flex: Same for here

tidy3d/components/lumped_element.py
dmarek-flex: I added examples for the lumped elements and a link to the forthcoming notebook. I think that should help newcomers get started.

tidy3d/components/lumped_element.py
caseyflex: looks good, thanks

caseyflex (approved):

tidy3d/components/lumped_element.py
```diff
@@ -3,100 +3,151 @@
...
+        dimensions of the lumped element. As a result, we need to ensure that staircasing will
+        lead to an accurate representation of the lumped element. We make sure to create a
+        :class:`.Box` aligned with the Yee grid that fully encloses the
+        electric field component which is parallel to the ``voltage_axis``.
+        """
+
+        snap_location = [SnapLocation.Boundary] * 3
+        snap_behavior = [SnapBehavior.Closest] * 3
+        snap_location[self.lateral_axis] = SnapLocation.Center
+        snap_behavior[self.lateral_axis] = SnapBehavior.Expand
```
QimingFlex, line 137: I understand that snapping to the center (H-field locations) eliminates staircasing errors in the structure, but why don’t we snap to the center along the normal direction as well? If we use one Yee cell to account for the box thickness, it should be centered around the E-fields. Is my understanding correct?

weiliangjin2021: My understanding is that box thickness is 0 along normal direction. It's snapped to the boundary so that Medium2D is not visible to normal E-field.

QimingFlex: Yeah, that's what I meant— I might have misunderstood the coordinates before. So, the coors along the x-direction are recording the positions for Ex, and it's the same for Ey and Ez, right?

weiliangjin2021: right, center coord along x-direction is for Ex. Medium2D is expected to be invisible to the E-field along normal axis, so it's placed on the boundary rather than the center.

QimingFlex (approved):

PR #1984, opened 2024-09-27
Autograd <-> PyTorch wrapper

### LINKED ISSUE #1945: PyTorch (and maybe JAX) plugins for wrapping objective functions
We need some plugins to convert objective functions defined using autograd (since that is what we use) to PyTorch and possibly JAX. Proposed structure:
```
tidy3d/plugins
├── pytorch/
└── jax/
```

For JAX, we can probably just use [agjax](https://github.com/mfschubert/agjax)

yaugenst-flex:
@tylerflex fyi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for arbitrary padding sizes for all padding modes in `tidy3d.plugins.autograd.functions.pad`.
 - `Expression.filter(target_type, target_field)` method for extracting object instances and fields from nested expressions.
 - Additional constraints and validation logic to ensure correct setup of optimization problems in `invdes` plugin.
+- `tidy3d.plugins.pytorch` to wrap autograd functions for interoperability with PyTorch via the `to_torch` wrapper.
 
 ### Changed
 - Renamed `Metric.freqs` --> `Metric.f` and made frequency argument optional, in which case all frequencies from the relevant monitor will be extracted.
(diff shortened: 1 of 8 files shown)


tylerflex (approved):
Looks great, thanks for adding this.

So I guess this lets people write their objective function using autograd and then use PyTorch to optimize it? 

And I guess it would also work if they define some of their objective function using autograd and some in PyTorch?

```py
@to_torch
def get_data_ag(p):
    return web.run(get_sim(p))

def objective_pytorch(...):
    data = get_data_ag(p)
    return postprocess_pytorch(data)
```

Would it also work if I did something like
`run = to_torch(web.run)`
and use this as a drop in for web.run? to would it break I suppose since the simulation doesn't know how to store pytorch tensors?

yaugenst-flex:
so mixing and matching is fine, but there is a tidy3d core that's autograd-only, which includes everything that concerns simulation definition as well as any post-processing that involves data-arrays or uses tidy3d internals (such as flux calculation or far field projection).
so the smallest "autograd core" would be something like
```python
@to_torch
def autograd_objective(params):
    sim = make_sim(params)
    sim_data = web.run(sim)
    # any code that calls tidy3d function on sim_data, for example:
    flux = sim_data["monitor"].flux
    return flux.values
```

but of course you could do custom field postprocessing in pytorch using field monitors if you return `sim_data["field_monitor"].Ex.values` for example

yaugenst-flex:
essentially, the wrapped function needs to receive (at least some) pytorch tensors as input and return a pytorch tensor or scalar

tylerflex:
cool thanks. eventually would be nice to get an example of this in some notebook. but feel free to merge for now!

PR #1987, opened 2024-09-27
Support uniform dimensions in `TopologyDesignRegion`

yaugenst-flex:


tidy3d/plugins/invdes/region.py
```diff
@@ -166,12 +173,10 @@ def _check_params(params: anp.ndarray = None):
     @property
     def params_shape(self) -> typing.Tuple[int, int, int]:
         """Shape of the parameters array in (x, y, z), given the ``pixel_size`` and bounds."""
-        # rmin, rmax = np.array(self.geometry.bounds)
-        # lengths = rmax - rmin
         side_lengths = np.array(self.size)
         num_pixels = np.ceil(side_lengths / self.pixel_size)
         # TODO: if the structure is infinite but the simulation is finite, need reduced bounds
```
momchil-flex, line 178: Can this TODO be removed?

More generally, in principle since we have the explicit `uniform` argument now, it probably doesn't make that much sense to set to single pixel if the design region is infinite. We should just span the entire simulation instead. Thoughts @yaugenst-flex @tylerflex ?

momchil-flex (approved):

tidy3d/plugins/invdes/region.py
yaugenst-flex: i think the problem with that was that the invdes region has no information about the simulation, i.e. it can't determine the array shape from inf bounds because it can't intersect with the simulation domain.
would require more changes to make that work somehow

momchil-flex: Then maybe this should be a method of InverseDesign not a property of TopologyDesignRegion, or it should be a method that takes in a Simulation?

Not necessary to fix here but something to think about. It's just that the `inf` behavior would be kind of unexpected to me especially compared to other Tidy3D components which we often send to inf when we just want to span the whole domain.

momchil-flex: More generally though it may be hard to come up with a simulation setup in which you want your topology region to span the whole domain.. Certainly seems fishy with PML, maybe in some cases with periodic boundaries?

yaugenst-flex: yeah agree I don't think this would ever be correct with PMLs. periodic could make sense yeah

tylerflex (approved):

PR #1988, opened 2024-09-30
fix autograd bug re: inclusion extraneous field mnts

tylerflex:
still not sure:
this isn't triggered by regular autograd code. might be something with the metrics

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -1093,6 +1093,10 @@ class FieldData(FieldDataset, ElectromagneticFieldData):
                 if not np.all(values == 0):
                     src_field_components[name] = ScalarFieldDataArray(values, coords=coords)
 
+            # dont include this source if no data
+            if all(fld_cmp is None for fld_cmp in src_field_components.values()):
+                continue
+
             # construct custom Current source
             dataset = FieldDataset(**src_field_components)
             custom_source = CustomCurrentSource(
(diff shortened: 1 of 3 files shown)


yaugenst-flex (approved):

tidy3d/components/data/monitor_data.py
```diff
@@ -1093,6 +1093,10 @@ def to_adjoint_field_sources(self, fwidth: float) -> List[CustomCurrentSource]:
                 if not np.all(values == 0):
                     src_field_components[name] = ScalarFieldDataArray(values, coords=coords)
 
+            # dont include this source if no data
+            if all(fld_cmp is None for _, fld_cmp in src_field_components.items()):
```
yaugenst-flex: `src_field_components.values()`?

yaugenst-flex (approved):

tidy3d/components/data/monitor_data.py
tylerflex: I think pylint used to complain about iterating over values, so I got in the habit of always doing over `items()`. but not really a strong preference.

yaugenst-flex: hm I don't see a reason for this. might have been an artifact of possibly unordered dicts, but that's not the case anymore as of python 3.7..

yaugenst-flex: e.g. ruff will complain by default about iterating over items if only keys or values are used. we might have disabled that rule 🤔

tylerflex:
fixed values() iteration but still am not able to reproduce this error. (basically the test I added passes even without the change) would be good (but not super high priority) to understand why exactly this popped up and reproduce it in a test

yaugenst-flex:
yeah I also can't reproduce it. I think it's because emulated run does not exactly reproduce what is happening in regular run, because the code will fail is using regular run

momchil-flex:
Merging but this would need further investigation

PR #1932, opened 2024-08-30
ComponentModeler.batch_data and pathdir enhancements

tylerflex:
messed up rebase of #1918 so made a new PR

PR #1990, opened 2024-10-01
fix bug when setting 'path_dir' in ComponentModeler

dmarek-flex:
Fixed a small bug that doesn't allow you to change the `path_dir` in the `ComponentModeler`. It is from a recent PR, so no changelog required.


tylerflex (approved):

weiliangjin2021 (approved):
Could you remind me why `path_dir` needs to match the internal values?

dmarek-flex:
> Could you remind me why `path_dir` needs to match the internal values?

Changes were made [here](https://github.com/flexcompute/tidy3d/pull/1932). I think just to remain consistent and simplify. And also to predict where the `BatchData` is saved.

PR #1992, opened 2024-10-02
`invdes` validators

yaugenst-flex:
Adds a bunch of validators for `invdes` as well as constraints on some fields.
Depends on #1988

### DIFF
--- a/tidy3d/plugins/autograd/functions.py
+++ b/tidy3d/plugins/autograd/functions.py
@@ -24,156 +24,87 @@ __all__ = [
 ]
 
 
-def _make_slices(rule: Union[int, slice], ndim: int, axis: int) -> Tuple[slice, ...]:
-    """Create a tuple of slices for indexing an array.
+def _pad_indices(n: int, pad_width: Tuple[int, int], mode: PaddingType) -> NDArray:
+    """Compute the indices to pad an array along a single axis based on the padding mode.
 
     Parameters
     ----------
-    rule : Union[int, slice]
-        The rule to apply on the specified axis.
-    ndim : int
-        The number of dimensions of the array.
-    axis : int
-        The axis to which the rule should be applied.
+    n : int
+        The size of the axis to pad.
+    pad_width : Tuple[int, int]
+        The number of values padded to the edges of the axis.
+    mode : PaddingType
+        The padding mode to use.
 
     Returns
     -------
-    Tuple[slice, ...]
-        A tuple of slices for indexing.
+    NDArray
+        The indices for padding along the axis.
     """
-    return tuple(slice(None) if d != axis else rule for d in range(ndim))
+    total_pad = pad_width[0] + pad_width[1]
+    if n == 0:
+        # If the axis length is zero, return zeros for indices
+        return np.zeros(total_pad, dtype=int)
+
+    idx = np.arange(-pad_width[0], n + pad_width[1])
+    if mode == "constant":
+        # For constant mode, indices outside the array bounds are ignored
+        pass
+    elif mode == "edge":
+        idx = np.clip(idx, 0, n - 1)
+    elif mode == "reflect":
+        period = 2 * n - 2 if n > 1 else 1
+        idx = np.mod(idx, period)
+        idx = np.where(idx >= n, period - idx, idx)
+    elif mode == "symmetric":
+        period = 2 * n if n > 1 else 1
+        idx = np.mod(idx, period)
+        idx = np.where(idx >= n, period - idx - 1, idx)
+    elif mode == "wrap":
+        idx = np.mod(idx, n)
+    else:
+        raise ValueError(f"Unsupported padding mode: {mode}")
+    return idx
 
 
-def _constant_pad(
-    array: NDArray, pad_width: Tuple[int, int], axis: int, *, constant_value: float = 0.0
+def _pad(
+    array: NDArray,
+    pad_width: Tuple[int, int],
+    axis: int,
+    *,
+    mode: PaddingType = "constant",
+    constant_value: float = 0.0,
 ) -> NDArray:
-    """Pad an array with a constant value along a specified axis.
+    """Pad an array along a specified axis with a given mode and pad width.
 
     Parameters
     ----------
     array : NDArray
         The input array to pad.
     pad_width : Tuple[int, int]
-        The number of values padded to the edges of each axis.
+        The number of values padded to the edges of the axis.
     axis : int
         The axis along which to pad.
+    mode : PaddingType = "constant"
+        The padding mode to use.
     constant_value : float = 0.0
-        The constant value to pad with.
+        The constant value to pad with when mode is 'constant'.
 
     Returns
     -------
     NDArray
         The padded array.
     """
-    p = [(pad_width[0], pad_width[1]) if ax == axis else (0, 0) for ax in range(array.ndim)]
-    return np.pad(array, p, mode="constant", constant_values=constant_value)
-
-
-def _edge_pad(array: NDArray, pad_width: Tuple[int, int], axis: int) -> NDArray:
-    """Pad an array using the `edge` mode along a specified axis.
-
-    Parameters
-    ----------
-    array : NDArray
-        The input array to pad.
-    pad_width : Tuple[int, int]
-        The number of values padded to the edges of each axis.
-    axis : int
-        The axis along which to pad.
-
-    Returns
-    -------
-    NDArray
-        The padded array.
-    """
-    left, right = (_make_slices(rule, array.ndim, axis) for rule in (0, -1))
-
-    cat_arys = []
-    if pad_width[0] > 0:
-        cat_arys.append(np.stack(pad_width[0] * [array[left]], axis=axis))
-    cat_arys.append(array)
-    if pad_width[1] > 0:
-        cat_arys.append(np.stack(pad_width[1] * [array[right]], axis=axis))
-
-    return np.concatenate(cat_arys, axis=axis)
-
-
-def _reflect_pad(array: NDArray, pad_width: Tuple[int, int], axis: int) -> NDArray:
-    """Pad an array using the `reflect` mode along a specified axis.
-
-    Parameters
-    ----------
-    array : NDArray
-        The input array to pad.
-    pad_width : Tuple[int, int]
-        The number of values padded to the edges of each axis.
-    axis : int
-        The axis along which to pad.
-
-    Returns
-    -------
-    NDArray
-        The padded array.
-    """
-    left, right = (
-        _make_slices(rule, array.ndim, axis)
-        for rule in (slice(pad_width[0], 0, -1), slice(-2, -pad_width[1] - 2, -1))
-    )
-    return np.concatenate([array[left], array, array[right]], axis=axis)
-
-
-def _symmetric_pad(array: NDArray, pad_width: Tuple[int, int], axis: int) -> NDArray:
-    """Pad an array using the `symmetric` mode along a specified axis.
-
-    Parameters
-    ----------
-    array : NDArray
-        The input array to pad.
-    pad_width : Tuple[int, int]
-        The number of values padded to the edges of each axis.
-    axis : int
-        The axis along which to pad.
-
-    Returns
-    -------
-    NDArray
-        The padded array.
-    """
-    left, right = (
-        _make_slices(rule, array.ndim, axis)
-        for rule in (
-            slice(pad_width[0] - 1, None, -1) if pad_width[0] > 0 else slice(0, 0),
-            slice(-1, -pad_width[1] - 1, -1),
-        )
-    )
-    return np.concatenate([array[left], array, array[right]], axis=axis)
-
-
-def _wrap_pad(array: NDArray, pad_width: Tuple[int, int], axis: int) -> NDArray:
-    """Pad an array using the `wrap` mode along a specified axis.
-
-    Parameters
-    ----------
-    array : NDArray
-        The input array to pad.
-    pad_width : Tuple[int, int]
-        The number of values padded to the edges of each axis.
-    axis : int
-        The axis along which to pad.
-
-    Returns
-    -------
-    NDArray
-        The padded array.
-    """
-    left, right = (
-        _make_slices(rule, array.ndim, axis)
-        for rule in (
-            slice(-pad_width[0], None) if pad_width[0] > 0 else slice(0, 0),
-            slice(0, pad_width[1]) if pad_width[1] > 0 else slice(0, 0),
-        )
-    )
-    return np.concatenate([array[left], array, array[right]], axis=axis)
+    n = array.shape[axis]
+    if mode == "constant":
+        p = [(0, 0)] * array.ndim
+        p[axis] = pad_width
+        return np.pad(array, p, mode="constant", constant_values=constant_value)
+    else:
+        idx = _pad_indices(n, pad_width, mode)
+        indexer = [slice(None)] * array.ndim
+        indexer[axis] = idx
+        return array[tuple(indexer)]
 
 
 def pad(
@@ -184,7 +115,7 @@ def pad(
     axis: Union[int, Iterable[int], None] = None,
     constant_value: float = 0.0,
 ) -> NDArray:
-    """Pad an array along a specified axis with a given mode and padding width.
+    """Pad an array along specified axes with a given mode and padding width.
 
     Parameters
     ----------
@@ -210,67 +141,40 @@ def pad(
     ------
     ValueError
         If the padding width has more than two elements or if padding is negative.
-    NotImplementedError
-        If padding larger than the input size is requested.
-    KeyError
-        If an unsupported padding mode is specified.
     IndexError
         If an axis is out of range for the array dimensions.
-
-    Examples
-    --------
-    >>> import numpy as np
-    >>> from tidy3d.plugins.autograd.functions import pad
-    >>> array = np.array([[1, 2], [3, 4]])
-    >>> pad(array, (1, 1), mode="constant", constant_value=0)
-    array([[0, 0, 0, 0],
-           [0, 1, 2, 0],
-           [0, 3, 4, 0],
-           [0, 0, 0, 0]])
-    >>> pad(array, 1, mode="reflect")
-    array([[4, 3, 4, 3],
-           [2, 1, 2, 1],
-           [4, 3, 4, 3],
-           [2, 1, 2, 1]])
     """
     pad_width = np.atleast_1d(pad_width)
-
     if pad_width.size == 1:
-        pad_width = np.array([pad_width[0], pad_width[0]])
-    elif pad_width.size != 2:
-        raise ValueError("Padding width must have one or two elements, got {pad_width.size}.")
+        pad_width = (pad_width[0], pad_width[0])
+    elif pad_width.size == 2:
+        pad_width = tuple(pad_width)
+    else:
+        raise ValueError(f"Padding width must have one or two elements, got {pad_width.size}.")
 
-    if any(any(p >= s for p in pad_width) for s in array.shape):
-        raise NotImplementedError("Padding larger than the input size is not supported.")
     if any(p < 0 for p in pad_width):
-        raise ValueError("Padding must be positive.")
+        raise ValueError("Padding must be non-negative.")
     if all(p == 0 for p in pad_width):
         return array
 
     if axis is None:
-        axis = range(array.ndim)
+        axes = range(array.ndim)
+    else:
+        axes = [axis] if isinstance(axis, int) else axis
 
-    _mode_map = {
-        "constant": _constant_pad,
-        "edge": _edge_pad,
-        "reflect": _reflect_pad,
-        "symmetric": _symmetric_pad,
-        "wrap": _wrap_pad,
-    }
-
-    try:
-        pad_fun = _mode_map[mode]
-    except KeyError as e:
-        raise KeyError(f"Unsupported padding mode: {mode}.") from e
-
-    for ax in np.atleast_1d(axis):
+    for ax in axes:
         if ax < 0:
             ax += array.ndim
         if ax < 0 or ax >= array.ndim:
-            raise IndexError(f"Axis out of range for array with {array.ndim} dimensions.")
+            raise IndexError(f"Axis {ax} out of range for array with {array.ndim} dimensions.")
 
-        array = pad_fun(array, pad_width, axis=ax)
-
+        array = _pad(
+            array,
+            pad_width,
+            axis=ax,
+            mode=mode,
+            constant_value=constant_value,
+        )
     return array
 
 
--- a/tidy3d/plugins/expressions/metrics.py
+++ b/tidy3d/plugins/expressions/metrics.py
@@ -1,11 +1,10 @@
-from typing import Any
+from typing import Any, Optional, Union
 
-import autograd.numpy as anp
+import autograd.numpy as np
 import pydantic.v1 as pd
 
 from tidy3d.components.monitor import ModeMonitor
 from tidy3d.components.types import Direction, FreqArray
-from tidy3d.exceptions import ValidationError
 
 from .types import NumberType
 from .variables import Variable
@@ -27,48 +26,51 @@ class ModeAmp(Metric):
     """
     Metric for calculating the mode coefficient from a ModeMonitor.
 
-    Parameters
-    ----------
-    monitor_name : str
-        The name of the mode monitor.
-    freqs : FreqArray
-        The frequency array.
-    direction : Direction = "+"
-        The direction of the mode.
-    mode_index : pd.NonNegativeInt = 0
-        The index of the mode.
-
     Examples
     --------
-    >>> monitor = ModeMonitor(name="monitor1", freqs=[1.0])
+    >>> monitor = ModeMonitor(name="monitor1", f=[1.0])
     >>> mode_coeff = ModeAmp.from_mode_monitor(monitor)
     >>> data = SimulationData()  # Assume this is a valid SimulationData object
     >>> result = mode_coeff.evaluate(data)
     """
 
-    monitor_name: str
-    freqs: FreqArray
-    direction: Direction = "+"
-    mode_index: pd.NonNegativeInt = 0
-
-    @pd.validator("freqs", always=True)
-    def _single_frequency(cls, val: FreqArray) -> FreqArray:
-        if len(val) != 1:
-            raise ValidationError("Only a single frequency is supported at the moment.")
-        return val
+    monitor_name: str = pd.Field(
+        ...,
+        title="Monitor Name",
+        description="The name of the mode monitor. This needs to match the name of the monitor in the simulation.",
+    )
+    f: Optional[Union[float, FreqArray]] = pd.Field(  # type: ignore
+        None,
+        title="Frequency Array",
+        description="The frequency array. If None, all frequencies in the monitor will be used.",
+    )
+    direction: Direction = pd.Field(
+        "+",
+        title="Direction",
+        description="The direction of propagation of the mode.",
+    )
+    mode_index: pd.NonNegativeInt = pd.Field(
+        0,
+        title="Mode Index",
+        description="The index of the mode.",
+    )
 
     @classmethod
-    def from_mode_monitor(cls, monitor: ModeMonitor):
-        return cls(monitor_name=monitor.name, freqs=monitor.freqs, mode_index=0)
+    def from_mode_monitor(
+        cls, monitor: ModeMonitor, mode_index: int = 0, direction: Direction = "+"
+    ):
+        return cls(
+            monitor_name=monitor.name, f=monitor.freqs, mode_index=mode_index, direction=direction
+        )
 
     def evaluate(self, *args: Any, **kwargs: Any) -> NumberType:
         data = super().evaluate(*args, **kwargs)
-        amps = (
-            data[self.monitor_name]
-            .amps.sel(direction=self.direction, mode_index=self.mode_index)
-            .isel(f=0)
+        amps = data[self.monitor_name].amps.sel(
+            direction=self.direction, mode_index=self.mode_index
         )
-        return anp.squeeze(amps.values.tolist())
+        if self.f is not None:
+            amps = amps.sel(f=list(self.f), method="nearest")
+        return np.squeeze(amps.values.tolist())
 
 
 class ModePower(ModeAmp):
@@ -77,7 +79,7 @@ class ModePower(ModeAmp):
 
     Examples
     --------
-    >>> monitor = ModeMonitor(name="monitor1", freqs=[1.0])
+    >>> monitor = ModeMonitor(name="monitor1", f=[1.0])
     >>> mode_power = ModePower.from_mode_monitor(monitor)
     >>> data = SimulationData()  # Assume this is a valid SimulationData object
     >>> result = mode_power.evaluate(data)
@@ -85,4 +87,4 @@ class ModePower(ModeAmp):
 
     def evaluate(self, *args: Any, **kwargs: Any) -> NumberType:
         amps = super().evaluate(*args, **kwargs)
-        return abs(amps) ** 2
+        return np.abs(amps) ** 2
--- a/tidy3d/plugins/invdes/design.py
+++ b/tidy3d/plugins/invdes/design.py
@@ -6,11 +6,14 @@ import abc
 import typing
 
 import autograd.numpy as anp
+import numpy as np
 import pydantic.v1 as pd
 
 import tidy3d as td
 import tidy3d.web as web
 from tidy3d.components.autograd import get_static
+from tidy3d.exceptions import ValidationError
+from tidy3d.plugins.expressions.metrics import Metric
 from tidy3d.plugins.expressions.types import ExpressionType
 
 from .base import InvdesBaseModel
@@ -118,6 +121,62 @@ class InverseDesign(AbstractInverseDesign):
     )
 
     _check_sim_pixel_size = check_pixel_size("simulation")
+
+    @pd.root_validator(pre=False)
+    def _validate_model(cls, values: dict) -> dict:
+        cls._validate_metric(values)
+        return values
+
+    @staticmethod
+    def _validate_metric(values: dict) -> dict:
+        metric_expr = values.get("metric")
+        if not metric_expr:
+            return values
+        simulation = values.get("simulation")
+        for metric in metric_expr.filter(Metric):
+            InverseDesign._validate_metric_monitor_name(metric, simulation)
+            InverseDesign._validate_metric_mode_index(metric, simulation)
+            InverseDesign._validate_metric_f(metric, simulation)
+        return values
+
+    @staticmethod
+    def _validate_metric_monitor_name(metric: Metric, simulation: td.Simulation) -> None:
+        """Validate that the monitor name of the metric exists in the simulation."""
+        monitor = next((m for m in simulation.monitors if m.name == metric.monitor_name), None)
+        if monitor is None:
+            raise ValidationError(
+                f"Monitor named '{metric.monitor_name}' associated with the metric not found in the simulation monitors."
+            )
+
+    @staticmethod
+    def _validate_metric_mode_index(metric: Metric, simulation: td.Simulation) -> None:
+        """Validate that the mode index of the metric is within the bounds of the monitor's ``ModeSpec.num_modes``."""
+        monitor = next((m for m in simulation.monitors if m.name == metric.monitor_name), None)
+        if metric.mode_index >= monitor.mode_spec.num_modes:
+            raise ValidationError(
+                f"Mode index '{metric.mode_index}' for metric associated with monitor "
+                f"'{metric.monitor_name}' is out of bounds. "
+                f"Maximum allowed mode index is '{monitor.mode_spec.num_modes - 1}'."
+            )
+
+    @staticmethod
+    def _validate_metric_f(metric: Metric, simulation: td.Simulation) -> None:
+        """Validate that the frequencies of the metric are present in the monitor."""
+        monitor = next((m for m in simulation.monitors if m.name == metric.monitor_name), None)
+        if metric.f is not None:
+            if len(metric.f) != 1:
+                raise ValidationError("Only a single frequency is supported for the metric.")
+            for freq in metric.f:
+                if not any(np.isclose(freq, monitor.freqs, atol=1.0)):
+                    raise ValidationError(
+                        f"Frequency '{freq}' for metric associated with monitor "
+                        f"'{metric.monitor_name}' not found in monitor frequencies."
+                    )
+        else:
+            if len(monitor.freqs) != 1:
+                raise ValidationError(
+                    f"Monitor '{metric.monitor_name}' must contain only a single frequency when metric.f is None."
+                )
 
     def is_output_monitor(self, monitor: td.Monitor) -> bool:
         """Whether a monitor is added to the ``JaxSimulation`` as an ``output_monitor``."""
--- a/tidy3d/plugins/invdes/initialization.py
+++ b/tidy3d/plugins/invdes/initialization.py
@@ -7,16 +7,19 @@ from typing import Optional, Union
 
 import numpy as np
 import pydantic.v1 as pd
+from numpy.typing import NDArray
 
+import tidy3d as td
 from tidy3d.components.base import Tidy3dBaseModel
 from tidy3d.components.types import ArrayLike
+from tidy3d.exceptions import ValidationError
 
 
 class AbstractInitializationSpec(Tidy3dBaseModel, ABC):
     """Abstract base class for initialization specifications."""
 
     @abstractmethod
-    def create_parameters(self, shape: tuple[int, ...]) -> np.ndarray:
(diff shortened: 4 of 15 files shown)


tidy3d/plugins/autograd/functions.py
```diff
@@ -241,7 +241,10 @@ def pad(
         raise ValueError("Padding width must have one or two elements, got {pad_width.size}.")
 
     if any(any(p >= s for p in pad_width) for s in array.shape):
-        raise NotImplementedError("Padding larger than the input size is not supported.")
+        raise NotImplementedError(
+            "Padding larger than the input size is not supported. "
+            "The convolution kernel is probably too large for the input."
```
tylerflex: can we display the relevant shapes here and maybe add more instruction on how to fix it?

tidy3d/plugins/expressions/metrics.py
```diff
@@ -66,9 +66,9 @@ def evaluate(self, *args: Any, **kwargs: Any) -> NumberType:
         amps = (
             data[self.monitor_name]
             .amps.sel(direction=self.direction, mode_index=self.mode_index)
-            .isel(f=0)
+            .sel(f=self.freqs[0], method="nearest")
```
tylerflex: I still dont quite understand the use of `0` here, doesn't this just ignore the frequency defined in the class?

tidy3d/plugins/invdes/design.py
```diff
@@ -119,6 +122,54 @@ class InverseDesign(AbstractInverseDesign):
 
     _check_sim_pixel_size = check_pixel_size("simulation")
 
+    @pd.root_validator(pre=False)
```
tylerflex, line 125: maybe could do this using the post init validators defined in Tidy3dBaseModel? would avoid needing to do values.get() stuff\

The basic idea is to overwrite this method and call your validation methods inside of it
https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/base.py#L150

tidy3d/plugins/invdes/initialization.py
```diff
@@ -29,17 +32,34 @@ class RandomInitializationSpec(AbstractInitializationSpec):
...
 
-    def create_parameters(self, shape: tuple[int, ...]) -> np.ndarray:
+    @pd.root_validator(pre=False)
+    def _validate_max_ge_min(cls, values):
+        """Ensure that max_value is greater than or equal to min_value."""
+        minval = values.get("min_value")
+        maxval = values.get("max_value")
+        if minval > maxval:
+            raise ValidationError(
+                f"'max_value' ({maxval}) must be greater or equal than 'min_value' ({minval})"
```
tylerflex, line 58: if they are equal, it might lead to some weird behavior in the projections? might want to verify

tidy3d/plugins/invdes/region.py
```diff
@@ -159,12 +175,69 @@ class TopologyDesignRegion(DesignRegion):
         "Supplying ``False`` will completely leave out the override structure.",
     )
 
+    def _validate_eps_values(self):
+        """Validate the epsilon values by evaluating the transformations."""
+        try:
+            x = self.initial_parameters
+            self.eps_values(x)
+        except Exception as e:
+            raise ValidationError(f"Could not evaluate transformations: {str(e)}") from e
```
tylerflex, line 186: will this print `e` twice though?

tylerflex (changes requested):
looks good, thanks. minor suggestions

tylerflex:
also note seemes that #1988 is included in this?

tidy3d/plugins/expressions/metrics.py
yaugenst-flex: sure yeah we can use `self.freqs`. it's guaranteed to be a single frequency through the validator

tidy3d/plugins/invdes/initialization.py
yaugenst-flex: it's fine, it's essentially the same as uniform initialization

tidy3d/plugins/invdes/design.py
yaugenst-flex: sure. I was thinking it'd be better to validate as early as possible, and here this works, i.e. you don't need an instance of the class to check the validation. I can change it

tidy3d/plugins/invdes/region.py
yaugenst-flex: no

yaugenst-flex: or, more specifically, the validation error will appear only once pretty printed in the logs, but the complete stack trace is preserved. in the stack trace, both ValidationError and the previous exception will show up

yaugenst-flex: we can also `raise from None` instead of `raise from e`, but then the traceback is lost, which I think would be bad for debugging?

yaugenst-flex:
> also note seemes that https://github.com/flexcompute/tidy3d/pull/1988 is included in this?

it is because this is branched off of that but has `develop` as a target. will be gone when that PR is merged and this is one rebased

tidy3d/plugins/autograd/functions.py
yaugenst-flex: done

tidy3d/plugins/invdes/design.py
yaugenst-flex: there are some other instance in this PR where that approach doesn't work and I use the post init validator

tidy3d/plugins/expressions/metrics.py
yaugenst-flex: fixed

tidy3d/plugins/invdes/design.py
tylerflex: ok, I'll leave it up to you, just letting you know so you know about it.

tidy3d/plugins/expressions/metrics.py
tylerflex: if self.freqs is a single frequency only maybe it should just be `f: float = None` (and if None we isel?)

yaugenst: not sure i understand? multifrequency support will be added later. currently there is only a validator that checks that len(freqs)==1, which we can remove once multifrequency works

yaugenst-flex: or do you mean making freqs optional and then isel if only a single freq is found in the monitor? but that'll be a pain to validate

e-g-melo (commented):
Thanks @yaugenst-flex! The new validators seem very detailed. Tomorrow, I will checkout your branch to run a real example.

e-g-melo:
@yaugenst-flex, I ran the [inverse design plugin example](https://docs.flexcompute.com/projects/tidy3d/en/v2.7.4/notebooks/InverseDesign.html) and got these errors in `design_region.to_structure()` and `inv_des.to_simulation()` when trying to call them without `params0`.

image

Should we make `param0` optional in `design_region.to_structure()`, `inv_des.to_simulation()`, and `inv_des.to_simulation_data()`?

tidy3d/plugins/invdes/region.py
```diff
@@ -159,12 +175,69 @@ class TopologyDesignRegion(DesignRegion):
...
+        """Validate the penalty values by evaluating the penalties."""
+        try:
+            x = self.initial_parameters
+            self.penalty_value(x)
+        except Exception as e:
+            raise ValidationError(f"Could not evaluate penalties: {str(e)}") from e
+
+    def _validate_gradients(self):
+        """Validate the gradients of the penalties and transformations."""
+        import warnings
```
momchil-flex: Should this just be top level?

momchil-flex (approved):

yaugenst-flex:
> Should we make param0 optional in design_region.to_structure(), inv_des.to_simulation(), and inv_des.to_simulation_data()?

hm good point, although it would only work for the initial parameters right? so if you want to give e.g. the optimized parameters, you still need to pass them explicitly. might be a bit confusing?

tidy3d/plugins/invdes/region.py
yaugenst-flex: sure can do. I guess I thought this is only used in the validator to catch an autograd warning, but maybe makes no sense to lazy import

e-g-melo:
> hm good point, although it would only work for the initial parameters right? so if you want to give e.g. the optimized parameters, you still need to pass them explicitly. might be a bit confusing?

Ok.. in that case, maybe we can simply modify the example and use `design_region.to_structure(design_region.initial_parameters)`

e-g-melo:
@yaugenst-flex, I made the filter radius larger than the `z` direction and set `uniform=(False, False, False)`. Then, I got this validation error. Is it the expected validation error? If that is the case, would it be possible to include a message that explicitly informs the user that the filter radius is larger than the design region in the `z` direction?
image

e-g-melo:
In the case we set `uniform=(False, False, True)` and set the projection filter radius larger than the `x, y` dimensions, this validation error is shown. Would it be possible to have a message that tells the user which component to change to solve the issue? For example:

ERROR: Could not evaluate transformations: **Conic filter radius larger than the 
             design region is not supported**. Padding widths: [240 240], Array      
             shape: (240, 252). Ensure that the conic filter radius is smaller than   
             the corresponding dimension of the design region.

image

e-g-melo:
I got this error in the 2nd iteration step when trying multiple frequencies in the metrics monitor but selecting a single one in metrics. Is it expected to work on this PR?

```py
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In[27], [line 1](vscode-notebook-cell:?execution_count=27&line=1)
----> [1](vscode-notebook-cell:?execution_count=27&line=1) result = optimizer.run()

File /mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:132, in AbstractOptimizer.run(self, post_process_fn, callback, params0)
    [117](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:117) """Run this inverse design problem from an optional initial set of parameters.
    [118](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:118) 
    [119](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:119) Parameters
   (...)
    [129](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:129)     Deprecated. Initial set of parameters. Use ``TopologyDesignRegion.intialization_spec`` instead.
    [130](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:130) """
    [131](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:131) starting_result = self.initialize_result(params0)
--> [132](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:132) return self.continue_run(
    [133](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:133)     result=starting_result,
    [134](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:134)     num_steps=self.num_steps,
    [135](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:135)     post_process_fn=post_process_fn,
    [136](https://vscode-remote
[InverseDesign_Validators3D.zip](https://github.com/user-attachments/files/17245123/InverseDesign_Validators3D.zip)
+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:136)     callback=callback,
    [137](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:137) )

File /mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:180, in AbstractOptimizer.continue_run(self, result, num_steps, post_process_fn, callback)
    [178](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:178) for step_index in range(done_steps, done_steps + num_steps):
    [179](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:179)     aux_data = {}
--> [180](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:180)     val, grad = val_and_grad_fn(params, aux_data=aux_data)
    [182](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:182)     if anp.allclose(grad, 0.0):
    [183](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:183)         td.log.warning(
    [184](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:184)             "All elements of the gradient are almost zero. This likely indicates "
    [185](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:185)             "a problem with the optimization set up. This can occur if the symmetry of the "
   (...)
    [193](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:193)             "function are contributing to the return value."
    [194](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/plugins/invdes/optimizer.py:194)         )

File ~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/wrap_util.py:20, in unary_to_nary.<locals>.nary_operator.<locals>.nary_f(*args, **kwargs)
     [18](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/wrap_util.py:18) else:
     [19](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/wrap_util.py:19)     x = tuple(args[i] for i in argnum)
---> [20](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/wrap_util.py:20) return unary_operator(unary_f, x, *nary_op_args, **nary_op_kwargs)

File ~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/differential_operators.py:143, in value_and_grad(fun, x)
    [139](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/differential_operators.py:139) if not vspace(ans).size == 1:
    [140](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/differential_operators.py:140)     raise TypeError("value_and_grad only applies to real scalar-output "
    [141](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/differential_operators.py:141)                     "functions. Try jacobian, elementwise_grad or "
    [142](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/differential_operators.py:142)                     "holomorphic_grad.")
--> [143](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/differential_operators.py:143) return ans, vjp(vspace(ans).ones())

File ~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:14, in make_vjp.<locals>.vjp(g)
---> [14](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:14) def vjp(g): return backward_pass(g, end_node)

File ~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:21, in backward_pass(g, end_node)
     [19](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:19) for node in toposort(end_node):
     [20](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:20)     outgrad = outgrads.pop(node)
---> [21](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:21)     ingrads = node.vjp(outgrad[0])
     [22](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:22)     for parent, ingrad in zip(node.parents, ingrads):
     [23](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:23)         outgrads[parent] = add_outgrads(outgrads.get(parent), ingrad)

File ~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:67, in defvjp.<locals>.vjp_argnums.<locals>.<lambda>(g)
     [64](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:64)         raise NotImplementedError(
     [65](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:65)             "VJP of {} wrt argnum 0 not defined".format(fun.__name__))
     [66](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:66)     vjp = vjpfun(ans, *args, **kwargs)
---> [67](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:67)     return lambda g: (vjp(g),)
     [68](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:68) elif L == 2:
     [69](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/autograd/core.py:69)     argnum_0, argnum_1 = argnums

File /mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:594, in _run_bwd.<locals>.vjp(data_fields_vjp)
    [591](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:591) def vjp(data_fields_vjp: AutogradFieldMap) -> AutogradFieldMap:
    [592](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:592)     """dJ/d{sim.traced_fields()} as a function of Function of dJ/d{data.traced_fields()}"""
--> [594](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:594)     sim_adj, adjoint_source_info = setup_adj(
    [595](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:595)         data_fields_vjp=data_fields_vjp,
    [596](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:596)         sim_data_orig=sim_data_orig,
    [597](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:597)         sim_fields_keys=sim_fields_keys,
    [598](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:598)     )
    [600](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:600)     # run adjoint simulation
    [601](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:601)     task_name_adj = str(task_name) + "_adjoint"

File /mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:748, in setup_adj(data_fields_vjp, sim_data_orig, sim_fields_keys)
    [745](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:745) data_vjp_paths = set(data_fields_vjp.keys())
    [747](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:747) num_monitors = len(sim_data_orig.simulation.monitors)
--> [748](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:748) adjoint_monitors = sim_data_orig.simulation.with_adjoint_monitors(sim_fields_keys).monitors[
    [749](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:749)     num_monitors:
    [750](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:750) ]
    [752](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:752) sim_adj, adjoint_source_info = sim_data_vjp.make_adjoint_sim(
    [753](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:753)     data_vjp_paths=data_vjp_paths, adjoint_monitors=adjoint_monitors
    [754](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:754) )
    [756](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/web/api/autograd/autograd.py:756) if _INSPECT_ADJOINT_FIELDS:

File /mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/simulation.py:3633, in Simulation.with_adjoint_monitors(self, sim_fields_keys)
   [3631](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/simulation.py:3631) mnts_fld, mnts_eps = self.make_adjoint_monitors(structure_indices=structure_indices)
   [3632](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/simulation.py:3632) monitors = list(self.monitors) + list(mnts_fld) + list(mnts_eps)
-> [3633](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/simulation.py:3633) return self.copy(update=dict(monitors=monitors))

File /mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:223, in Tidy3dBaseModel.copy(self, deep, **kwargs)
    [221](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:221) kwargs.update(deep=deep)
    [222](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:222) new_copy = pydantic.BaseModel.copy(self, **kwargs)
--> [223](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:223) return self.validate(new_copy.dict())

File ~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:711, in BaseModel.validate(cls, value)
    [708](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:708) value = cls._enforce_dict_if_root(value)
    [710](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:710) if isinstance(value, dict):
--> [711](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:711)     return cls(**value)
    [712](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:712) elif cls.__config__.orm_mode:
    [713](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:713)     return cls.from_orm(value)

File /mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:146, in Tidy3dBaseModel.__init__(self, **kwargs)
    [144](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:144) """Init method, includes post-init validators."""
    [145](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:145) log.begin_capture()
--> [146](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:146) super().__init__(**kwargs)
    [147](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:147) self._post_init_validators()
    [148](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev/tidy3d/tidy3d/components/base.py:148) log.end_capture(self)

File ~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:341, in BaseModel.__init__(__pydantic_self__, **data)
    [339](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:339) values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
    [340](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:340) if validation_error:
--> [341](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:341)     raise validation_error
    [342](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:342) try:
    [343](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Tidy3D_Dev_Test/test_serializable_ops/~/miniconda3/envs/tidy3d_dev/lib/python3.12/site-packages/pydantic/v1/main.py:343)     object_setattr(__pydantic_self__, '__dict__', values)

ValidationError: 1 validation error for Simulation
monitors
  'monitors' names are not unique, given ['MNT_0', 'MNT_1', 'MNT_2', 'adjoint_fld_4', 'adjoint_fld_4', 'adjoint_eps_4']. (type=value_error.setup)
```

tidy3d/plugins/expressions/metrics.py
tylerflex: No, I think what I'm confused about is that `ModePower.freqs` takes a list of freqs, but it only selects the first one. Why not just make it take a single frequency and call it `f`?

tylerflex: this would make it consistent with the xarray coords too. I think this is basically something I missed when reviewing the original metrics PR

yaugenst-flex: because we do want to support multiple freqs in the feature. it‘s just disabled right now

e-g-melo:
If I set `length_scale` to -1 I get this error, which is correct.

image

But if I set `length_scale` to 0, I get this other error. What should happen if `length_scale=0`? Should we validate it to `>0` as in the conic filter radius?

image

tidy3d/plugins/expressions/metrics.py
tylerflex: but in xarray we also would just overload the singular coordinate with a list, eg `f=[f1, f2]`. So maybe we should follow this approach here?

tylerflex: basically might be nice to be able to essentially copy the kwargs passed to `sel(**kwargs)` into `ModePower(**kwargs)`?

e-g-melo:
`override_structure_dl=0` passed the validation

image

tidy3d/plugins/invdes/region.py
tylerflex: I think how you have it is fine, I just get confused about exception handling.

e-g-melo:
`learning_rate=0` passed the validation. Should we validate it for `>0`?

image

tidy3d/plugins/expressions/metrics.py
yaugenst-flex: ah, i guess i was trying to resemble the modemonitor interface more than the dataarray one with the metrics

tylerflex: yea I think it should probably mirror the ModeAmpsDataArray, because the ModeMontior doesn't do `mode_index` for example? and this is essentially a wrapper for picking data out of the ModeAMpsDataARray right?

yaugenst-flex: yeah that's a good point. I think `**kwargs` will not be possible because we do need to validate all the fields, the `sel`s in the metrics won't be arbitrary. but we can definitely change the naming. might be better to do it in a separate PR though?

yaugenst-flex:
Hey @e-g-melo, these are great! Thanks a lot for catching all of this. Point by point:

> I made the filter radius larger than the z direction and set `uniform=(False, False, False)`. Then, I got this validation error. Is it the expected validation error?

No this is not expected, it looks like the parameters have the wrong shape. I'll have to look into this one.

> In the case we set `uniform=(False, False, True)` and set the projection filter radius larger than the x, y dimensions, this validation error is shown. Would it be possible to have a message that tells the user which component to change to solve the issue?

Ok I'll try to improve this. It's a bit tricky because scalar dimensions are removed before convolution, and the error is raised in the convolution, so there the information about the physical axis is lost. Maybe validation can happen a bit earlier.

> I got this error in the 2nd iteration step when trying multiple frequencies in the metrics monitor but selecting a single one in metrics. Is it expected to work on this PR?

In principle this should work. It looks like the `adjoint_fld_4` monitor is added twice, which causes simulation validation to fail. So I don't think this is a problem with the newly introduced validation. I'll investigate.

> But if I set `length_scale` to 0, I get this other error. What should happen if `length_scale=0`? Should we validate it to >0 as in the conic filter radius?

Good catch, fixed. `length_scale` must be positive.

> `override_structure_dl=0` passed the validation

Oooh this is interesting, and confusing. That field is being validated according to `typing.Union[pd.PositiveFloat, typing.Literal[False]]`, so I think `override_structure_dl=0` is being coerced into `False` in the validation. which disables the override structure. This _might_ actually be OK behavior? What do you think @tylerflex? To be fair we could also just change this to `pd.NonNegativeFloat`, I think this will lead to exactly the same behavior. `False` would be interpreted as `0` in that case.

> `learning_rate=0` passed the validation. Should we validate it for >0?

Absolutely, fixed.

tidy3d/plugins/expressions/metrics.py
tylerflex: no preference on the PR, but best to change before 2.7.5 I think

tylerflex:
> That field is being validated according to typing.Union[pd.PositiveFloat, typing.Literal[False]], so I think override_structure_dl=0 is being coerced into False in the validation. which disables the override structure. This might actually be OK behavior? What do you think @tylerflex? To be fair we could also just change this to pd.NonNegativeFloat, I think this will lead to exactly the same behavior. False would be interpreted as 0 in that case.

seems like expected behavior to me, although a bit confusing. I'd think from the user perspective, changing to `NonNegativeFloat` would be exactly the same? So maybe we can just leave it as is?

tylerflex:
something to note, we should update this to use the new initialization spec 

https://github.com/flexcompute/tidy3d/blob/ce648575a8c218e4a6a259ec406bd09446b40493/tidy3d/plugins/invdes/optimizer.py#L184

yaugenst-flex:
Ok, I think everything is addressed now except for the issue with the duplicate adjoint field monitors, which I think we need to investigate further, but it is not related to validation.

tidy3d/plugins/expressions/metrics.py
```diff
@@ -40,35 +40,35 @@ class ModeAmp(Metric):
...
-    >>> monitor = ModeMonitor(name="monitor1", freqs=[1.0])
+    >>> monitor = ModeMonitor(name="monitor1", f=[1.0])
     >>> mode_coeff = ModeAmp.from_mode_monitor(monitor)
     >>> data = SimulationData()  # Assume this is a valid SimulationData object
     >>> result = mode_coeff.evaluate(data)
     """
 
     monitor_name: str
-    freqs: FreqArray
+    f: FreqArray
```
tylerflex: can f take a float? I feel this would be potentially better for usability (similar to xarray coords). could add it later I suppose, but I feel making users do `f=[f]` when we only support a single frequency feels like it could be improved.

tylerflex: Also, could we make it optional? if `None` we just isel the single frequency?

tylerflex (changes requested):
Looks good overall, I still take some issue with the Metric.f thing, see comments. Basically I feel like it should operate roughly like xarray, so it's easy to take my existing postprocessing `sel()` code and transport it here, specitically
* allow single `f` value, not just array.
* if `f` is left out it automatically is selected (only works if single `f` in dataset because we dont support arrays).

Thoughts?

tidy3d/plugins/expressions/metrics.py
yaugenst-flex: pretty sure float works fine, pydantic will coerce into list on validation

yaugenst-flex: if we make it optional we can’t validate for single freq here. do you want that validation to happen somewhere else? in design would be possible along with the other post init stuff. but to me doesnt seem like the right place

yaugenst-flex:
```python
m = ModeAmp(f=1, monitor_name="")
print(m.f)  # prints [1.]
```

tylerflex: we can just raise an error when the metric is evaluated? and then the post init validation check should catch it automatically?

yaugenst-flex: no strong opinion there other than the fact that the validator would once again not show what actually failed in the log 😆 (except if you look at the stack trace)
but either way it's a bit of a crutch until we have multifrequency figured out for this, so it doesn't really matter

PR #1994, opened 2024-10-03
move adjoint source post norm from source info to simulation

momchil-flex:
Supersedes #1934 , rebased against develop.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -30,6 +30,7 @@ from .boundary import (
     PMCBoundary,
     StablePML,
 )
+from .data.data_array import FreqDataArray
 from .data.dataset import CustomSpatialDataType, Dataset
 from .geometry.base import Box, Geometry
 from .geometry.mesh import TriangleMesh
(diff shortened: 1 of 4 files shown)


tidy3d/components/simulation.py
```diff
@@ -210,6 +211,13 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
         "``autograd`` gradient processing.",
     )
 
+    post_norm: Union[float, FreqDataArray] = pydantic.Field(
+        1.0,
+        title="Post Normalization Values",
+        description="Factor to multiply the fields by after running, "
+        "given the adjoint source pipeline used. Note: this is used internally only.",
+    )
```
momchil-flex, line 219: Can we clarify this a little? I don't understand what `given the adjoint source pipeline used` means. Also, should we add the note about internal only to the `simulation_type` field too?

tylerflex: Yea maybe more clear is "Factor to multiply the fields by to incorporate the adjoint source amplitudes.".

And yea we can add this note to simulation type too

tylerflex (approved):
Approving my own changes :D

PR #2000, opened 2024-10-04
:memo: Basic example section introductions

### LINKED ISSUE #10: Docs Suggestion: Example section introductions
Kind of a minor comment but I think it will improve the docs a bit.

For each of our docs sections, there is a page with a bunch of links to the specific notebooks. For example, in the "design space exploration" section, it looks like this

https://docs.flexcompute.com/projects/tidy3d/en/v2.6.0rc1/notebooks/docs/features/parameter_sweep.html

I think what would improve things a bit is to put a few paragraphs of introduction on this page, explaining what these examples are showing and which ones might be best to check out depending on what you want to do. For example, this page could include some text above the example links like:

"
Tidy3D offers a few tools that facilitate and automate the exploration of a user-defined design space. The `design` plugin is a wrapper that allows users to define a general "parameter space" and sample this space through tidy3d simulations. The result is returned to the user in a `pandas.DataFrame` format for easy post processing and analysis. For a detailed tutorial, see [Design Plugin Tutorial Link].  We also have tools for batch processing of simulations, which can make custom parameter sweeps convenient. For an example of batching simulations and other tools for managing the results in a more custom way, see the [Parameter Sweep Tutorial Link] example.
"

we can also address a few other improvements here, such as describing the limitations of the various features, or linking to the feature README.md for more information.

daquinteroflex:
(On a side note I'm the lucky Y2K winner)

Ok finally the branch name is correct and syncing to rtd:

- [X] Basic example section introductions
- [X] Improved notebook list styling

This is a live version before I fixed the notebook colors - it'd be handy to check this with multiple platforms and browsers.
https://docs.flexcompute.com/projects/tidy3d/en/demo-test-basic_section_intro/notebooks/docs/features/autograd.html

Turns out I had misinterpreted (closes https://github.com/flexcompute/autoflex/issues/10) where I thought we wanted descriptions for each notebooks, whereas we just wanted section descriptions. So this implements section descriptions. The good thing is I now also know how to implement like notebooks links with thumbnails and descriptions rather than this list so if we want to do that we can.

image

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -29,6 +29,14 @@ iframe {width: 100%;border: unset}
 .toctree-wrapper li[class^=toctree-l3]>a {font-size: 0.9em; text-decoration: none}
 .toctree-wrapper li[class^=toctree-l4]>a {font-size: 0.9em; text-decoration: none}
 
+.toctree-wrapper.example-notebook-toc li[class^=toctree-l1] > a::before {
+    content: "📘 ";  /* Emoji before the link */
+}
+
+.toctree-wrapper.example-notebook-toc li[class^=toctree-l1] > a:hover::before {
+    content: "📖 ";  /* Change emoji on hover */
+}
+
 html[data-theme="light"] {
     --pst-color-primary: #b5445b;
     --pst-color-secondary: #4772ae;


FilipeFcp (approved):
I tested in Linux with Chrome, Opera, and Firefox. It looks great in all of them!

PR #2004, opened 2024-10-06
:fast_forward: Update latest notebooks

daquinteroflex:
So this needs to be merged  into https://github.com/flexcompute/tidy3d/pull/2003 or directly into `latest`. It depends if that notebook update requires any of the existing changes in the frontend `develop` branch or not. If not then we just have to update the submodule on the frontend `latest` branch @tylerflex fyi


daquinteroflex:
Decided to update this directly from `develop`, because there were some related invdes changes and the readthedocs sync fixes on the develop branches.

PR #2005, opened 2024-10-07
fix field projection discrepancy

dmarek-flex:
Small change to behavior when integration surface extends into PML region. We changed the behavior to closer match the server-side computations. Users should probably not be setting up simulations like this anyways, since field projection is only accurate when the field goes to 0 at the boundaries of the integration surface.

Backend [PR](https://github.com/flexcompute/tidy3d-core/pull/729) and related [issue](https://github.com/flexcompute/tidy3d-core/issues/679)


tylerflex (approved):

yaugenst-flex (approved):
Thanks!

PR #2006, opened 2024-10-07
Escaping web dependence in invdes plugin until needed

momchil-flex:


momchil-flex:
In the end, I didn't really end up needing the explicit `AbstractInverseDesign.run` functions on the backend, so I could just do the lazy import in the old code. But the lazy import did fix my test that the optimization can be loaded, and metadata processed, without needing web dependencies.

tidy3d/plugins/invdes/design.py
```diff
@@ -292,11 +305,10 @@ def to_simulation(self, params: anp.ndarray) -> dict[str, td.Simulation]:
         simulation_list = [design.to_simulation(params) for design in self.designs]
         return dict(zip(self.task_names, simulation_list))
 
-    def to_simulation_data(self, params: anp.ndarray, **kwargs) -> web.BatchData:
+    def to_simulation_data(self, params: anp.ndarray, **kwargs):  # -> "web.BatchData"
```
momchil-flex: I was a bit confused about this, I get an error if I uncomment (and also ruff complains), do we have a good way to do type annotation without importing? I remember there was something.

tylerflex: maybe just put it it as a string `'web.BatchData'`.

tylerflex (approved):
looks good to me

tidy3d/plugins/invdes/design.py
yaugenst-flex: that's strange, `from __future__ import annotations` should have exactly the same effect as wrapping the type hints in strings in python 3.9, and above it shouldn't even be necessary. what python version were you testing this with @momchil-flex?

tidy3d/plugins/invdes/design.py
```diff
@@ -100,6 +99,21 @@ def initial_simulation(self) -> td.Simulation:
         initial_params = self.design_region.initial_parameters
         return self.to_simulation(initial_params)
 
+    def run(self, simulation, **kwargs) -> td.SimulationData:
+        """Run a single tidy3d simulation."""
+        from tidy3d.web import run
+
+        kwargs.update({"verbose": self.verbose})
```
yaugenst-flex: Why `update` instead of `setdefault`? Is the idea to have the class field _always_ take precedence over the passed kwargs?

yaugenst-flex (approved):

tidy3d/plugins/invdes/design.py
yaugenst-flex: could it be that you were just seeing a lint error? probably ruff doesn't recognize this as a forward reference (it's not a type checker after all). I checked out your branch and uncommented and it's working minus the lint error, which in this case should just be disabled, so the following works for me:
```diff
-    def run_async(self, simulations, **kwargs):  # -> "web.BatchData"
+    def run_async(self, simulations, **kwargs) -> web.BatchData:  # noqa: F821
```

tidy3d/plugins/invdes/design.py
momchil-flex: Ah yeah actually I wanted to bring that up but forgot. But now that I think about this again, I think `setdefault` is better. I got confused that if some of our `run` functions set a default value for `verbose`, it will override the class field, which sounded backwards to me. But I realize that the `kwargs` come from the class `to_simulation_data` method, in which case it does make sense that they would take precedence.

tidy3d/plugins/invdes/design.py
momchil-flex: Yeah this does seem to work.

tidy3d/plugins/invdes/design.py
yaugenst-flex: yeah makes sense. generally my thinking was that explicit parameters (function arguments) should take precedence over implicit ones. as a user I would intuitively expect that

PR #2007, opened 2024-10-07
Added a fix to ensure remote folder is created before running batch o…

prashkh:
…f mode solver simulations.

One user reported that running batch solver for mode solver created multiple folders with the same name on our web platform. We figured out a way to fix this by creating the folder explicitly first before running the async call for batch solver.  Last time I had asked the user to just implement in their frontend code and it works so we just need to include this before the next release. Thank you!

### DIFF
--- a/tidy3d/web/api/mode.py
+++ b/tidy3d/web/api/mode.py
@@ -231,6 +231,10 @@ def run_batch(
 
     if verbose:
         console.log(f"[cyan]Running a batch of [deep_pink4]{num_mode_solvers} mode solvers.\n")
+
+        # Create the common folder before running the parallel computation
+        _ = Folder.create(folder_name=folder_name)
+
         with Progress(console=console) as progress:
             pbar = progress.add_task("Status:", total=num_mode_solvers)
             results = Parallel(n_jobs=max_workers, backend="threading")(


tidy3d/web/api/mode.py
```diff
@@ -231,6 +231,12 @@ def handle_mode_solver(index, progress, pbar):
 
     if verbose:
         console.log(f"[cyan]Running a batch of [deep_pink4]{num_mode_solvers} mode solvers.\n")
+
+        # Create the folder before running the parallel computation and wait one second before running the batch simulation job
+        folder = Folder.create(folder_name=folder_name)
+        import time
+        time.sleep(1)
```
momchil-flex: Put `from time import sleep` at the top level of this file and use `sleep(1)` here.

But separately, @magiWei , is there a better way to wait to make sure that the folder is created? This seems too hacky.

QingengWei: The current way to create a folder is https synchronous interaction, the interface will respond to success or failure. Therefore, getting the interface returned will let you know the interface created by the folder.

momchil-flex: You mean there should be no need for a `time.sleep()` even in the current code? @prashkh do you know why you had to add it?

prashkh:
@momchil-flex if you don't explicitly add this, multiple folders with the same name is created on our web GUI and the batch simulations are randomly distributed across those folders with the same name (puzzling) when solving a batch of mode solver simulations. 
The way I was able to fix this is by creating the folder explicitly first before running the async call for batch solver.

momchil-flex:
So I can see why that would happen without the explicit folder create call, but I'm wondering why you still need to wait 1s. Generally it sounds like `folder = Folder.create(folder_name=folder_name)` should only return when the folder is created. Did you observe otherwise?

prashkh:
I don't remember testing with no delay. I can do that today. My assumption was that the run_batch already calls run and each run function creates a folder (`ModeSolverTask.create(...)`), if there is none. So, if there is no delay between explicitly creating the folder and running the batch simulation, it would again have the same problems. But I can definitely test this today.  I used 1 sec delay just to be safe since in the `run` function I saw 0.5 sec delay being used.

```
    # Wait for task to finish
    prev_status = "draft"
    status = task.status
    while status not in ("success", "error", "diverged", "deleted"):
        if status != prev_status:
            log.log(log_level, f"Mode solver status: {status}")
            if verbose:
                console.log(f"Mode solver status: {status}")
            prev_status = status
        time.sleep(0.5)
        status = task.get_info().status
```

momchil-flex:
So I think what you say makes sense but the point is that you used to run the processes async and only then each process would try to make a folder. Then, multiple calls to the api will happen and it can be arbitrary when some folders will get made and some processes will start seeing them.

But now you call the folder creation *before* you start the async run, and since the folder creation itself is sync, I would think you don't need an explicit `sleep`.

prashkh:
@momchil-flex you were right! Just adding the `folder = Folder.create(folder_name=folder_name)` fixed it. No need to add time delay :)

momchil-flex:
Great! Could you update the PR?

momchil-flex:
I guess I'll do it myself and merge.

prashkh:
Thank you!

PR #2008, opened 2024-10-09
Convert several mode solver methods to staticmethods

weiliangjin2021:
The goal of this PR is to convert all methods involved in `_solve_single_freq` to staticmethods, so that calling `_solve_single_freq` in a multiprocessing pool will not serialize unrelated data in mode solver. This is very helpful when the simulation field in mode solver is huge.

### DIFF
--- a/tidy3d/plugins/mode/mode_solver.py
+++ b/tidy3d/plugins/mode/mode_solver.py
@@ -579,15 +579,16 @@ class ModeSolver(Tidy3dBaseModel):
         ]
         return np.stack(eps_tensor, axis=0)
 
+    @staticmethod
     def _tensorial_material_profile_modal_plane_tranform(
-        self, mat_data: ArrayComplex4D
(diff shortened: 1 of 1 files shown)


tidy3d/plugins/mode/mode_solver.py
```diff
@@ -579,27 +579,28 @@ def _get_epsilon(self, freq: float) -> ArrayComplex4D:
         ]
         return np.stack(eps_tensor, axis=0)
 
+    @staticmethod
     def _tensorial_material_profile_modal_plane_tranform(
-        self, mat_data: ArrayComplex4D
+        mat_data: ArrayComplex4D, normal_axis
```
tylerflex: type annotation for normal axis?

tylerflex (approved):
looks good to me

momchil-flex (approved):

PR #2014, opened 2024-10-11
Replace infinite coordinates for shapely operation

lucas-flexcompute:
This fixes the bug described in https://github.com/flexcompute/tidy3d-core/issues/726

The added test is the minimal failure case I could find, but to be honest, I'm not 100% sure why it fails in this instance when other shapes with infinite coordinates worked fine in shapely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+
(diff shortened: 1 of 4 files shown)


weiliangjin2021 (approved):

momchil-flex:
Thanks, just rebased to add this to the 2.7.5 changelog and fix one linting issue. By the way the pre-commit hook (`pre-commit install`) is very useful for avoiding these linting test failures.

PR #2015, opened 2024-10-11
efficiency improvements in autograd derivatives

tylerflex:
Orders of magnitude improvement to autograd gradient handling, especially for large field data and many structures.

Achieved through following:
1.  assuming sorted in field data interpolation for adjoint gradients.
2. not deep copying field data containing DerivativeInfo objects when passing around.

Still work in progress as I need to figure out where else I could apply these fixes.


yaugenst-flex (approved):
This is great! I guess the only thing I wonder is that `derivative_info` contents are never mutated? Basically after this we need to make sure that never happens, but seems like that shouldn't be the case anyways?

tylerflex:
it should never be the case. the derivative info is basically just a container to avoid having to pass Ex_fwd, Ey_fwd, etc into all of the derivative computation methods and then interpolate into them. We'd never mutate them

PR #2016, opened 2024-10-11
fix broken examples link in readme

tylerflex:


daquinteroflex (approved):
Thanks! I've done a manual check of all the remaining links and this one is the only one.

 I was recently chatting with MC about improving our broken links actually on the docs and I understand they run checks on that website so should be fine.

PR #2018, opened 2024-10-14
Add more robust gauge choice for high-order modes

lucas-flexcompute:
I'm porting this gauge fix from PhotonForge, where it's been working for a while now. It works for 2D and 1D mode domains and should not change the current gauge used for fundamental modes.

I'm attaching a notebook that more easily illustrates the problem. Just compare the results with and without this fix:
[gauge_test.zip](https://github.com/user-attachments/files/17365524/gauge_test.zip)

### DIFF
--- a/tidy3d/plugins/mode/mode_solver.py
+++ b/tidy3d/plugins/mode/mode_solver.py
@@ -41,6 +41,7 @@ from ...components.types import (
     ArrayComplex3D,
     ArrayComplex4D,
     ArrayFloat1D,
+    ArrayFloat2D,
     Ax,
     Axis,
     Axis2D,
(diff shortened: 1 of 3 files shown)


weiliangjin2021 (approved):

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -770,16 +775,59 @@ def _solve_single_freq_relative(
...
 
     def _rotate_field_coords(self, field: FIELD) -> FIELD:
         """Move the propagation axis=z to the proper order in the array."""
         f_x, f_y, f_z = np.moveaxis(field, source=3, destination=1 + self.normal_axis)
         return np.stack(self.plane.unpop_axis(f_z, (f_x, f_y), axis=self.normal_axis), axis=0)
 
+    @staticmethod
+    def _weighted_coord_max(
+        array: ArrayFloat2D, u: ArrayFloat1D, v: ArrayFloat1D
+    ) -> Tuple[int, int]:
```
momchil-flex, line 1269: Could you add a docstring

momchil-flex (approved):
Sorry, this had fallen through the cracks, but I think it looks good and would like to get it in. Just needs a rebase.

PR #2020, opened 2024-10-16
more polyslab gradient bug fixes

tylerflex:
Some minor but important polyslab gradient fixes. Surprised I missed this but glad I ended up catching it after all. Goes to show how tricky these things can be.

1. Found a sign error in the gradient contribution. This was not really affecting anything except in certain polarizations, it turned out to improve a customers' taper gradient accuracy, and fixed one of my numerical tests under certain conditions.

2. Also realized some condition for incorporating the slab_bounds into the area calculation was not being included as it was out of date with another PR. So after this fix + another tweak to the gradient norm, I am seeing POlySlab gradients matching numerical not only better than before but with the right magnitude.

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -2974,7 +2974,7 @@ class DiffractionData(AbstractFieldProjectionData):
         theta_data, phi_data = self.angles
         angle_sel_kwargs = dict(orders_x=int(order_x), orders_y=int(order_y), f=float(freq0))
         angle_theta = float(theta_data.sel(**angle_sel_kwargs))
-        angle_phi = 0 * np.pi + float(phi_data.sel(**angle_sel_kwargs))
+        angle_phi = float(phi_data.sel(**angle_sel_kwargs))
 
         # if the angle is nan, this amplitude is set to 0 in the fwd pass, so should skip adj
         if np.isnan(angle_theta):
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -1037,7 +1037,7 @@ class FieldData(FieldDataset, ElectromagneticFieldData):
 
             for freq0 in field_component.coords["f"]:
                 omega0 = 2 * np.pi * freq0
-                scaling_factor = 1 / (MU_0 * omega0)
+                scaling_factor = 33 / (MU_0 * omega0)
 
                 forward_amp = self.get_amplitude(field_component.sel(f=freq0))
 
@@ -1076,7 +1076,7 @@ class FieldData(FieldDataset, ElectromagneticFieldData):
             for name, field_component in self.field_components.items():
                 # get the VJP values at frequency and apply adjoint phase
                 field_component = field_component.sel(f=freq0)
-                values = -1j * field_component.values
+                values = 2 * -1j * field_component.values
 
                 # make source go backwards
                 if "H" in name:
@@ -2974,7 +2974,7 @@ class DiffractionData(AbstractFieldProjectionData):
         theta_data, phi_data = self.angles
         angle_sel_kwargs = dict(orders_x=int(order_x), orders_y=int(order_y), f=float(freq0))
         angle_theta = float(theta_data.sel(**angle_sel_kwargs))
-        angle_phi = np.pi + float(phi_data.sel(**angle_sel_kwargs))
+        angle_phi = 0 * np.pi + float(phi_data.sel(**angle_sel_kwargs))
 
         # if the angle is nan, this amplitude is set to 0 in the fwd pass, so should skip adj
         if np.isnan(angle_theta):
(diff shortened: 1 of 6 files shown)


tidy3d/components/geometry/polyslab.py
```diff
@@ -1438,12 +1438,9 @@ def compute_derivative_vertices(self, derivative_info: DerivativeInfo) -> Traced
...
         # correction to edge area based on sidewall distance along slab axis
-        dim_axis = "xyz"[self.axis]
-        field_coords_axis = derivative_info.E_der_map[f"E{dim_axis}"].coords[dim_axis]
-        if len(field_coords_axis) > 1:
-            slab_height = abs(float(np.squeeze(np.diff(self.slab_bounds))))
-            if not np.isinf(slab_height):
-                edge_areas *= slab_height
+        slab_height = abs(float(np.squeeze(np.diff(self.slab_bounds))))
+        if not np.isinf(slab_height):
+            edge_areas *= slab_height
```
yaugenst-flex, line 1443: Don't think I fully understand this one. So this would get ignored previously, or what was the issue?

tylerflex: I dont think I fully understand it either, I think the idea was to ignore the slab thickness in some cases where the slab was considered infinitely thin (or the simulation was 2D). but it ended up never getting triggered in 3D because we changed the adjoint field monitor to be planar (in the middle of the slab bounds).

tylerflex:
I tested this a bit more  (as mentioned) and I think it's good to go.

yaugenst-flex (approved):

tylerflex:
OK this is ready, freshly tested in notebooks and all seems good.

The changes:
1. Fixed a sign in the D field contribution to polyslab vertices gradient.
2. Included the `slab_bounds` in the differential area calculation for vertices adjoint, which was not being taken into account because of a condition that no longer was being triggered after earlier changes.
2. Fixed the norm of the FieldMonitor adjoint source for both point and volume monitors.
3. Fixed the phase of the DiffractionMonitor adjoint source, which is consistent with adjoint now and works for all polarizations and layer thicknesses in our notebook.
4. Revert an earlier "fix" to Box epsilon grabbing that made things worse

tidy3d/components/data/monitor_data.py
```diff
@@ -1037,7 +1037,7 @@ def to_adjoint_point_sources(self, fwidth: float) -> List[PointDipole]:
 
             for freq0 in field_component.coords["f"]:
                 omega0 = 2 * np.pi * freq0
-                scaling_factor = 1 / (MU_0 * omega0)
+                scaling_factor = 33 / (MU_0 * omega0)
```
momchil-flex, line 1040: Well that seems pretty arbitrary :D why `33`?

momchil-flex (approved):
If testing shows that the gradients are better, I trust that. It's really hard to make sure that all these phases, etc. are accounted for correctly in any other way but testing.

tidy3d/components/data/monitor_data.py
tylerflex: the honest answer is because I tested the PointDipole sources against every structure combination and the gradients were consistently 33x smaller than numerical ...

momchil-flex: A little worrying then. Have you tested different grid sizes at the monitor location? And the monitor being in different materials?

tylerflex: It's hard to test everything, I already had like a 10 x 4 grid of different combinations of things to test, but no I did not test grid size and medium. The grid size should be handled by using `interp` in the source, the medium could have an effect but I think it would be some scaling by the refractive index, the overall constant might still be the same. I can look into it some more if it would be useful

momchil-flex: Yeah I'm just trying to think about things that this "constant" might actually depend on, like the mesh step around the dipole for some reason, or the material epsilon.

tylerflex: it's very possible the gradients depend on these things in some way. This PR resolve the gradient direction and corrects the overall normalization so that the un-normalized gradients match pretty well (typically under 10% unnormalized and around 1% normalized), but under my test conditions. So There's another dimension to explore which is how these norms change under different background permittivity and grid resolution.. which gets pretty complicated. I think for sake of practical optimization what we have here is a good improvement, but I'll do a few more tests today to see if I can find some low hanging fruit to improve it

tylerflex: I tested a few different Simulation.medium and also a few different grid sizes. The grad norms were constant across each, maybe a tiny bit of variation with dl, but didnt seem linear with dl and  could be due to many things. Ultimately feel confident that these constants are good to give adj ~ numerical grads without normalization for a range of situations, and if not, we can revisit later. So I'll go ahead and merge this

PR #2022, opened 2024-10-17
Skip geometry validation in simulation subsection and mode solver reduction

weiliangjin2021:
This is part of efforts in speeding up mode solver when a geometry group contains large number of geometries. (Last piece in resolving https://github.com/flexcompute/tidy3d-core/issues/692)

- Add options in `updated_copy` to skip valiation.
- The geometry is validated when the simulation object is initialized, and they are not modified in subsection. Thus we can skip validators that are directly associated with geometry.
- For validators of other simulation fields associated with geometry, only the bounding box of the geometry matters. So to validate other simulation fields, we use an auxiliary simulation object whose geometry is replaced by its bounding box.

### DIFF
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -210,13 +210,17 @@ class Tidy3dBaseModel(pydantic.BaseModel):
         "by calling ``obj.json()``.",
     )
 
-    def copy(self, deep: bool = True, **kwargs) -> Tidy3dBaseModel:
-        """Copy a Tidy3dBaseModel.  With ``deep=True`` as default."""
+    def copy(self, deep: bool = True, validate: bool = True, **kwargs) -> Tidy3dBaseModel:
+        """Copy a Tidy3dBaseModel.  With ``deep=True`` and ``validate=True`` as default."""
         kwargs.update(deep=deep)
         new_copy = pydantic.BaseModel.copy(self, **kwargs)
-        return self.validate(new_copy.dict())
+        if validate:
+            return self.validate(new_copy.dict())
+        return new_copy
 
-    def updated_copy(self, path: str = None, deep: bool = True, **kwargs) -> Tidy3dBaseModel:
+    def updated_copy(
+        self, path: str = None, deep: bool = True, validate: bool = True, **kwargs
+    ) -> Tidy3dBaseModel:
         """Make copy of a component instance with ``**kwargs`` indicating updated field values.
 
         Note
@@ -231,7 +235,7 @@ class Tidy3dBaseModel(pydantic.BaseModel):
         """
 
         if not path:
-            return self._updated_copy(**kwargs, deep=deep)
+            return self._updated_copy(**kwargs, deep=deep, validate=validate)
 
         path_components = path.split("/")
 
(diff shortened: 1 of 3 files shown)


weiliangjin2021:
Total mode computation time in a test:
- develop branch: 1000s 
- pre/2.8 with recent parallel computing changes: 600s
- pre/2.8 + this PR: 200s

tidy3d/components/base.py
```diff
@@ -210,13 +210,17 @@ def _special_characters_not_in_name(cls, values):
         "by calling ``obj.json()``.",
     )
 
-    def copy(self, deep: bool = True, **kwargs) -> Tidy3dBaseModel:
-        """Copy a Tidy3dBaseModel.  With ``deep=True`` as default."""
+    def copy(self, deep: bool = True, validate=True, **kwargs) -> Tidy3dBaseModel:
```
tylerflex: add type annotation (bool)

tidy3d/components/simulation.py
```diff
@@ -1401,6 +1403,12 @@ def subsection(
         include_pml_cells : bool = False
             Keep PML cells in simulation boundaries. Note that retained PML cells will be converted
             to regular cells, and the simulation domain boundary will be moved accordingly.
+        skip_geometry_validation: bool = False
+            Skip geometry validation in generating new simulation object, since geometry is not modified
```
tylerflex: I'm still a bit unclear about this option. maybe something like:

"Skip validation for the geometries in the resulting simulation object. Simulation validators remain but only use the bounding box of the existing geometries. Used internally."

or something?

tidy3d/components/simulation.py
```diff
@@ -1401,6 +1403,12 @@ def subsection(
         include_pml_cells : bool = False
             Keep PML cells in simulation boundaries. Note that retained PML cells will be converted
             to regular cells, and the simulation domain boundary will be moved accordingly.
+        skip_geometry_validation: bool = False
```
tylerflex: maybe `validate_geometries : bool = True` would be more straightforward in the code and as an API?

tidy3d/components/simulation.py
```diff
@@ -1401,6 +1403,12 @@ def subsection(
...
+            Skip geometry validation in generating new simulation object, since geometry is not modified
+            in this process. Other validators remain, and assuming that they at most interact with
+            the bounding box of the geometry.
+        deep_copy: bool = True
```
tylerflex, line 1410: this option is only used in the mode solver internals. but I guess is it possible we just always pass `deep=False` in the subsection method?

Basically, when would we want `deep=True` vs `deep=False` and can we just simplify it?

tylerflex (commented):
HI Weiliang, a few higher level comments about the subsection options.

The new validate option looks good to me!

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -1564,9 +1564,21 @@ def reduced_simulation_copy(self):
...
+        structures = [
+            strc.updated_copy(geometry=strc.geometry.bounding_box, deep=False)
+            for strc in new_sim.structures
+        ]
+        # skip validation as it's validated already in subsection
+        aux_new_sim = new_sim.updated_copy(structures=structures, deep=False, validate=False)
+        # validate mode solver here where geometry is replaced by its bounding box
+        new_mode = self.updated_copy(simulation=aux_new_sim, deep=False)
+        # return full mode solver and skip validation
+        return new_mode.updated_copy(simulation=new_sim, deep=False, validate=False)
```
momchil-flex, line 1581: I'm pretty happy with this, I mean it definitely feels somewhat hacky but also it achieves the performance boost we want while still doing just about as little of validation skipping as possible.

Agree with @tylerflex 's comments otherwise.

momchil-flex (approved):

tidy3d/components/simulation.py
weiliangjin2021: While so far subsection is only used in mode solver internals, I feel it's a useful feature. So it's possible to be used elsewhere in the future. In that case, better to leave this option open?

tylerflex: sure, but my question is rather: when we use it in the future, would we generally just use `deep=False`? or are there instances where we might want to deep copy?

weiliangjin2021: the same reason for default `deep=True` in `updated_copy`, if you want a subsection object that can be edited completely separate from the original simulation?

tylerflex: ok sounds good then

tidy3d/components/simulation.py
weiliangjin2021: Sounds good, revised as:
> validate_geometries: bool = True
>             If ``False``, skip validation for the geometries in the resulting simulation object.
>             Simulation validators remain but only use the bounding box of the existing geometries.
>             Used internally.

tylerflex (approved):

dbochkov-flexcompute (approved):

PR #2023, opened 2024-10-18
Mode solver improvement for simulations at low-frequency containing metals

weiliangjin2021:
Two main improvements:

- For the eigenvalue problem with the matrix `PQ` where `P = p_mu + p_partial` and `Q = q_ep + q_partial`, where `p_mu` denotes the matrix containing `mu_parallel`, and `q_ep` for `epsilon_parallel`, and the `partial` term for the remaining matrix components that contains `Grad*ep(mu)^-1*Grad`. One observation is that if you write out the block form, one can find that `p_partial @ q_partial=0`. Since `Grad` scales `1/(k0* dl)`, in low-frequency/high-resolution regime, `Grad` can be huge, not mention that `p(q)_partial` scales as `Grad**2`. So even though `p_partial @ q_partial` is strictly 0, machine error can generate nonzero terms that completely ruin the eigenvalues of matrix `PQ`. This is responsible for the oscillating phenomenon of effective n as a function of frequency in the low-frequency regime.

  In this PR, we don't compute `PQ` with the fully assembled matrix `P` and `Q`, but `PQ = p_mu @ Q + p_partial @ q_ep` so that `p_partial @ q_partial` is for sure dropped.

- Let's take matrix `P` as an illustration (the same applies to `Q`), `p_partial` scales as `1/(k0* dl)**2`. So in low-frequency/high-resolution regime, `p_partial` term might grow so huge that `p_mu` term becomes negligible. But for PEC, its permittivity value should still be significant. So our default permittivity value for PEC should be scaled by the factor  `1/(k0* dl)**2`.

(Tests to be added in the solver part in a separate PR.)

### DIFF
--- a/tidy3d/plugins/mode/solver.py
+++ b/tidy3d/plugins/mode/solver.py
@@ -24,6 +24,9 @@ TOL_TENSORIAL = 1e-6
 TARGET_SHIFT = 10 * fp_eps
 # Preconditioner: "Jacobi" or "Material" based
 PRECONDITIONER = "Material"
+# PEC permittivity cut-off value. Let it be as large as possible so long as not causing overflow in
+# double precision. This value is very heuristic.
+PEC_CUT_OFF = 1e70
 
 
 class EigSolver(Tidy3dBaseModel):
@@ -325,11 +328,18 @@ class EigSolver(Tidy3dBaseModel):
             Permittivity characterization on the mode solver's plane.
         """
 
+        # In the matrices P and Q below, they contain terms ``epsilon_parallel`` or
+        # ``mu_parallel``, and also a term proportional to 1/(k0 * dl)**2. To make sure
+        # that permittivity of PEC is visible in low-frequency/high resolution, pec_val should be
+        # scaled by a factor max(1, max(1/k0 dl) **2).
+        pec_scaling = max(1, max([np.max(abs(f)) for f in der_mats]) ** 2)
+        pec_scaled_val = min(PEC_CUT_OFF, pec_scaling * abs(pec_val))
+
         # use a high-conductivity model for locations associated with a PEC
         def conductivity_model_for_pec(eps, threshold=0.9 * pec_val):
             """PEC entries associated with 'eps' are converted to a high-conductivity model."""
             eps = eps.astype(complex)
-            eps[eps <= threshold] = 1 + 1j * np.abs(pec_val)
+            eps[np.abs(eps) >= abs(threshold)] = 1 + 1j * pec_scaled_val
             return eps
 
         eps_tensor = conductivity_model_for_pec(eps_tensor)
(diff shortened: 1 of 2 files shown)


weiliangjin2021:
Here is a benchmark example from @dmarek-flex regarding a micostrip made of PEC. Its effective k should be 0, and effective n approaches a constant around 1.82 as frequency goes to 0. `double` precision is used.

- Before this PR: significant oscillation in low-frequency region and effective k is huge.
image

- Just scaling `pec_val`, equally bad.
image

- Just dropping `p_partial @ q_partial`, the curve becomes smooth, but incorrect
image

- With both: 
image

weiliangjin2021:
@prashkh welcome to test if you have other tests in mind

tidy3d/plugins/mode/solver.py
```diff
@@ -325,11 +328,18 @@ def solver_em(
...
+        # scaled by a factor max(1, max(1/k0 dl) **2).
+        pec_scaling = max(1, max([np.max(abs(f)) for f in der_mats]) ** 2)
+        pec_scaled_val = min(PEC_CUT_OFF, pec_scaling * abs(pec_val))
+
         # use a high-conductivity model for locations associated with a PEC
         def conductivity_model_for_pec(eps, threshold=0.9 * pec_val):
             """PEC entries associated with 'eps' are converted to a high-conductivity model."""
             eps = eps.astype(complex)
-            eps[eps <= threshold] = 1 + 1j * np.abs(pec_val)
+            eps[eps <= threshold] = 1 + 1j * pec_scaled_val
```
lucas-flexcompute: With `eps` being complex, I believe this inequality uses only the real part, correct? I'd be explicit about it just to make sure: `eps[eps.real <= threshold] = ...`

dmarek-flex (approved):
Results speak for themselves. Thanks @weiliangjin2021, I tried it out on my own example as well and it worked well down to 1 MHz, which is plenty low enough.

First time I get to remind someone to update the Changelog though!

tidy3d/plugins/mode/solver.py
weiliangjin2021: This is historical. I asked the same question when reviewing those lines last year. So it is quite confusing. I'll just change it to absolution value inequality.

weiliangjin2021:
> Results speak for themselves. Thanks @weiliangjin2021, I tried it out on my own example as well and it worked well down to 1 MHz, which is plenty low enough.

Well, soon enough we might need to handle <1MHz. But that will probably need a very different formulation.

> First time I get to remind someone to update the Changelog though!

Thanks for the reminder. I plan to add it after 1st round of review so that I'm confident enough to say that it works well in low-frequency.

lucas-flexcompute (approved):

weiliangjin2021:
@dmarek-flex CHANGELOG added.

PR #2025, opened 2024-10-18
Autograd Dataarray rework

### LINKED ISSUE #1940: FAQ section for autograd readme

yaugenst-flex:
Mostly working, needs more testing and cleanup

Keeping track of failing tests:
 - [x] `tests/test_components/test_custom.py::test_custom_sellmeier` (???)
 - [x] `tests/test_components/test_structure.py::test_custom_medium_to_gds`
 - [x] `tests/test_components/test_structure.py::test_lower_dimension_custom_medium_to_gds`
 - [x] `tests/test_plugins/test_adjoint.py::test_adjoint_pipeline_2d`
 - [x] `tests/test_plugins/test_adjoint.py::test_to_gds`
 - [x] `tests/test_plugins/test_microwave.py::test_voltage_integral_toggles`
 - [x] `tests/test_plugins/test_microwave.py::test_current_integral_toggles`
 - [x] `tests/test_plugins/test_microwave.py::test_tiny_voltage_path`
 - [x] `tests/test_plugins/test_microwave.py::test_impedance_accuracy`
 - [x] `tests/test_plugins/test_terminal_component_modeler.py::test_run_component_modeler`
 - [x] `tests/test_plugins/test_terminal_component_modeler.py::test_run_coaxial_component_modeler`
 - [x] `tests/test_components/test_field_projection.py::test_proj_clientside`
 - [x] `tests/test_components/test_field_projection.py::test_2d_proj_clientside`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Minor gradient direction and normalization fixes for polyslab, field monitors, and diffraction monitors in autograd.
 - Resolved an issue where temporary files for adjoint simulations were not being deleted properly.
+- Autograd functions can now be called directly on `DataArray` (e.g., `np.sum(data_array)`) in objective functions.
+
+### Changed
+- Improved autograd tracer handling in `DataArray`, resulting in significant speedups for differentiation involving large monitors.
+
+### Fixed
+- Resolve several edge cases where autograd boxes were incorrectly converted to numpy arrays.
+
 
 ## [2.7.5] - 2024-10-16
 
--- a/tidy3d/components/autograd/boxes.py
+++ b/tidy3d/components/autograd/boxes.py
@@ -1,0 +1,147 @@
+# Adds some functionality to the autograd arraybox
+# NOTE: this is not a subclass of ArrayBox since that would break autograd's internal checks
+
+import importlib
+from typing import Any, Callable, Dict, List, Tuple
+
+import autograd.numpy as anp
+from autograd.numpy.numpy_boxes import ArrayBox
+
+TidyArrayBox = ArrayBox  # NOT a subclass
+
+_autograd_module_cache = {}  # cache for imported autograd modules
+
+
+@classmethod
+def from_arraybox(cls, box: ArrayBox) -> TidyArrayBox:
+    """Construct a TidyArrayBox from an ArrayBox."""
+    return cls(box._value, box._trace, box._node)
+
+
+def __array_function__(
+    self: Any,
+    func: Callable,
+    types: List[Any],
+    args: Tuple[Any, ...],
+    kwargs: Dict[str, Any],
+) -> Any:
+    """
+    Handle the dispatch of NumPy functions to autograd's numpy implementation.
+
+    Parameters
+    ----------
+    self : Any
+        The instance of the class.
+    func : Callable
+        The NumPy function being called.
+    types : List[Any]
+        The types of the arguments that implement __array_function__.
+    args : Tuple[Any, ...]
+        The positional arguments to the function.
+    kwargs : Dict[str, Any]
+        The keyword arguments to the function.
+
+    Returns
+    -------
+    Any
+        The result of the function call, or NotImplemented.
+
+    Raises
+    ------
+    NotImplementedError
+        If the function is not implemented in autograd.numpy.
+
+    See Also
+    --------
+    https://numpy.org/doc/stable/reference/arrays.classes.html#numpy.class.__array_function__
+    """
+    if not all(t in TidyArrayBox.type_mappings for t in types):
+        return NotImplemented
+
+    module_name = func.__module__
+
+    if module_name.startswith("numpy"):
+        anp_module_name = "autograd." + module_name
+    else:
+        return NotImplemented
+
+    # Use the cached module if available
+    anp_module = _autograd_module_cache.get(anp_module_name)
+    if anp_module is None:
+        try:
+            anp_module = importlib.import_module(anp_module_name)
+            _autograd_module_cache[anp_module_name] = anp_module
+        except ImportError:
+            return NotImplemented
+
+    f = getattr(anp_module, func.__name__, None)
+    if f is None:
+        return NotImplemented
+
+    if f.__name__ == "nanmean":  # somehow xarray always dispatches to nanmean
+        f = anp.mean
+        kwargs.pop("dtype", None)  # autograd mean vjp doesn't support dtype
+
+    return f(*args, **kwargs)
+
+
+def __array_ufunc__(
+    self: Any,
+    ufunc: Callable,
+    method: str,
+    *inputs: Any,
+    **kwargs: Dict[str, Any],
+) -> Any:
+    """
+    Handle the dispatch of NumPy ufuncs to autograd's numpy implementation.
+
+    Parameters
+    ----------
+    self : Any
+        The instance of the class.
+    ufunc : Callable
+        The universal function being called.
+    method : str
+        The method of the ufunc being called.
+    inputs : Any
+        The input arguments to the ufunc.
+    kwargs : Dict[str, Any]
+        The keyword arguments to the ufunc.
+
+    Returns
+    -------
+    Any
+        The result of the ufunc call, or NotImplemented.
+
+    See Also
+    --------
+    https://numpy.org/doc/stable/reference/arrays.classes.html#numpy.class.__array_ufunc__
+    """
+    if method != "__call__":
+        return NotImplemented
+
+    ufunc_name = ufunc.__name__
+
+    anp_ufunc = getattr(anp, ufunc_name, None)
+    if anp_ufunc is not None:
+        return anp_ufunc(*inputs, **kwargs)
+
+    return NotImplemented
+
+
+def item(self):
+    if self.size != 1:
+        raise ValueError("Can only convert an array of size 1 to a scalar")
+    return anp.ravel(self)[0]
+
+
+TidyArrayBox._tidy = True
+TidyArrayBox.from_arraybox = from_arraybox
+TidyArrayBox.__array_namespace__ = lambda self, *, api_version=None: anp
+TidyArrayBox.__array_ufunc__ = __array_ufunc__
+TidyArrayBox.__array_function__ = __array_function__
+TidyArrayBox.__repr__ = str
+TidyArrayBox.real = property(anp.real)
+TidyArrayBox.imag = property(anp.imag)
+TidyArrayBox.conj = anp.conj
+TidyArrayBox.item = item
--- a/tidy3d/components/autograd/functions.py
+++ b/tidy3d/components/autograd/functions.py
@@ -132,7 +132,6 @@ def interpn(
         raise ValueError(f"Unsupported interpolation method: {method}")
 
     itrp = RegularGridInterpolator(points, values, method=method)
-    grid = anp.meshgrid(*xi, indexing="ij")
 
     # Prepare the grid for interpolation
     # This step reshapes the grid, checks for NaNs and out-of-bounds values
@@ -142,12 +141,12 @@ def interpn(
     #   - number of dimensions
     #   - boolean array indicating NaN positions
     #   - (discarded) boolean array for out-of-bounds values
-    grid, shape, ndim, nans, _ = itrp._prepare_xi(tuple(grid))
+    xi, shape, ndim, nans, _ = itrp._prepare_xi(xi)
 
     # Find the indices of the grid cells containing the interpolation points
     # and calculate the normalized distances (ranging from 0 at lower grid point to 1
     # at upper grid point) within these cells
-    indices, norm_distances = itrp._find_indices(grid.T)
+    indices, norm_distances = itrp._find_indices(xi.T)
 
     result = interp_fn(indices, norm_distances, values)
     nans = anp.reshape(nans, (-1,) + (1,) * (result.ndim - 1))
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -26,7 +26,7 @@ from ..exceptions import FileError
 from ..log import log
 from .autograd.types import AutogradFieldMap, Box
 from .autograd.utils import get_static
-from .data.data_array import AUTOGRAD_KEY, DATA_ARRAY_MAP, DataArray
+from .data.data_array import DATA_ARRAY_MAP, DataArray
 from .file_util import compress_file_to_gzip, extract_gzip_file
 from .types import TYPE_TAG_STR, ComplexNumber, Literal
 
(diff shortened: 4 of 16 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Improved autograd tracer handling in `DataArray`, resulting in significant speedups for differentiation involving large monitors. Also resolves several edge cases where autograd boxes were incorrectly converted to numpy arrays.
```
tylerflex: this is great, but either here or in separate bullets, we should probably list any API changes that might happen (eg `npa.sum(data_arr)` works, etc.

tidy3d/components/autograd/boxes.py
```diff
@@ -0,0 +1,53 @@
+# Adds some functionality to the autograd arraybox
+# NOTE: this is not a subclass of ArrayBox since that would break autograd's internal checks
+
+import autograd.numpy as anp
+from autograd.numpy.numpy_boxes import ArrayBox
+
+TidyArrayBox = ArrayBox  # NOT a subclass
```
tylerflex, line 10: so, this is interesting.. I guess it is functionally like forking autograd and adding these features in there?

tidy3d/components/autograd/boxes.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def from_arraybox(box: ArrayBox) -> TidyArrayBox:
+    """Convert an autograd arraybox to a tidy arraybox."""
+    return TidyArrayBox(box._value, box._trace, box._node)
+
+
+def __array_function__(self, func, types, args, kwargs):
+    f = getattr(anp, func.__name__, None)
```
tylerflex: docstring / types?

tidy3d/components/autograd/functions.py
```diff
@@ -142,12 +141,13 @@ def interpn(
     #   - number of dimensions
     #   - boolean array indicating NaN positions
     #   - (discarded) boolean array for out-of-bounds values
-    grid, shape, ndim, nans, _ = itrp._prepare_xi(tuple(grid))
+    # _, shape, ndim, nans, _ = itrp._prepare_xi(tuple(grid))
```
tylerflex: remove comment

tidy3d/components/base.py
```diff
@@ -965,27 +965,19 @@ def handle_value(x: Any, path: tuple[str, ...]) -> None:
...
-            elif isinstance(x, DataArray):
-                # try to grab the traced values from the `attrs` (if traced)
-                if AUTOGRAD_KEY in x.attrs:
-                    field_mapping[path] = x.attrs[AUTOGRAD_KEY]
-
-                # or just grab the static value out of the values
-                elif include_untraced_data_arrays:
-                    field_mapping[path] = get_static(x.values)
+            # for data arrays, need to be more careful as their tracers are stored in .data
+            elif isinstance(x, DataArray) and (isbox(x.data) or include_untraced_data_arrays):
```
tylerflex: much nicer..

tidy3d/components/base.py
```diff
@@ -1020,11 +1012,6 @@ def insert_value(x, path: tuple[str, ...], sub_dict: dict):
             sub_element = current_dict[final_key]
             if isinstance(sub_element, DataArray):
                 current_dict[final_key] = sub_element.copy(deep=False, data=x)
-                if isbox(x):
```
tylerflex, line 1023: yay, good riddance

tidy3d/components/base.py
```diff
@@ -1020,11 +1012,6 @@ def insert_value(x, path: tuple[str, ...], sub_dict: dict):
             sub_element = current_dict[final_key]
             if isinstance(sub_element, DataArray):
                 current_dict[final_key] = sub_element.copy(deep=False, data=x)
```
tylerflex, line 1014: is it possible to use another constructor here? or is this still the best one

tidy3d/components/data/data_array.py
```diff
@@ -221,6 +209,22 @@ def __eq__(self, other) -> bool:
                 return False
         return True
 
+    @property
+    def values(self):
+        if isbox(self.data):
```
tylerflex, line 223: docstring / type annotations

tidy3d/components/data/data_array.py
```diff
@@ -321,47 +325,183 @@ def interp(
...
+            return self._ag_interp(coords, method, assume_sorted, kwargs, **coords_kwargs)
 
-            missing_keys = set(coords) - set(self.coords)
-            if missing_keys:
-                raise KeyError(f"Cannot interpolate: {missing_keys} not in coords.")
+        return super().interp(coords, method, assume_sorted, kwargs, **coords_kwargs)
 
-            obj = self if assume_sorted else self.sortby(list(coords.keys()))
+    def _ag_interp(
```
tylerflex, line 354: this method is very dense, is there a way to simplify things?

tidy3d/components/data/data_array.py
```diff
@@ -321,47 +325,183 @@ def interp(
...
+                .difference(coords.keys())
+            )
+            indexers.update({d: ds.variables[d] for d in sdims})
+
+        obj = ds if assume_sorted else ds.sortby(list(coords))
+
+        def maybe_variable(obj, k):
+            try:
+                return obj._variables[k]
+            except KeyError:
```
tylerflex: can we just use `if k in obj._variables`?

tidy3d/components/data/data_array.py
```diff
@@ -321,47 +325,183 @@ def interp(
...
+            sdims = (
+                set(ds.dims)
+                .intersection(*[set(nx.dims) for nx in indexers.values()])
+                .difference(coords.keys())
+            )
+            indexers.update({d: ds.variables[d] for d in sdims})
+
+        obj = ds if assume_sorted else ds.sortby(list(coords))
+
+        def maybe_variable(obj, k):
```
tylerflex: docstring, just to help reader understand what's happening

tidy3d/components/data/data_array.py
```diff
@@ -321,47 +325,183 @@ def interp(
...
+
+        validated_indexers = {k: (maybe_variable(obj, k), v) for k, v in indexers.items()}
+
+        for k, v in validated_indexers.items():
+            obj, newidx = missing._localize(obj, {k: v})
+            validated_indexers[k] = newidx[k]
+
+        variables = {}
+        reindex = False
+        for name, var in obj._variables.items():
```
tylerflex, line 402: did a lot of this get copied from xarray or something?

tidy3d/plugins/autograd/README.md
```diff
@@ -65,17 +73,18 @@ There is also a `numpy` wrapper that can be similarly imported from `autograd.nu
...
+
 ```py
 import autograd.numpy as anp
 anp.sum(...)
-
 ```
 
-`Autograd` supports fewer features than `jax`. So, for example, the `has_aux` option is not supported in the `jax.grad()`, but one can write their own utilities to implement these features, as we show in the notebook examples.
+`Autograd` supports fewer features than `jax`.
+For example, the `has_aux` option is not supported in the `jax.grad()` function, but one can write their own utilities to implement these features, as we show in the notebook examples.
```
tylerflex: maybe this is actually supported with `ag.grad_with_aux`

tidy3d/plugins/autograd/README.md
```diff
@@ -97,61 +106,116 @@ import tidy3d as td
...
-- server-side gradient processing by passing `local_gradient=False` to the `web` functions. This can dramatically cut down on data storage time, just be careful about using this with multi-frequency monitors and large design regions as it can lead to large data storage on our servers.
+    # we can now use these just like regular numpy arrays
+    intensity = anp.abs(Ex) ** 2 + anp.abs(Ey) ** 2 + anp.abs(Ez) ** 2  # sim_data.get_intensity("field_monitor") would also work of course
+    norm_intensity = anp.linalg.norm(intensity)
+
+    return norm_intensity  # no .item() needed
+```
+
+## Feature Roadmap
+
```
tylerflex, line 182: link to the adjoint master plan issue maybe? with just a note that these are the things on our "to-do list" if you want to follow along. also is a good place for comments / discussions?

tidy3d/plugins/autograd/README.md
```diff
@@ -97,61 +106,116 @@ import tidy3d as td
...
+| ----------------------------------------------------------------- | ------------------------------------------------------- |
+| rectangular prisms                                                | `Box.center`, `Box.size`                                |
+| polyslab (including those with dilation or slanted sidewalls)     | `PolySlab.vertices`                                     |
+| regular mediums                                                   | `Medium.permittivity`, `Medium.conductivity`            |
+| spatially varying mediums (for topology optimization mainly)      | `CustomMedium.permittivity`, `CustomMedium.eps_dataset` |
+| groups of geometries with the same medium (for faster processing) | `GeometryGroup.geometries`                              |
+| complex and self-intersecting polyslabs                           | `ComplexPolySlab.vertices`                              |
+| dispersive materials                                              | `PoleResidue.eps_inf`, `PoleResidue.poles`              |
+| spatially dependent dispersive materials                          | `CustomPoleResidue.eps_inf`, `CustomPoleResidue.poles`  |
+
```
tylerflex, line 200: we also do `Cylinder.radius` and `Cylinder.center` (not along the axis dim though)

tidy3d/plugins/autograd/README.md
```diff
@@ -97,61 +106,116 @@ import tidy3d as td
...
+  However, exercise caution when using this feature with multi-frequency monitors and large design regions, as it may result in substantial data storage on our servers.
 
 We currently have the following restrictions:
 
-- Only 500 max structures containing tracers can be added to the `Simulation` to cut down on processing time. To bypass this restriction, use `GeometryGroup` to group structures with the same medium.
-- `web.run_async` for simulations with tracers does not return a `BatchData` but rather a `dict` mapping task name to `SimulationData`. There may be high memory usage with many simulations or a lot of data for each.
+- Only 500 max structures containing tracers can be added to the `Simulation` to cut down on processing time.
+  To bypass this restriction, use `GeometryGroup` to group structures with the same medium.
+- `web.run_async` for simulations with tracers does not return a `BatchData` but rather a `dict` mapping task name to `SimulationData`.
+  There may be high memory usage with many simulations or a lot of data for each.
```
tylerflex, line 228: should we add broadband? Basically:

Tidy3D can handle objective functions over a single simulation under any of the following conditions for the monitors that the objective function output depends on:
1. Several monitors, all withthe same frequency.
2. One monitor with many frequencies where the data is being extracted out of a single coordinate (eg. single mode_index or direction will work, multiple will not).

If your optimization does not fall into one of these categories, you must split it into separate simulations and run them with `web.run_asnyc`. In all cases, the adjoint simulation bandwidth will be the same as the forward simulation. These limitations allow us to avoid the need to use methods that combine all adjoint sources into one simulation, which have the potential to degrade accuracy and increase the run time and cost significantly. That being said, we plan to offer support for more flexible and general broadband adjoint in the future.

tidy3d/plugins/autograd/README.md
```diff
@@ -162,17 +226,17 @@ Next on our roadmap (targeting 2.8 and 2.9, fall 2024) is to support:
 
 ### Finally
 
-If you have feature requests or questions, please feel free to file an issue or discussion on the `tidy3d` front end repository.
+If you have feature requests or questions, please feel free to file an issue or discussion on the `tidy3d` front-end repository.
```
tylerflex: I was going to say to add a link but we're already here, so maybe "this `tidy3d` front-end repository"?

tylerflex (commented):
This looks great! I think the data array handling is very nice, simplifies things a lot. A couple high level questions:
1. Should we just consider forking autograd? Would that simplify anything? and can we do it without forcing users into using our fork? (for example, we do a few modifications of autograd internally (eg when we handle deepcopy) seems like we effectively fork autograd but n a hacky way? I might be missing a subtlety)
2. I'm having a lot of trouble following the interp logic, maybe how helpful for it be for me to go through that and is it from xarray internals?

tidy3d/components/autograd/boxes.py
yaugenst-flex: yeah more or less. it's somewhat hacky and modifies the underlying autograd arraybox too, although that doesn't really matter since we're not breaking anything. but yeah not ideal. an actual fork might still be something to consider...

tidy3d/components/base.py
yaugenst-flex: right :D just noticed the is instance check should still be done on `xr.DataArray` though...

tidy3d/components/base.py
yaugenst-flex: we can use the default constructor too, but I think copy is best since this preserves everything about the original object.

tidy3d/components/data/data_array.py
yaugenst-flex: not quite because it's a workaround to get a variable for a dimension without a coordinate.. xarray indexing is tricky

yaugenst-flex: but I can refactor it a bit, the function is not really necessary

tidy3d/components/data/data_array.py
yaugenst-flex: I wouldn't really know how 😬 especially since none of this is reused anywhere

tidy3d/components/data/data_array.py
yaugenst-flex: yes 😬 I'll explain a bit in the top-level comment

yaugenst-flex:
> Should we just consider forking autograd? Would that simplify anything? and can we do it without forcing users into using our fork? (for example, we do a few modifications of autograd internally (eg when we handle deepcopy) seems like we effectively fork autograd but in a hacky way? I might be missing a subtlety)

Yeah that would probably be a good idea, since we could move much faster with specific features that we need. We can't really do it without making people use our fork though. If it's a separate repository and a dependency of tidy3d though, it'd always get installed anyways?

> I'm having a lot of trouble following the interp logic, maybe how helpful for it be for me to go through that and is it from xarray internals?

Yeah a lot, or even most, of it is from xarray internals, so probably not super important to go through it. The main difficulty is that we need to make `DataArray.interp` hook into our own differentiable interp implementation while still matching coordinate/dimension handling of vanilla xarray. The latter is pretty involved as you can see, and it's complicated by the fact that most of `xr.DataArray.interp` is implemented in `xr.Dataset`, which we don't subclass, and then `xr.Dataset.interp` itself makes use of a lot of xarray internals until it reaches the actual interpolation function at some point. So basically my approach was to follow this logic as closely as possible, while keeping it within `DataArray`, replace some things with autograd-compatible versions, and finally use our own interpolation implementation where that would be called within `xarray.missing`.

tidy3d/components/autograd/boxes.py
tylerflex: sounds good. we can reconsider this later

tylerflex:
> We can't really do it without making people use our fork though.

So I guess more specifically, I think it's fine if we require people to install our fork (via our dependencies). but what I think would be cumbersome is if people needed to `import autograd as ag` from our fork. If we had a fork where we implemented these changes, would users be able to use regular autograd to work with objective functions involving our simulations?

Regarding interp, I think that's fine, I'm just worried about some major headaches if xarray or scipy change things down the line. Do you think it's unavoidable?

yaugenst-flex:
> If we had a fork where we implemented these changes, would users be able to use regular autograd to work with objective functions involving our simulations?

I don't think I'm following. If we fork autograd and implement additional functionality there, then _that_ would be a separate autograd version with things that the parent package does not support. So regular autograd would not work. If we just want to attach some things to regular autograd like we do here, then we can keep it here. But this approach is quite limited, i.e. I don't see how we could do an xarray wrapper for example.

> Regarding interp, I think that's fine, I'm just worried about some major headaches if xarray or scipy change things down the line. Do you think it's unavoidable?

Yeah we would need to test newer versions and fix things as we go. Don't think that's avoidable, but generally things might break if one of our dependencies decides to break them, so I don't think this is really any different

tidy3d/components/data/data_array.py
tylerflex: ok fair enough

tylerflex:
ok if regular autograd would not work then let's just keep this approach. Honestly other than the comments I already said, I think everything looks good and I don't have any complaints. I'll approve but let me know if you want me to take a detailed look at something

tylerflex (approved):

PR #2028, opened 2024-10-22
automatic mode rotation for bent waveguide

QimingFlex:
@momchil-flex This is a very initial draft and not ready for review yet, but could you take a quick look at the code structure to see if there are any major mistakes so I can address them early on? I added a field `bend_correction` in mode_spec, and the process for mode rotation is mostly in `data_raw` in `ModeSolver`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Differential operators `grad` and `value_and_grad` in `tidy3d.plugins.autograd` that behave similarly to the autograd operators but support auxiliary data via `aux_data=True` as well as differentiation w.r.t. `DataArray`.
 - `@scalar_objective` decorator in `tidy3d.plugins.autograd` that wraps objective functions to ensure they return a scalar value and performs additional checks to ensure compatibility of objective functions with autograd. Used by default in `tidy3d.plugins.autograd.value_and_grad` as well as `tidy3d.plugins.autograd.grad`.
 - Autograd support for simulations without adjoint sources in `run` as well as `run_async`, which will not attempt to run the simulation but instead return zero gradients. This can sometimes occur if the objective function gradient does not depend on some simulations, for example when using `min` or `max` in the objective.
+- `bend_angle_rotation` in `ModeSpec` to improve accuracy in some cases when both `bend_radius` and `angle_theta` are defined. This option is disabled by default but it can be tried when very high accuracy is required in `ModeSource` and `ModeMonitor` objects that are placed in bend and angled waveguides.
 
 ### Changed
 - `CustomMedium` design regions require far less data when performing inverse design by reducing adjoint field monitor size for dims with one pixel.
--- a/tidy3d/components/mode.py
+++ b/tidy3d/components/mode.py
@@ -129,6 +129,18 @@ class ModeSpec(Tidy3dBaseModel):
         "yz plane, the ``bend_axis`` is always 1 (the global z axis).",
     )
 
+    bend_angle_rotation: bool = pd.Field(
+        False,
+        title="Use fields rotation when both bend and angle are defined",
+        description="Defines how modes are computed when both a bend and an angle are defined. "
+        "If `False`, the two coordinate transformations are directly composed. "
+        "If `True`, the structures in the simulation are first rotated to compute a mode solution at "
+        "a reference plane normal to the bend's azimuthal direction. Then, the fields are rotated to align with "
+        "the mode plane, using the `n_eff` calculated at the reference plane. The second option can "
+        "produce more accurate results, but more care must be taken, for example, in ensuring that the "
+        "original mode plane intersects the correct geometries in the simulation with rotated structures.",
+    )
+
     track_freq: Union[TrackFreq, None] = pd.Field(
         "central",
         title="Mode Tracking Frequency",
--- a/tidy3d/plugins/mode/mode_solver.py
+++ b/tidy3d/plugins/mode/mode_solver.py
@@ -19,6 +19,7 @@ from ...components.boundary import PML, Absorber, Boundary, BoundarySpec, PECBou
 from ...components.data.data_array import (
     FreqModeDataArray,
     ModeIndexDataArray,
+    ScalarModeFieldCylindricalDataArray,
     ScalarModeFieldDataArray,
 )
 from ...components.data.monitor_data import ModeSolverData
@@ -30,8 +31,10 @@ from ...components.grid.grid import Grid
 from ...components.medium import FullyAnisotropicMedium, LossyMetalMedium
 from ...components.mode import ModeSpec
 from ...components.monitor import ModeMonitor, ModeSolverMonitor
+from ...components.scene import Scene
 from ...components.simulation import Simulation
 from ...components.source import ModeSource, SourceTime
+from ...components.structure import Structure
 from ...components.subpixel_spec import SurfaceImpedance
 from ...components.types import (
     TYPE_TAG_STR,
@@ -40,6 +43,7 @@ from ...components.types import (
     ArrayFloat1D,
     Ax,
     Axis,
+    Axis2D,
     Direction,
     EpsSpecType,
     FreqArray,
@@ -195,6 +199,13 @@ class ModeSolver(Tidy3dBaseModel):
         return plane_center
 
     @cached_property
+    def normal_axis_2d(self) -> Axis2D:
+        """Axis normal to the mode plane in a 2D plane that is normal to the bend_axis_3d."""
+        _, idx_plane = self.plane.pop_axis((0, 1, 2), axis=self.bend_axis_3d)
+
+        return idx_plane.index(self.normal_axis)
+
+    @cached_property
     def solver_symmetry(self) -> Tuple[Symmetry, Symmetry]:
         """Get symmetry for solver for propagation along self.normal axis."""
         mode_symmetry = list(self.simulation.symmetry)
@@ -333,6 +344,13 @@ class ModeSolver(Tidy3dBaseModel):
         if self.mode_spec.group_index_step > 0:
             return self._get_data_with_group_index()
 
+        if (
+            self.mode_spec.bend_angle_rotation
+            and self.mode_spec.bend_radius is not None
+            and np.abs(self.mode_spec.angle_theta) > 0
+        ):
+            return self.rotated_mode_solver_data
+
         # Compute data on the Yee grid
         mode_solver_data = self._data_on_yee_grid()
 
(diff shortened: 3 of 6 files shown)


tidy3d/components/mode.py
```diff
@@ -127,6 +127,15 @@ class ModeSpec(Tidy3dBaseModel):
...
     )
 
+    bend_correction: bool = pd.Field(
+        False,
+        title="Bend correction",
+        description="Applies a correction by rotating the mode solution from a reference plane to"
+        "the actual mode plane. This is done by first computing the mode solution at a reference"
+        "plane normal to the bend's azimuthal direction, then rotating the field solution to"
+        "align with the mode plane, using the n_eff calculated at the reference plane.",
+    )
```
momchil-flex, line 142: I would try to make both the name of this field and the description a bit clearer. For example, it is not clear that this would only apply to the cases when both a bend and an angle is specificed.

Maybe something like:

```
    bend_angle_rotation: bool = pd.Field(
        False,
        title="Use rotation when both bend and angle are defined",
        description="Defines how modes are computed when both a bend and an angle are defined. If `False`, the two coordinate transformations are directly composed. If `True`, the structures in the simulation are first rotated, to compute a mode solution at a reference plane normal to the bend's azimuthal direction. Then, the fields are rotated to align with the mode plane, using the `n_eff` calculated at the reference plane. The second option can produce more accurate results, but more care must be taken for example in ensuring that the original mode plane intersects the right geometries in the simulation with rotated structures.",
    )
```

tidy3d/components/mode.py
```diff
@@ -160,6 +169,17 @@ def bend_radius_not_zero(cls, val, values):
...
+    @pd.validator("bend_correction", always=True)
+    @skip_if_fields_missing(["bend_radius", "bend_axis"])
+    def validate_bend_correction_requirements(cls, val, values):
+        """Ensure that both ``bend_axis`` and ``bend_radius`` are provided if ``bend_correction`` is enabled."""
+        if val is True:
+            if values.get("bend_axis") is None or values.get("bend_radius") is None:
+                raise SetupError(
+                    "'bend_correction' can only be enabled when both 'bend_axis' and 'bend_radius' are provided."
+                )
+        return val
```
momchil-flex: Is this needed? We could just apply the method if a bend is defined and we don't have to do anything if not.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -317,27 +318,410 @@ def data_raw(self) -> ModeSolverData:
...
+        theta_ref = np.arctan2(
+            monitor_center[coord1] - bend_center[coord1],
+            monitor_center[coord2] - bend_center[coord2],
+        )
+        return theta_ref
+
+    def compute_bent_center(self):
+        bend_axis = self.mode_spec.bend_axis
+        id_inj, id_mnt_plane = self.pop_axis((0, 1, 2), axis=self.size.index(0.0))
+        id_bent = id_mnt_plane[bend_axis]
```
momchil-flex: I think you call `id_bent` here `idx_bend` above in the `data_raw` function, and then you call it `id_bend` in the `rotate_structure` function. Apart from keeping things consistent, it may be better to just gather this in a small property of the ModeSolver, similar to `self.normal_axis`. Note that `id_mnt_plane` are unused anywhere else in these functions, while `id_inj` is just `self.normal_axis`, so a property that just returns the `id_bend` (maybe call it `bend_axis_3d` or something like that) should suffice.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -317,27 +318,410 @@ def data_raw(self) -> ModeSolverData:
...
+            radius = self.mode_spec.bend_radius
+
+            rt_bent_center[self.normal_axis] = self.plane.center[self.normal_axis]
+            rt_bent_center[idx_bend] = self.plane.center[idx_bend]
+            rt_bent_center[id_rotate] = self.plane.center[id_rotate] - radius
+            cylindrical_solver_data_ref = self.process_car_2_cyn_field(
+                solver_data_ref, idx_bend, rt_bent_center
+            )
+
+            # To do: generate coords for the mode monitor
```
momchil-flex: What do you mean by that?

Also as a side note use `TODO:` in the comments, that's the standard keyword we use so it's good to search through the codebase for it, and also many IDEs have special handling for it, like they can show them to you nicely.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -317,27 +318,410 @@ def data_raw(self) -> ModeSolverData:
...
+            if self.mode_spec.track_freq and len(self.freqs) > 1:
+                mode_solver_data = mode_solver_data.overlap_sort(self.mode_spec.track_freq)
 
-        self._field_decay_warning(mode_solver_data.symmetry_expanded)
+            self._field_decay_warning(mode_solver_data.symmetry_expanded)
 
-        return mode_solver_data
+            return mode_solver_data
+        else:
+            # Create a reference solver with rotated geometries for interpolations
```
momchil-flex: Maybe slightly more readable would be to gather everything that is in this "else" into a `self.rotated_mode_data` method and have the logic here be something like

```
if self.mode_spec.bend_correction:
    return self.rotated_mode_data()

# then no change to the rest of the code since you don't need an ``else`` after a ``return``.
```

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -317,27 +318,410 @@ def data_raw(self) -> ModeSolverData:
...
+            rotrated_strucs = self.rotate_structure
+            rotated_sim = self.simulation.updated_copy(structures=rotrated_strucs)
+            rotated_mode_spec = self.mode_spec.updated_copy(
+                bend_correction=False, angle_theta=0, angle_phi=0
+            )
+            solver_ref = self.updated_copy(simulation=rotated_sim, mode_spec=rotated_mode_spec)
```
momchil-flex: I'm thinking we can have this in a separate method too, like `self.rotated_structures_copy`. Then in the documentation we could advise users that they can inspect this mode solver copy to make sure that the rotated simulation looks good.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -317,27 +318,410 @@ def data_raw(self) -> ModeSolverData:
...
+        theta = theta_map.get((id_inj, id_bend), 0)
+
+        # Get the translation values based on id_bent
+        translate_coords = [0, 0, 0]
+        translate_coords[id_bt_u] = mnt_center[id_bt_u]
+        translate_coords[id_bt_v] = mnt_center[id_bt_v]
+        translate_coords = tuple(translate_coords)
+
+        rotated_structures = []
+        for structure in Scene.intersecting_structures(self.plane, self.structures):
```
momchil-flex: It's probably slightly safer to avoid this and to use `self.reduced_simulation_copy` instead. I remember there were some intricacies where structures that are within a pixel's distance, but not intersecting the analytical plane, can have an effect on the computed permittivity.

momchil-flex (commented):
Adding some initial comments, overall the structure makes sense. So is this working currently? We'll obviously need to add a bunch of tests too.

QimingFlex:
> Adding some initial comments, overall the structure makes sense. So is this working currently? We'll obviously need to add a bunch of tests too.

Thanks! I’ve validated the process locally with several cases involving differently oriented bends, and I’ll add those to the tests. This PR isn’t fully working yet, but it’s close. I’ll wrap up the code based on your feedback.

tidy3d/components/mode.py
```diff
@@ -127,6 +127,18 @@ class ModeSpec(Tidy3dBaseModel):
         "yz plane, the ``bend_axis`` is always 1 (the global z axis).",
     )
 
+    bend_angle_rotation: bool = pd.Field(
+        False,
+        title="Use fields rotation when both bend and angle are defined",
+        description="Defines how modes are computed when both a bend and an angle are defined. "
+        " If `False`, the two coordinate transformations are directly composed. If `True`, the "
```
momchil-flex: double space (end of previous line and beginning of this one)

tidy3d/components/mode.py
```diff
@@ -127,6 +127,18 @@ class ModeSpec(Tidy3dBaseModel):
...
+        "structures in the simulation are first rotated, to compute a mode solution at a reference"
+        "plane normal to the bend's azimuthal direction. Then, the fields are rotated to align with"
+        "the mode plane, using the `n_eff` calculated at the reference plane. The second option can"
+        "produce more accurate results, but more care must be taken for example in ensuring that the"
```
momchil-flex: space at the end of all above lines is needed

tidy3d/components/mode.py
momchil-flex: I still think we should just remove this validator

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -339,6 +353,339 @@ def data_raw(self) -> ModeSolverData:
...
+        rotated_simulation = self.simulation.updated_copy(structures=rotated_structures)
+        rotated_mode_spec = self.mode_spec.updated_copy(
+            bend_angle_rotation=False, angle_theta=0, angle_phi=0
+        )
+
+        return self.updated_copy(simulation=rotated_simulation, mode_spec=rotated_mode_spec)
+
+    def _rotate_structure(self) -> List[Structure]:
+        "rotate the structures intersecting with modal plane by angle theta if bend_correction"
+        "is enabeled for bend simulations."
```
momchil-flex: Very minor but better to call it `_rotate_structures` since all structures are rotated. Also, use `"""` comment for function docstrings.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -317,6 +328,9 @@ def data_raw(self) -> ModeSolverData:
         if self.mode_spec.group_index_step > 0:
             return self._get_data_with_group_index()
 
+        if self.mode_spec.bend_angle_rotation:
+            return self.rotated_mode_solver_data
```
momchil-flex, line 352: So this skips everything else that happens in the `data_raw` function. Thinking about it, I think that's fine for everything *except* potentially the normalization. I'm hoping that calling `self._normalize_modes()` on the `rotated_mode_solver_data`, so as to renormalize *after* the interpolation, will fix the TE00 issue. This can be added in the `rotated_mode_solver_data` property.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -339,6 +353,339 @@ def data_raw(self) -> ModeSolverData:
...
+        rotated_bend_center = list(self.plane.center)
+        idx_rotate = 3 - self.normal_axis - self.bend_axis_3d
+        rotated_bend_center[idx_rotate] -= self.mode_spec.bend_radius
+
+        return rotated_bend_center
+
+    def _car_2_cyn(
+        self, mode_solver_data: ModeSolverData
+    ) -> Dict[Union[ScalarModeFieldCylindricalDataArray, ModeIndexDataArray]]:
+        """Convert Cartesian field data to cylindrical coordinates centered the
```
momchil-flex: centered *at* the

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -339,6 +353,339 @@ def data_raw(self) -> ModeSolverData:
...
+        rotated bend center."""
+
+        # Extract coordinate values from one of the six field components
+        pts = [mode_solver_data.Ex[name].values.copy() for name in ["x", "y", "z"]]
```
momchil-flex, line 520: I can't exactly follow what's happening in this function: do results depend on the specific choice to pick Ex?

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -339,6 +353,339 @@ def data_raw(self) -> ModeSolverData:
...
+
+    def _mode_rotation(
+        self,
+        solver_ref_data_cylindrical: Dict[
+            Union[ScalarModeFieldCylindricalDataArray, ModeIndexDataArray]
+        ],
+        solver_temp_data: ModeSolverData,
+    ) -> ModeSolverData:
+        """rotate the mode solver solution from the reference plane in cylindrical coordinates
+        to the practical plane"""
```
momchil-flex: Sorry minor again but generally we try to follow good punctuation, so start with a capital letter and end with a period.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -339,6 +353,339 @@ def data_raw(self) -> ModeSolverData:
...
+        x, y, z = solver_temp_data.Ex.x, solver_temp_data.Ex.y, solver_temp_data.Ex.z
+        axial = solver_ref_data_cylindrical["Er"].axial
+
+        # Initialize output arrays
+        shape = (x.size, y.size, z.size, f.size, mode_index.size)
+        rotated_field_data = {
+            name: np.zeros(shape, dtype=complex) for name in ("Ex", "Ey", "Ez", "Hx", "Hy", "Hz")
+        }
+
+        # offset coors by bend_center
```
momchil-flex: coords

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -339,6 +353,339 @@ def data_raw(self) -> ModeSolverData:
...
+        theta_rel = theta - self.theta_reference
+
+        cos_theta = pts[idx_u] / rho
+        sin_theta = pts[idx_v] / rho
+
+        cmp_normal, source_names = self.plane.pop_axis(("x", "y", "z"), axis=self.bend_axis_3d)
+        cmp_1, cmp_2 = source_names
+
+        for mode_idx in range(mode_index.size):
+            for f_idx, f_val in enumerate(f.values):
```
momchil-flex: I would think it should be possible to vectorize if not along both of these dims, then at least over the frequency one. This could be left as an optimization for later though.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -339,6 +353,339 @@ def data_raw(self) -> ModeSolverData:
...
+
+        axial_name, plane_names = self.plane.pop_axis(("x", "y", "z"), axis=self.bend_axis_3d)
+        cmp_1, cmp_2 = plane_names
+        sel_coords_func = {
+            cmp_1: lambda _rho: _rho * cos_theta + self.rotated_bend_center[idx_u],
+            cmp_2: lambda _rho: _rho * sin_theta + self.rotated_bend_center[idx_v],
+        }
+
+        # Transform fields from Cartesian to cylindrical coordinates
+        for i, _rho in enumerate(rho):
```
momchil-flex: I suspect this could be slow in some cases, but we may have no choice, or do you have any ideas about vectorizing?

tidy3d/plugins/mode/mode_solver.py
QimingFlex: The validator helps to error out such a situation when a regular mode monitor is defined but 'bend_mode_rotation' is set as 'True'.
image

momchil-flex: But why not just avoid erroring out but do nothing special in the handling, i.e. avoid the rotation? Basically I don't see the need for this if under the hood we only use the bend_angle_rotation only if something like `mode_spec.bend_radius is not None and np.abs(mode_spec.angle_theta) > 0` (note that mode_spec already has a validator that requires `bend_axis` to be defined if `bend_radius is not None`).

tidy3d/plugins/mode/mode_solver.py
QimingFlex: Do you mean to use the reduced_simulation object instead of the original simulation to get structures? I tried but it’s challenging to examine if the structures are correctly rotated with the reduced_simulation object.

momchil-flex: Hm. Ok. I think this is fine. We do warn users that using this approach means that things need to be well defined in a region around the mode plane, not just exactly at it.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -348,6 +366,421 @@ def data_raw(self) -> ModeSolverData:
...
+            theta_rel = theta - self.theta_reference
+
+            cos_theta = pts[idx_u] / rho
+            sin_theta = pts[idx_v] / rho
+
+            cmp_normal, source_names = self.plane.pop_axis(("x", "y", "z"), axis=self.bend_axis_3d)
+            cmp_1, cmp_2 = source_names
+
+            for mode_idx in range(mode_index.size):
+                for f_idx, f_val in enumerate(f):
```
momchil-flex: These two loops seem easy to vectorize out.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -348,6 +366,421 @@ def data_raw(self) -> ModeSolverData:
...
+                # Get colocation coordinates in the solver plane
+                normal_dim, plane_dims = self.plane.pop_axis("xyz", self.normal_axis)
+                colocate_coords = {}
+                for dim, sym in zip(plane_dims, self.solver_symmetry):
+                    coords = self.grid_snapped.boundaries.to_dict[dim]
+                    if len(coords) > 2:
+                        if sym == 0:
+                            colocate_coords[dim] = coords[1:-1]
+                        else:
+                            colocate_coords[dim] = coords[:-1]
```
momchil-flex: Let's separate this out in a separate private method that we can call both here and in `_colocate_fields` to avoid code duplication.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -348,6 +366,422 @@ def data_raw(self) -> ModeSolverData:
...
+
+        # Filter polarization if requested
+        if self.mode_spec.filter_pol is not None:
+            self._filter_polarization(mode_solver_data=rotated_mode_data)
+
+        # Sort modes if requested
+        if self.mode_spec.track_freq and len(self.freqs) > 1:
+            rotated_mode_data = rotated_mode_data.overlap_sort(self.mode_spec.track_freq)
+
+        self._field_decay_warning(rotated_mode_data.symmetry_expanded)
```
momchil-flex: These can be skipped since the filtering, sorting, and warning has already been done in the reference mode solver, and they are probably more accurate there too.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -348,6 +366,422 @@ def data_raw(self) -> ModeSolverData:
...
+        plane_name_lateral = axes[lateral_axis]
+
+        # Determine which coordinate transformation to use based on the lateral axis
+        if cmp_1 == plane_name_lateral:
+            lateral_coord_value = rho * cos_theta + self.rotated_bend_center[idx_u]
+        elif cmp_2 == plane_name_lateral:
+            lateral_coord_value = rho * sin_theta + self.rotated_bend_center[idx_v]
+
+        # Transform fields from cartesian to cylindrical
+        fields_at_point = {
```
momchil-flex: Guess this could be called something else now, like `fields_interp` or something.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -348,6 +366,422 @@ def data_raw(self) -> ModeSolverData:
...
+            grid_expanded=grid_expanded,
+            grid_primal_correction=grid_factors[0],
+            grid_dual_correction=grid_factors[1],
+            eps_spec=eps_spec,
+            **rotated_mode_fields,
+        )
+
+        # Colocate to grid boundaries if requested
+        if self.colocate:
+            rotated_mode_data = self._colocate_data(mode_solver_data=rotated_mode_data)
```
momchil-flex: If I'm reading `_mode_rotation` correctly, this can be skipped because the fields have already been created either colocated or not depending on `self.colocate`.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -348,6 +366,422 @@ def data_raw(self) -> ModeSolverData:
...
+            sin_theta = pts[idx_v] / rho
+
+            cmp_normal, source_names = self.plane.pop_axis(("x", "y", "z"), axis=self.bend_axis_3d)
+            cmp_1, cmp_2 = source_names
+
+            k0 = 2 * np.pi * f / C_0
+            n_eff = solver_ref_data_cylindrical["n_complex"].values
+            beta = k0[:, None] * n_eff * np.abs(self.mode_spec.bend_radius)
+
+            for axial_idx, axial_value in enumerate(axial):
```
momchil-flex: Seems like this should be easily vectorizable too?

momchil-flex (commented):
Almost there. One other thing is that this will go in 2.8, not in develop. One way to switch is something like:

```
git checkout -b backup/mode_rotation
git checkout qiming/mode_rotation
git fetch
git reset --hard origin/pre/2.8
git cherry-pick 8a718b6^..df06ea9 # I selected the sha-s of the three commits in this branch as of their current state but these could change if you e.g. rebase
git push -f origin qiming/mode_rotation # and then update the PR base here to be pre/2.8
```

tidy3d/plugins/mode/mode_solver.py
QimingFlex: Yeah, I've updated it with vectorization, before I was not quite familiar with xarray transpose. With this update, it's 80% faster compared to when `bend_angle_rotation=False`. Before, it was slower.

tidy3d/plugins/mode/mode_solver.py
QimingFlex: Yes you're correct. Sorry there was a bug in my code that led to non-colocated results. I found it was related to grid_expand, and it's now resolved and this has been removed.

QimingFlex:
> Almost there. One other thing is that this will go in 2.8, not in develop. One way to switch is something like:
> 
> ```
> git checkout -b backup/mode_rotation
> git checkout qiming/mode_rotation
> git fetch
> git reset --hard origin/pre/2.8
> git cherry-pick 8a718b6^..df06ea9 # I selected the sha-s of the three commits in this branch as of their current state but these could change if you e.g. rebase
> git push -f origin qiming/mode_rotation # and then update the PR base here to be pre/2.8
> ```

This is very helpful, much appreciated.

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Differential operators `grad` and `value_and_grad` in `tidy3d.plugins.autograd` that behave similarly to the autograd operators but support auxiliary data via `aux_data=True` as well as differentiation w.r.t. `DataArray`.
 - `@scalar_objective` decorator in `tidy3d.plugins.autograd` that wraps objective functions to ensure they return a scalar value and performs additional checks to ensure compatibility of objective functions with autograd. Used by default in `tidy3d.plugins.autograd.value_and_grad` as well as `tidy3d.plugins.autograd.grad`.
 - Autograd support for simulations without adjoint sources in `run` as well as `run_async`, which will not attempt to run the simulation but instead return zero gradients. This can sometimes occur if the objective function gradient does not depend on some simulations, for example when using `min` or `max` in the objective.
+- `bend_angle_rotation` in `mode_spec` to improve accuracy in some cases when both `bend_radius` and `angle_theta` are defined."
```
momchil-flex: There's an extra `"` at the end of this line. Also let's say `ModeSpec` and not `mode_spec`. And let's give a bit more explanation, say:

```
`bend_angle_rotation` in `ModeSpec` to improve accuracy in some cases when both `bend_radius` and `angle_theta` are defined. This option is disabled by default but it can be tried when very high accuracy is required in `ModeSource` and `ModeMonitor` objects that are placed in bend and angled waveguides.
```

momchil-flex (approved):
This looks ready, thanks for the great effort!

Just one small comment here and on the backend before merging.

PR #2029, opened 2024-10-22
fix for issue where subdivide was creating very small polygons

dmarek-flex:


dmarek-flex:
@tylerflex @lucas-flexcompute Just a friendly reminder. This bug is preventing the new tutorial notebook on lumped elements from working.

lucas-flexcompute (commented):
Sorry @dmarek-flex, for some reason I did not receive (or missed) the email with this request.

I see that `subdivide` is only used in the 2D material conversion. Is there any issue in removing the small pieces? Would it be possible to join them with larger parts? Note that I'm asking without understanding much of the code here. I don't quite understand the docstring (or why there are 2 docstrings):
```python
    """Subdivide Medium2D into pieces with homogeneous substrate / superstrate."""
    """Use the provided average grid size along the axis to search for neighbors."""
```

tylerflex (approved):
looks good, only slight comment that the chaneglog could be a bit more explicit. "Disallow very small polygons when subdividing 2D structures." or something?

dmarek-flex:
> I see that `subdivide` is only used in the 2D material conversion. Is there any issue in removing the small pieces? Would it be possible to join them with larger parts? Note that I'm asking without understanding much of the code here. I don't quite understand the docstring (or why there are 2 docstrings):

I'll fix the docstring, not sure what happened there! The implementation of 2D materials requires the materials above and below the 2D medium to be homogeneous.  `subdivide` is needed in the case that the structure associated with the 2D medium intersects with more than one structure above and/or below.  `subdivide` will divide the original 2D structure into partitions that each have only one medium above and one medium below. 

I don't think there is any issue in removing small pieces, since the threshold is very small. In my case, a polygon was created that was essentially a line segment, but one corner was slightly offset. I assume that it was due to numerical precision issues in `shapely` when performing the boolean polygon operations.

lucas-flexcompute (approved):

PR #2030, opened 2024-10-22
rename autograd_background_permittivity to background_permittivity

tylerflex:
the old name is too long. doesnt' require a new changelog item since it hasn't been released yet


yaugenst-flex (approved):
Two commits are probably not necessary? Otherwise looks good.

PR #2031, opened 2024-10-23
mode_spec.precision default to `auto`

weiliangjin2021:
Tests are added in the solver part

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added validation and repair methods for `TriangleMesh` with inward-facing normals.
 - Added `from_admittance_coeffs` to `PoleResidue`, which allows for directly constructing a `PoleResidue` medium from an admittance function in the Laplace domain.
 - Added material type `LossyMetalMedium` that has high DC-conductivity. Its boundaries can be modeled by a surface impedance boundary condition (SIBC). This is treated as a subpixel method, which can be switched by setting `SubpixelSpec(lossy_metal=method)` where `method` can be `Staircasing()`, `VolumetricAveraging()`, or `SurfaceImpedance()`.
+- Added mode solver option `precision='auto'` to automatically select single or double precision for optimizing performance and accuracy.
 
 ### Changed
 - Priority is given to `snapping_points` in `GridSpec` when close to structure boundaries, which reduces the chance of them being skipped.
--- a/tidy3d/components/mode.py
+++ b/tidy3d/components/mode.py
@@ -102,11 +102,13 @@ class ModeSpec(Tidy3dBaseModel):
         units=RADIAN,
     )
 
-    precision: Literal["single", "double"] = pd.Field(
-        "single",
-        title="single or double precision in mode solver",
+    precision: Literal["auto", "single", "double"] = pd.Field(
+        "auto",
+        title="single, double, or automatic precision in mode solver",
         description="The solver will be faster and using less memory under "
-        "single precision, but more accurate under double precision.",
+        "single precision, but more accurate under double precision. "
+        "With the default ``'auto'``, apply double precision if the simulation contains a good "
+        "conductor, single precision otherwise.",
     )
 
     bend_radius: float = pd.Field(
@@ -198,7 +200,7 @@ class ModeSpec(Tidy3dBaseModel):
                 )
 
             # multiply by 5 to be safe
-            if values["group_index_step"] < 5 * fp_eps and values["precision"] == "single":
+            if values["group_index_step"] < 5 * fp_eps and values["precision"] != "double":
                 log.warning(
                     "Group index step is too small! "
                     "The results might be fully corrupted by numerical errors. "
--- a/tidy3d/plugins/mode/solver.py
+++ b/tidy3d/plugins/mode/solver.py
@@ -24,9 +24,11 @@ TOL_TENSORIAL = 1e-6
 TARGET_SHIFT = 10 * fp_eps
 # Preconditioner: "Jacobi" or "Material" based
 PRECONDITIONER = "Material"
-# PEC permittivity cut-off value. Let it be as large as possible so long as not causing overflow in
+# Good conductor permittivity cut-off value. Let it be as large as possible so long as not causing overflow in
 # double precision. This value is very heuristic.
-PEC_CUT_OFF = 1e70
+GOOD_CONDUCTOR_CUT_OFF = 1e70
+# Consider a material to be good conductor if |ep| (or |mu|) > GOOD_CONDUCTOR_THRESHOLD * |pec_val|
+GOOD_CONDUCTOR_THRESHOLD = 0.9
 
 
 class EigSolver(Tidy3dBaseModel):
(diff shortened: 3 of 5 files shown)


tidy3d/plugins/mode/solver.py
```diff
@@ -1011,6 +1012,26 @@ def split_curl_field_postprocess_inverse(split_curl, E):
...
         raise RuntimeError("Split curl not yet implemented for relative mode solver.")
 
+    @staticmethod
+    def mode_plane_contain_good_metal(material_response) -> bool:
+        """Find out if epsilon on the modal plane contain very metallic materials whose permittivity
+        or permeability value is very large.
+        """
+        if material_response is None:
+            return False
+        return np.any(np.abs(material_response) > 0.9 * np.abs(pec_val))
```
momchil-flex: I think this 0.9 should still be either a function argument as previously in `any_pec`, or a `GOOD_CONDUCTOR_THRESHOLD` constant defined at the top of the file. Also: I wonder if we want it to be lower than 0.9?

momchil-flex (approved):

tidy3d/components/mode.py
```diff
@@ -102,11 +102,13 @@ class ModeSpec(Tidy3dBaseModel):
...
-    precision: Literal["single", "double"] = pd.Field(
-        "single",
+    precision: Literal["auto", "single", "double"] = pd.Field(
+        "auto",
         title="single or double precision in mode solver",
         description="The solver will be faster and using less memory under "
-        "single precision, but more accurate under double precision.",
+        "single precision, but more accurate under double precision. "
+        "With the default ``auto``, apply double precision if the simulation contains good "
+        "conductor, single precision otherwise.",
```
dmarek-flex: "... contains a good..."

tidy3d/plugins/mode/solver.py
```diff
@@ -1011,6 +1012,26 @@ def split_curl_field_postprocess_inverse(split_curl, E):
         """E has the shape (3, N, num_modes)"""
         raise RuntimeError("Split curl not yet implemented for relative mode solver.")
 
+    @staticmethod
+    def mode_plane_contain_good_metal(material_response) -> bool:
+        """Find out if epsilon on the modal plane contain very metallic materials whose permittivity
+        or permeability value is very large.
```
dmarek-flex: I think we can refer to these materials as "good" conductors, so it might help to use the same wording here too.

dmarek-flex (approved):
Just some small comments.

We should try to come up with a universal method for determining whether a material is a good conductor or not, at some point.

tidy3d/components/mode.py
```diff
@@ -102,11 +102,13 @@ class ModeSpec(Tidy3dBaseModel):
...
 
-    precision: Literal["single", "double"] = pd.Field(
-        "single",
+    precision: Literal["auto", "single", "double"] = pd.Field(
+        "auto",
         title="single or double precision in mode solver",
         description="The solver will be faster and using less memory under "
-        "single precision, but more accurate under double precision.",
+        "single precision, but more accurate under double precision. "
+        "With the default ``auto``, apply double precision if the simulation contains good "
```
tylerflex: could you put "auto" in quotes too? as its a string

tylerflex: also how do you define "good" conductor?

tidy3d/plugins/mode/solver.py
tylerflex: I'm not sure what the correct terminology here, but I feel "good" sounds a bit strange overall here? maybe "strong" conductors or something. again not super set on this but just a feeling I had. basically it feels a little vague to me

tylerflex (approved):
I'll approve with just a few wording issues. and just throwing out there that I'm not sure about using "good" but if it's a common terminology for this, no problem

lucas-flexcompute:
> I'll approve with just a few wording issues. and just throwing out there that I'm not sure about using "good" but if it's a common terminology for this, no problem

"Good conductor" is a used term in RF/MW: https://phys.libretexts.org/Bookshelves/Electricity_and_Magnetism/Electromagnetics_II_(Ellingson)/03%3A_Wave_Propagation_in_General_Media/3.11%3A_Good_Conductors

weiliangjin2021:
> > I'll approve with just a few wording issues. and just throwing out there that I'm not sure about using "good" but if it's a common terminology for this, no problem
> 
> "Good conductor" is a used term in RF/MW: https://phys.libretexts.org/Bookshelves/Electricity_and_Magnetism/Electromagnetics_II_(Ellingson)/03%3A_Wave_Propagation_in_General_Media/3.11%3A_Good_Conductors

Thanks Lucas! I'll change all other wordings to "good conductor".

tidy3d/plugins/mode/solver.py
weiliangjin2021: Any examples that you have in mind to have it lower than 0.9?

momchil-flex: Not really. It's more that our `pec_val` is somewhat arbitrary? Like should it be good conductor already at like 1000?

weiliangjin2021: In principle, I think good conductor here means it can be treated as impenetrable inside the conductor. So in time-stepping, the fields are set to 0; and in mode solver, sufficiently large permittivity. In our implementation, it's limited to materials that are defined with td.PEC or td.LossyMetalMedium. Otherwise, we cannot just arbitrarily boost its permittivity values.

momchil-flex: Ok, sounds good. So the threshold of 0.9 is just to make sure we capture PEC and everything that's effectively the same, up to some small tolerance.

weiliangjin2021: But as mentioned in the other PR, I'll revert all those and just use simulation.mediums to select precision.

weiliangjin2021:
Precision selection is now based on whether the reduced simulation contains PEC or LossyMetal with SIBC, so that it won't slow down metadata.

momchil-flex (approved):

tidy3d/plugins/mode/solver.py
```diff
@@ -64,6 +67,8 @@ def compute_modes(
             (Hertz) Frequency at which the eigenmodes are computed.
         mode_spec : ModeSpec
             ``ModeSpec`` object containing specifications of the mode solver.
+        precision : Literual["single", "double"]
```
dmarek-flex: typo "Literal"

dmarek-flex (approved):

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added validation and repair methods for `TriangleMesh` with inward-facing normals.
 - Added `from_admittance_coeffs` to `PoleResidue`, which allows for directly constructing a `PoleResidue` medium from an admittance function in the Laplace domain.
 - Added material type `LossyMetalMedium` that has high DC-conductivity. Its boundaries can be modeled by a surface impedance boundary condition (SIBC). This is treated as a subpixel method, which can be switched by setting `SubpixelSpec(lossy_metal=method)` where `method` can be `Staircasing()`, `VolumetricAveraging()`, or `SurfaceImpedance()`.
+- Added mode solver precision option `auto` to automatically select single or double precision for optimizing performance and accuracy.
```
tylerflex: please put auto in single quotes (inside of the backslash quotes too) just because it's a string. 

" add mode solver option `precision='auto'` to automatically ... "

tylerflex (approved):
looks good, just one minor changelog comment

PR #2032, opened 2024-10-23
fix: remove autograd temporary simulation files

yaugenst-flex:
Currently, autograd temp files are being created but not removed in `upload_sim_fields_keys` and `get_vjp_traced_fields`. This causes `/tmp` to eventually run out of space, especially when running longer optimizations.
Right now, the file is being created, deleted, and then re-created with the name of the temp file in the `.to_file()` and `download_file()` calls, respectively.
Ideally we would be able to use a context manager here, but since we are overwriting the file handle, that approach doesn't work. We should probably be passing around file handles instead of names internally.

### DIFF
--- a/tidy3d/web/api/autograd/autograd.py
+++ b/tidy3d/web/api/autograd/autograd.py
@@ -1,5 +1,6 @@
 # autograd wrapper for web functions
 
+import os
 import tempfile
 import typing
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


tidy3d/web/api/autograd/autograd.py
```diff
@@ -529,26 +530,32 @@ def postprocess_fwd(
...
+    try:
+        TracerKeys(keys=sim_fields_keys).to_file(fname)
+        upload_file(
+            task_id,
+            fname,
+            SIM_FIELDS_KEYS_FILE,
+            verbose=verbose,
+        )
+    finally:
+        os.unlink(fname)
```
momchil-flex, line 547: Looks good but can you explain why we couldn't just delete `data_file.name` with `os.unlink` or `shutil.rmtree` or something like that without changing the rest of the code?

yaugenst-flex: yeah that's fair, I only did this to be consistent with how we're handling temp files in base.py [here](https://github.com/flexcompute/tidy3d/blob/ce648575a8c218e4a6a259ec406bd09446b40493/tidy3d/components/base.py#L766), but I can change it

momchil-flex: No, that's fine, that's a good enough reason, I was just wondering.

But while you're at this could you actually browse through all the places we use `tempfile`? I think there may be others to fix too, like in the old adjoint plugin.

yaugenst-flex: ok yeah sounds good, will do

tylerflex (approved):

tylerflex (approved):
Just curious why no except logic here? Is it worth adding

yaugenst-flex:
good idea to add some, done

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: yup same problem in adjoint plugin, everywhere else looks fine

yaugenst-flex:
@momchil-flex can we merge this?

PR #2034, opened 2024-10-23
Polygon triangulation supports collinear vertices

lucas-flexcompute:
This is a fix for our ear clipping triangulation algorithm to work with polygons with collinear vertices.
It also adds a safeguard in the main loop to error if no progress is made, so that we don't get stuck jobs due to numerical issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Users can manually specify the background medium for a structure to be used for geometry gradient calculations by supplying `Structure.background_permittivity`. This is useful when there are overlapping structures or structures embedded in other mediums.
 
+### Changed
+- Triangulation of `PolySlab` polygons now supports polygons with collinear vertices.
+
 ### Fixed
 - Minor gradient direction and normalization fixes for polyslab, field monitors, and diffraction monitors in autograd.
 
--- a/tidy3d/components/geometry/triangulation.py
+++ b/tidy3d/components/geometry/triangulation.py
@@ -4,6 +4,7 @@ from typing import List, Tuple
 import numpy as np
 import shapely
 
+from ...exceptions import Tidy3dError
 from ..types import ArrayFloat1D, ArrayFloat2D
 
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Users can manually specify the background medium for a structure to be used for geometry gradient calculations by supplying `Structure.background_permittivity`. This is useful when there are overlapping structures or structures embedded in other mediums.
 
+### Changed
+- Triangulation of `PolySlab` poolygons now supports pollygons with collinear vertices.
```
momchil-flex: Two different typos in `polygons` :)

momchil-flex (approved):

CHANGELOG.md
lucas-flexcompute: I'm blaming that on a different keyboard that I'm using these past few days... :/

momchil-flex:
Awesome btw tested that this un-stucks the problematic sim! You probably did as well but I verified.

CHANGELOG.md
weiliangjin2021: It happens to me too with my apple magic keyboard that has been with me for 10 years. Now have to switch to a new one.

tidy3d/components/geometry/triangulation.py
```diff
@@ -41,15 +42,21 @@ def update_convexity(vertices: List[Vertex], i: int) -> None:
         Vertices of the polygon.
     i : int
         Index of the vertex to be updated.
+
+    Returns
+    -------
+    int
+        -1 if vertex was collinear, +1 if it became collinear, 0 otherwise.
```
weiliangjin2021: This part is hard to understand. Maybe add more explanation here?
My understanding by reading the code below is: -1 for vertex was collinear, and no longer collinear after the update. But what about +1 and 0?

weiliangjin2021 (approved):

tidy3d/components/geometry/triangulation.py
lucas-flexcompute: This return value is just to quickly update the counter of collinear vertices, i.e., if the update decreased the number of collinear vertices (a vertex was collinear and no longer is), it returns -1; if the update increased the number of collinears (the vertex was not collinear and now it is), it returns +1; else, the update doesn't change the count and it returns 0.

momchil-flex: Maybe just add this (or a short version of it) to the docstring @lucas-flexcompute ?

lucas-flexcompute: Done! I think it's more clear now with the note and comment.

momchil-flex: Thanks, could you squash and merge.

PR #2036, opened 2024-10-24
Add frequency/wavelength utilities

lucas-flexcompute:


tidy3d/frequencies.py
```diff
@@ -0,0 +1,216 @@
...
+            elif value < 3e3:
+                return ("radio wave", "ULF")
+            elif value < 30e3:
+                return ("radio wave", "VLF")
+            elif value < 300e3:
+                return ("radio wave", "LF")
+            elif value < 3e6:
+                return ("radio wave", "MF")
+            elif value < 30e6:
+                return ("radio waves", "HF")
```
dmarek-flex: typo `"radio wave"`

tidy3d/frequencies.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def classification(self, value: float) -> tuple[str]:
+        """Band classification for a given frequency/wavelength.
+
+        Args:
+            value: Value to classify.
+
+        Returns:
+            String tuple with classification.
+        """
```
dmarek-flex: Should fix style to follow other docstrings.

dmarek-flex (approved):
Looks good to me, thanks for including microwave! I suggest we make the units extremely clear in the class docstring perhaps. Frequency (Hz) is pretty obvious, but wavelength (microns) will be a bit unexpected for microwave applications.

Also for these utilities should we also derive from `Tidy3dBaseModel` and make use of `pydantic`?

tidy3d/frequencies.py
```diff
@@ -0,0 +1,216 @@
...
+
+O_BAND = (1.260, 1.360)
+E_BAND = (1.360, 1.460)
+S_BAND = (1.460, 1.530)
+C_BAND = (1.530, 1.565)
+L_BAND = (1.565, 1.625)
+U_BAND = (1.625, 1.675)
+
+
+class FrequencyUtils:
```
tylerflex: let's use pydantic for this (inherit from `Tidy3DBaseModel` and define `use_wavelength` as a `pd.Field`). Just for consistency and in case we want to include something like this in GUI.

tylerflex (changes requested):

lucas-flexcompute:
One question that might be worth discussing is the return values of the `*_band` functions. They return crescent wavelength values and decrescent frequencies, so that they match when converted without reversing the list. Is that ok?

tylerflex:
> One question that might be worth discussing is the return values of the *_band functions. They return crescent wavelength values and decrescent frequencies, so that they match when converted without reversing the list. Is that ok?

increasing / decreasing?  hm, I don't have a real strong preference. i think most of the time these will be used to set `monitor.freqs`, which i think doesnt care about ordering.

lucas-flexcompute:
True, in practice, it probably doesn't matter. I just implemented this way because thats the ordering we end up using the most in examples when we define
```python
lda = np.linspace(...)
freqs = td.C_0 / lda
```
So I think it's okay.

tylerflex (approved):

PR #2038, opened 2024-10-25
Pre-upload validating server autograd runs and optimizer

momchil-flex:
- Add `validate_pre_upload()` call to the forward simulation *with monitors* when called in a server autograd run. The adjoint simulation includes the monitors already.
- Modify the optimizer `validate_pre_upload()` method to call the corresponding method for the embedded simulation.


tylerflex (approved):
looks good, guess this answers my question on the core PR?

momchil-flex:
> looks good, guess this answers my question on the core PR?

Ah, yes, I answered that one there first before seeing this. :)

PR #2040, opened 2024-10-25
Limit max data size to 100 * simulation.num_cells

momchil-flex:
At some point we had a user submit many small simulations that each needed to upload GBs of data, which overloaded our servers. I discussed with Zongfu at the time that one solution would be to allow at most `100 * simulation.num_cells` bytes of data to be produced by a given simulation. So a simulation with 1M points can only have 100MB output, but a simulation with 500M points reaches our other hard limit of 50GB data total.

I am not sure if this will interfere with e.g. some of the adjoint examples, but I'm PR-ing this into pre/2.8, so we can see what happens during a notebook rerun.

Now that I think about it, this limit would mean recording all the fields in the entire simulation over 4 freqs only. It could be an issue in broadband 2D optimizations for example. But not sure how to both protect our infrastructure and not restrict users too much. In the particular problematic case the issue was due to a time monitor, which I'm more OK restricting.


momchil-flex:
Closing this as we will impose a flex credit charge on GB of data instead.

PR #2041, opened 2024-10-25
fix: xarray 2023.08.0 compatibility

yaugenst-flex:
fixes xarray compatibility issues caused by latest Dataarray changes

short explanation: older xarray has a check when assigning to values, which calls `astype` if `__array_namespace__` exists. since our new arraybox has that now, we need to put `astype` into the `autograd.numpy` namespace, otherwise it fails.


momchil-flex (approved):
So it just happened that all our GH test matrix used the xarray version where this was not an issue?

yaugenst-flex:
yup I assume they were just pulling latest. we're not really testing support of minimum versions....

PR #1913, opened 2024-08-21
:arrow_up: Looser 2.0 numpy functionality

daquinteroflex:
Closes https://github.com/flexcompute/tidy3d/issues/1912

- [x] Possibly add the trapz logic per numpy version
- [x] Work out how to create the locked resolved environment.
  - [X] Update jax dependency. Had to update jax due to a weird segmentation fault with the older version.
  - [X] ~Update pandas dependency to greater than 2.2.2 https://github.com/pandas-dev/pandas/issues/55519~  Not required below as not related to the error.
  - [x] Wait for the next autograd release with this PR https://github.com/HIPS/autograd/pull/618
  - [X] ~Increase boto3 version.~
  - [X] Identify the right gdstk dependencies (thanks Yannick) 
- [x] Get the CI tests to pass.
- [x] Proper backend testing too.
- [X] Finish the backend PR

PR #2043, opened 2024-10-27
try python 3.13

tylerflex:


tylerflex:
@lucas-flexcompute seems it is not compatible with python 3.13 because of `gdstk`

```
ERROR: No matching distribution found for gdstk
```

FYI

tylerflex:
good news is that otherwise I ran tidy3d successfully using python 3.13

lucas-flexcompute:
> @lucas-flexcompute seems it is not compatible with python 3.13 because of `gdstk`
> 
> ```
> ERROR: No matching distribution found for gdstk
> ```
> 
> FYI

Fixed: https://pypi.org/project/gdstk/#files

tylerflex:
thanks, it passed, looks like an issue with scipy now

tylerflex:
@daquinteroflex could you take a look? I'm getting stuck on jax now, which only supports python 3.13 after 0.4.35, but poetry lock doesnt like my pyproject.toml for some reason

daquinteroflex:
Sounds good! Possibly it gets resolved in https://github.com/flexcompute/tidy3d/pull/1913 with those updates. Seems like the backend is ready to be rebased into so will sort that out first so we can test this based on the numpy2 PR

yaugenst-flex:
Running into this. JAX added python 3.13 support in [0.4.34](https://jax.readthedocs.io/en/latest/changelog.html#jax-0-4-34-october-4-2024). Oh and JAX dropped support for python 3.9, which also causes problems :/

yaugenst-flex:
Ok after more fiddling around with this I'm giving up for now. I think we should deprecate python 3.9...

tylerflex:
would deprecating python 3.9 mean that it would still work for most users?  I suppose it would only be an issue for users of the adjoint plugin? / people using jax?

tylerflex:
note that it's still technically supported until October 2025
https://devguide.python.org/versions/

yaugenst-flex:
Closing since Python 3.13 is supported now.

PR #2048, opened 2024-10-31
fix: bug in local field projection for `far_field_approx=True`

yaugenst-flex:
Fixes a bug in the local field projection that was introduced in #1822. The bug occurred because autograd-compatible indexing of `J` and `M` was incorrect and a bit more involved than I thought...

The good news is that this PR also introduces autograd support for the kspace field projection.


momchil-flex (approved):
Confirmed fix with backend tests.

PR #2049, opened 2024-10-31
fix bug in autograd differentiation cylinder.center

tylerflex:
When the cylinder is differentiated w.r.t but not the radius, there's a bug that wasn't caught in tests.


momchil-flex (approved):

momchil-flex:
Looks good but tests are failing.

PR #2052, opened 2024-11-01
size=0 adjoint montiors for single pixel dims in custom mediums

tylerflex:
An optimization to reduce data usage for CustomMedium adjoint monitors. size=0 for any dims with just 1 pixel, since we only evaluate at centers.

Note; eventually would be realy nice to just have a point-cloud monitor and do these custom for each structure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Autograd support for local field projections using `FieldProjectionKSpaceMonitor`.
 - Function `components.geometry.utils.flatten_groups` now also flattens transformed groups when requested.
 
+### Changed
+- `CustomMedium` design regions require far less data when performing inverse design by reducing adjoint field monitor size for dims with one pixel.
+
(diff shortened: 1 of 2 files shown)


momchil-flex:
So this is probably nice to get in right now but in general these two things should not be coupled? Like I should be able to define my custom medium by a single pixel if I want it extruded in a given dimension, but potentially still integrate fields across that dimension which would be important e.g. for polarization mixing components?

tylerflex:
Not sure I see the problem with this?

Until we add the option to do full volume integration, I think it makes sense to just restrict the monitor size, like we do with PolySlab.

Agree that the point cloud monitor is kind of a separate issue. In principle all 3 are separate?

momchil-flex:
Yeah, we can merge this in right away, I was just thinking that we need to make sure this is removed when we incorporate full volume integration.

Have you done any tests that the gradients have not changed? I am not sure if this would be captured by our automatic testing.

tylerflex:
I tested with a couple different notebooks (mode converter and WDM) both looked the same as before, but were a bit faster

PR #2053, opened 2024-11-04
Validate that bend radius is not smaller than half the mode plane size

### LINKED ISSUE #1299: Validate that mode bend radius is larger than 1/2 of the mode plane size along the radial axis

momchil-flex:
This was a bit annoying because I need to validate both mode sources/monitors and mode solver init, which have different fields needed for the validation ("size" for mode sources and monitors, "plane" for mode solver). So I couldn't write a native pydantic validator but made it a post-init instead. I also decided to put the validation in Simulation so I could point to the exact monitor/source index, but on the other hand maybe it's nicer to validate as soon as the object is created? Let me know what you think. Fixes #1299

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -85,7 +85,11 @@ from .source import (
 from .structure import MeshOverrideStructure, Structure
 from .subpixel_spec import SubpixelSpec
 from .types import TYPE_TAG_STR, Ax, Axis, FreqBound, InterpMethod, Literal, Symmetry, annotate_type
-from .validators import assert_objects_in_sim_bounds, validate_mode_objects_symmetry
+from .validators import (
+    assert_objects_in_sim_bounds,
+    validate_mode_objects_symmetry,
+    validate_mode_plane_radius,
+)
 from .viz import (
     PlotParams,
     add_ax_if_none,
@@ -3245,6 +3249,24 @@ class Simulation(AbstractYeeGridSimulation):
         self._validate_tfsf_nonuniform_grid()
         self._validate_nonlinear_specs()
         self._validate_custom_source_time()
+        self._validate_mode_object_bends()
+
+    def _validate_mode_object_bends(self) -> None:
+        """Error if any mode sources or monitors with bends have a radius that is too small."""
+        for imnt, monitor in enumerate(self.monitors):
+            if isinstance(monitor, AbstractModeMonitor):
+                validate_mode_plane_radius(
+                    mode_spec=monitor.mode_spec,
+                    plane=monitor.geometry,
+                    msg_prefix=f"Monitor at 'monitors[{imnt}]' ",
+                )
+        for isrc, source in enumerate(self.sources):
+            if isinstance(source, ModeSource):
+                validate_mode_plane_radius(
+                    mode_spec=source.mode_spec,
+                    plane=source.geometry,
+                    msg_prefix=f"Source at 'sources[{isrc}]' ",
+                )
 
     def _validate_custom_source_time(self):
         """Warn if all simulation times are outside CustomSourceTime definition range."""
--- a/tidy3d/components/validators.py
+++ b/tidy3d/components/validators.py
@@ -8,6 +8,7 @@ from ..log import log
 from .base import DATA_ARRAY_MAP, skip_if_fields_missing
 from .data.dataset import Dataset, FieldDataset
 from .geometry.base import Box
+from .mode import ModeSpec
 from .types import Tuple
 
 """ Explanation of pydantic validators:
@@ -396,3 +397,21 @@ def validate_freqs_not_empty():
         return val
 
     return freqs_not_empty
+
+
+def validate_mode_plane_radius(mode_spec: ModeSpec, plane: Box, msg_prefix: str = ""):
+    """Validate that the radius of a mode spec with a bend is not smaller than half the size of
+    the plane along the radial direction."""
+
+    if not mode_spec.bend_radius:
+        return
+
+    # radial axis is the plane axis that is not the bend axis
+    _, plane_axs = plane.pop_axis([0, 1, 2], plane.size.index(0.0))
+    radial_ax = plane_axs[(mode_spec.bend_axis + 1) % 2]
+
+    if np.abs(mode_spec.bend_radius) < plane.size[radial_ax] / 2:
+        raise ValueError(
+            f"{msg_prefix} bend radius is smaller than half the mode plane size "
+            "along the radial axis, which can produce wrong results."
+        )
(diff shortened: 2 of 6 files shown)


tidy3d/components/simulation.py
```diff
@@ -3245,6 +3249,24 @@ def _post_init_validators(self) -> None:
...
+        self._validate_mode_object_bends()
+
+    def _validate_mode_object_bends(self) -> None:
+        """Error if any mode sources or monitors with bends have a radius that is too small."""
+        for imnt, monitor in enumerate(self.monitors):
+            if isinstance(monitor, AbstractModeMonitor):
+                validate_mode_plane_radius(
+                    mode_spec=monitor.mode_spec,
+                    plane=monitor.geometry,
+                    msg_prefix=f"Monitors[{imnt}] ",
```
tylerflex: do we want to put this in single quotes `'` and also use lowercase like `monitors[...]`?

tidy3d/components/simulation.py
```diff
@@ -3245,6 +3249,24 @@ def _post_init_validators(self) -> None:
         self._validate_tfsf_nonuniform_grid()
         self._validate_nonlinear_specs()
         self._validate_custom_source_time()
+        self._validate_mode_object_bends()
+
+    def _validate_mode_object_bends(self) -> None:
+        """Error if any mode sources or monitors with bends have a radius that is too small."""
+        for imnt, monitor in enumerate(self.monitors):
```
tylerflex, line 3256: not sure if it would make most sense to collect all `imnt` and `isrc` and do the validation in one go? displaying all of the offending indices?

tidy3d/components/validators.py
```diff
@@ -396,3 +396,21 @@ def freqs_not_empty(cls, val):
         return val
 
     return freqs_not_empty
+
+
+def validate_mode_plane_radius(mode_spec: "ModeSpec", plane: "Box", msg_prefix: str):  # noqa: F821
```
tylerflex: I think it's fine (and perhaps better) to import these types and use directly here.

tylerflex: maybe `ms_prefix: str =''` in case we dont want to require it, very minor though.

tidy3d/plugins/mode/mode_solver.py
```diff
@@ -173,6 +177,11 @@ def plane_in_sim_bounds(cls, val, values):
             raise SetupError("'ModeSolver.plane' must intersect 'ModeSolver.simulation'.")
         return val
 
+    def _post_init_validators(self) -> None:
+        validate_mode_plane_radius(
```
tylerflex, line 181: not sure if we put docstrings for these in other places, but a short one wouldnt hurt (validation check that runs after object initialized).

tylerflex (commented):
changelog item under fixed maybe?

tidy3d/components/simulation.py
momchil-flex: Hm it's not what we usually do I believe. Just error asap sounds fine to me.

tidy3d/components/validators.py
momchil-flex: That would be circular import as usually validators are imported in other components rather than the other way round. However, I don't know if this is the best place to put this function: it seems that everything else in this file creates an actual pydantic validator, while this is just a convenient method to use in a post-init validator.

tidy3d/components/simulation.py
momchil-flex: done

tidy3d/components/validators.py
tylerflex: i think validators already imports from `Box` https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/validators.py#L10 and ModeSpec doesnt import from validators, so is it circular?

tylerflex: in any case, while separating the validators into their own file was maybe not a bad idea, functionally it's probably fine if this validator was just defined inside of `simulation.py` or wherever its used.

tidy3d/components/simulation.py
tylerflex: sounds good.

tidy3d/components/validators.py
tylerflex: Here's my thought:

* if the type annotation causes circular imports, we probably should be rethinking where to place the function in the first place. So in this case, should this just go in simulation.py for example? what's the benefit of separating it into validators.py when it just gets used in simulation things?
* If we want to keep it validators.py just for some separation, I think it's probably ok to import Box and ModeSpec at least the way it looks now, but it does mean we can't import validators from here those modules in the future.

so maybe overall I think it makes a bit more sense to just put this in `simulation.py`? or nearby (eg in new module `base_sim/validators.py`?)

momchil-flex: > i think validators already imports from `Box` https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/validators.py#L10 and ModeSpec doesnt import from validators, so is it circular?

Oh interesting, sorry I was just thinking about it conceptually that this validators file would make sense to me to be at the top of the hierarchy. I'm not sure if there's any logic to it right now, it seems to me like we *could* at some point have a validator we may want to import in Box or ModeSpec and it just so happens that we don't have one now?

> in any case, while separating the validators into their own file was maybe not a bad idea, functionally it's probably fine if this validator was just defined inside of simulation.py or wherever its used.

So I put it here because I need to use it both in Simulation and in ModeSolver. Any better ideas?

tylerflex: > Oh interesting, sorry I was just thinking about it conceptually that this validators file would make sense to me to be at the top of the hierarchy.

Yea i'm also thinking about this out loud because I think it was never super well defined. Conceptually that makes sense, but also if these validators are going to use specific information about Box or ModeSpec components , they probably shouldnt be here. Maybe these more specific validators need to just go where they are used.

> So I put it here because I need to use it both in Simulation and in ModeSolver.

Yea maybe in `simulation.py` or actually a new file `base_sim/validators.py` makes some sense? the ModeSolver would import from `base_sim` right?

momchil-flex: I think the original idea of the `validators.py` file is exactly for things that need to be imported in multiple files. Generally it makes sense to me that there may be such validators but not sure if a more fine-grained splitting would make more sense.

tylerflex: yea it's just that there's some conflict between: being able to import these validators in many files vs. these validators being able to know about things in said files (eg. ModeSpec).

For now maybe my preference is make a new validators.py in `sim_base` and put this there (+ maybe other simulation specific validators that live in the current file?

momchil-flex: So I started doing this but actually I don't understand how this helps clarify the structure. What is the logic/difference between having validators in `components/validators.py` and `components/base_sim/validators.py`?

momchil-flex: Updated.

tylerflex (approved):

PR #2055, opened 2024-11-04
More autograd helpers

yaugenst-flex:
- Removes `DeprecationWarning` of `DataArray.values`.
- Adds `smooth_min`, `smooth_max`, and `least_squares` implementations to `plugins.autograd.functions`.
- Adds `@scalar_objective` decorator that extracts `DataArray.data` and performs some additional checks on the output data of objective functions. If an objective function is decorated with this, calling `DataArray.values` within the objective function is not necessary anymore. Our `grad` and `value_and_grad` implementations in `plugins.autograd` wrap objective functions with this decorator automatically.
- Some internal refactoring of the `pad` function.
- Some test refactoring to reduce the no. of function calls for grad checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Autograd support for local field projections using `FieldProjectionKSpaceMonitor`.
 - Function `components.geometry.utils.flatten_groups` now also flattens transformed groups when requested.
+- Differentiable `smooth_min`, `smooth_max`, and `least_squares` functions in `tidy3d.plugins.autograd`.
 
 ### Changed
 - `CustomMedium` design regions require far less data when performing inverse design by reducing adjoint field monitor size for dims with one pixel.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `CustomMedium` design regions require far less data when performing inverse design by reducing adjoint field monitor size for dims with one pixel.
+- Calling `.values` on `DataArray` no longer raises a `DeprecationWarning` during automatic differentiation
 
 ### Fixed
 - Regression in local field projection leading to incorrect results for `far_field_approx=True`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Autograd support for local field projections using `FieldProjectionKSpaceMonitor`.
 - Function `components.geometry.utils.flatten_groups` now also flattens transformed groups when requested.
 - Differentiable `smooth_min`, `smooth_max`, and `least_squares` functions in `tidy3d.plugins.autograd`.
+- Differential operators `grad` and `value_and_grad` in `tidy3d.plugins.autograd` that behave similarly to the autograd operators but support auxiliary data via `aux_data=True` as well as differentiation w.r.t. `DataArray`.
+- `@scalar_objective` decorator in `tidy3d.plugins.autograd` that wraps objective functions to ensure they return a scalar value and performs additional checks to ensure compatibility of objective functions with autograd. Used by default in `tidy3d.plugins.autograd.value_and_grad` as well as `tidy3d.plugins.autograd.grad`.
+
 
 ### Changed
 - `CustomMedium` design regions require far less data when performing inverse design by reducing adjoint field monitor size for dims with one pixel.
--- a/tidy3d/components/data/data_array.py
+++ b/tidy3d/components/data/data_array.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import warnings
 from abc import ABC
 from typing import Any, Dict, List, Mapping, Union
 
@@ -220,15 +219,7 @@ class DataArray(xr.DataArray):
         np.ndarray
             The values of the DataArray.
         """
-        if isbox(self.data):
-            warnings.warn(
-                "'DataArray.values' for automatic differentiation is deprecated, "
-                "please use 'DataArray.data' instead.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
-            return self.data
-        return super().values
+        return self.data if isbox(self.data) else super().values
 
     @values.setter
     def values(self, value: Any) -> None:
--- a/tidy3d/plugins/autograd/differential_operators.py
+++ b/tidy3d/plugins/autograd/differential_operators.py
@@ -1,65 +1,72 @@
-from typing import Any, Callable
+from typing import Callable
 
-from autograd import value_and_grad as value_and_grad_ag
 from autograd.builtins import tuple as atuple
 from autograd.core import make_vjp
 from autograd.extend import vspace
 from autograd.wrap_util import unary_to_nary
 from numpy.typing import ArrayLike
 
+from .utilities import scalar_objective
+
 __all__ = [
     "value_and_grad",
+    "grad",
 ]
 
 
 @unary_to_nary
-def value_and_grad(
-    fun: Callable, x: ArrayLike, *, has_aux: bool = False
-) -> tuple[tuple[float, ArrayLike], Any]:
-    """Returns a function that returns both value and gradient.
-
-    This function wraps and extends autograd's 'value_and_grad' function by adding
-    support for auxiliary data.
+def grad(fun: Callable, x: ArrayLike, *, has_aux: bool = False) -> Callable:
+    """Returns a function that computes the gradient of `fun` with respect to `x`.
 
     Parameters
     ----------
     fun : Callable
-        The function to differentiate. Should take a single argument and return
-        a scalar value, or a tuple where the first element is a scalar value if has_aux is True.
+        The function to differentiate. Should return a scalar value, or a tuple of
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -10,9 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Autograd support for local field projections using `FieldProjectionKSpaceMonitor`.
 - Function `components.geometry.utils.flatten_groups` now also flattens transformed groups when requested.
+- Differentiable `smooth_min`, `smooth_max`, and `least_squares` functions in `tidy3d.plugins.autograd`.
+- `@scalar_objective` decorator in `tidy3d.plugins.autograd` that wraps objective functions to ensure they return a scalar value and performs additional checks to ensure compatibility of objective functions with autograd. Used by default in `tidy3d.plugins.autograd.value_and_grad` as well as `tidy3d.plugins.autograd.grad`.
```
tylerflex, line 15: maybe i'd add another line separately to introduce the addition of the `grad` and `value_and_grad` (which automatically handle some tidy3d specific things under the hood).

CHANGELOG.md
```diff
@@ -10,9 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 
 ### Changed
 - `CustomMedium` design regions require far less data when performing inverse design by reducing adjoint field monitor size for dims with one pixel.
+- Calling `.values` on `DataArray` no longer raises a `DeprecationWarning` during automatic differentiation
```
tylerflex, line 20: i'll keep reading more, but I'm not sure we **need** to do this, but if it works just as well with values then that's fine.

tidy3d/components/data/data_array.py
```diff
@@ -220,15 +219,7 @@ def values(self):
...
-        if isbox(self.data):
-            warnings.warn(
-                "'DataArray.values' for automatic differentiation is deprecated, "
-                "please use 'DataArray.data' instead.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
-            return self.data
-        return super().values
+        return self.data if isbox(self.data) else super().values
```
tylerflex, line 222: so, no side effects internally from doing this? I think we discussed you were worried about it?

tidy3d/plugins/autograd/differential_operators.py
```diff
@@ -1,17 +1,53 @@
...
+            return grad_value, aux
+
+        return grad_with_aux
+    else:
+        return grad_ag(wrapped_fun)
+
+
 @unary_to_nary
 def value_and_grad(
     fun: Callable, x: ArrayLike, *, has_aux: bool = False
```
tylerflex: do we want to pass `**kwargs` to `scalar_objective`?

comment applies to scalar objective and `grad` too.

tidy3d/plugins/autograd/functions.py
```diff
@@ -575,3 +578,92 @@ def threshold(
...
+
+    Returns
+    -------
+    np.ndarray
+        The smooth minimum of the input array.
+    """
+    return -smooth_max(-x, tau, axis=axis)
+
+
+def least_squares(
```
tylerflex, line 627: i think a bit more explanation could be useful about what this returns. For example, what are the parameters being returned here? what does `func` accept (type annotation)? I assume `func` takes `params : tupl[float, ...]`? what about other args to `func` etc?

tidy3d/plugins/autograd/utilities.py
```diff
@@ -158,3 +162,69 @@ def chained(array: NDArray):
...
+    Callable
+        The decorated function.
+
+    Raises
+    ------
+    Tidy3dError
+        If the return value is not scalar or not real.
+    """
+
+    if func is None:
```
tylerflex, line 193: what is this for? if you want to do
 ```py
f = scalar_objective(has_aux=True)

@f
def objective(....)
```
?

File: tidy3d/plugins/autograd/utilities.py [hunk 7]
```diff
@@ -158,3 +162,69 @@ def chained(array: NDArray):
...
+        return lambda f: scalar_objective(f, has_aux=has_aux)
+
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        result = func(*args, **kwargs)
+        aux_data = None
+
+        # If the result is a tuple, unpack it
```
tylerflex: I'm wondering if all of this logic could be handled easier if it was just

```
try:
    result = result.item()
except ValueError as e:
    raise ValueError(...) from e

if not anp.isreal(result):
    raise Tidy3dError("An objective function's return value must be real.")

 return (result, aux_data) if aux_data is not None else result
```

but maybe there's some subtelty here?

tidy3d/plugins/autograd/utilities.py
```diff
@@ -158,3 +162,69 @@ def chained(array: NDArray):
         return reduce(lambda x, y: y(x), funcs, array)
 
     return chained
+
+
+def scalar_objective(func: Callable = None, *, has_aux: bool = False) -> Callable:
+    """Decorator to modify a function to return the `.data` attribute of an xarray.DataArray.
+
+    If the function returns an xarray.DataArray, this decorator extracts the `.data` attribute.
```
tylerflex: put `xarray.DataArray` in backslash quotes

tylerflex (commented):
Looks great! think at some point we should include all of these in our demo notebooks. Just a few minor comments.

tidy3d/components/data/data_array.py
yaugenst-flex: yeah this is fine. `values` simply does a `np.asarray(self.data)`, which we don't need to do here because if `data` is an `ArrayBox`, then it already is (or was) a numpy array and no casting is necessary.

tidy3d/plugins/autograd/functions.py
yaugenst-flex: improved the docstring and added an example

tidy3d/components/data/data_array.py
tylerflex: sounds good

tidy3d/plugins/autograd/differential_operators.py
yaugenst-flex: not really sure what you mean here? `grad_ag(wrapped_fun)` is still a callable to which you have to pass all function arguments. or am I misunderstanding?

tidy3d/plugins/autograd/utilities.py
yaugenst-flex: yeah, although I look at it mostly through the decorator syntax:
```python
@scalar_objective  # implicit has_aux=False
def objective(x):
    ...
```
vs
```python
@scalar_objective(has_aux=True)
def objective(x):
    ...
```

tidy3d/plugins/autograd/differential_operators.py
tylerflex: what I mean is that it seems like the other kwargs passable to `value_and_grad` and `grad` are not supported with the autograd wrapper here? is that right?

```
def grad(fun, argnum=0, *nary_op_args, **nary_op_kwargs)
```

yaugenst-flex: oooh yeah good point.. yeah this looks a bit tricky I'll try tomorrow. but agree this should be added

tidy3d/plugins/autograd/utilities.py
yaugenst-flex: this won't work for example if the return is a float.. I changed the logic a bit but overall I think most of these checks are necessary if we want to have this work with arbitrary functions

tidy3d/plugins/autograd/differential_operators.py
yaugenst-flex: ok I modified this to exactly follow the definitions of the original operators (e.g. [for grad](https://github.com/HIPS/autograd/blob/b8a8db276ae014220b91b4f4ee66a967ff02157c/autograd/differential_operators.py#L24)) but with support for aux data and the scalar objective wrapping

tidy3d/plugins/autograd/utilities.py
tylerflex: ah ok sounds good

tylerflex (approved):

e-g-melo:
Hi @yaugenst-flex! I wonder if we can serialize the `softmin` and `softmax` functions. It would be nice to include them in GUI.

yaugenst-flex:
Hey @e-g-melo, the thing is that we currently do not have support for n-ary operators. Currently, every metric is restricted to be scalar. I will look into it though, this is on the todo list.

e-g-melo:
> Hey @e-g-melo, the thing is that we currently do not have support for n-ary operators. Currently, every metric is restricted to be scalar. I will look into it though, this is on the todo list.

I've created a GUI version of [this notebook](https://www.flexcompute.com/tidy3d/examples/notebooks/InverseDesign/), which has a different `softmin` function.

image

In GUI, we can make it like this:

image

`3*(m1*exp(-m1)/(exp(-m1)+exp(-m2)+exp(-m3)) + m2*exp(-m2)/(exp(-m1)+exp(-m2)+exp(-m3)) + m3*exp(-m3)/(exp(-m1)+exp(-m2)+exp(-m3)))`

I wonder if we could have something simple, like `softmin([m1, m2, m3])`

e-g-melo:
Oh.. I saw your comment on Tom's notebook. Anyway, we can wait, no worries.

yaugenst-flex:
Yeah we can definitely add that, and it's also the goal to have these more generic functions. But I still have to figure out a good interface for them..

PR #2056, opened 2024-11-04
refactor gradient calculations to use `DerivativeSurfaceMesh`

### LINKED ISSUE #1111: Generalize VJP calculation
Potentially useful to speed things up, will prototype one possible approach:
* Get coordinates for all integration points in the structure. 
* Downsample E fwd and adj to the integration coordinates.
* Continue as normal, selecting instead of interpolating.

### LINKED ISSUE #1117: Adjoint `PolySlab.slab_bounds` support
Add the ability to differentiate with respect to the `PolySlab.slab_bounds`.

Will involve computing the VJP for this field using the adjoint shifting boundary formalism. 

Then this VJP value needs to be returned properly in the `JaxPolySlab.store_vjp()` method.

See some hints in the medium VJP functions for how to handle arbitrary shape evaluation.

Also, add a test to the backend adjoint tests to confirm the gradient accuracy.

### LINKED ISSUE #1985: autograd support for `PolySlab.slab_bounds`
Would be a good idea to include this, in case users want to compute sensitivity with respect to polyslab bounds.

Should be pretty straightforward (famous last words :P)

tylerflex:
Part 1 of an effort to generalize the VJP calculations so the Geometry and Medium just need to construct a single `DerivativeSurfaceMesh` for every field. The `DerivativeInfo` then has a method that computes derivatives given a `DerivativeSurfaceMesh`.

### DIFF
--- a/tidy3d/components/autograd/derivative_utils.py
+++ b/tidy3d/components/autograd/derivative_utils.py
@@ -5,6 +5,7 @@ import numpy as np
 import pydantic.v1 as pd
 import xarray as xr
 
+from ...constants import LARGE_NUMBER
 from ..base import Tidy3dBaseModel
 from ..data.data_array import ScalarFieldDataArray, SpatialDataArray
 from ..types import ArrayLike, Bound, tidycomplex
@@ -17,7 +18,30 @@ PermittivityData = dict[str, ScalarFieldDataArray]
 
 
 class DerivativeSurfaceMesh(Tidy3dBaseModel):
-    """Stores information about the surfaces of an object to be used for derivative calculation."""
+    """Stores information about the surfaces of an object to be used for derivative calculation.
+
+    User guide: this class is used to construct derivatives with respect to surface elements
+    given some forward and adjoint fields stored in a ``DerivativeInfo`` class. To use it,
+    you must specify the central locations of all the surface elements in ``centers``,
+    along with the area of each element in ``areas``. Then you need to specify three orthogonal
+    vectors (``normals``, ``perps1`` and ``perps2``). The derivative will be computed with
+    respect to a change in material along the ``normals`` direction. It is important to note
+    that the sign of these basis vectors is irrelevant since we end up multiplying the forward
+    and adjoint fields together in these bases.
+
+    The gradient is with respect to a change from the surface element permittivity from the
+    ``eps_in`` to ``eps_out`` field. So in principle it is always computing gradients with
+    respect to an infinitesimal outward shift in the normal direction. For example, for
+    ``PolySlab.vertices``, this means shifting each edge to the background.
+    For ``PolySlab.slab_bounds``, this means shifting the bound in either + or - direction if it
+    is index ``[1]`` or ``[0]`` respectively. This renders the sign of the normal vector
+    irrelevant.
+
+    After this ``DerivativeSurfaceMesh`` is constructed, given some ``DerivativeInfo`` in the
+    gradient calculation, a call to ``DerivativeInfo.grad_surfaces(x: DerivativeSurfaceMesh)``
+    will return the gradient with respect to a change in all of the provided surfaces.
+
+    """
 
     centers: ArrayLike = pd.Field(
         ...,
@@ -139,6 +163,13 @@ class DerivativeInfo(Tidy3dBaseModel):
         description="Bounds corresponding to the structure, used in ``Medium`` calculations.",
     )
 
+    bounds_intersect: Bound = pd.Field(
+        ...,
+        title="Geometry and Simulation Intersections Bounds",
+        description="Bounds corresponding to the minimum intersection between the "
+        "structure and the simulation it is contained in.",
+    )
+
     frequency: float = pd.Field(
         ...,
         title="Frequency of adjoint simulation",
@@ -191,16 +222,15 @@ class DerivativeInfo(Tidy3dBaseModel):
 
         if self.eps_no_structure is not None:
             return self.evaluate_eps(spatial_coords, is_inside=False)
-    
+
         return self.eps_out
 
-
-    @property
     def delta_eps(self, spatial_coords: np.ndarray) -> complex | np.ndarray:
+        """Change in the permittivity across interface (for E field grads)."""
         return self._eps_in(spatial_coords) - self._eps_out(spatial_coords)
 
-    @property
     def delta_eps_inv(self, spatial_coords: np.ndarray) -> complex | np.ndarray:
+        """Change in 1 / permittivity across interface (for D field grads)."""
         return 1.0 / self._eps_in(spatial_coords) - 1.0 / self._eps_out(spatial_coords)
 
     def grad_surfaces(self, surface_mesh: DerivativeSurfaceMesh) -> dict:
@@ -240,8 +270,8 @@ class DerivativeInfo(Tidy3dBaseModel):
         E_der_perp2 = E_fwd_perp2 * E_adj_perp2
 
         # approximate permittivity in and out
-        delta_eps_inv = self.delta_eps_inv
-        delta_eps = self.delta_eps
+        delta_eps_inv = self.delta_eps_inv(spatial_coords=spatial_coords)
+        delta_eps = self.delta_eps(spatial_coords=spatial_coords)
 
         # put together VJP using D_normal and E_perp integration
         vjps = 0.0
@@ -288,19 +318,20 @@ class DerivativeInfo(Tidy3dBaseModel):
 
         interp_kwargs = {}
         for dim, locations_dim in zip("xyz", (xs, ys, zs)):
-            # only include dims where the data has more than 1 coord, to avoid warnings and errors
-            if True or all(np.array(fld.coords).size > 1 for fld in fld_dataset.values()):
-                interp_kwargs[dim] = xr.DataArray(locations_dim, dims=edge_index_dim)
+            # filter out any infinity values to use LARGE_NUMBER and get 0 in the interp()
+            locations_dim = np.nan_to_num(locations_dim, posinf=LARGE_NUMBER, neginf=-LARGE_NUMBER)
+            interp_kwargs[dim] = xr.DataArray(locations_dim, dims=edge_index_dim)
 
         components = {}
         for fld_name, arr in fld_dataset.items():
             arr_interp = arr.interp(
                 **interp_kwargs,
                 assume_sorted=True,
-                kwargs={"bounds_error": False, "fill_value": None},
+                kwargs=dict(fill_value=None, bounds_error=False),
             )
             if "f" in arr_interp.coords:
                 arr_interp = arr_interp.sum("f")
+
             components[fld_name] = arr_interp
 
         return components
--- a/tidy3d/components/autograd/derivative_utils.py
+++ b/tidy3d/components/autograd/derivative_utils.py
@@ -7,13 +7,47 @@ import xarray as xr
 
 from ..base import Tidy3dBaseModel
 from ..data.data_array import ScalarFieldDataArray, SpatialDataArray
-from ..types import Bound, tidycomplex
+from ..types import ArrayLike, Bound, tidycomplex
 from .types import PathType
 from .utils import get_static
 
 # we do this because importing these creates circular imports
 FieldData = dict[str, ScalarFieldDataArray]
 PermittivityData = dict[str, ScalarFieldDataArray]
+
+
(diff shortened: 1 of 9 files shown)


tylerflex:
Let's also get this one in. Looks like I have a good amount of merge conflicts to sort out first, but I'll work on that today and re-test

tylerflex:
did a bit of testing of this and it seems like it still works. going to leave it unsquashed for now but should be ready to go after reviews @momchil-flex @yaugenst-flex

tylerflex:
I just noticed an issue with the epsilon grabbing from the other PR after the rebase, so don't merge this yet as I'll fix it tomorrow morning

tylerflex:
alright fixed it, turned out when there were `inf` in the `PolySlab.slab_bounds` or `Cylinder.length`, some weird things would happen after the rebase. These are handled properly now.

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -105,6 +162,14 @@ class DerivativeInfo(Tidy3dBaseModel):
         description="Bounds corresponding to the structure, used in ``Medium`` calculations.",
     )
 
+    bounds_intersect: Bound = pd.Field(
+        ...,
+        title="Geometry and Simulation Intersections Bounds",
+        description="Bounds corresponding to the minimum intersection between the "
+        "structure and the simulation it is contained in."
+        "",
```
yaugenst-flex: Additional quotes slipped int?

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -199,23 +314,22 @@ def evaluate_flds_at(
         """Compute the value of an dict with keys Ex, Ey, Ez at a set of spatial locations."""
 
         xs, ys, zs = spatial_coords.T
-        edge_index_dim = "edge_index"
+        edge_index_dim = "i"
```
yaugenst-flex: Feels like `edge_index` is better than `i`?

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -199,23 +314,22 @@ def evaluate_flds_at(
...
+            locations_dim = np.nan_to_num(locations_dim, posinf=LARGE_NUMBER, neginf=-LARGE_NUMBER)
+            interp_kwargs[dim] = xr.DataArray(locations_dim, dims=edge_index_dim)
 
         components = {}
         for fld_name, arr in fld_dataset.items():
             arr_interp = arr.interp(
-                **interp_kwargs,
-                assume_sorted=True,
-                kwargs={"bounds_error": False, "fill_value": None},
+                **interp_kwargs, assume_sorted=True, kwargs=dict(fill_value=None)
```
yaugenst-flex: I think this still needs `bounds_error=False`, because according to [xarray's docs](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.interp.html), `da.interp()` will choose the default value from either [interp1d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html) or [interpn](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interpn.html), the latter has a default of `bounds_error=True`, so this will fail for >1d interpolation.

tidy3d/components/data/sim_data.py
```diff
@@ -1115,11 +1115,21 @@ def process_adjoint_sources(self, adj_srcs: list[SourceType]) -> AdjointSourceIn
...
             log.info("Adjoint source creation: one unique frequency, no normalization.")
+            freqs_adj = self.simulation.freqs_adjoint
+
+            # if many adjoint freqs, but only 1 unique, need to mask out the non-contributors
+            if len(freqs_adj) > 1:
+                coords = dict(f=freqs_adj)
+                data = [1 if f == tuple(unique_freqs)[0] else 0 for f in freqs_adj]
+                post_norm = xr.DataArray(data, coords=coords)
+                return AdjointSourceInfo(sources=adj_srcs, post_norm=post_norm, normalize_sim=True)
+
```
yaugenst-flex: Unrelated to this PR, but this is not in `pre/2.8`, we should probably do another merge from develop? We probably shouldn't be doing that merge together with this PR.

tidy3d/components/geometry/base.py
```diff
@@ -295,8 +296,8 @@ def intersects(
             Whether the rectangular bounding boxes of the two geometries intersect.
         """
 
-        self_bmin, self_bmax = self.bounds
-        other_bmin, other_bmax = other.bounds
+        self_bmin, self_bmax = getval(self.bounds)
+        other_bmin, other_bmax = getval(other.bounds)
```
yaugenst-flex: Why is this necessary? Shouldn't the code below still work either way?

yaugenst-flex (commented):
Really nice to have this 🎉
Only some minor comments. Biggest issue for me is, this is basically a merge from develop right? I think we should do that separately, there are a lot of unrelated changes in here.

momchil-flex:
Yeah @tylerflex I think you need to drop the first 6 commits (then maybe clean up the rest?)

tidy3d/components/autograd/derivative_utils.py
tylerflex: ah yea, actually this snuck in because I was testing whether xarray was freaking out with a dim with more than one character, will revert.

tidy3d/components/geometry/base.py
tylerflex: this might have been a relic from when I was testing some validation errors when slab_bounds were traced. maybe we won't need them now.

tidy3d/components/autograd/derivative_utils.py
tylerflex: hm, but the thing is that I'm doing > 1d interpolation here,  and often out of bounds, so why wouldn't it be failing?

tidy3d/components/autograd/derivative_utils.py
tylerflex: fixed

tidy3d/components/autograd/derivative_utils.py
tylerflex: well it worked with `bounds_error=False` so I'll just put it in for extra security

tylerflex:
ok I cleaned up the commits into just 2, addressed Yannick's comments, but now ruff is complaining (but not locally, so that's a bit weird)

tidy3d/components/autograd/derivative_utils.py
yaugenst: Ok interesting, wouldn't have expected that. But still yeah I think better to keep it explicit.

yaugenst-flex (approved):

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -16,6 +17,62 @@
...
+    """Stores information about the surfaces of an object to be used for derivative calculation.
+
+    user guide: this class is used to construct derivatives with respect to surface elements
+    given some forward and adjoint fields stored in a ``DerivativeInfo`` class. To use it,
+    you must specify the central locations of all the surface elements in ``centers``,
+    along with the area of each element in ``areas``. Then you need to specify three orthogonal
+    vectors (``normals``, ``perps1`` and ``perps2``). The derivative will be computed with
+    respect to a change in material along the ``normals`` direction. It is important to note
+    that the sign of these basis vectors is irrelevant since we end up multiplying the forward
+    and adjoint fields together in these bases.
```
momchil-flex, line 30: Maybe move the last paragraph here to further clarify this sign/direction point? I first wrote a comment about it and then saw you've answered it in the last paragraph.

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -16,6 +17,62 @@
 PermittivityData = dict[str, ScalarFieldDataArray]
 
 
+class DerivativeSurfaceMesh(Tidy3dBaseModel):
+    """Stores information about the surfaces of an object to be used for derivative calculation.
+
+    user guide: this class is used to construct derivatives with respect to surface elements
```
momchil-flex: Capitalize `user`?

tidy3d/components/structure.py
```diff
@@ -248,34 +248,25 @@ def get_monitor_name(index: int, data_type: str) -> str:
         return monitor_name_map[data_type]
 
     def make_adjoint_monitors(
-        self, freqs: list[float], index: int
+        self, freqs: list[float], index: int, field_keys: list[str]
```
momchil-flex, line 251: Whenever changes are made to the frontend autograd components we need to be careful about the backend too. Seems like [this](https://github.com/flexcompute/tidy3d-core/blob/000aab3e18c9648fcee4ef93195bb76aaaed5a48/tidy3d_backend/metadata.py#L650) needs to be updated and I'm not sure how?

tidy3d/components/geometry/polyslab.py
```diff
@@ -137,7 +140,7 @@ def no_complex_self_intersecting_polygon_at_reference_plane(cls, val, values):
         For 2), we heal the polygon, and warn that the polygon has been cleaned up.
         """
         # no need to validate anything here
-        if isclose(values["dilation"], 0):
+        if np.isclose(values["dilation"], 0):
```
momchil-flex: So at some point we purposefully changed some validators to use `isclose` instead of `np.isclose` because the latter creates a numpy array which can have significant overhead if you e.g. have a metalens with a million polyslabs. We have [this test](https://github.com/flexcompute/tidy3d-core/blob/000aab3e18c9648fcee4ef93195bb76aaaed5a48/tests/test_simulation.py#L8-L38) on the backend about this (it could easily be moved to the frontend if we wanted). I tried a few times using `pre/2.8` and it would run in about 6.5-7s. With your branch, it runs in about 8.5-9s, which is potentially acceptable but still somewhat significant. Increasing the number of geometries seems to have a linear scaling so it's always about 30% slower.

This consideration was definitely much more important in some of the mesher code, where the scaling is with number of structures squared. But yeah even here it has an effect.

Is it necessary to change to numpy?

tidy3d/components/autograd/derivative_utils.py
tylerflex: sure, sounds good. I think I also ended up writing that last paragraph because after re-reading I realized it needed more clarification, so makes sense to put there.

tidy3d/components/autograd/derivative_utils.py
tylerflex: done

tidy3d/components/structure.py
tylerflex: ah, right. I think the fix is to pass the `sim_fields_keys` into the `make_adjoint_monitors`. Since you have the optimizer, that would mean getting

```py
sim = optimizer.design.simulation
traced_fields_sim = setup_run(simulation=sim)
sim_fields_keys = list(traced_fields_sim.keys())
```

momchil-flex: I don't think `setup_run` will work if this is not called inside a `grad` call?

tidy3d/components/geometry/polyslab.py
tylerflex: I changed them all back and things are passing, so fine by me. I think originally I had changed them because I was running into validation errors with `slab_bounds` and it was really hard to pinpoint what was failing. and so I ended up basically changing a ton of things and removing validators before I found it. this must have slipped into the PR.

tidy3d/components/structure.py
tylerflex: not sure, guess you could just do

```

    return simulation.strip_traced_fields(
        include_untraced_data_arrays=False, starting_path=("structures",)
    )

```
which is what setup run does.

momchil-flex: But the point is that there are no traced fields since it's not a simulation called in `grad` no?

momchil-flex: Basically what we're trying to do there is get the gradient monitors from the static definition of the InverseDesign object.

tylerflex: ah ok. yea I guess we dont know. but we can effectively "fake" it then, because we know it's a Box with a CustomMedium and we need the `permittivity` data.

So I think one approach could just be to pass
```
field_keys=(f'structures/{(len(sim.structures) - 1}/medium/permittivity', )
```
into the `make_adjoint_monitors` call. Or you can just make the monitors manually since again we know it's just a box so it needs a similarly sized field and permittivity monitor

momchil-flex: Using the first approach, works for now, obviously should be dealt with better in the future..

PR #2058, opened 2024-11-06
Use default `td.ModeSpec` in mode monitors

yaugenst-flex:
How do we feel about this? :smile:

### DIFF
--- a/tidy3d/components/monitor.py
+++ b/tidy3d/components/monitor.py
@@ -285,7 +285,7 @@ class AbstractModeMonitor(PlanarMonitor, FreqMonitor):
     """:class:`Monitor` that records mode-related data."""
 
     mode_spec: ModeSpec = pydantic.Field(
-        ...,
+        ModeSpec(),
         title="Mode Specification",
         description="Parameters to feed to mode solver which determine modes measured by monitor.",
     )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-[Unreleased]
+## [Unreleased]
+
+### Changed
+- `ModeMonitor` and `ModeSolverMonitor` now use the default `td.ModeSpec()` with `num_modes=1` when `mode_spec` is not provided.
 
 ### Fixed
-
 - NumPy 2.1 compatibility issue where `numpy.float64` values passed to xarray interpolation would raise TypeError.
 
 ## [2.8.0rc1] - 2024-12-17
--- a/tests/test_components/test_monitor.py
+++ b/tests/test_components/test_monitor.py
@@ -355,16 +355,16 @@ def test_monitor():
     m5 = td.ModeMonitor(
         size=(1, 1, 0), center=center, mode_spec=td.ModeSpec(), freqs=FREQS, name="test_mon"
     )
-    m6 = td.ModeSolverMonitor(
+    m6 = td.ModeMonitor(size=(1, 1, 0), center=center, freqs=FREQS, name="test_mon")
+    m7 = td.ModeSolverMonitor(
         size=(1, 1, 0),
         center=center,
-        mode_spec=td.ModeSpec(),
         freqs=FREQS,
         name="test_mon",
         direction="-",
     )
-    m7 = td.PermittivityMonitor(size=size, center=center, freqs=FREQS, name="perm")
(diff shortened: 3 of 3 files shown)


tidy3d/components/monitor.py
```diff
@@ -279,7 +279,7 @@ class AbstractModeMonitor(PlanarMonitor, FreqMonitor):
     """:class:`Monitor` that records mode-related data."""
 
     mode_spec: ModeSpec = pydantic.Field(
-        ...,
+        default_factory=ModeSpec,
```
tylerflex: generally we just do `ModeSpec()`  (no default factory). not sure if there's a substantial difference here, but i just dont know how the docs will handle default factories

momchil-flex:
I'm ok with this but would rather put it in 2.8. But simply changing the base makes a mess, could you try cherry-picking your commit?

yaugenst-flex:
Sure yeah, should be fine now.

tylerflex (approved):
i'm fine with it in 2.8 too. do we want to also add a `mode_index: int = None` and `direction : str = None` options? basically it would mean that we pre-select certain quantities from the data on the backend? and would make it more similar to ModeSource. If left None we just include all.

We could do this separately as well.

tidy3d/components/monitor.py
yaugenst-flex: Should I change it or do we want to check how the docs look? Generally I would've thought that it's best to instantiate defaults (if they are classes) only on instantiation of the class, otherwise 1. you get object creation on import, which slows imports down and 2. it leads to shared state between different mode monitor instances. All new mode monitors will share the same default instance of `mode_spec`, and if you happen to make a change that modifies this default instance (let's say you write something into `attrs`), then this change will affect all other mode monitors too.

yaugenst-flex:
> do we want to also add a mode_index: int = None and direction : str = None options? basically it would mean that we pre-select certain quantities from the data on the backend? and would make it more similar to ModeSource. If left None we just include all.

i would suggest we do this separately _if_ we decide to go for it. should probably avoid rushing that change, i'm also not so sure about it anymore...

tidy3d/components/monitor.py
tylerflex: i guess we could see how the docs looks, just might require some modification of [this](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/base.py#L1056-L1152)  method that sets the docstring, we'd need to add handling for `defaultfactory`. Also cc'ing @daquinteroflex just so he knows since he's working on a sphinx plugin related to this.

yaugenst-flex: I took a stab at this in 1eafd9d which I think would correctly handle cases where we just use a class as a default factory. It wouldn't handle lambdas nicely, although if necessary I can also add support for that (by just printing the lambda definition).

momchil-flex:
Gotta decide if we want to do this. I'm OK with it but would probably avoid adding `mode_index` and `direction` for now.

tylerflex:
yea I'm ok with it, just note PR needs probably to drop a commit before merging

tylerflex:
oh wait, I guess this default factory change is supposed to be in here? maybe slightly confused, but on making modeSpec optional, I'm ok with it if it makes the user experience better overall. so in summary:

give default value:
* 1 less field to specify 

keep required:
* ensures the user at least knows about ModeSpec so they can set more than 1 mode index if they desire. otherwise they might run a sim with default and only realize after that we only store 1st mode.

yaugenst-flex:
Ok how about the following:

1. I'll revert the `default_factory` change and instead initialize the default `ModeSpec` directly as we do elsewhere (we should revisit this though).
2. I will add a warning if no `ModeSpec` was supplied by the user and the default is used, something along the lines of `No "ModeSpec" was provided, using only fundamental mode`.

I am not 100% on board with 2 though, mostly because we _already_ allow the construction of a default `ModeSpec()` (without any arguments). This is actually what led me to propose this change in the first place.

momchil-flex:
Yeah I'm not too keen on the warning either, it kills the point a little because the way to suppress it is to do what we do now, i.e. pass ModeSpec() explicitly. So I wouldn't want to e.g. modify our examples to use default because it would then introduce warnings. :)

I think getting the fundamental mode amplitude covers enough use cases for me to be OK with this being a default even if it risks sometimes users not realizing they haven't set things right (e.g. if they want more modes).

yaugenst-flex:
Alright sounds good to me.

tylerflex:
alright I'm also on board with this and agree regarding avoiding that extra warning

tylerflex:
if users complain we can change it back

yaugenst-flex:
ok great. @momchil-flex could you approve, for completeness? 😄

momchil-flex (approved):

PR #2060, opened 2024-11-06
feat: add s3 config to download_file call.

pperanich:
Adding `_s3_config` specified in s3utils into the `download_file` call, similar to `upload_fileobj`.

### DIFF
--- a/tidy3d/web/core/s3utils.py
+++ b/tidy3d/web/core/s3utils.py
@@ -335,6 +335,7 @@ def download_file(
             Filename=str(to_file),
             Key=token.get_s3_key(),
             Callback=_callback,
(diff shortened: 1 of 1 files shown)


momchil-flex:
@magiWei could you take a look?

lei-flex (approved):
good catch. ship it.

yaugenst-flex:
Hi @pperanich,

Thanks for your contribution! I was wondering if you could provide some context on why you added the S3 config to the `download_file` call. Was there a specific issue or necessity on your end that prompted this change?

Just trying to understand the motivation so we can ensure everything is aligned properly. Appreciate your insights!

PR #2061, opened 2024-11-07
Support for adjoint simulations with no monitors (zero gradients)

yaugenst-flex:
Idea is to allow functions like `min()` and `max()` in an objective function where some simulations (for example in an adjoint batch) might not get adjoint sources.

Approach:
 - Generally allow `make_adjoint_sim` to return `None` instead of erroring.
 - We warn if:
   1. All simulations in `run_async` have no adjoint sources.
   2. The adjoint simulation in `run` has no adjoint sources.


tylerflex (approved):
Looks good, thanks for handling this 👍

PR #2064, opened 2024-11-08
fix polyslab edge gradient assignment

tylerflex:
There was a bug in polyslab, when transforming edge gradients to vertex gradients. The vertex assignments were shifted incorrectly by 1. This wouldnt affect finely discretized polyslabs but becomes a major issue with more coarse grained polylsab discretizations.


yaugenst-flex (approved):
Looks great. Took me a while to actually understand but can confirm, `shift=-1` connects current and next edge, which is misaligned with the vertices.

tylerflex:
Yea, basically to compute edges, we substract current vertex from vertex with shift=-1. then to get back to vertices, we need to apply shift=+1

PR #2065, opened 2024-11-12
Set minimum PML layers to 6

momchil-flex:
Reported by MC as a possible improvement. In practice unlikely anything smaller than 6 would make physical sense, and even 6 is probably too little. Had to update various tests.


tylerflex (approved):
Looks good, only thing I'd sugest is making the `6` a module level constant if it would help down the line if this changes.

PR #2067, opened 2024-11-13
Passing kwargs in Lorentz.from_nk

momchil-flex:
One last small thing noticed by MC...


tylerflex (approved):
looks good, maybe add kwargs to the docstring as well?

momchil-flex:
Done, they were not present in the docstring of any of the methods... (this may be an omission we have in various other places)

tylerflex:
I might be wrong but now it looks like the kwarg passing part of the code is no longer present?

tylerflex (approved):
I was wrong.  feel free to merge

PR #2068, opened 2024-11-13
Add warning to complex-field nonlinearities

caseyflex:
Some of the nonlinearities use complex fields to realize certain effects, which may be a good model in certain circumstances but can deviate from the underlying physics. I added some warnings to make sure that the user knows what they are doing (and doesn't inadvertently use KerrNonlinearity when NonlinearSusceptibility is more appropriate, for example). These warnings will always show whenever complex-field nonlinearities are present.

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -442,6 +442,14 @@ class TwoPhotonAbsorption(NonlinearModel):
 
     def _validate_medium(self, medium: AbstractMedium):
         """Check that the model is compatible with the medium."""
+        log.warning(
+            "Found a medium with a 'TwoPhotonAbsorption' nonlinearity. "
+            "This uses a phenomenological model based on complex fields, "
+            "so care should be taken in interpreting the results. For more "
+            "information on the model, see the documentation at "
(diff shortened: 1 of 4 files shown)


tidy3d/components/medium.py
```diff
@@ -442,6 +442,11 @@ def _hardcode_medium_freqs(
 
     def _validate_medium(self, medium: AbstractMedium):
         """Check that the model is compatible with the medium."""
+        log.warning(
+            "Found a medium with a 'TwoPhotonAbsorption' nonlinearity. "
+            "This uses a phenomenological model based on complex fields, "
+            "so care should be taken in interpreting the results."
```
momchil-flex: Can we reference something for more information?

momchil-flex (approved):

tomflexcompute (approved):
Thanks Casey.

FilipeFcp (approved):
Nice. This will be useful for the users.

tidy3d/components/medium.py
caseyflex: done

PR #2072, opened 2024-11-14
Fix potential self-intersection polygon in Scene._filter_structures_plane

weiliangjin2021:
This fixes the shapely/GEOS error in mode solver metadata pipeline when trying to find materials on the modal plane.

Hard to fabricate a test myself so I'll leave it out.

### DIFF
--- a/tidy3d/components/scene.py
+++ b/tidy3d/components/scene.py
@@ -637,7 +637,7 @@ class Scene(Tidy3dBaseModel):
                 if _shape.is_empty or not shape.intersects(_shape):
                     continue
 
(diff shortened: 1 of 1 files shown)


lucas-flexcompute (approved):
Happens all the time with shapely… I'm guessing `.buffer(0)` is better than `make_valid(...)` in this case, is that so?

weiliangjin2021:
> Happens all the time with shapely… I'm guessing `.buffer(0)` is better than `make_valid(...)` in this case, is that so?

`make_valid` does nothing in this case.

PR #2075, opened 2024-11-18
Show WebAPI system warning message.

QingengWei:


QingengWei:
@tylerflex @momchil-flex This requirement is from @Emerson: https://flow360.atlassian.net/browse/SCEM-8018

tidy3d/web/core/log.py
tylerflex, line 1: this is just copied directly from `tidy3d/log.py`?  I think this is not a good idea since we will need to maintain two identical files. is there a better approach?

QingengWei: Maybe we can move  tidy3d/log.py  to tidy3d/core/log.py.

tylerflex: The problem with this is that the log is used internally in the tidy3d components. So if we move it to `web` then we would be forced to import `web` (and do authentication and other things) to use `tidy3d`. It wold be nice if they were separate. What about just using the rich console to do the logging here? like we do in `web.monitor`?

momchil-flex: I am actually confused, why is this a problem? `tidy3d/log.py` is top level for both the `components` and `web` modules, why can't both just import it?

QingengWei: According to the original agreement, web/core is the bottom layer and cannot depend on other components, otherwise it will cause circular dependencies.

QingengWei: > The problem with this is that the log is used internally in the tidy3d components. So if we move it to `web` then we would be forced to import `web` (and do authentication and other things) to use `tidy3d`. It wold be nice if they were separate. What about just using the rich console to do the logging here? like we do in `web.monitor`?

I think a rich console is a good idea. It just be used in tidy3d/core

momchil-flex: > According to the original agreement, web/core is the bottom layer and cannot depend on other components, otherwise it will cause circular dependencies.

I see, but is this really a problem for the log or rather specifically for things in `components/` and `plugins/`. I could see logic behind things we put in the main `tidy3d` folder to be usable by all other modules, i.e. both components and web. @tylerflex ?

tylerflex: The way it was designed, the web API and tidy3d could be agnostic of one another. And just communicate through this "stub" class. In principle I think we could import log into web/core, but for now I'd prefer using rich console to not break this architecture unless we really need to. In the future we can figure out a better long term solution to decouple these things. probably the web API needs to be its own package and it might just need to import tidy3d as a dependency.

momchil-flex: Ok, now I see, I think then to me it sounds like this warning should ideally be propagated from `web/core` to the corresponding `web/api` function where it occurs, and just shown there using the usual log. But for the time being rich console sounds good.

QingengWei: Ok, I have changed to rich console.

QingengWei: > The way it was designed, the web API and tidy3d could be agnostic of one another. And just communicate through this "stub" class. In principle I think we could import log into web/core, but for now I'd prefer using rich console to not break this architecture unless we really need to. In the future we can figure out a better long term solution to decouple these things. probably the web API needs to be its own package and it might just need to import tidy3d as a dependency.

Great idea! We shouldn't embed web/* into tidy3d, as some users may not utilize the web functionality. Instead, we can separate web/* into its standalone package that depends on tidy3d. This way, users can choose the appropriate package according to their needs.

tylerflex (changes requested):
Hi @magiWei Looks good overall but see how we do the logger in other parts of web/core

https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/web/core/environment.py#L93-L106

let's use this `get_logger()` for consistency and so we can change the base config and have it reflect everywhere. 

Also wonder if momchil's suggestion about just handing the warning in the `web/api` side makes some sense? but for now maybe this is sufficient
Thanks

QingengWei:
> Hi @magiWei Looks good overall but see how we do the logger in other parts of web/core
> 
> https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/web/core/environment.py#L93-L106
> 
> let's use this `get_logger()` for consistency and so we can change the base config and have it reflect everywhere.
> 
> Also wonder if momchil's suggestion about just handing the warning in the `web/api` side makes some sense? but for now maybe this is sufficient Thanks

Ok, it makes sense to me. Changed.

tylerflex (approved):

PR #2076, opened 2024-11-18
fix gradient for single freq selection from multi freq mnt

tylerflex:


tidy3d/components/data/sim_data.py
```diff
@@ -1115,11 +1115,21 @@ def process_adjoint_sources(self, adj_srcs: list[SourceType]) -> AdjointSourceIn
...
         if num_unique_freqs == 1:
             log.info("Adjoint source creation: one unique frequency, no normalization.")
+            freqs_adj = self.simulation.freqs_adjoint
+
+            # if many adjoint freqs, but only 1 unique, need to mask out the non-contributors
+            if len(freqs_adj) > 1:
+                coords = dict(f=freqs_adj)
+                data = [1 if f == tuple(unique_freqs)[0] else 0 for f in freqs_adj]
+                post_norm = xr.DataArray(data, coords=coords)
+                return AdjointSourceInfo(sources=adj_srcs, post_norm=post_norm, normalize_sim=True)
```
yaugenst-flex, line 1131: So `post_norm` is either `0` or `1`, and it's only `1` if the frequency is the central frequency of the adjoint source? Is the adjoint `source_time.freq0` always equal to the single frequency that was selected out?

yaugenst-flex (commented):
Looks great!

tidy3d/components/data/sim_data.py
tylerflex: yes to all.

yaugenst-flex (approved):

PR #2080, opened 2024-11-19
auto grab permittivity inside and outside `PolySlab` and `Cylinder`

### LINKED ISSUE #1121: Grab boundary permittivity a few grid cells away automatically

tylerflex:
Fixes https://github.com/flexcompute/tidy3d/issues/1121

Approach is to call `Simulation.epsilon` without the structure present and then store this information in the `DerivativeInfo`. Then the gradient calculation can interpolate this data where it's needed, which gives a pretty good prox y to the permittivity outside the structture at that location.

Permittivity inside can still just use the `Structure.medium.eps_model`

Implemented for `PolySlab` but not yet for `Box`.

### DIFF
--- a/tidy3d/web/api/autograd/autograd.py
+++ b/tidy3d/web/api/autograd/autograd.py
@@ -857,10 +857,34 @@ def postprocess_adj(
 
         eps_in = np.mean(structure.medium.eps_model(freq_adj))
         eps_out = np.mean(sim_data_orig.simulation.medium.eps_model(freq_adj))
+        if structure.background_medium:
+            eps_background = structure.background_medium.eps_model(freq_adj)
+        else:
+            eps_background = None
 
         # manually override simulation medium as the background structure
-        if structure.background_medium is not None:
-            eps_out = structure.background_medium.eps_model(freq_adj)
+        if not isinstance(structure.geometry, td.Box):
+            # auto permittivity detection
+            sim_orig = sim_data_orig.simulation
+            plane_eps = eps_fwd.monitor.geometry
+
+            # get permittivity without this structure
+            structs_no_struct = list(sim_orig.structures)
+            structs_no_struct.pop(structure_index)
+            sim_no_structure = sim_orig.updated_copy(structures=structs_no_struct)
+            eps_no_structure = sim_no_structure.epsilon(box=plane_eps, coord_key="centers")
+
+            # get permittivity with structures on top of an infinite version of this structure
+            structs_inf_struct = list(sim_orig.structures)[structure_index + 1 :]
+            sim_inf_structure = sim_orig.updated_copy(
+                structures=structs_inf_struct,
+                medium=structure.medium,
+                monitors=[],
+            )
+            eps_inf_structure = sim_inf_structure.epsilon(box=plane_eps, coord_key="centers")
+
+        else:
+            eps_no_structure = eps_inf_structure = None
 
         derivative_info = DerivativeInfo(
             paths=structure_paths,
@@ -873,8 +897,11 @@ def postprocess_adj(
             eps_data=eps_fwd.field_components,
             eps_in=eps_in,
             eps_out=eps_out,
+            eps_background=eps_background,
             frequency=freq_adj,
             bounds=structure.geometry.bounds,  # TODO: pass intersecting bounds with sim?
+            eps_no_structure=eps_no_structure,
+            eps_inf_structure=eps_inf_structure,
         )
 
         vjp_value_map = structure.compute_derivatives(derivative_info)
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -1309,6 +1309,8 @@ def test_pole_residue(monkeypatch):
         eps_out=1.0,
         frequency=freq,
         bounds=((-1, -1, -1), (1, 1, 1)),
+        eps_no_structure=td.SpatialDataArray([[[1.0]]], coords=dict(x=[0], y=[0], z=[0])),
+        eps_inf_structure=td.SpatialDataArray([[[2.0]]], coords=dict(x=[0], y=[0], z=[0])),
     )
 
     grads_computed = pr.compute_derivatives(derivative_info=info)
@@ -1388,6 +1390,8 @@ def test_custom_pole_residue(monkeypatch):
         eps_out=1.0,
         frequency=freq,
         bounds=((-1, -1, -1), (1, 1, 1)),
+        eps_no_structure=td.SpatialDataArray([[[1.0]]], coords=dict(x=[0], y=[0], z=[0])),
+        eps_inf_structure=td.SpatialDataArray([[[2.0]]], coords=dict(x=[0], y=[0], z=[0])),
     )
 
     grads_computed = pr.compute_derivatives(derivative_info=info)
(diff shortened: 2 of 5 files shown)


tylerflex:
@yaugenst-flex this is causing the `TestFieldProjection` tests to give grads of `nan` and I dont understand why. Do you see anything in here that could be causing it?

yaugenst-flex:
it's because the coordinates are outside of the interpolation range,:
https://github.com/flexcompute/tidy3d/blob/06a513d1692d3ce4b8bac70ad02f03ccbe3d7b13/tidy3d/components/autograd/derivative_utils.py#L165-L168

where `self.eps_no_structure` only has `z=0.05`, but `spatial_coords` are sampled at `z=0`.

[`xarray.DataArray.interp()`](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.interp.html) will use the default `fill_value` specified in [`scipy.interpolate.interpn()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interpn.html) in that case, which is `nan`.

this can be avoided by setting `kwargs={"bounds_error": False, "fill_value": None}`, in which case the values will be extrapolated.

not sure if that is the correct solution though, probably the coordinates should be aligned before interpolation in this case. or we use `nearest` interpolation with `fill_value=None` in that situation.

tylerflex:
hm, ok I'll try this. is there a good notebook to test?

tylerflex:
OK, so I ended up with this:
* Automatically determines permittivity just inside and outside edges based on algorithm described earlier.
* Decided to leave the Box implementation for another PR. We already handle that pretty generally and it screwed up the cases where two Boxes were touching at an interface. #2084 

I think this is best of both worlds, because it leaves Box alone and enables PolySlab to work properly. Tested with PhC and Metalens notebooks.

tidy3d/web/api/autograd/autograd.py
```diff
@@ -862,6 +862,29 @@ def postprocess_adj(
...
+            sim_no_structure = sim_orig.updated_copy(structures=structs_no_struct)
+            eps_no_structure = sim_no_structure.epsilon(box=plane_eps, coord_key="centers")
+
+            # get permittivity without this structure
+            structs_inf_struct = list(sim_orig.structures)[structure_index + 1 :]
+            sim_inf_structure = sim_orig.updated_copy(
+                structures=structs_inf_struct,
+                medium=structure.medium,
+                monitors=[],
+            )
```
momchil-flex, line 883: Calling `updated_copy` on simulations always throws a red flag for me in terms of efficiency, because the simulation will get re-validated, which in some cases can be slow. In this particular case you're also doing this for every structure. I suspect this is going to get quite heavy e.g. in a metalens case where you have many structures so both re-validating each time may be costly, and it will be done many times (so at least O(N**2) scaling?)

momchil-flex: I think the answer here may be to refactor the `Simulation.epsilon` internals to have a function that can just get the epsilon of a list of structures without having to copy the simulation.

momchil-flex: Also computing over the entire simulation grid sounds like a giant overkill right? I mean, maybe we can't do it close to edges only, but at least within the structure bounding box + some spacing?

tidy3d/web/api/autograd/autograd.py
```diff
@@ -862,6 +862,29 @@ def postprocess_adj(
...
+            sim_orig = sim_data_orig.simulation
+            plane_eps = eps_fwd.monitor.geometry
+
+            # get permittivity without this structure
+            structs_no_struct = list(sim_orig.structures)
+            structs_no_struct.pop(structure_index)
+            sim_no_structure = sim_orig.updated_copy(structures=structs_no_struct)
+            eps_no_structure = sim_no_structure.epsilon(box=plane_eps, coord_key="centers")
+
+            # get permittivity without this structure
```
momchil-flex: comment is wrong

tidy3d/web/api/autograd/autograd.py
tylerflex: 1. It will be a performance hit for sure. We have plans to have an "autograd config" where we can toggle this on and off (off by default) depending on if the user needs this extra accuracy.
2. For things like metalens, it's likely to be a single structure since it's a geometry group. For example I tried this for both metalens and phc notebooks and it was negligible.
3. We can try refactoring simulation.epsilon, maybe adding a way to skip certain structures? for the inside one, I need to add the structure.medium to Simulation.medium, or make it infinitely large, which will require some modification.
4. It would be worth also considering adding a `validate: bool` argument to `updated_copy`? I thought Weiliang implemented this somewhere recently. 
5. The epsilon is not computed over the entire simulation, just over the region defined by the `.geometry` of the graidient permittivity monitor.

momchil-flex: Interesting, thanks for the clarification!

> For things like metalens, it's likely to be a single structure since it's a geometry group. For example I tried this for both metalens and phc notebooks and it was negligible.

I guess that's what we try to teach people to use, but they don't always - but fair point! So the inside and outside eps will be computed only once?

> It would be worth also considering adding a validate: bool argument to updated_copy? I thought Weiliang implemented this somewhere recently.

You're right, this is currently on pre/2.8 only. Actually should we put this in 2.8 too just to avoid unexpected side-effects? It seems like there could be some since who knows what edge cases we haven't thought about.

> The epsilon is not computed over the entire simulation, just over the region defined by the .geometry of the graidient permittivity monitor.

Ah ok I see.

tylerflex: > [For geometry group] So the inside and outside eps will be computed only once?

yes

> Actually should we put this in 2.8 too just to avoid unexpected side-effects?

I guess it's up to you, but that would be safer yea.  Maybe I should build in some logic to preferentially skip this calculation if the user still manually supplies `background_medium`.

momchil-flex: I thought that's what's gonna happen currently?

Yeah I think we should do 2.8 otherwise we're risking having to do another patch like few days later when we figure out some issue with this. :D

But wait for my merge before you change the base branch and cherry-pick this.

tylerflex:
oops, accidentally force pushed and erased your merge commit @yaugenst-flex

tidy3d/web/api/autograd/autograd.py
tylerflex: ok, no rush

yaugenst-flex:
> oops, accidentally force pushed and erased your merge commit @yaugenst-flex

yeah no problem that one was supposed to go anyways

tylerflex:
this has been rebased and tests fixed FYI

momchil-flex (approved):
Fine by me to merge after squashing the commits. @yaugenst-flex you good on it too?

yaugenst-flex (approved):
Looks good!

tylerflex:
this needs a bit more testing, dont merge it yet (just FYI)

tylerflex:
ok I think this is ready to go. tested on a handful of notebooks with different settings and it works. Btw here's the logic for how the outside permittivity is set

if Box:
* same logic as current, looks 2 grid cells from the interface for permittivity.

if PolySlab or Cylinder
* automatically grabs permittivity at interface with no structure (outside) and with infinite structure (inside)
* if the background_medium is supplied, uses this instead.

momchil-flex:
Sounds good to me!

PR #2081, opened 2024-11-20
If no sources set the runtime based on index of 1

momchil-flex:


tylerflex (approved):
Looks good. does it need a changelog?

PR #2084, opened 2024-11-20
fully automatic permittivity grabbing for adjoint gradients

tylerflex:


tylerflex:
closing this, may re-open later if we want it. for now box works fine and #2080 handles PolySlab and Cylinder

PR #2086, opened 2024-11-21
Suggest a better f_width for TerminalComponentModeler

QimingFlex:
Previous parameters result a central frequency out of the frequency range as shown, which uses f_min=0 and f_max = 50GHz. I took f_width=0.5*freq_0, which might not be optimal but helps centre the frequency at around 25GHz and improved S-parameters a bit.
image
Screenshot from 2024-11-21 16-58-17

### DIFF
--- a/tidy3d/plugins/smatrix/component_modelers/terminal.py
+++ b/tidy3d/plugins/smatrix/component_modelers/terminal.py
@@ -23,7 +23,7 @@ from ..ports.base_terminal import AbstractTerminalPort, TerminalPortDataArray
 from ..ports.coaxial_lumped import CoaxialLumpedPort
 from ..ports.rectangular_lumped import LumpedPort
 from ..ports.wave import WavePort
-from .base import FWIDTH_FRAC, AbstractComponentModeler, TerminalPortType
+from .base import AbstractComponentModeler, TerminalPortType
 
 
 class PortDataArray(DataArray):
@@ -177,9 +177,9 @@ class TerminalComponentModeler(AbstractComponentModeler):
     @cached_property
     def _source_time(self):
         """Helper to create a time domain pulse for the frequency range of interest."""
-        freq0 = np.mean(self.freqs)
-        fdiff = max(self.freqs) - min(self.freqs)
-        fwidth = max(fdiff, freq0 * FWIDTH_FRAC)
+        freq0 = (min(self.freqs) + max(self.freqs)) / 2
+        # This ensures the amplitude at the maximum frequency is around 0.1 of the amplitude at the central frequency
+        fwidth = freq0 * np.sqrt(2) / np.pi
         return GaussianPulse(
             freq0=freq0, fwidth=fwidth, remove_dc_component=self.remove_dc_component
         )


QimingFlex:
Screenshot from 2024-11-21 16-01-36
Screenshot from 2024-11-21 17-02-09

weiliangjin2021:
How about the case where `freqs` is not uniformly distributed: say freq0 is almost f_min, the new approach will result in too narrow bandwidth?

dmarek-flex:
@QimingFlex Can you also try the old way for choosing bandwidth, but use a smaller `shutoff` for the simulation. Simply using a smaller `shutoff` might remove those ripples as well.

QimingFlex:
> How about the case where `freqs` is not uniformly distributed: say freq0 is almost f_min, the new approach will result in too narrow bandwidth?

Good point! I guess just averaging and minimum and maximum by `freq0 = (min(self.freqs) + max(self.freqs)) / 2` will work?

QimingFlex:
> @QimingFlex Can you also try the old way for choosing bandwidth, but use a smaller `shutoff` for the simulation. Simply using a smaller `shutoff` might remove those ripples as well.

Screenshot from 2024-11-25 10-13-09
image (2)

Yes, it has indeed improved. These two figures show the original f_width and the updated f_width, both with a smaller shutoff factor.

weiliangjin2021:
> > How about the case where `freqs` is not uniformly distributed: say freq0 is almost f_min, the new approach will result in too narrow bandwidth?
> 
> Good point! I guess just averaging and minimum and maximum by `freq0 = (min(self.freqs) + max(self.freqs)) / 2` will work?

Sounds good. No edge cases come to my mind so far.

dmarek-flex:
> Yes, it has indeed improved. These two figures show the original f_width and the updated f_width, both with a smaller shutoff factor.

Great, so using the new `f_width` is definitely an improvement.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -177,9 +177,9 @@ def sim_dict(self) -> Dict[str, Simulation]:
     @cached_property
     def _source_time(self):
         """Helper to create a time domain pulse for the frequency range of interest."""
-        freq0 = np.mean(self.freqs)
-        fdiff = max(self.freqs) - min(self.freqs)
-        fwidth = max(fdiff, freq0 * FWIDTH_FRAC)
+        freq0 = (min(self.freqs) + max(self.freqs)) / 2
+        # This ensures the amplitude at the maximum frequency is around 0.1 of the amplitude at the central frequency
+        fwidth = freq0 * np.sqrt(2) / np.pi
```
dmarek-flex, line 182: This should be better in most cases in RF applications. However, what if the user wants a simulation, which truly has a lower bound on the minimum frequency? This `fwidth` is based only on the central frequency.

It seems like the ripple issue was because the bandwidth was larger than needed and lower frequencies had higher amplitudes. Why not use the `fdiff` calculation from the previous commit, but scale it appropriately so that the amplitude of the min and max frequency is set to some desired value, such as 0.1.

dmarek-flex (changes requested):
Also should update the changelog

momchil-flex:
What's the status here, can we move it forward?

tidy3d/plugins/smatrix/component_modelers/terminal.py
QimingFlex: Sorry, could you clarify what you mean by lower bound on the minimum frequency? I think fwidth can be determined by `freq0` alone to ensure a symmetrical spectrum and the desired levels at both min and max. It seems like `fdiff` might not be needed.

QimingFlex:
> What's the status here, can we move it forward?

Will complete by today.

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: Let's say the minimum frequency the user chooses is much larger than 0. There might be good reason for this, e.g., less padding needed for PML or maybe avoiding a resonance at lower frequencies. With this approach, changing the min frequency does not affect the bandwidth that much. As the min frequency approaches the max frequency, the signal bandwidth should drop, probably up to some minimum value ideally.

Basically using this new approach, the user cannot specify the frequency bandwidth of the input signal independent of the central frequency. 

I also realized that your comment does not seem correct for the amplitude of the maximum frequency. If you keep the max frequency the same but change the min frequency, the power spectral density of the signal at the max frequency will change, right?

I think a better approach is to calculate `freq0` as you proposed, but modify the `fwidth` calculation, so that it uses the min and max frequency to ensure some desired power spectral density at those frequencies, perhaps the value of 0.1 in your comment or 0.01 if it represents power.

dmarek-flex: I think the main issue causing ripples in the S parameters was that the bandwidth was simply too large for the central frequency. But the original general idea of calculating `fdiff` using the min and max frequency was ok, maybe just needs to be scaled differently.

QimingFlex: Ah, you are right! We need a smaller scale for `fdiff` then.

QimingFlex:
> What's the status here, can we move it forward?

@dmarek-flex @momchil-flex After a thorough investigation, I found that the main reason for the discrepancy in S-parameters is the voltage integral path. The current Gaussian source works properly, so for now, we don't need to make any changes. I will close this PR :).

weiliangjin2021:
> > What's the status here, can we move it forward?
> 
> @dmarek-flex @momchil-flex After a thorough investigation, I found that the main reason for the discrepancy in S-parameters is the voltage integral path. The current Gaussian source works properly, so for now, we don't need to make any changes. I will close this PR :).

You showed several figures in this PR, which seems to be related to fwidth?

QimingFlex:
> > > What's the status here, can we move it forward?
> > 
> > 
> > @dmarek-flex @momchil-flex After a thorough investigation, I found that the main reason for the discrepancy in S-parameters is the voltage integral path. The current Gaussian source works properly, so for now, we don't need to make any changes. I will close this PR :).
> 
> You showed several figures in this PR, which seems to be related to fwidth?

Yeah, but it seems that several factors, including the integral path, PMLs, mesh, and pulse, collectively contribute to the ripples in the S-parameters. Previously, I wasn't able to ensure the integral path, PMLs, and mesh were idealy set up. Now, with those optimized, the current pulse produces nice S-parameters as shown.
image
Screenshot from 2025-01-03 16-59-47

weiliangjin2021:
> Yeah, but it seems that several factors, including the integral path, PMLs, mesh, and pulse, collectively contribute to the ripples in the S-parameters. Previously, I wasn't able to ensure the integral path, PMLs, and mesh were idealy set up. Now, with those optimized, the current pulse produces nice S-parameters as shown.

That's good to know! Has the optimized integral path/PMLs/mesh been wrapped in a PR?

QimingFlex:
> > Yeah, but it seems that several factors, including the integral path, PMLs, mesh, and pulse, collectively contribute to the ripples in the S-parameters. Previously, I wasn't able to ensure the integral path, PMLs, and mesh were idealy set up. Now, with those optimized, the current pulse produces nice S-parameters as shown.
> 
> That's good to know! Has the optimized integral path/PMLs/mesh been wrapped in a PR?

PMLs simply need more layers (>50), and the mesh should better be uniform along the propagation direction. Could these be demonstrated in an example notebook for users? The tricky part is the integral path, and the voltage integral path plays a more important role in the S-parameters. The best path is to align the integral path along the grid line at the center of the copper thickness. 
image

momchil-flex:
> PMLs simply need more layers (>50)

I'm still a bit curious about this, wondering what causes it. The only cases I can think of so far where increasing the number of pml layers helps is simulations with periodic or Bloch boundaries in 2 dimensions and PML on the other, where oblique waves would not be well absorbed. This shouldn't be the case here though because there's PML in all 3 directions. But maybe somehow if one does a Fourier transform of the propagating mode in the plane parallel to the PML, there would be strong components close to oblique incidence? I really want to understand what exactly is causing this problem to require more PML layers, so we can make a more general recommendation for the users.

dmarek-flex:
> The best path is to align the integral path along the grid line at the center of the copper thickness.

That is an interesting observation, so in the inaccurate version of the results there was only a single cell to cover the metal thickness? As a result, the path would be snapped to the top or bottom boundary of that cell?

QimingFlex:
Yes, it seems there are edge effects around the PEC edges, which result in an inaccurate integral.

QimingFlex:
> > The best path is to align the integral path along the grid line at the center of the copper thickness.
> 
> That is an interesting observation, so in the inaccurate version of the results there was only a single cell to cover the metal thickness? As a result, the path would be snapped to the top or bottom boundary of that cell?

Yes, it seems there are edge effects around the PEC edges, which result in an inaccurate integral.

QimingFlex:
> > PMLs simply need more layers (>50)
> 
> I'm still a bit curious about this, wondering what causes it. The only cases I can think of so far where increasing the number of pml layers helps is simulations with periodic or Bloch boundaries in 2 dimensions and PML on the other, where oblique waves would not be well absorbed. This shouldn't be the case here though because there's PML in all 3 directions. But maybe somehow if one does a Fourier transform of the propagating mode in the plane parallel to the PML, there would be strong components close to oblique incidence? I really want to understand what exactly is causing this problem to require more PML layers, so we can make a more general recommendation for the users.

After conducting more tests, I conclude that the voltage integral path largely suppresses all other factors affecting the accuracy of S-parameters. Now, if the voltage integral path is chosen correctly, using a PML with 12 layers produces excellent results. But there is one issue with the PML: since it does not absorb the DC component, we use `remove_DC_component=True`, which causes an abrupt jump at low frequencies in S-parameters or other post-processing results. Although our main focus is on the high-frequency range, it would be better to smooth out the low frequencies. Any ideas on how to achieve this?

image

momchil-flex:
> But there is one issue with the PML: since it does not absorb the DC component, we use remove_DC_component=True, which causes an abrupt jump at low frequencies in S-parameters or other post-processing results.

Sorry, are the results bad if you do not remove the DC component? When I tested your notebook a little, what I noticed was that the fields never fully decay because of the DC part, but apart from that results looked the same?

In your plot, is the error / sharp jump due to the fact that there's very very little power at such low frequencies when DC component is removed? Or how should I understand it?

PR #2087, opened 2024-11-22
Minimum install for tidy3d

yaugenst-flex:
Shifting some things around to identify minimal set of packages needed to be able to validate a sample simulation file.

Currently requires:
```text
annotated-types   0.7.0
autograd          1.7.0
cycler            0.12.1
h5py              3.12.1
kiwisolver        1.4.7
matplotlib        3.9.2
numpy             2.1.3
packaging         24.2
pandas            2.2.3
pillow            11.0.0
pydantic          2.10.1
pydantic-core     2.27.1
pyparsing         3.2.0
pyroots           0.5.0
python-dateutil   2.9.0.post0
pytz              2024.2
pyyaml            6.0.2
rich              13.9.4
shapely           2.0.6
six               1.16.0
typing-extensions 4.12.2
xarray            2024.10.0
```

### DIFF
--- a/tidy3d/components/data/dataset.py
+++ b/tidy3d/components/data/dataset.py
@@ -9,8 +9,12 @@ from typing import Any, Callable, Dict, Tuple, Union
 import numpy as np
 import pydantic.v1 as pd
 import xarray as xr
-from matplotlib import pyplot as plt
-from matplotlib.tri import Triangulation
+
+try:
+    from matplotlib import pyplot as plt
+    from matplotlib.tri import Triangulation
+except ImportError:
+    pass
 
 from ...constants import PICOSECOND_PER_NANOMETER_PER_KILOMETER, inf
 from ...exceptions import DataError, Tidy3dNotImplementedError, ValidationError
--- a/tidy3d/components/viz.py
+++ b/tidy3d/components/viz.py
@@ -6,11 +6,18 @@ from functools import wraps
 from html import escape
 from typing import Any
 
-import matplotlib.pyplot as plt
-import matplotlib.ticker as ticker
 import pydantic.v1 as pd
-from matplotlib.patches import ArrowStyle, PathPatch
-from matplotlib.path import Path
+
+try:
+    import matplotlib.pyplot as plt
+    import matplotlib.ticker as ticker
+    from matplotlib.patches import ArrowStyle, PathPatch
+    from matplotlib.path import Path
+
+    # default arrow style
+    arrow_style = ArrowStyle.Simple(head_length=12, head_width=9, tail_width=4)
+except ImportError:
+    arrow_style = None
 from numpy import array, concatenate, inf, ones
 
 from ..constants import UnitScaling
@@ -160,10 +167,6 @@ MEDIUM_CMAP = [
 # colormap for structure's permittivity in plot_eps
 STRUCTURE_EPS_CMAP = "gist_yarg"
 STRUCTURE_HEAT_COND_CMAP = "gist_yarg"
-
-# default arrow style
-arrow_style = ArrowStyle.Simple(head_length=12, head_width=9, tail_width=4)
-
 
 """=================================================================================================
 Descartes modified from https://pypi.org/project/descartes/ for Shapely >= 1.8.0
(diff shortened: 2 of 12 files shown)


yaugenst-flex:
This version works with the following [requirements.txt](https://github.com/user-attachments/files/17948899/requirements.txt).

In an empty python environment, `pip install --no-deps --requirements requirements.txt`

I can succesfully do on the `SIM_FULL` (defined in the tests [here](https://github.com/flexcompute/tidy3d/blob/78dfeff490a9884e89786933708d6a64de980e62/tests/utils.py#L400)):
```python
import tidy3d as td

sim = td.Simulation.from_file("sim_to_validate.hdf5")
```

yaugenst-flex:
@majinge7846 can you check if this works for you? scipy is not necessary anymore for the interp validation.

So installing and running according to my [above comment](https://github.com/flexcompute/tidy3d/pull/2087#issuecomment-2493841860) should work.

tidy3d/components/data/dataset.py
```diff
@@ -9,8 +9,12 @@
...
 import pydantic.v1 as pd
 import xarray as xr
-from matplotlib import pyplot as plt
-from matplotlib.tri import Triangulation
+
+try:
+    from matplotlib import pyplot as plt
+    from matplotlib.tri import Triangulation
+except ImportError:
+    pass
```
daquinteroflex, line 17: Should we print a warning for users?

tidy3d/components/viz.py
```diff
@@ -6,11 +6,18 @@
...
+try:
+    import matplotlib.pyplot as plt
+    import matplotlib.ticker as ticker
+    from matplotlib.patches import ArrowStyle, PathPatch
+    from matplotlib.path import Path
+
+    # default arrow style
+    arrow_style = ArrowStyle.Simple(head_length=12, head_width=9, tail_width=4)
+except ImportError:
+    arrow_style = None
```
daquinteroflex, line 20: This won't break anything elsewhere?

daquinteroflex:
Just curious how you tested this fully worked? Trust that it does it's mainly a validation question.

tidy3d/components/data/dataset.py
yaugenst-flex: question goes out to @momchil-flex 😄
I think we discussed this previously and decided not to?

tidy3d/components/viz.py
yaugenst-flex: Yeah good point. Changed to a default that won't error in here (which is the only place where that default style is used):
https://github.com/flexcompute/tidy3d/blob/78dfeff490a9884e89786933708d6a64de980e62/tidy3d/components/geometry/base.py#L2232-L2239

In general the purpose of this change was to make this file not error on import if matplotlib isn't available, in which case the script otherwise would have failed when calling (an undefined) `ArrowStyle`. It's kind of expected that the code would still fail if the function that uses this `arrow_style` is called at any point. However, this should not happen because the only purpose of this minimal install PR is to give MC a way to import a version of tidy3d with minimal requirements and run validation.

yaugenst-flex:
> Just curious how you tested this fully worked? Trust that it does it's mainly a validation question.

Like this: https://github.com/flexcompute/tidy3d/pull/2087#issuecomment-2493841860

Note that none of this is supposed to be seen or used by users, and we definitely still need a better core packaging / refactor solution.

daquinteroflex:
> > Just curious how you tested this fully worked? Trust that it does it's mainly a validation question.
> 
> Like this: [#2087 (comment)](https://github.com/flexcompute/tidy3d/pull/2087#issuecomment-2493841860)
> 
> Note that none of this is supposed to be seen or used by users, and we definitely still need a better core packaging / refactor solution.

I see, so this just tests that a full simulation can be reloaded in terms of the new package imports. However, I guess I was curious if we should test for the plotting functionality at some point. But really this is just whether the environment properly works or not, and for a user without the packages installed it'd error whenever one of the plotting methods is called.

So this satisfies the packaging  import test, (and in principle the new import structure shouldn't affect the internal plotting methods if matplotlib is installed), and if matplotlib is not installed the plotting method will simply not work but we are not warning users explicitly until they call a method. I wonder if the error in this case when a user doesn't have matplotlib installed will be clear enough, or whether we just let them sort it out on their own. 

However, this PR is already good in terms of getting us closer to the core refactor plan! If we think we don't have to worry about these user plotting methods errors (it kind of depends how unclear and frustrating these can be?) then it's probably good to go! And if any of these plotting methods affect any other commonly used methods.

This said, think it's probably fine as is imo.

daquinteroflex:
Yannick and I spoke quickly and yeah all good on this! Now see how it fits in, just still catching up with things.

daquinteroflex (approved):

daquinteroflex:
Feel free to merge @yaugenst-flex whenever you want

PR #2088, opened 2024-11-25
Moving nonlinear warnings to simulation and consolidating

momchil-flex:
A user had 2000 structures each with the same nonlinear medium, which was producing a giant output of warnings.

Moving the warnings to `Simulation` ensures that they would only get printed once per unique medium. Any objections to this? Main difference seems to be that the warning will be shown a bit later (not immediately when constructing the medium), but still well in time to act on?


caseyflex (approved):
Thanks, yeah this sounds better!

PR #2089, opened 2024-11-25
PolySlab rotated and translated methods that return PolySlab

momchil-flex:


tidy3d/components/geometry/polyslab.py
```diff
@@ -1520,6 +1521,84 @@ def normalize_vect(arr: np.ndarray) -> np.ndarray:
...
+            Scaling factor along z.
+
+        Returns
+        -------
+        :class:`Geometry`
+            Scaled copy of this geometry.
+        """
+        scale_normal, scale_in_plane = self.pop_axis((x, y, z), axis=self.axis)
+        scaled_vertices = self.vertices * np.array(scale_in_plane)
+        print(self.vertices[:5, :])
```
lucas-flexcompute: Don't forget to remove the debug prints before merging.

lucas-flexcompute (approved):

tidy3d/components/geometry/polyslab.py
momchil-flex: Ah thanks.

PR #2090, opened 2024-11-26
Allow general web api handle mode solver tasks

dbochkov-flexcompute:
one annoying thing is that for mode solver tasks we need to upload/download `mode_solver.hdf5.gz`/`mode_solver_data.hdf5.gz` compared to simply `simulation.hdf5.gz`/`simulation_data.hdf5.gz` for all other solvers


momchil-flex:
This is great. Porting your comment from Slack, we need to decide between:

1. do not release for the current ModeSolver to public, just use it internally for debugging (as using regular web functions allows to specify, for example, solver version and worker group). Release it to public when we introduce 
[@Casey Wojcik](https://enreal.slack.com/team/U0444J7Q562)
’s refactor ModeSimulation https://github.com/flexcompute/tidy3d/pull/1838
2. release to public in the next minor/major version, and keep it separate from the mode solver own run and batch functions
3. release to public in the next minor/major version, and convert the mode solver own run and batch functions to wrappers of web.run() and web.Batch()

Certainly 2. sounds the most tempting as it's getting it out soon without much effort on our side. I guess the worry may be some complication or confusion that could a rise if a user mixes the two different usages. But I feel like... it should be fine?

momchil-flex:
Shall we PR this into pre/2.8 using the (simplest for us) approach 2? @tylerflex @daquintero ?

tylerflex:
I'm fine with 2, although I think if it's not too much extra effort, could be worth going for 3?

daquinteroflex:
Sounds good! Seems like 2 ultimately would lead to 3 too eventually to maintain compatibility anyways.

momchil-flex:
Closing in favor of #2115

PR #2091, opened 2024-11-26
changed the default choice of frequency when plotting permittivity

dmarek-flex:
The default choice is now the central frequency of the first source added to the simulation.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -439,7 +439,9 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
             position of plane in z direction, only one of x, y, z must be specified to define plane.
         freq : float = None
             Frequency to evaluate the relative permittivity of all mediums.
-            If not specified, evaluates at infinite frequency.
+            If not specified, the central frequency of sources in the simulation will be used.
+            If sources have different central frequencies, the relative permittivity will be evaluated
+            at infinite frequency.
         alpha : float = None
             Opacity of the structures being plotted.
             Defaults to the structure default alpha.
@@ -524,7 +526,9 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
             position of plane in z direction, only one of x, y, z must be specified to define plane.
         freq : float = None
             Frequency to evaluate the relative permittivity of all mediums.
-            If not specified, evaluates at infinite frequency.
+            If not specified, the central frequency of sources in the simulation will be used.
+            If sources have different central frequencies, the relative permittivity will be evaluated
+            at infinite frequency.
         reverse : bool = False
             If ``False``, the highest permittivity is plotted in black.
             If ``True``, it is plotteed in white (suitable for black backgrounds).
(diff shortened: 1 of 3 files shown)


tidy3d/components/simulation.py
```diff
@@ -549,7 +551,8 @@ def plot_structures_eps(
         hlim, vlim = Scene._get_plot_lims(
             bounds=self.simulation_bounds, x=x, y=y, z=z, hlim=hlim, vlim=vlim
         )
-
+        if freq is None and len(self.sources) > 0:
+            freq = self.sources[0].source_time.freq0
```
caseyflex: Maybe keep it as None if there are multiple freq0. Like this:
https://github.com/flexcompute/tidy3d/blob/2f2f0814b415aca77b8ba76307d354a40265936e/tidy3d/components/medium.py#L179

tomflexcompute (approved):
Thanks @dmarek-flex for the fast turnaround! Looks great. Since users are using `plot_eps` more often than `plot_structures_eps`, in the change log should we say something like "Default choice of frequency in `Simulation.plot_structures_eps` and `Simulation.plot_eps` is now the central frequency of the first source added to the simulation. "

If no source is present, freq will still be set up 1e50 right?

caseyflex (changes requested):
It does seem better to use source frequency than infinite frequency.

In the case where you do have to use infinite frequency (like when no sources are supplied or the freq0 are inconsistent), it would be good to give a more useful warning. You can see here how I silence the frequency range warning:
https://github.com/flexcompute/tidy3d/blob/9bb67c26307f8aba70ac6e7c0956a99beedd7beb/tidy3d/material_library/parametric_materials.py#L195

Then, instead, you could warn about how in the absence of supplied freq0 / source freq0, infinite frequency is used for plotting.

tidy3d/components/simulation.py
tomflexcompute: Agree with @caseyflex 's point. Rarelly plotting eps at infinite frequency would make physical sense or is what the user is looking for. Good to warn about it.

dmarek-flex:
> Then, instead, you could warn about how in the absence of supplied freq0 / source freq0, infinite frequency is used for plotting.

I like this idea, but don't know how best to implement. To silence the warnings, I would need to modify all media in the `Scene`. The other option is to modify `Scene.eps_bounds` so that it modifies the list of media it generates. It seems like  `Scene.eps_bounds`  is only used in the `Scene.plot_structures_eps` so it might make sense to emit a warning there.

Am I missing an easier way @caseyflex?

caseyflex:
> > Then, instead, you could warn about how in the absence of supplied freq0 / source freq0, infinite frequency is used for plotting.
> 
> I like this idea, but don't know how best to implement. To silence the warnings, I would need to modify all media in the `Scene`. The other option is to modify `Scene.eps_bounds` so that it modifies the list of media it generates. It seems like `Scene.eps_bounds` is only used in the `Scene.plot_structures_eps` so it might make sense to emit a warning there.
> 
> Am I missing an easier way @caseyflex?

I like having everything in simulation.py where you currently have it. I did a bit of digging, it seems like if you pass `freq=np.inf` instead of `freq=None` explicitly, this will silence those other out-of-range warnings?
https://github.com/flexcompute/tidy3d/blob/2f2f0814b415aca77b8ba76307d354a40265936e/tidy3d/components/medium.py#L115

caseyflex (approved):

PR #2096, opened 2024-11-28
Simplify interp validator

yaugenst-flex:
Changes the interp validator to only check for duplicate coordinates instead of explicitly interpolating and selecting and added a test.

Seems to be catching the same validation errors as before, but not 100% sure.

### DIFF
--- a/tidy3d/components/data/data_array.py
+++ b/tidy3d/components/data/data_array.py
@@ -9,7 +9,6 @@ import autograd.numpy as anp
 import dask
 import h5py
 import numpy as np
-import pandas
 import xarray as xr
 from autograd.tracer import isbox
 from xarray.core import alignment, missing
@@ -119,7 +118,7 @@ class DataArray(xr.DataArray):
         return val
 
     def _interp_validator(self, field_name: str = None) -> None:
-        """Make sure we can interp()/sel() the data.
+        """Ensure the data can be interpolated or selected by checking for duplicate coordinates.
 
         NOTE
         ----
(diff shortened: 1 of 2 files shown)


yaugenst-flex:
This change is copied over from #2087

momchil-flex (approved):
I think this is fine, all the cases that I know of that were problematic it was because of duplicated coordinates, if we encounter other cases in the future we'll have to expand this validator.

tidy3d/components/data/data_array.py
```diff
@@ -129,39 +128,13 @@ def _interp_validator(self, field_name: str = None) -> None:
...
-                        f"'{field_name}={field_name}.drop_duplicates(dim=\"{dim}\")' "
-                        f"and run 'sel()' with the new '{field_name}'). "
-                        "Plase make sure 'sel()' can be used on the 'DataArray'."
-                    ) from e
+        for dim, coord in self.coords.items():
+            if coord.to_index().duplicated().any():
+                raise DataError(
+                    f"Field '{field_name}' contains duplicate coordinates in dimension '{dim}'. "
+                    "You can remove duplicates by running "
+                    f"'{field_name}={field_name}.drop_duplicates(dim=\"{dim}\")'."
```
marc-flex: I'm not particularly bothered by this (I wrote the code you're replacing) but I thought we were avoiding second person statements? (like YOU should check etc.)
@momchil-flex , what's the policy for this kind of language?

marc-flex (approved):
Very cool! Thanks for taking care of this!

tidy3d/components/data/data_array.py
yaugenst-flex: Ah yeah good point, changed it.

momchil-flex: Yeah, sorry, this feels a bit odd to me so I have commented on it in the past, but I'm not 100% set against it and it may be hard to make sure we always enforce it...

PR #2097, opened 2024-12-02
better plot_grid

tylerflex:
uses `axhline` and `axvline` to Draw grid boundaries. and ignores grid boundaries for size 0 dims.

https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.axhline.html

before
image

after
image


momchil-flex (approved):
Nice, code looks much better!

PR #2098, opened 2024-12-02
Fix another geos error in _filter_structures_plane

weiliangjin2021:


weiliangjin2021:
Somehow `normalize()` fixes `shapely.errors.GEOSException: TopologyException: side location conflict at ..... This can occur if the input geometry is invalid.`. It works without `buffer(0`), but I just keep it to be safe.

I also shift things around to make it easier to follow.

tylerflex (approved):
looks good to me, but this kind of shapely manipulation is outside of my knowledge at this point so feel free to get another review if needed.

momchil-flex (approved):
Approving if tests pass and this fixes the error users encountered, but I don't understand this either. :D @lucas-flexcompute ?

lucas-flexcompute:
I've always disliked the `buffer(0)` hack, although that seems to be the only working option in some cases. And I guess `normalize` does more than the docs imply. There might be other places in the frontend where we use `buffer(0)` or `make_valid` that might also benefit from the `normalize` call. I don't think we need to test that in this PR, but we should remember that the next time we see GEOS/shapely errors.

lucas-flexcompute (approved):

PR #2099, opened 2024-12-03
fix `attrs` bug with `inf`

tylerflex:
Solution:
* make sure `Tidy3dBaseModel._json()` looks at / doesn't ignore `attrs` (?)
* [better] write custom encoder to properly handle infinity in the `json()` logic

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Silence warning in graphene from checking fit quality at large frequencies.
 - `CustomCurrentSource` now correctly validates its current dataset.
 - Better plotting of grids, respecting 2D and 1D simulation edge cases.
+- Bug when `td.inf` are in `attrs` and files saved and loaded to .json twice.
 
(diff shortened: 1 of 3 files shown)


momchil-flex (approved):
Looks good for now, could you squash - and is there an issue if we want to follow up with a better fix? Or is there nothing specific for now apart from a general refactor of our json IO?

tylerflex:
> Looks good for now, could you squash - and is there an issue if we want to follow up with a better fix? Or is there nothing specific for now apart from a general refactor of our json IO?

I think there are going to be better fixes out there, either general refactor of json IO using the proper json encoder technique, or even something to handle attrs more explicitly. I couldn't figure out why it was ignored actually. It might be because it's a mutable type (dict).

So I will make a "good first issue" but it might make sense to leave as 2 separate commits since the first just makes the test that fails, and the 2nd is a pretty hacky fix that we'll probably want to eventually just drop?

yaugenst:
Maybe worth to point out that pydantic v2 does this via https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.ser_json_inf_nan

PR #2100, opened 2024-12-04
Doping boxes

marc-flex:
This PR addresses a concerned raised by @lucas-flexcompute. It was suggested that we should have a way to produce Gaussian doping in a similar way as Lumerical does.

The current PR introduces the concept of doping box within which one can define a constant or gaussian doping in an analogous way to Lumerical. It is still not great but facilitates doping creation.

fyi @daquinteroflex


dbochkov-flexcompute:
I have concerns about plotting, it seems it's quite limited overall. For example:
- if there is a structure overlapping a doped structure, the doping will still be plotted over the structure on top
- if a structure has a non-box shape (say, a circle), then doping field will still be rendered as as a box
- it doesn't plot doping if it's in background medium

I do think that we should just extend current method for plotting eps https://github.com/flexcompute/tidy3d/blob/marc/dd-devsim-doping/tidy3d/components/scene.py#L736, which already properly takes care of plotting sampled fields like custom medium https://github.com/flexcompute/tidy3d/blob/marc/dd-devsim-doping/tidy3d/components/scene.py#L837-L840

tidy3d/components/heat_charge/doping.py
```diff
@@ -0,0 +1,98 @@
...
+from abc import ABC
+
+import numpy as np
+import pydantic.v1 as pd
+
+from ..base import Tidy3dBaseModel, cached_property
+from ..types import Bound, Union
+
+
+class AbstractDopingBox(ABC, Tidy3dBaseModel):
```
dbochkov-flexcompute: One possibility you can consider is to inherit actual class `Box` here. It follows how we typically define rectangular regions in Tidy3d (center + size), but still has convenience method `from_bounds` https://github.com/flexcompute/tidy3d/blob/marc/dd-devsim-doping/tidy3d/components/geometry/base.py#L1889

tidy3d/components/scene.py
```diff
@@ -1589,3 +1591,269 @@ def perturbed_mediums_copy(
...
+        for d, d_name in enumerate("xyz"):
+            if scene_bounds[0][d] == scene_bounds[1][d]:
+                coords_dict[d_name] = [scene_bounds[0][d]]
+            else:
+                coords_dict[d_name] = np.linspace(scene_bounds[0][d], scene_bounds[1][d], N)
+
+        size = [len(coords_x) for _, coords_x in coords_dict.items()]
+
+        # gen coordinate matrices
+        X, Y, Z = np.meshgrid(coords_dict["x"], coords_dict["y"], coords_dict["z"])
```
dbochkov-flexcompute: i think this needs `indexing="ij"`

marc-flex:
@dbochkov-flexcompute as with the back-end counterpart, this is now ready for re-review

The function `Scene:plot_structure_property()` can now be used to plot doping, which is a refactor of the `Scene:plot_structure_eps()`. This is all, of course, backwards compatible

The DopingBox class now has a function that evaluates the Gaussian contribution which can then be used in both front and back-end

dbochkov-flexcompute:
I think this works much better now. (right column: background medium is also doped. bottom row: alpha < 1)
plot_doping_old

There were still some issues and I pushed a small change trying to fix those (feel free to revert, modify, adjust, etc). It still has issues when we do alpha < 1 and background medium is a doping medium too (because shape has inf size, see bottom right figure), but perhaps it's not critical at the moment
plot_doping_new

marc-flex:
Looks very good. Thanks @dbochkov-flexcompute!

I'll merge this today @daquinteroflex

PR #2103, opened 2024-12-05
Avoiding matplotlib when importing/validating models

momchil-flex:


tidy3d/components/data/dataset.py
```diff
@@ -9,8 +9,12 @@
...
 import pydantic.v1 as pd
 import xarray as xr
-from matplotlib import pyplot as plt
-from matplotlib.tri import Triangulation
+
+try:
+    from matplotlib import pyplot as plt
+    from matplotlib.tri import Triangulation
+except ImportError:
+    pass
```
yaugenst-flex, line 17: I think the approach is generally fine for now as discussed but it might be a good idea to print a warning here instead of ignoring this silently and then failing unexpectedly when a user tries to call a function that relies on the missing import?

Something like
```python
td.log.warning("Matplotlib is not installed. Features requiring matplotlib will not work. Please install matplotlib if you need these features.")
```

yaugenst-flex (approved):

tidy3d/components/data/dataset.py
momchil-flex: So I thought about this too but since the point is it's for internal usage I thought it's maybe fine and in fact it will probably add extra complications. For example, one of the use cases is validating simulations, which captures all warnings and reports them to the user. I think probably things would have been imported by then but depending on how things work it may happen that this warning gets captured.

yaugenst-flex: Ok yeah fair enough. We can skip it. Tbh I think the default traceback when importing a module failed is pretty clear...

PR #2107, opened 2024-12-05
MeshOverrideStructure only Box

### LINKED ISSUE #2105: Override structures support cylinder, sphere, polyslab causing confusion
Our mesh override structure supports geometries besides Box such as Cylinder, Sphere, and PolySlab. This causes some confusion for users. For example, when using a Cylinder in the mesh override, the user thinks only the region within the cylinder will have a fine grid. However effectively we are only using the bounding box for the mesh override. Should we only allow Box as override?

tylerflex:
fixes #2105

### DIFF
--- a/tidy3d/components/structure.py
+++ b/tidy3d/components/structure.py
@@ -14,10 +14,12 @@ from ..constants import MICROMETER
 from ..exceptions import SetupError, Tidy3dError, Tidy3dImportError
 from ..log import log
 from .autograd.derivative_utils import DerivativeInfo
-from .autograd.types import AutogradFieldMap, Box
+from .autograd.types import AutogradFieldMap
+from .autograd.types import Box as AutogradBox
 from .autograd.utils import get_static
 from .base import Tidy3dBaseModel, skip_if_fields_missing
 from .data.data_array import ScalarFieldDataArray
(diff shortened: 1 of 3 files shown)


tidy3d/components/structure.py
```diff
@@ -678,5 +680,17 @@ class MeshOverrideStructure(AbstractStructure):
...
     )
 
+    @pydantic.validator("geometry")
+    def _box_only(cls, val):
+        """Ensure this is a box."""
+        if isinstance(val, Geometry):
+            if not isinstance(val, Box):
+                log.warning(
+                    "Override structures should be 'Box' as of 'tidy3d' version 2.8. "
+                    f"Given type of '{type(val)}'. Converting to bounding box."
```
momchil-flex: How about `Given type of '{type(val)}, using '{type(val)}.bounding_box' instead.`? Just as an extra hint to the user that if they pass `geometry.bounding_box` in their override structure definition they can avoid the warning.

tomflexcompute (approved):

tidy3d/components/structure.py
tylerflex: ok

tylerflex: changed

PR #2110, opened 2024-12-06
Adding 'fields' argument to ModeSolver and ModeSolverMonitor

momchil-flex:


caseyflex (approved):

tidy3d/components/data/monitor_data.py
```diff
@@ -1752,6 +1745,17 @@ def to_dataframe(self) -> DataFrame:
 
         return dataset.drop_vars(drop).to_dataframe()
 
+    def _check_fields_stored(self, components: list[EMField]):
+        """Check that all requested field components are stored in the data."""
+
+        # ModeData can either have all field components or none
```
lucas-flexcompute, line 1751: Isn't the goal of this PR to reduce the data size when mode solving for many frequencies? This seems to defeat the purpose if, for example, the user asks only for Ex and Ey, doesn't it?

momchil-flex: Ok so this is definitely confusing but basically `ModeData` is the result of a `ModeMonitor` in an FDTD simulation, which by default does not store any fields, but it could if `store_fields_direction` is set, in which case all fields are stored. My PR adds the ability to selectively store fields in `ModeSolverData`, which is the result of a `ModeSolver` call, or of a `ModeSolverMonitor` in an FDTD simulation. Because of various requests, e.g. to add option to store fields in `ModeData`, and now an option to not store fields in `ModeSolverData`, these two now look like they might make more sense to not be separate at all, but I don't think this can be easily done without breaking compatibility in some ways.

I guess I could add the `fields` argument to the `ModeMonitor` too, in case `store_fields_direction` is set?

lucas-flexcompute: I see! I think it might be fine this way, then, unless we have specific user requests for the `ModeMonitor` case too.

lucas-flexcompute (approved):

PR #2111, opened 2024-12-06
:wrench: Fix numpy2 2.1 depreciation error

daquinteroflex:
Say trying to generate a lockfile with numpy 2.1

```bash
± |dario/fix_numpy2_deps {1} ✓| → poetry lock
Updating dependencies
Resolving dependencies... (31.4s)

The current project's supported Python range (>=3.9,<4.0.0) is not compatible with some of the required packages Python requirement:
  - numpy requires Python >=3.10, so it will not be satisfied for Python >=3.9,<3.10
  - numpy requires Python >=3.10, so it will not be satisfied for Python >=3.9,<3.10
  - numpy requires Python >=3.10, so it will not be satisfied for Python >=3.9,<3.10
  - numpy requires Python >=3.10, so it will not be satisfied for Python >=3.9,<3.10
  - numpy requires Python >=3.10, so it will not be satisfied for Python >=3.9,<3.10

Because no versions of numpy match >2.1.0,<2.1.1 || >2.1.1,<2.1.2 || >2.1.2,<2.1.3 || >2.1.3,<2.2.0rc1 || >2.2.0rc1,<2.3
 and numpy (2.2.0rc1) requires Python >=3.10, numpy is forbidden.
And because numpy (2.1.0) requires Python >=3.10
 and numpy (2.1.1) requires Python >=3.10, numpy is forbidden.
And because numpy (2.1.2) requires Python >=3.10
 and numpy (2.1.3) requires Python >=3.10, numpy is forbidden.
So, because no versions of numpy match >=2.3
 and tidy3d depends on numpy (>=2.1.0), version solving failed.

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties
    
    For numpy, a possible solution would be to set the `python` property to ">=3.10,<4.0.0"
    For numpy, a possible solution would be to set the `python` property to ">=3.10,<4.0.0"
    For numpy, a possible solution would be to set the `python` property to ">=3.10,<4.0.0"
    For numpy, a possible solution would be to set the `python` property to ">=3.10,<4.0.0"
    For numpy, a possible solution would be to set the `python` property to ">=3.10,<4.0.0"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers


```

Generating the environment with `numpy<2.1` works for all our supported pythons.


yaugenst-flex:
Saw this because I was working on something related just now (https://github.com/flexcompute/tidy3d/pull/2143). I think we might also want to consider bumping minimum Python to 3.10

yaugenst-flex:
Is this still an issue @daquinteroflex?

PR #2113, opened 2024-12-06
Automatical mesh refinement in layered structures

weiliangjin2021:
(associated with https://github.com/flexcompute/tidy3d-core/issues/509)

## Background

In many applications, structures are built layer by layer. Materials can be considered uniform along the normal axis. 
- When there are metals (including PEC and lossy metal) present in the layer, it's important to snap and refine grid to the metal corners. We also want to refine around the corners/edges.
- Sometimes we want to resolve layer thickness. In most cases, we want lower or upper bound of the layer (or both) to be at grid.

In 2D, we can simply use shapely to find out corners. With corners, we can have boundaries (or edges). With the two, we can automatically generate snapping points and refinement mesh override structures.

## Features
- A dedicated [file](https://github.com/flexcompute/tidy3d/blob/9b3cd40ab34716f372848490dbd2a33589c9af2f/tidy3d/components/grid/corner_finder.py) for 2d corner finder;
- A `GridRefinement` class [here](https://github.com/flexcompute/tidy3d/blob/9b3cd40ab34716f372848490dbd2a33589c9af2f/tidy3d/components/grid/grid_spec.py#L834-L836) for local mesh refinement. It is to make it easier to generate mesh override structures based on the number of grid cells and refined grid size. The refined grid size can be defined by either the refinement factor w.r.t. vacuum grid size, or grid size directly.
- `LayerRefinementSpec` that defines the layer (it can be bounded on the inplane dimensions so that corners outside the bound are discarded) and various mesh generation specs.
- Added `layer_refinement_specs` field to `GridSpec`. The final snapping points and override structures will be internal ones + user supplied ones. The internal ones are generated by `LayerRefinementSpec`.
- When `dl_min` is not set by the user, we'll automatically generate a value based on the structures, simulation domain size, and layer_specs. 

## Examples

If we are fine with default meshing specs, `layer_refinement_spec` can be easily defined. First, let's defined a layer that is unbounded perpendicular to the normal axis:
```
layer = td.LayerRefinementSpec.from_unbounded(
    axis=2,
    bounds=(-patch_thickness/2, patch_thickness/2),
)
```
and grid_spec,
`grid_spec = td.GridSpec.auto(wavelength=wavelength, layer_refinement_specs=[layer])`

Consider a patch antenna but we drill a big hole on the left, with the above grid_spec, we get
image

As you can see, metal corners are all identified, and snapping points are applied at those corners. Small mesh refinement structures are placed at corners, too, so that the mesh is denser near those singularities.

Sometimes, you only want refine a portion of the layer, as the other portion is not affecting the simulation as much. E.g., we only want to refine the feeding region,
```
layer = td.LayerRefinementSpec(
    axis=2,
    center=(0, -10e3, 0),
    size=(10e3,10e3, patch_thickness),
)
```
image

### Major
- [x] Layer refinement specification class to define layer position and refinement parameters: defined [here](https://github.com/flexcompute/tidy3d/blob/ba8cd2e1ae03b27253f12e0e6b456af352abb027/tidy3d/components/grid/grid_spec.py#L535-L538)
- [x] Corner finder in 2d with shapely
- [x] Automatic snapping points generation
- [x] Automatically sets `dl_min` if not set yet:
  - [x] Compute dl_min from layers
  - [x] Compute dl_min from all structures
- [x] Automatic mesh refinement structures generation

### Minor
- [x] Allow snapping points to take `None` as coordinate, so that it's not snapped along that axis. Now snapping point is of type [CoordinateOptional](https://github.com/flexcompute/tidy3d/blob/ba8cd2e1ae03b27253f12e0e6b456af352abb027/tidy3d/components/types.py#L188). In visualization, snapping points will be `snapping line` if one of the inplane coordinate is `None`.
- [x] Shift `_filter_structures_plane` to geometry/utils.py, as it'll be used in 2d corner finder.

### Miscellaneous
- [x] Rebase once https://github.com/flexcompute/tidy3d/pull/2130 is merged. Many conflicts expected.
- [x] Add tests
- [x] Add CHANGELOG
- [x] Better visualization of snapping points that contain `None`

### DIFF
--- a/tidy3d/components/grid/corner_finder.py
+++ b/tidy3d/components/grid/corner_finder.py
@@ -1,0 +1,137 @@
+"""Find corners of structures on a 2D plane."""
+
+from typing import List, Literal, Optional
+
+import numpy as np
+import pydantic.v1 as pd
+
+from ...constants import inf
+from ..base import Tidy3dBaseModel
+from ..geometry.base import Box, ClipOperation
+from ..geometry.utils import merging_geometries_on_plane
+from ..medium import PEC, LossyMetalMedium
+from ..structure import Structure
+from ..types import ArrayFloat2D, Axis
+
+CORNER_ANGLE_THRESOLD = 0.1 * np.pi
+
+
+class CornerFinderSpec(Tidy3dBaseModel):
+    """Specification for corner detection on a 2D plane."""
+
+    medium: Literal["metal", "dielectric", "all"] = pd.Field(
+        "metal",
+        title="Material Type For Corner Identification",
+        description="Find corners of structures made of ``medium``, "
+        "which can take value ``metal`` for PEC and lossy metal, ``dielectric`` "
+        "for non-metallic materials, and ``all`` for all materials.",
+    )
+
+    angle_threshold: float = pd.Field(
+        CORNER_ANGLE_THRESOLD,
+        title="Angle Threshold In Corner Identification",
+        description="A vertex is qualified as a corner if the angle spanned by its two edges "
+        "is larger than the supplementary angle of "
+        "this threshold value.",
+        ge=0,
+        lt=np.pi,
+    )
+
+    distance_threshold: Optional[pd.PositiveFloat] = pd.Field(
+        None,
+        title="Distance Threshold In Corner Identification",
+        description="If not ``None`` and the distance of the vertex to its neighboring vertices "
+        "is below the threshold value based on Douglas-Peucker algorithm, the vertex is disqualified as a corner.",
+    )
+
+    def corners(
+        self,
+        normal_axis: Axis,
+        coord: float,
+        structure_list: List[Structure],
+    ) -> ArrayFloat2D:
+        """On a 2D plane specified by axis = `normal_axis` and coordinate `coord`, find out corners of merged
+        geometries made of `medium`.
+
+
+        Parameters
+        ----------
+        normal_axis : Axis
+            Axis normal to the 2D plane.
+        coord : float
+            Position of plane along the normal axis.
+        structure_list : List[Structure]
+            List of structures present in simulation.
+
+        Returns
+        -------
+        ArrayFloat2D
+            Corner coordinates.
+        """
+
+        # Construct plane
+        center = [0, 0, 0]
+        size = [inf, inf, inf]
+        center[normal_axis] = coord
+        size[normal_axis] = 0
+        plane = Box(center=center, size=size)
+
+        # prepare geometry and medium list
+        geometry_list = [structure.geometry for structure in structure_list]
+        # For metal, we don't distinguish between LossyMetal and PEC,
+        # so they'll be merged to PEC. Other materials are considered as dielectric.
+        medium_list = (structure.medium for structure in structure_list)
+        medium_list = [
+            PEC if (mat.is_pec or isinstance(mat, LossyMetalMedium)) else mat for mat in medium_list
+        ]
+        # merge geometries
+        merged_geos = merging_geometries_on_plane(geometry_list, plane, medium_list)
+
+        # corner finder
+        corner_list = []
+        for mat, shapes in merged_geos:
+            if self.medium != "all" and mat.is_pec != (self.medium == "metal"):
+                continue
+            polygon_list = ClipOperation.to_polygon_list(shapes)
+            for poly in polygon_list:
+                poly = poly.normalize().buffer(0)
+                if self.distance_threshold is not None:
+                    poly = poly.simplify(self.distance_threshold, preserve_topology=True)
+                corner_list.append(self._filter_collinear_vertices(list(poly.exterior.coords)))
+                # in case the polygon has holes
+                for poly_inner in poly.interiors:
+                    corner_list.append(self._filter_collinear_vertices(list(poly_inner.coords)))
+
+        if len(corner_list) > 0:
+            corner_list = np.concatenate(corner_list)
+        return corner_list
+
+    def _filter_collinear_vertices(self, vertices: ArrayFloat2D) -> ArrayFloat2D:
+        """Filter collinear vertices of a polygon, and return corners.
+
+        Parameters
+        ----------
+        vertices : ArrayFloat2D
+            Polygon vertices from shapely.Polygon. The last vertex is identical to the 1st
+            vertex to make a valid polygon.
+
+        Returns
+        -------
+        ArrayFloat2D
+            Corner coordinates.
+        """
+
+        def normalize(v):
+            return v / np.linalg.norm(v, axis=-1)[:, np.newaxis]
+
+        # drop the last vertex, which is identical to the 1st one.
+        vs_orig = np.array(vertices[:-1])
+        # compute unit vector to next and previous vertex
+        vs_next = np.roll(vs_orig, axis=0, shift=-1)
+        vs_previous = np.roll(vs_orig, axis=0, shift=+1)
+        unit_next = normalize(vs_next - vs_orig)
+        unit_previous = normalize(vs_previous - vs_orig)
+        # angle
+        angle = np.arccos(np.sum(unit_next * unit_previous, axis=-1))
+        ind_filter = angle <= np.pi - self.angle_threshold
+        return vs_orig[ind_filter]
--- a/tidy3d/components/grid/grid_spec.py
+++ b/tidy3d/components/grid/grid_spec.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import List, Tuple, Union
+from typing import Any, List, Literal, Optional, Tuple, Union
 
 import numpy as np
 import pydantic.v1 as pd
@@ -11,13 +11,29 @@ import pydantic.v1 as pd
 from ...constants import C_0, MICROMETER, inf
 from ...exceptions import SetupError
 from ...log import log
-from ..base import Tidy3dBaseModel
+from ..base import Tidy3dBaseModel, cached_property, skip_if_fields_missing
 from ..geometry.base import Box
 from ..source.utils import SourceType
 from ..structure import MeshOverrideStructure, Structure, StructureType
-from ..types import TYPE_TAG_STR, Axis, Coordinate, Symmetry, annotate_type
+from ..types import (
+    TYPE_TAG_STR,
+    ArrayFloat2D,
+    Axis,
+    Coordinate,
+    CoordinateOptional,
+    Symmetry,
+    annotate_type,
+)
+from .corner_finder import CornerFinderSpec
 from .grid import Coords, Coords1D, Grid
 from .mesher import GradedMesher, MesherType
+
+# Scaling factor applied to internally generated lower bound of grid size that is computed from
+# estimated minimal grid size
+MIN_STEP_BOUND_SCALE = 0.5
+
+# Default refinement factor in GridRefinement when both dl and refinement_factor are not defined
+DEFAULT_REFINEMENT_FACTOR = 2
 
 
 class GridSpec1d(Tidy3dBaseModel, ABC):
@@ -31,7 +47,7 @@ class GridSpec1d(Tidy3dBaseModel, ABC):
         periodic: bool,
         wavelength: pd.PositiveFloat,
         num_pml_layers: Tuple[pd.NonNegativeInt, pd.NonNegativeInt],
-        snapping_points: Tuple[Coordinate, ...],
+        snapping_points: Tuple[CoordinateOptional, ...],
     ) -> Coords1D:
         """Generate 1D coords to be used as grid boundaries, based on simulation parameters.
         Symmetry, and PML layers will be treated here.
@@ -49,7 +65,7 @@ class GridSpec1d(Tidy3dBaseModel, ABC):
             Free-space wavelength.
         num_pml_layers : Tuple[int, int]
             number of layers in the absorber + and - direction along one dimension.
-        snapping_points : Tuple[Coordinate, ...]
+        snapping_points : Tuple[CoordinateOptional, ...]
             A set of points that enforce grid boundaries to pass through them.
 
         Returns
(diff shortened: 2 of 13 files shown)


tidy3d/components/simulation.py
```diff
@@ -824,6 +824,26 @@ def plot_grid(
...
+                continue
+            if y_point is None:
+                ax.axvline(
+                    x=self._evaluate_inf(x_point),
+                    ymin=0.1,
+                    ymax=0.9,
+                    linewidth=2 * kwargs["linewidth"],
+                    color=plot_params.edgecolor,
+                )
+                continue
```
weiliangjin2021: @dmarek-flex @tylerflex here I enable `snapping point` to take `None` as a coordinate, which allows skipping this dimension in mesh generation. But I'm not sure what's the best way to visualization this. When one of the 2d coordinate is `None`, snapping points become "snapping line".  I plot a segment of line that doesn't cover the entire region, so that it doesn't get confused with grid lines. Otherwise, it's hard to tell if there is a grid line at the position of "snapping line".

weiliangjin2021: Here is the figure under current implementation where snapping point takes the form `(x, y, None)`
image
image

weiliangjin2021:
There are still a few items to be done, but they are likely to be minor. You are welcome to start reviewing now, e.g. the general flow, naming, etc.

tidy3d/components/simulation.py
```diff
@@ -1064,6 +1089,17 @@ def num_pml_layers(self) -> List[Tuple[float, float]]:
 
         return num_layers
 
+    @cached_property
+    def self_structure(self) -> Structure:
+        """The simulation background as a ``Structure``."""
+        geometry = Box(size=(inf, inf, inf), center=self.center)
+        return Structure(geometry=geometry, medium=self.medium)
```
dbochkov-flexcompute: it looks very similar to `self.scene.background_structure` https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/scene.py#L217-L221. Is `center=self.center` what distinguishes it from that?

tidy3d/components/simulation.py
dbochkov-flexcompute: would this look a bit unclear if the structure that generated this snapping line is actually outside of (0.1, 0.9) region? say, (0.1, 0.9) is entirely vacuum. 

another minor suggestions: 
- would it make sense to distinguish by color "external" and "internal" snapping points/lines?
- we can also have a flag to turn off/on (or an alpha parameter) snapping points/lines

dbochkov-flexcompute (commented):
Looks great! Just some minor comments so far

tidy3d/components/simulation.py
weiliangjin2021: Didn't realize that it's defined in Scene already. Will remove it, and move `all_structures` to Scene

weiliangjin2021: hmm, actually this part is historical. they are there already https://github.com/flexcompute/tidy3d/blob/bb62120818b7fe385de139c6879d9821be8d4356/tidy3d/components/simulation.py#L4290-L4299, I just shifted them to abstract simulation class.

weiliangjin2021: I'll add `all_structures` to scene, and change the two properties in Simulation to call scene properties.

tidy3d/components/simulation.py
weiliangjin2021: Thanks @dbochkov-flexcompute . Now internal snapping points/override structures are plotted in magenta, while external black. alpha for them has also been added. The linestyle has been changed to dashes. I increase the linewidth so that the dashes don't look too dense.
image
image

The denser dashes are from override-structures, while sparser one from snapping lines.

tidy3d/components/grid/grid_spec.py
```diff
@@ -665,14 +787,580 @@ def _preprocessed_structures(self, structures: List[StructureType]) -> List[Stru
...
+        title="Axis",
+        description="Specifies dimension of the layer normal axis (0,1,2) -> (x,y,z).",
+    )
+
+    min_steps_along_axis: Optional[pd.PositiveFloat] = pd.Field(
+        None,
+        title="Minimal Number Of Steps Along Axis",
+        description="If not ``None`` and the thickness of the layer is nonzero, set minimal "
+        "number of steps discretizing the layer thickness.",
+    )
```
momchil-flex, line 966: I assume this is not always enforced as it could lead to very small step size for thin layers?

tidy3d/components/grid/grid_spec.py
```diff
@@ -665,14 +787,580 @@ def _preprocessed_structures(self, structures: List[StructureType]) -> List[Stru
...
+        "of the layer along axis dimension. If `min_steps_along_axis` is also specified, "
+        "refinement here is only applied if it sets a smaller grid size.",
+    )
+
+    bounds_snapping: Optional[Literal["bounds", "lower", "upper", "center"]] = pd.Field(
+        "lower",
+        title="Placing Grid Snapping Point Along Axis",
+        description="If not ``None``, enforcing grid boundaries to pass through ``lower``, "
+        "``center``, or ``upper`` position of the layer; or both ``lower`` and ``upper`` with ``bounds``.",
+    )
```
momchil-flex, line 981: Same question, how does this work if the layer is too thin but ``bounds`` is selected?

tidy3d/components/grid/grid_spec.py
```diff
@@ -665,14 +787,580 @@ def _preprocessed_structures(self, structures: List[StructureType]) -> List[Stru
...
+        description="If ``True`` and ``corner_finder`` is not ``None``, enforcing inplane "
+        "grid boundaries to pass through corners of geometries specified by ``corner_finder``.",
+    )
+
+    corner_refinement: Optional[GridRefinement] = pd.Field(
+        GridRefinement(),
+        title="Inplane Mesh Refinement Factor Around Corners",
+        description="If not ``None`` and ``corner_finder`` is not ``None``, refine mesh around "
+        "corners of geometries  specified by ``corner_finder``. "
+        "This option is similar to edge refinement if the edge is axis-aligned.",
```
momchil-flex: Sorry what does edge refinement refer to here?

tidy3d/components/grid/grid_spec.py
```diff
@@ -665,14 +787,580 @@ def _preprocessed_structures(self, structures: List[StructureType]) -> List[Stru
...
+        """size must be finite along axis."""
+        if np.isinf(values["size"][val]):
+            raise SetupError("'size' must take finite values along 'axis' dimension.")
+        return val
+
+    @classmethod
+    def from_unbounded(
+        cls,
+        axis: Axis,
+        bounds: Tuple[float, float],
```
momchil-flex, line 1016: This is a bit nitpicky perhaps but the usage `from_unbounded(bounds=...)` struck me as a bit strange. It is not immediately clear what's unbounded and what is bounded. Do you think it would make sense to call it e.g. `from_inplane_unbounded` and even maybe `layer_bounds` instead of `bounds`?

momchil-flex (approved):
This seems like impressive work, note I mostly reviewed the usage rather than the low level since there seem to be many eyes on it.

tidy3d/components/grid/grid_spec.py
weiliangjin2021: It will not be enforced if `dl_min` is set. However, if `dl_min=0/None`, we'll internally set a `dl_min` based on material response and those refinement factors, meaning that it will be enforced for thin layers. In a way, that is expected because in many classes of simulation you'll want to resolve the thin layer.

tidy3d/components/grid/grid_spec.py
weiliangjin2021: if `dl_min` is set and it's larger than the layer thickness, then only one of the snapping points will take effect.

tidy3d/components/grid/grid_spec.py
weiliangjin2021: Grids are denser closer to the edge. Probably remove this sentence?

tidy3d/components/grid/grid_spec.py
weiliangjin2021: Agree, I also feel it reads strange. I like `layer_bounds`!

tidy3d/components/grid/grid_spec.py
momchil-flex: Ok, I can see how that's probably what's needed in RF. However one thing I'm a little wary about is if people put some arbitrary small thickness for layers that they want to say are "very thin". Like ideally they should put 0 thickness and use a 2D medium but they may try instead to put like a 10nm plate with a layer refinement because they don't know better, and then get a giant number of grid cells and time steps.

tidy3d/components/grid/grid_spec.py
momchil-flex: Do you mean it's similar to bounds refinement? I'm just not sure what you are referring to. :) What is the thing that it is similar to?

tidy3d/components/grid/grid_spec.py
lucas-flexcompute: It's not easy to differentiate between the 2 scenarios (intended vs. non-intended refinement). Maybe we could issue a warning if the simulation is in optical frequencies, but I'm not super convinced it would be useful, since we already have validations in place for the number of time steps (and cost)…

lucas-flexcompute (commented):
This looks good. From what I can tell, we can use the interface easily enough to build automatic meshes for RF. It would be good to have a more clear understanding of the `bounds_refinement` and `corner_refinement` fields, i.e., how exactly do they work (maybe with an illustration?)

Am I understanding it correctly that the `bounds_refinement` will create `GridRefinement.num_cells` at both bounds? Are the cells inside the layer or outside the layer?

And for `corner_refinement`, where are the cells created? Inside or outside the polygon?

tidy3d/components/grid/grid_spec.py
```diff
@@ -665,14 +787,587 @@ def _preprocessed_structures(self, structures: List[StructureType]) -> List[Stru
...
+
+    Corner detection is performed on a 2D plane sitting in the middle of the layer. If the layer is finite
+    along inplane axes, corners outside the bounds are discarded.
+
+    Note
+    ----
+
+    This class only takes effect when :class:`AutoGrid` is applied.
+
+    Example
```
dmarek-flex, line 949: Two notes here

dmarek-flex:
When a metal layer is very thin and we do not want to mesh both sides of the metal, so as to enable a larger delta t, would it be possible to automatically choose the best interface (top or bottom) for placing a grid boundary. Consider the case attached. Is it correct that placing a grid boundary along the black line would be more economical/accurate than at the red line, since it would reduce the number of times a control volume contains >2 materials? (Near the metal corners is where I am thinking)

image

tidy3d/components/grid/grid_spec.py
weiliangjin2021: > like ideally they should put 0 thickness and use a 2D medium but they may try instead to put like a 10nm plate with a layer refinement because they don't know better

Probably low chance, as now by default, thickness will not be refined, and we only snap one bound to the grid.

tidy3d/components/grid/grid_spec.py
weiliangjin2021: Imagine that you have a metallic polygon. You'll want finer meshing near the polygon edges to resolve the more rapid field variations.

weiliangjin2021:
> Am I understanding it correctly that the bounds_refinement will create GridRefinement.num_cells at both bounds? Are the cells inside the layer or outside the layer?

Yes, at both bounds. The refinement is applied in the region that centered at the bound, so it's both inside and outside. Do you see a need only outside/inside?

> And for corner_refinement, where are the cells created? Inside or outside the polygon?

They are centered at the vertices, e.g. those magenta dashed boxes below
image

lucas-flexcompute:
> > Am I understanding it correctly that the bounds_refinement will create GridRefinement.num_cells at both bounds? Are the cells inside the layer or outside the layer?
> 
> Yes, at both bounds. The refinement is applied in the region that centered at the bound, so it's both inside and outside. Do you see a need only outside/inside?

No, that's perfect!

lucas-flexcompute (approved):

weiliangjin2021:
> When a metal layer is very thin and we do not want to mesh both sides of the metal, so as to enable a larger delta t, would it be possible to automatically choose the best interface (top or bottom) for placing a grid boundary. Consider the case attached. Is it correct that placing a grid boundary along the black line would be more economical/accurate than at the red line, since it would reduce the number of times a control volume contains >2 materials? (Near the metal corners is where I am thinking)

When the thickness of metal layer is much thinner than the Yee cell size, it looks chances are equal with black/red line? Could you share a figure on the control volume on this (let's take this discussion off-line)?

Separately, it is interesting to study the behavior of conformal meshing in a simple microstrip when the grid line is placed at black/red. I think they will lead to slightly different averaged permittivity/permeability, as the red line go through 2 materials, while black 3 materials.

dmarek-flex:
This PR is pretty great! I am still playing around with all of the options to understand what they are doing. I found perhaps one non-ideal scenario. When I used a default `bounds_refinement=td.GridRefinement()` I get an overrefined grid. The calculated `dl` for the two created `MeshOverrideStructure` is ~0.6 mm, but the resulting grid is much more refined. 

Could it be because the two `MeshOverrideStructure` associated with the bounds of the layer are offset by a small amount?
 
Second pic is when `bounds_refinement=None`. 
image
image

tidy3d/components/grid/corner_finder.py
```diff
@@ -0,0 +1,134 @@
...
+            polygon_list = ClipOperation.to_polygon_list(shapes)
+            for poly in polygon_list:
+                poly = poly.normalize().buffer(0)
+                if self.distance_threshold is not None:
+                    poly = poly.simplify(self.distance_threshold, preserve_topology=True)
+                corner_list.append(self._filter_collinear_vertices(list(poly.exterior.coords)))
+                # in case the polygon has holes
+                for poly_inner in poly.interiors:
+                    corner_list.append(self._filter_collinear_vertices(list(poly_inner.coords)))
+        return np.concatenate(corner_list)
```
dmarek-flex: Bug here if the `LayerRefinementSpec` is setup in a dumb way that it only intersects with the background.

dmarek-flex (changes requested):
As I said, this is great! Still trying it on examples, but I think I found one bug.

tidy3d/components/grid/corner_finder.py
weiliangjin2021: found that too while trying to improve test coverage.

weiliangjin2021: should be fixed now.

tidy3d/components/grid/grid_spec.py
weiliangjin2021: I think two notes are fine, as I saw it somewhere else.

weiliangjin2021: e.g. multiple notes in this page https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.TwoPhotonAbsorption.html

weiliangjin2021:
> I found perhaps one non-ideal scenario. When I used a default bounds_refinement=td.GridRefinement() I get an overrefined grid. The calculated dl for the two created MeshOverrideStructure is ~0.6 mm, but the resulting grid is much more refined.

Could you share the code here so that I can reproduce?

tidy3d/components/grid/grid_spec.py
dmarek-flex: oh I see!

weiliangjin2021:
In `plot_grid`, to distinguish between lines from override structures and snapping lines, now I set the former to be dotted, and latter to be dashed. Here is an example on the visualization effects (the dash-dotted is coming from the overlap of the two types):
image

dmarek-flex (approved):

PR #2114, opened 2024-12-09
call metadata api after uploading file

QingengWei:


momchil-flex:
Thanks @magiWei 

It seems that some tests are failing because they are actually trying to connect to the server to estimate the cost? We need to mock that function in the tests.

The other thing is, we also discussed adding a check estimate api call in `web.start` before the `submit` api is called. Or maybe it should even be added directly in `task.submit`. And then this function also needs to be mocked in the tests.

QingengWei:
> Thanks @magiWei
> 
> It seems that some tests are failing because they are actually trying to connect to the server to estimate the cost? We need to mock that function in the tests.
> 
> The other thing is, we also discussed adding a check estimate api call in `web.start` before the `submit` api is called. Or maybe it should even be added directly in `task.submit`. And then this function also needs to be mocked in the tests.

About #2， it will check on API server side.

momchil-flex:
> > Thanks @magiWei
> > It seems that some tests are failing because they are actually trying to connect to the server to estimate the cost? We need to mock that function in the tests.
> > The other thing is, we also discussed adding a check estimate api call in `web.start` before the `submit` api is called. Or maybe it should even be added directly in `task.submit`. And then this function also needs to be mocked in the tests.
> 
> About #2， it will check on API server side.

Ah ok makes sense.

For the tests I think you also need to add `mock_get_info`.

QingengWei:
> > > Thanks @magiWei
> > > It seems that some tests are failing because they are actually trying to connect to the server to estimate the cost? We need to mock that function in the tests.
> > > The other thing is, we also discussed adding a check estimate api call in `web.start` before the `submit` api is called. Or maybe it should even be added directly in `task.submit`. And then this function also needs to be mocked in the tests.
> > 
> > 
> > About #2， it will check on API server side.
> 
> Ah ok makes sense.
> 
> For the tests I think you also need to add `mock_get_info`.

What is the most recent release version? Can we merge this feature into it?

momchil-flex:
We are preparing a 2.8 pre-release right now. Normally we're not planning on making more 2.7 patches.

momchil-flex:
@magiWei I also added a commit to remove the estimated cost message from `web.monitor` as now it will be duplicated during `web.run`. What do you think?

We should also add a changelog item, I can do it when we get to a final state of this.

QingengWei:
> @magiWei I also added a commit to remove the estimated cost message from `web.monitor` as now it will be duplicated during `web.run`. What do you think?
> 
> We should also add a changelog item, I can do it when we get to a final state of this.

It makes sense. I have no new commit.

PR #2115, opened 2024-12-09
Run the mode solver through the regular webapi functions

momchil-flex:
Supersedes #2090 : rebased to pre/2.8 and fixed tests and added a new testcase.


momchil-flex:
@tylerflex pinging you because of the web api mocking changes

tests/test_web/test_webapi_mode.py
tylerflex, line 1: all of the fixtures were copied in here from other tests?

we really need to re-do these web api tests :). I also am really confused how they work.

tylerflex (approved):
Looks good to me, but I've also lost a bit track of the workings of these web API calls. Wonder if we should get another pair of eyes on it or if you feel pretty confident about the change?

dbochkov-flexcompute (approved):

PR #2117, opened 2024-12-11
Allow double precision mode solver in EME.

caseyflex:
Previously we restricted EME to use the single-precision mode solver (not completely sure why), but Filipe's recent testing makes it clear that the double-precision solver is needed to resolve the attenuation constant in low-loss waveguides (such as bends with large radius).

I did quickly check that the metadata calculation in this case looks correct.


momchil-flex (approved):
Wonder why it wasn't before but yeah let's get this in and @FilipeFcp can do some more testing.

PR #2118, opened 2024-12-11
Bump pytorch version because <=2.1 does not have cpuonly py312 wheels

yaugenst-flex:


daquinteroflex (approved):
Looks good! I wonder why the torch-cpu source is `explicit` and if this makes a difference rather than `supplemental`, but looks correct anyway since it works.

yaugenst:
Because supplemental packages are only pulled if no other higher-priority channel (i.e. PyPI in our case) can satisfy the requirements, and pytorch is distributed on PyPI with cuda builds (which we don't want). Explicit forces that specific source for the package if the package is requested. At least I think that's how it works :D

PR #2122, opened 2024-12-13
Adding ElectromagneticFieldData._check_fields_stored

momchil-flex:
This appears in some FieldData checks but was only implemented for ModeData and ModeSolverData.


lucas-flexcompute (approved):
Looks fine. I'm not sure where the other implementations were. Have they been removed from Mode*Data?

momchil-flex:
> Looks fine. I'm not sure where the other implementations were. Have they been removed from Mode*Data?

No, as each has a different error message...

PR #2123, opened 2024-12-14
Generalized unstructured

dbochkov-flexcompute:
@marc-flex @daquinteroflex One of the aspects in charge PR that I thought needed some more ironing out was storing simulation results as unstructured arrays for multiple bias values. Specifically, our current Triangular and TetrahedralGridDataset's are designed to only store one field. We could store multiple fields as separate unstructured datasets, but it meas that we duplicate grid information many times which is not memory optimal. Specifically, they look like this:
```
class UnstructuredGridDataset:
    values: IndexedDataArray  # array of length N
    points: PointDataArray  # array of shape (3, N)
    cells: CellDataArray  # array of shape (3, M) or (4, M)
```
so, in fact memory to store `values` is less than memory for `points` + `cells`.

To avoid that Marc stores in charge monitors multiple fields in some sort of disassembled way: store one full unstructured array that contains grid info, and rest of fields of as separate IndexedDataArray's. So, when we need to get an unstructured array dataset for a given bias we combine a corresponding separate IndexedDataArray with that master dataset. This definitely works, but doesn't seem like a general solution.

To allow to manage this more easily, I took a stab at generalizing unstructured datasets to be as multidimensional as we need. It more or less follows xarray data arrays behavior and how we use them in Tidy3D. Now we can define unstructured datasets with inherently more dimensonality. For example, when we consider voltage dependent data as analog to
```
class DCSpatialDataArray(DataArray):
    _dims=("x", "y", "z", "v")
```
we can define
```
class DCIndexedDataArray(DataArray):
    _dims=("indexed", "v")

class DCTriangularGridDataset(TriangularGridDataset):
    values: DCIndexedDataArray

class DCTetrahedralGridDataset(TetrahedralGridDataset):
    values: DCIndexedDataArray
```
and all methods (slicing, dicing, selecting, interpolating, etc) will work automatically. And those will replace `HeatChargeDataset`

this PR needs more polishing and I will work on that, but let me know any thoughts you might have about this


marc-flex:
Thanks Daniil for looking into this! One thing that it's not clear to me (I didn't go through the changes in depth) is how we load data for multiple voltages, since these will be in different VTU files. Will we be able to open multiple files, read and combine? Or provide a bunch of files and field to load?

dbochkov-flexcompute:
> Thanks Daniil for looking into this! One thing that it's not clear to me (I didn't go through the changes in depth) is how we load data for multiple voltages, since these will be in different VTU files. Will we be able to open multiple files, read and combine? Or provide a bunch of files and field to load?

Yeah, need to load each file separately and combine them into a single array

tidy3d/components/data/unstructured/base.py
```diff
@@ -0,0 +1,1742 @@
...
+    def is_uniform(self):
+        """Whether each element is of equal value in ``values``."""
+        return self.values.is_uniform
+
+    @cached_property
+    def _values_coords_dict(self):
+        """Non-spatial dimensions are corresponding coordinate values of stored data."""
+        coord_dict = {dim: self.values.coords[dim].data for dim in self.values.dims}
+        _ = coord_dict.pop("index")
+        return coord_dict
```
marc-flex, line 258: Isn't this giving out an empty dictionary?

tests/test_data/test_datasets.py
```diff
@@ -301,11 +322,22 @@ def operation(arr):
     assert result.name == ds_name
 
 
+@pytest.mark.parametrize("dataset_type_ind", [0, 1])
 @pytest.mark.parametrize("ds_name", ["test123", None])
-def test_tetrahedral_dataset(log_capture, tmp_path, ds_name, no_vtk=False):
+def test_tetrahedral_dataset(log_capture, tmp_path, ds_name, dataset_type_ind, no_vtk=False):
```
daquinteroflex, line 327: I'd change making this index variables more verbose for clarity, but it's a stylistic preference anyway so not that important.

daquinteroflex:
Marc mentioned this was broadly ready, and I can't identify anything evidently wrong (without delving profoundly into the logic). So will go ahead and merge in order to do the rest of the refactor. Any specific issues that arise we can deal with after the refactor? I'd just like to check we have enough test coverage but looks alright broadly!

PR #2124, opened 2024-12-14
:zap: Charge API & SPICE/medium common types

daquinteroflex:
WIP.

Continuation of https://github.com/flexcompute/tidy3d/pull/2106 and https://github.com/flexcompute/tidy3d/pull/2077

- [X] Consolidates Marc's great work in a software structure that I think is more extensible and generic
- [ ]  Properly integrate RF and CHARGE classes 
  - [X] Explored exact integration points and data for the mediums. Proposed a strategy 
  - [X] Change the structure of chargespec to use the standard `Medium` parameters from the frontend
  - [X] Explore a shared stategy for the RF + ChargeMonitorData etc. Basically, whenever we begin doing frequency-dependent or non-static stuff we should then reuse the same monitors as the RF toolsets.
  - [x] Introduce an equivalent `SuperMedium`
  - [ ] Fully fix the validators to share the RF FDTD and TCAD models, depends on backend implementation.
  - [X] Rename classes in a more standard way
    - [X] Cleaner tcad model naming. 
    - [X] The ideal situation in terms of defining the equations would be to actually provide a set of serializable model equations using Yannick's technique and dealing with that on the backend, but we can use pre-defined models for now.
- [ ] Add theoretical documentation of what each class is and each parameter. (Determine functional and static data definitions accrodingly)
- [ ] Update all tests to new structure
- [ ] Write proper docs of all classes and how they work (happy to do this depending if we're happy on this approach)
- [ ] Implement the shared medium specification in a way that is compatible with the other solvers as discussed
- [ ] Unlike the previous PR, this one implements the upgrade with a considerably more progressive approach.
- [ ] Deal with the ruff issue from https://github.com/flexcompute/tidy3d/pull/2136#pullrequestreview-2518646777
- [ ] Maybe some renaming from https://github.com/flexcompute/tidy3d/pull/2123#issuecomment-2558114440

The existing structure is maintained and updated alongside the new structure in order to share the same names, and hence similar tests structures without a major refactor.

**Notes:**

- We need to do the PotentialData completely differently if we want to support multiple DC + RF sources. I'm not sure the current implementation will be nicely integrate-able with different sets of analysis.
- Should we have the VoltageBC also work as an iterative parameter? Part of me thinks the proper setup would be to create a boundary condition from a DC source rather than iterate directly from it.
- One reason why the static naming maybe makes sense is due to the temporal nature of our simulations, which when they are multi-physical then this changes. For example, free carriers may change on a static heat simulation, independent of a DC voltage/current. As such static represents this temporal relationship clearly I think.
- The issue I have with the definition of SemiConductor is that this does not imply the "active" nature of a material. Si, Si02 are all semiconductors, but they can also be treated as insulators, or conductors, depending on the simulation conditions. `Electronic` is probably not the answer. "ActiveSemiconductor" maybe sounds good to me.
- We need to decide on the threshold implementation here.
- Temporarily calling medium materials until we do the full migration.

Before continuing, things that we need to decide:
- [ ] The management of the electric_spec


marc-flex:
> One reason why the static naming maybe makes sense is due to the temporal nature of our simulations, which when they are multi-physical then this changes. For example, free carriers may change on a static heat simulation, independent of a DC voltage/current. As such static represents this temporal relationship clearly I think.

I'm not quite sure I follow. When you have time we should probably discuss

> The issue I have with the definition of SemiConductor is that this does not imply the "active" nature of a material. Si, Si02 are all semiconductors, but they can also be treated as insulators, or conductors, depending on the simulation conditions. `Electronic` is probably not the answer. "ActiveSemiconductor" maybe sounds good to me.

I think it's not a matter of what they are but how they're treated, and from Charge perspective we have either insulator or semiconductor. (conductor not yet supported). Another point is that I'd find the name "ActiveSemiconductor" a bit confusing. (like I'd ask myself, what's activating it?, how is it active?). If anything, I'd add `SemiconductorParameters` since that's what we're recording: band-gap energy (at 300K), electron/hole density of states in the conduction/valence bands, etc.

daquinteroflex:
Rebase turned out to be such a hell I've decided it's easier to start again. Closing this for the next branch, but also to keep it as a reference.

PR #2125, opened 2024-12-16
Thread-safe creation of temp dir

lsarra-ax:
Handle the exception triggered when the temp folders already exist in a multi-threaded setting 

See 
https://stackoverflow.com/questions/42544885/error-when-mkdir-in-multi-threads-in-python

### DIFF
--- a/tidy3d/web/cli/app.py
+++ b/tidy3d/web/cli/app.py
@@ -1,7 +1,7 @@
 """
 Commandline interface for tidy3d.
 """
-
+import errno
 import json
 import os.path
 import ssl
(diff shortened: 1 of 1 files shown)


tylerflex:
Thanks @lsarra-ax for contributing this. @lei-flex @momchil-flex could you take a look and see what you think?

tidy3d/web/cli/app.py
```diff
@@ -16,8 +16,13 @@
...
 
-if not os.path.exists(TIDY3D_DIR):
-    os.mkdir(TIDY3D_DIR)
+
+# Prevent race condition on threads
+try:
+    os.makedirs(TIDY3D_DIR)
+except OSError as e:
+    if e.errno != errno.EEXIST:
+        raise  # Reraise if failed for reasons other than existing already
```
momchil-flex: Thanks. This seems equivalent to `os.makedirs(TIDY3D_DIR, exist_ok=True)`. According to the link you shared, this version is only needed for python < 3.2, but we require >= 3.9, so let's just use the simple version. Could you update?

lsarra-ax: Thanks for your reply! Sure, this should work as well

momchil-flex (approved):
Thanks. Merging into my branch so I can squash the commits and add a changelog item before merging in the main code (pre/2.8, not develop, as we will be releasing the 2.8.0rc1 version today).

PR #2127, opened 2024-12-16
Convert JaxSimulation to Simulation if passed in mode solver

momchil-flex:


tylerflex (approved):
looks good. can you add a changelog too just to mention this was fixed?

PR #2129, opened 2024-12-17
Refactored DirectivityData to store far fields and flux as fields, and compute radiation parameters via properties

QimingFlex:


momchil-flex:
Does it make more sense to store the linearly polarized fields and add properties to get the circularly polarized ones?

QimingFlex:
> Does it make more sense to store the linearly polarized fields and add properties to get the circularly polarized ones?

Linearly polarized fields are the regular far-field components Etheta​ and Ephi​. I was previously thinking of putting E_left and E_right as properties of FarfieldProjectionMonitor, but doing so would require users to define both FarfieldProjectionMonitor and DirectivityMonitor to obtain all radiation characteristics. So I placed E_left​ and E_right​ under DirectivityMonitor, allowing users to access these radiation characteristics with just one monitor. Maybe a better way is to keep all six far-field components (Er to Hphi​) as fields in DirectivityData and introduce E_left​ and E_right as properties?

dmarek-flex:
> Does it make more sense to store the linearly polarized fields and add properties to get the circularly polarized ones?

I agree. For simplicity and extensibility, I think the `DirectivityData` should only store Etheta/Ephi far fields and total flux. Most antenna characteristics of interest can be derived from that data. We should not even need the radial components in the far field. (We could also compute H far fields on-the-fly given the background medium, but maybe simpler to keep it the way it is and store it explicitly.)

@QimingFlex can you think of any parameters of interest where the far field approximation is not used for antennas?

QimingFlex:
> > Does it make more sense to store the linearly polarized fields and add properties to get the circularly polarized ones?
> 
> I agree. For simplicity and extensibility, I think the `DirectivityData` should only store Etheta/Ephi far fields and total flux. Most antenna characteristics of interest can be derived from that data. We should not even need the radial components in the far field. (We could also compute H far fields on-the-fly given the background medium, but maybe simpler to keep it the way it is and store it explicitly.)
> 
> @QimingFlex can you think of any parameters of interest where the far field approximation is not used for antennas?

This sounds like a good plan. We store DirectivityData only with Etheta/Ephi, Htheta/Hphi, and total flux. Other characteristics, such as directivity, axial ratio, and polarized fields, are defined as properties. How do you think? @weiliangjin2021

weiliangjin2021:
> This sounds like a good plan. We store DirectivityData only with Etheta/Ephi, Htheta/Hphi, and total flux. Other characteristics, such as directivity, axial ratio, and polarized fields, are defined as properties. How do you think?

Sounds good to me.

momchil-flex:
Sorry, wouldn't this now be almost identical to a far field monitor? Should we consider unifying, e.g. deprectating DirectivityMonitor, and adding the directivity, axial ratio, etc. properties to the FarFieldAngleMonitor?

QimingFlex:
> Sorry, wouldn't this now be almost identical to a far field monitor? Should we consider unifying, e.g. deprectating DirectivityMonitor, and adding the directivity, axial ratio, etc. properties to the FarFieldAngleMonitor?

The difference lies in DirectivityMonitor inheriting FluxMonitor to compute directivity. Do you think it would be better for FarFieldAngleMonitor to inherit FluxMonitor and store flux as a field in FarFieldAngleData? This will allow DirectivitiMonitor to be depracted.

momchil-flex:
Ok, I think it makes sense as is.

Note that you're modifying the notebooks submodule.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added an option to simulate plane wave propagation at fixed angles by setting parameter `angular_spec=FixedAngleSpec()` when defining a `PlaneWave` source.
 - The universal `tidy3d.web` can now also be used to handle `ModeSolver` simulations.
 - Support for differentiation with respect to `PolySlab.slab_bounds`.
+- Polarized far field components to `DirectivityData`.
```
dmarek-flex: suggestion: `- DirectivityMonitor can now be used to decompose fields into circular polarization states.`

CHANGELOG.md
```diff
@@ -50,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Double precision mode solver is now supported in EME.
 - `estimate_cost` is now called at the end of every `web.upload` call.
 - Internal refactor of adjoint shape gradients using `GradientSurfaceMesh`.
+- Refactored `DirectivityData` fields: Now stores only far fields and flux. Radiation characteristics (e.g., directivity, axial ratio) are computed through properties.
```
dmarek-flex: Again I would suggest focusing on how the `DirectivityMonitor` has been changed. For example: The behavior of the `DirectivityMonitor` was changed, so that only far fields and flux are stored. Other radiation characteristics ...

tidy3d/components/data/monitor_data.py
```diff
@@ -2047,13 +2047,105 @@ class DirectivityData(MonitorData):
...
+
+        # Cap the axial ratio values at 1 / AXIAL_RATIO_CAP
+        axial_ratio_inverse = axial_ratio_inverse.where(
+            axial_ratio_inverse >= 1 / AXIAL_RATIO_CAP, 1 / AXIAL_RATIO_CAP
+        )
+
+        return 1 / axial_ratio_inverse
+
+    @property
+    def left_polarization(self):
```
dmarek-flex: `left_hand_polarization`
in the comments add parenthetically that it is the counter clockwise component of circular polarization

same for the other polarization

tests/test_data/test_monitor_data.py
```diff
@@ -190,8 +186,16 @@ def make_flux_data():
 
 
 def make_directivity_data():
+    data = make_far_field_data_array()
```
dmarek-flex, line 189: modify test to include the new fields `left_polarization` and `right_polarization`.

tidy3d/components/data/monitor_data.py
```diff
@@ -2047,13 +2047,105 @@ class DirectivityData(MonitorData):
         description="Monitor describing the angle-based projection grid on which to measure directivity data.",
     )
 
-    directivity: DirectivityDataArray = pd.Field(
-        ..., title="Directivity", description="Directivity with an angle-based projection grid."
+    flux: FluxDataArray = pd.Field(
+        ..., title="Flux", description="Flux values in the frequency-domain."
```
dmarek-flex: dont need a hyphen in "frequency-domain" for this case

dmarek-flex (changes requested):
Great, this looks a lot simpler @QimingFlex! Couple of simple things and yea just need to revert the notebooks module.

tests/test_data/test_monitor_data.py
QimingFlex: `left_polarization` and `right_polarization` are properties now rather than fields.

CHANGELOG.md
weiliangjin2021: Missing "are stored" in the revised version.

CHANGELOG.md
```diff
@@ -51,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `estimate_cost` is now called at the end of every `web.upload` call.
 - Internal refactor of adjoint shape gradients using `GradientSurfaceMesh`.
 - Enhanced progress bar display in batch operations with better formatting, colors, and status tracking.
+- The behavior of `DirectivityData` was changed, so that only far fields and flux are stored. Radiation characteristics (e.g., directivity, axial ratio) are computed through properties.
```
weiliangjin2021: "are computed" --> "can be accessed"

tidy3d/components/data/monitor_data.py
```diff
@@ -2028,13 +2027,14 @@ class DirectivityData(MonitorData):
...
     -------
-    >>> from tidy3d import DirectivityDataArray, AxialRatioDataArray
+    >>> from tidy3d import DirectivityDataArray, AxialRatioDataArray, FieldProjectionAngleDataArray
     >>> f = np.linspace(1e14, 2e14, 10)
     >>> r = np.atleast_1d(1e6)
     >>> theta = np.linspace(0, np.pi, 10)
     >>> phi = np.linspace(0, 2*np.pi, 20)
     >>> coords = dict(r=r, theta=theta, phi=phi, f=f)
     >>> values = np.random.random((len(r), len(theta), len(phi), len(f)))
+    >>> complex_values = (1+1j) * np.random.random((len(r), len(theta), len(phi), len(f)))
```
weiliangjin2021: Please update the example code here. Line 2041 seem outdated.

tidy3d/components/data/monitor_data.py
```diff
@@ -2047,14 +2047,104 @@ class DirectivityData(MonitorData):
...
+        description="Spatial distribution of theta-component of the magnetic field.",
+    )
+    Hphi: FieldProjectionAngleDataArray = pd.Field(
+        ...,
+        title="Hphi",
+        description="Spatial distribution of phi-component of the magnetic field.",
     )
 
+    @property
+    def directivity(self):
```
weiliangjin2021: return signature `->float`. The same for other properties

tidy3d/components/data/monitor_data.py
```diff
@@ -2047,14 +2047,104 @@ class DirectivityData(MonitorData):
...
+        flux_normed = self.flux / (4 * np.pi * self.monitor.proj_distance**2)
+
+        return power / flux_normed
+
+    @property
+    def axial_ratio(self):
+        """Axial Ratio (AR) in the frequency domain as a function of angles theta and phi.
+        AR is a dimensionless quantity defined as the ratio of the major axis to the minor
+        axis of the polarization ellipse.
+        Note
```
weiliangjin2021: I believe "Note" only works appropriately if you have a blank line above "Note"

tidy3d/components/data/monitor_data.py
```diff
@@ -2047,14 +2047,104 @@ class DirectivityData(MonitorData):
...
+        # Cap the axial ratio values at 1 / AXIAL_RATIO_CAP
+        axial_ratio_inverse = axial_ratio_inverse.where(
+            axial_ratio_inverse >= 1 / AXIAL_RATIO_CAP, 1 / AXIAL_RATIO_CAP
+        )
+
+        return 1 / axial_ratio_inverse
+
+    @property
+    def left_polarization(self):
+        "Left polarization (counter clockwise component of circular polarization) electric far field with an angle-based projection grid."
```
weiliangjin2021: The line here looks super long. Better to split the line. The same to the next property.

weiliangjin2021: Also please improve the sentence a bit. E.g. "counterclockwise"; "Left polarized far field component of electric field..."

weiliangjin2021 (commented):
Could you also add tests for left and right polarization?

tidy3d/components/data/monitor_data.py
QimingFlex: would it be better to use `DataArray`?

weiliangjin2021: ah sorry, it's DataArray

weiliangjin2021 (approved):
Approve once tests for right/left polarization have been added

tests/test_data/test_monitor_data.py
dmarek-flex: True, but make sure those properties are tested in `test_directivity_data():`

dmarek-flex (approved):
Looks great, approved! but yeah please add testing of the new properties

PR #2130, opened 2024-12-18
QuasiUniformGrid and min_steps_per_sim_size`

weiliangjin2021:
Two main features:

- `QuasiUniformGrid`: grid is roughly uniform, but locally snapped. This makes it more useful than `UniformGrid` in many simulations.
-  In `AutoGrid`, adds a field `min_steps_per_sim_size`. Now the grid step size is constrained by both `min_steps_per_sim_size` and `min_steps_per_wvl`. In many RF simulations, the entire simulation domain is deep subwavelength. Previously, one needs to set a dramatically large value of `min_steps_per_wvl` so that the simulation domain has more than 1 grid point. `min_steps_per_sim_size` will be more useful in those simulations.

### DIFF
--- a/tests/test_components/test_grid_spec.py
+++ b/tests/test_components/test_grid_spec.py
@@ -381,3 +381,46 @@ def test_custom_grid_boundaries():
         boundary_spec=td.BoundarySpec.all_sides(boundary=td.PML(num_layers=num_layers))
     )
     assert np.allclose(sim_pml.grid.boundaries.x, np.linspace(-3, 3, 31))
+
+
+def test_small_sim_with_min_steps_per_sim_size():
+    """Test if `min_steps_per_sim_size` takes effect for a small simulation domain."""
+
+    # only single grid because simulation domain too small
+    sim = td.Simulation(
+        size=(1, 1, 1),
+        run_time=1e-12,
+        grid_spec=td.GridSpec.auto(wavelength=1e4, min_steps_per_sim_size=1),
+        boundary_spec=td.BoundarySpec.pml(),
+    )
+    assert sim.num_cells == 1
+
+    # apply default min_steps_per_sim_size
+    sim = sim.updated_copy(grid_spec=td.GridSpec.auto(wavelength=1e4, min_steps_per_sim_size=10))
+    assert sim.num_cells > 500
+
+
+def test_quasiuniform_grid():
+    """Test grid is quasi-uniform that adjusts to structure boundaries."""
+    box = td.Structure(
+        geometry=td.Box(size=(0.5, 0.5, 0.5)),
+        medium=td.Medium(permittivity=25),
+    )
+    sim = td.Simulation(
+        size=(1, 1, 1),
+        grid_spec=td.GridSpec.quasiuniform(dl=0.1),
+        boundary_spec=td.BoundarySpec.pml(),
+        structures=[box],
+        run_time=1e-12,
+    )
+
+    # snapped to the boundary 0.25 because of box
+    assert any(np.isclose(sim.grid.boundaries.x, 0.25))
+
+    # add snapping points
+    pos = 0.281
+    snapping_points = [(pos, pos, pos)]
+    sim = sim.updated_copy(
+        grid_spec=td.GridSpec.quasiuniform(dl=0.1, snapping_points=snapping_points)
+    )
+    assert any(np.isclose(sim.grid.boundaries.x, pos))
--- a/tidy3d/components/grid/grid_spec.py
+++ b/tidy3d/components/grid/grid_spec.py
@@ -14,7 +14,7 @@ from ...log import log
 from ..base import Tidy3dBaseModel
 from ..geometry.base import Box
 from ..source.utils import SourceType
-from ..structure import Structure, StructureType
+from ..structure import MeshOverrideStructure, Structure, StructureType
 from ..types import TYPE_TAG_STR, Axis, Coordinate, Symmetry, annotate_type
 from .grid import Coords, Coords1D, Grid
 from .mesher import GradedMesher, MesherType
@@ -219,6 +219,9 @@ class UniformGrid(GridSpec1d):
     See Also
     --------
 
+    :class:`QuasiUniformGrid`
+        Specification for quasi-uniform grid along a given dimension.
+
     :class:`AutoGrid`
         Specification for non-uniform grid along a given dimension.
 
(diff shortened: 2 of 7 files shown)


weiliangjin2021:
Uniform grid vs Quasi-uniform grid (from [notebook](https://github.com/flexcompute/tidy3d-notebooks/blob/9449da386dc669b0c69852f49b0e6604c0d65cf2/AutoGrid.ipynb))
image
image

tests/test_components/test_grid_spec.py
```diff
@@ -381,3 +381,46 @@ def test_custom_grid_boundaries():
...
+    # snapped to the boundary 0.25 because of box
+    assert any(np.isclose(sim.grid.boundaries.x, 0.25))
+
+    # add snapping points
+    pos = 0.25
+    snapping_points = [(pos, pos, pos)]
+    sim = sim.updated_copy(
+        grid_spec=td.GridSpec.quasiuniform(dl=0.1, snapping_points=snapping_points)
+    )
+    assert any(np.isclose(sim.grid.boundaries.x, pos))
```
dmarek-flex: Might be a false positive, since the points should be there from the `box`?

tidy3d/components/grid/grid_spec.py
```diff
@@ -529,7 +520,144 @@ def _make_coords_initial(
         return np.array(bound_coords)
 
 
-GridType = Union[UniformGrid, CustomGrid, AutoGrid, CustomGridBoundaries]
+class QuasiUniformGrid(AbstractAutoGrid):
+    """Very simillar to :class:`UniformGrid` that generates uniform 1D grid, but grid point locations can
```
dmarek-flex: `similar` typo

tidy3d/components/grid/grid_spec.py
```diff
@@ -529,7 +520,144 @@ def _make_coords_initial(
...
+                for ind, dl_axis in enumerate(structure.dl):
+                    if dl_axis is None:
+                        dl[ind] = None
+            processed_structures.append(MeshOverrideStructure(geometry=structure.geometry, dl=dl))
+        return processed_structures
+
+    @property
+    def _dl_min(self) -> float:
+        """Lower bound of grid size."""
+        return 0.5 * self.dl
```
dmarek-flex: It might be nice for `dl_min` to be an optional field, like in `AutoGrid`.

tidy3d/components/grid/grid_spec.py
```diff
@@ -529,7 +520,144 @@ def _make_coords_initial(
...
+    **Notebooks:**
+        * `Using automatic nonuniform meshing <../../notebooks/AutoGrid.html>`_
+    """
+
+    dl: pd.PositiveFloat = pd.Field(
+        ...,
+        title="Grid Size",
+        description="Grid size for quasi-uniform grid generation.",
+        units=MICROMETER,
+    )
```
dmarek-flex: Add that it signifies the maximum grid size. in some places it might be smaller, but never greater than `dl`.

dmarek-flex (approved):
Great thanks @weiliangjin2021, just a couple of small comments.

I also think we should make it clear that `snapping_points` take priority over `override_structure` boundaries when they are closer than `dl_min`. Maybe add a small comment in the `snapping_points` field of `GridSpec`.

tests/test_components/test_grid_spec.py
weiliangjin2021: Good point. I copied wrong value from my notebook.

tidy3d/components/grid/grid_spec.py
```diff
@@ -529,7 +531,138 @@ def _make_coords_initial(
...
+        10.0,
+        title="Minimal Number of Steps Per Simulation Domain Size",
+        description="Minimal number of steps per longest edge length of simulation domain "
+        "bounding box. This is useful when the simulation domain size is subwavelength.",
+        ge=1.0,
+    )
+
+    def _dl_max(self, sim_size: Tuple[float, 3]) -> float:
+        """Upper bound of grid size, constrained by `min_steps_per_sim_size`."""
+        return max(sim_size) / self.min_steps_per_sim_size
```
QimingFlex, line 648: To make this more general, would it be better to take the smaller value between max(sim_size) / self.min_steps_per_sim_size and wavelength / self.min_steps_per_wvl?

weiliangjin2021: `wavelength / self.min_steps_per_wvl` can be different in different materials. Currently, the maximal step size is the minimal value of two.

weiliangjin2021: that part is implemented here: https://github.com/flexcompute/tidy3d/blob/dad7fc3e9eb1f0e06b1f8d81d77aee9d14cb06a7/tidy3d/components/grid/mesher.py#L549

QimingFlex:
Thanks @weiliangjin2021 for this great feature!  This is useful in CPW simulations when we want to avoid reflections at non-uniform mesh interfaces while still need to snap grid to edges. I have only one slight comment for now.

tidy3d/components/grid/grid_spec.py
QimingFlex: I see. That's what I was looking for. Thanks.

momchil-flex (commented):
So in text you describe the QuasiUniformGrid as a type of UniformGrid that still snaps to special points, but in code it seems to be much closer to an AutoGrid (subclassing from the same AbstractAutoGrid now). It makes me wonder about a few things:
- Is the name QuasiUniformGrid appropriate: did you come up with this or is it used somewhere else?
- Should this just be an option in AutoGrid as opposed to its own class? Like, ignore the structure media and use this specific dl, but still apply the auto gridding?
- And anyway if not, maybe at least also explain the connection to AutoGrid, i.e. it uses the same meshing but ignores the media information in favor for a user-defined `dl`.

Looks great otherwise!

weiliangjin2021:
> * Is the name QuasiUniformGrid appropriate: did you come up with this or is it used somewhere else?

Myself

> * Should this just be an option in AutoGrid as opposed to its own class? Like, ignore the structure media and use this specific dl, but still apply the auto gridding?

From user's perspective, it feels much closer to UniformGrid. So originally, I was thinking just adding a flag to UniformGrid to indicate whether to locally tune the grid. But yeah, implementation-wise, it's closer to AutoGrid. So I'm not sure what's the best approach.

momchil-flex:
Putting aside the external usage for now, in terms of internal implementation, would it make sense to define this simply as something like `AutoGrid(ignore_media=True)`, or is there more to it and it would get tangled up to handle it this way?

weiliangjin2021:
> Putting aside the external usage for now, in terms of internal implementation, would it make sense to define this simply as something like `AutoGrid(ignore_media=True)`, or is there more to it and it would get tangled up to handle it this way?

That's right. A minor additional part is that grid size can be defined more directly through `dl`.

weiliangjin2021:
To summarize: internally it's a special type of `AutoGrid` with `ignore_media=True`. From external user point of view, it's more natural to think of it as a special type of `UniformGrid` with `enable_snapping=True`. 

Both approaches are fine to me, although I slightly prefer the latter where one can easily switch between truly uniform grid and an almost uniform grid.

What do you think @momchil-flex @tylerflex @dmarek-flex ?

momchil-flex:
How does the default grid in CST work? What is usually defined? Can you make it take media into account and is that on by default or not?

dmarek-flex:
> Both approaches are fine to me, although I slightly prefer the latter where one can easily switch between truly uniform grid and an almost uniform grid.

I also prefer this approach. In that case `dl` is simply an upper bound on grid size.

weiliangjin2021:
> How does the default grid in CST work? What is usually defined? Can you make it take media into account and is that on by default or not?

Very similar to AutoGrid in Tidy3d, but with several additional refinement features such as edge/face refinement . The medium is taken into account by default, but you can untick the box to leave out the medium.

One main motivation for QuasiUniformGrid is that in some simulations it is preferable to apply uniform grid along a particular axis; but at the same time, you want grids to snap to material edges and lumped element. Currently, i think we just warn that mesh override structures won't take effect along uniform grid axis.

momchil-flex:
Sorry one last question about it. Let's think about how this will appear in our GUI. It sounds go me that if we make it an AutoGrid option like `ignore_media` or `use_media`, it will be much closer to what you described in CST? On the other hand if a user wants to switch between the two, the way it is now it requires changing the entire grid definition from AutoGrid to QuasiUniformGrid.

Eventually I'm fine either way if you think it makes sense, just wanted to pose this one last question.

tidy3d/components/grid/grid_spec.py
```diff
@@ -529,7 +536,140 @@ def _make_coords_initial(
         return np.array(bound_coords)
 
 
-GridType = Union[UniformGrid, CustomGrid, AutoGrid, CustomGridBoundaries]
+class QuasiUniformGrid(AbstractAutoGrid):
+    """Similar to :class:`UniformGrid` that generates uniform 1D grid, but grid positions
+    are locally fine tuned to be snaped to snapping points and the edges of structure bounding boxes.
+    Internally, it is using the same meshing method as :class:`AutoGrid`, but it ignores material information in
+    favor for a user-defined grid size.
```
weiliangjin2021, line 543: @momchil-flex The connection has been added here now.

PR #2132, opened 2024-12-19
add ability to not distribute lumped elements

dmarek-flex:
At high frequencies and mostly in the case of high impedances, distributing the lumped element over multiple grid cells breaks the assumption that the electric field is uniform within the lumped element. We now offer the ability to restrict the network portion of the lumped element to one grid cell and use automatically created PEC connections to connect the network portion of the lumped element to the bounds of the lumped element. We also have the ability restrict the network portion along the voltage axis but distribute the element **laterally**.

These PEC connections introduce parasitic inductances and capacitances that depend on the discretization used and which effectively modify the simulated network. In some cases, it will be possible to account for these, so we offer estimates for the values of these inductances and capacitances.

Example usage has been added to the [associated notebook tutorial](https://github.com/flexcompute/tidy3d-notebooks/pull/214).


dmarek-flex:
Here are some pictures of the different strategies we have for distributing lumped elements at the moment:

`DistributeSpec.Off `  and lateral size 0 gives wire type connections (high parasitic inductance)
image

`DistributeSpec.Off ` and lateral size  !=0 gives sheet type connections (lower inductance, higher capacitance)
image

`DistributeSpec.LaterallyOnly ` and lateral size  !=0 gives sheet type connections and distributes network portion (slightly less parasitic inductance)
image

`DistributeSpec.On ` this is the default approach and has the lowest parasitic inductance and capacitance
image

Overall `DistributeSpec.On ` seems to be the best approach in most cases. The exception is when you would like to model a network with a high inductance at high frequencies, then the parasitic inductance can be accounted for in the network model and its effect minimized.

dmarek-flex:
Let's start with an RF team round of reviews.

dmarek-flex:
And here is a circuit model for the parasitic effects that the connections add: 

image

$Y(s)$ is the network that you would like to model in the lumped element.

tidy3d/components/lumped_element.py
```diff
@@ -19,21 +22,50 @@
...
+    The lumped element is restricted to a single Yee cell.
+    """
+    LaterallyOnly = 1
+    """
+    The lumped element is only distributed along the lateral direction.
+    """
+    On = 2
+    """
+    The lumped element is distributed along the lateral and voltage axis directions.
+    """
```
weiliangjin2021: In our current codebase, we tend to use `Literal["Off", "LaterallyOnly", "On]` instead of `Enum`, and put those descriptions in the field description of `distribute_spec`. Not sure which one is more suitable here. @tylerflex what do you think?

tidy3d/components/lumped_element.py
```diff
@@ -884,11 +918,86 @@ class LinearLumpedElement(RectangularLumpedElement):
...
+            size[self.lateral_axis] = self.size[self.lateral_axis]
+        if self.dist_spec is DistributeSpec.On:
+            cell_center[self.voltage_axis] = self.center[self.voltage_axis]
+            size[self.voltage_axis] = self.size[self.voltage_axis]
+
+        cell_box = Box(center=cell_center, size=size)
+
+        snap_spec = self._snapping_spec
+        # Expand from zero size along the voltage and lateral axes
+        if size[self.voltage_axis] == 0:
```
weiliangjin2021, line 959: Will zero size along voltage axis occur in `LumpedResistor` and others?

tidy3d/components/lumped_element.py
```diff
@@ -884,11 +918,86 @@ class LinearLumpedElement(RectangularLumpedElement):
...
+
+    def _create_connection_boxes(
+        self, cell_box: Box, grid: Grid
+    ) -> tuple[Optional[Box], Optional[Box]]:
+        """Creates PEC structures that connect the network portion of the lumped element to the
+        boundaries of the lumped element.
+        """
+        element_box = self.to_geometry(grid)
+        element_min = list(element_box.bounds[0])
+        element_max = list(element_box.bounds[1])
```
weiliangjin2021: `element_min, element_max = element_box.bounds`

tidy3d/components/lumped_element.py
```diff
@@ -884,11 +918,86 @@ class LinearLumpedElement(RectangularLumpedElement):
...
+        cell_min = cell_box.bounds[0]
+        cell_max = cell_box.bounds[1]
```
weiliangjin2021: `cell_min, cell_max = cell_box.bounds`

tidy3d/components/lumped_element.py
```diff
@@ -884,11 +918,86 @@ class LinearLumpedElement(RectangularLumpedElement):
...
+
+        top_min = list(element_min)
+        top_min[self.voltage_axis] = cell_max[self.voltage_axis]
+        bottom_max = list(element_max)
+        bottom_max[self.voltage_axis] = cell_min[self.voltage_axis]
+
+        # Create "wires" if the size is 0 along the lateral axis
+        if isclose(self.size[self.lateral_axis], 0, rel_tol=fp_eps, abs_tol=fp_eps):
+            lateral_center = cell_box.center[self.lateral_axis]
+            width = max(fp_eps, fp_eps * abs(lateral_center))
```
weiliangjin2021, line 984: what happens if `width` here is set to 0?

weiliangjin2021 (commented):
First pass through lumped_element. All look quite clear! Will take another look after break

tidy3d/components/lumped_element.py
dmarek-flex: I wanted to keep the behavior of `LumpedResistor` as is. I could modify it to allow the same types of operations as the `LinearLumpedElement`, but thought to keep it around as sort of a legacy implementation that will be deprecated soon.

tidy3d/components/lumped_element.py
dmarek-flex: If `width` were allowed to be 0, the resulting operations in `_volumetric_structures_grid` would fail to create an equivalent for the 2D structure, since it relies on polygon boolean operations.

tidy3d/components/lumped_element.py
dmarek-flex: Need it as a list, but found a similar way to make it cleaner

QimingFlex:
Looks great, @dmarek-flex! I have a few questions, which I’ll address as I finish reviewing the code.

tidy3d/components/lumped_element.py
```diff
@@ -884,11 +918,84 @@ class LinearLumpedElement(RectangularLumpedElement):
...
+        "the lumped element.",
+    )
+
+    def _create_box_for_network(self, grid: Grid) -> Box:
+        """Creates a box for the network portion of the lumped element, where the equivalent
+        pole residue medium will be added.
+        """
+        # Snap center to closest electric field position
+        snap_location = 3 * [SnapLocation.Boundary]
+        snap_location[self.voltage_axis] = SnapLocation.Center
```
QimingFlex, line 944: Does this snap the point to the nearest magnetic field position, or am I mistaken?

tidy3d/components/lumped_element.py
```diff
@@ -898,10 +1005,110 @@ def to_structure(self, grid: Grid = None) -> Structure:
             other_component: Medium(permittivity=1),
         }
         return Structure(
-            geometry=box,
+            geometry=cell_box,
             medium=Medium2D(**medium_dict),
         )
 
+    def to_structure(self, grid) -> Optional[Structure]:
```
QimingFlex: Consider renaming this to something clearer? such as 'to_PEC_connections_structure`.

tidy3d/components/lumped_element.py
dmarek-flex: also interested in @yaugenst-flex's input. I like that `Enum` allows editors to easily autocomplete, but that is the main reason for me!

tidy3d/components/lumped_element.py
dmarek-flex: It basically finds the electric field component aligned with the voltage axis closest to the center of the `Box`. Due to the nature of the Yee grid,  the E component is situation on cell Boundaries, except along the voltage axis (field component axis), which is situated on the grid centers, aka, dual grid.

tidy3d/components/lumped_element.py
dmarek-flex: good point. to follow the base class I do kind of need a method called `to_structure`. But I reverse which one was which to make it clear that this method creates the PEC connections.

tidy3d/components/lumped_element.py
yaugenst-flex: Yeah I think I'll have to side with @weiliangjin2021 on this one 😅 I know we discussed enums previously and I still think they are fine in principle, but our convention for `xxxSpec` on the frontend is that they are `Tidy3dBaseModel`. `DistributeSpec` will expose a pretty different user-facing API with the current implementation, and consistency is king.

An alternative approach would be to have a `DistributeSpec(Tidy3dBaseModel)` that has a `DistributeLumpedEnum` or something like that as a field. But since there are only three choices and the enum does not contain any higher-level functionality to act on those choices, I do think that `Literal["Off", "LaterallyOnly", "On]` is more natural here. In fact, it should probably be defined as a type somewhere, e.g. `LumpedElementDistributeType`.

Regarding autocomplete, I think most editors should be able to pick up on `Literal` types and present you with the choices, but not 100% on this.

dmarek-flex: Fair enough, consistency is king!

dmarek-flex: Also yes, my editor does seem to be able to pick up the `Literal` type ... :man_facepalming:

tidy3d/components/simulation.py
```diff
@@ -1329,7 +1329,12 @@ def snap_to_grid(geom: Geometry, axis: Axis) -> Geometry:
         # Convert lumped elements into structures
         lumped_structures = []
         for lumped_element in self.lumped_elements:
-            lumped_structures.append(lumped_element.to_structure(self.grid))
+            if isinstance(lumped_element, LinearLumpedElement):
+                PEC_connection = lumped_element.to_PEC_connections(self.grid)
+                if PEC_connection is not None:
+                    lumped_structures.append(PEC_connection)
+            lumped_structure = lumped_element.to_structure(self.grid)
+            lumped_structures.append(lumped_structure)
```
weiliangjin2021: Minor: probably better to shift the logic to lumped_element method, as this part doesn't depend on Simulation class. E.g. you can add `to_structures` method to LumpedElement.

weiliangjin2021 (approved):

QimingFlex (approved):

dmarek-flex:
Got a basic structure question for @daquinteroflex @tylerflex @yaugenst-flex . You will notice in this PR I added a `tools/microwave` directory where I plan on adding small standalone type calculations for microwave specific applications. 

For example, I plan on moving things like the path integrals from the microwave plugin to this folder in the future. The reason being is that I find that I will need to use these calculations when adding basic components like Voltage or Current monitors, so making those tools a part of a plugin doesn't really make sense to me.

daquinteroflex:
~So I can't actually respond to this publicly so I've tagged the private issue to discuss this~

~One approach is simply stick it in the plugins (for now) till we begin this migration effort. This is my preference actually @dmarek-flex if that sounds ok. We can do the migration effort together in another PR when the new structure is more defined if that sounds ok? What's your timeline and we can plan to do it for this since Charge is wrapping up?~
Sorry I should have edited this comment after chatting with Damian.

momchil-flex:
> One approach is simply stick it in the plugins (for now) till we begin this migration effort. This is my preference actually @dmarek-flex if that sounds ok. We can do the migration effort together in another PR when the new structure is more defined if that sounds ok? What's your timeline and we can plan to do it for this since Charge is wrapping up?

I think so far we have 100% avoided importing in the main code (`components`) anything from the plugins. I think we shouldn't break this even temporarily. I'm more in favor of the `tools` folder as a temporary transition.

daquinteroflex:
Sorry I should have edited that comment after a slack chat with Damian.

So actually after discussing with Damian last night, things are more fleshed out on something that also aligns with the CHARGE structure: https://github.com/flexcompute/tidy3d-core/issues/862#issuecomment-2604123033 . Just waiting to see if this works out for Damian but I think it should?

yaugenst-flex:
Agree with @momchil-flex that if the `components` depend on anything, then it shouldn't go in plugins. For autograd, there are some custom functions that the `components` rely on, so these are defined `tidy3d.components.autograd.functions`. Everything that is not core functionality lives in `tidy3d.plugins.autograd`, but these extras _also_ import the "core" functions, so from a user perspective everything is available in the same namespace (`tidy3d.plugins.autograd` in this case).

daquinteroflex:
Structure looks good now! I'd probably call `functions/` -> `utils/` but this is more of a personal preference of common naming for this type of thing.

tidy3d/components/lumped_element.py
```diff
@@ -884,11 +905,84 @@ class LinearLumpedElement(RectangularLumpedElement):
...
+    dist_type: LumpedElementDistributeType = pd.Field(
+        "on",
+        title="Distribute Type",
+        description="When set to ``on``, the network portion of the lumped element is distributed "
+        "across the entirety of the lumped element's bounding box. When set to ``off``, the network "
+        "portion of the lumped element is restricted to one cell and PEC connections are used to "
+        "connect the network cell to the edges of the lumped element. A third option exists "
+        "``laterally_only``, where the network portion is only distributed along the lateral axis of "
+        "the lumped element.",
+    )
```
daquinteroflex: Would this benefit from the images you've put in the PR as part of the 
```
Notes 
-----
```

tidy3d/components/lumped_element.py
daquinteroflex, line 1: Stupid question. Is lumped element mostly microwave specific and should it be under `microwave/`? What would a photonic application look like out of curiosity?

tidy3d/components/microwave/formulas/circuit_parameters.py
```diff
@@ -0,0 +1,173 @@
...
+
+References
+----------
+[1]     F. E. Terman, Radio engineers' handbook, McGraw-Hill Book Company, Inc., 1943.
+
+[2]     E. B. Rosa, The self and mutal inductances of linear conductors, US Department of
+        Commerce and Labor, Bureau of Standards, 1908.
+
+[3]     Y. Y. Iossel, E. S. Kochanov, and M. G. Strunskii, “The Calculation of Electrical Capacitance,”
+        Foreign Technology Division Air Force Systems Command U.S. Air Force, 1971.
```
daquinteroflex, line 11: Also stupid question, have you checked how this renders in the docs? Should this be under the function description instead with the [corresponding notation](https://docutils.sourceforge.io/docs/user/rst/quickref.html#citations) `.. [1] etc. `?

tidy3d/components/microwave/formulas/circuit_parameters.py
```diff
@@ -0,0 +1,173 @@
...
+from ...geometry.base import Geometry
+from ...types import Axis
+
+
+def inductance_straight_rectangular_wire(
+    size: tuple[float, float, float], current_axis: Axis
+) -> float:
+    """Computes the self-inductance of a finite length of wire with a rectangular cross section.
+    Equation 26 [1] and Equation 21 [2].
+
```
daquinteroflex, line 26: It'd be nice if you can add a `.. math:: ` equation description to all of these

tidy3d/plugins/smatrix/ports/rectangular_lumped.py
```diff
@@ -15,14 +15,21 @@
 )
 from ....components.geometry.utils_2d import increment_float
 from ....components.grid.grid import Grid, YeeGrid
-from ....components.lumped_element import LumpedResistor
+from ....components.lumped_element import (
+    LinearLumpedElement,
+    LumpedResistor,
+    RLCNetwork,
```
daquinteroflex, line 21: Wonder if eventually it'll make sense to move the non-lumped electrical definitions to the `spice` directory in https://github.com/flexcompute/tidy3d/tree/marc/dd-devsim/tidy3d/components/spice

tidy3d/plugins/smatrix/ports/rectangular_lumped.py
```diff
@@ -54,7 +61,14 @@ class LumpedPort(AbstractLumpedPort, Box):
         "is always snapped to the grid along its injection axis.",
     )
 
-    _plane_validator = assert_plane()
+    dist_type: LumpedElementDistributeType = pd.Field(
```
daquinteroflex: Wonder if there's a more descriptive shorthand?

tidy3d/plugins/smatrix/ports/rectangular_lumped.py
```diff
@@ -54,7 +61,14 @@ class LumpedPort(AbstractLumpedPort, Box):
...
+        "on",
+        title="Distribute Type",
+        description="Optional field that is passed directly to the :class:`.LinearLumpedElement` used to model the port's load. "
+        "In most cases should be left ``on``.",
+    )
```
daquinteroflex: Also could this include the documentation you've added in the PR description?

tidy3d/plugins/smatrix/ports/rectangular_lumped.py
```diff
@@ -109,14 +123,17 @@ def to_load(self, snap_center: float = None) -> LumpedResistor:
...
             center[self.injection_axis] = snap_center
-        return LumpedResistor(
+
+        network = RLCNetwork(resistance=np.real(self.impedance))
+        return LinearLumpedElement(
             center=center,
             size=self.size,
             num_grid_cells=self.num_grid_cells,
-            resistance=np.real(self.impedance),
+            network=network,
```
daquinteroflex, line 142: I like this `network` generalisation, seems very powerful.

tidy3d/components/lumped_element.py
daquinteroflex: Would it be worth adding that explanation in the docstring?

daquinteroflex (commented):
Haven't gone too deep into the nitty gritty details, but broadly seems reasonable and good implementation!

tidy3d/components/lumped_element.py
weiliangjin2021: but not displaying an extra PEC grid, as that can be confusing. It's more of implementation details

dmarek-flex: @weiliangjin2021 you mean the extra half cell width on each side? agreed it is a bit confusing

weiliangjin2021: yes

tidy3d/components/microwave/formulas/circuit_parameters.py
dmarek-flex: These formulas are only for internal use, so they are not even part of the docs at the moment. I don't plan on users using them directly.

dmarek-flex: In public facing interface I show references like [here](https://docs.flexcompute.com/projects/tidy3d/en/v2.8.0rc1/_modules/tidy3d/components/lumped_element.html#RLCNetwork)

tidy3d/components/lumped_element.py
dmarek-flex: Probably would not be used in a photonic application, but it could pop up in one of these PIC simulations with the RF part. The reason is kind of historic, when I joined there was already the possibility of adding a simple lumped element added to the `Simulation`. Should I move now?

tidy3d/components/microwave/formulas/circuit_parameters.py
dmarek-flex: Do we have a convention for how we document private vs public API? Is that even such a thing?

tidy3d/components/lumped_element.py
dmarek-flex: lumped_dist_type

I'll add this to the lumped element notes

daquinteroflex: Ahh wow looks great! Sorry had to disconnect for a bit due to charge.

tidy3d/components/microwave/formulas/circuit_parameters.py
daquinteroflex: Nice that's great. Hmm, this is an interesting one. Personally, even for future extensibility, it can be useful to have complete documentation. However, this is likely something we'll deal with better with the refactor eventually. If not public maybe whatever you find most reasonable for now then.

tidy3d/components/lumped_element.py
daquinteroflex: So I do wonder if eventually it makes sense to have a `RFSimulation` for this type of specific application what do you think? All coupling is to be handled at the `MultiPhysicsSimulation` level imo in the future.

daquinteroflex: We can deal with that later though, and progressively depreciate.

weiliangjin2021:
@dmarek-flex since 2.8.0rc2 has been released, the new item in CHANGELOG should be moved to unreleased

PR #2136, opened 2024-12-20
Unstructured grid dataset

marc-flex:
@dbochkov-flexcompute I have added the generalization of `UnstructuredGridDataset` in this other PR. In case we don't want to move forward with these changes we can potentially merge #2123 so that we don't block @daquinteroflex

### DIFF
--- a/tidy3d/components/data/unstructured/base.py
+++ b/tidy3d/components/data/unstructured/base.py
@@ -255,8 +255,6 @@ class UnstructuredGridDataset(Dataset, np.lib.mixins.NDArrayOperatorsMixin, ABC)
         """Non-spatial dimensions are corresponding coordinate values of stored data."""
         coord_dict = {dim: self.values.coords[dim].data for dim in self.values.dims}
         _ = coord_dict.pop("index")
-        if "voltage" in coord_dict.keys():
-            _ = coord_dict.pop("voltage")
         return coord_dict
 
     @cached_property
--- a/tidy3d/components/data/unstructured/base.py
+++ b/tidy3d/components/data/unstructured/base.py
@@ -8,7 +8,6 @@ from typing import Tuple, Union
 
 import numpy as np
 import pydantic.v1 as pd
-import xarray as xr
 from xarray import DataArray as XrDataArray
 
 from ....constants import inf
@@ -21,6 +20,7 @@ from ..data_array import (
     DATA_ARRAY_MAP,
     CellDataArray,
     IndexedDataArray,
+    IndexedDataArrayTypes,
     PointDataArray,
 )
 from ..dataset import Dataset
@@ -39,7 +39,7 @@ class UnstructuredGridDataset(Dataset, np.lib.mixins.NDArrayOperatorsMixin, ABC)
         description="Coordinates of points composing the unstructured grid.",
     )
 
-    values: IndexedDataArray = pd.Field(
+    values: IndexedDataArrayTypes = pd.Field(
         ...,
         title="Point Values",
         description="Values stored at the grid points.",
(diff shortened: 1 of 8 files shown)


tidy3d/components/heat_charge/monitor_data.py
```diff
@@ -181,6 +179,22 @@ def warn_no_data(cls, val, values):
...
+        """Issue error if incorrect data type is used"""
+
+        mnt = values.get("monitor")
+
+        if isinstance(val, TetrahedralGridDataset) or isinstance(val, TriangularGridDataset):
+            if isinstance(val.values, IndexVoltageDataArray):
+                raise ValueError(
+                    f"Monitor {mnt} of type 'TemperatureMonitor' cannot be associated with data arrays "
+                    "of type 'IndexVoltageDataArray'."
+                )
```
dbochkov-flexcompute: since we are likely to add others indexed data arrays, probably easier to switch here to something like:
```
if not isinstance(val.values, IndexDataArray):
    raise ValueError(
        f"Unstructured grid dataset given in 'TemperatureData.temperature' must contain only one field as 'IndexedDataArray'."
    )
```

tidy3d/components/data/unstructured/base.py
```diff
@@ -255,6 +255,8 @@ def _values_coords_dict(self):
         """Non-spatial dimensions are corresponding coordinate values of stored data."""
         coord_dict = {dim: self.values.coords[dim].data for dim in self.values.dims}
         _ = coord_dict.pop("index")
+        if "voltage" in coord_dict.keys():
+            _ = coord_dict.pop("voltage")
```
dbochkov-flexcompute: the purpose of `_values_coords_dict` is to return all dimensions other than `index`. Why did you need to remove `voltage` too?

tidy3d/components/heat_charge/monitor_data.py
```diff
@@ -251,6 +265,22 @@ def warn_no_data(cls, val, values):
...
+        """Issue error if incorrect data type is used"""
+
+        mnt = values.get("monitor")
+
+        if isinstance(val, TetrahedralGridDataset) or isinstance(val, TriangularGridDataset):
+            if isinstance(val.values, IndexVoltageDataArray):
+                raise ValueError(
+                    f"Monitor {mnt} of type 'VoltageMonitor' cannot be associated with data arrays "
+                    "of type 'IndexVoltageDataArray'."
+                )
```
dbochkov-flexcompute: same here

tidy3d/components/simulation.py
```diff
@@ -4792,35 +4789,35 @@ def perturbed_mediums_copy(
...
-                new_values = IndexedDataArray(
-                    np.array(data.values.data).flatten(), coords=dict(index=data.values.index.data)
-                )
-                if num_voltages > 1:
-                    raise ValueError(
-                        f"The value entered for '{carrier}' contains field values associated to "
-                        f"{num_voltages} voltage values. Please select one before calling this function. "
-                        "This can be done with 'electron_data.sel(voltage=1)'"
+            if isinstance(data, TriangularGridDataset) or isinstance(data, TetrahedralGridDataset):
+                if "voltage" in data.values.dims:
```
dbochkov-flexcompute: for generality, we could probably switch this to 
```
if data._num_fields > 1:
```

tidy3d/components/data/unstructured/base.py
marc-flex: Yeah, I wasn't sure about this. I'll remove that

tidy3d/components/heat_charge/monitor_data.py
marc-flex: What happens if we have a `SpatialDataArray`? This would fail right? Since it doesn't have a `values` field?

tidy3d/components/simulation.py
marc-flex: Not sure I like this since we're later checking for the additional fields. In this case `voltage`.

daquinteroflex (approved):
Looks good! I'll deal with the ruff issue within the refactor PR if that sounds alright.

daquinteroflex:
@marc-flex I've been able to run tests after fixing the ruff issue and I think a tests patch needs to be made re this PR
 https://github.com/flexcompute/tidy3d/actions/runs/12447074050/job/34750177242
 
Ideally it'd be done onto https://github.com/flexcompute/tidy3d/pull/2137 than `marc/dd-devsim` because the rebase will be hell otherwise.

PR #2137, opened 2024-12-21
:zap: Charge API, `SPICE`, & `MultiPhysicsMedium` consolidation 

daquinteroflex:
Continuation of:
- https://github.com/flexcompute/tidy3d/pull/2106
- https://github.com/flexcompute/tidy3d/pull/2124
- https://github.com/flexcompute/tidy3d/pull/2077

This PR is initially rebased to `marc/dd-devsim` in order to keep track of the changes and be able to more easily see the differences.

Introduce new directories for new functionality:
```
tidy3d/
  components/
    tcad/
       # equivalent heat charge simulations
    spice/
       # electrical/electronic simulation functionality
    material/
      # ideally this should be a medium/ directory, but because that involves renaming, I will introduce this as such.
      # contains the `MultiPhysicsMedium` definition and integration scheme.
```

New functionality:
  - [x] Update `HeatChargeSimulation` accordingly.
  - [ ] Upgrade the existing simulation definition flow in examples and tests.
- [X] Consolidates Marc's great work in a software structure that I think is more extensible and generic.
  - [ ] Separate the internal heat-charge code functionality
- [x] Fix ruff literal issue
- [x] Introduce `MultiPhysicsMedium`
  - [x] Create new directory structure for future integrations.
  - [x] Implement backwards compatibility parameter definitions at the `AbstractMedium` level.
- [x] Introduce API documentation properly
  - [x] Add doctest examples
  - [x] Make sure all api docs have been renamed to the new form
  - [ ] Ideally make them self contained.  
- [ ] Remove code that is not used in this release
   - [ ] Time domain spice etc.

Renaming implemented:
- [X] `HeatChargeBoundaryConditionType` -> `HeatChargeBCTypes`
- [X] `SRHRecombination` -> `ShockleyReedHallRecombination`
- [X] `SlotboomNarrowingModel` -> `SlotboomNarrowingBandGap` 
- [X] `MobilityModelType` -> `MobilityModelTypes`
- [x] `RecombinationModelType` -> `RecombinationModelTypes`  
- [X] `BandgapNarrowingModelType` -> `BandGapModelTypes`
- [X] `VoltageMonitor` -> `SteadyVoltageMonitor`
- [x] `DCCapacitanceDataArray` -> `SteadyCapacitanceVoltageDataArray`
- [x] `VoltageData` -> `SteadyVoltageData`
- [x] `CapacitanceMonitor` -> `SteadyCapacitanceMonitor`
- [x] `HeatChargeMonitorType` -> `HeatChargeMonitorTypes`
- [x] `PotentialData` -> `SteadyPotentialData`
- [x] `FreeCarrierData` -> `SteadyFreeCarrierData`
- [x] `CapacitanceData` -> `SteadyCapacitanceData`
- [x] `HeatChargeMonitorDataType` -> `TCADMonitorDataTypes`
- [x] `DCIVCurveDataArray` -> `SteadyCurrentVoltageDataArray`
- [x] `HeatChargeSimulationType` -> `TCADAnalysisTypes`
- [x] `HeatChargeMonitorType` -> `HeatChargeMonitorTypes` (rather than `TCADMonitorTypes`)
- [x] `HeatChargeSourceType` -> `HeatChargeSourceTypes` (rather than `TCADSourceTypes`)
- [x] `ConductorSpec` -> `ChargeConductorMedium`
- [x] `InsulatorSpec` -> `ChargeInsulatorMedium`
- [x] `SemiConductorSpec` -> `SemiconductorMedium`
- [x] Rename internal `SemiconductorMedium` parameters.


daquinteroflex:
So my goal is to have this PR as a stable version of the refactor. This means that tests pass and the backend can be built accordingly. This will contain the new name and flow. Then in another PR which I will link here, the full `heat_charge` directory refactor will be implemented. However, the goal is that this PR can be used as a stepping stone for the full migration, kind of independently of the remaining frontend changes.

tidy3d/components/material/tcad/heat.py
marc-flex, line 1: For consistency with the charge mediums maybe we could name these `FluidMedium`, `SolidMedium` and then have something like
```
FluidSpec = FluidMedium
SolidSpec = SolidMedium
```
for backwards compatibility?

daquinteroflex: Yep that sounds good

tidy3d/components/spice/analysis/dc.py
```diff
@@ -0,0 +1,61 @@
...
+
+    This class sets some Charge tolerance parameters.
+
+    Example
+    -------
+    >>> import tidy3d as td
+    >>> charge_settings = td.ChargeToleranceSpec(abs_tol=1e8, rel_tol=1e-10, max_iters=30)
+    """
+
+    sources: Optional[MultiStaticTransferSourceDC] = []  # todo accept a single source
```
marc-flex: I'm unsure about having sources here. Every parameter except this one is related to convergence. Maybe we could have a convergence parameters class and so this transfer function only accepts sources and a settings instance?

tidy3d/components/spice/sources/dc.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+
+    step: Optional[pd.PositiveFloat] = pd.Field(
+        1.0,
+        title="Bias step.",
+        description="By default, a solution is computed at 0 bias. "
+        "If a bias different than 0 is requested, DEVSIM will start at 0 and increase bias "
+        "at 'dV' intervals until the required bias is reached. ",
+    )
+    # TODO units
```
marc-flex: So, this step is basically a convergence parameter. Not sure why it's here

tidy3d/components/spice/analysis/dc.py
daquinteroflex: So I'm imagining this as a standard electrical DC simulation (and trying to match it as much as possible to the SPICE standard defintions because that will streamline integration).

In my mind, this class is not a simulation setting: it is a simulation analysis setting. This means that this could be treated as the `independent` variable part of the transfer function, and as such, there is a strong relationship to the DC sweep iteration parameters related to convergence.

I'm following a standard definition for SPICE DC transfer functions:

> The following options control properties pertaining to DC and OP (operating point) analyses
> and algorithms. Since transient analysis (11.1.4) is based on OP, many of the options affect
> transient simulation as well. AC analysis (11.1.3) can be performed only when a stable operat-
> ing point has been found.
> ABSTOL=x resets the absolute current error tolerance of the program. The default value is 1
> pA.
> GMIN=x resets the value of GMIN, the minimum conductance allowed by the program. The
> default value is 1.0e-12.
> GMINSTEPS=x [*] sets the number of Gmin steps to be attempted. If the value is set to zero,
> the standard gmin stepping algorithm is skipped. The standard behavior is that gmin
> stepping is tried before going to the source stepping algorithm.
> 
> 1.3.2 .DC: DC Transfer Function
> General form:
> .dc srcnam vstart vstop vincr [src2 start2 stop2 incr2]
> Examples:
> .dc VIN 0.25 5.0 0.25
> .dc VDS 0 10 .5 VGS 0 5 1
> .dc VCE 0 10 .25 IB 0 10u 1u
> .dc RLoad 1k 2k 100
> .dc TEMP -15 75 5
> The .dc line defines the dc transfer curve source and sweep limits (with capacitors open and
> inductors shorted). srcnam is the name of an independent voltage or current source, a resistor,
> or the circuit temperature. vstart, vstop, and vincr are the starting, final, and incrementing
> values, respectively. The first example causes the value of the voltage source VIN to be swept
> from 0.25 Volts to 5.0 Volts with steps of 0.25 Volt. A second source (src2) may optionally
> be specified with its own associated sweep parameters. In such a case the first source is swept
> over its own range for each value of the second source. This option is useful for obtaining
> semiconductor device output characteristics. See the example on transistor characterization

marc-flex: Another question could be what do you envision having here. If we don't know maybe we could remove this field? 

As mentioned in another comment  `MultiStaticTransferSourceDC`  is another convergence parameter so, for the time being, I'd remove this field, add `step` here, and I'd add a `convergence` field that accepts a convergence class object, i.e.,
```
class TransferFunctionDC(Tidy3dBaseModel):
    convergence_settings: SteadyConvergenceSettings = pd.Field(
        SteadyConvergenceSettings(), title="...", descriptio="..."
    )
```
and in the future we can define more parameters here

daquinteroflex: Like this is how DC simualtions are normally run anyway. This is a good example that shows how it scales https://www.youtube.com/watch?v=6QzTZDm0guM

daquinteroflex: So basically as long as we can extend the data to be compatible with the spice simulation format in https://ngspice.sourceforge.io/docs/ngspice-43-manual.pdf then it'd be fine imo.

tidy3d/components/spice/sources/dc.py
daquinteroflex: So basically this would represent `vincr` in

```
.dc srcnam vstart vstop vincr [src2 start2 stop2 incr2]
```

daquinteroflex: I don't know how this could be better implemented yet be complete

marc-flex: I think the reason why this is difficult to implement is because we're mixing tidy3d setup with spice setup. The best way is to stick to tidy3d and then generate interface with spice

daquinteroflex:
```
source1=Source(parameters=[])
BC(source=source1)
structure(source=Source)?
electrical_analysis_0 = TransferFunction(
    input=[source1, source2],
    output=[CapacitanceMonitor, CurrentMonitor]
) 
# -> DataArray (CapacitanceVoltageDataArray? CurrentVoltageDataArray?)

CapacitanceMonitor(
    values
    ...index(Voltage)
)

electrical_analysis = TimeDomainRF(
    rf_source=[]
    dc_source=[],
    monitor=[CapacitanceMonitor]
)

HeatChargeSimualtion(
    structures=[],
    boundary_conditinos=[],
    electrical_analysis=electrical_analysis_1
)
```

tidy3d/components/spice/analysis/dc.py
```diff
@@ -0,0 +1,62 @@
...
+
+import pydantic.v1 as pd
+
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.components.spice.sources.dc import DCTransferSource, MultiDCTransferSource
+from tidy3d.components.tcad.types import HeatChargeMonitorTypes
+from tidy3d.components.types import annotate_type
+
+
+class TransferFunctionDC(Tidy3dBaseModel):
```
marc-flex: I'm going to add @dbochkov-flexcompute and @momchil-flex here since I believe this is kind of critical (this will affect how the user defines a heat-charge simulation). In summary, the current way of setting up a `HeatChargeSimulation` is as follows
```
sim = HeatChargeSimulation(
   structures=[],
   boundary_spec=[]
   monitors=[],
   ...
   electrical_analysis=analysis_type
) 
```
where the last input is of type `ElectricalAnalysisTypes` which for now is only `TransferFunctionDC`. The idea with this class is to give flexibility to the user with deciding what kind of simulation they want to run. This is: imagine you have defined the simulation object with multiple type of Voltage BCs (e.g., DC and sinusoidal) but initially you're only interested in the steady-state response, then you may define the electrical_analysis to be this `TransferFunctionDC` which will ignore unsteady BCs when the back-end is setup. 

Additionally, this `TransferFunctionDC` is also defining convergence parameters (which can potentially grouped in a class).

What do you think about the working principles of this? Dario made an interesting case for this but I think this is important enough to be decided with the wider team.

tidy3d/components/material/multi_physics.py
```diff
@@ -0,0 +1,26 @@
...
+    # TODO requires backwards compatibility.
+    name: Optional[str]
+    optical: Optional[OpticalMediumTypes]
+    electrical: Optional[ElectricalMediumTypes]
+    heat: Optional[HeatMediumTypes]
+    charge: Optional[ChargeMediumTypes]
+
+    @property
+    def heat_spec(self):
+        return self.heat
```
marc-flex: The reason some of the tests are failing is because `heat_spec()` should return `self.optical.heat_spec` if `self.heat` is `None`.  I'll push this change

tidy3d/components/spice/analysis/dc.py
```diff
@@ -0,0 +1,74 @@
...
+
+    rel_tol: pd.PositiveFloat = pd.Field(
+        default=1e-10,
+        title="Relative tolerance.",
+        description="Relative tolerance used as stop criteria when converging towards a solution.  Should be equivalent to the"
+        "SPICE RELTOL DC transfer parameter. TODO MARC units, TODO check equivalence. TODO what does this "
+        "mean?",
+    )
+
+    max_iters: pd.PositiveInt = pd.Field(
```
marc-flex, line 36: @daquinteroflex I have reverted these names to the short version. Basically, my new approach to this is: if the spellchecker doesn't complain and it's common enough I'll take the shortened version, otherwise, full names

tidy3d/components/spice/analysis/dc.py
```diff
@@ -0,0 +1,75 @@
...
+        default=30,
+        title="Maximum number of iterations.",
+        description="Indicates the maximum number of iterations to be run. "
+        "The solver will stop either when this maximum of iterations is met "
+        "or when the tolerance criteria has been met. Should be equivalent to the ngspice ITL1 parameter. TODO units, "
+        "TODO devsim ngspice equivalence",
+    )
+
+
+class SteadyDCAnalysis(Tidy3dBaseModel):
```
marc-flex: @daquinteroflex After some thought and reading through the ngspice manual I decided to change this to `SteadyDCAnalysis` (from `TransferFunctionDC`). The main reason for this is that, while reading the ngspice manual, it appeared that "transfer" is mostly used in relation to output whereas they're still calling it a DC analysis when talking about analysis type. So my suggestion would be, if you feel like changing it, to change the data array names that we have for the I-V and C-V curves. These could be called something like `CapacitanceDCTransferFunction` or something like that.

As always, we can discuss further with the broader team, including @dbochkov-flexcompute @momchil-flex and prash

PR #2144, opened 2025-01-07
Further charge docs

daquinteroflex:


tidy3d/components/material/tcad/charge.py
```diff
@@ -99,139 +99,170 @@ class SemiconductorMedium(AbstractChargeMedium):
...
+            \\end{equation}
 
-        .. math::
 
-           \\begin{equation}
-               \\nabla \\cdot \\mathbf{J_n} - q R = q \\frac{\\partial n}{\\partial t}
-           \\end{equation}
+    In this solver, we assume Boltzmann statistics. The electron and hole densities, :math:`n` and :math:`p`,
+    can be calculated from the conduction/valence bands and quasi-Fermi energy levels. An isothermal system is at a
+    constant temperature, and in our current implementation, the temperature is at :math:`T=300K`.
```
marc-flex: I would replace `In this solver, we assume Boltzmann statistics.` by `Currently, only Boltzmann statistics are supported when solving the Drift-Diffusion equations for semiconductors.` just to make sure people understand we can/will remove this constraint in the future. 

Also, n and p aren't computed from Fermi levels. Fermi-levels are computed once n and p are known. (n and p are our independent variables)

tidy3d/components/material/tcad/charge.py
```diff
@@ -99,139 +99,170 @@ class SemiconductorMedium(AbstractChargeMedium):
...
 
-        .. math::
+    .. math::
 
-           \\begin{equation}
-               -\\nabla \\cdot \\mathbf{J_p} - q R = q \\frac{\\partial p}{\\partial t}
-           \\end{equation}
+        \\begin{equation}
+            n = N_c \\exp\\left( \\frac{E_{Fn} - E_C}{k_B T} \\right)
+        \\end{equation}
```
marc-flex: This equation is valid in Boltzmann statistics. Maybe we should make sure to explicitly say something along the lines of `Since we're considering Boltzmann statistics (good approximation in non-degenerate semiconductors) the following relations apply`

tidy3d/components/material/tcad/charge.py
```diff
@@ -99,139 +99,170 @@ class SemiconductorMedium(AbstractChargeMedium):
...
+
+    .. math::
+
+        \\begin{equation}
+            q \\frac{\\partial p}{\\partial t} = -\\nabla \\cdot \\mathbf{J_p} - qR
+        \\end{equation}
+
+
+    The electron and hole current density :math:`\\mathbf{J_n}` & :math:`\\mathbf{J_p}` is solved for steady
+    temperature conditions as formalised in [1]_.
```
marc-flex: I feel like this would be better explained as you had it before. i.e.,
```
In semiconductor simulations we solve the equations
- Gauss equations
- Electron continuity equation
- Hole continuity equations

Where the electron and hole fluxes for iso-thermal cases are defined as 
Jn = ...
Jp = ...
where it has been assumed that:
1. Only the static electric potential field contributes to the effective electric field
2. The material is non-degenerate (gamma_n = gamma_p = 1)
```

tidy3d/components/spice/sources/dc.py
```diff
@@ -30,3 +30,8 @@ class DCVoltageSource(Tidy3dBaseModel):
     name: Optional[str]
     voltage: Union[pd.FiniteFloat, list[pd.FiniteFloat]] = pd.Field(title="Voltage")
     # units: Union[VOLT, AMP] = VOLT
+
+
+class DCCurrentSource(Tidy3dBaseModel):
+    name: Optional[str]
+    current: Union[pd.FiniteFloat, list[pd.FiniteFloat]] = pd.Field(title="Voltage")
```
marc-flex: This is for the CONDUCTION case, right?

tidy3d/components/tcad/boundary/charge.py
```diff
@@ -34,13 +35,14 @@ class CurrentBC(HeatChargeBC):
...
+    >>> current_source = td.DCCurrentSource(current=1)
+    >>> current_bc = CurrentBC(source=current_source)
     """
 
-    current_density: pd.FiniteFloat = pd.Field(
-        title="Current density",
-        description="Current density.",
+    source: CurrentSourceTypes = pd.Field(
+        title="Current Source",
+        description="A current source",
```
marc-flex, line 51: I like this way of the defining the BC. It's a bit more cumbersome but very generic, in my opinion.

tidy3d/components/tcad/mobility.py
```diff
@@ -4,16 +4,21 @@
...
 
+    Notes
+    -----
+        This class defines the parameters for the mobility model :math:`\\mu` of Caughey and Thomas [1]_ as defined below:
+
+        .. math::
+
+            \\mu = \\frac{\\mu_{max} - \\mu_{min}}{1 + \\left(N/N_{ref}\\right)^z} + \\mu_{min}
+
+    TODO temperature model.
```
marc-flex: We should probably specify that high (electric) field effects are not being yet considered but will be included in future versions?

Temperature effects are taken into account by specifying `mu_max(used)=mu_max*pow(T/Tref, thermal_exponent)` where the thermal exponents are specified below for each component

marc-flex (commented):
It's coming together quite nicely I'd say. Thanks Dario!

We can speak further. I can take over some of the comments if you'd like to split the work

tidy3d/components/material/tcad/charge.py
daquinteroflex: So wait it's unclear to me, you say `n` and `p` are our independent variables but we only define `N_c` and `N_v` really?

tidy3d/components/tcad/boundary/charge.py
daquinteroflex: So it's trying to standardise the SPICE integration.

daquinteroflex: So I'm still debating the concept of this as a "boundary" or a "source". We discussed it adds energy to the system. So it behaves more like an electrical node than anything else? I am curious whether this "boundary" implementation will nicely scale? Unclear to me.

tidy3d/components/spice/sources/dc.py
daquinteroflex: Yeah standardising

tidy3d/components/material/tcad/charge.py
daquinteroflex: Feel free to edit as you like! The equations I used before are in the other PR I understand, I'll just finish adding the example references for the other classes missing and wait till testing in deployment

tidy3d/components/material/tcad/charge.py
marc-flex: So Nc and Nv are material properties. But you don't know either the Fermi level nor n, p. You solve your transport equations i.e., electron and hole continuity equations (which couple with Gauss' eq) to get psi, n and p. Then, you can compute your Fermi level. In other words, the Fermi level is a result of solving the equations.

tidy3d/components/material/tcad/charge.py
marc-flex: OK. Let me take this on. Since I just updated mobility docs, maybe you can check the rest of the equations/models (which I think are OK for the most part)

tidy3d/components/tcad/boundary/charge.py
```diff
@@ -4,22 +4,28 @@
 
 import pydantic.v1 as pd
 
-from tidy3d.components.spice.sources.types import VoltageSourceTypes
+from tidy3d.components.spice.sources.types import CurrentSourceTypes, VoltageSourceTypes
 from tidy3d.components.tcad.boundary.abstract import HeatChargeBC
 from tidy3d.constants import CURRENT_DENSITY, VOLT
 
 
 class VoltageBC(HeatChargeBC):
```
daquinteroflex, line 12: @marc-flex I was doing the docs here, and realised that shouldn't this be called `SteadyVoltageBC`? Also https://www.notion.so/flexcompute/Analysis-Specification-Debate-175db96dc14e806a9533e64814d3eb7c

tidy3d/components/tcad/boundary/charge.py
```diff
@@ -30,25 +36,28 @@ class VoltageBC(HeatChargeBC):
 
 
 class CurrentBC(HeatChargeBC):
```
daquinteroflex, line 38: Should this be `SteadyCurrentDensityBC`? Or can we define it as the total current in a boundary `SteadyCurrentBC`? Just thinking of units here.

tidy3d/components/tcad/boundary/charge.py
marc-flex: Not necessarily. We call it voltage BC and then the source can be anything. Currently we have `SteadyVoltageSource` (or something like that) but it could be any other type, like, e.g., `SinusoidalVoltageSource` or whatever. At least this is how I understood it when we spoke about this.

tidy3d/components/tcad/boundary/charge.py
marc-flex: I think the same logic as that of the `VoltageBC` applies here

tidy3d/components/tcad/boundary/charge.py
```diff
@@ -4,22 +4,28 @@
...
-from tidy3d.components.spice.sources.types import VoltageSourceTypes
+from tidy3d.components.spice.sources.types import CurrentSourceTypes, VoltageSourceTypes
 from tidy3d.components.tcad.boundary.abstract import HeatChargeBC
 from tidy3d.constants import CURRENT_DENSITY, VOLT
 
 
 class VoltageBC(HeatChargeBC):
-    """Electric potential (voltage) boundary condition.
+    """
+    Constant electric potential (voltage) :math:`= \\text{V}` boundary condition.
```
marc-flex, line 14: This doesn't need to be constant. Eventually we could support  `VoltageSourceTypes` of many different types.

tidy3d/components/tcad/boundary/charge.py
```diff
@@ -4,22 +4,28 @@
...
+        In this case, a solution for each of these voltages will
+        be computed. TODO DEPRECIATE?
 
     Example
     -------
-    >>> bc1 = VoltageBC(source=2)
-    >>> bc2 = VoltageBC(source=[-1, 0, 1])
+    >>> import tidy3d as td
+    >>> voltage_source = td.DCVoltageSource(voltage=[-1, 0, 1])
+    >>> voltage_bc = td.VoltageBC(source=voltage_source)
```
marc-flex: This is already in `tidy3d/components/spice/sources/dc.py` if I'm not mistaken. I think this should be removed from here (maybe leave the example with a single voltage?

tidy3d/components/tcad/boundary/charge.py
```diff
@@ -30,25 +36,28 @@ class VoltageBC(HeatChargeBC):
...
     )
+    # TODO translation between currentsource amps and currentdensity, why not amps here?
 
 
 class InsulatingBC(HeatChargeBC):
     """Insulation boundary condition.
-    Ensures electric fields as well as the surface recombination current density
-    are set to zero.
+    Ensures electric fields :math:`\\mathbf{E} = - \\nabla \\psi  = 0` as well as the surface recombination current
+    density :math:`J_s = 0` are set to zero.
```
marc-flex: In reality grad(psi)·n = 0 and J·n=0 where is the normal to the domain

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -186,13 +204,12 @@ class HeatChargeSimulation(AbstractSimulation):
...
 
-    Warnings
-    --------
+    Note
+    ----
     There are some current limitation of the CHARGE solver:
 
     * The charge solver is currently limited to isothermal cases with T=300K.
     * Boltzmann statistics are assumed throughout (no degeneracy effects considered).
-
```
marc-flex: we can remove this from here since it's already explained in the semiconductor class

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -110,6 +110,25 @@ class HeatChargeSimulation(AbstractSimulation):
         determines the required computation physics according to the simulation scene definition.
         This is implemented in this way due to the strong multi-physics coupling.
```
marc-flex, line 112: The drift-diffusion equations are also solved using the FV method (couldn't highlight further up)

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -110,6 +110,25 @@ class HeatChargeSimulation(AbstractSimulation):
...
+
+    The steady-state electrical ``CONDUCTION`` equation depends on the electric conductivity (:math:`\\sigma`)  of a
+    medium, and the electric field (:math:`\\mathbf{E} = -\\nabla(\\psi)`) derived from electrical potential (:math:`\\psi`).
+    Currently, in this type of simulation, no current sources or sinks are supported.
+
+        .. math::
+
+            \\text{div}(\\sigma \\cdot \\nabla(\\psi)) = 0
+
+
```
marc-flex, line 131: After this bit, I'd add a comment like, `For details on what equations are solved in CHARGE simulations, refer to the SemiConductorMedium` class documentation.

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -240,11 +257,31 @@ class HeatChargeSimulation(AbstractSimulation):
         "Each element can be ``0`` (symmetry off) or ``1`` (symmetry on).",
     )
 
-    analysis_spec: AnalysisSpecType = pd.Field(
+    analysis_spec: ChargeAnalysisSpecTypes = pd.Field(
```
marc-flex: I think this is one of the things we should talk about. I think we should keep as it was (`AnalysisSpecType`) since we may want to specify a simulation regardless of BCs, sources etc. i.e., imagine a user has setup a simulation object with different types of BCs/sources/materials but wants to run a `HEAT` simulation only. In this case, we could just set the `analysis_type = HEAT` (or something along this lines).

Maybe I'm thinking that in the future it may be difficult to determine the simulation type solely based on setup

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -240,11 +257,31 @@ class HeatChargeSimulation(AbstractSimulation):
...
+        .. code::
+        
+            analysis(inputs, outputs) -> AnalysisData
+        
+        It operates on the defined output ``monitors`` for a given set of ``inputs`` in order to define specific data classes.
+        that correspond to the coupling.
+        
+        One of the goals of defining this ``analysis`` is that different types of physical domains
+        require different configuration parameters and this enables the proper definition of these terms.
+    """
```
marc-flex: this would still work with `AnalysisSpecType` except it'd be more general (if I'm not missing anything?)

tidy3d/components/tcad/simulation/heat_charge.py
daquinteroflex: Yeah I like this

tidy3d/components/tcad/simulation/heat_charge.py
daquinteroflex: Yeah it's the same?

tidy3d/components/tcad/boundary/charge.py
daquinteroflex: Sure! As long as it's not a pain to extend eventually, but yeah we did talk about it before.

tidy3d/components/tcad/simulation/heat_charge.py
marc-flex: Maybe not. If we run a HEAT simulation we won't output potential monitor even if it's specified...

daquinteroflex: Shall we catch up briefly?

tidy3d/components/tcad/boundary/charge.py
daquinteroflex: Solved by the plan for the future

tidy3d/components/spice/analysis/dc.py
```diff
@@ -47,30 +43,22 @@ class ChargeToleranceSpec(Tidy3dBaseModel):
...
-        title="Outputs",
-    )  # TODO this should be more generic, # TODO this should be a separate generic monitor class.
+    # TODO move.
 
     tolerance_settings: ChargeToleranceSpec = pd.Field(
         default=ChargeToleranceSpec(), title="Tolerance settings"
     )
 
-    dv: pd.PositiveFloat = pd.Field(
+    convergence_dv: pd.PositiveFloat = pd.Field(
```
daquinteroflex, line 57: fyi @marc-flex , sorry I think if we keep the original one people will genuinely get confused.

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -97,7 +81,24 @@ def field_name(self, val: str) -> str:
 
 
 class SteadyFreeCarrierData(HeatChargeMonitorData):
-    """Class that stores free carrier concentration in Charge simulations."""
+    """
+    This data class stores free-carrier concentration in charge simulations.
+
+    I guess this is this?
```
daquinteroflex, line 87: Also let's chat about this in another PR

daquinteroflex (commented):
Going to start rebasing and any furhter changes onto anotehr PR

tidy3d/components/spice/analysis/dc.py
marc-flex: I like this. Good idea!

tidy3d/components/tcad/data/monitor_data/charge.py
marc-flex: Yeah, we don't need that. Free carriers are just n and p

PR #2146, opened 2025-01-08
maint: Bump JAX version for numpy 2 compatibility

yaugenst-flex:
Currently potentially leads to invalid env as per https://github.com/flexcompute/tidy3d/pull/1913#issuecomment-2576595100

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -374,8 +374,8 @@ files = [
 jmespath = ">=0.7.1,<2.0.0"
 python-dateutil = ">=2.1,<3.0.0"
 urllib3 = [
-    {version = ">=1.25.4,<1.27", markers = "python_version < \"3.10\""},
     {version = ">=1.25.4,<2.2.0 || >2.2.0,<3", markers = "python_version >= \"3.10\""},
+    {version = ">=1.25.4,<1.27", markers = "python_version < \"3.10\""},
 ]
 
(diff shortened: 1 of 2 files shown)


daquinteroflex (approved):
We just have to decide if we need to do a new release

yaugenst-flex:
@momchil-flex adding you as a reviewer because of the release question. this is going into `develop` right now

momchil-flex:
@dmarek-flex was the issue restricted to trying to use our jax plugin, or was it somehow manifesting even without that?

If restricted to the jax plugin we're probably fine... If anyone complains in the meantime we can tell them to upgrade jax.

dmarek-flex:
> @dmarek-flex was the issue restricted to trying to use our jax plugin, or was it somehow manifesting even without that?

For me the issue manifested when running the [quickstart notebook](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/Autograd0Quickstart.html), since optax uses jax.

dmarek-flex:
Everything seems to be working as expected! In case it matters I found that on `develop` numpy>=2 will be installed in a fresh environment, while on `pre/2.8` numpy <2 is installed. I think that is what happened to my environment, switching between branches and running the poetry install commands updated my numpy version.

dmarek-flex (approved):

Issue #2141, opened 2025-01-02
Possibly functionalize the plotting functions?


daquinteroflex:
Break this up:

- [ ] Restructuring the plotting functions into `viz`
- [ ] On the fly import of matplotlib
- [ ] Eventually restructure the same functions to be composed using the  `(fig, axs)` mechanism.
- [ ] Possibly introduce standard `object.plot` functions using the `viz` module as an internal function call onto `self`.

Based on a conversation I had with Marc and my own personal experience:

Maybe it'd be nicer in terms of extensibility if we had purely functional plotting? Something on the lines of:
```
fig, axs = my_plotting_function(fig, axs, ...)
fig, axs = my_second_plotting_function(fig, axs, ...)
```

This can make it easier to represent information as just a set of operations between fig, axs and kwargs. We can move all the plotting functionality outside the classmethods and maybe make it more reusable? This way we avoid 

```
def current_plotting_function():
    fig, ax = self.another_function(...)
    ax.do_something()
    self.another_function_in_this_class(fig, ax) # can mutate the fig, ax object directly and sometimes not accessible at a higher level
    return fig, axs
```

momchil-flex:
I'm in favor of this, only curious about the exact scope especially from user perspective. If it's about moving all plotting functionality in a viz module while still keeping the same user-facing methods, then I think it sounds great. It's also in line with what we discussed about "properly" eliminating `matplotlib` as a core dependence.

daquinteroflex:
Yeah that'd be great actually, and change the functional definitions accordingly when we do this change.

```
my_pseudo_static_classes/
viz/
   all_my_internal_functions/
```

and then refactor those functions in that module to be used in:

```
class MyClass:

   my_plotting_function -> (fig, ax) = viz.my_function() # just like before but also 
```

In principle, for backwards compatibility we can make `my_plotting_function -> (fig, ax)` return the same data as before with the new structure.

momchil-flex:
Yeah, what I like about this apart from backwards compatibility is that otherwise it seems to me that we have two usage options, either:

```
viz.plot(object: Union[Simulation, Source, Monitor, ...])
# ugly-ish internal dispatch based on ``isinstance(object)``
```

or

```
viz.plot_simulation(simulation)
viz.plot_source(source)
viz.plot_monitor(monitor)
# a bit annoying to use
```

and both seem inferior to just `object.plot()`.

tylerflex:
[...]

PR #2147, opened 2025-01-08
feat[frontend]: implementation and testing for VisualizationSpec

groberts-flex:
Adding ability to have a visualization specification for a medium that overrides the plotting parameters in the structure plotting code so that structures in the gui can be specified to have fixed colors and not change when new structures are added. For task "VisualizationSpec for Structure / Medium".

Attached below are screenshots from local testing to show the ability to set different face color, edge color, and alpha as well as usage of default plotting parameters when no visualization specification is included in the medium

Magenta, Cyan, 0.1
magenta_cyan_alpha_p1
Red, Blue, 0.5
red_blue_alpha_p5
No visualization spec included in medium
no_viz_spec

### DIFF
--- a/tests/test_components/test_viz.py
+++ b/tests/test_components/test_viz.py
@@ -1,6 +1,7 @@
 """Tests visualization operations."""
 
 import matplotlib.pyplot as plt
+import pydantic.v1 as pd
 import pytest
 import tidy3d as td
 from tidy3d.components.viz import Polygon, set_default_labels_and_title
@@ -100,3 +101,151 @@ def test_set_default_labels_title():
         ax = set_default_labels_and_title(
             axis_labels=axis_labels, axis=2, position=0, ax=ax, plot_length_units="inches"
         )
+
+    plt.close()
+
+
+def test_make_viz_spec():
+    """
+    Tests core visualizaton spec creation.
+    """
+    viz_spec = td.VisualizationSpec(facecolor="red", edgecolor="green", alpha=0.5)
+    viz_spec = td.VisualizationSpec(facecolor="red", alpha=0.5)
+
+
+def test_unallowed_colors():
+    """
+    Tests validator for visualization spec for colors not recognized by matplotlib.
+    """
+    with pytest.raises(pd.ValidationError):
+        _ = td.VisualizationSpec(facecolor="rr", edgecolor="green", alpha=0.5)
+    with pytest.raises(pd.ValidationError):
+        _ = td.VisualizationSpec(facecolor="red", edgecolor="gg", alpha=0.5)
+
+
+def test_unallowed_alpha():
+    """
+    Tests validator for disallowed alpha values.
+    """
+    with pytest.raises(pd.ValidationError):
+        _ = td.VisualizationSpec(facecolor="red", edgecolor="green", alpha=-0.5)
+    with pytest.raises(pd.ValidationError):
+        _ = td.VisualizationSpec(facecolor="red", edgecolor="green", alpha=2.5)
+
+
+def test_plot_from_structure():
+    """
+    Tests visualization spec can be added to medium and structure plotting function can be run.
+    """
+    viz_spec = td.VisualizationSpec(facecolor="blue", edgecolor="pink", alpha=0.5)
+    medium = td.Medium(permittivity=2.25, viz_spec=viz_spec)
+    geometry = td.Box(size=(2, 0, 2))
+
+    structure = td.Structure(geometry=geometry, medium=medium)
+
+    structure.plot(z=0)
+    plt.close()
+
+
+def plot_with_viz_spec(alpha, facecolor, edgecolor=None, use_viz_spec=True):
+    """
+    Helper function for locally testing different visualization specs in structures through
+    structure plotting function.
+    """
+    if edgecolor is None:
+        viz_spec = td.VisualizationSpec(facecolor=facecolor, alpha=alpha)
+    else:
+        viz_spec = td.VisualizationSpec(facecolor=facecolor, edgecolor=edgecolor, alpha=alpha)
+
+    medium = td.Medium(permittivity=2.25)
+    if use_viz_spec:
+        medium = td.Medium(permittivity=2.25, viz_spec=viz_spec)
+
+    geometry = td.Box(size=(2, 4, 2))
+
+    structure = td.Structure(geometry=geometry, medium=medium)
+
+    structure.plot(z=1)
+    plt.show()
+
+
+def plot_with_multi_viz_spec(alphas, facecolors, edgecolors, rng, use_viz_spec=True):
+    """
+    Helper function for plotting simulations with multiple visulation specs via simluation
+    plotting function.
+    """
+    viz_specs = [
+        td.VisualizationSpec(
+            facecolor=facecolors[idx], edgecolor=edgecolors[idx], alpha=alphas[idx]
+        )
+        for idx in range(0, len(alphas))
+    ]
+    media = [td.Medium(permittivity=2.25) for idx in range(0, len(viz_specs))]
+    if use_viz_spec:
+        media = [
+            td.Medium(permittivity=2.25, viz_spec=viz_specs[idx])
+            for idx in range(0, len(viz_specs))
+        ]
+
+    structures = []
+    for idx in range(0, len(viz_specs)):
+        center = tuple(list(rng.uniform(-3, 3, 2)) + [0])
+        size = tuple(rng.uniform(1, 2, 3))
+        box = td.Box(center=center, size=size)
+
+        structures.append(td.Structure(geometry=box, medium=media[idx]))
+
+    sim = td.Simulation(
+        size=(10.0, 10.0, 10.0),
+        run_time=1e-12,
+        structures=structures,
+        grid_spec=td.GridSpec(wavelength=1.0),
+    )
+
+    sim.plot(z=0.0)
+    plt.show()
+
+
+@pytest.mark.skip(reason="Skipping test for CI, but useful for debugging locally with graphics.")
+def test_plot_from_structure_local():
+    """
+    Local test for visualizing output when specifying visualization spec.
+    """
+    plot_with_viz_spec(alpha=0.5, facecolor="red", edgecolor="blue")
+    plot_with_viz_spec(alpha=0.1, facecolor="magenta", edgecolor="cyan")
+    plot_with_viz_spec(alpha=0.9, facecolor="darkgreen", edgecolor="black")
+    plot_with_viz_spec(alpha=0.8, facecolor="brown", edgecolor="deepskyblue")
+    plot_with_viz_spec(alpha=0.2, facecolor="brown", edgecolor="deepskyblue")
+    plot_with_viz_spec(alpha=1.0, facecolor="green")
+    plot_with_viz_spec(alpha=0.75, facecolor="red", edgecolor="blue")
+    plot_with_viz_spec(alpha=0.75, facecolor="red", edgecolor="blue", use_viz_spec=False)
+
+    with pytest.raises(pd.ValidationError):
+        plot_with_viz_spec(alpha=0.5, facecolor="dark green", edgecolor="blue")
+    with pytest.raises(pd.ValidationError):
+        plot_with_viz_spec(alpha=0.5, facecolor="red", edgecolor="ble")
+    with pytest.raises(pd.ValidationError):
+        plot_with_viz_spec(alpha=1.5, facecolor="red", edgecolor="blue")
+    with pytest.raises(pd.ValidationError):
+        plot_with_viz_spec(alpha=-0.5, facecolor="red", edgecolor="blue")
+
+
+@pytest.mark.skip(reason="Skipping test for CI, but useful for debugging locally with graphics.")
+def test_plot_multi_from_structure_local(rng):
+    """
+    Local test for visualizing output when creating multiple structures with variety of
+    visualization specs.
+    """
+    plot_with_multi_viz_spec(
+        alphas=[0.5, 0.75, 0.25, 0.4],
+        facecolors=["red", "green", "blue", "orange"],
+        edgecolors=["black", "cyan", "magenta", "brown"],
+        rng=rng,
+    )
+    plot_with_multi_viz_spec(
+        alphas=[0.5, 0.75, 0.25, 0.4],
+        facecolors=["red", "green", "blue", "orange"],
+        edgecolors=["black", "cyan", "magenta", "brown"],
+        rng=rng,
+        use_viz_spec=False,
+    )
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -51,6 +51,7 @@ from ..viz import (
     ARROW_LENGTH,
     PLOT_BUFFER,
     PlotParams,
+    VisualizationSpec,
     add_ax_if_none,
     arrow_style,
     equal_aspect,
@@ -450,6 +451,7 @@ class Geometry(Tidy3dBaseModel, ABC):
         z: float = None,
         ax: Ax = None,
         plot_length_units: LengthUnit = None,
+        viz_spec: VisualizationSpec = None,
         **patch_kwargs,
     ) -> Ax:
         """Plot geometry cross section at single (x,y,z) coordinate.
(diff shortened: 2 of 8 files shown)


tests/test_components/test_visualization_spec.py
```diff
@@ -0,0 +1,95 @@
...
+import pytest
+import tidy3d as td
+
+
+def test_make_viz_spec():
+    viz_spec = td.VisualizationSpec(facecolor="red", edgecolor="green", alpha=0.5)
+
+
+def test_default_edgecolor():
+    viz_spec = td.VisualizationSpec(facecolor="red", alpha=0.5)
```
yaugenst-flex: I think these can be combined into a single test (if you want to test different permutations of arguments) or can also just be a single test to check if construction works.

tests/test_components/test_visualization_spec.py
```diff
@@ -0,0 +1,95 @@
...
+
+def test_unallowed_alpha():
+    with pytest.raises(pd.ValidationError):
+        _ = td.VisualizationSpec(facecolor="red", edgecolor="green", alpha=-0.5)
+    with pytest.raises(pd.ValidationError):
+        _ = td.VisualizationSpec(facecolor="red", edgecolor="green", alpha=2.5)
+
+
+def test_create_simple_medium():
+    medium = td.Medium(permittivity=2.25)
```
yaugenst-flex: Can be removed, already tested in the medium tests.

tests/test_components/test_visualization_spec.py
```diff
@@ -0,0 +1,95 @@
...
+
+def test_plot_from_structure():
+    viz_spec = td.VisualizationSpec(facecolor="blue", edgecolor="pink", alpha=0.5)
+    medium = td.Medium(permittivity=2.25, viz_spec=viz_spec)
+    geometry = td.Box(size=(2, 0, 2))
+
+    structure = td.Structure(geometry=geometry, medium=medium)
+
+    structure.plot(z=0)
+    plt.close()
```
yaugenst-flex: Since `test_create_medium` and `test_create_structure` are subsets of `test_plot_from_structure` let's only keep that last one.

tests/test_components/test_visualization_spec.py
yaugenst-flex, line 1: Looks good overall. Let's move these tests into `tidy3d/tests/test_components/test_viz.py` though. Also give each test a short docstring.

tidy3d/components/geometry/base.py
```diff
@@ -438,7 +439,13 @@ def _update_from_bounds(self, bounds: Tuple[float, float], axis: Axis) -> Geomet
...
     @add_ax_if_none
     def plot(
-        self, x: float = None, y: float = None, z: float = None, ax: Ax = None, **patch_kwargs
+        self,
+        x: float = None,
+        y: float = None,
+        z: float = None,
+        ax: Ax = None,
+        viz_spec: VisualizationSpec = None,
+        **patch_kwargs,
```
yaugenst-flex: Docstring of this function needs to be updated.

tidy3d/components/geometry/base.py
```diff
@@ -468,6 +475,8 @@ def plot(
         shapes_intersect = self.intersections_plane(x=x, y=y, z=z)
 
         plot_params = self.plot_params.include_kwargs(**patch_kwargs)
+        if viz_spec is not None:
+            plot_params = self.plot_params.override_with_viz_spec(viz_spec)
```
yaugenst-flex: Not completely sure on this one but I think it would make sense if explicitly passed parameters would take precedence over the `VisualizationSpec` defined in the `Medium`. It's a bit icky in the `Geometry` case because you are also explicitly passing `viz_spec` (so the choice is somewhat arbitrary), but when plotting simulations or structures, the `VisualizationSpec` of the `Medium` essentially acts as the default, which users should be able to override via `patch_kwargs`.

tidy3d/components/medium.py
```diff
@@ -637,6 +637,12 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         description="Modulation spec applied on top of the base medium properties.",
     )
 
+    viz_spec: Optional[VisualizationSpec] = pd.Field(
+        None,
+        title="Visualization Specification",
+        description="Plotting specification for visualizing medium in gui.",
```
yaugenst-flex: Not only for `gui` so let's just remove that part.

tests/test_components/test_visualization_spec.py
```diff
@@ -0,0 +1,167 @@
...
+
+# def test_plot_from_structure():
+#     viz_spec = td.VisualizationSpec(facecolor="blue", edgecolor="pink", alpha=0.5)
+#     medium = td.Medium(permittivity=2.25, viz_spec=viz_spec)
+#     geometry = td.Box(size=(2, 0, 2))
+
+#     structure = td.Structure(geometry=geometry, medium=medium)
+
+#     structure.plot(z=0)
+#     plt.close()
```
yaugenst-flex: Why are these all commented out?

groberts-flex:
multiple structures with no viz specs
multi_structure_no_viz_specs

multiple structures with multiple viz specs
multi_structure_with_viz_specs

Right now, the edge colors for the structures do not show up because the scene code used for plotting a simulation is plotting with linewidth=0. I believe it is straightforward to keep extending viz spec to include things like linewidth or other plotting parameters. The current behavior is if the viz spec is present in a medium, it overrides just the plot_params properties that it contains.

tests/test_components/test_visualization_spec.py
```diff
@@ -0,0 +1,167 @@
...
+                source_time=td.GaussianPulse(
+                    freq0=1e14,
+                    fwidth=1e12,
+                ),
+                name="my_dipole",
+            ),
+        ],
+        monitors=[
+            td.FieldMonitor(size=(0, 0, 0), center=(0, 0, 1), freqs=[1e14], name="point"),
+        ],
```
yaugenst-flex: fyi these can just be empty

tidy3d/components/scene.py
```diff
@@ -449,6 +449,9 @@ def _get_structure_plot_params(self, mat_index: int, medium: Medium) -> PlotPara
             facecolor = MEDIUM_CMAP[(mat_index - 1) % len(MEDIUM_CMAP)]
             plot_params = plot_params.copy(update={"facecolor": facecolor})
 
+        if medium.viz_spec is not None:
+            plot_params = plot_params.override_with_viz_spec(medium.viz_spec)
```
yaugenst-flex: Same comment as in `geometry/base` regarding precedence.

tidy3d/components/scene.py
```diff
@@ -1069,6 +1072,9 @@ def _get_structure_eps_plot_params(
             color = min(1, max(color, 0))  # clip in case of custom eps limits
             plot_params = plot_params.copy(update={"facecolor": str(color)})
 
+        if medium.viz_spec is not None:
+            plot_params = plot_params.override_with_viz_spec(medium.viz_spec)
```
yaugenst-flex: Here too.

tidy3d/components/scene.py
```diff
@@ -1282,6 +1289,9 @@ def _get_structure_heat_cond_plot_params(
             if alpha is not None:
                 plot_params = plot_params.copy(update={"alpha": alpha})
 
+        if medium.viz_spec is not None:
+            plot_params = plot_params.override_with_viz_spec(medium.viz_spec)
```
yaugenst-flex: Ditto.

tidy3d/components/viz.py
```diff
@@ -140,6 +145,53 @@ def to_kwargs(self) -> dict:
...
+        "",
+        title="Edge color",
+        description="Color applied to the edges in visualization.",
+    )
+
+    alpha: Optional[float] = pd.Field(
+        1.0,
+        title="Opacity",
+        description="Opacity/alpha value in plotting between 0 and 1.",
+    )
```
yaugenst-flex: Prefer to use a [constrained type](https://docs.pydantic.dev/1.10/usage/types/#constrained-types) here.

tidy3d/components/viz.py
```diff
@@ -140,6 +145,53 @@ def to_kwargs(self) -> dict:
...
+
+    @pd.validator("alpha", always=True)
+    def is_in_range(value: float) -> float:
+        if (value < 0.0) or (value > 1.0):
+            raise pd.ValidationError(f"{value} is not between 0 and 1")
+
+        return value
```
yaugenst-flex: Because then you can get rid of this validator.

yaugenst-flex (changes requested):
Nice thanks @groberts-flex, this is already looking really good. Most of the comments are pretty minor, so this should be good to go after a little bit of iteration!

tidy3d/components/geometry/base.py
groberts-flex: I see what you're saying, that makes sense! Changing this to first add parameters according to the `viz_spec` and then allowing any other parameters through `patch_kwargs` to override them.

tests/test_components/test_visualization_spec.py
groberts-flex: oops, no reason! just had commented out while testing the viz stuff locally

tidy3d/components/scene.py
groberts-flex: here, there isn't a `patch_kwargs` input for a user to override with. However, there are certain plot parameters being set manually depending on different media types. I'm assuming the use case of the visualization spec falls into the else condition where it is a regular medium and selecting a color automatically from the MEDIUM_CMAP structure (this is the use case I believe this was meant to address). I've changed this case so that the visualization spec can override the facecolor choice. I'm not sure if the visualization spec should also be overriding in the case where the medium is a Medium2D instance

tidy3d/components/scene.py
groberts-flex: moved the viz spec override towards the beginning so it can set the alpha if not otherwise specified as argument to this function. It cannot override face color currently because the permittivity is being plotted grayscale

tylerflex (changes requested):
Thanks @groberts-flex , this is high quality work and I think the feature will be very useful and timely for us. I had a couple high level comments but will leave the rest to yannick and dario.

1. could you add a short changelog item to describe the feature?
2. this PR introduces some more coupling between `tidy3d/components` and `tidy3d/viz.py`. Namely the VizSpec is defined in viz.py rather than in `components`. We are working towards making `matplotlib` an optional dependency of tidy3d to make it a more lightweight package. In this [PR](https://github.com/flexcompute/tidy3d/pull/2087) for example we are doing some changes to how viz are done for this goal. I think in the long run we want to also consider plotting as a "wrapper" around the core pydantic model definition. I think this PR accomplishes that more or less because of the `PlotParams.override_with_viz_spec` method. My question (largely for @yaugenst-flex too perhaps) is whether any of this architecture conflicts with our long term plan to decouple plotting from the data structure definition? and if so, do we want to for example move VizSpec into `components`?

tidy3d/components/scene.py
groberts-flex: changed similarly to permittivity plotting

groberts-flex:
Thanks @tylerflex for the comments! There is some explicit coupling to matplotlib in the `VisualizationSpec` class as it uses a function from matplotlib in a validator to ensure the user has specified a recognizable color for plotting. However, I'm sure this can be decoupled especially if there is a standard set of allowed colors to check the user input against. There are likely bigger considerations than just this, however, and I'll defer for now to @yaugenst-flex on the fuller answer to the question.

tidy3d/components/scene.py
yaugenst-flex: right that makes sense. and yeah let's maybe not override `Medium2D` here, so let's restrict the `viz_spec` override only to regular medium. eventually we'll want to move all of these defaults into a config, but that's out of scope for this PR.

tests/test_components/test_viz.py
```diff
@@ -79,3 +81,155 @@ def test_2d_boundary_plot():
...
+    structure.plot(z=1)
+    plt.show()
+
+
+def plot_with_multi_viz_spec(alphas, facecolors, edgecolors, use_viz_spec=True):
+    """
+    Helper function for plotting simulations with multiple visulation specs via simluation
+    plotting function.
+    """
+    np.random.seed(3726423)
```
yaugenst-flex: Prefer to use the `rng` fixture defined [here](https://github.com/flexcompute/tidy3d/blob/78dfeff490a9884e89786933708d6a64de980e62/tests/conftest.py#L27-L30). So this function would need an `rng` argument, as well as the tests that use it. No need to import it though that's then handled by pytest.

tests/test_components/test_viz.py
```diff
@@ -79,3 +81,155 @@ def test_2d_boundary_plot():
...
+        size=(10.0, 10.0, 10.0),
+        run_time=1e-12,
+        structures=structures,
+        grid_spec=td.GridSpec(wavelength=1.0),
+        boundary_spec=td.BoundarySpec(
+            x=td.Boundary.pml(num_layers=20),
+        ),
+        shutoff=1e-6,
+        courant=0.8,
+        subpixel=False,
```
yaugenst-flex: I think all of these parameters are also optional

yaugenst-flex (commented):
Ok nice this is looking good! One last thing I just noticed.. this should be merged into `pre/2.8` and not into `develop`. The plotting import in `viz.py` will need to be lazy then (you'll see). So I think maybe address those two minor comments I left and then apply the changes from your commits to the `pre/2.8` branch instead.

yaugenst-flex:
@tylerflex @groberts-flex regarding where `VisualizationSpec` should live: Yeah this is a bit tricky, actually @groberts-flex had this in a separate component initially. The problem with that is I don't think we should have a separate module just for this change, and it also doesn't really feel like this belongs in `medium.py`, especially if we end up adding this to `Structure` too. In that sense I think `viz.py` is the correct place. As long as the import is done lazily (will happen when this is rebased to `pre/2.8`), then this doesn't conflict at least with the current state of the minimal install.

In terms of long-term restructuring of visualization / repo I think this change is net neutral because this all needs to be revisited anyways (also e.g. for https://github.com/flexcompute/tidy3d/issues/2141).
So in my opinion this is fine as-is for now.

Maybe @daquinteroflex has additional thoughts?

yaugenst-flex:
> There is some explicit coupling to matplotlib in the VisualizationSpec class as it uses a function from matplotlib in a validator to ensure the user has specified a recognizable color for plotting.

Regarding this point, yeah this might be a bit of a problem. The idea behind the minimal install is to be able to validate simulations in a very constrained environment, ideally without matplotlib as a dependency.
We could either:
 1. Skip the color validator if matplotlib is not importable.
 2. Implement the color validation ourselves.
 3. Leave it as-is.

Option 1 is probably not great because it can lead to GUI bugs.
Option 2 seems like a pain, but actually solves the problem.
Option 3 might be ok too because the way that validation works (as far as I understand) is that if it fails in the minimal environment, it will be run on a full environment, which will still work. And since the `viz_spec` is `None` by default, the validation should work in a minimal environment in most cases.

I would be ok with both 2 or 3 for now, not sure how involved option 2 is.

daquinteroflex:
Nice work Greg, look forward to more of your contributions! `VizualizationSpec` as defined looks reasonable and should interface nicely with the GUI and yeah `viz.py` makes sense for now. Maybe in the future it'll be `viz/` after the refactor.

So yeah, the goal is to restructure the way we do plotting in the future so it is more "easier to extend", completely on-the-fly, clear on what's actually happening to the figure (easier to debug), and have it self contained/decoupled from the internal class parameters themselves as possible. In this sense, some of these methods may need to be restructured in the future. (Your help would be most welcome if you're interested eventually!)

So now that we can merge https://github.com/flexcompute/tidy3d/pull/2087, this PR rebase onto  that might identify some changes on the imports as Yannick and Tyler suggest. Re the validator, option 3 looks good if it can be tested too (we would really benefit from testing it eventually in this  https://github.com/flexcompute/tidy3d/pull/2073 but that'd be down the line.

groberts-flex:
Thanks Dario, that all sounds reasonable to me, thanks for the details on where viz stuff is headed! I rebased the changes and settled the merge conflicts in viz.py

prashkh:
@groberts-flex Thanks for implementing this feature so quickly. The way it is implemented at the medium level works great for implementation in PhotonForge. We don't have to change much! Thanks again.

yaugenst-flex (approved):
I'll approve this now, as far as I can see it's ready to go. We might want to add some additional plotting parameters to `VisualizationSpec` down the line though.

Only one tiny thing: Could you change the commit message @groberts-flex to explicitly mention the addition of `VisualizationSpec`?

e-g-melo:
Thanks @groberts-flex! It is a very nice new feature!

From the GUI side, it seems nice. I have no additional comments.

In GUI, users pick up colors from this interface. So, it should work very well.

image

PR #2152, opened 2025-01-09
:arrow_double_up: Data naming updates

daquinteroflex:
- [ ] ~Aliasing on the data classes naming~ For later.
- [x] Full coverage `tidy3d/components/tcad/data/monitor_data/abstract.py               53      1    98%   52`
- [x] Full coverage `tidy3d/components/tcad/data/monitor_data/charge.py                 79     28    65%   77-80, 115, 121-132, 138-148, 154-157, 165-168, 213-221, 225`
- [x] Full coverage `tidy3d/components/tcad/data/monitor_data/heat.py                   42      4    90%   87, 96-99`
- [x] Full coverage `tidy3d/components/tcad/data/sim_data.py                           107     16    85%   102-118, 185, 199, 205, 215, 250-253, 258, 281`
- [x] Full coverage `tidy3d/components/tcad/doping.py                                  113     89    21%   36-55, 157, 163-294`
- [x] Full coverage `tidy3d/components/tcad/monitors/abstract.py                        11      2    82%   36-37`
- [x] Full coverage `tidy3d/components/tcad/simulation/heat_charge.py                  577    100    83%   14-15, 292-293, 350, 481-485, 546, 554, 559, 576-577, 645, 719, 748-750, 757, 770, 835, 847, 931-932, 937, 952-964, 970-972, 1030, 1053-1067, 1074-1076, 1079, 1093, 1099-1107, 1154-1172, 1192, 1310, 1315-1316, 1339, 1373-1374, 1385-1386, 1406-1420, 1432-1439, 1484, 1500, 1507, 1521, 1528, 1535`
- [x] Full coverage `tidy3d/components/tcad/source/abstract.py                          21      1    95%   24`

So not all of these are actually fully covered, but they tend to be edge cases and I think it's already sufficiently high >90 for most of them. We have a separate issue for the final touches https://github.com/flexcompute/tidy3d/issues/2156

### DIFF
--- a/tests/test_components/test_heat_charge.py
+++ b/tests/test_components/test_heat_charge.py
@@ -169,7 +169,22 @@ def monitors():
         center=(0, 0.9, 0), size=(1.6, 0, 3), name="v_empty", unstructured=True, conformal=False
     )
 
-    return [temp_mnt1, temp_mnt2, temp_mnt3, temp_mnt4, volt_mnt1, volt_mnt2, volt_mnt3, volt_mnt4]
+    capacitance_mnt1 = td.SteadyCapacitanceMonitor(size=(1.6, 2, 3), name="cmnt_test")
+
+    free_carrier_mnt1 = td.SteadyFreeCarrierMonitor(size=(1.6, 2, 3), name="carrier_test")
+
+    return [
+        temp_mnt1,
+        temp_mnt2,
+        temp_mnt3,
+        temp_mnt4,
+        volt_mnt1,
+        volt_mnt2,
+        volt_mnt3,
+        volt_mnt4,
+        capacitance_mnt1,
+        free_carrier_mnt1,
+    ]
 
 
 @pytest.fixture(scope="module")
@@ -246,6 +261,96 @@ def conduction_simulation(mediums, structures, boundary_conditions, monitors, gr
 
 
 @pytest.fixture(scope="module")
+def voltage_capacitance_simulation(mediums, structures, boundary_conditions, monitors, grid_specs):
+    """
+    Creates a HeatChargeSimulation that focuses on voltage sweeping (for capacitance).
+    Specifically, we define a voltage BC with multiple voltage values (an array)
+    so that 'SteadyCapacitanceMonitor' can compute capacitance over this array.
+    """
+    # We will define our own VoltageBC with an array of voltages for the sweep
+    voltage_bc_array = td.VoltageBC(
+        source=td.DCVoltageSource(voltage=[0.0, 1.0, 2.0]),
+    )
+
+    # For illustration, we can reuse the insulator structure as background (like conduction).
+    # Suppose we want to set the voltage array at the simulation boundary
+    pl6 = td.HeatChargeBoundarySpec(
+        condition=voltage_bc_array,
+        placement=td.SimulationBoundary(),
+    )
+
+    # We can optionally define a second boundary condition if desired, e.g. an insulating BC:
+    bc_insulating = td.InsulatingBC()
+    pl7 = td.HeatChargeBoundarySpec(
+        condition=bc_insulating,
+        placement=td.StructureBoundary(structure="solid_structure"),
+    )
+
+    # Let’s pick a couple of monitors. We'll definitely include the CapacitanceMonitor
+    # (monitors[8] -> 'cap_mt1') so that we can measure capacitance. We can also include
+    # a potential monitor to see the fields, e.g. monitors[4] -> volt_mnt1 for demonstration.
+    cap_monitor = monitors[8]  # 'capacitance_mnt1'
+    volt_monitor = monitors[4]  # 'volt_mnt1'
+    chosen_monitors = [cap_monitor, volt_monitor]
+
+    # Build a new HeatChargeSimulation
+    voltage_cap_sim = td.HeatChargeSimulation(
+        medium=mediums["insulator_medium"],
+        structures=[structures["insulator_structure"], structures["solid_structure"]],
+        center=(0, 0, 0),
+        size=(2, 2, 2),
+        boundary_spec=[pl6, pl7],
+        grid_spec=grid_specs["uniform"],
+        sources=[],
+        monitors=chosen_monitors,
+    )
+
+    return voltage_cap_sim
+
+
+@pytest.fixture(scope="module")
+def current_voltage_simulation(mediums, structures, boundary_conditions, monitors, grid_specs):
+    """
+    Creates a HeatChargeSimulation for a scenario combining a current BC and a voltage BC.
+    This can be used to measure conduction properties and free carriers with different
+    monitors, e.g. potential monitors and free carrier monitors.
+    """
+    # We'll reuse bc_volt=boundary_conditions[3] and bc_current=boundary_conditions[4]
+    bc_volt = boundary_conditions[3]  # VoltageBC(source=td.DCVoltageSource(voltage=[1]))
+    bc_current = boundary_conditions[4]  # CurrentBC(source=td.DCCurrentSource(current=3e-1))
+
+    # Place the voltage BC at the simulation boundary
+    pl6 = td.HeatChargeBoundarySpec(
+        condition=bc_volt,
+        placement=td.SimulationBoundary(),
+    )
+    # Place the current BC at the boundary of the "insulator_structure" (arbitrary choice here)
+    pl7 = td.HeatChargeBoundarySpec(
+        condition=bc_current,
+        placement=td.StructureBoundary(structure="insulator_structure"),
+    )
+
+    # Pick a voltage monitor and a free carrier monitor
+    # e.g., monitors[5] -> 'volt_mnt2', monitors[9] -> 'free_carrier_mnt1'
+    volt_monitor = monitors[4]
+    free_carrier_monitor = monitors[9]
+    chosen_monitors = [volt_monitor, free_carrier_monitor]
+
+    current_volt_sim = td.HeatChargeSimulation(
+        medium=mediums["insulator_medium"],
+        structures=[structures["insulator_structure"], structures["solid_structure"]],
+        center=(0, 0, 0),
+        size=(2, 2, 2),
+        boundary_spec=[pl6, pl7],
+        grid_spec=grid_specs["uniform"],
+        sources=[],
+        monitors=chosen_monitors,
+    )
+
+    return current_volt_sim
+
+
+@pytest.fixture(scope="module")
 def temperature_monitor_data(monitors):
     """Creates different temperature monitor data."""
     temp_mnt1, temp_mnt2, temp_mnt3, temp_mnt4, *_ = monitors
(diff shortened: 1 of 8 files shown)


daquinteroflex:
@marc-flex what do you think? Before going deeper into the docs wanted to check this was valid broadly.

marc-flex:
:joy: I can't believe that it's going to be me that says I'd prefer a more verbose name :joy:

In devsim, the verbose option is used. Let me see what other solver use

daquinteroflex:
So it's mainly trying to be consistent with the rest of the API like the `SemiconductorMedium` etc. Think it is nicer and if people understand the physics via the docs more clearly since it is closer to the equations?

Mainly I'm concerned `electrons` or `holes` is a bit undefined in terms of ambiguity? Is it electron contentration or something else? Think this is more specific now?

Alternatively down for a longer name as long as it can be more specific exactly what it is mathematically?

daquinteroflex:
Hi Marc,

So sure, let's keep the old naming if we want per the reasons you've suggested privately. However, maybe this is a good case to introduce an alias since that way we maintain consistency with the rest of our shorthand API. It's not so much a thing about long names or short names, but it does look better if it can be very similar throughout the API imo.

marc-flex:
I like the alias solution!

- Users familiar with the equations have `n`/`p`
- Users who aren't have the more verbose/intuitive `electrons`/`holes`

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -84,7 +84,7 @@ class SteadyFreeCarrierData(HeatChargeMonitorData):
     """
     This data class stores free-carrier concentration in charge simulations.
 
-    I guess this is this?
+    TODO MARC this is this right?
 
     .. math::
```
marc-flex: We can remove these equations. Though going forward we may want to add a `BandStructureMonitor` so that users can visualize the band diagram

daquinteroflex: I guess we can just say these correspond to the `n` and `p` from the drift diffusion equations? Trying to mathematically illustrate what this means to the uninitiated - even if this is just for an boltzmann statistics case?

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -83,21 +83,8 @@ def field_name(self, val: str) -> str:
 class SteadyFreeCarrierData(HeatChargeMonitorData):
     """
     This data class stores free-carrier concentration in charge simulations.
```
marc-flex, line 85: Maybe we should delete this line, I think we're repeating below

daquinteroflex:
For reference these are the missing tests coverage:

```
tidy3d/components/tcad/data/monitor_data/abstract.py               53      1    98%   52
tidy3d/components/tcad/data/monitor_data/charge.py                 79     28    65%   77-80, 115, 121-132, 138-148, 154-157, 165-168, 213-221, 225
tidy3d/components/tcad/data/monitor_data/heat.py                   42      4    90%   87, 96-99
tidy3d/components/tcad/data/sim_data.py                           107     16    85%   102-118, 185, 199, 205, 215, 250-253, 258, 281
tidy3d/components/tcad/doping.py                                  113     89    21%   36-55, 157, 163-294
tidy3d/components/tcad/monitors/abstract.py                        11      2    82%   36-37
tidy3d/components/tcad/simulation/heat_charge.py                  577    100    83%   14-15, 292-293, 350, 481-485, 546, 554, 559, 576-577, 645, 719, 748-750, 757, 770, 835, 847, 931-932, 937, 952-964, 970-972, 1030, 1053-1067, 1074-1076, 1079, 1093, 1099-1107, 1154-1172, 1192, 1310, 1315-1316, 1339, 1373-1374, 1385-1386, 1406-1420, 1432-1439, 1484, 1500, 1507, 1521, 1528, 1535
tidy3d/components/tcad/source/abstract.py                          21      1    95%   24
```

tidy3d/components/tcad/data/monitor_data/charge.py
daquinteroflex: So this is the little summary that appears on the toctree so it's fine to repeat data

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -180,7 +169,21 @@ def field_name(self, val: str) -> str:
...
+    Notes
+    -----
+    The small signal-capacitance of electrons :math:`C_n` and holes :math:`C_p` is computed from the charge due to
+     electrons :math:`Q_n` and holes :math:`Q_p` at an applied voltage :math:`V` at a voltage difference
+    :math:`\\Delta V` between two simulations.
+
+    .. math::
+
+        C_{n,p} = \frac{Q_{n,p}(V + \\Delta V) - Q_{n,p}(V)}{\\Delta V}
+
```
marc-flex, line 188: This wording can be a bit confusing/misleading. What about something like this:
`Small signal capacitance is computed only when a steady-state solution has been computed. The computation of the small signal capacitance is done in the following manner

.. math::
        C_{n,p} = \frac{Q_{n,p}(V_{i+1}) - Q_{n,p}(V_i)}{V_{i+1} - V_i}

where V_i is and V_{i+1} are two consecutive voltage bias levels.`

Or something along these lines? What do you think?

daquinteroflex: So I'm confused how we determine V_i and V_{i+1}?

daquinteroflex: is this from the voltages list? Do we validate that this monitor only computes this informtation when there are more than one voltage BCs applied? What happens when there is only one?

marc-flex: Yes, they're the voltages from the list. When only 1 voltage is specified, the monitor will be an empty dict

daquinteroflex:
@marc-flex I'm probably being an idiot, and trying to debug exactly what's the deal here. 

Not super convinced this should fail to create a valid simulation so easily:
```

@pytest.fixture(scope="module")
def simulation_data(
    heat_simulation,
    conduction_simulation,
    temperature_monitor_data,
    voltage_monitor_data,
    capacitance_monitor_data,
    free_carrier_monitor_data,
):
    """Creates 'HeatChargeSimulationData' for both HEAT and CONDUCTION simulations."""
    heat_sim_data = td.HeatChargeSimulationData(
        simulation=heat_simulation,
        data=temperature_monitor_data,
    )

    cond_sim_data = td.HeatChargeSimulationData(
        simulation=conduction_simulation,
        data=voltage_monitor_data,
    )

    voltage_capacitance_sim_data = td.HeatChargeSimulationData(
        simulation=conduction_simulation,
        data=[capacitance_monitor_data, voltage_monitor_data],
    )

    current_voltage_sim_data = td.HeatChargeSimulationData(
        simulation=conduction_simulation,
        data=[voltage_monitor_data, free_carrier_monitor_data],
    )

    return [heat_sim_data, cond_sim_data, voltage_capacitance_sim_data, current_voltage_sim_data]

```

daquinteroflex:
Also Marc do we need to add a validator to check that there's always a `source` in a BC for a given HeatChargeSimulation with a `SteadyDCAnalysis`? Think so

marc-flex:
> Also Marc do we need to add a validator to check that there's always a source in a BC for a given HeatChargeSimulation with a SteadyDCAnalysis? Think so

The charge simulation checker (`check_charge_simulation`) makes sure `VoltageBC`s have been provided. I guess what we could do is make `source` compulsory, right?

Edit: I just updated `source` to be a required input in both `VoltageBC` and `CurrentBC`

tests/test_components/test_heat_charge.py
```diff
@@ -865,3 +1039,219 @@ def test_sim_structure_extent(box_size, log_level, log_capture):
     )
 
     assert_log_level(log_capture, log_level)
+
+
+def test_abstract_doping_box_default_size_and_center():
+    """Test default size and center for AbstractDopingBox."""
+    box = td.ConstantDoping()
+    assert box.size == (1, 1, 1), "Default size should be (1, 1, 1)."
+    assert box.center == (0, 0, 0), "Default center should be (0, 0, 0)."
```
marc-flex, line 1048: Do we need to test default parameters?

tests/test_components/test_heat_charge.py
```diff
@@ -865,3 +1039,219 @@ def test_sim_structure_extent(box_size, log_level, log_capture):
...
+    box_coords = ((-1, -1, -1), (1, 1, 1))
+    box = td.ConstantDoping(box_coords=box_coords)
+    assert box.size == (2, 2, 2), "Size should be calculated based on box_coords."
+    assert box.center == (0, 0, 0), "Center should be calculated based on box_coords."
+
+
+def test_constant_doping_initialization():
+    """Test initialization of ConstantDoping."""
+    box = td.ConstantDoping(concentration=1e18)
+    assert box.concentration == 1e18, "Concentration should be set to 1e18."
```
marc-flex, line 1062: we're also testing that class initialization works? Is this really required to accomplish 100% coverage? It feels a bit over the top

tests/test_components/test_heat_charge.py
```diff
@@ -865,3 +1039,219 @@ def test_sim_structure_extent(box_size, log_level, log_capture):
...
+    box = td.GaussianDoping(ref_con=1e15, concentration=1e18, width=0.1, source="xmin")
+    expected_sigma = np.sqrt(-(0.1**2) / (2 * np.log(1e15 / 1e18)))
+    assert np.isclose(box.sigma, expected_sigma), "Sigma calculation is incorrect."
+
+
+def test_gaussian_doping_get_contrib():
+    """Test get_contrib method in GaussianDoping."""
+    box = td.GaussianDoping(ref_con=1e15, concentration=1e18, width=0.1, source="xmin")
+    coords = {"x": [0], "y": [0], "z": [0]}
+    contrib = box.get_contrib(coords)
```
marc-flex: If you want I can add here an assertion to make sure that the `contrib` has the expected value

tests/test_components/test_heat_charge.py
daquinteroflex: So not trying 100%, but sometimes basic tests catch certain behaviour

tests/test_components/test_heat_charge.py
daquinteroflex: Go for it

tests/test_components/test_heat_charge.py
daquinteroflex: So maybe we want to avoid changing the user values accidentally when they use this in their scripts by default. It can help catching when its a concious effort to change.

marc-flex: That makes sense

tests/test_components/test_heat_charge.py
```diff
@@ -466,10 +627,6 @@ def test_heat_charge_bcs_validation(boundary_conditions):
     with pytest.raises(pd.ValidationError):
         td.VoltageBC(source=td.DCVoltageSource(voltage=[td.inf]))
 
-    # Invalid CurrentBC: infinite current density
-    with pytest.raises(pd.ValidationError):
-        td.CurrentBC(source=td.DCCurrentSource(current=td.inf))
-
```
marc-flex: No need for this?

daquinteroflex:
Sorry yeah should keep thst

PR #2153, opened 2025-01-10
:wrench: Minor frontend issues

daquinteroflex:
- [x] `PointDipole` docs
- [x] ~Pipeline overflow issue possible fix. Sporadic so it is hard to test.~ We've decided to fix this from the backend.
    - [x] ~Introduced constant value in `boto3.config` ~
    - [x] Need to check if necessary for requests~
- [x] Remove `version` from API docs.

So the problem is:
```
WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: flow360studio.s3.us-gov-west-1.amazonaws.com. Connection pool size: 10
```

So the pipeline overflow issue can be related to [this](https://stackoverflow.com/questions/53765366/urllib3-connectionpool-connection-pool-is-full-discarding-connection)

However, it seems to be an environment issue from the local testing so far (and that I've not been able to reproduce it with the current pre/2.8 environment?).


abdelq:
FYI, adding the following
```python
config=botocore.config.Config(
  max_pool_connections=42,  # A sensible default should be picked
)
```
in
https://github.com/flexcompute/tidy3d/blob/53e4ba4be5ea5075aa9230132be10429a9531d50/tidy3d/web/core/s3utils.py#L60-L67 is how we fixed the same warning.

daquinteroflex:
Thanks @abdelq , glad that worked for you!

We're working on fixing this at the server webpoint level as to maximise the amount of connections that we can open without these warnings. I'll update here when this is implemented without needing to change the boto3 configuration at the user api level.

PR #2161, opened 2025-01-14
fix[invdes]: handle pixel size validation for simulations without sources

yaugenst-flex:
Currently `InverseDesign` validation will error if the simulation contains no sources due to the `_check_pixel_size` validator trying to access `sim.wvl_mat_min`, which depends on the sources being present.
This PR skips the validator if the simulation contains no sources and instead warns the user about checking the pixel size. The reason for not erroring is that we probably still want to allow the construction of simulations without sources in the plugin since we also allow it in `td.Simulation` (and it can be useful in some cases).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - NumPy 2.1 compatibility issue where `numpy.float64` values passed to xarray interpolation would raise TypeError.
 - Support for quasi-uniform grid specifications via `QuasiUniformGrid` that subclasses from `GridSpec1d`. The grids are almost uniform, but can adjust locally to the edge of structure bounding boxes, and snapping points.
 - New field `min_steps_per_sim_size` in `AutoGrid` that sets minimal number of grid steps per longest edge length of simulation domain.
+- Validation error in inverse design plugin when simulation has no sources by adding a source existence check before validating pixel size.
 
 ## [2.8.0rc1] - 2024-12-17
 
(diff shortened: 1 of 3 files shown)


tylerflex:
Porting from other discussion: do we want to just fix `sim.wvl_mat_min` to use some default value even if sources are not defined? I think there was a related discussion about the RunTimeSpec not working with no sources present

yaugenst-flex:
@momchil-flex @tylerflex I think it might be best to raise an error in `sim.wvl_mat_min` in case sources are not present. Returning an arbitrary value seems incorrect. This fix would also work for the `invdes` plugin since the error message is propagated. So something along the lines of `Simulation contains no sources, cannot determine wavelength in material` would work fine.

e-g-melo (approved):
Thanks @yaugenst-flex!

I just gave it a try, and it works well. In which cases do you plan to run optimizations without sources? Eigenmode optimizations?

Emerson

yaugenst-flex:
> I just gave it a try, and it works well. In which cases do you plan to run optimizations without sources? Eigenmode optimizations?

Great! Actually it's not as much about running the optimization but just being able to set up optimizations, because a common pattern we use involves doing `base_something.updated_copy(field=new_items)`. It can also be useful when running locally to test the penalties maybe. But if we end up going for the `sim.wvl_mat_min` error then we'll error instead, which should also be fine. In that case we can give a good error message that there are no sources present, which should also be fine for GUI right?

e-g-melo:
> Great! Actually it's not as much about running the optimization but just being able to set up optimizations, because a common pattern we use involves doing `base_something.updated_copy(field=new_items)`. It can also be useful when running locally to test the penalties maybe. But if we end up going for the `sim.wvl_mat_min` error then we'll error instead, which should also be fine. In that case we can give a good error message that there are no sources present, which should also be fine for GUI right?

Got it!

Yes, it should also work for GUI.

momchil-flex (approved):

yaugenst-flex:
Merging now but will follow up with a fix for `sim.wvl_mat_min`

PR #2163, opened 2025-01-14
feat[frontend]: add angle validator to specify expected angle in radians

groberts-flex:
Provides a clearer validation message that the expected units for the `sidewall_angle` field are in radians to explain why the range is between +/- pi/2.

### DIFF
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -1696,8 +1696,6 @@ class Planar(SimplePlaneIntersection, Geometry, ABC):
         "along the ``axis`` direction; "
         "and ``-np.pi/2<sidewall_angle<0`` specifies an expanding cross section "
         "along the ``axis`` direction.",
-        gt=-np.pi / 2,
-        lt=np.pi / 2,
         units=RADIAN,
     )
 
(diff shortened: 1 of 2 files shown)


tidy3d/components/geometry/base.py
```diff
@@ -1712,6 +1710,17 @@ class Planar(SimplePlaneIntersection, Geometry, ABC):
         "``top`` refers to the positive side of the y-axis.",
     )
 
+    @pydantic.validator("sidewall_angle", always=True)
+    def validate_angle(cls, value: float) -> float:
+        lower_bound = -np.pi / 2
+        upper_bound = np.pi / 2
+        if (value <= lower_bound) or (value >= upper_bound):
+            raise ValidationError(
+                f"sidewall angle ({value}) must be between {lower_bound} and {upper_bound} rad"
```
yaugenst-flex: Capitalization and punctuation :)

Other than that, should we consider just writing `between -π/2 and π/2 rad` maybe? Don't think this bound is going to change, and the message would be even clearer.

yaugenst-flex (approved):
Thanks @groberts-flex, looks great! Don't forget the changelog though, even if it's a small change.

daquinteroflex (approved):

tidy3d/components/geometry/base.py
groberts-flex: Thanks for the catch! And yes, I like that idea and updated the string!

PR #2166, opened 2025-01-15
Adding reduced_simulation option to web functions

momchil-flex:
This is a missing piece from the special mode solver web functions which we want to deprecate.

Note that in our current web test mocking there's no good way for me to test if the simulation actually gets reduced since the upload and download are mocked...

### DIFF
--- a/tidy3d/web/api/webapi.py
+++ b/tidy3d/web/api/webapi.py
@@ -9,6 +9,7 @@ import pytz
 from requests import HTTPError
 from rich.progress import Progress
 
+from ...components.medium import AbstractCustomMedium
 from ...components.types import Literal
 from ...exceptions import WebError
 from ...log import get_logging_console, log
@@ -67,6 +68,7 @@ def run(
     worker_group: str = None,
     simulation_type: str = "tidy3d",
     parent_tasks: list[str] = None,
+    reduce_simulation: Literal["auto", True, False] = "auto",
 ) -> SimulationDataType:
     """
     Submits a :class:`.Simulation` to server, starts running, monitors progress, downloads,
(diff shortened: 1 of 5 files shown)


tidy3d/web/api/webapi.py
```diff
@@ -262,6 +272,53 @@ def upload(
...
+        the presence of custom mediums in the simulation.
+
+    Returns
+    -------
+    Simulation
+        The potentially reduced simulation object.
+    """
+
+    """
+    TODO: This only works for the mode solver, which is also why `simulation.simulation.scene` is
```
daquinteroflex, line 295: I see, implementation looks good as far as I can tell!

daquinteroflex (approved):

tidy3d/web/api/webapi.py
```diff
@@ -262,6 +272,53 @@ def upload(
...
+    used below. After refactor to use the new ModeSimulation, it should be possible to put the call
+    to this function outside of the MODE_SOLVER check in the upload function. We could implement
+    dummy `reduced_simulation_copy` methods for the other solvers or also implement reductions
+    there. Note that if we do the latter we may want to also modify the warning below to only
+    happen if there are custom media *and* they extend beyond the simulation domain.
+    """
+
+    if reduce_simulation == "auto":
+        sim_mediums = simulation.simulation.scene.mediums
+        contains_custom = any(isinstance(med, AbstractCustomMedium) for med in sim_mediums)
```
yaugenst-flex, line 305: I guess you could imagine a scenario where a polyslab has a ton of vertices outside of the simulation domain too...

yaugenst-flex (approved):
Looks good to me!

tidy3d/web/api/webapi.py
momchil-flex: Yeah, actually more pertinent may even be a large STL geometry that is completely outside.

PR #2167, opened 2025-01-15
Fix system-dependent floating-point issue in EMEGrid validation

caseyflex:
A validator was failing on certain systems because of the following comparison:
`if rmax > sim_rmax + fp_eps:`

`rmax` is a numpy float and `sim_rmax` is a python float. `fp_eps` is a numpy `float32`.

When `rmax` and `sim_rmax` are both 64-bit `6.45`, the comparison is False, as it should be.
When they are both cast to 32-bit, it is False.
When `sim_rmax` is 32-bit `6.44999980926513671875` and `rmax` is still 64-bit `6.45`, the comparison is True.

I fixed it by rewriting the comparison in the form
`np.float32(rmax) - np.float32(sim_rmax) > fp_eps`.
This ensures that everything is done using float32 datatype, and that the left-hand side is close to zero for safer comparison with fp_eps.

There may be other places in the code where `fp_eps` is used in an unsafe manner.

Thanks @e-g-melo and @tomflexcompute for finding this and helping pin down the issue. @e-g-melo can you see if this fixes your issue?

EDIT:
Instead of system-dependent, this seems to depend only on the numpy version and handling of promotions, see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for quasi-uniform grid specifications via `QuasiUniformGrid` that subclasses from `GridSpec1d`. The grids are almost uniform, but can adjust locally to the edge of structure bounding boxes, and snapping points.
 - New field `min_steps_per_sim_size` in `AutoGrid` that sets minimal number of grid steps per longest edge length of simulation domain.
 - Validation error in inverse design plugin when simulation has no sources by adding a source existence check before validating pixel size.
+- System-dependent floating-point precision issue in EMEGrid validation.
 
 ## [2.8.0rc1] - 2024-12-17
 
--- a/tidy3d/components/eme/grid.py
+++ b/tidy3d/components/eme/grid.py
@@ -231,11 +231,11 @@ class EMEExplicitGrid(EMEGridSpec):
         sim_rmin = center[axis] - size[axis] / 2
         sim_rmax = center[axis] + size[axis] / 2
         if len(self.boundaries) > 0:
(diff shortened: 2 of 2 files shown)


caseyflex:
This was introduced in `v2.7.8` likely due to numpy upgrade changing promotion behavior
https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion


https://github.com/flexcompute/tidy3d/pull/1913
@daquintero just so you are aware

e-g-melo (approved):
Thanks @caseyflex!

It worked for me. Now, the validation passes with no errors.

Emerson

tylerflex:
assigning @yaugenst-flex as he is effectively taking over from me for front end FYI @caseyflex

momchil-flex (approved):
Generally fine by me, only looks kind of bad to have to cast to 32 bit floats and I wonder if there isn't a better way.

caseyflex:
> Generally fine by me, only looks kind of bad to have to cast to 32 bit floats and I wonder if there isn't a better way.

Maybe the explicit casts aren't needed, since it should work whether it casts them implicitly both as 64 bit or 32 bit. @e-g-melo could I get you to check one more time if the new version works?

yaugenst-flex (approved):
The version without the explicit cast looks a bit less robust to me, but if @e-g-melo confirms that it works it's probably ok.

e-g-melo:
> Maybe the explicit casts aren't needed, since it should work whether it casts them implicitly both as 64 bit or 32 bit. @e-g-melo could I get you to check one more time if the new version works?

It worked!

PR #2168, opened 2025-01-15
Runner caching

yaugenst-flex:


yaugenst-flex:
As it turns out, the cached env is around 1gb, so restoring the cache takes almost as much time as just installing the dependencies. So this is not worth it, sadly.

PR #2173, opened 2025-01-16
Adding fp_eps buffer to polyslab 'inside' checks

momchil-flex:
The change to use shapely instead of matplotlib to compute `PolySlab.inside` was causing one backend test to fail due to numerical precision for points that are sitting exactly on the polygon edge. I am not sure if there's a better fix than this, any thoughts?

By the way I don't really understand why in other places we have `buffer(0)` but I checked that this does not fix the test, seems like a non-zero value is needed.

### DIFF
--- a/tidy3d/components/geometry/polyslab.py
+++ b/tidy3d/components/geometry/polyslab.py
@@ -527,7 +527,7 @@ class PolySlab(base.Planar):
 
             # vertical sidewall
             if math.isclose(self.sidewall_angle, 0):
-                face_polygon = shapely.Polygon(self.reference_polygon)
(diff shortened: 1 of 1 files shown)


lucas-flexcompute (approved):
`buffer(0)` is only used to clean-up possibly invalid geometries. By using `fp_eps` you are increasing the area of the original polyslab, but I guess this has no impact on the effective geometry (considering the magnitude of `fp_eps`). So, I think that it looks good.

yaugenst-flex:
@momchil-flex fyi the test fix is merged so if you rebase the tests should pass

weiliangjin2021 (approved):
Right, `buffer(0)` is similar to `make_valid`, but not exactly.

PR #2165, opened 2025-01-15
Test parallelization

yaugenst-flex:
Adds `pytest-xdist` for running tests in parallel and `pytest-env` to set `OMP_NUM_THREADS=1` while running tests. Locally I see a ~3x speedup for running the whole test suite. Some tests are faster with multithreading, so if needed we can try to tweak the resource allocation, although I feel like simple is better.

Waiting on the runners to see how much improvement there is for CI.

PR #2175, opened 2025-01-17
Fix file contention issues in parallel tests

yaugenst-flex:
This is a follow-up to https://github.com/flexcompute/tidy3d/pull/2165

Some test were sporadically failing due to I/O contention, so this fixes that. The PR also:
 - Introduces a `close_matplotlib` fixture that closes all matplotlib figures after testing a module so that these don't hang around for the whole test suite. Gets rid of a few warnings and reduces overall memory consumption.
 - Introduces a `reset_logger` that reset the logger for each test module, since most modules modify logger state.
 - Changes `pytest-xdist` distribution algorithm to "worksteal", which brings the time for the whole test suite further down from 160s to 50s (so ~8x improvement in total since https://github.com/flexcompute/tidy3d/pull/2165). This one should also be significant for CI. Let's see.
 - Removes the duplicate `rng` fixture in the autograd plugin.

This also includes the change to the test workflow we discussed @daquintero

### DIFF
--- a/tests/test_plugins/test_mode_solver.py
+++ b/tests/test_plugins/test_mode_solver.py
@@ -408,7 +408,7 @@ def test_mode_solver_simple(mock_remote_api, local):
 
 
 @responses.activate
-def test_mode_solver_remote_after_local(mock_remote_api):
+def test_mode_solver_remote_after_local(mock_remote_api, tmp_path):
     """Test that running a remote solver after a local one modifies the stored data. This is to
     catch a bug if ``_cached_properties["data"]`` is inadvertently used."""
 
@@ -436,7 +436,9 @@ def test_mode_solver_remote_after_local(mock_remote_api):
         direction="-",
     )
     data_local = ms.data
-    data_remote = msweb.run(ms)
+
+    data_remote = msweb.run(ms, results_file=tmp_path / "ms_remote.hdf5")
+
     assert np.all(data_local.n_eff != data_remote.n_eff)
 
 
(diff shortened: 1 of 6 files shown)


tests/test_plugins/test_mode_solver.py
```diff
@@ -482,7 +484,9 @@ def test_mode_solver_custom_medium(mock_remote_api, local, tmp_path):
             freqs=[freq0],
             direction="+",
         )
-        modes = ms.solve() if local else msweb.run(ms)
+        modes = (
+            ms.solve() if local else msweb.run(ms, results_file=tmp_path / "ms_custom_medium.hdf5")
```
momchil-flex, line 488: What is the takehome message here, we shouldn't have multiple tests writing something to the same file? Is it ok if it's across different test files?

yaugenst-flex: Not sure whether it's ok across different test files, the problem occurs in tests that write to the same object from multiple workers. I think this is more likely to happen with parameterized tests.

yaugenst-flex: What we really need to do is get rid of these global default paths in the tests, I think those are the root cause of the issue, and have each test always use its `tmp_path`. Currently (some of) the tests end up cluttering the base dir.

daquinteroflex: Maybe we can come up with a generic path management solution we can share for the back-end too re test file paths? Something on these lines and we'd just document it in some developers guide.

yaugenst-flex: Isn't that what the built-in `tmp_path` fixture is for? Or do you have something else in mind? I think we just need to do a better job of actually using it :smile:

daquinteroflex (approved):
Thanks Yannick! This is great.

 Mainly I'm thinking should we track this issue of making sure that all our tests paths follow this `tmp_path` convention both on the frontend and backend? Couldn't seem to find one already. If this can work nicely for the backend path management too. Happy to help on this if we want.

yaugenst-flex:
I am tracking general test refactoring but haven't managed to port this to the new projects tracking, just did not get to that yet. So yeah will tag you too. It's a big project :smile:

PR #2178, opened 2025-01-17
Reduce complex tolerance in mode solver

momchil-flex:
Currently, for a waveguide with conductivity than about 1e-8, our mode solver returns `k_eff=0`. This is due to the complex tolerance set to `fp_eps` regardless of requested precision.

I propose we reduce the complex tolerance to 1e-10 for both precisions. This is what I get then:

image

Alternative is to set it based on precision. However, first of all this requires some refactor as it is currently a unique constant, and the `ModeSpec.precision` will need to be passed around. Second, as shown above, keeping it lower than `fp_eps` even for single precision makes even single-precision simulations accurate for some of the scanned range.


weiliangjin2021 (approved):
Looks fine, as the frontend is usually with double-precision, so unlikely to confuse lossless to lossy one.

PR #2180, opened 2025-01-17
:zap: Review fixes

daquinteroflex:
Re https://github.com/flexcompute/tidy3d/pull/2077 review.

- [x] Minor docstring reviews
- [x] Minimal `MultiPhysicsMedium` silicon library.
- [x] Silicon example default everywhere.
- [x] Debugged weird R parameter bug?

### DIFF
--- a/tidy3d/components/material/multi_physics.py
+++ b/tidy3d/components/material/multi_physics.py
@@ -11,19 +11,64 @@ from tidy3d.components.material.solver_types import (
 
 class MultiPhysicsMedium(Tidy3dBaseModel):
     """
-    A multi-physics medium may contain multiple multi-physical properties, defined for each solver medium.
+    Contains multiple multi-physical properties as defined for each solver medium.
 
     Examples
     --------
-    >>> import tidy3d as td
-    >>> SiO2 = td.MultiPhysicsMedium(
-    ... optical=td.Medium(permittivity=3.9),
-    ...  charge=td.ChargeInsulatorMedium(permittivity=3.9), # redefining permittivity
-    ... name="SiO2",
-    ... )
+    For *silica* (:math:`SiO_2`):
+        >>> import tidy3d as td
+        >>> SiO2 = td.MultiPhysicsMedium(
+        ...   optical=td.Medium(permittivity=3.9),
+        ...   charge=td.ChargeInsulatorMedium(permittivity=3.9), # redefining permittivity
+        ...   name="SiO2",
+        ... )
+
+    For a silicon ``MultiPhysicsMedium`` composed of an optical model
+    from the material library and custom charge :class:`SemiconductorMedium`:
+        >>> import tidy3d as td
+        >>> default_multiphysics_Si = td.MultiPhysicsMedium(
+        ...     optical=td.material_library['cSi']['Green2008'],
+        ...     charge=td.SemiconductorMedium(
+        ...         N_c=2.86e19,
+        ...         N_v=3.1e19,
+        ...         E_g=1.11,
+        ...         mobility=td.CaugheyThomasMobility(
+        ...             mu_n_min=52.2,
+        ...             mu_n=1471.0,
+        ...             mu_p_min=44.9,
+        ...             mu_p=470.5,
+        ...             exp_t_mu=-2.33,
+        ...             exp_d_n=0.68,
+        ...             exp_d_p=0.719,
+        ...             ref_N=2.23e17,
+        ...             exp_t_mu_min=-0.57,
+        ...             exp_t_d=2.4,
+        ...             exp_t_d_exp=-0.146,
+        ...         ),
+        ...         R=[
+        ...             td.ShockleyReedHallRecombination(
+        ...                 tau_n=3.3e-6,
+        ...                 tau_p=4e-6
+        ...             ),
+        ...             td.RadiativeRecombination(
+        ...                 r_const=1.6e-14
+        ...             ),
+        ...             td.AugerRecombination(
+        ...                 c_n=2.8e-31,
+        ...                 c_p=9.9e-32
+        ...             ),
+        ...         ],
+        ...         delta_E_g=td.SlotboomBandGapNarrowing(
+        ...             v1=6.92 * 1e-3,
+        ...             n2=1.3e17,
+        ...             c2=0.5,
+        ...         ),
+        ...         N_a=0,
+        ...         N_d=0
+        ...     )
+        ... )
     """
 
-    # TODO requires backwards compatibility.
     name: Optional[str] = None
     optical: Optional[OpticalMediumType] = None
     electrical: Optional[ElectricalMediumType] = None
--- a/tidy3d/components/material/tcad/charge.py
+++ b/tidy3d/components/material/tcad/charge.py
@@ -10,14 +10,9 @@ from tidy3d.components.data.data_array import SpatialDataArray
 from tidy3d.components.medium import AbstractMedium
 from tidy3d.components.tcad.doping import DopingBoxType
 from tidy3d.components.tcad.types import (
-    AugerRecombination,
     BandGapNarrowingModelType,
-    CaugheyThomasMobility,
     MobilityModelType,
-    RadiativeRecombination,
     RecombinationModelType,
-    ShockleyReedHallRecombination,
-    SlotboomBandGapNarrowing,
 )
 from tidy3d.components.types import Union
 from tidy3d.constants import (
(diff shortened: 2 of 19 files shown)


tidy3d/components/material/tcad/charge.py
```diff
@@ -61,6 +60,7 @@ class ChargeInsulatorMedium(AbstractChargeMedium):
     """
     Insulating medium. Conduction simulations will not solve for electric
     potential in a structure that has a medium with this 'charge'.
+    Currently, permittivity is treated as a constant.
```
marc-flex: Shouldn't this comment be also in the rest of the `AbstractChargeMedium` derived classes?

tidy3d/components/spice/analysis/dc.py
```diff
@@ -54,6 +52,6 @@ class SteadyChargeDCAnalysis(Tidy3dBaseModel):
         title="Bias step.",
         description="By default, a solution is computed at 0 bias. If a bias different than "
         "0 is requested through a voltage source, the charge solver will start at 0 and increase bias "
-        "at 'convergence_dv' intervals until the required bias is reached. This is, therefore, a "
+        "at `convergence_dv` intervals until the required bias is reached. This is, therefore, a "
```
marc-flex, line 55: I thought code was highlighted with single quote (') rather than with `

tidy3d/components/tcad/data/sim_data.py
```diff
@@ -328,12 +328,13 @@ def plot_field(
 
 
 class HeatSimulationData(HeatChargeSimulationData):
-    """Wrapper for Heat simulation data. 'HeatSimulationData' is deprecated.
-    Consider using 'HeatChargeSimulationData' instead."""
+    """Wrapper for Heat simulation data. 'HeatSimulationData' is DEPRECIATED.
```
marc-flex: As I understand, the usual word for this is "deprecate" rather then "depreciate". I've searched throughout the code and "deprecated" appears much more frequently but "depreciated" does appear in 11 lines of code. Maybe @momchil-flex can clarify what's the preferred option.

marc-flex (commented):
Thanks @daquinteroflex for being on top of this. I'll finish the rest of the comments on Monday, when I'll probably have Daniil's feedback

tidy3d/components/tcad/data/sim_data.py
tylerflex: should be deprecate. but some people say depreciate :) https://www.grammarly.com/commonly-confused-words/depreciate-vs-deprecate

tidy3d/components/spice/analysis/dc.py
daquinteroflex: So it can follow standard markdown notation. But both forms are correct for `f` strings too.

tidy3d/components/tcad/data/sim_data.py
daquinteroflex: Haha oh yeah, didn't realise, will update to `DEPRECATED`

daquinteroflex:
~For the material library situation default, there are many combinations of a given Si `SemiconductorMedium`. This is due to the many recombination models, etc.~ There's one that makes most sense. I'm only going to create one combination of a `MultiPhysicsMedium(charge=SemiconductorMedium(our single instantiated parameters))` in order to have a minimal instantiation, and the remaining silicon defaults I will make them a docstring example on each class since that way people can simply create the higher level models they want from them. They will have to read the docs to determine the corresponding defaults. Fyi @dbochkov-flexcompute

tidy3d/components/material/multi_physics.py
```diff
@@ -11,19 +11,64 @@
...
+        ...                 tau_p=4e-6
+        ...             ),
+        ...             td.RadiativeRecombination(
+        ...                 r_const=1.6e-14
+        ...             ),
+        ...             td.AugerRecombination(
+        ...                 c_n=2.8e-31,
+        ...                 c_p=9.9e-32
+        ...             ),
+        ...         ]),
```
marc-flex: As far as I know you can pass a list and it's converted to tuple during initialization. I'll try later and remove additional brackets.

tidy3d/components/material/tcad/charge.py
```diff
@@ -32,7 +27,7 @@ class AbstractChargeMedium(AbstractMedium):
 
     conductivity: pd.PositiveFloat = pd.Field(
         1,
```
dbochkov-flexcompute: default conductivity = 1 is kind of random. Should this be conductivity = 0 like our optical mediums? Or maybe even better remove the default value altogether and make the user always set the conductivity value explicitly given it is a dedicated class for conductive materials

dbochkov-flexcompute:
> ~For the material library situation default, there are many combinations of a given Si `SemiconductorMedium`. This is due to the many recombination models, etc.~ There's one that makes most sense. I'm only going to create one combination of a `MultiPhysicsMedium(charge=SemiconductorMedium(our single instantiated parameters))` in order to have a minimal instantiation, and the remaining silicon defaults I will make them a docstring example on each class since that way people can simply create the higher level models they want from them. They will have to read the docs to determine the corresponding defaults. Fyi @dbochkov-flexcompute

looks great!

pyproject.toml
```diff
@@ -128,7 +128,6 @@ dev = [
   "coverage",
   'dill',
   'divparams',
-  'gdspy',
```
momchil-flex: Maybe fine but curious why?

tidy3d/components/spice/sources/dc.py
```diff
@@ -29,7 +29,7 @@
 
 class DCVoltageSource(Tidy3dBaseModel):
     """
-    This represents a DC voltage source in volts. An extended version of a SPICE model of steady-state independent
+    Represents a DC voltage source in volts. An extended version of a SPICE model of steady-state independent
```
momchil-flex: Just `A DC voltage source...`

tidy3d/components/spice/analysis/dc.py
```diff
@@ -9,7 +9,7 @@
 
 class ChargeToleranceSpec(Tidy3dBaseModel):
     """
-    This class sets some charge tolerance parameters relevant to multiple simulation analyis types.
+    Sets some charge tolerance parameters relevant to multiple simulation analysis types.
```
momchil-flex: Similarly just `Charge tolerance parameters...`

tidy3d/components/tcad/bandgap.py
```diff
@@ -7,11 +7,11 @@
 # Band-gap narrowing models
 class SlotboomBandGapNarrowing(Tidy3dBaseModel):
     """
-    This class specifies the parameters for the Slotboom model for band-gap narrowing.
+    Specifies the parameters for the Slotboom model for band-gap narrowing.
```
momchil-flex: `Parameters for the Slotboom...`

tidy3d/components/tcad/generation_recombination.py
```diff
@@ -5,7 +5,7 @@
 
 class AugerRecombination(Tidy3dBaseModel):
     """
-    This class defines the parameters for the Auger recombination model.
+    Defines the parameters for the Auger recombination model.
```
momchil-flex: `Parameters for the Auger...`

tidy3d/components/material/tcad/charge.py
momchil-flex: +1 for making it a required field.

tidy3d/components/tcad/bandgap.py
```diff
@@ -23,23 +23,29 @@ class SlotboomBandGapNarrowing(Tidy3dBaseModel):
...
+        description=r"$V_{1,bgn}$" + f" parameter in {VOLT}",
+        units=VOLT,
     )
 
     n2: pd.PositiveFloat = pd.Field(
-        1.3e17,
-        title="n2 parameter",
-        description="n2 parameter in cm^(-3)",
+        title=r"$N_{2,bgn}$ parameter",
+        description=r"$N_{2,bgn}$ parameter in cm^(-3)",
```
momchil-flex: Shouldn't units be in `units` rather than in the description?

tidy3d/components/spice/sources/dc.py
```diff
@@ -52,7 +52,7 @@ class DCVoltageSource(Tidy3dBaseModel):
 
 class DCCurrentSource(Tidy3dBaseModel):
     """
-    This represents a DC current source in amperes. An extended version of a SPICE model of steady-state independent current source.
+    Represents a DC current source in amperes. An extended version of a SPICE model of steady-state independent current source.
```
momchil-flex: `A DC current source...`

momchil-flex (commented):
Sorry this is getting nit-picky but some of the docstrings still sound unnecessarily verbose to me :)

pyproject.toml
daquinteroflex: Sorry we can remove this from this PR specifically.

Both @yaugenst-flex  and I are running into gdspy new environment problems in Linux (normally I recreate one for each PR so this was a temp fix). We need to do a depreciation PR I believe.

marc-flex:
Thanks everyone for your comments. I have addressed all of them.

I will merge this branch. Any additional comments can be dealt with in #2077

PR #2181, opened 2025-01-17
Allow MeshOverrideStructure to unshadow structures in overlapping region

weiliangjin2021:
Related to @lucas-flexcompute 's comment on https://github.com/flexcompute/tidy3d/pull/2113

When we automatically add mesh override structure that has the same geometry as regular structure, we don't want it to shadow physical structures if the physical structure sets a smaller grid size. 

With `shadow=False`, we drop the logic for checking if the override structure is covered or covering other structures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `VisualizationSpec` that allows `Medium` instances to specify color and transparency plotting attributes that override default ones.
 - `reduce_simulation` argument added to all web functions to allow automatically reducing structures only to the simulation domain, including truncating data in custom media, thus reducing the simulation upload size. Currently only implemented for mode solver simulation types.
+- Support for quasi-uniform grid specifications via `QuasiUniformGrid` that subclasses from `GridSpec1d`. The grids are almost uniform, but can adjust locally to the edge of structure bounding boxes, and snapping points.
+- New field `min_steps_per_sim_size` in `AutoGrid` that sets minimal number of grid steps per longest edge length of simulation domain.
+- New field `shadow` in `MeshOverrideStructure` that sets grid size in overlapping region according to structure list or minimal grid size.
 
 ### Changed
 - `ModeMonitor` and `ModeSolverMonitor` now use the default `td.ModeSpec()` with `num_modes=1` when `mode_spec` is not provided.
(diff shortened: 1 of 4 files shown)


weiliangjin2021:
Here is an example: 
image

There is a high-index box in the simulation domain. When we add a mesh override structure of larger grid size and of the same geometry, previous behavior is that this larger grid size is applied because it completely covers the physical structure, as illustrated in the middle figure.

Now with `shadow=False`, the smallest grid size is applied in this region (rightmost figure).

lucas-flexcompute (approved):
Looks good in general, but just for me to understand: Does the `shadow==False` setting only take effect when the previous structure is **completely** inside the override, or does it work  for partial intersections too? The documentation in the field description seems to imply intersections are supported, but the docs in the test function says otherwise.

weiliangjin2021:
> Looks good in general, but just for me to understand: Does the `shadow==False` setting only take effect when the previous structure is **completely** inside the override, or does it work for partial intersections too? The documentation in the field description seems to imply intersections are supported, but the docs in the test function says otherwise.

For partial intersection, it's already smaller grid size takes effect.

momchil-flex (approved):
Looks good, but just wondering what is the use case exactly? In your example, it is basically going to be a case where I don't really know or want to think about whether the physical structure has high or low index? So I use the override structure to set a minimum step, but if it needs to be even lower than that I allow it?

lucas-flexcompute:
I think the original use case is in RF simulations. We will use layer refinement to set the grid and those are implemented through override structures. Because they are created automatically, I asked Weiliang what would happen in partial and integral intersections. (I ended up sorting my layer refinements by size, so it wouldn't be a problem anyway, but applying the smallest grid step regardless of sorting makes sense for users from other tools, doesn't it?)

momchil-flex:
Yeah, it's just that in the example above, one could just remove the OverrideStructure, but I see the value of not having to manually think about it in some general cases.

weiliangjin2021:
> In your example, it is basically going to be a case where I don't really know or want to think about whether the physical structure has high or low index? So I use the override structure to set a minimum step, but if it needs to be even lower than that I allow it?

That's right. Let's say you have a layered structure, and you want to make sure that there are at least 5 grid points to resolve the layer thickness. The layer consist of many physical structures, so you don't really know grid size from `min_steps_per_wvl` until you processed all the structures where structure overlapping is figured out You might have a very high-index material but it can be entirely covered by other structures so it doesn't appear in the final structure. So it's hard to know if your MeshOverrideStructure is decreasing grid size, or the other way.

PR #2187, opened 2025-01-20
add tests for polyslab slab bounds gradient check

rahul-flex:


yaugenst-flex (approved):
Thanks @rahul-flex looks great!

PR #2189, opened 2025-01-21
feature[design]: allow design plugin to interface with pytorch via Result

groberts-flex:
This PR fixes a small bug in the design plugin wherein the optimization method returns the dimensions in a different order and so labeling with the `self.dims` parameter can end up being incorrect.

It also implements accessor and length functions for the Result class so that it can be used directly as a Dataset in PyTorch. This also may be useful generally.

Marking this as a draft for now to discuss if we want to make these changes to the plugin. The changes are currently needed for the surrogate notebooks draft PR. (https://github.com/flexcompute/tidy3d-notebooks/pull/228)

### DIFF
--- a/tests/test_plugins/test_design.py
+++ b/tests/test_plugins/test_design.py
@@ -782,3 +782,72 @@ def test_genalg_run_count():
 
     # Check that the result and the aux are the same value
     assert all(df["output"] == df["aux_key_0"])
+
+
+@pytest.mark.parametrize(
+    "dims, coords, values, multi_val",
+    [
+        (("a", "b", "c"), ((1, 0, 1), (2, 3, 4), (5, 1, 2)), (1, 2, 3), False),
+        (("a", "b", "c"), ((5, 2, 0), (1, 2, 3), (3, 4, 1)), ((3, 4), (4, 6), (6, 9)), True),
+        (("x", "y", "z", "t"), ((1, 2, 3, 4), (5, 6, 7, 8)), (5, 3), False),
+        (tuple("a"), (tuple("b"), tuple("c")), ("d", "e"), False),
+    ],
+)
+def test_result_accessor_and_len(dims, coords, values, multi_val):
+    """Test the accessor for Result lines up with regular indexing into the coordinates and values
+    and that the length matches that of the coordinates and values. Finally, since we have defined
+    accessors and length, ensure we can cast the Result class to a list and tuple"""
+
+    result = tdd.Result(dims=dims, values=values, coords=coords)
+
+    for idx in range(0, len(values)):
+        coord, value = result[idx]
+
+        assert all(coord == coords[idx])
+
+        if multi_val:
+            assert all(value == values[idx])
+        else:
+            assert value == values[idx]
+
+    assert len(result) == len(coords)
+    assert len(result) == len(values)
+
+    result_list = list(result)
+    result_tuple = tuple(result)
+
+
+def test_dim_coord_alignment():
+    """Test that the dims coming out of the optimization method align with the ordering of the coordinates.
+    The ordering of these dimensions from the optimization can differ from the initial ordering of dimensions
+    input into the DesignSpace."""
+    bay_opt = tdd.MethodBayOpt(initial_iter=5, n_iter=2, seed=1)
+
+    parameters_sweep = [
+        [
+            tdd.ParameterFloat(name="radius", span=(1, 5)),
+            tdd.ParameterFloat(name="alpha", span=(-5, -1)),
+        ],
+        [
+            tdd.ParameterFloat(name="alpha", span=(-5, -1)),
+            tdd.ParameterFloat(name="radius", span=(1, 5)),
+        ],
+    ]
+
+    for parameters in parameters_sweep:
+        design_space = tdd.DesignSpace(
+            parameters=parameters, method=bay_opt, name="bay opt", task_name="bayesian"
+        )
+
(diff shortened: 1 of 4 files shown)


groberts-flex:
converting now to a PR since the notebook PR is out

tidy3d/plugins/design/design.py
```diff
@@ -124,7 +124,7 @@ def _package_run_results(
         fn_args_coords_T = list(map(list, zip(*fn_args_coords)))
 
         return Result(
-            dims=self.dims,
+            dims=tuple(fn_args[0].keys()),
```
yaugenst-flex, line 127: Not too familiar with the design plugin, why is this change necessary?

tidy3d/plugins/design/result.py
```diff
@@ -427,3 +427,18 @@ def add(self, fn_args: Dict[str, float], value: Any) -> Result:
...
+    def __len__(self):
+        return len(self.coords)
+
+    def __getitem__(self, data_index):
+        features = self.coords[data_index]
+        labels = self.values[data_index]
+
+        return features, labels
+
+    def __getitems__(self, data_indices):
```
yaugenst-flex: Is this required? Since it's not a standard python getter, could this class also work without it? I assume pytorch would just automatically batch using `__getitem__` instead?

yaugenst-flex (commented):
Thanks @groberts-flex the accessors definitely are a great addition!

tidy3d/plugins/design/design.py
groberts-flex: Right now, the optimization methods don't necessarily return the values in the same order as they are in the DesignSpace dims attribute (which I believe is just set by the order the user inputs the parameters). I think the BayOpt for example returns with dimensions in alphabetical order. So self.dims is not always the correct key ordering for the result.

tidy3d/plugins/design/result.py
groberts-flex: good point on this! I thought it needed to be implemented, but actually just __getitem__ needs to be implemented. I think the __getitems__ can be used in batch sampling but if not available I think it just implements it using the underlying __getitem__ which is basically what I did here. will test this with the notebook, but think it can be removed.

tidy3d/plugins/design/design.py
yaugenst-flex: I see. Does this mean then that the `dims` of a `Result` don't necessarily align with the `dims` of the `DesignSpace` they were generated from? And if yes, is that intended behavior or should we try to align them?

groberts-flex: Yeah, they don't necessarily align (in fact they weren't aligning in one of the surrogate notebooks and it was causing the data to be interpreted backwards) because of the method outputting them in a different order.

I think it would be ideal if all the parameters stayed in the same order as originally specified (or always were shuffled to the same ordering). Right now, the Result is also packaged up with the optimizer that was used in the method. And a user can query that object to find the parameters associated with the optimal value. Right now, with the current change, the Result returned from here is consistent with the data and the optimizer. But if we rearrange here to match the DesignSpace, we should probably rearrange the optimizer parameters too, which I think is not straightforward since I believe that's coming from an external package.

yaugenst-flex: Alright, ok. Something to keep in mind then, but let's keep it as is for this PR then

tidy3d/plugins/design/result.py
groberts-flex: confirmed that we don't need getitems! also returning these as numpy arrays works better with PyTorch integration (and possibly generally). they come out as tensors from the dataloader now instead of a list of tensors which was causing issues before.

yaugenst-flex:
Thanks @groberts-flex I think this is good to go implementation-wise, but still needs tests and a changelog entry. For the tests, I think the accessor methods can be tested easily (check `len`, indexing, `list/tuple` conversion). For the dimension ordering, maybe identify one (or ideally more than one distinct) cases where this broke previously?

groberts-flex:
tests and changelog entry are in, how does this look for merging @yaugenst-flex?

tests/test_plugins/test_design.py
```diff
@@ -35,6 +35,28 @@
...
+    ),
+    dict(
+        dims=("x", "y", "z", "t"),
+        coords=((1, 2, 3, 4), (5, 6, 7, 8)),
+        values=(5, 3),
+        multi_val=False,
+    ),
+    dict(dims=tuple("a"), coords=(tuple("b"), tuple("c")), values=("d", "e"), multi_val=False),
+]
+
```
yaugenst-flex: since this isn't reused for other tests and the cases are relatively simple, i'd prefer this to be written out in the parametrization directly, i.e.
```python
@pytest.mark.parametrize("dims, coords, values, multi_val", [
    (("a", "b", "c"), ((1, 0, 1), (2, 3, 4), (5, 1, 2)), (1, 2, 3), False),
    # ... other cases
])
def test_result_accessor(dims, coords, values, multi_val):
    # ... test logic
```

this puts the data closer to the test, making it easier to see the test cases when looking at the test, and the function arguments are clearer too

yaugenst-flex (approved):
Thanks @groberts-flex I think this looks good, only a minor comment.

tests/test_plugins/test_design.py
groberts-flex: thanks, this makes sense! much clearer this way

PR #2190, opened 2025-01-21
add antenna gain and efficiency figures of merit

dmarek-flex:
added more fields to `DirectivityData` to allow for retrieving the partial radiation intensities, directivities, and helpers to compute gain given a supplied power. Radiated power is based on flux but now includes the direction of the projection monitor, when it is 2D.

Antenna parameters can be extracted by using the `TerminalComponentModeler`, where the concept of a port is needed to calculate the incident and reflected powers.

For more advanced antennas, we will need the ability to calculate results with a linear combination of ports excited.

Checklist:

- [x] Make gain and efficiency calculations available
- [x] Fix backend tests with new API
- [x] Simplify notebooks by using new API
- [x] Add post processing support for generating far field response using multiple ports with different complex amplitudes
- [x] Add tests and example to notebook demonstrating post processing step for multiple port excitations

Relevant PRs:
 - [Notebook PR](https://github.com/flexcompute/tidy3d-notebooks/pull/244) for how the changes can be used.
 - [~~Backend Test PR~~](https://github.com/flexcompute/tidy3d-core/pull/901) for changed tests.
 - [New backend test PR](https://github.com/flexcompute/testcases/pull/936)

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import warnings
 from abc import ABC
+from math import isclose
 from typing import Any, Callable, Dict, List, Tuple, Union
 
 import autograd.numpy as np
@@ -57,6 +58,7 @@ from ..types import (
     EpsSpecType,
     Literal,
     Numpy,
+    PolarizationBasis,
     Size,
     Symmetry,
     TrackFreq,
@@ -96,6 +98,8 @@ Coords1D = ArrayFloat1D
 # how much to shift the adjoint field source for 0-D axes dimensions
 SHIFT_VALUE_ADJ_FLD_SRC = 1e-5
 AXIAL_RATIO_CAP = 100
+# At this sampling rate, the computed area of a sphere is within ~1% of the true value.
+MIN_ANGULAR_SAMPLES_SPHERE = 10
 
 
 class MonitorData(AbstractMonitorData, ABC):
@@ -118,6 +122,31 @@ class MonitorData(AbstractMonitorData, ABC):
     def normalize(self, source_spectrum_fn: Callable[[float], complex]) -> Dataset:
         """Return copy of self after normalization is applied using source spectrum function."""
         return self.copy()
+
+    def scale_fields_by_freq_array(
+        self, freq_array: FreqDataArray, method: str = None
+    ) -> MonitorData:
+        """Scale fields in :class:`.MonitorData` by an array of values stored in a :class:`.FreqDataArray`.
+
+        Parameters
+        ----------
+        freq_array : FreqDataArray
+            Array containing the scaling factors in the frequency domain.
+        method : str = None
+            Interpolation method to use when selecting frequency values. If None, uses default xarray
+            method. Passed to xarray's sel() method.
+
+        Returns
+        -------
+        :class:`.MonitorData`
+            A new instance of :class:`.MonitorData` with scaled field values.
+        """
+
+        # Reuse the normalize method, so we need the inverse of the scaling amplitude
+        def amplitude_fn(freq: list[float]) -> complex:
+            return 1.0 / freq_array.sel(f=freq, method=method).values
+
+        return self.normalize(amplitude_fn)
 
     def _updated(self, update: Dict) -> MonitorData:
         """Similar to ``updated_copy``, but does not actually copy components, for speed.
@@ -2518,6 +2547,94 @@ class FieldProjectionAngleData(AbstractFieldProjectionData):
         # compute updated fields and their coordinates
         return self.make_renormalized_data(phase, proj_distance)
 
+    @property
+    def tangential_dims(self) -> list[str]:
+        """Tangential dimensions to a spherical surface in the spherical coordinate system."""
+        tangential_dims = ["theta", "phi"]
+        return tangential_dims
+
+    @staticmethod
+    def _check_coords_sorted(coord: np.ndarray, name: str):
+        """Helper for checking whether an array is sorted and raises an exception if it is not."""
+        is_sorted = np.all(np.diff(coord) >= 0)
+        if not is_sorted:
+            raise ValueError(f"{name} was not provided as a sorted array.")
+
+    def _check_integration_suitability(self):
+        """Checks whether the sampling of ``theta`` and ``phi`` is suitable for
+        integrating over a spherical surface."""
+        if (
+            len(self.theta) < MIN_ANGULAR_SAMPLES_SPHERE
+            or len(self.phi) < 2 * MIN_ANGULAR_SAMPLES_SPHERE
+        ):
+            raise ValueError(
+                "There are not enough sampling points along `theta` or `phi` for accurate integration. "
+                f"Currently, {len(self.theta)} samples for `theta` and {len(self.phi)} samples for `phi`. "
+                f"Consider using, at the very least, {MIN_ANGULAR_SAMPLES_SPHERE} samples for `theta` and "
+                f"{2*MIN_ANGULAR_SAMPLES_SPHERE} samples for `phi`."
+            )
+        self._check_coords_sorted(self.theta, "theta")
+        self._check_coords_sorted(self.phi, "phi")
+        if not isclose(self.theta[0], 0) or not isclose(self.theta[-1], np.pi):
+            raise ValueError(
+                "Chosen limits for `theta` are not appropriate for integration. "
+                "`theta` must range from 0 to π."
+            )
+        if not isclose(self.phi[0], 0) or not isclose(self.phi[-1], 2 * np.pi):
+            raise ValueError(
+                "Chosen limits for `phi` are not appropriate for integration. "
+                "`phi` must range from 0 to 2π."
+            )
+
+    def flux_from_projected_fields(self) -> FluxDataArray:
+        """Flux calculated by integrating the projected fields on a spherical surface.
+
+        Returns
+        -------
+        :class:`.FluxDataArray`
+            Flux in the frequency domain.
+        """
+        self._check_integration_suitability()
+        d_solid_angle = np.sin(self.Etheta.theta)
+        integrand = (self.power * d_solid_angle).sel(r=self.monitor.proj_distance)
+        flux = self.monitor.proj_distance**2 * integrand.integrate(self.tangential_dims)
+        return FluxDataArray(flux)
+
+    @staticmethod
+    def get_phi_slice(
+        field_array: FieldProjectionAngleDataArray, phi: float, symmetric: bool = False
+    ) -> FieldProjectionAngleDataArray:
+        """Get a planar slice of the :class:`.FieldProjectionAngleDataArray` along a given phi angle.
+        Extends theta range from [0, π] to [0, 2π] to create a full slice.
+
+        Parameters
+        ----------
+        field_array : :class:`.FieldProjectionAngleDataArray`
+            Field array to slice.
+        phi : float
+            Angle phi in radians to slice at.
+        symmetric : bool = False
+            If True, uses same data for both halves. If False, takes opposite phi angle
+            for back half.
+
+        Returns
+        -------
+        :class:`.FieldProjectionAngleDataArray`
+            2D slice with theta going from 0 to 2π.
+        """
+        slice_phi = field_array.sel(phi=phi, method="nearest")
+        slice_phi = slice_phi.where(slice_phi.theta < np.pi)
+        if symmetric:
+            slice_opposite_phi = field_array.sel(phi=phi, method="nearest")
+        else:
+            slice_opposite_phi = field_array.sel(phi=phi + np.pi, method="nearest")
+        slice_opposite_phi = slice_opposite_phi.where(slice_opposite_phi.theta > 0)
+        slice_opposite_phi = slice_opposite_phi.assign_coords(
+            theta=(2 * np.pi - slice_opposite_phi.theta)
+        )
+        data_array = xr.concat((slice_phi, slice_opposite_phi), dim="theta").sortby("theta")
+        return FieldProjectionAngleDataArray(data_array)
+
 
 class FieldProjectionCartesianData(AbstractFieldProjectionData):
     """Data associated with a :class:`.FieldProjectionCartesianMonitor`: components of
@@ -2627,7 +2744,7 @@ class FieldProjectionCartesianData(AbstractFieldProjectionData):
 
     @cached_property
     def flux(self) -> FluxDataArray:
-        """Flux for projecteded field data corresponding to a Cartesian field projection monitor."""
+        """Flux for projected field data corresponding to a Cartesian field projection monitor."""
         flux = self.poynting.integrate(self.tangential_dims)
         return FluxDataArray(flux)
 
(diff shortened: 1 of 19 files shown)


dmarek-flex:
Let's start of the review with RF. Not too complicated of changes, but maybe there is room for improvement on how we package the data. See [this notebook PR](https://github.com/flexcompute/tidy3d-notebooks/pull/244) for how the changes can be used.

weiliangjin2021:
1st round of review. All look nice!

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -410,3 +611,172 @@ def _check_port_impedance_sign(self, Z_numpy: np.ndarray):
...
+        Ephi = update_dict["Ephi"]
+        Htheta = update_dict["Htheta"]
+        Hphi = update_dict["Hphi"]
+        # Compute flux by integrating average power over spherical surface
+        power_theta = 0.5 * np.real(Etheta * Hphi.conj())
+        power_phi = 0.5 * np.real(-Ephi * Htheta.conj())
+        power = power_theta + power_phi
+        integrand = power * d_solid_angle
+        flux_values = r_proj**2 * integrand.integrate(("theta", "phi"))
+        flux = FluxDataArray(flux_values.squeeze().values, coords={"f": Etheta.f.values})
```
dbochkov-flexcompute: looks like this kind of integration comes up multiple times in this PR and in array factor. What do you think is the best place for a function like that which is used everywhere? I guess one option we could consider is to add into either `DirectivityData` or `FieldProjectionAngleData` and also make a convenience method either `DirectivityData.from_projection_data(projection_data)` or `FieldProjectionAngleData.to_directivity_data` that calculates flux based on fields in the projection data. That is, which would work something like this:
```
        update_dict = {}
        # Combine the field data
        for (key1, field1), (key2, field2) in zip(
            mon_data_1.field_components.items(), mon_data_2.field_components.items()
        ):
            assert key1 == key2
            update_dict[key1] = field1.copy() + field2.copy()
        projection_data_combined = mon_data_1.updated_copy(**update_dict)
        return projection_data_combined.to_directivity_data

Comment (dmarek-flex) [reply tidy3d/plugins/smatrix/component_modelers/terminal.py]: Yes, great idea! Would you like me to add that to this PR, since I am already modifying DirectivityData a lot?

Comment (dbochkov-flexcompute) [reply tidy3d/plugins/smatrix/component_modelers/terminal.py]: that would be great, thanks! I can just rebase that other PR onto this one at some point

Comment (dmarek-flex) [reply tidy3d/plugins/smatrix/component_modelers/terminal.py]: Ok got the `DirectivityData.from_field_projection_angle_data` method working. After some rearranging I think it is quite nice and simple now. DirectivityData now inherits directly from `FieldProjectionAngleData` and can be generated in mainly 2 ways (from `DirectivityMonitor` or `from_field_projection_angle_data`)

Comment (dmarek-flex) [reply tidy3d/plugins/smatrix/component_modelers/terminal.py]: ok maybe this is not exactly what we need :cry:. It is causing a lot of headache because I need to carry the monitor around which is of different types.

Comment (dmarek-flex) [reply tidy3d/plugins/smatrix/component_modelers/terminal.py]: although it might work fine for your purposes.

Comment (dbochkov-flexcompute) [reply tidy3d/plugins/smatrix/component_modelers/terminal.py]: ah, yeah, that's annoying. If having `from_field_projection_angle_data` turns out to be a so-so idea after all, I think having `flux_from_projected_field` that can be called from either `FieldProjectionAngleData` and `DirectivityData` is great already

Comment (dmarek-flex) [reply tidy3d/plugins/smatrix/component_modelers/terminal.py]: Yea, I decided to get rid of that small feature, since it would be probably a pain to maintain. Now, you can create `DirectivityData` fairly easily with a monitor/field_dataset pair.

**Comment by dmarek-flex** [comment]
@weiliangjin2021 @dbochkov-flexcompute Assuming all tests pass I am done the main changes I wanted to make to enable these features. Anything else I do will be typos and docstring improvements if I see any.

File: tidy3d/plugins/smatrix/data/terminal.py [hunk 2]
```diff
@@ -0,0 +1,195 @@
...
+    power_reflected: FreqDataArray = pd.Field(
+        ...,
+        title="Power reflected",
+        description="Array of values representing power reflected due to an impedance mismatch with the antenna.",
+    )
+
+    @staticmethod
+    def from_directivity_data(
+        dir_data: DirectivityData, power_inc: FreqDataArray, power_refl: FreqDataArray
+    ):
```
weiliangjin2021: return type hint `-> AntennaParametersData`

tidy3d/plugins/smatrix/data/terminal.py
```diff
@@ -0,0 +1,195 @@
...
+    def radiation_efficiency(self) -> FreqDataArray:
+        """The radiation efficiency of the antenna."""
+        return self.calc_radiation_efficiency(self.power_incident - self.power_reflected)
+
+    @property
+    def reflection_efficiency(self) -> FreqDataArray:
+        """The reflection efficiency of the antenna, which is due to an impedance mismatch."""
+        total_power = self.power_incident + self.power_reflected
+        reflection_efficiency = self.power_incident / total_power
+        return reflection_efficiency
```
weiliangjin2021: I thought reflection_efficiency = `(self.power_incident - self.power_reflected)/self.power_incident`, as the total power is the incident power?

tidy3d/components/data/monitor_data.py
```diff
@@ -2426,6 +2444,70 @@ def renormalize_fields(self, proj_distance: float) -> FieldProjectionAngleData:
...
+    def _check_integration_suitability(self):
+        """Checks whether the sampling of ``theta`` and ``phi`` is suitable for
+        integrating over a spherical surface."""
+        if (
+            len(self.theta) < MIN_ANGULAR_SAMPLES_SPHERE
+            or len(self.phi) < 2 * MIN_ANGULAR_SAMPLES_SPHERE
+        ):
+            raise ValueError(
+                "There are not enough sampling points along `theta` or `phi` for accurate integration."
+            )
```
weiliangjin2021, line 2575: More informative msg on how many sampling points are present, and how many are required.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -399,14 +435,147 @@ def _set_port_data_array_attributes(data_array: PortDataArray) -> PortDataArray:
...
+                scale_factor = amplitude / a_raw
+            a = scale_factor * a
+            b = scale_factor * b
+
+            # Combine the possibly scaled directivity data and the power wave amplitudes
+            if combined_directivity_data is None:
+                combined_directivity_data = directivity_data
+            else:
+                combined_directivity_data = self._combine_directivity_data(
+                    combined_directivity_data, directivity_data
```
weiliangjin2021: Probably looks cleaner by defining the operator + for directivity data?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -399,14 +435,147 @@ def _set_port_data_array_attributes(data_array: PortDataArray) -> PortDataArray:
...
+            a_raw = a.sel(port=port.name)
+
+            if amplitude is None:
+                directivity_data = sim_data_port[rad_mon.name]
+                scale_factor = 1.0
+            else:
+                directivity_data = self._monitor_data_at_port_amplitude(
+                    port, sim_data_port, radiation_data, amplitude
+                )
+                scale_factor = amplitude / a_raw
```
weiliangjin2021, line 598: A naive question: why is there a normalization of a_raw here?

weiliangjin2021 (commented):
Very remarkable work! A few minor comments below

tidy3d/plugins/smatrix/data/terminal.py
dmarek-flex: Yikes :face_with_diagonal_mouth: ! Yea at some point I switched the meaning of the power fields I think in `AntennaParametersData`.  Reflection efficiency should be the same as the transmission coefficient so you are correct!

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: The simulations launched in the `TerminalComponentModeler` do not create port excitations with unit power wave amplitude. So I have to normalize the results first and then scale by the desired port amplitude. Setting the port excitation amplitude in the `TerminalComponentModeler` is a feature we need, but is not absolutely necessary at the moment.

dmarek-flex: It's related to the question I asked you previously about how to setup `UniformCurrentSource` with a desired total current.

weiliangjin2021: make sense!

weiliangjin2021:
One architecture question: `AntennaParametersData` seems more suitable to be placed in `monitor_data.py`?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -399,14 +434,132 @@ def _set_port_data_array_attributes(data_array: PortDataArray) -> PortDataArray:
...
+        The method computes the radiated far fields and port excitation power wave amplitudes
+        for a superposition of port excitations, which can be used to analyze antenna radiation
+        characteristics.
+
+        Parameters
+        ----------
+        port_amplitudes : dict[str, complex] = None
+            Dictionary mapping port names to their desired excitation amplitudes. For each port,
+            :math:`\frac{1}{2}|a|^2` represents the incident power from that port into the system.
+            If None, uses only the first port without any scaling of the raw simulation data.
```
weiliangjin2021, line 544: What's the practical use case of `None`?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -399,14 +434,132 @@ def _set_port_data_array_attributes(data_array: PortDataArray) -> PortDataArray:
...
+            # Select a possible subset of frequencies
+            a = a.sel(f=f)
+            b = b.sel(f=f)
+            a_raw = a.sel(port=port.name)
+
+            if amplitude is None:
+                directivity_data = sim_data_port[rad_mon.name]
+                scale_factor = 1.0
+            else:
+                directivity_data = self._monitor_data_at_port_amplitude(
```
weiliangjin2021: probably clearer to call it `scaled_directivity_data`, to indicate that they are scaled by scale_factor/a_raw.

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: It is just a sensible shorthand for antennas with only one port, which is most likely the general use case. Many parameters of interest like efficiency and gain do not depend on the port amplitude, so we do not need to normalize the simulation data.

weiliangjin2021: but they still need to normalized by a_raw?

dmarek-flex:
> One architecture question: `AntennaParametersData` seems more suitable to be placed in `monitor_data.py`?

I think that is not a bad idea, but it seems like monitor data is usually associated with a monitor that directly creates it, which is the main reason I did not put it there. In some sense, `AntennaParametersData` is created by a monitor plus a `TerminalComponentModeler`. I am leaning towards leaving it in the microwave plugin. Alternatively, maybe I could add it to the `components/microwave` folder to keep it separate from the other monitor data.

@daquinteroflex @yaugenst-flex Any suggestions from outside the RF world?

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: No, since the main figures of merit (radiation efficiency/reflection efficiency/gain/realized gain) are all independent of the incident power when there is only one port. It is only when making linear combinations of the port amplitudes when you need to normalize the incident power to ensure they are being combined with the correct amplitudes/phase. 

Of course if the user specifically wants a certain port excitation amplitude, they must supply it. Then the raw simulation data will be normalized to get the desired result.

daquinteroflex:
So there's a few reasons I'll text privately, but basically everything RF specific like this, imo should go under 
```
tidy3d/components/microwave/
    data/
       ...
       # however we want to structure it in here.
```

Basically imagine that directory is a small world of its own that represents the current "tidy3d/components" and should be extended in a kind of self contained domain if possible

dmarek-flex:
@daquinteroflex @weiliangjin2021 Added a new commit improving docs and moving/renaming `AntennaMetricsData`. Does that seem consistent with other solver stuff? Later I will move the path integral code as a tool into that folder as well I think.

The one unique thing about `AntennaMetricsData` is that it is not directly created by a `Monitor`, but I ended up using the same kind of file naming scheme as other types of `MonitorData`. A little unsure about that.

daquinteroflex:
Nice, the reorganization makes sense! So for now, unless there's going to be a set of mixed monitor + other things data types - this still fits in logically within `microwave/data/monitor_data.py`.

In terms of naming, it's still a "data" container as such. If we believe this concept of "metrics data" might be extended, this could be a good name as you have it? It also kind of makes sense because it's like a "DeviceMetricsData" type. Can always refactor file location in the future depending how this changes too to `device_data.py` or `metrics_data.py` too.

dmarek-flex:
@daquinteroflex Yea I like metrics since it is shorter than what I had before and could be used in the future for other specific types of devices.

tidy3d/components/data/monitor_data.py
```diff
@@ -2518,6 +2536,73 @@ def renormalize_fields(self, proj_distance: float) -> FieldProjectionAngleData:
...
+        """Checks whether the sampling of ``theta`` and ``phi`` is suitable for
+        integrating over a spherical surface."""
+        if (
+            len(self.theta) < MIN_ANGULAR_SAMPLES_SPHERE
+            or len(self.phi) < 2 * MIN_ANGULAR_SAMPLES_SPHERE
+        ):
+            raise ValueError(
+                "There are not enough sampling points along `theta` or `phi` for accurate integration. "
+                f"Currently, {len(self.theta)} samples for `theta` and {len(self.phi)} samples for `phi`. "
+                "Consider using at least 45 samples for `theta` and 90 samples for `phi`, which "
```
weiliangjin2021: The value here looks different from `MIN_ANGULAR_SAMPLES_SPHERE = 10` defined at the beginning?

tidy3d/components/data/monitor_data.py
```diff
@@ -2518,6 +2536,73 @@ def renormalize_fields(self, proj_distance: float) -> FieldProjectionAngleData:
         # compute updated fields and their coordinates
         return self.make_renormalized_data(phase, proj_distance)
 
+    @property
+    def tangential_dims(self):
```
weiliangjin2021: return type hints, and docstring

weiliangjin2021 (approved):

weiliangjin2021 (approved):

dbochkov-flexcompute (approved):
A little bit of nitpicking, technically functions
- `scale_fields_by_freq_array`
- `get_phi_slice`
- `from_spherical_field_dataset`
- `calc_radiation_efficiency`
- `from_directivity_data`
- `compute_power_wave_amplitudes_at_each_port`
- `compute_port_VI`
- `compute_power_wave_amplitudes`
- `compute_power_delivered_by_port`

are exposed to the user, so ideally they should have `Parameters` and `Returns` sections in docstrings. Or you could just add underscore to those ones that don't really need to be accessible by the user. Otherwise everything looks great to me

PR #2191, opened 2025-01-22
fix[autograd]: ensure consistent task name mapping for batch autograd

yaugenst-flex:
https://github.com/flexcompute/tidy3d/pull/2061 introduced a bug that causes a mismatch between forward and adjoint simulation task names in `run_async` due to incorrect use of [rstrip](https://docs.python.org/3/library/stdtypes.html#str.rstrip). So whenever the forward task name ends with any character in `_adjoint`, `run_async` would break during autodiff. This PR fixes that and updates the autograd `run_async` test to check for this.


momchil-flex (approved):

momchil-flex:
Oh note that you've added the Changelog item in the wrong place.

PR #2192, opened 2025-01-22
fix[autograd]: ensure consistent task name mapping for batch autograd

yaugenst-flex:
Cherry-pick of https://github.com/flexcompute/tidy3d/pull/2191 into `pre/2.8` because this fix is important.


yaugenst-flex:
closing because we are merging instead

PR #2194, opened 2025-01-22
Merge/develop 2 8

momchil-flex:


yaugenst-flex (approved):
LGTM!

yaugenst-flex:
Seems like there are unresolved conflicts though?

momchil-flex:
I think for some reason this is how it shows when I make a merge branch PR, I have previously merged these locally and pushed and then the PR gets resolved. Will try.

yaugenst-flex:
oooh i think this just needs to be a merge PR instead of rebase and it looks fine. just that rebase is default so you have to manually switch. wasnt aware of that

image

momchil-flex:
Right, only then I don't want to do it through github because the default changes to merge instead of rebase (not sure if for me or for everyone) and I sometimes forget to change back for the next PR :)

PR #2199, opened 2025-01-23
:wrench: fix duplicate web folders

### LINKED ISSUE #2198: 🐛 Bug creating new folders with the Design plugin
**Describe the bug**
When a custom folder is used in the DesignSpace plugin, it seems like there is a multithreading issue when launching the run command where multiple folders are created.

**To Reproduce**
```python
design_space = tdd.DesignSpace(parameters=[param_l1, param_l2], method=method, path_dir="output_design",folder_name="pruc_cell_design2")

results = design_space.run(pre, post)
``` 

Gave me a couple of errors:

```shell
11:12:35 EST Running 40 Simulations                                             
11:12:36 EST ERROR: The name already exists.                                    
             ERROR: The name already exists.                                    
             ERROR: The name already exists.                                    
             ERROR: The name already exists. 
``` 
And also created duplicate folders on the web:

Image

**Expected behavior**
No errors and only a single folder is created on the web.

dmarek-flex:
My attempt at fixing this issue, but first time working in the Web area of the code!

### DIFF
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -19,6 +19,7 @@ from ...exceptions import DataError
 from ...log import get_logging_console, log
 from ..api import webapi as web
 from ..core.constants import TaskId, TaskName
+from ..core.task_core import Folder
 from ..core.task_info import RunInfo, TaskInfo
 from .tidy3d_stub import SimulationDataType, SimulationType
 
@@ -31,6 +32,18 @@ BATCH_MONITOR_PROGRESS_REFRESH_TIME = 0.02
 
 class WebContainer(Tidy3dBaseModel, ABC):
     """Base class for :class:`Job` and :class:`Batch`, technically not used"""
+
+    @staticmethod
+    def _check_path_dir(path_dir: str) -> None:
+        """Make sure ``path_dir`` exists and create one if not."""
+        location = os.path.dirname(path_dir)
+        if len(location) > 0 and not os.path.exists(location):
+            os.makedirs(location, exist_ok=True)
+
+    @staticmethod
+    def _check_folder(folder_name: str) -> None:
+        """Make sure ``folder_name`` exists on the web UI and create it if not."""
+        Folder.get(folder_name, create=True)
 
 
 class Job(WebContainer):
@@ -220,6 +233,7 @@ class Job(WebContainer):
         """The task ID for this ``Job``. Uploads the ``Job`` if it hasn't already been uploaded."""
         if self.task_id_cached:
             return self.task_id_cached
+        self._check_folder(self.folder_name)
         return self._upload()
 
     def _upload(self) -> TaskId:
@@ -290,6 +304,7 @@ class Job(WebContainer):
         ----
         To load the data after download, use :meth:`Job.load`.
         """
+        self._check_path_dir(path_dir=path)
         web.download(task_id=self.task_id, path=path, verbose=self.verbose)
 
     def load(self, path: str = DEFAULT_DATA_PATH) -> SimulationDataType:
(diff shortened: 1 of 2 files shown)


yaugenst-flex (commented):
Thanks @dmarek-flex this is great! The solution seems solid. However, after looking into these classes a bit I am noticing quite a few pitfalls in the current implementation (not due to your changes). In particular, the `_check_folder` and `_check_path_dir` checks are only performed during `run`, which means that when calling the I/O functions separately (`download`/`load` for example), they might break. Therefore I would suggest the following changes:

- Add `_check_path_dir` and `_check_folder` as `staticmethod` to `WebContainer`
- Add `self._check_path_dir` to both the `Job`'s as well as the `Batch`'s `download` and `load` methods
- Add `_check_folder` to the `Job`'s `task_id` property and to `Batch`'s `jobs` property (both after checking cache)

Sorry for blowing up your PR a bit but I think these changes would help make things more robust?

dmarek-flex:
Take a look. I no longer experience the issue, but I am a little confused whether it is safe to put `_check_folders` inside the `jobs` property, since a few times the property is accessed within a multithreaded loop. However, as long as the property is first  cached outside a loop then it should be fine.

yaugenst-flex:
Yeah I also considered this, but as you said it's fine as long as the `jobs` property is called at least once outside of the multithreaded loop, which is actually [true by definition](https://github.com/flexcompute/tidy3d/blob/e9298627568b6aaaa1e4e37089fd074b6a6565c3/tidy3d/web/api/container.py#L650) because `self.jobs` needs to be called in order to assign tasks to workers. I don't think caching plays a role here because the only thing we're trying to avoid is sending multiple folder creation requests to the web API simultaneously when the folder does not exist yet. Sending them after the fact is fine.

tidy3d/web/api/container.py
```diff
@@ -32,6 +33,18 @@
...
+    def _check_path_dir(path_dir: str) -> None:
+        """Make sure ``path_dir`` exists and create one if not."""
+
+        if not os.path.exists(path_dir):
+            os.makedirs(path_dir, exist_ok=True)
+
+    @staticmethod
+    def _check_folder(folder_name: str) -> None:
+        """Make sure ``folder_name`` exists on the web UI and create it if not."""
+        _ = Folder.get(folder_name, create=True)
```
yaugenst-flex: nit pick but I don't see a reason to do an assignment at all here

tidy3d/web/api/container.py
```diff
@@ -305,6 +320,7 @@ def load(self, path: str = DEFAULT_DATA_PATH) -> SimulationDataType:
         Union[:class:`.SimulationData`, :class:`.HeatSimulationData`, :class:`.EMESimulationData`]
             Object containing simulation results.
         """
+        WebContainer._check_path_dir(path_dir=path)
```
yaugenst-flex: use `self`

tidy3d/web/api/container.py
```diff
@@ -635,6 +645,7 @@ def num_jobs(self) -> int:
     def upload(self) -> None:
         """Upload a series of tasks associated with this ``Batch`` using multi-threading."""
 
+        WebContainer._check_folder(self.folder_name)
```
yaugenst-flex: I'm fine with having it either in `upload` or in `jobs`. The reason I picked jobs was I assumed there might be other methods that depend on the existence of the folder where the same issue might crop up, but it'll always depend on `jobs`? Not sure if that reasoning is sound :D

yaugenst-flex (approved):
Looks good! Prefer `self` over `WebContainer` everywhere but otherwise looks good.

tidy3d/web/api/container.py
dmarek-flex: I think I cannot put it in `jobs`. If I do a bunch of tests fail, since it tries access the web without api key (my rough understanding).

PR #2201, opened 2025-01-23
add flux monitor support

groberts-flex:


groberts-flex:
One short term addition to this before we can trace this automatically would be to let users turn off field creation for flux monitors they will not be using in the adjoint computation. We could create field monitors by default as we do here but give users the ability to at least turn that off if they have a bunch of flux monitors that aren't used in the adjoint computation. In practice, if they ended up using one of those flux monitors anyways, then they would get the error we currently throw that we can't differentiate through a flux monitor.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for quasi-uniform grid specifications via `QuasiUniformGrid` that subclasses from `GridSpec1d`. The grids are almost uniform, but can adjust locally to the edge of structure bounding boxes, and snapping points.
 - New field `min_steps_per_sim_size` in `AutoGrid` that sets minimal number of grid steps per longest edge length of simulation domain.
 - New field `shadow` in `MeshOverrideStructure` that sets grid size in overlapping region according to structure list or minimal grid size.
+- `FluxData` can now be used in autograd. The `FluxMonitor` is replaced with a `FieldMonitor` to create the proper adjoint source.
```
yaugenst-flex, line 16: I'd make this a bit clearer from a user-facing side, along the lines of "Autograd support for `FluxMonitor`". I think the `FieldMonitor` thing is more of an implementation detail (but you can mention it if you want). Actually, might be best to update the [autograd readme](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/plugins/autograd/README.md) and add a short explanation for the implementation of `FluxMonitor`.

yaugenst-flex (commented):
Nice @groberts-flex this will be great to have! It has proven to be a pain point for some users in the past.

Only thing I'm a bit uneasy with regarding the organization is whether we really want to have a dedicated `make_adjoint_proxy_monitor` method in the base class from which most of our monitors derive while it is really only needed by a single monitor (seems like an unnecessary abstraction). Should we maybe instead just check for `FluxMonitor` type explicitly in `_run_primitive`?

tidy3d/web/api/autograd/autograd.py
```diff
@@ -400,6 +400,18 @@ def _run_primitive(
...
+    for monitor_ in sim_original.monitors:
+        mnt_proxy = monitor_.make_adjoint_proxy_monitor()
+
+        adjoint_proxy_monitors.append(mnt_proxy)
+
+    proxy_monitors = [
+        sim_original.monitors[idx].make_adjoint_proxy_monitor()
+        for idx in range(0, len(sim_original.monitors))
+    ]
+    sim_original = sim_original.copy(update=dict(monitors=proxy_monitors))
```
tylerflex, line 413: huh, so this just works? I guess basically when the user tries to extract their FluxData, they get a FieldData instead and the `.flux` call just works? neat!

yaugenst-flex:
> One short term addition to this before we can trace this automatically would be to let users turn off field creation for flux monitors they will not be using in the adjoint computation. We could create field monitors by default as we do here but give users the ability to at least turn that off if they have a bunch of flux monitors that aren't used in the adjoint computation. In practice, if they ended up using one of those flux monitors anyways, then they would get the error we currently throw that we can't differentiate through a flux monitor.

That does sound like a good idea, just wondering how best to implement it. I guess a general solution for all monitors (until we figure out how to do it automatically) would be to have, e.g., an optional `autograd_freqs` field that could be set to `False` (no tracing) or to a list of frequencies to be considered for autograd. Let's discuss offline, I think that's a bit out of scope for this PR.

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: that's what i thought :smile:

tylerflex: the original way I had thought about implementing this, we'd essentially duplicate the monitors (so add an extra set of "auxiliary" monitors) for anything needing them). And then when writing the FluxData adjoint source, it would just grab the field profile for the source from the auxiliary FieldData and multiply by the `complex` vjp value in the FluxData vjp. The idea here being that users wouldn't get confused if their data comes back as a FieldData. I think either way works, my approach is more effort, but just something to keep in mind.

tylerflex: (sorry, not duplicate monitors, but rather I guess keep the original one but add an extra FieldData, if it wasn't clear)

tylerflex: After thinking about it some more, I'm not sure if what I wrote above would work actually. What we would need to do if the user calls `FluxData.flux` in their FoM is essentially replace that call with a call to the auxilary `FieldData.flux` and use that value instead. Not sure if there's any clean way to do this. maybe the way it's done here is the only way?

groberts-flex: Thanks for the suggestions and thoughts on this. That was what I was getting stuck on when I tried to implement this with an auxiliary monitor. It does seem better to me to allow the user to get FluxData back if they put in a FluxMonitor, but I was having trouble using the FieldData in place of the FluxData because there wasn't a gradient to use there since all the fields in the backwards pass in that data were 0 (since flux was not called on that auxiliary data). So there may be a way still to do this, but I was really struggling to figure out how to do it cleanly!

tylerflex: so in this implementation, will **all** FluxMonitors be replaced by FieldMonitors assuming the user is running an objective function though autograd? this might not be ideal. for example, say the user adds a FluxMonitor just for looking at the spectrum (not included in the objective function), it will then have to download all of the FieldData?

yaugenst-flex: yes that is the case, but I think that would also be the case in the other approach

tylerflex: hm, well either way maybe it's not ideal? for many users this could increase their download unexpectedly, potentially quite a lot? unfortunately I dont see a way around It though if we want to try to support FluxMonitor. 

One option could be to add a special flag to FluxMonitor to explicitly allow us to convert to a FieldMonitor under the hood? but if the user is already toggling this flag, maybe they could also just use FieldMonitor?

yaugenst-flex: hence the option to set traced freqs as a field up next 😄

yaugenst-flex: I think it's better to support it and point to that if there is a problem rather than not supporting `FluxMonitor`

momchil-flex: Yeah it's actually nice that it works out so simply. :) Wondering if we should add a warning about the conversion but that would make it kind of annoying to use then.

groberts-flex:
Closing this PR for now until we can better solve this problem

PR #2202, opened 2025-01-24
Add twin axis plot for n and k

txdai:
Normally n and k have widely different ranges, making plotting them with a single y axis difficult. This update uses the twin axis functionality of matplotlib to make this clear. For example this figure from the documentation clearly shows this difficulty.

### DIFF
--- a/tidy3d/plugins/dispersion/fit.py
+++ b/tidy3d/plugins/dispersion/fit.py
@@ -476,17 +476,20 @@ class DispersionFitter(Tidy3dBaseModel):
         medium: PoleResidue = None,
         wvl_um: ArrayFloat1D = None,
         ax: Ax = None,
+        dual_axis: bool = False,
     ) -> Ax:
         """Make plot of model vs data, at a set of wavelengths (if supplied).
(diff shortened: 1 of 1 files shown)


groberts-flex:
Thank you for submitting this PR, it will be a very useful feature for visualizing permittivity data! I put a couple of minor comments on the code and a slightly larger request is if this could be updated so that the dual axis behavior is turned on with an optional flag argument to the function. The reason being that while the twin axis is extremely helpful in seeing the imaginary component value magnitudes, it does create a more crowded plot and in some cases users may still want to be able to see the relative magnitude of the real and imaginary components in which case it is helpful for them to be plotted on the same axis scaling. I would tend towards the dual axis behavior to be off by default with a flag to turn it on.

txdai:
I just included the options for enabling the dual axis and set the default to being "False" so it does not affect current behavior

groberts-flex:
Thank you for the changes, this looks great! I was able to test it locally and see the different plot options turn on and off as expected. Appreciate your contribution!

PR #2205, opened 2025-01-27
Constant doping box foxed

marc-flex:


daquinteroflex (approved):
Nice! Just to check this just consolidates the doping concentration distribution for gaussian or constant doping then.

marc-flex:
> Nice! Just to check this just consolidates the doping concentration distribution for gaussian or constant doping then.

Yes, I still need to add a test for constant doping

tidy3d/components/tcad/doping.py
```diff
@@ -56,6 +56,47 @@ def check_bounds(cls, values):
 
         return values
 
+    def _get_indices_in_box(self, coords: dict, meshgrid: bool = True):
```
momchil-flex, line 29: So every geometry (so including `Box`) has an [inside](https://github.com/flexcompute/tidy3d/blob/948098a3940500b7dee71cdefe5bd94600717378/tidy3d/components/geometry/base.py#L89) and [inside_meshgrid](https://github.com/flexcompute/tidy3d/blob/948098a3940500b7dee71cdefe5bd94600717378/tidy3d/components/geometry/base.py#L179) method. Can these be used directly instead of this? It seems like this handles some special cases but wondering if you've considered those.

tidy3d/components/scene.py
```diff
@@ -1866,10 +1866,9 @@ def _pcolormesh_shape_doping_box(
                 struct_doping[n] = struct_doping[n] + contrib
             if isinstance(doping, tuple):
                 for doping_box in doping:
-                    if isinstance(doping_box, ConstantDoping):
-                        contrib = np.ones(X.shape) * doping_box.concentration
-                        struct_doping[n] = struct_doping[n] + contrib
-                    if isinstance(doping_box, GaussianDoping):
+                    if isinstance(doping_box, ConstantDoping) or isinstance(
+                        doping_box, GaussianDoping
+                    ):
```
momchil-flex: Just fyi this can be isinstance(doping_box, (ConstantDoping, GaussianDoping)). Same on backend.

tidy3d/components/tcad/doping.py
marc-flex: Didn't know about those. Will check how I can make it work

tidy3d/components/scene.py
marc-flex: I'll change it now

tidy3d/components/tcad/doping.py
```diff
@@ -70,11 +70,30 @@ def _get_indices_in_box(self, coords: dict, meshgrid: bool = True):
...
+    def check_dimensions(cls, values):
+        """Make sure dimensionality is specified correctly. I.e.,
+        a 2D box must be defined with an inf size in the normal direction."""
+
+        size = values["size"]
+        for dim in range(3):
+            if size[dim] == 0:
+                dim_name = "xyz"[dim]
+                raise SetupError(
+                    f"The doping box has been setup with 0 size in the {dim_name} direction. "
```
momchil-flex: set up

tidy3d/components/tcad/doping.py
```diff
@@ -70,11 +70,30 @@ def _get_indices_in_box(self, coords: dict, meshgrid: bool = True):
...
+                    "If this was intended to be a 2D box, the box must have inf size in the "
```
momchil-flex: 2D still sounds a bit undefined to me, whether it means infinitely extended or infinitely thin. How about something like "if this was intended to be translationally invariant, the box must have ..."

tidy3d/components/tcad/doping.py
```diff
@@ -70,11 +70,30 @@ def _get_indices_in_box(self, coords: dict, meshgrid: bool = True):
...
+                )
+
+        return values
+
 
 class ConstantDoping(AbstractDopingBox):
     """
     Sets constant doping :math:`N` in the specified box with a :parameter`size` and :parameter:`concentration`.
 
+    For 2D behavior, the box must have infinite size in the homogenous direction.
```
momchil-flex: Same here and below, instead of `2D behavior` maybe `translationally invariant`?

momchil-flex (approved):
Looks good for now, to be revisited later maybe.

daquinteroflex (approved):
Nice yeah the usage is more consistent with box

PR #2208, opened 2025-01-28
Add ModeSimulation classes

caseyflex:
Another attempt at this old PR: https://github.com/flexcompute/tidy3d/pull/1838. I rewrote it pretty extensively to try to minimize the size of the changes and make this PR more manageable. (I think I incorporated all changes made since my original attempt.) Now the `ModeSimulation` is implemented as a fairly thin layer on top of the existing `ModeSolver`. And there is no backend PR currently. If we want more convenience features maybe we can implement them in a future PR just so this doesn't get too big again?

Motivation: 

The current `ModeSolver` is not unified with our other simulation types, which results in several annoying interface issues. It needs an FDTD simulation, which has extra data like run_time and sources that isn't needed for the mode simulation. Also, the resulting data `ModeSolverData` doesn't know about the original simulation, which makes plotting difficult.

Summary of changes:

- Moves local ModeSolver from plugins to components (but left the plugin interface as before)
- Adds `ModeSimulation` and `ModeSimulationData` classes.
- Adds the following key methods: `ModeSimulation.mode_solver` and `ModeSimulation.from_mode_solver` for easy conversion to / from `ModeSolver`. `ModeSimulation.from_simulation` for easy construction from an FDTD or EME simulation.`ModeSimulation.reduced_simulation_copy` for efficiency. `ModeSimulation.run_local` for local mode solving directly (could get rid of this since ModeSolver plugin would work). And a bunch of plotting functions.
- `ModeSolver.simulation` can now be a `ModeSimulation`.

Some subtleties:

- There are currently two sets of plotting functions: those inherited from `Simulation`, and those that wrap around `ModeSolver` plotting, which I named `plot_mode_plane` and the like. There seem to be two key differences: the `ModeSolver` plotting only plots the mode solving plane, compared to `sim.plot(y=0)` which can plot any cross-section; and the `ModeSolver` plotting shows the PML from the mode_spec (which is the relevant one), rather than the PML from the boundary_spec.
- The web api is just wrapping around the `ModeSolver` web api support, which is very convenient and simple but maybe not optimal long-term. For example it still saves the `ModeSolverData` in the file rather than the `ModeSimulationData`, which makes loading a bit clunky as shown below. I'd rather save changing this for the future though as it was a sticking point last time I tried to make this PR.

Example:

```
# standard imports
import tidy3d as td
from tidy3d.plugins.mode import ModeSolver
import tidy3d.web as web

# basic simulation parameters
sim_size = (3, 3, 3)
lambda0 = 1.55
freq0 = td.C_0 / lambda0
si = td.material_library["cSi"]["Li1993_293K"]
sio2 = td.material_library["SiO2"]["Horiba"]
wg = td.Structure(
    geometry=td.Box(size=(.22, .5, td.inf)),
    medium=si
)
mode_plane = td.Box(size=(td.inf, td.inf, 0))
mode_spec = td.ModeSpec(num_pml=(12, 12))
grid_spec = td.GridSpec.auto(wavelength=lambda0, min_steps_per_wvl=30)

# construct and plot simulation
mode_sim = td.ModeSimulation(
    size=sim_size,
    medium=sio2,
    structures=[wg],
    grid_spec=grid_spec,
    plane=mode_plane,
    mode_spec=mode_spec,
    freqs=[freq0]
)
mode_sim.plot(z=0)
mode_sim.plot_mode_plane()

# run it locally
mode_sim_data_local = mode_sim.run_local()
mode_sim_data_local.plot_field("Ey")
print(float(mode_sim_data_local.modes.n_eff))

# run it remotely via standard web api
mode_sim_data = web.run(mode_sim, task_name="mode_solver")
mode_sim_data.plot_field("Ey")
print(float(mode_sim_data.modes.n_eff))

# load saved data from file
mode_sim_data = td.ModeSimulationData(
    simulation=mode_sim, 
    modes=td.ModeSolverData.from_file("simulation_data.hdf5")
)

```

TODO:

- [x] Clean up tests and improve test coverage
- [x] Check over everything again


caseyflex:
I'm still polishing some things (especially the unit tests) but wanted to get on the same page about what this PR will look like. Open to feedback and discussion at this point, especially regarding the webapi and plotting subtleties I mentioned above.

tidy3d/components/mode/mode_solver.py
```diff
@@ -0,0 +1,2205 @@
+"""Solve for modes in a 2D cross-sectional plane in a simulation, assuming translational
+invariance along a given propagation axis.
+"""
```
momchil-flex, line 3: I'm confused about this file as github is showing it as entirely new, but it seems pretty much the same, are there any other changes except supporting `ModeSimulation`?

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,455 @@
...
+    @pd.validator("grid_spec", always=True)
+    def _validate_auto_grid_wavelength(cls, val, values):
+        """Handle the case where grid_spec is auto and wavelength is not provided."""
+        # this is handled instead post-init to ensure freqs is defined
+        return val
+
+    @cached_property
+    def mode_solver(self):
+        """Convert the :class:`.ModeSimulation` to a :class:`.ModeSolver`."""
+        from .mode_solver import ModeSolver
```
momchil-flex: I guess to avoid circular imports?

momchil-flex (commented):
Thanks Casey. I think having the web api return a ModeSolverData rather than a ModeSimulationData is definitely not great. I am wondering how to handle this so it doesn't spiral out of control again. On one hand I feel like we really want to move away from just wrapping the existing ModeSolver, but on the other hand I know how hard it can be to carry such a PR around as other changes happen (we have smaller projects on the mode solver). Maybe we start merging things without making ModeSimulation officially announced/documented, and only do that when we are happy with the final state. For example, I think we may only do that once we support proper ModeSimulation pipelines as was originally done in the backend PR, which upload a ModeSimulation and return a ModeSimulationData. This could also be used by the GUI team to start building the native support for it. Then going beyond that, do you see a path to incrementally completely extricate ModeSimulation from ModeSolver?

tidy3d/components/mode/mode_solver.py
caseyflex: Yeah, it's almost exactly the same file. The only changes are the addition of the validator `_validate_mode_simulation` which ensures consistency between the ModeSimulation and the ModeSolver, the addition of ModeSimulation to MODE_SIMULATION_TYPE and ModeSimulationData to MODE_SIMULATION_DATA_TYPE, and the function `sim_data` is modified if simulation is a ModeSimulation

caseyflex:
> Thanks Casey. I think having the web api return a ModeSolverData rather than a ModeSimulationData is definitely not great. I am wondering how to handle this so it doesn't spiral out of control again. On one hand I feel like we really want to move away from just wrapping the existing ModeSolver, but on the other hand I know how hard it can be to carry such a PR around as other changes happen (we have smaller projects on the mode solver). Maybe we start merging things without making ModeSimulation officially announced/documented, and only do that when we are happy with the final state. For example, I think we may only do that once we support proper ModeSimulation pipelines as was originally done in the backend PR, which upload a ModeSimulation and return a ModeSimulationData. This could also be used by the GUI team to start building the native support for it. Then going beyond that, do you see a path to incrementally completely extricate ModeSimulation from ModeSolver?

Thanks for the feedback. I don't think there is any problem with just wrapping the ModeSolver, actually. We can think of it as containing the actual computational technical details, while ModeSimulation provides the interface and can use ModeSolver as a black box. The user won't see any of this and can just interact with ModeSimulation. We would need to modify the web api / change the pipelines to allow direct submission of a ModeSimulation / return of a ModeSimulationData. I just remember this being a pain last time, MC didn't want to have a new pipeline for it separate from ModeSolver pipeline. Maybe you could help negotiate this...

I like the idea of merging this before the pipeline is fully setup. You can see the way I set up the webapi handling in frontend currently, it's not very elegant but it would be a minor change to make it more consistent once the pipelines are set up, and at least this lets us run everything server-side already.

tidy3d/components/mode/simulation.py
caseyflex: Yeah, exactly

momchil-flex:
Ok, I'll do a deeper dive in this at some point. Generally my desire to fully separate them eventually is because I suspect if we don't things will keep being unnecessarily tangled, there may be code duplication (like if we want to add a new attribute to ModeSolver/ModeSimulation), and just potential confusions about the conversion (even from a developer point of view, given that we manage to completely hide it from the user). I think we should also use ModeSimulation as a way to fix some of the poor design choices we made previously. For example sticking `num_pml` in the mode_spec should likely be eventually deprecated in favor of the proper `ModeSimulation.boundary_spec`. Similarly we need to think hard about how exactly we want to handle the frequency for the auto grid so that we both don't need to carry around a meaningless source but also the discretization matches that of a corresponding FDTD simulation under typical usage. None of those things may require decoupling from the `ModeSolver`, at least for some time, but I think doing so eventually will be cleaner.

caseyflex:
I agree some things like PML handling and autogrid can be improved. In this version, in API at least, autogrid is unified because sources are not allowed, so only the simulation freqs can be used. But PML handling cannot be unified so simply, because in an FDTD sim used in mode solver, both PMLs do something but they do different things. So that’s what I do here too.

let me know if you think any of this handling can be improved in a clean way, otherwise my preference is to make these possibly backwards-compatibility breaking changes in a future PR.

The way it’s currently set up, btw, if you wanted to add a new function to mode solver and mode simulation, you would add the implementation to mode solver and then make a wrapper in ModeSimulation that calls modesim.mode_solver

momchil-flex:
> let me know if you think any of this handling can be improved in a clean way, otherwise my preference is to make these possibly backwards-compatibility breaking changes in a future PR.

Yeah just to clarify from the start I'm doing some preliminary discussion here just to say while I feel like eventually we may want to fully separate things and clean up some of these design choices, but yeah I'm fine with leaving this for the future. As a side note, apart from being able to use the standard web calls, which is now actually enabled for the mode solver, what were some of the other pain points that we encountered that we were hoping your refactor will fix? Are they addressed here or will some also need to wait for future PRs? I remember every now and then we'd discuss about something that it can be handled in the refactor, but I don't remember what. :D

> I agree some things like PML handling and autogrid can be improved. In this version, in API at least, autogrid is unified because sources are not allowed, so only the simulation freqs can be used.

As far as I understand, it needs to be a single frequency provided in `freqs`, or effectively if a list they need to be identical? I feel like what may make more sense is either just having `wavelength` being mandatory in AutoGrid for `ModeSimulation`-s, or taking something like the average of the `freqs`? Then there are two things we need to consider: turning ModeSimulation to a Simulation (e.g. to get the grid) -> instead of adding sources, maybe we should explicitly set the AutoGrid wavelength to whatever choice we make for the ModeSimulation; and turning Simulation to a ModeSimulation (e.g. when a Simulation contains a ModeMonitor) -> then we probably want to set the ModeSimulation AutoGrid wavelength to the same thing that would be used [here](https://github.com/flexcompute/tidy3d/blob/c627db39d51de5ae1cb2a2dee5eb4ecab79ec40a/tidy3d/components/grid/grid_spec.py#L843-L845).

> But PML handling cannot be unified so simply, because in an FDTD sim used in mode solver, both PMLs do something but they do different things. So that’s what I do here too.

The FDTD sim in the ModeSolver, in principle, doesn't do anything apart from help with meshing, plotting, etc, right? But anyway I probably know what you're driving at, I would say that the fundamental issue may be that in FDTD simulations containing mode objects, we need a way to define everything for the associated mode solver inside e.g. the ModeMonitor object. So this normally is done through the `mode_spec`, plus a few other things that are just regular `ModeMonitor` attributes, like the `freqs` list. So now the question is how should we define a `boundary_spec` for the mode solver. It can't be the one of the Simulation obviously. But we also don't want to put it in `mode_spec` because it will be confusing if ModeSimulation has both a `boundary_spec` and a `mode_spec.boundary_spec` attribute. There are various options here, including redesigning `mode_spec` to be purely a `ModeSource`/`ModeMonitor` thing, while all its fields are unpacked in a `ModeSimulation`. I'm just brainstorming here, there could be a lot of options.

> The way it’s currently set up, btw, if you wanted to add a new function to mode solver and mode simulation, you would add the implementation to mode solver and then make a wrapper in ModeSimulation that calls modesim.mode_solver

Yeah that's clear, it just doesn't make me super happy. :D

caseyflex:
I have some of the pain points at the top of this PR. Some big ones are 1. The need for an FDTD simulation, which requires extra data not needed in the ModeSimulation. 2. The lack of a good plot_field function, which needs the ModeSimulationData to know about the original ModeSimulation. 3. You currently need to run an FDTD simulation just to get permittivity data. This PR doesn't solve that right away, but later we can add PermittivityMonitor to a ModeSimulation and charge as such.

Currently, if the freqs are all equal, it uses that for the autogrid, and otherwise, the `wavelength` field in autogrid is mandatory. This is very reasonable behavior for the primary use case where the grid doesn't matter too much and just needs to be fine enough, although there are two cases that it misses. 1. The user has an FDTD sim and wants the grid to match exactly, then they should set the `wavelength` field in autogrid. 2. The user doesn't care about the grid but has multiple frequencies, then it's a bit annoying to have to set `wavelength` manually. We could handle the second case by using the `min(wavelength)`. But it's a bit at odds with the first case, where we want to warn the user that they might care about the grid matching exactly.

What I meant about the PML is that the `boundary_spec` actually changes the geometry of the simulation, for example it is not uncommon for `ModeSolver.plane` to extend into the FDTD PML layers. Also, the presence of the PML layers can show up subtly in things related to grid, symmetry, colocation, and subsection -- I had a real headache trying to deal with this in the earlier PR. For example even if the mode plane just extended to the edge of the simulation and not past it, it included a couple of grid points in PML for colocation, which was hard to reproduce without geometric PML layers in the ModeSimulation. Easier to just match the FDTD geometry / setup here. And then the `mode_spec` api can also stay the same as before, and give the same results. I couldn't think of a better way to unify it because `boundary_spec` actually adds Yee cells to the `AbstractYeeGridSimulation`.

caseyflex:
The way the autogrid is setup now mirrors EME. We could change it to use `min(td.C_0 / freqs)` if multiple freqs are supplied. Independently from that, we could log a warning (or just an info) about how the wavelength is being inferred and can be set manually if it matters. In the context of the ModeSimulation, it might be appropriate to log this regardless of how many freqs are supplied. In the context of EMESimulation, it may be less appropriate since EMESimulation is less closely related to FDTD Simulation than ModeSimulation is.

momchil-flex:
> What I meant about the PML is that the `boundary_spec` actually changes the geometry of the simulation, for example it is not uncommon for `ModeSolver.plane` to extend into the FDTD PML layers. Also, the presence of the PML layers can show up subtly in things related to grid, symmetry, colocation, and subsection -- I had a real headache trying to deal with this in the earlier PR. For example even if the mode plane just extended to the edge of the simulation and not past it, it included a couple of grid points in PML for colocation, which was hard to reproduce without geometric PML layers in the ModeSimulation. Easier to just match the FDTD geometry / setup here. And then the `mode_spec` api can also stay the same as before, and give the same results. I couldn't think of a better way to unify it because `boundary_spec` actually adds Yee cells to the `AbstractYeeGridSimulation`.

Ah yes I see this is quite a headache indeed. I guess not even in the case of PML: the fact that the mode solve is effectively done on the *intersection* between the simulation domain and the defined plane makes it pretty tricky. For example it's a common use case in 2D simulations to define the plane to be infinite in the extruded dimension. Now that I think about it, I was kind of imagining the `ModeSimulation` will no longer have a `plane` argument but instead the plane would be implicitly defined by the `center` and `size`. But I can see how there can be various issues with that. So instead are you thinking that the current setup will be how it would be in the long-term, namely that the `ModeSimulation` comes with some 3D size that kind of looks like an FDTD simulation, + a plane that defines where exactly to do the mode solve?

caseyflex:
I think the current setup is good -- it actually allows both options. You can have a 3D simulation and specify the `plane`. Or, you can have a 2D simulation and leave `plane=None`, in which case the validator makes it equal to the entire `simulation.geometry`. It errors if `plane=None` and the geometry is not 2D. This does complicate the plotting a bit (like I said above, there are basically two versions of plotting -- one to plot it as a 3D simulation, and one to plot the mode plane), but it follows the philosophy of minimizing the changes relative to the current ModeSolver api (for example, the GUI actually works fine for `ModeSimulation` remote runs already, it just thinks it is a `ModeSolver`.)

Thanks for the discussions btw! Good to decide on these things. I generally will push for keeping this PR minimalistic though, to avoid breaking backwards compatibility, to avoid confusing the user, to simplify the development for frontend / backend / web / GUI, and to keep it from getting stuck in limbo again.

momchil-flex:
> Thanks for the discussions btw! Good to decide on these things. I generally will push for keeping this PR minimalistic though, to avoid breaking backwards compatibility, to avoid confusing the user, to simplify the development for frontend / backend / web / GUI, and to keep it from getting stuck in limbo again.

Yeah, I mean thanks for bearing with me, you've obviously thought about this a lot and I'm trying to catch up. Will do another deep dive in this in the coming days.

tylerflex:
Let's also get @daquinteroflex to take a look at this, as it factors into the long term multiphsics solver plan. They've had quite a push lately for charge solver, so we can take our time on this

caseyflex:
> Let's also get @daquinteroflex to take a look at this, as it factors into the long term multiphsics solver plan. They've had quite a push lately for charge solver, so we can take our time on this

Yeah, no need to rush it out, I just want to keep the scope of this PR manageable.

daquinteroflex:
Thanks Casey for such an interesting and thought out PR! We should definitely keep the scope limited to get the ball rolling on this. I've done a skim through the code and just writing general thoughts. There's much to think about too from the discussion and code so still getting my head around things, this is just what comes to mind so far.

**In terms of future thoughts/not necessarily this PR**: I think this PR and Momchil/Casey discussion raises an important question: when and how do we break compatibility? If we consider carrying the tricky `ModeSolver` code is going to cause us more problems in the long run (and it kind of looks like already we're trying to get around issues with it), maybe we need to seriously consider either removing it or properly refactoring it into something that is more compatible with our current/future plans. This of course has an effect for MC, backend, users. However, it could enable us to be better placed to extend cleanly. We know users can have issues with the ModeSolver existing toolset, so if we can progressively work to replace it fully that's probably good. 

**In terms of this PR**: I really like how this PR is focused just on the minimal aspects towards standardization even in a progressive direction. Also on the same lines to Momchil's thoughts, it'd be nice if we can extirpate the `ModeSolver/ModeSolverData` old implementation given we'd be carrying around code that could complicate future extensibility. In this sense, I am unsure whether we should release this to users immediately, because it depends on how we want to handle the `ModeSolver` in the future or just fully migrate to this new approach without this PR becoming a growing snowball. In this sense, until we decide this not super sure we want to release to users yet? Imo, if we can at least have a decoupled "dataclass" for a `ModeSimulation`, then it's much more straightforward to extend of build upon it both on our existing flow and future multiphysics integration. If we decide we will get rid of the `ModeSolver` as such, then it can be more straightforward to have even that "computational" implementation like Casey suggests align with the existing monitor/data strategy or not having to come around with workarounds? This also raises a question how we want to handle relationships between simulations `Simulation <-> ModeSimulation`, but we probably want to separate data classes as much as possible in order to have a clear hierarchy and then coupling/relationships methods/classes between them like already implemented on the `ModeSimulation.from_simulation`. This should enable functional extensibility.

If we don't make this class available to users currently, this PR can give us the structure to build and modify the `ModeSolver` in the consolidated approach progressively. If we can eventually have a `ModeSimulation/ModeSimulationData` implementation that is unlinked from if-we-depreciate the `ModeSolverData etc.` this would be ideal possibly. 

As a scope of the PR to be the first step towards the new structure, broadly sounds great though! Assuming we can have flexibility to refine these questions without having to support user backwards compatibility.

**Brainstorming ideas on keeping compatibility?**
I'm not fully sure/convinced these make sense, but for the sake of argument just mentioning them. 

* It's also a question whether we want/need to fully unlink from the `ModeSolverData` and just restructure it into the new flow, but I guess this kind of led to any change keeping growing from Casey's old PR. Which is why this one is interesting, because maybe it gives us a framework to build upon whilst everything else gets worked on. 
* Even if we have to add legacy compatibility methods, this is imo still alright - as long as the data and the methods are separate in terms of how it gets solved within the new `ModeSimulation/ModeSolver` structure and we're not causing a problem for us down the line due to mantaining backwards compatibility.
* Could we split up the `ModeSolver/ModeSolverData` as a general monitor/data function such as `modesolver(ModeSimulation, monitors=[stuff inside ModeSolver]) -> ModeSimulationData(data=MonitorData structure etc.)`? That way the `ModeSimulationData` would not require to declare the mode solver old class directly and we could use it part of the `data` parameter.

caseyflex:
@daquinteroflex Thanks for the feedback. I'm not sure I fully understand, please correct me if I missed your point.

I don't think we will get rid of `ModeSolverMonitor` or the corresponding `ModeSolverData` classes, because these can be quite useful in an FDTD simulation.

The way it is currently set up, a `ModeSimulation` corresponds to a single run of the `ModeSolver`. If you wanted to run multiple mode solves in different planes in the same 3D simulation, you would need to use a Batch. This is different from an FDTD simulation, where you could have as many `ModeSolverMonitor` objects as you wanted in a single simulation. Setting it up this way also aligns with the philosophy of minimally extending the `ModeSolver` and not breaking anything.

I think the current set up captures the primary use case and frustration, which was not with `ModeSolverMonitor` functionality but rather with the `ModeSolver` functionality, namely the fact that it is not handled as an `AbstractSimulation` subclass that outputs a `AbstractSimulationData`. It does miss out on the use case where someone wants to effectively run an FDTD simulation but only perform the mode solves, with multiple `ModeSolverMonitors`. However I think this is not that common, when it does occur it's probably only a couple and can be handled using `Batch`, and possibly we can extend to offer this functionality in the future, gradually, after this PR.

Regarding decoupled dataclasses -- this PR does provide that with `ModeSimulation` and `ModeSimulationData` -- it just still uses `ModeSolver` internally (it doesn't have to be exposed to the user, although it might be convenient to do so), and it uses `ModeSolverData` in the natural way (although in `sim_data.modes` rather than `sim_data.data` for the reason discussed above).

weiliangjin2021:
Nice work! A subpixel-related question once we have a standalone mode simulation class:

- Mode solver is 2D, while FDTD is 3D. So people will tend to define the simulation domain size and structures to be 0 in out-of-plane dimension, or structures terminate exactly at the modal plane. However, right now we are still perform volumetric averaging, so you might find the results totally off. In principle, we can apply area averaging, but that can generate different results when we run mode solver as part of FDTD.

daquinteroflex:
> I think the current set up captures the primary use case and frustration, which was not with `ModeSolverMonitor` functionality but rather with the `ModeSolver` functionality, namely the fact that it is not handled as an `AbstractSimulation` subclass that outputs a `AbstractSimulationData`. It does miss out on the use case where someone wants to effectively run an FDTD simulation but only perform the mode solves, with multiple `ModeSolverMonitors`. However I think this is not that common, when it does occur it's probably only a couple and can be handled using `Batch`, and possibly we can extend to offer this functionality in the future, gradually, after this PR.
>
> Regarding decoupled dataclasses -- this PR does provide that with `ModeSimulation` and `ModeSimulationData` -- it just still uses `ModeSolver` internally (it doesn't have to be exposed to the user, although it might be convenient to do so), and it uses `ModeSolverData` in the natural way (although in `sim_data.modes` rather than `sim_data.data` for the reason discussed above).

Ahh I see! Still getting familiar with how everything works in more detail.

Based on the code and how you've described the way it will work, I can currently see this would align nicely as a self-contained simulation within the multi physics composition plans especially if the ` ModeSolver` and `ModeSolverData` is something we'll want to keep around to some level anyways.

Will have a more thorough look and think in a bit later!

tidy3d/components/mode/mode_solver.py
```diff
@@ -0,0 +1,2200 @@
...
+
+# Warning for field intensity at edges over total field intensity larger than this value
+FIELD_DECAY_CUTOFF = 1e-2
+
+# Maximum allowed size of the field data produced by the mode solver
+MAX_MODES_DATA_SIZE_GB = 20
+
+MODE_SIMULATION_TYPE = Union[Simulation, EMESimulation, ModeSimulation]
+MODE_SIMULATION_DATA_TYPE = Union[SimulationData, EMESimulationData, ModeSimulationData]
+MODE_PLANE_TYPE = Union[Box, ModeSource, ModeMonitor, ModeSolverMonitor]
```
daquinteroflex: Maybe we can name this in CamelCase to keep it standard with the rest of the solvers? Like `HeatChargeSimulationDataType`

tidy3d/components/mode/data/sim_data.py
```diff
@@ -0,0 +1,98 @@
...
+        title="Modes",
+        description=":class:`.ModeSolverData` containing the field and effective index data.",
+    )
+
+    data: Tuple[None, ...] = pd.Field(
+        (),
+        title="Monitor Data",
+        description="List of monitor data "
+        "associated with the monitors of the original :class:`.ModeSimulation`.",
+    )
```
daquinteroflex, line 37: Again apologies if this is stupid, I'm still getting familiar with everything in more detail with the exiting implementation/examples.

>The way it is currently set up, a `ModeSimulation` corresponds to a single run of the ModeSolver. If you wanted to run multiple mode solves in different planes in the same 3D simulation, you would need to use a Batch. This is different from an FDTD simulation, where you could have as many ModeSolverMonitor objects as you wanted in a single simulation. Setting it up this way also aligns with the philosophy of minimally extending the ModeSolver and not breaking anything.

Kind of curious how this `data` would extend conceptually in terms of I guess in a `ModeSimulation` . Say we're solving for four modes at multiple frequencies in the same plane in a `ModeSimulation` (this corresponds to a single `ModeSolver.run()` I understand), I guess this `data` would be indexed eventually to each of the `mode in modes` and `freq in freqs`.  But then I'm curious how this fits in with for example our say `PermittivityMonitor`s -> `PermittivityData` given this also gets defined for a set of `freqs` as for an FDTD simulation. Part of me feels we're abstracting each individual computational task in an array, but then now we're trying to integrate them with different simulation flows how can this be managed better?

I guess the ethos would be to maintain the way that `monitors -> data` is defined similar to `FDTDSimulation`, but maybe we're running into problems that the way we define and index the parameters in our simulation data is inherently different for different simulations such as a `ModeSimulation` or `FDTDSimulation`.  And so I'm wondering about interoperability between any existing classes in this sense. 

 I don't know, I'm still grasping how this all works so apologies if this is something obvious I'm just not super familiar with.

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,462 @@
...
+    "simulation_type",
+    "post_norm",
+]
+
+
+class ModeSimulation(AbstractYeeGridSimulation):
+    """
+    Simulation class for solving electromagnetic eigenmodes in a 2D plane with
+    translational invariance in the third dimension.
+
```
daquinteroflex: It'd be nice if we could add the equations and limitations of this too! I'll see if I can contribute to this based on the notebooks we already have, but think it'd be good to have it all consolidated in the same place on the lines of how we did the [HeatChargeSimulation](https://docs.flexcompute.com/projects/tidy3d/en/v2.8.0rc2/api/_autosummary/tidy3d.HeatChargeSimulation.html) docs

caseyflex: Just to clarify, none of the Mode Solver code is new, I just moved it from plugins to components

caseyflex:
> Based on the code and how you've described the way it will work, I can't currently see any reason why this wouldn't align nicely as a self-contained simulation within the multi physics composition plans especially if the ` ModeSolver` and `ModeSolverData` is something we'll want to keep around to some level anyways.

I think the `ModeSolverData` will definitely stay around. I think the `ModeSolver` interface will be deprecated eventually in favor of `ModeSimulation`, but the internal workings can stay around. In anticipation of this, I think I will make the `ModeSimulation.mode_solver` -> `ModeSimulation._mode_solver` so it's clearly only for internal use.

tidy3d/components/mode/simulation.py
daquinteroflex: Yeah no I know, this is mainly just a documentation improvement we could possibly do for new users. Happy to contribute to this in either this or another PR though.

caseyflex: but yeah, I can elaborate a bit here

tidy3d/components/mode/mode_solver.py
caseyflex: I just moved this code from plugins to components, I don't think there's any reason to change it at the moment as it's basically "internal workings" code

tidy3d/components/mode/data/sim_data.py
caseyflex: Yeah, the general set up is `monitors` -> `data` which is inspired by FDTD. In EME, we keep that setup for things like field monitors, mode solver monitors, etc, but we add the field `EMESimulationData.smatrix` because we figure that in every case a user runs an EME simulation they will want to know the smatrix.

I thought it was similar here, like there could be PermittivityMonitors and maybe other monitors down the line, but in every case a user runs a `ModeSimulation`, they will primarily be interested in `ModeSimulationData.modes`, the result of the mode solve.

caseyflex:
> Nice work! A subpixel-related question once we have a standalone mode simulation class:
> 
> * Mode solver is 2D, while FDTD is 3D. So people will tend to define the simulation domain size and structures to be 0 in out-of-plane dimension, or structures terminate exactly at the modal plane. However, right now we are still perform volumetric averaging, so you might find the results totally off. In principle, we can apply area averaging, but that can generate different results when we run mode solver as part of FDTD.

Good point. Currently we warn if structures extend exactly to the simulation edge, would that catch this case and warn the user that they should extend infinitely in the out-of-plane dimension? If I check that this warning is triggering properly, do you think that would give consistent results (extending the structures infinitely)?

That's all for a 2D ModeSimulation, note that 3D ModeSimulation with specified plane is also supported

tidy3d/components/mode/mode_solver.py
caseyflex: but sure, can't hurt to rename it

tidy3d/components/mode/simulation.py
caseyflex: maybe I'll just copy over the `ModeSolver` docstring "See Also" for now, we can add equations later but again, it's at least as clear as the existing mode solver

tidy3d/components/mode/mode_solver.py
caseyflex: actually it's easier to leave it for now

tidy3d/components/mode/__init__.py
daquinteroflex, line 1: Nice structure! Goes towards the self-contained "solver" structure

daquinteroflex (approved):
I've had another pass and seems broadly good to me and well implemented! I'm still getting more familiar with how this all works anyways and the conversation with Casey was very helpful to understand how this all fits in progressively. As far as I understand this aligns with the multiphysics structure

tidy3d/web/api/webapi.py
```diff
@@ -148,6 +150,10 @@ def run(
     :meth:`tidy3d.web.api.container.Batch.monitor`
         Monitor progress of each of the running tasks.
     """
+    is_mode_sim = isinstance(simulation, ModeSimulation)
+    if is_mode_sim:
+        mode_sim = simulation
+        simulation = simulation._mode_solver.updated_copy(simulation=simulation._as_fdtd_sim)
```
momchil-flex: Note that this also needs to happen in `upload` as it can be called independently from `run`.

I also have some broader thoughts about how to handle web stuff in my general PR comment.

tidy3d/web/api/webapi.py
```diff
@@ -166,9 +172,12 @@ def run(
         worker_group=worker_group,
     )
     monitor(task_id, verbose=verbose)
-    return load(
+    sim_data = load(
         task_id=task_id, path=path, verbose=verbose, progress_callback=progress_callback_download
     )
+    if is_mode_sim:
+        sim_data = ModeSimulationData(simulation=mode_sim, modes=sim_data)
+    return sim_data
```
momchil-flex: Similarly the `load` function should be changed directly so as it can be called in a standalone way outside of `web.run`.

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,475 @@
...
+        description="Specifications for the simulation grid along each of the three directions.",
+        validate_default=True,
+    )
+
+    plane: MODE_PLANE_TYPE = pd.Field(
+        None,
+        title="Plane",
+        description="Cross-sectional plane in which the mode will be computed. "
+        "If ``None``, the simulation must be 2D, and the plane will be the entire "
+        "simulation geometry.",
```
momchil-flex, line 178: If provided, the computational plane will be the intersection between the provided `plane`, and the simulation geometry.

tidy3d/components/mode/simulation.py
momchil-flex: Also add some notes about the interaction between the simulation size and the `plane`, and maybe mention here already the meaning of the `boundary_spec` for the mode simulation. We probably also need to explain how the `symmetry` argument works. [Namely](https://github.com/flexcompute/tidy3d/blob/078e56c0450906677677af616d1d797843294e3f/tidy3d/components/mode/mode_solver.py#L251-L259), it is only applied if the plane center and the simulation center coincide.

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,475 @@
...
+        description="Tuple of monitors in the simulation. "
+        "Note: monitor names are used to access data after simulation is run.",
+    )
+
+    sources: Tuple[None, ...] = pd.Field(
+        (),
+        title="Sources",
+        description="Sources in the simulation. Note: sources are not supported in mode "
+        "simulations.",
+    )
```
momchil-flex, line 163: Why have this at all? It shouldn't be needed for the grid since it is computed from `self._as_fdtd_sim`.

Edit: after looking through the code more, I think the reason is so that the auto grid obtained from `from_simulation` works correctly.  I am inclined to recommend removing this field and instead setting the `wavelength` in the auto grid when doing this construction, based on the sources if it's an FDTD simulation. However I'm not sure if this has implications I'm missing.

tidy3d/components/mode/simulation.py
```diff
@@ -0,0 +1,475 @@
...
+
+    grid_spec: GridSpec = pd.Field(
+        GridSpec(),
+        title="Grid Specification",
+        description="Specifications for the simulation grid along each of the three directions.",
+        validate_default=True,
+    )
```
momchil-flex, line 169: So this adds `validate_default=True` compared to what's inherited from `AbstractYeeGridSimulation`, what's the logic exactly?

momchil-flex (commented):
Note: I was starting to get tangled in all my thoughts so wanted to share this and hear your feedback, I hope most of it makes sense. :)

I think it's clear that this PR is a good initial step. It fixes a few issues as outlined in the top comment and allows for easy creation of both standalone ModeSimulation-s and mode simulations that are coupled to FDTD sims (previously, `ModeSover(simulation=... **kwargs)`, now `ModeSimulation.from_simulation(simulation=..., **kwargs)`)

I will try to summarize a few things that it seems to me are worth discussing further, either as part of this PR but more likely in terms of how they should be handled in the future. As a general point, a lot of the "pain" comes from the tight coupling between the FDTD simulations and the mode solver, specifically the use case in which a mode solver is placed in an FDTD simulation and has currently been made to use the exact same grid, and also to only span an intersection between the mode solver plane and the FDTD simulation domain. I am not sure if this was really mandatory to get good accuracy. Back in the day, a long time ago, I *seem* to remember seeing e.g. cleaner injection by a ModeSource when the solver is defined on the FDTD grid, e.g. as measured by the backward-injected power (which should be 0 ideally). But I am not 100% sure that this cannot be improved with other means even if the grids are mismatched. It would be much cleaner if we could separate the mode simulation from any FDTD details, but I think at the very least given that they are currently so tightly bound, we may just have to work around it anyway.

- Is there a good reason to allow `ModeSolver.simulation` to be a `ModeSimulation`? It seems to me that it would be nice if `ModeSimulation` entirely wraps `ModeSolver`, as opposed to the two-way coupling that currently happens. `ModeSimulation._mode_solver` could create a mode solver with `simulation = ModeSimulation._as_fdtd_sim`. No import of anything `ModeSimulation` related in `ModeSolver`. I don't know if you had anything in mind when implementing the circular dependence, but one issue that I do see is related to my code comments around the `web` calls, e.g. when trying to load from web just through a task_id, currently this would always be downloading a `ModeSolverData`, so having the ability to know if this was constructed using a `ModeSimulation`, and to reconstruct a `ModeSimulationData` from it, may require storing the `ModeSimulation`. So this is something that's not currently happening in the code, but I think it needs to. But is there another reason I'm not seeing?

  If not, and we only need this so that we can keep track in `web` functions, I think the better solution really would be to implement the new `MODE` task type (as opposed to `MODE_SOLVER`), upload `ModeSimulation` directly, and download `ModeSimulationData` directly - like you had already set up in your previous PR. This would achieve a few things: a) eliminate the cirular imports if we can make `ModeSolver` agnostic of `ModeSimulation`; b) avoid some confusion users may sometimes hit in which something they submitted as `ModeSimulation` has somehow converted to a `ModeSolver`: for example, if they go to GUI and download the simulation file and c) it is just likely that's how we just want to handle in on the long run, so question is only how soon we want to go for it. Regarding GUI, if we were to go for this approach, it would mean that `ModeSimulation` would *not* be direclty supported, until the GUI team implements support for that specific schema. However, I kind of feel like within the current setup in which we convert to `ModeSolver`, they will have no incentive to start working on it, or even good means for testing. So one proposal that I'm thinking about is this: we make the switch already to `task_type="MODE"`, and adjust pipelines and everything correspondingly. We don't yet broadly advertise `ModeSimulation` (e.g. we won't modify our notebook examples for now to use it) while we keep polishing rough edges if any, and while GUI works on support for it. Once that is done, we deprecate direct usage of `ModeSolver` in favor of `ModeSimulation`. We should discuss further with Emerson and MC.

- Potential confusion between the boundaries of `ModeSimulation` and the `plane`. Don't see much we can do about it now except add some more explanation about it in the documentation.

- Potential confusion between `ModeSimulation.boundary_spec` and `ModeSimulation.mode_spec.num_pml`. Again, we probably just need to accept this for now. But as a separate point we probably want to deprecate `num_pml` in favor of a `ModeSpec.boundary_spec`, for more fine-grained control of the PML coefficients as well as to allow setting e.g. Periodic or Bloch boundaries when we allow for that (e.g. to analyze gratings).

- What to do with the auto grid. I do feel like it would be convenient to not have to supply wavelength when using the `ModeSimulation` as a standalone tool. I know we discussed this a bit above but trying to understand a use case in which taking e.g. `td.C_0 / max(freqs)` by default will be problematic. If created `from_simulation`, we can make it so that the wavelength is set as needed (see my comment on `sources`), so the default will not take effect. Is there any problem then?

- Finally, again longer-term thinking, but do we eventually want to get rid of `ModeSpec` altogether, or more generally how does this play in with our multi-physics plans. I think the mode monitor / mode source is a perfect example of the types of coupling between solvers that we need to handle more and more. You have a component in one simulation that requires data from another simulation. The way this was set up is for the `ModeSpec` to kind of contain all the information that is needed by the mode solver, and this can be passed to FDTD components like the mode monitor and mode source so it's clear how to define the mode simulation. It seems to me that this assumes a sort of solver heirarchy, in which the FDTD simulation sits on top and the mode simulations are just auxiliary to it, which is also what led to some of the problems we're dealing with here, namely a mode simulation relying too heavily on a notion of some FDTD simulation that "contains it". Presumably, when moving to a `MultiPhysicsSimulation`, we want solvers to be on equal footing. What could be an alternative way to devine `ModeSource` and `ModeMonitor` in such a framework? @daquintero

tidy3d/components/mode/simulation.py
caseyflex: Actually, it's not related to auto grid. The reason is just that the AbstractSimulation class has this field, so we have this field even though we don't allow sources.

tidy3d/components/mode/simulation.py
caseyflex: I copied from EME simulation, I don't remember why it was there. I can try removing it.

caseyflex:
@momchil-flex Thanks for the very detailed feedback.

I do think it can be important to match the mode solver grid to the FDTD grid in some circumstances.

1. We could change it so `ModeSolver.simulation` cannot be a `ModeSimulation`. Currently `_as_fdtd_sim` is only used to generate the grid to get it to match exactly; the benefit of having `ModeSolver.simulation` be a `ModeSimulation` is that the plotting functions will only show the relevant data. But, if instead of adding sources in `_as_fdtd_sim` to control auto grid generation, I modify the grid_spec directly, then I suppose the plotting of the FDTD sim would be fine already.

Previously when I tried to get a "MODE" pipeline separate from the "MODE_SOLVER" pipeline, I got some pushback from MC; they didn't want to have two pipelines for essentially equivalent tasks. The alternative is to extend the "MODE_SOLVER" pipeline to support `ModeSimulation` -> `ModeSimulationData`. But then the other parts would all need to be aware of that too (GUI, etc.)

The current webapi setup is definitely a temporary workaround. If we want something more permanent, I definitely need your help negotiating it.

2. I agree with all your other points. I will update the auto grid logic as you suggest.

tidy3d/components/mode/simulation.py
momchil-flex: Oh I see! Then at least maybe the type can be something like `Literal[Tuple[]]`, I mean something that prevents anything that's not `()` from being passed? I'm not sure what the exact syntax would be.

tidy3d/web/api/webapi.py
```diff
@@ -257,6 +264,7 @@ def upload(
             console.log(f"View task using web UI at [link={url}]'{url}'[/link].")
 
     remote_sim_file = SIM_FILE_HDF5_GZ
+    # for ModeSimulation, keep original simulation. User can manually reduce if needed.
```
momchil-flex: why?

caseyflex: My reasoning was that the resulting sim_data.simulation will be whatever was uploaded to the server here, and it may be surprising if that is the reduced_simulation_copy. If the user wants to upload a smaller simulation, they would call the function reduced_simulation_copy explicitly, and then they would expect the sim_data to only know the reduced simulation.

open for discussion though

momchil-flex: Ah, I see. It's a good point, however the reason for generally trying to reduce the simulation is because in various cases (specifically with custom media, e.g. when you do a thermal simulation, get some medium perturbation over the entire simulation domain, and then want to do a mode solve in some specific location), the simulation file can be quite large and uploading it in its entirety is quite a waste (and then downloading it back in the `ModeSimulationData`...)

I think given that the user can turn off the reduction in the `web.run` arguments, we should still by default reduce as we do for `ModeSolver`. What do you think?

caseyflex: Yeah, that makes sense! I’ll update the webapi then to do that

momchil-flex (approved):

e-g-melo:
@caseyflex @momchil-flex, some users have complained about the PML position within the mode plane. They would expect to see it outward the mode plane. I understand that it is related to some technical challenges, but I wonder if we could consider this as an improvement for the next months.

caseyflex:
> @caseyflex @momchil-flex, some users have complained about the PML position within the mode plane. They would expect to see it outward the mode plane. I understand that it is related to some technical challenges, but I wonder if we could consider this as an improvement for the next months.

Is this just for ModeSimulation or is this for ModeSolver also? (I thought the treatment was the same). Can you explain a bit more about what you mean, or give an example where the behavior is not as expected?

e-g-melo:
> Is this just for ModeSimulation or is this for ModeSolver also? (I thought the treatment was the same). Can you explain a bit more about what you mean, or give an example where the behavior is not as expected?

Currently, when we set the PML, it is placed within the simulation plane. For example:

image

So, if the user set a large number of PML layers, we can have situations like this:

image

Users would expect the PML surrounding the mode plane.

image

caseyflex:
@e-g-melo thanks for the explanation. So this is unrelated to this PR and just an issue with ModeSolver right? This is the behavior of num_pml in ModeSpec? And usually we don’t see it because the simulation has its own pml or at least sufficient room between device and boundary. Could you create a new github issue?

e-g-melo:
Yes, it is related to num_pml. I will create a new issue. Thanks!

PR #2211, opened 2025-01-28
maint[tests]: constrain number of cores used for parallel testing

yaugenst-flex:
This PR fixes some issues with auto-assignment of `pytest-xdist` workers by:
 1. Limiting the no. of cores used based on available system memory (we need ~1GB per core)
 2. Keeping one core unused as a courtesy to other programs :smirk:
 3. Not doing 2. if running in CI

Note: We already install `psutil` because it's a dependency of `memory_profiler` and `ipykernel`, so adding it to `pyproject.toml` is mostly a formality.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,8 @@
+import os
+
 import matplotlib.pyplot as plt
 import numpy as np
+import psutil
 import pytest
 import tidy3d as td
 from tidy3d.log import DEFAULT_LEVEL, set_logging_console, set_logging_level
@@ -23,3 +26,26 @@ def reset_logger():
         del td.log.handlers["console"]
     set_logging_console()
     set_logging_level(DEFAULT_LEVEL)
+
+
+def pytest_xdist_auto_num_workers(config):
+    """Return the number of workers for pytest-xdist auto mode based on CPU cores and available memory.
+
+    Each worker requires approximately 1GB of memory, so the number of workers is limited by both
+    the number of physical CPU cores and available system memory.
+    """
+    try:
+        cores = psutil.cpu_count(logical=False)
+    except Exception:
+        cores = os.cpu_count()
+
+    available_mem_gb = psutil.virtual_memory().available / (1024**3)
+
+    # allow 1.2gb per core to provide some buffer
+    mem_limited_cores = int(available_mem_gb / 1.2)
+
+    cores = min(cores, mem_limited_cores)
+
+    if os.getenv("GITHUB_ACTIONS"):
+        return cores
+    return max(1, cores - 1)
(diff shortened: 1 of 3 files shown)


yaugenst-flex:
@daquinteroflex @momchil-flex could you review if you get the chance, this fixes some of the testing issues people have been experiencing

tests/conftest.py
```diff
@@ -23,3 +26,26 @@ def reset_logger():
...
+def pytest_xdist_auto_num_workers(config):
+    """Return the number of workers for pytest-xdist auto mode based on CPU cores and available memory.
+
+    Each worker requires approximately 1GB of memory, so the number of workers is limited by both
+    the number of physical CPU cores and available system memory.
+    """
+    try:
+        cores = psutil.cpu_count(logical=False)
+    except Exception:
+        cores = os.cpu_count()
```
daquintero, line 40: Nice! I'm interested why they're not the same actually.

daquinteroflex, line 40: Nice! I'm interested why they're not the same actually.

daquinteroflex (approved):
Looks good to me broadly!

tests/conftest.py
yaugenst-flex: because `os.cpu_count()` does not account for hyperthreading

tests/conftest.py
```diff
@@ -23,3 +26,26 @@ def reset_logger():
...
     set_logging_console()
     set_logging_level(DEFAULT_LEVEL)
+
+
+def pytest_xdist_auto_num_workers(config):
+    """Return the number of workers for pytest-xdist auto mode based on CPU cores and available memory.
+
+    Each worker requires approximately 1GB of memory, so the number of workers is limited by both
+    the number of physical CPU cores and available system memory.
+    """
```
momchil-flex, line 36: What's the 1GB of memory per worker estimate based on? Is it system dependent? Does it depend on the test suite?

momchil-flex (commented):
I believe this will not change anything for me because `psutil.cpu_count(logical=False)` returns 16, and I have 32GB of memory, so it will still use 16 workers, which is what seems to freeze my system already.

tests/conftest.py
yaugenst-flex: Yeah just from running the test suite. It's also something we might want to look into to optimize in the future, don't think it should be that much...

yaugenst-flex:
Yeah, this is based on some other reports too, but I have not been able to reproduce that cpu count issue that you're seeing. Not sure it's fixable in the tests...

momchil-flex (approved):
Maybe when my computer freezes it was because a lot of the RAM was already in use actually, so maybe this will help my case too. Anyway, better than nothing!

PR #2212, opened 2025-01-29
Doping 0-size message and bugs

marc-flex:
- confusing error message 
- Fix bug in scene plot for doping
- Make some semiconductor quantities required input fields

### DIFF
--- a/tests/test_components/test_heat_charge.py
+++ b/tests/test_components/test_heat_charge.py
@@ -1183,7 +1183,15 @@ def test_2D_doping_box():
     _ = td.ConstantDoping(size=(1, 1, np.inf), concentration=1)
 
     with pytest.raises(pd.ValidationError):
+        _ = td.ConstantDoping(size=(0, 1, 1), concentration=1)
+
+    with pytest.raises(pd.ValidationError):
+        _ = td.ConstantDoping(size=(1, 0, 1), concentration=1)
+
+    with pytest.raises(pd.ValidationError):
         _ = td.ConstantDoping(size=(1, 1, 0), concentration=1)
+
+    _ = td.ConstantDoping.from_bounds(rmin=(-td.inf, -1, -1), rmax=(td.inf, 1, 1), concentration=1)
 
 
 def test_edge_case_boundary_conditions():
--- a/tidy3d/components/tcad/doping.py
+++ b/tidy3d/components/tcad/doping.py
@@ -78,11 +78,12 @@ class AbstractDopingBox(Box):
         size = values["size"]
         for dim in range(3):
             if size[dim] == 0:
-                dim_name = "xyz"[dim]
+                zero_dim_name = "xyz"[dim]
+
                 raise SetupError(
-                    f"The doping box has been set up with 0 size in the {dim_name} direction. "
-                    "If this was intended to be translationally invariant, the box must have inf size in the "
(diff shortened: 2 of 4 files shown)


momchil-flex (approved):

tidy3d/components/tcad/doping.py
```diff
@@ -76,14 +76,28 @@ def check_dimensions(cls, values):
...
+            for n, dim in enumerate("xyz"):
+                if dim == zero_dim_name:
+                    size_str = size_str + "td.inf"
+                else:
+                    size_str = size_str + dim
+                if n == 2:
+                    size_str = size_str + ")"
+                else:
+                    size_str = size_str + " ,"
+
```
momchil-flex: Sorry I clicked approve but actually I don't really follow the logic here...

marc-flex: This is so that the user can read the output as `(x, y, z)` with `td.inf` wherever the user has a 0

momchil-flex: Sorry, still not getting it. It seems that `zero_dim_name` is exactly one of x, y, or z. And we already print it in the first sentence of the error message. So can't the `size_str` just read `td.inf`?

marc-flex: Yes, so basically, if `y = 0` the message would print something like ` ...by setting the size to (x, td.inf, z)` instead of just `(x, y, z)`

momchil-flex: Ohh. It seems a bit over-engineered honestly. How about just something like `If this was intended to be translationally invariant, the box must have a large or infinite ('td.inf') size in the perpendicular direction.` ?

marc-flex: :joy: Yeah OK. I just wanted to be super explicit.

tidy3d/components/tcad/doping.py
```diff
@@ -78,11 +78,12 @@ def check_dimensions(cls, values):
...
-                dim_name = "xyz"[dim]
+                zero_dim_name = "xyz"[dim]
+
                 raise SetupError(
-                    f"The doping box has been set up with 0 size in the {dim_name} direction. "
-                    "If this was intended to be translationally invariant, the box must have inf size in the "
-                    "perpendicular direction."
+                    f"The doping box has been set up with 0 size in the {zero_dim_name} direction. "
+                    "If this was intended to be translationally invariant, the box must have a large "
+                    "or infinite ('td.inf') size in the perpendicular direction."
```
daquinteroflex, line 86: nice!

tests/test_components/test_heat_charge.py
```diff
@@ -1182,6 +1182,12 @@ def test_2D_doping_box():
 
     _ = td.ConstantDoping(size=(1, 1, np.inf), concentration=1)
 
+    with pytest.raises(pd.ValidationError):
+        _ = td.ConstantDoping(size=(0, 1, 1), concentration=1)
+
+    with pytest.raises(pd.ValidationError):
```
daquinteroflex, line 1188: Maybe you can add a few tests `.from_bounds` like the notebook to cover that too?

marc-flex: good idea!

daquinteroflex (approved):

PR #2217, opened 2025-01-30
fix bug in the resistor and coax resistor variants of lumped elements

dmarek-flex:
Recent PR merge broke the existing implementations of `LumpedResistor` and `CoaxialLumpedResistor`. Fixed and also expanded unit tests.

I didn't add CHANGELOG entry because it was introduced by previous merge.


daquinteroflex (approved):
Looks reasonable, do you think these tests cover all the previous backwards compatibility cases?

dmarek-flex:
> Looks reasonable, do you think these tests cover all the previous backwards compatibility cases?

Yes, `LumpedResistor` and `CoaxialLumpedResistor` are the only previous implementations. The bug was happening in the volumetric conversion, which I wasn't testing directly previously.

weiliangjin2021 (approved):

PR #2219, opened 2025-01-31
Fix duplicate CI runs on non-develop branches

yaugenst-flex:
Currently the test workflow is triggered twice on `pre/*` because merging a PR is both a `push` as well as a `pull_request` event. Was confused why this doesn't happen on `develop` but apparently github automatically optimizes this away on default branches (but not on others). The more you know...


daquinteroflex (approved):
Good sir :raised_hands:

PR #2220, opened 2025-01-31
A meshing bug that skips the domain coordinate

weiliangjin2021:
The added test previously would error:

> AutoGrid coordinates along axis 2 do not match the simulation domain, indicating an unexpected error in the meshing. Please create a github issue so that the problem can be investigated. In the meantime, switch to uniform or custom grid along 2.

Now in `filter_min_step`, make sure that the last coordinate is kept.

### DIFF
--- a/tidy3d/components/grid/mesher.py
+++ b/tidy3d/components/grid/mesher.py
@@ -673,7 +673,9 @@ class GradedMesher(Mesher):
         coords_filter = [interval_coords[0]]
         steps_filter = []
         for coord_ind, coord in enumerate(interval_coords[1:]):
-            if coord - coords_filter[-1] >= min_step:
+            if coord - coords_filter[-1] >= min_step or (
+                coord_ind == len(interval_coords) - 2 and not isclose(coord, coords_filter[-1])
+            ):
                 coords_filter.append(coord)
                 steps_filter.append(max_steps[coord_ind])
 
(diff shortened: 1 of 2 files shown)


tidy3d/components/grid/mesher.py
```diff
@@ -673,7 +673,9 @@ def filter_min_step(
         coords_filter = [interval_coords[0]]
         steps_filter = []
         for coord_ind, coord in enumerate(interval_coords[1:]):
-            if coord - coords_filter[-1] >= min_step:
+            if coord - coords_filter[-1] >= min_step or (
+                coord_ind == len(interval_coords) - 2 and not isclose(coord, coords_filter[-1])
+            ):
```
weiliangjin2021, line 678: @momchil-flex i lost track here on why we are applying this method to filter small steps. It looks like handled already in `insert_bbox`?

momchil-flex: I guess at the very least there is this? https://github.com/flexcompute/tidy3d/blob/be7ededc7554a01e3319efa074c03a5107f735d2/tidy3d/components/grid/mesher.py#L669-L670

I do not remember either right now if this is the only reason or there was another use case. But it seems like it should be done even just because of this?

weiliangjin2021: i see, that makes sense

weiliangjin2021:
@momchil-flex good to merge?

PR #2221, opened 2025-02-03
User friendly error for wvl_mat_min when no sources present

groberts-flex:
Raise an error when the user tries to get the minimum wavelength in material, but there are no sources present in the simulation.

Added associated test to make sure we are catching this condition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -`HeatChargeSimulation` supersedes `HeatSimulation`. Though both of the can be used for Heat simulation interchangeably, the latter has been deprecated and will disappear in the future. 
 - `FluidSpec` and `SolidSpec` are now deprecated in favor of `FluidMedium` and `SolidMedium`. Both can still be used interchangeably.
 - Exclude `FluxMonitor` frequencies from adjoint design region gradient monitors since we do not differentiate through `FluxData`
+- Check for empty source list in `wvl_mat_min` in `Simulation` and raise user friendly error if no sources exist in the simulation.
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed magnitude of gradient computation in `CustomMedium` by accounting properly for full volume element when permittivity data is defined over less dimensions than the medium.
 - Fixed key ordering in design plugin when returning `Result` from an optimization method run.
 - Bug in `Batch`, where duplicate folders might be created in the Web UI if the folder did not already exist.
+- Check for empty source list in `wvl_mat_min` in `Simulation` and raise user friendly error if no sources exist in the simulation.
```
yaugenst-flex: I think this should go under "Changed", since this only adds a message, but it's failing in the same way as before technically.

tidy3d/components/simulation.py
```diff
@@ -4642,6 +4642,12 @@ def wvl_mat_min(self) -> float:
         float
             Minimum wavelength in the material (microns).
         """
+        if len(self.sources) == 0:
```
yaugenst-flex, line 4645: Nitpick, but the more "idiomatic" way would be `if not self.sources`. But actually let's keep it like this, I think it's more explicit.

yaugenst-flex (approved):
Thanks @groberts-flex!

tidy3d/components/simulation.py
groberts-flex: sounds good, good to know though!

CHANGELOG.md
groberts-flex: I see, that makes sense!

PR #2223, opened 2025-02-03
Grid snapping off for unshadowed MeshOverrideStructure unless it refines mesh

weiliangjin2021:
Also fixed a bug.

The main motivation is that when the automatic mesh refinement structure is having a larger grid size, e.g. larger than the grid size from substrate material, we shouldn't snap its bounding box. This override structure should take no effect at all in meshing. To achieve that, now I shift all those unshadowed structures to the beginning of structure list (but after simulation structure) so that it won't affect intervals generated by other structures. The main logic for not inserting coordinates of bounding box is added to `insert_bbox`.

## Example

Consider a substrate of permittivity = 4.4, and default mesh refinement factor = 2, now the auto-generated override structures take no effect:
image

If the refinement factor is 3, grids are snapped, and grid size refined:
image

### DIFF
--- a/tests/test_components/test_meshgenerate.py
+++ b/tests/test_components/test_meshgenerate.py
@@ -855,3 +855,43 @@ def test_override_unshadowed():
     )
     sizes = sim_unshadow.grid.sizes.to_list[2]
     assert sizes[len(sizes) // 2] < 0.1
+
+
+def test_override_unshadowed_snapping():
+    """Test that an unshadowed override structure won't snap to grid if it doesn't reduce grid size in the overriding
+    region.
+    """
+
+    sim = td.Simulation(
+        size=(3, 3, 6),
+        grid_spec=td.GridSpec.auto(wavelength=WAVELENGTH),
+        run_time=1e-13,
+        structures=[
+            BOX1,
+        ],
+    )
+
+    # override structure takes no effect (including bounding box snapping) if its grid size is
+    # larger than the existing ones in the overriding region.
+    override_structure = td.MeshOverrideStructure(
+        geometry=td.Box(size=(1, 1, 1)),
+        dl=[0.2, 0.2, 0.2],
+        shadow=False,
+    )
+    sim_shadow = sim.updated_copy(
+        grid_spec=td.GridSpec.auto(wavelength=WAVELENGTH, override_structures=[override_structure])
+    )
+    assert sim_shadow.num_cells == sim.num_cells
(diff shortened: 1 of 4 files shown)


tidy3d/components/grid/mesher.py
```diff
@@ -230,23 +230,23 @@ def parse_structures(
...
 
-        # Special attention needs to be paid to enforced overrideStructures.
-        # They shouldn't be overridden by other structures;
-        # for overlapping enforced structures, the grid size of the overlapped
-        # region is determined by the last enforced structure. We take two
-        # steps to implement the feature:
+        # Special attention needs to be paid to enforced or unshadowed overrideStructures.
+        # 1) enforced structures shouldn't be overridden by other structures; for overlapping enforced
+        #    structures, the grid size of the overlapped region is determined by the last enforced structure.
+        # 2) unshadowed structures doesn't override other structures based on the structure list order,
```
momchil-flex: don't override

tidy3d/components/grid/mesher.py
```diff
@@ -378,11 +388,14 @@ def insert_bbox(
...
+        structure_steps: ArrayFloat1D,
+        unshadowed: bool,
     ) -> Dict[str, List]:
         """Figure out where to place the bounding box coordinates of current structure.
         For both the left and the right bounds of the structure along the meshing direction,
-        we check if they are not too close to an already existing coordinate, and if the
-        structure is not completely covered by another structure at that location.
+        we check if they are not too close to an already existing coordinate, if the
+        structure is not completely covered by another structure at that location, and if
+        the structure is unshadowed and it refine grid size,
```
momchil-flex: it refines the grid size

momchil-flex (approved):
I think it makes sense.

tests/test_components/test_meshgenerate.py
```diff
@@ -855,3 +855,43 @@ def test_override_unshadowed():
...
+        size=(3, 3, 6),
+        grid_spec=td.GridSpec.auto(wavelength=WAVELENGTH),
+        run_time=1e-13,
+        structures=[
+            BOX1,
+        ],
+    )
+
+    # override structure takes no effect (including bounding box snapping) if its grid size is
+    # larger than the exisiting ones in the overriding region.
```
dmarek-flex: "existing"

dmarek-flex (approved):
Thanks for this fix/feature! I do find the terminology (`enforce` and `shadow`) a bit confusing, but that's outside the scope of this PR.

PR #2226, opened 2025-02-04
maint[web]: Use default S3 TransferConfig for up/downloads

yaugenst-flex:
The current `TransferConfig` limits chunking to 25kb and uses a pretty large no. of threads, which can become problematic in large batches. It's probably best to just use the S3 defaults ([see here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.TransferConfig)).

### DIFF
--- a/tidy3d/web/core/s3utils.py
+++ b/tidy3d/web/core/s3utils.py
@@ -173,12 +173,7 @@ def _get_progress(action: _S3Action):
     )
 
 
-_s3_config = TransferConfig(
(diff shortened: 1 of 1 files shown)


momchil-flex (approved):
It's fine by me but I don't understand this well.

QingengWei (approved):

yaugenst-flex:
Been testing this over the past week and haven't run into any issues, downloads seem to be consistently faster than before. Merging.

PR #2227, opened 2025-02-05
Bumping version to 2.8.0rc3

momchil-flex:


daquinteroflex (approved):

daquinteroflex:
The failed test looks like the sporadic mode solver issue yannick is fixing?

PR #2228, opened 2025-02-05
Array factor

dbochkov-flexcompute:
Currently I implemented this functionality as stand-alone functions in the `microwave` plugin. Specifically, it includes:
- `make_antenna_array`: converts a single antenna simulation into an array simulation
- `rectangular_antenna_array_factor`: comptutes array factor for 1D/2D/3D rectangular arrays
- `monitor_data_from_array_factor`: converts a single antenna field projection monitor data into one for an antenna array
- `simulation_data_from_array_factor`: coverts an entire simulation data for a single antenna into simulation data of an antenna array

Example of usage here:
https://github.com/flexcompute/tidy3d-notebooks/blob/daniil/array-factor/DipoleAntenna.ipynb

This lacks tests, but I wanted to first confirm if this organization/naming looks good. I guess, an alternative is to wrap these functions into a class, for example, `RectangularAntennaArrayCalculator`:
```
array_calculator = RectangularAntennaArrayCalculator(array_size=array_size, spacings=spacings, phase_shifts=phase_shifts)

array_factor = array_calculator.array_factor(theta=theta, phi=phi, f=freqs)
sim_array = array_calculator.make_array_simulation(sim_unit)
sim_array_data = array_calculator.simulation_data_from_array_factor(sim_unit_data)
```
this would be easier to generalize to `HexagonalAntennaArrayCalculator`, `CircularAntennaArrayCalculator`, etc

### DIFF
--- a/tests/test_plugins/test_array_factor.py
+++ b/tests/test_plugins/test_array_factor.py
@@ -1,0 +1,721 @@
+"""Test the array factor functions."""
+
+import numpy as np
+import pydantic.v1 as pydantic
+import pytest
+import tidy3d as td
+import tidy3d.plugins.microwave as mw
+import tidy3d.plugins.smatrix as smatrix
+
+from ..utils import AssertLogLevel
+
+
+def analytical_array_factor(array_size, spacings, phase_shifts, theta, phi, freqs, medium):
+    """Analytical array factor for a rectangular array."""
+    n, k = medium.nk_model(freqs)
+    k = 2 * np.pi * freqs * n / td.C_0
+    n_x, n_y, n_z = array_size
+    d_x, d_y, d_z = spacings
+    phi_x, phi_y, phi_z = phase_shifts
+    psi_x = k[None, :] * d_x * np.sin(theta[:, None]) * np.cos(phi[:, None]) - phi_x
+    psi_y = k[None, :] * d_y * np.sin(theta[:, None]) * np.sin(phi[:, None]) - phi_y
+    psi_z = k[None, :] * d_z * np.cos(theta[:, None]) - phi_z
+    af_x = np.sin(n_x * psi_x / 2) / np.sin(psi_x / 2)
+    af_y = np.sin(n_y * psi_y / 2) / np.sin(psi_y / 2)
+    af_z = np.sin(n_z * psi_z / 2) / np.sin(psi_z / 2)
+    af_exact = af_x * af_y * af_z
+    return af_exact
+
+
+def test_rectangular_array_calculator_basic():
+    """Test the array factor for a rectangular array."""
+
+    n_x = 1
+    n_y = 2
+    n_z = 3
+
+    d_x = 0.4
+    d_y = 0.5
+    d_z = 0.6
+
+    phi_x = np.pi / 6
+    phi_y = np.pi / 4
+    phi_z = np.pi / 3
+
+    array_calculator = mw.RectangularAntennaArrayCalculator(
+        array_size=(n_x, n_y, n_z),
+        spacings=(d_x, d_y, d_z),
+        phase_shifts=(phi_x, phi_y, phi_z),
+    )
+
+    # test ._antenna_locations
+    assert np.allclose(
+        array_calculator._antenna_locations,
+        np.array(
+            [
+                [0.0, -d_y / 2, -d_z],
+                [0.0, -d_y / 2, 0.0],
+                [0.0, -d_y / 2, d_z],
+                [0.0, d_y / 2, -d_z],
+                [0.0, d_y / 2, 0.0],
+                [0.0, d_y / 2, d_z],
+            ]
+        ),
+    )
+
+    # test ._antenna_amps
+    assert np.allclose(array_calculator._antenna_amps, np.ones(n_x * n_y * n_z))
+
+    # test ._antenna_phases
+    assert np.allclose(
+        array_calculator._antenna_phases,
+        np.array(
+            [
+                0,
+                phi_z,
+                2 * phi_z,
+                phi_y + 0,
+                phi_y + phi_z,
+                phi_y + 2 * phi_z,
+            ]
+        ),
+    )
+
+    # test ._antenna_nominal_size
+    assert np.allclose(
+        array_calculator._antenna_nominal_size,
+        np.array([(n_x - 1) * d_x, (n_y - 1) * d_y, (n_z - 1) * d_z]),
+    )
+
+    # test ._antenna_nominal_center
+    assert np.allclose(array_calculator._antenna_nominal_center, np.array([0, 0, 0]))
+
+    # test ._extend_dims
+    assert array_calculator._extend_dims == [1, 2]
+
+    with pytest.raises(pydantic.ValidationError):
+        # Test invalid array size
+        mw.RectangularAntennaArrayCalculator(
+            array_size=(0, 4, 5), spacings=(0.5, 0.5, 0.5), phase_shifts=(0, 0, 0)
+        )
+
+    with pytest.raises(pydantic.ValidationError):
+        # Test invalid spacings
+        mw.RectangularAntennaArrayCalculator(
+            array_size=(3, 4, 5), spacings=(-0.5, 0.5, 0.5), phase_shifts=(0, 0, 0)
+        )
+
+    with pytest.raises(pydantic.ValidationError):
+        # Test wrong length of amp_multipliers
+        mw.RectangularAntennaArrayCalculator(
+            array_size=(3, 4, 5),
+            spacings=(0.5, 0.5, 0.5),
+            phase_shifts=(0, 0, 0),
+            amp_multipliers=(0.5, 0.5),
+        )
+
+    for i in range(3):
+        with pytest.raises(pydantic.ValidationError):
+            # Test wrong length of amp_multipliers components
+            amps = [np.ones(3), np.ones(4), np.ones(5)]
+            amps[i] = np.ones(10)
+            mw.RectangularAntennaArrayCalculator(
+                array_size=(3, 4, 5),
+                spacings=(0.5, 0.5, 0.5),
+                phase_shifts=(0, 0, 0),
+                amp_multipliers=amps,
+            )
+
+
+def test_rectangular_array_calculator_array_factor():
+    """Test the array factor for a rectangular array."""
+
+    n_x = 1
+    n_y = 2
+    n_z = 3
+
+    d_x = 0.4
+    d_y = 0.5
+    d_z = 0.6
+
+    phi_x = np.pi / 6
+    phi_y = np.pi / 4
+    phi_z = np.pi / 3
+
+    array_calculator = mw.RectangularAntennaArrayCalculator(
+        array_size=(n_x, n_y, n_z),
+        spacings=(d_x, d_y, d_z),
+        phase_shifts=(phi_x, phi_y, phi_z),
+    )
+
+    # Test basic array factor calculation
+    theta = np.linspace(0, np.pi, 10)
+    phi = np.linspace(0, 2 * np.pi, 10)
+    theta_grid, phi_grid = np.meshgrid(theta, phi)
+    theta_grid = theta_grid.flatten()
+    phi_grid = phi_grid.flatten()
+    medium = td.Medium(permittivity=1)
+    freqs = np.array([1e9, 2e9, 3e9])
+
+    af = array_calculator.array_factor(theta_grid, phi_grid, freqs, medium)
+    assert af.shape == (100, 3)
+
+    af_exact = analytical_array_factor(
+        (n_x, n_y, n_z), (d_x, d_y, d_z), (phi_x, phi_y, phi_z), theta_grid, phi_grid, freqs, medium
+    )
+
+    assert np.allclose(np.abs(af), np.abs(af_exact))
+
+    # Test array factor with amplitude multipliers
+    array_calculator_amps = mw.RectangularAntennaArrayCalculator(
+        array_size=(n_x, n_y, n_z),
+        spacings=(d_x, d_y, d_z),
+        phase_shifts=(phi_x, phi_y, phi_z),
+        amp_multipliers=(0.5 * np.ones(n_x), 0.4 * np.ones(n_y), 0.3 * np.ones(n_z)),
+    )
+    af_amps = array_calculator_amps.array_factor(theta_grid, phi_grid, freqs, medium)
+    assert af_amps.shape == (100, 3)
+    assert np.allclose(np.abs(af_amps), np.abs(af_exact) * 0.5 * 0.4 * 0.3)
+
+    # Test array factor with non-uniform amplitude multipliers
+    # we will use amplitude multipliers that make the array effective size different from nominal size
+    n_x = 5
+    n_y = 8
+    n_z = 9
+    array_calculator_amps_nonuniform = mw.RectangularAntennaArrayCalculator(
+        array_size=(n_x, n_y, n_z),
+        spacings=(d_x, d_y, d_z),
+        phase_shifts=(phi_x, phi_y, phi_z),
+        amp_multipliers=(
+            np.int64(np.arange(n_x) % 2 == 0),
+            np.int64(np.arange(n_y) % 2 == 0),
+            np.int64(np.arange(n_z) % 2 == 0),
+        ),
+    )
+
+    af_amps_nonuniform = array_calculator_amps_nonuniform.array_factor(
+        theta_grid, phi_grid, freqs, medium
+    )
+    af_amps_nonuniform_exact = analytical_array_factor(
+        (3, 4, 5),
+        (2 * d_x, 2 * d_y, 2 * d_z),
+        (2 * phi_x, 2 * phi_y, 2 * phi_z),
+        theta_grid,
+        phi_grid,
+        freqs,
+        medium,
+    )
+    assert af_amps_nonuniform.shape == (100, 3)
+    assert np.allclose(np.abs(af_amps_nonuniform), np.abs(af_amps_nonuniform_exact))
+
+    # Test validation
+    with pytest.raises(ValueError):
+        # Test mismatched theta/phi lengths
+        array_calculator.array_factor(theta, phi[:5], freqs, medium)
+
+
+def make_antenna_sim():
+    # Scaling used for millimeters
+    mm = 1e3
+
+    # Frequency range
+    freq_start = 5e9
+    freq_stop = 15e9
+
+    freq0 = (freq_start + freq_stop) / 2  # Centeral frequency
+    freqs_target = np.linspace(freq_start, freq_stop, 5)
+    fwidth = freq_stop - freq_start  # Bandwidth
+
+    # Wavelength of centeral frequency in Vaccum
+    wavelength0 = td.C_0 / freq0
+
+    # Frequency sweep for S-parameters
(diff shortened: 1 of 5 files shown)


tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,430 @@
...
+
+    Parameters:
+    ----------
+    simulation : Simulation
+        The simulation specification describing a single antenna setup.
+    array_size : Tuple[int, int, int]
+        The number of antennas in the x, y, and z directions.
+    spacings : Tuple[float, float, float]
+        The spacing between antennas in the x, y, and z directions (in micrometers).
+    phase_shifts : Tuple[float, float, float]
```
dmarek-flex: You may want to allow nonuniform amplitudes for the array, e.g. Binomial Arrays

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,430 @@
...
+    ) -> ArrayLike:
+    """
+    Compute the array factor for a 3D antenna array.
+
+    Parameters:
+    -----------
+    array_size : Tuple[PositiveInt, PositiveInt, PositiveInt]
+        Number of elements in the x, y, and z directions.
+    spacings : Tuple[NonNegativeFloat, NonNegativeFloat, NonNegativeFloat]
+        Spacing between elements in the x, y, and z directions (in micrometers).
```
dmarek-flex: Again not sure if there are additional challenges for 2D or 3D arrays, but also allowing for nonuniform amplitudes should not be too hard

dmarek-flex:
I think the organization looks good. We do have a new `microwave` folder in `components` for microwave-specific operations. The operations you support in this PR seem like they would be a better fit for a plugin though.

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,430 @@
...
+                    # add the new lumped element to the list of lumped elements
+                    array_lumped_elements.append(new_lumped_element)
+
+                for override in simulation.grid_spec.override_structures:
+                    # create a copy of the original override by translating it in x, y, and z directions
+                    new_override = override.updated_copy(
+                        geometry=override.geometry.translated(*translation_vector).bounding_box
+                    )
+                    # add the new override to the list of overrides
+                    array_overrides.append(new_override)
```
weiliangjin2021: Now also need to duplicate layer_refinement_spec.

weiliangjin2021 (commented):
Nice work! Regarding the organization, I also feel that it can clearer with class, especially that those functions share lots of arguments. A few other comments:

- It seems that `make_antenna_array` can be used for two separate purposes: 1) set up a large simulation to simulate the complete phase array; 2) for creating the monitor data from single antenna simulation with array factor. There is also the hybrid usage of 1) and 2) when we want to set up an embeded antenna simulation where the antenna is embeded in an array, and you use that to multiply to the array factor.

  So we can probably separate it into two classes/functions based on the two purposes , but not duplicating the code as one can call the other through API? 


- Regarding angular sampling, shall we also check if the sampling density is not dense enough? Shall we add a symmetry variable, so that the angle doesn't need to span the entire range?

- Regarding this entire flux computation issue, I wonder if we can use adaptive integration internally. We keep an internal sampling angles that are separate from the angles in user-defined directivity monitor.

weiliangjin2021:
Maybe just even sampling in the 1st release, and enable adaptive or the spherical harmonic bases in the next release.

dbochkov-flexcompute:
> It seems that make_antenna_array can be used for two separate purposes: 1) set up a large simulation to simulate the complete phase array; 2) for creating the monitor data from single antenna simulation with array factor. There is also the hybrid usage of 1) and 2) when we want to set up an embeded antenna simulation where the antenna is embeded in an array, and you use that to multiply to the array factor.

how about we keep it a single class, just add a flag `duplicate_sources: bool = True` into `RectangularAntennaArrayCalculator.make_array_simulation(simulation: Simulation, duplicate_sources: bool = True)` which should cover the hybrid case?

weiliangjin2021:
> how about we keep it a single class, just add a flag `duplicate_sources: bool = True` into `RectangularAntennaArrayCalculator.make_array_simulation(simulation: Simulation, duplicate_sources: bool = True)` which should cover the hybrid case?

That should work.

dbochkov-flexcompute:
> > how about we keep it a single class, just add a flag `duplicate_sources: bool = True` into `RectangularAntennaArrayCalculator.make_array_simulation(simulation: Simulation, duplicate_sources: bool = True)` which should cover the hybrid case?
> 
> That should work.

Actually, what I ended up doing is implementing @dmarek-flex's suggestion on adding tapering amplitudes option (currently tapering along each axis independently) and that seems to cover this functionality as well. That is, to create a simulation with only one source we just need to set amplitudes of all sources except at a desired location to zero and those source will not be created. The reason I think that should be a good approach is that having just one `duplicate_sources: bool = True` flag is quite limiting. For example, in rectangular arrays it could be of interest to place source either in the middle of array, or on the edge, or anywhere in between to extract particular embedded array pattern. I could have alternatively added an argument `active_antenna: Union[Tulple[int, int, int], Literal["all"]] = "all"` into 
```make_antenna_array(simulation: Simulation, active_antenna: Union[Tulple[int, int, int], Literal["all"]] = "all")```, but in this case `make_antenna_array` would have had different signatures for different array arrangements (rectangular, hexagonal, circular, etc)

Overall, I refactored and organized this antenna calculator class so that it's ready for extension into different array types whenever is needed

> Regarding angular sampling, shall we also check if the sampling density is not dense enough? Shall we add a symmetry variable, so that the angle doesn't need to span the entire range?
>
> Regarding this entire flux computation issue, I wonder if we can use adaptive integration internally. We keep an internal sampling angles that are separate from the angles in user-defined directivity monitor.
>
> Maybe just even sampling in the 1st release, and enable adaptive or the spherical harmonic bases in the next release.

For now, I piggy-backed off of @dmarek-flex implementation of flux integration. But I agree that having harmonic expansion would be great in future

File: tidy3d/plugins/microwave/array_factor.py [hunk 4]
```diff
@@ -0,0 +1,661 @@
...
+                update_dict["flux"] = monitor_data.flux_from_projected_fields()
+            except ValueError as e:
+                log.warning(
+                    "Could not recalculate flux for a 'DirectivityData' due to the following "
+                    f"reason: {e} This monitor will not be included in the resulting data."
+                )
+                return None
+
+        else:
+            update_dict["projection_surfaces"] = monitor.projection_surfaces
```
dmarek-flex: Now `DirectivityData` inherits from `FieldProjectionAngleData`, so I think setting the projection surfaces should apply to `DirectivityData`  as well

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,661 @@
...
+    phase_shifts: Tuple[float, float, float] = pd.Field(
+        (0, 0, 0),
+        title="Phase Shifts",
+        description="Phase-shifts between antennas along x, y, and z directions.",
+    )
+
+    amp_multipliers: Tuple[Optional[ArrayLike], Optional[ArrayLike], Optional[ArrayLike]] = (
+        pd.Field(
+            (None, None, None),
+            title="Apmlitude Multipliers",
```
dmarek-flex: typo amplitude

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,661 @@
...
+        psi_z = k * self.spacings[2] * np.cos(theta_array) - self.phase_shifts[2]
+
+        amp_x = 1.0 if self.amp_multipliers[0] is None else self.amp_multipliers[0][:, None, None]
+        amp_y = 1.0 if self.amp_multipliers[1] is None else self.amp_multipliers[1][:, None, None]
+        amp_z = 1.0 if self.amp_multipliers[2] is None else self.amp_multipliers[2][:, None, None]
+
+        # Calculate the array factor in the x, y, and z directions
+        af_x = np.sum(
+            amp_x
+            * np.exp(-1j * np.arange(self.array_size[0])[:, None, None] * psi_x[np.newaxis, :]),
```
dmarek-flex, line 759: `None` is equivalent using `np.newaxis`? Just curious

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,661 @@
...
+        phase_shifts_per_dim = [
+            np.arange(self.array_size[dim]) * self.phase_shifts[dim] for dim in range(3)
+        ]
+
+        phase_shifts_grid = np.meshgrid(*phase_shifts_per_dim)
+
+        return np.ravel(sum(p for p in phase_shifts_grid))
+
+    @property
+    def _extend_dims(self):
```
dmarek-flex: add type hint

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,661 @@
...
+            for multiplier, size in zip(self.amp_multipliers, self.array_size)
+        ]
+
+        amps_grid = np.meshgrid(*amps_per_dim)
+
+        return np.ravel(amps_grid[0] * amps_grid[1] * amps_grid[2])
+
+    @property
+    def _antenna_phases(self) -> ArrayLike:
+        """Locations of antennas' centers in an array."""
```
dmarek-flex: docstring wrong

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,661 @@
...
+        y = (np.arange(self.array_size[1]) - (self.array_size[1] - 1) / 2) * self.spacings[1]
+        z = (np.arange(self.array_size[2]) - (self.array_size[2] - 1) / 2) * self.spacings[2]
+
+        X, Y, Z = np.meshgrid(x, y, z)
+
+        return np.transpose([X.ravel(), Y.ravel(), Z.ravel()])
+
+    @property
+    def _antenna_amps(self) -> ArrayLike:
+        """Locations of antennas' centers in an array."""
```
dmarek-flex: docstring wrong

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,661 @@
...
+                        f"Object {obj.name} (type: {obj.type}) extends beyond simulation bounds along"
+                        f" '{'xyz'[dim]}' axis. Please check your antenna setup for correctness."
+                    )
+                # update minimum and maximum bounds in each dimension
+                antenna_bounds[0][dim] = min(antenna_bounds[0][dim], obj_bounds[0][dim])
+                antenna_bounds[1][dim] = max(antenna_bounds[1][dim], obj_bounds[1][dim])
+
+        return antenna_bounds
+
+    def make_antenna_array(self, simulation: Simulation):
```
dmarek-flex: type hint

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,661 @@
...
+        in the input simulation and uses this information to compute the dimensions
+        of the resulting antenna array simulation. All structures, sources, lumped elements,
+        and mesh override structures are duplicated, while monitors are extended in size.
+        Only projection monitors are transferred into the resulting simulation.
+
+        Parameters:
+        ----------
+        simulation : Simulation
+            The simulation specification describing a single antenna setup.
+        active_antenna : bool = True
```
dmarek-flex: no longer present as parameter

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,661 @@
...
+                    )
+                    array_sources.append(new_source)
+
+            for lumped_element in simulation.lumped_elements:
+                new_lumped_element = lumped_element.updated_copy(
+                    center=tuple(lumped_element.center + translation_vector),
+                    name=f"{lumped_element.name}_{ind}",
+                )
+                array_lumped_elements.append(new_lumped_element)
+
```
dmarek-flex: We will need the ability to add "lumped ports" to a simulation without using the `TerminalComponentModeler` or add a new type of source. But that is for some other PR and this step should still work.

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,661 @@
...
+from ...components.monitor import AbstractFieldProjectionMonitor
+from ...components.simulation import Simulation
+from ...components.structure import Structure
+from ...components.types import ArrayLike
+from ...constants import C_0, inf
+from ...log import log
+
+
+class AbstractAntennaArrayCalculator(Tidy3dBaseModel, ABC):
+    """Abstract base for phased array calculators."""
```
dmarek-flex, line 27: I think it would be helpful to add an explanation of the two overall strategies for calculating the far field of the array. Either from array factor or from direct simulation of the entire array.  We could also add explanation for why a user would use one strategy of the other.

Not sure where to put this maybe in a notebook or as a docstring here.

tests/test_plugins/test_array_factor.py
```diff
@@ -0,0 +1,691 @@
...
+    # We integrate the base simulation with the LumpedPort using the TerminalComponentModeler.
+    modeler = smatrix.TerminalComponentModeler(
+        simulation=sim,
+        ports=[port],
+        freqs=freqs,
+        verbose=True,
+        remove_dc_component=False,  # Include DC component for more accuracy at low frequencies
+    )
+
+    sim_unit = list(modeler.sim_dict.values())[0]
```
dmarek-flex, line 365: Yea this is what I mean about lumped port in Simulation :smile:. It should be easier to do this!

dmarek-flex (commented):
It all looks great to me! Just some typos and missing type hints for the most part.

tidy3d/plugins/microwave/array_factor.py
dbochkov-flexcompute: yeah, I've used them interchangeably, but mostly `None` since it's less to type lol

tidy3d/plugins/microwave/array_factor.py
dbochkov-flexcompute: added this note in class description:
```
    In addition, this class provides a convenience method to create an antenna array simulation
    from a single antenna simulation. This can be used to compute the behavior (near-field and/or
    far-field) of the full antenna array directly without any approximations. Such a simulation setup
    can be obtained by directly calling the `make_antenna_array` function or from the `SimulationData`
    object returned by the `simulation_data_from_array_factor` method.
```

weiliangjin2021: Clearer with bullet point?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,668 @@
...
+                extends_beyond_min = obj_bounds[0][dim] < sim_bounds[0][dim]
+                extends_beyond_max = obj_bounds[1][dim] > sim_bounds[1][dim]
+                if extends_beyond_min or extends_beyond_max:
+                    # in case of a box, we just ignore it, since we will be able to extend it later
+                    if (
+                        isinstance(obj, Structure)
+                        and isinstance(obj.geometry, Box)
+                        and extends_beyond_min
+                        and extends_beyond_max
+                    ):
```
weiliangjin2021, line 100: The box is treated differently here. My understanding is that it's easier to adjust the size of a box. Does it mean the handling of other geometries has some caveats?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,668 @@
...
+                    or structure.geometry.bounds[1][dim] > sim_bounds[1][dim]
+                    for dim in extend_dims
+                )
+            )
+
+            if duplicate:
+                for ind, translation_vector in enumerate(locations):
+                    new_structure = structure.updated_copy(
+                        geometry=structure.geometry.translated(*translation_vector),
+                        name=None if structure.name is None else f"{structure.name}_{ind}",
```
weiliangjin2021: Maybe we don't need to worry about it this time, but custom medium also needs to be translated. Probably add a warning here for this release?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,668 @@
...
+
+                # add the new structure to the list of structures
+                array_structures.append(new_str)
+
+        for ind, (translation_vector, phase_shift, amp_multiplier) in enumerate(
+            zip(locations, phases, amps)
+        ):
+            if amp_multiplier != 0.0:
+                for source in simulation.sources:
+                    new_source = source.updated_copy(
```
weiliangjin2021: Just double check: custom source should be fine here, as the coordinate of the data is defined w.r.t. the center of the source.

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -0,0 +1,668 @@
...
+
+    The array factor is calculated using the standard array factor formula for rectangular arrays,
+    which accounts for the spatial distribution of antennas and their relative phases and amplitudes.
+    This can be used to analyze beam steering, sidelobe levels, and other array characteristics.
+
+    In addition, this class provides a convenience method to create an antenna array simulation
+    from a single antenna simulation. This can be used to compute the behavior (near-field and/or
+    far-field) of the full antenna array directly without any approximations. Such a simulation setup
+    can be obtained by directly calling the `make_antenna_array` function or from the `SimulationData`
+    object returned by the `simulation_data_from_array_factor` method.
```
weiliangjin2021: `simulation_data_from_array_factor` method is not a full simulation, but with the approximation?

weiliangjin2021 (commented):
`make_antenna_array`  contains quite many lines of code. Probably better to split it into several methods, e.g. `_extend_structures`, `_extend_monitors`, etc.

tidy3d/plugins/microwave/array_factor.py
dbochkov-flexcompute: yes, it converts `SimulationData` for one antenna into `SimulationData` for antenna array, and when doing so it modifies `SimulationData.simulation` accordingly. So, then someone can do `sim_array_direct_data = web.run(sim_array_data.simulation, task_name="exact_array")` to get exact results. I can more explanations about that

tidy3d/plugins/microwave/array_factor.py
dbochkov-flexcompute: I think it just doesn't apply to other geometries. With box, we only modify size instead of duplicating if we see that it is infinite in dimension that we extend antenna. For example, an infinite substrate. We could potentially to try to detect similar cases with other geometries, like a cylinder of a very large radius, or rectangular made as a polyslab, but that might significantly increase complexity of the code

tidy3d/plugins/microwave/array_factor.py
dbochkov-flexcompute: Thanks for catching that. Added a warning

tidy3d/plugins/microwave/array_factor.py
dbochkov-flexcompute: yeah, indeed, source fields are with respect to source center

dbochkov-flexcompute:
> make_antenna_array contains quite many lines of code. Probably better to split it into several methods, e.g. _extend_structures, _extend_monitors, etc.

Implemented this reorg. Also, realized that layer refinement specs, override structures, and lumped elements (this one is very unlikely I guess) could also followthe same logic as for usual box structures, that is, expand instead of duplicate.

weiliangjin2021 (approved):

dmarek-flex (approved):

PR #2230, opened 2025-02-07
Introducing AnalyticBeam components

momchil-flex:
This introduces `PlaneWaveAnalyticBeam`, `GaussianAnalyticBeam`, and `AstigmaticGaussianAnalyticBeam` components, which can be used to generate the fields that correspond to such beams in a given cross-section. The main use case for this is to compute overlap integrals with recorded field data.

I'm not a fan of the `Analytic` part in the name, but simply `GaussianBeam` is already the name of our source. Ideas for other names are welcome!

@tomflexcompute could you check if this works in your example, and maybe share it with me so we can add something related to it as a test?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `layer_refinement_specs` field in `GridSpec` that takes a list of `LayerRefinementSpec` for automatic mesh refinement and snapping in layered structures. Structure corners on the cross section perpendicular to layer thickness direction can be automatically identified. Mesh is automatically snapped and refined around those corners.
 - New field `drop_outside_sim` in `MeshOverrideStructure` to specify whether to drop an override structure if it is outside the simulation domain, but it overlaps with the simulation domain when projected to an axis.
 - Function `translated_copy` in `ElectromagneticFieldData` to facilitate computing overlaps between field data at different locations.
+- `PlaneWaveBeamProfile`, `GaussianBeamProfile` and `AstigmaticGaussianBeamProfile` components to compute field data associated to such beams based on their corresponding analytical expressions, and compute field overlaps with other data.
+- `num_freqs` argument to the `PlaneWave` source.
 
 ### Changed
 - The coordinate of snapping points in `GridSpec` can take value `None`, so that mesh can be selectively snapped only along certain dimensions.
 - Grid snapping for `MeshOverrideStructure` with `shadow=False` is only on if the structure refines the mesh.
 - Renamed `SkinDepthFitterParam` --> `SurfaceImpedanceFitterParam` used in `LossyMetalMedium`. `plot` in `LossyMetalMedium` now plots complex-valued surface impedance.
 - Changed plot_3d iframe url to tidy3d production environment.
+- `num_freqs` is now set to 3 by default for the `PlaneWave`, `GaussianBeam`, and `AnalyticGaussianBeam` sources, which makes the injection more accurate in broadband cases.
 
 ### Fixed
 - Make gauge selection for non-converged modes more robust.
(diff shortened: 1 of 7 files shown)


tomflexcompute:
For `freqs` if I pass in a list like `[freq0]` it will error when calculating `field_data`. Using a numpy array works. 

It seems like the propagation axis is fixed to `z`. Maybe we should allow it to change? 

Also do you want to add these to the `__init__.py` so users don't have to do `from tidy3d.components.beam import GaussianAnalyticBeam`?

momchil-flex:
> For `freqs` if I pass in a list like `[freq0]` it will error when calculating `field_data`. Using a numpy array works.

Ah ok! Will fix.

> It seems like the propagation axis is fixed to `z`. Maybe we should allow it to change?

No, just like for sources, the propagation axis is determined by the dimension along which the size is 0.

> Also do you want to add these to the `__init__.py` so users don't have to do `from tidy3d.components.beam import GaussianAnalyticBeam`?

Good point, and also to the api docs.

tomflexcompute:
> > It seems like the propagation axis is fixed to `z`. Maybe we should allow it to change?
> 
> No, just like for sources, the propagation axis is determined by the dimension along which the size is 0.
> 
Oh yeah indeed that's the case. 

Here is a simple demo that I made for a user and I modified the Gaussian beam part to use your analytical implementation. Results are good. Once this PR is merged I can polish this notebook and use it as a public tutorial. 
[ModeOverlap.zip](https://github.com/user-attachments/files/18709232/ModeOverlap.zip)

momchil-flex:
Updated with all revisions including a test that the overlap between two beams can be computed.

yaugenst-flex (approved):
Thanks @momchil-flex this is great, I was actually looking for functionality like this a while back :smile:

This all looks good to me, very minor question / comment I have is about the `self.size.index(0.0)` check... the `PlanarSource` has an [injection_axis](https://github.com/flexcompute/tidy3d/blob/898d1669883a83e6fc5668e2ca6ab11b0bcd736c/tidy3d/components/source.py#L702-L705) property, should we maybe have something similar here?

momchil-flex:
> Thanks @momchil-flex this is great, I was actually looking for functionality like this a while back 😄
> 
> This all looks good to me, very minor question / comment I have is about the `self.size.index(0.0)` check... the `PlanarSource` has an [injection_axis](https://github.com/flexcompute/tidy3d/blob/898d1669883a83e6fc5668e2ca6ab11b0bcd736c/tidy3d/components/source.py#L702-L705) property, should we maybe have something similar here?

Yeah.. not really sure. I think in this case is more because the 3D TFSF source also has an injection_axis, which is a user argument as it cannot be inferred by the size, so this property is a way to handle a default value + a user-supplied one. We do have this `size.index(0.0)` popping in various places though, like [here](https://github.com/flexcompute/tidy3d/blob/898d1669883a83e6fc5668e2ca6ab11b0bcd736c/tidy3d/plugins/mode/mode_solver.py#L185-L188) and [here](https://github.com/flexcompute/tidy3d/blob/898d1669883a83e6fc5668e2ca6ab11b0bcd736c/tidy3d/components/geometry/base.py#L2095), on one hand it's a fairly short thing to write but on the other it could be put into a property, just not sure where? In `Box` maybe but in principle it should only be used for 2D boxes, so also add a check for that in the poperty?

yaugenst-flex:
I don't think we necessarily _have_ to change it, just thought it might be a bit cleaner. And yeah `Box` seems like a good candidate, technically this already exists as `Box.zero_dims`. Not sure if that ends up being nicer to use in the end, but I wouldn't want to add a `normal_dir` to `Box` that only works for 2D boxes... so I guess instead of `self.size.index(0.0)` it could be `self.zero_dims[0]`. Maybe not worth it.

momchil-flex:
@tomflexcompute will wait for a final green light from you.

tomflexcompute:
> @tomflexcompute will wait for a final green light from you.

Looks great. Feel free to merge. I'll polish the tutorial notebook and PR that later this week.

tomflexcompute (approved):

PR #2231, opened 2025-02-07
:zap: Fossum model for carrier lifetime

marc-flex:
- [x] Added Fossum model for carrier lifetime
- [x] Split reference doping into n and p dopings (`CaugheyThomasMobility`)

This is the only thing that needed to be added to the front-end to improve the 2D results. 

FYI: There will soon be a back-end PR associated to this plus some other fixes


tidy3d/components/tcad/generation_recombination.py
```diff
@@ -1,6 +1,55 @@
...
+        ...   N0=7.1e15,
+        ...   A=1,
+        ...   B=0,
+        ...   C=1,
+        ...   alpha=1
+        ... )
+
+    """
+
+    tau300: pd.PositiveFloat = pd.Field(
```
daquinteroflex: Would propose `tau_300` for consistency, or should it even be at a reference temperate based on the simulation steady state temperature definition?

tidy3d/components/tcad/mobility.py
```diff
@@ -151,11 +155,18 @@ class CaugheyThomasMobility(Tidy3dBaseModel):
...
     )
 
-    # reference doping
-    ref_N: pd.PositiveFloat = pd.Field(
+    # reference doping n
+    ref_N_n: pd.PositiveFloat = pd.Field(
+        ...,
+        title="Reference doping for electrons",
+        description="Reference doping for electrons at reference temperature (300K) in #/cm^3.",
+    )
```
daquinteroflex: Nice! Yeah it's kind of related to our previous conversation whether n and p should have separate mobility models (ie different instances of the same `Mobility` subclass). This seems good though in line with our current approach!

daquinteroflex (approved):
Looking good!

tidy3d/components/tcad/mobility.py
marc-flex: At this point I think having two different models would be beneficial and since it's a pre-release we're still allowed to do it, right @momchil-flex?

momchil-flex: Yes, now's the time.

tidy3d/material_library/material_library.py
```diff
@@ -1886,18 +1886,25 @@ def medium(self, optical_axis: Axis):
...
+                mu_min=52.2,
+                mu=1471.0,
+                ref_N=9.68e16,
+                exp_N=0.68,
+                exp_1=-0.57,
+                exp_2=-2.33,
+                exp_3=2.4,
+                exp_4=-0.146,
+            ),
+            mobility_p=CaugheyThomasMobility(
```
daquinteroflex, line 1899: Nice! Sounds like you've already begun having a check that all the docstring examples also pass and the notebook with the new changes too.

tidy3d/components/tcad/mobility.py
```diff
@@ -58,23 +54,26 @@ class CaugheyThomasMobility(Tidy3dBaseModel):
...
+       * - :math:`\\alpha_N`
+         - ``exp_N``
+         - Exponent for doping dependence.
        * - :math:`\\alpha_4`
-         - ``exp_t_d_exp``
-         - Exponent for the temperature dependence of the exponent :math:`\\alpha_n` and :math:`\\alpha_p`
+         - ``exp_4``
+         - Exponent for the temperature dependence of the exponent :math:`\\alpha_N`
+       * - :math:`N_{ref}`
+         - ``ref_N_n``, ``ref_N_p``
```
momchil-flex: Should these be removed now?

momchil-flex (approved):
Looks good, but for my understanding where exactly do the specific values come from in the two separate mobility models now?

marc-flex:
>  where exactly do the specific values come from in the two separate mobility models now?

The Fossum parameters can be found in the original paper by Fossum et al. The mobility model is basically the same model as before but by splitting up the entry into n and p we avoid having to duplicate naming (e.g., mu_n, mu_p)

I'll add some references.

marc-flex:
@momchil-flex @daquinteroflex if you both are OK with this I'll merge by the end of the day

daquinteroflex:
Makes sense to me!

PR #2232, opened 2025-02-07
Allow setting of numerical tolerance for autograd check_grads function

groberts-flex:
Certain tests are failing because of tight numerical tolerances in autograds check_grads function (https://github.com/HIPS/autograd/blob/67f341fd108521cc2a0ec6e89796a0c6f800dd72/autograd/test_util.py). That test boils down eventually to a call to scalar_close at the top of the file which uses hardcoded tolerances defined as globals in the file. 

I'm not sure if this is the best way to go about it, but I've wrapped the function in check_grads such that we can override those variables, call autograd's function, and then reset the tolerances before returning. The default values I put as the same as what autograd is currently using.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,13 @@
 import os
 
+import autograd
 import matplotlib.pyplot as plt
 import numpy as np
 import psutil
 import pytest
 import tidy3d as td
+from autograd.test_util import check_grads
+from autograd.wrap_util import unary_to_nary
 from tidy3d.log import DEFAULT_LEVEL, set_logging_console, set_logging_level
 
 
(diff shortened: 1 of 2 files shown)


yaugenst-flex (commented):
Thanks @groberts-flex! This probably works but has some potential pitfalls where if multiple tests in the same session run `check_grads_with_tolerance`, the test that sets the tolerance last "wins". I'd suggest an alternative approach where we monkeypatch the tolerance and then use the regular `autograd.test_util.check_grads` within a context manager. This ensures that things are always scoped correctly.

So a fixture in `conftest.py`:
```python
@pytest.fixture
def with_autograd_tol(monkeypatch)
  def _apply_tol(tol=1e-6, rtol=1e-6):
    monkeypatch.setattr(autograd.test_util, "TOL", tol)
    monkeypatch.setattr(autograd.test_util, "RTOL", rtol)
    # monkeypatch automatically reverts at the end of fixture scope
  return _apply_tol
```

And then for usage in tests where the tolerance requires changing:
```python
def test_something(with_autograd_tol):
  with with_autograd_tol(tol=1e-3, rtol=1e-3):
    check_grads(...)(x)
```

Alternatively, if this concerns only a particular test, we could also use a monkeypatch context in the test explicitly like so:
```python
def test_something(monkeypatch):
  with monkeypatch.context() as m:
    m.setattr(autograd.test_util, "TOL", new_tol)
    check_grads(...)(x)
```

groberts-flex:
thanks @yaugenst-flex, this is really helpful! I ended up doing a little bit of a hybrid of these, let me know if this looks good to you. I think it's good to have it available for other tests in case there are more failures for tolerance, but I was having some issues getting the first approach to work in the form above. I was needing to decorate it as a context manager and then use `yield` in the function. I was less familiar with what was going on when I added those so I went with wrapping the second approach into a fixture.

yaugenst-flex (approved):
Thanks @groberts-flex looks good. Having a fixture return a callable is a bit unusual but it's functionally equivalent so I'm ok with it here.

PR #2233, opened 2025-02-07
Bug fix for unshadowed override structure extending outside the simulation domain

weiliangjin2021:


lucas-flexcompute (approved):
This works for my examples. I think that it looks good!

PR #2235, opened 2025-02-08
New flag in override structure on whether to be dropped when it's outside the simulation

weiliangjin2021:
A feature requested by @lucas-flexcompute

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Advanced option `dist_type` that allows `LinearLumpedElement` to be distributed across grid cells in different ways. For example, restricting the network portion to a single cell and using PEC wires to connect to the desired location of the terminals. 
 - New `layer_refinement_specs` field in `GridSpec` that takes a list of `LayerRefinementSpec` for automatic mesh refinement and snapping in layered structures. Structure corners on the cross section perpendicular to layer thickness direction can be automatically identified. Mesh is automatically snapped and refined around those corners.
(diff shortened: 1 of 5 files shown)


momchil-flex (commented):
So as far as I understand, this already happens by default for regular structures right https://github.com/flexcompute/tidy3d/blob/8f3195cc0d3be3a8c94cac83058afdf5608a60c0/tidy3d/components/grid/grid_spec.py#L606

Is there a use case in which we *don't* want it to apply to override structures?

lucas-flexcompute:
> So as far as I understand, this already happens by default for regular structures right
> 
> https://github.com/flexcompute/tidy3d/blob/8f3195cc0d3be3a8c94cac83058afdf5608a60c0/tidy3d/components/grid/grid_spec.py#L606
> 
> Is there a use case in which we _don't_ want it to apply to override structures?

When we use the `LayerRefinementSpec` to find the corners in the geometry, they can be outside the domain. The simplest case would be a straight waveguide drawn with a single box. the corners are outside to make sure they expand beyond the boundary absorbers, but we still need the refinement.

lucas-flexcompute (approved):

momchil-flex:
Ok, I think it makes sense, also that the flag affects *only* mesh override structures - as it's hard to imagine a case where you need a physical structure outside of the simulation domain to affect the meshing, right?

lucas-flexcompute:
Right, I cannot imagine a case for that. If the refinement is due to an existing geometry, it would only be needed when the geometry is inside the domain…

weiliangjin2021:
Right, only makes sense for non-physical structures. Any suggestions on the naming of those fields?

PR #2237, opened 2025-02-11
In LossyMetal rename SkinDepthFitterParam to SurfaceImpedanceFitterParam and plot impedance

weiliangjin2021:
The original term `SkinDepthFitterParam` is very nonstandard, and people might get more confused in the future when we extend it to more complicated imperfect metals, e.g. including surface roughness. Luckily, it was rolled out in the pre-release, so let's fix that before it's going formal.

Even though internally we are fitting `Z/(-j omega)` where `Z` is the surface impedance, we plot surface impedance as it's more relevant to users.

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -23,6 +23,7 @@ from ..constants import (
     C_0,
     CONDUCTIVITY,
     EPSILON_0,
+    ETA_0,
     HBAR,
     HERTZ,
     MICROMETER,
@@ -100,7 +101,7 @@ NONLINEAR_DEFAULT_NUM_ITERS = 5
 # Lossy metal
 LOSSY_METAL_DEFAULT_SAMPLING_FREQUENCY = 20
 LOSSY_METAL_SCALED_REAL_PART = 10.0
-LOSSY_METAL_DEFAULT_MAX_POLES = 3
+LOSSY_METAL_DEFAULT_MAX_POLES = 5
 LOSSY_METAL_DEFAULT_TOLERANCE_RMS = 1e-3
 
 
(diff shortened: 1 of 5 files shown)


weiliangjin2021:
An example of the plot:
image

tidy3d/components/medium.py
```diff
@@ -5105,22 +5105,25 @@ def _validate_frequency_range(cls, val):
...
 
-        # let's use scaled `skin_depth` in fitting: minimal real part equals ``SCALED_REAL_PART``
-        min_skin_depth_real = np.min(skin_depth.real)
-        if min_skin_depth_real <= 0:
-            raise SetupError("Physical skin depth cannot be non-positive. Something is wrong.")
+        # let's use scaled quantity in fitting: minimal real part equals ``SCALED_REAL_PART``
+        min_real = np.min(scaled_impedance.real)
+        if min_real <= 0:
+            raise SetupError(
+                "The real part of scaled surface impedance cannot be non-positive. Something is wrong."
```
lucas-flexcompute: "cannot be non-negative" → "must be positive"

Also, "something is wrong" is not helpful to the user. Either we say something to help solve the issue, or don't say anything. "Something is wrong" might be frustrating…

lucas-flexcompute (approved):

tidy3d/components/medium.py
```diff
@@ -5179,16 +5181,20 @@ def plot(
...
+        surface_impedance_model = (
+            -1j
+            * self.Hz_to_angular_freq(frequencies)
+            * self.scaled_surface_impedance_model.eps_model(frequencies)
+        )
+        ax.plot(frequencies, surface_impedance_model.real, label="Real (model)")
+        ax.plot(frequencies, surface_impedance_model.imag, label="Imag (model)")
 
-        ax.set_ylabel("Skin depth")
+        ax.set_ylabel(r"Surface impedance ($\Omega$)")
```
dmarek-flex, line 5216: Please double-check the units for "Surface impedance". I would think it would be Ohms/um^2

dmarek-flex (approved):
Thanks, surface impedance is a clearer name.

tidy3d/components/medium.py
weiliangjin2021: This isn't supposed to happen for passive medium. I'll add a validation to make sure it's passive, and change the msg to "Please create a github issue so that the problem can be investigated."

PR #2238, opened 2025-02-11
Add translated_copy function to ElectromagneticFieldData

### LINKED ISSUE #1526: Translation of field data for easier overlap computation

caseyflex:
This PR solves this issue:
https://github.com/flexcompute/tidy3d/issues/1526

It should replace this previous PR:
https://github.com/flexcompute/tidy3d/pull/1916

It allows field data to be translated so that overlaps can be taken between data from different locations.

There are some subtleties:

- The grid_expanded is translated along with the fields. The translated grid_expanded won't in general align with the original Yee grid
- The monitor is translated along with the fields. This was needed for stuff like `_diff_area` which used the monitor geometry, but it does mean the monitor is not the same as the original simulation monitor. This is related to the comment in `_isel` in `ModeSolverData`, where the function is kept private because the `_isel` is not applied to the monitor settings. There are some differences though: here, at least the translated monitor  could correspond to some hypothetical translated simulation, whereas with `_isel`, it's not always even possible to change the monitor configuration to give you whatever field data you selected.
- A rather important one: I updated the logic in `_symmetry_update_dict` basically to use linear interpolation rather than nearest-neighbor. The reason for this was that it was colocating to cell boundaries, and using nearest-neighbor interpolation meant it was basically choosing arbitrarily between the two neighboring values. It seems more robust and consistent to use linear interpolation for colocation throughout, including here. In particular it is more robust to translations (overlaps being translation-invariant etc.)
- Another robustness improvement: in `dot`, I changed from `_colocated_tangential_fields` to _interpolated_tangential_fields` for `fields_other` (this was already how `outer_dot` had it). This might be a bit slower sometimes due to caching of the former, but the former was not working robustly with translations, again because of numerical grid alignment issues.

### DIFF
--- a/tests/test_data/test_monitor_data.py
+++ b/tests/test_data/test_monitor_data.py
@@ -168,9 +168,9 @@ def make_permittivity_data(symmetry: bool = True):
     sim = SIM_SYM if symmetry else SIM
     return PermittivityData(
         monitor=PERMITTIVITY_MONITOR,
-        eps_xx=make_scalar_field_data_array("Ex", symmetry),
-        eps_yy=make_scalar_field_data_array("Ey", symmetry),
-        eps_zz=make_scalar_field_data_array("Ez", symmetry),
+        eps_xx=make_scalar_field_data_array("Ex", symmetry, colocate=False),
+        eps_yy=make_scalar_field_data_array("Ey", symmetry, colocate=False),
+        eps_zz=make_scalar_field_data_array("Ez", symmetry, colocate=False),
         symmetry=sim.symmetry,
         symmetry_center=sim.center,
         grid_expanded=sim.discretize_monitor(PERMITTIVITY_MONITOR),
@@ -234,7 +234,7 @@ def test_field_data():
     # Compute flux as dot product with itself
     flux2 = np.abs(data_2d.dot(data_2d))
     # Assert result is the same
-    assert np.all(flux1 == flux2)
+    assert np.allclose(flux1, flux2)
 
 
 def test_field_data_to_source():
@@ -267,7 +267,7 @@ def test_mode_solver_data():
     # Compute flux as dot product with itself
     flux2 = np.abs(data.dot(data))
     # Assert result is the same
-    assert np.all(flux1 == flux2)
+    assert np.allclose(flux1, flux2)
     # Compute dot product with a field data
     field_data = make_field_data_2d()
     dot = data.dot(field_data)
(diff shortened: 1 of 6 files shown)


tests/test_data/test_monitor_data.py
```diff
@@ -611,6 +611,75 @@ def isel(data, freqs):
...
+        mode_data.dot(mode_data),
+        mode_data_translated.translated_copy(vector=[-v for v in vector]).dot(mode_data),
+        atol=atol,
+    )
+
+    assert np.allclose(
+        mode_data.outer_dot(mode_data),
+        mode_data_translated.translated_copy(vector=[-v for v in vector]).outer_dot(mode_data),
+        atol=atol,
+    )
```
momchil-flex, line 672: These tests make sure that `.dot` and `.outer_dot` work with translated data, that translating two sets of data by the same amount returns the same overlap, and that translating back recovers the original data. That's great, but is there also a way to test that translation does what we expect? For example, you could do two mode solves for the same waveguide, but the waveguide is translated, and then check that the overlap between the translated data and the second data is close to 1? This could be added to the mode solver tests.

tidy3d/components/data/monitor_data.py
```diff
@@ -272,9 +273,32 @@ def _symmetry_update_dict(self) -> Dict:
...
+                scalar_data_arrays = []
+                if len(scalar_data.coords[dim_name]) == 1:
+                    scalar_data = scalar_data.sel(**{dim_name: coords_interp}, method="nearest")
+                else:
+                    if len(flip_inds) > 0:
+                        scalar_data_flip = scalar_data.interp(
+                            **{dim_name: coords_interp[flip_inds]},
+                            method="linear",
+                            kwargs={"fill_value": "extrapolate"},
+                        )
```
momchil-flex: Also add `assume_sorted=True` for performance, here and below.

momchil-flex (commented):
Thanks @caseyflex this looks great!

> - The grid_expanded is translated along with the fields. The translated grid_expanded won't in general align with the original Yee grid
> - The monitor is translated along with the fields. This was needed for stuff like _diff_area which used the monitor geometry, but it does mean the monitor is not the same as the original simulation monitor. This is related to the comment in _isel in ModeSolverData, where the function is kept private because the _isel is not applied to the monitor settings. There are some differences though: here, at least the translated monitor could correspond to some hypothetical translated simulation, whereas with _isel, it's not always even possible to change the monitor configuration to give you whatever field data you selected.

I think these are fine.

> - A rather important one: I updated the logic in _symmetry_update_dict basically to use linear interpolation rather than nearest-neighbor. The reason for this was that it was colocating to cell boundaries, and using nearest-neighbor interpolation meant it was basically choosing arbitrarily between the two neighboring values. It seems more robust and consistent to use linear interpolation for colocation throughout, including here. In particular it is more robust to translations (overlaps being translation-invariant etc.)
> - Another robustness improvement: in dot, I changed from _colocated_tangential_fields to _interpolated_tangential_fieldsforfields_other(this was already howouter_dot` had it). This might be a bit slower sometimes due to caching of the former, but the former was not working robustly with translations, again because of numerical grid alignment issues.

Yeah so both of those are hitting an important point. I agree that interpolating would be much more robust and it's nice to enable `dot` to work with field datasets which are not necessarily on the same coords. However indeed this was originally done due to performance. I had to put quite some effort in getting our postprocessing time down in cases where e.g. there are mode monitors with many spatial points or at many frequencies or modes. To be fair, I do not know or remember in the end of the day how important each change was. Maybe especially if you add `assume_sorted=True` it will actually work well (I think I only figured this out later). Could you test the change in time in computing the `dot` product for colocated data (could just be `mode_data.dot(mode_data)`) when the data is big? If it's significant, we might want to have some way to also keep the old behavior for performance in our post-processing (e.g. introduce a private `_dot` function or something).

tidy3d/components/data/monitor_data.py
```diff
@@ -272,9 +273,32 @@ def _symmetry_update_dict(self) -> Dict:
...
+                        scalar_data_arrays.append(scalar_data_no_flip)
+                    scalar_data = xr.concat(scalar_data_arrays, dim=dim_name)
 
-                scalar_data = scalar_data.sel(**{dim_name: coords_interp}, method="nearest")
-                scalar_data = scalar_data.assign_coords({dim_name: coords})
+                # scalar_data = scalar_data.interp(**{dim_name: coords_interp}, method="linear", kwargs={"fill_value": "extrapolate"})
+                # scalar_data = scalar_data.sel(**{dim_name: coords_interp}, method="nearest")
+                # scalar_data = scalar_data.assign_coords({dim_name: coords})
```
yaugenst-flex: This should probably be removed?

yaugenst-flex (approved):
Thanks @caseyflex this is super useful. Looks good to me minus @momchil-flex comments about performance. Anecdotally, `assume_sorted` has been a bottleneck for autograd too, so might be all that is needed.

caseyflex:
@momchil-flex

I did a little bit of benchmarking like you said. In my example problem (which wasn't too big), `sel` took about .14 seconds, while my new `interp` code took about .22 seconds. Adding `assume_sorted` helped a bit, taking about .19 seconds.

However I also realized something -- the reason `sel` made sense here is that we are assuming that `monitor.colocate` agrees with the actual colocation of the field data. For example, if `monitor.colocate = True`, then the field data (unexpanded) should already be colocated, so in this symmetry expansion when we `sel` at the colocation coords (cell boundaries), it shouldn't be doing any further interpolation.

The issue was rather in `tests/test_data/test_data_arrays.py::get_xyz`. The field data was being constructed without colocation, regardless of the value of `monitor.colocate`. So this was actually violating the assumption in `symmetry_expanded_copy`, which was therefore doing additional zeroth-order interpolation.

I updated this function in the tests to respect the value of `monitor.colocate`. I changed `_symmetry_update_dict` to use the original `sel` implementation if no interpolation is needed; if interpolation is needed, then it uses the less efficient implementation but gives a warning (since this indicates incorrect setup of data / monitor pairing). Alternatively, we could just go back to the `sel` implementation entirely, but I thought it would be helpful to see if there was ever such a mismatch through a warning (and have a more accurate fallback interpolation). Another alternative would be to raise an error here instead of the graceful fallback + warning.

I also added some tests for this and the mode solver translated dot test you described. I fixed up the colocation for the permittivity monitor data tests too.

momchil-flex (approved):
Ah nice I think this all looks good now! We actually don't need to modify the symmetry expansion (away from .sel) to achieve the more general behavior of .dot (not requiring field datasets to live on the same grid), which is perfect.

@tomflexcompute I think when you plan an example with the Gaussian beam overlap, you should also showcase this new feature!

PR #2241, opened 2025-02-12
Support for automatically running multiple adjoint simulations from single fwd

yaugenst-flex:
**Summary of Changes**:

- **Multiple Adjoint Simulations**: Each forward simulation can now spawn multiple adjoint simulations, grouped by either port or frequency (whichever yields fewer adjoint simulations).  
- **Batch Execution**: All adjoint simulations are now combined and run in a single batch—this applies to both `run` (1 forward → N adjoint) and `run_async` (N forward → M adjoint).  
- **`max_num_adjoint_sims`**: Introduced a limit on the maximum number of adjoint simulations to run automatically (defaults to `10`) to prevent unexpectedly large batched runs.  
- **Testing**: Verified on the `AutogradWDM` notebook and various numerical test scenarios.  
- **Known Edge Case**: When grouping by port for broadband sources, some setups with overlapping frequencies in the adjoint sources still don't work. This needs some modified handling in the post-normalization amplitudes that I'll implement in a separate PR, but this is not a regression so I think it's ok to not include here.

I also removed the stale broadband fitting code for now, until we can explore that implementation again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Function `translated_copy` in `ElectromagneticFieldData` to facilitate computing overlaps between field data at different locations.
 - `PlaneWaveBeamProfile`, `GaussianBeamProfile` and `AstigmaticGaussianBeamProfile` components to compute field data associated to such beams based on their corresponding analytical expressions, and compute field overlaps with other data.
 - `num_freqs` argument to the `PlaneWave` source.
+- Support for running multiple adjoint simulations from a single forward simulation in adjoint pipeline depending on monitor configuration.
 
 ### Changed
 - The coordinate of snapping points in `GridSpec` can take value `None`, so that mesh can be selectively snapped only along certain dimensions.
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -151,7 +151,7 @@ def use_emulated_run(monkeypatch):
     import tidy3d
 
     if TEST_MODE in ("pipeline", "speed"):
-        task_id_fwd = "task_fwd"
+        task_name_fwd = "task_fwd"
         AUX_KEY_SIM_FIELDS_KEYS = "sim_fields_keys"
 
         cache = {}
@@ -168,7 +168,7 @@ def use_emulated_run(monkeypatch):
 
         def emulated_run_fwd(simulation, task_name, **run_kwargs) -> td.SimulationData:
             """What gets called instead of ``web/api/autograd/autograd.py::_run_tidy3d``."""
-            task_id_fwd = task_name
+            task_name_fwd = task_name
             if run_kwargs.get("simulation_type") == "autograd_fwd":
                 sim_original = simulation
                 sim_fields_keys = run_kwargs["sim_fields_keys"]
@@ -186,28 +186,28 @@ def use_emulated_run(monkeypatch):
                 )
 
                 # cache original and fwd data locally for test
-                cache[task_id_fwd] = copy.copy(aux_data)
-                cache[task_id_fwd][AUX_KEY_SIM_FIELDS_KEYS] = sim_fields_keys
+                cache[task_name_fwd] = copy.copy(aux_data)
+                cache[task_name_fwd][AUX_KEY_SIM_FIELDS_KEYS] = sim_fields_keys
                 # return original data only
-                return aux_data[AUX_KEY_SIM_DATA_ORIGINAL], task_id_fwd
+                return aux_data[AUX_KEY_SIM_DATA_ORIGINAL], task_name_fwd
             else:
-                return run_emulated(simulation, task_name=task_name), task_id_fwd
+                return run_emulated(simulation, task_name=task_name), task_name_fwd
 
         def emulated_run_bwd(simulation, task_name, **run_kwargs) -> td.SimulationData:
             """What gets called instead of ``web/api/autograd/autograd.py::_run_tidy3d_bwd``."""
 
-            task_id_fwd = task_name[:-8]
+            task_name_fwd = "".join(task_name.partition("_adjoint")[:-2])
 
             # run the adjoint sim
             sim_data_adj = run_emulated(simulation, task_name="task_name")
 
             # grab the fwd and original data from the cache
-            aux_data_fwd = cache[task_id_fwd]
+            aux_data_fwd = cache[task_name_fwd]
             sim_data_orig = aux_data_fwd[AUX_KEY_SIM_DATA_ORIGINAL]
             sim_data_fwd = aux_data_fwd[AUX_KEY_SIM_DATA_FWD]
 
             # get the original traced fields
-            sim_fields_keys = cache[task_id_fwd][AUX_KEY_SIM_FIELDS_KEYS]
+            sim_fields_keys = cache[task_name_fwd][AUX_KEY_SIM_FIELDS_KEYS]
 
             # postprocess (compute adjoint gradients)
             traced_fields_vjp = postprocess_adj(
@@ -225,9 +225,9 @@ def use_emulated_run(monkeypatch):
             for task_name, simulation in simulations.items():
                 if sim_fields_keys_dict is not None:
                     run_kwargs["sim_fields_keys"] = sim_fields_keys_dict[task_name]
-                sim_data_orig, task_id_fwd = emulated_run_fwd(simulation, task_name, **run_kwargs)
+                sim_data_orig, task_name_fwd = emulated_run_fwd(simulation, task_name, **run_kwargs)
                 batch_data_orig[task_name] = sim_data_orig
-                task_ids_fwd[task_name] = task_id_fwd
+                task_ids_fwd[task_name] = task_name_fwd
 
             class EmulatedBatchData(web.BatchData):
                 def load_sim_data(self, task_name):
@@ -251,7 +251,6 @@ def use_emulated_run(monkeypatch):
 
         monkeypatch.setattr(webapi, "run", run_emulated)
         monkeypatch.setattr(tidy3d.web.api.autograd.autograd, "_run_tidy3d", emulated_run_fwd)
-        monkeypatch.setattr(tidy3d.web.api.autograd.autograd, "_run_tidy3d_bwd", emulated_run_bwd)
         monkeypatch.setattr(
             tidy3d.web.api.autograd.autograd, "_run_async_tidy3d", emulated_run_async_fwd
         )
@@ -752,7 +751,7 @@ def test_run_zero_grad(use_emulated_run):
         sim_data = run(sim, task_name="adjoint_test", verbose=False)
         return 0 * postprocess(sim_data)
 
-    with AssertLogLevel("WARNING", contains_str="no sources"):
+    with AssertLogLevel("WARNING", contains_str="fields are zero"):
         grad = ag.grad(objective)(params0)
 
 
@@ -1678,7 +1677,7 @@ def test_multi_freq_edge_cases(use_emulated_run, structure_key, label, check_fn,
         return postprocess_fn(data)
 
     if label == "src_2_freq_2_mon_2":
-        with pytest.raises(NotImplementedError):
+        with pytest.raises(ValueError):
             g = ag.grad(objective)(params0)
     else:
         g = ag.grad(objective)(params0)
--- a/tidy3d/components/data/data_array.py
+++ b/tidy3d/components/data/data_array.py
@@ -248,8 +248,7 @@ class DataArray(xr.DataArray):
         """Load an DataArray from an hdf5 file with a given path to the group."""
         if ".hdf5" not in fname:
             raise FileError(
-                "DataArray objects must be written to '.hdf5' format. "
-                f"Given filename of {fname}."
+                f"'DataArray' objects must be written to '.hdf5' format. Given filename of {fname}."
             )
         return cls.from_hdf5(fname=fname, group_path=group_path)
 
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -1033,48 +1033,67 @@ class SimulationData(AbstractYeeGridSimulationData):
 
         return sim_data_original, sim_data_fwd
 
-    def make_adjoint_sim(
+    def make_adjoint_sims(
         self,
         data_vjp_paths: set[tuple],
         adjoint_monitors: list[Monitor],
-    ) -> Simulation | None:
-        """Make the adjoint simulation from the original simulation and the VJP-containing data."""
+    ) -> list[Simulation]:
+        """Make the adjoint simulations from the original simulation and the VJP-containing data."""
+
+        if not data_vjp_paths:
+            return []
 
         sim_original = self.simulation
 
         # generate the adjoint sources {mnt_name : list[Source]}
         sources_adj_dict = self.make_adjoint_sources(data_vjp_paths=data_vjp_paths)
+        if not sources_adj_dict:
+            return []
+
         adj_srcs = []
         for src_list in sources_adj_dict.values():
             adj_srcs += list(src_list)
 
-        if not any(adj_srcs):
-            return None
+        adjoint_source_infos = self.process_adjoint_sources(adj_srcs=adj_srcs)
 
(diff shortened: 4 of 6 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Advanced option `dist_type` that allows `LinearLumpedElement` to be distributed across grid cells in different ways. For example, restricting the network portion to a single cell and using PEC wires to connect to the desired location of the terminals. 
 - New `layer_refinement_specs` field in `GridSpec` that takes a list of `LayerRefinementSpec` for automatic mesh refinement and snapping in layered structures. Structure corners on the cross section perpendicular to layer thickness direction can be automatically identified. Mesh is automatically snapped and refined around those corners.
 - New field `drop_outside_sim` in `MeshOverrideStructure` to specify whether to drop an override structure if it is outside the simulation domain, but it overlaps with the simulation domain when projected to an axis.
+- Support for running multiple adjoint simulations from a single forward simulation in adjoint pipeline depending on monitor configuration.
```
tylerflex, line 17: could you also add some mention of this in the autograd/README.md FAQ?

tests/test_components/test_autograd.py
```diff
@@ -196,7 +196,7 @@ def emulated_run_fwd(simulation, task_name, **run_kwargs) -> td.SimulationData:
         def emulated_run_bwd(simulation, task_name, **run_kwargs) -> td.SimulationData:
             """What gets called instead of ``web/api/autograd/autograd.py::_run_tidy3d_bwd``."""
 
-            task_id_fwd = task_name[:-8]
+            task_id_fwd = "".join(task_name.partition("_adjoint")[:-2])
```
tylerflex: should this be `[:-1]`? if the forward task name is `"something"` then the adjoint task name is `"something_adjoint"`, wouldn't `"something_adjoint".partition("_adjoint")` give `["something", "_adjoint"]` and therefore we should select up until the `-1` element? I don't remember exactly how this was set up.

tylerflex: should the variable be `"task_name_fwd"`?

tests/test_components/test_autograd.py
```diff
@@ -752,7 +752,7 @@ def objective(*args):
         sim_data = run(sim, task_name="adjoint_test", verbose=False)
         return 0 * postprocess(sim_data)
 
-    with AssertLogLevel("WARNING", contains_str="no sources"):
```
tylerflex, line 755: could you explain why this is removed? what warning gets raised?

tidy3d/components/data/data_array.py
```diff
@@ -248,8 +248,7 @@ def from_file(cls, fname: str, group_path: str) -> Self:
         """Load an DataArray from an hdf5 file with a given path to the group."""
         if ".hdf5" not in fname:
             raise FileError(
-                "DataArray objects must be written to '.hdf5' format. "
-                f"Given filename of {fname}."
+                f"DataArray objects must be written to '.hdf5' format. Given filename of {fname}."
```
tylerflex: let's put data `DataArray` in single quotes.

tidy3d/components/data/sim_data.py
```diff
@@ -1033,48 +1033,67 @@ def split_original_fwd(self, num_mnts_original: int) -> Tuple[SimulationData, Si
...
 
-    def make_adjoint_sim(
+    def make_adjoint_sims(
         self,
         data_vjp_paths: set[tuple],
         adjoint_monitors: list[Monitor],
-    ) -> Simulation | None:
-        """Make the adjoint simulation from the original simulation and the VJP-containing data."""
+    ) -> list[Simulation] | None:
+        """Make the adjoint simulations from the original simulation and the VJP-containing data."""
```
tylerflex, line 1041: remind me why we return `None` in these edge cases instead of empty list?

tidy3d/components/data/sim_data.py
```diff
@@ -1033,48 +1033,67 @@ def split_original_fwd(self, num_mnts_original: int) -> Tuple[SimulationData, Si
...
-        # fields to update the 'fwd' simulation with to make it 'adj'
-        sim_adj_update_dict = dict(
-            sources=adjoint_source_info.sources,
-            boundary_spec=bc_adj,
-            monitors=adjoint_monitors,
-            post_norm=adjoint_source_info.post_norm,
-        )
+        adj_sims = []
+        for adjoint_source_info in adjoint_source_infos:
+            # only include monitors with the same freqs as the adjoint sources
```
tylerflex, line 1073: so this basically filters out situations where the gradient monitors have more frequencies than were needed in the gradient calculation for this specific adjoint sim? because the objective function only depends on a subset of them?

tidy3d/components/data/sim_data.py
```diff
@@ -1033,48 +1033,67 @@ def split_original_fwd(self, num_mnts_original: int) -> Tuple[SimulationData, Si
...
+                boundary_spec=bc_adj,
+                monitors=monitors,
+                post_norm=adjoint_source_info.post_norm,
+            )
+
+            if not adjoint_source_info.normalize_sim:
+                sim_adj_update_dict["normalize_index"] = None
+
+            # set the ADJ grid spec wavelength to the original wavelength (for same meshing)
+            grid_spec_original = sim_original.grid_spec
```
tylerflex: does this need to be computed each time in the loop over adjoint_source_infos?

tidy3d/web/api/autograd/autograd.py
```diff
@@ -33,6 +34,7 @@
 SIM_FIELDS_KEYS_FILE = "autograd_sim_fields_keys.hdf5"
 
 MAX_NUM_TRACED_STRUCTURES = 500
+MAX_NUM_ADJOINT_SIMS = 10
```
tylerflex: might want to be more specific that this is "per fwd sim", otherwise could sound like we restrict the number of adjoint sims eg. for batches

tidy3d/web/api/autograd/autograd.py
```diff
@@ -91,6 +95,7 @@ def run(
     simulation_type: str = "tidy3d",
     parent_tasks: list[str] = None,
     local_gradient: bool = LOCAL_GRADIENT,
+    max_num_adjoint_sims: int = MAX_NUM_ADJOINT_SIMS,
```
tylerflex: we really need a config for this and local_gradient

tidy3d/web/api/autograd/autograd.py
```diff
@@ -125,6 +130,8 @@ def run(
     local_gradient: bool = False
         Whether to perform gradient calculation locally, requiring more downloads but potentially
         more stable with experimental features.
+    max_num_adjoint_sims: int = MAX_NUM_ADJOINT_SIMS
```
tylerflex: not sure how we handle these in docstrings, but maybe it needs to be either an f-string or just hard coded for now.

tidy3d/components/data/sim_data.py
```diff
@@ -1191,115 +1209,6 @@ def _make_post_norm_amps(adj_srcs: list[SourceType]) -> xr.DataArray:
         amps_complex = np.array(amps_complex)
         return xr.DataArray(amps_complex, coords=coords)
 
-    """ FITTING APPROACH """
```
tylerflex, line 1194: sad to lose all of this code, but I suppose we'll probably need to just write it from scratch eventually

tidy3d/web/api/autograd/autograd.py
```diff
@@ -612,31 +646,63 @@ def vjp(data_fields_vjp: AutogradFieldMap) -> AutogradFieldMap:
             )
             return {k: 0 * v for k, v in sim_fields_original.items()}
 
-        # run adjoint simulation
-        task_name_adj = str(task_name) + "_adjoint"
+        # Run adjoint simulations in batch
+        task_names_adj = [f"{task_name}_adjoint_{i}" for i in range(len(sims_adj))]
```
tylerflex, line 653: so in the typical, non-batched case, will all adjoint sims now just be called "task_name_fwd_adjoint_0"? maybe in that case we should just remove the _0? up to you

tidy3d/web/api/autograd/autograd.py
```diff
@@ -612,31 +646,63 @@ def vjp(data_fields_vjp: AutogradFieldMap) -> AutogradFieldMap:
...
+            # Run all adjoint sims in batch
+            td.log.info("Starting local batch adjoint simulations")
+            path = run_kwargs.pop("path")
+            path_dir = str(Path(path).parent.resolve())
+
+            batch_data_adj, _ = _run_async_tidy3d(sims_adj_dict, path_dir=path_dir, **run_kwargs)
+            td.log.info("Completed local batch adjoint simulations")
+
+            # sum partial derivatives from each adjoint simulation
+            for k, sim_data_adj in batch_data_adj.items():
```
tylerflex: variable `k` is redefined in the loop over `vjp_fields`.

tylerflex (commented):
looks great, my comments are pretty minor. Think we definitely will want to test this a bit in some of the other notebooks, for example [this one](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/Autograd21GaPLightExtractor.html) to make sure the mapping from fwd to adj is 100% correct, but it seems like it should be. Thanks Yannick!

tests/test_components/test_autograd.py
yaugenst-flex: the adjoint tasks are named slightly differently now, they look like `task_name_adjoint_idx` so that we can keep track of the adjoint sim. so the partition will return `["task_name", "_adjoint", "_0"]` for instance, which is where the `[:-2]` comes from

yaugenst-flex: sure, i just didnt change the variable name. it is the forward task name yeah

tests/test_components/test_autograd.py
yaugenst-flex: oh because it's actually a different warning now, there is an additional check before the adjoint sim construction which returns early if all vjps are zero. i can change the string to use the new warning

tidy3d/components/data/sim_data.py
yaugenst-flex: fair enough yeah this could return an empty list

tidy3d/components/data/sim_data.py
yaugenst-flex: correct, so for the adjoint sim we now only place adjoint monitors with the relevant frequencies. this is basically what we discussed and it would be nice to also be able to do this for the forward sim, but that's more involved. in the adjoint it's easy to do

tidy3d/components/data/sim_data.py
yaugenst-flex: yeah right this can be computed outside, good point.

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: `max_num_adjoint_per_fwd`? explicit but very verbose :D any better ideas?

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: yup was thinking the same

tidy3d/components/data/sim_data.py
yaugenst-flex: yeah i made a task with a link to the PR that introduced it, so we still have a reference internally

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: oops yeah. on that note, do you think 10 is a good default? i feel like this should cover most cases and is not a crazy number

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: wait i'm not sure. do you mean when there is a single fwd and single bwd? in that case yeah it'll be zero but i don't see a reason to introduce special handling just for that case? adjoint sims will always run in a batch, there is no non-batch adjoint anymore

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: fair enough :smile:

tests/test_components/test_autograd.py
tylerflex: yea, my mistake originally

tidy3d/web/api/autograd/autograd.py
tylerflex: not sure even if we need this. what is a situation where it might be super large?

tidy3d/web/api/autograd/autograd.py
tylerflex: ok that's fine, just will change behavior but since it's automatic anyway and behind the scenes probably fine

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: a field monitor with > 10 frequencies :smile:
or basically any time you have multiple monitors that contain multiple frequencies, although there you have to have >10 monitors and >10 freqs, otherwise there will always be one grouping which is below the threshold

tylerflex: ok, yea multifrequency field monitors is certainly one.. just wonder if maybe we should catch that specific case and basically warn if we encounter a situation where we need to split a FieldData into several adjoint sources.. but I think what you implemented already is probably fine and already takes care of that edge case

tylerflex: 10 seems reasonable. I guess if the number is too high, people would be upset about running too many sims without knowing. if it's too low, they might be upset about the error.

CHANGELOG.md
yaugenst-flex: right yeah, also updated some other outdated info

yaugenst-flex:
Thanks @tylerflex! The light extractor notebook actually doesn't use this single fwd -> multi bwd functionality since it runs a batch of distinct forward simulations. However, it still uncovered a problem - since adjoint is a batch now, the adjoint batch will conflict with the fwd batch (because we hard-code `batch.hdf5`...). I changed this behavior now so that the adjoint batch will always create a directory named `adjoint_data/` in whichever directory the forward simulation data is saved (specified either by `path` or `path_dir`). This way, when you run the gradient calculation locally, the adjoint data will always be clearly separated from the forward field/batch data.

We'll need to rework the paths similar (and maybe in conjunction with) the configs eventually, but I think that's out of scope for this PR.

Other than that, the notebook seems to run fine. I'll check a few more cases to be sure.

tylerflex:
> I changed this behavior now so that the adjoint batch will always create a directory named adjoint_data/ in whichever directory the forward simulation data is saved (specified either by path or path_dir). This way, when you run the gradient calculation locally, the adjoint data will always be clearly separated from the forward field/batch data.

hm. I would have thought maybe the solution would be to create a `batch_adjoint.hdf5` in the original data path dir? do you agree / how possible is this without a major Batch refactor?

yaugenst-flex:
So one problem there is that we actually don't expose a way of setting the file name for a batch. It's not a huge project but definitely requires a little refactoring. That should be addressed, but I'm not sure this PR is the right place for that. In particular because I think:
```
batch.hdf5
fb847d33-f50c-45c7-be9d-9c9b0e7e8b6d.hdf5
2f838e26-8828-4e54-ae95-7ac213ba1a0c.hdf5
e41600cf-cb36-4a09-bfcb-4df2453498db.hdf5
adjoint_data/
├── batch.hdf5
├── ac962576-8d21-4acd-9ef8-b82c8d7bdaad.hdf5
├── 9fcddd4b-f64b-4d0a-8fb0-9fa9845752ce.hdf5
└── 2116be94-01db-4f02-9c81-e2bab8d7a5cc.hdf5
```

is much cleaner / easier to work with than

```
batch.hdf5
batch_adjoint.hdf5
fb847d33-f50c-45c7-be9d-9c9b0e7e8b6d.hdf5
2f838e26-8828-4e54-ae95-7ac213ba1a0c.hdf5
e41600cf-cb36-4a09-bfcb-4df2453498db.hdf5
ac962576-8d21-4acd-9ef8-b82c8d7bdaad.hdf5
9fcddd4b-f64b-4d0a-8fb0-9fa9845752ce.hdf5
2116be94-01db-4f02-9c81-e2bab8d7a5cc.hdf5
```

So even if we could set the batch name, I think I'd still prefer this version..

tylerflex (approved):

PR #2243, opened 2025-02-12
Update the ModeSolver by adding perturbation-based group index and group velocity dispersion calculatoion method

vasilyzabelin:


vasilyzabelin (commented):
This is the draft version of the new perturbation theory based method to calculate the group index and the group dispersion velocity.

PR #2244, opened 2025-02-12
Fix extremely long run_time in the presence of lossy metal

weiliangjin2021:
Encountered a validation error on run_time being too large, which is related to how index of lossy metal is applied to the estimation of runtime.


tidy3d/components/simulation.py
```diff
@@ -4679,7 +4679,13 @@ def num_computational_grid_points(self):
...
         """List of refractive indices in the simulation at a given frequency."""
 
-        eps_values = [structure.medium.eps_model(freq) for structure in self.static_structures]
+        # set n = 1 for special medium
+        eps_values = [
+            1.0 + 0j
+            if isinstance(structure.medium, LossyMetalMedium) or structure.medium.is_pec
+            else structure.medium.eps_model(freq)
+            for structure in self.static_structures
+        ]
```
momchil-flex, line 4686: So this is fine but remind me how this enters in other places, e.g. the mesher? Do we also skip these media / assign a value of 1? Do we want to have like a method of all `Medium` types like `eps_meshing(freq)` which returns `eps_model(freq)` for usual ones, and `1` for lossy metal and PEC, and use this everywhere where needed instead of `eps_model`?

tylerflex: was going to say same comment

tylerflex (commented):
changelog reminder :)

tidy3d/components/simulation.py
weiliangjin2021: yeah, that's a good idea

momchil-flex (approved):

tylerflex (approved):

PR #2247, opened 2025-02-13
Replace complex fields with real fields in all nonlinearities

caseyflex:
Previously, `KerrNonlinearity` (n2) and `TwoPhotonAbsorption` (TPA, FCA, FCPD) were implemented using complex fields. This works fine to capture some effects, but it misses some other physical effects that are captured if they are implemented using the physical real fields.

Here, these are replaced with "real fields" implementations. If complex fields are used in the simulation for some other reason (like Bloch boundary condition), then the real and imaginary parts of all fields are treated separately here (they do not interact nonlinearly).

The new implementation does change the results of simulations containing these nonlinearities. For example, previously `KerrNonlinearity` would not capture third-harmonic generation, while `NonlinearSusceptibility` would capture this; now, `KerrNonlinearity` agrees exactly with `NonlinearSusceptibility` for the right choice of parameters (`KerrNonlinearity` can be more convenient to use, as it automatically detects `n0` and does the conversion for you). `TwoPhotonAbsorption` similarly might have different results under the new implementation. Thus, I have added the option `use_complex_fields` (False by default) to these nonlinearities so that a user is still able to obtain the old results. I have also added some warnings discouraging this. It is unfortunate to make a breaking change, but it is more important that the results be more physical. (Also, complex values of `beta` and `n2` are no longer supported if `use_complex_fields=False`; previously they resulted in a warning).

I have updated the docstrings to reflect this new behavior. I have also tried to give helpful warnings and error messages where possible. There are a couple other minor changes related to edge case handling and test coverage.

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -218,6 +218,8 @@ class NonlinearModel(ABC, Tidy3dBaseModel):
         freqs: List[pd.PositiveFloat],
     ) -> complex:
         """Get a single value for n0."""
+        if freqs is None:
+            freqs = []
         freqs = np.array(freqs, dtype=float)
         ns, ks = medium.nk_model(freqs)
         nks = ns + 1j * ks
@@ -256,7 +258,7 @@ class NonlinearModel(ABC, Tidy3dBaseModel):
     @property
     def complex_fields(self) -> bool:
         """Whether the model uses complex fields."""
-        pass
+        return False
 
 
 class NonlinearSusceptibility(NonlinearModel):
@@ -322,11 +324,6 @@ class NonlinearSusceptibility(NonlinearModel):
             )
         return val
 
-    @property
-    def complex_fields(self) -> bool:
-        """Whether the model uses complex fields."""
-        return False
-
 
 class TwoPhotonAbsorption(NonlinearModel):
     """Model for two-photon absorption (TPA) nonlinearity which gives an intensity-dependent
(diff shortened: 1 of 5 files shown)


momchil-flex:
Thanks @caseyflex, this makes sense overall, will look in detail soon. Meanwhile I'm wondering if you reran the nonlinear notebook and if the results looked (sufficiently) the same?

yaugenst-flex (approved):
Thanks @caseyflex lgtm!

tidy3d/components/medium.py
```diff
@@ -322,53 +324,67 @@ def _validate_numiters(cls, val):
...
 
 class TwoPhotonAbsorption(NonlinearModel):
     """Model for two-photon absorption (TPA) nonlinearity which gives an intensity-dependent
     absorption of the form :math:`\\alpha = \\alpha_0 + \\beta I`.
     Also includes free-carrier absorption (FCA) and free-carrier plasma dispersion (FCPD) effects.
     The expression for the nonlinear polarization is given below.
 
-    Note
+    Notes
     ----
```
momchil-flex: In RST you need the dashes underneath to be at least as long as the title, so add one more dash.

tidy3d/components/medium.py
```diff
@@ -322,53 +324,67 @@ def _validate_numiters(cls, val):
...
     -------
     >>> tpa_model = TwoPhotonAbsorption(beta=1)
     """
 
+    use_complex_fields: bool = pd.Field(
+        False,
+        title="Use complex fields",
+        description="Whether to use the old deprecated complex-fields implementation. "
+        "The default real-field implementation is more physical and is always "
+        "recommended; this option is only available for backwards-compatibility "
```
momchil-flex: Since you're removing the documentation of the old method, let's mention here explicitly something like "backwards compatibility with Tidy3D version < 2.8." Then people can go to our docs and pull up the old documentation if they need to. Also in the other places this argument is defined.

momchil-flex (approved):
Nice

PR #2249, opened 2025-02-14
Speed up layer refinement by caching and skipping exterior structures

weiliangjin2021:
For addressing two issues encountered in meshing a complicated PCB:
- Corner detection can be faster when one only intends to refine a region rather than all structures. Now in corner detection, we drop structures outside the layer when the layer is bounded in inplane axis;
- Previously, internal snapping points and override structures are not cached. Whenever we can call plot_grid, it will need to find corners again.


weiliangjin2021:
I removed cached_property `grid` in `Simulation` class, as it almost looks identical to the one in `AbstractYeeGridSimulation` class, except that it's using self.static_structures rather than self.structures. Since we also called `self.static_structures` in `AbstractYeeGridSimulation` class, shall we move `self.static_structures` to `AbstractYeeGridSimulation`? @yaugenst

tidy3d/components/grid/grid_spec.py
```diff
@@ -1226,8 +1226,8 @@ def center_axis(self) -> float:
 
     @cached_property
     def _is_inplane_unbounded(self) -> bool:
-        """Whether the layer is unbounded in inplane dimensions."""
-        return np.isinf(self.size[(self.axis + 1) % 3]) and np.isinf(self.size[(self.axis + 2) % 3])
+        """Whether the layer is unbounded in any of inplane dimensions."""
```
momchil-flex: `any in-plane dimension` or `any of the in-plane dimensions`.

tidy3d/components/grid/grid_spec.py
```diff
@@ -1625,22 +1632,30 @@ def internal_snapping_points(self, structures: List[Structure]) -> List[Coordina
...
+    ) -> List[CoordinateOptional]:
         """Internal and external snapping points. External snapping points take higher priority.
         So far, internal snapping points are generated by `layer_refinement_specs`.
 
         Parameters
         ----------
         structures : List[Structure]
             List of physical structures.
+        cached_internal_snapping_points : List[CoordinateOptional]
+            If not `None`, skip recomputing internal snapping points.
```
momchil-flex: I guess this is fine, but to me it seems to suggest that some reorganization where there is some cached component stored in Simulation would make sense. Like instead of calling `self.grid_spec.make_grid` to make the grid, it could be something like a `self.grid_maker` cached property that initializes some `GridMaker` object that takes both inputs from `Simulation` like `structures`, etc. and the `grid_spec`? Then these `internal_snapping_points` are cached properties in that component and they don't need to be passed around its internal methods.

If we keep it as is for now, I think we don't really need the `cached_` prefix? Generally it is a common use case to allow something optional called e.g. `internal_snapping_points` and re-compute it if not provided.

momchil-flex (approved):
Ok for now, some comments, and yeah @yaugenst should comment on the static structures stuff, I did not give this any though.

yaugenst-flex:
Hey @weiliangjin2021, @momchil-flex, I think the `.static_structures` look fine to me. We need to do this on anything grid-related since we do not support tracing through grid setup (and it doesn't really make sense either...). We do test this though so as long as the tests pass this should be fine.

tidy3d/components/grid/grid_spec.py
weiliangjin2021: that's a good idea, since grid generation takes several inputs from simulation. But I'll leave that to a separate PR.

tidy3d/components/grid/grid_spec.py
```diff
@@ -1625,22 +1632,30 @@ def internal_snapping_points(self, structures: List[Structure]) -> List[Coordina
...
+    ) -> List[CoordinateOptional]:
         """Internal and external snapping points. External snapping points take higher priority.
         So far, internal snapping points are generated by `layer_refinement_specs`.
 
         Parameters
         ----------
         structures : List[Structure]
             List of physical structures.
+        internal_snapping_points : List[CoordinateOptional]
+            If not `None`, skip recomputing internal snapping points.
```
dmarek-flex: Maybe clearer to say:

> If `None`, recomputes internal snapping points.

tidy3d/components/grid/grid_spec.py
```diff
@@ -1688,17 +1707,22 @@ def all_override_structures(
             List of structures, with the simulation structure being the first item.
         wavelength : pd.PositiveFloat
             Wavelength to use for minimal step size in vaccum.
+        internal_override_structures : List[MeshOverrideStructure]
+            If not `None`, skip recomputing internal override structures.
```
dmarek-flex: Same here, double-negative always confuses me :smile:

dmarek-flex (approved):
Looks good to me, just thought the docstring could be a bit clearer by avoiding double-negative

tidy3d/components/grid/grid_spec.py
```diff
@@ -1772,6 +1798,10 @@ def make_grid(
             List of sources.
         num_pml_layers : List[Tuple[float, float]]
             List containing the number of absorber layers in - and + boundaries.
+        internal_override_structures : List[MeshOverrideStructure]
+            If not `None`, skip recomputing internal override structures.
+        internal_snapping_points : List[CoordinateOptional]
+            If not `None`, skip recomputing internal snapping points.
```
dmarek-flex: Here too

PR #2252, opened 2025-02-17
add batch_id for Batch

QingengWei:


tidy3d/web/api/container.py
```diff
@@ -476,6 +484,12 @@ class Batch(WebContainer):
         * `Inverse taper edge coupler <../../notebooks/EdgeCoupler.html>`_
     """
 
+    batch_id: str = pd.Field(
+        f"bat-{uuid.uuid4()}",
+        title="BatchId",
+        description="The batchId to run as a batch.",
+    )
+
```
yaugenst-flex, line 492: This implementation will currently lead to the same default batch id for every batch in the same python session, i.e. if you submit multiple batches, they will all have the same batch id.

Also more generally I'm not a big fan of using uuid's here because they won't be reproducible across multiple sessions. The batch id's should be unique and deterministic, e.g. based on the hashes of the simulations that are contained in the batch.

yaugenst-flex (changes requested):

PR #2254, opened 2025-02-18
Fixed bug and added warning when a lumped element is placed outside the Simulation.

dmarek-flex:
Lumped elements are no longer added as a volumetric structure when they partially intersect with the simulation domain, which is a bit more restrictive than regular structures. Now if the lumped element is not fully contained within the simulation domain a warning is emitted and it is not added.


weiliangjin2021 (approved):

tomflexcompute (approved):
Thanks @dmarek-flex

PR #2257, opened 2025-02-19
Add ModeSolver remote data patching

caseyflex:
This fixes a bug where some methods in `ModeSolver`, like `plot_field`, were not updated with the results of the remote mode solve, and thus recomputed / used the local results.

It already worked correctly in the legacy mode webapi; now, it works correctly in `web.run` and when using `Job` and `Batch`.

Caveats:

- It's not a nice solution; it might be unexpected that `web.run(mode_solver)` modifies the typically immutable `mode_solver` object. This is just a workaround until we fully switch to `ModeSimulation`.
- It modifies `cached_properties` which may not be robust, but is the same as the legacy mode webapi.
- When `Batch` is used, it would need to load each data into memory to patch the simulations, while keeping all the simulations in memory (and thus their patched data too). This may cause difficulty in memory-intensive situations
- There are still some webapi functions that would not work correctly. If you call `web.load` manually, it doesn't know about the `ModeSolver` so it can't automatically patch, but you can manually patch using the private `ModeSolver._patch_data` method. Similarly if you download from the GUI, the `ModeSolver` object might need you to call `_patch_data` manually. I don't think these could be automated without more serious overhaul.
- In the legacy mode webapi, if you call `get_modesolver`, the mode solver would need manual patching with the remote data. We could fix this too by making `get_modesolver` also run the remote mode solve and patch the data, but that's kind of a big change to the semantics of this function.


lucas-flexcompute (approved):

tomflexcompute (approved):

tidy3d/web/api/container.py
```diff
@@ -944,7 +948,14 @@ def load(self, path_dir: str = DEFAULT_DATA_DIR) -> BatchData:
             task_paths[task_name] = self._job_data_path(task_id=job.task_id, path_dir=path_dir)
             task_ids[task_name] = self.jobs[task_name].task_id
 
-        return BatchData(task_paths=task_paths, task_ids=task_ids, verbose=self.verbose)
+        data = BatchData(task_paths=task_paths, task_ids=task_ids, verbose=self.verbose)
+
+        for task_name, job in self.jobs.items():
+            if isinstance(job.simulation, ModeSolver):
+                job_data = data[task_name]
+                job.simulation._patch_data(data=job_data)
```
momchil-flex, line 956: This also propagates to `batch.simulations[...]` right? So a user would be able to do `batch.simulations["1"].plot_fields`?

momchil-flex (approved):
Yeah, definitely messy, but again it seems like we're just maintaining the same functionality - and messiness - of the existing ModeSolver run, and will eventually be fixed by fully migrating to ModeSimulation.

> When Batch is used, it would need to load each data into memory to patch the simulations, while keeping all the simulations in memory (and thus their patched data too). This may cause difficulty in memory-intensive situations

This is the one that worried me the most, but, again: this is equivalent to the existing `run_batch` that Prash added. There too, all the data will be loaded.

tidy3d/web/api/container.py
caseyflex: the answer is yes, but thanks for the comment, I added a test for this

momchil-flex:
Great, seems like it needs a rebase but feel free to merge after!

PR #2258, opened 2025-02-19
Add option to use dispersion fitter without rich.progress

caseyflex:
The `rich` package is not supported by GUI, so they requested the option to use the fitter without this package.

### DIFF
--- a/tidy3d/components/dispersion_fitter.py
+++ b/tidy3d/components/dispersion_fitter.py
@@ -6,11 +6,10 @@ from typing import Optional
 
 import numpy as np
 from pydantic.v1 import Field, NonNegativeFloat, PositiveFloat, PositiveInt, validator
-from rich.progress import Progress
 
 from tidy3d.constants import fp_eps
 from tidy3d.exceptions import ValidationError
-from tidy3d.log import get_logging_console, log
+from tidy3d.log import Progress, get_logging_console, log
 
 from .base import Tidy3dBaseModel, cached_property, skip_if_fields_missing
 from .types import ArrayComplex1D, ArrayComplex2D, ArrayFloat1D, ArrayFloat2D
@@ -822,12 +821,11 @@ def fit(
 
     Returns
     -------
-    Tuple[Tuple[float, ArrayComplex1D, ArrayComplex1D], float]
+    tuple[tuple[float, ArrayComplex1D, ArrayComplex1D], float]
         Best fitting result: (dispersive medium parameters, weighted RMS error).
         The dispersive medium parameters have the form (resp_inf, poles, residues)
         and are in the original unscaled units.
     """
-
     if max_num_poles < min_num_poles:
         raise ValidationError(
             "Dispersion fitter cannot have 'max_num_poles' less than 'min_num_poles'."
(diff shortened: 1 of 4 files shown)


yaugenst-flex (commented):
Thanks @caseyflex yeah I think we need to generally streamline sources/sinks for logging etc., this is kind of related to that.

A lazy import for `rich` is the right choice (as we do elsewhere), what I'm a bit concerned about is the duplication of logic in the `fit()` function, it's more surface area for bugs to creep in. I think a more robust approach might be to keep the current logic as-is, but introduce a dummy progress bar that exposes all the same methods but does nothing if the `rich` import fails. This could in principle also be reused in other functions. We could either add it here, or perhaps better in `tidy3d/log.py` (although we import `rich` there too so this ends up being a bit of a rabbit hole...).

Something like:
```py
class NoOpProgress:
  def __enter__(self): return self
  def __exit__(self, *args, **kwargs): pass
  # any other methods that are called

try:
  from rich.progress import Progress
except ImportError:
  Progress = NoOpProgress
```

Depending on your bandwidth I think it'd be ok to define the dummy within the context of the function for now (this is mostly an MC thing right?), but we should aim for a more complete solution down the line. What do you think?

caseyflex:
Thanks @yaugenst-flex 

Converting this to draft as we need to clarify the requirements here. I now see that even `base.py` imports rich, so it might be that MC wants all dependencies on Tidy3d classes removed from the fitter. I have some thoughts about improving the fitter speed at the same time so it might be worth it, although it is a large effort. But first we need to clarify exactly what they want. Scipy is a pretty essential dependency at the moment too, which may be too heavy.

Regarding the general streamlining -- let's say we don't worry about imports for now. It would be nice to introduce classes in `log.py` to wrap rich `Progress`. In particular it would be nice to control whether a progress bar appears or not in the same way as our `log_level`, although possibly with a different setting because it's a bit independent from the rest of the logging.

caseyflex:
@yaugenst-flex Qingkun would like to get this in to the next version. It is sufficient for their purposes. Do you think it is good enough for now?

tidy3d/plugins/dispersion/fit_fast.py
```diff
@@ -145,15 +149,18 @@ def constant_loss_tangent_model(
...
-        return material
+        params, _ = constant_loss_tangent_model(
+            eps_real=eps_real,
+            loss_tangent=loss_tangent,
+            frequency_range=frequency_range,
+            max_num_poles=max_num_poles,
+            number_sampling_frequency=number_sampling_frequency,
+            tolerance_rms=tolerance_rms,
+            scale_factor=HBAR,
+        )
```
yaugenst-flex: Is `show_progress` not included here on purpose?

tidy3d/components/dispersion_fitter.py
```diff
@@ -866,88 +864,86 @@ def make_configs():
...
-                if best_model.rms_error < tolerance_rms:
-                    progress.update(
-                        task,
-                        completed=len(configs),
-                        description=f"Best weighted RMS error: {best_model.rms_error:.3g}",
-                        refresh=True,
-                    )
+                log.info(
+                    "Found optimal fit with weighted RMS error %.3g",
+                    best_model.rms_error,
```
yaugenst-flex: Consider turning into f-string?

tidy3d/components/dispersion_fitter.py
```diff
@@ -866,88 +864,86 @@ def make_configs():
...
-                    if best_model.show_unweighted_rms:
-                        log.info(
-                            "Unweighted RMS error %.3g",
-                            best_model.unweighted_rms_error,
-                        )
-                    return (
-                        best_model.pole_residue,
-                        best_model.rms_error,
+                        "Unweighted RMS error %.3g",
+                        best_model.unweighted_rms_error,
```
yaugenst-flex: Same here

yaugenst-flex (commented):
Thanks @caseyflex I think this all looks good, had some minor comments. Would be nice to add a test for `show_progress=True/False` though. Something along these lines:
```py
console_out = io.StringIO()
test_console = rich.console.Console(file=console_out)
with mock.patch("tidy3d.log.get_logging_console", return_value=test_console):
  constant_loss_tangent_model(..., show_progress=True)
  with_progress = console_out.getvalue()

  console_out.truncate(0)
  console_out.seek(0)

  constant_loss_tangent_model(..., show_progress=False)
  without_progress = console_out.getvalue()

assert len(with_progress) > len(without_progress)
```

Also, does this warrant a changelog entry or is it for internal use only?

tidy3d/plugins/dispersion/fit_fast.py
caseyflex: Yeah, I think for the plugin we can always use the default `show_progress=True`. Same thing with the changelog -- this change is really only for internal use (to make it lighter for use in GUI)

caseyflex: eh, but it can't hurt to include it here, so I might put it in

caseyflex:
Thanks @yaugenst-flex for the comments and especially the unit test suggestion. All should be addressed now

yaugenst-flex (approved):
LGTM!

PR #2259, opened 2025-02-19
Fix modeler.to_file when batch is empty

weiliangjin2021:


weiliangjin2021:
Currently, for a modeler that is initiated but without batch simulation data, `to_file` will error:

> Traceback (most recent call last):
>   File "..../tidy3d/plugins/smatrix/component_modelers/base.py", line 140, in to_file
>     jobs_cached = batch_cached._cached_properties.get("jobs")
>                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'NoneType' object has no attribute '_cached_properties'

tylerflex (approved):
looks good. just checking, do we need to raise a warning if the batch is none? or does it work as expected now without any warning needed?

weiliangjin2021:
I think if it's None, it'll submit the simulations to get batch data. My use case here is to modify a modeler and save it to file, and run later on.

dmarek-flex (approved):

PR #2263, opened 2025-02-20
Improve meshing and snapping points in lumped elements

weiliangjin2021:
Still many tests to fix and add, and CHANGELOG to add, but @dmarek-flex could you take a look first to see if the change makes sense?

Two main changes:

- Meshing part: snapping points along normal axis, and one override structure for the entire lumped element, with default grid number to 1.
- Mesh is generated automatically by moving lumped element meshing to grid_spec. The generated override structures and snapping points are considered internal.


tidy3d/components/lumped_element.py
```diff
@@ -159,49 +170,42 @@ def _snapping_spec(self) -> SnappingSpec:
...
+        plane of lumped element, and along normal axis. In the normal direction, we'll make sure there
         are at least 2 cell layers above and below whose size is the same as the in-plane cell
-        size in the override region. Also, to ensure that the port itself is aligned with a grid
-        boundary in the normal direction, two separate override regions are defined, one above
-        and one below the analytical port region.
+        size in the override region.
         """
 
         mesh_overrides = []
         if self.num_grid_cells:
```
dmarek-flex: Since we are making changes, I think it make sense to improve this comparison:

`if self.num_grid_cells is not None:`

tidy3d/components/lumped_element.py
```diff
@@ -380,24 +379,13 @@ def to_mesh_overrides(self) -> list[MeshOverrideStructure]:
             dl = (self.outer_diameter - self.inner_diameter) / self.num_grid_cells / 2
```
dmarek-flex: also change the comparison above here

tidy3d/components/lumped_element.py
```diff
@@ -159,49 +170,42 @@ def _snapping_spec(self) -> SnappingSpec:
...
-
-            override_center_below = list(self.center)
-            override_center_below[self.normal_axis] -= dl
+            override_size[self.normal_axis] = 4 * dl
             mesh_overrides.append(
                 MeshOverrideStructure(
-                    geometry=Box(center=override_center_below, size=override_size),
-                    dl=override_dl,
+                    geometry=Box(center=self.center, size=override_size),
+                    dl=(dl, dl, dl),
```
dmarek-flex: Hmm, I don't think we need to refine along the normal axis by default, as long as the snapping point is enabled correctly. But I do not have any evidence supporting my claim.

tidy3d/components/lumped_element.py
```diff
@@ -355,23 +359,18 @@ class CoaxialLumpedResistor(LumpedElement):
         "the orientation of the circles making up the coaxial lumped element.",
     )
 
-    def to_mesh_overrides(self) -> list[MeshOverrideStructure]:
-        """Creates a suggested :class:`MeshOverrideStructure` list that could be added to the
-        :class:`Simulation`.
-
-        Note
-        ----
+    def to_snapping_points(self) -> list[CoordinateOptional]:
```
dmarek-flex, line 371: As discussed off-line, to ensure accuracy of the voltage integral we could add snapping points at the min/max corners of the bounding boxes enclosing circles defined by the inner and outer diameters.

dmarek-flex (changes requested):
Looks like a nice improvement! Thanks. Please consider the changes to the coaxial lumped resistor.

tidy3d/components/lumped_element.py
weiliangjin2021: Previously, you mentioned that it's for more accurate current computation?

dmarek-flex: yes, I do vaguely remember that. Let's keep it the way it is then and modify if needed in the future!

dmarek-flex (approved):

weiliangjin2021:
@dmarek-flex I also updated LumpedPort to pass `enable_snapping_points`.

dbochkov-flexcompute (approved):
Looks great, didn't spot any issues. Sounds like this is an important step towards being able to properly define lumped ports without invoking a terminal modeler?

dmarek-flex (approved):

PR #2265, opened 2025-02-21
Increase the refresh time.

QingengWei:


QingengWei:
@momchil-flex The thread sleep time to small, the request will put a lot of pressure on the server.

PR #2266, opened 2025-02-24
Fix to _interpolated_tangential_fields when data has only one coordinate

momchil-flex:
The data translation #1526 PR had broken one of our mode solver backend tests, namely when the simulation is 2D. This is because calling interpolation over data that has a single coordinate along a given dimension returns nan values.


caseyflex (approved):
I see. Thanks for the fix!

PR #2268, opened 2025-02-24
Co- and cross-polarization for linear polarization

weiliangjin2021:


tidy3d/components/data/monitor_data.py
```diff
@@ -3519,6 +3519,20 @@ def right_polarization(self) -> FieldProjectionAngleDataArray:
         """
         return (self.Etheta - 1j * self.Ephi) / np.sqrt(2)
 
+    def linear_co_polarization(self, tilt_angle: float) -> DataArray:
```
dmarek-flex: This will do the trick, but a much more powerful feature would be to also allow for calculating partial directivity and partial gain with the same polarization definition. Currently, we support `linear` or `circular` polarization bases for these measurements, but if you combine them with a tilt angle you could cover all cases.

dmarek-flex: And by currently, I mean the PR that is not even merged yet :smile:

dmarek-flex: So one idea is adding the tilt angle as an argument to all the partial methods in `DirectivityData` and `AntennaParametersData` so that they can be computed for a desired reference polarization reference

weiliangjin2021: Good idea!

dmarek-flex: Awesome looks great!

weiliangjin2021:
@dmarek-flex I shifted one of your CHANGELOG item to `Added`, and amend it with the change in this PR.

dbochkov-flexcompute (approved):

tidy3d/components/data/monitor_data.py
```diff
@@ -3363,33 +3363,52 @@ def normalize(self, source_spectrum_fn: Callable[[float], complex]) -> Directivi
         return self.copy(update=dict(fields_norm, flux=new_flux))
 
     @staticmethod
-    def _check_valid_pol_basis(pol_basis: PolarizationBasis):
+    def _check_valid_pol_basis(pol_basis: PolarizationBasis, tilt_angle: float):
         if pol_basis != "linear" and pol_basis != "circular":
             raise ValueError("``pol_basis`` must be either 'linear' or 'circular'")
```
dmarek-flex: can you fix my typo here, I think I should have used single quotes around `pol_basis`

dmarek-flex (approved):

PR #2269, opened 2025-02-25
Refactored `bend_angle_rotation` to support general angle rotation

QimingFlex:


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Nonlinear models `KerrNonlinearity` and `TwoPhotonAbsorption` now default to using the physical real fields instead of complex fields.
 - Added warning when a lumped element is not completely within the simulation bounds, since now lumped elements will only have an effect on the `Simulation` when they are completely within the simulation bounds.
 - Allow `ModeData` to be passed to path integral computations in the `microwave` plugin.
+- `bend_angle_rotation` to `angle_rotation`, enabling general angle rotation for both straight and bend structures.
```
momchil-flex: Instead of adding a new item here, could you modify the existing `bend_angle_rotation` item? Basically, we are going to consolidate the rc changelogs into a single 2.8.0 changelong, so we don't need to track changes that happened while we were still developing this.

tidy3d/components/mode_spec.py
```diff
@@ -129,14 +129,14 @@ class ModeSpec(Tidy3dBaseModel):
         "yz plane, the ``bend_axis`` is always 1 (the global z axis).",
     )
 
-    bend_angle_rotation: bool = pd.Field(
+    angle_rotation: bool = pd.Field(
         False,
-        title="Use fields rotation when both bend and angle are defined",
-        description="Defines how modes are computed when both a bend and an angle are defined. "
-        "If `False`, the two coordinate transformations are directly composed. "
+        title="Use fields rotation when angle are defined",
```
momchil-flex: when 'angle_theta' is not zero

tidy3d/components/mode_spec.py
```diff
@@ -129,14 +129,14 @@ class ModeSpec(Tidy3dBaseModel):
...
+        description="Defines how modes are computed when an angle are defined. "
```
momchil-flex: when 'angle_theta' is not zero

tidy3d/components/mode_spec.py
```diff
@@ -129,14 +129,14 @@ class ModeSpec(Tidy3dBaseModel):
...
+        "If `False`, the coordinate transformations are directly composed. "
```
momchil-flex: If 'False', a coordinate transformation is applied through the permittivity and permeability tensors.

tidy3d/components/mode_spec.py
```diff
@@ -129,14 +129,14 @@ class ModeSpec(Tidy3dBaseModel):
...
         "If `True`, the structures in the simulation are first rotated to compute a mode solution at "
```
momchil-flex: If 'True'

' instead of `

momchil-flex (commented):
Thanks!

Some small comments on the wording. To make sure this works, could you update the backend test as well and add some tests for the case with only angle but no bend?

QimingFlex:
[rotation_test.zip](https://github.com/user-attachments/files/19028404/rotation_test.zip)
I added a notebook comparing mode source and custom source using the mode solver solution for 0 injection. The latter resulted in a small-level backward propagating wave. This should be the reason of unclean injections in some cases.

momchil-flex (approved):

PR #2270, opened 2025-02-25
Add testing for the monitor's frequency ranges to be consistent with the sources frequency ranges

vasilyzabelin:
This PR updates the validator for the monitor's frequency ranges to be consistent with the frequency range defined by the sources. 
It generates the warning if the monitor frequency is too high or too low.
The corresponding pre-load validator generates the error in this case.


tidy3d/components/simulation.py
```diff
@@ -2729,19 +2731,26 @@ def _warn_monitor_simulation_frequency_range(cls, val, values):
...
-        freq_min = min((freq_range[0] for freq_range in source_ranges), default=0.0)
-        freq_max = max((freq_range[1] for freq_range in source_ranges), default=0.0)
+        freq_range_min = (
+            min((freq_range[0] for freq_range in source_ranges), default=0.0)
+            * MIN_MONITOR_FREQUENCY_RANGE_PARAMETER
+        )
+        freq_range_max = (
+            max((freq_range[1] for freq_range in source_ranges), default=0.0)
+            * MAX_MONITOR_FREQUENCY_RANGE_PARAMETER
+        )
```
momchil-flex: To clarify my initial description of how this should be implemented, the warning behavior should stay the same. We should always warn if the monitor frequencies are outside the source bandwidth, because it's likely that there isn't much power injected there and the results will be more strongly subject to numerical artifacts. Still, it's only a warning, because in some cases - e.g. we allow nonlinear materials, so there could be frequency conversion - this could be a correct setup.

On the other hand, we want to raise the pre-upload error based on the `MAX_MONITOR_FREQUENCY_RANGE_PARAMETER`, because if the monitor frequency is e.g. 10x away from the source bandwidth, then there really isn't any currently supported correct setup.

Finally, on a related note, the test should be updated to make sure that a warning is issues when the mismatch is small, and an error is issued when it is large.

tidy3d/components/simulation.py
```diff
@@ -3570,6 +3580,33 @@ def _validate_time_monitors_num_steps(self) -> None:
...
+
+        for monitor in self.monitors:
+            if not isinstance(monitor, FreqMonitor):
+                continue
+
+            freqs = np.array(monitor.freqs)
+            if freqs.min() < freq_range_min or freqs.max() > freq_range_max:
+                raise SetupError(
+                    f"Time monitor '{monitor.name}' contains frequencies "
```
momchil-flex: Just `Monitor`, or `Frequency-domain monitor` if you want to specify something. This doesn't apply to time monitors.

tidy3d/components/simulation.py
```diff
@@ -3570,6 +3580,33 @@ def _validate_time_monitors_num_steps(self) -> None:
...
+                    f"outside of the simulation frequency range ({freq_range_min:2e}, {freq_range_max:2e})"
```
momchil-flex, line 3844: We should also specify the margin, i.e. the `MAX_MONITOR_FREQUENCY_RANGE_PARAMETER`.

tidy3d/components/simulation.py
```diff
@@ -2729,19 +2731,26 @@ def _warn_monitor_simulation_frequency_range(cls, val, values):
...
                     continue
 
                 freqs = np.array(monitor.freqs)
-                if freqs.min() < freq_min or freqs.max() > freq_max:
+                if freqs.min() < freq_range_min or freqs.max() > freq_range_max:
                     consolidated_logger.warning(
                         f"monitors[{monitor_index}] contains frequencies "
-                        f"outside of the simulation frequency range ({freq_min:2e}, {freq_max:2e})"
+                        f"outside of the simulation frequency range ({freq_range_min:2e}, {freq_range_max:2e})"
+                        f"outside of the simulation frequency range ({freq_range_min:2e}, {freq_range_max:2e})"
```
tylerflex: duplicate line here?

PR #2271, opened 2025-02-25
:wrench: :robot: Consolidate coverage tests

daquinteroflex:


.github/workflows/run_tests.yml
```diff
@@ -78,9 +78,27 @@ jobs:
...
       env:
         PYTHONUNBUFFERED: "1"
       run: |
-        poetry run pytest -rF --tb=short tests/_test_data/_test_datasets_no_vtk.py
-        poetry run pytest -rF --tb=short tests
+        poetry run pytest --cov=tidy3d -rF --tb=short tests/_test_data/_test_datasets_no_vtk.py
+        poetry run pytest --cov=tidy3d -rF --tb=short tests
+        poetry run coverage report -m
+        echo "total=$(poetry run coverage report --format=total)" >> $GITHUB_ENV
+        echo '### Total coverage: ${{ env.total }}%'
```
yaugenst-flex: Sure this works correctly? The test run output looks like this:
```
TOTAL                                                        31230   3048    90%
### Total coverage: %
```

yaugenst-flex (approved):
Thanks @daquinteroflex! Maybe double-check whether the env var is set correctly for the total coverage but otherwise looks great!

.github/workflows/run_tests.yml
daquinteroflex: Apparently our previous setup worked like [this too](https://github.com/flexcompute/tidy3d/actions/runs/13517970772/job/37770604302#step:8:45138) but yeah, will fix that to be sure! It might be because the job environment action hasn't finished, and it's definitely a development bug likely

yaugenst-flex: Ah I see, yeah maybe the github envs are only updated after a step is complete? Not sure how it works

PR #2272, opened 2025-02-25
Frequency-dependent rotation to constant in-plane k beam

momchil-flex:
Another fix to how we get the beam fields. Now, `points` has been extended to have shape `[3, Npoints, Nfreqs]`, and in the case of constant in-plane k wave the points are rotated with the actual theta at every frequency. This made things cleaner and actually allowed me to clean up all of the `np.outer` and `[:, None]` type of handling we had in the scalar field functions, as everything just works with regular broadcasting. And of course the main reason for this was to fix the plane wave injection with backward direction, which indeed is now fixed.

One last thing that I don't understand about fixed angle is why we don't rotate the points, but then we rotate the fields? That's not super intuitive to me and ideally I would like to generalize things so that the points are always rotated (propagation direction is always z). I actually can understand why we don't rotate the points - in some sense the "actual angle" is zero (and we just modify the kz). But why do we rotate the fields by `angle_theta`?

### DIFF
--- a/tidy3d/components/beam.py
+++ b/tidy3d/components/beam.py
@@ -174,19 +174,21 @@ class BeamProfile(Box):
     ) -> Numpy:
         """Analytic beam with all the beam parameters but assuming ``z`` as the normal axis."""
 
+        # Add a frequency dimension to points
+        points = np.repeat(points[:, :, np.newaxis], len(self.freqs), axis=2)
+
         # Rotate points to axes where propagation is along z
-        points_prop_z = self._rotate_points_z(points)
+        points_prop_z = self._rotate_points_z(points, background_n)
 
         # Reflection at the z = 0 plane for negative direction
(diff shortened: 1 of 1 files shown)


dbochkov-flexcompute (approved):
Looks great! When I was looking at it also thought that maybe we need to make points frequency dependent.

Regarding rotation of fields for fixed angle case, I think we could give the following explanantion: 

What we would like to model is injection of $\vec{A} \exp(j \vec{r}\cdot \vec{k})$, where information about angle sits both in $\vec{A}$ and $\vec{k}$. The fixed angle treatment only modifies $\exp(j \vec{r}\cdot \vec{k})$ part, while $\vec{A}$ remains rotated. Maybe it's not the best way to think about it in terms of not rotating points but rotating fields, but it provides the desired outcome

PR #2274, opened 2025-02-26
Added validation for the monitors frequency range to be consistent with the frequency range defined by the sources.

vasilyzabelin:
Added validation for the monitors frequency range to be consistent with the frequency range defined for the sources.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -158,6 +158,8 @@ WARN_MONITOR_DATA_SIZE_GB = 10
 MAX_MONITOR_INTERNAL_DATA_SIZE_GB = 50
 MAX_SIMULATION_DATA_SIZE_GB = 50
 WARN_MODE_NUM_CELLS = 1e5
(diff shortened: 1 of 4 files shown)


tidy3d/components/simulation.py
```diff
@@ -2994,7 +2996,7 @@ def _warn_monitor_simulation_frequency_range(cls, val, values):
                 freqs = np.array(monitor.freqs)
                 if freqs.min() < freq_min or freqs.max() > freq_max:
                     consolidated_logger.warning(
-                        f"monitors[{monitor_index}] contains frequencies "
+                        f"Monitors[{monitor_index}] contains frequencies "
```
momchil-flex: This was actually not capitalized before it refers to `Simulation.monitors`, but let's put quotes around it: `'monitors[{monitor_index}]'` to make it clear it refers to a variable.

momchil-flex (commented):
Looks good, only one test is failing. When you fix that please also squash into single commit.

momchil-flex (approved):

PR #2280, opened 2025-02-27
add viz_spec property for plotting structures that have no medium

groberts-flex:
Fixes error when plotting a structure with no medium where a viz_spec was trying to be pulled from the nonexistent medium. This defaults an abstract structure to have no viz_spec and ones with underlying mediums to get their viz_spec from their medium.


momchil-flex (approved):
Awesome, thanks

PR #2282, opened 2025-02-27
Fix adjoint source creation

groberts-flex:
Currently, in `setup_adj`, the traced fields are inserted into the original simulation data. The fields in this simulation data control what adjoint sources are being created based on the monitor data functions `make_adjoint_sources`. Since the other fields are not zero'd, what is currently happening is that the traced field gradients replace the current simulation data and create correct adjoint sources, but the other simulation data that is not zero'd also is creating adjoint sources. 

The fix here is to zero out the simulation data when no traced field is present and to keep the insertion of the traced gradients the same so that we are only creating the adjoint sources we want.

### DIFF
--- a/tidy3d/web/api/autograd/autograd.py
+++ b/tidy3d/web/api/autograd/autograd.py
@@ -861,8 +861,19 @@ def setup_adj(
     # immediately filter out any data_vjps with all 0's in the data
     data_fields_vjp = {key: get_static(value) for key, value in data_fields_vjp.items()}
 
+    # start with the full simulation data structure and either zero out the fields
+    # that have no tracer data for them or insert the tracer data
+    full_sim_data_dict = sim_data_orig.strip_traced_fields(
+        include_untraced_data_arrays=True, starting_path=("data",)
+    )
+    for path in full_sim_data_dict.keys():
+        if path in data_fields_vjp:
+            full_sim_data_dict[path] = data_fields_vjp[path]
+        else:
+            full_sim_data_dict[path] *= 0
+
     # insert the raw VJP data into the .data of the original SimulationData
-    sim_data_vjp = sim_data_orig.insert_traced_fields(field_mapping=data_fields_vjp)
+    sim_data_vjp = sim_data_orig.insert_traced_fields(field_mapping=full_sim_data_dict)
 
     # make adjoint simulation from that SimulationData
     data_vjp_paths = set(data_fields_vjp.keys())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed `ComponentModeler.to_file` when its batch is empty.
 - In web api, mode solver is patched with remote data so that certain methods like `plot_field` show remote data.
 - Added `viz_spec` property to `AbstractStructure` to fix error when plotting structures that have no `medium`.
+- Fixed invalid adjoint source creation by disallowing sources made from non-traced fields.
 
 ## [2.8.0rc2] - 2025-01-28
 


groberts-flex:
Getting some test failures here that I wasn't seeing on another test branch! Looking into them!

tylerflex (commented):
Thanks Greg, I'll take a closer look tomorrow. But I thought this was already taken care of? Otherwise it would seem the existing code would cause a significant error in the gradient accuracy? Do you have a test showing this in action? also how was this uncovered?

groberts-flex:
Thanks for taking a look Tyler! I found this when I was testing out a sample optimization in the inverse design plugin as part of testing the refactoring I've been doing. I was trying an optimization where I was maximizing flux for one frequency through a small zone and flux for another frequency through a different zone (free space color splitter type of design). The gradient from the autograd calculation was causing the objective value to decrease regardless of the step size I was using. I started simplifying and removing from the plugin to isolate and simplified to a case of what should be dipole adjoint sources. Originally I found that if I included more than one frequency in the point monitor but used only one, I got a different gradient than if I included just the one I was using. That's when I started looking at the adjoint sources that were getting created. From here, I also noticed that when I was just using an objective function that depended on Ex, I was also seeing Jy and Jz sources come up. I eventually tracked this down to the place where I made the fix above. I've attached some code to reproduce this (had to attach as a .txt file because GitHub wouldn't let me upload a .py)
[dipole_adjoint_source_error.txt](https://github.com/user-attachments/files/19019481/dipole_adjoint_source_error.txt)

groberts-flex:
Attached are examples of the computed gradient for the above code with the same objective when including just one frequency in the monitors used in the objective function calculation versus when I include additional ones that are not used.
gradient_additional_freq_in_monitors
gradient_just_one_freq_in_monitors

groberts-flex:
And attached below is a screenshot of one of the adjoint sources created in Ez with nonzero amplitude.

ez_dipole_for_ex_obj_function

tylerflex:
Wow, great catch!

tidy3d/web/api/autograd/autograd.py
```diff
@@ -861,8 +861,19 @@ def setup_adj(
     # immediately filter out any data_vjps with all 0's in the data
     data_fields_vjp = {key: get_static(value) for key, value in data_fields_vjp.items()}
 
+    # start with the full simulation data structure and either zero out the fields
+    # that have no tracer data for them or insert the tracer data
+    full_sim_data_dict = sim_data_orig.strip_traced_fields(
```
tylerflex, line 866: could this be an alternative approach? (only new code is "clearing" the `data` of untraced fields)

```py
# get all of the adjoint source data
data_fields_vjp = {key: get_static(value) for key, value in data_fields_vjp.items()}

# remove all of the untraced data
traced_data = [data for data in sim_data_orig.data if data.monitor.name in data_fields_vjp]
sim_data_vjp = sim_data_orig.updated_copy(data=traced_data, deep=False)

# insert the VJP (as before)
sim_data_vjp = sim_data_vjp.insert_traced_fields(field_mapping=data_fields_vjp)
```

I suppose a downside of this is that it introduces an additional copy, but maybe because it is shallow it is not so bad? Upside is that it's bit simpler to understand.

Just a thought, I think what you have also works.

groberts-flex: Thanks for the suggestion, I like the upside of easier to understand.

The issue I'm running into with this approach is we actually are needing to clear certain fields in each monitor data according to the data_fields_vjp keys instead of just the whole data associated with a monitor. The monitor name is not a key into the data_fields_vjp because that key contains information on a specific field in the monitor data (i.e. - an example of one of the keys in data_fields_vjp looks like ('data', 0, 'Ex'))

tylerflex: oh right, I've already forgotten a bit how this data is stored. I guess it could look rather like

```py
traced_indices = {index for _, index, * in data_fields_vjp.keys()}
traced_data = [sim_data_orig.data[i] for i in traced_indices]
sim_data_vjp = sim_data_orig.updated_copy(data=traced_data, deep=False)
```

or does that not work either?

tylerflex: side note: might need to make sure `traced_indices` is sorted before iterating through it to avoid potential issues.

groberts-flex: `traced_indices = {index for _, index, * in data_fields_vjp.keys()}`
Is the `*` here intended to also be a `_`?

This isn't working when I test it either because while it strips the index correctly to pull out a specific monitor data from the simulation data, it still doesn't zero out fields inside that monitor data that are not present in the vjp data. Those non zero'd fields then get turned into adjoint sources.

tylerflex: sorry, it should be a `*_`, basically to unpack everything after the `index`

> it still doesn't zero out fields inside that monitor data that are not present in the vjp data

oh, interesting. Yea maybe in that case what you propose is ultimately the correct approach. thanks for going through that with me. I think it makes sense now. I've started somewhat forgetting how all of this works after spending more time away from it.

tylerflex (approved):

tidy3d/web/api/autograd/autograd.py
groberts-flex: Gotcha, that makes sense on the syntax - I hadn't used that unpack before!

And definitely, no problem! I'm definitely still getting up to speed with a lot of how all this stuff is passed around and connects to the autograd library. It's quite the feat how seamless this stuff is for the user! I'm open to other approaches to this as well, but keeping with how the current adjoint source creation works inside monitor data, this was the best way I could come up with for now that seemed to get the job done. Yannick and I discussed this morning a potential different approach this morning in our 1:1 of using the vjp data directly to create the adjoint sources instead of filling in the monitor data as an intermediate.

yaugenst-flex (approved):
Awesome, thanks @groberts-flex! Looks good to me as we just discussed.

PR #2285, opened 2025-02-28
Simulation.grid_info property for important grid information

weiliangjin2021:
A convenient property to show all relevant meshing information that I frequently use in benchmark. However, the print doesn't look neat. Any suggestions?

```
>>> print(sim.grid_info)

<xarray.Dataset> Size: 56B
Dimensions:                   ()
Data variables:
    Nx                        int64 8B 41
    Ny                        int64 8B 41
    Nz                        int64 8B 7
    total_cells               int64 8B 11767
    min_grid_size             float64 8B 1.429e+03
    max_grid_size             float64 8B 2.997e+03
    computational_complexity  float64 8B 8.237
```


dmarek-flex:
> Any suggestions?

Nice! I was using a Pandas dataframe for the lobe measuring tool, which is nicely formatted

weiliangjin2021:
> Nice! I was using a Pandas dataframe for the lobe measuring tool, which is nicely formatted

It asks for indexing, but here it's all scalar data. Maybe a simple python Dict is good enough?

weiliangjin2021:
Now rendered quite well in jupyter:
image

yaugenst-flex (approved):
I think this is fine, just one thing we could think about adding is a custom `__str__` method for `Grid`. So instead of relying on the `info` property, `__str__` could return a nicely formatted string assembled from the `info`. Then you could simply call `print(grid)` and it would give you this output. You'd then also not need to add the additional grid-specific method to `Simulation`, but instead you'd just `print(sim.grid)`. We could think about similar "nice" representations of the data to other classes where it makes sense.

weiliangjin2021:
> I think this is fine, just one thing we could think about adding is a custom `__str__` method for `Grid`. So instead of relying on the `info` property, `__str__` could return a nicely formatted string assembled from the `info`. Then you could simply call `print(grid)` and it would give you this output. You'd then also not need to add the additional grid-specific method to `Simulation`, but instead you'd just `print(sim.grid)`. We could think about similar "nice" representations of the data to other classes where it makes sense.

That's a good idea. I'll revise it along this line!

weiliangjin2021:
Hmm, perhaps I'll leave it as it is for now, as sometimes it's more useful to obtain values than just viewing the string?

yaugenst-flex:
Oh yeah I agree, I meant keeping the dict but adding the `__str__`. The str method would get its info from the dict. But it would remove the need for a modification in the simulation class. And if you really need to retrieve the grid dict, you could always call it as `sim.grid.grid_info`.
But no strong opinion on it in this context (already approved ;) ). Just think we should probably consider better printing in general.

PR #2287, opened 2025-02-28
refc[adjoint]: filter zero VJPs in setup_adj

yaugenst-flex:
This moves the zero VJP filtering into `setup_adj` so both `_run_bwd` and `_run_async_bwd` reuse the same logic.

Note no changelog item for this one since it's mostly an internal refactor.


groberts-flex (approved):
This looks good to me! I like having this logic all in one spot!

PR #2288, opened 2025-03-01
optimize abort task

QingengWei:


momchil-flex (commented):
In which case might the task not be found?

This doesn't really fix SCEM-9503 does it?

QingengWei:
> In which case might the task not be found?
> 
> This doesn't really fix SCEM-9503 does it?

#1. The task has been deleted or it's not in DB or due to some reasons (network problems), the query cannot be found from the DB.
#2 No need to fix, I can't reproduce it.

This PR is to make the error message more friendly.

momchil-flex:
Ok, thanks. Note we've merged pre/2.8 into develop so PRs should now go into develop branch.

PR #2289, opened 2025-03-03
Adjusting corss-medium check to SolidMedium

marc-flex:
This fix should address one of the issues in https://github.com/flexcompute/compute/pull/998 where `test_conduction_2d` test would fail if a Multiphysics medium is used instead of the SolidMedium.

### DIFF
--- a/tidy3d/components/material/multi_physics.py
+++ b/tidy3d/components/material/multi_physics.py
@@ -102,7 +102,8 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
     def heat_spec(self):
         if self.heat is not None:
             return self.heat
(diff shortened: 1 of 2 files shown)


tidy3d/components/material/multi_physics.py
```diff
@@ -102,7 +102,8 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
     def heat_spec(self):
         if self.heat is not None:
             return self.heat
-        elif self.optical is not None:
+        
+        if self.optical is not None:
```
momchil-flex, line 106: Ah, the joy of carrying backward compatibility...

momchil-flex (approved):

PR #2291, opened 2025-03-04
optimize abort task

QingengWei:


QingengWei:
As discussed with Leixin, Calling this interface is to try to abort the task. The final result can be viewed on the GUI. So some guidance words need to be added.

PR #2293, opened 2025-03-04
Raise an error if far_field_approx=False enabled for 2D simulations.

QimingFlex:


tidy3d/components/simulation.py
```diff
@@ -3258,6 +3258,12 @@ def _projection_mnts_2d(cls, val, values):
                         f"Monitor '{monitor.name}' is not supported in 1D simulations."
                     )
 
+                if not monitor.far_field_approx:
+                    raise SetupError(
+                        f"Exact far-field projection for 2D simulations is not yet available for Monitor '{monitor.name}'. "
+                        "Currently, only `far_field_approx = True` is supported."
```
momchil-flex: We don't use markdown syntax in error messages, so just `'far_field_approx = True'`.

momchil-flex (approved):
Thanks @QimingFlex! Just a tiny syntax note.

dmarek-flex:
@momchil-flex This should have been merged into `develop` right?

momchil-flex:
Good catch, I'll cherry-pick

PR #2295, opened 2025-03-06
fix[autograd]: add validation for missing frequency-domain monitors

yaugenst-flex:
Also adds `validate_pre_upload()` check for remote gradient runs to `run_async` (previously only in `run`)

### DIFF
--- a/tidy3d/web/api/autograd/autograd.py
+++ b/tidy3d/web/api/autograd/autograd.py
@@ -16,6 +16,7 @@ from tidy3d.components.autograd import AutogradFieldMap, get_static
 from tidy3d.components.autograd.derivative_utils import DerivativeInfo
 from tidy3d.components.types import Literal
 
+from ....exceptions import AdjointError
 from ...core.s3utils import download_file, upload_file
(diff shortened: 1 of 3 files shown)


tidy3d/web/api/autograd/autograd.py
```diff
@@ -61,6 +61,15 @@ def is_valid_for_autograd(simulation: td.Simulation) -> bool:
...
         return False
 
+    # if no frequency-domain data (e.g. only field time monitors), raise an error
+    if not simulation.freqs_adjoint:
+        msg = (
+            "No frequency-domain data found in simulation, but found traced structures. "
+            "For an autograd run, you must have at least one frequency-domain monitor."
+        )
+        td.log.error(msg)
+        raise ValueError(msg)
```
momchil-flex: What's the thinking here about not using tidy3d Exceptions?

yaugenst-flex: Ah good catch, there is none :sweat_smile:

momchil-flex (approved):

PR #2299, opened 2025-03-07
a minimum refactor of optimizer and design region

tylerflex:
just a sketch

1. every InverseDesign has a list of design regions.
2. Optimizer now is the container for the entire optimization.
3. Each design region has an OptimizationSpec, corresponding to specifics of optimization updates.

to do potentially:
combine InverseDesign and INverseDesignMulti into one class, handle simulations as a list always but maybe provide nicer api


yaugenst-flex:
Closing since this was only for discussion.

PR #2303, opened 2025-03-10
Fix corner filtering in LayerRefinementSpec

weiliangjin2021:


lucas-flexcompute (approved):
I don't usually like double negatives... it'd be clearer (in my opinion) to have this as `_is_inplane_bounded`. This being a private method, you might be allowed to change it. This is just my opinion, though.

weiliangjin2021:
> I don't usually like double negatives... it'd be clearer (in my opinion) to have this as `_is_inplane_bounded`. This being a private method, you might be allowed to change it. This is just my opinion, though.

Good point, I'll change that.

PR #2304, opened 2025-03-11
Add function to obtain cell values in UnstructuredGridDataset

marc-flex:
Adding a function to compute cell values for a given `UnstructuredGridDataset`.  This is related to issue https://github.com/flexcompute/tidy3d-core/issues/880

Since this function `get_cell_values()` returns the cell values and volumes associated with each cell, integration of a quantity over the volume is a straightforward dot-product.

### DIFF
--- a/tidy3d/components/data/unstructured/base.py
+++ b/tidy3d/components/data/unstructured/base.py
@@ -641,6 +641,20 @@ class UnstructuredGridDataset(Dataset, np.lib.mixins.NDArrayOperatorsMixin, ABC)
 
         return values
 
+    def get_cell_values(self, **kwargs):
+        """This function returns the cell values for the fields stored in the UnstructuredGridDataset.
+        If multiple fields are stored per point, like in an IndexedVoltageDataArray, cell values
+        will be provided for each of the fields unless a selection argument is provided, e.g., voltage=0.2
+        """
+
+        values = self.values.sel(**kwargs)
+
+        return values[self.cells].mean(dim="vertex_index").values
+
+    @abstractmethod
+    def get_cell_volumes(self):
+        """Get the volumes associated to each cell."""
+
     """ Grid operations """
 
     @requires_vtk
--- a/tests/test_data/test_datasets.py
+++ b/tests/test_data/test_datasets.py
@@ -667,3 +667,75 @@ def test_triangular_dataset_uniform():
 
     tri_grid = tri_grid.updated_copy(values=tri_grid_values)
     assert not tri_grid.is_uniform
+
+
+def test_cell_values():
+    """Test whether the cell values are correctly calculated"""
+    import tidy3d as td
+
+    # start with a triangle grid
+    tri_grid_points = td.PointDataArray(
+        [[0.0, 0.0], [1.0, 0.0], [0.0, 1.0], [1.0, 1.0]],
+        dims=("index", "axis"),
+    )
+
+    tri_grid_cells = td.CellDataArray(
+        [[0, 1, 2], [1, 2, 3]],
+        dims=("cell_index", "vertex_index"),
+    )
+
+    tri_grid_values = td.IndexedVoltageDataArray(
+        [[0.0, 0.0], [0, 0], [3, -3], [3, -3]],
+        coords=dict(index=np.arange(4), voltage=[-1, 1]),
+        name="test",
+    )
+
+    tri_grid = td.TriangularGridDataset(
+        normal_axis=1,
+        normal_pos=0,
+        points=tri_grid_points,
+        cells=tri_grid_cells,
+        values=tri_grid_values,
+    )
+
+    cell_values = tri_grid.get_cell_values(voltage=-1)
+    cell_vols = tri_grid.get_cell_volumes()
+    assert np.dot(cell_values, cell_vols) == 1.5
+
+    # Now repeat for a tet mesh
+    tet_grid_points = td.PointDataArray(
+        [
+            [0.0, 0.0, 0.0],
+            [1.0, 0.0, 0.0],
+            [0.0, 1.0, 0.0],
+            [1.0, 1.0, 0.0],
+            [0.0, 0.0, 1.0],
+            [1.0, 0.0, 1.0],
+            [0.0, 1.0, 1.0],
+            [1.0, 1.0, 1.0],
+        ],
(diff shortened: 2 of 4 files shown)


dbochkov-flexcompute (changes requested):
looks good so far
- does this work for datasets with multiple fields? like when we have several voltages in one dataset?
- I would probably separate volume calculation from `get_cell_values` into a stand-alone `get_cell_volumes`. For example, currently in backend everytime you call `get_cell_values` it calculates both cell values and volumes, but volumes doesn't change between different voltages and different fields.
- Do you know if cell values are just averages of corner values of a cell? I think we can easily calculate it without asking vtk for that as `cell_values = self.values[self.cells].mean(dim="vertex_index")` and that should work automatically for datasets with multiple voltages as well, here's a quick example:
  ```
  import tidy3d as td
  import numpy as np
  
  tri_grid_points = td.PointDataArray(
      [[0.0, 0.0], [1.0, 0.0], [0.0, 1.0], [1.0, 1.0]],
      dims=("index", "axis"),
  )
  
  tri_grid_cells = td.CellDataArray(
      [[0, 1, 2], [1, 2, 3]],
      dims=("cell_index", "vertex_index"),
  )
  
  tri_grid_values = td.IndexedVoltageDataArray(
      np.transpose([[0.1, 0.2, 3.0, 3.4], [-0.1, -0.2, -3.0, -3.4]]), coords=dict(index=np.arange(4), voltage=[-1, 1]), name="test"
  )
  
  tri_grid = td.TriangularGridDataset(
      normal_axis=1,
      normal_pos=0,
      points=tri_grid_points,
      cells=tri_grid_cells,
      values=tri_grid_values,
  )
  
  print(tri_grid_values[tri_grid_cells].mean(dim="vertex_index"))
  ```

dbochkov-flexcompute:
Likewise, we can calculate areas of triagular cells as
```
v0 = self.points[self.cells.sel(vertex_index=0)]
e01 = self.points[self.cells.sel(vertex_index=1)] - v0
e02 = self.points[self.cells.sel(vertex_index=2)] - v0

cell_areas = 0.5 * np.abs(np.cross(e01, e02))
```
and volumes of tetrahedral cells as
```
v0 = self.points[self.cells.sel(vertex_index=0)]
e01 = self.points[self.cells.sel(vertex_index=1)] - v0
e02 = self.points[self.cells.sel(vertex_index=2)] - v0
e03 = self.points[self.cells.sel(vertex_index=3)] - v0

cell_volumes = np.abs(np.sum(np.cross(e01, e02) * e03, axis=1)) / 6
```

dbochkov-flexcompute:
or if we were to try to use the least number of lines :sweat_smile:
```
edges = self.points[self.cells.sel(vertex_index=[1, 2])] - self.points[self.cells.sel(vertex_index=0)]
cell_areas = 0.5 * np.abs(np.cross(edges.isel(vertex_index=0), edges.isel(vertex_index=1)))
```
```
edges = self.points[self.cells.sel(vertex_index=[1, 2, 3])] - self.points[self.cells.sel(vertex_index=0)]
cell_volumes = np.abs(np.sum(np.cross(edges.isel(vertex_index=0), edges.isel(vertex_index=1)) * edges.isel(vertex_index=2), axis=1)) / 6
```

marc-flex:
> Do you know if cell values are just averages of corner values of a cell? I think we can easily calculate it without asking vtk for that as cell_values = self.values[self.cells].mean(dim="vertex_index") and that should work automatically for datasets with multiple voltages as well, here's a quick example

vtk seems to be doing linear interpolation which is why I'd probably leave it with vtk. 

I didn't try with multiple fields.  Let me experiment with it a little and I'll address the rest of the comments as well

marc-flex:
Correction: vtk does seem to be doing means, so I'll change to what you suggest @dbochkov-flexcompute

marc-flex:
With this latest commit, when we have multiple fields like when we have `IndexedVoltageDataArrays` in values, if nothing is specified, it will output cell values for each of the fields. However, I have added `kwargs` so that one can select, e.g., `voltage=0.1`, in which case only the cell values for that field will be output.

tidy3d/components/data/unstructured/base.py
```diff
@@ -641,6 +641,42 @@ def _get_values_from_vtk(
...
+
+        return values[self.cells].mean(dim="vertex_index").values
+
+    def get_cell_volumes(self):
+        """Get the volumes of each cell in the vtk object."""
+
+        # determine whether the UnstructuredGridDataset is 2D or 3D
+        is_3D = True
+        if self._point_dims() == 2:
+            is_3D = False
```
dbochkov-flexcompute: More of a nitpicking. I think we should make `get_cell_volumes` an abstract function with concrete implementations in `TriangularGridDataset` and `TetrahedralGridDatasets`

tidy3d/components/data/unstructured/base.py
```diff
@@ -641,6 +641,42 @@ def _get_values_from_vtk(
...
+    def get_cell_values(self, **kwargs):
+        """This function returns the cell values for the fields stored in the UnstructuredGridDataset.
+        If multiple fields are stored per point, like in an IndexedVoltageDataArray, cell values
+        will be provided for each of the fields unless a selection argument is provided, e.g., voltage=0.2
+        """
+
+        values = self.values
+        if len(self._fields_shape) > 0:
+            if kwargs is not None:
+                values = self.values.sel(**kwargs)
```
dbochkov-flexcompute: I wonder if we should execute it this even if `len(self._fields_shape) == 0`, so that it fails if we try to provide any sel kwargs. Otherwise, it can happen that we provide something, it doesn't error, and we conclude that it does what we thought it should, but in reality it just ignored our arguments

dbochkov-flexcompute (commented):
I think this is very close, just a couple of minor comments

tidy3d/components/data/unstructured/base.py
marc-flex: That makes a lot of sense, actually. Thanks Daniil!

tidy3d/components/data/unstructured/base.py
marc-flex: That's a good point :+1:

dbochkov-flexcompute (approved):

momchil-flex (approved):

PR #2306, opened 2025-03-12
Unstructured mesh refinement regions

momchil-flex:


CHANGELOG.md
```diff
@@ -22,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Validation step that raises a `ValueError` when no frequency-domain monitors are present, preventing invalid adjoint runs.
 - Metal surface roughness models: modified Hammerstad, Huray Snowball, and Cannonball-Huray.
 - Support for Fermi-Dirac statistics in Charge solver. This option can be activated when defining the analysis type `IsothermalSteadyChargeDCAnalysis` with `fermi_dirac=True`. This option will provide more accurate results in simulations where very high doping may lead the pseudo-Fermi energy levels to approach either the conduction or the valence energy levels.
+- Add refinement regions and refinement lines for the mesh generation for teh CHARGE solver.
```
marc-flex: This can also be used to Heat simulations so I would change the sentence to this:
`Unstructured grids now support refinement regions (including 1D refinement lines in arbitrary directions)`

tidy3d/components/tcad/grid.py
```diff
@@ -60,6 +64,91 @@ class UniformUnstructuredGrid(UnstructuredGrid):
...
+
+    r1: Coordinate = pd.Field(
+        (0.0, 0.0, 0.0),
+        title="Start point of the line",
+        description="Start point of the line in x, y, and z.",
+        units=MICROMETER,
+    )
+
+    r2: Coordinate = pd.Field(
+        (0.0, 0.0, 0.0),
```
marc-flex: I think we should make this a required input and not provide a default.

tidy3d/components/tcad/grid.py
```diff
@@ -126,6 +215,18 @@ class DistanceUnstructuredGrid(UnstructuredGrid):
...
+        title="Refinement regions",
+        description="List of regions for which the mesh refinement will be applied",
+    )
+
+    refinement_lines: Tuple[GridRefinementLine, ...] = pd.Field(
+        (),
+        title="Refinement lines",
+        description="List of lines for which the mesh refinement will be applied",
+    )
+
```
marc-flex: Now that I see this I'm thinking that maybe `refinement_regions` could accept both `GridRefinementRegion` and `GridRefinementLine`. Conceptually they're both refinement regions in a way.

Of course this implies updating the back-end.

momchil-flex: Makes sense, could be a single argument `refinements` even.

dbochkov-flexcompute: there is also the option to reuse name `override_structures` from fdtd grid spec

vasilyzabelin:
There is a new branch for this PR: vasily/refinement_region_update

It should be merged into this PR

tidy3d/components/tcad/grid.py
marc-flex: that makes a lot of sense

vasilyzabelin:
I have created a new clean PR with the mesh refinement structures:
[#2398](https://github.com/flexcompute/tidy3d/pull/2398)
It turned out easier that to clean the existing PR from unnecessary commits.

momchil-flex:
closing in favor of #2398

PR #2309, opened 2025-03-12
Add aux field monitors

caseyflex:
Adds monitor class `AuxFieldTimeMonitor` and corresponding data classes `AuxFieldTimeData` and `AuxFieldTimeDataset`. Currently it is modeled after `FieldTimeMonitor` -- you specify some aux field components to download, and it puts those in the corresponding data class. The default is no aux field components. Any field components which don't correspond to nonlinearities which are present in the region / material / simulation are just stored as 0.

Todo:

- [x] Add frontend unit tests
- [x] Finalize api
- [x] Maybe validate against requesting nonlinear aux field components for nonlinearities not present in the simulation. This was the purpose of the function `_aux_fields` in `Simulation`.

### DIFF
--- a/tidy3d/components/data/dataset.py
+++ b/tidy3d/components/data/dataset.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, Union
+from typing import Any, Callable, Dict, Optional, Union
 
 import numpy as np
 import pydantic.v1 as pd
@@ -149,32 +149,32 @@ EMScalarFieldType = Union[
 class ElectromagneticFieldDataset(AbstractFieldDataset, ABC):
     """Stores a collection of E and H fields with x, y, z components."""
 
-    Ex: EMScalarFieldType = pd.Field(
+    Ex: Optional[EMScalarFieldType] = pd.Field(
         None,
         title="Ex",
         description="Spatial distribution of the x-component of the electric field.",
     )
-    Ey: EMScalarFieldType = pd.Field(
+    Ey: Optional[EMScalarFieldType] = pd.Field(
         None,
         title="Ey",
         description="Spatial distribution of the y-component of the electric field.",
     )
-    Ez: EMScalarFieldType = pd.Field(
+    Ez: Optional[EMScalarFieldType] = pd.Field(
         None,
         title="Ez",
         description="Spatial distribution of the z-component of the electric field.",
     )
-    Hx: EMScalarFieldType = pd.Field(
+    Hx: Optional[EMScalarFieldType] = pd.Field(
         None,
         title="Hx",
         description="Spatial distribution of the x-component of the magnetic field.",
     )
-    Hy: EMScalarFieldType = pd.Field(
+    Hy: Optional[EMScalarFieldType] = pd.Field(
         None,
         title="Hy",
         description="Spatial distribution of the y-component of the magnetic field.",
     )
-    Hz: EMScalarFieldType = pd.Field(
+    Hz: Optional[EMScalarFieldType] = pd.Field(
         None,
         title="Hz",
         description="Spatial distribution of the z-component of the magnetic field.",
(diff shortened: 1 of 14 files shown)


yaugenst-flex (approved):
Thanks @caseyflex this looks really solid so code-wise I don't have any comments, more of a general question - is this something that we expect users to use or is it mostly for internal purposes? Also, what would happen if we were to add support for more auxiliary fields in the future? I wonder if these fields could just be stored in the regular `FieldTimeData`?

caseyflex:
> Thanks @caseyflex this looks really solid so code-wise I don't have any comments, more of a general question - is this something that we expect users to use or is it mostly for internal purposes? Also, what would happen if we were to add support for more auxiliary fields in the future? I wonder if these fields could just be stored in the regular `FieldTimeData`?

Yeah, I think users will use it. Already knowing the free-carrier density when using TwoPhotonAbsorption nonlinearity is quite useful.

Regarding future aux fields -- I was thinking that we could just add more fields to `AuxFieldTimeData` every time we need to add a new field. Would that cause backwards compatibility issues? This is related to the last point of just using `FieldTimeData` -- in principle we could use that here too, but we would need to add fields to it which would clutter that dataset (currently it has Ex, Ey, Ez, Hx, Hy, Hz and nothing else). Also electromagnetic field data is treated specially in other places; you can compute poynting vectors, overlaps, and other things that don't make sense for aux fields.

tidy3d/components/data/dataset.py
```diff
@@ -312,6 +312,88 @@ def apply_phase(self, phase: float) -> AbstractFieldDataset:
         return self
 
 
+class AuxFieldDataset(AbstractFieldDataset, ABC):
+    """Stores a collection of aux fields with x, y, z components."""
+
+    Nfx: EMScalarFieldType = pd.Field(
+        None,
```
yaugenst-flex: Actually just noticed, slight nitpick: This should be `typing.Optional[EMScalarFieldType]`. Pydantic is flexible enough to accept the implicit optional here but a type checker generally won't. Similar everywhere else. I know we do it this way in a few places but since this is new code I think it'd be better to be more explicit. If you have the extra two minutes might as well update the `FieldDataset` and `FieldTimeDataset` types aswell...

tidy3d/components/monitor.py
```diff
@@ -270,6 +271,46 @@ def _storage_size_solver(self, num_cells: int, tmesh: ArrayFloat1D) -> int:
         return solver_data_size
 
 
+class AbstractAuxFieldMonitor(Monitor, ABC):
+    """:class:`Monitor` that records electromagnetic field data as a function of x,y,z."""
+
+    fields: Tuple[AuxField, ...] = pydantic.Field(
+        [],
```
yaugenst-flex: This here should be `()`, otherwise it's a mutable default.

yaugenst-flex:
> Regarding future aux fields -- I was thinking that we could just add more fields to AuxFieldTimeData every time we need to add a new field. Would that cause backwards compatibility issues?

No I think that'd be fine as long as all fields are optional.

> Also electromagnetic field data is treated specially in other places; you can compute poynting vectors, overlaps, and other things that don't make sense for aux fields.

Yeah I agree. Let's keep it like this for now.

tidy3d/components/data/dataset.py
caseyflex: thanks! done

tidy3d/components/monitor.py
caseyflex: ah nice catch

tidy3d/components/monitor.py
```diff
@@ -270,6 +271,46 @@ def _storage_size_solver(self, num_cells: int, tmesh: ArrayFloat1D) -> int:
...
 
 
+class AbstractAuxFieldMonitor(Monitor, ABC):
+    """:class:`Monitor` that records electromagnetic field data as a function of x,y,z."""
+
+    fields: Tuple[AuxField, ...] = pydantic.Field(
+        (),
+        title="Aux Field Components",
+        description="Collection of auxiliary field components to store in the monitor.",
+    )
```
momchil-flex, line 286: How can the user know what's allowed to enter here?

I mean, I kind of understand that the check can only happen on the Simulation level. So my question is maybe mostly documentation wise. If I'm trying to understand what to put here, I need to discover two things: the possible values and what they mean, for which I need to look at the AuxFieldTimeMonitorData, and what is allowed in my simulation, for which I need to look at the `Medium.aux_fields` in the mediums I'm using.

How do we assemble all that in coherent instructions to the user?

caseyflex: Hmm, yeah. Currently, when building the monitor, you can at least see the list of all possible aux fields (including those used in nonlinearities not present in the simulation). Then when building the simulation, it gives a warning if some of these aux fields are not actually present in the simulation -- in which case the monitor just records them as zero.

I couldn't really think of anything cleaner -- the Monitor can't really know about the Simulation. I can make `Simulation.aux_fields` public (it lists all the aux fields actually present in the simulation, compiled from all the Medums).

caseyflex: I guess I can also improve this docstring to state this explicitly

momchil-flex: Yeah maybe explicitly reference the data components where the various docstrings for the aux fields live?

tidy3d/components/data/dataset.py
```diff
@@ -312,6 +312,88 @@ def apply_phase(self, phase: float) -> AbstractFieldDataset:
...
+
+    @property
+    def symmetry_eigenvalues(self) -> Dict[str, Callable[[Axis], float]]:
+        """Maps field components to their (positive) symmetry eigenvalues."""
+
+        return dict(
+            Nfx=lambda dim: +1,
+            Nfy=lambda dim: +1,
+            Nfz=lambda dim: +1,
+        )
```
momchil-flex, line 360: I took another look at this PR to see if the symmetry expansion, and grid locations would be correctly handled and I'm pleasantly surprised that this seems to have been set up in a pretty expandable way previously! :D I was worried about it.

PR #2311, opened 2025-03-13
fix: automatically create batch output directories

### LINKED ISSUE #2305: `batch.run` not creating new folders in 2.8
**Describe the bug**

When calling the `batch.run(path_dir="newDir")` in 2.8, if 'newDir' doesn't exist, it won't be created and will cause an error. This behavior is different from 2.7.8, where the folder would be created.

**To Reproduce**

Just run the [bandstructure](https://www.flexcompute.com/tidy3d/examples/notebooks/Bandstructure/) example with  `batch.run(path_dir="newDir")`

**Expected behavior**
The folder should be created

yaugenst-flex:
Fixes #2305 

Since `os.path.dirname` always returns the _parent_ directory, the directory creation works fine for paths like `a/b/c.txt` (where it returns `a/b`), but if we specify an output _folder_ as we do in `Batch`, then only folders up to the parent folder will get created. To fix this, I made a separate `_check_path_dir` implementation for `Batch` and `Job` and added a test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Bug in `LayerRefinementSpec` that refines grids outside the layer region when one in-plane dimension is of size infinity.
 - Querying tasks was sometimes erroring unexpectedly.
+- Fixed automatic creation of missing output directories.
 
 ## [2.8.0] - 2025-03-04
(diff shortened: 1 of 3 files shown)


FilipeFcp (approved):
Great. Thank you @yaugenst-flex!

momchil-flex (approved):

yaugenst-flex:
@dmarek-flex this is mostly for your reference

PR #2312, opened 2025-03-13
Convenience function to add charge perturbations with Nedeljkovic Soref Mashanovich model

marc-flex:
This PR adds a convenience class that allows to easily specify medium perturbation using the model in [here](https://ieeexplore.ieee.org/document/6051462). It addresses issue https://github.com/flexcompute/tidy3d-core/issues/846

For instance, in the Charge notebook one can now define the perturbation medium like this
```python
perturbation_model = td.NedeljkovicSorefMashanovich(ref_freq=freq0)

si_perturb = td.PerturbationMedium.from_unperturbed(
    medium=si_non_perturb,
    perturbation_spec=td.IndexPerturbation(
        delta_n=perturbation_model.delta_n(),
        delta_k=perturbation_model.delta_k(),
        freq=freq0,
    )
)
```
The class `NedeljkovicSorefMashanovich` derives from a new class `AbstractDeltaModel` so potentially more perturbation models could be implemented following this approach (or used as template).

### DIFF
--- a/tidy3d/components/parameter_perturbation.py
+++ b/tidy3d/components/parameter_perturbation.py
@@ -12,11 +12,12 @@ except ImportError:
     pass
 import numpy as np
 import pydantic.v1 as pd
+import xarray as xr
 
 from ..components.data.validators import validate_no_nans
 from ..components.types import TYPE_TAG_STR, ArrayLike, Ax, Complex, FieldVal, InterpMethod
 from ..components.viz import add_ax_if_none
-from ..constants import CMCUBE, EPSILON_0, HERTZ, KELVIN, PERCMCUBE, inf
+from ..constants import C_0, CMCUBE, EPSILON_0, HERTZ, KELVIN, PERCMCUBE, inf
 from ..exceptions import DataError
 from ..log import log
 from .base import Tidy3dBaseModel, cached_property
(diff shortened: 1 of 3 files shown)


tidy3d/components/parameter_perturbation.py
```diff
@@ -1227,6 +1228,389 @@ def _sample_delta_eps_delta_sigma(
...
+
+    M. Nedeljkovic, R. Soref and G. Z. Mashanovich, "Free-Carrier Electrorefraction and Electroabsorption
+    Modulation Predictions for Silicon Over the 1–14- μm Infrared Wavelength Range," in IEEE Photonics
+    Journal, vol. 3, no. 6, pp. 1171-1180, Dec. 2011, doi: 10.1109/JPHOT.2011.2171930
+
+    Example
+    -------
+    """
+
+    perturb_coeffs = xr.Dataset(
```
dbochkov-flexcompute, line 1255: I wonder if we are better off with not having `perturb_coeffs` field in this class and store outside of class as, say, `NSM_DATA`

tidy3d/components/parameter_perturbation.py
```diff
@@ -1227,6 +1228,389 @@ def _sample_delta_eps_delta_sigma(
...
+    )
+
+    @pd.root_validator(skip_on_failure=True)
+    def _check_freq_in_range(cls, values):
+        """Check that the given frequency is within validity range.
+        If not, issue a warning.
+        """
+
+        freq = values.get("ref_freq")
+        freq_range = (C_0 / 14, C_0 / 1.3)
```
dbochkov-flexcompute: probably better to use `np.min(coords.wvl)` and `np.max(coords.wvl)` so that's it's clear where they are coming from

tidy3d/components/parameter_perturbation.py
```diff
@@ -1227,6 +1228,389 @@ def _sample_delta_eps_delta_sigma(
...
+        return self.perturb_coeffs.interp(wvl=self.ref_wavelength)
+
+    def delta_k(self) -> ChargePerturbationType:
+        """Return the perturbation range of the model."""
+
+        # create free-carrier mesh
+        # NOTE: this range (0, 1e22) should work in most cases. If need be we can add a new field
+        # to the model to allow for custom ranges.
+        Ne_range = np.concatenate(([0], np.logspace(-6, 22, num=200)))
+        Nh_range = np.concatenate(([0], np.logspace(-6, 22, num=200)))
```
dbochkov-flexcompute: yeah, I would say lets go ahead and add two new fields in this class `electron_concentration_grid = np.logspace(-6, 22, num=200)` and `hole_concentration_grid = np.logspace(-6, 22, num=200)`

tidy3d/components/parameter_perturbation.py
```diff
@@ -1380,3 +1764,7 @@ def _sample_delta_eps_delta_sigma(
             delta_sigma = delta_sigma * EPSILON_0
 
         return delta_eps, delta_sigma
+
+    def from_perturbation_delta_model(cls, deltas_model: AbstractDeltaModel) -> IndexPerturbation:
+        """Create an IndexPerturbation from a DeltaPerturbationModel."""
+        pass
```
dbochkov-flexcompute: does this do anything?

dbochkov-flexcompute (commented):
not sure what would be the best organization of this kind of model, given it so specific to a single material. I guess, one alternative I see is to simply have something like `IndexPerturbation.from_nsm_si_model(f: float)` class method instead of a new class. One concern, neither that nor the current implementation easily allow to include heat perturbations too

tidy3d/components/parameter_perturbation.py
marc-flex: Apologies. I left this one for the end and I forgot to implement it.

tidy3d/components/parameter_perturbation.py
marc-flex: Why do you want to have this outside the class? I created the class precisely for this :laughing: 

I guess the alternative is to not have a class at all. But then all this specific stuff needs to be put somewhere else

marc-flex:
I agree this model is very specific. That's why I made it derive from a more generic base class. This base class, in principle would allow to implement other models for other use cases. It would also allow users to create their own and use them with 
```python
def from_perturbation_delta_model(cls, deltas_model: AbstractDeltaModel) -> IndexPerturbation:
    """Create an IndexPerturbation from a DeltaPerturbationModel."""
   return IndexPerturbation(delta_n=deltas_model.delta_n, delta_k=deltas_model.delta_k)
```

Precisely because of it being so specific I'm not super keen on having something like `IndexPerturbation.from_nsm_si_model(f: float)`. I'd prefer to have the specific functions somewhere else.

momchil-flex (approved):
I think I agree with @marc-flex here, @dbochkov-flexcompute if you find a quick moment to give final thoughts if we can merge this as is?

dbochkov-flexcompute:
yeah, sounds good!

dbochkov-flexcompute (approved):

PR #2315, opened 2025-03-14
Add component selection to plot_eps

### LINKED ISSUE #2085: Add component selection to plot_eps
# Current
When we call `plot_eps` and the simulation has anisotropic material models, the `eps` limits are calculated using `eps_model` which in turn, takes the average between the `eps` components on the main diagonal. So, it is not possible to visualize a specific permittivity component.

For example, we have this anisotropic material model with a main diagonal of `[15, 10, 5]`.
image
 
When we plot it, the maximum permittivity is 10. 
image

# Requirement
Considering we have introduced some anisotropic materials in our material library and we have some users working with LiNbO3, it would be interesting to allow the users to specify which `eps` component they want to plot. We could keep the average of the mais `eps` diagonal as the default for backward compatibility. Maybe something like `plot_eps(component="xx")` and the user can select between `xx`, `yy`, `zz`, or `None`. If the medium is fully anisotropic, maybe we could allow for `xx`, `xy`, `xz`, `yx`, `yy`, `yz`, `zx`, `zy`, `zz`, or `None`.

bzhangflex:
Closes https://github.com/flexcompute/tidy3d/issues/2085

Adds "component" argument to simulation.plot_eps(), where component = "xx", "yy", "zz", "xy", "yz", "xz". Affects plotting of AnisotropicMedium, FullyAnisotropicMedium and CustomAnisotropicMedium.

my first PR :')

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `fill` and `fill_structures` argument in `td.Simulation.plot_structures()` and `td.Simulation.plot()` respectively to disable fill and plot outlines of structures only.
 - New subpixel averaging option `ContourPathAveraging` applied to dielectric material boundaries.
 - A property `interior_angle` in `PolySlab` that stores angles formed inside polygon by two adjacent edges.
+- `eps_component` argument in `td.Simulation.plot_eps()` to optionally select a specific permittivity component to plot (eg. `"xx"`).
 
 ### Fixed
 - Compatibility with `xarray>=2025.03`.
--- a/tests/test_components/test_simulation.py
+++ b/tests/test_components/test_simulation.py
@@ -1,11 +1,15 @@
 """Tests the simulation and its validators."""
 
+import shutil
+from pathlib import Path
+
 import gdstk
 import matplotlib.pyplot as plt
 import numpy as np
 import pydantic.v1 as pydantic
 import pytest
 import tidy3d as td
+from matplotlib.testing.decorators import check_figures_equal
 from tidy3d.components import simulation
 from tidy3d.components.scene import MAX_GEOMETRY_COUNT, MAX_NUM_MEDIUMS
 from tidy3d.components.simulation import MAX_NUM_SOURCES
@@ -691,6 +695,208 @@ def test_plot_eps_bounds():
     plt.close()
     _ = SIM_FULL.plot_eps(x=0, hlim=[-0.45, 0.45], vlim=[-0.45, 0.45])
     plt.close()
+
+
+class TestAnisotropicPlotting:
+    """Tests for plotting anisotropic media"""
+
+    diag_comps = ["xx", "yy", "zz"]
+    offdiag_comps = ["xy", "yx", "xz", "zx", "yz", "zy"]
+    allcomps = diag_comps + offdiag_comps
+
+    medium_diag = td.AnisotropicMedium(
+        xx=td.Medium(permittivity=5), yy=td.Medium(permittivity=10), zz=td.Medium(permittivity=15)
+    )
+
+    medium_fullyani = td.FullyAnisotropicMedium(permittivity=[[6, 2, 3], [2, 7, 4], [3, 4, 8]])
+
+    @pytest.fixture(scope="class")
+    def medium_customani(self):
+        """based this custom medium on
+        https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.CustomAnisotropicMedium.html
+        """
+        Nx, Ny, Nz = 100, 100, 100
+        x = np.linspace(-1, 1, Nx)
+        y = np.linspace(-1, 1, Ny)
+        z = np.linspace(-1, 1, Nz)
+        coords = dict(x=x, y=y, z=z)
+        permittivity = td.SpatialDataArray(2 * np.ones((Nx, Ny, Nz)), coords=coords)
+        conductivity = td.SpatialDataArray(np.ones((Nx, Ny, Nz)), coords=coords)
+        medium_xx = td.CustomMedium(permittivity=permittivity, conductivity=conductivity)
+        medium_yy = td.CustomMedium(permittivity=2 * permittivity, conductivity=conductivity)
+
+        # make the zz component a spatially varying medium
+        # define coordinate array
+        x_mesh, y_mesh, _ = np.meshgrid(x, y, z, indexing="ij")
+        r_mesh = np.sqrt(x_mesh**2 + y_mesh**2)  # radial distance
+
+        # index of refraction array
+        # assign the refractive index value to the array according to the desired profile
+        n_data = np.ones((Nx, Ny, Nz))
+        n0 = 2
+        A = 0.5
+        r = 1
+        n_data[r_mesh <= r] = n0 * (1 - A * r_mesh[r_mesh <= r] ** 2)
+        # convert to dataset array
+        n_dataset = td.SpatialDataArray(n_data, coords=dict(x=x, y=y, z=z))
+        medium_zz = td.CustomMedium.from_nk(n_dataset, interp_method="nearest")
+
+        return td.CustomAnisotropicMedium(xx=medium_xx, yy=medium_yy, zz=medium_zz)
+
+    @pytest.fixture(scope="module", autouse=True)
+    def cleanup_figures(self):
+        yield  # Run tests first
+        fig_dir = Path().resolve() / "result_images"
+        shutil.rmtree(fig_dir, ignore_errors=False)
+
+    def make_sim(self, medium):
+        L = 5
+
+        source = td.UniformCurrentSource(
+            center=(0, 0, -L / 3),
+            size=(L, L / 2, 0),
+            polarization="Ex",
+            source_time=td.GaussianPulse(
+                freq0=td.C_0,
+                fwidth=10e14,
+            ),
+        )
+        structures = (td.Structure(geometry=td.Sphere(center=(0, 0, 0), radius=1), medium=medium),)
+
+        return td.Simulation(
+            size=(L, L, L),
+            grid_spec=td.GridSpec.uniform(dl=0.01),
+            structures=structures,
+            sources=[source],
+            run_time=1e-12,
+        )
+
+    @pytest.mark.parametrize("eps_comp", ("xyz", "123", "", 5))
+    def test_bad_eps_arg(self, eps_comp):
+        """Tests that an incorrect component raises the proper exception."""
+        with pytest.raises(ValueError, match=f"eps_component '{eps_comp}' is not supported. "):
+            self.make_sim(self.medium_diag).plot_eps(x=0, eps_component=eps_comp)
+
+    @pytest.mark.parametrize(
+        "eps_comp",
+        [
+            None,
+        ]
+        + diag_comps,
+    )
+    def test_plot_anisotropic_medium(self, eps_comp):
+        """Test plotting diagonal components of a diagonally anisotropic medium succeeds or not.
+        diagonal components and ``None`` should succeed.
+        """
+        self.make_sim(self.medium_diag).plot_eps(x=0, eps_component=eps_comp)
+
+    @pytest.mark.parametrize("eps_comp", offdiag_comps)
+    def test_plot_anisotropic_medium_offdiagfail(self, eps_comp):
+        """Tests that plotting off-diagonal components of a diagonally anisotropic medium raises an exception."""
+        with pytest.raises(
+            ValueError,
+            match=f"Plotting component '{eps_comp}' of a diagonally-anisotropic permittivity tensor is not supported",
+        ):
+            self.make_sim(self.medium_diag).plot_eps(x=0, eps_component=eps_comp)
+
+    @pytest.mark.parametrize(
+        "eps_comp1,eps_comp2",
+        (
+            pytest.param("xx", "yy", marks=pytest.mark.xfail),
+            pytest.param("xx", "zz", marks=pytest.mark.xfail),
+            pytest.param("yy", "zz", marks=pytest.mark.xfail),
+        ),
+    )
+    @check_figures_equal(extensions=("png",))
+    def test_plot_anisotropic_medium_diff(self, fig_test, fig_ref, eps_comp1, eps_comp2):
+        """Tests that the plots of different components of an AnisotropicMedium are actually different."""
+        sim = self.make_sim(self.medium_diag)
+
+        ax1 = fig_test.add_subplot()
+        sim.plot_eps(x=0, eps_component=eps_comp1, ax=ax1)
+        ax2 = fig_ref.add_subplot()
+        sim.plot_eps(x=0, eps_component=eps_comp2, ax=ax2)
+
+    @pytest.mark.parametrize(
+        "eps_comp",
+        [
+            None,
+        ]
+        + diag_comps
+        + offdiag_comps,
+    )
+    def test_plot_fully_anisotropic_medium(self, eps_comp):
+        """Test plotting all components of a fully anisotropic medium.
+        All plots should succeed.
+        """
+        sim = self.make_sim(self.medium_fullyani)
+        sim.plot_eps(x=0, eps_component=eps_comp)
+
+    # Test parameters for comparing plots of a FullyAnisotropicMedium
+    fullyani_testplot_diff_params = []
+    for eps_comp1 in allcomps:
+        for eps_comp2 in allcomps:
+            if eps_comp1 == eps_comp2 or eps_comp1[::-1] == eps_comp2:
+                # Same components, or transposed components (eg. xy and yx) should plot the same
+                fullyani_testplot_diff_params.append((eps_comp1, eps_comp2))
+            else:
+                # All other component pairs should plot differently
+                fullyani_testplot_diff_params.append(
+                    pytest.param(eps_comp1, eps_comp2, marks=pytest.mark.xfail)
+                )
+
+    @pytest.mark.parametrize("eps_comp1,eps_comp2", fullyani_testplot_diff_params)
+    @check_figures_equal(extensions=("png",))
+    def test_plot_fully_anisotropic_medium_diff(self, fig_test, fig_ref, eps_comp1, eps_comp2):
+        """Tests that the plots of different components of a FullyAnisotropicMedium are actually different."""
+        sim = self.make_sim(self.medium_fullyani)
+
+        ax1 = fig_test.add_subplot()
+        sim.plot_eps(x=0, eps_component=eps_comp1, ax=ax1)
+        ax2 = fig_ref.add_subplot()
+        sim.plot_eps(x=0, eps_component=eps_comp2, ax=ax2)
+
+    @pytest.mark.parametrize(
+        "eps_comp",
+        [
+            None,
+        ]
+        + diag_comps,
+    )
+    def test_plot_customanisotropic_medium(self, eps_comp, medium_customani):
+        """Test plotting diagonal components of a diagonally anisotropic custom medium.
+        diagonal components and ``None`` should succeed.
+        """
+        self.make_sim(medium_customani).plot_eps(x=0, eps_component=eps_comp)
+
+    @pytest.mark.parametrize("eps_comp", offdiag_comps)
+    def test_plot_customanisotropic_medium_offdiagfail(self, eps_comp, medium_customani):
+        """Tests that plotting off-diagonal components of a diagonally anisotropic custom medium raises an exception."""
+        with pytest.raises(
+            ValueError,
+            match=f"Plotting component '{eps_comp}' of a diagonally-anisotropic permittivity tensor is not supported.",
+        ):
+            self.make_sim(medium_customani).plot_eps(x=0, eps_component=eps_comp)
+
+    @pytest.mark.parametrize(
+        "eps_comp1,eps_comp2",
+        (
+            pytest.param("xx", "yy", marks=pytest.mark.xfail),
+            pytest.param("xx", "zz", marks=pytest.mark.xfail),
+            pytest.param("yy", "zz", marks=pytest.mark.xfail),
+        ),
+    )
+    @check_figures_equal(extensions=("png",))
+    def test_plot_customanisotropic_medium_diff(
+        self, fig_test, fig_ref, eps_comp1, eps_comp2, medium_customani
+    ):
+        """Tests that the plots of different components of an AnisotropicMedium are actually different."""
+        sim = self.make_sim(medium_customani)
+
+        ax1 = fig_test.add_subplot()
+        sim.plot_eps(x=0, eps_component=eps_comp1, ax=ax1)
+        ax2 = fig_ref.add_subplot()
+        sim.plot_eps(x=0, eps_component=eps_comp2, ax=ax2)
 
 
 def test_plot():
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -398,6 +398,7 @@ start_node = VJPNode.new_root()
 tracer = new_box(1.0, 0, start_node)
 tracer_arr = new_box(np.array([[[1.0]]]), 0, start_node)
 
+
 SIM_FULL = td.Simulation(
     size=(8.0, 8.0, 8.0),
     run_time=1e-12,
@@ -469,6 +470,12 @@ SIM_FULL = td.Simulation(
         td.Structure(
             geometry=td.Box(size=(1, 1, 1), center=(-1, 0, 0)),
             medium=td.AnisotropicMedium(xx=td.PEC, yy=td.Medium(), zz=td.Medium()),
+        ),
+        # Test a fully anistropic medium
+        td.Structure(
+            geometry=td.Box(size=(1, 1, 1), center=(-1, 0, 0)),
+            medium=td.FullyAnisotropicMedium(permittivity=[[6, 2, 3], [2, 7, 4], [3, 4, 9]]),
+            name="fully_anisotropic_box",
         ),
         td.Structure(
             geometry=td.GeometryGroup(geometries=[td.Box(size=(1, 1, 1), center=(-1, 0, 0))]),
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -84,6 +84,7 @@ from .types import (
     FreqBound,
     InterpMethod,
     Literal,
+    PermittivityComponent,
     PoleAndResidue,
     TensorReal,
 )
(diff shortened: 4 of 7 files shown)


e-g-melo:
Thanks @bzhangflex! This feature will be handy to help users working with anisotropic materials.

I created an `AnisotropicMedium` like below.

```py
permittivity_main = (15, 10, 5)
conductivity_main = (0.01, 0.02, 0.03)

medium_diag = td.AnisotropicMedium(
    xx=td.Medium(permittivity=permittivity_main[0], conductivity=conductivity_main[0]),
    yy=td.Medium(permittivity=permittivity_main[1], conductivity=conductivity_main[1]),
    zz=td.Medium(permittivity=permittivity_main[2], conductivity=conductivity_main[2]),        
)
```

And it works very well for component = "xx", "yy", or "zz". However, when I tried component = "xy", I got the following plot. It seems quite strange. Maybe can confuse users.



I wonder if we could, for example, ignore `component`, then plot the average of the main diagonal permittivity and show a warning if the user tries to plot an off-diagonal element from `AnisotropicMedium` or `CustomAnisotropicMedium`. If we go in this direction, could we handle that in simulations using `AnisotropicMedium` and `FullyAnisotropicMedium` simultaneously? not likely to happen, though :)

tidy3d/components/scene.py
```diff
@@ -769,6 +772,10 @@ def plot_structures_eps(
             The x range if plotting on xy or xz planes, y range if plotting on yz plane.
         vlim : Tuple[float, float] = None
             The z range if plotting on xz or yz planes, y plane if plotting on xy plane.
+        component : str = None
+            Material polarization component to plot for anisotropic materials,
+            e.g. "xx", "yy", "zz", "xy", "yz", ...
```
tylerflex: what happens if I pass something like "foo"? We should raise a `ValueError` with a useful message.

tidy3d/components/scene.py
```diff
@@ -984,15 +999,37 @@ def _add_cbar_eps(eps_min: float, eps_max: float, ax: Ax = None) -> None:
...
+    ) -> Tuple[float, float]:
         """Compute range of (real) permittivity present in the mediums at frequency "freq"."""
         medium_list = [medium for medium in medium_list if not medium.is_pec]
-        # regular medium
-        eps_list = [
-            np.real(medium.eps_model(freq))
-            for medium in medium_list
-            if not isinstance(medium, AbstractCustomMedium) and not isinstance(medium, Medium2D)
-        ]
+        if component is None:
```
tylerflex: this if/else block to me signifies that we should consider refactoring a bit.

For example, I think we need some method `Medium._eps_comp(freq, component=None)` that would handle this logic of component selection internally for anisotropic mediums. (not sure about best name for this, could also put everything in existing `eps_comp`).

Note also it could return `None` for `AbstractCustomMedium` and `Medium2D`. 

Potentially then, the eps_list creation could be as simple as

```py
eps_comp_list = [medium._eps_comp(freq=freq, component=component) for medium in medium_list]
eps_list = [np.real(m) for m eps_comp_list if m is not None]
```

tidy3d/components/scene.py
```diff
@@ -769,6 +772,10 @@ def plot_structures_eps(
             The x range if plotting on xy or xz planes, y range if plotting on yz plane.
         vlim : Tuple[float, float] = None
             The z range if plotting on xz or yz planes, y plane if plotting on xy plane.
+        component : str = None
```
tylerflex: should we consider naming this something more specific? like `pol_component` to indicate that it's polarization? could go either way, but worth discussing.

tidy3d/components/scene.py
```diff
@@ -984,15 +999,37 @@ def _add_cbar_eps(eps_min: float, eps_max: float, ax: Ax = None) -> None:
...
+                if not isinstance(
+                    medium,
+                    (AbstractCustomMedium, Medium2D, AnisotropicMedium, FullyAnisotropicMedium),
+                )
+            ]
+            # anisotropic medium with user selected component
+            comp2indx = {"x": 0, "y": 1, "z": 2}
+            eps_list += [
+                medium.eps_comp(
+                    row=comp2indx[component[0]], col=comp2indx[component[1]], frequency=freq
```
tylerflex: assumes too much of `component`. Would give a hard to understand KeyError if eg. `component="fg"`. we should check right away I think to ensure that `len(component) == 2 and all(c in 'xyz' for c in component)` if `component is not None`

tidy3d/components/scene.py
tylerflex: could you put the string in double backslash quotes, eg ``"xx"``? It will render better for the docs. see https://www.notion.so/flexcompute/Python-Style-Guide-afac5dbbf949463e89c18fe39b7453c9?pvs=4#72d1e0acdd8b42b6872ba944a4599efd

note, check this for all other related docstrings too.

tidy3d/components/scene.py
```diff
@@ -1083,18 +1125,25 @@ def _pcolormesh_shape_custom_medium_structure_eps(
...
-
-                    if isinstance(eps_mean, TetrahedralGridDataset):
+                    if component == "xx":
+                        eps = eps_diag[0]
+                    elif component == "yy":
+                        eps = eps_diag[1]
+                    elif component == "zz":
+                        eps = eps_diag[2]
+                    else:
+                        eps = (eps_diag[0] + eps_diag[1] + eps_diag[2]) / 3
```
tylerflex, line 1128: hm, I would think now that we have component, we should only do this averaging if `component is None` for backwards compatibility.

I'd actually say maybe `if component is not None` and it's not one of the ii components, we should consider either erroring or just biting the bullet and doing the averaging between eg. [0] and [1] for "xy"?

```py
if component is None:
    eps = np.mean(eps_diag) # I think this should work?
else:
    eps_indices = ['xyz'.index(c) for c in component]
    eps_values = [eps_diag[i] for i in eps_indices]
    eps = np.mean(eps_values)
```

tidy3d/components/scene.py
```diff
@@ -1160,8 +1209,18 @@ def _pcolormesh_shape_custom_medium_structure_eps(
             normal_axis: [normal_position],
         }
         coord_shape = Coords(**coord_dict)
-        # interpolate permittivity and take the average over components
-        eps_shape = np.mean(medium.eps_diagonal_on_grid(frequency=freq, coords=coord_shape), axis=0)
+
+        # interpolate permittivity and pick the component to plot
+        eps_shape = medium.eps_diagonal_on_grid(frequency=freq, coords=coord_shape)
+        if component == "xx":
```
tylerflex: similar comment to above. should we just generalize tis to do the the `mean` using the off diagonal components too? or in this case will it create issues if the shape is not the same between eg. x and y permittivity grids?

tidy3d/components/scene.py
```diff
@@ -1209,8 +1269,15 @@ def _get_structure_eps_plot_params(
             # 2d material
             plot_params = plot_params.copy(update={"edgecolor": "k", "linewidth": 1})
         else:
-            # regular medium
-            eps_medium = medium.eps_model(frequency=freq).real
+            if component and isinstance(medium, (AnisotropicMedium, FullyAnisotropicMedium)):
```
tylerflex: this seems like duplicated code from above. we should refactor to put it somewhere it can be re-used, potentially we can just put it into `eps_comp` as I mentioned in the earlier comment.

tylerflex (changes requested):
Thanks @bzhangflex ! Implementation looks good, I had a few comments to hopefully simplify the code a bit and handle some edge cases.

tidy3d/components/scene.py
bzhangflex: I like the idea of renaming it `pol_component`

tidy3d/components/scene.py
bzhangflex: good call, will add that in

tidy3d/components/scene.py
bzhangflex: That's a great idea, agreed that would clean up the logic

tidy3d/components/scene.py
bzhangflex: Should we error then? Since technically the off-diagonal permittivity is zero, so it doesn't really make sense to plot

bzhangflex:
> I wonder if we could, for example, ignore `component`, then plot the average of the main diagonal permittivity and show a warning if the user tries to plot an off-diagonal element from `AnisotropicMedium` or `CustomAnisotropicMedium`. If we go in this direction, could we handle that in simulations using `AnisotropicMedium` and `FullyAnisotropicMedium` simultaneously? not likely to happen, though :)

Thanks @e-g-melo! @tylerflex suggested either erroring or plotting the average of the two diagonal components (eg "xy" would average "xx" and "yy"). I'm leaning more towards erroring, since in those cases the off-diagonal elements should all be zero so it would be strange to plot them.

tidy3d/components/scene.py
tylerflex: btw, feel free to mark this (and others) as "Resolve conversation"  once the code is in.

tidy3d/components/scene.py
tylerflex: @yaugenst-flex thoughts?

tidy3d/components/scene.py
tylerflex: ah, so I'm a little confused. this is for plotting custom medium components, which have to be diagonally isotropic? in that case yea I think it would be reasonable to just error, good point

bzhangflex: Yea, currently only diagonal permittivity is implemented for custom mediums

tidy3d/components/scene.py
yaugenst-flex: hmm I think `pol_component` is not really appropriate. I mean it's not fundamentally wrong but I think it conflates field polarization (a response) with tensor entries (a material property). it can be interpreted as specifying an input field polarization instead of which element of the permittivity tensor to plot, which gets particularly confusing when you have cross-coupling between polarizations. i'd say either `eps_component` or just `component` would make sense. since the function is already named `plot_eps` or `plot_structures_eps`, I think just `component` is appropriate.

on that note, I would also update the docstring to read `Component of the permittivity tensor` or just `Permittivity component` instead of `Material polarization component`. I think material polarization is a misnomer here because the permittivity is not an inherent material polarization but it describes how a material becomes polarized under an applied field.

tests/test_components/test_simulation.py
yaugenst-flex, line 1: Should be testing all of `AnisotropicMedium`, `FullyAnisotropicMedium` and `CustomAnisotropicMedium` here, especially since `CustomAnisotropicMedium` is supposed to error for off-diagonal elements? Try to catch as many edge cases as possible in tests. Also, this test is testing multiple things (plotting of `SIM_FULL` and a custom sim?), and for each permittivity component. Ideally each test should cover one specific piece of functionality, e.g., the permittivity components could be [parametrized](https://docs.pytest.org/en/stable/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions), and we could make a separate test for custom medium off-diagonal elements. Also don't be afraid of long test function names. Something like:
 - test_plot_fully_anisotropic_medium
 - test_plot_anisotropic_medium
 - test_plot_anisotropic_medium_offdiagonal_error
 - test_plot_custom_anisotropic
 - test_plot_custom_anisotropic_offdiagonal_error

If many of these tests end up sharing the same code, you can group them into a [test class](https://docs.pytest.org/en/stable/getting-started.html#group-multiple-tests-in-a-class).

More generally I guess it's a bit tricky for plotting to test much more than "does this error/not error?", but there are some testing utilities [in matplotlib](https://matplotlib.org/stable/api/testing_api.html#module-matplotlib.testing.compare) which we could use to e.g. verify that plotting different permittivity components actually yields different plots.

tidy3d/components/scene.py
```diff
@@ -738,6 +740,7 @@ def plot_structures_eps(
         hlim: Tuple[float, float] = None,
         vlim: Tuple[float, float] = None,
         grid: Grid = None,
+        component: str = None,
```
yaugenst-flex: This shouldn't just be `str` but instead should be a `typing.Literal` that enumerates all choices. We can introduce a new type for that, e.g. `PermittivityComponent = Literal["xx", "yy", "zz", ...]`.
It would also make the documentation less ambiguous if the function signature is `component: Optional[PermittivityComponent] = None`.

yaugenst-flex: On that note, all these types that are `None` by default should be `Optional`...

tidy3d/components/scene.py
```diff
@@ -769,6 +772,10 @@ def plot_structures_eps(
             The x range if plotting on xy or xz planes, y range if plotting on yz plane.
         vlim : Tuple[float, float] = None
             The z range if plotting on xz or yz planes, y plane if plotting on xy plane.
+        component : str = None
+            Material polarization component to plot for anisotropic materials,
+            e.g. "xx", "yy", "zz", "xy", "yz", ...
+            Defaults to None, which returns the average of the diagonal.
```
yaugenst-flex: Same comment as Tyler's above, `None` should be ` ``None`` `

yaugenst-flex (commented):
Thanks @bzhangflex! Some comments and probably I'll have some more after the you've addressed @tylerflex's comments, but let's address those first.

tidy3d/components/scene.py
bzhangflex: Ended up adding a `_eps_forplotting` method for this

tests/test_components/test_simulation.py
bzhangflex: Added a `TestAnisotropicPlotting` class in `test_simulation.py`

tidy3d/components/simulation.py
```diff
@@ -513,6 +519,17 @@ def plot_eps(
             * `Visualizing geometries in Tidy3D: Plotting Permittivity <../../notebooks/VizSimulation.html#Plotting-Permittivity>`_
         """
 
+        # check that eps_component is one of the allowed values, otherwise raise an error
+        if eps_component is not None:
+            # I'm not sure what the best way to check that eps_component is one of the
+            # values in the PermittivityComponent literal
+            if eps_component not in PermittivityComponent.__args__:
```
bzhangflex: I wasn't sure what the best way to check if a value is one of the allowed options in a Literal, if there's a better way please let me know.

tidy3d/components/scene.py
bzhangflex: Updated to use a literal, but I wasn't sure what the best way to check that the value passed in is one of the allowed Literal values. If there's a better way than what I have please let me know, I added a comment where this was updated

bzhangflex:
Looks like some tests are failing, I'll need to fix before requesting a new review

bzhangflex:
I think this is ready for a second review

tidy3d/components/simulation.py
tylerflex: hm, I think the "best" way to do it is to separate the tuple from the literal..

```py
PermittivityComponents = ("blah", "blah", ...)
PermittivityComponent = Literal[PermittivityComponents]
```

and then import and use the `PermittivityComponents`. It's not pretty but probably better than using the __args__ in this case, just in case that changes.

tests/utils.py
```diff
@@ -470,6 +471,12 @@ def make_custom_data(lims, unstructured):
             geometry=td.Box(size=(1, 1, 1), center=(-1, 0, 0)),
             medium=td.AnisotropicMedium(xx=td.PEC, yy=td.Medium(), zz=td.Medium()),
         ),
+        # # Test a fully anistoropic medium
```
tylerflex: should we add this back?

tidy3d/components/medium.py
```diff
@@ -1053,6 +1054,30 @@ def eps_comp(self, row: Axis, col: Axis, frequency: float) -> complex:
             return self.eps_model(frequency)
         return 0j
 
+    def _eps_forplotting(
```
tylerflex: could we rename this? maybe just `_eps_plot` is good enough?

tidy3d/components/medium.py
```diff
@@ -1053,6 +1054,30 @@ def eps_comp(self, row: Axis, col: Axis, frequency: float) -> complex:
...
+        self, frequency: float, eps_component: Optional[PermittivityComponent] = None
+    ):
+        """Returns epsilon for plotting. A specific component of the epsilon tensor can
```
tylerflex: might want to mention "real part"

tidy3d/components/medium.py
```diff
@@ -1053,6 +1054,30 @@ def eps_comp(self, row: Axis, col: Axis, frequency: float) -> complex:
             return self.eps_model(frequency)
         return 0j
 
+    def _eps_forplotting(
+        self, frequency: float, eps_component: Optional[PermittivityComponent] = None
+    ):
```
tylerflex: please add return type, probably just `): -> float`

tidy3d/components/medium.py
```diff
@@ -1053,6 +1054,30 @@ def eps_comp(self, row: Axis, col: Axis, frequency: float) -> complex:
...
+        """Returns epsilon for plotting. A specific component of the epsilon tensor can
+        be selected for anisotropic medium.
+
+        Parameters
+        ----------
+        frequency : float
+            Frequency to evaluate permittivity at.
+        eps_component : PermittivityComponent
+            Component of the permittivity tensor to plot
+            e.g. ``xx``, ``yy``, ``zz``, ``xy``, ``yz``, ...
```
tylerflex: let's put the strings in inner quotes too. so they look like strings in the docs, eg.


``"xx"``

tidy3d/components/medium.py
```diff
@@ -1530,6 +1555,51 @@ def eps_diagonal(self, frequency: float) -> Tuple[complex, complex, complex]:
...
+        -------
+        Tuple[float, float]
+            The min and max values of the permittivity for the selected component and evaluated at ``frequency``.
+        """
+        eps_dataarray = self.eps_dataarray_freq(frequency)
+
+        comps = ["xx", "yy", "zz"]
+        if eps_component in comps:
+            # Return the bounds of a specific component
+            eps = _get_numpy_array(np.real(eps_dataarray[comps.index(eps_component)])).ravel()
```
tylerflex: think it might be worth defining a function in the scope of this method.

```py
def get_eps_values(x: np.ndarray) -> np.ndarray:
    """Grab the real part of the values in array"""
    return _get_numpy_array(np.real(x)).ravel()
```

because this is repeated a few times.

tidy3d/components/types.py
```diff
@@ -215,6 +215,8 @@ def __modify_schema__(cls, field_schema):
 FreqBoundMin = float
 FreqBound = Tuple[FreqBoundMin, FreqBoundMax]
 
+PermittivityComponent = Literal["xx", "xy", "xz", "yx", "yy", "yz", "zx", "zy", "zz", None]
```
tylerflex: typically, we just leave out `None` in the `Literal` and handle it separately. 

For example, if the argument is optional

```py
def f(c: PermittivityComponent = None)
```

or you can also do 
```py
comp : typing.Optional[PermittivityComponent]
```

tidy3d/components/scene.py
```diff
@@ -1112,18 +1118,30 @@ def _pcolormesh_shape_custom_medium_structure_eps(
                             axis=normal_axis_ind, pos=normal_position
                         )
                 else:
-                    eps_mean = (eps_diag[0] + eps_diag[1] + eps_diag[2]) / 3
+                    if eps_component == "xx":
```
tylerflex: There's some logic duplication below. I think similar to the comp2ind you had earlier, you can define one for the double index components 

```
comp2ind = {dim+dim: index for dim, index in zip('xyz', range(3))}
```

and then,

```py
if eps_component is None:
    eps = ...
elif eps_component in comp2ind:
    eps = eps_diag[comp2ind[eps_component]]
else
    ....
```

tylerflex (changes requested):
Hi @bzhangflex thanks for the changes. it's looking much better. Just had a few more kind of minor things to look at.

tests/utils.py
bzhangflex: I added this at first because I was using `SIM_FULL` for testing, but I'm no longer using `SIM_FULL` anymore. I added it back in now in case it's good to have, or we can remove it if it's not needed

tests/utils.py
```diff
@@ -470,6 +471,12 @@ def make_custom_data(lims, unstructured):
             geometry=td.Box(size=(1, 1, 1), center=(-1, 0, 0)),
             medium=td.AnisotropicMedium(xx=td.PEC, yy=td.Medium(), zz=td.Medium()),
         ),
+        # Test a fully anistropic medium
+        td.Structure(
```
bzhangflex, line 475: I added this at first when I was using `SIM_FULL` for pytests, but now that I'm no longer using SIM_FULL I'm not sure if we want to keep this or not.

bzhangflex:
Thanks @tylerflex I pushed some more updates, let me know how it looks when you get the chance. One of Weiliang's commits slipped in when I rebased, hopefully that doesn't cause confusion. I will clean up the commit history before this is ready to merge

tests/utils.py
yaugenst-flex: no harm in having it in there i think

tests/test_components/test_simulation.py
```diff
@@ -693,6 +694,200 @@ def test_plot_eps_bounds():
...
+    offdiag_comps = ["xy", "yx", "xz", "zx", "yz", "zy"]
+    allcomps = diag_comps + offdiag_comps
+
+    medium_diag = td.AnisotropicMedium(
+        xx=td.Medium(permittivity=5), yy=td.Medium(permittivity=10), zz=td.Medium(permittivity=15)
+    )
+
+    medium_fullyani = td.FullyAnisotropicMedium(permittivity=[[6, 2, 3], [2, 7, 4], [3, 4, 8]])
+
+    def medium_customani(self):
```
yaugenst-flex, line 714: minor suggestion: this could be a class-scoped fixture so that you don't instantiate the medium in every parametrized test. might be unnecessarily slow otherwise and should be safe since it's a static component

so `@pytest.fixture(scope="class")`

tests/test_components/test_simulation.py
```diff
@@ -693,6 +694,200 @@ def test_plot_eps_bounds():
...
+    def test_plot_fully_anisotropic_medium_diff(self, fig_test, fig_ref, eps_comp1, eps_comp2):
+        """Tests that the plots of different components of a FullyAnisotropicMedium are actually different."""
+        sim = self.make_sim(self.medium_fullyani)
+
+        ax1 = fig_test.add_subplot()
+        sim.plot_eps(x=0, eps_component=eps_comp1, ax=ax1)
+        ax2 = fig_ref.add_subplot()
+        sim.plot_eps(x=0, eps_component=eps_comp2, ax=ax2)
+
+    # TODO: test CustomAnisotropicMedium
```
yaugenst-flex: TODO outdated?

tests/test_components/test_simulation.py
yaugenst-flex, line 1: Nice tests!

tidy3d/components/types.py
```diff
@@ -215,6 +215,9 @@ def __modify_schema__(cls, field_schema):
 FreqBoundMin = float
 FreqBound = Tuple[FreqBoundMin, FreqBoundMax]
 
+PermittivityComponents = ("xx", "xy", "xz", "yx", "yy", "yz", "zx", "zy", "zz")
+PermittivityComponent = Literal[PermittivityComponents]
```
yaugenst-flex: Not a big fan of this pattern. `PermittivityComponents` is not a type so technically doesn't even belong in this file. The canonical way to do this would be via [`typing.get_args`](https://docs.python.org/3/library/typing.html#typing.get_args), so when checking if a value is in a literal you would do `if x in get_args(type)`

tidy3d/components/simulation.py
```diff
@@ -513,6 +520,17 @@ def plot_eps(
             * `Visualizing geometries in Tidy3D: Plotting Permittivity <../../notebooks/VizSimulation.html#Plotting-Permittivity>`_
         """
 
+        # check that eps_component is one of the allowed values, otherwise raise an error
+        if eps_component is not None:
+            # I'm not sure what the best way to check that eps_component is one of the
+            # values in the PermittivityComponent literal
+            if eps_component not in PermittivityComponents:
```
yaugenst-flex: see comment above

`if eps_component not in get_args(PermittivityComponent)`

tidy3d/components/types.py
tylerflex: my bad, I suggested this (we use it elsewhere). get_args works, didnt know about it.

tests/utils.py
tylerflex: SIM_FULL is passed around in a lot of other tests and so it's good to have an instance of every kind object in there as it can catch edge case errors. so I think best to leave it in.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `fill` and `fill_structures` argument in `td.Simulation.plot_structures()` and `td.Simulation.plot()` respectively to disable fill and plot outlines of structures only.
+- New subpixel averaging option `ContourPathAveraging` applied to dielectric material boundaries.
```
tylerflex, line 12: looks like this and the `docs/api/subpixel_averaging.rst` got overwritten or something?

tidy3d/__init__.py
tylerflex, line 1: note, seems like unrelated changes

tidy3d/components/subpixel_spec.py
tylerflex, line 1: note, seems like this snuck in

tylerflex (approved):
thanks @bzhangflex , looks good to me now once you figure out how to resolve the sub pixel commits sneaking in here. thanks a lot!

tests/test_components/test_simulation.py
bzhangflex: thanks 😄

CHANGELOG.md
bzhangflex: I think I managed to rebase successfully, please let me know how it looks

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `fill` and `fill_structures` argument in `td.Simulation.plot_structures()` and `td.Simulation.plot()` respectively to disable fill and plot outlines of structures only.
 - New subpixel averaging option `ContourPathAveraging` applied to dielectric material boundaries.
 
+- `eps_component` argument in `td.Simulation.plot_eps()` to optionally select a specific permittivity component to plot.
```
bzhangflex: I added this to the changelog, let me know if it looks ok

tests/test_components/test_simulation.py
```diff
@@ -693,6 +697,209 @@ def test_plot_eps_bounds():
...
+        r = 1
+        n_data[r_mesh <= r] = n0 * (1 - A * r_mesh[r_mesh <= r] ** 2)
+        # convert to dataset array
+        n_dataset = td.SpatialDataArray(n_data, coords=dict(x=x, y=y, z=z))
+        medium_zz = td.CustomMedium.from_nk(n_dataset, interp_method="nearest")
+
+        return td.CustomAnisotropicMedium(xx=medium_xx, yy=medium_yy, zz=medium_zz)
+
+    @pytest.fixture(scope="class", autouse=True)
+    def cleanup_figures(self):
```
bzhangflex, line 747: @yaugenst-flex It turns out I can't specify a save path for matplotlib test figures, this was the solution chatgpt suggested 😂 It works pretty well but let me know if you see issues with it

bzhangflex:
Thanks @tylerflex and @yaugenst-flex ! Hopefully these are the finishing touches, but please let me know if there's any additional changes I can make.

CHANGELOG.md
tylerflex: looks good!

CHANGELOG.md
tylerflex: looks good, I would consider adding a tiny bit more `eg. "xx".` just to make it clear.

e-g-melo (approved):

tests/test_components/test_simulation.py
```diff
@@ -693,6 +698,211 @@ def test_plot_eps_bounds():
...
+        yield  # Run tests first
+        plt.close("all")  # Close all figures before deleting the directory
+        time.sleep(0.5)
+        fig_dir = "result_images"
+        if os.path.exists(fig_dir):
```
bzhangflex: dont check if exists

tests/test_components/test_simulation.py
bzhangflex: Think I got it to work, changing the fixture scope from "class" to "module" did the trick. I guess it was deleting the folder too early, before all the tests were run.

PR #2319, opened 2025-03-14
fix[adjoint]: fix tuple handling in autograd gradient calculations

yaugenst-flex:
This PR addresses an issue where tuple-type gradients were not properly
handled during gradient calculations. The fix ensures
that:
1. Zero gradients maintain proper tuple structure instead of being converted to empty tuples when early-returning from adjoint pipeline (no adjoint sim).
2. Gradient accumulation from multiple adjoint simulations properly combines tuple values element-wise
3. Refactored gradient accumulation code to live outside of `local_gradient=True/False`, i.e. this happens in one place only now.

### DIFF
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -813,6 +813,83 @@ def test_autograd_async(use_emulated_run, structure_key, monitor_key):
     assert anp.all(grad != 0.0), "some gradients are 0"
 
 
+class TestTupleGrads:
+    center0 = (0.0, 0.0, 0.0)
+    size0 = (0.5, 1.0, 1.5)
+
+    @staticmethod
+    def make_simulation(center: tuple, size: tuple) -> td.Simulation:
+        wavelength = 1.0
+        freq0 = td.C_0 / wavelength
+
+        src = td.PointDipole(
+            center=(-1.4, 0, 0),
+            source_time=td.GaussianPulse(freq0=freq0, fwidth=freq0 / 10),
+            polarization="Ex",
+        )
+
+        mnt = td.FieldMonitor(
+            size=(0, 0, 1),
+            center=(1.4, 0, 0),
+            freqs=[freq0, freq0 + freq0 / 50],
+            name="fields",
+        )
+
+        scatterer = td.Structure(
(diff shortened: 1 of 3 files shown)


yaugenst-flex:
Note this is blocked by #2311

tests/test_components/test_autograd.py
```diff
@@ -813,6 +813,83 @@ def objective(*args):
...
+        val, (dp_dcenter, dp_dsize) = d_power(self.center0, self.size0)
+
+        assert len(dp_dcenter) == 3
+        assert len(dp_dsize) == 3
+
+        if zero:
+            assert np.allclose(dp_dcenter, 0)
+            assert np.allclose(dp_dsize, 0)
+        else:
+            assert not np.allclose(dp_dcenter, 0)
```
groberts-flex, line 889: without the fix, are these either zero or empty tuples?

groberts-flex:
Good catch/fix here! Added one comment/question mostly for my own understanding of what's going wrong here. Only general question I have on this is if we suspect things like this could be happening in other places as well and if so, if there's a way to more generally handle this source of error where the type could be list, tuple, or numpy array/traced fields?

yaugenst-flex:
> if there's a way to more generally handle this source of error where the type could be list, tuple, or numpy array/traced fields?

So this was specific to the gradient accumulation in the traced fields, where we were sometimes accumulating tuples like `dict[key]: tuple += other_tuple`, which ends up being `(*old_tuple, *new_tuple)`, which is of course incorrect. I don't think this happens outside of the gradient accumulation code though. There is some other weirdness with types that I've noticed like the grads for `center` ending up as `(array(x), array(y), array(z))` instead of just a traced tuple `(x, y, z)`, which is something that I'd like to look into a bit but it doesn't seem to be breaking anything.

groberts-flex:
sounds good, thanks for the explanation! this looks good to me

groberts-flex (approved):

tests/test_components/test_autograd.py
yaugenst-flex: yup exactly. test fails on previous commit

Issue #2277, opened 2025-02-26
Improve the PML placement in mode solver


e-g-melo:
# Current

When we set the PML in mode solver, it is placed within the simulation plane. For example, here `num_pml = [7, 12]`.

<img width="343" alt="image" src="https://github.com/user-attachments/assets/25db832b-95b4-464b-9205-00dbac190e61" />

So, if the user sets a large number of PML layers, for example `num_pml = [7, 20]`, we can have situations like this:

<img width="343" alt="image" src="https://github.com/user-attachments/assets/a5e18003-f790-46b5-a2fa-45dc393d1bd3" />

Users would not expect it, and if they don't know how the PML is placed, they might have issues with the results.

# Requirement
Place the PML surrounding the mode plane. For example:

<img width="343" alt="image" src="https://github.com/user-attachments/assets/01a95960-9fc7-4fa5-b715-c69bccd78910" />

caseyflex:
Thanks @e-g-melo, I think we need to discuss exactly how to handle this, but it would be good to unify BoundarySpec PML with ModeSpec num_pml, in ModeMonitors / ModeSolver and in ModeSimulation.

One option is to add a 2D mode_boundary_spec field to ModeMonitor, ModeSolver, and ModeSimulation. This might be better than adding it to ModeSpec because it modifies the actual geometry and grid. This field would be treated in the mode solver like the boundary_spec pml - extend the simulation and grid (assuming structures extend into it), then superimpose PML on the new region.

If a ModeMonitor doesn’t extend across the entire FDTD simulation, would the resulting grid in the PML region of the ModeMonitor agree with the grid in the surrounding region of the FDTD simulation?

Of course not all boundary types would be supported - just PEC and PML I think.

Then we could deprecate num_pml, add an option “auto” to just use the mode_boundary_spec pml (maybe as default), and keep manual num_pml for backwards compatibility.

There are a lot of different use cases here to consider, as well as implementation subtleties.

e-g-melo:
@caseyflex, these are all good points.

Maybe the ModeMonitor, ModeSolverMonitor (and ModeSource?) would be the most critical cases. I think we should not perturb the FDTD simulation grid because it would be difficult for users to understand why the grid changes when they add a monitor at another position. 

So that could mean we need to agree with the surrounding grid or create a kind of "artificial" PML grid surrounding the mode plane only when calculating the modes.

[...]

PR #2320, opened 2025-03-14
Warn when mode solver pml covers a significant portion of the mode plane

caseyflex:
Partially addresses this issue:

https://github.com/flexcompute/tidy3d/issues/2277

Adds a warning when the PML covers a significant portion of the mode plane, and makes it easy to get the mode PML thickness (e.g. for GUI plotting).

Doesn't change the mode PML behavior, API, or plotting.

### DIFF
--- a/tests/test_components/test_simulation.py
+++ b/tests/test_components/test_simulation.py
@@ -14,6 +14,7 @@ from tidy3d.components import simulation
 from tidy3d.components.scene import MAX_GEOMETRY_COUNT, MAX_NUM_MEDIUMS
 from tidy3d.components.simulation import MAX_NUM_SOURCES
 from tidy3d.exceptions import SetupError, Tidy3dError, Tidy3dKeyError
+from tidy3d.plugins.mode import ModeSolver
 
 from ..utils import (
     SIM_FULL,
@@ -2749,7 +2750,11 @@ def test_sim_subsection(unstructured, nz):
     sim_red = SIM_FULL.subsection(
         region=region, boundary_spec=td.BoundarySpec.all_sides(td.Periodic())
     )
-    sim_red = SIM_FULL.subsection(region=region, sources=[], grid_spec=td.GridSpec.uniform(dl=20))
+    sim_red = SIM_FULL.subsection(
+        region=region,
+        sources=[],
+        grid_spec=td.GridSpec.uniform(dl=20),
+    )
     assert len(sim_red.sources) == 0
     sim_red = SIM_FULL.subsection(region=region, monitors=[])
     assert len(sim_red.monitors) == 0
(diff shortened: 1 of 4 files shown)


tests/test_components/test_simulation.py
```diff
@@ -2543,7 +2544,12 @@ def test_sim_subsection(unstructured, nz):
     sim_red = SIM_FULL.subsection(
         region=region, boundary_spec=td.BoundarySpec.all_sides(td.Periodic())
     )
-    sim_red = SIM_FULL.subsection(region=region, sources=[], grid_spec=td.GridSpec.uniform(dl=20))
```
caseyflex, line 2752: @dbochkov-flexcompute I saw from git blame that you wrote this originally, so I wanted to check with you if it's ok to change it, or if I'm missing something about simulation subsection. I removed the mode monitors in the subsection call in this test. When checking these mode monitors to warn if the pml was too large, the `ModeSolver._grid_snapped` was generated (just to get the exact size). But this was erroring in `Grid.snap_to_box_zero_dim` because the mode monitor was completely outside of the simulation subsection.

```
tests/test_components/test_simulation.py:2547: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tidy3d/components/simulation.py:1786: in subsection
    return self.updated_copy(**new_sim_dict, deep=deep_copy)
tidy3d/components/base.py:248: in updated_copy
    return self._updated_copy(**kwargs, deep=deep, validate=validate)
tidy3d/components/base.py:293: in _updated_copy
    return self.copy(update=kwargs, deep=deep, validate=validate)
tidy3d/components/base.py:225: in copy
    return self.validate(new_copy.dict())
/home/casey/virtualenvs/tidy3d/lib/python3.12/site-packages/pydantic/v1/main.py:711: in validate
    return cls(**value)
tidy3d/components/base.py:148: in __init__
    self._post_init_validators()
tidy3d/components/simulation.py:3534: in _post_init_validators
    self._warn_mode_object_pml()
tidy3d/components/simulation.py:3543: in _warn_mode_object_pml
    ModeSolver._warn_thick_pml(
tidy3d/components/mode/mode_solver.py:225: in _warn_thick_pml
    size_no_pml = cls._mode_plane_size_no_pml(
tidy3d/components/mode/mode_solver.py:2259: in _mode_plane_size_no_pml
    pml_thickness = cls._pml_thickness(simulation=simulation, plane=plane, mode_spec=mode_spec)
tidy3d/components/mode/mode_solver.py:2214: in _pml_thickness
    grid_snapped = cls._grid_snapped(simulation=simulation, plane=plane)
tidy3d/components/mode/mode_solver.py:407: in _grid_snapped
    return simulation._snap_zero_dim(grid_snapped)
tidy3d/components/simulation.py:1188: in _snap_zero_dim
    return grid.snap_to_box_zero_dim(Box(center=self.center, size=size_snapped))


E                   ValueError: Cannot snap grid to box center outside of grid domain.
```

@momchil-flex fyi

caseyflex:
@momchil-flex fyi. You mentioned changing the plotting for mode monitors and mode sources to show the mode pml. I didn't include this here because I wasn't sure about it -- it would clutter the plotting, and maybe it's not that critical most of the time? But we could consider including that or unifying the plotting between the `ModeSolver` and the mode objects in FDTD simulations (currently `ModeSolver` shows the pml while FDTD mode objects do not). It would be easy to include in this PR if we decide to make that plotting change.

tests/test_components/test_simulation.py
dbochkov-flexcompute: If the user sets up a mode solver monitor and places it outside of simulation box from the beginning (without invoking subsection function), will it error in the same way?

caseyflex: Thanks Daniil.

No, in that case the error is more helpful, that the monitor is outside the simulation domain. Here, there is a grid subtlety, the simulation subsection is at .02 while the mode monitor grid is at 0 in its normal direction. Do you think it is an error in `Simulation.subsection` or something introduced by this PR? Either way I'll take a closer look tomorrow.

caseyflex: I think I tracked down the issue, this is resolved now, I'll comment separately on the relevant code

yaugenst-flex (approved):
Thanks @caseyflex looks good to me! We really need a config though for these globals... 😄

tidy3d/components/mode/mode_solver.py
```diff
@@ -353,8 +392,22 @@ def _get_data_with_group_index(self) -> ModeSolverData:
...
+    @classmethod
+    def _grid_snapped(cls, simulation: Simulation, plane: Box) -> Grid:
+        """The solver grid snapped to the plane normal and to simulation 0-sized dims if any."""
+        solver_grid = cls._get_solver_grid(
+            simulation=simulation,
+            plane=plane,
+            keep_additional_layers=False,
+            truncate_symmetry=True,
+        )
+        # snap to plane center along normal direction
```
caseyflex, line 410: @momchil-flex @dbochkov-flexcompute  fyi

I tracked down the issue to this PR:
https://github.com/flexcompute/tidy3d/pull/851

It's pretty minor but it comes up when the simulation only has one grid cell in a direction and has a different center than the mode plane. Before that PR, the normal dim was snapped to mode plane and 0 size tangential dims were snapped to simulation. After that PR, the normal dim was snapped to mode plane and all 0 size dims (including normal dim) were snapped to simulation.

tidy3d/components/mode/mode_solver.py
```diff
@@ -2131,12 +2190,82 @@ def plot_grid(
...
+        """
+        # Get the mode plane normal axis, center, and limits.
+        _, _, _, t_axes = cls._center_and_lims(simulation=simulation, plane=plane)
+
+        solver_symmetry = cls._solver_symmetry(simulation=simulation, plane=plane)
+        grid_snapped = cls._grid_snapped(simulation=simulation, plane=plane)
+
+        # Mode plane grid.
+        plane_grid = grid_snapped.boundaries.to_list
+        coord_0 = plane_grid[t_axes[0]]
```
caseyflex, line 2215: @e-g-melo I modified this logic a bit to handle some edge cases where the simulation is small in a tangential direction. I changed `centers -> boundaries` in plane_grid, removed the `[1:-1]` from `coord_0` and `coord_1`, and upper bounded `num_pml_0` by `len(coord_0) - 1` and likewise for `num_pml_1`. It should only affect the plotting and this new warning. Let me know if you think this change is OK.

caseyflex: @e-g-melo here is the PR where you added this function:
https://github.com/flexcompute/tidy3d/issues/1397

caseyflex:
@e-g-melo can we merge this?

e-g-melo (approved):
Hi @caseyflex! Thanks for this work!

I am ok with the changes to plotting. 

After trying it, I suggest reducing that threshold to 50% of the mode plane.

momchil-flex:
Sorry, one thing this is making me think about is that e.g. in RF CPW simulations, the mesh could be much much finer around the waveguide than at the outer edges of the mode plane. So while the *physical* area of the PML may appear large, there is no problem because the majority of the grid *cells* are outside of PML. I feel like generally checking on the grid points level rather than physical area might be better?

caseyflex:
@momchil-flex I reworked it so the warning is now based on num_cells (although the area functions are still available if needed)

momchil-flex (approved):
Looks good!

PR #2323, opened 2025-03-18
chore: add dependency group for `design` plugin

### LINKED ISSUE #2322: Broken dependency in design space Bayesian optimization
**Describe the bug**
The Bayesian Optimizer `MethodBayOpt` relies on an external library `bayesian-optimization`. The external library changed their API in an update that breaks dependency with `MethodBayOpt`. The most recent version of `bayesian-optimization` that still works is `1.5.1`. 

Version history: https://github.com/bayesian-optimization/BayesianOptimization/releases
Update that broke the dependency: https://github.com/bayesian-optimization/BayesianOptimization/pull/447

**To Reproduce**
1. Install the latest version of `bayesian-optimization` library
2. Open a notebook that uses `MethodBayOpt` or set up a dummy `DesignSpace`.
3. Execute the design space using `run()`. 

**Expected behavior**
The design space `run()` fails with an error message about missing `bayes_opt` library. 

**Additional context**
The version `1.5.1` of `bayesian-optimization` has been tested to work successfully without error. Since this is an external library that the user is expected to install by themselves, we can update our documentation to emphasize that they should install the older `1.5.1` version.

yaugenst-flex:
Fixes #2322 by introducing a `design` plugin dependency group installable via `pip install tidy3d[design]` and updating the warning message if `bayesian-optimization` is not installed as suggested in https://github.com/flexcompute/tidy3d/issues/2322#issuecomment-2730640910.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -171,6 +171,63 @@ Some users or systems may require a more specialized installation, which we will
 
         See the development installation `instructions <../development/index.html>`_. You can install ``tidy3d`` within reproducible environment guaranteed by the developers using the ``poetry.lock`` installation and the ``poetry`` toolchain.
 
+Optional Dependencies
+=====================
+
+Tidy3D provides several optional dependency groups that you can install based on your specific needs.
+
+Installing Optional Core Dependencies
+-------------------------------------
+
+Tidy3D has several optional dependencies that provide additional functionality. You can install these using the following syntax:
+
+.. code-block:: bash
+
+    pip install "tidy3d[dependency_group]"
+
+Where ``dependency_group`` is one of the following:
+
+- ``gdspy``: Adds support for GDS export using `gdspy <https://github.com/heitzmann/gdspy>`_.
+- ``gdstk``: Adds support for GDS export using `gdstk <https://github.com/heitzmann/gdstk>`_.
+- ``trimesh``: Support for more complex mesh handling and manipulation.
+- ``vtk``: Support for working with unstructured data.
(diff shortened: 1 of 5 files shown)


daquinteroflex (approved):
Looks great! Should the getting started installation instructions be updated with a quick line on the docs?

yaugenst-flex:
Thanks @daquinteroflex I added a section to `install.rst` to document optional dependencies, if you could have a look?

docs/install.rst
```diff
@@ -171,6 +171,62 @@ Some users or systems may require a more specialized installation, which we will
...
+.. code-block:: bash
+
+    pip install "tidy3d[dependency_group]"
+
+Where ``dependency_group`` is one of the following:
+
+- ``gdspy``: Adds support for GDS export using `gdspy <https://github.com/heitzmann/gdspy>`_.
+- ``gdstk``: Adds support for GDS export using `gdstk <https://github.com/heitzmann/gdstk>`_.
+- ``trimesh``: Support for more complex mesh handling and manipulation.
+- ``vtk``: Support for working with unstructured data.
```
daquinteroflex, line 193: Would just add like that from the heat and charge solvers

docs/install.rst
daquinteroflex, line 1: Nice this is really good thanks! Seems the rst is formatted properly too, we can check when it builds the develop docs

yuanshen-flexcompute (approved):

docs/install.rst
yaugenst-flex: updated, added a `heatcharge` dependency group

Issue #1372, opened 2024-01-10
Unify and test `eps_model`


momchil-flex:
Currently `eps_model` works for a list of frequencies in some cases, and not in others. I encountered this before as well, in fact I think I fixed it for anisotropic media, but it's still an issue for custom media. The method docstring actually suggests that only a float should be passed, but I believe there are many instances where we pass an array instead. Ideally we should make this consistent across all media, or it will keep popping up. We can discuss whether to make it work for a list of frequencies or not (I am not sure what would make most sense for CustomMedium), but basically we should either

- Make it accept single frequency only like the docstrings currently suggest, and fix all instances where we pass a list/array of freqs (not the most convenient, and may be inefficient too)
- Make all versions of that method work with list of freqs, update the docstrings, and add tests or update existing ones.

tylerflex:
probably a good idea to nail this down yea.

I dont think it should be too hard to make things work with floats and arrays, and if we can get that to work, we can probably refactor things to take lists too (convert to arrays, run, and convert back).

For testing purposes we could just utilize SIM_FULL, which should have all mediums in it ideally.

Just iterate through the structures in SIM_FULL and call `structure.medium.eps_model()` passing a float, list, and array?

PR #2324, opened 2025-03-18
Add metal surface roughness models

weiliangjin2021:


Frank-DingYC:
I think the code is good to me.
I use Grok to search for the Hammerstad Surface Roughness model. [Here](https://grok.com/share/bGVnYWN5_502c82c2-b67c-4588-b752-add1742d47eb) is the link. If it's described correctly, we can suggest the user use it as the Grok described.
image

weiliangjin2021:
@Frank-DingYC we can add the feasible regime to the description, but I don't think we should fill in default roughness values.

tidy3d/components/medium.py
```diff
@@ -5180,6 +5181,190 @@ class SurfaceImpedanceFitterParam(Tidy3dBaseModel):
...
+        -----
+            The roughness correction factor should be causal. It is multiplied to the
+            surface impedance of the lossy metal to account for the effects of surface roughness.
+
+        Parameters
+        ----------
+        frequency : float
+            Frequency to evaluate roughness correction factor at (Hz).
+        skin_depths : float
+            Skin depths of the lossy metal that is frequency-dependent.
```
dbochkov-flexcompute: these can be arrays, right?

tidy3d/components/medium.py
```diff
@@ -5180,6 +5181,190 @@ class SurfaceImpedanceFitterParam(Tidy3dBaseModel):
...
+
+    Note
+    ----
+    This model is based on:
+
+        J. Eric Bracken, "A Causal Huray Model for Surface Roughness", DesignCon, 2012.
+    """
+
+    relative_area: pd.PositiveFloat = pd.Field(
+        1,
```
dbochkov-flexcompute, line 5312: is it bounded by 0 and 1?

dbochkov-flexcompute (commented):
looks great, just some minor comments

tidy3d/components/medium.py
weiliangjin2021: yes

tidy3d/components/medium.py
weiliangjin2021: I read somewhere that "In general, set Amatte/Aflat = 1 , and it is set larger than one only for very high roughness". So it can exceed 1.

tidy3d/components/medium.py
weiliangjin2021: But I'm following the convention in this file, e.g. `eps_model(self, frequency: float)` can take arrays too.

weiliangjin2021:
For your reference, here are the doc pages for the two models:
image
image

tidy3d/components/medium.py
dbochkov-flexcompute: I think it was unintentional there https://github.com/flexcompute/tidy3d/issues/1372

weiliangjin2021: ha, i see. Forgot that it was assigned to me!

weiliangjin2021: changed to Array1D now

tidy3d/components/medium.py
```diff
@@ -5180,6 +5182,203 @@ class SurfaceImpedanceFitterParam(Tidy3dBaseModel):
     )
 
 
+class AbstractSurfaceRoughness(Tidy3dBaseModel):
+    """Abstract class for modelling surface roughness of lossy metal."""
```
dmarek-flex: We should use American spelling for `modeling`.

tidy3d/components/medium.py
```diff
@@ -5180,6 +5182,203 @@ class SurfaceImpedanceFitterParam(Tidy3dBaseModel):
...
+    Note
+    ----
+
+        The power loss compared to smooth surface is described by:
+
+        .. math::
+
+            1 + (RF-1) \\frac{2}{\\pi}\\arctan(1.4\\frac{R_q^2}{\\delta^2})
+
+        where :math:`\\delta` is skin depth.
```
dmarek-flex: name the other terms in the equation as well

tidy3d/components/medium.py
```diff
@@ -5180,6 +5182,203 @@ class SurfaceImpedanceFitterParam(Tidy3dBaseModel):
...
+    """Huray surface roughness model.
+
+    Note
+    ----
+
+        The power loss compared to smooth surface is described by:
+
+        .. math::
+
+            \\frac{A_{matte}}{A_{flat}} + \\frac{3}{2}\\sum_i f_i/[1+\\frac{\\delta}{r_i}+\\frac{\\delta^2}{2r_i^2}]
```
dmarek-flex, line 5296: Also describe each symbol here, since they don't exactly match the reference symbol usage.

dmarek-flex (approved):
Nice, excited to see further improvements to benchmarking with this feature! Just a few minor changes please.

tidy3d/components/medium.py
```diff
@@ -5210,6 +5409,13 @@ class LossyMetalMedium(Medium):
         1.0, title="Permittivity", description="Relative permittivity.", units=PERMITTIVITY
     )
 
+    roughness: SurfaceRoughnessType = pd.Field(
+        None,
+        title="Surface Roughness Model",
+        description="Surface roughness model.",
```
dmarek-flex: Perhaps add a little more description here explaining that surface roughness is modeled as a frequency-dependent scaling factor which increases the effective surface impedance.

PR #2325, opened 2025-03-19
Min size autodetect

dbochkov-flexcompute:
I am not sure about naming `SmallestFeatureSpec` and also how deep it is tucked away:
```
td.Simulation(
    ...
    grid_spec=td.GridSpec.auto(
        layer_refinement_specs=[
            td.LayerRefinementSpec(
                corner_finder=td.CornerFinderSpec(
                    dl_min_spec=td.SmallestFeatureSpec(
                        concave_resolution=1,
                        convex_resolution=2,
                        mixed_resolution=3,
                    )
                )
            )
        ]
    )
)
```
would appreciate any suggestions

### DIFF
--- a/tidy3d/components/grid/corner_finder.py
+++ b/tidy3d/components/grid/corner_finder.py
@@ -1,17 +1,17 @@
 """Find corners of structures on a 2D plane."""
 
-from typing import List, Literal, Optional
+from typing import List, Literal, Optional, Tuple
 
 import numpy as np
 import pydantic.v1 as pd
 
 from ...constants import inf
-from ..base import Tidy3dBaseModel
+from ..base import Tidy3dBaseModel, cached_property
 from ..geometry.base import Box, ClipOperation
 from ..geometry.utils import merging_geometries_on_plane
 from ..medium import PEC, LossyMetalMedium
 from ..structure import Structure
-from ..types import ArrayFloat2D, Axis
+from ..types import ArrayFloat1D, ArrayFloat2D, Axis
 
 CORNER_ANGLE_THRESOLD = 0.1 * np.pi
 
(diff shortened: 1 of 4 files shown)


tidy3d/components/grid/corner_finder.py
```diff
@@ -1,21 +1,51 @@
...
+
+    concave_resolution: pd.NonNegativeInt = pd.Field(
+        2,
+        title="Concave Region Resolution.",
+        description="Specifies number of steps to use for determining `dl_min` based on concave featues."
+        "If set to 0, then the corresponding `dl_min` reduction is not applied.",
+    )
+    convex_resolution: pd.NonNegativeInt = pd.Field(
+        0,
+        title="Concave Region Resolution.",
```
weiliangjin2021: convex

tidy3d/components/grid/corner_finder.py
```diff
@@ -1,21 +1,51 @@
...
 
 
+class SmallestFeatureSpec(Tidy3dBaseModel):
+    """Specification for automatic detection of smallest geometric features in layered structures."""
+
+    concave_resolution: pd.NonNegativeInt = pd.Field(
+        2,
+        title="Concave Region Resolution.",
+        description="Specifies number of steps to use for determining `dl_min` based on concave featues."
+        "If set to 0, then the corresponding `dl_min` reduction is not applied.",
```
weiliangjin2021: Not sure if we should use `0` or `None` when not applied

weiliangjin2021:
Regarding deeply nested field, maybe `concave_resolution` etc. can be part of CornerFinderSpec fields directly? We have a field `distance_threshold` already, that might also have some impact on the minimal feature.

Another questions is what would be good default values for those resolutions?

tidy3d/components/grid/corner_finder.py
```diff
@@ -136,5 +221,12 @@ def normalize(v):
         inner_product = np.where(inner_product > 1, 1, inner_product)
         inner_product = np.where(inner_product < -1, -1, inner_product)
         angle = np.arccos(inner_product)
+        num_vs = len(vs_orig)
+        cross_product = np.cross(
```
weiliangjin2021, line 219: `np.cross` seems to have issues in photonforge: https://github.com/flexcompute/tidy3d/pull/2317

weiliangjin2021 (commented):
Went through it again, and now makes more sense to me: your default value in the code is probably good enough.

A high-level question: this PR looks also partially resolve the short-issue. I guess they will work together?

dbochkov-flexcompute:
> Went through it again, and now makes more sense to me: your default value in the code is probably good enough.
> 
> A high-level question: this PR looks also partially resolve the short-issue. I guess they will work together?

yeah, it does partially resolve it, but not as robustly as the other one. For example, if structure has rounded corners everywhere this autodetection feature would not see anything. But given that the other approach is more computationally intensive, might still be good to have a rough estimation like this one

weiliangjin2021:
Just tested the microstrip: I removed lumped elements and substrate layer refinement so that they don't set the minimal grid size. Before this PR, I get 
image

And with the default parameters in this PR, I get
image

Quite nice meshing!

tidy3d/components/grid/corner_finder.py
dbochkov-flexcompute: should be ok here because that warning is only for 2d vectors, and we have full 3d coordinates here

weiliangjin2021 (approved):

PR #2326, opened 2025-03-20
Improve accuracy of normal E-field computation in mode solver

weiliangjin2021:
Consider a CPW at 1 GHz, previously the modal field profile is as follows (x is normal axis). Ex should be 0, but it's very noisy and large instead. `wg TE fraction` should be 1, but it's 0.407814.
image

This PR drops a term that is analytically 0, and now Ex is much smaller, and `wg TE fraction` is computed to 1.
image

### DIFF
--- a/tidy3d/components/mode/solver.py
+++ b/tidy3d/components/mode/solver.py
@@ -660,7 +660,12 @@ class EigSolver(Tidy3dBaseModel):
         Hx = h_field[:N, :] / (1j * neff - keff)
         Hy = h_field[N:, :] / (1j * neff - keff)
         Hz = inv_mu_zz.dot(dxf.dot(Ey) - dyf.dot(Ex))
-        Ez = inv_eps_zz.dot(dxb.dot(Hy) - dyb.dot(Hx))
+        # Ez = inv_eps_zz.dot(dxb.dot(Hy) - dyb.dot(Hx))
+
+        # Ez = -inv_eps_zz * div^H J H_xy, while Hxy = vals^-1 * qmat * Exy;
+        # Note that div^H J q_partial = 0, so Ez = -vals^-1 inv_eps_zz * div^H J q_ep Exy
+        h_partial_field = q_ep.dot(vecs) / (1j * neff - keff)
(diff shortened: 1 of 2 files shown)


weiliangjin2021:
@dbochkov-flexcompute related to your comment in https://github.com/flexcompute/tidy3d-core/issues/918 that shows very large and noisy normal E-field.

tidy3d/components/mode/solver.py
```diff
@@ -660,7 +660,12 @@ def conditional_inverted_vec(eps_vec, threshold=1):
         Hx = h_field[:N, :] / (1j * neff - keff)
         Hy = h_field[N:, :] / (1j * neff - keff)
         Hz = inv_mu_zz.dot(dxf.dot(Ey) - dyf.dot(Ex))
-        Ez = inv_eps_zz.dot(dxb.dot(Hy) - dyb.dot(Hx))
+        # Ez = inv_eps_zz.dot(dxb.dot(Hy) - dyb.dot(Hx))
+
+        # Ez = -inv_eps_zz * div^H J H_xy, while Hxy = vals^-1 * qmat * Exy;
+        # Note that div^H J q_partial = 0, so Ez = -vals^-1 inv_eps_zz * div^H J q_ep Exy
+        h_partial_field = q_ep.dot(vecs) / (1j * neff - keff)
+        Ez = inv_eps_zz.dot(dxb.dot(h_partial_field[N:, :]) - dyb.dot(h_partial_field[:N:, :]))
```
lucas-flexcompute: Just curious: is the index slice notation `:N:` same as `:N`?

lucas-flexcompute (approved):
I'm assuming the eqs are right, coming from @weiliangjin2021, so, really, no comments here… awesome results!

tidy3d/components/mode/solver.py
weiliangjin2021: sorry, I meant to code `:N`. Now surprised that it works in my tests

weiliangjin2021: haha, but you are right, `:N:` is equivalent to `:N`.

momchil-flex (approved):
Amazing!

momchil-flex:
I updated this so I can merge it right now as I'll be making the 2.8.1 patch soon.

PR #2328, opened 2025-03-20
fix: deletion of tasks using `web.delete(task_id)`

yaugenst-flex:
Fixes task deletion not working properly in 2.8.0.
I added an optional argument `versions: bool` to either delete only the version associated with a specific task ID or to delete all versions in the task's group.

```py
import tidy3d.web as web
task_id = "fdve-a124a6f9-33ce-4eac-9189-85fe2ba81dac"
web.delete(task_id)
# web.delete(task_id, versions=True)  # will delete all versions
```

image

-->

image


momchil-flex (approved):
Nice, the delete group tasks addition seems great!

PR #2330, opened 2025-03-20
:package: :robot: Include py3.13 tests

daquinteroflex:


yaugenst-flex:
@daquinteroflex looks like it's working!

pyproject.toml
```diff
@@ -21,9 +22,9 @@ documentation = "https://docs.flexcompute.com/projects/tidy3d/en/latest/"
 "Bug Tracker" = "https://github.com/flexcompute/tidy3d/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<4.0.0"
+python = ">=3.9,<3.14"
```
momchil-flex, line 25: This is always a dilemma: restrict it to 4.0, and maybe it will just work with the next python version, but we risk people coming back to us with something being broken. On the other hand, if we restrict directly we risk people coming back to us saying they can't install and they will only be able to do it when we patch (or we have to explain to them how to use a python venv, or how to install from source and modify this file).

momchil-flex (approved):
Looks good, one comment about python restriction, I really don't know what's better though.

pyproject.toml
momchil-flex: For what it's worth, with other dependencies we try not to restrict the upper version. But I guess python is quite specific in that e.g. we actually run tests for each version that we officially support.

daquinteroflex: Yeah maybe it can be more frustrating if it doesn't work when we say it "should" by specifying <4? Than to say, "it's not officially supported yet till we do a patch and test". So inclined to restrict personally

daquinteroflex:
Thanks @yaugenst-flex looks great!

pyproject.toml
yaugenst-flex: I kind of agree that you should be able to at least try to install on a newer python version even if it's not officially supported, so that's a bit of a dilemma. what about something like this in our `__init__`?
```py
if sys.version_info >= (3, 14):
    warnings.warn("This package has not been tested with Python 3.14+. Proceed at your own risk.")
```

yaugenst-flex: On the other hand, I don't think we've ever _not_ had issues when including a new Python version, so maybe there is not really a point in allowing it? If someone wants to do it anyways and is willing to tinker around then they can do that by cloning the repo?

momchil-flex: Yeah I'm pretty sure we've been oscillating about this. Before it was set to <4 it was actually restricted and then people were like hey I'm using the new version (I think it was 3.12 yeah) and we were like oh I guess let's not restrict it for the next one. :)

But I agree chances are it won't run right off the bat... So I'm fine restricting.

yaugenst-flex: Ok let's keep it like this for now then, and just keep up with the official Python release cycle. I like explicit.

PR #2331, opened 2025-03-21
Add periodic repetition in EME simulations

caseyflex:
Major changes:

- Adds a field `num_reps` to every `EMEGridSpec` class, which can be used to periodically repeat that subgrid while reusing mode solves, overlaps, and interface s-matrices
- Adds a class `EMEPeriodicitySweep` which allows sweeping the `num_reps` of every `EMEGridSpec` in the `EMESimulation.eme_grid_spec`. This is currently done by assigning names to `EMEGridSpec` objects, and then passing `num_reps` as a list of dictionaries, one for each sweep_index, and each of which is a mapping from the name of the subgrid which will be repeated to the desired number of repetitions of that subgrid.

Minor changes:

- When multiple freqs are provided, uses the largest for automeshing instead of raising an error.
- Adds methods `EMEExplicitGrid.from_structures` and `EMECompositeGrid.from_structure_groups` which place grid boundaries at structure bounds
- Increases `MAX_NUM_FREQS` to 100

Todo:

- [x] Check / extend test coverage

### DIFF
--- a/tidy3d/components/eme/grid.py
+++ b/tidy3d/components/eme/grid.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import List, Literal, Tuple, Union
+from typing import Dict, List, Literal, Optional, Tuple, Union
 
 import numpy as np
 import pydantic.v1 as pd
@@ -14,11 +14,13 @@ from ..base import Tidy3dBaseModel, skip_if_fields_missing
 from ..geometry.base import Box
 from ..grid.grid import Coords1D
 from ..mode_spec import ModeSpec
+from ..structure import Structure
 from ..types import ArrayFloat1D, Axis, Coordinate, Size, TrackFreq
 
 # grid limits
 MAX_NUM_MODES = 100
 MAX_NUM_EME_CELLS = 100
+MAX_NUM_REPS = 100000
 
 
 class EMEModeSpec(ModeSpec):
@@ -96,6 +98,29 @@ class EMEGridSpec(Tidy3dBaseModel, ABC):
     in the simulation.
     """
 
+    num_reps: pd.PositiveInt = pd.Field(
+        1,
+        title="Number of Repetitions",
+        description="Number of periodic repetitions of this EME grid. Useful for "
+        "efficiently simulating long periodic structures like Bragg gratings. "
+        "Instead of explicitly repeating the cells, setting 'num_reps' allows "
+        "the EME solver to reuse the modes and cell interface scattering matrices.",
+    )
+
+    name: Optional[str] = pd.Field(
+        None, title="Name", description="Name of this 'EMEGridSpec'. Used in 'EMEPeriodicitySweep'."
+    )
+
+    @pd.validator("num_reps", always=True)
+    def _validate_num_reps(cls, val):
+        """Check num_reps is not too large."""
+        if val > MAX_NUM_REPS:
+            raise SetupError(
+                f"'EMEGridSpec' has 'num_reps={val:.2e}'; "
+                f"the largest value allowed is '{MAX_NUM_REPS}'."
+            )
+        return val
+
     @abstractmethod
     def make_grid(self, center: Coordinate, size: Size, axis: Axis) -> EMEGrid:
         """Generate EME grid from the EME grid spec.
@@ -115,6 +140,46 @@ class EMEGridSpec(Tidy3dBaseModel, ABC):
             An EME grid dividing the EME simulation into cells, as defined
             by the EME grid spec.
         """
+
+    @property
+    def real_cell_indices(self) -> int:
+        """The cell indices inside this EME grid, starting at 0,
+        not including periodic repetition of cells."""
+        return np.arange(self.num_real_cells)
+
+    @property
+    @abstractmethod
+    def num_real_cells(self) -> int:
+        """Number of real cells in this EME grid spec."""
+
+    @property
+    def virtual_cell_indices(self) -> int:
+        """The cell indices inside this EME grid, starting at 0
+        and including periodic repetition of cells with ``num_reps``."""
+        return list(self.real_cell_indices) * self.num_reps
+
+    @property
+    def num_virtual_cells(self) -> int:
+        """Number of virtual cells in this EME grid spec."""
+        return len(self.virtual_cell_indices)
+
+    def _updated_copy_num_reps(self, num_reps: Dict[str, pd.PositiveInt]) -> EMEGridSpec:
+        """Update ``num_reps`` of named subgrids."""
+        if self.name is not None:
+            new_num_reps = num_reps.get(self.name)
+            if new_num_reps is not None:
+                return self.updated_copy(num_reps=new_num_reps)
+        return self
+
+    @property
+    def _cell_index_pairs(self) -> List[pd.NonNegativeInt]:
+        """Pairs of adjacent cell indices."""
+        cell_indices = self.virtual_cell_indices
+        pairs = []
+        for i in range(len(cell_indices) - 1):
+            if (cell_indices[i], cell_indices[i + 1]) not in pairs:
+                pairs.append((cell_indices[i], cell_indices[i + 1]))
+        return pairs
 
 
 class EMEUniformGrid(EMEGridSpec):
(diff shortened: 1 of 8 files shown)


tomflexcompute:
That's great @caseyflex . At the conference this week so hard to find a good chuck of time to carefully review it. Maybe @momchil-flex can take a look?

yaugenst-flex (commented):
Thanks @caseyflex, the code looks good to me, no comments there, only some high level ones regarding documentation.

This might just be my unfamiliarity with EME but from reading through the PR it wasn't really obvious to me when and how you would use this functionality, and for example what is the difference between a `EMEModeSpec(num_reps=N)` and an `EMEPeriodicitySweep`? A more detailed docstring at least for the latter would help greatly I think. Also perhaps a quick motivation in the docstring ("Periodic repetition [...] for X purpose") would go a long way. Is there a related example notebook?

caseyflex:
> Thanks @caseyflex, the code looks good to me, no comments there, only some high level ones regarding documentation.
> 
> This might just be my unfamiliarity with EME but from reading through the PR it wasn't really obvious to me when and how you would use this functionality, and for example what is the difference between a `EMEModeSpec(num_reps=N)` and an `EMEPeriodicitySweep`? A more detailed docstring at least for the latter would help greatly I think. Also perhaps a quick motivation in the docstring ("Periodic repetition [...] for X purpose") would go a long way. Is there a related example notebook?

Ah thanks @yaugenst-flex , I forgot the docstring. I'll add some details there. I think @tomflexcompute will make an example notebook once he has a bit of time.

Currently the api is like this:

```

unit_cell_grid_spec = td.EMEExplicitGrid(
    mode_specs=[mode_spec]*2,
    # boundaries=[0, p*ff, p],
    boundaries=[p*ff],
    name="unit cell")

port_cell_grid_spec = td.EMEUniformGrid(
    mode_spec=mode_spec,
    num_cells=1
)

eme_grid_spec = td.EMECompositeGrid(
    subgrids=[port_cell_grid_spec, unit_cell_grid_spec, port_cell_grid_spec],
    subgrid_boundaries=[0, p]
)

```

this defines a composite grid with two port subgrids (each having one cell) and a unit cell subgrid consisting of two cells which is meant to be periodically repeated, see the image below. The unit cell grid is named

Then there are two main ways you could use this. You could just set `unit_cell_grid_spec.num_reps` directly -- this would effectively repeat the unit cell that many times, and simulate a single structure. Alternatively, you can use `EMEPeriodicitySweep` like this:

```
n_list = [1, 50, 100]
sweep_spec = td.EMEPeriodicitySweep(num_reps=[{"unit cell": n} for n in n_list])
```

then this would simulate three different structures at once, with varying number of repetitions of the unit cell.

Let me know your thoughts on the API!

image

yaugenst-flex (dismissed):
Thanks for that explanation @caseyflex, that's really helpful, and the interface makes sense to me.

tomflexcompute (dismissed):
Thanks @caseyflex for the great addition of the EME capability!

tidy3d/components/eme/grid.py
```diff
@@ -249,6 +319,26 @@ def make_grid(self, center: Coordinate, size: Size, axis: Axis) -> EMEGrid:
             mode_specs=self.mode_specs,
         )
 
+    @classmethod
+    def from_structures(
+        cls, structures: List[Structure], axis: Axis, mode_spec: EMEModeSpec, **kwargs
+    ) -> EMEExplicitGrid:
+        """Create an explicit EME grid with boundaries aligned with
+        structure bounding boxes."""
```
momchil-flex: Since this is user-facing it should have a proper doc string with parameters and return, and maybe an example.

tidy3d/components/eme/grid.py
```diff
@@ -116,6 +141,46 @@ def make_grid(self, center: Coordinate, size: Size, axis: Axis) -> EMEGrid:
...
+    def updated_copy_num_reps(self, num_reps: Dict[str, pd.PositiveInt]) -> EMEGridSpec:
+        """Update ``num_reps`` of named subgrids."""
+        if self.name is not None:
+            new_num_reps = num_reps.get(self.name)
+            if new_num_reps is not None:
+                return self.updated_copy(num_reps=new_num_reps)
+        return self
+
+    @property
+    def cell_index_pairs(self) -> List[pd.NonNegativeInt]:
```
momchil-flex: Should this and `updated_copy_num_reps` be private? Or is there a reason we might want them user-exposed?

tidy3d/components/eme/grid.py
```diff
@@ -380,6 +470,108 @@ def make_grid(self, center: Coordinate, size: Size, axis: Axis) -> EMEGrid:
...
+    def from_structure_groups(
+        cls,
+        structure_groups: List[List[Structure]],
+        axis: Axis,
+        mode_specs: List[EMEModeSpec],
+        names: List[str] = None,
+        num_reps: List[pd.PositiveInt] = None,
+    ) -> EMECompositeGrid:
+        """Create a composite EME grid with boundaries aligned with
+        structure bounding boxes."""
```
momchil-flex: Same comment about full docstring.

tidy3d/components/eme/simulation.py
```diff
@@ -752,6 +784,15 @@ def _validate_monitor_setup(self):
...
                 )
+            if isinstance(monitor, EMEFieldMonitor):
+                if not np.array_equal(
+                    self.eme_grid_spec.virtual_cell_indices, self.eme_grid_spec.real_cell_indices
+                ):
+                    raise SetupError(
+                        f"Monitor at 'monitors[{i}]' is an 'EMEFieldMonitor', "
+                        "which is not compatible with periodic repetition "
+                        "('num_reps != 1' in any 'EMEGridSpec'.)"
+                    )
```
momchil-flex, line 796: Do we want to lift this restriction eventually?

tidy3d/components/eme/sweep.py
```diff
@@ -81,4 +83,46 @@ def num_sweep(self) -> pd.PositiveInt:
...
+    scattering matrices.
+
+    Compared to setting ``num_reps`` directly in the ``eme_grid_spec``,
+    this sweep spec allows varying the number of repetitions,
+    effectively simulating multiple structures in a single EME simulation.
+
+    Example
+    -------
+    >>> n_list = [1, 50, 100]
+    >>> sweep_spec = EMEPeriodicitySweep(num_reps=[{"unit_cell": n} for n in n_list])
```
momchil-flex, line 99: I'm not sure how I feel about using the `name` attribute for functional purposes. But this is not the first place in the code this happens. Just curious - is it used anywhere else in EME, and did you consider any alternatives?

momchil-flex (commented):
Thanks @caseyflex , just some small comments/questions

tomflexcompute:
Btw the current limit on the number of frequencies is still a little limited I find. Just wondering if the limit is due to cost consideration? The credit cost scales linearly with number of frequencies right?

caseyflex:
> Btw the current limit on the number of frequencies is still a little limited I find. Just wondering if the limit is due to cost consideration? The credit cost scales linearly with number of frequencies right?

@momchil-flex it was actually just based on simulations taking too long, I remember you weighed in on setting it to 20, shall we relax it to maybe 500?

tidy3d/components/eme/sweep.py
caseyflex: It’s only here. I considered forcing the sweep num_reps to be a list of tuples, each of which contained a num_reps for the current grid spec, as well as a list of num_reps for the sub grids if it is composite (as a recursive data structure in the same format as the grid spec). But this was way too cumbersome

tidy3d/components/eme/simulation.py
caseyflex: No, because the thing is the field may vary cell to cell on the repeated grid, so there is no way of capturing it on the original EME grid (before repetition)

tidy3d/components/eme/grid.py
caseyflex: Yeah, fair

momchil-flex:
> > Btw the current limit on the number of frequencies is still a little limited I find. Just wondering if the limit is due to cost consideration? The credit cost scales linearly with number of frequencies right?
> 
> @momchil-flex it was actually just based on simulations taking too long, I remember you weighed in on setting it to 20, shall we relax it to maybe 500?

Yeah I'm trying to remember now, not really sure but I think there was something to the effect that in many cases the EME time and cost would become (much?) larger than an FDTD run? So people might be surprised/unhappy. Do we want to be proactive about this @tomflexcompute or just let users do whatever they want? Maybe just a warning if more than e.g. 20 frequencies saying that EME time and cost scales with number of frequencies, while FDTD does not?

tomflexcompute:
I think it's much better to allow more frequency points and warn the users than not allowing it. Most people would probably do more than 20 frequency points. Maybe we can warn above 100 or something like that.

momchil-flex (approved):

PR #2332, opened 2025-03-21
`xarray>=2025.03` compatibility

yaugenst-flex:
Depends on https://github.com/flexcompute/tidy3d/pull/2330

Introduces a `tidy3d/compat.py` module where we can sort out these compatibility issues between packages in the future.


momchil-flex (approved):
Makes sense!

daquinteroflex (approved):

PR #2335, opened 2025-03-24
feat: add validation for units

rahul-flex:
Added setup errors for grid spacing `dl` and frequency (attached).

dl_error
freq_error

### DIFF
--- a/tidy3d/components/grid/grid_spec.py
+++ b/tidy3d/components/grid/grid_spec.py
@@ -276,6 +276,19 @@ class UniformGrid(GridSpec1d):
         units=MICROMETER,
     )
 
+    @pd.validator("dl", always=True)
+    def _validate_dl(cls, val):
+        """
+        Ensure 'dl' is not too small.
+        """
+        if val < 1e-7:
+            raise SetupError(
+                f"Uniform grid spacing 'dl' is {val} µm. "
+                "Please check your units! For more info on Tidy3D units, see: "
+                "https://docs.flexcompute.com/projects/tidy3d/en/latest/faq/docs/faq/What-are-the-units-used-in-the-simulation.html"
+            )
+        return val
+
     def _make_coords_initial(
         self,
         axis: Axis,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Validation check for unit error in grid spacing.
+
 ### Changed
 - Supplying autograd-traced values to geometric fields (`center`, `size`) of simulations, monitors, and sources now logs a warning and falls back to the static value instead of erroring.
 - Attempting to differentiate server-side field projections now raises a clear error instead of silently failing.
--- a/tests/test_components/test_grid_spec.py
+++ b/tests/test_components/test_grid_spec.py
@@ -1,6 +1,7 @@
 """Tests GridSpec."""
 
 import numpy as np
+import pydantic.v1 as pydantic
 import pytest
 import tidy3d as td
 from tidy3d.exceptions import SetupError
@@ -515,3 +516,28 @@ def test_domain_mismatch():
         boundary_spec=td.BoundarySpec.pml(),
     )
     z = sim.grid.boundaries.z
+
+
+@pytest.mark.parametrize(
+    ("dl", "expect_exception"),
+    [
+        (1e-8, True),  # Below 1e-7 => fail
+        (1e-7, False),  # Exactly at lower bound => pass
+        (0.0, True),  # Zero => fail
+    ],
+)
+def test_uniform_grid_dl_validation(dl, expect_exception):
+    """Test the validator that checks 'dl' is between 1e-7 and 3e8 µm."""
+    if expect_exception:
+        with pytest.raises(pydantic.ValidationError):
+            _ = td.Simulation(
+                size=(1, 1, 1),
+                grid_spec=td.GridSpec.uniform(dl=dl),
+                run_time=1e-12,
+            )
+    else:
+        _ = td.Simulation(
+            size=(1, 1, 1),
+            grid_spec=td.GridSpec.uniform(dl=dl),
+            run_time=1e-12,
+        )


yaugenst-flex:
We encountered users specifying the wrong units in their simulation setup, so the motivation here is to introduce range-based validators and link to the relevant docs page to make sure users can quickly identify and correct unit issues. While the validators impose strict bounds, the idea is not to necessarily allow those values but simply catch setups that are clearly off by orders of magnitude. We had a stab at selecting bounds that would cover that but if you could double check @momchil-flex and @tomflexcompute?

momchil-flex:
I think it makes sense, the main concern is RF freqs, it seems like the warning will currently be raised for a wavelength larger than 1 meter, which is probably fine for all applications we're currently targeting? But may be close. But probably for a center frequency it's fine.

I remember we added a similar validator at some point, maybe only for the mode solver... But I can't find it now. @weiliangjin2021 do you remember?

tidy3d/components/grid/grid_spec.py
```diff
@@ -276,6 +276,20 @@ class UniformGrid(GridSpec1d):
         units=MICROMETER,
     )
 
+    @pd.validator("dl", always=True)
+    def _validate_dl(cls, val):
+        """
+        Ensure 'dl' is within a physically sensible range to catch common
+        unit mistakes.
+        """
```
yaugenst-flex: This fits on a single line, can remove everything after `sensible range`.

tidy3d/components/grid/grid_spec.py
```diff
@@ -276,6 +276,20 @@ class UniformGrid(GridSpec1d):
...
+        if not (1e-6 <= val <= 5e4):
```
yaugenst-flex: Probably `1e4` is fine here? Not that it matters much.

tidy3d/components/grid/grid_spec.py
```diff
@@ -276,6 +276,20 @@ class UniformGrid(GridSpec1d):
...
+            raise SetupError(
+                f"Uniform grid spacing 'dl' is {val} µm."
```
yaugenst-flex: Space missing here: `µm."` -> `µm. "`

tidy3d/components/source/time.py
```diff
@@ -95,6 +95,23 @@ class Pulse(SourceTime, ABC):
...
     )
 
+    @pydantic.validator("freq0", always=True)
+    def _validate_freq0(cls, val):
+        """
+        Ensure 'freq0' is within a physically sensible range to catch
+        common unit mistakes.
+        """
+        min_freq = C_0 / (1e6)
+        max_freq = C_0 / (1e-6)
```
yaugenst-flex: Can remove the parantheses here

tidy3d/components/source/time.py
```diff
@@ -95,6 +95,23 @@ class Pulse(SourceTime, ABC):
         ge=2.5,
     )
 
+    @pydantic.validator("freq0", always=True)
+    def _validate_freq0(cls, val):
+        """
+        Ensure 'freq0' is within a physically sensible range to catch
+        common unit mistakes.
+        """
```
yaugenst-flex: Same comment as above, I'd put this on a single line.

yaugenst-flex (changes requested):
Thanks @rahul-flex this looks good! Left some minor comments. Make sure to also add tests for these validators and a changelog entry (under the "Changed" section probably).

yaugenst-flex:
> it seems like the warning will currently be raised for a wavelength larger than 1 meter, which is probably fine for all applications we're currently targeting? But may be close.

Hmm I guess that also means we might want `dl` > 1cm? @momchil-flex

tomflexcompute (commented):
This is nice but I think 1m wavelength is actually a bit too restrictive for some use cases like radar and long range communication, even though these use cases are not common within our current user base.

yaugenst-flex:
@tomflexcompute I guess that would put us in the ~100m wavelength range?

tomflexcompute:
> @tomflexcompute I guess that would put us in the ~100m wavelength range?

yeah potentially. It's probably fine for now but in the future if we have more users from RF and microwave it might be a problem.

weiliangjin2021:
We have a validator [here](https://github.com/flexcompute/tidy3d/blob/14f326d2201488376b24a81fbcb720bf59c74d02/tidy3d/components/validators.py#L417-L435) checking the minimal frequency. The lowest frequency is 1e5 Hz ([here](https://github.com/flexcompute/tidy3d/blob/14f326d2201488376b24a81fbcb720bf59c74d02/tidy3d/components/validators.py#L46-L47)).

It's applied to FreqMonitor [here](https://github.com/flexcompute/tidy3d/blob/14f326d2201488376b24a81fbcb720bf59c74d02/tidy3d/components/monitor.py#L103), and source [here](https://github.com/flexcompute/tidy3d/blob/14f326d2201488376b24a81fbcb720bf59c74d02/tidy3d/components/source/base.py#L61-L65).

rahul-flex:
Thanks for the comments @yaugenst-flex @momchil-flex @tomflexcompute @weiliangjin2021. Should we make upper bound of dl > 1cm (say 10 cm?) and lower bound of frequency:  1e5 Hz?

weiliangjin2021:
> Should we make upper bound of dl > 1cm (say 10 cm?) and lower bound of frequency: 1e5 Hz?

I wonder if we really need a validator for grid size. If the grid size is too large, you'll get just one grid in that dimension, and it should be easy to spot the issue? And if the grid size is too small, I think there is already a validator that checks the total number of grid points below a threshold value.

Now if we go for a validator for grid size, with lower bound frequency 1e5 Hz, corresponding to wavelength 300m, the upper bound of grid size will be 30 m.

rahul-flex:
> And if the grid size is too small, I think there is already a validator that checks the total number of grid points below a threshold value.

@weiliangjin2021 Thanks for the bound values. 
I’m not aware of an existing validator for grid size. Currently, the error for fine grid
seems to be due to high memory allocation, as reported by a user ( screenshot attached).
Screenshot 2025-03-25 at 7 32 44 PM

weiliangjin2021:
> > And if the grid size is too small, I think there is already a validator that checks the total number of grid points below a threshold value.
> 
> @weiliangjin2021 Thanks for the bound values. I’m not aware of an existing validator for grid size. Currently, the error for fine grid seems to be due to high memory allocation, as reported by a user ( screenshot attached). Screenshot 2025-03-25 at 7 32 44 PM

See https://github.com/flexcompute/tidy3d/blob/14f326d2201488376b24a81fbcb720bf59c74d02/tidy3d/components/simulation.py#L3701-L3709. But it seems it errorred already before reaching here.

rahul-flex:
> . But it seems it errorred already before reaching here.

Ohh, I see! Thanks!

tidy3d/components/source/time.py
```diff
@@ -95,6 +95,22 @@ class Pulse(SourceTime, ABC):
...
+        max_freq = C_0 / 1e-6
+        if not (min_freq <= val <= max_freq):
+            raise SetupError(
+                f"Pulse central frequency 'freq0' is {val:.3e} Hz, which corresponds "
+                f"to a free-space wavelength of {C_0 / val:.3e} µm. "
+                "Check your units! For more info on Tidy3D units, see: "
+                "https://docs.flexcompute.com/projects/tidy3d/en/latest/faq/docs/faq/What-are-the-units-used-in-the-simulation.html"
+            )
+        return val
+
```
yaugenst-flex: As @weiliangjin2021 pointed out, there already is a frequency validator which we could reuse. Would it make sense to also add an upper frequency bound to that validator?

yaugenst-flex:
> > > And if the grid size is too small, I think there is already a validator that checks the total number of grid points below a threshold value.
> > 
> > 
> > @weiliangjin2021 Thanks for the bound values. I’m not aware of an existing validator for grid size. Currently, the error for fine grid seems to be due to high memory allocation, as reported by a user ( screenshot attached). Screenshot 2025-03-25 at 7 32 44 PM
> 
> See
> 
> https://github.com/flexcompute/tidy3d/blob/14f326d2201488376b24a81fbcb720bf59c74d02/tidy3d/components/simulation.py#L3701-L3709
> 
> . But it seems it errorred already before reaching here.

I think the size validator is not quite the same though, or at least if there is an opportunity to catch a bad `dl` earlier that'd still be beneficial. Actually we could also just change the `dl` validator have no upper but only a lower bound.

yaugenst-flex (approved):
Thanks @rahul-flex 

@weiliangjin2021 fyi we ended up going conservative and only check for tiny grid spacing now. This check would've already caught the error the user that prompted this PR was seeing.

weiliangjin2021:
> @weiliangjin2021 fyi we ended up going conservative and only check for tiny grid spacing now. This check would've already caught the error the user that prompted this PR was seeing.

Looks nice! A minor: "if not (1e-7 <= val)"  --> "if val <1e-7" would be easier to read

rahul-flex:
Thanks @weiliangjin2021 and @yaugenst-flex for the comments. Simplified the condition.

yaugenst-flex:
Thanks @rahul-flex looks good to go but could you please squash this into a single commit before we merge and make sure the changelog entry is in the right place. currently it's listed as a changelog item under 2.8.2 (should go under "unreleased")

momchil-flex:
@yaugenst-flex @rahul-flex a user just had some simulations erroring because they had obviously set units wrong, which resulted in a grid spacing on the scale of 1e-9. However, this was still an auto grid (they had also set source frequency wrong), so it was not caught by this validator. A large number of our users probably use auto grid instead of uniform grid - it would be good to have a similar validator there. However, in that case it should be a Simulation `post_init` validator, as the grid needs to be constructed to examine the smallest dl.

PR #2336, opened 2025-03-24
Structure priority option in structure overlapping region

weiliangjin2021:
In many RF smulations, metallic structures are usually very thin: thickness much smaller than grid size. We can still accurately simulate them by applying PEC conformal meshing and SIBC techniques. However, since we slightly expand the structures in the solver (not really now, but a similar strategy), when the metallic structures are not added later compared to the touching dielectric structures in the structure list, those metallic structures will be completely missing in Yee-grid representation. It is a very common issues when people setup an RF simulation, even among our internal developers.

As discussed [here](https://github.com/flexcompute/tidy3d-core/issues/943), the scope of this PR:

- [x] Add a priority field to `AbstractStructure`. The default value is `None`.
- [x] Add a field `structure_priority_mode` to `Scene`. It sets priority of structures that have `priority=None`. Right now, there are two modes:
  - [x] `equal` that sets `priority=0`;
  - [x] `conductor` that sets `priority=90` for lossyMetaMedium, and 100 for PEC, and 0 for others
- [x] A cached_property `sorted_structures` in Scene to be applied in places where `sim.structures` was used
- [x] MeshOverideStructures will have default priority=0, regardless of the value of  `structure_priority_mode`.
- [x] In meshing, `structures = sorted_structure + sorted_mesh_override_structures`; internal mesh override structures have priority=-1
- [x] Priority of 2d medium: no different from the bulk.
- [x] Add tests

### DIFF
--- a/tidy3d/components/scene.py
+++ b/tidy3d/components/scene.py
@@ -54,6 +54,7 @@ from .types import (
     InterpMethod,
     LengthUnit,
     PermittivityComponent,
+    PriorityMode,
     Shapely,
     Size,
 )
@@ -106,8 +107,20 @@ class Scene(Tidy3dBaseModel):
         (),
         title="Structures",
         description="Tuple of structures present in scene. "
-        "Note: Structures defined later in this list override the "
-        "simulation material properties in regions of spatial overlap.",
+        "Note: In regions of spatial overlap between structures, "
+        "material properties are dictated by structure of higher priority. "
+        "The priority for structure of `priority=None` is set automatically "
+        "based on `structure_priority_mode`. For structures of equal priority, "
+        "the structure added later to the structure list takes precedence.",
+    )
+
+    structure_priority_mode: PriorityMode = pd.Field(
+        "equal",
+        title="Structure Priority Setting",
+        description="This field only affects structures of `priority=None`. "
+        "If `equal`, the priority of those structures is set to 0; if `conductor`, "
+        "the priority of structures made of `LossyMetalMedium` is set to 90, "
+        "`PECMedium` to 100, and others to 0.",
     )
 
     plot_length_units: Optional[LengthUnit] = pd.Field(
@@ -246,6 +259,17 @@ class Scene(Tidy3dBaseModel):
         return {medium: index for index, medium in enumerate(self.mediums)}
 
     @cached_property
+    def sorted_structures(self) -> List[Structure]:
+        """Returns a list of sorted structures based on their priority.In the sorted list,
+        latter added structures take higher priority.
+
+        Returns
+        -------
+        List[:class:`.Structure`]
+        """
+        return Structure._sort_structures(self.structures, self.structure_priority_mode)
+
+    @cached_property
     def background_structure(self) -> Structure:
         """Returns structure representing the background of the :class:`.Scene`."""
         geometry = Box(size=(inf, inf, inf))
@@ -254,7 +278,7 @@ class Scene(Tidy3dBaseModel):
     @cached_property
     def all_structures(self) -> List[Structure]:
         """List of all structures in the simulation including the background."""
-        return [self.background_structure] + list(self.structures)
+        return [self.background_structure] + self.sorted_structures
 
     @staticmethod
     def intersecting_media(
@@ -443,7 +467,7 @@ class Scene(Tidy3dBaseModel):
         """
 
         medium_shapes = self._get_structures_2dbox(
-            structures=self.to_static().structures, x=x, y=y, z=z, hlim=hlim, vlim=vlim
+            structures=self.to_static().sorted_structures, x=x, y=y, z=z, hlim=hlim, vlim=vlim
         )
         medium_map = self.medium_map
         for medium, shape in medium_shapes:
@@ -889,7 +913,7 @@ class Scene(Tidy3dBaseModel):
             The supplied or created matplotlib axes.
         """
 
-        structures = self.structures
+        structures = self.sorted_structures
 
         # alpha is None just means plot without any transparency
         if alpha is None:
@@ -1466,7 +1490,7 @@ class Scene(Tidy3dBaseModel):
             The supplied or created matplotlib axes.
         """
 
-        structures = self.structures
+        structures = self.sorted_structures
 
         # alpha is None just means plot without any transparency
         if alpha is None:
@@ -1744,7 +1768,7 @@ class Scene(Tidy3dBaseModel):
         """
 
         scene_dict = self.dict()
-        structures = self.structures
+        structures = self.sorted_structures
         array_dict = {
             "temperature": temperature,
             "electron_density": electron_density,
@@ -1795,7 +1819,7 @@ class Scene(Tidy3dBaseModel):
         acceptors_lims = [1e50, -1e50]
         donors_lims = [1e50, -1e50]
 
-        for struct in [self.background_structure] + list(self.structures):
+        for struct in self.all_structures:
             if isinstance(struct.medium.charge, SemiconductorMedium):
                 electric_spec = struct.medium.charge
                 for doping, limits in zip(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `eps_lim` keyword argument to `Simulation.plot_eps()` for manual control over the permittivity color limits.
 - Added `thickness` parameter to `LossyMetalMedium` for computing surface impedance of a thin conductor.
+- `priority` field in `Structure` and `MeshOverrideStructure` for setting the behavior in structure overlapping region. When its value is `None`, the priority is automatically determined based on the material property and simulation's `structure_priority_mode`.
 
 ### Changed
 - Relaxed bounds checking of path integrals during `WavePort` validation.
--- a/tests/test_components/test_scene.py
+++ b/tests/test_components/test_scene.py
@@ -307,3 +307,48 @@ def test_max_geometry_validation():
     ]
     with pytest.raises(pd.ValidationError, match=f" {MAX_GEOMETRY_COUNT + 2} "):
         _ = td.Scene(structures=not_fine)
+
+
+def test_structure_manual_priority():
+    """make sure structure is properly orderd based on the priority settings."""
+
+    box = td.Structure(
+        geometry=td.Box(size=(1, 1, 1), center=(0, 0, 0)),
+        medium=td.Medium(permittivity=2.0),
+    )
+    structures = []
+    priorities = [2, 4, -1, -4, 0]
+    for priority in priorities:
+        structures.append(box.updated_copy(priority=priority))
+    scene = td.Scene(
+        structures=structures,
+    )
+
+    sorted_priorities = [s.priority for s in scene.sorted_structures]
+    assert all(np.diff(sorted_priorities) >= 0)
+
+
+def test_structure_automatic_priority():
+    """make sure metallic structure has the highest priority in `conductor` mode."""
+
+    box = td.Structure(
+        geometry=td.Box(size=(1, 1, 1), center=(0, 0, 0)),
+        medium=td.Medium(permittivity=2.0),
+    )
+    box_pec = box.updated_copy(medium=td.PEC)
+    box_lossymetal = box.updated_copy(
+        medium=td.LossyMetalMedium(conductivity=1.0, frequency_range=(1e14, 2e14))
+    )
+    structures = [box_pec, box_lossymetal, box]
+    scene = td.Scene(
+        structures=structures,
+        structure_priority_mode="equal",
+    )
+
+    # in equal mode, the order is preserved
+    scene.sorted_structures == structures
+
+    # conductor mode
+    scene = scene.updated_copy(structure_priority_mode="conductor")
+    assert scene.sorted_structures[-1].medium == td.PEC
+    assert isinstance(scene.sorted_structures[-2].medium, td.LossyMetalMedium)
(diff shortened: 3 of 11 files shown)


yaugenst-flex:
Haven't had time to review properly yet but a quick thought: Maybe it would make sense if we introduced a numbers-based priority to structures in general? By default they'd all have the same, and then priority is decided on ordering, but giving a larger number would mean higher priority. This is a pretty common approach I think, for example matplotlib does it for `zorder`. When dealing with metals, these could get a higher priority by default for example? This seems like it would be a useful feature to have in general. Hopefully it would also obviate the need for a `sim.sorted_structures`?

weiliangjin2021:
> Haven't had time to review properly yet but a quick thought: Maybe it would make sense if we introduced a numbers-based priority to structures in general? By default they'd all have the same, and then priority is decided on ordering, but giving a larger number would mean higher priority. This is a pretty common approach I think, for example matplotlib does it for `zorder`. When dealing with metals, these could get a higher priority by default for example? This seems like it would be a useful feature to have in general. Hopefully it would also obviate the need for a `sim.sorted_structures`?

That's actually my initial plan. But the concern is that this feature will barely be used, and people will only use the automated priority approach. @tomflexcompute also brought up this when he is using a certain photonic simulation software.

weiliangjin2021:
> When dealing with metals, these could get a higher priority by default for example? This seems like it would be a useful feature to have in general. Hopefully it would also obviate the need for a sim.sorted_structures

For backward-compatibility, we cannot by default assign higher priority to metal. So we will still need an additional field like the one added in this PR that sets the automation behavior: by structure list order or material property. To that regard, I think we can just have this additional field for automation for now, and adds the manual priority number when requested in the future.

Now regardless of the approach, I think we'll still need `.sorted_structures` internally. Otherwise, there will be even more changes to the code: significant changes in plotting, mesher, solver processing.

tidy3d/components/scene.py
```diff
@@ -107,8 +110,19 @@ class Scene(Tidy3dBaseModel):
         (),
         title="Structures",
         description="Tuple of structures present in scene. "
-        "Note: Structures defined later in this list override the "
-        "simulation material properties in regions of spatial overlap.",
+        "Note: In regions of spatial overlapping between structures, "
```
momchil-flex: I think either "of spatial overlap" or "with spatial overlapping" sounds better

tidy3d/components/scene.py
```diff
@@ -107,8 +110,19 @@ class Scene(Tidy3dBaseModel):
...
+        "material properties are dictated by structure of higher priority "
+        "defined by ``overlapping_priority``.",
+    )
+
+    overlapping_priority: OverlappingPriority = pd.Field(
+        "latter",
+        title="Resolution Strategy In Structure Overlapping Region",
+        description="In regions of spatial overlapping between structures, "
+        "if ``overlapping_priority=latter``, structures defined later in the list haver higher "
+        "priority. If ``overlapping_priority=metal``, metallic structures of higher conductivity "
```
momchil-flex: By `metallic` we explicitly mean PEC or SIBC but not e.g. usual dispersive metals?

tidy3d/components/scene.py
```diff
@@ -246,6 +260,36 @@ def medium_map(self) -> Dict[StructureMediumType, pd.NonNegativeInt]:
...
+            # so that it's lower than that of any metal.
+            sigmas = [-1, -1]
+            for ind, mat in enumerate(mat_list):
+                if mat.is_pec:
+                    sigmas[ind] = LARGE_NUMBER
+                elif isinstance(mat, LossyMetalMedium):
+                    sigmas[ind] = mat.conductivity
+            return sigmas[0] - sigmas[1]
+
+        return sorted(self.structures, key=cmp_to_key(structure_comparator))
```
momchil-flex: There was some rearrangement done for 2D media too, right? I think it was to put the equivalent volumetric structures on top. I wonder how this plays with these changes, and if it's at all possible to unify these rearrangements more - also to clarify even better to the user what exactly happens in various cases (metals + 2D media? very strange case I guess but still)

momchil-flex (commented):
Yeah I'm not sure honestly, I see the advantage of the current approach in that it's easier for the user to set, but it does feel like we'll just keep running into the need to have a priority setting in the structures. It seems like it could already make at least the internal code a bit better, and then if we have it internally, we might expose it anyway.

If we were to do that, how might the `metallic` option combine with it?

weiliangjin2021:
> but it does feel like we'll just keep running into the need to have a priority setting in the structures.

Could you give more contexts on what other occasions that requires a manual priority setting?

weiliangjin2021:
> it seems like it could already make at least the internal code a bit better, and then if we have it internally, we might expose it anyway.

What's your thought on improving the internal codes with manual priority setting? I think we'll still need the `sorted_structures` to be used anywhere that requires the structure lists so that we don't sort it again in those places?

> If we were to do that, how might the metallic option combine with it?

Yeah, that's my question too. In this approach, I think the priority field in structure will be `Union[float, "auto"]`, and we set the default as `auto`. Then we'll still need an additional field in simulation class to indicate the behavior of `auto`.

tidy3d/components/scene.py
weiliangjin2021: we don't have any special subpixel methods applied to usual dispersive metals. So we only mean "good conductor" by "metallic" here.

momchil-flex:
> > but it does feel like we'll just keep running into the need to have a priority setting in the structures.
> 
> Could you give more contexts on what other occasions that requires a manual priority setting?

I'm actually not sure off the top of my head but both Lumerical and CST Studio support this which makes me think that there are use cases that we haven't yet thought of.

Btw note [what CST does](https://space.mit.edu/RADIO/CST_online/mergedProjects/VBA_3D/common_vbasolido/common_vbasolido_solid_object.htm), which could also be one possibility for us

```
SetAutomeshPriority 

This method specifies how the specified solid should be treated by the automatic mesh generation. If the solid is more important than others, a priority can be given for it. Generally the priority for all objects (apart from wires, lumped elements, discrete ports) equals to zero. In the case that two fixpoints are so close to each other that the ratiolimit prohibits a mesh line for each point, the mesh lines will be merged. However if one of the fixpoints has been created by an object of higher priority the mesh lines will be placed on this fixpoint.

Because wires, lumped elements and discrete ports are very sensitive to their start and endpoints, they have a priority of 1000. Otherwise it would be possible that they get disconnected during the mesh generation.
```

I was thinking along these lines too when reviewing your PR, instead of introducing a literal "metal" setting, if we were to go with the priority instead, structures with metal media can come with a high default value.

yaugenst-flex:
I think the priority should just be an `int`, with some number, e.g. `0`, by default for everything. This would be identical to the current behavior and backward-compatible. I guess one use case is simply ergonomics. Sometimes it's nice to not have to worry about the order in which you put your structures into a list. Another one that I can think of is when we are inserting new structures into simulations. In inverse design, we often create a base simulation into which we then insert structures. You might want to insert a structure that is not at the very end of the list but insert it such that it covers some part of the base simulation but gets clipped by another part. That's something that would probably be easier to do with these priorities.

tidy3d/components/scene.py
weiliangjin2021: Now it's renamed to conductor

weiliangjin2021:
@momchil-flex One question on meshing: in this part, we have both physical structures and mesh override structures. There are two directions for us to handle this:
- We don't distinguish between the two structures, and sort them all together;
- We sort physical structures and mesh override structures separately, and structure list will be `sorted physical structures + sorted override structures`. So the override structures take higher precedence.

Personally, I prefer the 2nd approach. For the 1st approach, since for now the default priority of override structures is 0, and metallic physical structures have higher priority, override structures might never take effect.

weiliangjin2021:
Another question to @dmarek-flex and @caseyflex: in the `conductor` mode, I plan to set priority=90 for lossyMetaMedium, 100 for PEC, and 0 for others. For Medium2D, and lumped elements that generate Medium2D, does it make more sense to set highest priority for Medium2D, like 110? In that case, those lumped elements won't be overridden by the bulk metal parts.

momchil-flex:
> @momchil-flex One question on meshing: in this part, we have both physical structures and mesh override structures. There are two directions for us to handle this:
> 
> * We don't distinguish between the two structures, and sort them all together;
> * We sort physical structures and mesh override structures separately, and structure list will be `sorted physical structures + sorted override structures`. So the override structures take higher precedence.
> 
> Personally, I prefer the 2nd approach. For the 1st approach, since for now the default priority of override structures is 0, and metallic physical structures have higher priority, override structures might never take effect.

2nd approach makes sense to me too. I think for some very esoteric situations where someone wants to have a regular structure above a meshing override one, they can just also add the regular structures in the override structures list.

caseyflex:
> Another question to @dmarek-flex and @caseyflex: in the `conductor` mode, I plan to set priority=90 for lossyMetaMedium, 100 for PEC, and 0 for others. For Medium2D, and lumped elements that generate Medium2D, does it make more sense to set highest priority for Medium2D, like 110? In that case, those lumped elements won't be overridden by the bulk metal parts.

It looks to me like currently, when a Medium2D is encountered, it stays where it is in the list of structures. This choice was made to allow defining structures like in this notebook:

https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/Bandstructure.html

where you place a uniform medium (which could be a Medium2D), and then put air holes into it

dmarek-flex:
> Another question to @dmarek-flex and @caseyflex: in the `conductor` mode, I plan to set priority=90 for lossyMetaMedium, 100 for PEC, and 0 for others. For Medium2D, and lumped elements that generate Medium2D, does it make more sense to set highest priority for Medium2D, like 110? In that case, those lumped elements won't be overridden by the bulk metal parts.

Good question, current behavior of lumped elements is that they are always appended to the `all_structures` list during  `volumetric_structures`, so none of your changes should change that. I also think that it is ideal to prioritize lumped elements in this way.

For general `Medium2D`, I think they should be prioritized in the same manner as regular `Medium` are in this PR. So giving priority to `PEC`/`LossyMetal`, if desired by the user. But otherwise they maintain their order based on when they are added to the structures list.

weiliangjin2021:
> Good question, current behavior of lumped elements is that they are always appended to the `all_structures` list during `volumetric_structures`, so none of your changes should change that. I also think that it is ideal to prioritize lumped elements in this way.

In `conductor` mode, the position of lumped elements in `all_structures` list doesn't guarantee higher priority, because internally we will sort the structure list based on the priority. If we don't assign higher priority to them, they can be overridden by PEC etc.. So perhaps I'll assign them the same priority as PEC? When structures are of the same priority, the latter ones take precedence.

> For general `Medium2D`, I think they should be prioritized in the same manner as regular `Medium` are in this PR. So giving priority to `PEC`/`LossyMetal`, if desired by the user. But otherwise they maintain their order based on when they are added to the structures list.

OK, we will not assign them special priority. Their priority will only be boosted, if they are 2D PEC.

weiliangjin2021:
Should be ready for review now

momchil-flex:
@weiliangjin2021 we should probably keep this for 2.9 too, what do you think?

Edit: actually I guess this is still not expected to produce any different results if the default is used, is it?

weiliangjin2021:
> @weiliangjin2021 we should probably keep this for 2.9 too, what do you think?
> 
> Edit: actually I guess this is still not expected to produce any different results if the default is used, is it?

That's right

tidy3d/components/scene.py
```diff
@@ -253,7 +277,7 @@ def background_structure(self) -> Structure:
     @cached_property
     def all_structures(self) -> List[Structure]:
         """List of all structures in the simulation including the background."""
-        return [self.background_structure] + list(self.structures)
+        return [self.background_structure] + list(self.sorted_structures)
```
yaugenst-flex: Since `self.sorted_structures` is a `list` I'd remove the cast here.

tidy3d/components/scene.py
```diff
@@ -1788,7 +1812,7 @@ def doping_bounds(self):
         acceptors_lims = [1e50, -1e50]
         donors_lims = [1e50, -1e50]
 
-        for struct in [self.background_structure] + list(self.structures):
+        for struct in [self.background_structure] + list(self.sorted_structures):
```
yaugenst-flex: Should this be `self.all_structures`?

yaugenst-flex (dismissed):
Thanks @weiliangjin2021 LGTM! Only some nitpicks.

tidy3d/components/scene.py
weiliangjin2021: I think so. I think @marc-flex originally wrote `for struct in [self.background_structure] + list(self.structures)`, and I simply replace `.structures` with `.sorted_structures`. @marc-flex: should be equivalent to `for struct in self.all_structures`?

marc-flex: I think so? The order here doesn't affect the visibility, correct?

weiliangjin2021: The order matters when two structures overlap. Here it doesn't seem to matter, as it's just computing the limits, and doesn't deal with structure overriding.

marc-flex: Right. Then it should be OK :)

tidy3d/components/scene.py
weiliangjin2021: In this PR and also previously, 2D media is not handled differently from 3D medium. They stay the same order / priority.

caseyflex: Only PEC is reordered? Is this clear to the user?

weiliangjin2021: if one sets `structure_priority_mode='conductor'`, 2D PEC is reordered. That is expected behavior for RF users.

caseyflex: Ah right. Makes sense then

yaugenst-flex (approved):
Looks good from my side!

PR #2337, opened 2025-03-25
Add energy bandgap monitors for the CHARGE simulations.

vasilyzabelin:
Add energy bandgap monitors for the CHARGE simulations.

### DIFF
--- a/tidy3d/components/tcad/data/monitor_data/charge.py
+++ b/tidy3d/components/tcad/data/monitor_data/charge.py
@@ -17,10 +17,13 @@ from tidy3d.components.data.utils import TetrahedralGridDataset, TriangularGridD
 from tidy3d.components.tcad.data.monitor_data.abstract import HeatChargeMonitorData
 from tidy3d.components.tcad.monitors.charge import (
     SteadyCapacitanceMonitor,
+    SteadyEnergyBandMonitor,
     SteadyFreeCarrierMonitor,
     SteadyPotentialMonitor,
 )
-from tidy3d.components.types import TYPE_TAG_STR, annotate_type
+from tidy3d.components.types import TYPE_TAG_STR, Ax, Axis, annotate_type
+from tidy3d.components.viz import add_ax_if_none
+from tidy3d.exceptions import DataError
 from tidy3d.log import log
 
 FieldDataset = Union[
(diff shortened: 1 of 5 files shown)


tidy3d/components/tcad/monitors/charge.py
```diff
@@ -40,6 +40,26 @@ class SteadyFreeCarrierMonitor(HeatChargeMonitor):
     )
 
 
+class SteadyEnergyBandMonitor(HeatChargeMonitor):
+    """
+    Free-carrier monitor for Charge simulations.
```
momchil-flex, line 45: Free-carrier -> Energy band

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -172,6 +175,209 @@ def field_name(self, val: str = "") -> str:
...
+            Ec=new_Ec,
+            Ev=new_Ev,
+            Ei=new_Ei,
+            Efn=new_Efn,
+            Efp=new_Efp,
+            symmetry=(0, 0, 0),
+        )
+
+    def field_name(self, val: str = "") -> str:
+        """Gets the name of the fields to be plot."""
```
momchil-flex, line 294: `plotted`

I see that this is just copied from above, so could you correct it everywhere?

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -172,6 +175,209 @@ def field_name(self, val: str = "") -> str:
...
+        if val == "abs^2":
+            return "|Ec|², |Ev|², |Ei|², |Efn|², |Efp|²"
+        else:
+            return "Ec, Ev, Ei, Efn, Efp"
+
+    @add_ax_if_none
+    def plot_slice(self, axis: Axis, pos: float, voltage: float, ax: Ax = None) -> Ax:
```
momchil-flex, line 301: As a general comment (@marc-flex ) in the EM data `plot_field` we do not explicitly pass things like frequency, mode inde, etc. We just bundle all unknown arguments into [**sel_kwargs](https://github.com/flexcompute/tidy3d/blob/954a43aa73f514d714f937bc4c6316e8344eea3e/tidy3d/components/mode/data/sim_data.py#L54) and pass this to `.sel`, which allows a bit more freedom. So here, `voltage` would go into these not-explicitly-defined arguments. Then someone could even do `plot_slice(... , voltage=1, method="nearest")` to avoid errors that might sometimes happen if they pass a value that is slightly off from an actual recorded value (sometimes just due to numerical precision...)

momchil-flex:
Thanks! Should also add a test somewhere around [here](https://github.com/flexcompute/tidy3d/blob/954a43aa73f514d714f937bc4c6316e8344eea3e/tests/test_components/test_heat_charge.py#L563-L582)

vasilyzabelin:
Updated the code and created the new PR #2344

marc-flex:
I think we can close this PR, correct?

momchil-flex:
Pretty sure that yes.

PR #2338, opened 2025-03-26
Correct scaling for adjoint sources created from field monitors

groberts-flex:
Corrects the normalization for field sources created when differentiating through field monitors. Here, depending on the size of the source, we normalize by the volume, area, or linear element associated with the mesh the source is created on. Originally, I thought we would have to pass in more information about the mesh surrounding the monitor, but we can use the coordinates already inside of the monitor data for the normalization.

An associated numerical test is also added that does not get run unless the --run-numerical-adjoint flag is passed into pytest.

I'm verifying this one more time and will post a plot of the finite difference vs. adjoint gradient with this change after that runs!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Compatibility with `xarray>=2025.03`.
 - Inaccurate gradient when auto-grabbing permittivities for structures using `td.PolySlab` when using dispersive material models.
+- Fixed scaling for adjoint sources when differentiating with respect to `FieldData` to account for the mesh size of the monitor and thus the created source. This aligns adjoint gradient magnitudes with numerical finite difference gradients for field data.
 
 ## [2.8.1] - 2025-03-20
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,8 @@
 import os
+from pathlib import Path
 
 import autograd
+import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
 import psutil
@@ -68,3 +70,33 @@ def pytest_xdist_auto_num_workers(config):
     if os.getenv("GITHUB_ACTIONS"):
         return cores
     return max(1, cores - 1)
+
+
+@pytest.fixture
+def mpl_config_noninteractive():
+    """Configure matplotlib non-interactive backend for all tests in this module."""
+    original_backend = mpl.get_backend()
+    mpl.use("Agg")
+    yield
+    plt.close("all")
+    mpl.use(original_backend)
+
+
+@pytest.fixture
+def mpl_config_interactive():
+    """Configure matplotlib interactive backend for all tests in this module."""
+    original_backend = mpl.get_backend()
+    mpl.use("TkAgg")
+    yield
+    mpl.use(original_backend)
+
+
+@pytest.fixture
+def dir_name(request):
+    return request.param
+
+
+@pytest.fixture
+def create_directory(dir_name):
+    if dir_name is not None:
+        directory = Path(dir_name).mkdir(parents=True, exist_ok=True)
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -13,7 +13,7 @@ import xarray as xr
 from pandas import DataFrame
 from xarray.core.types import Self
 
-from ...constants import C_0, ETA_0, MICROMETER, MU_0
+from ...constants import C_0, EPSILON_0, ETA_0, MICROMETER
 from ...exceptions import DataError, SetupError, Tidy3dNotImplementedError, ValidationError
 from ...log import log
 from ..base import TYPE_TAG_STR, cached_property, skip_if_fields_missing
@@ -1158,53 +1158,8 @@ class FieldData(FieldDataset, ElectromagneticFieldData):
 
     def make_adjoint_sources(
         self, dataset_names: list[str], fwidth: float
-    ) -> List[Union[CustomCurrentSource, PointDipole]]:
+    ) -> List[CustomCurrentSource]:
         """Converts a :class:`.FieldData` to a list of adjoint current or point sources."""
-
-        if np.allclose(self.monitor.size, 0):
-            return self.to_adjoint_point_sources(fwidth=fwidth)
-
-        return self.to_adjoint_field_sources(fwidth=fwidth)
-
-    def to_adjoint_point_sources(self, fwidth: float) -> List[PointDipole]:
-        """Create adjoint point dipole source if this field data contains one item."""
-
-        sources = []
-
-        for polarization, field_component in self.field_components.items():
-            if field_component is None:
-                continue
-
-            for freq0 in field_component.coords["f"]:
-                omega0 = 2 * np.pi * freq0
-                scaling_factor = 33 / (MU_0 * omega0)
-
-                forward_amp = self.get_amplitude(field_component.sel(f=freq0))
-
-                if forward_amp == 0.0:
-                    continue
-
-                adj_phase = np.pi + np.angle(forward_amp)
-                adj_amp = scaling_factor * forward_amp
-
-                src_adj = PointDipole(
-                    center=self.monitor.center,
-                    polarization=polarization,
-                    source_time=GaussianPulse(
-                        freq0=freq0,
-                        fwidth=fwidth,
-                        amplitude=abs(adj_amp),
-                        phase=adj_phase,
-                    ),
-                    interpolate=True,
-                )
-
-                sources.append(src_adj)
-
-        return sources
-
-    def to_adjoint_field_sources(self, fwidth: float) -> List[CustomCurrentSource]:
-        """Create adjoint custom field sources if this field data has some dimensionality."""
 
         sources = []
         source_geo = self.monitor.geometry
@@ -1221,12 +1176,24 @@ class FieldData(FieldDataset, ElectromagneticFieldData):
                 if "H" in name:
                     values *= -1
 
-                # make coords that are shifted relative to geometry (0,0,0) = geometry.center
                 coords = dict(field_component.coords.copy())
+                grid_coords = Coords(**{key: coords[key] for key in "xyz"})
+
+                size_element = grid_coords.cell_size_meshgrid
+
+                # make coords that are shifted relative to geometry (0,0,0) = geometry.center
                 for dim, key in enumerate("xyz"):
                     coords[key] = np.array(coords[key]) - source_geo.center[dim]
+
                 coords["f"] = np.array([freq0])
                 values = np.expand_dims(values, axis=-1)
+
+                size_element = np.reshape(size_element, values.shape)
+
+                omega0 = 2 * np.pi * freq0
+                scaling_factor = 0.5 * omega0 * EPSILON_0 / size_element
+
+                values *= scaling_factor
 
                 # ignore zero components
                 if not np.all(values == 0):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -275,7 +275,10 @@ ignore = [
 
 [tool.pytest.ini_options]
 # TODO: remove --assert=plain when https://github.com/scipy/scipy/issues/22236 is resolved
-addopts = "--cov=tidy3d --doctest-modules -n auto --dist worksteal --assert=plain"
+addopts = "--cov=tidy3d --doctest-modules -n auto --dist worksteal --assert=plain -m 'not numerical' "
+markers = [
+    "numerical: marks numerical tests for adjoint gradients that require running simulations (deselect with '-m \"not numerical\"')",
+]
 env = ["MPLBACKEND=Agg", "OMP_NUM_THREADS=1"]
 doctest_optionflags = "NORMALIZE_WHITESPACE ELLIPSIS"
 norecursedirs = [
--- a/tests/test_components/test_autograd_numerical.py
+++ b/tests/test_components/test_autograd_numerical.py
@@ -1,0 +1,370 @@
+# test autograd and compares to numerically computed finite difference gradients
+
+import operator
+import sys
+
+import autograd as ag
+import matplotlib.pylab as plt
+import numpy as np
+import pytest
+import tidy3d as td
+import tidy3d.web as web
+from scipy.ndimage import gaussian_filter
+
+PLOT_FD_ADJ_COMPARISON = False
+NUM_FINITE_DIFFERENCE = 10
+SAVE_FD_ADJ_DATA = False
+SAVE_FD_LOC = 0
+SAVE_ADJ_LOC = 1
+LOCAL_GRADIENT = False
+VERBOSE = False
+NUMERICAL_RESULTS_DATA_DIR = "./numerical_field_test/"
+SHOW_PRINT_STATEMENTS = False
+
+RMS_THRESHOLD = 0.25
+
+if PLOT_FD_ADJ_COMPARISON:
+    pytestmark = pytest.mark.usefixtures("mpl_config_interactive")
+else:
+    pytestmark = pytest.mark.usefixtures("mpl_config_noninteractive")
+
+if SHOW_PRINT_STATEMENTS:
+    sys.stdout = sys.stderr
+
+FINITE_DIFF_PERM_SEED = 1.5**2
+MESH_FACTOR_DESIGN = 30.0
+
+
+def get_sim_geometry(mesh_wvl_um):
+    return td.Box(size=(5 * mesh_wvl_um, 5 * mesh_wvl_um, 7 * mesh_wvl_um), center=(0, 0, 0))
+
+
+def make_base_sim(
+    mesh_wvl_um,
+    adj_wvl_um,
+    monitor_size_wvl,
+    box_for_override,
+    monitor_bg_index=1.0,
+    run_time=1e-11,
+):
+    sim_geometry = get_sim_geometry(mesh_wvl_um)
+    sim_size_um = sim_geometry.size
+    sim_center_um = sim_geometry.center
+
+    boundary_spec = td.BoundarySpec(
+        x=td.Boundary.pml(),
+        y=td.Boundary.pml(),
+        z=td.Boundary.pml(),
+    )
+
+    dl_design = mesh_wvl_um / MESH_FACTOR_DESIGN
+
+    mesh_overrides = []
+    mesh_overrides.extend(
+        [
+            td.MeshOverrideStructure(
+                geometry=box_for_override,
+                dl=[dl_design, dl_design, dl_design],
+            ),
+        ]
+    )
+
+    src_size = sim_size_um[0:2] + (0,)
+
+    wl_min_src_um = 0.9 * adj_wvl_um
+    wl_max_src_um = 1.1 * adj_wvl_um
+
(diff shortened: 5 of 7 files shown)


groberts-flex:
Screenshot 2025-03-27 at 1 49 12 PM
a6a361" />
Screenshot 2025-03-27 at 1 49 35 PM
Screenshot 2025-03-27 at 1 49 19 PM

groberts-flex:
Screenshot 2025-03-27 at 1 49 12 PM
Screenshot 2025-03-27 at 1 49 19 PM
Screenshot 2025-03-27 at 1 49 27 PM
Screenshot 2025-03-27 at 1 49 35 PM

tidy3d/components/data/monitor_data.py
```diff
@@ -1221,13 +1179,47 @@ def to_adjoint_field_sources(self, fwidth: float) -> List[CustomCurrentSource]:
                 if "H" in name:
                     values *= -1
 
-                # make coords that are shifted relative to geometry (0,0,0) = geometry.center
+                size_element = np.ones(values.shape)
+                meshgrid_elements = []
+
                 coords = dict(field_component.coords.copy())
```
groberts-flex, line 1179: Ended up being simpler than passing in the discretized monitor because coords had enough information in the end. I thought we were needing to normalize by the grid surrounding the monitor, but we just need to normalize by the correct dimension based on the monitor. For for 2D monitors, we normalize by area, for 1D monitors, we normalize by the linear mesh element, etc. I think @tylerflex actually suggested this would be the case when we all discussed in slack and it turned out to be right

groberts-flex:
The above images show the results of running the whole unit test (which is **not** on by default) and plotting the finite difference vs adjoint. I believe it tests all relevant combinations of monitor sizes (0D, 1D, 2D, 3D), background indices on the monitor, wavelengths, mesh sizes, and a couple of different objective functions (mostly tests intensity, but for 2D we also test flux).

tidy3d/components/data/monitor_data.py
```diff
@@ -1221,13 +1179,47 @@ def to_adjoint_field_sources(self, fwidth: float) -> List[CustomCurrentSource]:
...
                 for dim, key in enumerate("xyz"):
                     coords[key] = np.array(coords[key]) - source_geo.center[dim]
+
                 coords["f"] = np.array([freq0])
                 values = np.expand_dims(values, axis=-1)
 
+                size_element = np.reshape(size_element, values.shape)
+
+                omega0 = 2 * np.pi * freq0
+                scaling_factor = 0.5 * omega0 * EPSILON_0 / size_element
```
groberts-flex, line 1194: is this the right place for the EPSILON_0 factor? I would suspect it needs to come in for the adjoint simulation types  (i.e. - non-field cases) as well when dealing with dielectrics. I didn't want to put it somewhere more general at this time because I didn't want to mess up the current scaling for the other adjoint sources. But if this is applied in all cases, maybe it makes sense to move out of the source definition and into the place where we compute the final gradient?

tidy3d/components/data/monitor_data.py
tylerflex: that's a relief :)

tidy3d/components/data/monitor_data.py
tylerflex: unclear to me where ultimately it should go. I think in general what really matters is where the ω's go. Maybe not worth checking right now, but was only FieldData output affected by this scaling issue? or did it also pop up for eg. ModeData?

It's possible the ε0 would be needed in the more general gradient calculation, but I think for now putting in the source is good enough.

tylerflex:
The test results look great!

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Compatibility with `xarray>=2025.03`.
 
+### Fixed
+- Fixed scaling for adjoint sources when differentiating with respect to field monitor data to account for the mesh size of the monitor and thus the created source. This aligns adjoint gradient magnitudes with numerical finite difference gradients for field data.
```
tylerflex: wouldn't hurt to just replace "field monitor data" with `FieldData` to be explicit.

tests/conftest.py
```diff
@@ -68,3 +68,12 @@ def pytest_xdist_auto_num_workers(config):
     if os.getenv("GITHUB_ACTIONS"):
         return cores
     return max(1, cores - 1)
+
+
+def pytest_addoption(parser):
```
tylerflex: a little docstring explaining what this does / how to use? I'm unfamiliar with it.

tests/test_components/test_numerical_field_data.py
```diff
@@ -0,0 +1,319 @@
+# test numerical autograd for field monitors and field data
```
tylerflex: time to consider creating a new directory `tests/test_components/test_autograd/` for the autograd related tests?

tidy3d/components/data/monitor_data.py
```diff
@@ -1221,13 +1179,47 @@ def to_adjoint_field_sources(self, fwidth: float) -> List[CustomCurrentSource]:
...
+
+                size_elements = {}
+                for dim in "xyz":
+                    if len(coords[dim]) > 1:
+                        diff = coords[dim][1:].values - coords[dim][0:-1].values
+
+                        diff_left = np.pad(diff, ((1, 0)), mode="edge")
+                        diff_right = np.pad(diff, ((0, 1)), mode="edge")
+
+                        size_elements[dim] = 0.5 * (diff_left + diff_right)
```
tylerflex: slight nitpick, but consider making it's own variable?
```py
diff_avg = 0.5 * (diff_left + diff_right)
size_elements[dim] = diff_avg
 meshgrid_elements.append(diff_avg)
```

tidy3d/components/data/monitor_data.py
```diff
@@ -1221,13 +1179,47 @@ def to_adjoint_field_sources(self, fwidth: float) -> List[CustomCurrentSource]:
                 if "H" in name:
                     values *= -1
 
-                # make coords that are shifted relative to geometry (0,0,0) = geometry.center
+                size_element = np.ones(values.shape)
+                meshgrid_elements = []
+
                 coords = dict(field_component.coords.copy())
+
+                size_elements = {}
```
tylerflex: I feel like this whole `size_elements` calculation might come in handy in other places. did you consider separating to it's own utility function somewhere? maybe in `tidy3d/components/grid/`?

tylerflex (commented):
Looks great, super nice to have this working properly, thanks @groberts-flex ! amazing work

tidy3d/components/data/monitor_data.py
yaugenst-flex: Agree with @tylerflex I think in the context of this PR let's just keep it like this. Sometime soon I think we should also double check mode data and if we discover there is something off there we can come back to this then.

tests/test_components/test_numerical_field_data.py
yaugenst-flex: we discussed this but decided it'd be best to keep with the current structure for now. a bigger test refactor is coming up soon(ish). I really want to move to a test structure that mirrors the actual repository structure for unit tests. integration tests should go in a separate directory, and for autograd numerical we might just want to move those to the backend

yaugenst-flex: that being said, I would maybe rename this to `test_autograd_numerical.py`

tests/conftest.py
```diff
@@ -68,3 +68,12 @@ def pytest_xdist_auto_num_workers(config):
...
+    parser.addoption(
+        "--run-numerical-adjoint",
```
yaugenst-flex: I'd prefer the marker not be a verb. I'd follow the convention in the [pytest docs](https://docs.pytest.org/en/stable/example/markers.html) where the markers describe properties of the tests (e.g. "slow", "webtest", ...). So here it'd just be `numerical-adjoint` or maybe better `adjoint-numerical`

tests/test_components/test_numerical_field_data.py
```diff
@@ -0,0 +1,319 @@
...
+        eval_fn_names.append("flux")
+
+    return eval_fns, eval_fn_names
+
+
+@pytest.mark.skipif(
+    "not config.getoption('--run-numerical-adjoint')",
+    reason="Only run when --run-numerical-adjoint is given",
+)
+def test_finite_difference(rng):
```
yaugenst-flex: Looks like the `rng` here is unused?

tests/test_components/test_numerical_field_data.py
```diff
@@ -0,0 +1,319 @@
...
+LOCAL_GRADIENT = False
+VERBOSE = False
+SIM_PATH_DIR = "./data/"
+NUMERICAL_RESULTS_DATA_DIR = "./numerical_field_test/"
+SHOW_PRINT_STATEMENTS = False
+
+RMS_THRESHOLD = 0.25
+
+Path(SIM_PATH_DIR).mkdir(parents=True, exist_ok=True)
+Path(NUMERICAL_RESULTS_DATA_DIR).mkdir(parents=True, exist_ok=True)
```
yaugenst-flex: I'd suggest using fixtures for directory setup (and cleanup when the test session is done). I guess you probably want to store some of the data here? If not, an even better option might be to just use `tmp_path` (https://docs.pytest.org/en/stable/how-to/tmp_path.html)

tests/test_components/test_numerical_field_data.py
```diff
@@ -0,0 +1,319 @@
...
+
+if PLOT_FD_ADJ_COMPARISON:
+    matplotlib.use("TkAgg")
```
yaugenst-flex: We have a matplotlib setup and teardown fixture [in `test_viz.py`](https://github.com/flexcompute/tidy3d/blob/954a43aa73f514d714f937bc4c6316e8344eea3e/tests/test_components/test_viz.py#L14-L21), might make sense to move that to `conftest.py` and reuse?

tests/test_components/test_numerical_field_data.py
```diff
@@ -0,0 +1,319 @@
...
+        eval_fns.append(flux)
+        eval_fn_names.append("flux")
+
+    return eval_fns, eval_fn_names
+
+
+@pytest.mark.skipif(
+    "not config.getoption('--run-numerical-adjoint')",
+    reason="Only run when --run-numerical-adjoint is given",
+)
```
yaugenst-flex: You should register these markers properly with pytest and then just mark the test like so:
```py
@pytest.mark.adjoint_numerical
def test(): ...
```

or perhaps just `@pytest.mark.numerical`

We could even mix and match e.g. `slow`, `autograd`, `numerical` markers for more fine grained control. That could look like:
```py
@pytest.mark.adjoint
@pytest.mark.numerical
@pytest.mark.parametrize(...)
def test(): ...
```

tests/test_components/test_numerical_field_data.py
```diff
@@ -0,0 +1,319 @@
...
+
+        box_for_override = td.Box(
+            center=(0, 0, 0), size=sim_geometry.size[0:2] + (thickness_um + mesh_wvl_um,)
+        )
+
+        for monitor_size_wvl in monitor_sizes_3d_wvl:
+            eval_fns, eval_fn_names = make_eval_fns(monitor_size_wvl)
+
+            for monitor_bg_index in background_indices:
+                for eval_fn_idx, eval_fn in enumerate(eval_fns):
```
yaugenst-flex: I'd strongly prefer using `pytest.mark.parametrize` over these nested for-loops here since these are all different test configurations.

tidy3d/components/data/monitor_data.py
```diff
@@ -1156,56 +1156,14 @@ def to_source(
             field_dataset=dataset, source_time=source_time, center=center, size=size, **kwargs
         )
 
+    #
+    # pass relevant information to make_adjoint_sources
+    #
```
yaugenst-flex: I think this comment is unnecessary?

tidy3d/components/data/monitor_data.py
yaugenst-flex: Maybe something like a `cell_sizes` property?

yaugenst-flex (commented):
Thanks @groberts-flex this is awesome! Left some comments but nothing too major once the tests are passing :)

tests/test_components/test_numerical_field_data.py
groberts-flex: I had originally tried this technique to use the pytest.mark, but the issue I was running into was the mark allows me to run only tests that are marked with certain keywords, but it didn't skip those tests when no marker was requested. As in, if it was marked as pytest.mark.adjoint, I could choose to only run that test in the file by specifying -m adjoint, but if I didn't specify any makers, the test also was run. So this is why I added the custom flag to skip the test unless the flag is present.

tests/test_components/test_numerical_field_data.py
groberts-flex: good catch, passed it in and forgot to use it and called numpy random instead! will change to use rng

tests/conftest.py
groberts-flex: sounds good, will change to adjoint-numerical!

tests/test_components/test_numerical_field_data.py
groberts-flex: sounds good, happy to rename the file. I was thinking of creating more of these tests for different types of monitors which is why I kept this one specific to field data in the title. But I can also put all the other monitor types in one file

tests/test_components/test_numerical_field_data.py
yaugenst-flex: I'd change it in the `pyproject.toml` via `addopts = -m "not numerical"` for example. But yeah don't think there is a way to do this within a session.

groberts-flex: oh cool, that makes sense! changing to use mark and updating .toml file

CHANGELOG.md
groberts-flex: good call!

tests/conftest.py
groberts-flex: good point, was unfamiliar with it too. ended up taking it out in favor of Yannick's suggestion on how to get the pytest mark to work

tidy3d/components/data/monitor_data.py
groberts-flex: Sounds good, that makes sense to me, I'll keep this as is for now and will circle back when I have a fuller picture of how these factors come in for different monitors/sources! My plan is to create tests like these for mode monitors next, but haven't checked them to this point. I'll get to that after I conclude the initial PEC gradient work and get that into a PR. I noticed the epsilon thing possibly being out of place here based on the PEC gradients because for that gradient we scale the magnetic field component by mu instead of epsilon. So to get that gradient on the right scale when using field data, I actually need to multiply by mu_0/eps_0 in the surface integral calculation for H and not scale the E component by anything. That part is what made me think the epsilon factor may need to get put somewhere that's aware of the material being used.

tidy3d/components/data/monitor_data.py
groberts-flex: great point, definitely!

tidy3d/components/data/monitor_data.py
groberts-flex: definitely, good catch!

tests/test_components/test_numerical_field_data.py
groberts-flex: changed this to use tmp_path for the sim data which can get cleaned up afterwards and added a fixture to create a directory for the data I want to save and not clean up after the test

tidy3d/components/data/monitor_data.py
groberts-flex: good idea! separated these out into the Coords class

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Compatibility with `xarray>=2025.03`.
 
+### Fixed
+- Fixed scaling for adjoint sources when differentiating with respect to FieldData to account for the mesh size of the monitor and thus the created source. This aligns adjoint gradient magnitudes with numerical finite difference gradients for field data.
+
```
yaugenst-flex: Seems a duplicate "Fixed" slipped in here

tests/conftest.py
```diff
@@ -68,3 +70,35 @@ def pytest_xdist_auto_num_workers(config):
...
+    mpl.use(original_backend)
+
+
+@pytest.fixture
+def mpl_config_interactive():
+    """Configure matplotlib interactive backend for all tests in this module."""
+    original_backend = mpl.get_backend()
+    mpl.use("TkAgg")
+    yield
+    mpl.use(original_backend)
```
yaugenst-flex, line 91: Duplicate code?

yaugenst-flex: oooh i see nevermind

yaugenst-flex:
@groberts-flex https://github.com/flexcompute/tidy3d/pull/2348 was merged so you can rebase and push and hopefully the tests pass

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Compatibility with `xarray>=2025.03`.
+- Fixed scaling for adjoint sources when differentiating with respect to FieldData to account for the mesh size of the monitor and thus the created source. This aligns adjoint gradient magnitudes with numerical finite difference gradients for field data.
```
tylerflex: could you put `FieldData` in single bakslash quotes?

tylerflex (approved):
Looks good to me, just a minor note about changelog. thanks @groberts-flex !

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Compatibility with `xarray>=2025.03`.
 - Inaccurate gradient when auto-grabbing permittivities for structures using `td.PolySlab` when using dispersive material models.
+- Fixed scaling for adjoint sources when differentiating with respect to FieldData to account for the mesh size of the monitor and thus the created source. This aligns adjoint gradient magnitudes with numerical finite difference gradients for field data.
+- Fixed scaling for adjoint sources when differentiating with respect to `FieldData` to account for the mesh size of the monitor and thus the created source. This aligns adjoint gradient magnitudes with numerical finite difference gradients for field data.
```
yaugenst-flex: Duplicate lines?

tests/conftest.py
```diff
@@ -68,3 +70,35 @@ def pytest_xdist_auto_num_workers(config):
...
+def dir_name(request):
+    return request.param
+
+
+@pytest.fixture
+def create_directory(dir_name):
+    if dir_name is not None:
+        directory = Path(dir_name).mkdir(parents=True, exist_ok=True)
+
+    yield
```
yaugenst-flex: The `yield` here is unnecessary. Also, just to confirm, this is for creating the output data that you want to inspect after the test run right?

yaugenst-flex (approved):
Nice found some minor things but otherwise looks good to go!

CHANGELOG.md
groberts-flex: thanks for catching, not sure why that happened! deleted extra line

tests/conftest.py
groberts-flex: thanks, makes sense! yeah these are for saving the data for inspection. For data we don't need afterwards, I used the tmp path fixture.

PR #2343, opened 2025-03-31
fix[testing]: fixed test_geometry bounds checking tests

groberts-flex:
In test_geometry, the polyslab bounds checking was failing depending on the number of workers. @yaugenst-flex helped me identify what was going on here. Updated here also to check against actual expected bounds.

### DIFF
--- a/tests/test_components/test_geometry.py
+++ b/tests/test_components/test_geometry.py
@@ -12,7 +12,6 @@ import pytest
 import shapely
 import tidy3d as td
 import trimesh
-from tidy3d.components.geometry.base import Planar
 from tidy3d.components.geometry.mesh import AREA_SIZE_THRESHOLD
 from tidy3d.components.geometry.utils import (
     SnapBehavior,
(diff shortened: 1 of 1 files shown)


tests/test_components/test_geometry.py
```diff
@@ -36,6 +35,9 @@
 SPHERE = td.Sphere(radius=1)
 CYLINDER = td.Cylinder(axis=2, length=1, radius=1)
 
+EXPECTED_POLYSLAB_BOUNDS = ((0.0, 0.0, -0.5), (1.0, 1.0, 0.5))
+EXPECTED_CYLINDER_BOUNDS = ((-1.0, -1.0, -0.5), (1.0, 1.0, 0.5))
```
yaugenst: ~~Since these are only used in the respective tests they should go there I think.~~
Actually disregard for comment below

tests/test_components/test_geometry.py
```diff
@@ -215,7 +217,7 @@ def test_intersections_plane_inf():
 
 
 def test_bounds_base():
-    assert all(a == b for a, b in zip(Planar.bounds.fget(POLYSLAB), POLYSLAB.bounds))
+    assert all(a == b for a, b in zip(td.PolySlab.bounds.fget(POLYSLAB), EXPECTED_POLYSLAB_BOUNDS))
```
yaugenst: Since these are essentially the same test now but parametrized, let's make it so:
```py
@pytest.mark.parametrize("component,expected_bounds", [(CYLINDER, cylinder_bounds), (POLYSLAB, polyslab_bounds)])
def test_bounds(component, expected_bounds): ...
```

also don't use `td.PolySlab.bounds.fget(POLYSLAB)` but `component.bounds` instead

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Compatibility with `xarray>=2025.03`.
+- Fixed bounds checking tests in test_geometry that were failing depending on number of pytest workers and updated them to check against expected bounds.
```
yaugenst: Generally we don't mention implementation details of the test suite in the changelog, so I'd just remove this.

yaugenst (changes requested):
Thanks @groberts-flex!

yaugenst-flex:
needs a rebase and should pass now since https://github.com/flexcompute/tidy3d/pull/2348 got merged

PR #2344, opened 2025-04-01
Vasily/bandgap_monitor_update

vasilyzabelin:
Update of the energy bandgap monitor for the CHARGE solver

### DIFF
--- a/tidy3d/components/tcad/data/monitor_data/charge.py
+++ b/tidy3d/components/tcad/data/monitor_data/charge.py
@@ -17,10 +17,13 @@ from tidy3d.components.data.utils import TetrahedralGridDataset, TriangularGridD
 from tidy3d.components.tcad.data.monitor_data.abstract import HeatChargeMonitorData
 from tidy3d.components.tcad.monitors.charge import (
     SteadyCapacitanceMonitor,
+    SteadyEnergyBandMonitor,
     SteadyFreeCarrierMonitor,
     SteadyPotentialMonitor,
 )
-from tidy3d.components.types import TYPE_TAG_STR, annotate_type
+from tidy3d.components.types import TYPE_TAG_STR, Ax, annotate_type
+from tidy3d.components.viz import add_ax_if_none
+from tidy3d.exceptions import DataError
 from tidy3d.log import log
 
 FieldDataset = Union[
@@ -73,7 +76,7 @@ class SteadyPotentialData(HeatChargeMonitorData):
         return self.updated_copy(potential=new_potential, symmetry=(0, 0, 0))
 
     def field_name(self, val: str) -> str:
-        """Gets the name of the fields to be plot."""
+        """Gets the name of the fields to be plotted."""
         if val == "abs^2":
             return "|V|²"
         else:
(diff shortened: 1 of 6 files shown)


momchil-flex (commented):
Seems like it still needs a test to be added as per my comment on the other PR.

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -165,13 +168,235 @@ def symmetry_expanded_copy(self) -> SteadyFreeCarrierData:
...
+        return dict(Ec=self.Ec, Ev=self.Ev, Ei=self.Ei, Efn=self.Efn, Efp=self.Efp)
+
+    @pd.root_validator(skip_on_failure=True)
+    def check_correct_data_type(cls, values):
+        """Issue error if incorrect data type is used"""
+
+        mnt = values.get("monitor")
+        field_data = {field: values.get(field) for field in ["Ec", "Ev", "Ei", "Efn", "Efp"]}
+
+        for field, data in field_data.items():
```
marc-flex, line 247: This `field_data` seems redundant? The for-loop could be over the fields, i.e., `for field in ["Ec", "Ev", "Ei", "Efn", "Efp"]:` as you're doing below.

Not particularly bother by it though

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -165,13 +168,235 @@ def symmetry_expanded_copy(self) -> SteadyFreeCarrierData:
...
+        method: Literal[None, "nearest", "pad", "ffill", "backfill", "bfill"] = None
+            Method to use in xarray sel() function.
+        **sel_kwargs : dict
+            Keyword arguments to pass to the xarray sel() function.
+        """
+
+        if not isinstance(self.Ec, TriangularGridDataset):
+            raise DataError(
+                "Bandgap monitor slice plot can be done only for a 2D unstructured dataset."
+            )
```
marc-flex: Do we really need this restriction? If we had a 3D monitor we would have a `TetrahedralGridDataset` that we could slice twice?

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -165,13 +168,235 @@ def symmetry_expanded_copy(self) -> SteadyFreeCarrierData:
...
+        -------
+        SteadyEnergyBandData
+            The resulting slice.
+        """
+
+        if not isinstance(self.Ec, TetrahedralGridDataset):
+            raise DataError(
+                "Bandgap monitor plane slice can be done only for a 3D unstructured dataset."
+            )
+
```
marc-flex: So if I understand correctly, the user would first need to create a copy of the data using this `plane_slice(axis, pos)` and then `plot_slice(axis, pos)`. Wouldn't it make sense to have a plot function that would take all? So for instance, in the case of a 3D monitor it could be `plot_diagram(x=1,y=3)` and when a 2D `plot_diagram(x=1)`?

marc-flex (commented):
Thanks Vasily for this! My main comment would be the platting function and how this would work for a 3D monitor. But other than that looks good to me

tidy3d/components/tcad/data/monitor_data/charge.py
vasilyzabelin: Here I have the loop where I need co sweep "data" and "data" variables together. I thought that this way is better to preserve that I they are updated each iteration of the loop.

tidy3d/components/tcad/data/monitor_data/charge.py
vasilyzabelin: The logic of the bandgap monitor is that we gather all fields (Ec, Ev, Efn, etc.) together and do operations like plane_slice() and plot_slice() on all variables. So, plot_slice() works only for a 2D data and plots the 1D cross-section for all fields, and plane_slice() works only for a 3D data and make a 2D crossection -> new bandgap monitor data object that we can use for plot_slice() etc.

vasilyzabelin: If user wants to do more detailed analysis, the original data fields (and all related operations) are available.

tidy3d/components/tcad/data/monitor_data/charge.py
vasilyzabelin: It can be done this way.
But in the current implementation we need only two specific operations for the bandgap monitor with the syntax that is very close to operations for .sel() function and plane_slice(). So it is already familiar for users.

vasilyzabelin:
Regarding the unit tests. I want to add one, but I don't understand the logic of the existing unit tests for SteadyPotentialMonitor, SteadyFreeCarrierMonitor, etc. It looks like they just test creation of the new data.

tidy3d/components/tcad/data/monitor_data/charge.py
marc-flex: I think it'd be nicer to have it all in one function so that the user doesn't need to go through a series of steps instead of doing it in a single step.

marc-flex: So basically:

- If data is 3D
    - user specifies only one coordinate (e.g., x=2) -> raise error
    - user specifies 2 coordinates (e.g., x=2, z=4) -> OK
- if data is 2D
    - user specifies only one coordinate (e.g., x=2) -> OK
    - user specifies 2 coordinates (e.g., x=2, z=4) -> raise error

momchil-flex: Yeah that would also make sense and is along the lines of our field plotting, again bringing up the `sel_kwargs` there:
https://github.com/flexcompute/tidy3d/blob/382f2041dd192f113e4f7b212d42f1be66bb7836/tidy3d/components/data/sim_data.py#L488-L493

In that case, we expect exactly two spatial coordinates to remain after all the selection is done. Here, we expect one.

vasilyzabelin: OK. I will update the code.

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -298,103 +298,81 @@ def field_name(self, val: str = "") -> str:
...
-        self,
-        axis: Axis,
-        pos: float,
-        ax: Ax = None,
-        method: Literal["None", "nearest", "pad", "ffill", "backfill", "bfill"] = None,
-        **sel_kwargs,
-    ) -> Ax:
-        """Plot the energy bands along the slice.
+    def plot_data(self, ax: Ax = None, **sel_kwargs) -> Ax:
+        """Plot the 1D cross-section of the energy bandgap diagram.
```
marc-flex: This is a bit of a personal taste but I feel like `plot_data` sounds generic. What about `plot_energy_band_diagram`? Or simply `plot` since this is already a `SteadyEnergyBandMonitor` and it'd be shorter

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -298,103 +298,81 @@ def field_name(self, val: str = "") -> str:
...
-            )
+        selection_data = dict()
 
-        if axis == self.Ec.normal_axis:
-            raise DataError(
-                f"Triangular grid (normal: {self.Ec.normal_axis}) cannot be sliced by a parallel "
-                "plane."
-            )
+        if "voltage" not in sel_kwargs:
+            raise DataError("'voltage' is not selected for the plot.")
```
marc-flex: We need to be a bit careful with this. If you run for a single voltage you should be able to get away without selecting a `voltage`

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -298,103 +298,81 @@ def field_name(self, val: str = "") -> str:
...
-        Ev_slice.plot(ax=ax, label="Ev")
-        Ei_slice.plot(ax=ax, label="Ei")
-        Efn_slice.plot(ax=ax, label="Efn")
-        Efp_slice.plot(ax=ax, label="Efp")
-        ax.legend()
+        N_coords = 0
+        for coord_name in ["x", "y", "z"]:
+            if coord_name in sel_kwargs:
+                N_coords += 1
+                selection_data[coord_name] = sel_kwargs[coord_name]
```
marc-flex: what about
```python
selection_data = {coord: sel_kwarg[coord] for coord in "xyz" if coord in sel_kwarg.keys() }
N_coords = len(selection_data.keys())
```

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -298,103 +298,81 @@ def field_name(self, val: str = "") -> str:
...
+        Ei_data = self.Ei
+        Efn_data = self.Efn
+        Efp_data = self.Efp
+
+        for coord_name, coord_val in selection_data.items():
+            Ec_data = Ec_data.sel(**{coord_name: coord_val}, method="nearest")
+            Ev_data = Ev_data.sel(**{coord_name: coord_val}, method="nearest")
+            Ei_data = Ei_data.sel(**{coord_name: coord_val}, method="nearest")
+            Efn_data = Efn_data.sel(**{coord_name: coord_val}, method="nearest")
+            Efp_data = Efp_data.sel(**{coord_name: coord_val}, method="nearest")
```
marc-flex, line 366: would `Efp_data = Efp_data.sel(selection_data, method="nearest")` work?

marc-flex (commented):
Thanks Vasily!
For the naming of the function maybe `plot` would suffice? Other than that I think this looks pretty much ready

marc-flex:
@vasilyzabelin with regards to the test, in those tests you're referring to I think we simply make sure we can create the structures. You can also test that the correct error is raised, For instance, you can check that if your data is of type `TetrahedralGridDataset`  and only one selection coordinate is given, a `DataError` is raised.

tidy3d/components/tcad/data/monitor_data/charge.py
vasilyzabelin: I've changed the name of the function

tidy3d/components/tcad/data/monitor_data/charge.py
vasilyzabelin: This code add more items to the selection_data, not just overriding it. So I think this logic is more clear.

tidy3d/components/tcad/data/monitor_data/charge.py
vasilyzabelin: I tried it, but it doesn't work Ans the code in `tidy3d/components/data/sim_data.py` , `line 582 - 584` also performs selection as a loop over variables. So, it looks like the only way that works.

vasilyzabelin:
I've added the unit test  for the new monitor type.

tidy3d/components/tcad/data/monitor_data/charge.py
marc-flex: Sorry, you mean because of the voltage, right? I mean, you can still have it in 3 lines
```python
selection_data = {coord: sel_kwarg[coord] for coord in "xyz" if coord in sel_kwarg.keys() }
N_coords = len(selection_data.keys())
selection_data["voltage"] = sel_kwargs["voltage"]
```

not a deal breaker though

marc-flex (approved):
Overall this looks good but it'd be great if you could update the test to capture the error when voltage isn't provide but we have fields with multiple voltages

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -165,13 +168,215 @@ def symmetry_expanded_copy(self) -> SteadyFreeCarrierData:
...
+        matplotlib.axes._subplots.Axes
+            The supplied or created matplotlib axes.
+        """
+
+        selection_data = dict()
+
+        if ("voltage" not in sel_kwargs) and (self.Ec.values.coords.sizes["voltage"] > 1):
+            raise DataError(
+                "'voltage' is not selected for the plot with multiple voltage data points."
+            )
```
marc-flex, line 325: Can you check that this error is actually raised in your tests?

vasilyzabelin: I tried to do it, but I can't figure out the proper way. Can we have a meeting to discuss the logic of the unit tests here?

tidy3d/components/tcad/data/monitor_data/charge.py
vasilyzabelin: Updated the code.

tidy3d/components/tcad/data/monitor_data/charge.py
marc-flex: Here's an example of what I mean: https://github.com/flexcompute/tidy3d/blob/3dc27f16bc272101d285df3b2f4bb3a42deaef2b/tests/test_components/test_heat_charge.py#L1320
though there are more examples in that same file. Basically, we set up the component in the way we want the error to arise and check that it does. We can of course discuss it in a meeting. Feel free to schedule one

vasilyzabelin: I will try. As far as I see, the monitor data in various tests here is artificially generated, and not a result of an actual simulation, that is done in this function:
https://github.com/flexcompute/tidy3d/blob/3dc27f16bc272101d285df3b2f4bb3a42deaef2b/tests/test_components/test_heat_charge.py#L199

So, I need to create the monitor data with the desired error?

marc-flex: yes, you can create the data. There's examples in here: https://github.com/flexcompute/tidy3d/blob/3dc27f16bc272101d285df3b2f4bb3a42deaef2b/tests/test_data/test_datasets.py#L672

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -165,13 +168,213 @@ def symmetry_expanded_copy(self) -> SteadyFreeCarrierData:
...
+
+            for index, coord_name in enumerate(["x", "y", "z"]):
+                if coord_name in selection_data:
+                    axis = index
+                    continue
+
+            if axis == self.Ec.normal_axis:
+                raise DataError(
+                    f"Triangular grid (normal: {self.Ec.normal_axis}) cannot be sliced by a parallel plane."
+                )
```
marc-flex, line 353: Let's add a test for these errors too

marc-flex (changes requested):
This is close to finished. I guess at this point is only a matter of adding corresponding tests

Issue #2154, opened 2025-01-10
Improve error handling for `web.load` when a non-existent task ID is provided.


FilipeFcp:
A user pointed out that when an incorrect task-id is provided to the `web.load` function, it leads to a Python error, with a lot of output messages that are not useful for the user.

If possible, a clear message that the task-id doesn't exist could be more informative.

priority: small

momchil-flex:
What do you think @QingengWei

QingengWei:
> What do you think @QingengWei

Web API returns a 404 NOT FOUND error, but the Python client changes the error to None: https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/web/core/http_util.py#L132-L133

momchil-flex:
That's funny, who knows where this comes from you or maybe even Lei Zheng in the first days of the web api? Could you propose a change?

QingengWei:
> That's funny, who knows where this comes from you or maybe even Lei Zheng in the first days of the web api? Could you propose a change?
![image](https://github.com/user-attachments/assets/d5c4fd04-95ac-4c8c-a477-151bb76996a0)
Do you know who they are?

I changed the error code from Web API, and now the error info in Python client is as follows:
![image](https://github.com/user-attachments/assets/7c0e1e86-c462-44e6-9a51-dffe53843189)

yaugenst-flex:
We should probably just handle status code correctly in the client though I'd say?

QingengWei:
> We should probably just handle status code correctly in the client though I'd say?

I don't know the historical reason. I handle it from the web api and the client does not make any changes.

yaugenst-flex:
I understand the reasoning, but returning a different error code just to avoid triggering `try/except None` creates two issues:
 1. The frontend still does not handle 404 errors correctly.
 2. The web API now returns an incorrect or unexpected status code when a simulation is not found.

Instead of introducing workarounds, we should address the root cause by fixing the frontend to properly handle `404` responses. This ensures both the API and client behave as expected without compromising on correctness.

QingengWei:
> I understand the reasoning, but returning a different error code just to avoid triggering `try/except None` creates two issues:
> 
> 1. The front end still does not handle 404 errors correctly.
> 2. The web API now returns an incorrect or unexpected status code when a simulation is not found.
> 
> Instead of introducing workarounds, we should address the root cause by fixing the front end to properly handle `404` responses. This ensures both the API and client behave as expected without compromising on correctness.

[...]

PR #2345, opened 2025-04-01
Fix:Error message for invalid task ID

rahul-flex:
Previously, when web.load received a 404 response for an invalid or non-existent task ID, the client would return None and produce unhelpful error messages. This commit changes that behavior: it now raises a TaskNotFoundError with a clear, user-friendly explanation.
Screenshot 2025-03-31 at 11 22 56 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Supplying autograd-traced values to geometric fields (`center`, `size`) of simulations, monitors, and sources now logs a warning and falls back to the static value instead of erroring.
 - Attempting to differentiate server-side field projections now raises a clear error instead of silently failing.
+- Improved error message and handling when attempting to load a non-existent task ID.
 
 ## [2.8.3] - 2025-04-24
 
--- a/tidy3d/web/core/exceptions.py
+++ b/tidy3d/web/core/exceptions.py
@@ -11,3 +11,9 @@ class WebError(Exception):
         log = get_logger()
         super().__init__(message)
         log.error(message)
+
+
+class WebNotFoundError(WebError):
+    """A generic error indicating an HTTP 404 (resource not found)."""
+
+    pass
--- a/tidy3d/web/core/http_util.py
+++ b/tidy3d/web/core/http_util.py
@@ -25,7 +25,7 @@ from .constants import (
 )
 from .core_config import get_logger
 from .environment import Env
-from .exceptions import WebError
+from .exceptions import WebError, WebNotFoundError
 
 REINITIALIZED = False
 
@@ -131,7 +131,7 @@ def http_interceptor(func):
 
         if resp.status_code != ResponseCodes.OK.value:
             if resp.status_code == ResponseCodes.NOT_FOUND.value:
-                return None
+                raise WebNotFoundError("Resource not found (HTTP 404).")
             json_resp = resp.json()
             if "error" in json_resp.keys():
                 raise WebError(json_resp["error"])
--- a/tidy3d/web/core/task_core.py
+++ b/tidy3d/web/core/task_core.py
@@ -12,12 +12,14 @@ import pydantic.v1 as pd
 from botocore.exceptions import ClientError
 from pydantic.v1 import Extra, Field, parse_obj_as
 
+import tidy3d as td
+
 from . import http_util
 from .cache import FOLDER_CACHE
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Compatibility with `xarray>=2025.03`.
+- Error message for invalid task ID.
```
yaugenst-flex: I'd put this under "Changed"

tidy3d/web/core/http_util.py
```diff
@@ -131,7 +131,9 @@ def wrapper(*args, **kwargs):
 
         if resp.status_code != ResponseCodes.OK.value:
             if resp.status_code == ResponseCodes.NOT_FOUND.value:
-                return None
+                raise TaskNotFoundError(
+                    "The requested task ID does not exist. Please verify your task ID."
+                )
```
yaugenst-flex: This is much more general than not finding a task. Any request that returns a 404 status code will be caught here, so this should also raise a more generic exception. Maybe something like `WebNotFoundError(WebError)`.

The error then needs to be handled properly by the caller. We can focus on tasks for now since that was the original intention. A good place to catch this is probably in `tidy3d/web/core/task_core.py`

yaugenst-flex (changes requested):
Thanks @rahul-flex! This is going in the right direction but we should try to handle 404 errors more generally.

rahul-flex:
Updated error message:
Screenshot 2025-04-01 at 8 47 36 PM

CHANGELOG.md
yaugenst-flex: could you reorder this to Added/Changed/Fixed

tidy3d/web/core/exceptions.py
```diff
@@ -11,3 +11,15 @@ def __init__(self, message: str = None):
         log = get_logger()
         super().__init__(message)
         log.error(message)
+
+
+class TaskNotFoundError(WebError):
+    """Raised when a requested task ID does not exist on the server."""
+
+    pass
```
yaugenst-flex: in principle i'm not opposed to having a fine grained error like this but on the other hand it seems like just raising a `WebNotFoundError` should be fine. so maybe remove this one?

tidy3d/web/core/task_core.py
```diff
@@ -261,7 +261,10 @@ def get(cls, task_id: str, verbose: bool = True) -> SimulationTask:
             :class:`.SimulationTask` object containing info about status,
              size, credits of task and others.
         """
-        resp = http.get(f"tidy3d/tasks/{task_id}/detail")
+        try:
+            resp = http.get(f"tidy3d/tasks/{task_id}/detail")
+        except WebNotFoundError:
+            raise TaskNotFoundError(f"The requested task ID '{task_id}' does not exist.")
```
yaugenst-flex: i'd probably do the following instead:
```py
except WebNotFoundError as e:
  td.log.error("error message")
  raise e
```

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `fill` and `fill_structures` argument in `td.Simulation.plot_structures()` and `td.Simulation.plot()` respectively to disable fill and plot outlines of structures only.
 - New subpixel averaging option `ContourPathAveraging` applied to dielectric material boundaries.
 
+### Changed
+- Error message for invalid task ID.
```
yaugenst-flex: `Improved error message and handling when attempting to load a non-existent task ID.`

yaugenst-flex:
Closes https://github.com/flexcompute/tidy3d/issues/2154

yaugenst-flex (approved):
Thanks @rahul-flex looks good to go from my side, minus the little changelog clarification.

yaugenst-flex:
@QingengWei could you have a look? This addresses the discussion in #2154. The `http_interceptor` now raises an error on 404 instead of returning `None`, which means that the API should now be able to return 404 status which we then need to handle properly here on the client.

The old `if resp is None` checks are still in place to keep compatibility but I think we should move to returning "normal" status codes on the API side?

CHANGELOG.md
yaugenst-flex: @rahul-flex could you add this change

yaugenst-flex:
@rahul-flex I think this is good to go after the small change in the changelog. could you rebase?

CHANGELOG.md
yaugenst-flex: also make sure to put it under "unreleased" after you rebase

PR #2346, opened 2025-04-01
Add validator to ensure that angle_phi is a multiple of pi when angle…

momchil-flex:
…_rotation is used


momchil-flex:
Also opened an issue: https://github.com/flexcompute/tidy3d/issues/2347

lucas-flexcompute (approved):

PR #2348, opened 2025-04-01
Fix polyslab tests and add interior_angle property to PolySlab

weiliangjin2021:


caseyflex:
Thanks @weiliangjin2021. I think the idea makes sense, but it looks like `interior_angle` is in radians while the cutoff in the test is in degrees.

tests/test_components/test_sidewall.py
```diff
@@ -277,16 +289,20 @@ def test_max_erosion_polygon(execution_number):
...
     dilation = -max_dist + 1e-10
     # avoid polygon splitting etc. case
-    if not s._edge_events_detection(s.reference_polygon, dilation, ignore_at_dist=False):
-        s = setup_polyslab(vertices, dilation, angle, bounds)
-        assert np.isclose(minimal_edge_length(s.reference_polygon), 0, atol=1e-4)
+    if s._edge_events_detection(s.reference_polygon, dilation, ignore_at_dist=False):
+        return
+    # avoid checking if interior angle too small
+    if any(s.interior_angle < 1):
+        return
```
yaugenst-flex: I'd avoid early returns in tests. If for some reason e.g. the `s._edge_events_detection` function is somehow broken this whole test will effectively not run but look like it's passing. Ideally there would be specific tests for those edge cases but just having the information that the test was skipped would already be good enough, so basically a `pytest.skip("Reason")` instead of `return`.

weiliangjin2021: thanks, done

weiliangjin2021:
> Thanks @weiliangjin2021. I think the idea makes sense, but it looks like `interior_angle` is in radians while the cutoff in the test is in degrees.

Thanks for catching that! Changed to 1e-2 in the test.

yaugenst-flex (approved):
Thanks @weiliangjin2021 looks good to me! We should merge this asap, the failing test is blocking some other PRs

caseyflex (approved):

PR #2351, opened 2025-04-02
added ability to define a wave port using only a single voltage or current path

dmarek-flex:
An important step on the road to automating wave ports is the ability to define wave ports with only a single path integral (voltage or current). Typically it is recommended to rely on the current version, which results in a characteristic impedance which is more stable vs frequency.

To simplify these steps, and future steps, I switched from recording fields at each port using a field monitor to a mode monitor. Voltage and current quantities are now calculated using the modal fields along with the modal amplitudes returned in the `ModeData`.

### DIFF
--- a/tests/test_plugins/smatrix/terminal_component_modeler_def.py
+++ b/tests/test_plugins/smatrix/terminal_component_modeler_def.py
@@ -251,6 +251,8 @@ def make_coaxial_component_modeler(
         CoaxialLumpedPort,
         CoaxialLumpedPort,
     ),
+    use_current: bool = True,
+    use_voltage: bool = True,
     **kwargs,
 ):
     if not length:
@@ -279,6 +281,25 @@ def make_coaxial_component_modeler(
             voltage_center[0] += mean_radius
             voltage_size = [Router - Rinner, 0, 0]
 
+            voltage_integral = None
+            if use_voltage:
+                voltage_integral = microwave.VoltageIntegralAxisAligned(
+                    center=voltage_center,
+                    size=voltage_size,
+                    extrapolate_to_endpoints=True,
+                    snap_path_to_grid=True,
+                    sign="+",
+                )
+            current_integral = None
+            if use_current:
+                current_integral = microwave.CustomCurrentIntegral2D.from_circular_path(
+                    center=center,
+                    radius=mean_radius,
+                    num_points=41,
+                    normal_axis=2,
+                    clockwise=direction != "+",
+                )
+
             port = WavePort(
                 center=center,
                 size=[2 * Router, 2 * Router, 0],
@@ -286,20 +307,8 @@ def make_coaxial_component_modeler(
                 name=name,
                 mode_spec=td.ModeSpec(num_modes=1),
                 mode_index=0,
-                voltage_integral=microwave.VoltageIntegralAxisAligned(
-                    center=voltage_center,
-                    size=voltage_size,
-                    extrapolate_to_endpoints=True,
-                    snap_path_to_grid=True,
-                    sign="+",
-                ),
-                current_integral=microwave.CustomCurrentIntegral2D.from_circular_path(
-                    center=center,
-                    radius=mean_radius,
-                    num_points=41,
-                    normal_axis=2,
-                    clockwise=direction != "+",
-                ),
+                voltage_integral=voltage_integral,
+                current_integral=current_integral,
             )
         return port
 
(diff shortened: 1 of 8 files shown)


tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -188,7 +209,7 @@ def _validate_path_integrals_within_port(cls, val, values):
     def _check_voltage_or_current(cls, val, values):
```
weiliangjin2021, line 210: add `@skip_if_fields_missing(["voltage_integral"])` in case it failed validation and the error message is then confusing

tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -94,33 +113,22 @@ def to_source(self, source_time: GaussianPulse, snap_center: float = None) -> Mo
 
     def to_field_monitors(
         self, freqs: FreqArray, snap_center: float = None, grid: Grid = None
-    ) -> list[FieldMonitor]:
-        """Field monitor to compute port voltage and current."""
+    ) -> list[ModeMonitor]:
```
weiliangjin2021, line 116: also need to update the docstring in base_terminal.py

tests/test_plugins/smatrix/test_terminal_component_modeler.py
```diff
@@ -413,17 +413,29 @@ def test_make_coaxial_component_modeler_with_wave_ports(tmp_path):
...
 
-def test_run_coaxial_component_modeler_with_wave_ports(monkeypatch, tmp_path):
+@pytest.mark.parametrize("voltage_enabled", [False, True])
+@pytest.mark.parametrize("current_enabled", [False, True])
+def test_run_coaxial_component_modeler_with_wave_ports(
+    monkeypatch, tmp_path, voltage_enabled, current_enabled
+):
     """Checks that the terminal component modeler runs with wave ports."""
+    if not (voltage_enabled or current_enabled):
+        pytest.skip("Skip invalid parameters.")
```
weiliangjin2021: check whether the error is raised instead?

weiliangjin2021 (commented):
Nice work! Just some minor comments.

tidy3d/plugins/smatrix/ports/wave.py
dmarek-flex: I wonder if I should break backwards compatibility and rename this method `to_mode_monitors` to be more clear. In fact by returning a `ModeMonitor` it is already broken a bit. The lumped elements use this method as well.

weiliangjin2021: I think we shouldn't break backwards compatibility. Maybe add a new `to_monitors`, and add warnings to `to_field_monitors` saying that it is deprecated.

dmarek-flex: Excellent, I always forget that is an option!

weiliangjin2021 (approved):

tests/test_plugins/smatrix/terminal_component_modeler_def.py
```diff
@@ -251,6 +251,8 @@ def make_coaxial_component_modeler(
         CoaxialLumpedPort,
         CoaxialLumpedPort,
     ),
+    current_enabled=True,
+    voltage_enabled=True,
```
yaugenst-flex: nit: type hint

yaugenst-flex: also: I wonder if `use_current` and `use_voltage` might be clearer here? just a thought, not really sure what makes more sense. if in doubt it's fine as is

tidy3d/plugins/microwave/impedance_calculator.py
```diff
@@ -72,15 +72,15 @@ def compute_impedance(self, em_field: MonitorDataTypes) -> IntegralResultTypes:
         if not self.voltage_integral:
             flux = em_field.flux
             if isinstance(em_field, FieldTimeData):
-                voltage = flux / current
+                voltage = flux.abs / current
```
yaugenst-flex, line 75: Interesting, why is `abs` needed here?

yaugenst-flex (approved):
Thanks @dmarek-flex looks good! Really only one tiny typing-related comment otherwise looks good to go.

tidy3d/plugins/microwave/impedance_calculator.py
dmarek-flex: Really, it was a simple way to ensure that the calculated impedance always has a positive real part. When not using `abs`, the issue was that "flux" using data from the `ModeMonitor` will be relative to `store_fields_direction`, whereas the flux recorded by `FieldMonitor` is always relative to the "+" direction (since they do not have a field for controlling directionality).

yaugenst-flex: ah I see! thanks for clarifying

PR #2352, opened 2025-04-02
fix[adjoint]: Pass correct frequency for epsilon calculation in adjoint postprocessing

yaugenst-flex:
The automatic permittivity grabbing was not passing the frequency to `simulation.epsilon()`, leading to wrong permittivity and bad gradients.


groberts-flex:
Good catch! This looks good to me

groberts-flex (approved):

tylerflex (approved):

Issue #2354, opened 2025-04-03
Heat conductivity plotting error


alec-flexcompute:
**Describe the bug**
`plot_property` in HeatChargeSimulations has an error when plotting heat conductivity. See the attached code

**Screenshots**
![Image](https://github.com/user-attachments/assets/71d2eb65-40db-49d3-bf27-6806c0388ea5)

`import numpy as np
import matplotlib.pyplot as plt
import gdstk
import os

import tidy3d as td
import tidy3d.web as web

print(td.__version__)

medium_al = td.MultiPhysicsMedium(heat=td.SolidMedium(capacity=1, conductivity=236e-6), charge=td.ChargeConductorMedium(conductivity=3.407e1), name="medium_al")
medium_w = td.MultiPhysicsMedium(heat=td.SolidMedium(capacity=1, conductivity=168e-6), charge=td.ChargeConductorMedium(conductivity=1.82e1))
medium_tin = td.MultiPhysicsMedium(heat=td.SolidMedium(capacity=1, conductivity=29e-6), charge=td.ChargeConductorMedium(conductivity=757576e-6))
medium_si = td.MultiPhysicsMedium(heat=td.SolidMedium(capacity=1, conductivity=148e-6), charge=td.ChargeInsulatorMedium())
medium_sio2 = td.MultiPhysicsMedium(heat=td.SolidMedium(capacity=1, conductivity=1.38e-6), charge=td.ChargeInsulatorMedium(), name="medium_sio2")
medium_sin = td.MultiPhysicsMedium(heat=td.SolidMedium(capacity=1, conductivity=18.5e-6), charge=td.ChargeInsulatorMedium())
medium_air_solid = td.MultiPhysicsMedium(heat=td.SolidMedium(capacity=1, conductivity=0.0263e-6), charge=td.ChargeInsulatorMedium())
medium_air_fluid = td.MultiPhysicsMedium(heat=td.FluidMedium(), charge=td.ChargeInsulatorMedium(), name="medium_air_fluid")

oxide = td.Structure(geometry=td.Box(center=(0, 0, 0), size=(50, 50, 50)), medium=medium_sio2)
tr = td.Structure(geometry=td.Box(center=(0, 0, 0), size=(20, 1, 0.11)), medium=medium_tin)
plus = td.Structure(geometry=td.Box(center=(-10, 0, 0), size=(2, 4, 1)), medium=medium_al, name="plus")
[...]

PR #2355, opened 2025-04-03
Fix heat conductivity plot

marc-flex:
This PR should address issue https://github.com/flexcompute/tidy3d/issues/2354

@alec-flexcompute do try this out


alec-flexcompute:
Works in the test case, thanks for the ultra-fast turnaround @marc-flex!

Issue #2307, opened 2025-03-12
Better repr representation of Tidy3D Medium when calling print


prashkh:
Currently when I look at the optical or electrical medium this is what I see. It's hard to say by looking at the PoleResidue model what this material is. For example. this is SiO2 from our material library. 

`{'optical': PoleResidue(attrs={}, name=None, frequency_range=(59958491600000.0, 1998616386666666.8), allow_gain=False, nonlinear_spec=None, modulation_spec=None, viz_spec=None, heat_spec=None, type='PoleResidue', eps_inf=1.5385442336875639, poles=(((-11504139.374277674-1.595196740783775e+16j), (7507685.43042605+4535416182817100j)), ((-249390.3565044153-172280738540723.53j), (46272.506981344035+99704543223121.88j)))),
 'electrical': Medium(attrs={}, name='SiO2', frequency_range=None, allow_gain=False, nonlinear_spec=None, modulation_spec=None, viz_spec=None, heat_spec=None, type='Medium', permittivity=3.9, conductivity=0.0)}`

daquinteroflex:
Yeah agreed! It'd be really cool if we can make our representations more graphical and even in a notebook to be more visually interactive in some form

tylerflex:
could try pretty print? https://docs.python.org/3/library/pprint.html

yaugenst-flex:
i'm not sure how much we want to do for `Medium` itself, since most of the information there can be quite useful when you're interested in actually printing the material model, so i'd second tyler's pprint suggestion (`rich.print` is even prettier and comes with tidy3d too..)

however, i think where we really might want to address this is in the material library. i looked through that a bit today and the data structures in there are not super obvious. ideally i think we would be able to `print(material_library)`, `print(material_library["mat"])`, `print(material_library["mat"]["variant"])`, with each of these printing relevant and easily digestible info. or at least have a `.summary()` method for each of these. however, with how the data structures currently work, this is not really possible. `material_library` is a `dict`, and `material_library["mat"]["variant"]` returns the underlying `td.Medium` (although this is only part of the datastructure, which is really a `VariantItem`).
since the material library is used heavily in GUI (and a ton of other places too i assume), i'm a bit hesitant to "overhaul" how these work, but i'm not completely sure what a good solution might look like either without doing that.
we could provide helper functions to do the pretty printing? although i doubt that would solve prash's issue..

[...]

PR #2356, opened 2025-04-03
feature[frontend]: improve printing of materials and material library

groberts-flex:
Improving of material printing to address feature request #2307.

```
import tidy3d as td

mat = td.material_library["W"]["RakicLorentzDrude1998"]
slab = td.Structure(
        geometry=td.Box(
            center=(0, 0, 0),
            size=(1, 1, 1),
        ),
        medium=mat,
    )

print("\n\nPrinting mat:")
print(mat)
print("\n\nPrinting slab with mat as medium:")
print(slab)

medium_dict = {"medium" : mat}
print("\n\nPrinting dict holding medium:")
print(medium_dict)

print("\n\n")
```

**Output:**

Printing mat:
```
td.PoleResidue(
	eps_inf=1.0, 
	poles=((0j, (4.2732115514080845e+17+0j)), ((-97233116671752.14+0j), (-4.2732115514080845e+17+0j)), ((-402605873718973.75-1471252666401400j), 7403002173803196j), ((-973090800441519.4-2745063931489722.5j), 1.2197111799530032e+16j), ((-2531099568361548-4814146946972908j), 2.9579221430831016e+16j), ((-4433222413252700-1.0493429699239636e+16j), 4.978330061510858e+16j)), 
	frequency_range=(24176811129032.258, 1208986804855426.0))
```

Printing slab with mat as medium:
```
attrs={} geometry=Box(attrs={}, type='Box', center=(0.0, 0.0, 0.0), size=(1.0, 1.0, 1.0)) name=None background_permittivity=None background_medium=None type='Structure' medium=W_RakicLorentzDrude1998
```

Printing dict holding medium:
```
{'medium': W_RakicLorentzDrude1998}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Classmethod `from_frequency_range` in `GaussianPulse` for generating a pulse whose amplitude in the frequency_range [fmin, fmax] is maximized, which is particularly useful for running broadband simulations.
 - Differentiable function `td.plugins.autograd.interpolate_spline` for 1D linear, quadratic, and cubic spline interpolation, supporting differentiation with respect to the interpolated values (`y_points`) and optional endpoint derivative constraints.
 - `SteadyEnergyBandMonitor` in the Charge solver.
+- Pretty printing enabled with `rich.print` for the material library, materials, and their variants. In notebooks, this can be accessed using `rich.print` or `display`, or by evaluating the material library, a material, or a variant in a cell.
 
 ### Changed
 - Performance enhancement for adjoint gradient calculations by optimizing field interpolation.
 - Auto grid in EME simulations with multiple `freqs` provided uses the largest instead of raising an error.
+- Named mediums now display by name for brevity; materials/variants print concise summaries including references.
 
 ### Fixed
 - Fixed `reverse` property of `td.Scene.plot_structures_property()` to also reverse the colorbar.
--- a/tests/test_material_library/test_material_library.py
+++ b/tests/test_material_library/test_material_library.py
@@ -1,0 +1,132 @@
+"""Tests material library functions and pretty printing"""
+
+import tidy3d as td
+from tidy3d.material_library.material_library import MaterialItemUniaxial
+
+
+def test_material_library_summary():
+    """Test to make sure we can print the material library without error."""
+    print(td.material_library)
+
+
+def test_material_summary():
+    """Test the string method for each material in the material library."""
+
+    for _, material in td.material_library.items():
+        print(material)
+
+
+def test_variant_summary():
+    """Test the string method for each variant in the material library."""
+
+    for _, material in td.material_library.items():
+        # graphene in the material library is run differently than the other materials and
+        # doesn't have the variant structure so we exclude any materials that are in this
+        # format
+        if hasattr(material, "variants"):
+            for variant in material.variants:
+                print(variant)
+
+
+def test_material_library_medium_repr():
+    """Test the new repr method does not error for material library variants."""
+
+    for material_key, material in td.material_library.items():
+        if hasattr(material, "variants"):
+            for variant_key in material.variants:
+                mat = td.material_library[material_key]
+                med = td.material_library[material_key][variant_key]
+                print(med)
+                print(repr(med))
+
+                if (type(med) is not td.MultiPhysicsMedium) and (
+                    type(mat) is not MaterialItemUniaxial
+                ):
+                    assert med.__repr__() == med.name, "Expected repr to return just the name"
+
+
+def test_medium_repr():
+    """Test the new repr method does not error for regular media with and without names and
+    that names are returned correctly by repr when they exist."""
+
+    material_name = "material"
+    test_media = [
+        td.Medium(permittivity=1.5**2),
+        td.Medium(permittivity=1.5**2, name=material_name),
+        td.material_library["SiO2"]["Horiba"].updated_copy(name=None),
+    ]
+    noname_medium_in_dict = {"medium": test_media[0]}
+
+    str_noname_medium = str(test_media[0])
+    repr_noname_medium = test_media[0].__repr__()
+    str_noname_medium_dict = str(noname_medium_in_dict)
+
+    assert (
+        "type='Medium' permittivity=2.25 conductivity=0.0" in str_noname_medium
+    ), "Expected medium information in string"
+    assert (
+        "Medium(attrs={}, name=None, frequency_range=None" in repr_noname_medium
+    ), "Expcted medium information in repr"
+    assert repr_noname_medium in str_noname_medium_dict, "Expected repr in dictionary string"
+
+    for medium in test_media:
+        repr_str = medium.__repr__()
+
+    assert (
+        test_media[1].__repr__() == material_name
+    ), "Expected repr to return just the material name."
+
+
+def test_variant_str():
+    """Test one of the materials for some expected output in variant printing."""
+
+    printed_SiO2 = str(td.material_library["SiO2"].variants["Horiba"])
+
+    assert "eps_inf: 1.0" in printed_SiO2, "Expected eps_inf in SiO2 printed string"
+    assert "poles: 1" in printed_SiO2, "Expected 1 pole in SiO2 printed string"
+
+    printed_SiO2_Palik_lossless = str(td.material_library["SiO2"].variants["Palik_Lossless"])
+
+    assert (
+        "eps_inf: 1.5385442336875639" in printed_SiO2_Palik_lossless
+    ), "Expected eps_inf in SiO2 printed string"
+    assert "poles: 2" in printed_SiO2_Palik_lossless, "Expected 1 pole in SiO2 printed string"
+
+    printed_SiO2_Palik_lossy = str(td.material_library["SiO2"].variants["Palik_Lossy"])
+
+    assert (
+        "eps_inf: 2.1560362571240765" in printed_SiO2_Palik_lossy
+    ), "Expected eps_inf in SiO2 printed string"
+    assert "poles: 5" in printed_SiO2_Palik_lossy, "Expected 1 pole in SiO2 printed string"
+
+
+def test_material_str():
+    """Test one of the materials for some expected output in variant printing."""
+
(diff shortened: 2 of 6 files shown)


tidy3d/material_library/util.py
```diff
@@ -0,0 +1,86 @@
+from typing import List
```
groberts-flex, line 2: thanks for starting me off with this util.py file @yaugenst-flex. Made some minor modifications, but overall it works well!

tidy3d/material_library/material_library.py
```diff
@@ -1926,7 +2046,13 @@ def medium(self, optical_axis: Axis):
...
 )
 
-material_library = dict(
+
+class MaterialLibrary(dict):
+    def __str__(self):
+        return summarize_material_library(self)
+
+
+material_library = MaterialLibrary(
```
yaugenst-flex: Neat! Didn't think of this approach, but that looks great.

tidy3d/material_library/material_library.py
```diff
@@ -70,6 +71,13 @@ class AbstractVariantItem(Tidy3dBaseModel):
         "model is fitted.",
     )
 
+    @property
+    def summarize_mediums(self) -> Dict[str, Union[PoleResidue, Medium2D, MultiPhysicsMedium]]:
+        return {}
```
yaugenst-flex, line 84: Bit confused by this property, what's its purpose?

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `fill` and `fill_structures` argument in `td.Simulation.plot_structures()` and `td.Simulation.plot()` respectively to disable fill and plot outlines of structures only.
 - New subpixel averaging option `ContourPathAveraging` applied to dielectric material boundaries.
 - A property `interior_angle` in `PolySlab` that stores angles formed inside polygon by two adjacent edges.
+- Methods for pretty printing the `material_library` as well as materials and their variants via their string methods.
+
+### Changed
+- Updated `__repr__` for `AbstractMedium` instances that have a specified name field to use that name instead of the full string representation of the medium to reduce clutter when printing models that have these mediums.
```
yaugenst-flex: I feel like this is a bit too technical for this change, maybe focus more on the impact this has on usability. Maybe something like "Named mediums now display by name for brevity; materials/variants print concise summaries including references."

tests/test_material_library/test_material_library.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def test_medium_repr():
+    """Test the new repr method does not error for regular media with and without names."""
+
+    test_media = [
+        td.Medium(permittivity=1.5**2),
+        td.Medium(permittivity=1.5**2, name="material"),
+        td.material_library["SiO2"]["Horiba"].updated_copy(name=None),
+    ]
```
yaugenst-flex: Not sure what this test does? `__repr__` is never called here no?

tidy3d/material_library/util.py
yaugenst-flex, line 1: Now that I'm thinking about this some more I feel like we should do two, optionally three things:
 1. A concise `__str__` representation that is pretty similar to the one currently. Maybe even shorter, like just a list of the material names. Not sure. Tbh if we have the below, might also just skip it entirely for the material library and just print the dict instead.
 2. A `.show()`, `.describe()`, or `.pretty_print()` (just throwing some names out there) method that prints a nicely formatted summary using `rich`. For the material library as a whole that could be e.g. a [table](https://rich.readthedocs.io/en/stable/tables.html). We already use `rich` in other places so this would not be an additionaly dependency.
 3. Optionally (but definitely nice), a `_repr_pretty_` method (see [here](https://ipython.readthedocs.io/en/stable/config/integrating.html#id1)) to ensure nicely formatted output in jupyter notebooks too. I think this can largely reuse whatever we do in 2.

tests/test_material_library/test_material_library.py
yaugenst-flex, line 1: These are mostly smoke tests, which are good to have but it'd be good if we would also test some functionality, e.g. if the material names are what we think they are, whether the `__repr__` returns the string, what it does for `None`...
The variants and material library are a bit more annoying, we should still check if e.g. some expected strings are in there or not.

yaugenst-flex (changes requested):

yaugenst-flex (changes requested):
Thanks @groberts-flex, this is much needed! I think the changes to the `Medium` work well as-is, and thanks for adding all the material names! Regarding pretty printing within the material library itself I added some more comments because I think we should try to make this output as pretty and useful as we can if we're already working on it. But all in all it's looking pretty good!

tidy3d/material_library/material_library.py
groberts-flex: This is to get the mediums that exist inside a variant. Usually, it's a medium, but in the case of uniaxial there are ordinary and extraordinary mediums. This helps display each of these by associating their material properties with a name for each of these.

tests/test_material_library/test_material_library.py
groberts-flex: oops thanks for the catch, changed to call __repr__

tests/test_material_library/test_material_library.py
groberts-flex: Sounds good, added more tests. I'm not sure if I'm following what you mean by "what it does for `None`"?

groberts-flex:
Material library rich printed output (works with display call and repr_pretty in Jupyter notebook as well)

image

groberts-flex:
image

Example material item and variant with rich printing (also works with display and _repr_pretty_ in Jupyter notebooks)

groberts-flex:
Example regular printing...
```
print(f"{td.material_library}\n\n")
print(f"{td.material_library["SiO2"]}\n\n")
print(td.material_library["SiO2"].variants["Horiba"])
```
image

image

tidy3d/material_library/util.py
groberts-flex: I believe I added all of these in the most recent set of changes, but let me know if I missed something! I included some screenshots below to show what some of this looks like for different styles of printing. There's regular printing, rich printing, and _repr_pretty_ that uses the rich output for nice Jupyter notebook printing!

groberts-flex:
image

Added lines between table entries

tests/test_material_library/test_material_library.py
yaugenst-flex: Ah I meant if the medium doesn't have a name, then the default is `None`. What does that look like?

yaugenst-flex:
> Added lines between table entries

Ok I'll let you decide here but I actually think the version without the horizontal lines looks much cleaner 😄

yaugenst-flex (approved):
Thanks @groberts-flex I think this looks incredible!

CHANGELOG.md
```diff
@@ -14,10 +14,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - 'ModeSimulation' now supports 'PermittivityMonitor'.
 - Classmethod `from_frequency_range` in `GaussianPulse` for generating a pulse whose amplitude in the frequency_range [fmin, fmax] is maximized, which is particularly useful for running broadband simulations.
 - Differentiable function `td.plugins.autograd.interpolate_spline` for 1D linear, quadratic, and cubic spline interpolation, supporting differentiation with respect to the interpolated values (`y_points`) and optional endpoint derivative constraints.
+- Pretty printing enabled for the material library, materials, and their variants.
```
yaugenst-flex: Could you add a note on how this is used? For example in notebooks, would you just use `print`?

groberts-flex:
> > Added lines between table entries
> 
> Ok I'll let you decide here but I actually think the version without the horizontal lines looks much cleaner 😄

Yeah, I agree too! I put the lines in in case it helped people line up entries a bit better, but aesthetically I like the other one better and like that it's more compact. I'll switch back!

CHANGELOG.md
groberts-flex: Yes, will do. There are a few ways this will pop up in notebooks. For example, with the material library, if you call print(td.material_library) you will get the non-rich print. But if you call
rich.print(td.material_library), display(td.material_library), or just evaluate a cell with td.material_library in it (such that repr would normally get called), you will see the pretty print output

tests/test_material_library/test_material_library.py
groberts-flex: I see, good point! It just falls back to the repr from before. We could override in this case as well - possibly we could  pick a couple of relevant properties of the material to use as summarizing information. For now, I added a case to make sure the None case is falling back to default repr

yaugenst-flex: nah that sounds good, just wanted to make sure that case is handled

CHANGELOG.md
yaugenst-flex: sounds good, yeah. I think in a notebook context you'd just `td.material_library` etc

daquinteroflex (approved):
Nice! Thanks Greg and Yannick for this pretty big effort, looks quite nice! I also agree without horizontal lines looks nicer. Just worth mentioning maybe when we start moving more towards the `MultiPhysicsMedium` library, we might have to build upon this a bit more - but should be able to use a lot of this. My only tiny suggestion is that it'd be nice if the rich colors could also be the same color scheme as the new website maybe? Maybe on 2.9 we actually update the docs to those new colors too.

groberts-flex:
> Nice! Thanks Greg and Yannick for this pretty big effort, looks quite nice! I also agree without horizontal lines looks nicer. Just worth mentioning maybe when we start moving more towards the `MultiPhysicsMedium` library, we might have to build upon this a bit more - but should be able to use a lot of this. My only tiny suggestion is that it'd be nice if the rich colors could also be the same color scheme as the new website maybe? Maybe on 2.9 we actually update the docs to those new colors too.

Thanks Dario, that makes sense wrt the `MultihysicsMedium` library! I'm happy to update the color schemes - is there a place I can get the color info that we use on the new website (i.e. - hex colors or rgb values)?

tomflexcompute (approved):
This is very nice. Thanks to the great efforts @groberts-flex and all. I wish we have pretty prints everywhere in tidy3d.

groberts-flex:
updated with the Flexcompute colors (thanks to Dario sending me those via slack!). 

image

image

prashkh:
Thanks a lot for working on this! Really appreciate it!

prashkh (approved):

PR #2359, opened 2025-04-04
Better validation of TFSF sources, and enabling broadband

momchil-flex:


tidy3d/components/source/field.py
```diff
@@ -700,6 +700,16 @@ class TFSF(AngledFieldSource, VolumeSource):
         "to the ``injection_axis`` by ``angle_theta`` and ``angle_phi``.",
     )
 
+    num_freqs: int = pydantic.Field(
```
dmarek-flex: Might be simpler to have `BroadbandSource` default to `num_freqs=3`, and then make the special case `ModeSource` which uses 1 currently. Although not sure of the exact purpose of this difference.

dmarek-flex (approved):
Looks good

tidy3d/components/source/field.py
momchil-flex: Yeah fair enough. The reason it's 1 in the ModeSource is that if there are e.g. mode crossings or disappearing modes somewhere in the source frequency range, the results will be messed up everywhere. So it's kind of dangerous. For analytic sources, there's no such danger.

PR #2362, opened 2025-04-04
feat: autograd support for rotated Box

rahul-flex:
Autograd support for the `Transformed` class when the base geometry is a `Box` under rotation. Key changes are:

**Rotated Normals:**
Box face derivatives now accept rotation matrix, ensuring that the face derivative computed use rotated normals.

**Derivative Computation:**
The derivative with respect to rotated faces is computed using the surface mesh and gradient surfaces.

**Testing:**
Finite difference checks to validate autograd computations.
Emulated simulation test cases to verify rotation.
Test results attached for 7 cases: box, box with background structure, 0 deg rotation, 90 deg rotation about z, 45 deg about y, 45 deg about x, and 45 deg about z.
rel_error_center-x
rel_error_center-y
rel_error_center-z
rel_error_size-x
rel_error_size-y
rel_error_size-z

### DIFF
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -10,7 +10,6 @@ from typing import Any, Callable, List, Tuple, Union
 import autograd.numpy as np
 import pydantic.v1 as pydantic
 import shapely
-import xarray as xr
 
 try:
     from matplotlib import patches
@@ -28,7 +27,10 @@ from ...exceptions import (
 from ...log import log
 from ...packaging import check_import, verify_packages_import
 from ..autograd import AutogradFieldMap, TracedCoordinate, TracedSize, get_static
-from ..autograd.derivative_utils import DerivativeInfo, integrate_within_bounds
+from ..autograd.derivative_utils import (
+    DerivativeInfo,
+    DerivativeSurfaceMesh,
+)
 from ..base import Tidy3dBaseModel, cached_property
 from ..transformation import RotationAroundAxis
 from ..types import (
@@ -61,6 +63,7 @@ from ..viz import (
 )
 
 POLY_GRID_SIZE = 1e-12
+_NUM_PTS_DIM_BOX_FACE = 200
 
 
 _shapely_operations = {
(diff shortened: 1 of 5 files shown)


tidy3d/web/api/autograd/autograd.py
```diff
@@ -1013,32 +1013,28 @@ def postprocess_adj(
                 eps_background = None
 
             # manually override simulation medium as the background structure
-            if not isinstance(structure.geometry, td.Box):
```
yaugenst-flex, line 1016: Why was this removed? Because `Box` would use the same auto-detection for permittivity as the other structures now?

tidy3d/components/geometry/base.py
```diff
@@ -1637,6 +1640,125 @@ def __invert__(self):
             operation="difference", geometry_a=Box(size=(inf, inf, inf)), geometry_b=self
         )
 
+    def build_box_face_mesh(
```
yaugenst-flex, line 1643: So if I understand this function correctly it currently assumes that the box's center is at the origin? What happens if it's not?

tidy3d/components/geometry/base.py
```diff
@@ -2522,18 +2647,26 @@ def derivatives_center_size(vjps_faces: Bound) -> dict[str, Coordinate]:
...
 
         # change in permittivity between inside and outside
         vjp_faces = np.zeros((2, 3))
 
         for min_max_index, _ in enumerate((0, -1)):
             for axis in range(3):
+                if rotation_matrix is not None:
+                    rotation_matrix = rotation_matrix
+                else:
+                    rotation_matrix = None
```
yaugenst-flex: This looks strange? Lines 2660 & 2661 don't appear to do anything?

tidy3d/components/geometry/base.py
```diff
@@ -2923,6 +2982,45 @@ def _update_from_bounds(self, bounds: Tuple[float, float], axis: Axis) -> Transf
...
+
+            transformed_center_gradient = np.array(
+                transformed_geometry_derivatives.get(("center",), (0.0, 0.0, 0.0))
+            )
+            transformed_size_gradient = np.array(
+                transformed_geometry_derivatives.get(("size",), (0.0, 0.0, 0.0))
+            )
+            derivative_map[("geometry", "center")] = transformed_center_gradient
+            derivative_map[("geometry", "size")] = transformed_size_gradient
+        derivative_map[("transform",)] = np.zeros((4, 4))
```
yaugenst-flex: So this returns a zero array even though it's not really implemented right? Is this field needed for book-keeping and used somewhere else?

tidy3d/components/geometry/base.py
```diff
@@ -61,6 +63,7 @@
 )
 
 POLY_GRID_SIZE = 1e-12
+_NUM_PTS_DIM_BOX_FACE = 200
```
yaugenst-flex: Seems a bit strange to let all box faces always have the same no. of points..

tidy3d/components/geometry/base.py
```diff
@@ -1637,6 +1640,125 @@ def __invert__(self):
...
+        perps2 = np.tile(t2_local, (centers.shape[0], 1))
+
+        surface_mesh = DerivativeSurfaceMesh(
+            centers=centers,
+            areas=areas,
+            normals=normals,
+            perps1=perps1,
+            perps2=perps2,
+        )
+        return surface_mesh, n_local
```
yaugenst-flex: What's `n_local` used for? It seems to be discarded later.

yaugenst-flex (commented):
Thanks @rahul-flex! Left some comments. I think the biggest thing we should think about is if maybe now would be a good time to consider reusing the polyslab routines for `Box` too, since #2418 introduces more robust polyslab handling. That code handles a lot of edge cases that we currently don't handle for `Box`, such as intersection with the simulation bounds support for 2d simulations, and adaptive grid sampling.

tidy3d/components/geometry/base.py
rahul-flex: Steps in the function are:
1. takes box center and size, then it generates face corners

-  It considers center location for generation of corners. So it should work for arbitrary centers. https://github.com/flexcompute/tidy3d/blob/rahul-flex/rotated-box-derivatives/tidy3d/components/geometry/base.py#L1691-L1692

2. uses rotation matrix to rotate the face corners and generates points for integration:

- here rotation is always about global origin cause the rotation in forward pass is always about the global origin:

https://github.com/flexcompute/tidy3d/blob/rahul-flex/rotated-box-derivatives/tidy3d/components/transformation.py#L114-L127

tidy3d/web/api/autograd/autograd.py
rahul-flex: Yes—that special-case guard was removed because we now auto-detect the background permittivity for all geometries, including td.Box. In other words, every compute-derivatives path (the ones that use grad_surfaces) will pull the correct permittivity from the surrounding medium automatically. With the condition check in place,  we’d end up with the wrong background permittivity.

tidy3d/components/geometry/base.py
rahul-flex: Thanks for catching that. It does not do anything. I will remove it.

tidy3d/components/geometry/base.py
rahul-flex: Yes, it not doing anything when transform parameters are not differentiable. I will remove it.

tidy3d/components/geometry/base.py
rahul-flex: I followed along the older polyslab surface integration, which had a constant number of points for all faces. Should we change it?

tidy3d/components/geometry/base.py
rahul-flex: Yes, currently it is not being used. In the next PR: derivative wrt rotation angle, we need n_local. That uses the same function build_box_face_mesh(), that's why kept the signature same.

yaugenst-flex:
Closing in favor of https://github.com/flexcompute/tidy3d/pull/2475/

@rahul-flex maybe just make sure that these changes are in sync? not sure if there are updates on this branch that havent been propagated to the other PR yet?

PR #2370, opened 2025-04-09
Change doping ramp-up default

marc-flex:


momchil-flex (approved):
Thanks. Needs a changelog, but I can add it in my 2_8_2 PR where I need to update the changelog anyway.

PR #2371, opened 2025-04-09
Validating that bend_radius is strictly larger than half the mode plane size

momchil-flex:
The case when the two are equal could actually be particularly problematic as the bend center lies on a grid point, which is a singularity in our coordinate transformation.

So maybe I should even leave a bit of buffer extra?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- The `bend_radius` for a mode solve must be strictly larger than half the mode plane size (previously it could be equal).
+
 ## [2.8.2] - 2025-04-09
 
 ### Added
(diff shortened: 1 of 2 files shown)


weiliangjin2021:
as for extra buffer, you mean
`np.abs(mode_spec.bend_radius) <= plane.size[radial_ax] / 2 * (1-FP_EPS)`

weiliangjin2021 (dismissed):

momchil-flex:
Yeah something like that, I'm also wondering if it should even be larger than `FP_EPS` though. Basically in the trasnform [here](https://github.com/flexcompute/tidy3d/blob/3dc27f16bc272101d285df3b2f4bb3a42deaef2b/tidy3d/components/mode/transforms.py#L47-L57), the jacobian has a division by zero if the radius lies on a grid coordinate (either primal or dual). So to avoid some very large values, might be better to require it to be even a bit further away.

weiliangjin2021:
The safest is probably to use those `np.nextafter` (e.g. in this function https://github.com/flexcompute/tidy3d/blob/3dc27f16bc272101d285df3b2f4bb3a42deaef2b/tidy3d/components/geometry/utils_2d.py#L18-L21 from @dmarek-flex )

PR #2372, opened 2025-04-09
fix: cleanup temp figures after running anisotropic epsilon plotting tests

bzhangflex:
Fixes "folder not found" error when cleaning up figures temporarily created by matplotlib for anisotropic epsilon plotting tests.
Also added another fixture for closing pyplot figures after each test is ran, which should save memory when all the tests are ran together.

### DIFF
--- a/tests/test_components/test_simulation.py
+++ b/tests/test_components/test_simulation.py
@@ -1,7 +1,6 @@
 """Tests the simulation and its validators."""
 
-import shutil
-from pathlib import Path
+import uuid
 
 import gdstk
 import matplotlib.pyplot as plt
@@ -9,7 +8,7 @@ import numpy as np
 import pydantic.v1 as pydantic
 import pytest
 import tidy3d as td
-from matplotlib.testing.decorators import check_figures_equal
+from matplotlib.testing.compare import compare_images
 from tidy3d.components import simulation
 from tidy3d.components.scene import MAX_GEOMETRY_COUNT, MAX_NUM_MEDIUMS
 from tidy3d.components.simulation import MAX_NUM_SOURCES
@@ -744,12 +743,6 @@ class TestAnisotropicPlotting:
 
         return td.CustomAnisotropicMedium(xx=medium_xx, yy=medium_yy, zz=medium_zz)
 
-    @pytest.fixture(scope="module", autouse=True)
-    def cleanup_figures(self):
-        yield  # Run tests first
-        fig_dir = Path().resolve() / "result_images"
-        shutil.rmtree(fig_dir, ignore_errors=False)
-
     def make_sim(self, medium):
         L = 5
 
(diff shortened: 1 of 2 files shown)


tests/test_components/test_simulation.py
```diff
@@ -748,7 +749,14 @@ def medium_customani(self):
...
         fig_dir = Path().resolve() / "result_images"
-        shutil.rmtree(fig_dir, ignore_errors=False)
+        if os.path.exists(fig_dir):
+            shutil.rmtree(fig_dir, ignore_errors=False)
+
+    @pytest.fixture(autouse=True)
+    def close_figures(self):
+        """close figures after each matplotlib test"""
+        yield
+        plt.close()
```
yaugenst-flex: Should we maybe just change the scope of [this fixture](https://github.com/flexcompute/tidy3d/blob/e1cd386fa400b4cc7c31944142d7175b72dd5957/tests/conftest.py#L22-L24) instead?

tests/test_components/test_simulation.py
```diff
@@ -748,7 +749,14 @@ def medium_customani(self):
     def cleanup_figures(self):
         yield  # Run tests first
         fig_dir = Path().resolve() / "result_images"
-        shutil.rmtree(fig_dir, ignore_errors=False)
+        if os.path.exists(fig_dir):
+            shutil.rmtree(fig_dir, ignore_errors=False)
```
yaugenst-flex: This still seems problematic to me. If you have 2 processes running tests, then 2 of them will be writing files to this directory. Once one finishes, it will run this fixture and remove the directory, leaving the other process with no directory to work with. I guess this only "works" if the other process is also already done with writing figures to file. This is making me think it might be better to use [`image_comparison`](https://matplotlib.org/stable/api/testing_api.html#matplotlib.testing.decorators.image_comparison) instead of `check_figures_equal` because that one gives direct control over the image paths and we can just use [`tmp_path`](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://docs.pytest.org/en/stable/how-to/tmp_path.html&ved=2ahUKEwiLzOq35MyMAxXI2gIHHTzZKeoQFnoECAkQAQ&usg=AOvVaw0tHQ_7MW_l2b_TDg6xFsPl).

yaugenst-flex (changes requested):

tests/test_components/test_simulation.py
bzhangflex: I didn't know about that fixture haha, will give it a try

tests/test_components/test_simulation.py
bzhangflex: That does seem better, thanks. I'll give it a try

bzhangflex: re-implemented the tests with image_comparison instead of check_figures_equal

tests/test_components/test_simulation.py
```diff
@@ -772,6 +766,29 @@ def make_sim(self, medium):
...
+
+        # plot and save epsilon component 2
+        fname2 = tmp_path / Path(str(uuid.uuid4()) + ".png")
+        f2, ax2 = plt.subplots()
+        sim.plot_eps(x=0, eps_component=eps_comp2, ax=ax2)
+        f2.savefig(fname2)
+
+        # close the figures now that they have been saved to images
+        plt.close(f1)
+        plt.close(f2)
```
yaugenst-flex: I guess this should be removed since the fixture is taking care of it?

tests/test_components/test_simulation.py
```diff
@@ -772,6 +766,29 @@ def make_sim(self, medium):
             run_time=1e-12,
         )
 
+    def compare_eps_images(self, tmp_path, eps_comp1, eps_comp2, expected, medium):
+        """Asserts that two epsilon component plots are different"""
+        sim = self.make_sim(medium)
+
+        # plot and save epsilon component 1
+        fname1 = tmp_path / Path(str(uuid.uuid4()) + ".png")
```
yaugenst-flex: `fname1 = tmp_path / (str(uuid.uuid4()) + ".png")`

no `Path` instance necessary

yaugenst-flex (commented):
Thanks @bzhangflex this is much better!

tests/test_components/test_simulation.py
bzhangflex: I found the fixture wasn't handling it so I added those in

bzhangflex: I think because the fixture only closes the last active figure, but in these tests I open two figures at a time. Actually I could probably just reuse the same figure object for both plots

bzhangflex: hmm reusing the same figure causes some errors, not sure why. Probably easiest to keep it like this

yaugenst-flex: `plt.close("all")` in the fixture?

bzhangflex: that works for me locally, I'm just worried that if multiple tests are running in parallel, figures might be closed before tests are done

yaugenst-flex: there is only ever one test running at a time on a single worker, closing plots on one worker does not affect others

yaugenst-flex (approved):

PR #2375, opened 2025-04-10
Add effective DOS classes for the Charge solver

vasilyzabelin:
Add effective DOS classes for the Charge solver


tidy3d/components/tcad/effective_DOS.py
```diff
@@ -0,0 +1,159 @@
...
+import numpy as np
+import pydantic.v1 as pd
+
+from tidy3d.components.base import Tidy3dBaseModel
+
+from ...exceptions import DataError
+
+# constants definition
+k_B = 1.380649e-23  # Boltzmann constant in J/K
+m_e = 9.1093837139e-31  # electron mass in kg
```
marc-flex: we actually have some constants already here: https://github.com/flexcompute/tidy3d/blob/0b68e6d30229b28ab73c3d7882212f81a87d39c7/tidy3d/constants.py#L49

If you need additional feel free to add them there. Careful with the units though. Tidy3d works in $\mu m$ so most units will be modified accordingly. In particular, in charge we have densities in the usual units (#/$cm^3$) but distances are in microns

tidy3d/components/tcad/effective_DOS.py
```diff
@@ -0,0 +1,159 @@
...
+
+        return self._calc_eff_DOS(T)
+
+    def get_effective_DOS_derivative(self, T: float):
+        if T <= 0:
+            raise DataError(
+                f"Incorrect temperature value ({T}) for the effectve density of states calculation."
+            )
+
+        return self._calc_eff_DOS_derivative(T)
```
marc-flex: not sure we'll need these here since we formulation is in the back-end. We can leave them there for now though.  We can decide once we know how this all will play out in the back-end.

tidy3d/components/tcad/effective_DOS.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class IsotropicEffectiveDOS(EffectiveDOS):
+    """Effective density of states model that assumes single valley and isotropic effective mass.
+    The model assumes the standard equation for the 3D semiconductor with parabolic energy dispersion:
+
+    .. math::
+
+        \\begin{equation}
+             \\mathbf{N_eff} = 2 * (\\m_eff \\m_e \\k_B \\T / (2 \\pi \\hbar^2))^(3/2)
```
marc-flex: why the double slash before m_eff, _e, k_B and T?

tidy3d/components/tcad/effective_DOS.py
```diff
@@ -0,0 +1,159 @@
...
+        \\end{equation}
+    """
+
+    m_eff: pd.NonNegativeFloat = pd.Field(
```
marc-flex: shouldn't this be a positive float? Is the mass allowed to be 0?

marc-flex (changes requested):
Thanks Vasily for dealing with this. My main concern is that except for the `ConstantEffectiveDOS` model which is basically a float, the rest are variations of the same model. 

In particular, ´IsotropicEffectiveDOS´ and `MultiValleyEffectiveDOS` can be combined into a single model in which we change the effective mass model. That is, the effective mass should not be a float but rather a model. This is also explained [here](https://github.com/flexcompute/tidy3d-core/issues/934) in points 1 to 3. That way we can also add temperature dependence to the effective mass.

`DualValleyEffectiveDOS` can stay as is except that we should probably add the masses as models to allow temperature dependence.

tidy3d/components/tcad/effective_DOS.py
vasilyzabelin: Yes, it must be a positive value. I will change it.

tidy3d/components/tcad/effective_DOS.py
vasilyzabelin: It's a copy-paste typo. Just a placeholder description here. I will update it later for the final release.

tidy3d/components/tcad/effective_DOS.py
vasilyzabelin: So far I added the temperature derivative for experimentation with the non-isothermal simulations. We should think about the final implementation later together with the model development.

tidy3d/components/tcad/effective_DOS.py
vasilyzabelin: I can move the constants to the common file.

vasilyzabelin:
I see here two ways: several simple DOS models that covers traditional cases (this is implemented now), or one complex model that cover all cases, but require many parameters. It's a matter of personal choice what to prefer.

I don't think that we need to implement a special model for the effective mass (at least now). I've never seen that people do fine tuning of the effective mass as a function of the temperature, I believe it is a negligible effect. More useful case is the mass dependence on the strain in the material, this is important for quantum wells and for some Si transistor simulations. But we are not going to cover these cases now. So I think that we may stay with the constant effective mass assumption now and change it later, if necessary.

tidy3d/components/tcad/effective_DOS.py
vasilyzabelin: I updated the used constants, but the electron mass definition is quite exotic (in order to be compatible with tidy3D constant's units), so it is better not to expose it to users.

marc-flex:
`MultiValleyEffectiveDOS` is non isotropic, right?

`ConstantEffectiveDOS` I think can be removed since we already accept floats for the effective DOS. 

So I think we should definitely merge `IsotropicEffectiveDOS` and `MultiValleyEffectiveDOS`. This class could be called `EffectiveDOS` and current `EffectiveDOS` could be named `BaseEffectiveDOS`.

With regards to `DualValleyEffectiveDOS`: this is basically a summation of isotropic `EffectiveDOS` with one term per valley. So in the effective density of states ( [N_c](https://github.com/flexcompute/tidy3d/blob/74d5be32de7a15670273416c50246608cf6cc45c/tidy3d/components/material/tcad/charge.py#L259 )) we could accept also a tuple of `EffectiveDOS` where `N_valley=1` for each `EffectiveDOS`.

So, in summary:

- `EffectiveDOS` -> `BaseEffectiveDOS`
- `IsotropicEffectiveDOS` + `MultiValleyEffectiveDOS` -> `EffectiveDOS`
- Remove `DualValleyEffectiveDOS` and validate that when [N_c](https://github.com/flexcompute/tidy3d/blob/74d5be32de7a15670273416c50246608cf6cc45c/tidy3d/components/material/tcad/charge.py#L259 ) is a tuple of `EffectiveDOS` , `N_valley=1`

Does that make sense? 

What do you think @dbochkov-flexcompute @momchil-flex

vasilyzabelin:
This is basically goes down to choose between a list of several simple models, that are traditional for semiconductor physics and device simulation, or to implement one flexible class that can cover all models. I, personally, like the first option, but the second one is also good in terms of the program logic.

I can redesign the classes and merge everything into one universal class, if this is our common decision.

What is your opinion @momchil-flex and @dbochkov-flexcompute ?

tidy3d/components/tcad/effective_DOS.py
```diff
@@ -0,0 +1,158 @@
...
+um_3_to_cm_3 = 1e12  # conversion factor from micron^(-3) to cm^(-3)
+
+DOS_aux_const = 2.0 * np.power((m_e_eV * K_B) / (2 * np.pi * HBAR * HBAR), 1.5) * um_3_to_cm_3
+
+
+class EffectiveDOS(Tidy3dBaseModel, ABC):
+    """Abstract class for the effective density of states"""
+
+    @abstractmethod
+    def _calc_eff_DOS(self, T: float):
```
dbochkov-flexcompute: lower case all letters `_calc_eff_dos` here and everywhere else

tidy3d/components/tcad/effective_DOS.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+class MultiValleyEffectiveDOS(EffectiveDOS):
+    """Effective density of states model that assumes multiple equivalent valleys and anisotropic effective mass.
+    The model assumes the standard equation for the 3D semiconductor with parabolic energy dispersion:
+
+    .. math::
+
+        \\begin{equation}
+             \\mathbf{N_eff} = 2 * N_valley (\\frac{(m_{eff_long} * m_{eff_trans} * m_{eff_trans})^(1/2) * m_e * k_B * T}{2 \\pi * \\hbar^2})^(3/2)
```
dbochkov-flexcompute: does this have right units? I see that `(m_{eff_long} * m_{eff_trans} * m_{eff_trans})^(1/2)` comes instead of `m_eff` compared to `IsotropicEffectiveDOS`

dbochkov-flexcompute (commented):
regarding class organization, no strong preference, but slightly lean towards having it more explicit, that is, keeping separate `IsotropicEffectiveDOS`, `MultiValleyEffectiveDOS`, and `DualValleyEffectiveDOS`. I imagine the simpler the model the more usage it could potentially get? In that sense, it's maybe not worth making the user to define an isotropic effective dos by setting up twice the same parameter (if I understand this correctly):
```
medium = SemiconductorMedium(
    Nc=MultiValleyEffectiveDOS(m_eff_long=1, m_eff_trans=1, N_valley=1),
    ...
)
```
instead of

```
medium = SemiconductorMedium(
    Nc=IsotropicEffectiveDOS(m_eff=1),
    ...
)
```
Also, if we implement dual valley effective dos as a tuple, then we would have to provide additional documentation that this is how you set it up and make sure the user can easily find that information. With having explicit `DualValleyEffectiveDOS` I think it should be easier for the user to understand that this is an option and how to set it up

tidy3d/components/tcad/effective_DOS.py
```diff
@@ -0,0 +1,158 @@
...
+import numpy as np
+import pydantic.v1 as pd
+
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.constants import C_0, HBAR, K_B
+
+from ...exceptions import DataError
+
+# constants definition
+m_e_C_square = 0.51099895069e6  # (electron mass * C_0^2) in eV
```
marc-flex, line 12: is that constant used anywhere?

vasilyzabelin:
I have updated the code according to your comments. As far as I understand, Daniil also prefers to keep several simple classes instead of a single complex one,
@dbochkov-flexcompute @marc-flex Are you OK with the current version of the code?

marc-flex:
I am mostly OK with the code. I still see, however, an unanswered question by Daniil (regarding units). 

I guess my only comment would be about the `ConstantEffectiveDOS` class which I think it's redundant since we can define Nc and Nv with a float. So I guess the options would be to either remove `ConstantEffectiveDOS` or remove the possibility of defining Nc/Nv with floats (i.e., force Nc&Nv to be `EffectiveDOS`

vasilyzabelin:
I personally prefer to remove possibility to define the DOS using float. It is more consistent when we have a special type for the effective DOS and it simplifies the code by removing additional checks of the type of the passes parameters. And I believe that most of the users will use the temperature-dependent DOS definition anyway.
The only negative side is backward compatibility, that will be broken if we remove possibility to use float numbers. But because we are still in the early stage of the device simulation development and probably there are not so many external users, we can make this change.

tidy3d/components/tcad/effective_DOS.py
vasilyzabelin: I have updated the code to resolve this issue

marc-flex:
> And I believe that most of the users will use the temperature-dependent DOS definition anyway.

That's fair. So let's remove the possibility of defining Nc/Nv with a float. We still need to check types in the back-end though.

> The only negative side is backward compatibility, that will be broken if we re...

Yeah, don't worry about compatibility at this stage.

momchil-flex:
This can be closed right @marc-flex

marc-flex:
> This can be closed right @marc-flex

Yes. This has already been merged onto non isothermal

momchil-flex:
Closing in favor of #2425

PR #2377, opened 2025-04-11
feat: plot3d support stl

majinge7846:


yaugenst-flex (changes requested):
Thanks @majinge7846! I think we can avoid the file I/O here by writing to an in-memory buffer. We do something similar for hashing [here](https://github.com/flexcompute/tidy3d/blob/0b68e6d30229b28ab73c3d7882212f81a87d39c7/tidy3d/components/base.py#L138-L142). In practice I think this would look something like:
```py
hdf5_buffer = io.BytesIO()
sim.to_hdf5(hdf5_buffer)
hdf5_bytes = hdf5_buffer.getvalue()
base64_encoded = base64.b64encode(hdf5_bytes)
base64 = base64_encoded.decode("utf-8")
```

This would be faster and remove the need for cleanup. Could you test this on your end?

majinge7846:
@lucas-flexcompute @heitzmann  it seems that to to_hdf5_gz can speed up too? https://github.com/flexcompute/tidy3d/blob/8a836435f83ff728486bb71b0bb3bff7c61691f3/tidy3d/components/base.py#L666

lucas-flexcompute:
@yaugenst-flex From my understanding of the docs, using BytesIO seems have some issues (not clear what, though). Maybe we should upgrade the dependency to 3.13 and use their own memory files: https://docs.h5py.org/en/stable/high/file.html#in-memory-files.

yaugenst-flex:
@lucas-flexcompute that's a great suggestion, I wasn't aware of `h5py.File.in_memory()`. I agree we should probably use that, however it would require modifying `Tidy3dBaseModel.to_hdf5()` because we'd need to make sure the file handle isn't lost once the context manager exits. Maybe it's easier to stick with `BytesIO` for this PR and move to in-memory hdf5 files in a separate PR?

lucas-flexcompute:
> @lucas-flexcompute that's a great suggestion, I wasn't aware of `h5py.File.in_memory()`. I agree we should probably use that, however it would require modifying `Tidy3dBaseModel.to_hdf5()` because we'd need to make sure the file handle isn't lost once the context manager exits. Maybe it's easier to stick with `BytesIO` for this PR and move to in-memory hdf5 files in a separate PR?

Sure, no rush at all with this, as I don't think the `BytesIO` caveats really affect us right now. The comment was more to let you know the option exists for the future (I also just found out about it a couple of weeks ago).

yaugenst-flex (approved):

PR #2379, opened 2025-04-11
fix: lazy import scipy and matplotlib

yaugenst-flex:
MC ran into an issue with this since scipy and matplotlib are not installed in the minimal tidy3d install and tidy3d still needs to be importable.

@caseyflex @dmarek-flex fyi

### DIFF
--- a/tidy3d/components/dispersion_fitter.py
+++ b/tidy3d/components/dispersion_fitter.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 from typing import Optional, Tuple
 
 import numpy as np
-import scipy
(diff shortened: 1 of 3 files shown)


yaugenst-flex:
hmm actually the mode solver has many more scipy dependencies, not really sure how we should handle that

yaugenst-flex:
according to @majinge7846 might be fine as is but we can keep the PR as draft for now in case anything else comes up

frederikschubertflex:
@yaugenst-flex  At least the `scipy` issue should be resolved by https://github.com/flexcompute/tidy3d/pull/2543?

yaugenst-flex:
Yup this is actually just a dummy PR for MC, we can close.

PR #2380, opened 2025-04-11
enh[adjoint]: speed up interpolation of `evaluate_flds_at`

yaugenst-flex:
This replaces the xarray interpolation (which is comparatively slow) used in `evaluate_flds_at` with an equivalent approach using scipy's [`map_coordinates`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.map_coordinates.html).

Benchmarks using the [metalens notebook](https://github.com/flexcompute/tidy3d-notebooks/blob/develop/Autograd7Metalens.ipynb).

Before:
image

After:
image

### DIFF
--- a/tidy3d/components/autograd/derivative_utils.py
+++ b/tidy3d/components/autograd/derivative_utils.py
@@ -313,26 +313,39 @@ class DerivativeInfo(Tidy3dBaseModel):
     ) -> dict[str, ScalarFieldDataArray]:
         """Compute the value of an dict with keys Ex, Ey, Ez at a set of spatial locations."""
 
-        xs, ys, zs = spatial_coords.T
-        edge_index_dim = "edge_index"
+        from scipy.interpolate import RegularGridInterpolator
 
-        interp_kwargs = {}
-        for dim, locations_dim in zip("xyz", (xs, ys, zs)):
-            # filter out any infinity values to use LARGE_NUMBER and get 0 in the interp()
-            locations_dim = np.nan_to_num(locations_dim, posinf=LARGE_NUMBER, neginf=-LARGE_NUMBER)
-            interp_kwargs[dim] = xr.DataArray(locations_dim, dims=edge_index_dim)
-
+        coords = np.nan_to_num(spatial_coords, posinf=LARGE_NUMBER, neginf=-LARGE_NUMBER)
         components = {}
-        for fld_name, arr in fld_dataset.items():
-            arr_interp = arr.interp(
-                **interp_kwargs,
-                assume_sorted=True,
-                kwargs=dict(fill_value=None, bounds_error=False),
-            )
-            if "f" in arr_interp.coords:
-                arr_interp = arr_interp.sum("f")
+        edge_index_dim = "edge_index"
+        n_points = coords.shape[0]
 
-            components[fld_name] = arr_interp
+        for fld_name, arr in fld_dataset.items():
+            data = arr.values
+            points = tuple(arr.coords[dim].values for dim in "xyz")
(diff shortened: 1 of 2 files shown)


tidy3d/components/autograd/derivative_utils.py
```diff
@@ -313,26 +313,49 @@ def evaluate_flds_at(
...
+                    elif f_dim_idx == 1:
+                        data_slice = data[:, f_idx, :, :]
+                    elif f_dim_idx == 2:
+                        data_slice = data[:, :, f_idx, :]
+                    else:
+                        data_slice = data[:, :, :, f_idx]
+
+                    result += map_coordinates(
+                        data_slice, coords_for_interp, order=1, mode="nearest", prefilter=False
+                    )
```
yaugenst-flex: I am aware this chunk looks pretty horrible but this seemed to be by far the most performant.

momchil-flex: The if-else bit just looks like `np.take(data, f_idx, axis=f_dim_idx)`, did you try that? I would be really surprised if it's slower.

momchil-flex, line 337: Curious, why `map_coordinate` as opposed to scipy's `interp` methods? I mean the latter are used by xarray under the hood so there's probably a reason but I hadn't heard about `map_coordinate` and wondering if you understand what's up.

yaugenst-flex: `np.take` returns a copy instead of a view. So if we use that, we end up spending 16 seconds in `np.fromnumeric` in my test case, just benchmarked it. I ended up creating the slices more dynamically though so at least it looks a bit nicer now.

yaugenst-flex: In my experience, `map_coordinates` is slightly faster. I think for the most part it's just less general and performs fewer checks, but the difference ended up being smaller than I thought. I just made a different version using `RegularGridInterpolator` and while the version using `map_coordinates` is slightly faster (I see a ~10% difference), I decided to go with the `RegularGridInterpolator` instead because that is what `xarray` ends up using in this case, so less chance of edge cases where results don't match. The reason why `xarray` is so slow is because it does a lot of type checking and coercion on the data, creates a bunch of intermediate objects in the process, decides which interpolation method to use, etc. etc... it goes through four or so modules before ever getting to the interpolation :smile:

momchil-flex: Yeah, makes sense.

momchil-flex (approved):
Looks good!

tylerflex (commented):
My only comment is that I would prefer probably if this special interp was moved somewhere where it could be re-used.  What do you think?

yaugenst-flex:
> My only comment is that I would prefer probably if this special interp was moved somewhere where it could be re-used. What do you think?

You mean essentially moving this `staticmethod` somewhere else?

tylerflex:
just not sure if we'll want to reuse this faster interpolation function somewhere else. if so, it might make sense to refactor it now so it's more generalized. but I'll leave it up to you

yaugenst-flex:
> just not sure if we'll want to reuse this faster interpolation function somewhere else. if so, it might make sense to refactor it now so it's more generalized. but I'll leave it up to you

i'll merge this now because i don't think this interpolation is much more custom than it was before this change. if anything, we might want to consider moving the `evaluate_flds_at` method somewhere outside of `derivative_utils`, but we also have to consider that this method is not differentiable, so probably should only be used in the context of the derivative utils.

PR #2384, opened 2025-04-15
Change return values to None instead of ValueError

marc-flex:
AbstractMedium was returning a `ValueError` instead of raising it. This seems like a bit too hard of a behavior so instead returning `None`. 

One question that I have (for @daquinteroflex ) is whether we should return `seff` in the `optical` case?

I have also added some minor improvements to the docs as pointed out by Xin Lei.

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -867,24 +867,19 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
 
     @property
     def charge(self):
-        return ValueError(f"A `charge` medium does not exist in this Medium definition: {self}")
+        return None
 
     @property
     def electrical(self):
-        return ValueError(
-            f"An `electrical` medium does not exist in this Medium definition: {self}"
-        )
+        return None
 
     @property
     def heat(self):
-        if self.heat_spec:
-            return self.heat_spec
-        else:
-            return ValueError(f"A `heat` medium does not exist in this Medium definition: {self}")
+        return self.heat_spec
 
     @property
     def optical(self):
-        return ValueError(f"An `optical` medium does not exist in this Medium definition: {self}")
+        return None
 
     @pd.validator("modulation_spec", always=True)
     @skip_if_fields_missing(["nonlinear_spec"])
--- a/tidy3d/components/material/tcad/charge.py
+++ b/tidy3d/components/material/tcad/charge.py
@@ -142,7 +142,7 @@ class SemiconductorMedium(AbstractChargeMedium):
             \\mathbf{F_{n,p}} = \\nabla \\psi
         \\end{equation}
 
-    i.e., we are not considering the effect of band-gab narrowing and degeneracy on the effective
+    i.e., we are not considering the effect of band-gap narrowing and degeneracy on the effective
     electric field :math:`\\mathbf{F_{n,p}}`. This is a good approximation for non-degenerate semiconductors.
 
     Let's explore how material properties are defined as class parameters or other classes.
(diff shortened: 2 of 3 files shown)


tidy3d/components/medium.py
```diff
@@ -867,24 +867,19 @@ def _validate_modulation_spec_post_init(self):
...
-        if self.heat_spec:
-            return self.heat_spec
-        else:
-            return ValueError(f"A `heat` medium does not exist in this Medium definition: {self}")
+        return None
 
     @property
     def optical(self):
-        return ValueError(f"An `optical` medium does not exist in this Medium definition: {self}")
+        return None
```
daquinteroflex, line 882: So because its in AbstractMedium I would not return self due to this https://github.com/flexcompute/tidy3d/blob/825d71f4efd60a4f3f36cf7c77b87dce4f122355/tidy3d/components/material/tcad/charge.py

daquinteroflex (approved):
Yeah we discussed about this separately

momchil-flex:
So `return ValueError` is definitely strange. But shouldn't the change be to `raise ValueError` rather than to return `None`? These two have pretty different implications for future code modifications.

marc-flex:
The idea was to follow the same behavior of `heat_spec`. When there isn't `heat_spec`, `heat_spec=None` is returned. In essence this is what we're doing with charge except we're doing it through exception management.

momchil-flex (approved):
Ok, let's give it a try.

tidy3d/components/medium.py
```diff
@@ -867,24 +867,19 @@ def _validate_modulation_spec_post_init(self):
...
-        if self.heat_spec:
-            return self.heat_spec
-        else:
-            return ValueError(f"A `heat` medium does not exist in this Medium definition: {self}")
+        return self.heat_spec
 
     @property
     def optical(self):
-        return ValueError(f"An `optical` medium does not exist in this Medium definition: {self}")
+        return None
```
yaugenst-flex, line 882: What's the rationale for having a `heat` property that returns the `heat_spec` field? Shouldn't there be _either_ a model field _or_ a property for these? Or maybe the other way around, why is it not `charge_spec`, `electrical_spec`, etc.?

marc-flex: The idea is that `heat` replaces `heat_spec` as we go multi-physics. It still needs to stay there for backward compatibility reasons, though.

momchil-flex: @yaugenst-flex does that make sense to you?

yaugenst-flex: > @yaugenst-flex does that make sense to you?

Not returning an error, absolutely :smile: 

I guess the part I don't understand is why these are properties instead of model fields. Does that mean that in the future these will be initialized by some other model fields? Because I assume we're not going to end up hard-coding what `property(optical)` returns right? How will these be initialized?

Or is the thinking more that we want to have these be properties right now and change the schema later (i.e., move these from a property to a model field) once things are properly implemented?

momchil-flex: Oh they are fields in [MultiPhysicsMedium](https://github.com/flexcompute/tidy3d/blob/74d5be32de7a15670273416c50246608cf6cc45c/tidy3d/components/material/multi_physics.py#L12). This is so that legacy media also work directly, e.g. calling .optical on a MultiPhysicsMedium or a SemiConductorMedium or a simple Medium works without having to type check.

yaugenst-flex: Ok great thanks that clarifies it!

PR #2385, opened 2025-04-15
Differentiable spline interpolation

yaugenst-flex:
This PR introduces a new differentiable function, `interpolate_spline`, to the `tidy3d.plugins.autograd` module. This function enables 1D spline interpolation (linear, quadratic, or cubic) of data points (`x_points`, `y_points`) onto a specified number of evaluation points. Crucially, it supports automatic differentiation via `autograd` with respect to the `y_points` input.

## Changes

*   **New Primitive:** Added `tidy3d.plugins.autograd.interpolate_spline`.
    *   Supports `order=1` (linear), `order=2` (quadratic), and `order=3` (cubic).
    *   Accepts optional `endpoint_derivatives` tuple `(left, right)`.
        *   Order 1: Both ignored (warning issued if provided).
        *   Order 2: Only `left` is used (warning issued if `right` is provided).
        *   Order 3: Both are used if provided.
*   **Refactoring:** Moved `gaussian_filter` and its test into a new `primitives` sub-package within `autograd` for better organization.
*   **Testing:** Added tests for `interpolate_spline` covering:
    *   Gradient correctness (`check_grads`) for all orders and various configurations.
    *   Interpolation accuracy.
    *   Error handling for unsupported orders.
    *   Warning generation for unused endpoint derivatives.

## Example Usage

```python
from autograd import value_and_grad
import autograd.numpy as anp
from tidy3d.plugins.autograd import interpolate_spline

# Define original points
x = anp.linspace(0, 1, 5)
y = anp.array([0.0, 0.8, 1.0, 0.5, 0.0])

# Define a function that uses spline interpolation
def objective(y_pts):
    # Interpolate y_pts onto 10 points using a cubic spline
    y_interp = interpolate_spline(x, y_pts, num_points=10, order=3)
    # Example: minimize the integral (sum as approximation)
    return anp.sum(y_interp**2)

# Calculate value and gradient
val, grad = value_and_grad(objective)(y)

print("Objective Value:", val)
print("Gradient wrt y:", grad)

### CODE REVIEW

File: tidy3d/plugins/autograd/primitives/interpolate.py [hunk 1]
```diff
@@ -0,0 +1,701 @@
...
+        Spline coefficients
+    x_eval : np.ndarray
+        Points at which to evaluate the spline
+
+    Returns
+    -------
+    np.ndarray
+        Evaluated spline values
+    """
+    if order == 1:
```
bhnzhang, line 637: maybe this is too nitpicky but is `order` needed as an argument or can it be gotten from `len(coeffs)`

tidy3d/plugins/autograd/primitives/interpolate.py
```diff
@@ -0,0 +1,701 @@
...
+
+    x_min, x_max = x_points[0], x_points[-1]
+    x_eval = np.linspace(x_min, x_max, num_points)
+
+    coeffs = compute_spline_coeffs(x_points, y_points, endpoint_derivatives, order)
+    y_eval = evaluate_spline(order, x_points, coeffs, x_eval)
+    return y_eval
+
+
+def interpolate_spline_y_vjp(ans, x_points, y_points, num_points, endpoint_derivatives, order):
```
bhnzhang: looks like `ans` is unused here, is that intentional?

bhnzhang (commented):
Thanks! That's a lot of code haha, looks good to me. Had a couple of minor comments

bzhangflex:
whoops didn't realize I was logged into my personal github

tidy3d/plugins/autograd/primitives/interpolate.py
yaugenst-flex: yeah this is the signature that autograd expects for a vjp, even if it's unused

tidy3d/plugins/autograd/primitives/interpolate.py
yaugenst-flex: yeah good catch, changed the function signature

tidy3d/plugins/autograd/primitives/interpolate.py
```diff
@@ -0,0 +1,707 @@
...
+        raise NotImplementedError(f"Derivatives for spline order {order} not implemented.")
+
+
+@primitive
+def interpolate_spline(
+    x_points: NDArray,
+    y_points: NDArray,
+    num_points: int,
+    endpoint_derivatives: tuple[Optional[float], Optional[float]] = (None, None),
+    order: int = 3,
```
groberts-flex: is there a reason for this default choice of order? it might be better to have the user specify this rather than having it default

tidy3d/plugins/autograd/primitives/interpolate.py
```diff
@@ -0,0 +1,707 @@
...
+        left_deriv = endpoint_derivatives[0]
+        return get_quadratic_derivative_wrt_y(x_points, y_points, left_deriv)
+    elif order == 3:
+        return get_cubic_derivative_wrt_y(x_points, y_points, endpoint_derivatives)
+    else:
+        raise NotImplementedError(f"Derivatives for spline order {order} not implemented.")
+
+
+@primitive
+def interpolate_spline(
```
groberts-flex, line 685: it might be useful to also return the x points corresponding to the new y points that this function returns

groberts-flex (commented):
cool feature @yaugenst-flex! Left a couple of comments

tidy3d/plugins/autograd/primitives/interpolate.py
yaugenst-flex: I do think there should be a default, but usually it's linear (if you look at the scipy interp methods). I chose cubic because I feel like this is what would be most often used in the context of shape optimization, and it's the only order where both `endpoint_derivatives` work / are meaningful (and this is a non-standard thing). So yeah not sure. I did consider no default too, I'm not sure what would be best. Opinions?

tidy3d/plugins/autograd/primitives/interpolate.py
yaugenst-flex: yeah actually noticed it's a bit annoying without. will add

rahul-flex:
Thanks @yaugenst-flex for adding the feature! A couple of minor comments:
1. Should there be a test for checking the length mismatch of x and y points passed to the interpolation function?
2. Also, for different orders of interpolation, do we need to check for the minimum number of points/knots required. For example in cubic spline we will need at least 4.

yaugenst-flex:
> Thanks @yaugenst-flex for adding the feature! A couple of minor comments:
> 
> 1. Should there be a test for checking the length mismatch of x and y points passed to the interpolation function?
> 2. Also, for different orders of interpolation, do we need to check for the minimum number of points/knots required. For example in cubic spline we will need at least 4.

1. yeah you're right that is an edge case that is currently not explicitly handled, probably good idea to add
2. the code works with all orders for >=2 input points. for two knots, quadratic ends up being a straight line if no endpoint derivative is specified, and with an endpoint you get the unique parabola that goes through both points. in the case of cubic, we always get a well-conditioned natural spline system. for two knots it collapses to quadratic (and follows above), and with three or more you always get cubic pieces. so basically if you don't supply enough points, the resulting curve will be of lower effective order, which I think is fine. I believe this is pretty common behavior.

yaugenst-flex:
ok I think I implemented all the comments. I also switched the solve for the cubic splines from a dense solve to a tridiagonal solve which grows linearly w.r.t. the number of points instead of cubically, so you can do _really_ large polygons with this

tidy3d/plugins/autograd/primitives/interpolate.py
groberts-flex: Thanks for the explanation on this. I am thinking about the case of using this in shape optimization to more densely sample a long, straight line segment like the edge of a box. For a cubic interpolation, the points wouldn't be evenly spaced along that line segment, which I would expect as a default behavior. I thinks I would lean slightly towards the lowest order interpolation as default or not have a default, but I can definitely see the argument for picking the cubic default if it's the most common case! I'm good with whatever way you go on this

tidy3d/plugins/autograd/primitives/interpolate.py
```diff
@@ -0,0 +1,786 @@
...
+
+import numpy as np
+from autograd.extend import defvjp, primitive
+from numpy.typing import NDArray
+from scipy.linalg import solve_banded
+
+from tidy3d.log import log
+
+
+def _assert_strictly_increasing(x: NDArray) -> None:
```
groberts-flex: can be done in the future, but should be able to support strictly decreasing x values as well by flipping x, interpolating, and then flipping back?

groberts-flex (dismissed):

rahul-flex (dismissed):

bzhangflex (dismissed):

tidy3d/plugins/autograd/primitives/interpolate.py
yaugenst-flex: I guess technically they don't need to be sorted at all, we can just do it...

tidy3d/plugins/autograd/primitives/interpolate.py
yaugenst-flex: > For a cubic interpolation, the points wouldn't be evenly spaced along that line segment

why would they not be evenly spaced? or do you mean they wouldn't sit on a straight line? hmm yeah.. maybe we shouldn't specify an order

tidy3d/plugins/autograd/primitives/interpolate.py
yaugenst-flex: ok the unsorted thing is bit annoying and the function is already quite over engineered as is, so I'll just require monotonicity instead as you suggested

tidy3d/plugins/autograd/primitives/interpolate.py
yaugenst-flex: remove the order.

groberts-flex: I was considering the following example:
`x = anp.linspace(0, 1, 3)`
`y = x`

x: array([0. , 0.5, 1. ])

`y_interp = interpolate_spline(x, y, num_points=5, order=1)`
y_interp: array([0.  , 0.25, 0.5 , 0.75, 1.  ])

`y_interp = interpolate_spline(x, y, num_points=5, order=3)`
y_interp: array([0.    , 0.1875, 0.5   , 0.6875, 1.    ])

They are both sampled along the straight line, but the spacing between points are different. I think this makes sense from the standpoint that there is nothing wrong with the implementation you have, but if you're a long edge that is a single straight segment, I would assume by default you would want to have all the vertices spaced evenly along the edge.

groberts-flex (approved):

PR #2386, opened 2025-04-16
fix[adjoint]: remove frequency summing in polyslab gradient calculation

groberts-flex:
Fixes error in grad_surfaces where the fields were being summed in frequency when boundary fields were being queried. The frequency summing is moved outside of the grad_surfaces and box gradient calls and gradients at each frequency are summed over.

We also add a polyslab numerical testing with mode sources that can be run by specifying "-m numerical" in pytest call.

### DIFF
--- a/tests/test_components/test_autograd_numerical.py
+++ b/tests/test_components/test_autograd_numerical.py
@@ -31,6 +31,7 @@ else:
 if SHOW_PRINT_STATEMENTS:
     sys.stdout = sys.stderr
 
+
 FINITE_DIFF_PERM_SEED = 1.5**2
 MESH_FACTOR_DESIGN = 30.0
 
@@ -255,9 +256,6 @@ def test_finite_difference_field_data(field_data_test_parameters, rng, tmp_path,
         "test_number",
     )(field_data_test_parameters)
 
-    mesh_wvl_um = mesh_wvls_um[idx]
-    adj_wvl_um = adj_wvls_um[idx]
-
     dim_um = mesh_wvl_um
     dim_um = mesh_wvl_um
     thickness_um = 0.5 * mesh_wvl_um
@@ -296,7 +294,7 @@ def test_finite_difference_field_data(field_data_test_parameters, rng, tmp_path,
 
     obj_val_and_grad = ag.value_and_grad(objective)
 
-    perm_init = FINITE_DIFF_PERM_SEED**2 * np.ones((dim, dim, Nz))
+    perm_init = FINITE_DIFF_PERM_SEED * np.ones((dim, dim, Nz))
 
     obj, adj_grad = obj_val_and_grad([perm_init])
 
--- a/tidy3d/components/autograd/derivative_utils.py
+++ b/tidy3d/components/autograd/derivative_utils.py
@@ -249,10 +249,18 @@ class DerivativeInfo(Tidy3dBaseModel):
         D_adj = self.D_adj
 
         # compute the E and D fields at the edge centers
-        E_fwd_at_coords = self.evaluate_flds_at(fld_dataset=E_fwd, spatial_coords=spatial_coords)
-        E_adj_at_coords = self.evaluate_flds_at(fld_dataset=E_adj, spatial_coords=spatial_coords)
-        D_fwd_at_coords = self.evaluate_flds_at(fld_dataset=D_fwd, spatial_coords=spatial_coords)
-        D_adj_at_coords = self.evaluate_flds_at(fld_dataset=D_adj, spatial_coords=spatial_coords)
+        E_fwd_at_coords = self.evaluate_flds_at(
+            fld_dataset=E_fwd, spatial_coords=spatial_coords, freq=self.frequency
+        )
+        E_adj_at_coords = self.evaluate_flds_at(
+            fld_dataset=E_adj, spatial_coords=spatial_coords, freq=self.frequency
+        )
+        D_fwd_at_coords = self.evaluate_flds_at(
+            fld_dataset=D_fwd, spatial_coords=spatial_coords, freq=self.frequency
+        )
+        D_adj_at_coords = self.evaluate_flds_at(
+            fld_dataset=D_adj, spatial_coords=spatial_coords, freq=self.frequency
+        )
 
         # project the relevant field quantities into their respective basis for gradient calculation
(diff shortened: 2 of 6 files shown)


groberts-flex:
Results from running the numerical unit test and plotting the finite difference vs. adjoint gradients.

Screenshot 2025-04-15 at 8 51 24 PM
Screenshot 2025-04-15 at 8 51 32 PM
Screenshot 2025-04-15 at 8 51 50 PM
Screenshot 2025-04-15 at 8 52 01 PM
Screenshot 2025-04-15 at 8 52 11 PM

tests/test_components/test_autograd_numerical.py
```diff
@@ -255,9 +256,6 @@ def test_finite_difference_field_data(field_data_test_parameters, rng, tmp_path,
         "test_number",
     )(field_data_test_parameters)
 
-    mesh_wvl_um = mesh_wvls_um[idx]
```
groberts-flex, line 258: quick fix here to make sure we are picking up the test parameters. reran the tests to make sure everything still passes.

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -303,42 +311,35 @@ def evaluate_eps(
...
-                    slicer = [slice(None)] * data.ndim
-                    slicer[f_dim_idx] = f_idx
-                    interpolator = RegularGridInterpolator(
-                        points, data[tuple(slicer)], **interp_kwargs
-                    )
-                    result += interpolator(coords)
-            else:
-                interpolator = RegularGridInterpolator(points, data, **interp_kwargs)
-                result = interpolator(coords)
+            interpolator = RegularGridInterpolator(points, data, **interp_kwargs)
```
yaugenst-flex: Since you remove the other interpolator, could you move the `interp_kwargs` into the function call directly?

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -303,42 +311,35 @@ def evaluate_eps(
...
         coords = np.nan_to_num(spatial_coords, posinf=LARGE_NUMBER, neginf=-LARGE_NUMBER)
         components = {}
+
         edge_index_dim = "edge_index"
         n_points = coords.shape[0]
 
         for fld_name, arr in fld_dataset.items():
             data = arr.values
+            if "f" in arr.dims:
+                data = arr.sel(f=freq).values
```
yaugenst-flex: Let's do `data = arr.sel(f=freq).values if "f" in arr.dims else arr.values` here

tidy3d/web/api/autograd/autograd.py
```diff
@@ -969,11 +969,11 @@ def postprocess_adj(
         eps_adj = sim_data_adj.get_adjoint_data(structure_index, data_type="eps")
 
         # post normalize the adjoint fields if a single, broadband source
-        fwd_flds_normed = {}
+        adj_flds_normed = {}
```
yaugenst-flex, line 980: oops

tidy3d/web/api/autograd/autograd.py
```diff
@@ -988,78 +988,78 @@ def postprocess_adj(
...
-        for structure_path, vjp_value in vjp_value_map.items():
-            sim_path = tuple(["structures", structure_index] + list(structure_path))
-            sim_fields_vjp[sim_path] = vjp_value
+            # extract VJPs and put back into sim_fields_vjp AutogradFieldMap
+            for structure_path, vjp_value in vjp_value_map.items():
+                sim_path = tuple(["structures", structure_index] + list(structure_path))
+                if freq_idx == 0:
+                    sim_fields_vjp[sim_path] = vjp_value
+                else:
+                    sim_fields_vjp[sim_path] += vjp_value
```
yaugenst-flex: Not necessarily related to this PR but I wonder if `vjp_value` can ever be something other than a scalar or a numpy array, in which case this would silently fail. We ran into this in #2319. Might be good to double check vs `Box` center & size derivatives because that method does return tuples for the VJPs, see [here](https://github.com/flexcompute/tidy3d/blob/74d5be32de7a15670273416c50246608cf6cc45c/tidy3d/components/geometry/base.py#L2520-L2523)...

yaugenst-flex (commented):
Thanks @groberts-flex this is great! I guess the new test functions could use some more comments (e.g., why was `fd_step = 0.2 * mesh_wvl_um` chosen?), but nothing major.

tidy3d/web/api/autograd/autograd.py
groberts-flex: Good point, didn't think of this when implementing. Tried with the box to see how this would ultimately happen in the code and I think this only operates on scalar values in that case. The center/size derivatives come back as a dictionary of tuples as you pointed out, but those get packaged up into a dictionary that identifies each gradient by its component in either center or size. That dictionary has keys that look like ('geometry', 'size', 0) which select out the attribute and index into the attribute for the derivative and then the vjp value is just a scalar.

For the vertices, this ends up looking like ('geometry', 'vertices') and then the vertices are being summed. If those vertices came back as a list or tuple (or some object using this code path had that) though instead of an array, this could be a problem. I've added a change to catch this situation. One thing I'm wondering about the solution from the #2319 that didn't occur to me before is if the list is not flat. For example,

`x = [ [ 1, 2 ], [ 3, 4 ] ]`
`y = [ [ 5, 6 ], [ 7, 8 ] ]`
`type(x)(x + y for x, y in zip(x, y))`
[[1, 2, 5, 6], [3, 4, 7, 8]]

instead of [[6, 8], [10, 12]]

yaugenst-flex: yeah you're right. I guess the question is, do we ever run into nested tuples? if all tuples map to regular arrays, we could cast to numpy, to the math there, and then cast back. otherwise we'd have to recursively map through the nesting. I feel like that would be over-engineering it if there are never nested tuples in the vjps in the first place

groberts-flex: Yeah, not sure if we have a lot of possibility for it. The reason it came up in my head is the vertices are in a nested form when they come into this function (they are a list of coordinate pairs). They are not coming in as a list/tuple and are summing completely fine as arrays, so not sure if there's a case where things could possibly come in that way. I would assume that at that point, there would be a dimension mismatch down the line and some sort of error.

PR #2388, opened 2025-04-17
Unsteady heat support

marc-flex:
Adding support for unsteady heat simulations.

### DIFF
--- a/tidy3d/components/material/tcad/heat.py
+++ b/tidy3d/components/material/tcad/heat.py
@@ -31,17 +31,15 @@ class AbstractHeatMedium(ABC, Tidy3dBaseModel):
 
     @property
     def charge(self):
-        return ValueError(f"A `charge` medium does not exist in this Medium definition: {self}")
+        raise ValueError(f"A `charge` medium does not exist in this Medium definition: {self}")
 
     @property
     def electrical(self):
-        return ValueError(
-            f"An `electrical` medium does not exist in this Medium definition: {self}"
-        )
+        raise ValueError(f"An `electrical` medium does not exist in this Medium definition: {self}")
 
     @property
     def optical(self):
-        return ValueError(f"An `optical` medium does not exist in this Medium definition: {self}")
+        raise ValueError(f"An `optical` medium does not exist in this Medium definition: {self}")
 
 
 class FluidMedium(AbstractHeatMedium):
--- a/tidy3d/components/material/tcad/heat.py
+++ b/tidy3d/components/material/tcad/heat.py
@@ -9,6 +9,7 @@ import pydantic.v1 as pd
 
 from tidy3d.components.base import Tidy3dBaseModel
 from tidy3d.constants import (
+    DENSITY,
     SPECIFIC_HEAT_CAPACITY,
     THERMAL_CONDUCTIVITY,
 )
@@ -69,8 +70,9 @@ class SolidMedium(AbstractHeatMedium):
     """
 
     capacity: pd.PositiveFloat = pd.Field(
+        None,
         title="Heat capacity",
-        description=f"Volumetric heat capacity in unit of {SPECIFIC_HEAT_CAPACITY}.",
+        description=f"Specific heat capacity in unit of {SPECIFIC_HEAT_CAPACITY}.",
         units=SPECIFIC_HEAT_CAPACITY,
     )
 
@@ -79,6 +81,32 @@ class SolidMedium(AbstractHeatMedium):
         description=f"Thermal conductivity of material in units of {THERMAL_CONDUCTIVITY}.",
         units=THERMAL_CONDUCTIVITY,
     )
+
+    density: pd.PositiveFloat = pd.Field(
+        None,
+        title="Density",
+        description=f"Mass density of material in units of {DENSITY}.",
+        units=DENSITY,
+    )
+
+    def from_si_units(
+        conductivity: pd.PositiveFloat,
+        capacity: pd.PositiveFloat = None,
+        density: pd.PositiveFloat = None,
+    ):
+        """Create a SolidMedium using SI units"""
+        new_conductivity = conductivity * 1e-6  # Convert from W/(m*K) to W/(um*K)
+        new_capacity = capacity
+        new_density = density
+
+        if density is not None:
+            new_density = density * 1e-18
+
+        return SolidMedium(
+            capacity=new_capacity,
+            conductivity=new_conductivity,
+            density=new_density,
+        )
 
 
 class SolidSpec(SolidMedium):
--- a/tidy3d/components/tcad/monitors/heat.py
+++ b/tidy3d/components/tcad/monitors/heat.py
@@ -1,7 +1,19 @@
 """Objects that define how data is recorded from simulation."""
 
+import pydantic.v1 as pd
+
 from tidy3d.components.tcad.monitors.abstract import HeatChargeMonitor
 
 
 class TemperatureMonitor(HeatChargeMonitor):
     """Temperature monitor."""
+
+    interval: pd.PositiveInt = pd.Field(
+        1,
+        title="Interval",
+        description="Sampling rate of the monitor: number of time steps between each measurement. "
+        "Set ``interval`` to 1 for the highest possible resolution in time. "
+        "Higher integer values down-sample the data by measuring every ``interval`` time steps. "
+        "This can be useful for reducing data storage as needed by the application."
+        "NOTE: this is only relevant for unsteady (transient) Heat simulations. ",
+    )
--- a/tidy3d/components/tcad/simulation/heat_charge.py
+++ b/tidy3d/components/tcad/simulation/heat_charge.py
@@ -363,8 +363,7 @@ class HeatChargeSimulation(AbstractSimulation):
                     isinstance(medium.heat_spec, SolidMedium) for medium in medium_set
                 )
                 crosses_elec_spec = any(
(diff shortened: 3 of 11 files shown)


dbochkov-flexcompute (commented):
a couple of big picture comments:
- Not sure about having output frequency as a parameter in `UnsteadySpec`. I think it could be more useful if sampling frequency can be defined for each monitor separately, similar to fdtd time monitors. Say, if I want to have just few snapshots in time for a big monitor, but then also record temperature for each step at a given point.
- For initial condition, I imagine at some point we might need to extend it to non-uniform distribution. While I would say this is out of scope for this PR, I think it would be worth ironing out how that would look like. I see two options
  - `initial_temperature: Union[pd.PositiveFloat, SpatialDataArray, UnstructuredGridDataset]`
  - `initial_temperature: Union[UniformDistibution, CustomDistribution]`
  
  probably the first option is a way to go, unless I am missing something that would require introduction of new specs?

marc-flex:
Thanks @daniil for looking into this. As per usual these are great suggestions. 

> Not sure about having output frequency as a parameter in UnsteadySpec. I think it could be more useful if sampling frequency can be defined for each monitor separately, similar to fdtd time monitors. Say, if I want to have just few snapshots in time for a big monitor, but then also record temperature for each step at a given point.

I'll implement this as done in fdtd. I guess my only comment is that depending on the simulation the amount of data generated during the run could be quite big since we'd be saving every time-step and then post-process only the required steps. On the other hand, once the simulation is done and post-processed, all those temp files are removed, so it is only a temporary problem?

> For initial condition, I imagine at some point we might need to extend it to non-uniform distribution. While I would say this is out of scope for this PR, I think it would be worth ironing out how that would look like. I see two options...

I like this. This may be related to [this issue](https://github.com/flexcompute/tidy3d-core/issues/929) where we want to interpolate data to a mesh. I guess once I resolve that issue we can revisit this?

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -834,6 +836,34 @@ def estimate_charge_mesh_size(cls, values):
...
+        if isinstance(analysis_type, UnsteadyHeatAnalysis):
+            monitors = values.get("monitors")
+            for mnt in monitors:
+                if isinstance(mnt, TemperatureMonitor):
+                    if not mnt.unstructured:
+                        raise SetupError(
+                            f"Unsteady simulations require the temperature monitor '{mnt.name}' to be unstructured."
+                        )
+            # additionaly check that the SolidSpec has capacitance and density defined
+            # NOTE: not sure this is needed. Is capacitance a required field in SolidMedium?
```
momchil-flex: Why do we need capacitance defined for heat simulations?

momchil-flex (approved):
Looks great!

The main thing that I think remains (and can't entirely be done now) is imposing some restrictions similar to what we have [for fdtd sims](https://github.com/flexcompute/tidy3d/blob/cd3b0df87d463c7d034dd5c316d7e8ed45e906a7/tidy3d/components/simulation.py#L151-L162). Maybe the only one we can really do now is just setting a limit on the maximum number of time steps in the simulation (and maybe some ball park limits on the time step). Basically users will inevitably create something that has way too many time steps for some reason, unless we limit it.

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -85,14 +85,16 @@
...
+
 HEAT_CHARGE_BACK_STRUCTURE_STR = "<<<HEAT_CHARGE_BACKGROUND_STRUCTURE>>>"
 
 HeatBCTypes = (TemperatureBC, HeatFluxBC, ConvectionBC)
 HeatSourceTypes = (UniformHeatSource, HeatSource, HeatFromElectricSource)
 ChargeSourceTypes = ()
 ElectricBCTypes = (VoltageBC, CurrentBC, InsulatingBC)
 
-AnalysisSpecType = ElectricalAnalysisType
+AnalysisSpecType = Union[ElectricalAnalysisType, UnsteadyHeatAnalysis]
```
momchil-flex, line 97: This [descirption](https://github.com/flexcompute/tidy3d/blob/58799600418fc51471b4a20296b30a252e34cbff/tidy3d/components/tcad/simulation/heat_charge.py#diff-b059a461f69d59f8db0b976d761063f21b17b70652284253ee880aa7960b6f60L302-L303) doesn't sound quite right now? And even previously it reads a bit weird - the purpose of the spec isn't really to *validate* anything, is it?

tidy3d/components/tcad/simulation/heat_charge.py
marc-flex: This what happens when one starts juggling heat and charge at the same time. I meant to say capacity as in volumetric heat capacity. I'll update this :+1:

tidy3d/components/tcad/simulation/heat_charge.py
marc-flex: So it serves the user specify the type of simulation they really want to run. For us, it also helps us make sure we have all the required parameters/input.

momchil-flex: Right, don't you think the description should be more along the lines of the first part, defining what simulation people really want to run?

marc-flex: I agree. Not sure why we even wrote the previous in the first place.

tidy3d/components/material/tcad/heat.py
```diff
@@ -69,6 +70,7 @@ class SolidMedium(AbstractHeatMedium):
     """
 
     capacity: pd.PositiveFloat = pd.Field(
+        None,
         title="Heat capacity",
         description=f"Volumetric heat capacity in unit of {SPECIFIC_HEAT_CAPACITY}.",
```
dbochkov-flexcompute: I guess now it is not `Volumetric heat capacity` but rather `Specific heat capacity`?

tidy3d/constants.py
```diff
@@ -194,6 +194,11 @@
 Joules per (kilogram Kelvin).
 """
 
+DENSITY = "kg/m^3"
```
dbochkov-flexcompute: we use `um` in all other variables, including heat related ones `THERMAL_CONDUCTIVITY`, `HEAT_FLUX`, `VOLUMETRIC_HEAT_RATE`, and `HEAT_TRANSFER_COEFF`. I wonder if we should also make it `"kg/um^3"`, although all tables typically provide it in `"kg/m^3"`

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -834,6 +839,68 @@ def estimate_charge_mesh_size(cls, values):
             )
         return values
 
+    @pd.root_validator(skip_on_failure=True)
+    def check_transient_heat(cls, values):
```
dbochkov-flexcompute, line 836: add `@skip_if_fields_missing(["analysis_spec", "monitors", "structures", "size"])` like here https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/simulation.py#L369C5-L369C44. This will skip execution of this validator if any of the above fields fail before that. This help to avoid meaningless errors

tidy3d/components/tcad/monitors/heat.py
```diff
@@ -1,7 +1,18 @@
...
 from tidy3d.components.tcad.monitors.abstract import HeatChargeMonitor
 
 
 class TemperatureMonitor(HeatChargeMonitor):
     """Temperature monitor."""
+
+    interval: pd.PositiveInt = pd.Field(
+        1,
+        title="Interval",
+        description="Sampling rate of the monitor: number of time steps between each measurement. "
```
dbochkov-flexcompute, line 14: add that this is only relevant for transient simulations

dbochkov-flexcompute (approved):
Looks great! A few minor comments

tidy3d/constants.py
marc-flex: I have mixed feelings about this. I think it's useful to have units readily available in the literature though this forces us to change the units internally. Either way I don't have a strong preference for this.

Another popular unit for density in the literature is g/cm³ . 

@momchil-flex @dbochkov-flexcompute should leave it to a vote? Or should I just change it to kg/um³ which is consistent with the rest of the units and is convenient for us?

momchil-flex: Yeah I had mixed feelings about this too when reading @dbochkov-flexcompute 's comment. It's weird for density to be in kg/um**3. But, it's also weird for it not to be when every other length unit in Tidy3D is in um. Because of the latter, I tend to lean towards kg/um**3. But I am wondering if there were similar considerations when defining `THERMAL_CONDUCTIVITY`, etc. - I would imagine those are also not standard units? I could have probably imagined a path in which we only define lengths in um but material constants that have length scales in their definitions don't need to be converted to um.

dbochkov-flexcompute: yeah, I didn't think it through back when I introduced `THERMAL_CONDUCTIVITY` and `HEAT_TRANSFER_COEFF`, now looking at it I think more standard units would have made more sense

marc-flex: I's too late to change now, is it? We're going to go with kg/um³ I think

momchil-flex: I guess something we can change in tidy3d 3.0 :)

marc-flex: OK, so it is now defined using um. As a convenience function, I added the a `SolidMedium.from_si_units()` with which I think we kind of have both ways

PR #2390, opened 2025-04-17
Gap automeshing

dbochkov-flexcompute:
Integrated gap meshing in an iterative way. Now `LayerRefinementSpec` has two additional parameters:
- `gap_meshing_iters: NonNegativeInt = 1` to set the number of iteration to perform for attempting resolving all gaps
- `dl_min_from_gap_width: bool = True` whether or not to reduce `dl_min` reduce to the minimal detected gap width.

A few of points to discuss:
- Doing just one iterations seems to be performing quite well, so that could be a good default. But maybe we should switch to a higher number and just let it iterate until no multiple crossings are detected in a single cell?
- Currently it does not distinguish between thin gaps and thin strips. Do we want to control mesher behavior for the two types
- while we use snapping planes for achieving the desired result, we currently do not store them, and, as a consequence, we cannot plot them. Should we worry about that?

examples of meshing thin strips and gaps:
Screenshot_20250417_231504
Screenshot_20250417_231652

### DIFF
--- a/tidy3d/components/grid/grid_spec.py
+++ b/tidy3d/components/grid/grid_spec.py
@@ -8,11 +8,11 @@ from typing import Any, List, Literal, Optional, Tuple, Union
 import numpy as np
 import pydantic.v1 as pd
 
-from ...constants import C_0, MICROMETER, inf
+from ...constants import C_0, MICROMETER, dp_eps, inf
 from ...exceptions import SetupError
 from ...log import log
 from ..base import Tidy3dBaseModel, cached_property, skip_if_fields_missing
-from ..geometry.base import Box
+from ..geometry.base import Box, ClipOperation
 from ..lumped_element import LumpedElementType
 from ..source.utils import SourceType
 from ..structure import MeshOverrideStructure, Structure, StructureType
@@ -35,6 +35,9 @@ MIN_STEP_BOUND_SCALE = 0.5
 
 # Default refinement factor in GridRefinement when both dl and refinement_factor are not defined
 DEFAULT_REFINEMENT_FACTOR = 2
+
+# Tolerance for distinguishing pec/grid intersections
+GAP_MESHING_TOL = 1e-3
 
 
 class GridSpec1d(Tidy3dBaseModel, ABC):
@@ -62,6 +65,9 @@ class GridSpec1d(Tidy3dBaseModel, ABC):
         symmetry : Tuple[Symmetry, Symmetry, Symmetry]
             Reflection symmetry across a plane bisecting the simulation domain
             normal to each of the three axes.
+        periodic : bool
+            Apply periodic boundary condition or not.
+            Only relevant for autogrids.
         wavelength : float
             Free-space wavelength.
         num_pml_layers : Tuple[int, int]
@@ -1037,6 +1043,20 @@ class LayerRefinementSpec(Box):
         "and the projection of the simulation domain overlaps.",
     )
 
+    gap_meshing_iters: pd.NonNegativeInt = pd.Field(
+        1,
+        title="Gap Meshing Iterations",
+        description="Number of recursive iterations for resolving thin gaps. "
+        "The underlying algorithm detects gaps contained in a single cell and places a snapping plane at the gaps's centers.",
+    )
+
+    dl_min_from_gap_width: bool = pd.Field(
+        True,
+        title="Set ``dl_min`` from Estimated Gap Width",
+        description="Take into account autodetected minimal PEC gap width when determining ``dl_min``. "
+        "This only applies if ``dl_min`` in ``AutoGrid`` specification is not set.",
+    )
+
     @pd.validator("axis", always=True)
     @skip_if_fields_missing(["size"])
     def _finite_size_along_axis(cls, val, values):
@@ -1057,6 +1077,8 @@ class LayerRefinementSpec(Box):
         corner_snapping: bool = True,
         corner_refinement: GridRefinement = GridRefinement(),
         refinement_inside_sim_only: bool = True,
+        gap_meshing_iters: pd.NonNegativeInt = 1,
+        dl_min_from_gap_width: bool = True,
     ):
         """Constructs a :class:`LayerRefiementSpec` that is unbounded in inplane dimensions from bounds along
         layer thickness dimension.
@@ -1082,6 +1104,10 @@ class LayerRefinementSpec(Box):
             Inplane mesh refinement factor around corners.
         refinement_inside_sim_only : bool = True
             Apply refinement only to features inside simulation domain.
+        gap_meshing_iters : bool = True
+            Number of recursive iterations for resolving thin gaps.
+        dl_min_from_gap_width : bool = True
+            Take into account autodetected minimal PEC gap width when determining ``dl_min``.
 
 
         Example
@@ -1103,6 +1129,8 @@ class LayerRefinementSpec(Box):
             corner_snapping=corner_snapping,
             corner_refinement=corner_refinement,
             refinement_inside_sim_only=refinement_inside_sim_only,
+            gap_meshing_iters=gap_meshing_iters,
+            dl_min_from_gap_width=dl_min_from_gap_width,
         )
 
     @classmethod
@@ -1118,6 +1146,8 @@ class LayerRefinementSpec(Box):
         corner_snapping: bool = True,
         corner_refinement: GridRefinement = GridRefinement(),
         refinement_inside_sim_only: bool = True,
+        gap_meshing_iters: pd.NonNegativeInt = 1,
+        dl_min_from_gap_width: bool = True,
     ):
         """Constructs a :class:`LayerRefiementSpec` from minimum and maximum coordinate bounds.
 
@@ -1145,6 +1175,10 @@ class LayerRefinementSpec(Box):
             Inplane mesh refinement factor around corners.
         refinement_inside_sim_only : bool = True
             Apply refinement only to features inside simulation domain.
+        gap_meshing_iters : bool = True
+            Number of recursive iterations for resolving thin gaps.
+        dl_min_from_gap_width : bool = True
+            Take into account autodetected minimal PEC gap width when determining ``dl_min``.
 
 
         Example
@@ -1166,6 +1200,8 @@ class LayerRefinementSpec(Box):
             corner_snapping=corner_snapping,
             corner_refinement=corner_refinement,
             refinement_inside_sim_only=refinement_inside_sim_only,
+            gap_meshing_iters=gap_meshing_iters,
+            dl_min_from_gap_width=dl_min_from_gap_width,
         )
 
     @classmethod
@@ -1180,6 +1216,8 @@ class LayerRefinementSpec(Box):
         corner_snapping: bool = True,
         corner_refinement: GridRefinement = GridRefinement(),
         refinement_inside_sim_only: bool = True,
+        gap_meshing_iters: pd.NonNegativeInt = 1,
+        dl_min_from_gap_width: bool = True,
     ):
         """Constructs a :class:`LayerRefiementSpec` from the bounding box of a list of structures.
 
@@ -1205,6 +1243,10 @@ class LayerRefinementSpec(Box):
             Inplane mesh refinement factor around corners.
         refinement_inside_sim_only : bool = True
             Apply refinement only to features inside simulation domain.
+        gap_meshing_iters : bool = True
+            Number of recursive iterations for resolving thin gaps.
+        dl_min_from_gap_width : bool = True
+            Take into account autodetected minimal PEC gap width when determining ``dl_min``.
 
         """
 
@@ -1226,6 +1268,8 @@ class LayerRefinementSpec(Box):
             corner_snapping=corner_snapping,
             corner_refinement=corner_refinement,
             refinement_inside_sim_only=refinement_inside_sim_only,
+            gap_meshing_iters=gap_meshing_iters,
+            dl_min_from_gap_width=dl_min_from_gap_width,
         )
 
     @cached_property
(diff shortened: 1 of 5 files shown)


dbochkov-flexcompute:
one more point:
- currently we also take into account all structures. Should we worry about making it work for a selected set of structures in this PR?

weiliangjin2021:
Thanks for this great feature!

> Doing just one iterations seems to be performing quite well, so that could be a good default. But maybe we should switch to a higher number and just let it iterate until no multiple crossings are detected in a single cell?

Agree that we can set 1 as default if ti works quite well already. Then we can add an option like `inf` or a string to indicate it will iterate until no crossings are detected?

> Currently it does not distinguish between thin gaps and thin strips. Do we want to control mesher behavior for the two types

Right now, we don't have good conformal scheme for thin strips, so probably no need right now. 

> while we use snapping planes for achieving the desired result, we currently do not store them, and, as a consequence, we cannot plot them. Should we worry about that?

Not too concerned, but I wonder that it's not hard to store them?

> currently we also take into account all structures. Should we worry about making it work for a selected set of structures in this PR?

My understanding is that metallic structures are more of a concern. So we probably only need to distinguish between metallic (PEC, lossyMetal) and dielectric.

tidy3d/components/grid/grid_spec.py
```diff
@@ -1921,6 +2114,102 @@ def make_grid(
...
+            List of sources.
+        num_pml_layers : List[Tuple[float, float]]
+            List containing the number of absorber layers in - and + boundaries.
+        lumped_elements : List[LumpedElementType]
+            List of lumped elements.
+        internal_override_structures : List[MeshOverrideStructure]
+            If `None`, recomputes internal override structures.
+        internal_snapping_points : List[CoordinateOptional]
+            If `None`, recomputes internal snapping points.
+
```
dmarek-flex, line 2605: missing some of the parameters in the docstring

dmarek-flex: looks like missing parameters in the original function's docstring as well

tidy3d/components/grid/grid_spec.py
```diff
@@ -1024,6 +1024,19 @@ class LayerRefinementSpec(Box):
...
     )
 
+    gap_meshing_iters: pd.NonNegativeInt = pd.Field(
+        1,
+        title="Gap Meshing Iterations",
+        description="Number of recursive iterations for resolving thin gaps. "
+        "The underlying algorithm detects gaps contained in a single cell and places a snapping plane at the gaps's centers.",
+    )
+
+    dl_min_from_gap_width: bool = pd.Field(
```
weiliangjin2021, line 1053: Now that we have dl_min from quite a few places: gap refinement, corner refinement, thickness refinement, and minimal feature, shall we allow controling them individually and combine them into a new class?

tidy3d/components/grid/grid_spec.py
```diff
@@ -1497,6 +1534,162 @@ def _override_structures_along_axis(
                 override_structures += refinement_structures
         return override_structures
 
+    def _find_vertical_intersections(self, x, y, vertices):
+        """Detect intersection points of single polygon and vertical grid lines."""
+
+        cells_ij = []
+        cells_dy = []
+        h_inds = np.argmax(x[:, None] > vertices[None, :, 0], axis=0)
+        h_inds[vertices[:, 0] > x[-1]] = len(x)
```
weiliangjin2021: More informative variable name for `h_inds`

tidy3d/components/grid/grid_spec.py
```diff
@@ -1497,6 +1534,162 @@ def _override_structures_along_axis(
...
+        for ind_beg, ind_end, v_beg, v_end in zip(
```
weiliangjin2021, line 1580: add a comment that `beg` and `end` refers to the two vertices on the edge of the polygon

tidy3d/components/grid/grid_spec.py
```diff
@@ -1497,6 +1534,162 @@ def _override_structures_along_axis(
...
+            h_inds, np.roll(h_inds, -1), vertices, np.roll(vertices, axis=0, shift=-1)
+        ):
+            # sort vertices in ascending order
+            if ind_beg > ind_end:
+                ind_beg, ind_end, v_beg, v_end = ind_end, ind_beg, v_end, v_beg
+
+            if ind_end > ind_beg:
```
weiliangjin2021: clearer with `if ind_end == ind_beg: continue`, and undent the rest? This brings up the question on how the close-to-vertical polygon edge is handled. I guess they are handled when looking at horizontal intersection?

tidy3d/components/grid/grid_spec.py
```diff
@@ -1497,6 +1534,162 @@ def _override_structures_along_axis(
...
+        if len(cells_dy) > 0:
+            cells_dy = np.concatenate(cells_dy)
+
+        return cells_ij, cells_dy
+
+    def _process_poly(self, x, y, vertices):
+        """Detect intersection points of single polygon and grid lines."""
+        v_cells_ij, v_cells_dy = self._find_vertical_intersections(x, y, vertices)
+        # reuse the same command but flip dimensions
+        h_cells_ij, h_cells_dx = self._find_vertical_intersections(y, x, vertices[:, [1, 0]])
```
weiliangjin2021: use np.flip?

tidy3d/components/grid/grid_spec.py
```diff
@@ -1497,6 +1534,162 @@ def _override_structures_along_axis(
...
+        if len(h_cells_ij) > 0:
+            h_cells_ij = np.roll(h_cells_ij, axis=1, shift=1)
```
weiliangjin2021, line 1729: more explanation here?

tidy3d/components/grid/grid_spec.py
```diff
@@ -1921,6 +2114,102 @@ def make_grid(
...
+                    if layer_spec.gap_meshing_iters > ind:
+                        one_layer_snapping_lines, gap_width = layer_spec._resolve_gaps(
+                            structures, old_grid
+                        )
+                        new_snapping_lines = new_snapping_lines + one_layer_snapping_lines
+                        if layer_spec.dl_min_from_gap_width:
+                            min_gap_width = min(min_gap_width, gap_width)
+
+                if len(new_snapping_lines) == 0:
+                    log.info("Grid is no longer changing. Stopping recursive refinement.")
```
weiliangjin2021: include the information on the number of iteration where it stops.

weiliangjin2021 (commented):
Half way through the code. Looks quite nice, although taking quite a bit efforts in understanding the logic. Some high-level description in each function will be very helpful. Some minor comments so far:

dbochkov-flexcompute:
sprinkled around some more comments in the code, had to recall myself what I was doing :sweat_smile:

tidy3d/components/grid/grid_spec.py
dmarek-flex: `periodic` and `min_dl_from_gaps`

tidy3d/components/grid/grid_spec.py
```diff
@@ -1497,6 +1534,253 @@ def _override_structures_along_axis(
...
+        v_cells_dy = []
+
+        # cells that contain intersections of horizontal grid lines
+        h_cells_ij = []
+        # locations of those intersections (along x axis)
+        h_cells_dx = []
+
+        # loop over all shapes
+        for mat, shapes in merged_geos:
+            if not mat.is_pec:
```
dmarek-flex, line 1771: We should support `LossyMetalMedium` as well, I need to make the same check in my next PR. Maybe we can add `is_metal_conductor` to `AbstractMedium`

tidy3d/components/grid/grid_spec.py
```diff
@@ -1497,6 +1534,253 @@ def _override_structures_along_axis(
                 override_structures += refinement_structures
         return override_structures
 
+    def _find_vertical_intersections(self, x, y, vertices):
```
dmarek-flex: add some type hints for the return value of these methods

tidy3d/components/grid/grid_spec.py
```diff
@@ -1497,6 +1534,253 @@ def _override_structures_along_axis(
...
+        # 1. locations of those intersections along y axis
+        # 2. cells that contain intersections of horizontal grid lines
+        # 3. locations of those intersections along x axis
+        v_cells_ij, v_cells_dy, h_cells_ij, h_cells_dx = self._process_slice(x, y, plane_slice)
+
+        detected_gap_width = inf
+
+        # generate horizontal snapping lines
+        snapping_lines_y = []
+        if len(v_cells_ij) > 0:
```
dmarek-flex: Seems possible to avoid duplicating the code for the horizontal and vertical lines (as you did for `_find_vertical_intersections`

tidy3d/components/grid/grid_spec.py
```diff
@@ -1497,6 +1534,253 @@ def _override_structures_along_axis(
...
         return override_structures
 
+    def _find_vertical_intersections(self, x, y, vertices):
+        """Detect intersection points of single polygon and vertical grid lines."""
+
+        cells_ij = []
+        cells_dy = []
+
+        # for each polygon vertex find the index of the first grid line on the right
+        grid_lines_on_right = np.argmax(x[:, None] > vertices[None, :, 0], axis=0)
```
dmarek-flex: It might make sense to use a small tolerance, so that we can distinguish between almost exactly coincident with the cell boundary and certainly greater than.

dmarek-flex: That might minimize the number of false positives, could also result in some of the asymmetry present in the resulting grids.

dmarek-flex (commented):
Another great feature! I found a couple of small things in the associated comments. The only real sticking point for me is that it would be nice to get a symmetric grid, when the structure itself is symmetric. 

I think it came up in the meeting and maybe there is no easy solution. But for the structure below it seems to always give an asymmetric grid. I thought maybe part of the reason could be due to how intersections are found with grid boundaries, which does not make use of a tolerance. The grid boundary and the polygon vertex might be very close. 

image

And here is the original grid without gap refinement. (setting gap_meshing_iters=0)

Screenshot from 2025-04-23 15-28-32

tidy3d/components/grid/grid_spec.py
dbochkov-flexcompute: actually we rely here on that the function that constructed `merged_geos` also converted all `LossyMetalMedium`'s into pec https://github.com/flexcompute/tidy3d/blob/daniil/gap-meshing/tidy3d/components/grid/corner_finder.py#L109-L116 

Not ideal, as it is very easy not to realize that, but I added a couples of notes to point that out

tidy3d/components/grid/grid_spec.py
dbochkov-flexcompute: done

tidy3d/components/grid/grid_spec.py
dbochkov-flexcompute: done

tidy3d/components/grid/grid_spec.py
dbochkov-flexcompute: done

tidy3d/components/grid/grid_spec.py
dbochkov-flexcompute: good idea, added special treatment for such cases

dbochkov-flexcompute:
should be all addressed now. Some notable changes:
- Regarding @dmarek-flex point on doing this with some tolerance allowance, I implemented the following: intersections that are very close to cell boundaries are duplicated to the neighboring edges across those boundaries. That is, if we have a PEC boundary segment passing through a grid node (up to some tolerance) , we consider that this segment crosses both edges around this grid node. This seems to nicely improve the quality of resolving gaps and strips. Here's the circularly polarized antenna with base resolutions 6 (needed 2 iterations), 10, 15, and 20
Screenshot_20250506_220238
antenna10
antenna15
antenna20

- Additionally, now I also explicitly filter out intersections that happen due to very small features slightly protruding into cells. This prevents `dl_min` from becoming unreasonably small. If we do want to resolve such feature in certain cases, it can be done by turning on `convex_resolution` in `CornerFinderSpec`
- Changed the strategy of how snapping points are generated: previously, I would create only 1 snapping line per problematic grid spacing. Now, for a given column (or row) of cells, I look for the one with max number of found intersections and create (num_intersections - 1) snapping points to resolve all gaps/stripes with one go. This seems to reduce the total number of iteration to converge the grid.
Screenshot_20250506_213243

tidy3d/components/simulation.py
```diff
@@ -1142,10 +1147,40 @@ def grid(self) -> Grid:
             internal_override_structures=self.internal_override_structures,
         )
 
+        # This would AutoGrid the in-plane directions of the 2D materials
```
weiliangjin2021, line 1232: some typos?

tidy3d/components/grid/grid_spec.py
```diff
@@ -1037,6 +1043,19 @@ class LayerRefinementSpec(Box):
...
+        1,
+        title="Gap Meshing Iterations",
+        description="Number of recursive iterations for resolving thin gaps. "
+        "The underlying algorithm detects gaps contained in a single cell and places a snapping plane at the gaps's centers.",
+    )
+
+    dl_min_from_gap_width: bool = pd.Field(
+        True,
+        title="Set ``dl_min`` from Gap Width",
+        description="Take into account autodetected minimal PEC gap width when determining ``dl_min``.",
```
weiliangjin2021: add description that it only takes effect if dl_min of AutoGrid is not set

tidy3d/components/simulation.py
```diff
@@ -1118,7 +1123,7 @@ def set_plot_params(boundary_edge, lim, side, thickness):
     #    return plot_sim_3d(self, width=width, height=height)
 
     @cached_property
-    def grid(self) -> Grid:
+    def _grid_and_snapping_lines(self) -> Grid:
```
weiliangjin2021: return type to be updated

tidy3d/components/grid/grid_spec.py
weiliangjin2021: (not to be considered in this PR)

weiliangjin2021 (approved):
Code looks good! Just one part not fully understand: how re-entering subcell features are filtered. Could you provide some visual explanations?

tidy3d/components/grid/grid_spec.py
```diff
@@ -1510,6 +1553,346 @@ def _override_structures_along_axis(
...
+                if len(data[0]) > 0:
+                    v_cells_ij.append(data[0])
+                    v_cells_dy.append(data[1])
+
+                if len(data[2]) > 0:
+                    h_cells_ij.append(data[2])
+                    h_cells_dx.append(data[3])
+
+                # in case the polygon has holes
+                for poly_inner in poly.interiors:
```
dmarek-flex, line 1796: Add a unit test to test this case

dmarek-flex (approved):
Everything looks pretty good! Just one improvement for code coverage, and friendly reminder for Changelog.

Also I tried another notebook with the gap meshing. [WidebandBeamSteerableReflectarrayWithPRUC.zip](https://github.com/user-attachments/files/20172145/WidebandBeamSteerableReflectarrayWithPRUC.zip)

I assumed it would place snapping lines in the gap between simulation boundary and the arrows. Instead it gives:

image

And when I use many more iterations:

image

Is this expected?

dbochkov-flexcompute:
> Is this expected?

yeah, currently it ignores boundary conditions. I should add that. In this example, it's periodic I assume? I think need to add for PEC and PMC boundary conditions as well

dmarek-flex:
> > Is this expected?
> 
> yeah, currently it ignores boundary conditions. I should add that. In this example, it's periodic I assume? I think need to add for PEC and PMC boundary conditions as well

yea periodic

dbochkov-flexcompute:
> > > Is this expected?
> > 
> > 
> > yeah, currently it ignores boundary conditions. I should add that. In this example, it's periodic I assume? I think need to add for PEC and PMC boundary conditions as well
> 
> yea periodic

Actually, just realized, that even with proper boundary treatment there won't be any additional grid lines in this example. Because  the edges across periodic sides are already separated by a grid line (=simulation domain edge)

dbochkov-flexcompute:
> Code looks good! Just one part not fully understand: how re-entering subcell features are filtered. Could you provide some visual explanations?

@weiliangjin2021 Damian's case is a good example of that: note those sharp corners that cross the same grid segment before crossing anything else, this is what I call a small re-entry feature. 

Screenshot_20250512_161047

When we look for intersections between PEC polygons and grid lines, we go around the polygon's perimeter one segment after another. So, when such a small feature intersects the same edge twice, the array that stores indices of intersected cells will have two consecutive elements with the same value. Such elements are removed from array of intesections.

Btw, in Damian's comment above these small features were still resolved after two iterations :sweat_smile:, but I already found a bag, and now grid stay the same after the first iteration

tidy3d/components/simulation.py
dbochkov-flexcompute: hmm, not sure what this means exactly, copied it from the old function

dbochkov-flexcompute:
Should be ready to go. This is how it works now for boundaries:
- refine is PEC or PMC
Screenshot_20250512_230429

- no refine if periodic and separated
Screenshot_20250512_161047

- refine if periodic and no gap on one side
Screenshot_20250512_230414

dmarek-flex:
Great I think its ready, thanks for handling BCs!

PR #2391, opened 2025-04-18
:timer_clock: Reduce CI tests trigger frequency

daquinteroflex:
When draft
image

When opened
image
image

Note it triggers correctly in a review
image

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -3,8 +3,12 @@ name: "tidy3d-frontend-tests"
 on:
   workflow_dispatch:
     inputs:
-      run_tests:
-        description: 'Run test suite'
+      remote_tests:
+        description: 'remote-tests'
+        type: boolean
+        default: true
+      local_tests:
+        description: 'local-tests'
         type: boolean
         default: false
   push:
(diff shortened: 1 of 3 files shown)


lei-flex (dismissed):
ship it

.github/workflows/run_tests.yml
```diff
@@ -3,8 +3,12 @@ name: "tidy3d-frontend-tests"
...
   workflow_dispatch:
     inputs:
-      run_tests:
-        description: 'Run test suite'
+      remote_tests:
+        description: 'remote Test Suite'
+        type: boolean
+        default: true
+      local_tests:
+        description: 'local Test suite'
```
yaugenst-flex: super nitpicky but the inconsistent spelling irks me 😅

.github/workflows/run_tests.yml
```diff
@@ -32,15 +108,69 @@ jobs:
...
+    - name: Run tests with coverage
+      env:
+        PYTHONUNBUFFERED: "1"
+      run: |
+        poetry run pytest --cov=tidy3d -rF --tb=short tests/_test_data/_test_datasets_no_vtk.py
+        poetry run pytest --cov=tidy3d -rF --tb=short tests
+        poetry run coverage report -m
+        TOTAL_COVERAGE=$(poetry run coverage report --format=total)
+        echo "total=$TOTAL_COVERAGE" >> "$GITHUB_ENV"
+        echo "### Total coverage: ${TOTAL_COVERAGE}%"
```
yaugenst-flex, line 163: I guess we probably don't need the coverage in local tests? Doesn't hurt to keep either, but could be a little optimization.

.github/workflows/run_tests.yml
```diff
@@ -32,15 +108,69 @@ jobs:
...
+    # Run tests on a push event OR a workflow dispatch with remote_tests
+    needs: determine-test-scope 
+    if: needs.determine-test-scope.outputs.remote_tests == 'true'
     name: Python ${{ matrix.python-version }} - ${{ matrix.platform }}
     runs-on: ${{ matrix.platform }}
     strategy:
       matrix:
         python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
-        platform: [ubuntu-latest, windows-latest, macos-latest]
+        platform: [windows-latest, macos-latest]
```
yaugenst-flex: Oh interesting. So we don't run linux here? because it was run before already or do they still get submitted but to our workers?

.github/workflows/run_tests.yml
```diff
@@ -14,12 +18,84 @@ on:
       - latest
       - develop
       - 'pre/*'
+    types: ['opened', 'reopened', 'synchronize', 'ready_for_review']
+  pull_request_review:
+    types: [submitted]
```
yaugenst-flex, line 23: So here the test suite will trigger if the review was "changes request" or just "comment" I think. We could add something like `if: github.event.review.state == 'approved'` to make the tests only trigger if the PR is approved?

yaugenst-flex (dismissed):
Thanks a lot @daquinteroflex left some comments but this looks good to me. I'll try to debug the dev1 performance a bit but I think the workflow wouldn't need changes.

.github/workflows/run_tests.yml
daquinteroflex: Haha I feel very similarly, this was a Friday evening rushjob so fixing it now before merging it

.github/workflows/run_tests.yml
daquinteroflex: So I've brought it back for now. I had removed it because the "local-tests" would be equivalent ubuntu tests originally for all python versions. But since dev1 was already clogged up, I reduced the scope of the local tests now. This means for full coverage adding ubuntu back into the "remote-tests" makes sense. Maybe in the future if we get more worker bandwidth we can move all ubuntu tests to be local.

.github/workflows/run_tests.yml
daquinteroflex: Thanks for raising this because it made me realise it was set up incorrectly originally, because I had customised this here. The intended behaviour is that yeah we skip running tests on anything but an approval, and then run full tests. However, I understand it is not possible to trigger the action based on a pull request only approval event. So this is why this line here is meant to filter out any pull-request-review events that are not approvals and skip accordingly. https://github.com/flexcompute/tidy3d/blob/ece2309aa7d6775776709f1991cf19f280dc7f2a/.github/workflows/run_tests.yml#L68-L77

daquinteroflex:
Fyi @yaugenst-flex https://github.com/flexcompute/tidy3d/actions/runs/14570333503/job/40866349516?pr=2391 I guess this is what you'll debug in dev1 from the xdist timeout error? I'll comment out this error for now till you return from holiday as the rest of the PR is ready as far as I've been able to test.

.github/workflows/run_tests.yml
daquinteroflex: So was thinking coverage tests were going to run by default anyway due to the pyproject.toml? So it'd be mainly the report which is pretty fast after the tests have run anyway? Since you may be looking at this when you're back tomorrow when testing in dev1 for speedup maybe feel free to change from what you find then?

daquinteroflex:
So seems like the enforcement logic is valid https://github.com/flexcompute/tidy3d/actions/runs/14571373946 . Full tests would trigger on "final approval". We can always change it so let's try this out for now. Sending a message in the slack explaining it:

> This is now enforced as was required by infra, we're working on improving it anyways. Again, open to changing it, but this seemed to make enable what we wanted and make it work. So a few things to note:
> 
> Running tests via Github will become a lot more annoying now. Encourage to test locally first, and when those pass then un-draft the PR. local-tests queue times might be considerable until we can run on more workers when we enable the docker CI. 
> local-tests will run ubuntu tests on our workers. remote-tests run the entire test suite on all the OS platforms we support. You can run both at the same time, or separately, as they are non-exclusive.
> There is a blocking check called pr-requirements-pass that will stop any PRs from being merged unless the remote-tests and lint checks have passed. These have to be manually triggered from Github actions  or will be automatically triggered when a PR is approved. The latest master state need to be updated or rebased on each PR too. 
> Unlike before, a final approval on the last commit state will be required to merge. This final approval is what will trigger the pr-requirements-pass flag to turn green, since this will block merging. Without enforcing this, we would enable "fast checks green-tick PRs" to be merged without running the full suite. Problem is that there's no other way to manually trigger the full check when ready to merge from what I saw..
> 
> We can modify this, again open to feedback. I just think this satisfies the requirements of what we want to do as I understand them. But we can see what we feel like as we try it out.

PR #2393, opened 2025-04-21
GaussianPulse.from_frequency_range for maximizing amplitude in frequency range of interest

weiliangjin2021:
When DC component is removed, Gaussian pulse is highly asymmetric in frequency. This is particularly an issue when the frequency bandwidth is comparable to central frequency.

This PR finds optimal `freq0` and `fwidth` from frequency range (`fmin` and `fmax`) for maximizing the amplitude in the frequency range of interest. An example is shown below: in frequency range [1, 10] GHz, naively setting `freq0=5.5 GHz` and `fwidth=4.5 GHz` will result in tiny amplitude at 1GHz, and the pulse is very asymmetric in the frequency range; the other pulse from the new classmethod leads to a more symmetric pulse and higher amplitude. The implementation is based on the notes [here](https://github.com/flexcompute/tidy3d-technotes/blob/main/012_optimal_gaussian_parameters.pdf).

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Periodic repetition of EME subgrids via `num_reps` or `EMEPeriodicitySweep`.
 - Methods `EMEExplicitGrid.from_structures` and `EMECompositeGrid.from_structure_groups` to place EME cell boundaries at structure bounds.
 - 'ModeSimulation' now supports 'PermittivityMonitor'.
+- Classmethod `from_frequency_range` in `GaussianPulse` for generating a pulse whose amplitude in the frequency_range [fmin, fmax] is maximized, which is particularly useful for running broadband simulations.
 
 ### Changed
 - Performance enhancement for adjoint gradient calculations by optimizing field interpolation.
--- a/tests/test_components/test_source.py
+++ b/tests/test_components/test_source.py
@@ -93,6 +93,47 @@ def test_source_times():
     assert abs(dc_comp) ** 2 > 1e-32
 
 
+def test_gaussian_from_frequency_range():
+    # error with negative fmin
+    with pytest.raises(ValueError):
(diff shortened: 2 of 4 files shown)


weiliangjin2021:
@momchil-flex I'm also hoping to modify https://github.com/flexcompute/tidy3d/blob/981c0ebb22ba4dd317818ebc3729edb1e0005f2f/tidy3d/components/source/time.py#L103-L121 based on the cutoff amplitude. The current formula is only valid for remove_dc=False or narrowband source.

momchil-flex:
> @momchil-flex I'm also hoping to modify
> 
> https://github.com/flexcompute/tidy3d/blob/981c0ebb22ba4dd317818ebc3729edb1e0005f2f/tidy3d/components/source/time.py#L103-L121
> 
> based on the cutoff amplitude. The current formula is only valid for remove_dc=False or narrowband source.

Could you explore what this would affect? Is it e.g. just some warnings that may start/stop appearing, or is there also something physical? If the latter then we could postpone it for 2.9

momchil-flex (approved):
Great!

yuanshen-flexcompute (approved):
Looks good. Should this be the default method for GaussianPulse creation? Naively, it appears that it should also work for high-frequency, narrow-band.

dmarek-flex (approved):
Thanks for the notes!

weiliangjin2021:
> Could you explore what this would affect? Is it e.g. just some warnings that may start/stop appearing, or is there also something physical? If the latter then we could postpone it for 2.9

I'm most concerned about Chebyshev grid: https://github.com/flexcompute/tidy3d/blob/373739c4fcb79268cb4317fe6b844d57c9facff9/tidy3d/components/source/field.py#L111-L114 We can postpone it to 2.9

weiliangjin2021:
> Looks good. Should this be the default method for GaussianPulse creation? Naively, it appears that it should also work for high-frequency, narrow-band.

Yes, it works for high-frequency narrow-band. But people might already get used to using freq0 and fwidth in that region.

dmarek-flex (approved):

PR #2395, opened 2025-04-22
:wrench: PR requirements pass on `local-tests`

daquinteroflex:
- [x] Enables venv based testing (non-poetry)
- [x] Extends the workers we can use to test.


daquinteroflex:
Feel it's ready btw

yaugenst-flex (approved):
Thanks @daquinteroflex look good!

PR #2397, opened 2025-04-22
feat[FieldData]: export to ZBF

### LINKED ISSUE #2321: Add zbf file export in 2.9
After discussion with tyler, momchil and all we think it might be worth it to officially add the zbf file export to our stable version in 2.9, even though we can't fully test it. At the same time we can add a warning like "this is an experimental feature, if you encounter any issues with the exported file, please contact us".

bzhangflex:
- Adds a `to_zbf()` method to `FieldData` for exporting E fields to a ZBF file.
- Adds a `from_zbf()` method to `FieldDataset` for importing E fields from a ZBF file.
- Adds a `ZBFData` class for reading ZBF files

Closes https://github.com/flexcompute/tidy3d/issues/2321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Differentiable function `td.plugins.autograd.interpolate_spline` for 1D linear, quadratic, and cubic spline interpolation, supporting differentiation with respect to the interpolated values (`y_points`) and optional endpoint derivative constraints.
 - `SteadyEnergyBandMonitor` in the Charge solver.
 - Pretty printing enabled with `rich.print` for the material library, materials, and their variants. In notebooks, this can be accessed using `rich.print` or `display`, or by evaluating the material library, a material, or a variant in a cell.
+- `FieldData` and `ModeData` support exporting E fields to a Zemax Beam File (ZBF) with `.to_zbf()` (warning: experimental feature).
+- `FieldDataset` supports reading E fields from a Zemax Beam File (ZBF) with `.from_zbf()` (warning: experimental feature).
 
 ### Changed
 - Performance enhancement for adjoint gradient calculations by optimizing field interpolation.
--- a/tests/test_data/test_monitor_data.py
+++ b/tests/test_data/test_monitor_data.py
@@ -21,9 +21,11 @@ from tidy3d.components.data.monitor_data import (
     ModeData,
     PermittivityData,
 )
+from tidy3d.components.data.zbf import ZBFData
+from tidy3d.constants import UnitScaling
 from tidy3d.exceptions import DataError
 
-from ..utils import AssertLogLevel
+from ..utils import AssertLogLevel, run_emulated
 from .test_data_arrays import (
     AUX_FIELD_TIME_MONITOR,
     DIFFRACTION_MONITOR,
@@ -865,3 +867,190 @@ def test_no_nans():
     )
     with pytest.raises(pydantic.ValidationError):
         td.CustomMedium(eps_dataset=eps_dataset_nan)
+
+
+class TestZBF:
+    """Tests exporting field data to a zbf file"""
+
+    freq0 = td.C_0 / 0.75
+    freqs = (freq0, freq0 * 1.01)
+
+    def simdata(self, monitor) -> td.SimulationData:
+        """Returns emulated simulation data"""
+        source = td.PointDipole(
+            center=(-1.5, 0, 0),
+            source_time=td.GaussianPulse(freq0=self.freq0, fwidth=self.freq0 / 10.0),
+            polarization="Ey",
+        )
+        sim = td.Simulation(
+            size=(4, 3, 3),
+            grid_spec=td.GridSpec.auto(min_steps_per_wvl=10),
+            structures=[],
+            sources=[source],
+            monitors=[monitor],
+            run_time=120 / self.freq0,
+        )
+        return run_emulated(sim)
+
+    @pytest.fixture(scope="class")
+    def field_data(self) -> td.FieldData:
+        """Make random field data from an emulated simulation run."""
+        monitor = td.FieldMonitor(
+            size=(td.inf, td.inf, 0),
+            freqs=self.freqs,
+            name="fields",
+            colocate=True,
+        )
+        return self.simdata(monitor)["fields"]
+
+    @pytest.fixture(scope="class")
+    def mode_data(self) -> td.ModeData:
+        """Make random ModeData from an emulated simulation run."""
+        monitor = td.ModeMonitor(
+            size=(td.inf, td.inf, 0),
+            freqs=self.freqs,
+            name="modes",
+            colocate=True,
+            mode_spec=td.ModeSpec(num_modes=2, target_neff=4.0),
+            store_fields_direction="+",
+        )
+        return self.simdata(monitor)["modes"]
+
+    @pytest.mark.parametrize("background_index", [1, 2, 3])
+    @pytest.mark.parametrize("freq", list(freqs) + [None])
+    @pytest.mark.parametrize("n_x", [2**5, 2**6])
+    @pytest.mark.parametrize("n_y", [2**5, 2**6])
+    @pytest.mark.parametrize("units", ["mm", "cm", "in", "m"])
+    def test_fielddata_tozbf_readzbf(
+        self, tmp_path, field_data, background_index, freq, n_x, n_y, units
+    ):
+        """Test that FieldData.to_zbf() -> ZBFData.read_zbf() works"""
+        zbf_filename = tmp_path / "testzbf.zbf"
+
+        # write to zbf and then load it back in
+        ex, ey = field_data.to_zbf(
+            fname=zbf_filename,
+            background_refractive_index=background_index,
+            freq=freq,
+            n_x=n_x,
+            n_y=n_y,
+            units=units,
+        )
+        zbfdata = ZBFData.read_zbf(zbf_filename)
+
+        assert zbfdata.background_refractive_index == background_index
+
+        unitscaling = UnitScaling[units]
+
+        if freq is not None:
+            assert np.isclose(zbfdata.wavelength / unitscaling, td.C_0 / freq)
+        else:
+            assert np.isclose(
+                zbfdata.wavelength / unitscaling,
+                td.C_0 / np.mean(field_data.monitor.freqs),
+            )
+
+        assert zbfdata.nx == n_x
+        assert zbfdata.ny == n_y
+
+        # check that fields are close
+        assert np.allclose(ex.values, zbfdata.Ex)
+        assert np.allclose(ey.values, zbfdata.Ey)
+
+    @pytest.mark.parametrize("mode_index", [0, 1])
+    def test_tozbf_modedata(self, tmp_path, mode_data, mode_index):
+        """Tests ModeData.to_zbf()"""
+        zbf_filename = tmp_path / "testzbf_modedata.zbf"
+
+        # write to zbf and then load it back in
+        ex, ey = mode_data.to_zbf(
+            fname=zbf_filename,
+            background_refractive_index=1,
+            freq=self.freq0,
+            mode_index=mode_index,
+            n_x=32,
+            n_y=32,
+            units="mm",
+        )
+        zbfdata = ZBFData.read_zbf(zbf_filename)
+
+        # check that fields are close
+        assert np.allclose(ex.values, zbfdata.Ex)
+        assert np.allclose(ey.values, zbfdata.Ey)
+
+    def test_tozbf_modedata_fails(self, tmp_path, mode_data):
+        """Asserts that Modedata.to_zbf() fails if mode_index is not specified"""
+        with pytest.raises(ValueError) as e:
+            _ = mode_data.to_zbf(
+                fname=tmp_path / "testzbf_modedata_fail.zbf",
+                background_refractive_index=1,
+                freq=self.freq0,
+                mode_index=None,
+                n_x=32,
+                n_y=32,
+                units="mm",
+            )
+
+    @pytest.mark.parametrize("n_x", [16, 2**14, 33])
+    @pytest.mark.parametrize("n_y", [16, 2**14, 33])
+    def test_tozbf_nxny_fails(self, tmp_path, field_data, n_x, n_y):
+        """Asserts that to_zbf() fails when n_x and n_y are invalid values."""
+        with pytest.raises(ValueError) as e:
+            _ = field_data.to_zbf(
+                fname=tmp_path / "testzbf_nxny_fail.zbf",
+                background_refractive_index=1,
+                freq=self.freq0,
+                n_x=n_x,
+                n_y=n_y,
+                units="mm",
+            )
+
+    @pytest.mark.parametrize("units", ["mmm", "123"])
+    def test_tozbf_units_fails(self, tmp_path, field_data, units):
+        """Asserts that to_zbf() fails when units are invalid."""
+        with pytest.raises(ValueError) as e:
+            _ = field_data.to_zbf(
+                fname=tmp_path / "testzbf_nxny_fail.zbf",
+                background_refractive_index=1,
+                freq=self.freq0,
+                n_x=32,
+                n_y=32,
+                units=units,
+            )
+
+    def test_from_zbf(self, tmp_path, field_data):
+        """Tests creating a field dataset from a zbf"""
+        zbf_filename = tmp_path / "testzbf.zbf"
+        # write to zbf and then load it back in
+        ex, ey = field_data.to_zbf(
+            fname=zbf_filename,
+            background_refractive_index=1,
+            n_x=32,
+            n_y=32,
+            units="mm",
+        )
+
+        # create a field dataset from the zbf file
+        fd = td.FieldDataset.from_zbf(filename=zbf_filename, dim1="x", dim2="y")
+
+        # compare loaded field data to saved data
+        assert np.allclose(ex.values, fd.Ex.values.squeeze())
+        assert np.allclose(ey.values, fd.Ey.values.squeeze())
+
+    @pytest.mark.parametrize(
+        "dim1,dim2", [("x", "x"), ("y", "y"), ("z", "z"), ("1", "2"), ("c", "z")]
+    )
+    def test_from_zbf_dimsfail(self, tmp_path, field_data, dim1, dim2):
+        """Tests fail cases when the dimensions to populate are wrong."""
+        zbf_filename = tmp_path / "testzbf.zbf"
+        # write to zbf and then load it back in
+        _, _ = field_data.to_zbf(
+            fname=zbf_filename,
+            background_refractive_index=1,
+            n_x=32,
+            n_y=32,
+            units="mm",
+        )
+        # this should fail
+        with pytest.raises(ValueError) as e:
+            _ = td.FieldDataset.from_zbf(filename=zbf_filename, dim1=dim1, dim2=dim2)
--- a/tidy3d/components/data/dataset.py
+++ b/tidy3d/components/data/dataset.py
@@ -3,17 +3,17 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, Optional, Union
+from typing import Any, Callable, Dict, Optional, Union, get_args
 
 import numpy as np
 import pydantic.v1 as pd
 import xarray as xr
 
-from ...constants import PICOSECOND_PER_NANOMETER_PER_KILOMETER
+from ...constants import C_0, PICOSECOND_PER_NANOMETER_PER_KILOMETER, UnitScaling
 from ...exceptions import DataError
 from ...log import log
 from ..base import Tidy3dBaseModel
-from ..types import Axis
+from ..types import Axis, xyz
 from .data_array import (
     DataArray,
     EMEScalarFieldDataArray,
@@ -28,6 +28,7 @@ from .data_array import (
     TimeDataArray,
     TriangleMeshDataArray,
 )
+from .zbf import ZBFData
 
 DEFAULT_MAX_SAMPLES_PER_STEP = 10_000
 DEFAULT_MAX_CELLS_PER_STEP = 10_000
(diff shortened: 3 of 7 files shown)


tests/test_data/test_monitor_data.py
yaugenst-flex, line 1: Nice! Two comments on the tests:

1. It looks like you already implemented all the logic for reading a zbf file here. Should we also add a `FieldDataset.from_zbf()`? That'd make it really easy to create custom field sources from zbf data.
2. Instead of re-implementing dummy field data generation here it's probably cleaner to set up a simulation and use the emulated run to get the field data? That'd also be a bit more representative of the actual workflow I guess.

tidy3d/components/data/monitor_data.py
```diff
@@ -1224,6 +1225,127 @@ def make_adjoint_sources(
 
         return sources
 
+    def to_zbf(
+        self,
+        fname: str,
+        background_refractive_index: float,
+        freq: Optional[float] = None,
+        mode_index: Optional[int] = 0,
```
yaugenst-flex: Looks like this is not really `Optional` but would always expect an `int`?

yaugenst-flex (commented):
Nice, looks great! Only some minor comments.

tests/test_data/test_monitor_data.py
```diff
@@ -865,3 +867,225 @@ def test_no_nans():
...
+
+class TestZBF:
+    """Tests exporting field data to a zbf file"""
+
+    simulation = SIM
+    monitor = FIELD_MONITOR_2D
+
+    def read_beam_file(self, filename: str) -> tuple:
+        """Read in a Zemax Beam file
+        Adapted from https://community.zemax.com/code-exchange-10/python-class-to-open-zbf-file-4845
```
lucas-flexcompute: Should we worry about the (lack of) license in the original code? Is it assumed to be public domain because it was shared in a forum? I'm not sure if that's how it works.

tests/test_data/test_monitor_data.py
```diff
@@ -865,3 +867,225 @@ def test_no_nans():
...
+        efield : 4-tuple of 2D lists, (Ex_real, Ex_imag, Ey_real, Ey_imag)
+            a tuple containing two dimensional lists with the real and
+            imaginary parts of the x and y polarizations of the beam
+
+        """
+        # will need to get rid of during automation.
+        # beamfilename = self.open_zbf()
+        f = open(filename, "rb")
+        # zemax version number
+        version = unpack("i", f.read(4))[0]
```
lucas-flexcompute: This use of `unpack` is not [setting alignment and size](https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment), so it will only work if the native settings match the ZBF specs. In my original implementation I fixed the sizes and left native order (`=`) because the manual did not specify it, but it should probably be set little-endian (`<`).

Also, it could probably be a lot more efficient if it unpacked more values at once:
```python
version, nx, ny, ispol, units = unpack("<5i", f.read(4 * 5))
```

lucas-flexcompute (approved):

tidy3d/components/data/monitor_data.py
```diff
@@ -1224,6 +1225,127 @@ def make_adjoint_sources(
...
+        fname : str
+            Full path to the .zbf file to be written.
+        background_refractive_index : float
+            Refractive index of the medium surrounding the monitor.
+        freq : float
+            Field frequency selection. If ``None``, the central one is used.
+        mode_index : int
+            Mode index selection.
+        num_samples: int
+            Number of samples to use. Must be a power of 2
```
tylerflex: could you explain a bit more what is being sampled? and what does the default value correspond to?

tidy3d/components/data/monitor_data.py
```diff
@@ -1224,6 +1225,127 @@ def make_adjoint_sources(
...
+
+        # Use the central frequency if freq is not specified
+        if freq is None:
+            freq = e_x.coords["f"].values[e_x.coords["f"].size // 2]
+
+        mode_area = mode_area.sel(f=freq, drop=True)
+        e_x = e_x.sel(f=freq, drop=True)
+        e_y = e_y.sel(f=freq, drop=True)
+
+        if "mode_index" in e_x.coords:
```
tylerflex: what case does this cover? I guess if the object is a `ModeSolverData`?  

If the `mode_index` is only used in this case, maybe it does actually make sense for the default value to be `None`? And then only do the `isel` here if it's not `None`?

@yaugenst-flex

tidy3d/components/data/monitor_data.py
```diff
@@ -1224,6 +1225,127 @@ def make_adjoint_sources(
...
+
+        # Using file-local coordinates x, y for the tangential components
+        e_x = self._tangential_fields["E" + dim1]
+        e_y = self._tangential_fields["E" + dim2]
+        x = e_x.coords[dim1].values
+        y = e_x.coords[dim2].values
+
+        # Use the central frequency if freq is not specified
+        if freq is None:
+            freq = e_x.coords["f"].values[e_x.coords["f"].size // 2]
```
tylerflex: do we need to worry if the number of frequency coords is even? wonder if we should rather interpolate to the mean frequency?

tidy3d/components/data/monitor_data.py
```diff
@@ -1224,6 +1225,127 @@ def make_adjoint_sources(
...
+
+        if "mode_index" in e_x.coords:
+            mode_area = mode_area.isel(mode_index=mode_index, drop=True)
+            e_x = e_x.isel(mode_index=mode_index, drop=True)
+            e_y = e_y.isel(mode_index=mode_index, drop=True)
+
+        # Header info
+        version = 1
+        polarized = 1
+        units = 0  # "mm": 0, "cm": 1, "in": 2, "m": 3
```
tylerflex: should this be something that the user can configure through the kwargs?

tidy3d/components/data/monitor_data.py
```diff
@@ -1224,6 +1225,127 @@ def make_adjoint_sources(
...
+        lda = C_0 / freq * 1e-3
```
tylerflex: this 1e-3 floating around here and below makes me a bit nervous. I think it would be better if we had something like 

```py
# kwarg length_unit = "mm"

length_scale_map = dict(mm=1e-3, cm=1e-2, in=...) # maps kwarg to length scale factor

if length_unit not in length_scale_map:
    raise KeyError(...)

length_scale = length_scale_map[length_unit]

```

and then use `length_scale` in place of 1e-3.

If there are free units floating around like this it can lead to some unexpected bugs.

tidy3d/components/data/monitor_data.py
```diff
@@ -1224,6 +1225,127 @@ def make_adjoint_sources(
...
+        # Pilot (reference) beam waist: use the mode area to approximate the expected value
+        w_x = (mode_area.item() / np.pi) ** 0.5 * 1e-3
+        w_y = w_x
+
+        # Pilot beam Rayleigh distance (ignored on input)
+        r_x = 0
+        r_y = 0
+
+        # Pilot beam z position w.r.t. the waist
+        z_x = 0 * 1e-3
```
tylerflex: what's with the 0 * 1e-3? are we planning to allow this to be modified eventually?

tidy3d/components/data/monitor_data.py
```diff
@@ -1224,6 +1225,127 @@ def make_adjoint_sources(
...
+        z_y = 0 * 1e-3
+
+        # Number of samples: n_x = 2 ** a, a >= 5 and a <= 13 (32-bit), or a <= 14 (64-bit)
+        if num_samples is not None:
+            # custom user defined number of samples
+            # Raise an error if the number of samples is not a power of 2
```
tylerflex: can you explain why the num_samples needs to be a factor of 2? in principle we could interpolate using however points they provide?

tidy3d/components/data/monitor_data.py
```diff
@@ -1224,6 +1225,127 @@ def make_adjoint_sources(
...
+            if (num_samples & (num_samples - 1)) != 0:
+                raise ValueError("num_samples must be a power of 2.")
+            n_x = num_samples
+            n_y = num_samples
+        else:
+            n_x = 2 ** min(13, max(5, int(np.log2(x.size) + 1)))
```
tylerflex: should these `5` and `13` be hardcoded somewhere? eg as a module level constant? or configurable? not sure exactly where they come from. is it a requirement of zemax?

CHANGELOG.md
tylerflex, line 1: it might be worth also adding this to the FAQ section of the docs, as we recently had some users unaware of our export to MATLAB. so probably some extra information about this capability in the docs wouldn't hurt. Also, not sure if it applies, but it might be nice also to add it to at least one notebook example.

tylerflex (changes requested):
Thanks @bzhangflex , this looks great! Will be super useful and a highly requested feature!

I had a few minor comments. I'm not familiar with zemax configuration so take what I say with a bit of a grain of salt.

tomflexcompute:
Since this is a bit experimental and largely untested, there might be cases that it will not work properly. Should we add a warning when it's called? Something like "This is an experimental feature. If you run into any issues, contact us at flexcompute.com/tidy3d/technical-support/"

tests/test_data/test_monitor_data.py
bzhangflex: For 2. are you referring to this for the emulated run? https://github.com/flexcompute/tidy3d/blob/b848f5da93b82bb765fbb86e165db52c89650042/tests/utils.py#L920

tests/test_data/test_monitor_data.py
bzhangflex: Good question, I'm not sure either 😅

tidy3d/components/data/monitor_data.py
bzhangflex: Good question, I assumed it was for `ModeSolverData` but I haven't tested it yet. I'll test it out and also change the default value

tidy3d/components/data/monitor_data.py
bzhangflex: good catch! I'm not fully sure what the pilot beam is used for (seems like it helps [guide the actual beam](https://community.zemax.com/got-a-question-7/pilot-and-beam-width-549)) but it sounds like a good idea to let the user define values for it.

tidy3d/components/data/monitor_data.py
bzhangflex: it's a requirement set by zemax. My guess is it might have to do with FFTs in their beam propagation algorithm.

Should we not enforce it here and let the user choose an arbitrary number? That could be fine too if Zemax emits a clear error message, or changes behavior in the future

tidy3d/components/data/monitor_data.py
bzhangflex: hmm it was left in from @lucas-flexcompute's code, I wonder if there was a specific reason for the values. If not maybe we don't specify any default `n_x` or `n_y`?

tidy3d/components/data/monitor_data.py
lucas-flexcompute: This is a limitation in Zemax. It must be a power of 2.

tidy3d/components/data/monitor_data.py
lucas-flexcompute: The Zemax manual I have (dated July 8, 2011) defines that the number of samples must be a power of 2 between 32 and 8192 (inclusive)

tidy3d/components/data/monitor_data.py
tylerflex: it's probably fine to enforce it here. maybe we can just mention in the error message that this is a requirement for zemax

tidy3d/components/data/monitor_data.py
tylerflex: if so maybe we can mention this in the changelog too

tests/test_data/test_monitor_data.py
yaugenst-flex: yea exactly

tests/test_data/test_monitor_data.py
bzhangflex: I reworked the read zbf function and put it here https://github.com/flexcompute/tidy3d/blob/88a7a20fb11e9a5f6cd84dd04815e8b1feccb747/tidy3d/components/data/zbf.py#L8

tidy3d/components/data/monitor_data.py
bzhangflex: now it interpolates to the mean frequency https://github.com/flexcompute/tidy3d/blob/fe6faac2f3d5b27c573866db970d2e5edd28fb83/tidy3d/components/data/monitor_data.py#L1159

bzhangflex:
Pushed quite a few changes
- Created a read_zbf() utility function https://github.com/flexcompute/tidy3d/blob/fe6faac2f3d5b27c573866db970d2e5edd28fb83/tidy3d/components/data/zbf.py#L8
- Added `FieldDataset.from_zbf()` for creating a new `FieldDataset` from a zbf file
- In `to_zbf()`, require the user to choose the # of field samples `n_x` and `n_y` separately. Checks that they are powers of two between 2^5 and 2^13 per Zemax manual
- If `freq` in `to_zbf()` is not specified, default to the mean of the frequencies
- `units` is now a kwarg
- User has the option to specify pilot beam parameters. Defaults are 0
- Updated changelog
- Updated warning messages to include support url
- use emulated_run in the pytests

CHANGELOG.md
bzhangflex: maybe dumb question, how do I update the FAQ?

tylerflex: I think you add one [here](https://github.com/flexcompute/tidy3d-faq/tree/9b8479c00a7b2d5fb81a90dd21c7b8e92423ef58/_faqs) and then link it in the proper .rst [file](https://github.com/flexcompute/tidy3d-faq/tree/9b8479c00a7b2d5fb81a90dd21c7b8e92423ef58/docs)

tidy3d/components/data/dataset.py
```diff
@@ -257,6 +258,90 @@ class FieldDataset(ElectromagneticFieldDataset):
...
+            Tangential field component to map the y-axis of the zbf data to.
+
+        Returns
+        -------
+        :class:`.FieldDataset`
+            A :class:`.FieldDataset` object with two tangential E field components populated
+            by zbf data.
+        """
+        log.warning(
+            "`FieldDataset.from_zbf()` is currently an experimental feature."
```
tylerflex: it's confusing but for console outputs you will want the code to be in regular single quotes like '

tidy3d/components/data/dataset.py
```diff
@@ -257,6 +258,90 @@ class FieldDataset(ElectromagneticFieldDataset):
...
+            " If any issues are encountered, please contact Flexcompute support https://www.flexcompute.com/tidy3d/technical-support/"
```
tylerflex: let's also put the url in the single quotes

tidy3d/components/data/dataset.py
tylerflex: this comes up in the ValueErrors and things below, so just double check all of them 👍

tylerflex: https://www.notion.so/flexcompute/Python-Style-Guide-afac5dbbf949463e89c18fe39b7453c9

tidy3d/components/data/monitor_data.py
```diff
@@ -1066,6 +1068,157 @@ def translated_copy(self, vector: Coordinate) -> ElectromagneticFieldData:
             **field_kwargs,
         )
 
+    def to_zbf(
+        self,
+        fname: str,
+        background_refractive_index: float,
+        n_x: int,
+        n_y: int,
+        units: UnitsZBF,
```
tylerflex: do we want to set defaults for units or background ref_index? (not sure just wondering)

tidy3d/components/data/monitor_data.py
```diff
@@ -1066,6 +1068,157 @@ def translated_copy(self, vector: Coordinate) -> ElectromagneticFieldData:
...
+        ----------
+        fname : str
+            Full path to the .zbf file to be written.
+        background_refractive_index : float
+            Refractive index of the medium surrounding the monitor.
+        n_x : int
+            Number of field samples along x. Must be a power of 2, between 2^5 and 2^13 inclusive per Zemax's requirements.
+        n_y : int
+            Number of field samples along y. Must be a power of 2, between 2^5 and 2^13 inclusive per Zemax's requirements.
+        units : `UnitsZBF`
```
tylerflex: double check the style guide. I think we use double backslash quotes inside of the docstrings or anything rendered by docs basically.  Also would be good to add to `.zbf` where it comes up.

tidy3d/components/data/monitor_data.py
```diff
@@ -1066,6 +1068,157 @@ def translated_copy(self, vector: Coordinate) -> ElectromagneticFieldData:
...
+        mode_area = self.mode_area
+        dim1, dim2 = self._tangential_dims
+
+        # Using file-local coordinates x, y for the tangential components
+        e_x = self._tangential_fields["E" + dim1]
+        e_y = self._tangential_fields["E" + dim2]
+        x = e_x.coords[dim1].values
+        y = e_x.coords[dim2].values
+
+        # Use the mean frequency if freq is not specified
```
tylerflex, line 1150: would it be worth warning in this case? or not needed?

tidy3d/components/data/monitor_data.py
```diff
@@ -1066,6 +1068,157 @@ def translated_copy(self, vector: Coordinate) -> ElectromagneticFieldData:
...
+            freq = np.mean(e_x.coords["f"].values)
+
+        mode_area = mode_area.interp(f=freq)
+        e_x = e_x.interp(f=freq)
+        e_y = e_y.interp(f=freq)
+
+        # If the data is ModeData, choose one of the modes to save
+        if "mode_index" in e_x.coords:
+            if mode_index is None:
+                raise ValueError("`mode_index` is required for `ModeData.to_zbf()`")
```
tylerflex: regular single quotes

tidy3d/components/data/zbf.py
```diff
@@ -0,0 +1,128 @@
+"""ZBF utilities"""
+
+from struct import unpack
+
+import numpy as np
+
+
+def read_zbf(filename: str):
```
tylerflex: add return type annotation, just `dict` is probably fine

tidy3d/components/data/zbf.py
```diff
@@ -0,0 +1,128 @@
...
+        Ey_real = np.asarray(rawy[0::2]).reshape(nx, ny)
+        Ey_imag = np.asarray(rawy[1::2]).reshape(nx, ny)
+    else:
+        Ey_real = np.zeros((nx, ny))
+        Ey_imag = np.zeros((nx, ny))
+
+    Ex = Ex_real + 1j * Ex_imag
+    Ey = Ey_real + 1j * Ey_imag
+
+    return {
```
tylerflex: this is probably not a "must do" but more a "highly recommended": To make things more robust, I'd create a simple data structure for it in pydantic, 

```py
class ZBFData(Tidy3DBaseModel):
    version : str = pd.Field(..., title='Version', description=...)
    ...
```
and then you can return an instance of this instead of a `dict`

basically it offloads what wrote out by hand docstring to a model that we can reference in the doc and will do some validation and things. In general it's a nice approach we use to keep it more "tidy" when passing data with a lot of fields around. otherwise it can get quite complex

tylerflex (changes requested):
Looking like it's almost there! mostly minor comments

tidy3d/constants.py
```diff
@@ -238,6 +238,7 @@
         "mm": 1e-3,
         "cm": 1e-4,
         "m": 1e-6,
+        "in": 1.0 / 25400,
```
tylerflex, line 241: this data structure is a bit unfamiliar to me. I was confused about what it was mapping. I guess for a key `k`, the value is basically how much 1 um is in units of `k`? maybe can you add a short sentence explaining a bit more?

tidy3d/components/data/monitor_data.py
lucas-flexcompute: Just a minor comment: in the previous implementation, they did have default values, which I think is great to reduce user overload. Same for `n_x` and `n_y`. However, they were provided based on "reasonable defaults", which can be improved depending on the use case.

tidy3d/components/data/dataset.py
bzhangflex: Thanks! I think I've corrected all the formatting issues

tidy3d/components/data/monitor_data.py
bzhangflex: good point, added the defaults back in, they are probably good to have

tidy3d/components/data/monitor_data.py
bzhangflex: good idea, I added a warning

tidy3d/components/data/zbf.py
bzhangflex: good call, I created a pydantic data structure, let me know if it looks ok https://github.com/flexcompute/tidy3d/blob/7533e425c22dfcd540a9d4514c5ca1cc3a2f7cd8/tidy3d/components/data/zbf.py#L13

tidy3d/constants.py
bzhangflex: yup, updated the docstring

bzhangflex:
Thanks @tylerflex!, let me know how it looks now

CHANGELOG.md
bzhangflex: Created a [PR](https://github.com/flexcompute/tidy3d-faq/pull/29), keeping as a draft until this is merged and the docs are updated, so I can link to the source in the FAQ

tylerflex (approved):
looks good, thanks @bzhangflex !

tidy3d/components/data/dataset.py
```diff
@@ -257,6 +258,92 @@ class FieldDataset(ElectromagneticFieldDataset):
...
+        Returns
+        -------
+        :class:`.FieldDataset`
+            A :class:`.FieldDataset` object with two tangential E field components populated
+            by zbf data.
+
+        See Also
+        --------
+        :class:`.ZBFData':
+            A class containing data read in from a `.zbf` file.
```
tylerflex: double backslash here.

tidy3d/components/data/dataset.py
```diff
@@ -257,6 +258,92 @@ class FieldDataset(ElectromagneticFieldDataset):
         description="Spatial distribution of the z-component of the magnetic field.",
     )
 
+    def from_zbf(filename: str, dim1: xyz, dim2: xyz) -> FieldDataset:
+        """Creates a :class:`.FieldDataset` from a Zemax Beam File (`.zbf`).
```
tylerflex: double backslash here.

tidy3d/components/data/dataset.py
bzhangflex: oops thanks

PR #2398, opened 2025-04-23
Vasily/refinement_region_clean

vasilyzabelin:
Clean implementation of the mesh refinement structures

### DIFF
--- a/tidy3d/components/tcad/grid.py
+++ b/tidy3d/components/tcad/grid.py
@@ -5,11 +5,15 @@ from __future__ import annotations
 from abc import ABC
 from typing import Tuple, Union
 
+import numpy as np
 import pydantic.v1 as pd
 
 from tidy3d.components.base import Tidy3dBaseModel, skip_if_fields_missing
 from tidy3d.constants import MICROMETER
 from tidy3d.exceptions import ValidationError
+
+from ..geometry.base import Box
+from ..types import Coordinate, annotate_type
 
 
 class UnstructuredGrid(Tidy3dBaseModel, ABC):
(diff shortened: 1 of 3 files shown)


tidy3d/components/tcad/grid.py
```diff
@@ -126,6 +216,12 @@ class DistanceUnstructuredGrid(UnstructuredGrid):
         "``dl_bulk`` is used instead.",
     )
 
+    mesh_refinements: Tuple[Union[GridRefinementRegion, GridRefinementLine], ...] = pd.Field(
```
dbochkov-flexcompute: does this cause any sporadic error messages? We typically need to add `annotate_type` in such situations like here https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/simulation.py#L2289, otherwise pydantic would loop through all allowed classes when initializing this field and some of which fail and show an error message

dbochkov-flexcompute (commented):
This PR needs unit tests which checks that new classes can be successfully instantiated and that they also fail when expected

marc-flex (approved):

tidy3d/components/tcad/grid.py
vasilyzabelin: I can create a special type for `Union[GridRefinementRegion, GridRefinementLine]` like it is done in other places, if it is better.

dbochkov-flexcompute: no need for that,   just meant to wrap `Union[GridRefinementRegion, GridRefinementLine]` into `annotate_type(Union[GridRefinementRegion, GridRefinementLine])`

PR #2399, opened 2025-04-23
fix[autograd]: add boolean flag indicating adjoint runs

yaugenst-flex:


momchil-flex (approved):
Looks good and tested that it works in a pipeline test.

PR #2401, opened 2025-04-23
enh[web]: make pay type selection case insensitive

yaugenst-flex:
Small change that makes payment type selection case-insensitive. Would be nice to have in 2.8.3 so we don't have to change the public-facing interface later.


momchil-flex (approved):
Makes sense.

tomflexcompute (approved):

PR #2406, opened 2025-04-24
Making normal requirement for tilted plane intersections less stringent

momchil-flex:


tidy3d/components/geometry/base.py
```diff
@@ -1686,7 +1686,7 @@ def intersections_tilted_plane(
         """
 
         # Check if normal is a special case, where the normal is aligned with an axis.
-        if normal.count(0.0) == 2:
+        if np.sum(np.isclose(normal, 0.0)) == 2:
             axis = np.nonzero(normal)[0][0]
```
lucas-flexcompute: `nonzero` will return almost zero as valid directions. It is probably better to replace `axis = np.nonzero(normal)[0][0]` with `axis = np.argmax(np.abs(normal)).item()`

lucas-flexcompute (approved):

tidy3d/components/geometry/base.py
momchil-flex: Nice, great catch.

daquinteroflex:
Is it fine to rebase again @momchil-flex ? Then I can reapprove to test the change.

momchil-flex:
> Is it fine to rebase again @momchil-flex ? Then I can reapprove to test the change.

Need to implement the change Lucas suggested but I'll also wait for Weiliang's review first.

weiliangjin2021:
`normal` is constructed here https://github.com/flexcompute/tidy3d/blob/3366a7117d7656c144ff64c9b3034bd07a795f84/tidy3d/components/geometry/base.py#L257-L264 , so it should always have 2 "0"s. Is it corrupted by machine rounding error?

weiliangjin2021 (approved):

momchil-flex:
> `normal` is constructed here
> 
> https://github.com/flexcompute/tidy3d/blob/3366a7117d7656c144ff64c9b3034bd07a795f84/tidy3d/components/geometry/base.py#L257-L264
> 
> , so it should always have 2 "0"s. Is it corrupted by machine rounding error?

It's because the [normal for transformed geometries is also transformed](https://github.com/flexcompute/tidy3d/blob/3366a7117d7656c144ff64c9b3034bd07a795f84/tidy3d/components/geometry/base.py#L2755), which can introduce a machine rounding error yeah even for transformations that are meant to preserve it.

PR #2410, opened 2025-04-25
Add support for 2D EME simulations

caseyflex:
EME should support 2D simulations for consistency with FDTD and mode simulations.

To that end, this PR:

- Makes the mode planes in an EME simulation infinite; the mode solving still occurs in the intersection with the simulation geometry
- Allows EME simulations to have non-PEC boundary conditions. In general this can allow for more consistency with FDTD and mode simulations. The specific use case here though is periodic boundary conditions on size 0 dimensions.
- Adds the `boundaries_for_zero_dims` validator from `Simulation` to `EMESimulation` and `ModeSimulation`
- Keeps the default boundary as PEC for EME simulations and PML for mode simulations

This doesn't break backwards compatibility for EME simulations. It does break backwards compatibility for mode simulations, but those have not yet been "officially released" anyway. Specifically, if you try to create a 2D FDTD simulation and run the mode solver on a 1D cross-section, the FDTD simulation will fail validation if the boundary condition is PEC and not periodic in the 0-thickness direction. However, if you tried to create a mode simulation with the PEC boundary condition (a rather unlikely thing to do as the default is PML), it would previously have passed validation and then been converted to periodic boundary internally in `subsection`. Now, it will fail validation, consistently with the other setup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improved error message and handling when attempting to load a non-existent task ID.
 - `ClipOperation` now fails validation if traced fields are detected.
 - Warn if more than 20 frequencies are used in EME, as this may lead to slower or more expensive simulations.
+- EME now supports 2D simulations.
 
 ## [2.8.3] - 2025-04-24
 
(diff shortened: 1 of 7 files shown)


caseyflex:
@tylerflex FYI, we will be able to support 2D EME simulations after this (requires new solver version though)

tylerflex:
Yay! Thanks @caseyflex

yaugenst-flex (approved):
LGTM, thanks @caseyflex!

momchil-flex (approved):

PR #2413, opened 2025-04-25
Avoiding overflow in Simulation.num_cells

momchil-flex:
For wrong settings, it was possible to get overflow in the `Simulation.num_cells` even with 64-bit precision integers.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -4821,7 +4821,7 @@ class Simulation(AbstractYeeGridSimulation):
             Number of yee cells in the simulation.
         """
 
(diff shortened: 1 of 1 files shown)


yaugenst-flex (approved):
I guess this relates to the min grid spacing validator. from #2335

momchil-flex:
Well, kind of, but do those validators only check absolute bounds? In the problematic simulation I was investigating, the grid spacing was as if for ~1um wavelength, but the simulation domain was on the order of tens of mm in each direction. Neither of these is prohibited on its own, but the combination produces an immense amount of grid points. If it were not for overflow making the total number of cells negative, the simulation would have hit various validators, at the very least the pre-upload one.

Issue #2157, opened 2025-01-13
Reflection operation for geometries


tomflexcompute:
Currently we already support transformations including rotation, translation, and scaling. Another operation that users request is reflection off a plane or point. Would if be possible to add this transformation?

lucas-flexcompute:
Hi Tom, if I'm not mistaken, reflection is just scaling with a negative factor. The original implementation supported that, but I'm not sure autograd does. @yaugenst thoughts?

yaugenst-flex:
Yeah we are working on autograd support for geometric transformations, shouldn't take too long. I guess reflection is mostly a question about API, since yeah scaling is more general. But it might make sense to introduce a `reflected` method that just uses `scaled` under the hood?

lucas-flexcompute:
Sure, I see no problem there. Do you think Rahul could tacke that together with his work on autograf support? With PhotonForge close to release, I just don't have the bandwidth to work on it.

yaugenst-flex:
Sounds good!

momchil-flex:
> Another operation that users request is reflection off a plane or point

Just a quick note for future reference, reflection off a point doesn't make sense I think, but yeah I think usually the general way to define this is w.r.t. a plane passing through the origin, similar to how we define rotations around an axis passing through the origin. The reflection matrix is orthogonal like the rotation matrix but with determinant -1. It's only in the special case of reflecting w.r.t. a plane normal to one of the x, y, or z axes that we get a scaling. But more generally we could handle it similarly to rotation where an axis is provided, which defines the plane as passing through the origin and normal to the axis. The corresponding orthogonal matrix is defined here. https://en.wikipedia.org/wiki/Reflection_(mathematics)#Reflection_through_a_hyperplane_in_n_dimensions

momchil-flex:
Relevant code for [other supported transformations](https://github.com/flexcompute/tidy3d/blob/b848f5da93b82bb765fbb86e165db52c89650042/tidy3d/components/geometry/base.py#L2802-L2858), plus anywhere those are used, an addition for the reflection case might be needed.

tylerflex:
can we close this @momchil-flex ? assuming so and closing for now.

PR #2414, opened 2025-04-25
:sparkles: FEAT: Add reflection transformation with verification tests

damianofranzo:
This PR addresses feature request #2157 by adding the capability to apply reflection transformations to geometry objects.

### Key changes include:

- Implementation of reflection logic via `AbstractReflection` and `ReflectionFromPlane` classes, calculating the reflection matrix from a normal vector with input validation.
- Addition of the method `Transformed.reflection(normal)` to provide the 4x4 reflection matrix.
- Addition of the method `.reflected(normal)` to the class `Geometry` to return the reflected copy of the geometry.
- Addition of `test_general_reflection` to verify the reflection transformation is invariant to the scale and sign of the normal vector.
- Addition of other tests to verify the correct reflection of several geometries.

Includes necessary imports and type definitions.

### Considerations

- The rotation matrix is simply derived from the [definition from Wikipedia](https://en.wikipedia.org/wiki/Reflection_(mathematics)#Reflection_through_a_hyperplane_in_n_dimensions): $R = I - 2 \mathbf{\hat{n}} \mathbf{\hat{n}}^T $, where $\mathbf{\hat{n}}$ is the normal vector of the reflection plane. The input normal vector is automatically normalized internally. An error is raised only if the input normal's magnitude is close to zero.
- The class `AbstractReflection` is basically a copy of `AbstractRotation`, with the difference that a reflection can never be an identity, and as a consequence, `isidentity` is omitted.
- The class `ReflectionFromPlane` only accepts a normal in the format `(x, y, z)`. Initially, I thought the input could also have been `0`, `1`, or `2` to specify YZ, XZ, and XY planes, respectively, but then I realized that this could have caused some confusion when used as API.
- This PR focuses solely on adding reflection and is conceived as atomic. Therefore, it avoided any kind of refactoring (let's say, merging `AbstractReflection` and `AbstractRotation` into a more general abstraction, adding a possible center of reflection/rotation, or operations of this type).
- The testing covers sufficient cases, but let me know if you want me to extend it to additional scenarios.


momchil-flex (commented):
Thanks! This is looking good!

Just one thing I'm wondering: have you checked the codebase for places where transformations are used and where the reflection might also need to be incorportaed?

damianofranzo:
@momchil-flex Thank you!

The geometrical transformations are mainly checked in `test_geometry.py`, so I incorporated the changes into that file. In particular, I added `reflected` or `reflection` to 
- `test_planar_transform` to ensure that the transformed geometry does not project outside its defining plane
- `test_transforms` to ensure that the reflection works under some controlled scenarios.
- `test_general_reflection` to check fundamental properties of the reflection matrix

Regarding other uses of geometrical transformations:
- `test_structure.py` and `test_autograd.py`. I noticed `Polyslab` has a custom `rotated` method, and I suppose I could benefit from having its own `reflected` method. I will have a look at it.
- `test_medium.py`. I saw some usage of `rotationAroundAxis` in it, but it seems that it was mainly used for dealing with anisotropic properties, and I did not feel the need to add reflection there.
- `mode_solver.py`. I found the part you referred to where you do translation, rotation, and inverse translation. Since this is part of a plugin, I suppose further discussion is needed before adding new APIs.

 Let me know what you think.

damianofranzo:
I implemented the specialized `reflected` method for `PolySlab` with additional tests covering its usage.

tests/test_components/test_structure.py
```diff
@@ -138,6 +138,31 @@ def test_invalid_polyslab(axis):
...
+    geo10 = ps.reflected(n1)
+    _ = td.Structure(geometry=geo10, medium=medium)
+
+    geo11 = ps.reflected(n2)
+    with pytest.raises(pd.ValidationError):
+        _ = td.Structure(geometry=geo11, medium=medium)
+
+    geo12 = td.GeometryGroup(geometries=[ps]).reflected(n2)
+    with pytest.raises(pd.ValidationError):
+        _ = td.Structure(geometry=geo12, medium=medium)
```
momchil-flex, line 161: Can you remind me what happens here? It seems like it is automatically not allowed to reflect w.r.t. a non-zero component of the slab axis. This is good, because in general we don't want to have slanted polyslabs for efficiency (if something like that is needed, the user can use our general STL mesh geometry). I'm surprised that this worked right off the bat though rather than requiring an extra validator. What is the error exactly that gets raised here / what's the validator that's triggered?

momchil-flex (commented):
Nice, indeed, from the things you identified, materializing the transformation immediately for the PolySlab for efficiency seems the only thing that needed to be done, thanks for getting it in already!

tests/test_components/test_structure.py
damianofranzo: The error I get is a `ValidationError` with an error message `Slanted PolySlabs are not allowed to be rotated out of the slab plane.`.

This error occurs because reflecting a `PolySlab` off a plane whose normal is not perpendicular to the slab's axis results in a `Transformed` object.

The validation check (`validate_no_transformed_polyslabs` in`components/geometry/utils.py` ) then calls `base.Transformed.preserves_axis(transform, geometry.axis)`. This returns `false` for such transformations, triggering the validation error, as this state is considered an invalid rotation out of the slab plane for `PolySlabs`.

momchil-flex: Ok, makes sense, the validator is general enough already.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- For efficiency, `reflected` `PolySlab`-s now return updated `PolySlab` objects rather than a `Transformed` object, except when the normal of the plane of reflection has a non-zero component along the slab axis, in which case `Transformed` is still returned.
+- The method `Geometry.reflected` can be used to create a reflected copy of any geometry off a plane.
```
momchil-flex: This is very nitpicky but I would just merge these two items into one because no "change" was made to the `reflected` operation as it didn't exist before. Maybe something like

"The method `Geometry.reflected` can be used to create a reflected copy of any geometry off a plane. As for other transformations, for efficiency, `reflected` `PolySlab`-s directly return an updated `PolySlab` objects rather than a `Transformed` object, except when the normal of the plane of reflection has a non-zero component along the slab axis, in which case `Transformed` is still returned."

momchil-flex (approved):
Thanks @damianofranzo this basically looks good for me to integrate into our codebase, awesome job.

CHANGELOG.md
damianofranzo: Hi, Momchil. I updated this line with your indication so you can eventually merge these commits into your codebase. For the second sentence, the only slight difference is that I used the singular for consistency. Let me know if that works for you.

I am looking forward to working with you.

PR #2416, opened 2025-04-28
Add PermittivityMonitor to EMESimulation

caseyflex:
A small extension of the earlier PR adding PermittivityMonitor to ModeSimulation. It is also convenient to allow it directly in EMESimulation.


caseyflex:
@tylerflex  fyi, this will allow PermittivityMonitor in EMESimulation, but requires a new solver version

yaugenst-flex (approved):

PR #2417, opened 2025-04-28
:books: Fix and upgrade documentation of `latest`

daquinteroflex:
So this needs a merge commit right.


momchil-flex (approved):

daquinteroflex:
Actually I think I need to update the notebooks submodule in develop back to what it was

PR #2425, opened 2025-05-02
Non-isothermal charge simulations

marc-flex:
## Greptile Summary

Added support for non-isothermal charge simulations by introducing `SteadyChargeDCAnalysis` base class and refactoring the existing `IsothermalSteadyChargeDCAnalysis`, enabling temperature variation effects in charge simulations.

- Split analysis classes in `spice/analysis/dc.py` to support both temperature-varying and fixed-temperature simulations while maintaining backward compatibility
- Added validation in `tcad/simulation/heat_charge.py` to ensure structures have both electrical and thermal properties for non-isothermal simulations
- Enhanced test fixtures in `test_heat_charge.py` with comprehensive thermal conductivity and medium components
- Modified package exports in `__init__.py` to include new `SteadyChargeDCAnalysis` class


yaugenst-flex:
@greptileai

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -951,6 +955,44 @@ def check_transient_heat(cls, values):
...
 
+    @pd.root_validator(skip_on_failure=True)
+    def check_non_isothermal_is_possible(cls, values):
+        """Make sure that when a non-isothermal case is defined the structrures
+        have both electrical and thermal properties."""
+
+        analysis_spec = values.get("analysis_spec")
+        if isinstance(analysis_spec, SteadyChargeDCAnalysis) and not isinstance(
+            analysis_spec, IsothermalSteadyChargeDCAnalysis
+        ):
```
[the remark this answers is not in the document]
yaugenst-flex: this is suggesting a change that is literally the same thing as the code that it's commenting. not useful.

tidy3d/components/tcad/effective_DOS.py
```diff
@@ -0,0 +1,153 @@
...
+
+        return self.calc_eff_dos(T)
+
+    def get_effective_DOS_derivative(self, T: float):
+        if T <= 0:
+            raise DataError(
+                f"Incorrect temperature value ({T}) for the effectve density of states calculation."
+            )
+
+        return self.calc_eff_dos_derivative(T)
```
marc-flex: I can see why one would like to inspect the density of states at different temperatures but I can't see why would like to see it's temperature derivative. 

I'm thinking of removing this. @damiano-flex @momchil-flex do you agree with this?

damiano-flex: Yes you can remove it.

tests/test_components/test_heat_charge.py
```diff
@@ -144,9 +146,9 @@ def mediums():
             ),
         ),
         charge=td.SemiconductorMedium(
-            N_c=1e10,
-            N_v=1e10,
-            E_g=1,
+            N_c=td.ConstantEffectiveDOS(N=1e10),
+            N_v=td.ConstantEffectiveDOS(N=1e10),
+            E_g=td.ConstantEnergyBandGap(eg=1),
```
momchil-flex, line 151: How about adding some of the other DOS types, and the Varshni E_G?

momchil-flex (approved):
Had another quick look over this, exciting to get it in.

tests/test_components/test_heat_charge.py
marc-flex: You're right. I'll take care of it :+1:

PR #2428, opened 2025-05-06
Add Tidy3D Matplotlib Styles on Import

### LINKED ISSUE #1713: :art: Custom `tidy3d.rcparams` that activates automatically for  nicer custom standard plots
Just an idea that popped in the other day, if we wanted to make all our plots look the same and nicer  than some default matplotlib options (say to format all our notebook example plots, and data-generated ones). It could be an interesting aesthetic improvement.

frederikschubertflex:
This PR addresses https://github.com/flexcompute/tidy3d/issues/1713 and sets the global Matplotlib styles via the `rcParams` to a basic set of styles that follow that Flexcompute design brand.

before_style_update
after_style_update

The `reset_previous_style` function is not yet exported as part of the public API.

Note: The PR also contains a refactoring commit that converts the Flexcompute color map to simple hex strings and refactors the single place that used the hex numbers (tidy3d/material_library/util.py) to simplify the code.

Closes https://github.com/flexcompute/tidy3d/issues/1713


frederikschubertflex:
This PR might need more testing over a broader set of notebooks to see if the new styles break anything. The example is taken from https://github.com/flexcompute/tidy3d-notebooks/blob/develop/Simulation.ipynb. Additionally, the style presets are quite minimal for now and can (should) be extended to improve the display of the plots.

daquinteroflex:
Overall looking great @frederikschubertflex thanks for this!

@tomflexcompute do you have any other preferences for an improved standard tidy3d style? Like thicker axes, etc. We can add this in here.

daquinteroflex:
So the reset styles functionality needs to be added to the changelog.md and also to somewhere in the sphinx documentation too

tests/test_components/test_viz.py
```diff
@@ -317,3 +319,8 @@ def test_sim_plot_structures_fill():
     for patch in structure_patches[:1]:
         assert patch.get_fill(), "Should be filled when True"
         assert patch.get_facecolor() != "none", "Face color should be set"
+
+def test_tidy3d_matplotlib_style_application_on_import():
+    assert mpl.rcParams.get("axes.edgecolor") == "#ECEBEA" 
+    reset_previous_style()
+    assert mpl.rcParams.get("axes.edgecolor") == mpl.rcParamsDefault.get("axes.edgecolor")
```
daquinteroflex: Nice!

tidy3d/material_library/util.py
```diff
@@ -147,28 +142,24 @@ def summarize_material_library_rich(matlib):
...
 
     table = Table(
-        title=f"[bold {get_hex_formatted_string('brand_black')}]Material Library Summary[/]",
+        title=f"[bold {FLEXCOMPUTE_COLORS['brand_black']}]Material Library Summary[/]",
         show_header=True,
-        header_style=f"bold {get_hex_formatted_string('brand_black')}",
-        border_style=f"{get_hex_formatted_string('brand_black')}",
+        header_style=f"bold {FLEXCOMPUTE_COLORS['brand_black']}",
+        border_style=f"{FLEXCOMPUTE_COLORS['brand_black']}",
         expand=False,  # Don't force table to full terminal width
```
daquinteroflex, line 149: Do we have a test for this? I guess not necessary just wondering how to avoid breaking it accidentally though.

frederikschubertflex: ~~I will create one that asserts that the color is correct when printing the material library.~~

The best way would be to add tests to the material library that test their rich representation `__rich__()`. Rich itself validates the styles on creation of the `rich.Column`.

frederikschubertflex: Added the test.

frederikschubertflex:
> So the reset styles functionality needs to be added to the changelog.md and also to somewhere in the sphinx documentation too

I added an entry in the CHANGELOG.md but we need to discuss where to document it in the sphinx docs.

tidy3d/components/viz.py
```diff
@@ -224,6 +225,52 @@ def validate_and_copy_color(value: str, values: Dict[str, Any]) -> str:
...
+# Tidy3D default plotting style parameters
+_TIDY3D_STYLE_PARAMS = {
+    "axes.prop_cycle": cycler(
+        color=[
+            FLEXCOMPUTE_COLORS["brand_green"],
+            FLEXCOMPUTE_COLORS["brand_tan"],
+            FLEXCOMPUTE_COLORS["brand_blue"],
+            FLEXCOMPUTE_COLORS["brand_purple"],
+        ]
+    ),
```
yaugenst-flex: Hi @frederikschubertflex, just passing by, great to see this PR!

I have a more general question that we might want to discuss @daquinteroflex @tomflexcompute @e-g-melo  - the prop cycler here has only four colors, compared to the default matplotlib one which has 10. So if a plot contains >4 lines, then the colors start repeating. That seems problematic. Do you know whether this would happen often, or if plots with that many lines are rare? We might need to consider broadening the palette a bit.

Another thing to consider is how do these colors actually look in a plot with multiple lines? Are they easily distinguishable? We should make sure that these colors produce good-looking plots as it will affect all of our users.

tomflexcompute: That's a great point @yaugenst-flex . I think 4 colors are definitely a bit limited. Even in our examples there are multiple cases where we plot more than 4 curves.

frederikschubertflex: These are good points! I worked on a set of palettes and came up with these:
color_maps

I will look into their effect on the different notebooks and plots. What do you think?

yaugenst-flex: That looks awesome, this will be super cool to have! Only two comments:
1. I assume we will use the categorical one for the cycler, so we'd probably want to order them such that the "main" brand colors appear first (most plots are only a single line after all). The light gray might also be a bit too light for line colors. And the green appears pretty bright? Might just be me.
2. The diverging ones have this pretty jarring zero-crossing at the moment. can we smoothen the transition there? There are some nice examples here: https://www.fabiocrameri.ch/colourmaps/

Ultimately however it will come down to how the plots using these look, so we should make some..

daquinteroflex: Yeah ideally the main cycler colors work nicely with the main colors in the new docs https://docs.flexcompute.com/projects/tidy3d/en/develop/ like deep green, the gold, the magenta, the yellow, the blue and then the light I'd suggest or something like that - whatever seems more reasonable.

frederikschubertflex: Thank you @yaugenst-flex  for the hint to Crameri! I looked into his construction of the palettes and saw that he uses a closed spline in the `CAM02-UCS` color space to interpolate for the cyclic palettes. Because we are kind of commited to one color palette, I tried to blend his approach and the existing color tints. This is my current proposal:

output

yaugenst-flex: Magnificent! 🤩

tidy3d/components/viz/flex_color_paletes.py
yaugenst-flex, line 1: `paletes` -> `palettes`

tidy3d/components/viz/flex_style.py
```diff
@@ -0,0 +1,51 @@
...
+                    _ORIGINAL_PARAMS[key] = plt.rcParams[key]
+            plt.rcParams.update(_TIDY3D_STYLE_PARAMS)
+        except Exception as e:
+            log.error(f"Failed to apply Tidy3D plotting style on import. Error: {e}")
+            _ORIGINAL_PARAMS = {}
+    except ImportError:
+        pass
+
+
+def reset_previous_style():
```
yaugenst-flex: Not sure about the naming of this. I think as a user it wouldn't be clear to me what this function does just from its name. Maybe something like `reset_matplotlib_style()`, `restore_rcparams()` or something along those lines?

tidy3d/components/viz/flex_style.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    Resets matplotlib rcParams to the values they had before the Tidy3D
+    style was automatically applied on import.
+    """
+    if not _ORIGINAL_PARAMS:
+        log.warning("No previous Matplotlib style state found to reset to.")
+        return
```
yaugenst-flex: Good behavior here might be to default to:
```py
matplotlib.style.use("default")
```

yaugenst-flex (commented):
Thanks @frederikschubertflex overall looks great, just a couple of minor comments.

One general comment/question I had is whether it would make sense to create [a custom matplotlib style sheet](https://matplotlib.org/stable/users/explain/customizing.html#defining-your-own-style) that we could just load on import? Not sure if that's better or not.

tidy3d/components/viz/flex_style.py
frederikschubertflex: Yes, maybe `restore_matplotlib_rcparams`?

frederikschubertflex:
> One general comment/question I had is whether it would make sense to create [a custom matplotlib style sheet](https://matplotlib.org/stable/users/explain/customizing.html#defining-your-own-style) that we could just load on import? Not sure if that's better or not.

Good point! I was also thinking about this but did not want to add the complexity of distributing a separate style file. This was the approach with the least amount of friction, but I am happy to change it if you feel more comfortable with the external style sheet approach.

yaugenst-flex:
> Good point! I was also thinking about this but did not want to add the complexity of distributing a separate style file. This was the approach with the least amount of friction, but I am happy to change it if you feel more comfortable with the external style sheet approach.

Is there something that needs to be done beyond just having it live in the `viz/` folder? But yeah I agree if it significantly changes in how we have to package then it's not worth it (for the moment at least).

And actually one more question: Do we have some sample plots (or notebook runs) that show the new styles?

tidy3d/components/viz/flex_color_palletes.py
yaugenst-flex, line 1: `palletes` -> `palettes` 😄

frederikschubertflex: My eyes... :D Thanks for catching this!

frederikschubertflex:
> Is there something that needs to be done beyond just having it live in the viz/ folder? But yeah I agree if it significantly changes in how we have to package then it's not worth it (for the moment at least).

No, you are correct! We can use it as `mpl.style.use("tidy3d.viz.style")` if we put a file at `tidy3d/viz/style.mplstyle`. So its not a large change. What do you think?

frederikschubertflex:
I will add a few examples. Note, that many plots such as field data and simulation structures are not affected due to their manual color palettes.
# VizData
## Old
VizData_old
## New
VizData_new
# FullyAnisotropic
## Old
FullyAnisotropic_old
## New
FullyAnisotropic_new

yaugenst-flex:
> No, you are correct! We can use it as mpl.style.use("tidy3d.viz.style") if we put a file at tidy3d/viz/style.mplstyle. So its not a large change. What do you think?

I feel like it might be worth it then. It'll be easier to reuse down the line, since the style might make it into other packages too or perhaps we can even expose it to users if they want to use it.

frederikschubertflex:
I changed the setup to use the `mplstyle` file. I had to add an entry to the `pyproject.toml` file but I think it should be fine.

pyproject.toml
```diff
@@ -245,7 +246,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.ruff]
 target-version = "py39"
-line-length = 100
+line-length = 101
```
yaugenst-flex: 👀

frederikschubertflex: I do not know how that got in there. Sorry, will revert. I had some strange issues with the vscode vim extension today which led to weird formatting.

tidy3d/style.mplstyle
```diff
@@ -0,0 +1,4 @@
+axes.prop_cycle : cycler(color=["#176737", "#FF7B0D", "#979BAA", "#F44E6A", "#0062FF", "#26AB5B", "#6D3EF2", "#F59E0B"])
```
daquinteroflex, line 1: Sorry maybe I missed this is there a reason why these cycler colors are not the same as FLEXCOMPUTE_COLORS?

daquinteroflex (approved):
Thanks for this very comprehensive upgrade! Looks good to me apart from the question I asked on the cycler, to be sure those are the colors we want?

daquinteroflex:
We'll of course need to do the notebooks PR after this is merged to develop maybe before the next release.

tidy3d/style.mplstyle
frederikschubertflex: We needed more colors for plots with more than 6 series. That's why I went down this road of creating our own palettes.

daquinteroflex:
Happy to merge whenever imo! Maybe we should run notebooks after #2433  is merged though to catch any issues with that

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed issue with `CustomMedium` gradients where other frequencies would wrongly contribute to the gradient.
 - Fixed bug when computing `PolySlab` bounds in plotting functions.
 - Fixed bug in broadband adjoint source creation when forward simulation had a pulse amplitude greater than 1 or a nonzero pulse phase.
+- Automatically apply `matplotlib` styles when importing `tidy3d` which can be reverted via the `td.restore_matplotlib_rcparams()` function
```
yaugenst-flex: Changelog item needs to move up to "Unreleased"

frederikschubertflex: Moved it!

yaugenst-flex (approved):
LGTM!

PR #2429, opened 2025-05-06
Fix bug when symmetry + non-symmetric boundary conditions are defined

momchil-flex:
This fixes a bug a user encountered which was brought to my attention by @FilipeFcp. Basically, when a simulation is created with symmetry applied along a given axis, but with boundary conditions that differ on the left and right side, physically we just ignore the left side. However, in the case of PML/Absorber boundaries with a different number of layers, the simulation grid that would be created would still use the left-side number of layers, and create a non-symmetric grid on which the "expanded fields" are defined.

The initial fix I tried (the code change here) just fixes the grid generation to use whatever number of layers is on the right. This fixes the issue that was observed in the recorded field data, however I realize that the current state can still be kind of confusing. For example, if I plot such a simulation, I get this:

image

Obviously something is going wrong here too as the overlaid "symmetry region" extends into the large absorber region on the right. Ideally, if we are neglecting the boundary definition on the left, we should actually incorporate this everywhere, and e.g. display the simulation with the right-side boundary on the left, too.

I'm afraid doing this automatically may be a bit frustrating though. For example, one option is to create a `_boundary_spec` property that mirrors `boundary_spec`, but replaces any minus-side boundaries with the plus-side ones, if a symmetry is applied - then use this everywhere internally. The issue though is that what is written to the json will not correspond to what is e.g. displayed, which can again be confusing. Maybe better is to take the route of introducing a validator that would modify the `boundary_spec` upon creation - but might that be confusing to the user too? Meaning that their input will be modified - but in some sense, it will be "made right".

The last alternative is to just add a validator to require that the user makes the boundary specs symmetric, if symmetry is applied. More strict from a user perspective, but at least fool- and confusion-proof.

After writing all this out, maybe the best is to add the dynamic validator that modifies `boundary_spec` if needed. What do you think @yaugenst ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - The method `Geometry.reflected` can be used to create a reflected copy of any geometry off a plane. As for other transformations, for efficiency, `reflected` `PolySlab` directly returns an updated `PolySlab` object rather than a `Transformed` object, except when the normal of the plane of reflection has a non-zero component along the slab axis, in which case `Transformed` is still returned.
 - Validation check for unit error in grid spacing.
+- Validation that when symmetry is imposed along a given axis, the boundary conditions on each side of the axis are identical.
 
 ### Changed
 - Supplying autograd-traced values to geometric fields (`center`, `size`) of simulations, monitors, and sources now logs a warning and falls back to the static value instead of erroring.
--- a/tests/test_components/test_simulation.py
+++ b/tests/test_components/test_simulation.py
@@ -585,6 +585,33 @@ def test_validate_zero_dim_boundaries():
     )
 
 
+def test_validate_symmetry_boundaries():
+    # simulation with symmetry along an axis should have the same boundaries defined on both sides
+    td.Simulation(
+        size=(1, 1, 1),
+        symmetry=(0, 1, 0),
+        grid_spec=td.GridSpec.uniform(dl=0.1),
+        run_time=1e-12,
+        boundary_spec=td.BoundarySpec(
+            x=td.Boundary.periodic(),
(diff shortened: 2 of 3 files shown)


FilipeFcp:
Hi @momchil-flex.
In my opinion, using a validator with a clear error message would be a good option, since it's difficult to know whether the user truly intended to define a PML or an absorber.

In this specific case, I believe the user intended to apply absorbers on both sides but simply assigned it to one side without realizing it.

FilipeFcp:
Hi @momchil-flex,
In my opinion, users shouldn't be allowed to create this kind of configuration.

In this specific case, I believe the user intended to apply absorbers on both sides but simply assigned it to one side without realizing it. So maybe just don't allowing it, with a clear error message would be a good option.

momchil-flex:
I guess that makes sense. I will work to update it in that direction unless @yaugenst has a different commetn.

yaugenst-flex:
Yeah I agree with @FilipeFcp. I think allowing this can have a lot of unintended consequences. Physically the simulation is not symmetric in these cases so I'm not sure if we want to introduce a bunch of additional logic just to "fix" it after the fact (and we can only ever guess what the user intended).

momchil-flex:
Ok, I've updated it as discussed. I guess for completeness let me put here the reasoning why I was considering "fixing" the setup (but, generally, I agree that just validating it out as done here is likely best).

- We already "disregard" any e.g. structures that are in the left symmetry space. When a user does a plot, those geometries would be shown. But if they are not symmetrically defined, it would not matter (the plot will be wrong in that sense) as physically the geometries in the right space are symmetrically applied. Or rather, we only compute fields in the right space, so nothing in the left space matters.
- Some other tools consider the symmetry to be a boundary condition in itself, so you cannot simultaneously define a symmetry and a boundary condition on the left.

yaugenst-flex (approved):

PR #2431, opened 2025-05-07
Delegate PEC check from MultiPhysicsMedium to its Optical Medium

frederikschubertflex:
The check is used for plotting the permittivity in the https://github.com/flexcompute/tidy3d/blame/9c8ca5ff9310a681352ae0c2624c9a142275cd6f/tidy3d/components/scene.py#L1040C9-L1040C20 `_eps_bounds` method. I am not sure whether this should depend solely on the optical medium or if the other media are relevant as well.

Fixes https://github.com/flexcompute/tidy3d-core/issues/956

### DIFF
--- a/tidy3d/components/material/multi_physics.py
+++ b/tidy3d/components/material/multi_physics.py
@@ -103,7 +103,7 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
         if self.optical is not None:
             return self.optical.is_pec
         else:
-            return True
+            return False
 
     @property
     def heat_spec(self):
--- a/tidy3d/components/material/multi_physics.py
+++ b/tidy3d/components/material/multi_physics.py
@@ -100,7 +100,15 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
 
     def __getattr__(self, name: str):
         """Delegates calls to a fixed set of missing attributes to their respective inner medium."""
-        DELEGATED_ATTRIBUTES = {"is_pec": self.optical, "_eps_plot": self.optical}
+        IGNORED_ATTRIBUTES = ["__deepcopy__"]
+        if name in IGNORED_ATTRIBUTES:
+            return None
+
+        DELEGATED_ATTRIBUTES = {
+            "is_pec": self.optical,
+            "_eps_plot": self.optical,
+            "viz_spec": self.optical,
+        }
         # NOTE(frederikschubertflex): this dictionary might need to be extended as we transition our code towards using this `MultiPhysicsMedium`.
 
         if name in DELEGATED_ATTRIBUTES:
--- a/tidy3d/components/material/multi_physics.py
+++ b/tidy3d/components/material/multi_physics.py
@@ -107,7 +107,7 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
             return getattr(DELEGATED_ATTRIBUTES[name], name)
         else:
             raise ValueError(
-                f"MultiPhysicsMedium has no attribute called {name}. Did you meant to access the attribute of one of the optical, heat or charge media?"
+                f"MultiPhysicsMedium has no attribute called {name}. Did you mean to access the attribute of one of the optical, heat or charge media?"
             )
 
     @property
--- a/tidy3d/components/material/multi_physics.py
+++ b/tidy3d/components/material/multi_physics.py
@@ -99,13 +99,14 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
     )
 
     def __getattr__(self, name: str):
-        """Delegates calls to missing attributes to the inner `optical` medium."""
-        if self.optical is not None:
+        """Delegates calls to missing attributes to the inner media."""
+        if self.optical is not None and hasattr(self.optical, name):
             return getattr(self.optical, name)
-        else:
-            raise ValueError(
-                f"Tried to access {name} of MultiPhysicsMedium without optical medium definition."
-            )
+        if self.heat is not None and hasattr(self.heat, name):
+            return getattr(self.heat, name)
+        if self.charge is not None and hasattr(self.charge, name):
+            return getattr(self.charge, name)
+        return None
 
     @property
     def heat_spec(self):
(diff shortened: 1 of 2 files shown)


tidy3d/components/material/multi_physics.py
```diff
@@ -98,6 +98,13 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
         None, title="Charge properties", description="Specifies properties for Charge simulations."
     )
 
+    @property
+    def is_pec(self):
+        if self.optical is not None:
+            return self.optical.is_pec
+        else:
+            return True
```
momchil-flex: Why would this return True if the `optical` medium is not defined?

frederikschubertflex: That's why I marked this as draft. I am not yet very deep into the multi physics part of tidy3d. Looking at the other Medium's default, this probably should be False.
Or would you prefer None as it is indeterminate?

momchil-flex: No problem! I just got the ping that you assigned me as reviewer and took a look. :)

Probably False makes sense. We can explicitly set it as True when needed.

frederikschubertflex: Thank you! 😊 ~~Do you think it makes sense to also check the other media, i.e. if any of them returns true for the `is_pec` property?~~

Defining the property differently does not make sense so I leave it at that.

tidy3d/components/material/multi_physics.py
```diff
@@ -98,6 +98,16 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
         None, title="Charge properties", description="Specifies properties for Charge simulations."
     )
 
+    def __getattr__(self, name: str):
+        """Delegates calls to missing attributes to the inner media."""
+        if self.optical is not None and hasattr(self.optical, name):
```
frederikschubertflex: I just noticed that this precedence order of `optical`, `heat` and `charge` might be sub-optimal and we should check whether there is a conflict between the media. For example the optical permittivity might be different than the one of the charge medium. Should we raise an error in this case?

daquinteroflex: Yeah so as we just chatted, ultimately this PR will discover the issues that were inherited or coupled between solvers. So this generic delegation might cause errors or different behaviour users might not expect. So as we chatted maybe to have it per indivdual fucntions to reduce the scope.

momchil-flex: For my understanding what is you guys' thinking now? Also should we get this in today for the testing of the next patch or better leave it for 2.9?

frederikschubertflex: So the `MultiPhysicsMedium` is used in places where the rest of the code expects a subclass of the `AbstractMedium`. Thus, its missing some functions and properties that are expected, such as the `is_pec` property or `_eps_plot` function.
The current implementation of the fix delegates all calls to missing properties/functions to one of the respective inner media. This could backfire if two properties/functions are called the same but have different semantics. I feared that this might be the case, so I wanted to build in a check. But this still is not fail-safe as per discussion with @daquinteroflex .
Thus, I will change this implementation to be restricted to a fixed  set of properties that are delegated to their respective medium. I will add a note that this set might need to be extended in the future.

momchil-flex: Got it, indeed that makes sense to me!

tidy3d/components/material/multi_physics.py
```diff
@@ -98,6 +98,26 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
         None, title="Charge properties", description="Specifies properties for Charge simulations."
     )
 
+    def __getattr__(self, name: str):
+        """Delegates calls to a fixed set of missing attributes to their respective inner medium."""
+        IGNORED_ATTRIBUTES = ["__deepcopy__"]
+        if name in IGNORED_ATTRIBUTES:
+            return None
```
momchil-flex, line 133: What is the logic of ignored attributes? What goes there?

tidy3d/components/material/multi_physics.py
```diff
@@ -98,6 +98,26 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
...
+            "viz_spec": self.optical,
+        }
+        # NOTE(frederikschubertflex): this dictionary might need to be extended as we transition our code towards using this `MultiPhysicsMedium`.
+
+        if name in DELEGATED_ATTRIBUTES:
+            return getattr(DELEGATED_ATTRIBUTES[name], name)
+        else:
+            raise ValueError(
+                f"MultiPhysicsMedium has no attribute called {name}. Did you mean to access the attribute of one of the optical, heat or charge media?"
+            )
```
momchil-flex, line 146: This seems like it would error for all attributes that are not in IGNORED_ or DELEGATED_? So anything that's actually defined in MultiPhysicsMedium would error?

tidy3d/components/material/multi_physics.py
frederikschubertflex: There are a few attributes that we do not want to delegate such as `__deepcopy__` which `deepcopy` checks before applying its default deep copy function.

tidy3d/components/material/multi_physics.py
frederikschubertflex: No, the `__getattr__` is only called for missing attributes on the class where `__getattribute__` returns nothing. So the normal access of e.g. `medium.name` still works.

I added a line to the unit test that verifies this.

tidy3d/components/material/multi_physics.py
frederikschubertflex: But you are right that the variable name is not as clear as it could be. I will add a comment explaining the intention.

tidy3d/components/material/multi_physics.py
daquinteroflex: I see, I like how generic this is actually and nicely extensible for other missing properties or methods

daquinteroflex (approved):
Nice thanks Frederik!

momchil-flex (approved):
Great, thanks!

PR #2433, opened 2025-05-08
FXC-3378 Pydantic v2

### LINKED ISSUE #1239: Full Pydantic v2 upgrade ⏫
We ported to pydantic 2.* by importing `pydantic.v1`, but it would be good to do a full port.

There are many internals that utilize v1 features, so it's an opportunity to dive pretty deep into the front end code, convert them to v2 syntax, and potentially simplify them if they are not needed anymore or there are simpler ways to handle them.

I took a stab at it [here](https://github.com/flexcompute/tidy3d/pull/1030) but couldn't get it 100% working and there are some loose ends I'm not 100% sure about. It's also possible these changes messed with our docs build, I didn't check that yet.

@daquinteroflex depending on how much time you have, maybe after the docs merging and poetry, we can try wrapping this up and that will help us future-proof tidy3d.

yaugenst-flex:
## Summary
Migrates the tidy3d python client from **Pydantic v1 → v2** across models, validators, serialization, and schema generation.

This PR also includes the CI/tooling updates needed to keep the v2 codebase clean (type-only import handling and mypy checks).

### Pydantic v2 migration
- Standardize imports on `pydantic` / `pydantic_settings`; remove v1 compatibility imports.
- Replace `@validator` / `@root_validator` with `@field_validator` / `@model_validator` (incl. `mode='before'|'after'|'wrap'` where appropriate), updating signatures (`ValidationInfo`) and using `object.__setattr__` when mutating `frozen` models.
- Replace `.dict()/.json()/.parse_obj()` with `model_dump*` / `model_validate`, and `.schema()` with `model_json_schema()`.
- Migrate model configuration to `model_config = ConfigDict(...)`.
- Migrate `BaseSettings` usage to `pydantic_settings.BaseSettings`.
- Update custom types to the v2 hooks:
  - `__get_pydantic_core_schema__` / `__get_pydantic_json_schema__`
  - refactors toward `Annotated[...]` validators/serializers (`BeforeValidator`, `PlainSerializer`, etc.)
- Regenerate and commit updated schema artifacts under `schemas/`.

### Pydantic-v2-specific fixes + regressions locked by tests
- Fix `ValueMap` validation bug caused by `Mapping.keys/values` name collisions; add regression tests.
- Add regression test for lazy proxy materialization that touches `__pydantic_*` internals (guards future Pydantic bumps).
- Preserve YAML serialization behavior for `Infinity` / `-Infinity` / `NaN` (string-based, JSON-aligned); add regression test and document why `to_yaml()` intentionally uses a JSON round-trip.
- Guard `info.data[...]` cross-field access in field validators to avoid `KeyError` when prerequisite fields fail validation; add regression tests.

### Tooling / CI
- Add/update pre-commit + CI steps around type-only import organization and mypy type-signature checking (needed to keep the v2 migration maintainable).

## Compatibility notes
- Goal: no intentional public API drift. The generated JSON schemas are treated as part of the public API and were regenerated/committed as part of the migration.
- tidy3d continues to raise domain-specific exceptions (`tidy3d.exceptions.ValidationError`, `SetupError`, etc.) for many validation failures; Pydantic core parse errors may differ slightly from v1 for invalid inputs.

## How to test
- `poetry run pre-commit run --all-files`
- `poetry run pytest -q`

## Follow-ups (post-merge)
- Refactor validator sprawl in `Simulation`/`Medium` into orchestrated helpers while preserving warning/error ordering.
- Audit load-only aliases and consider `validation_alias` / `serialization_alias` where appropriate (may require schema review).
- Evaluate removing JSON encode/decode round-trips only where behavior stays identical.

---

> [!NOTE]
> **Medium Risk**
> Primarily CI/tooling and dependency constraint changes; moderate risk due to expanding mypy coverage to all of `tidy3d` and introducing a new import-rewrite gate that could fail PRs or require code changes.
> 
> **Overview**
> Adds a new CI and pre-commit verification (`scripts/move_type_imports.py`) that enforces type-only imports are moved under `if TYPE_CHECKING`, and wires this into the workflow’s required checks.
> 
> Updates verification tooling to standardize on Python `3.11` via a shared `VERIFICATIONS_PY_VERSION`, adjusts `pytest` config formatting, and broadens `mypy` to run across the full `tidy3d` package.
> 
> Tightens the Pydantic v2 dependency spec (`pydantic >=2.9,<3`) and adds `pydantic-settings`, plus includes `libcst` in dev extras to support the new import-guard tooling; also ignores `simulation.json` in git.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bdf3a7ab5a7ebe861e95c67c3637bd45280befc1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


momchil-flex:
It seems like fundamentally there's just a few things that change? Maybe I'm missing something important/tricky. Could you explain if there's anything like that, and what changes going forward?

Also very important - we will certainly need to update the backend python code too and might need to update denormalizer, metadata api, and who knows what else.

yaugenst-flex:
Yes the biggest changes are related to custom types and serialization, everything else is relatively straightforward. I'll write something up.

yaugenst-flex:
This is ready for initial review. `plugins/adjoint` still has failing tests but everything else is passing and rebased. i'll work on fixing adjoint and incrementally fixing warnings.

tests/test_components/test_IO.py
```diff
@@ -24,6 +24,7 @@
 
 # Store an example of every minor release simulation to test updater in the future
 SIM_DIR = "tests/sims"
+SIM_STATIC = SIM.to_static()
```
tylerflex, line 29: is there a reason we test with sim static? might be worth a comment for future reference

yaugenst-flex: right yeah it's because i fixed (i think?) the equality check (`Tidy3dBaseModel.__eq__`). so when you compare a model with and without tracers, then they will not compare equal. i think this behavior is correct?

basically if we run these tests with `SIM` instead of `SIM_STATIC`, then we get the following failures:
```
FAILED tests/test_components/test_IO.py::test_simulation_load_export_hdf5 - AssertionError: original and loaded simulations are not the same
FAILED tests/test_components/test_IO.py::test_simulation_load_export_hdf5_explicit - AssertionError: original and loaded simulations are not the same
FAILED tests/test_components/test_IO.py::test_simulation_load_export - AssertionError: original and loaded from hdf5 simulations are not the same
FAILED tests/test_components/test_IO.py::test_simulation_load_export_hdf5_gz - AssertionError: original and loaded simulations are not the same
FAILED tests/test_components/test_IO.py::test_simulation_load_export_hdf5_gz_explicit - AssertionError: original and loaded simulations are not the same
```

pyproject.toml
```diff
@@ -291,7 +291,11 @@ required-imports = ["from __future__ import annotations"]
 
 [tool.pytest.ini_options]
 # TODO: remove --assert=plain when https://github.com/scipy/scipy/issues/22236 is resolved
-addopts = "--cov=tidy3d --doctest-modules -n auto --dist worksteal --assert=plain -m 'not numerical' "
+# TODO: <yaugenst-flex> - un-ignore warnings & adjoint tests
+addopts = """
+--doctest-modules -n auto --dist worksteal --assert=plain -m 'not numerical' \
+-W ignore::UserWarning tests --ignore=tests/test_plugins/test_adjoint.py
```
tylerflex: not sure if previous comment got through , are we planning to ignore these warnings indefinitely?

frederikschubertflex:
I wrote a little script for some deprecated symbols but was not sure whether you have already removed them or whether they are relevant. If not, you can ignore this comment @yaugenst-flex, otherwise I think its faster if you fix them? 

### parse_obj_as – deprecated; prefer TypeAdapter
./tidy3d/web/core/task_core.py
54:            parse_obj_as(
127:            parse_obj_as(
313:        return parse_obj_as(list[SimulationTask], resp)

./tidy3d/web/api/material_libray.py
44:        return parse_obj_as(list[MaterialLibray], resp) if resp else None

./tidy3d/web/api/material_library.py
61:        return parse_obj_as(list[MaterialLibrary], resp) if resp else None

### json_encoders – replaced by serializers
./tidy3d/plugins/adjoint/components/base.py
45:        json_encoders={

### parse_raw – removed; use model_validate_json()
./tidy3d/plugins/dispersion/web.py
297:        best_medium = PoleResidue.parse_raw(run_result["message"])

yaugenst-flex:
thanks @frederikschubertflex that's great. the adjoint plugin one is left intentionally but i missed those `parse_` methods!

tidy3d/components/base.py
```diff
@@ -1051,119 +1147,110 @@ def to_static(self) -> Tidy3dBaseModel:
...
-            name=TYPE_TAG_STR,
-            value=value,
-            annotation=annotation,
-            class_validators=None,
-            config=cls.__config__,
-        )
-        cls.__fields__[TYPE_TAG_STR] = tag_field
-
     @classmethod
     def generate_docstring(cls) -> str:
```
daquinteroflex, line 1640: Thanks for migrating this! I've triggered the docs to see how it goes. 

Did a small before and after check on the terminal:

```
Rectangular prism.
       Also base class for :class:`Simulation`, :class:`Monitor`, and :class:`Source`.

    Parameters
    ----------
    attrs : dict = {}
        Dictionary storing arbitrary metadata for a Tidy3D object. This dictionary can be freely used by the user for storing data without affecting the operation of Tidy3D as it is not used internally. Note that, unlike regular Tidy3D fields, ``attrs`` are mutable. For example, the following is allowed for setting an ``attr`` ``obj.attrs['foo'] = bar``. Also note that `Tidy3D`` will raise a ``TypeError`` if ``attrs`` contain objects that can not be serialized. One can check if ``attrs`` are serializable by calling ``obj.json()``.
    center : Union[tuple[Union[float, autograd.tracer.Box], Union[float, autograd.tracer.Box], Union[float, autograd.tracer.Box]], Box] = (0.0, 0.0, 0.0)
        [units = um].  Center of object in x, y, and z.
    size : Union[tuple[Union[pydantic.v1.types.NonNegativeFloat, autograd.tracer.Box], Union[pydantic.v1.types.NonNegativeFloat, autograd.tracer.Box], Union[pydantic.v1.types.NonNegativeFloat, autograd.tracer.Box]], Box]
        [units = um].  Size in x, y, and z directions.

    Example
    -------
    >>> b = Box(center=(1,2,3), size=(2,2,2))
```

after
```
>>> print(td.Box(size=(1,1,1)).__doc__)
Rectangular prism.
       Also base class for :class:`Simulation`, :class:`Monitor`, and :class:`Source`.

    Parameters
    ----------
    attrs : dict = dict()
        Dictionary storing arbitrary metadata for a Tidy3D object. This dictionary can be freely used by the user for storing data without affecting the operation of Tidy3D as it is not used internally. Note that, unlike regular Tidy3D fields, ``attrs`` are mutable. For example, the following is allowed for setting an ``attr`` ``obj.attrs['foo'] = bar``. Also note that `Tidy3D`` will raise a ``TypeError`` if ``attrs`` contain objects that can not be serialized. One can check if ``attrs`` are serializable by calling ``obj.json()``.
    center : Optional[tuple[Union[float, Annotated[autograd.tracer.Box, BeforeValidator(func=<function traced_alias.<locals>._validate_box_or_container at 0x734e1f5099e0>, json_schema_input_type=PydanticUndefined), PlainSerializer(func=<function traced_alias.<locals>.<lambda> at 0x734e1f509b20>, return_type=PydanticUndefined, when_used='json')], Annotated[object, BeforeValidator(func=<function traced_alias.<locals>._validate_box_or_container at 0x734e1f5099e0>, json_schema_input_type=PydanticUndefined)]], Union[float, Annotated[autograd.tracer.Box, BeforeValidator(func=<function traced_alias.<locals>._validate_box_or_container at 0x734e1f5099e0>, json_schema_input_type=PydanticUndefined), PlainSerializer(func=<function traced_alias.<locals>.<lambda> at 0x734e1f509b20>, return_type=PydanticUndefined, when_used='json')], Annotated[object, BeforeValidator(func=<function traced_alias.<locals>._validate_box_or_container at 0x734e1f5099e0>, json_schema_input_type=PydanticUndefined)]], Union[float, Annotated[autograd.tracer.Box, BeforeValidator(func=<function traced_alias.<locals>._validate_box_or_container at 0x734e1f5099e0>, json_schema_input_type=PydanticUndefined), PlainSerializer(func=<function traced_alias.<locals>.<lambda> at 0x734e1f509b20>, return_type=PydanticUndefined, when_used='json')], Annotated[object, BeforeValidator(func=<function traced_alias.<locals>._validate_box_or_container at 0x734e1f5099e0>, json_schema_input_type=PydanticUndefined)]]]] = None
        [units = um].  Center of object in x, y, and z.
    size : tuple
        [units = um].  Size in x, y, and z directions.

    Example
    -------
    >>> b = Box(center=(1,2,3), size=(2,2,2))

```

The parameter table reference change is kind of expected given the nature of the size of the changes.  Just got to see how it renders, but I suspect there might be issues. I'm actually keen to try to straight up upgrade our docs APIs to pydantic v2 extension based on this branch so trying that now separately. This table was actually a bit of a hack so it's time to sort it out properly.

yaugenst-flex: oooh yeah the type annotations are now much more extensive and probably need to be trimmed down significantly. a lot of the validation is attached to the type annotations

tidy3d/components/autograd/types.py
```diff
@@ -25,32 +28,96 @@
 
 Box.__copy__ = lambda v: _copy(v)
 Box.__deepcopy__ = lambda v, memo: _deepcopy(v, memo)
+Box.__str__ = lambda self: f"{self._value} <{type(self).__name__}>"
+Box.__repr__ = Box.__str__
+
+
+def traced_alias(base_alias, *, name: Optional[str] = None) -> TypeAlias:
```
tylerflex: what is this doing?

tylerflex (commented):
I went through a bunch of files, but it's becoming clear I'm definitely not going to be able to go through all of this line by line. One question for now: what if any are the changes that a user would experience from this PR? A few that come to mind right now are:
 a. I guess if the ArrayLike field is initialized as a list it will become np.ndarray instead of a tuple? 
 b. users will need to also use the pydantic v2 methods instead of the v1 methods, so their scripts could break if they rely on those?

are there any others? and have you run notebooks from this branch yet (out of curiosity...)?

pyproject.toml
yaugenst-flex: no this needs to go before 2.9, it's just for the rc's currently

mahlau-flex:
I have added a commit where all of the conflicts are resolved that can be resolved automatically. This is a bit tricky, because I had to make this a commit that does not resolve the other merge conflicts with some git magic (create two clones of the repo, make changes in one and copy them to the other as a "normal" commit). With the same strategy we should be able to bring this branch back on track piece by piece. In the end, we can squash all of these commits into a single rebase.

@yaugenst-flex if this sounds good to you, I will continue to make more commits here that resolve more and more conflicts. For some of them I will need help though, because at first glance it does not always seem clear what the best strategy for merging is.

tidy3d/web/core/task_core.py
marcorudolphflex: should be `model_dump`?

tidy3d/web/api/webapi.py
marcorudolphflex: should be `model_dump` instead of `dict`?

marcorudolphflex:
we should also update the readme here:
https://github.com/flexcompute/tidy3d/blob/d2bc9d4b7cf4663487e9a38c992872e51bccb90e/tidy3d/components/README.md?plain=1#L137
with `model_dump_json`

tidy3d/components/medium.py
```diff
@@ -6930,10 +7019,14 @@ def perturbed_copy(
...
             return self
 
-        new_dict = self.dict(
+        new_dict = self.model_dump(
             exclude={"eps_inf_perturbation", "poles_perturbation", "perturbation_spec", "type"}
         )
 
+        if all(x is None for x in [temperature, electron_density, hole_density]):
+            new_dict.pop("subpixel")
+            return PoleResidue.parse_obj(new_dict)
```
marcorudolphflex: replace `parse_obj` with `model_validate`?

tidy3d/components/data/dataset.py
marcorudolphflex: `__fields__` -> `model_fields`

tidy3d/web/core/task_core.py
mahlau-flex: which line exactly? git seems to be overwhelmed with this PR and does not show where you made the comment

tidy3d/web/api/webapi.py
mahlau-flex: same here, which line?

tidy3d/components/medium.py
mahlau-flex: good catch!

tidy3d/components/data/dataset.py
mahlau-flex: it should actually be self.__class__.model_fields, self.model_fields is deprecated and will be removed in pydanticV3

tidy3d/web/api/webapi.py
marcorudolphflex: https://github.com/flexcompute/tidy3d/blob/b68df7cf26bade95022edc8587d321ba6bf3aea3/tidy3d/components/base_sim/data/sim_data.py#L172

tidy3d/web/core/task_core.py
marcorudolphflex: https://github.com/flexcompute/tidy3d/blob/b68df7cf26bade95022edc8587d321ba6bf3aea3/tidy3d/web/core/task_core.py#L365

tidy3d/web/api/webapi.py
marcorudolphflex: oopsie, meant https://github.com/flexcompute/tidy3d/blob/ca282a89d8d52795bfd3c6d1ba6fc9fa585a9657/tidy3d/web/api/webapi.py#L960

marcorudolphflex: still both are valid

docs/api/plugins/index.rst
```diff
@@ -1,6 +1,11 @@
 Plugins
 =======
 
+.. warning::
+
+   |:warning:| The 'adjoint' plugin (legacy JAX-based adjoint plugin)
+   was deprecated in Tidy3D 'v2.7.0' and is disabled as of 'v2.9.0'.
```
daquinteroflex: We removed it in 2.10.0 should this diff be here? https://github.com/flexcompute/tidy3d/blob/develop/CHANGELOG.md#removed-1

daquinteroflex (approved):
Massive effort 🎉 💪

PR #2434, opened 2025-05-08
fix[frontend]: fix computation of finite axis length for PolySlab

groberts-flex:
I believe this was only affecting the plotting function, but came from debugging the issue Filipe was seeing when plotting PolySlab with an infinite slab bound. Since the bounds are capped at +/- LARGE_NUMBER, we need the finite length to be able to be larger than LARGE_NUMBER and at a maximum up to 2 * LARGE_NUMBER.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed issue with `CustomMedium` gradients where other frequencies would wrongly contribute to the gradient.
+- Fixed bug when computing `PolySlab` bounds in plotting functions.
 
 ## [2.8.3] - 2025-04-24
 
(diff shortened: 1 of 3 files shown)


yaugenst-flex (commented):
Nice! I wonder, since this only adds a `property`, where would this actually be used in the code?

groberts-flex:
This property is normally inherited from the Planar class in geometry/base.py. It ends up getting used to find the intersections of the geometry with a plotting plane. Possibly it is better to put this into Planar directly instead of its default implementation?

yaugenst-flex:
Ooooh I see, thanks for clarifying. Ok yeah I think your fix makes perfect sense and I also think that it's the right place to change it, since for other geometries the current behavior works. The only comment I would have would be to make the one-liner a bit more readable because currently it's a bit difficult to parse (at least for me):
```py
z_min, z_max = self.slab_bounds
z_min = max(z_min, -LARGE_NUMBER)
z_max = min(z_max,  LARGE_NUMBER)
return z_max - z_min
```

groberts-flex:
good point, like the readability! changed!

yaugenst-flex (approved):

PR #2436, opened 2025-05-09
feat: autograd support for angle gradients of rotated box

rahul-flex:
Autograd support for derivative w.r.t rotation angle for box geometry in the Transformed class. Key changes are:

**Derivative Computation:**

- Applies the chain rule in two stages:

 Uses adjoint sensitivities at face‑mesh sample points to obtain ∂J/∂X_rot. Then derivative of sample points coordinates wrt rotation matrix elements.

**Testing:**
Finite‑difference vs Autograd checks for seven representative angles across all three rotation axes (x, y, z).
grad_theta_fd_vs_ad_axis_y
grad_theta_fd_vs_ad_axis_z
grad_theta_fd_vs_ad_axis_x


yaugenst-flex:
Closing in favor of https://github.com/flexcompute/tidy3d/pull/2475/

PR #2444, opened 2025-05-09
add a TerminalSpec for RF specific mode information

dmarek-flex:
Calculating impedance automatically:

1. Find all conductors in mode plane and merge into isolated conductors, each possible carrying a conduction current.
2. Place current integrals around each conductor (right now just using bounding box)
3. Calculate total current using a new type of current integral (`CompositeCurrentIntegral`) made up of multiple current integrals. The total current flowing in a transmission line is net 0 because of current flowing in and out . To isolate these two terms, we split the currents based on their phase (using an initial reference phase). Then we choose the maximum of the two at the end. We choose the max, because it is possible that some current is flowing back in the PEC boundary or ground plane, which are not totally enclosed by the modal plane and their contribution will be missing.

Need at least one small [backend](https://github.com/flexcompute/compute/pull/1699) change

Todo on:
- [ ] Update Changelog
- [ ] Check coverage and improve

## Greptile Summary

Implements automated impedance calculation for RF transmission lines by introducing a new TerminalSpec system that detects conductors and computes characteristic impedance using voltage/current path integrals.

- Added `tidy3d.components.microwave.terminal_spec.TerminalSpec` class for automated transmission line impedance calculation using conductor detection
- Introduced `CompositeCurrentIntegral` to handle complex current flows with phase-based current isolation for accurate impedance measurements
- Added `PathSpecGenerator` in `components/microwave/path_spec_generator.py` to automatically detect conductors and generate appropriate current paths
- Enhanced `ModeSolver` with characteristic impedance (Z0) calculations integrated with the new terminal system
- Improved code organization by moving visualization code to dedicated `components/microwave/viz.py` module

### DIFF
--- a/tests/test_components/test_microwave.py
+++ b/tests/test_components/test_microwave.py
@@ -26,12 +26,13 @@ from tidy3d.components.microwave.formulas.circuit_parameters import (
 )
 from tidy3d.components.microwave.path_integrals.path_integral_factory import (
     make_current_integral,
+    make_path_integrals,
     make_voltage_integral,
 )
 from tidy3d.components.microwave.path_integrals.path_spec_generator import PathSpecGenerator
 from tidy3d.components.types import Ax, Shapely
 from tidy3d.constants import EPSILON_0
-from tidy3d.exceptions import ValidationError
+from tidy3d.exceptions import SetupError, ValidationError
 
 from ..test_data.test_monitor_data import make_directivity_data
 
@@ -43,6 +44,9 @@ if MAKE_PLOTS:
     from matplotlib import use
 
     use("TkAgg")
+
+COAX_R1 = 0.04
+COAX_R2 = 0.5
 
 
 def make_mw_sim(
@@ -211,7 +215,7 @@ def make_mw_sim(
 
 
 def plot_auto_path_spec(
-    path_spec: tidy3d.components.microwave.path_integrals.current_spec.CompositeCurrentIntegralSpec,
+    path_spec: td.CompositeCurrentIntegralSpec,
     geoms: list[Shapely],
     ax: Ax = None,
 ) -> Ax:
@@ -318,6 +322,53 @@ def test_antenna_parameters():
         antenna_params.partial_gain("invalid")
     with pytest.raises(ValueError):
         antenna_params.partial_realized_gain("invalid")
+
+
+def test_path_integral_plotting():
+    """Test that all types of path integrals correctly plot themselves."""
+
+    mean_radius = (COAX_R2 + COAX_R1) * 0.5
+    size = [COAX_R2 - COAX_R1, 0, 0]
+    center = [mean_radius, 0, 0]
+
+    voltage_integral = td.VoltageIntegralAxisAlignedSpec(
+        center=center, size=size, sign="-", extrapolate_to_endpoints=True, snap_path_to_grid=True
+    )
+
+    current_integral = td.CustomCurrentIntegral2DSpec.from_circular_path(
+        center=(0, 0, 0), radius=0.4, num_points=31, normal_axis=2, clockwise=False
+    )
+
+    ax = voltage_integral.plot(z=0)
+    current_integral.plot(z=0, ax=ax)
+    plt.close()
+
+    # Test off center plotting
+    ax = voltage_integral.plot(z=2)
+    current_integral.plot(z=2, ax=ax)
+    plt.close()
+
+    # Plot
+    voltage_integral = td.CustomVoltageIntegral2DSpec(
+        axis=1, position=0, vertices=[(-1, -1), (0, 0), (1, 1)]
+    )
+
+    current_integral = td.CurrentIntegralAxisAlignedSpec(
+        center=(0, 0, 0),
+        size=(2, 0, 1),
+        sign="-",
+        extrapolate_to_endpoints=False,
+        snap_contour_to_grid=False,
+    )
+
+    ax = voltage_integral.plot(y=0)
+    current_integral.plot(y=0, ax=ax)
+    plt.close()
+
+    # Test off center plotting
+    ax = voltage_integral.plot(y=2)
+    current_integral.plot(y=2, ax=ax)
+    plt.close()
 
 
 @pytest.mark.parametrize("colocate", [False, True])
@@ -451,20 +502,10 @@ def test_auto_path_spec_validation():
 def test_composite_current_integral_validation():
     """Ensures that the CompositeCurrentIntegralSpec is validated correctly."""
 
-    current_spec = (
-        tidy3d.components.microwave.path_integrals.current_spec.CurrentIntegralAxisAlignedSpec(
-            center=(1, 2, 3), size=(0, 1, 1), sign="-"
-        )
-    )
-    voltage_spec = (
-        tidy3d.components.microwave.path_integrals.voltage_spec.VoltageIntegralAxisAlignedSpec(
-            center=(1, 2, 3), size=(0, 0, 1), sign="-"
-        )
-    )
-    path_spec = (
-        tidy3d.components.microwave.path_integrals.current_spec.CompositeCurrentIntegralSpec(
-            center=(1, 2, 3), size=(0, 1, 1), path_specs=[current_spec], sum_spec="sum"
-        )
+    current_spec = td.CurrentIntegralAxisAlignedSpec(center=(1, 2, 3), size=(0, 1, 1), sign="-")
+    voltage_spec = td.VoltageIntegralAxisAlignedSpec(center=(1, 2, 3), size=(0, 0, 1), sign="-")
+    path_spec = td.CompositeCurrentIntegralSpec(
+        center=(1, 2, 3), size=(0, 1, 1), path_specs=[current_spec], sum_spec="sum"
     )
 
     with pytest.raises(pd.ValidationError):
@@ -491,12 +532,10 @@ def test_path_integral_creation():
     )
     current_integral = make_current_integral(path_spec)
 
-    path_spec = tidy3d.components.microwave.path_integrals.voltage_spec.CustomVoltageIntegral2DSpec(
-        vertices=[(0, 1), (0, 4)], axis=1, position=2
-    )
+    path_spec = td.CustomVoltageIntegral2DSpec(vertices=[(0, 1), (0, 4)], axis=1, position=2)
     voltage_integral = make_voltage_integral(path_spec)
 
-    path_spec = tidy3d.components.microwave.path_integrals.current_spec.CustomCurrentIntegral2DSpec(
+    path_spec = td.CustomCurrentIntegral2DSpec(
         vertices=[
             (0, 1),
             (0, 4),
@@ -509,17 +548,15 @@ def test_path_integral_creation():
     _ = make_current_integral(path_spec)
 
     with pytest.raises(pd.ValidationError):
-        path_spec = (
-            tidy3d.components.microwave.path_integrals.current_spec.CustomCurrentIntegral2DSpec(
-                vertices=[
-                    (0, 1, 3),
-                    (0, 4, 5),
-                    (3, 4, 5),
-                    (3, 1, 5),
-                ],
-                axis=1,
-                position=2,
-            )
+        path_spec = td.CustomCurrentIntegral2DSpec(
+            vertices=[
+                (0, 1, 3),
+                (0, 4, 5),
+                (3, 4, 5),
+                (3, 1, 5),
+            ],
+            axis=1,
+            position=2,
         )
 
 
@@ -529,12 +566,8 @@ def test_microwave_mode_spec_validation():
     _ = td.MicrowaveModeSpec()
     _ = td.MicrowaveModeSpec(voltage_spec=None, current_spec=None)
 
-    v_spec = tidy3d.components.microwave.path_integrals.voltage_spec.VoltageIntegralAxisAlignedSpec(
-        center=(1, 2, 3), size=(0, 0, 1), sign="-"
-    )
-    i_spec = tidy3d.components.microwave.path_integrals.current_spec.CurrentIntegralAxisAlignedSpec(
-        center=(1, 2, 3), size=(0, 1, 1), sign="-"
-    )
+    v_spec = td.VoltageIntegralAxisAlignedSpec(center=(1, 2, 3), size=(0, 0, 1), sign="-")
+    i_spec = td.CurrentIntegralAxisAlignedSpec(center=(1, 2, 3), size=(0, 1, 1), sign="-")
 
     # All valid methods
     _ = td.MicrowaveModeSpec(voltage_spec=(v_spec,), current_spec=(i_spec,))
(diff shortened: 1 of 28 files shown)


dmarek-flex:
Still very rough, and I am not particularly happy about needing to define all of these different path spec types. But right now it is the best way I found to get around the circular dependency issue I am having. Basically, I cannot import path integral code into `TerminalSpec`, or I get circular dependencies. 

 @weiliangjin2021 @dbochkov-flexcompute Any ideas to solve this? 
 
 Otherwise you can retrieve impedance info by doing something as simple as:
 
` mode_spec = td.ModeSpec(num_modes=5, target_neff=1.8, terminal_spec = TerminalSpec())`

dmarek-flex:
~Hmm, might be possible if I just add TerminalSpec directly to the various monitors and solvers.~

Nah, having path specs seems to be the best. I'll just add some factory class to convert path specs to the correct path integral version.

dmarek-flex:
Ready for review finally! The main feature this PR provided is automatically computing impedance and adding it to the `ModeData` or `ModeSolverData`. By itself, the feature is not so complicated, but I had a lot of challenges with getting around circular dependencies and organizing the classes. To get impedance information, all the user needs to do is pass ModeSpec like:

`mode_spec = td.ModeSpec(num_modes=4, target_neff=1.8, terminal_spec=td.TerminalSpec())`

to some object that calculates and returns mode data. The simulation will not be affected if used in a `ModeSource`. The user can also pass their own voltage and current specifications to the `TerminalSpec` for a custom definition of impedance. The mode data now has a new field `TerminalSpec` which is populated by the `ModeSpec`, perhaps automatically. This allows users to swap the  `TerminalSpec` to a new definition in a post-processing step, if they desire.

Now, to get around all of the circular dependence stuff, I had to split the path integral classes in the microwave plugin into the specification part, and the part that actually computes the integral. At some later point we will most likely move the everything out of the microwave plugin, since these path integrals are needed for many things in RF. If anyone has suggestions on how to better get around these issues, please let me know! The main issue was the need to import the monitor data module in the path integral modules.

Finally, the auto generated paths are bounding boxes around each conductor, which will work for 99% of transmission lines, but there might be improvements needed in the future to auto generate the line string type path integrals for handling more complex structures.

dmarek-flex:
@yuanshen-flexcompute It would be nice to test out this automatic version of impedance calculation on some of your test cases to make sure I did not miss anything.

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -334,64 +223,82 @@ def compute_current(self, em_field: MonitorDataTypes) -> IntegralResultTypes:
...
+            # Wrap phase difference to [-pi, pi]
+            phase_diff = np.mod(phase_diff + np.pi, 2 * np.pi) - np.pi
+
+            # Check phase consistency across frequencies
+            freq_axis = term.get_axis_num("f")
+            consistent_phase = np.all(
+                np.logical_or(
+                    abs(phase_diff) <= np.pi / 2,  # in phase
+                    abs(phase_diff) > np.pi / 2,  # out of phase
+                ),
```
dbochkov-flexcompute: hmm, is it not always an array of `True`'s? seems like it is doing `(a or (not a))`?

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -334,64 +223,82 @@ def compute_current(self, em_field: MonitorDataTypes) -> IntegralResultTypes:
...
+        current_out_phase = xr.zeros_like(first_term)
+
+        # Get reference phase from first non-zero current
+        phase_reference = None
+        for path in current_integrals:
+            term = path.compute_current(em_field)
+            if np.any(abs(term) > 0):
+                phase_reference = np.angle(term)
+                break
+        assert phase_reference is not None, "No non-zero current found"
```
dbochkov-flexcompute: any reason to use `assert` instead of `log.error`?

tidy3d/plugins/microwave/impedance_calculator.py
```diff
@@ -71,20 +78,20 @@ def compute_impedance(self, em_field: MonitorDataTypes) -> IntegralResultTypes:
         # a time signal, then it is real and flux corresponds to the instantaneous power. Otherwise
         # the input field is in frequency domain, where flux indicates the time-averaged power
         # 0.5*Re(V*conj(I)).
-        if not self.voltage_integral:
+        if self.voltage_integral is None:
             flux = em_field.flux
             if isinstance(em_field, FieldTimeData):
-                voltage = flux.abs / current
+                impedance = flux.abs / current**2
```
dbochkov-flexcompute: I forgot, do we return complex or real values in time monitors in case of bloch bc simulations (is it even applicable here?)?

tidy3d/components/data/monitor_data.py
```diff
@@ -2003,6 +2011,28 @@ def pol_fraction_waveguide(self) -> xr.Dataset:
 
         return xr.Dataset(data_vars={"te": te_frac, "tm": tm_frac})
 
+    @cached_property
+    def line_impedance(self) -> FreqModeDataArray:
```
dbochkov-flexcompute: add a doctring

tidy3d/components/data/monitor_data.py
```diff
@@ -2003,6 +2011,28 @@ def pol_fraction_waveguide(self) -> xr.Dataset:
...
+        from tidy3d.components.microwave.path_integral_factory import (
+            make_current_integral,
+            make_voltage_integral,
+        )
+        from tidy3d.plugins.microwave.impedance_calculator import ImpedanceCalculator
```
dbochkov-flexcompute: any reason not to import these at the top level?

dbochkov-flexcompute (commented):
Nice work! Gave it a first pass, didn't spot anything major, just some minor comments/questions so far

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: It does look wrong. The intention was to make sure the phase is the same over all frequencies. But it can be different for each mode

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: No, just forgot to update

tidy3d/components/data/monitor_data.py
dmarek-flex: circular dependency :cry: Many of those popped up in this PR

daquinteroflex: Yes I think this is part of why we need to do the RF refactor soon on the lines of what we chatted back on Friday too.

tidy3d/plugins/microwave/impedance_calculator.py
dmarek-flex: hmm, I have not considered this. It looks like the time monitor stores a complex field, for example the poynting calculation. I'll try to confirm though. Would it be enough for me to simply take the real part of the current/voltage in the case of a time domain signal?

https://github.com/flexcompute/tidy3d/blob/c955b0ed99e85fe853b0d4d383d065d7e0be533a/tidy3d/components/data/monitor_data.py#L1427-L1436

dbochkov-flexcompute: yeah, I believe just taking real part of the field is enough, as the imaginary part can be thought of as an auxiliary field that facilitates simulation of bloch bc

dmarek-flex:
Planning on changing the name to `ImpedanceSpec`, which makes more sense. A terminal would correspond with a single conductor, so doesn't really make sense here.

dbochkov-flexcompute:
> Planning on changing the name to `ImpedanceSpec`, which makes more sense. A terminal would correspond with a single conductor, so doesn't really make sense here.

That does make more sense. Are there any other impedance specs that might be introduced in future? Just wondering if it would also make sense to expand it to `ModeImpedanceSpec` or `LineImpedanceSpec` to prevent any possible name collisions in future (field name `impedance_spec` doesn't need to be expanded though)

dmarek-flex:
> That does make more sense. Are there any other impedance specs that might be introduced in future? Just wondering if it would also make sense to expand it to `ModeImpedanceSpec` or `LineImpedanceSpec` to prevent any possible name collisions in future (field name `impedance_spec` doesn't need to be expanded though)

Yea I think `ModeImpedanceSpec` would be a good expansion for it, that is its main purpose.

yaugenst-flex:
@greptileai

tidy3d/components/mode/mode_solver.py
```diff
@@ -29,6 +29,7 @@
 from tidy3d.components.geometry.base import Box
 from tidy3d.components.grid.grid import Coords, Grid
 from tidy3d.components.medium import FullyAnisotropicMedium, LossyMetalMedium
+from tidy3d.components.microwave.path_integral_factory import make_path_integrals
```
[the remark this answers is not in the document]
yaugenst-flex: imports are sorted automatically using ruff/isort, do not suggest changes to them

tests/test_components/test_microwave.py
```diff
@@ -3,24 +3,224 @@
...
+    make_voltage_integral,
+)
+from tidy3d.components.microwave.path_spec_generator import PathSpecGenerator
+from tidy3d.components.types import Ax, Shapely
 from tidy3d.constants import EPSILON_0
+from tidy3d.exceptions import ValidationError
 
 from ..test_data.test_monitor_data import make_directivity_data
 
+mm = 1e3
```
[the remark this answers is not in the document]
yaugenst-flex: we use micrometers as the base unit and this is well understood in our code base & domain. leave this as-is

tidy3d/components/geometry/utils.py
```diff
@@ -38,6 +47,46 @@
...
+    for geom in geoms:
+        if geom.is_empty:
+            continue
+        if isinstance(geom, keep_types):
+            flat.append(geom)
+        elif isinstance(geom, (MultiPolygon, MultiLineString, MultiPoint, GeometryCollection)):
+            flat.extend(flatten_shapely_geometries(geom.geoms, keep_types))
+        elif isinstance(geom, BaseGeometry) and hasattr(geom, "geoms"):
+            flat.extend(flatten_shapely_geometries(geom.geoms, keep_types))
+    return flat
```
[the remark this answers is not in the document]
dmarek-flex: That type of error would not be handled in this function, so best to not check either.

dmarek-flex:
@dbochkov-flexcompute I finally settled on `MicrowaveModeSpec`, the reason being is that it really is a bit more than simply defining how impedance is calculated. It also defines how voltage/current should be related to the waveguide mode, which will be used in the WavePort. Although easy enough to change.

Also in the future, I can see how it might provide additional features/specification fields, like making linear combinations of modes to excite a specified conductor, instead of relying on the set of modes returned by the mode solver. Also combining modes to give a desired polarization direction that is based on the voltage/current paths.

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -238,6 +285,44 @@ def get_port_by_name(self, port_name: str) -> Port:
...
+        """Abstract property for all matrix indices that will be used to collect data."""
+
+    @cached_property
+    def matrix_indices_source(self) -> tuple[IndexType, ...]:
+        """Tuple of all the source matrix indices, which may be less than the total number of ports."""
+        if self.run_only is not None:
+            return self.run_only
+        return self.matrix_indices_monitor
+
+    @cached_property
```
daquinteroflex, line 214: So I'm trying to get familiar with the workflow of why these cached_properties are required? Since they introduce mutability

I'm aware some of this functionality has alrady been merged into develop

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -95,12 +169,70 @@ def plot_sim_eps(
...
+            sim_dict[task_name] = sim_with_src
 
+        # Check final simulations for grid size at ports
+        for _, sim in sim_dict.items():
+            TerminalComponentModeler._check_grid_size_at_ports(sim, self._lumped_ports)
+            TerminalComponentModeler._check_grid_size_at_wave_ports(sim, self._wave_ports)
+
+        return sim_dict
+
+    @cached_property
```
daquinteroflex, line 272: Same here re the cached_properties usage

Just curious how we're checking there are no internal state changes from when they get computed and that they are really pure and immutable

tidy3d/components/data/monitor_data.py
```diff
@@ -1607,11 +1607,37 @@ class ModeData(ModeSolverDataset, ElectromagneticFieldData):
...
+    voltage_amps: Optional[FreqModeDataArray] = pd.Field(
+        None,
+        title="Mode Voltage Amplitudes",
+        description="Optional quantity calculated for transmission lines, which associates "
+        "a voltage-like quantity with each mode profile. "
+        "The mode voltages are only calculated when a :class:`MicrowaveModeSpec` "
+        "is provided to the :class:`ModeSpec` associated with this data.",
+    )
+
+    current_amps: Optional[FreqModeDataArray] = pd.Field(
```
daquinteroflex: amps could also mean Amperes. Maybe something like voltage_magnitude is fine?

tidy3d/components/data/monitor_data.py
```diff
@@ -1639,7 +1665,7 @@ def overlap_sort(
         frequency according to their overlap values with the modes at the previous frequency.
         That is, it attempts to rearrange modes in such a way that a given ``mode_index``
         corresponds to physically the same mode at all frequencies. Modes with overlap values over
-        ``overlap_tresh`` are considered matching and not rearranged.
+        ``overlap_thresh`` are considered matching and not rearranged.
```
daquinteroflex, line 1688: Haha why not three more letters and `overlap_threshold`?

tidy3d/components/microwave/path_integral_factory.py
```diff
@@ -0,0 +1,148 @@
+"""Factory functions for creating current and voltage path integrals from path specifications."""
```
daquinteroflex, line 1: This file makes sense as a scope combiner

tidy3d/components/microwave/path_spec.py
```diff
@@ -0,0 +1,770 @@
...
+    @staticmethod
+    def _compute_dl_component(coord_array: xr.DataArray, closed_contour=False) -> np.array:
+        """Computes the differential length element along the integration path."""
+        dl = np.gradient(coord_array)
+        if closed_contour:
+            # If the contour is closed, we can use central difference on the starting/end point
+            # which will be more accurate than the default forward/backward choice in np.gradient
+            grad_end = np.gradient([coord_array[-2], coord_array[0], coord_array[1]])
+            dl[0] = dl[-1] = grad_end[1]
+        return dl
```
daquinteroflex: Can also live in a utils separate file, and just use it where you need. Helpful for reusability

tidy3d/components/microwave/path_spec.py
```diff
@@ -0,0 +1,770 @@
...
+from .viz import (
+    ARROW_CURRENT,
+    plot_params_current_path,
+    plot_params_voltage_minus,
+    plot_params_voltage_path,
+    plot_params_voltage_plus,
+)
+
+
+class AbstractAxesRH(Tidy3dBaseModel, ABC):
```
daquinteroflex: Does it make sense to move all these axes aligned specs to another file to avoid making this file very big?

tidy3d/components/microwave/path_spec.py
```diff
@@ -0,0 +1,770 @@
...
+        """
+        for path_spec in self.path_specs:
+            ax = path_spec.plot(x=x, y=y, z=z, ax=ax, **path_kwargs)
+        return ax
+
+
+VoltagePathSpecTypes = Union[VoltageIntegralAxisAlignedSpec, CustomVoltageIntegral2DSpec]
+CurrentPathSpecTypes = Union[
+    CurrentIntegralAxisAlignedSpec, CustomCurrentIntegral2DSpec, CompositeCurrentIntegralSpec
+]
```
daquinteroflex: Maybe if you look at the refactor PR, something like:

```
path_integrals/
  __init__.py
  axes_align.py
  voltage.py
  current.py
  types.py # these types live here so they are manageable in scope
``

File: tidy3d/components/data/monitor_data.py (hunk 12 above)
Comment (dmarek-flex) [reply tidy3d/components/data/monitor_data.py:1688]: I was just trying to fix a typo in the original doc string :smile:. I am going to make a PR for this little fix, which is a separate little bug I found.

File: tidy3d/components/microwave/path_spec_generator.py [hunk 17]
```diff
@@ -0,0 +1,326 @@
...
+
+    This class analyzes the geometry of conductors in a simulation cross-section to determine
+    appropriate paths for computing current line integrals. These paths are typically used
+    for setting up a :class:`.MicrowaveModeSpec` automatically.
+
+    The paths are chosen by:
+    1. Finding and merging conductor surfaces that intersect with the plane
+    2. Creating current paths that enclose isolated conductors
+    """
+
```
daquinteroflex: Is this class just a collection of static methods? why not simply have standalone static methods and use them where required?

tidy3d/plugins/microwave/path_integrals.py
daquinteroflex, line 1: Yeah it does make sense to move a bunch of this stuff to components as long as the __init__ files remain the same for now.

tidy3d/components/microwave/path_spec_generator.py
dmarek-flex: Originally, I thought there might be some kind of options for controlling how these steps take place. I'll consider making them as standalone functions

PR #2448, opened 2025-05-12
:warning:  RF License Validator Warning

daquinteroflex:
- [x] Adds warning as part of the user level validation of RF components.
- [x] Appears only once on instantiation of the first RF component. Originally, I had it print also which class caused the warning, but realised it would be very cumbersome. Unclear
- [x] Improved simulation logging and component breakdown
- [x] Make sure this only covers RF-specific classes, and not shared optical classes too
    - [x] Consolidate warnings after discussion
- [x] Implement docs 
- [ ] (Future once this is fully merged) )Make notebooks warning update PR (just run a few notebooks with the warning).
- ~Think how we would want, if we want, to do any testing.~

@weiliangjin2021 so was thinking we can combine this and your PR so it's "nicer" for users to be informed as they define their simulations and problems. There's a few ways to do this.

1. On each class-type instantiation, it can print a FutureWarning that specifies that a class is RF related and will be subject to licensing issues like below. It will only print once per class type.
image

2. We can make the RF licensing warning appear only once for any RF component, just so they're aware.
image

I've updated the simulation logging so it appears like this:
image

Docs are building 

My concern with option 1 is that it will become very overwhelming with so many classes and instantiations, so I'm inclined for option 2 and then maybe we combine it with your logging detailed breakdown? 

See an example notebook on how the logs are shown, seems good to me as it only prints once when a RF component has been instantiated and should cover the entire scope without cluttering.
[AntennaCharacteristics.zip](https://github.com/user-attachments/files/20186389/AntennaCharacteristics.zip)

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -15,6 +15,7 @@ try:
 except ImportError:
     pass
 
+
 import pydantic.v1 as pydantic
 import xarray as xr
 
@@ -173,6 +174,9 @@ PML_HEIGHT_FOR_0_DIMS = inf
 
 # additional (safety) time step reduction factor for fixed angle simulations
 FIXED_ANGLE_DT_SAFETY_FACTOR = 0.9
+
+# RF frequency warning
+RF_FREQ_WARNING = 1e12
 
 
 def validate_boundaries_for_zero_dims():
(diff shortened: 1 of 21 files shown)


tidy3d/components/simulation.py
```diff
@@ -3599,37 +3601,47 @@ def _post_init_validators(self) -> None:
...
         for mat in self.scene.mediums:
             if isinstance(mat, LossyMetalMedium):
-                comp_msg += ["it contains 'LossyMetalMedium'"]
+                rf_component_breakdown_msg += "\n - Contains a 'LossyMetalMedium'"
                 break
 
         # 2) lumped elements
         if len(self.lumped_elements) > 0:
-            comp_msg += ["it contains lumped elements"]
+            rf_component_breakdown_msg += "\n - Contains a 'LossyMetalMedium'"
```
weiliangjin2021: LossyMetalMedium --> lumped element

tidy3d/components/simulation.py
```diff
@@ -3599,37 +3601,47 @@ def _post_init_validators(self) -> None:
...
 
         # 3) source frequency is in RF range
         if self.frequency_range[0] < RF_FREQ_WARNING:
-            comp_msg += ["frequency of some sources is in the RF range"]
+            rf_component_breakdown_msg += (
+                f"\n - Contains a sources defined for RF wavelengths: below {RF_FREQ_WARNING} Hz."
```
weiliangjin2021: We probably don't want to display the threshold frequency value in the warning.

yaugenst-flex:
Instead of adding this validator to so many classes, would it make sense to add it once as a pre upload check instead?

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -307,3 +308,11 @@ def sim_data_by_task_name(self, task_name: str) -> SimulationData:
...
         config.logging_level = log_level_cache
         return sim_data
+
+    @pd.root_validator(pre=False)
+    def _warn_rf_license(cls, values):
+        warnings.warn(
+            "ℹ️ ⚠️ RF simulations are subject to new license requirements in the future. You are have instantiated a RF-specific component.",
+            FutureWarning,
+        )
+        return values
```
weiliangjin2021: move to terminal.py

weiliangjin2021:
> Instead of adding this validator to so many classes, would it make sense to add it once as a pre upload check instead?

In simulation level, we have it already in sim.post-init validator. But we also need to validate a higher-level class that is generating simulation class, or postprocessing simulation data.

dmarek-flex:
Missing classes:

- RF Material library (maybe)
- Should also include `DirectivityMonitor`

weiliangjin2021 (approved):

PR #2452, opened 2025-05-13
relax wave port bounds check

dmarek-flex:
Relaxed validation step of wave port bounds. As long as the path is almost coincident with the boundaries of the mode field data the voltage/current calculations will work accurately.

Also need to modify the `does_cover` helper, to enable the same type of check. I moved some `Bound` related functions to their own file to avoid a circular dependency.

TODO:

- [x] Changelog


weiliangjin2021 (approved):

yaugenst-flex (approved):
Thanks @dmarek-flex LGTM! Sorry for getting to this one so late. Needs a rebase and (probably?) changelog, otherwise good to go.

yaugenst-flex:
Everything passing as per https://github.com/flexcompute/tidy3d/actions/runs/15155392963

PR #2453, opened 2025-05-13
feat: add validation for grid spacing in AutoGrid

### LINKED ISSUE #2449: Validate units after AutoGrid generation
Issue to follow up on my comment on #2335 :

A user just had some simulations erroring because they had obviously set units wrong, which resulted in a grid spacing on the scale of 1e-9. However, this was still an auto grid (they had also set source frequency wrong), so it was not caught by this validator. A large number of our users probably use auto grid instead of uniform grid - it would be good to have a similar validator there. However, in that case it should be a Simulation post_init validator, as the grid needs to be constructed to examine the smallest dl.

rahul-flex:
Extend `Simulation._post_init_validators` to compute the minimum cell size (dl) after AutoGrid construction and validate it against threshold [1e-7 µm].  Addresses this [issue](https://github.com/flexcompute/tidy3d/issues/2449).

Note: The pytest takes a lot of time and memory to run locally, because it builds the full grid before running the dl check.

## Greptile Summary

Added validation to prevent numerical instability by enforcing minimum grid spacing (dl) of 1e-7 µm in AutoGrid calculations.

- Added minimum grid spacing validation in `Simulation._post_init_validators` to catch problematic small cell sizes
- Added unit test `test_autogrid_min_spacing` with both passing (freq0=1e14) and failing (freq0=3e21) test cases
- Validation checks cell-boundary differences in each dimension, ignoring zero-length dimensions
- Updated CHANGELOG.md to document this validation feature under 'Added' section

### DIFF
--- a/tests/test_components/test_grid_spec.py
+++ b/tests/test_components/test_grid_spec.py
@@ -541,3 +541,41 @@ def test_uniform_grid_dl_validation(dl, expect_exception):
             grid_spec=td.GridSpec.uniform(dl=dl),
             run_time=1e-12,
         )
+
+
+@pytest.mark.parametrize(
+    ("freq0", "expect_error"),
+    [
+        (3e21, True),  # dl<1e-6 → fail
+        (1e14, False),  # dl>1e-7 → pass
+    ],
+)
+def test_autogrid_min_spacing(freq0, expect_error):
+    """
+    Test that Simulation.post_init catches too‐fine spacing produced by AutoGrid
+    (i.e. min grid spacing < 1e-7 µm).
+    """
+    L = 10.0
+    buffer = 1.0
+    source = td.PointDipole(
+        center=(-L / 2 + buffer, 0, 0),
+        source_time=td.GaussianPulse(freq0=freq0, fwidth=freq0 / 10.0),
+        polarization="Ez",
+    )
+
+    kwargs = dict(
+        size=(L, L, L),
+        sources=[source],
+        grid_spec=td.GridSpec(
+            grid_x=td.AutoGrid(min_steps_per_wvl=10),
+            grid_y=td.AutoGrid(min_steps_per_wvl=10),
+            grid_z=td.AutoGrid(min_steps_per_wvl=10),
+        ),
+        run_time=1e-11,
+    )
+
+    if expect_error:
+        with pytest.raises(SetupError):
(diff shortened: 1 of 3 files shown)


tests/test_components/test_grid_spec.py
```diff
@@ -541,3 +541,41 @@ def test_uniform_grid_dl_validation(dl, expect_exception):
...
         )
+
+
+@pytest.mark.parametrize(
+    ("freq0", "expect_error"),
+    [
+        (3e21, True),  # dl<1e-6 → fail
+        (1e14, False),  # dl>1e-7 → pass
+    ],
+)
```
momchil-flex, line 552: if you just make the simulation size 1e7 smaller wouldn't that avoid the long runtime of the test?

tests/test_components/test_grid_spec.py
```diff
@@ -541,3 +541,41 @@ def test_uniform_grid_dl_validation(dl, expect_exception):
             grid_spec=td.GridSpec.uniform(dl=dl),
             run_time=1e-12,
         )
+
+
+@pytest.mark.parametrize(
+    ("freq0", "expect_error"),
+    [
+        (3e21, True),  # dl<1e-6 → fail
+        (1e14, False),  # dl>1e-7 → pass
```
momchil-flex, line 550: 1e-6 for both

tidy3d/components/simulation.py
```diff
@@ -3596,6 +3619,9 @@ def _post_init_validators(self) -> None:
         self._validate_custom_source_time()
         self._validate_mode_object_bends()
         self._warn_mode_object_pml()
+        _validate_min_grid_spacing(
+            self.grid, threshold=1e-7, msg_prefix=f"'{self.__class__.__name__}'"
+        )
```
momchil-flex, line 3624: Let's do 1e-6, even that sounds too large, but to be consistent with the uniform grid validator. Or maybe we should increase the threshold for both? I don't know if I missed anything in the other discussion but wouldn't even 1e-5 be more than reasonable (hard to imagine any simulation that needs to go even below 1e-4). @yaugenst ?

tests/test_components/test_grid_spec.py
rahul-flex: Yes, it reduces the runtime. Thanks!

PR #2455, opened 2025-05-13
adjust adjoint source fwidth to decay before zero frequency when possible

groberts-flex:
For sources that run a single frequency simulation, this chooses a source width according to decay naturally before f=0. For sources that run a broadband simulation and are custom current sources, we attempt to adjust the source width to decay before f=0 while covering all of the adjoint frequencies in the pulse. When this is not possible, a warning is generated.

The results for a custom current source before and after the change are below for when we have a forward simulation with fwidth=f0

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Internal adjoint helper methods are now prefixed with an underscore to separate them from the public API.
 - Drop the dependency on `gdspy`, which has been unmaintained for over two years. Interfaces previously relying on `gdspy` now use its maintained successor, `gdstk`, with equivalent functionality.
 - Small (around 1e-4) numerical precision improvements in EME solver.
+- Adjoint source frequency width is adjusted to decay sufficiently before zero frequency when possible to improve accuracy of simulation normalization when using custom current sources.
 
 ## [2.8.4] - 2025-05-15
 
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -1225,6 +1225,147 @@ def test_no_freq_adjoint(monkeypatch, use_emulated_run):
         ag.grad(objective)(params0)
 
 
+def test_adjoint_src_width():
+    """Test the adjoint source width for single sources decays by f=0."""
+
+    f0 = td.C_0 / 1.55
+    fwidth = f0
+
+    fwidths = f0 * np.linspace(0.1, 1.0, 5)
+
+    adj_srcs = [
+        td.PointDipole(
+            center=(0, 0, 0),
+            source_time=td.GaussianPulse(freq0=f0, fwidth=fwidth),
+            polarization="Ex",
+        )
+        for fwidth in fwidths
+    ]
+
+    adj_srcs_fwidth = td.SimulationData._adjoint_src_width_single(adj_srcs)
+
+    for src in adj_srcs_fwidth:
+        assert np.isclose(
+            (src.source_time.freq0 - f0) / f0, 0.0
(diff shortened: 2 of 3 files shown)


groberts-flex:
There is also minor improvement for mode sources when fwidth=f0 as seen below:

image

tidy3d/components/data/sim_data.py
```diff
@@ -1187,9 +1206,32 @@ def process_adjoint_sources_broadband(
...
+        # width of source to sufficiently decay by zero frequency
+        decay_by_f0_fwidth = middle_f0 / NUM_ADJOINT_FWIDTH_TO_ZERO
+        # width of source to sufficiently cover all adjoint frequencies
+        fwidth_to_min_f0 = (middle_f0 - min_f0) / NUM_ADJOINT_FWIDTH_TO_FMIN
+
+        # log warning if the adjoint pulse width is not sufficiently decayed by zero frequency
+        # which may cause some issues in the adjoint accuracy when using field sources
+        if (fwidth_to_min_f0 > decay_by_f0_fwidth) and isinstance(adj_srcs[0], CustomCurrentSource):
+            log.warning(
+                "Adjoint source generated with a frequency spectrum that extends to or overlaps with 0 Hz."
```
yaugenst-flex: Space missing: `Hz.` -> `Hz. `

yaugenst-flex (commented):
Thanks for this fix @groberts-flex, code looks good overall. Can we add a test for this behavior? I think it should be relatively simple to just construct a couple of adjoint sources and run them through `_make_broadband_source`?

groberts-flex:
> Thanks for this fix @groberts-flex, code looks good overall. Can we add a test for this behavior? I think it should be relatively simple to just construct a couple of adjoint sources and run them through `_make_broadband_source`?

Yes, definitely, great point! Added a couple of tests for this to make sure the single source widths were being set correctly as well as the broadband source (and that we issue a warning for custom field sources that still end up wide bandwidth based on the adjoint frequencies being computed for)

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Warn if more than 20 frequencies are used in EME, as this may lead to slower or more expensive simulations.
 - EME now supports 2D simulations.
 - 'EMESimulation' now supports 'PermittivityMonitor'.
+- Change `VisualizationSpec` validator for checking validity of user specified colors to only issue a warning if matplotlib is not installed instead of an error.
```
yaugenst-flex, line 38: Did this slip in?

yaugenst-flex: Did this slip in?

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Adjoint source frequency width is adjusted to decay sufficiently before zero frequency when possible to improve accuracy of simulation normalization when using custom current sources.
```
yaugenst-flex: Changelog needs to go under "Unreleased"

tests/test_components/test_autograd.py
yaugenst-flex, line 1: Looks great!

CHANGELOG.md
groberts-flex: Hmm yeah I'm not sure what happened here because I'm not seeing this VisualizationSpec changelog entry on develop but the PR with it was merged a few days ago.  Looking back at that PR, I think I had put that entry in the wrong location as well (2.8.4 instead of unreleased).

CHANGELOG.md
groberts-flex: thank you, good catch!

yaugenst-flex (approved):

PR #2424, opened 2025-05-01
ABC boundary

dbochkov-flexcompute:
- Introduces the ability to set first order absorbing boundary conditions on simulation domain boundaries using either `ABCBoundary` or `ModeABCBoundary`. 
  - In case of `ABCBoundary` the user my provide effective permittivity and conductivity to use in one-way wave equation. Otherwise, Tidy3D will autodetect the meidum on the boundary and use its parameters (if multiple mediums or anisotropic medium, it will throw an error)
  - In case of `ModeABCBoundary` the user need to specify the plane and mode spec for finding the mode to absorb. If there is already a modal source or a monitor, `ModeABCBoundary.from_source()` and `ModeABCBoundary.from_monitor()` can be used.
- Introduces field `frame` in `ModeSource` (default=None) and `WavePort` (default: `PECFrame()`), which could currently be set to `PECFrame(length=...)` to automatically place a thin PEC frame of a given length around them.
  - Automatic creation of frames is done in `Simulation.with_port_frames`, which could be used to visualize where they are placed exactly. This is what backend also uses for running actual simulation
- Introduces `PortAbsorber` class that represent a plane inside the simulation domain where first order absorbing boundary conditions will be placed.
  - Port absorbers are passed to `Simulation` through the field `absorbers`.
  - In simulation plots they are drawn similar to sources but in blue color.
  - `WavePort` now have a field `absorber: bool = True` that allows to automatically put an absorber in the port.
  - Each `PortAbsorber` automatically gets a pec frame of length one and also PEC backing to separate absorbed field from surroundings.
- Adds an option in `ModeMonitors` and `WavePort` to use non-conjugated dot product for mode decomposition with defaults 

backend https://github.com/flexcompute/compute/pull/1445 

Screenshot_20250719_114118

## Greptile Summary

This PR introduces comprehensive support for Absorbing Boundary Conditions (ABC) in Tidy3D electromagnetic simulations. The implementation adds first-order absorbing boundary conditions as an alternative to PML boundaries, particularly useful for scenarios involving dispersive materials or structures intersecting boundaries at angles where PML might cause divergence issues.

The key additions include three main boundary condition types:
- `ABCBoundary`: Standard ABC with optional permittivity/conductivity parameters, with automatic medium detection when parameters aren't provided
- `ModeABCBoundary`: Mode-specific ABC that absorbs specific waveguide modes based on plane and mode specifications
- `InternalAbsorber`: Absorbing planes placed inside the simulation domain rather than at boundaries

[...]

PR #2456, opened 2025-05-14
mode solver pec frame

dbochkov-flexcompute:


dbochkov-flexcompute:
implemented in https://github.com/flexcompute/tidy3d/pull/2424

PR #2457, opened 2025-05-14
fix[web]: use dedicated endpoint for `web.delete_old()`

yaugenst-flex:


tidy3d/web/api/webapi.py
```diff
@@ -846,6 +844,7 @@ def delete(task_id: TaskId, versions: bool = False) -> TaskInfo:
 
 @wait_for_connection
 def delete_old(
+    *,
```
tylerflex: I think this breaks backwards compatibility. not sure. I get 
```
 python3 -c "import tidy3d.web as web; web.delete_old(1)"
  File "<string>", line 1, in <module>
    import tidy3d.web as web; web.delete_old(1)
                              ~~~~~~~~~~~~~~^^^
  File "/Users/twhughes/Documents/Flexcompute/tidy3d/tidy3d/web/api/connect_util.py", line 34, in web_fn_wrapped
    return web_fn(*args, **kwargs)
TypeError: delete_old() takes 0 positional arguments but 1 was given
```
whereas this would work before

tylerflex: once I resolved this I still got that connection error from before

yaugenst-flex: yeah I added the keyword only argument because I ran into wrong argument order multiple times, but I can revert that change too. otherwise this runs fine for me though

momchil-flex (approved):

PR #2459, opened 2025-05-14
Last updates for 2.8.4

momchil-flex:
The time has (almost) come!


tests/_test_notebooks/full_test_notebooks.py
daquinteroflex, line 1: Fyi @frederikschubertflex

momchil-flex: Should add some more to the "basic" ones, like some RF.

yaugenst-flex (approved):

PR #2460, opened 2025-05-14
:wrench: :warning: Fix RF scope warning

daquinteroflex:


momchil-flex (commented):
Should this string be a single variable that can be imported? :D

daquinteroflex:
Haha so I intentionally didn't do that because:
1. The warning should be temporary. 
2. Didn't want to create a `tidy3d.warnings` and extra import file because of that.
3. We can update them all with the IDE. Easier to do a straightforward implementation that can be replaced with a find-all.

PR #2463, opened 2025-05-14
split validators to help avoid circular dependencies

dmarek-flex:
Can we do something like in this PR to help avoid circular dependencies? In my upcoming feature branch `ModeSpec` is importing my own module which itself importing validators. I already resolved many other circular dependencies by splitting my own modules, but this seems like a fairly reasonable split to me.

### DIFF
--- a/tidy3d/components/mode/mode_solver.py
+++ b/tidy3d/components/mode/mode_solver.py
@@ -60,9 +60,11 @@ from ..types import (
 from ..validators import (
     validate_freqs_min,
     validate_freqs_not_empty,
-    validate_mode_plane_radius,
 )
 from ..viz import make_ax, plot_params_pml
+from .validators import (
+    validate_mode_plane_radius,
+)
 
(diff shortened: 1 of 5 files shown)


dmarek-flex:
@yaugenst-flex For your Pydantic knowledge

yaugenst-flex:
One way to get around this is to just import the module _within_ the validator. It's a relatively common pattern in Python, but it does hide the module dependencies somewhat. I was also getting pretty confused with all the different places where we put validators in https://github.com/flexcompute/tidy3d/pull/2433 so eventually this all should be cleaned up.

In that light, I think this change is ok for now if you need it. We have a lot of circular dependencies that we generally manage to get around by just throwing everything into one big file...

momchil-flex (approved):
Ah. Well this is fine by me and is kind of how we've been doing things. :D

dmarek-flex:
great ill just make these changes directly in my other PR

PR #2466, opened 2025-05-14
remove pulse amplitude and phase of forward source time in broadband adjoint source

groberts-flex:
For a broadband source, the post norm values are set assuming the adjoint source time has amplitude=1 and phase=0, but the source_time is being extracted from the forward simulation. If the user has a phase or amplitude in it, then we aren't normalizing properly with the post norm. We could either update the post norm or just remove this amplitude and phase, which is what I did. Examples of numerical gradients are below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed issue with `CustomMedium` gradients where other frequencies would wrongly contribute to the gradient.
 - Fixed bug when computing `PolySlab` bounds in plotting functions.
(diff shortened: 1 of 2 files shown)


groberts-flex:
before this change, here is what happens to the computed gradients if you add an amplitude=10 in the forward source (first image is amplitude=1 and second is amplitude=10)

image
image

groberts-flex:
and here is what happens to the computed gradients if you add phase=np.pi / 4 in the forward source (first image is phase=0 and second is phase=np.pi / 4)

image

image

groberts-flex:
after the change here are all three cases (first is (amplitude=1, phase=0), second is (amplitude=10, phase=0), third is (amplitude=1, phase=np.pi/4.))

image

image


image

yaugenst-flex (approved):
Great catch @groberts-flex, thanks!

PR #2467, opened 2025-05-14
adding eps_lim argument to plot_eps

FilipeFcp:
Hi all,
We've noticed that the option to control the color limits in the `Simulation.plot_eps` method is missing, and it could be handy sometimes. Since the `Scene` plot functions already support this argument, it seemed straightforward to implement, so I gave it a shot.

[Here's](https://github.com/FilipeFcp/Tests/blob/main/plot_limit_test.ipynb) a notebook to test it out.

I'm happy to hear any feedback.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -541,6 +541,7 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
         vlim: Tuple[float, float] = None,
         ax: Ax = None,
         eps_component: Optional[PermittivityComponent] = None,
+        eps_lim: Tuple[Union[float, None], Union[float, None]] = (None, None),
     ) -> Ax:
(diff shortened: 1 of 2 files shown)


tidy3d/components/simulation.py
```diff
@@ -541,6 +541,7 @@ def plot_eps(
         vlim: Tuple[float, float] = None,
         ax: Ax = None,
         eps_component: Optional[PermittivityComponent] = None,
+        eps_lim: Tuple[Union[float, None], Union[float, None]] = (None, None),
```
yaugenst-flex, line 544: For future reference: `Union[float, None]` should be `Optional[float]`

yaugenst-flex (approved):
Thanks @FilipeFcp, looks great!

tomflexcompute (approved):

yaugenst-flex:
@FilipeFcp could you also add a changelog entry and rebase to develop? should be good to go then

PR #2468, opened 2025-05-15
fix multiphysics medium attribute lookup

yaugenst-flex:
The custom `__getattr__` from https://github.com/flexcompute/tidy3d/pull/2431 intercepts every missing-attribute lookup, including internals such as `_cached_properties`. During model construction or when a `@cached_property` is accessed, Pydantic tries to read that attribute, hits this handler, and raises, breaking any code path that touched cached properties. This surfaced in https://github.com/flexcompute/tidy3d/pull/2433 because that PR makes the `_cached_properties` fields a `PrivateAttr`, but this problem is also present on `develop` for any Pydantic internal field.

This PR addresses that by delegating all unresolved lookups (fields, cached props, dunders) back to Pydantic before our own logic runs.

### DIFF
--- a/tidy3d/components/material/multi_physics.py
+++ b/tidy3d/components/material/multi_physics.py
@@ -128,6 +128,12 @@ class MultiPhysicsMedium(Tidy3dBaseModel):
         Extend that mapping as additional cross-medium shim behaviour becomes
         necessary.
         """
+        # first check whether the attribute is already present
+        try:
+            return super().__getattr__(name)
+        except AttributeError:
+            pass
+
         IGNORED_ATTRIBUTES = ["__deepcopy__"]
         if name in IGNORED_ATTRIBUTES:
             return None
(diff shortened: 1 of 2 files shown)


tidy3d/components/material/multi_physics.py
```diff
@@ -128,6 +128,12 @@ def __getattr__(self, name: str):
         Extend that mapping as additional cross-medium shim behaviour becomes
         necessary.
         """
+        # first check whether the attribute is present (e.g. cached properties, dunders...)
+        try:
+            return super().__getattr__(name)
+        except AttributeError:
+            pass
+
         IGNORED_ATTRIBUTES = ["__deepcopy__"]
```
frederikschubertflex, line 137: This was my intention with the ignored attributes list but I guess its not feasible to cover all attributes that pydantic checks etc.  Maybe we should simply not delegate any private (i.e. `_*`) attributes?

frederikschubertflex:
That this was not caught by the unit tests should be motivation for us to improve on that :blush:

yaugenst-flex:
Noticed another issue: if `self.optical is None` (which is the default), then `DELEGATED_ATTRIBUTES[name] is None` and `getattr(None, name)` raises. Fixed that case too (well not "fixed" but the error message should be clear).

daquinteroflex (approved):
Thanks a lot Yannick!

frederikschubertflex (approved):

frederikschubertflex:
~~Can we add some tests for this?~~ I will add some tests!

tidy3d/components/material/multi_physics.py
yaugenst-flex: I think that runs into issues with `_eps_plot` then. But yeah conceptually that's basically the same thing right? When `MultiPhysicsMedium` gets more fleshed out this needs to be revisited anyways..

yaugenst-flex:
@frederikschubertflex oh sorry didnt see your edit, I added some tests now and force pushed! feel free to add more, I got a little bit confused. actually I still don't understand the root cause of the error I was seeing on my pydantic v2 branch. there, the `_cached_properties` lookup fails for `MultiPhysicsMedium`, but on `develop` it doesn't. I added a test to catch this but it currently passes on `develop`. still good to have probably. I did add tests for the other fix too though.

frederikschubertflex:
No, worries! I just figured that you have other things on your plate :) I will look into this in more depth later!

yaugenst-flex:
ok maybe just to conclude this discussion:
it looks like pydantic v1 handled `PrivateAttr` in `__getattribute__`, v2 does not, so it falls back to `__getattr__` which now gets caught by `MultiPhysicsMedium`'s override. another difference is that `__dir__` in v1 includes private attrs, and v2 omits them..

PR #2472, opened 2025-05-16
option for PMC boundary in mode solver

dbochkov-flexcompute:


tidy3d/components/mode/derivatives.py
```diff
@@ -6,13 +6,15 @@
...
-def make_dxf(dls, shape, pmc):
+def make_dxf(dls, shape, pmc_neg, pmc_pos):
     """Forward derivative in x."""
     Nx, Ny = shape
     if Nx == 1:
         return sp.csr_matrix((Ny, Ny))
     dxf = sp.csr_matrix(sp.diags([-1, 1], [0, 1], shape=(Nx, Nx)))
-    if not pmc:
+    if pmc_pos:
+        dxf[-1, -1] = 0.0
```
momchil-flex, line 16: So I don't think this is right, this will just impose PEC on the next-to-last grid boundary, right?

To impose PMC, this can be placed in the backward derivatives instead. However, it is still tricky: there it will impose it at the last grid *center*, not at the last grid *boundary*. I think it's not possible to impose it at the last boundary without expanding by one pixel, and then doing something like seetting a value of 2 for the derivative, like we do for pmc on the negative side (but there we can do without an extra pixel). I think this might have been the reason I only had it working on the negative side in the first place...

dbochkov-flexcompute: Ah, yeah, this is not quite correct. I believe what I did here is to impose PMC in the middle of the last cell: specifically here by setting this last coefficient to zero we effectively impose the normal derivative of tangential E field to be zero. I did forget to set tangential H to zero there (by setting `dxb[-1, -1] = 0.0` and `dyb[-1, -1] = 0.0`), but somehow it did not matter

momchil-flex: Ah, you are right. I think probably both of those decouple that last dual location from the rest of the simulation, so the only thing is if you don't do both, you might get spurious modes that are purely localized to that location? In fact I see we do both of those for PEC yet we still had to add [this](https://github.com/flexcompute/tidy3d/blob/4f267b80b351ff2a7bab681f744ad9d3d0a1c62b/tidy3d/components/mode/solver.py#L947-L951) to avoid such modes. And we don't do it for the right boundary, because there's no actual field there, but if the PMC right boundary is defined at the last dual location, then will probably be needed at least in the tensorial case where we solve for both E and H?

PR #2475, opened 2025-05-16
feat: autograd support for chained affine transforms: rotation, scaling, translation

rahul-flex:
**Key changes**

**Unified affine gradients** – one routine now returns gradients for the full linear block ( rotation and scaling) and for translation, replacing the old rotation‑only path.

**Correct normal handling** – face normals are transformed with the proper inverse‑transpose rule; mesh builder now accepts a linear transform plus translation.

**Per‑step gradients** – chain‑rule logic recovers a separate gradient for every transform in the sequence.

**Validation**

Finite‑difference versus autograd tests for scale, translate, size, rotate, and centre parameters.


yaugenst-flex:
Closing in favor of https://github.com/flexcompute/tidy3d/pull/2490

PR #2476, opened 2025-05-16
Added resistance calculation

marc-flex:
Added resistance calculation. Also, allowing ndarrays to be provided as voltage input.

Addresses https://github.com/flexcompute/tidy3d-core/issues/961

### DIFF
--- a/tidy3d/components/spice/sources/dc.py
+++ b/tidy3d/components/spice/sources/dc.py
@@ -19,12 +19,14 @@ Examples:
 
 """
 
-from typing import Optional, Union
+from typing import Optional
 
 import pydantic.v1 as pd
 
 from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.components.types import ArrayFloat1D
 from tidy3d.constants import AMP, VOLT
+from tidy3d.constants import inf as td_inf
 
 
 class DCVoltageSource(Tidy3dBaseModel):
@@ -45,11 +47,19 @@ class DCVoltageSource(Tidy3dBaseModel):
     """
 
     name: Optional[str]
-    voltage: Union[pd.FiniteFloat, list[pd.FiniteFloat]] = pd.Field(
+    voltage: ArrayFloat1D = pd.Field(
+        ...,
         title="Voltage",
         description="DC voltage usually used as source in 'VoltageBC' boundary conditions.",
+        units=VOLT,
     )
-    units: str = VOLT
+
+    @pd.validator("voltage")
+    def check_voltage(cls, val):
+        for v in val:
+            if v == td_inf:
+                raise ValueError(f"Voltages must be finite. Currently  voltage={val}.")
+        return val
 
 
 class DCCurrentSource(Tidy3dBaseModel):
--- a/tidy3d/components/tcad/data/sim_data.py
+++ b/tidy3d/components/tcad/data/sim_data.py
@@ -73,6 +73,14 @@ class DeviceCharacteristics(Tidy3dBaseModel):
         description="Device steady DC current-voltage relation for the device.",
     )
 
+    steady_dc_resistance_voltage: Optional[SteadyVoltageDataArray] = pd.Field(
+        None,
+        title="Small signal resistance",
+        description="Steady DC computation of the small signal resistance. This is computed "
+        "as the derivative of the current-voltage relation, delta(V)/delta(I) and the result "
+        "is given in Ohms. Note that in 2D the resistance is given in :math:`\\Omega \\mu`.",
+    )
+
 
 class HeatChargeSimulationData(AbstractSimulationData):
     """Stores results of a :class:`HeatChargeSimulation`.
--- a/tests/test_components/test_heat_charge.py
+++ b/tests/test_components/test_heat_charge.py
@@ -711,6 +711,9 @@ def test_heat_charge_bcs_validation(boundary_conditions):
     with pytest.raises(pd.ValidationError):
         td.CurrentBC(source=td.DCCurrentSource(current=td.inf))
 
+    with pytest.raises(pd.ValidationError):
+        td.VoltageBC(source=td.DCVoltageSource(voltage=np.array([td.inf, 0, 1])))
+
 
 def test_heat_charge_monitors_validation(monitors):
     """Checks for no name and negative size in monitors."""
@@ -790,6 +793,7 @@ def test_device_characteristics():
         steady_dc_hole_capacitance=capacitance,
         steady_dc_electron_capacitance=capacitance,
         steady_dc_current_voltage=current_voltage,
+        steady_dc_resistance_voltage=current_voltage,
     )
 
 
--- a/tidy3d/components/tcad/simulation/heat_charge.py
+++ b/tidy3d/components/tcad/simulation/heat_charge.py
@@ -434,11 +434,8 @@ class HeatChargeSimulation(AbstractSimulation):
             for bc in bounday_spec:
                 if isinstance(bc.condition, VoltageBC):
                     if isinstance(bc.condition.source, DCVoltageSource):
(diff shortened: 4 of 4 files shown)


tidy3d/components/spice/sources/dc.py
```diff
@@ -45,12 +46,20 @@ class DCVoltageSource(Tidy3dBaseModel):
     """
 
     name: Optional[str]
-    voltage: Union[pd.FiniteFloat, list[pd.FiniteFloat]] = pd.Field(
+    voltage: Union[pd.FiniteFloat, list[pd.FiniteFloat], np.ndarray] = pd.Field(
```
momchil-flex: Can you clarify this change? Also @yaugenst-flex might have something to say about it in view of pydantic v2.

tidy3d/components/tcad/data/sim_data.py
```diff
@@ -73,6 +73,14 @@ class DeviceCharacteristics(Tidy3dBaseModel):
         description="Device steady DC current-voltage relation for the device.",
     )
 
+    steady_dc_resistance_voltage: Optional[SteadyVoltageDataArray] = pd.Field(
+        None,
+        title="Small signal resistance",
+        description="Steady DC computation of the small signal resistance. This is computed "
+        "as the derivative of the current-voltage relation, delta(V)/delta(I) and the result "
+        "is given in Ohms. Note that in 2D the resistance is given in Ohm-um.",
```
momchil-flex: Do you mean Ohm/m?

tidy3d/components/spice/sources/dc.py
yaugenst-flex: Eh you can ignore my v2 PR, I'm rebasing and fixing merge conflicts anyways with any PR that is coming in at the moment 😄
That being said - we should not use raw `np.ndarray` in our types, we have dedicated array types for that in `types.py`. This looks like a case for `ArrayFloat1D`.

marc-flex: Sometimes users want to include voltages as a numpy array instead of passing a list. I'm introducing this to allow it.
Basically now users will be able to do `DCVoltageSource(voltage=np.linspace(-5, 1, 10))`

momchil-flex: Ah yeah, as @yaugenst-flex says you can use our ArrayFloat1D and it will work for this usage but is the way we do it internally for a few reasons (immutable, we check number of dimensions, etc.)

tidy3d/components/tcad/data/sim_data.py
marc-flex: nah, it's Ohm-um. 
Explanation; R = V/I and I = J * A -> R = V / (J * A) 
In 2D R = V / (J * L)  so you have 1 length dimension less in the denominator. You need to divide by your length to get Ohms

tidy3d/components/spice/sources/dc.py
yaugenst-flex: I'm not sure if this works correctly at the moment but in my v2 port I think this whole type annotation should just be `ArrayFloat1D`. Passing scalars or lists/tuples will just coerce these into arrays. This way we a guarantee for the type that we're working with _within_ the module.

marc-flex: This also means I don't need the validator, correct?

tidy3d/components/tcad/data/sim_data.py
momchil-flex: Ah ok thanks that makes sense!

@daquintero do we use latex in these descriptions? If not, I know that there's no good way to write this in plain text, but maybe `Ohm * um` (with the code formatting) is a bit better than Ohm-um?

tidy3d/components/spice/sources/dc.py
yaugenst-flex: I say keep it for now, but if you give me the green light then I'll change it in my v2 PR when I rebase

marc-flex: I was actually changing it myself.

One note, if I remove the validator tests fail because one could create the object with `[inf, 0, 1]`

marc-flex: OK, so new commit. This gets rid of `np.ndarray` and passes the tests I already had. 
Let me know if that's OK

yaugenst-flex: Yes I'm not sure whether this works correctly for the current state of things.

yaugenst-flex: > OK, so new commit. This gets rid of `np.ndarray` and passes the tests I already had. Let me know if that's OK

Perfect! Thanks

marc-flex: what would then?

marc-flex: One related question.  So now voltage can be defined by either a float or an array which could be anything, right? So in the a different class, can I do something like this?  `isinstance(myVoltage, ArrayFloat1D)` and that would work for any array-like data type?

yaugenst-flex: Yes that would work. Ideally though we should just be dealing with a well-known validated type in the module internals so that such checks are not necessary. Or is there fundamentally different logic required for handling scalars and arrays here?

marc-flex: Yes, I have a case where I actually need to make sure there at least an array with more than 1 element

yaugenst-flex: But that would be guaranteed if the input is always coerced into an array right?

marc-flex: No. I need it to have more than 1 element

yaugenst-flex: The point is that you would never need an `isinstance` check, you can just check the length of the array directly because it is safe to assume that it's an array.

marc-flex: ahh right, so you mean I need to remove `voltage: Union[pd.FiniteFloat, ArrayFloat1D] = pd.Field(` and instead have `voltage: ArrayFloat1D = pd.Field(`. Right?

yaugenst-flex: Correct yeah. Basically the `Union` is only necessary if you _really_ need to be able to distinguish scalar from array-like _types_ within the class internals.

marc-flex: unfortunately this requires a major revision because I'm counting on `isinstance(voltages, float)` in many places

yaugenst-flex: Yes that's fine you don't need to change it, I just need to know that I'm allowed to 😄 I'm overhauling a lot of this machinery in https://github.com/flexcompute/tidy3d/pull/2433

marc-flex: That doesn't look at the back-end, does it?

yaugenst-flex: Eventually it will but right now it's only python client.

tidy3d/components/tcad/data/sim_data.py
daquinteroflex: Oh I drafted a response but never sent it.

So it depends how it wants to be rendered. For the docs, it can just follow rst math formatting `:math:`\Omega \mu m``, and it should render fine. See `coeffs` parameter in https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.Drude.html . If we want it to render nicely beyond docs, like in the terminal help, that's a differnet game and we'd need to eventually support this maybe with a docs schema

marc-flex:
So all comments have been tackled. Do have a last look so we can merge

@yaugenst-flex I dealt with the changes related using array1D also in the backend,

momchil-flex (approved):

yaugenst (approved):

PR #2477, opened 2025-05-16
Remove gdspy

### LINKED ISSUE #2465: 🕐 Deprecate `gdspy`
It is a pretty big pain now it's obsolete, especially installing in recent ubuntu.

Remove from pyproject.toml and any code references,  tests, and docs.

frederikschubertflex:
This PR addresses https://github.com/flexcompute/tidy3d/issues/2465 and removes the dependency and all references to `gdspy`.


yaugenst-flex (commented):
Sweet! Haven't reviewed the code in depth yet but just wondering more generally whether we should keep the old methods around for the time being and let users know that this is not supported anymore and redirect to gdstk? I guess we should probably have had some transition period too but ¯\\_(ツ)_/¯

frederikschubertflex:
I was wondering this as well but understood the issue as removing `gdspy`. I can also change this to only include warnings for the users for now. Then we can remove this later (maybe in 2.9)?

tylerflex (approved):

yaugenst-flex (approved):
LGTM! Thanks @frederikschubertflex

daquinteroflex (approved):
Thanks Frederik! The time has come.

PR #2478, opened 2025-05-16
Improve mode solver handling of 1D simulations

### LINKED ISSUE #2471: Improve the user experience for 1D mode simulations
# Current
We have the following issues with 1D Mode simulations:
1. We need a two-dimensional mode plane to pass validation. In GUI, it seems a bit weird to create a 2D mode plane for a 1D mode simulation domain.
2. There are validation errors related to using default boundary settings (PML) in zero-dimensional simulation domain (FDTD) directions. While for Python users, that is only a minor inconvenience, since they are setting a Mode simulation and concerned about mode plane boundary conditions, in the GUI, that is critical because we don't have FDTD boundary settings in the Mode workbench.

# Requirement
1. Support a 0-size mode plane on directions with a simulation domain size of 0 in Mode simulations.
2. Avoid validation errors related to using PML in zero-dimensional simulation domain directions in Mode simulations.  *Potential solution: Use periodic boundary conditions automatically for 0-size dimensions, regardless of the user’s boundary condition settings, and provide a warning/info message in the solver log in Mode simulations.*

caseyflex:
Addresses this issue:
https://github.com/flexcompute/tidy3d/issues/2471

Changes:

- Instead of erroring when a simulation has 0 thickness dimension and the boundary is not Periodic or Bloch, the boundary is automatically replaced with Periodic (giving a warning)
- The mode plane can be inferred in mode solver and mode simulation for 1D and 2D simulations (previously, it was inferred in 2d mode simulations only)

## Greptile Summary

Enhances mode solver functionality for 1D/2D simulations by improving boundary condition handling and plane inference capabilities.

- Automatically converts non-Periodic/Bloch boundaries to Periodic (with warning) for zero-thickness dimensions in `components/simulation.py`
- Added automatic mode plane inference for both 1D and 2D simulations in `components/mode/mode_solver.py`
- Made 'plane' parameter optional in mode solver, letting it be inferred from simulation geometry
- Added comprehensive test coverage in `tests/test_components/test_mode.py` for new 1D/2D simulation handling
- Updated validation logic in `components/mode/simulation.py` to support these lower-dimensional cases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Zero-size dimensions automatically receive periodic boundary conditions, raising a warning where previously they would error.
+- Plane used for mode solving can be automatically inferred for 1D and 2D simulations.
+
 ## [2.8.4] - 2025-05-15
 
 ### Added
--- a/tests/test_components/test_simulation.py
+++ b/tests/test_components/test_simulation.py
@@ -560,7 +560,7 @@ def test_validate_zero_dim_boundaries():
         pol_angle=0.0,
     )
 
-    with pytest.raises(pydantic.ValidationError):
+    with AssertLogLevel("WARNING", contains_str="Periodic"):
         td.Simulation(
             size=(1, 1, 0),
             run_time=1e-12,
--- a/tidy3d/components/mode/mode_solver.py
+++ b/tidy3d/components/mode/mode_solver.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from functools import wraps
 from math import isclose
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pydantic.v1 as pydantic
@@ -14,7 +14,7 @@ import xarray as xr
 from matplotlib.collections import PatchCollection
 from matplotlib.patches import Rectangle
 
-from ...constants import C_0
+from ...constants import C_0, inf
 from ...exceptions import SetupError, ValidationError
 from ...log import log
 from ..base import Tidy3dBaseModel, cached_property, skip_if_fields_missing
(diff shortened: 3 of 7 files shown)


e-g-melo (approved):
Hi @caseyflex, thanks a lot for this work!

1. I wonder if we could suppress this warning in `ModeSimulation` because the FDTD boundary conditions are not directly exposed in its definition. For the old `ModeSolver` that would be fine, though.

image

2. The plotting functions would need some adjustments to suppress these warnings. This is not directly related to the original motivation for this PR, but it would be nice for Python users. 

image

3. It would be nice to automatically detect the 1D domain and change the `.plot_field()` function into a line plot. Currently, it throws an error.

```py
ERROR: Data after selection has 1 coordinates (['y']), must be 2   
             spatial coordinates for plotting on plane. Please add keyword      
             arguments to `plot_field()` to select out the other coords. 
```

tidy3d/components/mode/mode_solver.py
```diff
@@ -188,9 +188,45 @@ def _convert_to_simulation(cls, val):
...
+        elif num_zero_dims == 2:
+            new_size[new_size.index(0.0)] = inf
+            nonzero_dims.pop(new_size.index(0.0))
+            log.warning(
+                "The simulation geometry is 1D; the mode solver 'plane' "
+                "is taken by default as the span of the nonzero-size dimension "
+                "together with the first zero-size dimension "
+                f"(the {nonzero_dims}-plane)."
+            )
+        sim_geom = sim_geom.updated_copy(size=new_size)
```
yaugenst-flex, line 220: Not sure if this is validated elsewhere (not for `Box` at least..) but `num_zero_dims == 3` falls through to the 2d branch here.

tests/test_components/test_simulation.py
```diff
@@ -560,7 +560,7 @@ def test_validate_zero_dim_boundaries():
         pol_angle=0.0,
     )
 
-    with pytest.raises(pydantic.ValidationError):
```
yaugenst-flex, line 563: This works but it might be good to also assert whether the boundary actually became periodic.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Zero-size dimensions automatically receive periodic boundary conditions, raising a warning where previously they would error.
+- Plane used for mode solving can be automatically inferred for 1D and 2D simulations.
```
yaugenst-flex, line 12: Might be good to explicitly state that `plane` is now an optional argument since it technically is a breaking change in the public API.

yaugenst-flex (commented):
Thanks @caseyflex LGTM, only a couple of minor comments/nits.

Issue #2286, opened 2025-02-28
Variable Height Structure


tylerflex:
Ability to create a `TriangleMesh` representing geometry with a variable height described by a function returning thickness as a function of (x,y).

I believe we have this capabability, but would be nice to support officially as a @classmethod constructor of `TriangleMesh`, for example.

@tomflexcompute @yaugenst-flex @dbochkov-flexcompute

tomflexcompute:
@dbochkov-flexcompute just to confirm you haven't added the function to the frontend already?

dbochkov-flexcompute:
@tomflexcompute yes, that's correct, I haven't added this into the python API yet. Happy to take care of it if needed or it could be a nice and easy first-time task for newly joining members.

momchil-flex:
@dbochkov-flexcompute could you just comment the code here for now?

dbochkov-flexcompute:
[height_function.ipynb.zip](https://github.com/user-attachments/files/19251284/height_function.ipynb.zip)

PR #2479, opened 2025-05-17
Adding function to generate trimesh geometry from height p…

tomflexcompute:
Adding functions to generate trimesh geometry from a height profile to address https://github.com/flexcompute/tidy3d/issues/2286 since we have a recent use case that could benefit from this feature. 

It seems like in the current implementation if the height is negative, it will not generate the surface @dbochkov-flexcompute . Not sure what the ideal behavior should be, say the height expression is a sine function and the base is at 0. 

Also do we need a test for it?

FYI @FilipeFcp .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `thickness` parameter to `LossyMetalMedium` for computing surface impedance of a thin conductor.
 - `priority` field in `Structure` and `MeshOverrideStructure` for setting the behavior in structure overlapping region. When its value is `None`, the priority is automatically determined based on the material property and simulation's `structure_priority_mode`.
 - Automatically apply `matplotlib` styles when importing `tidy3d` which can be reverted via the `td.restore_matplotlib_rcparams()` function.
+- Added `TriangleMesh.from_height_expression` class method to create a mesh from an analytical height function defined on a 2D grid and `TriangleMesh.from_height_grid` class method to create a mesh from height values sampled on a 2D grid.
 
 ### Fixed
 - Fixed bug in broadband adjoint source creation when forward simulation had a pulse amplitude greater than 1 or a nonzero pulse phase.
--- a/tests/test_components/test_geometry.py
+++ b/tests/test_components/test_geometry.py
@@ -1163,3 +1163,86 @@ def test_triangulation_with_collinear_vertices():
     xr = np.linspace(0, 1, 6)
     a = np.array([[x, -0.5] for x in xr] + [[x, 0.5] for x in xr[::-1]])
     assert len(td.components.geometry.triangulation.triangulate(a)) == 10
+
+
+def test_triangle_mesh_from_height():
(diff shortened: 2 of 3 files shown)


dbochkov-flexcompute:
> It seems like in the current implementation if the height is negative, it will not generate the surface @dbochkov-flexcompute . Not sure what the ideal behavior should be, say the height expression is a sine function and the base is at 0.

Yeah, I think we need to add a check that all `height` values are `>= 0`. And to handle cases when it goes exactly to zero in some places we can just clip height values by a small number from below
```
flat_height = np.clip(flat_height, fp_eps * np.max(flat_height), inf)
```

caseyflex (changes requested):
Looks good, this will be useful! I do think we should add a couple unit tests. Also maybe `from_height_function` is clearer than `from_height_expression`.

tomflexcompute:
Thanks @dbochkov-flexcompute @caseyflex . Not very familiar with our test suite. Is it just adding a test function to `test_geometry.py`? Can revisit it a bit later after some time-sensitive tasks but if you guys think it's relatively easy feel free to add something directly.

caseyflex:
> Not very familiar with our test suite. Is it just adding a test function to `test_geometry.py`?

Yeah, just something simple -- if you have a simple vectorized function in mind that these functions would operate on to produce a trimesh, then you can include a `test_from_height_function` or similar in `test_geometry.py` that at least runs through this code once with a working example, and possibly checks anything obvious you can think of in terms of properties of the resulting trimesh. Ideally you would also add some lines in the test to check your edge-case handling, aiming for close to 100% line coverage...

Getting good coverage in the tests can be very helpful in preventing easy but annoying bugs from popping up.

an example:

```
def test_intersections_plane_inf():
    a = (
        td.Cylinder(radius=3.2, center=(0.45, 9, 0), length=td.inf)
        + td.Box(center=(0, 0, 0), size=(0.9, 24, td.inf))
        + td.Box(center=(0, 0, 0), size=(7.3, 18, td.inf))
    )
    b = td.Cylinder(radius=2.9, center=(-0.45, 9, 0), length=td.inf)
    c = a - b
    assert len(c.intersections_plane(y=0)) == 1

def test_center_not_inf_validate():
    with pytest.raises(pydantic.ValidationError):
        _ = td.Box(center=(td.inf, 0, 0))
    with pytest.raises(pydantic.ValidationError):
        _ = td.Box(center=(-td.inf, 0, 0))
```

but you could combine it all in one test function

caseyflex (approved):
Thanks looks great! You can rebase and squash together all the commits into one, then I think it is good to merge

tomflexcompute:
Thanks @caseyflex ! I think @dbochkov-flexcompute is not feeling well at the moment? Feel free to take your time on the review there is no rush.

dbochkov-flexcompute (approved):
looks great, thank you!

tomflexcompute:
Some tests seem to not go through after rebase. Is it fine actually?

caseyflex:
> Some tests seem to not go through after rebase. Is it fine actually?

I’ll just rerun them, let’s see if it was just a fluke

caseyflex:
@tomflexcompute looks good now!

PR #2485, opened 2025-05-19
fix: NedeljkovicSorefMashanovich serialization

### LINKED ISSUE #2483: Fix `NedeljkovicSorefMashanovich` serialization
The [`NedeljkovicSorefMashanovich`](https://github.com/flexcompute/tidy3d/blob/d079b2e0eb6ee92d8e0d5aa4fc347a11338fd1ee/tidy3d/components/parameter_perturbation.py#L1243) perturbation model currently is not JSON serializable as it uses a `xr.Dataset` field type. The following errors:
```py
m = NedeljkovicSorefMashanovich(ref_freq=1e14)
print(m.json())
```

yaugenst-flex:
`NedeljkovicSorefMashanovich` currently uses `xr.Dataset` for which we have no serializer. This PR introduces a `PerturbationCoefficientDataArray` and uses that as the type for `perturb_coeffs`. This is technically a breaking change because `perturb_coeffs` are now a `DataArray` instead of a `Dataset`. It seems to me like that change is _ok_ though and I fixed the places in the code where that caused problems.
Also added a test that currently fails on `develop` and is fixed by this PR.


yaugenst-flex:
Note that this only works because the coefficients represent a regular array. If we at any point need to start dealing with ragged arrays then probably there will be no way around writing a serializer for `xr.Dataset`.

tidy3d/components/parameter_perturbation.py
```diff
@@ -1252,277 +1258,261 @@ class NedeljkovicSorefMashanovich(AbstractDeltaModel):
...
+            coords={
+                "wvl": np.array([1.3, 1.55] + list(np.arange(2, 14.5, 0.5))),
+                "coeff": ["a", "b", "c", "d", "p", "q", "r", "s"],
+            },
+            name="perturb_coeffs",
+        )
     )
 
     ref_freq: pd.NonNegativeFloat = pd.Field(
-        ...,
```
marc-flex, line 1518: Aren't the dots necessary for the field to be required?

marc-flex (approved):
Thanks @yaugenst-flex

tidy3d/components/parameter_perturbation.py
yaugenst: No. A field that has no default value is a required field by definition, even in pydantic v1.

yaugenst-flex: No. A field that has no default value is a required field by definition, even in pydantic v1.

PR #2486, opened 2025-05-19
Upgrade ruff 0.5.5 -> 0.10.11 & review rules

### LINKED ISSUE #2481: Ruff upgrade & additional lint rules
Our ruff version is lagging significantly behind, so it's due for an update. We should also review existing linting rules.
There are also at least two more rules that we should enable:

- [`TID252`](https://docs.astral.sh/ruff/rules/relative-imports/): Only allow relative imports from siblings. Maybe it's just me but I find the following very hard to deal with:
  ```py
  from ....exceptions import AdjointError
  from ...core.s3utils import download_file, upload_file
  from ...core.types import PayType
  ```
  We'll exclude this rule in tests.
- [`UP006`](https://docs.astral.sh/ruff/rules/non-pep585-annotation/): Replace `typing` generics with standard library types when available. We have dropped support for Python 3.8, so it's time to do that.

yaugenst-flex:
The most controversial rule change in here is _probably_ the addition of [`C408`](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/), which enforces `{"a": a, "b": b}` over `dict(a=a, b=b)`.
Following reasons why I think it's a good change:
 1. **Speed**: `dict()` is 2x slower than a literal because it needs to do a global namespace lookup.
 2. **Edge cases**: `dict()` only works if keys are valid identifiers.
 3. **Clarity**: Personally I find it easier to identify dicts at a glance when they are literals. I know this is subjective...

I kept all commits in here grouped by the rule and the code changes associated with it, so it's easy to inspect what a specific rule does and if we might want to revert it. I'll keep this PR open and merge _before_ https://github.com/flexcompute/tidy3d/pull/2433 and resolve any conflicts that arise there.

There are many lines of changes but really only the rule additions need reviewing and whether you agree with them or not.

**NOTE:** I _really_ want to enable [`B008`](https://docs.astral.sh/ruff/rules/function-call-in-default-argument/) because function calls in default arguments are a significant anti-pattern and they are so common in our codebase that they do affect import time significantly. However, [I had to revert](https://github.com/flexcompute/tidy3d/pull/2486/commits/62356ee2f854cc4005fc7217596545ab24b48651) the addition because my fix broke `tests/test_components/test_layerrefinement.py::test_grid_spec_with_layers` because it seems like many of the `grid_spec` functions have undocumented behavior when it comes to passing `None` for `corner_finder` and `corner_refinement`. If you could have a look @weiliangjin2021 that would be great.

---

### Net Changes
* **Enabled**: `C4*`, `NPY*`, `RUF*`, `ISC*`, `PIE*`, `RSE*`, `TID*`, `PLE*`, `PLC*`
* **Disabled**: legacy `C*`, standalone `NPY201`
* **Newly ignored**: `RUF001-003`, `RUF012`, `RUF015`, `NPY002`
* **Now enforced**: `C408`, `B904`, `B028`, `UP006`, `UP038`, `UP035`
* **Restructured** the `[tool.ruff]` section in `pyproject.toml`:
    - `select` -> `extend-select` to keep default rules, similarly for `ignore`
    - Moved test-specific ruff config from `tests/ruff.toml` to `[tool.ruff.lint.per-file-ignores]` in `pyproject.toml`

### Additions
| Area | What | Why |
|------|------------|----------------|
| **Rule families** | `C4`, `NPY`, `RUF`, `ISC`, `PIE`, `RSE`, `TID`, `PLE`, `PLC` in **`extend-select`** | Modern comprehensions, full NumPy checks, Ruff built-ins, pylint parity, etc. |
| **isort** | `required-imports = ["from __future__ import annotations"]` | Auto-adds future annotations → no longer need to ignore `UP006`. |
| **Ignore list** | `RUF001-003`, `RUF012`, `RUF015`, `NPY002` via **`extend-ignore`** | Low-value or transitional warnings. |
| **Per-file ignores** | Relaxed rules for `tests/**/*` (`B015`, `E402`, `TID252`, …) | Tests don't need to be as strict. |

### Removed / No-Longer-Ignored
| Removed from config | Effect |
|---------------------|--------|
| `typing-modules` hack | Ruff ≥ 0.4 now infers `Literal` types; custom list deleted. |
| Old **`ignore`** items: `C408`, `B904`, `B028`, `UP006`, `UP038`, `UP035` | These rules are **enforced** again. `UP006` safe thanks to required imports. |
| Obsolete selections: `C` prefix, lone `NPY201` | Replaced by `C4*` and full `NPY*` suites. |


frederikschubertflex (approved):

weiliangjin2021:
> However, [I had to revert](https://github.com/flexcompute/tidy3d/pull/2486/commits/62356ee2f854cc4005fc7217596545ab24b48651) the addition because my fix broke tests/test_components/test_layerrefinement.py::test_grid_spec_with_layers because it seems like many of the grid_spec functions have undocumented behavior when it comes to passing None for corner_finder and corner_refinement. If you could have a look @weiliangjin2021 that would be great.

Currently, `None` means disabling this feature. E.g. `corner_refinement=None`: don't apply refinement around corners. The default value, on the other hand, indicates a good initial choice. E.g. `corner_finder=CornerFinderSpec()` means you can start with the default value in `CornerFinderSpec` if you want the corner finder feature, and it's very likely that you don't need to tune the default values. Any better options you have in mind?

yaugenst-flex:
Thanks @weiliangjin2021 for clarifying. I think my main source of confusion was that this behavior doesn't seem to be documented, at least I wasn't able to infer it [from the field descriptions](https://github.com/flexcompute/tidy3d/blob/3a9ec0622447b1b0167c6e88e6929e3f4007c8e8/tidy3d/components/grid/grid_spec.py#L1016-L1035). The defaults for the fields can be solved by using default factories, which I am changing across the board in https://github.com/flexcompute/tidy3d/pull/2433. For the classmethods it seems a bit more tricky. One option would be to change the field to `Union[Optional[CornerFinderSpec], Literal[False]]`, where `False` explicitly turns it off and `None` would mean default, which I think is more explicit. If not, then an alternative approach might be to introduce a sentinel (e.g. `pydantic.Undefined`) in the class constructors. That would look something like:
```py
@classmethod
def from_layer_bounds(cls, ..., corner_finder: Union[CornerFinderSpec, None, object] = Undefined):
    if corner_finder is Undefined:
        corner_finder = ...  # handle default
```

momchil-flex (approved):
Sounds good to me. Note that the backend uses the same ruff definitions like the frontend so will need an update too...

> * **Now enforced**: `C901`, `C408`, `B904`, `B028`, `UP006`, `UP038`, `UP035`

Seems like C901 is still ignored?

yaugenst-flex:
> Seems like C901 is still ignored?

Oops you're right. I disabled `C` in favor of the `C4` group (which is the actual group..). It's good this way though don't think we want `C901` 😅 

Edited the PR description.

weiliangjin2021:
> Thanks @weiliangjin2021 for clarifying. I think my main source of confusion was that this behavior doesn't seem to be documented, at least I wasn't able to infer it [from the field descriptions](https://github.com/flexcompute/tidy3d/blob/3a9ec0622447b1b0167c6e88e6929e3f4007c8e8/tidy3d/components/grid/grid_spec.py#L1016-L1035).

Actually it's in the description: https://github.com/flexcompute/tidy3d/blob/3a9ec0622447b1b0167c6e88e6929e3f4007c8e8/tidy3d/components/grid/grid_spec.py#L1023-L1035. Maybe not explicit enough, but it does indicate that snapping and refinement are only enabled if corner_finder is not None.

> The defaults for the fields can be solved by using default factories, which I am changing across the board in #2433. For the classmethods it seems a bit more tricky. One option would be to change the field to `Union[Optional[CornerFinderSpec], Literal[False]]`, where `False` explicitly turns it off and `None` would mean default, which I think is more explicit. If not, then an alternative approach might be to introduce a sentinel (e.g. `pydantic.Undefined`) in the class constructors. That would look something like:

How do we deal with back-compatibility, as right now `None` means turning it off.

yaugenst-flex:
> How do we deal with back-compatibility, as right now None means turning it off.

Yes that's fair, in that case it's probably be to go with the sentinel approach in the classmethods, that should be fully backwards-compatible.

yaugenst-flex:
@weiliangjin2021 implemented the change in 64f7fbfa. tests are passing, so just fyi

daquinteroflex (approved):
Nice, gone through each rule change and do feel things will be clearer and more standardised by enfocing these. Thanks Yannick.

tylerflex:
RIP `dict()`. sad!

Issue #2179, opened 2025-01-17
Support `MultiPhysicsMedium` in `td.Simulation`


daquinteroflex:
Basically make this not break:
```
td.Simulation(
    medium=td.MultiPhysicsMedium(...)
    structures=[
         td.Structure(
              medium=td.MultiPhysicsMedium(...)
         )
    ]
)
```

PR #2487, opened 2025-05-19
fix: add delegation for necessary properties of a Simulation with MultiPhysicsMedium structures

frederikschubertflex:
This is the fix for https://github.com/flexcompute/tidy3d/issues/2179. I added a test that actually accesses the missing property `_has_incompatibilities` and added the remaining delegations.


daquinteroflex (approved):
Looks good, and you're increasing the test coverage in a different PR right? Just wondering how to cover tests the entire set of cases that can accept this medium.

frederikschubertflex:
Good point! ~~It is a larger change but the most straight forward way for this would be to parametrize all tests that use some Medium to also use the `MultiPhysicsMedium` wrapper. @yaugenst-flex should I wait with this until the pydantic v2 refactor is merged?~~ See comment below.

frederikschubertflex:
I found a simpler solution by just adding the `MultiPhysicsMedium` structure to the `FULL_SIM` simulation that is used in many different tests.

frederikschubertflex:
Solves https://github.com/flexcompute/tidy3d/issues/2179

PR #2488, opened 2025-05-19
only warning on VisualizationSpec color validation when matplotlib is not installed

groberts-flex:
I'm not quite sure the best way to test this in CI. I've tested locally to make sure the warning pops up if I fake that the import couldn't happen, but I couldn't figure out a way to make python think matplotlib doesn't exist in the unit tests. Open to ideas here!

### DIFF
--- a/tests/test_components/test_viz.py
+++ b/tests/test_components/test_viz.py
@@ -10,6 +10,8 @@ from tidy3d.components.viz import Polygon, set_default_labels_and_title
 from tidy3d.constants import inf
 from tidy3d.exceptions import Tidy3dKeyError
 
+from ..utils import AssertLogLevel
+
 pytestmark = pytest.mark.usefixtures("mpl_config_noninteractive")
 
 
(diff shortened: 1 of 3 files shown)


yaugenst-flex (commented):
Thanks @groberts-flex I think the approach looks fine. For testing, the simplest thing would probably be to mock `MATPLOTLIB_IMPORTED` in the test function:
```py
monkeypatch.setattr(viz, "MATPLOTLIB_IMPORTED", False)
```

A more rigorous check would be to mock out `matplotlib` from `sys.modules` entirely and re-import `viz`. There is a similar approach [here](https://github.com/flexcompute/tidy3d/blob/3a9ec0622447b1b0167c6e88e6929e3f4007c8e8/tests/test_package/test_compat.py#L22-L31).

tests/test_components/test_viz.py
```diff
@@ -218,6 +218,24 @@ def plot_with_multi_viz_spec(alphas, facecolors, edgecolors, rng, use_viz_spec=T
...
+    viz_spec = td.VisualizationSpec(facecolor="green")
+
+    warning_list = td.log.captured_warnings()
+
+    EXPECTED_WARNING_MSG_PIECE = (
+        "matplotlib was not successfully imported, but is required to validate colors"
+    )
+    warnings_for_matplotlib = [w for w in warning_list if EXPECTED_WARNING_MSG_PIECE in w["msg"]]
+
+    assert len(warnings_for_matplotlib) > 0, "Expected warning for mocked matplotlib import failure"
```
yaugenst-flex: why not `AssertLogLevel`? https://github.com/flexcompute/tidy3d/blob/3a9ec0622447b1b0167c6e88e6929e3f4007c8e8/tests/test_package/test_log.py#L278-L279

groberts-flex: exactly what I was looking for with this, thank you! changed!

yaugenst-flex (approved):

PR #2494, opened 2025-05-20
Add local subpixel integration

caseyflex:
Lays the foundation for local subpixel integration:

- Adds `config.use_local_subpixel`, default to `None`
- Lays foundation for local subpixel integration once released, but shouldn't change any behavior until then.
- A few code improvements in relative mode solver

### DIFF
--- a/tidy3d/components/mode/mode_solver.py
+++ b/tidy3d/components/mode/mode_solver.py
@@ -62,6 +62,7 @@ from tidy3d.components.viz import make_ax, plot_params_pml
 from tidy3d.constants import C_0
 from tidy3d.exceptions import SetupError, ValidationError
 from tidy3d.log import log
+from tidy3d.packaging import supports_local_subpixel, tidy3d_extras
 
 # Importing the local solver may not work if e.g. scipy is not installed
 IMPORT_ERROR_MSG = """Could not import local solver, 'ModeSolver' objects can still be constructed
@@ -316,7 +317,7 @@ class ModeSolver(Tidy3dBaseModel):
             _, plane_inds = Box.pop_axis([0, 1, 2], normal_axis)
             for dim, sym in enumerate(solver_symmetry):
                 if sym != 0:
-                    span_inds[plane_inds[dim], 0] += np.diff(span_inds[plane_inds[dim]]) // 2
+                    span_inds[plane_inds[dim], 0] += np.diff(span_inds[plane_inds[dim]])[0] // 2
 
         return simulation._subgrid(span_inds=span_inds)
 
@@ -1355,18 +1356,21 @@ class ModeSolver(Tidy3dBaseModel):
 
     def _solver_eps(self, freq: float) -> ArrayComplex4D:
         """Diagonal permittivity in the shape needed by solver, with normal axis rotated to z."""
-
         # Get diagonal epsilon components in the plane
         eps_tensor = self._get_epsilon(freq)
         # tranformation
         return self._tensorial_material_profile_modal_plane_tranform(eps_tensor, self.normal_axis)
 
+    @supports_local_subpixel
     def _solve_all_freqs(
         self,
         coords: tuple[ArrayFloat1D, ArrayFloat1D],
         symmetry: tuple[Symmetry, Symmetry],
     ) -> tuple[list[float], list[dict[str, ArrayComplex4D]], list[EpsSpecType]]:
         """Call the mode solver at all requested frequencies."""
+        if tidy3d_extras["use_local_subpixel"]:
+            subpixel_ms = tidy3d_extras["mod"].SubpixelModeSolver.from_mode_solver(self)
+            return subpixel_ms._solve_all_freqs(coords=coords, symmetry=symmetry)
 
         fields = []
         n_complex = []
@@ -1380,6 +1384,7 @@ class ModeSolver(Tidy3dBaseModel):
             eps_spec.append(eps_spec_freq)
         return n_complex, fields, eps_spec
 
+    @supports_local_subpixel
     def _solve_all_freqs_relative(
         self,
         coords: tuple[ArrayFloat1D, ArrayFloat1D],
@@ -1387,6 +1392,11 @@ class ModeSolver(Tidy3dBaseModel):
         basis_fields: list[dict[str, ArrayComplex4D]],
     ) -> tuple[list[float], list[dict[str, ArrayComplex4D]], list[EpsSpecType]]:
         """Call the mode solver at all requested frequencies."""
+        if tidy3d_extras["use_local_subpixel"]:
+            subpixel_ms = tidy3d_extras["mod"].SubpixelModeSolver.from_mode_solver(self)
+            return subpixel_ms._solve_all_freqs_relative(
+                coords=coords, symmetry=symmetry, basis_fields=basis_fields
+            )
 
         fields = []
         n_complex = []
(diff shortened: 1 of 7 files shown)


tidy3d/packaging.py
```diff
@@ -20,6 +21,8 @@
     "numpy_to_vtk": None,
 }
 
+tidy3d_extras = {"mod": None, "use_local_subpixel": False}
```
yaugenst-flex, line 26: Mostly just making a note that we _really_ need a better way to do config...

tidy3d/components/mode/mode_solver.py
```diff
@@ -1356,18 +1357,21 @@ def _diagonal_material_profile_modal_plane_tranform(
...
+    @supports_local_subpixel
     def _solve_all_freqs(
         self,
         coords: Tuple[ArrayFloat1D, ArrayFloat1D],
         symmetry: Tuple[Symmetry, Symmetry],
     ) -> Tuple[List[float], List[Dict[str, ArrayComplex4D]], List[EpsSpecType]]:
         """Call the mode solver at all requested frequencies."""
+        if tidy3d_extras["use_local_subpixel"]:
+            subpixel_ms = tidy3d_extras["mod"].SubpixelModeSolver.from_mode_solver(self)
+            return subpixel_ms._solve_all_freqs(coords=coords, symmetry=symmetry)
```
yaugenst-flex, line 1373: This pattern is repeated 3 times, and the number might increase over time. Would it make sense to centralize this dispatch logic into something like a `SubpixelModeSolverProxy` class? Maybe it's a bit overkill, not sure.

tidy3d/components/mode/simulation.py
```diff
@@ -251,6 +251,9 @@ def run_local(self):
         """Run locally."""
         from .data.sim_data import ModeSimulationData
 
+        # repeat the calculation every time, in case use_local_subpixel changed
+        self._cached_properties.pop("_mode_solver", None)
```
yaugenst-flex: In principle, other cached properties might develop a dependency on this flag in the future. So it might make sense to instead introduce something like a `AbstractYeeGridSimulation._invalidate_subpixel_caches()`?

tidy3d/packaging.py
```diff
@@ -173,3 +176,33 @@ def get_numpy_major_version(module=np):
     major_version = int(module_version.split(".")[0])
 
     return major_version
+
+
+def supports_local_subpixel(fn):
+    """When decorating a method, checks that 'tidy3d-extras' is availabe,
```
yaugenst-flex: `availabe` -> `available`

yaugenst-flex (commented):
Thanks @caseyflex I think the direction looks good and it's a good workable initial implementation. Mostly left some nits and high-level comments to discuss.

tidy3d/packaging.py
caseyflex: I can add a comment, but what do you have in mind here?

tidy3d/components/mode/mode_solver.py
caseyflex: hmm, it's used twice for `SubpixelModeSolver` and once for `SubpixelSimulation`. I don't anticipate this number growing that much actually.

Also I like the suggestion but I'm not sure I fully understand it. We have the decorator `supports_local_subpixel`; maybe an option would be to extend that to actually implement this pattern as well? But it would need some `isinstance` checking to handle `ModeSolver` and `Simulation` separately, and it would also need some logic for checking which function it is decorating and calling the corresponding function in the corresponding `Subpixel` class.

tidy3d/components/mode/simulation.py
caseyflex: This is an interesting idea, again not quite sure of the details. One thing to note is that `run_local` is very different from other properties and functions -- it is the only one that actually runs the solver. In `simulation.py`, `epsilon_on_grid` is similar; it doesn't run the whole solver but it does now run the subpixel averaging. But you're right that the line can start to blur; in fact maybe we should separate `epsilon_on_grid` from something like `run_subpixel_averaging` to make it clear that one operation may be heavier than the other. (although epsilon_on_grid can already be quite heavy)

Also note that we don't just invalidate this if `use_local_subpixel` changed, we actually invalidate it every time, essentially just requiring the `_mode_solver` (which caches `data` and `data_raw`) to run again every time `run_local` is called. If the `ModeSolver` is used directly, these caches will not be invalidated every time `use_local_subpixel` changes; if the user wants to use that old API and play with `use_local_subpixel`, they will have to invalidate the caches manually. In my mind this is actually one of the faults of the old `ModeSolver` interface -- it computes the data and stores it as a cached property, while `ModeSimulation` is just a simulation type with corresponding data type `ModeSimulationData`, and just one explicit function `run_local` that computes one from the other (and should always bypass caches).

tidy3d/components/mode/mode_solver.py
yaugenst-flex: Oh you're right I missed that one was for `SubpixelSimulation`... yeah then I think my suggestion would just be overengineering it. Putting it in the decorator is interesting too but probably falls into the same trap. Seems fine as-is.

tidy3d/packaging.py
yaugenst-flex: Oh sorry maybe I wasn't clear, I meant that this along with many other options we currently set globally should be configurable, but it's out of scope here.

tidy3d/components/mode/simulation.py
yaugenst-flex: Thanks for the additional context, I think this all makes sense. I'm mainly worried about discoverability. Say we introduce other things to cache, like `epsilon_on_grid`, or perhaps parts of that. Then one has to remember which of those fields are invalidated by `use_local_subpixel` and which aren't. So basically what I'm proposing is to:
```py
# in AbstractYeeGridSimulation
def _invalidate_solver_cache(self) -> None:
    """Clear cached attributes that become stale when subpixel changes."""
    self._cached_properties.pop("_mode_solver", None)
    # future caches go here
```

Then you call this private helper instead of manipulating the `_cached_properties` directly in whichever functions are affected. Maybe it's a bit unnecessary right now but would just future-proof the design a little (plus I think the naming is a bit clearer). Behavior stays exactly the same but there is a single obvious place to extend if new caches appear.

tidy3d/packaging.py
caseyflex: I see. For example, it would be nice to have a context manager and set a config option only within a context, e.g. to silence warnings for a certain block of code, or to disable subpixel for a certain block of code

caseyflex:
thanks @yaugenst-flex , revised from your feedback and added a unit test for the packaging. Also checked the other tests, found a couple that fail because they assume subpixel is not used. Fixed them. But this raises the question, @daquinteroflex should we update the github actions to test on systems with and without tidy3d_extras installed?

caseyflex:
@yaugenst-flex there is something weird going on when I run the full pytest-xdist here with the `tidy3d_extras` package installed. I get web API key error messages from the tidy3d web api (not the new license check), from a random selection of the tests (mostly in test_simulation and test_autograd). Not really sure what change is breaking the thread safety

caseyflex:
> @yaugenst-flex there is something weird going on when I run the full pytest-xdist here with the `tidy3d_extras` package installed. I get web API key error messages from the tidy3d web api (not the new license check), from a random selection of the tests (mostly in test_simulation and test_autograd). Not really sure what change is breaking the thread safety

I suppose this is actually coming from the new license check which uses the tidy3d web api libraries. I guess we usually use mock webapi unit tests in pytest-xdist so we don't run into issues due to web api not being threadsafe. In that case, maybe it is best to leave the unit tests not using local subpixel, and count on the unit tests in that library to check that it works. We can disable it globally for pytest

caseyflex:
ok @yaugenst-flex , added a fixture to disable local subpixel for test suite. The tests in that separate package should suffice for local subpixel integration down the line. Also clarified in the docstrings that it is not yet supported -- we can change this and add a CHANGELOG entry once it is supported. This should be ready for final review and then merging?

daquinteroflex:
Sorry @caseyflex missed this comment, I've actually replied privately on the other PR

yaugenst-flex (approved):
Thanks @caseyflex LGTM. Thread-safety of webapi is an issue but for this I agree we can just disable it for now.

caseyflex:
coverage diff report is incorrect, it's still 91% after this PR. Some of the checks are failing sometimes, also unrelated to this PR. So I'll just rerun them until they pass and then merge

PR #2497, opened 2025-05-21
Allow kwargs to be used with autograd primitives

bzhangflex:
`tidy3d.plugins.autograd.interpolate_spline()` and `tidy3d.plugins.autograd.add_at()` can now be called with autograd traceable keyword arguments (previously required positional args).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Adjoint source frequency width is adjusted to decay sufficiently before zero frequency when possible to improve accuracy of simulation normalization when using custom current sources.
 - Change `VisualizationSpec` validator for checking validity of user specified colors to only issue a warning if matplotlib is not installed instead of an error.
 
+### Changed
+- `tidy3d.plugins.autograd.interpolate_spline()` and `tidy3d.plugins.autograd.add_at()` can now be called with keyword arguments during tracing.
+
 ## [2.8.4] - 2025-05-15
 
 ### Added
--- a/tests/test_plugins/autograd/primitives/test_interpolate.py
+++ b/tests/test_plugins/autograd/primitives/test_interpolate.py
@@ -1,6 +1,7 @@
 import autograd.numpy as np
 import numpy.testing as npt
 import pytest
+from autograd import grad
 from autograd.test_util import check_grads
 from tidy3d.plugins.autograd import interpolate_spline
 
@@ -34,6 +35,22 @@ def test_interpolate_spline_grads(rng, order, num_points, endpoint_derivs, x_dis
     check_grads(interpolate_spline, modes=["rev"], order=1, argnum=1)(
         x, y, num_points, order, endpoint_derivs
     )
+
+
+def test_interpolate_spline_grads_kwargs(rng):
+    """Test interpolate_spline function can be called with kwargs."""
+    x = np.linspace(0, 1, 10)
+    y = rng.random(x.size)
+    # this should not error
+    grad(
+        lambda y_: interpolate_spline(
+            x_points=x,
+            y_points=y_,
+            num_points=10,
+            order=3,
+            endpoint_derivatives=(None, None),
+        )[1][0]
+    )(y)
 
 
 @pytest.mark.parametrize("order", [1, 2, 3])
--- a/tests/test_plugins/autograd/test_functions.py
+++ b/tests/test_plugins/autograd/test_functions.py
@@ -3,6 +3,7 @@ import numpy.testing as npt
 import pytest
 import scipy.interpolate
 import scipy.ndimage
+from autograd import grad
 from autograd.test_util import check_grads
 from scipy.signal import convolve as convolve_sp
 from tidy3d.plugins.autograd import (
@@ -385,6 +386,15 @@ class TestAddAt:
         x, y = self.generate_x_y(rng, shape, indices)
         check_grads(lambda x: add_at(x, indices, y), modes=["fwd", "rev"], order=2)(x)
         check_grads(lambda y: add_at(x, indices, y), modes=["fwd", "rev"], order=2)(y)
+
+
+def test_add_at_grad_kwargs(rng):
+    """Test add_at function for different array dimensions and indices, with kwargs."""
+    indices = (0,)
+    x = rng.uniform(-1, 1, (10,))
+    y = rng.uniform(-1, 1, x[tuple(indices)].shape)
+    # this should not error
+    grad(lambda y_: add_at(x=x, y=y_, indices_x=indices)[0])(y)
 
 
 @pytest.mark.parametrize("shape", [(5,), (5, 5), (5, 5, 5)])
--- a/tidy3d/components/autograd/functions.py
+++ b/tidy3d/components/autograd/functions.py
@@ -198,6 +198,34 @@ def trapz(y: NDArray, x: NDArray = None, dx: float = 1.0, axis: int = -1) -> flo
 
 
 @primitive
+def _add_at(x: NDArray, indices_x: tuple, y: NDArray) -> NDArray:
+    """
+    Add values to specified indices of an array.
+
+    Autograd requires that arguments to primitives are passed in positionally.
+    ``add_at`` is the public-facing wrapper for this function,
+    which allows keyword arguments in case users pass in kwargs.
+    """
+    out = np.copy(x)  # Copy to preserve 'x' for gradient computation
+    out[tuple(indices_x)] += y
+    return out
+
+
+defvjp(
+    _add_at,
+    lambda ans, x, indices_x, y: unbroadcast_f(x, lambda g: g),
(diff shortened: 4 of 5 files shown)


tests/test_plugins/autograd/primitives/test_interpolate.py
```diff
@@ -36,6 +37,52 @@ def test_interpolate_spline_grads(rng, order, num_points, endpoint_derivs, x_dis
...
+)
+@pytest.mark.parametrize(
+    "x_distribution",
+    [
+        (lambda n, rng: np.linspace(0, 1, n)),  # uniform ascending
+        (lambda n, rng: np.sort(rng.random(n))),  # random ascending
+        (lambda n, rng: np.linspace(1, 0, n)),  # uniform descending
+    ],
+)
+def test_interpolate_spline_grads_kwargs(rng, order, num_points, endpoint_derivs, x_distribution):
```
bzhangflex: wasn't sure if this was the best way to do the tests. check_grads only takes in positional args, though I suppose an alternative could be to use check_grads on a wrapper that takes in positional args, then internally calls `interpolate_spline` with kwargs?

bzhangflex: update: modified the tests to use check_grads

yaugenst-flex: Actually I think using `check_grads` is a bit overkill here. We just need to make sure that both `grad(interpolate_spline)(x, y)` as well as `grad(interpolate_spline)(x_points=x, y_points=y)` do not error, so a minimal test would just be a single call to both. And then again, since the positional style is used elsewhere in the tests, we really only need to make sure that the kwargs one doesn't error.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `eps_lim` keyword argument to `Simulation.plot_eps()` for manual control over the permittivity color limits.
 
+### Changed
+- `tidy3d.plugins.autograd.interpolate_spline()` and `tidy3d.plugins.autograd.add_at()` can now be called with autograd traceable keyword arguments.
```
yaugenst-flex: maybe reword: "can now be called with keyword arguments during tracing"

tidy3d/components/autograd/functions.py
```diff
@@ -198,7 +198,7 @@ def trapz(y: NDArray, x: NDArray = None, dx: float = 1.0, axis: int = -1) -> flo
 
 
 @primitive
-def add_at(x: NDArray, indices_x: tuple, y: NDArray) -> NDArray:
+def _add_at(x: NDArray, indices_x: tuple, y: NDArray) -> NDArray:
     """
     Add values to specified indices of an array.
```
yaugenst-flex, line 204: I'd remove the docstring here and just refer to `add_at`, which is the public-facing function. This docstring should probably just mention the rationale of why this is done.

tidy3d/plugins/autograd/primitives/interpolate.py
```diff
@@ -682,7 +682,7 @@ def get_spline_derivatives_wrt_y(
 
 
 @primitive
-def interpolate_spline(
+def _interpolate_spline(
```
yaugenst-flex, line 685: Same as with `_add_at`

yaugenst-flex (commented):
Thanks for spotting and resolving @bzhangflex! Other than simplifying the tests I think this is good to go.

tests/test_plugins/autograd/primitives/test_interpolate.py
bzhangflex: How does this look? `grad(func)` requires that `func` returns a scalar, and can only be called with positional args so I wrapped `interpolate_spline` inside of a simple objective function

yaugenst-flex: why not just:
```py
def test_interpolate_spline_grads_kwargs(rng):
  x = np.linspace(0, 1, 10)
  y = rng.random(x.size)
  # this should not error
  grad(interpolate_spline, 1)(x_points=x, y_points=y, order=3)
```

yaugenst-flex: i would expect this test to fail without your changes, right?

bzhangflex: > why not just:
> 
> ```python
> def test_interpolate_spline_grads_kwargs(rng):
>   x = np.linspace(0, 1, 10)
>   y = rng.random(x.size)
>   # this should not error
>   grad(interpolate_spline, 1)(x_points=x, y_points=y, order=3)
> ```

I've tried but it fails with this error because `grad` expects positional args
```python
@wrap_nary_f(fun, unary_operator, argnum)
    def nary_f(*args, **kwargs):
        @wraps(fun)
        def unary_f(x):
            if isinstance(argnum, int):
                subargs = subvals(args, [(argnum, x)])
            else:
                subargs = subvals(args, zip(argnum, x))
            return fun(*subargs, **kwargs)
        if isinstance(argnum, int):
>           x = args[argnum]
E           IndexError: tuple index out of range
```

`grad` with positional args also fails because `grad` expects a scalar-output function
```python
grad(interpolate_spline, 1)(x, y, num_points, order, endpoint_derivs)

>           raise TypeError("Grad only applies to real scalar-output functions. "
                            "Try jacobian, elementwise_grad or holomorphic_grad.")
E           TypeError: Grad only applies to real scalar-output functions. Try jacobian, elementwise_grad or holomorphic_grad.
```

 but using `elementwise_grad` works
```python
elementwise_grad(interpolate_spline, 1)(x, y, num_points, order, endpoint_derivs)
```

`elementwise_grad` doesn't work with kwargs though haha. Same error as `grad` with kwargs above.

This is the best I can think of
```python
    x = x_distribution(num_points, rng)
    y = rng.random(num_points)
    
    def interpspline_args2kwargs(x, y, num_points, order, endpoint_derivs):
        return interpolate_spline(
            num_points=num_points,
            order=order,
            endpoint_derivatives=endpoint_derivs,
            x_points=x,
            y_points=y,
        )
    # check that this doesn't error
    elementwise_grad(interpspline_args2kwargs, 1)(x, y, num_points, order, endpoint_derivs)
```

bzhangflex: > i would expect this test to fail without your changes, right?

Yes good point! I tried using `_interpolate_spline` in those tests and it failed, should we add another test that asserts that it fails or is that unnecessary?

yaugenst-flex: > grad with positional args also fails because grad expects a scalar-output function

ooh oops my bad. this?:
```py
def test_interpolate_spline_grads_kwargs(rng):
  x = np.linspace(0, 1, 10)
  y = rng.random(x.size)
  # this should not error
  grad(lambda y_: interpolate_spline(x_points=x, y_points=y_, order=3)[0])(y)
```

bzhangflex: oh yeah that works and is much cleaner, thanks!

bzhangflex: @yaugenst-flex updated the tests, thanks!

tests/test_plugins/autograd/primitives/test_interpolate.py
```diff
@@ -36,6 +37,41 @@ def test_interpolate_spline_grads(rng, order, num_points, endpoint_derivs, x_dis
...
+    ],
+)
+@pytest.mark.parametrize(
+    "x_distribution",
+    [
+        (lambda n, rng: np.linspace(0, 1, n)),  # uniform ascending
+        (lambda n, rng: np.sort(rng.random(n))),  # random ascending
+        (lambda n, rng: np.linspace(1, 0, n)),  # uniform descending
+    ],
+)
```
yaugenst-flex: oh sorry maybe i wasn't clear, i think we should also get rid of the parametrization here. we don't need to test these as they are already tested elsewhere so this would just kind of bloats the test suite i think, since it doesn't really have anything to do with the behavior we're testing for here

bzhangflex: ah good point, updated

tests/test_plugins/autograd/test_functions.py
```diff
@@ -386,6 +387,16 @@ def test_add_at_grad(self, rng, shape, indices):
         check_grads(lambda x: add_at(x, indices, y), modes=["fwd", "rev"], order=2)(x)
         check_grads(lambda y: add_at(x, indices, y), modes=["fwd", "rev"], order=2)(y)
 
+    def test_add_at_grad_kwargs(self, rng, shape, indices):
```
bzhangflex: @yaugenst-flex Should I simplify this too so it doesn't run the parametrizations for the class?

yaugenst-flex: ah right. yeah that would probably be best

bzhangflex: done!

yaugenst-flex (approved):

PR #2500, opened 2025-05-21
Thin conductor thickness in LossyMetalMedium

weiliangjin2021:
Not sure how useful this feature is, but it seems to be a common option. The implementation here is based on the formula  [here](https://www.jpier.org/ac_api/download.php?id=0007011).

When I try computing attenuation constant of a CPW of 1 micron thickness, I need to set the thickness here to be 0.25 micron to get a good match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `eps_lim` keyword argument to `Simulation.plot_eps()` for manual control over the permittivity color limits.
+- Added `thickness` parameter to `LossyMetalMedium` for computing surface impedance of a thin conductor.
 
 ### Changed
 - Relaxed bounds checking of path integrals during `WavePort` validation.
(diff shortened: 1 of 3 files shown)


prashkh (approved):
Thanks for adding this feature! Looks good and helps us to have feature parity to what you get for finite conductivity BC in other software.

dmarek-flex:
> When I try computing attenuation constant of a CPW of 1 micron thickness, I need to set the thickness here to be 0.25 micron to get a good match.

At all frequencies?

My concern is that it will not be accurate for surfaces that are not normal to the "thickness" of the metal, but I suppose the large surfaces normal to the thickness direction will contribute more, so overall this is probably closer to reality.

dmarek-flex (approved):
add unit test?

weiliangjin2021:
> My concern is that it will not be accurate for surfaces that are not normal to the "thickness" of the metal, but I suppose the large surfaces normal to the thickness direction will contribute more, so overall this is probably closer to reality.

Yeah, that's the main assumption under this formulation: the wave is assumed to mainly vary along the normal direction.

weiliangjin2021:
> At all frequencies?

You are right, not at all frequencies. To match lower frequency, the thickness should be closer to physical thickness.

PR #2501, opened 2025-05-22
chore: add coverage report action to local tests

frederikschubertflex:
I added this Github [action](https://github.com/insightsengineering/coverage-action?tab=readme-ov-file) to post a comment with the coverage report to the PR when running the `local_tests`. What do you think about this and do you want it to fail if the coverage falls below a certain threshold?


yaugenst-flex:
Yeah I think this is good. Another option would be to fail if a PR decreases coverage?

frederikschubertflex:
Yes, I thought about this as well, but I think you need some margin around that in case a PR adds simple boilerplate. So a coverage of over 90% is fine I guess and we should look into the integration or e2e tests. If the unit test suite succeeds at 90% coverage but the integration/e2e tests fail, we should take this as a signal that the unit test suite needs to be improved upon.

daquinteroflex:
This is good and can integrate for now! It'd be nice if we can also generate coverage for the diffed files per PR so people know what they need to test more if they can.

daquinteroflex (approved):

frederikschubertflex:
> It'd be nice if we can also generate coverage for the diffed files per PR so people know what they need to test more if they can.

Good idea! I am going to add this as it should not be too difficult.

daquinteroflex:
This is great!

frederikschubertflex:
https://github.com/flexcompute/tidy3d/pull/2501#issuecomment-2901052420 is an example of the report that is going to be generated whenever there are `*.py` files changed in the project.

yaugenst-flex (approved):

PR #2503, opened 2025-05-22
fix shaping of custom medium derivatives

groberts-flex:
Rahul ran into a problem last week (I believe when implementing the clip operations) where the eps_data he was using sometimes had more than one entry in the frequency dimension and then the shaping of the overall derivative was incorrect when using _derivative_field_cmp. This wasn't an issue in the current adjoint flow as far as I'm aware, but to be safe, this PR updates to only use the spatial coordinates now to get the shape for times when multifrequency datasets are passed into the function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed bug in broadband adjoint source creation when forward simulation had a pulse amplitude greater than 1 or a nonzero pulse phase.
+- Fixed shaping of `CustomMedium` gradients when permittivity data includes a frequency dimension with multiple entries.
 
 ### Changed
 - Relaxed bounds checking of path integrals during `WavePort` validation.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed issue with `CustomMedium` gradients where other frequencies would wrongly contribute to the gradient.
 - Fixed bug when computing `PolySlab` bounds in plotting functions.
 - Fixed bug in broadband adjoint source creation when forward simulation had a pulse amplitude greater than 1 or a nonzero pulse phase.
+- Fixed shaping of `CustomMedium` gradients when permittivity data includes a frequency dimension with multiple entries.
```
yaugenst-flex: Needs to go under "Unreleased"

yaugenst-flex (approved):
Thanks!

CHANGELOG.md
```diff
@@ -43,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed issue with `CustomMedium` gradients where other frequencies would wrongly contribute to the gradient.
 - Fixed bug when computing `PolySlab` bounds in plotting functions.
+- Fixed bug in broadband adjoint source creation when forward simulation had a pulse amplitude greater than 1 or a nonzero pulse phase.
```
yaugenst-flex: needs to be removed

groberts-flex: thanks, removed! not sure why these repeated ones keep getting into the changelog

PR #2507, opened 2025-05-23
Fix to DC voltage source units for compatibility

momchil-flex:


tidy3d/components/spice/sources/dc.py
```diff
@@ -76,5 +80,9 @@ class DCCurrentSource(Tidy3dBaseModel):
     current: pd.FiniteFloat = pd.Field(
         title="Current",
         description="DC current usually used as source in 'CurrentBC' boundary conditions.",
+        units=AMP,
     )
-    units: str = AMP
+
+    # TODO: This should have always been in the field above but was introduced wrongly as a
+    # standalone field. Keeping for compatibility, remove in 3.0.
+    units: Literal[AMP] = AMP
```
marc-flex, line 88: I'll create an issue to remind us of that

marc-flex (approved):

PR #2510, opened 2025-05-26
chore(github): add script  to determine test scope

frederikschubertflex:
This step results in a full test suite run if an already approved PR receives changes to prevent breaking the main branch.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -236,7 +236,7 @@ jobs:
       MPLBACKEND: agg
     
     concurrency:
-      group: ${{ github.workflow }}-${{ github.event.pull_request.number }}-${{ matrix.platform }}-remote
+      group: ${{ github.workflow }}-${{ github.event.pull_request.number }}-${{ matrix.platform }}-${{ matrix.python-version }}-remote
       cancel-in-progress: true
 
     steps:
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -24,7 +24,7 @@ on:
 
 jobs:
   determine-test-scope:
-    runs-on: [ inhouse ] 
+    runs-on: [ inhouse ]
     if: |
       github.event.pull_request.draft == false ||
       github.ref == 'refs/heads/develop' ||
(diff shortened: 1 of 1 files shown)


daquinteroflex:
Yes this is much better to cover the important cases thanks Frederik. Technically, we'll run full tests more, but I think it's valid in this type of scenario so we can justify the extra compute given the few breaking changes.

.github/workflows/run_tests.yml
```diff
@@ -33,6 +33,22 @@ jobs:
...
     steps:
+      - name: check-current-approval-status
+        id: approval
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const {owner, repo} = context.repo
+            const number = context.payload.pull_request.number
+            const reviews = await github.rest.pulls.listReviews({owner, repo, pull_number: number})
```
daquinteroflex, line 44: Does it require the rest api? Or is there a way to access it from the action event manager, like could it just get that info from the github event if triggered by a pull request? Just out of curiosity if you looked into it.

daquinteroflex: https://stackoverflow.com/questions/69493260/get-approval-status-in-github-action

daquinteroflex: Maybe we can reduce the complexity using that and only cover PR changes, within the determine test scope job section.

daquinteroflex (changes requested):
Just to integrate your first block with the second one if we can use the github event manager? Happy to help test this when ready.

frederikschubertflex:
> Technically, we'll run full tests more, but I think it's valid in this type of scenario so we can justify the extra compute given the few breaking changes.

We should also probably cancel concurrent workflows to reduce the number of runs. I will add this.

.github/workflows/run_tests.yml
frederikschubertflex: Yes, I also thought that this solution was a little too complex. But my understanding is that `github.event.review.state == 'approved'` is only available in the `pull_request_review` event. It will not be available in the `push` etc. events and evaluate to `false`.

frederikschubertflex: And my current approach requires the REST API as the Github script is only a thin [wrapper](https://github.com/actions/github-script) around it. I like it because it does not introduce extra dependencies such as the Github CLI.

daquinteroflex: Ahh ok sounds good then!

daquinteroflex (approved):

PR #2511, opened 2025-05-26
fix contains check for lumped element, which was too strict

dmarek-flex:
I introduced this bug a couple of months ago, when fixing a bug related to lumped elements outside the simulation domain.

We should allow lumped elements to touch the boundary.


yaugenst-flex (approved):

yaugenst-flex:
@dmarek-flex looks like some test is failing?

dmarek-flex:
yes, I am working on fixing this, and I will also move the new test to be part of that failing test

dmarek-flex:
OK so 2D structures, like lumped elements should not be against boundaries on their dimension of zero size, but they can touch it otherwise

dmarek-flex:
@yaugenst-flex The coverage report does not make complete sense to me, but this is ready to go?

PR #2404, opened 2025-04-23
:robot: :package: Tests in Docker Image

daquinteroflex:
- [x] Isolates the environment of each worker in a docker image
- [x] Implements uv as part of testing suite~
- [x] Sets up image correctly.

PR #2512, opened 2025-05-26
chore(github): use diff-cover for easy local diff coverage computation

frederikschubertflex:
This changes the computation of the changed or added line coverage to use [diff-cover](https://github.com/Bachmann1234/diff_cover) in CI which can also be used on the command line locally.

```bash
poetry run diff-cover --compare-branch=origin/develop coverage.xml
```

CC @marc-flex


daquinteroflex:
What do you think we could make this part of the development CLI suite in `tidy3d.cli.develop`? Just that way people don't have to remember most of the command and can just run a simpler one automatically? Or if not, how could we best document it? Maybe we can make this part of the development docs upgrade.
```
tidy3d develop diff-coverage
```

frederikschubertflex:
I like the idea with the CLI suite! But I would propose to use `pre-commit` hooks for this. We already use it, its self-documenting and scriptable. What do you think about this? 

The hook can be called via `pre-commit run --hook-stage manual test_local`.

daquinteroflex:
Sure yeah that's good too!

daquinteroflex:
Can you do a little demo that it works and then we merge?

frederikschubertflex:
Yes, but I tested this pre-commit approach and it has the shortcoming that you need diffed files for it to trigger, even with the manual state. So I just extracted it into a script `scripts/test_coverage.sh` that you can run and which runs the unit test suite with coverage and computes the diff coverage.

The output at looks like this:
```
...
Wrote XML report to coverage.xml
-------------
Diff Coverage
Diff: origin/develop...HEAD, staged and unstaged changes
-------------
No lines with coverage information in this diff.
-------------
```

I think your proposal of a `tidy3d.cli.develop` module is the right choice in the long run to align the CI with our local dev flow. But I would leave it out of scope for this PR.

daquinteroflex (approved):

daquinteroflex:
Implemented in #2404

PR #2514, opened 2025-05-27
Adding heat, charge and conduction simulations to test/utils.py

marc-flex:


yaugenst-flex:
@frederikschubertflex fyi coverage diff here is breaking because the PR did not add/remove anything that affects coverage

frederikschubertflex:
This should be fixed after https://github.com/flexcompute/tidy3d/pull/2512 is merged!

marc-flex:
Thanks @frederikschubertflex with that last merge this is now ready to be merged

momchil-flex (approved):
Thanks @marc-flex ! I added these simulations to the `sample.py` script that we run every time we release, as well as to the IO test. Let me know if this looks good.

marc-flex:
Looks good to me. These simulations is yet another thing to keep in mind as we add more features, though.

momchil-flex:
Yes, we haven't been very strict about it, but at least now this will catch inadvertent schema changes for the existing components!

PR #2515, opened 2025-05-27
Fix bug in EME simulations with bends

caseyflex:
A bug was introduced by a combination of the validator here:
https://github.com/flexcompute/tidy3d/pull/2053
and the change made here:
https://github.com/flexcompute/tidy3d/pull/2410

The validator compares the bend radius to the mode plane size, without taking the simulation size into account. Thus when the mode plane in EME was made infinite (so it was still a plane for 2D EME simulations), no bend radius was allowed. This broke EME simulations with nonzero bend radius in v2.8.4; they would just give a solver error.


momchil-flex (approved):
Thanks for the fix!

Somehow I feel like I had thought about this but apparently not. Maybe it was in light of some other mode solver PRs/considerations that I had thought about the need to intersect mode plane and simulation...

yaugenst-flex (approved):
Thanks @caseyflex

PR #2516, opened 2025-05-27
Minor improvement to import check for local subpixel

caseyflex:
A bit more robust handling in case there are some residual files after uninstalling tidy3d-extras


yaugenst-flex (approved):
LGTM, thanks!

PR #2521, opened 2025-05-28
Disable local subpixel by default

caseyflex:
Disabling local subpixel by default seems to be safer than automatically setting it based on the presence of the python package.

And this replaces the previous PR, eliminates the need for `subpixel=False` (the alternative there would have been carefully disabling local subpixel wherever needed. But better to just have it off by default)

### DIFF
--- a/tidy3d/config.py
+++ b/tidy3d/config.py
@@ -38,7 +38,7 @@ class Tidy3dConfig(pd.BaseModel):
     )
 
     use_local_subpixel: Optional[bool] = pd.Field(
-        None,
+        False,
         title="Whether to use local subpixel averaging. If 'None', local subpixel "
         "averaging will be used if 'tidy3d-extras' is installed and not used otherwise. "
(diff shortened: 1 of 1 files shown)


tidy3d/config.py
```diff
@@ -38,7 +38,7 @@ class Config:
     )
 
     use_local_subpixel: Optional[bool] = pd.Field(
-        None,
+        False,
```
yaugenst-flex, line 41: This is fine, just wondering whether it still makes sense to keep it `Optional` then? Or will there still be a use-case for `None` in the future?

yaugenst-flex (approved):

momchil-flex (approved):

tidy3d/config.py
caseyflex: Maybe let's keep it Optional for now, just to get it in and fix the solver tests. We can change it before release

PR #2523, opened 2025-05-29
Fix adjoint in API docs

daquinteroflex:
Just making this draft PR to track fixes required for API docs to build with the recent changes. Feel free to copy them over to your branch manually or can merge when I've confirmed it's all looking ok on docs if anything else is missing.


daquinteroflex:
@yaugenst-flex feel free to either close if you want to implement it directly or merge into your PR. Without this docs don't even build.

yaugenst-flex:
great thanks @daquinteroflex. i think this should go directly into my PR because build failures seem to be due to adjoint plugin

PR #2524, opened 2025-05-29
Adding monitor validations for Conduction simulations

marc-flex:
I'm adding validators to check that VoltageBC have single values for `voltage` in `Conduction` simulations. 

I'm also adding some more validators to avoid situations where we can submit a heat simulation without a monitor.

### DIFF
--- a/tidy3d/components/tcad/simulation/heat_charge.py
+++ b/tidy3d/components/tcad/simulation/heat_charge.py
@@ -788,6 +788,55 @@ class HeatChargeSimulation(AbstractSimulation):
         return values
 
     @pd.root_validator(skip_on_failure=True)
+    def check_heat_sim(cls, values):
+        """Make sure that heat simulations have at least one monitor defined."""
+
+        simulation_types = cls._check_simulation_types(values=values)
+
+        if TCADAnalysisTypes.HEAT in simulation_types:
+            monitors = values.get("monitors")
+            if not any(isinstance(mnt, TemperatureMonitor) for mnt in monitors):
+                raise SetupError(
+                    "Heat simulations require the definition of, at least, one "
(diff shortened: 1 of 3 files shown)


tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -778,6 +778,49 @@ def check_coupling_source_can_be_applied(cls, values):
 
         return values
 
+    @pd.root_validator(skip_on_failure=True)
+    def check_heat_sim(cls, values):
+        """Make sure that heat simulations have at least one monitor defined."""
+
+        simulation_types = cls._check_simulation_types(values=values)
+        print(simulation_types)
```
dbochkov-flexcompute: a stray `print`

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -778,6 +778,49 @@ def check_coupling_source_can_be_applied(cls, values):
...
+
+        simulation_types = cls._check_simulation_types(values=values)
+
+        if TCADAnalysisTypes.CONDUCTION in simulation_types:
+            monitors = values.get("monitors")
+            if not any(isinstance(mnt, SteadyPotentialMonitor) for mnt in monitors):
+                log.warning(
+                    "A Conduction simulation has been defined but no "
+                    "SteadyPotentialMonitor has been defined. "
+                )
```
dbochkov-flexcompute: do we also need to error if additionally there is no conduction->heat coupling?

marc-flex: Good point. 
At some point I'd like to refactor a lot of these validators. Maybe for tidy3d v3.0

momchil-flex (approved):

PR #2526, opened 2025-05-29
Fix field colocation in EMEModeSolverMonitor

caseyflex:
When `colocate=True`, previously the field was not properly colocated in an `EMEModeSolverMonitor`.

Also fixes `grid_expanded` in `EMESimulationData._extract_mode_solver_data`.


yaugenst-flex (approved):
Thanks!

PR #2533, opened 2025-06-02
Limiting number of pytest workers to 4

bryance-flexcompute:
Avoid crashing self-hosted runners by limiting to 4 workers max.

Fixing: https://github.com/flexcompute/tidy3d/actions/runs/15340356898/attempts/1 and https://flow360.atlassian.net/browse/EMA-70

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -138,7 +138,7 @@ jobs:
     needs: determine-test-scope
     if: needs.determine-test-scope.outputs.local_tests == 'true'
     name: Python ${{ matrix.python-version }} - self-hosted-runner 
-    runs-on: [ slurm-runner ]
+    runs-on: [ deploynode ]
     container: ghcr.io/astral-sh/uv:debian 
     strategy:
       matrix:
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -138,7 +138,7 @@ jobs:
     needs: determine-test-scope
     if: needs.determine-test-scope.outputs.local_tests == 'true'
     name: Python ${{ matrix.python-version }} - self-hosted-runner 
-    runs-on: [ deploynode ]
+    runs-on: [ slurm-runner ]
     container: ghcr.io/astral-sh/uv:debian 
     strategy:
       matrix:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -71,9 +71,10 @@ def pytest_xdist_auto_num_workers(config):
 
     cores = min(cores, mem_limited_cores)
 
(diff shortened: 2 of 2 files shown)


yaugenst-flex (changes requested):
#Thanks @bryance-flexcompute for identifying this as an issue on the self hosted runners! I don't think setting 4 workers as a hard upper limit is the right solution though, since this will be unnecessarily limiting on other deployments.
Do you know why 4 is the magic number? As far as I'm aware there are also runners with plenty more cores than that. Is the script not correctly identifying the hardware there?
Is there a way to fix this?
There is also logic in that function to set workers by available memory, which might be another thing to look into whether it's working correctly. Is `psutil` available?


Alternatively, maybe we could set an env var on the self-hosted runners? This could be done in the workflow directly to identify them as self-hosted, or maybe there is already something available we could use?

bryance-flexcompute:
Hi @yaugenst-flex I talked with Lei and Yuwei about this yesterday, and Lei suggested a 4 to 8 worker limit. The reasoning was: in this [failure](https://github.com/flexcompute/tidy3d/actions/runs/15340356898/job/43165203694) for example, the passing tests running on github runners only use 2 workers: eg [here](https://github.com/flexcompute/tidy3d/actions/runs/15340356898/job/43165203881#step:7:27), so he didn't believe more much more was necessary. Yuwei also thought the failure is maybe the network got overloaded: there is a gap in the telemetry data for that runner at the time of failure on our grafana graphs, so it may not be an out-of-memory issue that I initially thought: it's hard to diagnose though as there's the gap in data.

Is the issue that particular script is used in other pytests that are currently running fine with many more workers?

bryance-flexcompute:
Hi @yaugenst-flex, I discussed this more with Lei today:
- He is fine with the environment var to check if it's self-hosted, and limit the cores based on that: I updated the code for this. But he does want some kind of hard resource limit.
- With 8 cores, [job finishes in 6-7 min](https://github.com/flexcompute/tidy3d/actions/runs/15426603868/job/43414984393#step:5:22) with `runs-on: [ deploynode  ]`
- Currently, running the job does block the entire runner (other CI jobs cannot run) even if a max limit to cores is artificially imposed, so it is good to have the job finish as quickly as possible to unclog the CI. Lei is ok with this for now. In the future, we will move everything to slurm workers and be able to specify num of cpus in the runs-on: [].
- Reason we can't use slurm workers right now: webhook is needed to launch a slurm worker, and github repo can only have 1 webhook, which is currently already in use. This will need to be worked around, e.g. make another webhook that forks to the existing one and the launch the slurm worker.

yaugenst-flex (approved):
Thanks @bryance-flexcompute LGTM!

Issue #2531, opened 2025-06-02
Error when `angle_rotation` is used with geometries with infinite dimension


momchil-flex:
Our geometry transformations (or at least, rotation) error if a geometry with an infinite dimension needs to be transformed. This results in an error that currently only happens server-side after a simulation starts running, if a mode object with `angle_rotation=True` is defined, but the mode plane intersects geometries with an infinite extent. This is because the mode solver tries to rotate these geometries to compute the rotated fields, but then the transformation error is hit.

Probably the best solution is to construct the rotated mode solver (without solving it) for every mode object that has `angle_rotation = True` in a post-init validator.

PR #2534, opened 2025-06-03
Validating that structure rotation in the mode solver can be done

momchil-flex:
Adding a `.rotated` method to optical media which errors for all except a simple `td.Medium`.

This addresses #2531 and also @lucas-flexcompute 's observation that using `angle_rotation` with anisotropic media can produce wrong results. For now we just validate this out - in the future, if a `rotated` method is implemented for a given medium, it will then work automatically.

EDIT: In principle, translation should also be added for Custom-type media. However, I am not sure how much effort we should put into this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed field colocation in `EMEModeSolverMonitor`.
 - Solver error for EME simulations with bends, introduced when support for 2D EME simulations was added.
 - Internal interpolation errors with some versions of `xarray` and `numpy`.
+- If `ModeSpec.angle_rotation=True` for a mode object, validate that the structure rotation can be successfully done. Also, error if the medium cannot be rotated (e.g. anisotropic or custom medium), which would previously have just produced wrong results.
 
 ### Changed
 - Relaxed bounds checking of path integrals during `WavePort` validation.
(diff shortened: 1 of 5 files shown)


tidy3d/components/medium.py
```diff
@@ -1973,6 +1991,23 @@ def _derivative_eps_complex_volume(
 
         return vjp_value.sum("f")
 
+    def rotated(self, angle: float, axis: Union[Axis, Coordinate]) -> Medium:
```
lucas-flexcompute: Isn't this going to be inherited by CustomIsotropicMedium too? My guess is that only uniform, isotropic medium should use this method.

momchil-flex: Yeah, good point.

momchil-flex:
So I fully reworked this PR to remove the `Medium.rotated` method because as @lucas-flexcompute pointed out there was an issue with this. In fact there was more than one issue - while `CustomIsotropicMedium` would not have raised an error, isotropic but dispersive media, i.e. all `PoleResidueModel` media, would have. For now, I instead just check whether the medium that needs to be rotated is an instance of the isotropic uniform media set.

I also introduced something new: an FDTD simulation with mode objects (source/monitor) will now try to construct the associated mode solver. This is a bit of a bigger change, but it feels like it is worth it to know that things will not error on the server, where these mode solvers will have to be constructed. It was also hard to come up with a way around this to validate the structure rotation, and since I was doing it for the structure rotation check anyway, it felt like I should just always do it. This allowed me to remove two other validators which now are just part of the ModeSolver validation.

With this, a few tests fail, but they are in the jax adjoint plugin set, and our pydantic v2 PR might be deprecating that...

@dbochkov @weiliangjin2021 fyi - am I forgetting some potentially unwanted consequence in validating FDTD mode objects like [this](https://github.com/flexcompute/tidy3d/blob/340ce9cb78d408667e53eec55df2769874db18c7/tidy3d/components/simulation.py#L3705-L3733)?

lucas-flexcompute (approved):

PR #2536, opened 2025-06-03
update pre-commit hook

dmarek-flex:


daquinteroflex:
Curious how it failed?

daquinteroflex (approved):

dmarek-flex:
> Curious how it failed?

It does not fail, it just says that we are using legacy alias, so I though I would update 

> [INFO] This may take a few minutes...
> ruff (legacy alias)......................................................Passed
> ruff format..............................................................Passed
>

Issue #1072, opened 2023-08-16
Add option to transpose plot


tylerflex:
by default, plotting on a plane puts the lowest transverse dimension (in 'xyz') on the x axis and the highest on the y axis.

It would be useful to be able to switch this convention with an extra kwarg `transpose: bool = False` passed to `Simulation.plot()`

```
Sim.plot(x=0) # yz plot
Sim.plot(x=0, transpose=False) # also yz plot
Sim.plot(x=0, transpose=True)  # zy plot
```

tylerflex:
@m-bone since you've already been digging in the plotting code a bit, can you take this one on as well? Could be a good 2nd warmup for our development flow so you can get a little more practice. hopefully the code bits are pretty minor. Thanks!

jewettaijfc:
> This [notebook](https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/develop/docs/source/notebooks/VizSimulation.ipynb) can be useful for additional testing.

Note-to-self: The notebook that Tyler recommended to use for testing has been moved [here](https://github.com/flexcompute/tidy3d-notebooks/blob/develop/VizSimulation.ipynb)

jewettaijfc:
This issue is due to a fundamental limitation of the [`pop_axis()` function](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/geometry/base.py#L725) which is invoked in at least 45 places throughout the code.  That function needs to be replaced with something more general.  I will try hard not to break too much existing functionality as I do that.

jewettaijfc:
I can think of two ways of specify this information through the argument list:
- a `transpose: bool` argument (as described), OR
- an `axis_order: str` argument.  *(It's allowed values would be `xy`, `yx`, `yz`, `zy`, `xz`, `zx`.  The first character specifies the horizontal axis (`x`,`y`,or `z`).  The second character specifies the vertical axis.)*

Either choice is equally easy to implement.  Which do you think would be easier for the user to understand?  *(Or should I call it something else besides `axis_order`?)*

tylerflex:
So one important piece of information is that we already use the dimension (x,y,or z) to select out the plotting plane. eg `sim.plot(z=0)` plots on the z=0 line.  So there are really only two plotting options fundamentally, regular or transpose.

So if we go with axis_order, that means that 4 of the 6 options are invalid and would lead to errors. So in that sense, passing `transpose` would be simpler. 

[...]

PR #2537, opened 2025-06-04
XY transposable sim.plot() plots (issue1072)

jewettaijfc:
This PR addresses he feature request from issue #1072 (adding the ability to swap the horizontal and vertical axes of a simulation plot).

This includes:
- transposing all of the coordinate geometry
- swapping the X and Y axis labels

(I'm not yet sure whether I should swap the horizontal and vertical plot limits, because it might confuse the user.  Will ask for feedback regarding this detail.)

------
## UPDATE: There is an alternative version of this PR, ([PR2544](https://github.com/flexcompute/tidy3d/pull/2544)).

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -451,6 +451,7 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
         hlim: Optional[tuple[float, float]] = None,
         vlim: Optional[tuple[float, float]] = None,
         fill_structures: bool = True,
+        transpose: bool = False,
         **patch_kwargs,
     ) -> Ax:
         """Plot each of simulation's components on a plane defined by one nonzero x,y,z coordinate.
@@ -515,7 +516,7 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
             bounds=self.simulation_bounds, ax=ax, x=x, y=y, z=z, hlim=hlim, vlim=vlim
         )
         ax = self.plot_boundaries(ax=ax, x=x, y=y, z=z)
-
+        Geometry.transpose_axis_info(ax, swap_axis_labels=transpose)  # Swap axis labels if needed
         return ax
 
     @equal_aspect
@@ -930,6 +931,7 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
         vlim: Optional[tuple[float, float]] = None,
         override_structures_alpha: float = 1,
         snapping_points_alpha: float = 1,
+        transpose: bool = False,
         **kwargs,
     ) -> Ax:
         """Plot the cell boundaries as lines on a plane defined by one nonzero x,y,z coordinate.
@@ -1064,6 +1066,7 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
         ax = Box.add_ax_labels_and_title(
             ax=ax, x=x, y=y, z=z, plot_length_units=self.plot_length_units
         )
+        Geometry.transpose_axis_info(ax, swap_axis_labels=transpose)  # Swap axis labels if needed
         return ax
 
     @equal_aspect
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -478,6 +478,8 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
             The x range if plotting on xy or xz planes, y range if plotting on yz plane.
         vlim : Tuple[float, float] = None
             The z range if plotting on xz or yz planes, y plane if plotting on xy plane.
+        transpose: bool = False
+            Optional: Swap the horizontal and vertical axes.
 
         Returns
         -------
@@ -492,7 +494,7 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
 
         """
         hlim, vlim = Scene._get_plot_lims(
-            bounds=self.simulation_bounds, x=x, y=y, z=z, hlim=hlim, vlim=vlim
+            bounds=self.simulation_bounds, x=x, y=y, z=z, hlim=hlim, vlim=vlim, transpose=transpose,
         )
 
         ax = self.scene.plot(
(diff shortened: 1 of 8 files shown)


jewettaijfc:
Although the basic functionality works *(for `sim.plot()` and `sim.plot_grid()`)*, this pull-request is not yet ready for review.  (I was hoping to make more progress today.)

As Tyler mentioned, there are a bunch of other "plot" functions in the code.

I'm going through them to find the ones where I should add a "transpose" argument.

tylerflex:
Let's limit the scope to `Simulation.plot()`, `Simulation.plot_eps()`, and any sub-functions that they call where we will need to pass `transpose` in. (so for example, dont worry about plotting data outputs or HeatSimulation, etc)

and yes I think plot limits should also be transposed. You'll see there are some functions that compute vlims and hlims, I think it should be possible to pass `transpose` through to those and handle it.

Thanks!

jewettaijfc:
Note to self: At a minimum, I still need to add the `transpose` argument to:
- [Geometry.plot()](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/geometry/base.py#L491).  *(Note: That function just invokes `plot_shape()`, which I already modified.)*
- [Source.plot()](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/source/base.py#L71)
- [Monitor.plot()](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/monitor.py#L355)
- [Geometry._plot_arrow()](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/geometry/base.py#L2190)

I still need to test:
- The grid points are also transposed.
- Arrow directions are transposed

jewettaijfc:
> Let's limit the scope to `Simulation.plot()`, `Simulation.plot_eps()`, and any sub-functions that they call where we will need to pass `transpose` in. (so for example, dont worry about plotting data outputs or HeatSimulation, etc)

Thanks Tyler!

> and yes I think plot limits should also be transposed. You'll see there are some functions that compute vlims and hlims, I think it should be possible to pass transpose through to those and handle it.

To clarify, here's a concrete example, plotting a simulation in the XY plane:
```
ax = sim.plot(z=0, ax=ax, hlim=[-4,4], vlim=[-8,8])
```
It will look like a vertical brick because the vertical limits are bigger than the horizontal limits.

Now suppose, we set `transpose=True`:
```
ax = sim.plot(z=0, ax=ax, hlim=[-4,4], vlim=[-8,8], transpose=True)
```
In that case, the entire plot will indeed be transposed.  *(It will look like a brick laid horizontally.)*  But it also means that the ***vertical*** range of the plot will be controlled by `hlim` (=[-4,4]).  *(And, similarly, the horizontal range of the plot will be controlled by `vlim` (=[-8,8]))*.  Is that what you want?

tylerflex:
oh I see what you mean. I think for `vlim` and `hlim` submitted by user, they should **not** be transposed. eg. it should always just correspond to the vertical and horizontal limits of the **output** (transposed or not)

However, when these values are computed **internally** (ie when not supplied by user), let's take the transpose into account.

Does that seem reasonable?

jewettaijfc:
> oh I see what you mean. I think for `vlim` and `hlim` submitted by user, they should **not** be transposed. eg. it should always just correspond to the vertical and horizontal limits of the **output** (transposed or not)
> 
> However, when these values are computed **internally** (ie when not supplied by user), let's take the transpose into account.
> 
> Does that seem reasonable?

Yeah!  I'll take a look at that.  Thanks!

Update: 
- Working on rotating the grid points.  (It's not working yet, but I haven't tried everything).
- In `geometry/base/`, I am updating `plot_sources()`, and `plot_monitors()`, `plot_pml()`, `plot_lumped_elements()`, `plot_boundaries()`.  (These are all referenced by `Simulation.plot()`).
- Reading about [`plot_symmetries()`](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/base_sim/simulation.py#L404), and deciding how to update it.

jewettaijfc:
*(I just committed the changes that I made to this branch today in order to backup this code.  It runs without crashing, but it has some weird quirks, which will require debugging.)*

Hi Tyler: **I don't plan on continuing with this branch today.**  As I mentioned on slack, I'm going to start a new PR that uses a different approach.  (If we decide that approach ultimately works better, we can abandon this PR.)

jewettaijfc:
I already said this to Tyler in our Slack-chat, but Dario (if you are reading this), here's the reason I am deprecating this PR (for now):  There were just too many places in the code where I had to add something like this:
```
if transpose:
    do this custom thing
else:
    do the regular thing
```
Many different functions and types of plots needed a different "custom thing" .  I just wasn't sure in every case that the "custom thing" I was adding was correct.   Testing each custom thing I added was hard.

I'm about 80% done with creating a  newPR which uses a different approach.  Going to bed.

tidy3d/components/simulation.py
```diff
@@ -1184,6 +1211,11 @@ def set_plot_params(boundary_edge, lim, side, thickness):
         ax = Box.add_ax_labels_and_title(
             ax=ax, x=x, y=y, z=z, plot_length_units=self.plot_length_units
         )
+        if transpose:
+            import sys
+            sys.stderr.write(
```
tylerflex, line 1216: can use` raise NotImplementedError(...)` here

tylerflex:
@jewettaijfc actually I think this PR is basically in line with what I expect. 

The only thing I'm a little confused about is why the new

```py
Geometry.transpose_axis_info
```
vs passing `transpose` to 
```py
Box.add_ax_labels_and_title
```
and handling it there?

otherwise I think the if/else complications are simply unavoidable due to the nature of this problem. Although there were a few areas I saw where it could probably be cleaned up slightly for example

```py
           if transpose:
                ax.plot(ys, xs, color=plot_params.facecolor, linewidth=plot_params.linewidth)
            else:
                ax.plot(xs, ys, color=plot_params.facecolor, linewidth=plot_params.linewidth)
```
could be 
```py
           if transpose:
                xs, ys = ys, xs
           ax.plot(xs, xs, color=plot_params.facecolor, linewidth=plot_params.linewidth)
```

tidy3d/components/simulation.py
jewettaijfc: Thanks.  (If we decide to continue with this PR, I'll need to come back and fix this part of the code.)

jewettaijfc:
> ```python
>            if transpose:
>                 xs, ys = ys, xs
>            ax.plot(xs, xs, color=plot_params.facecolor, linewidth=plot_params.linewidth)
> ```

Good suggestion!

jewettaijfc:
> The only thing I'm a little confused about is why the new
> ```python
> Geometry.transpose_axis_info
> ```
> vs passing `transpose` to
> ```python
> Box.add_ax_labels_and_title
> ```
> and handling it there?

Ahh.  Good question.  In my first few commits, I added this line at the end of my `plot_sim()` and `plot_grid()` functions to swap the axis labels.  Later, I realized I need to update some of the other `plot_` functions as well.  Remembering which functions I need to modify was a cognitive burden.  So it was just easier to apply the same pattern to all of the `plot` functions:

That is: Whenever `self.add_ax_labels_and_title()` is invoked, follow it with `Geometry.transpose_axis_info()`.

I can stop doing that in `Box.add_ax_labels_and_title()` and other low-level functions, if you prefer.

(The other PR I'm working does not use this pattern.)

jewettaijfc:
After reading your comment, I also noticed that one of the `add_ax_labels_and_title()` functions mistakenly has `transpose` as an argument.  Thanks for catching that!  I will take that out.

jewettaijfc:
For the record, the other PR is [#2544](https://github.com/flexcompute/tidy3d/pull/2544).

jewettaijfc:
The other PR (#2544 ) is working well and is much further along.  I think I will close this PR for now so that people don't get confused.  We can always reopen it later if we want to.

PR #2538, opened 2025-06-04
Avoid interpolation errors with some xarray versions

### LINKED ISSUE #2535: TypeError on mode solver on phase_primal.interp
**Describe the bug**
Using a recent version of numpy, 2.x.x, there is a bug using the mode solver where we get a TypeError: <class 'numpy.float64'> from https://github.com/flexcompute/tidy3d/blob/v2.8.4/tidy3d/components/mode/mode_solver.py#L1684 This is due to a lack of casting to a float. There was a patch https://github.com/flexcompute/tidy3d/pull/2143/files which addressed a similar issue but this issue still remains.

**To Reproduce**
Run a mode solver using a non cloud mode solver

**Expected behavior**
The program should not crash when using the local mode solver

**Screenshots**

 - OS: linux

**Additional context**
Stack Trace:
```
/python3.12/site-packages/tidy3d/components/mode/mode_solver.py in ?(self)
    313             "Use the remote mode solver with subpixel averaging for better accuracy through "
    314             "'tidy3d.web.run(...)' or the deprecated 'tidy3d.plugins.mode.web.run(...)'.",
    315             log_once=True,
    316         )
--> 317         return self.data

/python3.12/site-packages/tidy3d/components/base.py in ?(self)
     53 
     54         if stored_value is not None:
     55             return stored_value
     56 
---> 57         computed_value = prop(self)
     58         self._cached_properties[prop_name] = computed_value
     59         return computed_value

/python3.12/site-packages/tidy3d/components/mode/mode_solver.py in ?(self)
   1207         -------
   1208         ModeSolverData
   1209             :class:`.ModeSolverData` object containing the effective index and mode fields.
   1210         """
-> 1211         mode_solver_data = self.data_raw
   1212         return mode_solver_data.symmetry_expanded_copy

/python3.12/site-packages/tidy3d/components/base.py in ?(self)
     53 
     54         if stored_value is not None:
     55             return stored_value
     56 
---> 57         computed_value = prop(self)
     58         self._cached_properties[prop_name] = computed_value
     59         return computed_value

/python3.12/site-packages/tidy3d/components/mode/mode_solver.py in ?(self)
    372         if self.mode_spec.angle_rotation and np.abs(self.mode_spec.angle_theta) > 0:
    373             return self.rotated_mode_solver_data
    374 
    375         # Compute data on the Yee grid
--> 376         mode_solver_data = self._data_on_yee_grid()
    377 
    378         # Colocate to grid boundaries if requested
    379         if self.colocate:

/python3.12/site-packages/tidy3d/components/mode/mode_solver.py in ?(self)
   1022             )
   1023             data_dict[field_name] = scalar_field_data
   1024 
   1025         # finite grid corrections
-> 1026         grid_factors = solver._grid_correction(
   1027             simulation=solver.simulation,
   1028             plane=solver.plane,
   1029             mode_spec=solver.mode_spec,

/python3.12/site-packages/tidy3d/components/mode/mode_solver.py in ?(simulation, plane, mode_spec, n_complex, direction)
   1619         phase_dual = np.exp(1j * k_vec * (normal_dual - normal_pos))
   1620 
   1621         # Fields are modified by a linear interpolation to the exact monitor position
   1622         if normal_primal.size > 1:
-> 1623             phase_primal = phase_primal.interp(**{normal_dim: normal_pos})
   1624         else:
   1625             phase_primal = phase_primal.squeeze(dim=normal_dim)
   1626         if normal_dual.size > 1:

/python3.12/site-packages/xarray/core/dataarray.py in ?(self, coords, method, assume_sorted, kwargs, **coords_kwargs)
   2379         if self.dtype.kind not in "uifc":
   2380             raise TypeError(
   2381                 f"interp only works for a numeric type array. Given {self.dtype}."
   2382             )
-> 2383         ds = self._to_temp_dataset().interp(
   2384             coords,
   2385             method=method,
   2386             kwargs=kwargs,

/python3.12/site-packages/xarray/core/dataset.py in ?(self, coords, method, assume_sorted, kwargs, method_non_numeric, **coords_kwargs)
   4122         if kwargs is None:
   4123             kwargs = {}
   4124 
   4125         coords = either_dict_or_kwargs(coords, coords_kwargs, "interp")
-> 4126         indexers = dict(self._validate_interp_indexers(coords))
   4127 
   4128         if coords:
   4129             # This avoids broadcasting over coordinates that are both in

/python3.12/site-packages/xarray/core/dataset.py in ?(self, indexers)
   2933                     yield k, IndexVariable((k,), v, attrs=self.coords[k].attrs)
   2934                 else:
   2935                     raise AssertionError()  # Already tested by _validate_indexers
   2936             else:
-> 2937                 raise TypeError(type(v))

TypeError: <class 'numpy.float64'>
```

momchil-flex:
Fixes #2535 , as well as a few other instances where the `TypeError` occurs. I found that this is specific to older `xarray` versions and has been fixed in newer ones, but I guess we can get these fixes in anyway to avoid user confusion.

### DIFF
--- a/tidy3d/components/parameter_perturbation.py
+++ b/tidy3d/components/parameter_perturbation.py
@@ -942,6 +942,10 @@ class CustomChargePerturbation(ChargePerturbation):
         if isinstance(h_vals, UnstructuredGridDataset):
             h_vals = h_vals.values
 
+        # Needed to avoid error in some xarray / numpy versions
+        e_vals = e_vals.item() if e_vals.size == 1 else e_vals
+        h_vals = h_vals.item() if h_vals.size == 1 else h_vals
+
         # note that the dimensionality of this operation differs depending on whether xarrays
         # or simple unlabeled arrays are provided:
         # - for unlabeled arrays, values are broadcasted
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bug when generating a grid with snapping points near the simulation boundaries.
 - Fixed field colocation in `EMEModeSolverMonitor`.
 - Solver error for EME simulations with bends, introduced when support for 2D EME simulations was added.
+- Internal interpolation errors with some versions of `xarray` and `numpy`.
 
 ### Changed
 - Relaxed bounds checking of path integrals during `WavePort` validation.
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -1146,6 +1146,13 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
                 "'freq' was not specified for 'FieldData.to_zbf()'. Defaulting to the mean frequency of the dataset."
             )
             freq = np.mean(e_x.coords["f"].values)
+        else:
+            freq = np.array(freq)
+
+        if freq.size > 1:
+            raise ValueError("'freq' must be a single value, not an array.")
+        else:
+            freq = freq.item()
 
         mode_area = mode_area.interp(f=freq)
         e_x = e_x.interp(f=freq)
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -583,6 +583,9 @@ class AbstractYeeGridSimulationData(AbstractSimulationData, ABC):
 
         # select the extra coordinates out of the data from user-specified kwargs
         for coord_name, coord_val in sel_kwargs.items():
+            interp_val = np.array(coord_val)
+            if interp_val.size == 1:
+                interp_val = interp_val.item()
             if (
                 field_data.coords[coord_name].size <= 1
                 or coord_name == "eme_port_index"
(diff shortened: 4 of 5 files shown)


yaugenst-flex (approved):
LGTM, considering I fixed a similar issue a while back.

tidy3d/components/parameter_perturbation.py
```diff
@@ -942,6 +942,12 @@ def sample(
         if isinstance(h_vals, UnstructuredGridDataset):
             h_vals = h_vals.values
 
+        # Needed to avoid error in some xarray / numpy versions
+        if e_vals.size == 1:
```
tylerflex: just a heads up, not sure of the shape of `e_vals`, but if I do this float casting I get a deprecation warning (both in numpy 1 and 2)

```py
>>> import numpy as np
>>> a = np.array([[[[1]]]])
>>> float(a)
<python-input-3>:1: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
  float(a)
1.0

```

perhaps it would be safer to `float(np.squeeze(e_vals))` or I think `e_vals.item()` is perhaps the standard way.

tylerflex (changes requested):
In general looks good, but a few things:
1. can be sure that the passed coordinates are `size==1` in the first 3 instances of this fix? otherwise we'll get an obscure error.
2. `.item()` might be a more standard way to handle extraction of a scalar from a size==1 numpy array as it seems `float(a)` is deprecated if a.size==1 but it has some shape.

tidy3d/components/parameter_perturbation.py
momchil-flex: Yeah using .item() sounds good!

yaugenst-flex:
~~`item()` returns a scalar numpy array type and not a standard Python type. If the type is what is causing the issue then that might not work.~~

ah nevermind apparently only for longdouble, i misremembered https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.item.html

tidy3d/components/parameter_perturbation.py
momchil-flex: Actually I guess we need to be sure that it's an np.array and not a float already... So maybe np.squeeze is better.

tylerflex: yea good point

momchil-flex:
Alright did some fixes in line with @tylerflex 's comments.

tylerflex (approved):

Issue #2529, opened 2025-05-31
Convenience class for handling frequency/wavelength conversion and assignment


dbochkov-flexcompute:
**Is your feature request related to a problem? Please describe.**
Setting up an EM simulation requires specification of frequency ranges and sample points for sources and monitors. Depending on the context the user can start with different ways of describing their frequency range of interest:
- [interval of wavelengths](https://docs.simulation.cloud/projects/tidy3d/en/latest/notebooks/MetasurfaceBIC.html) 
```
lda0 = 0.8  # central wavelength
freq0 = td.C_0 / lda0  # central frequency
ldas = np.linspace(0.7, 0.9, 201)  # wavelength range
freqs = td.C_0 / ldas  # frequency range
fwidth = 0.5 * (np.max(freqs) - np.min(freqs))  # width of the source frequency range
```
- [interval of frequencies](https://docs.simulation.cloud/projects/tidy3d/en/latest/notebooks/MicrowaveFrequencySelectiveSurface.html) 
```
freq0 = 12 * GHz  # central frequency
freqs = np.linspace(10, 14, 500) * GHz  # frequency range of interest

fwidth = 0.5 * (np.max(freqs) - np.min(freqs))  # width of the source spectrum

lda0 = td.C_0 / freq0  # central wavelength
```
- [central frequency + width of the interval](https://docs.simulation.cloud/projects/tidy3d/en/latest/notebooks/OptimizedL3.html) 
```
# Central frequency around which we'll look for the cavity mode (Hz)
wavelength = 1.55
freq0 = td.C_0 / wavelength

# Source bandwidth (Hz)
fwidth = 1e13
```
- etc

In any case the user typically has to manually convert their frequency range representation into the central frequency `freq0` and the width of interval `fwitdh` to pass to a `GaussianPulse` when creating a source, and arrays of frequency sample points to pass to monitors. This is a repetitive (and even annoying at times) task that could be avoided. 

**Describe the solution you'd like**
One way of unifying and simplifying this could be introduction of a convenience class called, for example, `FreqRange` that automatically does the required conversions. That is, it could have fields `freq0` and `fwidth`
```
freq_range = td.FreqRange(freq0=f0, fwidth=fwidth)
```
and additional class methods to initialize from different representations:
```
freq_range = td.FreqRange.from_interval(fmin=fmin, fmax=fmax)
freq_range = td.FreqRange.from_wavelength(wvl0=wvl0, wvl_width=wvl_width)
[...]

PR #2542, opened 2025-06-05
Add FreqRange utility class and unit tests (related issue #2529)

George-Guryev:
## Summary 
This pull request implements a new utility class `FreqRange` for convenient handling of frequency and wavelength conversions in simulations.

### Related issue
Convenience class for handling frequency/wavelength conversion and assignment #2529

### What was implemented
- Created FreqRange class to:
   - Accept central frequency and bandwidth as input (SI units).
   - Provide computed fmin, fmax, freqs, and ldas.
   - Support construction from wavelength ranges and intervals.
   - Include sampling method for generating frequency arrays.
   - Provide compatibility with GaussianPulse.
   
- Added unit tests in tests/test_FreqRange.py to validate:
   - Constructor and class methods (frequency- and wavelength-based initializations)
   - Sampling behavior
   - Instantiation of GaussianPulse with given central frequency and bandwidth

## Greptile Summary

Added new `FreqRange` utility class for standardized handling of frequency/wavelength conversions in simulations, with support for various initialization methods and Gaussian pulse compatibility.

- Added `tidy3d/components/source/freq_range.py` implementing core frequency range handling with computed properties for fmin, fmax, freqs, and wavelengths
- Potential floating-point precision issues in frequency calculations need attention (using == for comparisons)
- Factory methods provide flexible initialization from frequency/wavelength ranges
- Added comprehensive test suite in `tests/test_components/test_FreqRange.py` covering core functionality
- Documentation and changelog entries properly reflect the new feature addition

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Implemented `FreqRange` utility class for frequency/wavelength handling with constructor methods `from_freq_interval()`, `from_wavelength()`, and `from_wvl_interval()`. 
 - Add support for `np.unwrap` in `tidy3d.plugins.autograd`.
 - Add Nunley variant to germanium material library based on Nunley et al. 2016 data.
 - Add `PointDipole.sources_from_angles()` that constructs a list of `PointDipole` objects needed to emulate a dipole oriented at a user-provided set of polar and azimuthal angles.
--- a/tidy3d/components/source/freq_range.py
+++ b/tidy3d/components/source/freq_range.py
@@ -1,0 +1,228 @@
+"""Utility class ``FreqRange`` for frequency and wavelength handling."""
+
+from __future__ import annotations
+
+import numpy as np
+import pydantic.v1 as pydantic
+from numpy.typing import NDArray
+
+from tidy3d import constants as td_const
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.components.source.time import GaussianPulse
+
+
+class FreqRange(Tidy3dBaseModel):
+    """
+    Convenience class for handling frequency/wavelength conversion; it simplifies specification
+    of frequency ranges and sample points for sources and monitors.
+
+    Notes
+    -----
+        Depending on the context the user can define desired frequency range by specifying:
+        - central frequency ``freq0`` and frequency bandwidth ``fwidth``;
+        - frequency interval [``fmin``,``fmax``];
+        - central wavelength ``wvl0`` and wavelength range ``wvl_width``;
+        - wavelength interval [``wvl_min``, ``wvl_max``].
+
+    Example
+    -------
+    >>> import tidy3d as td
+    >>> freq0  = 1e12
+    >>> fwidth = 1e11
+    >>> freq_range = td.FreqRange(freq0=freq0, fwidth=fwidth)
+    >>> central_freq = freq_range.freqs(num_points=1)
+    >>> freqs = freq_range.freqs(num_points=11)
+    >>> source = freq_range.to_gaussian_pulse()
+    """
+
+    freq0: pydantic.PositiveFloat = pydantic.Field(
+        ...,
+        title="Central frequency",
+        description="Real-valued positive central frequency.",
+        units="Hz",
+    )
+
+    fwidth: pydantic.PositiveFloat = pydantic.Field(
+        ...,
+        title="Frequency bandwidth",
+        description="Real-valued positive width of the frequency range (bandwidth).",
+        units="Hz",
+    )
+
+    @property
+    def fmin(self) -> float:
+        """Infer lowest frequency ``fmin`` from central frequency ``freq0`` and bandwidth ``fwidth``."""
+        return self.freq0 - self.fwidth
+
+    @property
+    def fmax(self) -> float:
+        """Infer highest frequency ``fmax`` from central frequency ``freq0`` and bandwidth ``fwidth``."""
+        return self.freq0 + self.fwidth
+
+    @property
+    def lda0(self) -> float:
+        """Get central wavelength from central frequency and bandwidth."""
+        lmin = td_const.C_0 / (self.freq0 + self.fwidth)
+        lmax = td_const.C_0 / (self.freq0 - self.fwidth)
+        return 0.5 * (lmin + lmax)
+
+    @classmethod
+    def from_freq_interval(cls, fmin: float, fmax: float) -> FreqRange:
+        """
+        method ``from_freq_interval()`` creates instance of class ``FreqRange`` from frequency interval
+        defined by arguments  ``fmin`` and ``fmax``.
+
+        NB: central frequency never corresponds to central wavelength!
+        ``freq0 = (fmin + fmax) / 2`` implies that ``lda0 != (lda_min + lda_max) / 2`` and vise-versa.
+
+        Parameters
+        ----------
+        fmin : float
+            Lower bound of frequency of interest.
+        fmax : float
+            Upper bound of frequency of interest.
+
+        Returns
+        -------
+        FreqRange
+            An instance of ``FreqRange`` defined by frequency interval [``fmin``, ``fmax``].
+        """
+
+        # extract frequency-related info
+        freq0 = 0.5 * (fmax + fmin)  # extract central freq
+        fwidth = 0.5 * (fmax - fmin)  # extract bandwidth
+        return cls(freq0=freq0, fwidth=fwidth)
+
+    @classmethod
+    def from_wavelength(cls, wvl0: float, wvl_width: float) -> FreqRange:
+        """
+        method ``from_wavelength()`` updated instance of class ``FreqRange`` by reassigning new
+        frequency- and wavelength-related parameters.
+
+        NB: central frequency never corresponds to central wavelength!
+        ``lda0 = (lda_min + lda_max) / 2`` implies that ``freq0 != (fmin + fmax) / 2`` and vise versa.
+
+        Parameters
+        ----------
+        wvl0 : float
+            Real-valued central wavelength.
+        wvl_width : float
+            Real-valued wavelength range.
+
+        Returns
+        -------
+        FreqRange
+            An instance of ``FreqRange`` defined by central wavelength ``wvl0`` and wavelength range ``wvl_width``.
+        """
+
+        # calculate lowest and highest frequencies
+        fmin = td_const.C_0 / (wvl0 + wvl_width)
+        fmax = td_const.C_0 / (wvl0 - wvl_width)
+
+        return cls.from_freq_interval(fmin=fmin, fmax=fmax)
+
+    @classmethod
+    def from_wvl_interval(cls, wvl_min: float, wvl_max: float) -> FreqRange:
+        """
+        method ``from_wvl_interval()`` updated instance of class ``FreqRange`` by reassigning new
+        frequency- and wavelength-related parameters.
+
+        NB: central frequency never corresponds to central wavelength!
+        ``lda0 = (lda_min + lda_max) / 2`` implies that ``freq0 != (fmin + fmax) / 2``.
+
+        Parameters
+        ----------
+        wvl_min : float
+            The lowest wavelength of interest.
+        wvl_max : float
+            The longest wavelength of interest.
+
+        Returns
+        -------
+        FreqRange
+            An instance of ``FreqRange`` defined by the wavelength interval [``wvl_min``, ``wvl_max``].
+        """
+
+        # convert wavelength intervals to frequency interval
+        fmax = td_const.C_0 / wvl_min
+        fmin = td_const.C_0 / wvl_max
+
+        return cls.from_freq_interval(fmin=fmin, fmax=fmax)
+
+    def freqs(self, num_points: int) -> NDArray[np.float64]:
+        """
+        method ``freqs()`` returns a numpy array of ``num_point`` frequencies uniformly
+        sampled from the specified frequency range;
+        if ``num_points == 1`` method returns the central frequency ``freq0``.
+
+        Parameters
+        ----------
+        num_points : int
+            Number of frequency points in a frequency range of interest.
+
+        Returns
+        -------
+        np.ndarray
+            a numpy array of uniformly distributed frequency samples in a frequency range of interest.
+        """
+
+        if num_points == 1:  # return central frequency
+            return np.array([self.freq0])
+        else:
+            # calculate frequency points and corresponding wavelengths
+            return np.linspace(self.fmin, self.fmax, num_points)
+
+    def ldas(self, num_points: int) -> NDArray[np.float64]:
+        """
+        method ``ldas()`` returns a numpy array of ``num_points`` wavelengths uniformly
+        sampled from the range of wavelengths;
+        if ``num_points == 1`` the method returns central wavelength ``lda0``.
+
+        Parameters
+        ----------
+        num_points : int
+            Number of wavelength points in a range of wavelengths of interest.
+
+        Returns
+        -------
+        np.ndarray
+            a numpy array of uniformly distributed wavelength samples in ascending order.
+        """
+        if num_points == 1:  # return central wavelength
+            return np.array([self.lda0])
+        else:
+            # define shortest and longest wavelengths
+            lmin = td_const.C_0 / self.fmax
+            lmax = td_const.C_0 / self.fmin
+
+            # generate array of wavelengths (in ascending order)
+            return np.linspace(lmin, lmax, num_points)
+
+    def to_gaussian_pulse(self, **kwargs) -> GaussianPulse:
+        """
+        method ``to_gaussian_pulse()`` returns instance of class ``GaussianPulse``
+        with frequency-specific parameters defined in ``FreqRange``.
+
+        Parameters
+        ----------
+        kwargs : dict
+            Keyword arguments passed to ``GaussianPulse()``, excluding ``freq0`` & ``fwidth``.
+
+        Returns
+        -------
+        GaussianPulse
+            A ``GaussianPulse`` that maximizes its amplitude in the frequency range [``fmin``, ``fmax``].
+        """
+
+        duplicate_keys = {"fmin", "fmax"} & kwargs.keys()
+        if duplicate_keys:
+            is_plural = len(duplicate_keys) > 1
+            keys_str = ", ".join(f"'{key}'" for key in sorted(duplicate_keys, reverse=True))
+            raise ValueError(
+                f"Keyword argument{'s' if is_plural else ''} {keys_str} "
+                f"conflict{'' if is_plural else 's'} with values already set in the 'FreqRange' object. "
+                f"Please exclude {'them' if is_plural else 'it'} from the 'to_gaussian_pulse()' call."
+            )
+
+        # create an instance of GaussianPulse class with defined frequency params
+        return GaussianPulse.from_frequency_range(fmin=self.fmin, fmax=self.fmax, **kwargs)
--- a/tests/test_components/test_freq_range.py
+++ b/tests/test_components/test_freq_range.py
@@ -1,0 +1,188 @@
+"""Test class ``FreqRange`` for frequency and wavelength handling."""
+
+from __future__ import annotations
+
+import numpy as np
+
+import tidy3d.constants as td_const
+from tidy3d.components.source.freq_range import FreqRange
+from tidy3d.components.source.time import GaussianPulse
+
+
+def test_constructor():
+    """
+    test construction of ``FreqRange`` from central frequency
+    ``freq0`` and bandwidth ``fwidth``.
+    """
+    # set initial params
+    freq0 = 1e9  # set central frequency
+    fwidth = 1e8  # set one-side bandwidth
+
+    # construct instance of FreqRange class
+    freq_range = FreqRange(freq0=freq0, fwidth=fwidth)
+
+    fmin = freq0 - fwidth
+    fmax = freq0 + fwidth
+
+    lmin = td_const.C_0 / fmax
+    lmax = td_const.C_0 / fmin
+
+    # validated if class atributes were initialized correctly
+    assert freq_range.fmin == fmin
+    assert freq_range.fmax == fmax
+    assert freq_range.lda0 == (lmin + lmax) / 2
+    assert freq_range.freq0 == freq0
+    assert freq_range.fwidth == fwidth
+
+
+def test_from_freq_interval():
+    """
+    test construction of ``FreqRange`` from frequency interval.
+    """
+
+    fmin = 1e10  # new min frequency
+    fmax = 1e11  # new max frequency
+
+    lmin = td_const.C_0 / fmax
+    lmax = td_const.C_0 / fmin
+
+    # update object given new frequencies
+    freq_range = FreqRange.from_freq_interval(fmin=fmin, fmax=fmax)
+
+    # validate if frequencies were updated correctly
+    assert freq_range.fmin == fmin
+    assert freq_range.fmax == fmax
+    assert np.isclose(freq_range.freq0, (fmin + fmax) / 2, rtol=1e-14)
+    assert np.isclose(freq_range.fwidth, (fmax - fmin) / 2, rtol=1e-14)
+    assert np.isclose(freq_range.lda0, (lmin + lmax) / 2, rtol=1e-14)
+
+
+def test_from_wavelength():
+    """
+    test construction of ``FreqRange`` from a central wavelength and a wavelength.
+    """
+    # set initial params
+    wvl0 = 1
+    wvl_width = 0.1
+
+    # get the shortest and the longest wavelengths
+    wvl_min = wvl0 - wvl_width
+    wvl_max = wvl0 + wvl_width
+
+    # define frequency range
+    freq_range = FreqRange.from_wavelength(wvl0=wvl0, wvl_width=wvl_width)
+
+    #
+    assert freq_range.lda0 == wvl0
+
+    fmin = td_const.C_0 / wvl_max
+    fmax = td_const.C_0 / wvl_min
+
+    assert np.isclose(freq_range.fmin, fmin, rtol=1e-14)
+    assert np.isclose(freq_range.fmax, fmax, rtol=1e-14)
+    assert np.isclose(freq_range.freq0, 0.5 * (fmin + fmax), rtol=1e-14)
+    assert np.isclose(freq_range.fwidth, 0.5 * (fmax - fmin), rtol=1e-14)
+
+
+def test_from_wvl_interval():
+    """
+    test construction of ``FreqRange`` from an interval of wavelengths.
+    """
+    # set initial params
+    wvl_min = 0.1
+    wvl_max = 1
+    fmin = td_const.C_0 / wvl_max
+    fmax = td_const.C_0 / wvl_min
+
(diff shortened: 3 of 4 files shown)


tidy3d/freq_range.py
```diff
@@ -0,0 +1,237 @@
...
+        return self.freqs
+    
+    def gaussian_pulse(self):
+        """
+            method gaussian_pulse() returns instance of class GaussianPulse
+            with central frequency freq0 and width of the source frequency fwidth
+        """
+
+        # create an instance of GaussianPulse class with defined frequency params
+        return GaussianPulse(freq0 = self.freq0, fwidth = self.fwidth)
```
momchil-flex: Better to use [from_frequency_range](https://github.com/flexcompute/tidy3d/blob/16607e9985e7ba4664ed5330368f1e7a911be622/tidy3d/components/source/time.py#L189) here because this method ensures a good amount of power in the pulse signal across the frequency range.

tidy3d/freq_range.py
```diff
@@ -0,0 +1,237 @@
...
+            Parameters
+            ----------
+            freq0 - central frequency [Hz]
+            fwidth - one-sided bandwidth [Hz]
+        """
+
+        if not isinstance(freq0, (int, float)):
+            raise TypeError("Central frequency freq0 must be a number (int or float)")
+        if not isinstance(fwidth, (int, float)):
+            raise TypeError("Frequency bandwidth fwidth must be a number (int or float)")
```
momchil-flex: Generally, Tidy3D follows a specific development pattern for all components. Everything is a subclass of `Tidy3dBaseModel`, which is immutable and does not store an internal state. There is also no init method. The input fields are defined as `pydantic` fields (which takes care of the type validation), and there are no private attributes. See e.g. [here](https://github.com/flexcompute/tidy3d/blob/16607e9985e7ba4664ed5330368f1e7a911be622/tidy3d/components/time.py#L21-L30), but any component, really.

So here, you should decide what would be the fundamental input arguments: for example, `freq0` and `fwidth`, and everything else should be `@property` rather than assigned in the `__init__` method. Similarly there would either be no `self.num_points`, or if there is one, then it should be static and should be removed as an argument in the `samples` method.

I think one thing that is missing in the implementation is to be able to have the points either equally spaced in frequency, or equally spaced in wavelength. This could either be an argument passed to the `FreqRange`, or, maybe better, just an argument to `samples`.

momchil-flex (commented):
Hey, this is looking really good, and the tests are thorough!

I do have some comments which are mainly related to the way things work in Tidy3D specifically. You have used a more generic way to define things in python, while our components use `pydantic` to define arguments, and inherit from `Tidy3dBaseModel`.

Regarding using SI units: since everything in Tidy3D uses Tidy3D units, I think so should the `FreqRange` and all its methods.

tests/test_FreqRange.py
tylerflex, line 1: We generally separate these tests into different sub directories. for this test, I suggest putting inside of `tests/test_components/`

tests/test_FreqRange.py
```diff
@@ -0,0 +1,134 @@
+import numpy as np
```
tylerflex: we also generally add a short module level docstring to describe what this file is testing, see other test files for examples.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Implemented `FreqRange` utility class for frequency/wavelength handling; the default constructor and `from_interval()` class method assume, that the central frequency `freq0` is defined to be `freq0 = (fmin + fmax) / 2`; class methods `from_wavelength()` and `from_wavelegth_interval()` assume that the central wavelength `lda0` is defined to be `lda0 = (lda_min + lda_max) / 2`.
```
tylerflex: I think all three of these bullets could be combined into one, potentially. something like 

```
- Utility class `FreqRange` for frequency and wavelength handling. With constructor methods x, y, and z to initialize frequency data.
```

I think the details about the arguments can be left out of the changelog for now since they are part of the API reference already. And we wont need to mention that we added unit tests in the changelog.

tidy3d/freq_range.py
tylerflex, line 1: Please add a module level docstring at the top of the file to briefly describe what is in this module.

tidy3d/freq_range.py
```diff
@@ -0,0 +1,237 @@
...
+from tidy3d.components.source.time import GaussianPulse
+
+
+
+def um_to_m(val):
+    """um_to_m() is a simple function for convertion from [um] to [m]"""
+    return val * 1e-6
+
+
+class FreqRange:
```
tylerflex: as Momchil mentioned, we use pydantic to define classes that store data, which also handles initialization. see this for an example https://github.com/flexcompute/tidy3d/blob/d354c666967fe792340769318952cc494f384927/tidy3d/components/apodization.py#L16

tidy3d/freq_range.py
```diff
@@ -0,0 +1,237 @@
...
+        self.num_points = 1        # set default number of wavelengths/frequencies to 1
+        self.freqs = np.array([self.freq0])    # until num_points is updated set freqs to carrier frequency/central frequency  
+        self.lda0  = um_to_m(td_const.C_0  / self.freq0)  # set central wavelength 
+        self.ldas  = np.array([self.lda0])                   # set wavelength range to central wavelength
+
+    #-----------------------------------
+    #       class methods 
+    #-----------------------------------
+    @classmethod
+    def from_interval(clf, fmin, fmax):
```
tylerflex: please be sure to add type hints to these and other function signatures, eg `fmin: float`

Also, in python we usually call the class `cls`.

tidy3d/freq_range.py
```diff
@@ -0,0 +1,237 @@
...
+        freq0 = um_to_m(td_const.C_0) / wvl0   
+
+        fmin = um_to_m(td_const.C_0) / (wvl0 + wvl_width)
+        fmax = um_to_m(td_const.C_0) / (wvl0 - wvl_width)
+
+        fwidth = 0.5 * (fmax - fmin)
+
+        freq_range = clf(freq0, fwidth)
+
+        freq_range.fmin = fmin
```
tylerflex: these are already set in the init method? or is there a need to overwrite their values here?

tidy3d/freq_range.py
```diff
@@ -0,0 +1,237 @@
...
+
+        else:                   # otherwise 
+
+            if self.fmax <= self.fmin:
+                raise ValueError("fmin has to be larger than fmin")
+            if self.fmin <=0:
+                raise ValueError("frequencies have to be positive values")
+
+            # calculate frequency points and corresponding wavelengths
+            self.freqs = np.linspace(self.fmin, self.fmax, self.num_points)
```
tylerflex: is there a need to store `self.freqs` and `self.ldas`? if `samples` is meant to simply return a numpy array of frequencies, perhaps we shouldn't be modifying self at all?

it might be worth considering a design where instead of `samples(num_points)` we simply expose two methods `FreqRange.freqs(num_points)` and `FreqRange.wavelengths(num_points)`?

tidy3d/freq_range.py
```diff
@@ -0,0 +1,237 @@
...
+            # define shortest and longest wavelengths
+            lmin = um_to_m(td_const.C_0 / self.fmax)
+            lmax = um_to_m(td_const.C_0 / self.fmin) 
+
+            # update array of wavelengths (wavelengths would be in descending order) 
+            self.ldas  = um_to_m(td_const.C_0 / self.freqs)
+
+        return self.freqs
+    
+    def gaussian_pulse(self):
```
tylerflex: what about if I want to pass other arguments that get passed to `GaussianPulse`? such as the amplitude?

tylerflex (changes requested):
Thanks @George-Guryev , looks good overall! A few things we can work on to make the design more in line with our existing design

tylerflex (commented):
please also be sure to add the `FreqRange` class to `tidy3d/__init__.py`, otherwise it wont be able to be imported from the `tidy3d` namespace (eg `import tidy3d as td; td.FreqRange`)

tidy3d/freq_range.py
George-Guryev: Thank you for pointing this out! Fixed it

tidy3d/freq_range.py
George-Guryev: Rewrote class using pydantic. Looks cleaner now. Thank you for walking me through the development pattern, it was helpful!

tests/test_FreqRange.py
George-Guryev: Makes sense, fixed it!

tidy3d/freq_range.py
George-Guryev: Done!

tidy3d/freq_range.py
George-Guryev: Thank you for pointing these issues out. Fixed both

tidy3d/freq_range.py
George-Guryev: There was an issue with an ambiguous definition of what "central wavelength" and "central frequency" were: 

Naturally one interprets central frequency f0 = (fmin + fmax)/ 2 and central wavelength lda0 = (lda_min + lda_max)/2, i.e. both quantities are points in the middle of respective frequency and wavelength ranges.
If we assume that the dispersion equation holds: |k| = omega / c, then wavelength would be inverse-proportional to the corresponding frequency:    lambda  =  c / f.  In this case central wavelength does NOT map to the central frequency (and vice versa). 

However, the first two lines in the ticket #2529 suggest the following definition:

lda0 = 0.8  # central wavelength
freq0 = td.C_0 / lda0  # central frequency

therefore the "central frequency" freq0 that corresponds to lda0 is not a true central frequency (i.e. it does not lie in the center of the frequency range [fmin, fmax].  

### Assumption in first commit

In first commit I decided to stick with the ambiguity in definition of "central frequency" and "central wavelength": if user constructs object with frequency args, freq0 is the true central frequency, while  lda0 is defined as lda0 = c / freq0, but this "central wavelength" is not a true central wavelength;  if user specifies wavelengths - reverse the argument.

To reflect that I had to modify fmin after constructor, as one could not sample frequencies  with fwidth without reference frequency. 

### Assumption in second commit
I was not entirely happy with the solution and decided to change the assumption: 
Instead I used the intuitive definitions for central wavelength and central frequency and did not enforce lda0 = c / freq0. This removes unnecessary ambiguity in definition and the need to reassign properties after object construction

tidy3d/freq_range.py
George-Guryev: That's a great suggestion! I've replaced `samples(num_points)`  with `freqs(num_points)` and `ldas(num_points)`. Looks cleaner now.

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # calculate lowest and highest frequencies
+        fmin = td_const.C_0 / (wvl0 + wvl_width)
+        fmax = td_const.C_0 / (wvl0 - wvl_width)
+
+        # find central frequency and bandwidth
+        freq0  = 0.5 * (fmax + fmin)
+        fwidth = 0.5 * (fmax - fmin)
+
+        return cls(freq0=freq0, fwidth=fwidth)
```
tylerflex: I would prefer maybe to just call `cls.from_interval` using `fmin` and `fmax`, to avoid duplicating operations

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,204 @@
...
+        """ Get central wavelength from central frequency and bandwidth"""
+        lmin = td_const.C_0 / (self.freq0 + self.fwidth)
+        lmax = td_const.C_0 / (self.freq0 - self.fwidth)
+        return 0.5 * (lmin + lmax)
+
+    #-----------------------------------
+    #       class methods 
+    #-----------------------------------
+    @classmethod
+    def from_interval(cls, fmin: float, fmax: float) -> FreqRange:
```
tylerflex: we have `from_interval` and `from_wavelength` but both construct the `FreqRange` out of intervals. Perhaps to unify the naming, we should call these `from_freq_interval` and `from_wvl_interval` or similar?

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,204 @@
...
+        fmax = td_const.C_0 / wvl_min
+        fmin = td_const.C_0 / wvl_max
+
+        freq0 = 0.5 * (fmax + fmin)      # get central frequency
+        fwidth = 0.5 * (fmax - fmin)     # define width of source frequency range
+
+        return cls(freq0=freq0, fwidth=fwidth)
+    
+
+    def freqs(self, num_points: int)->NDArray[np.float64]:
```
tylerflex: nice, I think this api looks a lot cleaner 👍

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,204 @@
...
+            return np.array([self.lda0])
+        else:
+            # define shortest and longest wavelengths
+            lmin = td_const.C_0 / self.fmax
+            lmax = td_const.C_0 / self.fmin 
+
+            # generate array of wavelengths (in ascending order) 
+            return np.linspace(lmin, lmax, num_points)
+    
+    def gaussian_pulse(self, **kwargs)->GaussianPulse:
```
tylerflex: let's call this `to_gaussian_pulse()` (just because internally we tend to use to_ and from_ for these kinds of conversion methods).

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+            method gaussian_pulse() returns instance of class GaussianPulse
+            with frequency-specific parameters defined in 'FreqRange'
+        """
+        # create an instance of GaussianPulse class with defined frequency params
+        return GaussianPulse.from_frequency_range(self.fmin, self.fmax, **kwargs)
```
tylerflex: let's pass things using the keywords, eg (from_frequency_range(fmin=self.fmin, ...) for extra robustness.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Implemented `FreqRange` utility class for frequency/wavelength handling with constructor methods `from_interval()`, `from_wavelength()`, and `from_wavelength_interval()`. 
+- Added methods `freqs(num_points)` and `ldas(num_points)` return numpy arrays of frequency and wavelength points uniformly sampled from the specified frequency range.
```
tylerflex: I think we can just remove these second two bullets, the first should be sufficient. and users can check the API reference in the docs for more details.

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,204 @@
...
+from tidy3d.components.source.time import GaussianPulse
+from tidy3d.components.base import Tidy3dBaseModel
+
+class FreqRange(Tidy3dBaseModel):
+
+    freq0: pydantic.PositiveFloat = pydantic.Field(
+        1.0, title="Central frequency", description="Real-valued positive central frequency", units="Hz")
+    
+    fwidth: pydantic.PositiveFloat = pydantic.Field(
+        1.0, title="Frequency bandwidth", description="Real-valued positive central frequency", units="Hz")
```
tylerflex: description needs updating too.

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+            method from_wavelength() updated instance of class FreqRange by reassigning new 
+            frequency- and wavelength-related parameters;
+
+            NB: central frequency never corresponds to central wavelength!
+            lda0 = (lda_min + lda_max) / 2; it implies that freq0 != (fmin + fmax) / 2 and vise versa
+
+            Input Parameters
+            ----------------
+            wvl0      - central wavelength
```
tylerflex: we'll need to update the docstrings to be consistent with our formatting.

tylerflex (changes requested):
Thanks George!

I'm going to leave my 2nd round of comments here just so they are recorded and I dont lose them. But for the purposes of this exercise, feel free to leave the PR as is (unless you feel compelled to make the changes), we can always pick this up later.

tidy3d/components/source/freq_range.py
George-Guryev: Updated docstrings

tidy3d/components/source/freq_range.py
George-Guryev: Updated description

tidy3d/components/source/freq_range.py
George-Guryev: renamed to ``to_gaussian_pulse()``

tidy3d/components/source/freq_range.py
George-Guryev: Yes, that looks more natural. Renamed class methods to ``from_freq_intervals`` and ``from_wvl_interval``

tidy3d/components/source/freq_range.py
George-Guryev: Makes sense. Removed redundancies

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,202 @@
...
+from tidy3d.components.base import Tidy3dBaseModel
+
+class FreqRange(Tidy3dBaseModel):
+
+    freq0: pydantic.PositiveFloat = pydantic.Field(
+        1.0, title="Central frequency", description="Real-valued positive central frequency", units="Hz")
+    
+    fwidth: pydantic.PositiveFloat = pydantic.Field(
+        1.0, title="Frequency bandwidth", description="Real-valued positive width of the frequency range (bandwidth)", units="Hz") 
+    """
```
tylerflex, line 15: this docstring should go below the `class` definition above I think.

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Implemented `FreqRange` utility class for frequency/wavelength handling with constructor methods `from_freq_interval()`, `from_wavelength()`, and `from_wvl_interval()`.   
+
+## [0.1.0] - 2025-07-02
```
tylerflex: remove. And you can just put your bullet point under the existing ### Added section in the chanbgelog along with all of the most recent added items, eg here

https://github.com/flexcompute/tidy3d/blob/e11b187ba1693654e97f70ba007a96515044d317/CHANGELOG.md?plain=1#L16

might be easiest to just restore the state of `CHANGELOG` in develop and add the single line.

tests/test_components/test_FreqRange.py
```diff
@@ -0,0 +1,186 @@
+"""Test class FreqRange for frequency and wavelength handling"""
```
tylerflex: Make sure any code is in double backslash quotes, just for consistency. more info
https://www.notion.so/flexcompute/Python-Style-Guide-afac5dbbf949463e89c18fe39b7453c9

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,202 @@
+""" Utility class `FreqRange` for frequency and wavelength handling."""
```
tylerflex: please use double backslash quotes here.

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,202 @@
...
+    fwidth: pydantic.PositiveFloat = pydantic.Field(
+        1.0, title="Frequency bandwidth", description="Real-valued positive width of the frequency range (bandwidth)", units="Hz") 
+    """
+        Convenience class for handling frequency/wavelength conversion and assignment
+
+        Example
+        -------
+        >>> freq0  = 1e12
+        >>> fwidth = 1e11
+        >>> freq_range = FreqRange(freq0=freq0, fwidth=fwidth)
```
tylerflex: I think it would be helpful to also add a few lines to the example showing off how you can use `freq_range`. Like very simply calling some of the properties, `freq_range.freqs` for example, or `freq_range.to_gaussian_pulse()`

tylerflex: Similar comment, adding a bit more explanation to the docstring above. not a lot, but just motivate what this class is intended to be used for would be helpful for people stumbling upon this.

tests/test_components/test_FreqRange.py
tylerflex: please check all of the other strings, eg in the Field descriptions

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,202 @@
...
+            -------
+            FreqRange
+                An instance of ``FreqRange`` defined by the wavelength interval [wvl_min, wvl_max]
+        """
+
+        # convert wavelength intervals to frequency interval
+        fmax = td_const.C_0 / wvl_min
+        fmin = td_const.C_0 / wvl_max
+
+        return cls.from_freq_interval(fmin, fmax)
```
tylerflex: it's often times recommended to call functions in python with the argument names too. like `cls.from_freq_interval(fmin=fmin, fmax=fmax)`. it seems verbose, but the reason is that if we add more arguments to these functions, this makes sure that we aren't relying on the order being correct for this function to work (makes it more robust). also in some instances where the variable name is different from the argument name, it can be easier to read. Eg `my_function(x)` is ambiguous whereas calling `my_function(frequency=x)` makes it clear we are passing a frequency.

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,202 @@
...
+            ----------
+            kwargs : dict
+                Keyword arguments passed to ``GaussianPulse()``, excluding ``freq0`` & ``fwidth``.
+
+            Returns
+            -------
+            GaussianPulse
+                A ``GaussianPulse`` that maximizes its amplitude in the frequency range [fmin, fmax].
+        """
+        # create an instance of GaussianPulse class with defined frequency params
```
tylerflex, line 227: let's add a simple check here. raise a `ValueError` if `kwargs` contains `fmin` or `fmax` with a helpful error message. it's a rare edge case, but might be worth catching and just some extra practice.

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+
+        # convert wavelength intervals to frequency interval
+        fmax = td_const.C_0 / wvl_min
+        fmin = td_const.C_0 / wvl_max
+
+        return cls.from_freq_interval(fmin, fmax)
+    
+
+    def freqs(self, num_points: int)->NDArray[np.float64]:
```
tylerflex: @yaugenst-flex what's the recommended return type for numpy arrays these days?

tylerflex (changes requested):
Overall looks good! Mostly formatting and docstring comments, thanks @George-Guryev-flxcmp

tidy3d/components/source/freq_range.py
George-Guryev: Makes sense. Done!

tidy3d/components/source/freq_range.py
George-Guryev: Thank you for walking me through the logic of using argument names! Makes perfect sense.

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
+## Added
```
tylerflex: just FYI we will want this to remain as a level 3 header `###`. it can be helpful to do a final review of the files changed view to just quickly run through and check for these little things.

https://github.com/flexcompute/tidy3d/pull/2542/files

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,217 @@
...
+
+            Parameters
+            ----------
+            num_points : int
+                Number of frequency points in a frequency range of interest.
+
+            Returns
+            -------
+            np.ndarray
+                a numpy array of uniformly distributed frequency samples in a frequency range of interest
```
tylerflex: please do one last readtthough and make sure that there are periods at the end of the field and argument description sentences like this one.

(I know it's super tedious :) , soon we should have an AI bot that will actually catch these for us!)

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,217 @@
...
+            kwargs : dict
+                Keyword arguments passed to ``GaussianPulse()``, excluding ``freq0`` & ``fwidth``.
+
+            Returns
+            -------
+            GaussianPulse
+                A ``GaussianPulse`` that maximizes its amplitude in the frequency range [``fmin``, ``fmax``].
+        """
+
+        if {"fmin", "fmax"} & kwargs.keys():
```
tylerflex: a little nicer might be

```py
duplicate_keys = {"fmin", "fmax"} & kwargs.keys()
if duplicate_keys:
    raise ValueError(f"was passed keyword arguments {duplicate_keys} that are overwritten by the values in this 'FreqRange' object. Please exclude these from the 'to_gaussian_pulse()' call.")
```
or something to this effect. This way the users see which fields were specifically passed.

tylerflex (changes requested):
Looking almost ready! just a few minor comments again, thanks @George-Guryev

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,224 @@
...
+            Returns
+            -------
+            GaussianPulse
+                A ``GaussianPulse`` that maximizes its amplitude in the frequency range [``fmin``, ``fmax``].
+        """
+
+        duplicate_keys = {"fmin", "fmax"} & kwargs.keys()
+        if duplicate_keys:
+            is_plural = len(duplicate_keys) > 1
+            keys_str  = ", ".join(sorted(duplicate_keys, reverse=True))
```
tylerflex: very minor point but let's take sure the keys have single quotes around them

tylerflex (approved):
Hi @George-Guryev in terms of content, looks good to me! just a really really minor formatting thing.

Make sure you squash commits to one and wait for all tests to finish. can coordinate with @dbochkov-flexcompute for the final merge.

Thanks a lot!

tidy3d/components/source/freq_range.py
George-Guryev: Thanks for pointing that out! I've updated the message to enclose the keys in single quotes as recommended.

tidy3d/components/source/freq_range.py
```diff
@@ -0,0 +1,224 @@
...
+        >>> field_mnt  = td.FieldMonitor(..., freqs=freq_range.freqs(num_points=1))
+        >>> mode_mnt   = td.ModeMonitor(..., freqs=freq_range.freqs(num_points=11))
+        >>> source     = td.PlaneWave(..., source_time=freq_range.to_gaussian_pulse())
+    """
+
+    freq0: pydantic.PositiveFloat = pydantic.Field(
+        1.0, title="Central frequency", description="Real-valued positive central frequency.", units="Hz")
+    
+    fwidth: pydantic.PositiveFloat = pydantic.Field(
+        1.0, title="Frequency bandwidth", description="Real-valued positive width of the frequency range (bandwidth).", units="Hz")
```
dbochkov-flexcompute: One small request from me: let's make `freq0` and `fwidth` mandatory to specify, i.e. replace `1.0` with `...` for both fields. So that the user doesn't accidentally do `freq_range = td.FreqRange(freq=1e11)`, which would create a super narrow range

George-Guryev: Makes sense! Replaced the defaults.

PR #2547, opened 2025-06-06
Setting default mode precision to 'double', except in EME solver

### LINKED ISSUE #2518: Automatically sets `ModeSpec` precision to `double` when `group_index_step` is not `False`
This is a suggestion, but I think would be good to automatically set the precision to double  when the the group index is calculated. The single precision usually generate noise data, specially for the group index dispersion. Altough it is straightforward to orient the user on using double precision, I think it may cause a bad impression for users using Tidy3D for the first time. A user report this using single precision:

Image

Just changing to double precision it gets much better:

Image

The dispersion gets even better tuning the `group_step_index`.

If, for any reason, this option isn’t convenient, I can think of ways to make it clearer in the documentation.

I’d consider this a low-priority suggestion.

momchil-flex:
Fixes #2518 (well, and it goes beyond it to set 'double' by default)

@caseyflex note that I changed the `EMEModeSpec_to_mode_spec` method to be more stable if we e.g. add a new field in the `ModeSpec` that we just want to directly propagate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Upon initialization, an FDTD `Simulation` will now try to create all `ModeSolver` objects associated to `ModeSource`-s and `ModeMonitor`-s so they can be validated.
 - `tidy3d.plugins.autograd.interpolate_spline()` and `tidy3d.plugins.autograd.add_at()` can now be called with keyword arguments during tracing.
 - Zero-size dimensions automatically receive periodic boundary conditions instead of raising an error.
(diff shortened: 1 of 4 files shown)


tomflexcompute (approved):

caseyflex (approved):
Great! Yeah we have some other functions like `_to_mode_spec` that could be made more robust like this in the future. Like in `mode/simulation.py`; we have `MODE_SIM_YEE_SIM_SHARED_ATTRS` which should probably be handled like this, just popping the properties that are specific to `ModeSimulation`. `MODE_SIM_MODE_SOLVER_SHARED_ATTRS` probably can't be handled like this, since those classes aren't related by simple inheritance

FilipeFcp (approved):

weiliangjin2021 (approved):

PR #2553, opened 2025-06-06
Automatically use periodic boundary on zero-size dimensions

caseyflex:
This is mainly to address the annoying issue with `ModeSimulation`, where if it is 2D, the boundary must be manually set to periodic in the normal direction. It replaces some of this PR for 1D mode solving (https://github.com/flexcompute/tidy3d/pull/2478), so that we can get it in before all the issues are resolved there.

### DIFF
--- a/tidy3d/components/mode/simulation.py
+++ b/tidy3d/components/mode/simulation.py
@@ -8,7 +8,7 @@ import numpy as np
 import pydantic.v1 as pd
 
 from tidy3d.components.base import cached_property
-from tidy3d.components.boundary import Boundary, BoundarySpec
+from tidy3d.components.boundary import BoundarySpec
 from tidy3d.components.geometry.base import Box
 from tidy3d.components.grid.grid import Grid
 from tidy3d.components.grid.grid_spec import GridSpec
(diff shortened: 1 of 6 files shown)


caseyflex:
@momchil-flex @yaugenst-flex 

The boundary is silently replaced (per @e-g-melo 's suggestion) for EMESimulation and ModeSimulation. For FDTD Simulation, it gives a warning. Previously, it would error for all simulation types if the boundary was not periodic or bloch.

caseyflex:
@momchil-flex let's try to get this in 2.9.0rc1?

tidy3d/components/mode/simulation.py
```diff
@@ -213,18 +213,6 @@ def plane_in_sim_bounds(cls, val, values):
...
-    def boundaries_for_zero_dims(cls, val, values):
-        """Replace with periodic boundary along zero-size dimensions."""
-        boundaries = [val.x, val.y, val.z]
-        size = values.get("size")
-
-        for dim, size_dim in enumerate(size):
-            if size_dim == 0:
-                boundaries[dim] = Boundary.periodic()
-
-        return BoundarySpec(x=boundaries[0], y=boundaries[1], z=boundaries[2])
```
momchil-flex, line 226: Ah so this validator actually existed but it wasn't taking effect because `_boundaries_for_zero_dims = validate_boundaries_for_zero_dims()` was also included (and probably executed first?)

momchil-flex (approved):
Makes sense, I tested locally and it works, merging now so I can test rc1.

tidy3d/components/mode/simulation.py
caseyflex: yeah I think that's right

PR #2557, opened 2025-06-10
account for weighting of adjoint field source current when simulation uses symmetry

groberts-flex:
This accounts for when the forward simulation has symmetry and we are creating adjoint field sources. The overall normalization is affected depending on the location of the source with respect to the symmetry axes.

We also warn in the case where we are splitting up adjoint simulations and might be breaking the forward symmetry.

For symmetry, the adjoint gradient isn't exactly the same in both cases with and without symmetry I believe due to the padding around the source and slight differences in the mesh normalization depending on if the # of mesh cells is even or odd.  The results of the tests are below for the two figure of merit types in the unit test.

Test 1 (intensity):
adj_test1_symm_x
adj_test1_symm_y
adj_test1_symm_xy

Test 2 (flux):
adj_test2_symm_x
adj_test2_symm_y
adj_test2_symm_xy

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -1358,6 +1358,12 @@ class FieldData(FieldDataset, ElectromagneticFieldData):
                 field_component = field_component.sel(f=freq0)
                 values = 2 * -1j * field_component.values
 
+                # accounts for the effective size of the source when injecting into a
+                # simulation with symmetry
+                symmetry_factor = np.prod(values.shape) / np.prod(
+                    self.symmetry_expanded_copy.field_components[name].sel(f=freq0).values.shape
+                )
+
                 # make source go backwards
                 if "H" in name:
                     values *= -1
@@ -1379,7 +1385,8 @@ class FieldData(FieldDataset, ElectromagneticFieldData):
                 omega0 = 2 * np.pi * freq0
                 scaling_factor = 0.5 * omega0 * EPSILON_0 / size_element
 
-                values *= scaling_factor
+                values *= scaling_factor * symmetry_factor
+                values = np.nan_to_num(values, nan=0.0)
 
                 # ignore zero components
                 if not np.all(values == 0):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Internal interpolation errors with some versions of `xarray` and `numpy`.
 - If `ModeSpec.angle_rotation=True` for a mode object, validate that the structure rotation can be successfully done. Also, error if the medium cannot be rotated (e.g. anisotropic or custom medium), which would previously have just produced wrong results.
 - Characteristic impedance calculations in the `ImpedanceCalculator` using definitions that rely on flux, which were giving incorrect results for lossy transmission lines.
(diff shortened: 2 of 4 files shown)


groberts-flex:
The numerical results of the test:

Testing intensity objective
Symmetry comparison: none, x
RMS error (normalized): 0.023567554375240524

Testing intensity objective
Symmetry comparison: none, y
RMS error (normalized): 0.06464612699977461

Testing intensity objective
Symmetry comparison: none, xy
RMS error (normalized): 0.05452025029584866

.Testing flux objective
Symmetry comparison: none, x
RMS error (normalized): 0.02154225911715167

Testing flux objective
Symmetry comparison: none, y
RMS error (normalized): 0.021522638860377995

Testing flux objective
Symmetry comparison: none, xy
RMS error (normalized): 0.0430668572667713

tidy3d/components/data/monitor_data.py
```diff
@@ -1358,6 +1358,12 @@ def _make_adjoint_sources(
                 field_component = field_component.sel(f=freq0)
                 values = 2 * -1j * field_component.values
 
+                # accounts for the effective size of the source when injecting into a
+                # simulation with symmetry
+                symmetry_factor = np.prod(values.shape) / np.prod(
+                    self.symmetry_expanded_copy.field_components[name].sel(f=freq0).values.shape
+                )
```
momchil-flex, line 1365: Is this trying to account for both cases in which the monitor is either sitting on a symmetry plane or away from it? Would this work if it is only partially sitting on it as opposed to either being centered exactly on the axis, or being completely outside of it?

groberts-flex: this is accounting for either overlapping the axis or away from it (i.e. - symmetric about it but not crossing the axis). I think if it's a symmetric setup, we don't really have cases where there is more monitor on one side of the symmetry than the other. I guess this comes down to what happens in the symmetry expansion function when we have an asymmetric monitor with respect to a symmetric axis?

momchil-flex: Yeah so this is not allowed for a ModeMonitor but it is allowed for a FieldMonitor. I believe in that case, the symmetry_expanded_data would still be exactly the same as if you ran the monitor without the symmetry imposed.

Not sure how deep we want to go into this. One option could be to try to detect if such a partial overlap is the case and then error...

For the scale factor, due to the fact that the shape might indeed not be exactly a factor of 2 larger, it could make more sense to define it as something like `2 ** np.prod(np.sum(self.symmetry != 0))`. But actually more checks would be needed as this would only apply to dimensions along which the monitor is not size 0, and only when it sits on the symmetry center... so maybe the shortcut you've taken is fine.

groberts-flex: I see - so the symmetry expanded monitor in that case wouldn't just be twice the size of the non symmetry expanded field data?

I see, yeah I considered manually calculating the size difference as you suggested, but with the additional corner cases I figured it would lead to simpler code to just use the ratio of data sizes. I think worst case is it is off by one mesh cell and in the unit test, we only have a few percent error. For adjoint, at least for now, a few percent off in gradient should be ok especially since it look like in most of the error plots, it is mostly a small overall scale factor rather than a directional error.

momchil-flex (approved):
Ok well this looks good for me to merge right now since it's obviously an improvement.

Issue #2545, opened 2025-06-06
upgrade inverse design FAQ to refer to autograd not jax


tylerflex:
https://docs.flexcompute.com/projects/tidy3d/en/v2.7.8/faq/docs/inverse-design.html

tomflexcompute:
@FilipeFcp would you be able to work on this?

FilipeFcp:
Sure.

yaugenst-flex:
Completed in https://github.com/flexcompute/tidy3d-faq/pull/35

PR #2559, opened 2025-06-10
Jax plugin deprecation

momchil-flex:
I took most of the code from the pydantic v2 PR but just made it a warning for now.

@tylerflex also when should we do [this](https://github.com/flexcompute/tidy3d/issues/2545)?


tylerflex (approved):
looks good. since this doesn't break anything we can just roll it out for 2.9.0rc1?

PR #2563, opened 2025-06-11
add grid refinement option and validation for WavePorts

dmarek-flex:
Improvement for issue reported [here](https://flow360.atlassian.net/browse/SCRF-715?atlOrigin=eyJpIjoiNjYwNTUwODEzYWM4NGZlN2I0YWE0MGY3M2I5N2UwNTYiLCJwIjoiaiJ9).

Added grid refinement around WavePorts that defaults to a small number of grid cells. Also a validation step to make sure that there is a sufficient number of grid cells across WavePorts.

## Greptile Summary

Added grid refinement capabilities to WavePorts to ensure accurate mode calculations, with built-in validation for grid resolution requirements.

- Added `num_grid_cells` parameter to `WavePort` in `plugins/smatrix/ports/wave.py` defaulting to 5 cells (minimum 3) for transverse plane refinement
- Implemented `to_mesh_overrides()` method in WavePort class to generate mesh refinement structures
- Added validation in `plugins/smatrix/component_modelers/terminal.py` to ensure sufficient grid cells across WavePorts
- Extended test coverage in `test_terminal_component_modeler.py` to verify grid refinement functionality and validation

### DIFF
--- a/tidy3d/plugins/smatrix/ports/wave.py
+++ b/tidy3d/plugins/smatrix/ports/wave.py
@@ -18,13 +18,17 @@ from tidy3d.components.monitor import ModeMonitor
 from tidy3d.components.simulation import Simulation
 from tidy3d.components.source.field import ModeSource, ModeSpec
 from tidy3d.components.source.time import GaussianPulse
-from tidy3d.components.types import Direction, FreqArray
+from tidy3d.components.structure import MeshOverrideStructure
+from tidy3d.components.types import Axis, Direction, FreqArray
 from tidy3d.constants import fp_eps
 from tidy3d.exceptions import ValidationError
 from tidy3d.plugins.microwave import CurrentIntegralTypes, ImpedanceCalculator, VoltageIntegralTypes
 from tidy3d.plugins.mode import ModeSolver
 
 from .base_terminal import AbstractTerminalPort
+
+DEFAULT_WAVE_PORT_NUM_CELLS = 5
+MIN_WAVE_PORT_NUM_CELLS = 3
 
 
 class WavePort(AbstractTerminalPort, Box):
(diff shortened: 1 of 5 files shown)


tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -63,6 +67,14 @@ class WavePort(AbstractTerminalPort, Box):
         description="Definition of current integral used to compute current and the characteristic impedance.",
     )
 
+    num_grid_cells: Optional[pd.PositiveInt] = pd.Field(
+        DEFAULT_WAVE_PORT_NUM_CELLS,
+        title="Number of Grid Cells",
+        description="Number of mesh grid cells in the transverse plane of the `WavePort`. "
+        "Used in generating the suggested list of :class:`.MeshOverrideStructure` objects. "
+        "Must be greater than or equal to 3. When set to `None`, no grid refinement is performed.",
+    )
```
dbochkov-flexcompute: seems like now all derivatives of `AbstractTerminalPort` have field `num_grid_cells`? Maybe it would make sense to move into `AbstractTerminalPort`?

tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -227,3 +264,17 @@ def validate_current_integral_sign(cls, val, values):
                 f"'current_integral' sign must match the '{name}' direction '{direction}'."
             )
         return val
+
+    @pd.validator("num_grid_cells", always=True)
+    def _validate_num_grid_cells(cls, val):
```
dbochkov-flexcompute: you can use parameter `ge=` in pydantic field to do this validation automatically (like here https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/medium.py#L1806)

tidy3d/plugins/smatrix/ports/wave.py
dmarek-flex: Yea it could, but the default number of cells needs to be different, so I think I need to redefine the field anyways?

tidy3d/plugins/smatrix/ports/wave.py
dmarek-flex: In this case, not sure how to get it to work, since the field is using `Optional` , pydantic is giving errors when I try using `ge=` unless I switch it to a plain `int`

yaugenst-flex:
@greptileai

tidy3d/plugins/smatrix/ports/wave.py
dbochkov-flexcompute: ah, I see

tidy3d/plugins/smatrix/ports/wave.py
dbochkov-flexcompute: you mean `num_grid_cells: Optional[int]` or `num_grid_cells: int`? If the former than I think it's fine as int + `ge=MIN_WAVE_PORT_NUM_CELLS` is stricter than `PositiveInt`

dmarek-flex: Ah ok I thought `num_grid_cells: Optional[int]` was not working, but now it seems to work.

dbochkov-flexcompute (approved):

PR #2360, opened 2025-04-04
Surface monitors

dbochkov-flexcompute:
see https://github.com/flexcompute/compute/pull/1239

PR #2566, opened 2025-06-12
Electric field monitor for Charge

marc-flex:
## Greptile Summary

Adds electric field monitoring capabilities to the Charge simulation module, enabling visualization and analysis of steady-state electric fields with support for both tetrahedral and triangular mesh grids.

- Added new `SteadyElectricFieldMonitor` class with Ex, Ey, Ez component monitoring and proper unit handling (V/μm)
- Implemented robust VTK file support with `ignore_invalid_cells` parameter and cell-to-point data conversion for better visualization
- Reorganized mesh-related documentation and APIs under a dedicated `/mesh` section for improved code organization
- Enhanced `VolumeMesher` with mesh refinement capabilities and grid validation for wave ports
- Improved boundary layer validation with new MIN_NUM_* constants and centralized warning system

### DIFF
--- a/tidy3d/components/data/data_array.py
+++ b/tidy3d/components/data/data_array.py
@@ -1175,9 +1175,9 @@ class SteadyVoltageDataArray(DataArray):
 
 
 class PointDataArray(DataArray):
-    """A two-dimensional array that stores coordinates of a collection of points.
+    """A two-dimensional array that stores coordinates/field components for a collection of points.
     Dimension ``index`` denotes the index of a point in the collection, and dimension ``axis``
-    denotes the point's coordinate along that axis.
+    denotes the field component (or point coordinate) in that direction.
 
     Example
     -------
@@ -1188,6 +1188,14 @@ class PointDataArray(DataArray):
     >>> point3 = point_array.sel(index=3)
     >>> # get x coordinates of all points
     >>> x_coords = point_array.sel(axis=0)
+    >>>
+    >>> field_da = PointDataArray(
+    ...     np.random.random((120, 3)), coords=dict(index=np.arange(120), axis=np.arange(3)),
+    ... )
+    >>> # get field of point number 90
+    >>> field_point90 = field_da.sel(index=90)
+    >>> # get z component of all points
+    >>> z_field = field_da.sel(axis=2)
     """
 
     __slots__ = ()
@@ -1265,6 +1273,20 @@ class IndexedTimeDataArray(DataArray):
     _dims = ("index", "t")
 
 
+class IndexedFieldVoltageDataArray(DataArray):
+    """Stores indexed values of vector fields for different voltages. It is typically used
+    in conjuction with a ``PointDataArray`` to store point-associated vector data.
+    Example
+    -------
+    >>> indexed_array = IndexedFieldVoltageDataArray(
+    ...     (1+1j) * np.random.random((4,3,2)), coords=dict(index=np.arange(4), axis=np.arange(3), voltage=[-1, 1])
+    ... )
+    """
+
+    __slots__ = ()
+    _dims = ("index", "axis", "voltage")
+
+
 class SpatialVoltageDataArray(AbstractSpatialDataArray):
     """Spatial distribution with voltage mapping.
 
@@ -1319,6 +1341,7 @@ DATA_ARRAY_TYPES = [
     PointDataArray,
     CellDataArray,
     IndexedDataArray,
+    IndexedFieldVoltageDataArray,
     IndexedVoltageDataArray,
     SpatialVoltageDataArray,
     PerturbationCoefficientDataArray,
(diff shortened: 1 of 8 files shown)


dbochkov-flexcompute:
Looks good overall, but I wanted to discuss whether we should store Ex, Ey, and Ez separately or in a single Triangular/Tetrahedral-GridDataset. In FDTD field monitors we do store them separately, but I guess one big reason to do that there is that Ex, Ey, and Ez can live on different grids (in fact, it was default up until some point). Moreover, storing cartesian grid info typically doesn't require much memory in comparison to actual field data. For unstructured datasets storing grid info does actually take more memory that a single field, and it seems like at least in this PR Ex, Ey, and Ez are stored at the same locations. Do you see any other arguments towards storing them separately or as a single xarray?

In my surface monitor PR I decided to go the single-array route and use indexed data array with additional coord `"axis"` https://github.com/flexcompute/tidy3d/pull/2360/files#diff-834da9ec0c6f019fc8dde7c9b3cd1d2eafb780dadfadb21e94c0a313be526076R1256-R1285. One other smaller advantage is that it makes calculating the magnitude of vector field and producing a new unstructured dataset for that very straightforward https://github.com/flexcompute/tidy3d/blob/2116575a6b13e23d97d6352d2143645e9ea29f7a/tidy3d/components/data/unstructured/base.py#L410-L412

marc-flex:
Yes, that makes sense! Will do that. Thanks @dbochkov-flexcompute

marc-flex:
@dbochkov-flexcompute  I just realized that we're already storing multiple solution in each of the `Ex, Ey, Ez` fields since it can have multiple voltages. Though currently it only stores one voltage in Conduction, it can store multiple in Charge. I'm wondering what approach we should follow here. Add another index to the data so that we have `voltage` and `component`?

dbochkov-flexcompute:
> @dbochkov-flexcompute I just realized that we're already storing multiple solution in each of the `Ex, Ey, Ez` fields since it can have multiple voltages. Though currently it only stores one voltage in Conduction, it can store multiple in Charge. I'm wondering what approach we should follow here. Add another index to the data so that we have `voltage` and `component`?

yeah, I think that would make sense

marc-flex:
@dbochkov-flexcompute I updated this to use the same data array class you mentioned. There will probably be conflict when either this or your branch gets merged though but it should be pretty straightforward.

tidy3d/components/data/data_array.py
```diff
@@ -1265,6 +1265,39 @@ class IndexedTimeDataArray(DataArray):
     _dims = ("index", "t")
 
 
+class IndexedFieldDataArray(PointDataArray):
```
momchil-flex: So this looks identical to a `PointDataArray` from a functional perspective. Did you just introduce it so that we can distinguish between the use cases depending on the name, but in principle a `PointDataArray` would also work? Would we want e.g. a single `IndexedDataArray` as a class that covers both?

tidy3d/components/tcad/data/monitor_data/charge.py
```diff
@@ -460,3 +463,82 @@ def symmetry_expanded_copy(self) -> SteadyCapacitanceData:
...
+
+        mnt = values.get("monitor")
+        E = values.get("E")
+
+        if isinstance(E, TetrahedralGridDataset) or isinstance(E, TriangularGridDataset):
+            AcceptedTypes = (IndexedFieldVoltageDataArray, IndexedFieldDataArray)
+            if not isinstance(E.values, AcceptedTypes):
+                raise ValueError(
+                    f"In the data associated with monitor {mnt}, the field E does not contain "
+                    "data associated to any voltage value."
```
momchil-flex: Is this message right? It seems like `IndexedFieldDataArray` is allowed and that object is not associated to any voltage value.

momchil-flex (commented):
Thanks. Minor comments, also wondering what is your plan for the related current density monitor?

tidy3d/components/data/data_array.py
marc-flex: That right. Here I'm only adding a new name that makes more sense for fields. 

I also like your suggestion of renaming it to `IndexedDataArray` but this may break backwards compatibility? @dbochkov-flexcompute any thoughts?

marc-flex: I could also just have our field stored as `PointDataArray` just thought having the word Field in there would have made sense

tidy3d/components/tcad/data/monitor_data/charge.py
marc-flex: Oops! Yeah I need to update the comment!

tidy3d/components/data/data_array.py
dbochkov-flexcompute: yeah, `IndexedDataArray` or `IndexedVectorDataArray` would have been better in hindsight. From backward compatibility, we do have to keep `PointDataArray`. So, it's either just using `PointDataArray` for indexed vector arrays, or add a second class like you did here with the intention of retiring `PointDataArray` in future. 

I think either is fine, unless we want eventually to maybe come up with a different system for this kinds of types? It seems like with introducing eme, heat, charge, etc we have an explosion of them, which may or may not be sustainable

marc-flex: OK so in an attempt to keep backwards compatibility and saving us from having another data type, I'll use `PointDataArray` but I'll add a little more documentation

momchil-flex (approved):

PR #2572, opened 2025-06-13
autograd compatible s-matrix calculation

tylerflex:
Objective functions involving s-matrix calculations using `tidy3d.plugins.smatrix.ComponentModeler` are now differentiable with autograd.

Here is an example:

```py
import autograd as ag
import autograd.numpy as np
from tidy3d.plugins.smatrix import ComponentModeler, Port

port_in = Port(
    size=sim.sources[0].size,
    center=sim.sources[0].center,
    direction='+',
    name='in',
)

port_out = Port(
    size=sim.monitors[0].size,
    center=sim.monitors[0].center,
    direction='-',
    name='out',
)

modeler = ComponentModeler(
    simulation=sim.updated_copy(monitors=[], sources=[]),
    ports=[port_in, port_out],
    freqs=[freq0]
)

# ...

def power(center: tuple, size: tuple, eps: float) -> float:
    """Compute power transmitted into 0th order mode given a set of scatterer parameters."""
    sim = make_simulation(center=center, size=size, eps=eps)
    traced_modeler = modeler.updated_copy(simulation=sim.updated_copy(monitors=[], sources=[]))
    smatrix = traced_modeler.run()
    amp = smatrix.sel(port_in='in', port_out='out')
    return np.sum(np.abs(amp)**2).item()

grad = ag.grad(power)(...)
```

## Greptile Summary

Added autograd compatibility to S-matrix calculations in `ComponentModeler`, enabling gradient-based optimization for photonic component design through automatic differentiation.

- Modified `plugins/smatrix/component_modelers/base.py` to use `autograd.numpy` and `run_async` instead of non-differentiable `batch.run`
- Updated `plugins/smatrix/component_modelers/modal.py` to use mask-based indexing for S-matrix construction instead of `loc` indexing
- Added example in `plugins/autograd/README.md` demonstrating gradient computation for power transmission objectives
- Documented feature in CHANGELOG.md under [Unreleased] section

### DIFF
--- a/tests/test_data/test_data_arrays.py
+++ b/tests/test_data/test_data_arrays.py
@@ -4,9 +4,11 @@ from __future__ import annotations
 
 from typing import Optional
 
-import numpy as np
+import autograd as ag
+import autograd.numpy as np
 import pytest
 import xarray.testing as xrt
+from autograd.test_util import check_grads
 
 import tidy3d as td
 from tidy3d.exceptions import DataError
@@ -468,3 +470,49 @@ def test_interp(method, scalar_index):
     xr_interp = data.interp(f=f)
     ag_interp = data._ag_interp(f=f)
     xrt.assert_allclose(xr_interp, ag_interp)
+
+
+def test_with_updated_data_grad():
+    """Check the ``DataArray.with_updated_data()`` method."""
+
+    arr = td.SpatialDataArray(
+        np.ones((2, 3, 4, 5), dtype=np.complex128),
+        coords={"x": [0, 1], "y": [1, 2, 3], "z": [2, 3, 4, 5], "w": [0, 1, 2, 3, 4]},
+    )
+
+    data = np.zeros((1, 1, 1, 5))
+
+    coords = {"x": 0, "y": 2, "z": 3}
+
+    arr2 = arr._with_updated_data(data=data, coords=coords)
+
+    data_expected = np.ones(arr.shape) + 0j
+    data_expected[0, 1, 1, :] = 0.0 + 0j
+    assert np.all(arr2.data == data_expected), "DataArray.with_updated_copy() failed"
+
+    def f(x):
+        arr2 = arr._with_updated_data(data=x, coords=coords)
+        return np.abs(np.sum(arr2.data))
+
+    # grad should just be all 1s because of sum, so check that this is true
+    g = ag.grad(f)(data)
+    assert np.all(g == np.ones_like(data))
+
+    check_grads(f, order=1, modes=["rev"])(data)
+
+
+def test_with_updated_data_shape():
+    """Check the ``DataArray.with_updated_data()`` method."""
+
+    arr = td.SpatialDataArray(
+        np.ones((2, 3, 4, 5), dtype=np.complex128),
+        coords={"x": [0, 1], "y": [1, 2, 3], "z": [2, 3, 4, 5], "w": [0, 1, 2, 3, 4]},
+    )
+
+    # wrong shape
+    data = np.zeros((1, 1, 1, 3))
+
+    coords = {"x": 0, "y": 2, "z": 3}
+
+    with pytest.raises(ValueError):
+        arr2 = arr._with_updated_data(data=data, coords=coords)
--- a/tidy3d/components/data/data_array.py
+++ b/tidy3d/components/data/data_array.py
@@ -489,6 +489,40 @@ class DataArray(xr.DataArray):
                 result = result.transpose(*out_dims)
         return result
 
+    def _with_updated_data(self, data: np.ndarray, coords: dict[str, Any]) -> DataArray:
+        """Make copy of ``DataArray`` with ``data`` at specified ``coords``, autograd compatible
+
+        Constraints / Edge cases:
+            - `coords` must map to a specific value eg {x: '1'}, does not broadcast to arrays
+            - `data` will be reshaped to try to match `self.shape` except where `coords` present
+        """
+
+        # make mask
+        mask = xr.zeros_like(self, dtype=bool)
+        mask.loc[coords] = True
+
+        # reshape `data` to line up with `self.dims`, with shape of 1 along the selected axis
+        old_data = self.data
+        new_shape = list(old_data.shape)
+        for i, dim in enumerate(self.dims):
+            if dim in coords:
+                new_shape[i] = 1
+        try:
+            new_data = data.reshape(new_shape)
+        except ValueError as e:
+            raise ValueError(
+                "Couldn't reshape the supplied 'data' to update 'DataArray'. The provided data was "
+                f"of shape {data.shape} and tried to reshape to {new_shape}. If you encounter this "
+                "error please raise an issue on the tidy3d github repository with the context."
+            ) from e
+
+        # broadcast data to repeat data along the selected dimensions to match mask
+        new_data = new_data + np.zeros_like(old_data)
+
+        new_data = np.where(mask, new_data, old_data)
+
+        return self.copy(deep=True, data=new_data)
+
 
 class FreqDataArray(DataArray):
     """Frequency-domain array.
--- a/tidy3d/plugins/smatrix/component_modelers/base.py
+++ b/tidy3d/plugins/smatrix/component_modelers/base.py
@@ -6,7 +6,7 @@ import os
 from abc import ABC, abstractmethod
 from typing import Generic, Optional, TypeVar, Union, get_args
 
-import numpy as np
+import autograd.numpy as np
 import pydantic.v1 as pd
 
 from tidy3d.components.base import Tidy3dBaseModel, cached_property
@@ -23,11 +23,15 @@ from tidy3d.plugins.smatrix.ports.coaxial_lumped import CoaxialLumpedPort
 from tidy3d.plugins.smatrix.ports.modal import Port
 from tidy3d.plugins.smatrix.ports.rectangular_lumped import LumpedPort
 from tidy3d.plugins.smatrix.ports.wave import WavePort
+from tidy3d.web import run_async
 from tidy3d.web.api.container import Batch, BatchData
 
 # fwidth of gaussian pulse in units of central frequency
 FWIDTH_FRAC = 1.0 / 10
 DEFAULT_DATA_DIR = "."
+
+# whether to run gradient calculation for component modeler locally
+LOCAL_GRADIENT = False
 
 LumpedPortType = Union[LumpedPort, CoaxialLumpedPort]
 TerminalPortType = Union[LumpedPortType, WavePort]
@@ -251,7 +255,26 @@ class AbstractComponentModeler(ABC, Generic[IndexType, ElementType], Tidy3dBaseM
     @cached_property
     def batch_data(self) -> BatchData:
         """The :class:`.BatchData` associated with the simulations run for this component modeler."""
-        return self.batch.run(path_dir=self.path_dir)
+
+        # NOTE: uses run_async because Batch is not differentiable.
+        batch = self.batch
+        run_async_kwargs = batch.dict(
+            exclude={
+                "type",
+                "path_dir",
+                "attrs",
+                "solver_version",
+                "jobs_cached",
+                "num_workers",
+                "simulations",
+            }
+        )
+        return run_async(
+            batch.simulations,
+            **run_async_kwargs,
+            local_gradient=LOCAL_GRADIENT,
+            path_dir=self.path_dir,
+        )
 
     def get_path_dir(self, path_dir: str) -> None:
         """Check whether the supplied 'path_dir' matches the internal field value."""
--- a/tidy3d/plugins/smatrix/component_modelers/modal.py
+++ b/tidy3d/plugins/smatrix/component_modelers/modal.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from typing import Optional
 
-import numpy as np
+import autograd.numpy as np
 import pydantic.v1 as pd
 
 from tidy3d.components.base import cached_property
@@ -237,14 +237,15 @@ class ComponentModeler(AbstractComponentModeler[MatrixIndex, Element]):
                 )
                 source_norm = self._normalization_factor(port_in, sim_data)
                 s_matrix_elements = np.array(amp.data) / np.array(source_norm)
-                s_matrix.loc[
-                    {
-                        "port_in": port_name_in,
-                        "mode_index_in": mode_index_in,
-                        "port_out": port_name_out,
-                        "mode_index_out": mode_index_out,
-                    }
-                ] = s_matrix_elements
+
+                coords_set = {
+                    "port_in": port_name_in,
+                    "mode_index_in": mode_index_in,
+                    "port_out": port_name_out,
+                    "mode_index_out": mode_index_out,
+                }
+
+                s_matrix = s_matrix._with_updated_data(data=s_matrix_elements, coords=coords_set)
 
         # element can be determined by user-defined mapping
         for (row_in, col_in), (row_out, col_out), mult_by in self.element_mappings:
@@ -259,12 +260,14 @@ class ComponentModeler(AbstractComponentModeler[MatrixIndex, Element]):
 
             port_out_to, mode_index_out_to = row_out
             port_in_to, mode_index_in_to = col_out
+
+            elements_from = mult_by * s_matrix.loc[coords_from].values
             coords_to = {
                 "port_in": port_in_to,
                 "mode_index_in": mode_index_in_to,
                 "port_out": port_out_to,
                 "mode_index_out": mode_index_out_to,
             }
-            s_matrix.loc[coords_to] = mult_by * s_matrix.loc[coords_from].values
+            s_matrix = s_matrix._with_updated_data(data=elements_from, coords=coords_to)
 
         return s_matrix
--- a/tidy3d/plugins/smatrix/component_modelers/terminal.py
+++ b/tidy3d/plugins/smatrix/component_modelers/terminal.py
@@ -291,9 +291,10 @@ class TerminalComponentModeler(AbstractComponentModeler[NetworkIndex, NetworkEle
             port, mode_index = self.network_dict[source_index]
             sim_data = batch_data[self._task_name(port=port, mode_index=mode_index)]
             a, b = self.compute_power_wave_amplitudes_at_each_port(port_impedances, sim_data)
-            indexer = {"f": a.f, "port_out": a.port, "port_in": source_index}
-            a_matrix.loc[indexer] = a
-            b_matrix.loc[indexer] = b
+
+            indexer = {"port_in": source_index}
+            a_matrix = a_matrix._with_updated_data(data=a.data, coords=indexer)
+            b_matrix = b_matrix._with_updated_data(data=b.data, coords=indexer)
 
         # If excitation is assumed ideal, a_matrix is assumed to be diagonal
         # and the explicit inverse can be avoided. When only a subset of excitations
@@ -315,7 +316,8 @@ class TerminalComponentModeler(AbstractComponentModeler[NetworkIndex, NetworkEle
                 "port_in": col_out,
                 "port_out": row_out,
             }
-            s_matrix.loc[coords_to] = mult_by * s_matrix.loc[coords_from].values
+            data = mult_by * s_matrix.loc[coords_from].data
+            s_matrix = s_matrix._with_updated_data(data=data, coords=coords_to)
 
         return s_matrix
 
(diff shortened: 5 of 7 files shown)


tylerflex:
NOTE: this doesn't cover the terminal component modeler, that will require some modifications inspired by this. I didnt want to mess with it in case I broke something.

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -196,7 +197,21 @@ def batch_path(self) -> str:
...
+                "type",
+                "path_dir",
+                "attrs",
+                "solver_version",
+                "jobs_cached",
+                "num_workers",
+                "simulations",
+            }
+        )
+        return run_async(batch.simulations, **run_async_kwargs, path_dir=self.path_dir)
```
yaugenst-flex: huh, any idea how autograd tracers interact with the cached property? this seems like it could potentially be a  source of errors. what happens if you use the same data to differentiate twice (not second order, just reuse somewhere else)?

tylerflex: I think cached properties dont discriminate. so in principle they could contain autograd tracers. 

However, after any updated copy, the cached properties should be cleared, so I think practically you'd get the expected behavior. For example

```py
def f(p):

    a = td.Component(p)
    b = a
    return a.cached_property + b.cached_property
```

should trace through both a and b's traced property (is this what you're getting at?)



Note: I dont think this discussion affects the code above. I just `exclude` various fields of `Batch` that dont get passed to `run_async`

tidy3d/components/data/data_array.py
```diff
@@ -489,6 +489,25 @@ def _ag_interp_func(var, indexes_coords, method, **kwargs):
                 result = result.transpose(*out_dims)
         return result
 
+    def with_updated_data(
```
groberts-flex: will this work if you're setting multiple values at once instead of selecting a single coordinate value? in the terminal component case, I needed to select for multiple frequencies at once and ended up using xarray directly to create the mask.

```
 indexer = {"f": V_out.f, "port": port_out.name}
 mask = xarray.zeros_like(V_matrix, dtype=bool) 
 mask.loc[indexer] = True
```

tylerflex: Unfortunately it doesnt work with setting arrays yet. And i kind of ran out of time to implement that. I tried modifying the terminal.py to use it and ran into issues because of this. We can revisit it

tylerflex: @groberts-flex I tried changing this PR to use xarray.DataArray as a mask but didnt get very far. Do you have an idea of how to make this work within this `with_updated_data()` method?

groberts-flex: not sure off the top of my head, but will have some time later this afternoon to take a look! I'll let you know if I can get it to work

groberts-flex: ended up getting to this today and think I'm close to a solution. I believe I have a version of this that is working for the terminal component modeler, but not working for the modal which exposed a flaw in my logic. I'll hopefully be able to get it finished up in the morning and will send you the code snippet

groberts-flex:
```
def with_updated_data(
    self,
    data,
    coords: dict[str, Any]) -> DataArray:
    """Make copy of ``DataArray`` with ``data`` at specified ``coords``, autograd-approved."""

    mask = xr.zeros_like(self, dtype=bool)
    mask.loc[coords] = True

    old_values = self.values
    replacement_data = np.zeros_like(old_values)
    axes_to_fill = [self.get_axis_num(key) for key in coords]

    reshape_guide = [1] * len(old_values.shape)
    for idx, dim in enumerate(axes_to_fill):
        reshape_guide[dim] = data.shape[idx]

    replacement_data = replacement_data + data.reshape(reshape_guide)
    modified_values = np.where(mask, replacement_data, old_values)

    return self.copy(deep=True, data=modified_values)
```

This requires the input in data to be shaped according to the dimensions in the coords. So the other change this would require is in modal.py

```
reshaped_smatrix_data = np.reshape(s_matrix_elements, (1, 1, 1, 1))
s_matrix = s_matrix.with_updated_data(data=reshaped_smatrix_data, coords=coords_set)
```

and

```
reshaped_elements_from = np.reshape(elements_from, (1, 1, 1, 1))
s_matrix = s_matrix.with_updated_data(data=reshaped_elements_from, coords=coords_to)
```

I tested this for modal using the waveguide bend example notebook and adding an smatrix calculation to its objective function. I also tested with terminal and for my use case there it is working.

tylerflex: I got 
```
FAILED tests/test_plugins/smatrix/test_component_modeler.py::test_run_component_modeler - ValueError: cannot reshape array of size 2 into shape (1,1,1,1)
```
when trying this in the unit tests

tylerflex: I think I got it working by reshaping to 
```
(1, 1, 1, -1)
```

tylerflex: I just pushed this change to the latest commit, do you mind trying it again on your test?

tylerflex: confirmed that this change works for my test case (the Autograd1 notebook, implemented with the ComponentModeler)

groberts-flex: interesting, I think I interpreted the data as always being one-dimensional in that part of the s matrix because it was a single port and mode to another single port and mode, so I'm confused why it ended up having two entries in that test

groberts-flex: but also can confirm that your change also works in my modal s matrix autograd test!

tidy3d/plugins/smatrix/component_modelers/modal.py
```diff
@@ -339,12 +341,16 @@ def _internal_construct_smatrix(self, batch_data: BatchData) -> ModalPortDataArr
...
+            elements_from = mult_by * s_matrix.loc[coords_from].values
             coords_to = {
                 "port_in": port_in_to,
                 "mode_index_in": mode_index_in_to,
                 "port_out": port_out_to,
                 "mode_index_out": mode_index_out_to,
             }
-            s_matrix.loc[coords_to] = mult_by * s_matrix.loc[coords_from].values
+
+            reshaped_elements_from = np.reshape(elements_from, (1, 1, 1, -1))
```
yaugenst-flex: do we know that this is always the correct shape? seems a bit brittle to me.
i think this reshaping should ideally be done within `with_updated_data`, where you identify the dims that are being indexed and then make sure that the data shape matches the non-indexed dims?

tidy3d/components/data/data_array.py
tylerflex: Maybe could you add your changes to terminal.py as a commit to this branch? and then we can merge everything together in this PR

tylerflex: > I think I interpreted the data as always being one-dimensional in that part of the s matrix because it was a single port and mode to another single port and mode, so I'm confused why it ended up having two entries in that test

yea it does seem strange, I will follow up on this.

groberts-flex: yes, happy to put the terminal changes in here as well! just pushed the commit for this

tidy3d/plugins/smatrix/component_modelers/modal.py
groberts-flex: originally I had tried to do some reshaping inside `with_updated_data` but quickly it was adding a good bit of logic to that function and I worried that would end up being a little brittle so I thought it would be better if it was required that the input data to this function matched the shape of the coords passed in as well (if we stick with that route, maybe a double check of this assumption and a good error message would be useful to have). That being said, I'm super open for this to be done the other way with reshaping done inside of `with_updated_data`, but wanted to give the reasoning for why I left it out!

yaugenst-flex: it does make sense, although it puts the burden of reshaping correctly on the caller of the function, which is my main concern. specifically because this is a new `DataArray` method i think it would be best to adhere to xarray conventions. in any case, i was thinking along these lines in `with_updated_data`:
```py
# target shape should be 1 for indexed dims and the original size for non-indexed dims
target_shape = []
data_shape_iter = iter(data.shape)
for dim_name in self.dims:
    if dim_name in coords:
        target_shape.append(1)
    else:
        # should probably throw in a check to ensure data shape is same as coord shape here...
        target_shape.append(next(data_shape_iter))
reshaped_data = data.reshape(target_shape)
```

yaugenst-flex: but actually now that i'm thinking about this some more, why are we not simply creating an autograd-compatible `DataArray.assign_coords()` (https://docs.xarray.dev/en/stable/generated/xarray.DataArray.assign_coords.html)? like it seems to me that that's fundamentally what's being attempted here? @tylerflex 

so essentially follow this pattern we already use:
```py
class DataArray:
  def assign_coords(coords=None, **coords_kwargs):
    if hasbox(coords):
      return self._ag_assign_coords(coords, **coords_kwargs)
  return super().assign_coords(coords, **cords_kwargs)
```

yaugenst-flex: actually i wonder if `assign_coords` is autograd-compatible out of the box...

tylerflex: I dont quite understand the use of assign coords here, could you explain more? I do agree about moving the reshaping logic to `with_updated_copy`

yaugenst-flex: maybe i misunderstand but doesn't `assign_coords` do essentially the same thing as `with_updated_copy`? it assigns updated/new values to specific coordinates and returns a copy

tylerflex: I thought it only reassigns the values of the coordinates themselves, not the data

yaugenst: how did i miss that 😂 ehm yeah you're right.. somehow i misread it `{coords: values}` as being values at coords... oops. that being said, maybe `assign_coords` can still be useful for the custom reshaping logic

yaugenst-flex: how did i miss that 😂 ehm yeah you're right.. somehow i misread it `{coords: values}` as being values at coords... oops. that being said, maybe `assign_coords` can still be useful for the custom reshaping logic

tylerflex: I think I figured out a good solution that covers everything, will push my commit soon.

tylerflex: @groberts-flex I needed to remove some of the items from the `indexer` dictionaries passed to `with_updated_copy` because they were referring to some dimensions that were not being selected out. Then I used a combination of the logic from each of you to reshape the data.

I'm going to try a pure xarray version soon because I think that's a bit cleaner (eg where `data` is now just the DataArray itself not a numpy array)

tylerflex: something weird is happening in the terminal component modeler when I try to generalize this

```py
    def _internal_construct_smatrix(self, batch_data: BatchData) -> TerminalPortDataArray:
        """Post process :class:`.BatchData` to generate scattering matrix, for internal use only."""

        port_names = [port.name for port in self.ports]

        a_values = np.zeros(
            (len(self.freqs), len(port_names), len(port_names)),
            dtype=complex,
        )
        b_values = np.zeros(
            (len(self.freqs), len(port_names), len(port_names)),
            dtype=complex,
        )
        coords = {
            "f": np.array(self.freqs),
            "port_out": port_names,
            "port_in": port_names,
        }
        a_matrix = TerminalPortDataArray(a_values, coords=coords)
        b_matrix = TerminalPortDataArray(b_values, coords=coords)

        # Tabulate the reference impedances at each port and frequency
        port_impedances = self._port_reference_impedances(batch_data=batch_data)

        for _port_in_idx, port_in in enumerate(self.ports):
            sim_data = batch_data[self._task_name(port=port_in)]
            a, b = self.compute_power_wave_amplitudes_at_each_port(port_impedances, sim_data)
            indexer = {"port_in": port_in.name}
            a_matrix = a_matrix.with_updated_data(data=a, coords=indexer)
            b_matrix = b_matrix.with_updated_data(data=b, coords=indexer)

        s_matrix = self.ab_to_s(a_matrix, b_matrix)
```

it seems like `a` has dimensions of `('f', 'port_out', 'port_in', 'port')` whereas the DataArray getting updated `a_matrix` only has dimensions of `('f', 'port_out', 'port_in')`. so in this case it seems there's an extra dimension of `port` not in the target DataArray, which I'm not sure how to handle. can we just drop this dim?

groberts-flex: interesting, that does seem odd! seems like things are a little messy in how these indexers are being used. following to where a, b are generated from in compute_power_wave_amplitudes_at_each_port, I'm actually only expecting them to have dimensions of 'f' and 'port' but not necessarily 'port_out' and 'port_in'.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -284,8 +288,8 @@ def compute_power_wave_amplitudes_at_each_port(
         for port_out in self.ports:
             V_out, I_out = self.compute_port_VI(port_out, sim_data)
             indexer = {"port": port_out.name}
-            V_matrix.loc[indexer] = V_out
-            I_matrix.loc[indexer] = I_out
+            V_matrix = V_matrix.with_updated_data(data=V_out.data, coords=indexer)
+            I_matrix = V_matrix.with_updated_data(data=I_out.data, coords=indexer)
```
groberts-flex: a typo I introduced with my commit. The I_matrix line should be I_matrix.with_updated_data 

sorry about that, tested things on my antenna example so I missed this fix

tidy3d/plugins/smatrix/component_modelers/modal.py
tylerflex: oops yea it's only `f` and  `port` actually, the port in and port out came from my own code where I was broadcasting. So if we want to update `a_matrix` with `a`, what dimension of `a_matrix` does `a.port` index to?

tidy3d/components/data/data_array.py
```diff
@@ -489,6 +489,38 @@ def _ag_interp_func(var, indexes_coords, method, **kwargs):
...
+    def with_updated_data(self, data: xr, coords: dict[str, Any]) -> DataArray:
+        """Make copy of ``DataArray`` with ``data`` at specified ``coords``, autograd-approved."""
+
+        old_values = self.values
+        shape_in = data.shape
+        shape_in_new = []
+        i = 0
+        for dim in self.dims:
+            if dim in coords:
+                shape_in_new.append(1)
```
groberts-flex: I was thinking of this flipped around and that if the dim is in the coords that we would want to use the dim in shape_in_new and otherwise would want to have append 1

tidy3d/plugins/smatrix/component_modelers/modal.py
groberts-flex: a.port corresponds to a_matrix.port_out as I understand it (and based on the original indexer that was in there of `indexer = {"f": a.f, "port_in": port_in.name, "port_out": a.port}`)

tidy3d/plugins/smatrix/component_modelers/terminal.py
tylerflex: thanks, fixed!

tidy3d/components/data/data_array.py
tylerflex: hm, I just assumed that if the dim was in the coords, then the `data` was effectively missing that dim? 

I tried an algorithm where
1. I add any missing dims to the `data` (new data), assuming it had been selected out already.
2. I added the expanded dim `data` to 0 * the self.values, which essentially broadbasts everything to self.

Note in the latest commit I simply use xarray to handle everything, still testing that one but it passes the unit tests.

tylerflex: I reverted my previous commit, it broke things in autograd..

groberts-flex: I see - I interpreted it the opposite, that if the dim was in coords that that dim was one of the ones that existed in the data. But it's entirely possible I've just gotten myself a little turned around on which coords is which! I think if there is a way to get xarray to do this it would be great! tried at one point to use the xarray where functionality and it did break somewhere for me, but now am not remembering exactly where. let me know if you want me to run any versions of this with my terminal component test case!

groberts-flex:
Testing with terminal in my patch antenna optimization, and one other spot we need a change is on line 319/320 in terminal.py, to replace
```
V_matrix.loc[indexer] = V_out
I_matrix.loc[indexer] = I_out
```
with
```
V_matrix = V_matrix._with_updated_data(data=V_out.data, coords=indexer)
I_matrix = I_matrix._with_updated_data(data=I_out.data, coords=indexer)
```

I'm re-running the whole optimization currently, but so far it looks like it is working with this change!

tidy3d/components/data/data_array.py
```diff
@@ -489,6 +489,33 @@ def _ag_interp_func(var, indexes_coords, method, **kwargs):
                 result = result.transpose(*out_dims)
         return result
 
+    def _with_updated_data(self, data: np.ndarray, coords: dict[str, Any]) -> DataArray:
+        """Make copy of ``DataArray`` with ``data`` at specified ``coords``, autograd compatible
+
+        Constraints / Edge cases:
+            - `coords` must map to a specific value eg {x: '1'}, does not broadcast to arrays
```
groberts-flex, line 496: will this case error in the function if you try and broadcast to an array like this? if not, can we catch that case and error on it?

groberts-flex (commented):
I was able to verify with my optimization that uses the terminal component modeler after locally adding the other _with_updated_data call. Initially, though, the optimization wasn't working and I realized it was because I need local_gradient=True on so that the gradient gets computed through my branch (for PEC changes). I'm not sure the best way to enable this, but I added another field to AbstractComponentModeler that was local_gradient with a default of False, and then when the run_async call gets made, I pass in local_gradient=self.local_gradient.

tylerflex:
@groberts-flex  thanks I'll revisit your other comments after my meetings this morning, but I added the initial fix in

https://github.com/flexcompute/tidy3d/pull/2572/commits/44ff766d4c76042fd8ccb88d862c89dda5f90acf

tylerflex:
I think we could consider either adding `local_gradient : bool = False` as a `ComponentModeler` field or as a kwarg to `ComponentModeler.run(local_gradient=False)` could make sense. We could also just add `**run_kwargs` there. Unless @yaugenst-flex objects. (basically all extra kwargs get passed to `run_async`)

tidy3d/components/data/data_array.py
tylerflex: it will error. we could catch that, it might be more clean that way. We can basically raise an exception discussing the requirements on the data shapes being passed to this function and do this `from e` where e is the original exception so they get the full trace.

yaugenst-flex:
i think it definitely should not be a model field but rather an argument to the run function. the default can be set via config soon. but it should not be part of the schema

tidy3d/components/data/data_array.py
tylerflex: implemented here
https://github.com/flexcompute/tidy3d/pull/2572/commits/3463a9f172dd38de617d63dc8dacbbb747400486

tylerflex:
> i think it definitely should not be a model field but rather an argument to the run function. the default can be set via config soon. but it should not be part of the schema

@yaugenst-flex unfortunately the way things are organized right now, it seems like it needs to be a model field unless we want to break backwards compatibility. Maybe in light of the RF changes to the TerminalComponentModeler, we could consider waiting, but here's my stab at it https://github.com/flexcompute/tidy3d/pull/2572/commits/47130cf9d48c7d0536f909d1e811e3866578b045

Basically the `.batch` and `.batch_data` are currently @properties of the modeler, so we can't pass arguments to them without changing the syntax or messing around with mutability. as far as I can tell.

tylerflex:
@yaugenst-flex want to take a final look?

tests/test_data/test_data_arrays.py
```diff
@@ -468,3 +470,49 @@ def test_interp(method, scalar_index):
...
+    data_expected[0, 1, 1, :] = 0.0 + 0j
+    assert np.all(arr2.data == data_expected), "DataArray.with_updated_copy() failed"
+
+    def f(x):
+        arr2 = arr._with_updated_data(data=x, coords=coords)
+        return np.abs(np.sum(arr2.data))
+
+    # grad should just be all 1s because of sum, so check that this is true
+    g = ag.grad(f)(data)
+    assert np.all(g == np.ones_like(data))
```
yaugenst-flex, line 499: Is fine but also redundant due to `check_grads()` call below

yaugenst-flex (approved):
LGTM!

tests/test_data/test_data_arrays.py
tylerflex: yea. I realized this too but felt why not leave it in

PR #2573, opened 2025-06-13
add selective simulation capabilities to TerminalComponentModeler

### LINKED ISSUE #2392: Option to not excite all ports in TerminalComponentModeler
Suggestion to include an option to not excite all ports in a TerminalComponentModeler simulation. 
Rationale is that:

- Sometimes the user is only interested in running sim for 1 port
- Sometimes the device is symmetric so only 1 sim is needed to extract multiple s-parameters

A simple addition would be to include a 'run_only' option similar to ComponentModeler. This allows the user to control which ports to excite.

However, in the case that the device is symmetric, this does not automatically populate the entire S-matrix. (E.g. The user runs sim with port 1 excitation to get S11, but S22 does not get updated automatically.) 

A slightly more complicated feature would be a 'port_symmetry' feature where the user can specify port symmetry pairs. For example, in a 4-port device, one can specify that the device is symmetric w.r.t. ports 2 and 3. Then the simulation only excites one of ports 2 and 3, and sets S22=S33, S23=S32 etc. 

Below is a screenshot of how CST implements it.
Image

dmarek-flex:
Follows the same approach as the `ComponentModeler` where users can choose to `run_only` a subset of simulations and also may provide `element_mappings` to fill in unknown entries in the S matrix using other known entries.

The major difference is that the `coords` used in the final matrix `DataArray` are flattened so I can mix different types of Ports (Lumped and Wave), where some support modes and others do not.

## Greptile Summary

Extends `TerminalComponentModeler` with selective simulation capabilities, enabling users to compute only specific S-matrix columns and derive missing elements through mapping relationships.

- Added `run_only` parameter to `TerminalComponentModeler` for selectively computing S-matrix columns, reducing simulation time
- Implemented `element_mappings` feature to derive unknown S-matrix elements from known entries
- Modified matrix indexing to support mixed port types (Lumped and Wave) through flattened coordinates
- Added comprehensive test coverage in `test_terminal_component_modeler.py` validating selective simulations and mappings
- Updated port naming convention in test definitions to clearly distinguish between 'coax' and 'wave' port types

### DIFF
--- a/tidy3d/plugins/smatrix/component_modelers/terminal.py
+++ b/tidy3d/plugins/smatrix/component_modelers/terminal.py
@@ -30,8 +30,11 @@ from tidy3d.web.api.container import BatchData
 
 from .base import AbstractComponentModeler, TerminalPortType
 
+NetworkIndex = str  # the 'i' in S_ij
+NetworkElement = tuple[NetworkIndex, NetworkIndex]  # the 'ij' in S_ij
 
-class TerminalComponentModeler(AbstractComponentModeler):
+
+class TerminalComponentModeler(AbstractComponentModeler[NetworkIndex, NetworkElement]):
     """Tool for modeling two-terminal multiport devices and computing port parameters
     with lumped and wave ports."""
 
@@ -47,6 +50,18 @@ class TerminalComponentModeler(AbstractComponentModeler):
         title="Radiation Monitors",
         description="Facilitates the calculation of figures-of-merit for antennas. "
         "These monitor will be included in every simulation and record the radiated fields. ",
+    )
+
+    assume_ideal_excitation: bool = pd.Field(
+        False,
+        title="Assume Ideal Excitation",
+        description="If ``True``, only the excited port is assumed to have incident power, so the "
+        "vector of incident power wave amplitudes (a) is assumed to be all zeros except for the "
+        "entry associated with the excited port. This choice simplifies the calculation of the "
+        "scattering matrix. If ``False``, every entry in the vector of incident power wave "
+        "amplitudes (a) is calculated explicitly. This choice requires a matrix inversion when "
+        "calculating the scattering matrix, but may lead to more accurate scattering parameters "
+        "when there are reflections from simulation boundaries. ",
     )
 
     @pd.root_validator(pre=False)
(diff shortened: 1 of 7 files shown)


tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -148,33 +262,27 @@ def sim_dict(self) -> dict[str, Simulation]:
...
-        # Check final simulations for grid size at lumped ports
-        for _, sim in sim_dict.items():
-            TerminalComponentModeler._check_grid_size_at_ports(sim, self._lumped_ports)
-
-        # Now, create simulations with wave port sources and mode solver monitors for computing port modes
-        for wave_port in self._wave_ports:
+        return sim_wo_source.copy(update=update_dict)
+
+    def _add_source_to_sim(
+        self, source_index: NetworkIndex, sim_wo_source: Simulation
```
dbochkov-flexcompute: is `sim_wo_source == self.base_sim` always? maybe remove the argument then?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -95,12 +119,102 @@ def plot_sim_eps(
...
+
+    @cached_property
+    def matrix_indices_source(self) -> tuple[NetworkIndex, ...]:
+        """Tuple of all the source matrix indices, which may be less than the total number of ports."""
+        if self.run_only is not None:
+            return self.run_only
+        return self.matrix_indices_monitor
+
+    @cached_property
+    def matrix_indices_run_sim(self) -> tuple[NetworkIndex, ...]:
```
dbochkov-flexcompute: looks like `matrix_indices_source` and `matrix_indices_run_sim` are exactly the same as in modal component modeler. Maybe can move them to the base class?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -208,14 +320,34 @@ def _internal_construct_smatrix(self, batch_data: BatchData) -> TerminalPortData
...
 
         # loop through source ports
-        for port_in in self.ports:
-            sim_data = batch_data[self._task_name(port=port_in)]
+        for source_index in source_indices:
+            port, mode_index = self.network_dict[source_index]
+            sim_data = batch_data[self._task_name(port=port, mode_index=mode_index)]
             a, b = self.compute_power_wave_amplitudes_at_each_port(port_impedances, sim_data)
-            indexer = {"f": a.f, "port_in": port_in.name, "port_out": a.port}
+            # a = a.where(a.port.isin([str(source_index)]), 0)
```
dbochkov-flexcompute: stray commented code

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -208,14 +320,34 @@ def _internal_construct_smatrix(self, batch_data: BatchData) -> TerminalPortData
...
             a_matrix.loc[indexer] = a
             b_matrix.loc[indexer] = b
 
-        s_matrix = self.ab_to_s(a_matrix, b_matrix)
+        if len(monitor_indices) == len(source_indices):
+            s_matrix = self.ab_to_s(a_matrix, b_matrix)
+        else:
+            a_diag = np.diagonal(a_matrix, axis1=1, axis2=2)
+            # Scale each column by the corresponding diagonal entry
+            s_matrix = b_matrix / a_diag[:, np.newaxis, :]
```
dbochkov-flexcompute, line 306: is this an approximation to s-matrix? or is it still exact one? If it is an approximation we should probably show a warning

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -208,14 +320,34 @@ def _internal_construct_smatrix(self, batch_data: BatchData) -> TerminalPortData
...
+            coords_from = {
+                "port_in": col_in,
+                "port_out": row_in,
+            }
+            coords_to = {
+                "port_in": col_out,
+                "port_out": row_out,
+            }
+            s_matrix.loc[coords_to] = mult_by * s_matrix.loc[coords_from].values
+
```
dbochkov-flexcompute, line 319: I am not an expert in this, so maybe that's why I don't see this immediately, but do we get identical s-matrices if we run a component modeler directly without any mappings and with appropriate mappings? I could see that if element mappings were used to populate a and b until they are full, and then the s-matrix were computed as usual

dmarek-flex: The element mappings are when the user has some additional knowledge of the component, like that it is symmetric so that port 1 and port 2 are identical. You could also assume that the system is reciprocal which forces the s matrix itself to be symmetric. 

So maybe theoretically you should get identical s matrices, but in practice there will be some numerical error that might change the values.

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: Note sure what the best approach is here. In an ideal simulation, the two techniques would be equivalent because the `a_matrix` would be a diagonal matrix. In practice though, even if you only excite one Port (say a `ModeSource`) there might be reflections from PMLs or other affects that make nonzero non-diagonal entries enter into the `a_matrix` at the ports that are turned off. The `ComponentModeler` takes the approach that assumes the off-diagonal entries are 0, and I believe it is the only way for calculating partial S matrices.

I'm leaning towards making the default in `TerminalComponentModeler` to be the same as `ComponentModeler`, so just taking the diagonal of `a_matrix`. But I might run some tests, to see how S parameters change when using this approach.

weiliangjin2021 (approved):
All look good to me.

dmarek-flex:
Going to be conflicts with https://github.com/flexcompute/tidy3d/pull/2663, so I might wait for that to merge first.

FYI @daquinteroflex

daquinteroflex:
Actually I'd argue it's probably better to merge this if it's ready and continue the refactor using this

dbochkov-flexcompute (approved):

Issue #2575, opened 2025-06-15
Arrows are scaled incorrectly when `ax.set_aspect()` is used.


jewettaijfc:
**Describe the bug**

Currently, the length of arrows are incorrectly scaled causing them to be too long to fit in the graph window.  This happens in the following situation:
1) The user invokes matplotlib's `ax.set_aspect()` function to override the aspect ratio.  (Many of the example notebooks do this to display objects that are very long or very thin.)
2) The arrow you want to view is *not* pointing in the horizontal direction.  *(Note: Arrows pointing in the horizontal direction are currently scaled correctly.)*

**Screenshots**
![Image](https://github.com/user-attachments/assets/8663a250-a60b-4b98-acc6-def048d519f1)

**To Reproduce**
Here is a simple notebook that I used to produce the image above: 

[test_arrow_autoscale.ipynb.zip](https://github.com/user-attachments/files/20747274/test_arrow_autoscale.ipynb.zip)

Download it, unzip it, and run the cells.

*(NOTE: I disabled the plot on the right because it depends on [PR2544](https://github.com/flexcompute/tidy3d/pull/2544).  That PR might not be merged by the time you are reading this.  But it's not necessary to show this plot to see the problem.  It just makes it more obvious.)*

**Expected behavior**
Ideally we want both the 2nd and 3rd plots to 1) have reasonable arrow sizes, and 2) look consistent (ie. they should look identical except for swapping the horizontal and vertical axes).

**Desktop (please complete the following information):**
 - OS: ubuntu24.04
 - Browser: chrome
 - Version: 137.0.7151.55

**Additional context**
This problem is making it hard to implement [PR2544](https://github.com/flexcompute/tidy3d/pull/2544).  But it is a problem which predates that PR.

jewettaijfc:
## I created a [PR to fix this problem (#2576)](https://github.com/flexcompute/tidy3d/pull/2576)
It is a hacky solution, but it helps.

tylerflex:
closing as fixed by #2576

PR #2544, opened 2025-06-05
XY transposable plots V3 part 2: Updated all remaining plot functions (issue1072)

jewettaijfc:
This PR addresses the feature request from issue [1072](https://github.com/flexcompute/tidy3d/issues/1072), adding the ability to swap the horizontal and vertical axes of *all* Tidy3D plots.

## Reviewer instructions

This PR has been based on top of [PR2687](https://github.com/flexcompute/tidy3d/pull/2687).  You must land that PR before this one.

## Automatic Tests
```
poetry run pytest tests
```
## Manual Tests
- Open any of the [tidy3d-notebook examples](https://github.com/flexcompute/tidy3d-notebooks/) with jupyter.
- Evaluate the cells to verify that the plots using `transpose=True` are flipped (with respect to the plots that omit the `transpose` argument).

I have tested this PR with over 150 examples (out of 191).  It's possible that some advanced plot types which are not covered in these examples may have visual glitches or don't work when `transpose=True`.

But all plots that display simulations, scenes, structures, sources, monitors work with `transpose=True`.  (Including geometry transformations and clipping.)

PR #2576, opened 2025-06-15
Prevent arrow lengths from being too long

jewettaijfc:
### This PR is an attempt to address [issue #2575](https://github.com/flexcompute/tidy3d/issues/2575).

After this PR, arrows will always fit within plot window, regardless of their direction and regardless of the aspect ratio of the plot (`ax.set_aspect()`).

**Screenshots**

***Before this PR***

Before this PR, the vertical arrows could easily grow too long to fit within the plot window:
Image

***After this PR***

After this PR, the green arrow is visible, regardless of it's orientation and regardless of the aspect ratio:
fix_arrow_autoscale_after

**Test files**
The plot above was made using this file:
[test_arrow_autoscale.ipynb.zip](https://github.com/user-attachments/files/20747335/test_arrow_autoscale.ipynb.zip)

Download it, unzip it, and run the cells.

(NOTE: I disabled the plot on the right because it depends on https://github.com/flexcompute/tidy3d/pull/2544. That PR might not be merged by the time you are reading this. But it's not necessary to show this plot to see the problem. It just makes it more obvious.)

**Notes:**
- Previously, autoscaling worked in the horizontal direction, but not the vertical direction.
- Now it works in both horizontal and vertical directions.
- This PR fixes the problem of ***long*** arrows.  In this example, the other arrow (the grey arrow) is now too short to see clearly.  But I'm less concerned about short arrows because all arrows are still visible, regardless of how short they are. *(To see they grey arrow in this example, you just need to make the 2nd graph wider.  You can do that by adding `hlim=[-20,20]` to the 2nd plot() argument list.  The short grey arrow will resemble a small sideways triangle.)*  If you want to preserve the length of *short arrows* as well, a [different approach](https://docs.google.com/document/d/1JZ1dAv_0V_B7NaobwO1DcLnuR9rtFLDb3M7WdMuaHww/edit?usp=drive_link) is needed.  (I did not try this because it is more complicated.  See [below](https://github.com/flexcompute/tidy3d/pull/2576#issuecomment-2975060699).)
- ***More notes are listed in the comments below.***

**Additional context**

[This problem](https://github.com/flexcompute/tidy3d/issues/2575) is making it hard to implement [PR2544](https://github.com/flexcompute/tidy3d/pull/2544).

## Greptile Summary

Fixed visualization issue in `tidy3d/components/geometry/base.py` where arrows could extend beyond plot boundaries by modifying `_arrow_shape_cb` to consider both horizontal and vertical scales when computing arrow lengths.

- Updated arrow length calculation in Box class to use both axes scales instead of just horizontal
- Ensures arrows remain visible within plot window regardless of direction/aspect ratio
- Trade-off: While long arrows are now properly constrained, some short arrows may appear smaller
- Visual improvement verified through test notebook with different aspect ratios
- Critical fix needed to unblock implementation of PR#2544

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Arrow lengths are now scaled consistently in the X and Y directions,
+  and their lengths no longer exceed the height of the plot window.
+
 ## [2.9.0rc1] - 2025-06-10
 
 ### Added
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -2292,7 +2292,9 @@ class Box(SimplePlaneIntersection, Centered):
             event.canvas.mpl_disconnect(arrow.set_shape_cb[0])
 
             transform = arrow.axes.transData.transform
-            scale = transform((1, 0))[0] - transform((0, 0))[0]
+            scale_x = transform((1, 0))[0] - transform((0, 0))[0]
+            scale_y = transform((0, 1))[1] - transform((0, 0))[1]
+            scale = max(scale_x, scale_y)  # <-- Hack: This is a somewhat arbitrary choice.
             arrow_length = ARROW_LENGTH * event.canvas.figure.get_dpi() / scale
 
             if bend_radius:


jewettaijfc:
> ## tidy3d/components/geometry/base.py
> ```
>   2295             transform = arrow.axes.transData.transform
> ! 2296             scale_x = transform((1, 0))[0] - transform((0, 0))[0]
> ! 2297             scale_y = transform((0, 1))[1] - transform((0, 0))[1]
> ! 2298             scale = math.sqrt(scale_x**2 + scale_y**2)
>   2299             arrow_length = ARROW_LENGTH * event.canvas.figure.get_dpi() / scale
> ```

- I confess that the choice of equation used for `scale`  (eg `scale = math.sqrt(scale_x**2 + scale_y**2)`) was not well motivated.  I tried several equations.  This hacky equation seems to work reasonably well at all orientations and all aspect ratios.  (The arrow never grows by more than a factor of $\sqrt(2)$.  If you want to make sure the arrow length never grows at all, then a simpler choice is to use `scale=max(scale_x, scale_y)` instead.)  ***<--UPDATE: Later decided to use this equation instead.***
- Feel free to pick a different equation if you find something that works better.  I am happy as long as the arrows always fit within the plot window (and point in the correct direction).

jewettaijfc:
I updated the equation for `scale`:
```
scale = max(scale_x, scale_y)
```
This is probably not the optimal choice either.  But at least it guarantees that arrows remain within the plot window.
- Again, feel free to suggest a different equation or a different approach.
- *(Sorry if I was sloppy with this PR.  I am in a hurry to get back to work on a different problem.)*

jewettaijfc:
## IMPORTANT: This PR does *not* preserve arrow lengths on screen when the aspect ratio is changed.

Do we want that?

### If you want that, you must use a different approach.

...That approach is explained [**here**](https://docs.google.com/document/d/1JZ1dAv_0V_B7NaobwO1DcLnuR9rtFLDb3M7WdMuaHww/edit?usp=drive_link).  I tested that google-gemini code and actually works.  I did not implement it, because I could not figure out how to make it work with the custom `bend_radius` code we are using.  But if I can get someone to explain what the `bend_radius` code does, I am happy to solve this problem the better way (so that arrow lengths on screen are not affected by the aspect ratio).  This would prevent arrows from getting shorter as the aspect ratio changes.  It would guarantee that all arrows remain easy to see.

***Currently, this PR is a quick hack which makes the minimal code changes to make sure that the arrows are not too large to fit in the plot window.***

lucas-flexcompute (approved):

tylerflex:
Thanks @jewettaijfc, this looks good to me overall, but I'll feel safer if we check to make sure the notebooks that use this in plotting still look ok (if you haven't already).

tylerflex:
To be clear: no need to check every notebook but maybe just look at around 5 and re-run (even without transpose) just to make sure nothing looks weird. Thanks!

tylerflex:
Please squash commits into one (using `git rebase -i`)

and also please add a line under "fixed" in the `CHANGELOG.md`

jewettaijfc:
> Thanks @jewettaijfc, this looks good to me overall, but I'll feel safer if we check to make sure the notebooks that use this in plotting still look ok (if you haven't already).

I've checked 25 notebooks that use it this code today.  I tested them while I was testing [PR2544](https://github.com/flexcompute/tidy3d/pull/2544).  Not all of these notebooks have examples that display arrows.  But the examples that do have arrows look okay (in my opinion).

------------------------------------
Here's the only weird side effect of this PR that I've found so far in my tests:

Sometimes arrows that were originally finite length become very short (as a result of this PR) when the aspect ratio gets very large.  This doesn't happen very often, but it does occasionally happen.  The [alternative solution I mentioned earlier](https://github.com/flexcompute/tidy3d/pull/2576#issuecomment-2975060699) tries to solve that that issue by preserving the (visual) length of each arrow.  But I don't know if it's worth the effort.  I personally don't mind seeing short arrows (especially when the aspect ratio is 100).

jewettaijfc:
- I think I squashed it.
- I added my changes to the CHANGELOG under `### Fixed` inside `## [2.9.0rc1] - 2025-06-10`, and I updated the date to today's date `2025-06-16`.)  (Please let me know if I was not supposed to alter the date.)

CHANGELOG.md
tylerflex, line 1: Kind of a special case because we just released 2.9.0rc1, but you should make a new section for `### Fixed`  under the `## [Unreleased]` section (and above `## [2.9.0rc1] - 2025-06-10`). Essentially we are going to be building up features under this unreleased section until the next release goes out.

The date should be reverted back to June 10, because that's the date we released 2.9.0rc1. 

You can read more about changelog conventions here
https://keepachangelog.com/en/1.1.0/

tylerflex (commented):
oops forgot to hit submit!

tylerflex (approved):

PR #2582, opened 2025-06-18
feat[adjoint]: Add conductivity gradient for `CustomMedium`

yaugenst-flex:
### Summary

This PR introduces the capability to compute gradients with respect to the `conductivity` of a `CustomMedium`.

### Implementation Details

* Generalizes the `_derivative_field_cmp` helper method with a new `component` parameter that can be set to `'real'`, `'imag'`, or `'complex'`.
* Implements the Vector-Jacobian Product (VJP) for conductivity (`component='imag'`) by scaling the imaginary part of the complex permittivity derivative by `1 / (omega * epsilon_0)`.
* Updates the `_compute_derivatives` dispatcher to handle the new `"conductivity"` parameter path and invoke the generalized helper method.
* Adds `test_custom_medium_conductivity_only_gradient` to validate the gradient calculation for a `CustomMedium` with variable conductivity and constant permittivity.
* Adds a test case structure, `custom_med_with_conductivity`, that contains both variable permittivity and conductivity.

## Greptile Summary

Implements conductivity gradient calculations for `CustomMedium` by extending the permittivity gradient framework to handle real, imaginary, and complex components, enabling material conductivity optimization.

- Added `component` parameter to `_derivative_field_cmp` to support 'real', 'imag', and 'complex' gradient types
- Implemented conductivity Vector-Jacobian Product (VJP) by scaling imaginary permittivity derivative by `1/(omega * epsilon_0)`
- Added new test case `custom_med_with_conductivity` to validate combined permittivity and conductivity gradients
- Refactored test code in `test_autograd.py` to improve maintainability with shared coordinate dictionaries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,11 +18,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `InternalAbsorber` class for placing first-order absorbing boundary conditions on planes inside the simulation domain. Internal absorbers are automatically wrapped in a PEC frame with a backing PEC plate on the non-absorbing side.
 - Added `absorber` field (default: `True`) to `WavePort` for automatically placing an absorber behind the port.
 - Added `conjugated_dot_product` field in `ModeMonitor` (default: `True`) and `WavePort` (default: `False`) to allow selecting the conjugated or non-conjugated dot product for mode decomposition.
+- Support for gradients with respect to the `conductivity` of a `CustomMedium`.
 
 ### Changed
 - Validate mode solver object for large number of grid points on the modal plane.
 - Adaptive minimum spacing for `PolySlab` integration is now wavelength relative and a minimum discretization is set for computing gradients for cylinders.
 - The `TerminalComponentModeler` defaults to the pseudo wave definition of scattering parameters. The new field `s_param_def` can be used to switch between either pseudo or power wave definitions.
+- Add support for `np.unwrap` in `tidy3d.plugins.autograd`.
 
 ### Fixed
 - Fixed missing amplitude factor and handling of negative normal direction case when making adjoint sources from `DiffractionMonitor`.
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -319,17 +319,34 @@ def make_structures(params: anp.ndarray) -> dict[str, td.Structure]:
     eps_arr = 1.01 + 0.5 * (anp.tanh(matrix @ params).reshape(DA_SHAPE) + 1)
 
     nx, ny, nz = eps_arr.shape
+    da_coords = {
+        "x": np.linspace(-0.5, 0.5, nx),
+        "y": np.linspace(-0.5, 0.5, ny),
+        "z": np.linspace(-0.5, 0.5, nz),
+    }
 
     custom_med = td.Structure(
         geometry=box,
         medium=td.CustomMedium(
             permittivity=td.SpatialDataArray(
                 eps_arr,
-                coords={
-                    "x": np.linspace(-0.5, 0.5, nx),
-                    "y": np.linspace(-0.5, 0.5, ny),
-                    "z": np.linspace(-0.5, 0.5, nz),
-                },
+                coords=da_coords,
+            ),
+        ),
+    )
+
+    # custom medium with variable permittivity and conductivity data
+    conductivity_arr = 0.01 * (anp.tanh(matrix @ params).reshape(DA_SHAPE) + 1)
+    custom_med_with_conductivity = td.Structure(
+        geometry=box,
+        medium=td.CustomMedium(
+            permittivity=td.SpatialDataArray(
+                eps_arr,
+                coords=da_coords,
+            ),
+            conductivity=td.SpatialDataArray(
+                conductivity_arr,
+                coords=da_coords,
             ),
         ),
     )
@@ -337,12 +354,7 @@ def make_structures(params: anp.ndarray) -> dict[str, td.Structure]:
     # custom medium with vector valued permittivity data
     eps_ii = td.ScalarFieldDataArray(
         eps_arr.reshape(nx, ny, nz, 1),
-        coords={
-            "x": np.linspace(-0.5, 0.5, nx),
-            "y": np.linspace(-0.5, 0.5, ny),
-            "z": np.linspace(-0.5, 0.5, nz),
-            "f": [td.C_0],
-        },
+        coords=da_coords | {"f": [td.C_0]},
     )
 
     custom_med_vec = td.Structure(
@@ -484,6 +496,7 @@ def make_structures(params: anp.ndarray) -> dict[str, td.Structure]:
         "center_list": center_list,
         "size_element": size_element,
         "custom_med": custom_med,
+        "custom_med_with_conductivity": custom_med_with_conductivity,
         "custom_med_vec": custom_med_vec,
         "polyslab": polyslab,
         "polyslab_dispersive": polyslab_dispersive,
@@ -581,6 +594,7 @@ structure_keys_ = (
     "center_list",
     "size_element",
     "custom_med",
+    "custom_med_with_conductivity",
     "custom_med_vec",
     "polyslab",
     "complex_polyslab",
@@ -1738,7 +1752,7 @@ def test_custom_pole_residue(monkeypatch):
     monkeypatch.setattr(
         td.CustomPoleResidue,
         "_derivative_field_cmp",
-        lambda self, E_der_map, eps_data, dim: dJ_deps / 3.0,
+        lambda self, E_der_map, spatial_data, dim, freqs, component="real": dJ_deps / 3.0,
     )
 
     import importlib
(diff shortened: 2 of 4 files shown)


tests/test_components/test_autograd.py
```diff
@@ -2351,3 +2377,52 @@ def objective(x):
...
+            sim,
+            path=str(tmp_path / "sim_test.hdf5"),
+            task_name="conductivity_only_grad_test",
+            verbose=False,
+        )
+        return postprocess(data, data[monitor.name])
+
+    val, grad = ag.value_and_grad(objective)(params0)
+
+    assert anp.all(grad != 0.0), "gradients are 0 for conductivity-only test"
```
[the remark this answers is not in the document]
yaugenst-flex: wow stickler for the details

tidy3d/components/medium.py
```diff
@@ -2912,10 +2930,25 @@ def _derivative_field_cmp(
...
 
-        # TODO: probably this could be more robust. eg if the DataArray has weird edge cases
-        E_der_dim_interp = (
-            E_der_dim_interp.interp(**coords_interp, assume_sorted=True).fillna(0.0).real.sum("f")
-        )
+        E_der_dim_interp_complex = E_der_dim_interp.interp(
+            **coords_interp, assume_sorted=True
+        ).fillna(0.0)
+
+        if component == "imag":
```
tylerflex: just to avoid extra operations in the code, I'd recommend utilizing one of these to do the conversions involving omega

https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/medium.py#L1205-L1249

tidy3d/components/medium.py
```diff
@@ -2866,15 +2881,18 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
     def _derivative_field_cmp(
         self,
         E_der_map: ElectromagneticFieldDataset,
-        eps_data: PermittivityDataset,
+        spatial_data: PermittivityDataset,
```
tylerflex: what's the reason for renaming this arg?

yaugenst-flex: Because we are passing both `self.permittivity` as well as `self.conductivity` to this method now

tidy3d/components/medium.py
yaugenst-flex: not really sure how to reuse that here? we are applying the scaling to the derivatives, not to the material

tylerflex: these two functions are reciprocal to each other, so basically I think you can just use 
```py
for freq in freqs:
    vjp = td.Medium.eps_sigma_to_eps_complex(eps_real=0, sigma=E_der_dim_interp.sel(f=freq, freq: freq)
    E_der_dim_interp.loc[{"f": freq}] = vjp.imag
```

yaugenst-flex: i understand that it works out but i find that harder to reason about actually. if i'm reading this for the first time i would not understand why we'd be calling `td.Medium.eps_sigma_to_eps_complex` on a derivative here

tylerflex: it's because the real permittivity and conductivity are computed from `eps_complex_to_eps_sigma` right? so the VJP basically needs to back propagate through this function by passing through the inverse. 

it's harder to reason about but it just feels more fragile to me to hard-code this derivative. what if we change `eps_complex_to_eps_sigma` to eg include more kwargs in the future?

ultimately it's a pretty minor point but I just feel uneasy about writing out the mathematical operations in the VJP functions when the forward passes are calling methods that have well defined inverses.

tidy3d/components/medium.py
tylerflex: sorry still dont understand, so `eps_data` is basically now either a (real-valued) dataset passing real permittivity or sigma now? even still `spatial_data` seems a bit strange, since it's still passing a `PermittivityDataset`? not eg just spatial coordinates, which is what it sounds like

yaugenst-flex: i mean, that type hint is not correct. the function receives a `SpatialDataArray` (via `CustomSpatialDataTypeAnnotated`)

tylerflex: ah ok, guess we should rename that then. makes more sense now.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
```
tylerflex: extra line added here FYI, greptile where were you on this one?

tidy3d/components/medium.py
```diff
@@ -2912,10 +2930,25 @@ def _derivative_field_cmp(
             # if sizes_list is empty, then reduce() fails
             d_vol = np.array(1.0)
 
-        # TODO: probably this could be more robust. eg if the DataArray has weird edge cases
-        E_der_dim_interp = (
-            E_der_dim_interp.interp(**coords_interp, assume_sorted=True).fillna(0.0).real.sum("f")
-        )
+        E_der_dim_interp_complex = E_der_dim_interp.interp(
+            **coords_interp, assume_sorted=True
```
tylerflex, line 2997: what happened to the sum over `f`? is it needed?

yaugenst-flex: it‘s still there but moved a few lines further down because in the conductivity case the omega factor needs to be applied per frequency before the sum happens

tidy3d/components/medium.py
yaugenst-flex: ok added the change

tylerflex (approved):
Looks good to me!

a really minor comment is that I notice that we `sum("f")` the `E_der_dim_interp` in each of the handlings of `component`. could be slightly more robust if w just sum that after constructing `E_der_dim_interp`.

yaugenst-flex:
> a really minor comment is that I notice that we `sum("f")` the `E_der_dim_interp` in each of the handlings of `component`. could be slightly more robust if w just sum that after constructing `E_der_dim_interp`.

good point, fixed

tests/test_components/test_autograd.py
```diff
@@ -2351,3 +2365,52 @@ def objective(x):
...
         g = ag.grad(objective)(1.0)
+
+
+def test_custom_medium_conductivity_only_gradient(rng, use_emulated_run, tmp_path):
+    """Test conductivity gradients for CustomMedium with constant permittivity."""
+
+    monitor, postprocess = make_monitors()["field_point"]
+
+    def objective(params):
+        """Objective function testing only gonductivity gradient (constant permittivity)."""
```
groberts-flex: gonductivity -> conductivity

groberts-flex (approved):
looks good! not sure this would be in the scope of the PR or not but would love to have a numerical test for conductivity as well. I'm happy to take this on at some point, think I could modify the field and/or mode numerical tests we have to check conductivity (right now, they use custom medium for permittivity, but should be quick change to also test conductivity.

daquinteroflex:
Hello! Is the plan to merge this by 2.9.0 ? We can add the `.1` tag possbly if not?

tidy3d/components/medium.py
```diff
@@ -2898,13 +2900,27 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
-            if field_path == ("permittivity",):
+            if field_path[0] == "permittivity":
                 vjp_array = 0.0
                 for dim in "xyz":
                     vjp_array += self._derivative_field_cmp(
                         E_der_map=derivative_info.E_der_map,
-                        eps_data=self.permittivity,
+                        spatial_data=self.permittivity,
                         dim=dim,
+                        freqs=np.atleast_1d(derivative_info.frequencies),
```
groberts-flex: do we need these atleast_1d calls anymore since derivative_info.frequencies is an array?

tidy3d/components/medium.py
```diff
@@ -2972,10 +2993,26 @@ def _derivative_field_cmp(
...
+        E_der_dim_interp_complex = E_der_dim_interp.interp(
+            **coords_interp, assume_sorted=True
+        ).fillna(0.0)
+
+        if component == "imag":
+            # convert from derivative w.r.t. complex permittivity to derivative w.r.t. conductivity
+            E_der_dim_interp = E_der_dim_interp_complex.imag
+            # frequency-dependent scaling must be applied before summing over frequencies
+            for freq in freqs:
+                vjp_imag = Medium.eps_sigma_to_eps_complex(
```
groberts-flex: thinking about this more and the imaginary gradient part is being computed with the scaling for the sigma gradient. I would tend towards moving that scaling to the part of the code that's specifically computing the sigma gradient from the imaginary gradient component. But I see the complication here of needing the frequency value to compute the sigma component and we are summing over frequencies in this function in the end. Maybe this doesn't matter that much, but would it be good to rename the `component` field here to indicate that this is `sigma` instead of `imag`?

groberts-flex (commented):
overall looks good, couple of small comments. 

did you want me to send you the fixed up numerical test to include here as well? or do you want to save that for a separate PR?

tidy3d/components/medium.py
yaugenst-flex: yeah i think you're right. i added a `sigma` component too. what do you think?

groberts-flex: thanks, looks great to me!

PR #2583, opened 2025-06-18
Fix to mode solver simulation reduction in web.upload

momchil-flex:
The simulation reduction was happening only *after* the simulation was already uploaded. :facepalm:

I don't know if this never worked right, or if it broke at some point - I think the latter.

We have a test [here](https://github.com/flexcompute/tidy3d/blob/82bf4baca3e2a6169eb5980f432e36fd435e84b0/tests/test_web/test_webapi_mode.py#L268-L269) that the reduction function works, but not that the reduced simulation is actually uploaded. I'm not sure if the latter is possible within our mock web tests?  

## Greptile Summary

Fixed a critical performance issue in mode solver simulations where the simulation reduction was incorrectly occurring after upload instead of before, causing unnecessary data transfer. This bug impacted the efficiency of mode solver tasks by uploading full-sized simulations when reduced versions would suffice.

- Modified `tidy3d/web/api/webapi.py` to perform mode solver simulation reduction before Tidy3dStub initialization and upload
- Existing test confirms reduction function works but lacks verification of reduced simulation upload
- Improves server-side efficiency by transmitting only the necessary simulation data

### DIFF
--- a/tests/test_web/test_webapi_mode.py
+++ b/tests/test_web/test_webapi_mode.py
@@ -108,10 +108,21 @@ def mock_upload(monkeypatch, set_api_key):
         status=200,
     )
 
+    # store the uploaded stub for verification
+    uploaded_stub = {}
+
+    def mock_upload_simulation(self, stub, **kwargs):
+        uploaded_stub["stub"] = stub
+
     def mock_upload_file(*args, **kwargs):
         pass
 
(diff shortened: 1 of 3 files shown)


yaugenst-flex (approved):
Huh, interesting 😄 LGTM

PR #2585, opened 2025-06-18
refc[adjoint]: Refactor PoleResidue derivative calculation

yaugenst-flex:
This PR removes the "mysterious" factors in the adjoint gradient calculation for `CustomPoleResidue` and refactors the derivative logic for both `PoleResidue` and `CustomPoleResidue`. Tested on the [gold nanoantenna notebook](https://github.com/flexcompute/tidy3d-notebooks/blob/develop/Autograd15Antenna.ipynb).

The PR removes more code than it adds so you know it's good 😉

## Greptile Summary

Refactors adjoint gradient calculations for PoleResidue models to use analytical derivatives instead of automatic differentiation, improving mathematical precision and code clarity.

- Replaced autograd with analytical derivative implementation in `tidy3d/components/medium.py` for PoleResidue calculations
- Introduced `_get_vjps_from_params` helper method to consolidate derivative logic for both PoleResidue and CustomPoleResidue
- Updated derivative testing in `tests/test_autograd.py` to properly handle complex conjugates
- Removed arbitrary numerical factors from CustomPoleResidue derivative calculations

### DIFF
--- a/tests/test_components/test_autograd.py
+++ b/tests/test_components/test_autograd.py
@@ -1650,18 +1650,18 @@ def test_custom_pole_residue(monkeypatch):
     custom_med_pole_res = td.CustomPoleResidue(eps_inf=eps_inf, poles=poles)
 
     def J(eps):
-        return anp.sum(abs(eps))
+        return anp.sum(anp.abs(eps))
 
     freq = 3e8
     pr = td.CustomPoleResidue(eps_inf=eps_inf, poles=poles)
     eps0 = pr.eps_model(freq)
 
-    dJ_deps = ag.holomorphic_grad(J)(eps0)
+    dJ_deps = np.conj(ag.holomorphic_grad(J)(eps0))
 
     monkeypatch.setattr(
         td.CustomPoleResidue,
         "_derivative_field_cmp",
-        lambda self, E_der_map, eps_data, dim, freqs: dJ_deps,
+        lambda self, E_der_map, eps_data, dim, freqs: dJ_deps / 3.0,
     )
 
     import importlib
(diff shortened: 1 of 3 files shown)


tidy3d/components/medium.py
```diff
@@ -3425,50 +3423,52 @@ def loss_upper_bound(self) -> float:
...
+                    deps_da = c_val / (jw + a_val) ** 2
+                    dJ_da = dJ_deps_complex * deps_da
+                    vjps[("poles", i, 0)] = dJ_da
+                if ("poles", i, 1) in requested_paths:
+                    deps_dc = -1 / (jw + a_val)
+                    dJ_dc = dJ_deps_complex * deps_dc
+                    vjps[("poles", i, 1)] = dJ_dc
+
+        return vjps
+
```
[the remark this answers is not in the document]
yaugenst-flex: Yes. That's what this does.

tests/test_components/test_autograd.py
```diff
@@ -1650,18 +1650,18 @@ def test_custom_pole_residue(monkeypatch):
...
     eps0 = pr.eps_model(freq)
 
-    dJ_deps = ag.holomorphic_grad(J)(eps0)
+    dJ_deps = np.conj(ag.holomorphic_grad(J)(eps0))
 
     monkeypatch.setattr(
         td.CustomPoleResidue,
         "_derivative_field_cmp",
-        lambda self, E_der_map, eps_data, dim, freqs: dJ_deps,
+        lambda self, E_der_map, eps_data, dim, freqs: dJ_deps / 3.0,
```
tylerflex, line 1664: where does this mysterious 3 come from ? :D

yaugenst-flex: This one isn't mysterious it's just correct 😄 `_compute_derivatives` calls `_derivative_field_cmp` three times and sums. So since we're feeding it a known total `dJ_deps`, the 1/3 factor is necessary to account for that.

yaugenst-flex: oooooooooh _that's_ where the mysterious 3 in the `CustomPoleResidue` derivative came from! it fixed the mistake in the test!

tylerflex: yea that sounds right! the mystery is solved!

tylerflex (approved):
approved, thanks yannick!

groberts-flex (approved):
great, I like this! like the simplification and re-use!

tests/test_components/test_autograd.py
```diff
@@ -1703,17 +1703,18 @@ def f(eps_inf, poles):
...
     assert np.allclose(grads_computed[("eps_inf",)], grad_eps_inf)
 
+    gfn = ag.holomorphic_grad(lambda x: f(eps_inf.values, x))
+    grad_poles = gfn(poles_complex)
+
     for i in range(len(poles)):
         for j in range(2):
             field_path = ("poles", i, j)
-            assert np.allclose(grads_computed[field_path], grad_poles[i][j])
+            assert np.allclose(grads_computed[field_path], np.conj(grad_poles[i][j]))
```
groberts-flex, line 1717: actually, quick question I forgot to ask is the reason for the conj here?

yaugenst-flex: yeah that's a good point. so fundamentally the reason is that the implementation uses standard complex derivatives, but autograd's `holomorphic_grad` uses the convention `df/dx - i df/dy`, so for real valued functions these are related by complex conjugation. this is _only_ needed for the test because we mock the field computations with `holomorphic_grad`, but the implementation in `components/` is self-consistent as far as i can tell

groberts-flex: cool, thanks for the explanation on this, makes sense

PR #2589, opened 2025-06-20
perf(autograd): optimize grey_dilation with striding

yaugenst-flex:
The previous implementation of `grey_dilation` was based on convolution, which was slow for both the forward and backward passes.

This PR replaces it with a high-performance implementation that uses NumPy's [`sliding_window_view`](https://numpy.org/doc/2.1/reference/generated/numpy.lib.stride_tricks.sliding_window_view.html#numpy.lib.stride_tricks.sliding_window_view) to create sliding window views of the input array. I also wrote a custom VJP that uses the same striding technique to make the backward pass faster too.

I also simplified the implementation of `grey_erosion` so that `grey_dilation` is now the only function that does the heavy lifting.

Benchmarks show speedups of 10-100x depending on the array and kernel size.

dilation_image_scaling
dilation_kernel_scaling

This should make these ops much more usable in topopt @groberts-flex

## Greptile Summary

Significant performance optimization of the `grey_dilation` morphological operation by replacing convolution-based implementation with NumPy's `sliding_window_view` for strided array operations.

- Replaced convolution-based implementation with strided array approach in `tidy3d/plugins/autograd/functions.py`, achieving 10-100x speedup
- Added custom VJP (vector-Jacobian product) for efficient backpropagation using the same striding technique
- Simplified `grey_erosion` by expressing it through duality with `grey_dilation`
- Updated morphology test cases to use first-order gradients and include kernel structure testing
- Added comprehensive benchmarks showing performance improvements scaling with array and kernel sizes

### DIFF
--- a/tidy3d/plugins/autograd/functions.py
+++ b/tidy3d/plugins/autograd/functions.py
@@ -208,6 +208,8 @@ def _get_footprint(size, structure, maxval):
     """Helper to generate the morphological footprint from size or structure."""
     if size is None and structure is None:
         raise ValueError("Either size or structure must be provided.")
+    if size is not None and structure is not None:
+        raise ValueError("Cannot specify both size and structure.")
     if structure is None:
         size_np = onp.atleast_1d(size)
         shape = (size_np[0], size_np[-1]) if size_np.size > 1 else (size_np[0], size_np[0])
@@ -238,8 +240,11 @@ def grey_dilation(
         The input array to perform grey dilation on.
     size : Union[Union[int, tuple[int, int]], None] = None
         The size of the structuring element. If None, `structure` must be provided.
+        If a single integer is provided, a square structuring element is created.
+        For 1D arrays, use a tuple (size, 1) or (1, size) for horizontal or vertical operations.
     structure : Union[np.ndarray, None] = None
         The structuring element. If None, `size` must be provided.
+        For 1D operations on 2D arrays, use a 2D structure with one dimension being 1.
     mode : PaddingType = "reflect"
         The padding mode to use.
     maxval : float = 1e4
@@ -286,6 +291,10 @@ def _vjp_maker_dilation(ans, array, size=None, structure=None, *, mode="reflect"
     is_max_mask = (dilated_windows == output_reshaped).astype(onp.float64)
 
     # normalize the gradient for cases where multiple elements are the maximum.
+    # When multiple elements in a window equal the maximum value, the gradient
+    # is distributed equally among them. This ensures gradient conservation.
+    # Note: Values can never exceed maxval in the output since we add structure
+    # values (capped at maxval) to the input array values.
     multiplicity = onp.sum(is_max_mask, axis=(-2, -1), keepdims=True)
     is_max_mask /= onp.maximum(multiplicity, 1)
 
--- a/tidy3d/plugins/autograd/functions.py
+++ b/tidy3d/plugins/autograd/functions.py
@@ -4,9 +4,13 @@ from collections.abc import Iterable
 from typing import Callable, Literal, Union
 
 import autograd.numpy as np
+import numpy as onp
 from autograd import jacobian
+from autograd.extend import defvjp, primitive
 from autograd.scipy.signal import convolve as convolve_ag
 from autograd.scipy.special import logsumexp
+from autograd.tracer import getval
+from numpy.lib.stride_tricks import sliding_window_view
 from numpy.typing import NDArray
 
 from tidy3d.components.autograd.functions import add_at, interpn, trapz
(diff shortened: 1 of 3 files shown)


tidy3d/plugins/autograd/functions.py
```diff
@@ -238,9 +204,27 @@ def convolve(
     return convolve_ag(array, kernel, axes=axes, mode=mode)
 
 
+def _get_footprint(size, structure, maxval):
+    """Helper to generate the morphological footprint from size or structure."""
+    if size is None and structure is None:
+        raise ValueError("Either size or structure must be provided.")
+    if structure is None:
+        size_np = onp.atleast_1d(size)
+        shape = (size_np[0], size_np[-1]) if size_np.size > 1 else (size_np[0], size_np[0])
```
groberts-flex, line 215: in the case of a 1D structuring element or size, does this mean the dilation/erosion gets applied in 2D and then just a single dimension is extracted at the end?

tidy3d/plugins/autograd/functions.py
```diff
@@ -238,9 +204,27 @@ def convolve(
     return convolve_ag(array, kernel, axes=axes, mode=mode)
 
 
+def _get_footprint(size, structure, maxval):
+    """Helper to generate the morphological footprint from size or structure."""
+    if size is None and structure is None:
```
groberts-flex, line 209: possibly the other case to catch is that both the size and the structure are specified

tidy3d/plugins/autograd/functions.py
```diff
@@ -269,27 +253,81 @@ def grey_dilation(
...
-        nb = np.zeros_like(structure)
-    else:
-        nb = np.copy(structure)
-        nb[structure == 0] = -maxval
+    padded_array = pad(array, (h // 2, h // 2), mode=mode, axis=0)
+    padded_array = pad(padded_array, (w // 2, w // 2), mode=mode, axis=1)
 
+    padded_array_np = getval(padded_array)
+
+    windows = sliding_window_view(padded_array_np, window_shape=(h, w))
```
groberts-flex, line 271: checking my understanding - this replaces the convolve call in the previous version which was being used to essentially create this same view but doing so through a bunch of unnecessary computation with an identity kernel?

tidy3d/plugins/autograd/functions.py
```diff
@@ -269,27 +253,81 @@ def grey_dilation(
...
+    dilated_windows = windows + nb
+
+    output_reshaped = ans[..., None, None]
+    is_max_mask = (dilated_windows == output_reshaped).astype(onp.float64)
+
+    # normalize the gradient for cases where multiple elements are the maximum.
+    multiplicity = onp.sum(is_max_mask, axis=(-2, -1), keepdims=True)
+    is_max_mask /= onp.maximum(multiplicity, 1)
```
groberts-flex, line 299: not fully understanding this part - is it possible for values to come out of the operation as the max_val?

groberts-flex (commented):
thanks for this implementation, the speed up looks awesome especially for a function that will be in a lot of robust optimizations!


left some comments/questions, some just for my own understanding!

tidy3d/plugins/autograd/functions.py
yaugenst-flex: The operation is still applied to the full 2D array, but with a 1D-like structuring element. You always need to define a 2D structuring element of shape like `(1, size)` or `(size, 1)`. The dilation/erosion
  operation slides this structuring element across the entire 2D array, effectively performing the operation along rows or columns respectively. I updated the docstring to clarify this, because we don't explicitly handle 1d structuring elements.

tidy3d/plugins/autograd/functions.py
yaugenst-flex: yeah good catch. i think this was discussed previously at some point and we went for silently having precedence of structuring element over sizes, but since it came up again i added a check instead to forbid this

tidy3d/plugins/autograd/functions.py
yaugenst-flex: yes, exactly. the previous implementation would create the sliding windows (which we now do directly), multiply each window element-wise with the kernel, and then sum the results. you can see how this scales horrendously with image and kernel sizes 😄

tidy3d/plugins/autograd/functions.py
yaugenst-flex: no, values cannot equal `max_val` in the output. i added a comment explaining this:
```
# Note: Values can never exceed maxval in the output since we add structure
# values (capped at maxval) to the input array values.
```

groberts-flex (approved):
thanks @yaugenst-flex this is great! I understand now much better the multiplicity/multiple maximum part reading through the updated comment. thanks for the clarifications and changes.

PR #2590, opened 2025-06-20
Partially reverting #2534 and reorganizing mode validation

momchil-flex:
## Greptile Summary

Reorganizes mode validation logic across the codebase by extracting validation from ModeSolver into dedicated functions and consolidating checks for mode sources and monitors.

- Added `validate_mode_object` helper in `tidy3d/components/simulation.py` to centralize validation for both mode sources and monitors
- Removed redundant mode plane radius validation from `ModeSimulation._post_init_validators` in favor of consolidated checks
- Updated `test_validation_from_simulation()` to validate anisotropic media and infinite geometries against mode sources instead of monitors
- Simplified structure rotation validation in `ModeSolver` by extracting it into a reusable static helper function

### DIFF
--- a/tests/test_components/test_mode.py
+++ b/tests/test_components/test_mode.py
@@ -89,6 +89,11 @@ def test_validation_from_simulation():
         monitors=[],
     )
 
+    reg_geometry = td.Structure(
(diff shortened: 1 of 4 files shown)


yaugenst-flex (approved):
Awesome thanks, already ported over to the v2 branch

PR #2592, opened 2025-06-20
Fix to plane wave with negative azimuth angle

### LINKED ISSUE #2591: Negative polar angle leads to incorrect bloch BCs
**Describe the bug**
Plane wave sources with a negative polar angle seem to mess up the definition of bloch boundaries. Making the polar angle the same but positive results in expected results

**To Reproduce**
This simulation for example:
https://tidy3d.simulation.cloud/workbench?taskId=pa-65fb5c76-4106-451f-a550-683aee990ddc

**Expected behavior**
This type of wave propagation
Image

**Screenshots**
Image

momchil-flex:
Fixes #2591 but requires a solver that's built with the fix.

## Greptile Summary

Fixed plane wave source handling of negative azimuth angles, particularly important for Bloch boundary conditions.

- Modified `_angle_theta_actual` in `tidy3d/components/beam.py` to preserve sign in arcsin calculation for correct propagation with negative polar angles
- Change requires corresponding solver update to function properly
- Directly addresses issue #2591 affecting plane wave sources with Bloch boundaries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- Arrow lengths are now scaled consistently in the X and Y directions,
-  and their lengths no longer exceed the height of the plot window.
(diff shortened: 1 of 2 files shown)


alec-flexcompute (approved):
Looks good @momchil-flex! Straightforward fix

PR #2596, opened 2025-06-23
clean up polygon intersection glitches VERSION 2 (using only shapely functions) (issue #2560)

jewettaijfc:
## This PR addresses issue #2560 

This diff uses several different functions from the `shapely` to cleanup polygons with thin or tiny hair-like protrusions.

Such features are often generated when computing intersections of polygons *(using `(ClipOperation.intersections_plane()`)*:  At certain rotation angles, polygons which would overlap exactly fail to be aligned due to floating-point rounding errors that occur during rotation.  This was causing visual glitches mentioned in [that issue](https://github.com/flexcompute/tidy3d/issues/2560), and shown below:

## *Before this PR*
```
box  = td.Box(center = (0,-9.01,10),
              size = (10,0.01,10))
cylinder = td.Cylinder(radius = 9,
                       center = (0,-9.01,0),
                       length=0.01,
                       axis = 1)
ax = (box-cylinder).rotated(np.pi/2,axis = 0).plot(x=0)
ax.set_aspect('auto') 
```
Image

## *After this PR*
tidy3d_pr2581_after

NOTE: An earlier PR attempted to solve this problem [PR2581](https://github.com/flexcompute/tidy3d/pull/2581), but I decided to pursue this approach instead.

## Test files
Here is a file with more tests: 
[test_pr2596.ipynb.zip](https://github.com/user-attachments/files/20929827/test_pr2596.ipynb.zip)

(You must unzip it first.)

## Greptile Summary

Implements polygon cleanup using `shapely.set_precision()` to eliminate thin triangles from polygon boundaries in 2D plots, offering a simpler but slower alternative to PR#2581.

- Added `cleanup_shapely_polygon()` function in `tidy3d/components/geometry/base.py` using shapely's built-in precision control
- Modified `ClipOperation.to_polygon_list()` with optional cleanup parameter for removing visualization artifacts
- Added comprehensive test suite in `tests/test_components/test_geometry.py` covering edge cases
- Performance impact.  The complicated benchmark example now requires 0.23 seconds, vs. 0.064s before this PR (or 0.1s using PR#2581).

### DIFF
--- a/tests/test_components/test_geometry.py
+++ b/tests/test_components/test_geometry.py
@@ -14,6 +14,8 @@ import shapely
 import trimesh
 
 import tidy3d as td
+from tidy3d.compat import _shapely_is_older_than
+from tidy3d.components.geometry.base import cleanup_shapely_object
 from tidy3d.components.geometry.mesh import AREA_SIZE_THRESHOLD
 from tidy3d.components.geometry.utils import (
     SnapBehavior,
@@ -1249,3 +1251,50 @@ def test_triangle_mesh_from_height():
     error_message = str(excinfo.value)
     assert f"shape {expected_shape}" in error_message
     assert "shape (3, 3)" in error_message
+
+
+def test_cleanup_shapely_object():
+    if _shapely_is_older_than("2.1"):
+        # (Old versions of shapely don't support `shapely.make_valid()` with the correct arguments.
+        # However older alternatives like `.buffer(0)` are not as robust.  `.buffer(0)` is likely
+        # to generate polygons which look correct, but have extra vertices, causing test to fail.
+        # So if `shapely.make_valid()` is not supported, the safest thing to do is skip this test.)
+        pytest.skip("This test requires `shapely` version 2.1 or later")
+
+    # Test 1: A square containing a triangular hole, and an infinitley thin hole.
+    square_with_spikes_5x5 = np.array(
+        (
+            (0, 0),
+            (5, 0),
+            (5, 0),
+            (5, 10),  # this vertex creates an outward spike and should be removed
+            (5, 5),
+            (0, 5 - 1e-13),  # this vertex should be rounded to (0, 5)
+            (3, 3),  # this vertex creates an inward spike and should be removed
+            (0, 5 + 1e-13),  # this vertex should be removed because it duplicates (0, 5 - 1e-13)
+        )
+    )
+    triangle_empty_tails = np.array(((1, 1), (3, 1), (2, 2), (2.5, 2.5), (0.5, 0.5)))
+    triangle_collinear = np.array(((4, 2), (3, 3), (2, 4), (4, 2)))  # has zero area
+    # NOTE: Test will fail for self intersecting polys like: ((0,0), (1,1), (2,-1), (3,1), (4,0))
+    # Now build a shapely polygon with the 4 small polygons enclosed by big_square_5x5
+    exterior_coords = square_with_spikes_5x5
+    interior_coords_list = [
+        triangle_empty_tails,
+        triangle_collinear,  # this triangle should be eliminated
+    ]
+    # Test using a non-empty exterior polygon (big_square_5x5)
+    orig_polygon = shapely.Polygon(exterior_coords, interior_coords_list)
+    new_polygon = cleanup_shapely_object(orig_polygon, tolerance_ratio=1e-12)
+    # Delete any nearby or overlapping vertices (cleanup_shapely_object() does not do this).
+    new_polygon = shapely.simplify(new_polygon, tolerance=1e-10)
+    # Now `new_polygon` should only contain the coordinates of the square (with a duplicate at end).
+    assert len(new_polygon.exterior.coords) == 5  # squares have 4 vertices but shapely adds 1
+    assert len(new_polygon.interiors) == 1  # only the "triangle_empty_tails" interior hole survives
+    assert len(new_polygon.interiors[0].coords) == 4  # triangles have 3 vertices but shapely adds 1
+    # Test 2: An infinitely thin triangle exterior with some holes (which should be deleted)
+    exterior_coords = triangle_collinear  # has zero area
+    orig_polygon = shapely.Polygon(exterior_coords)
+    new_polygon = cleanup_shapely_object(orig_polygon, tolerance_ratio=1e-12)
+    new_polygon = shapely.simplify(new_polygon, tolerance=1e-10)
+    assert len(new_polygon.exterior.coords) == 0  # empty / collinear polygons should get deleted
--- a/tidy3d/compat.py
+++ b/tidy3d/compat.py
@@ -2,9 +2,23 @@
 
 from __future__ import annotations
 
+import importlib
+
+from packaging.version import parse as parse_version
+
 try:
     from xarray.structure import alignment
 except ImportError:
     from xarray.core import alignment
+
+
+_SHAPELY_VERSION = parse_version(importlib.metadata.version("shapely"))
+
+
+def _shapely_is_older_than(version: str) -> bool:
+    if _SHAPELY_VERSION < parse_version(version):
+        return True
+    return False
+
 
 __all__ = ["alignment"]
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -17,6 +17,7 @@ try:
 except ImportError:
     pass
 
+from tidy3d.compat import _shapely_is_older_than
 from tidy3d.components.autograd import AutogradFieldMap, TracedCoordinate, TracedSize, get_static
 from tidy3d.components.autograd.derivative_utils import DerivativeInfo, integrate_within_bounds
 from tidy3d.components.base import Tidy3dBaseModel, cached_property
@@ -63,6 +64,7 @@ from tidy3d.packaging import verify_packages_import
 from .bound_ops import bounds_intersection, bounds_union
 
 POLY_GRID_SIZE = 1e-12
+POLY_TOLERANCE_RATIO = 1e-12
 
 
 _shapely_operations = {
@@ -2862,27 +2864,45 @@ class ClipOperation(Geometry):
         return val
 
     @staticmethod
-    def to_polygon_list(base_geometry: Shapely) -> list[Shapely]:
+    def to_polygon_list(base_geometry: Shapely, cleanup: bool = False) -> list[Shapely]:
         """Return a list of valid polygons from a shapely geometry, discarding points, lines, and
-        empty polygons.
+        empty polygons, and empty triangles within polygons.
 
         Parameters
         ----------
         base_geometry : shapely.geometry.base.BaseGeometry
             Base geometry for inspection.
+        cleanup: bool = False
+            If True, removes extremely small features from each polygon's boundary.
+            This is useful for removing artifacts from 2D plots displayed to the user.
 
         Returns
         -------
         List[shapely.geometry.base.BaseGeometry]
             Valid polygons retrieved from ``base geometry``.
         """
+        unfiltered_geoms = []
         if base_geometry.geom_type == "GeometryCollection":
-            return [p for geom in base_geometry.geoms for p in ClipOperation.to_polygon_list(geom)]
+            unfiltered_geoms = [
+                p for geom in base_geometry.geoms for p in ClipOperation.to_polygon_list(geom)
+            ]
         if base_geometry.geom_type == "MultiPolygon":
-            return [p for p in base_geometry.geoms if not p.is_empty]
+            unfiltered_geoms = [p for p in base_geometry.geoms if not p.is_empty]
         if base_geometry.geom_type == "Polygon" and not base_geometry.is_empty:
-            return [base_geometry]
-        return []
+            unfiltered_geoms = [base_geometry]
+        geoms = []
+        if cleanup:
+            # Optional: "clean" each of the polygons (by removing extremely small or thin features).
+            for geom in unfiltered_geoms:
+                geom_clean = cleanup_shapely_object(geom)
+                if geom_clean.geom_type == "Polygon":
+                    geoms.append(geom_clean)
+                if geom_clean.geom_type == "MultiPolygon":
+                    geoms += [p for p in geom_clean.geoms if not p.is_empty]
+                # Ignore other types of shapely objects (points and lines)
+        else:
+            geoms = unfiltered_geoms
+        return geoms
 
     @property
     def _shapely_operation(self) -> Callable[[Shapely, Shapely], Shapely]:
@@ -2931,7 +2951,10 @@ class ClipOperation(Geometry):
         b = self.geometry_b.intersections_tilted_plane(normal, origin, to_2D)
         geom_a = shapely.unary_union([Geometry.evaluate_inf_shape(g) for g in a])
         geom_b = shapely.unary_union([Geometry.evaluate_inf_shape(g) for g in b])
-        return ClipOperation.to_polygon_list(self._shapely_operation(geom_a, geom_b))
+        return ClipOperation.to_polygon_list(
+            self._shapely_operation(geom_a, geom_b),
+            cleanup=True,
+        )
 
     def intersections_plane(
         self, x: Optional[float] = None, y: Optional[float] = None, z: Optional[float] = None
@@ -2958,7 +2981,10 @@ class ClipOperation(Geometry):
         b = self.geometry_b.intersections_plane(x, y, z)
         geom_a = shapely.unary_union([Geometry.evaluate_inf_shape(g) for g in a])
         geom_b = shapely.unary_union([Geometry.evaluate_inf_shape(g) for g in b])
-        return ClipOperation.to_polygon_list(self._shapely_operation(geom_a, geom_b))
+        return ClipOperation.to_polygon_list(
+            self._shapely_operation(geom_a, geom_b),
+            cleanup=True,
+        )
 
     @cached_property
     def bounds(self) -> Bound:
@@ -3291,6 +3317,94 @@ class GeometryGroup(Geometry):
             grad_vjps[field_path] = grad_vjp_values[0]
 
         return grad_vjps
+
+
+def cleanup_shapely_object(obj: Shapely, tolerance_ratio: float = POLY_TOLERANCE_RATIO) -> Shapely:
+    """Remove small geometric features from the boundaries of a shapely object including
+    inward and outward spikes, thin holes, and thin connections between larger regions.
+
+    Parameters
+    ----------
+    obj : shapely
+        a shapely object (typically a ``Polygon`` or a ``MultiPolygon``)
+    tolerance_ratio : float = ``POLY_TOLERANCE_RATIO``
+        Features on the boundaries of polygons will be discarded if they are smaller
+        or narrower than ``tolerance_ratio`` multiplied by the size of the object.
+
+    Returns
+    -------
+    Shapely
+        A new shapely object whose small features (eg. thin spikes or holes) are removed.
+
+    Notes
+    -----
+    This function does not attempt to delete overlapping, nearby, or collinear vertices.
+    To solve that problem, use ``shapely.simplify()`` afterwards.
+    """
+    if _shapely_is_older_than("2.1"):
+        log.warning(
+            "Using old versions of the shapely library (prior to v2.1) may cause "
+            "plot errors.  This can be solved by upgrading to Python 3.10 "
+            "(or later) and reinstalling Tidy3d.",
+            log_once=True,
+        )
+        return obj
+    if obj.is_empty:
+        return obj
+    centroid = obj.centroid
+    object_size = min(obj.bounds[2] - obj.bounds[0], obj.bounds[3] - obj.bounds[1])
+    if object_size == 0.0:
+        return shapely.Polygon([])
+    # In order to prevent numerical overflow or underflow errors, we first subtract
+    # the centroid and divide by (rescale) the size of the object so it is not too big.
+    normalized_obj = shapely.affinity.affine_transform(
+        # https://shapely.readthedocs.io/en/stable/manual.html#affine-transformations
+        obj,
+        matrix=[
+            1 / object_size,
+            0.0,
+            0.0,
+            1 / object_size,
+            -centroid.x / object_size,
+            -centroid.y / object_size,
+        ],
+    )
+    # Important: Remove any self intersections beforehand using `shapely.make_valid()`.
+    valid_obj = shapely.make_valid(normalized_obj, method="structure", keep_collapsed=False)
+    # To get rid of small thin features, erode(shrink), dilate(expand), and erode again.
+    eroded_obj = shapely.buffer(  # This removes outward spikes
+        valid_obj,
+        distance=-tolerance_ratio,
+        cap_style="square",  # (optional parameter to reduce computation time)
+        quad_segs=3,  # (optional parameter to reduce computation time)
+    )
+    dilated_obj = shapely.buffer(  # This removes inward spikes and tiny holes
+        eroded_obj,
+        distance=2 * tolerance_ratio,
+        cap_style="square",
+        quad_segs=3,
+    )
+    cleaned_obj = dilated_obj
+    # Optional: Now shrink the polygon back to the original size.
+    cleaned_obj = shapely.buffer(
+        cleaned_obj,
+        distance=-tolerance_ratio,
+        cap_style="square",
+        quad_segs=3,
+    )
+    # Revert to the original scale and position.
+    rescaled_clean_obj = shapely.affinity.affine_transform(
+        cleaned_obj,
+        matrix=[
+            object_size,
+            0.0,
+            0.0,
+            object_size,
+            centroid.x,
+            centroid.y,
+        ],
+    )
+    return rescaled_clean_obj
 
 
 from .utils import GeometryType, from_shapely, vertices_from_shapely  # noqa: E402
--- a/tidy3d/components/viz/descartes.py
+++ b/tidy3d/components/viz/descartes.py
@@ -35,7 +35,10 @@ class Polygon:
     @property
     def exterior(self):
         """Get polygon exterior."""
-        return getattr(self.context, "exterior", None) or self.context[0]
+        value = getattr(self.context, "exterior", None)
+        if value is None:
+            value = self.context[0]
+        return value
 
     @property
     def interiors(self):
@@ -53,9 +56,13 @@ def polygon_path(polygon):
     def coding(obj):
         # The codes will be all "LINETO" commands, except for "MOVETO"s at the
         # beginning of each subpath
-        n = len(getattr(obj, "coords", None) or obj)
+        crds = getattr(obj, "coords", None)
+        if crds is None:
+            crds = obj
+        n = len(crds)
         vals = ones(n, dtype=Path.code_type) * Path.LINETO
-        vals[0] = Path.MOVETO
+        if len(vals) > 0:
+            vals[0] = Path.MOVETO
         return vals
 
     ptype = polygon.geom_type
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,17 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for `np.unwrap` in `tidy3d.plugins.autograd`.
 - Add Nunley variant to germanium material library based on Nunley et al. 2016 data.
 
-### Changed
-- Switched to an analytical gradient calculation for spatially-varying pole-residue models (`CustomPoleResidue`).
-
-### Changed
-- Significantly improved performance of the `tidy3d.plugins.autograd.grey_dilation` morphological operation and its gradient calculation. The new implementation is orders of magnitude faster, especially for large arrays and kernel sizes.
-
 ### Fixed
 - Arrow lengths are now scaled consistently in the X and Y directions, and their lengths no longer exceed the height of the plot window.
 - Bug in `PlaneWave` defined with a negative `angle_theta` which would lead to wrong injection.
+- Plots of objects defined by shape intersection logic will no longer display thin line artifacts.
 
 ### Changed
+- Switched to an analytical gradient calculation for spatially-varying pole-residue models (`CustomPoleResidue`).
+- Significantly improved performance of the `tidy3d.plugins.autograd.grey_dilation` morphological operation and its gradient calculation. The new implementation is orders of magnitude faster, especially for large arrays and kernel sizes.
 - `GaussianBeam` and `AstigmaticGaussianBeam` default `num_freqs` reset to 1 (it was set to 3 in v2.8.0) and a warning is issued for a broadband, angled beam for which `num_freqs` may not be sufficiently large.
 - Set the maximum `num_freqs` to 20 for all broadband sources (we have been warning about the introduction of this hard limit for a while).
 


jewettaijfc:
I just found an example (described in [issue 2598](https://github.com/flexcompute/tidy3d/issues/2598)) causing the assertion I made in this PR to fail.  The first line of code can split a polygon into mulitple polygons:
```
cleaned_polygon = shapely.set_precision(shapely.Polygon(vertices).buffer(0), POLY_GRID_SIZE) 
assert isinstance(cleaned_polygon, shapely.Polygon)
```
...so the assertion on the next line line will fail in this case.  (In that example `cleaned_polygon` will be a `shapely.MultiPolygon`, not a `shapely.Polygon`.)  I need to add a few lines of code to deal with this issue.

jewettaijfc:
I fixed the previous issue.  It's okay to review this PR.

jewettaijfc:
### New failure type
I've been testing the PR for several hours today, and I found a more complicated example where both of approaches I've tried so far (PR2596 and PR2581) fail at certain rotation angles.
- Start with a 2x3 box.
- Remove a 1x2 box from it's lower-left corner ("bite1").
- Remove a 1x2 box from it's upper-right corner ("bite2").

Image
```
box2x3_outer = td.Box(center = (1, 1.5, 0), size = (2,3,1))
box1x2_bite1 = td.Box(center = (0.5, 1, 0), size = (1,2,1))
box1x2_bite2 = td.Box(center = (1.5, 2, 0), size = (1,2,1))

split_box = (box2x3_outer - box1x2_bite1) - box1x2_bite2

split_box.rotated(0.23*np.pi/2, axis = 2).plot(z=0)
```
You *should* see two squares.  Unfortunately, this is what we sometimes see:

tidyed_pr2596_failure1
tidy3d_pr2596_failure2

I tried increasing the grid size used for geometry simplification, but it did not help that much for this example.  There were always some angles where this example had visual glitches.

## Possible issues with our existing code?


As Weiliang points out, I am currently using [the same approach we use elsewhere in Tidy3D](https://github.com/flexcompute/tidy3d/pull/2581#issuecomment-2993325491) to cleanup polygons:
```
shapely.set_precision(shapely.Polygon(vertices).buffer(0), POLY_GRID_SIZE)
```
This trick helps most of the time, but fails in the example above.

**UPDATE: I no longer use this approach.  (See below.)**

jewettaijfc:
## UPDATE: I fixed this problem by changing the approach

To solve these problems, I now use buffering (a.k.a. "erosion", "dilation") with *non-zero distance* parameters (ie 1e-12).

What that means in simple language:

1) First I "erode" (srhink) the polygons by a small distance (1e-12) to get rid of outward spikes.
2) Then I "dilate" (expand) the (previously eroded) polygons (by a ***larger*** distance 2e-12) to get rid of inward spikes.
3) Then I erode/shrink the polygons again (by 1e-12) back to the original size.

Here is a simplified version of the relevant code.
```
eroded = shapely.buffer(original_polygon, distance=-resolution)  # removes outward spikes
dilated = shapely.buffer(eroded, distance=2 * resolution)  # removes inward spikes
cleaned = shapely.buffer(dilated, distance=resolution)  # expands it back to the original size
```

This will get rid of spikes that protrude either *outward or inward*, and it will handle changes in topology like the two-square example above.

This seems to work really well.

(Perhaps we should update the rest of the other polygon-cleaning code to do it this way?)

jewettaijfc:
I just noticed the unit test failures.  I'll try and fix that in the next hour.

*(UPDATE: "next hour"   ha ha)*

jewettaijfc:
I finally fixed all of the issues which were causing unit test failures.  (Lots of failures!)
This PR is ready for review again!

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3319,40 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+    """
+    exterior_coords = np.asarray(p.exterior.coords)
+    interior_coords_list = [np.asarray(interior_ring.coords) for interior_ring in p.interiors]
+    cleaned_polygon = shapely.set_precision(
+        shapely.Polygon(exterior_coords, interior_coords_list).buffer(0),
+        resolution,
+    )
+    # Assertion fails when polygon intersects itself (eg: ((0,0), (1,1), (2,-1), (3,1), (4,0)))
+    assert isinstance(cleaned_polygon, shapely.Polygon)
+    return cleaned_polygon
```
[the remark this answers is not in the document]
jewettaijfc: ## *UPDATE: THIS CODE EXCERPT IS OUTDATED.*
Please ignore that code excerpt.  (Sorry.  A lot of changes have been made since this PR was first posted.)

tests/test_components/test_geometry.py
```diff
@@ -211,6 +214,10 @@ def test_intersections_plane(component):
     assert len(component.intersections_plane(x=10000)) == 0
 
 
+def test_intersections_plane_DELETEME():  # DEBUG - REMOVE THIS FUNCTION BY 2025-6-25 -Andrew
+    assert len(SYM_DIFFERENCE.intersections_plane(z=0.2)) > 0
```
yaugenst-flex: Delete I guess?

tests/test_components/test_geometry.py
```diff
@@ -14,6 +14,9 @@
 import trimesh
 
 import tidy3d as td
+from tidy3d.components.geometry.base import (
+    cleanup_shapely_object,
```
yaugenst-flex: nit: remove comma after `cleanup_shapely_object`

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3319,92 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+            -centroid.y / object_size,
+        ],
+    )
+    # Important: Remove any self intersections beforehand using `shapely.make_valid()``
+
+    try:
+        valid_obj = shapely.make_valid(normalized_obj, method="structure", keep_collapsed=False)
+    except (TypeError, AttributeError):
+        # (Old versions of shapely don't support `shapely.make_valid()`.  Use buffer(0) instead.)
+        valid_obj = normalized_obj.buffer(0)
```
yaugenst-flex: This seems strange. As far as I can tell, `.make_valid()` was introduced in shapely `1.8a1 (2021-03-03)`, and we pin `shapely = "^2.0"`. Could you double check whether this is really the reason that the tests fail on python 3.9? You can `poetry env use python3.9` to set up your env with a specific python version.

yaugenst-flex: Same comment for the tests too.

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3319,92 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
         return grad_vjps
 
 
+def cleanup_shapely_object(obj: Shapely, resolution: float = POLY_GRID_SIZE) -> Shapely:
```
yaugenst-flex: The function assumes that `obj` is not empty, not sure if the function would crash if it is? Might be good to add a guard like
```py
if obj.is_empty:
  return obj
```

tests/test_components/test_geometry.py
```diff
@@ -1249,3 +1256,59 @@ def wrong_shape_height_func(x, y):
...
+    new_polygon = cleanup_shapely_object(orig_polygon, resolution=1e-12)
+    try:
+        new_polygon = shapely.make_valid(
+            shapely.set_precision(new_polygon, 1e-10),
+            method="structure",
+            keep_collapsed=False,
+        )
+    except (TypeError, AttributeError):
+        # (Old versions of shapely don't support `shapely.make_valid()`.)
+        return
```
yaugenst-flex: This will never be reached if `shapely.make_valid()` errors. The comment below regarding `shapely.make_valid()` notwithstanding, this should probably be revisited.

tests/test_components/test_geometry.py
```diff
@@ -1249,3 +1256,59 @@ def wrong_shape_height_func(x, y):
...
+    # Delete any nearby or overlapping vertices (cleanup_shapely_object() does not do this).
+    try:
+        new_polygon = shapely.make_valid(
+            shapely.set_precision(new_polygon, 1e-10),
+            method="structure",
+            keep_collapsed=False,
+        )
+    except (TypeError, AttributeError):
+        # (Old versions of shapely don't support `shapely.make_valid()`.)
+        return
```
yaugenst-flex: Don't `return` here but instead [`pytest.skip`](https://docs.pytest.org/en/stable/how-to/skipping.html#skipping-test-functions) with the reason, otherwise it looks like the test passed.

tidy3d/components/geometry/base.py
```diff
@@ -2862,27 +2862,47 @@ def _geometries_untraced(cls, val):
         return val
 
     @staticmethod
-    def to_polygon_list(base_geometry: Shapely) -> list[Shapely]:
+    def to_polygon_list(base_geometry: Shapely, cleanup_afterwards: bool = False) -> list[Shapely]:
```
yaugenst-flex: I think the `afterwards` here is an implementation detail that the caller of the function does not need to know about. I'd suggest just naming this `cleanup`.

tidy3d/components/geometry/base.py
```diff
@@ -2862,27 +2862,47 @@ def _geometries_untraced(cls, val):
...
+            (Vertices will be rounded to the nearest grid point with spacing ``POLY_GRID_SIZE``,
+            and any zero-area triangles that result will be deleted.)  This is useful for removing
+            artifacts from 2D plots displayed to the user.
 
         Returns
         -------
         List[shapely.geometry.base.BaseGeometry]
             Valid polygons retrieved from ``base geometry``.
         """
+        unfiltered_geoms: list[Shapely] = []
```
yaugenst-flex: We typically don't type hint within functions. Doesn't hurt I guess.

tidy3d/components/viz/descartes.py
```diff
@@ -35,7 +35,10 @@ def __init__(self, context):
     @property
     def exterior(self):
         """Get polygon exterior."""
-        return getattr(self.context, "exterior", None) or self.context[0]
+        value = getattr(self.context, "exterior", None)
+        if value is None:
+            value = self.context[0]
+        return value
```
yaugenst-flex, line 41: Why change this?

tidy3d/components/viz/descartes.py
```diff
@@ -53,9 +56,13 @@ def polygon_path(polygon):
     def coding(obj):
         # The codes will be all "LINETO" commands, except for "MOVETO"s at the
         # beginning of each subpath
-        n = len(getattr(obj, "coords", None) or obj)
+        crds = getattr(obj, "coords", None)
+        if crds is None:
+            crds = obj
+        n = len(crds)
```
yaugenst-flex, line 62: Same here. Looks like the previous implementation was fine?

tidy3d/components/viz/descartes.py
```diff
@@ -53,9 +56,13 @@ def polygon_path(polygon):
...
         vals = ones(n, dtype=Path.code_type) * Path.LINETO
-        vals[0] = Path.MOVETO
+        if len(vals) > 0:
+            vals[0] = Path.MOVETO
```
yaugenst-flex, line 65: Can you explain this change?

yaugenst-flex (commented):
Thanks @jewettaijfc this look pretty good overall! Left a couple of comments but most of the points are relatively minor. I think the performance impact is justifiable, since it only concerns plotting. Is there any worst-case scenario where the slowdown is really noticeable?

tests/test_components/test_geometry.py
jewettaijfc: Thanks for catching that!

jewettaijfc:
Thanks for all the comments!

The things you thought were weird are weird.  Those changes were added yesterday to get around all the strange little problems I encountered.

I am addressing your comments now.  I'll comment as I go through each one..

tidy3d/components/geometry/base.py
jewettaijfc: Thanks for telling me about `poetry env use python3.9`.  I tried it and I'm fighting with conda and poetry.lock file conflicts right now.  But it's a good learning experience for me.  Hold on while I figure this out...

jewettaijfc: Okay.  It looks like this is necessary.
- Although `make_valid()` has been around since `1.8a1 (2021-03-03)`, the `method` and 'keep_collapsed' arguments were not added to the `make_valid()` function until shapely version `2.1.0`.
- In a python 3.9 environment, when I use `poetry install -E dev` inside the `tidy3d` directory, it installs `shapely.__version__='2.0.7'`.  *(That raises a `TypeError`.  Even older versions of shapely/python that don't support `make_valid()` will throw an `AttributeError`.)*

**Background information:**

`buffer(0)` is deprecated and `make_valid()` is it's replacement.   The `method="structure"` and `keep_collapsed=False`arguments simplify the output of the `make_valid()` function, but I suppose we don't have to use these arguments.  Either way, I attempted to harden the code so that nothing terrible will happen, regardless whether we use `make_valid()` or `buffer(0)`.

yaugenst-flex: I see. If the arguments are not necessary then I think the cleaner approach would be to just not use them, or would that significantly complicate the code? Or I guess, maybe the other way round is worth considering too: Is `make_valid()` necessary, i.e., does it add significant value beyond just `buffer(0)`? Because if it doesn't, we could just use that instead.

tidy3d/components/geometry/base.py
jewettaijfc: I'll try to avoid them in the future.  But in this particular example, I like the type hint.  *(It clarifies that `unfiltered_geoms` is a list of general `Shapely` objects.  I remember I was a little bit confused about that detail when I first read the code for this function.  At first, I thought everything was of type `shapely.Polygon`.)*

But if you want me to remove these type hints here too, I'm happy to do that.

tidy3d/components/viz/descartes.py
jewettaijfc: The `exterior` and `coding` functions in the `decartes.py` file were crashing whenever the polygon (self.context.exterior) was empty.  *(See "Details" below.)*

I noticed this because the `cleanup_shapely_object()` function (introduced in this PR) can remove all of the vertices from a polygon, resulting in empty polygons..

This problem happens rarely.  (The unit tests did not catch it.)  But at some point it was crashing here when I ran one of the complex examples from `tidy3d-notebooks`.  That's why I updated both `exterior` and `coding` functions so they can handle empty polygons better.

-----------------------------------------
**Details:**

This happens when `self.context` is an empty `shapely.Polygon`.  Let's take a look at the original function:
```
    @property
    def exterior(self):
        return getattr(self.context, "exterior", None) or self.context[0]
```

If `self.context.exterior` is empty, it will evaluate to `False` when inside an `or` statement, even though `self.context.exterior` is not the same as `None`.  This returns `self.context[0]` which crashes since `self.context` is of type `shapely.Polygon`.  I don't think this was the intended behavior.

Incidentally, I am imitating [the code from the existing `interiors` function](https://github.com/flexcompute/tidy3d/blob/994bdc253041a9fd1a887b315ed16a38012cf3a5/tidy3d/components/viz/descartes.py#L43) here.

tidy3d/components/viz/descartes.py
jewettaijfc: It was not.  This was brittle.  See my comment above.

tests/test_components/test_geometry.py
jewettaijfc: Awesome.  Thanks for the suggestion.  I'll keep that in mind for the future.

I just realized that I can put
```
new_polygon = new_polygon.buffer(0)
```
here, and that also works fine.  *(I was afraid the test would fail if I did that, but it passes.)*

tidy3d/components/viz/descartes.py
jewettaijfc: Sure!  It's related to the other changes I made in this file.
- The new `cleanup_shapely_object()` function can remove all of the vertices from a polygon, increasing the risk that `vals` could be empty.
- So I inserted a `if len(vals) > 0:` check before the `vals[0] = Path.MOVETO`.

tidy3d/components/geometry/base.py
jewettaijfc: ***If*** we decide to use `make_valid()`, then then we should use those arguments *(`method="structure"` and `keep_collapsed=False`)* to maintain type consistency.  If my understanding is correct, these arguments ensure that `make_valid()` will only return objects of type `shapely.Polygon` or `shapely.MultiPolygon` (as does `buffer(0)`).  Otherwise [make_valid() might return `shapely.Point` or `shapely.LineString`, or other 1-D or 0-D objects](https://shapely.readthedocs.io/en/stable/reference/shapely.make_valid.html).  (After my recent code changes, *I think* that would still be okay.  But I have seen tidy3d break in unexpected places when that happens.)

If we want a cleaner, simpler solution, then we can just use `buffer(0)` inside  `cleanup_shapely_object()` function as a replacement for `make_valid()`.  Regarding that issue, this is what gemini.google.com told me:

*"It's important to note that while buffer(0) is a common "trick" for cleaning, it's not always foolproof for all types of invalid geometries. For more robust validation and repair, shapely.make_valid() is often preferred in newer shapely versions (Shapely 2.0+)."*

I have noticed that `buffer(0)` is more likely to generate polygons with duplicate vertices.  But these are visually harmless.   However `buffer(0)` *does not* work in `test_cleanup_shapely_object()`, because that unit test is very sensitive to the number of vertices in the final polygon.  So I skip that test when `make_valid()` is not available.  (Thanks for showing me how to do that, btw.)

jewettaijfc: I decided to go with `buffer(0)`.

tidy3d/components/geometry/base.py
jewettaijfc: removed.

jewettaijfc:
This PR is ready for review again.

(Ignore the unit test failures.  Our testing infrastructure has been having issues with docker all day today.)

weiliangjin2021:
> Polygon + rotation really poses lots of challenges!

Yeah!

>> I tried increasing the grid size used for geometry simplification, but it did not help that much for this example.
> From performance perspective, is `set_precision` faster? In my test, `set_precision` works with grid size 1e-10. If it's just for plotting, a grid size around single float epsilon 1e-7 should be sufficient enough.

I haven't tested the speed of `set_precision()`.  I suspect it is very fast.  But I no longer use `set_precision()` in the main code for this PR.

### Speed comparison
Fortunately, the extra code I added since your suggesttion did not slow things down that much.

The last cell of [**this jupyter file**](https://github.com/user-attachments/files/20929827/test_pr2596.ipynb.zip) contains a benchmark example.  (You must unzip it first.)
On my laptop, after my most recent changes
- [**THIS PR 2596**](https://github.com/flexcompute/tidy3d/pull/2596) now requires 0.23 seconds.
- Previously it required 0.195 seconds. *(It was using `set_precision(polygon.buffer(0), POLY_GRID_SIZE)`.  I suspect most of the computation time was due to `buffer(0)`)*
- The `develop` branch requires 0.064 seconds to draw the plot in the benchmark example
- [**PR 2581**](https://github.com/flexcompute/tidy3d/pull/2581) requires 0.098 seconds.  *(But it was not a robust solution, so I closed that PR.  Indeed, `shapely` functions are more robust.)*

*(All times reported are the best/fastest time after 10 attempts.)*

**Why I don't use set_precision()**

`set_precision()` is very helpful to get rid of overlapping vertices.  It also helps for rectangular polygons which are aligned with the x,y,z axes.  But it's not as useful for vertices on lines at arbitrary angles.  The [second example](https://github.com/flexcompute/tidy3d/pull/2596#issuecomment-2998424331) (after rotation) has vertices along the line that are *almost* collinear, after rotation.  Using `set_precision()` did not help in that case because the line was not pointing in the x, y, or z direction.

I still use `set_precision()` in the `text_cleanup_shapely_object()` function, because the square polygon from that test example is aligned with the x,y,z axes.

Thank you again, Weiliang for being part of this discussion!  In addition to `set_precision()`, you also suggested I look at `buffer()`.  This is what saved me.  I use `buffer()` in this PR.  Your comments encouraged me to take a serious look at what shapely provides.

tidy3d/components/viz/descartes.py
yaugenst-flex: ah ok, yeah that makes sense! thanks

tidy3d/components/geometry/base.py
yaugenst-flex: ok i see. i guess ideally we'd want to pin `shapely>=2.1` then but it seems they dropped support for python 3.9, which we'll keep supporting until october..
i think it's fine as-is then for now. probably not great to default to a less robust method. the only thing i'd consider is wrapping `make_valid` for `shapely<2.1` in `tidy3d/compat.py` and importing from there. makes it easier to remember to remove later once we drop support for python 3.9, and we won't have to touch `cleanup_shapely_object`. so something like:
```py
# compat.py

from importlib.metadata import version

from packaging.version import parse as parse_version
from shapely import make_valid as _make_valid_base

if parse_version(version("shapely")) >= parse_version("2.1"):
    make_valid = _make_valid_base
else:
    def make_valid(geom, *arg, **kwargs):
        return geom.buffer(0)
```

```
# base.py

from tidy3d.compat import make_valid
```

jewettaijfc: Thanks Yannick!  I'll do that now.  Hold on..

jewettaijfc: > Thanks Yannick! I'll do that now. Hold on.

This is a really cool trick by the way.
But the code breaking in weird places if I do that.
But I kind of want to pursue why it's doing that.  Because it shouldn't be.

jewettaijfc: After I implemented this change to the compat.py file, we now have a failing unit test.(`test_validate_polyslab_vertices_valid()`.  Recall that this change replaces `shapely.make_valid()`  with `buffer(0)` when python 3.9 is in use).

I don't know why it failed this time *(and not previously when i was also using buffer(0))*.

***But I'm glad it did.***

It convinced me that perhaps `buffer(0)` is not very robust.  Perhaps I should stop using it inside `cleanup_shapely_object()`.

Getting `cleanup_shapely_object()` to work with python3.9 has taken me an entire day of work.  (It's not your fault.  Both `make_valid()` and `buffer(0)` are really finicky.)  Do you want me to keep working on this?

I'm inclined to disable `cleanup_shapely_object()` when using shapely version < 2.1 (python3.9).  If our users complain about visual glitches in their plots, they can upgrade to python 3.11.

jewettaijfc:
## Removed support for pretty plots in python 3.9
(More specifically, I removed support for versions of `shapely` below 2.1.  Shapely 2.1 is incompatible with python 3.9.)

Getting this PR working with python 3.9 was taking too long.
(Thanks again, Yannick for telling me we will no longer be supporting 3.9 soon.)

jewettaijfc:
## This PR is ready for review again.

tidy3d/components/geometry/base.py
jewettaijfc: >  I'm inclined to disable cleanup_shapely_object() when using shapely version < 2.1 (python3.9).

Done.
(Thanks for letting me know we are no longer supporting python 3.9, btw.  That gave me the courage to make this decision.)

jewettaijfc:
> ## Removed support for python 3.9
> (More specifically, I removed support for versions of `shapely` below 2.1. Shapely 2.1 is incompatible with python 3.9.)
> 
> Getting this PR working with python 3.9 was taking too long. (Thanks again, Yannick for telling me we will no longer be supporting 3.9 soon.)

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+    To solve that problem, use shapely.simplify() afterwards.  (Warning: This can be slow.)
+    """
+    if SHAPELY_IS_OLD:
+        pyver = platform.python_version()
+        log.warning(
+            f"Using python {pyver} with the shapely library (prior\n"
+            " to v2.1) may cause plot errors.  Consider upgrading to\n"
+            " python 3.11 (or later) and reinstalling tidy3d.",
+            log_once=True,
+        )
```
yaugenst-flex: The message is not quite accurate, it only depends on the shapely version no?

yaugenst-flex: The message is not quite accurate, it only depends on the shapely version no?

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+        return obj  # Do not attempt to modify the object.
```
yaugenst-flex: Remove comment

tidy3d/components/geometry/base.py
```diff
@@ -63,6 +66,7 @@
 from .bound_ops import bounds_intersection, bounds_union
 
 POLY_GRID_SIZE = 1e-12
+SHAPELY_IS_OLD = parse_version(importlib.metadata.version("shapely")) < parse_version("2.1")
```
yaugenst-flex: I'd prefer moving this logic into `tidy3d/compat.py`

tidy3d/components/geometry/base.py
```diff
@@ -2862,27 +2866,47 @@ def _geometries_untraced(cls, val):
...
 
         Parameters
         ----------
         base_geometry : shapely.geometry.base.BaseGeometry
             Base geometry for inspection.
+        cleanup: bool = False
+            If True, then extremely thin or tiny features on the polygon's boundary will be removed.
+            (Vertices will be rounded to the nearest grid point with spacing ``POLY_GRID_SIZE``,
+            and any zero-area triangles that result will be deleted.)  This is useful for removing
+            artifacts from 2D plots displayed to the user.
```
yaugenst-flex: Let's make this more declarative:
```
If True, removes extremely small features from the polygon's boundary. It achieves this by rounding vertices to the nearest grid point, defined by the ``POLY_GRID_SIZE`` spacing, and subsequently deleting any resulting zero-area triangles. This process is effective for eliminating visual artifacts in 2D plots.
```

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
 
 
+def cleanup_shapely_object(obj: Shapely, resolution: float = POLY_GRID_SIZE) -> Shapely:
+    """Remove small geometric features from the boundaries of a shapely object including
+    inward and outward spikes, thin holes, and thin connections between larger regions.
+
+    Parameters
+    ----------
+    obj : shapely
+        a shapely object (typically a Polygon or a MultiPolygon)
```
yaugenst-flex: ` ``Polygon`` ` and ` ``MultiPolygon`` `

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+    resolution : float = ``POLY_GRID_SIZE``
+        Features that are smaller than ``resolution`` (relative to the object's size)
+        will be discarded.
```
yaugenst-flex: Reading this it's not immediately obvious to me what `(relative to the object's size)` means here.

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+
+    Returns
+    -------
+    Shapely
+        A new shapely object whose small features (eg. thin spikes or holes) are removed.
+
+    Notes
+    -----
+    This function does not attempt to delete overlapping, nearby, or collinear vertices.
+    To solve that problem, use shapely.simplify() afterwards.  (Warning: This can be slow.)
```
yaugenst-flex: ` ``shapely.simplify()`` ` and remove `(Warning: This can be slow.)`

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+            " to v2.1) may cause plot errors.  Consider upgrading to\n"
+            " python 3.11 (or later) and reinstalling tidy3d.",
+            log_once=True,
+        )
+        return obj  # Do not attempt to modify the object.
+    if obj.is_empty:
+        return obj
+    centroid = obj.centroid
+    object_size = max(obj.bounds[2] - obj.bounds[0], obj.bounds[3] - obj.bounds[1])
+    if object_size == 0.0:  # edge case
```
yaugenst-flex: Unnecessary comment, remove.

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+        return shapely.Polygon([])
+    # In order to prevent numerical overflow or underflow errors, we first subtract
+    # the centroid and divide by (rescale) the size of the object so it is not too big.
+    # (If we don't do this then some of the unit tests with huge polygons will fail.)
```
yaugenst-flex: Remove `()` comment

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+    normalized_obj = shapely.affinity.affine_transform(
+        # https://shapely.readthedocs.io/en/stable/manual.html#affine-transformations
+        obj,
+        matrix=[
+            1 / object_size,  # multiply x coordinates by 1 / object_size
+            0.0,
+            0.0,
+            1 / object_size,  # multiply y coordinates by 1 / object_size
+            -centroid.x / object_size,  # subtract the centroid (after dividing by object_size)
```
yaugenst-flex: These comments are duplicates of what was already explained in the comment block above. Let's keep one or the other.

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+            -centroid.y / object_size,
+        ],
+    )
+    # Important: Remove any self intersections beforehand using `shapely.make_valid()`.
+    valid_obj = shapely.make_valid(normalized_obj, method="structure", keep_collapsed=False)
+    # To get rid of small thin features, erode(shrink), dilate(expand), and erode again.
+    # (Alternatively, you could dilate, erode, and dilate again.  That works too.)
```
yaugenst-flex: Remove `()` comment, serves no purpose in the context of the code here.

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+        eroded_obj,
+        distance=2 * resolution,
+        cap_style="square",
+        quad_segs=3,
+    )
+    cleaned_obj = dilated_obj
+    # Optional: Now shrink the polygon back to the original size.
+    # Note: This is not really necessary, and we could reduce computation time by skipping it.
+    # However I decided to do it anyway.  I do it because am paranoid weird things could happen
+    # with other parts of the code which make assumptions about the polygon's size and topology.
```
yaugenst-flex: Can we condense this comment into a single line? Most of it isn't adding much information for whoever is reading the code.

tidy3d/components/geometry/base.py
```diff
@@ -3293,4 +3323,99 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+    cleaned_obj = shapely.buffer(
+        cleaned_obj,
+        distance=-resolution,
+        cap_style="square",
+        quad_segs=3,
+    )
+    # Earlier we subtracted the centroid and divided by the size of the object.
+    # Now we undo that.
```
yaugenst-flex: More concise: `# Revert to the original scale and position.`

yaugenst-flex (commented):
Thanks @jewettaijfc, this is getting very close! For the most part only the comments need some polish. Generally comments should be concise and add necessary information for someone unfamiliar with the code, and avoid adding noise (the code itself is "self-documenting" to some extent).

tests/test_components/test_geometry.py
```diff
@@ -1249,3 +1252,50 @@ def wrong_shape_height_func(x, y):
     error_message = str(excinfo.value)
     assert f"shape {expected_shape}" in error_message
     assert "shape (3, 3)" in error_message
+
+
+def test_cleanup_shapely_object():
+    if parse_version(importlib.metadata.version("shapely")) < parse_version("2.1"):
```
yaugenst-flex: This should probably reuse what we implemented for `compat`

tidy3d/compat.py
```diff
@@ -2,9 +2,32 @@
 
 from __future__ import annotations
 
+# xarray
```
yaugenst-flex: remove

tidy3d/compat.py
```diff
@@ -2,9 +2,32 @@
...
 try:
     from xarray.structure import alignment
 except ImportError:
     from xarray.core import alignment
 
+# shapely
```
yaugenst-flex: remove

tidy3d/compat.py
```diff
@@ -2,9 +2,32 @@
...
+import importlib
+
+from packaging.version import parse as parse_version
+
+from tidy3d.log import log
```
yaugenst-flex: imports to the top

tidy3d/compat.py
```diff
@@ -2,9 +2,32 @@
...
+
+_SHAPELY_VERSION = parse_version(importlib.metadata.version("shapely"))
+
+
+def _shapely_is_older_than(version: str) -> bool:
+    if _SHAPELY_VERSION < parse_version(version):
+        log.warning(
+            "Using old versions of the shapely library (prior to v2.1) may cause\n"
+            "plot errors.  This can be solved by upgrading to python 3.11\n"
```
yaugenst-flex: replace `\n` with a space. Superfluous space in `errors.  This`

Python 3.11 should be 3.10

also, reword: "Using versions versions of shapely prior to 2.1 [...]"

`tidy3d -> Tidy3D`

`python -> Python`

also, i'd actually not mention reinstalling tidy3d. i don't even know how you would do the upgrade without having to reinstall everything

yaugenst-flex (commented):
Thanks, almost there!

jewettaijfc:
(I guess our testing infrastructure is overwhelmed during normal working hours.  So I ran the unit tests manually using `poetry run pytest --tb=long  tests` using python 3.9 just now.  All unit tests are passing for both 3.11 and 3.9.)

tylerflex (approved):

PR #2599, opened 2025-06-23
feat: add Nunley variant to germanium material library

yaugenst-flex:
## Summary
- Add Nunley variant to germanium material library based on Nunley et al. 2016 data
- 20-pole fit with RMS error of 0.0132 (1.32%)
- Valid wavelength range: 0.188 - 2.480 μm

## Fit Quality
The fit achieves excellent results for the refractive index (mean error 0.04%) and good results for the extinction coefficient. The main challenge is in the very low loss region at longer wavelengths (>1.5 μm) where k values drop below 10⁻⁵.

germanium_fit_quality

## Changes
- Added `Ge_Nunley` variant to material library
- Added reference for Nunley et al. 2016 publication
- Updated CHANGELOG.md


e-g-melo (approved):
Thanks @yaugenst-flex! 

The fit seems very good. I would be happy with a bit larger RMS error if we can reduce the number of poles. However, I understand that sometimes this can be tricky to adjust. I am adding @xin-flex here, as she would use this medium for a new GUI example.

caseyflex (approved):
Looks great, thanks! One suggestion is if you want to prioritize fitting the extinction coefficient, especially in the very low loss region, you can play around with the AdvancedFastFitterParam weights, and play with adding more data points in that region, to change the fitter rms objective function

yaugenst-flex:
> The fit seems very good. I would be happy with a bit larger RMS error if we can reduce the number of poles.

Thanks @e-g-melo, I played around with the fit a bit more and was able to get a good model using 16 (instead of 20) poles.

> One suggestion is if you want to prioritize fitting the extinction coefficient, especially in the very low loss region, you can play around with the AdvancedFastFitterParam weights, and play with adding more data points in that region, to change the fitter rms objective function

Thanks for the suggestion @caseyflex, I did try some of this but in the end the fit didn't improve dramatically (went all the way up to 5x weighting of the low frequency region), resampling etc... the experimental data is actually also just much sparser in that region so (~3% of data at k<10e-4) so I think it's probably fine as-is.

Issue #2571, opened 2025-06-13
Gaussian beam with `num_freqs=3` (default) produces incorrect field at oblique incidence


tomflexcompute:
Currently `num_freqs=3` is the default for `GaussianBeam`. In oblique incidence, it actually produces a strange field. Here I ran an empty simulation with a single `GaussianBeam` source. With the default, the field looks like

<img width="860" alt="Image" src="https://github.com/user-attachments/assets/f710c84a-104e-4239-b85f-4b91ef91015e" />

Using `num_freqs=1` gives the correct beam profile

<img width="858" alt="Image" src="https://github.com/user-attachments/assets/b413f672-8dae-4eaf-99e2-6e83602d0546" />

You can use this [model](https://tidy3d.simulation.cloud/workbench?taskId=pa-42fdfb9f-d1f6-41be-b484-315f2eb59f3f&version=v9) to reproduce the observed phenomenon.

momchil-flex:
So this seems indeed quite bad and has slipped our attention. The tests we have passed after the change but I realize that e.g. a lot of our backend tests have been testing num_freqs=1 vs. num_freqs=15 as we have supported broadband GaussianBeam for a long time. We just changed the default num_freqs=3 in 2.8 as it helps a lot for broadband PlaneWave, and the thinking was that it should be helping for GaussianBeam too. But it seems that for that, it is trickier.

My understanding is that this is because the in-plane *phase* of the GaussianBeam fields at different frequencies actually differ quite strongly. Here's a plot of the absolute value (left) and the real part (right)

![Image](https://github.com/user-attachments/assets/964d9e9e-24af-4e48-bc19-67cfd465d508)

This is not the case for either PlaneWave or ModeSource, but we have not accounted for this in the GaussianBeam. This then means that the Chebyshev interpolation is bad when only a few frequencies are used - and can actually improve if we add more.

So in your case, if you look at lambda = 400nm, 550nm, 700nm, with num_freqs=1, this is what we see - importantly, the beam angle is frequency-dependent, which is ideally what we want to avoid

![Image](https://github.com/user-attachments/assets/12c39a68-423d-419e-a48c-fd861f309c1b)

With num_freqs=3, indeed things look really bad

![Image](https://github.com/user-attachments/assets/9c406be5-8282-4ebf-a435-22e1bc2f3591)

[...]

PR #2603, opened 2025-06-24
Revert to GaussianBeam.num_freqs = 1 by default

momchil-flex:
This addresses #2571 although it is not a complete fix. The changes made:

- Revert default `num_freqs=1` for guassian beam type sources
- Add a validator that checks if the frequency variation is too fast and warns the user to try decreasing the bandwidth, or increasing `num_freqs`
- Sets the hard limit of 20 for all broadband sources since using more than that tends to lead to time signatures which do not decay in the beginning and end of the simulation (even 20 seems to be problematic). We have been warning that this is upcoming for a while.

The problem in the linked issue can't really be solved by increasing `num_freqs` because it then hits the time decay issue. However, I tested that the warning is meaningful for the same setup but smaller bandwidth. Below are plots at low, middle, and high frequency.

`num_freqs = 3`, warning issued:
image

`num_freqs = 6`, no warning
image

`num_freqs = 12`, no warning
image

Visually, the result with the warning shows some deviation towards the end, while the results with 6 and 12 frequencies seem to match.

Obviously this is very heuristic. A better solution will be worked on in the future.

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -3618,6 +3618,53 @@ class Simulation(AbstractYeeGridSimulation):
                                 "A fixed angle plane wave can only be injected into a homogeneous isotropic"
                                 "dispersionless medium."
                             )
+                    # check if broadband angled gaussian beam frequency variation is too fast
+                    if (
+                        isinstance(source, (GaussianBeam, AstigmaticGaussianBeam))
+                        and np.abs(source.angle_theta) > 0
+                        and source.num_freqs > 1
+                    ):
+
+                        def radius(waist_radius, waist_distance, k0):
+                            """Gaussian beam radius at a given waist distance and k0."""
+                            z_r = waist_radius**2 * k0 / 2
+                            return waist_radius * np.sqrt(1 + (waist_distance / z_r) ** 2)
+
+                        # A slanted GaussianBeam will accumulate a phase that's frequency-dependent
+                        # like phi = K f, with the derivative dphi / df = K = 2 * pi * n * r * sin(theta) / c_0.
+                        # Here, we compute the maximum value of this coefficient computed at the waist radius
+                        # and over all frequencies. Then we compare this to the frequency spacing to
+                        # determine whether the frequency dependence is too fast, and issue a warning.
+                        optical_path_length = []
+                        freqs = source.frequency_grid
+                        for freq in freqs:
+                            n_freq, _ = src_medium.nk_model(frequency=freq)
+                            k0 = 2 * np.pi * n_freq * freq / C_0
+                            if isinstance(source, GaussianBeam):
+                                rad = radius(source.waist_radius, source.waist_distance, k0)
+                            else:
+                                rad = max(
+                                    radius(source.waist_sizes[0], source.waist_distances[0], k0),
+                                    radius(source.waist_sizes[1], source.waist_distances[1], k0),
+                                )
+                            optical_path_length.append(n_freq * rad * np.sin(source.angle_theta))
+                        # Maximum value of the path length over all freqs
+                        max_path_length = np.max(optical_path_length)
+                        # Maximum value of the phase difference
+                        max_phase_diff = max_path_length * 2 * np.pi * (freqs[-1] - freqs[0]) / C_0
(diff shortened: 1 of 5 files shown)


tomflexcompute (approved):

tidy3d/components/source/field.py
```diff
@@ -90,14 +88,14 @@ def _dir_vector(self) -> tuple[float, float, float]:
...
     num_freqs: int = pydantic.Field(
-        3,
+        1,
         title="Number of Frequency Points",
         description="Number of points used to approximate the frequency dependence of the injected "
-        "field. Default is 3, which should cover even very broadband sources. For simulations "
-        "which are not very broadband and the source is very large (e.g. metalens simulations), "
-        "decreasing the value to 1 may lead to a speed up in the preprocessing.",
+        "field. A Chebyshev interpolation is used, thus, only a small number of points is "
+        "typically sufficient to obtain converged results. Note that larger values of 'num_freqs' "
```
dbochkov-flexcompute, line 96: do we use ` or ' in this kind of docstrings?

tidy3d/components/simulation.py
```diff
@@ -3618,6 +3618,53 @@ def _source_homogeneous_isotropic(cls, val, values):
...
+                            """Gaussian beam radius at a given waist distance and k0."""
+                            z_r = waist_radius**2 * k0 / 2
+                            return waist_radius * np.sqrt(1 + (waist_distance / z_r) ** 2)
+
+                        # A slanted GaussianBeam will accumulate a phase that's frequency-dependent
+                        # like phi = K f, with the derivative dphi / df = K = 2 * pi * n * r * sin(theta) / c_0.
+                        # Here, we compute the maximum value of this coefficient computed at the waist radius
+                        # and over all frequencies. Then we compare this to the frequency spacing to
+                        # determine whether the frequency dependence is too fast, and issue a warning.
+                        dph_df = []
```
dbochkov-flexcompute: a totally optional suggestion to improve readability: 
instead of `dph_df` we can compute `optical_path_length.append(n * r * sin(theta))`, then later 
```
max_path_length = np.max(optical_path_length)
max_phase_difference = max_path_length * 2 * pi * (freqs[-1] - freqs[0]) / c_0
if not max_phase_difference / source.num_freqs < 1:
    ...
```

tidy3d/components/simulation.py
```diff
@@ -3618,6 +3618,53 @@ def _source_homogeneous_isotropic(cls, val, values):
...
+                        # Maximum value of the coefficient over all freqs
+                        max_dph_df = np.amax(dph_df)
+                        # Compare this in magnitude to the frequency spacing assuming uniform
+                        # spacing. This is heuristic since in reality we use a Chebyshev grid,
+                        # but it should be a good rule of thumb. Because the Chebyshev interpolation
+                        # is much better than simple interpolation, we don't require << 1, just < 1
+                        if not max_dph_df * (freqs[-1] - freqs[0]) / source.num_freqs < 1:
+                            log.warning(
+                                f"Broadband, angled {source.type} source has a phase dependence "
+                                "with frequency that might be too fast for the provided number of "
```
dbochkov-flexcompute: a variant: `on frequency that might be under-resolved by the provided number of sample frequencies`?

tidy3d/components/simulation.py
```diff
@@ -3618,6 +3618,53 @@ def _source_homogeneous_isotropic(cls, val, values):
...
+                                "frequencies to capture. Consider reducing the source bandwidth, "
+                                "or increasing the number of frequencies and verify the source "
```
dbochkov-flexcompute: I think it's worth specifying which parameter exactly: `... the number of frequencies ('num_freqs') ...`

dbochkov-flexcompute (approved):

tidy3d/components/source/field.py
momchil-flex: Good catch, here it should be `, while we use ' in warning and error messages.

tidy3d/components/simulation.py
momchil-flex: Sure, why not. :)

PR #2605, opened 2025-06-25
Fix EME port modes symmetry expansion

caseyflex:
This fixes two related issues:

- The port modes were not symmetry expanded. This was introduced in this PR: https://github.com/flexcompute/tidy3d/pull/2526
- The function `EMESimulationData._extract_mode_solver_data` did not retain symmetry information.

### DIFF
--- a/tidy3d/components/eme/data/sim_data.py
+++ b/tidy3d/components/eme/data/sim_data.py
@@ -39,12 +39,22 @@ class EMESimulationData(AbstractYeeGridSimulationData):
         None, title="S Matrix", description="Scattering matrix of the EME simulation."
     )
 
-    port_modes: Optional[EMEModeSolverData] = pd.Field(
+    port_modes_raw: Optional[EMEModeSolverData] = pd.Field(
         None,
         title="Port Modes",
         description="Modes associated with the two ports of the EME device. "
-        "The scattering matrix is expressed in this basis.",
+        "The scattering matrix is expressed in this basis. "
+        "Note: these modes are not symmetry expanded; use 'port_modes' instead.",
     )
+
+    @cached_property
+    def port_modes(self):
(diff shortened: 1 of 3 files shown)


lucas-flexcompute (approved):
My only comment is to maybe define `port_modes_raw` as an internal property `_port_modes_raw` if we don't want the user poking there. Unless this is handled similarly for other monitors, then just keep it consistent…

caseyflex:
> My only comment is to maybe define `port_modes_raw` as an internal property `_port_modes_raw` if we don't want the user poking there. Unless this is handled similarly for other monitors, then just keep it consistent…

Thanks for the comment. In `ModeSimulation` we also use `modes_raw` and `modes` as the symmetry expanded version. I think there is an issue with the way pydantic handles fields starting with an underscore (it doesn't treat them as actual fields).

tidy3d/components/eme/data/sim_data.py
```diff
@@ -39,13 +39,23 @@ class EMESimulationData(AbstractYeeGridSimulationData):
...
+    port_modes_raw: Optional[EMEModeSolverData] = pd.Field(
         None,
         title="Port Modes",
         description="Modes associated with the two ports of the EME device. "
-        "The scattering matrix is expressed in this basis.",
+        "The scattering matrix is expressed in this basis. "
+        "Note: these modes are not symmetry expanded; use 'port_modes' instead.",
     )
 
+    @property
```
momchil-flex: `@cached_property` would be more in line with how things work in the mode solver with `data` and `data_raw`, and also could speed up things in some use cases.

momchil-flex (approved):

tidy3d/components/eme/data/sim_data.py
caseyflex: changed. thanks!

Issue #2584, opened 2025-06-18
Minimum number of PML layers is 6


sumukhvaidya:
I'm trying to follow the tutorial 6 in the FDTD tutorials. When I set the num_layers = 2 for the PML, I get the following error. I understand that the default is 12 to eliminate all reflection. Is the lower limit 6 intended behaviour? 

ValidationError: 1 validation error for PML
num_layers
  ensure this value is greater than or equal to 6 (type=value_error.number.not_ge; limit_value=6)

To reproduce:
 set num_layers = 2 in the command 
pml = td.Boundary.pml(num_layers=2)

Thanks for your time!

momchil-flex:
Ah, indeed we introduced that validation at some point since fewer than ~6 layers of PML would pretty much always produce numerical artifacts that are unwanted in typical simulations. The tutorial is essentially exploiting this to illustrate what exactly PML does, and it was created before we added the validator.

Not sure what we should do about this - @tylerflex @weiliangjin2021 @yaugenst ?

tylerflex:
Maybe we can switch to a strongly-worded warning?

sumukhvaidya:
@ Tyler, I agree with you. 
The documentation for the td.Bounday.pml method mentions `num_layers: pydantic.v1.types.NonNegativeInt = 12`,
A warning would be better, just to keep things consistent. Getting that error straightaway caused some confusion since I did use a nonnegative integer.

momchil-flex:
@jewettaijfc could you take on this small change?

jewettaijfc:
Sure!

jewettaijfc:
[PR 2606](https://github.com/flexcompute/tidy3d/pull/2606) just landed.
I took the liberty to close this issue.

Thank you @sumukhvaidya for your well documented, excellent bug-report/feature-request!

PR #2606, opened 2025-06-26
a warning message is generated when num_layers <= 6 (instead of an error)

jewettaijfc:
## This PR addresses #2584 

`MIN_NUM_PML_LAYERS` was reduced to 1.  However a warning message is now generated if the `num_layers` argument for the following functions is less than 6
- `Boundary.pml()`
- `Boundary.stable_pml()`
- `Boundary.absorber()`

## Test instructions
```
import tidy3d as td
pml = td.Boundary.pml(num_layers=1)
```
...prints this warning:
```
23:29:21 EDT WARNING: Simulations with fewer than 6 absorbing layers are        
             likely to have unwanted numerical artifacts.  (You have only 1.)   
```
*(Let me know if I should try and fit this message on one line.)*
- No warning is generated for `td.Boundary.pml(num_layers=6)`
- An exception is raised for  `td.Boundary.pml(num_layers=0)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GaussianBeam` and `AstigmaticGaussianBeam` default `num_freqs` reset to 1 (it was set to 3 in v2.8.0) and a warning is issued for a broadband, angled beam for which `num_freqs` may not be sufficiently large.
 - Set the maximum `num_freqs` to 20 for all broadband sources (we have been warning about the introduction of this hard limit for a while).
 - Significantly improved performance of the `tidy3d.plugins.autograd.grey_dilation` morphological operation and its gradient calculation. The new implementation is orders of magnitude faster, especially for large arrays and kernel sizes.
+- Warnings are now generated (instead of errors) when instantiating `PML`, `StablePML`,
+or `Absorber` classes (or when invoking `pml()`, `stable_pml()`, or `absorber()` functions)
+with fewer layers than recommended.
 
 ### Fixed
 - Arrow lengths are now scaled consistently in the X and Y directions, and their lengths no longer exceed the height of the plot window.
--- a/tests/test_components/test_boundaries.py
+++ b/tests/test_components/test_boundaries.py
@@ -7,6 +7,9 @@ import pytest
 
 import tidy3d as td
 from tidy3d.components.boundary import (
+    MIN_NUM_ABSORBER_LAYERS,
+    MIN_NUM_PML_LAYERS,
+    MIN_NUM_STABLE_PML_LAYERS,
     PML,
     Absorber,
     BlochBoundary,
@@ -73,7 +76,7 @@ def test_boundaryedge_types(plane_wave_dir):
 
 
 def test_boundary_validators():
-    """Test the validators in class `Boundary`"""
+    """Test the validators in class ``Boundary``"""
 
     bloch = BlochBoundary(bloch_vec=1)
     pec = PECBoundary()
@@ -91,20 +94,20 @@ def test_boundary_validators():
 
 @pytest.mark.parametrize("boundary, log_level", [(PMCBoundary(), None), (Periodic(), "WARNING")])
 def test_boundary_validator_warnings(boundary, log_level):
-    """Test the validators in class `Boundary` which should show a warning but not an error"""
+    """Test the validators in class ``Boundary`` which should show a warning but not an error"""
     with AssertLogLevel(log_level):
         _ = Boundary(plus=PECBoundary(), minus=boundary)
 
 
 @pytest.mark.parametrize("boundary, log_level", [(PMCBoundary(), None), (Periodic(), "WARNING")])
 def test_boundary_validator_warnings_switched(boundary, log_level):
-    """Test the validators in class `Boundary` which should show a warning but not an error"""
+    """Test the validators in class ``Boundary`` which should show a warning but not an error"""
     with AssertLogLevel(log_level):
         _ = Boundary(minus=PECBoundary(), plus=boundary)
 
 
 def test_boundary():
-    """Test that the various classmethods and combinations for Boundary work correctly."""
+    """Test that the various classmethods and combinations for ``Boundary`` to work correctly."""
 
     # periodic
     boundary = Boundary.periodic()
@@ -144,7 +147,7 @@ def test_boundary():
 
 
 def test_boundaryspec_classmethods():
-    """Test that the classmethods for BoundarySpec work correctly."""
+    """Test that the classmethods for ``BoundarySpec`` work correctly."""
 
     # pml
     boundary_spec = BoundarySpec.pml(x=False, y=True, z=True)
@@ -188,3 +191,26 @@ def test_boundaryspec_classmethods():
     assert all(
         isinstance(boundary, PML) for boundary_dim in boundaries for boundary in boundary_dim
     )
+
+
+@pytest.mark.parametrize("absorber_type", [PML, StablePML, Absorber])
+def test_num_layers_validator(absorber_type):
+    """Test the Field validators that enforce ``num_layers>0``."""
+    with pytest.raises(pydantic.ValidationError):
+        _ = absorber_type(num_layers=0)
+
+
+@pytest.mark.parametrize(
+    "absorber_type, num_layers",
+    [
+        (PML, MIN_NUM_PML_LAYERS),
+        (StablePML, MIN_NUM_STABLE_PML_LAYERS),
+        (Absorber, MIN_NUM_ABSORBER_LAYERS),
+    ],
+)
+def test_num_layers_validator_warning(absorber_type, num_layers):
+    """Test the validators in ``PML`` which should display a warning not an error."""
+    with AssertLogLevel(None):
+        _ = absorber_type(num_layers=num_layers)
+    with AssertLogLevel("WARNING"):
+        _ = absorber_type(num_layers=num_layers - 1)
--- a/tidy3d/components/boundary.py
+++ b/tidy3d/components/boundary.py
@@ -18,6 +18,30 @@ from .source.field import TFSF, GaussianBeam, ModeSource, PlaneWave
 from .types import TYPE_TAG_STR, Axis, Complex
 
 MIN_NUM_PML_LAYERS = 6
+MIN_NUM_STABLE_PML_LAYERS = 6
+MIN_NUM_ABSORBER_LAYERS = 6
+
+
+def warn_num_layers_factory(min_num_layers: int, descr: str):
+    """Several similar classes defined have a ``num_layers`` data member, and they generate
+    similar warning messages when ``num_layers`` is too small.  This function creates a pydantic
+    validator which can be shared with all of these classes to create these warning messages."""
+
+    @pd.validator("num_layers", allow_reuse=True, always=True)
+    def _warn_num_layers(cls, val):
+        if val < min_num_layers:
+            cls_name = cls.__name__
+            log.warning(
+                f"A {descr} ({cls_name}) boundary was created with {val} layers. "
+                f"{cls_name}s with less than {min_num_layers} layers are highly subject to "
+                "unwanted numerical artifacts. We strongly recommend "
+                "increasing the number of layers. For more details refer to: "
+                f"'https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.{cls_name}.html'.",
+                log_once=True,
+            )
+        return val
+
+    return _warn_num_layers
 
 
 class BoundaryEdge(ABC, Tidy3dBaseModel):
@@ -267,7 +291,7 @@ class AbsorberSpec(BoundaryEdge):
         ...,
         title="Number of Layers",
         description="Number of layers of standard PML.",
-        ge=MIN_NUM_PML_LAYERS,
+        ge=1,
     )
     parameters: AbsorberParams = pd.Field(
         ...,
@@ -384,13 +408,17 @@ class PML(AbsorberSpec):
         12,
         title="Number of Layers",
         description="Number of layers of standard PML.",
-        ge=MIN_NUM_PML_LAYERS,
+        ge=1,
     )
 
     parameters: PMLParams = pd.Field(
         DefaultPMLParameters,
         title="PML Parameters",
         description="Parameters of the complex frequency-shifted absorption poles.",
+    )
+
+    _warn_num_layers = warn_num_layers_factory(
+        min_num_layers=MIN_NUM_PML_LAYERS, descr="perfectly-matched layer"
     )
 
 
@@ -422,13 +450,17 @@ class StablePML(AbsorberSpec):
         40,
         title="Number of Layers",
         description="Number of layers of 'stable' PML.",
-        ge=MIN_NUM_PML_LAYERS,
+        ge=1,
     )
 
     parameters: PMLParams = pd.Field(
         DefaultStablePMLParameters,
         title="Stable PML Parameters",
         description="'Stable' parameters of the complex frequency-shifted absorption poles.",
+    )
+
+    _warn_num_layers = warn_num_layers_factory(
+        min_num_layers=MIN_NUM_STABLE_PML_LAYERS, descr="stable perfectly-matched layer"
     )
 
 
@@ -475,13 +507,17 @@ class Absorber(AbsorberSpec):
         40,
         title="Number of Layers",
         description="Number of layers of absorber to add to + and - boundaries.",
-        ge=MIN_NUM_PML_LAYERS,
+        ge=1,
     )
 
     parameters: AbsorberParams = pd.Field(
         DefaultAbsorberParameters,
         title="Absorber Parameters",
         description="Adiabatic absorber parameters.",
+    )
+
+    _warn_num_layers = warn_num_layers_factory(
+        min_num_layers=MIN_NUM_ABSORBER_LAYERS, descr="absorber"
     )
 
 


tidy3d/components/boundary.py
```diff
@@ -275,6 +276,13 @@ class AbsorberSpec(BoundaryEdge):
         description="Parameters to fine tune the absorber profile and properties.",
     )
 
+    def warn_low_num_layers(num_layers: int) -> None:
```
tylerflex: please write this as a pydntic validator. you can check out some examples all over the code base, such as this

https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/apodization.py#L51-L58

and here's the reference on pedantic's API site (note we are still using pydantic v1 for now!)

https://docs.pydantic.dev/1.10/usage/validators/

tylerflex: also if we are only throwing this warning in `PML`, I think the validator could go there instead of at this `BoundaryEdge` class

tidy3d/components/boundary.py
```diff
@@ -275,6 +276,13 @@ class AbsorberSpec(BoundaryEdge):
...
+        if num_layers < MIN_NUM_PML_LAYERS_WARN:
+            log.warning(
+                f"Simulations with fewer than {MIN_NUM_PML_LAYERS_WARN} absorbing layers are\n"
```
tylerflex: We previously errored, so I think a verbose message is totally warranted here. It could be even more verbose.

A few minor comments about the message.

"Simulations with ..." : this is specifically for PML (absorber) layers, so I think we can replace this with something like

```
f"A perfectly-matched layer (PML) was initialized with {num_layers}. PMLs with under {MIN_NUM_PML_LAYERS_WARN} layers are highly subject to unwanted numerical artifacts. We strongly recommend increasing the number of layers. For more details refer to 'https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.PML.html'."
```

tidy3d/components/boundary.py
```diff
@@ -17,7 +17,8 @@
 from .source.field import TFSF, GaussianBeam, ModeSource, PlaneWave
 from .types import TYPE_TAG_STR, Axis, Complex
 
-MIN_NUM_PML_LAYERS = 6
+MIN_NUM_PML_LAYERS = 1
```
tylerflex: See the definition of the `num_layers` field in the `PML` class
https://github.com/flexcompute/tidy3d/blob/be65108f47a0dc1d9a5398b3b097eb50304e26ac/tidy3d/components/boundary.py#L391-L396

we can set the keyword argument here `ge` to determine that the value should be greater than or equal to this number, or else pydantic throws a warning. 

For this issue, I think it makes sense that we can just keep `ge=MIN_NUM_PML_LAYERS`

tylerflex (changes requested):
We will need also 
- a changelog item, perhaps under ## changed
- unit tests. 

You basically wrote the tests in your GitHub comment, just need to translate these into  unit tests. Validate that the proper behavior is observed in the 3 use cases.

https://github.com/flexcompute/tidy3d/blob/be65108f47a0dc1d9a5398b3b097eb50304e26ac/tests/test_components/test_boundaries.py#L75-L96

You can see how we ensure that both exceptions and warnings are raised in this code snippet (that's also the best file to put the unit tests).

General note: this PR touches on some concepts that you'll see all of the time in other projects (pydantic models and fields, validators, unit tests) so worth understanding the details of all of these changes.

Thanks!

tidy3d/components/boundary.py
jewettaijfc: > and here's the reference on pedantic's API site (note we are still using pydantic v1 for now!)

I'm happy to do that.  But to confirm: are we only throwing warnings for `PML`?  What about `StablePML`?

For reference, here are the 3 classes which inherit from `AbsorberSpec`
- `class PML(AbsorberSpec)`
- `class StablePML(AbsorberSpec)`
- `class Absorber(AbsorberSpec) `
You want me to generate the warning for `PML` and `StablePML`, but not `Absorber`?

-------
*(Thanks for the link to the pydantic 1.0 documentation, btw.  It helped a lot.)*

tidy3d/components/boundary.py
jewettaijfc: > ge to determine that the value should be greater than or equal to this number, or else pydantic throws a warning.

That sentence clashes with my limited understanding of pydantic Fields.  (I confess I am sometimes even a little confused by the pydantic documentation.)

**I am quite confused here.**

------------------
(Incidentally, I left this line alone: `ge=MIN_NUM_PML_LAYERS`, but `MIN_NUM_PML_LAYERS=1` now.)

Do you want me leave `MIN_NUM_PML_LAYERS=6` (instead of 1)?

If I do that, the `pd.Field()` will raise a `ValidatorException` before the `PML` class (or `AbsorberSpec` class) is even instantiated.  I could catch that exception, but it remains true that the PML class will never instantiated.  What use would a `@validator` be in that case?  (Am I misunderstanding the pydantic control flow?  Can a `@validator` override the behavior of `Field()`?)

jewettaijfc: If it helps to explain my thinking, here is (my first draft) of a validator to generate a warning message.  I use the `MIN_NUM_PML_LAYERS_WARN` here.  Please critique it, if it's not what you wanted.
```py
    @classmethod
    @pd.validator("num_layers")
    def warn_low_num_layers(cls, v) -> None:
        if v < MIN_NUM_PML_LAYERS_WARN:
            log.warning(
                f"A perfectly-matched layer (PML) was created with {v} layers.\n"
                f"PMLs with under {MIN_NUM_PML_LAYERS_WARN} layers are highly subject to\n"
                "unwanted numerical artifacts. We strongly recommend\n"
                "increasing the number of layers. For more details refer to:\n"
                "'https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.PML.html'."
            )
        # (Note: `Field(ge=MIN_NUM_PML_LAYERS)` above should throw an
        #  excpetion if num_layers < MIN_NUM_PML_LAYERS.)
        return v
```

tylerflex: looks pretty good to me! just be aware that we dont need these newlines. also make sure there's a period after each space, even at the end of the string like the first line (`"layers. "`) and after the `:` on the 2nd to last line.

tylerflex: > Do you want me leave MIN_NUM_PML_LAYERS=6 (instead of 1)?

No sorry, I was probably unclear. We can leave it as 1.

> If I do that, the pd.Field() will raise a ValidatorException before the PML class (or AbsorberSpec class) is even instantiated. I could catch that exception, but it remains true that the PML class will never instantiated. What use would a @validator be in that case? (Am I misunderstanding the pydantic control flow? Can a @validator override the behavior of Field()?)

So originally I said that pydantic would "warn" if the provided value was not `ge=1`, that was incorrect. It will error. This is the expected behavior, namely

* `num_layers < 1` -> validation error. handled by pydantic through `ge=1` automatically in the `Field()`
* `num_layers <=  6` -> `log.warning()` handled by a custom validator that only warns.
* `num_layers > 6` -> nothing happens

does that make sense? I think the logic in your validator is fine. Maybe you would want to do `if v <= MIN_NUM_PML_LAYERS_WARN` instead of `<` though.

tidy3d/components/boundary.py
tylerflex: Let's just put it in PML for now since all of these other absorbers likely have different thresholds for warnings. It will be easy to change later if we want.

tidy3d/components/boundary.py
jewettaijfc: Thanks for all of the great replies.

If we implement this as a validator, then a warning message will be printed every time a `PML` or `StablePML` class is instantiated.  If we use the `pml()` or `stable_pml()` functions, it will be printed twice.  That's because `pml()` instantiates PML twice.  Here's the definition of the `pml()` function:

```py
    @classmethod
    def pml(cls, num_layers: pd.NonNegativeInt = 12, parameters: PMLParams = DefaultPMLParameters):
        plus = PML(num_layers=num_layers, parameters=parameters)
        minus = PML(num_layers=num_layers, parameters=parameters)
        return cls(plus=plus, minus=minus)
```
This prints out:
```
>>> td.Boundary.pml(num_layers=4)
23:52:26 EDT WARNING: A perfectly-matched layer (PML) was created  
             with 4 layers. PMLs with under 6 layers are highly subject to      
             unwanted numerical artifacts. We strongly recommend increasing the 
             number of layers. For more details refer to:                       
             'https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosu
             mmary/tidy3d.PML.html'.                                            
             WARNING: A perfectly-matched layer (PML) was created  
             with 4 layers. PMLs with under 6 layers are highly subject to      
             unwanted numerical artifacts. We strongly recommend increasing the 
             number of layers. For more details refer to:                       
             'https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosu
             mmary/tidy3d.PML.html'.
```
One possible solution is to invoke `log.warning()` function with `log_once=True`.  In that case, it will only print one warning message.  If the user runs `pml(num_layers=3)` a second time in a different jupyter cell, it will not print the warning message a second time.

That's okay with me.  How do you feel about that?

tylerflex: `log_once=True` sounds good to me!

tidy3d/components/boundary.py
jewettaijfc: > "Let's just put it in PML for now since all of these other absorbers likely have different thresholds for warnings. It will be easy to change later if we want."

That would mean after this PR, `PML` and `StablePML` no longer complain at all if they are instantiated with `num_layers=1`.  *(For reference, the default value of `num_layers` is 40.)*

How about this:  I created different validators for `PML` and `StablePML` and `Absorber` with (slightly) different warning messages and distinctly named threshold variables:
```
MIN_NUM_PML_LAYERS_WARN = 7
MIN_NUM_STABLE_PML_LAYERS_WARN = 7
MIN_NUM_ABSORBER_LAYERS_WARN = 7
```
We can customize these thresholds later.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `GaussianBeam` and `AstigmaticGaussianBeam` default `num_freqs` reset to 1 (it was set to 3 in v2.8.0) and a warning is issued for a broadband, angled beam for which `num_freqs` may not be sufficiently large.
 - Set the maximum `num_freqs` to 20 for all broadband sources (we have been warning about the introduction of this hard limit for a while).
+- Warnings are now generated (instead of errors) when invoking `pml()`, `stable_pml()`,
```
tylerflex: it's rather when initializing `PML`, `StablePML` and `Absorber` classes, right? Also maybe instead of being specific about `num_layers<7` we can just say "when invoking x, y, z with fewer layers than recommended. (because we have variable layer requirements for each)

tidy3d/components/boundary.py
```diff
@@ -393,6 +395,19 @@ class PML(AbsorberSpec):
         description="Parameters of the complex frequency-shifted absorption poles.",
     )
 
+    @pd.validator("num_layers")
+    def warn_low_num_layers(cls, v) -> None:
```
tylerflex: as an exercise for working with validators, why dont you make a single function to generate each of these validators as a function of the number of layers. Then we can reuse this in each of the 3 classes.

See this function as an example of how to do this. You can add a new validator generator function in this file

https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/validators.py#L126-L136

You can use the `cls` to figure out what the link and class name are supposed to look like in the warning string.

tests/test_components/test_boundaries.py
```diff
@@ -188,3 +191,51 @@ def test_boundaryspec_classmethods():
     assert all(
         isinstance(boundary, PML) for boundary_dim in boundaries for boundary in boundary_dim
     )
+
+
+def test_pml_validator():
```
tylerflex: we can just parameterize this num_layers=0 over the 3 classes

tests/test_components/test_boundaries.py
```diff
@@ -188,3 +191,51 @@ def test_boundaryspec_classmethods():
...
+        _ = PML(num_layers=num_layers)
+
+
+def test_stable_pml_validator():
+    """Test the validators in class ``StablePML``"""
+    with pytest.raises(pydantic.ValidationError):
+        _ = StablePML(num_layers=0)
+
+
+@pytest.mark.parametrize(
```
tylerflex: you can nest parameterize decorators, eg if you want to loop through multiple dimensions. as an exercise why dont you I've it a try, looping over the 3 classes and their corresponding minimum layers values so these 3 tests become one

tylerflex (changes requested):
Thanks @jewettaijfc , I think functionality wise it's looking good. I just had some minor comments and thought it would be worth trying to generalize some of these tests and validators so you can practice with that.

tests/test_components/test_boundaries.py
jewettaijfc: That's cool!  For the record, I got it working:
```
@pytest.mark.parametrize("absorber_type")
@pytest.mark.parametrize("num_layers, log_level", [
        (MIN_NUM_PML_LAYERS, None),
        (MIN_NUM_PML_LAYERS - 1, "WARNING"),
    ],
)
```
But in the end, I went with this instead, because we might want to use different thresholds for PML, StablePML, and Absorber.:
```
@pytest.mark.parametrize(
    "absorber_type, num_layers, log_level",
    [
        (PML, MIN_NUM_PML_LAYERS, None),
        (PML, MIN_NUM_PML_LAYERS - 1, "WARNING"),
        (StablePML, MIN_NUM_STABLE_PML_LAYERS, None),
        (StablePML, MIN_NUM_STABLE_PML_LAYERS - 1, "WARNING"),
        (Absorber, MIN_NUM_ABSORBER_LAYERS, None),
        (Absorber, MIN_NUM_ABSORBER_LAYERS - 1, "WARNING"),
    ],
)
```

tidy3d/components/boundary.py
jewettaijfc: Thanks for teaching me about this.  Without you pointing this out to me, I would have feared this code if I ever stumbled upon it in a dark alley.
I'm sorry implementing this is taking so long.  (Several hours so far...)

jewettaijfc: > "You can use the cls to figure out what the link and class name are supposed to look like in the warning string."

I spent some time on doing it that way, ~~but I could not make it work~~.

After a long time, I figured out that I was not accessing `cls` correctly from within the validator factory.  This took me far longer than it should have.  But it works now.

Sigh.  A learning experience...

jewettaijfc: Done.

tylerflex: yea these functions that return functions are tough to wrap one's head about but good to grapple with :)

tylerflex: just wait until you have to do triple or quadruple nested functions :P

tests/test_components/test_boundaries.py
```diff
@@ -188,3 +191,27 @@ def test_boundaryspec_classmethods():
...
+
+
+@pytest.mark.parametrize("absorber_type", [PML, StablePML, Absorber])
+def test_num_layers_validator(absorber_type):
+    """Test the Field validators that enforce ``num_layers>0``."""
+    with pytest.raises(pydantic.ValidationError):
+        _ = absorber_type(num_layers=0)
+
+
+@pytest.mark.parametrize(
```
tylerflex, line 203: now that I look at this potentially this could be one way to go

```py
pytest.mark.parameterize(
    "absorber_type, num_layers",
    [
        (PML, MIN_NUM_PML_LAYERS),
        (StablePML, MIN_NUM_STABLE_PML_LAYERS),
        (Absorber, MIN_NUM_ABSORBER_LAYERS),
    ],
)
def test_num_layers_validator_warning(absorber_type, num_layers):
    """Test the validators in ``PML`` which should display a warning not an error."""
    with AssertLogLevel(None):
        _ = absorber_type(num_layers=num_layers)
    with AssertLogLevel("WARNING"):
        _ = absorber_type(num_layers=num_layers-1)

```

tidy3d/components/boundary.py
```diff
@@ -17,7 +17,28 @@
...
 from .types import TYPE_TAG_STR, Axis, Complex
 
-MIN_NUM_PML_LAYERS = 6
+MIN_NUM_PML_LAYERS = 7
+MIN_NUM_STABLE_PML_LAYERS = 7
+MIN_NUM_ABSORBER_LAYERS = 7
+
+
+# Helper function to create the validator
+# It now takes the min_num_layers for the specific class as an argument
```
tylerflex: please remove these floating comments and just add a docstring to the function itself describing what it does.

tidy3d/components/boundary.py
```diff
@@ -431,6 +454,10 @@ class StablePML(AbsorberSpec):
         description="'Stable' parameters of the complex frequency-shifted absorption poles.",
     )
 
+    _warn_num_layers = warn_num_layers_factory(
+        MIN_NUM_STABLE_PML_LAYERS, "stable perfectly-matched layer"
```
tylerflex: in general we will pass arguments with their keywords, such as `fn(min_num_layers= MIN_NUM_STABLE_PML_LAYERS, ...)`. it is technically not required but makes it easier to read and reason about.

tylerflex (changes requested):
Thanks @jewettaijfc , getting quite close, just a few pedantic comments from me! feel free to re-request review when you're ready.

tidy3d/components/boundary.py
jewettaijfc: mu ha ha.  gemini wrote that!

jewettaijfc:
> Thanks @jewettaijfc , getting quite close, just a few pedantic comments from me!

*pydantic*

jewettaijfc:
> feel free to re-request review when you're ready.

Requested!

tests/test_components/test_simulation.py
```diff
@@ -1151,9 +1151,9 @@ def test_sim_structure_gap(box_size, log_level):
             structures=[box],
             sources=[src],
             boundary_spec=td.BoundarySpec(
```
tylerflex: out of curiosity, why was this changed?

tylerflex (changes requested):

tests/test_components/test_simulation.py
jewettaijfc: Previously the minimum number of layers was 6.  But during one of our conversations, I got the impression that you wanted to generate a warning if `num_layers<=6` (effectively increasing the minimum to 7).  This caused a test in `test_simulation.py` to fail which was using 6 layers.

I can change the minimum number of layers back to 6 again (down from 7 currently).

tylerflex: OK I see. Must have been a miscommunication from me. Let's keep the old number.

Previously we required num_layers>=6

So I think we can warn now if < 6 and error if < 1.

jewettaijfc:
> _pydantic_

That's as witty as it gets from me.

jewettaijfc:
Ready for review again!

CHANGELOG.md
```diff
@@ -23,6 +23,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Plots of objects defined by shape intersection logic will no longer display thin line artifacts.
 - Fixed incorrect gradient computation in PyTorch plugin (`to_torch`) for functions returning multi-element arrays.
 
+### Changed
```
tylerflex: can you double check this? looks like a lot of extra items slipped in, is it expected?

This seems to add a whole new ### Changed section. Please add your change to the existing one.

tidy3d/components/boundary.py
```diff
@@ -18,6 +18,25 @@
 from .types import TYPE_TAG_STR, Axis, Complex
 
 MIN_NUM_PML_LAYERS = 6
+MIN_NUM_STABLE_PML_LAYERS = 6
+MIN_NUM_ABSORBER_LAYERS = 6
+
+
+def warn_num_layers_factory(min_num_layers: int, descr: str):
```
tylerflex, line 25: please add the docstrings to this (and the validator function returned to). Just something to help other people know what this function does.

tidy3d/components/boundary.py
```diff
@@ -18,6 +18,25 @@
...
+    @pd.validator("num_layers", allow_reuse=True, always=True)
+    def _warn_num_layers(cls, val):
+        if val < min_num_layers:
+            log.warning(
+                f"A {descr} ({cls.__name__}) boundary was created with {val} layers. "
```
tylerflex: nitpick but in the future I would just extract the name into a variable  and use it within the string
```py
class_name = cls.__name__
```

tylerflex (commented):
Just a few comments. 

It would help the reviewer out if you do a final sanity check of the PR, such as the changelog, for some easy to spot things. You can get a good summary of the diff in the  "files changed" [view](https://github.com/flexcompute/tidy3d/pull/2606/files).

tidy3d/components/boundary.py
jewettaijfc: Pydantic code readability is something that bothered me at first.   Reading this code is a headache for people haven't seen this pydantic programming pattern before.  Maybe docstrings can help.  I put a lengthy docstring for the `warn_num_layers_factory()` function.  *(After my long explanation, the `_warn_num_layers()` function was kind of self-explanatory, so I did not think it needed an additional docstring.)*

CHANGELOG.md
jewettaijfc: Oops!   Yannick asked me to cleanup the CHANGELOG.md file in this PR.  Multiple disconnected `### Changed` items were floating around in the `## [Unreleased]` section.  Yannick wanted me to consolidate them.  We weren't thinking it would take long to get this PR approved.  But it looks like in the last day somebody else also did that too.  So when I rebased this PR on those changes, we now have duplicate `### Changed` sections.  I fixed it.  Thanks for pointing this out!

jewettaijfc:
> Just a few comments.
> 
> It would help the reviewer out if you do a final sanity check of the PR, such as the changelog, for some easy to spot things. You can get a good summary of the diff in the "files changed" [view](https://github.com/flexcompute/tidy3d/pull/2606/files).

Sorry.  I usually do this about once a day.  The commits that altered the CHANGELOG.md (and conflicted with my changes) landed in the last 24 hours.  I think I should avoid making changes to any portion of the CHANGELOG.md file that isn't strictly related to my own changes unless my PR is just about to land.  Too many people are modifying that file frequently.

tylerflex (approved):
Thanks @jewettaijfc it looks good to me. We can merge once all the tests pass, assuming we dont have issues with our runners that is!

jewettaijfc:
...and in the future, I'll do it before I click the "Re-review" button.

jewettaijfc:
> Thanks @jewettaijfc it looks good to me. We can merge once all the tests pass, assuming we dont have issues with our runners that is!

Yay!!

PR #2607, opened 2025-06-26
fix(pytorch): Correct gradient for array-valued functions in wrapper

txdai:
The `to_torch` wrapper, which connects `autograd` functions to PyTorch's autograd system, failed to compute correct gradients for functions that returned multi-element arrays.

The root cause was in the `_Wrapper.backward` method:
1.  The vector-Jacobian product function (`vjp`) was called with an array of ones instead of the true upstream gradient (`grad_output`).
2.  The result was then incorrectly multiplied by `grad_output` again.

This worked by coincidence for scalar outputs, where the upstream gradient is often `1.0`, but produced incorrect gradients for array outputs.

This commit corrects the implementation by passing the NumPy-converted `grad_output` directly to the `vjp` function and removing the subsequent redundant multiplication. The wrapper now correctly supports differentiation through functions that return tensors of any shape.

### DIFF
--- a/tidy3d/plugins/pytorch/wrapper.py
+++ b/tidy3d/plugins/pytorch/wrapper.py
@@ -1,10 +1,7 @@
-from __future__ import annotations
-
 import inspect
 
 import torch
 from autograd import make_vjp
-from autograd.extend import vspace
 
 
 def to_torch(fun):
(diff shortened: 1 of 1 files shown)


yaugenst-flex:
Thanks @txdai this is great! I confirmed your fix works and also added a test. However, I can't push to this branch (since it's coming from a fork), so I am closing this one in favor of https://github.com/flexcompute/tidy3d/pull/2608

PR #2608, opened 2025-06-26
fix(pytorch): Correct gradient for array-valued functions in wrapper

yaugenst-flex:
The `to_torch` wrapper, which connects `autograd` functions to PyTorch's autograd system, failed to compute correct gradients for functions that returned multi-element arrays.

The root cause was in the `_Wrapper.backward` method:
1.  The vector-Jacobian product function (`vjp`) was called with an array of ones instead of the true upstream gradient (`grad_output`).
2.  The result was then incorrectly multiplied by `grad_output` again.

This worked by coincidence for scalar outputs, where the upstream gradient is often `1.0`, but produced incorrect gradients for array outputs.

This commit corrects the implementation by passing the NumPy-converted `grad_output` directly to the `vjp` function and removing the subsequent redundant multiplication. The wrapper now correctly supports differentiation through functions that return tensors of any shape.


yaugenst-flex:
@txdai thanks a lot for bringing up this issue and implementing a fix! I kept your original commit as-is and added a test that fails without the fix. Could you have a look if everything is in order?

@tylerflex I can't review my own PR so tagging you :smile:

tylerflex (approved):
LGTM!

PR #2610, opened 2025-06-27
Change coordinates when Charge simulations aren't in the xy plane

marc-flex:
## Greptile Summary

Added coordinate transformation functionality to rotate 2D charge simulations from z-plane to y-plane orientation with comprehensive handling of structure components and validation.

- Added `_create_charge_copy_xy()` method in `tidy3d/components/tcad/simulation/heat_charge.py` to enable z-plane to y-plane rotation of 2D charge simulations
- Added validation to prevent rotation of 1D charge simulations in `_create_charge_copy_xy()`
- Implemented coordinate transformations for semiconductor mediums, doping boxes, boundary conditions, and monitors during rotation
- Added comprehensive test case `test_charge_copy_xy` in `tests/test_components/test_heat_charge.py` to verify rotation functionality

### DIFF
--- a/tidy3d/components/tcad/simulation/heat_charge.py
+++ b/tidy3d/components/tcad/simulation/heat_charge.py
@@ -1875,6 +1875,20 @@ class HeatChargeSimulation(AbstractSimulation):
 
                         new_boundary_spec.append(boundary.updated_copy(placement=new_placement))
 
+                    # Monitors
+                    new_monitors = []
+                    for mnt in self.monitors:
+                        new_center = list(mnt.center)
+                        new_center[zero_dims[0]] = mnt.center[2]
+                        new_center[2] = mnt.center[zero_dims[0]]
+
+                        new_size = list(mnt.size)
+                        new_size[zero_dims[0]] = mnt.size[2]
+                        new_size[2] = mnt.size[zero_dims[0]]
+
+                        new_monitors.append(mnt.updated_copy(center=new_center, size=new_size))
+
+                    # simulation size
                     new_size = list(self.size)
                     new_size[zero_dims[0]] = self.size[2]
                     new_size[2] = self.size[zero_dims[0]]
@@ -1883,6 +1897,7 @@ class HeatChargeSimulation(AbstractSimulation):
                         structures=new_structures,
                         boundary_spec=new_boundary_spec,
                         size=new_size,
+                        monitors=new_monitors,
                     )
                 else:
                     return self
--- a/tidy3d/components/tcad/simulation/heat_charge.py
+++ b/tidy3d/components/tcad/simulation/heat_charge.py
@@ -1792,123 +1792,122 @@ class HeatChargeSimulation(AbstractSimulation):
 
         sim_types = self._get_simulation_types()
 
-        if TCADAnalysisTypes.CHARGE in sim_types:
-            zero_dims = self.zero_dims
-            if len(zero_dims) == 1:
-                if zero_dims[0] != 2:
-                    # check doping boxes
-                    # the following dictionary associates a structure with a modified medium
-                    struct_medium = {}
-                    for structure in self.structures:
-                        if isinstance(structure.medium.charge, SemiconductorMedium):
-                            sc_medium = copy.deepcopy(structure.medium.charge)
-
-                            dopants = {"N_a": sc_medium.N_a, "N_d": sc_medium.N_d}
-                            for key, dopant in dopants.items():
-                                if isinstance(dopant, SpatialDataArray):
-                                    new_dopant = (
-                                        dopant.rename({"z": "z_tmp"})
-                                        .rename({["x", "y"][zero_dims[0]]: "z"})
-                                        .rename({"z_tmp": ["x", "y"][zero_dims[0]]})
-                                    )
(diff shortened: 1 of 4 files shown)


tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1778,3 +1786,129 @@ def _useHeatSourceFromConductionSim(self):
...
+
+    def _create_charge_copy_xy(self) -> HeatChargeSimulation:
+        """If a Charge simulation is not in the x-y plane we need to rotate it to that plane."""
+
+        sim_types = self._get_simulation_types()
+
+        if TCADAnalysisTypes.CHARGE in sim_types:
+            zero_dims = self.zero_dims
+            if len(zero_dims) == 1:
+                if zero_dims[0] != 2:
```
dbochkov-flexcompute: for better readability:
```
if TCADAnalysisTypes.CHARGE not in sim_types:
    return self

zero_dims = self.zero_dims

if len(zero_dims) > 1:
    raise SetupError("Charge simulation can only be 2- or 3-D")

if len(zero_dims) == 0 or zero_dims[0] == 2:
    return self
```

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1778,3 +1786,129 @@ def _useHeatSourceFromConductionSim(self):
...
+                                                )
+                                            )
+                                        else:
+                                            new_boxes.append(doping_box)
+                                        sc_medium = sc_medium.updated_copy(**{key: new_boxes})
+                            struct_medium[structure] = structure.medium.updated_copy(
+                                charge=sc_medium
+                            )
+                        else:
+                            struct_medium[structure] = structure.medium
```
dbochkov-flexcompute: let's also invert here:
```
if not isinstance(structure.medium.charge, SemiconductorMedium):
    struct_medium[structure] = structure.medium
else:
    ...
```
otherwise hard to remember what this `else` was for

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1778,3 +1786,129 @@ def _useHeatSourceFromConductionSim(self):
...
+                        new_size[zero_dims[0]] = structure.geometry.size[2]
+                        new_size[2] = structure.geometry.size[zero_dims[0]]
+                        new_structures.append(
+                            structure.updated_copy(
+                                geometry=structure.geometry.updated_copy(
+                                    center=new_center, size=new_size
+                                ),
+                                medium=struct_medium[structure],
+                            )
+                        )
```
dbochkov-flexcompute: does this work for non-Box structures?

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1778,3 +1786,129 @@ def _useHeatSourceFromConductionSim(self):
...
+                    for mnt in self.monitors:
+                        new_center = list(mnt.center)
+                        new_center[zero_dims[0]] = mnt.center[2]
+                        new_center[2] = mnt.center[zero_dims[0]]
+
+                        new_size = list(mnt.size)
+                        new_size[zero_dims[0]] = mnt.size[2]
+                        new_size[2] = mnt.size[zero_dims[0]]
+
+                        new_monitors.append(mnt.updated_copy(center=new_center, size=new_size))
```
dbochkov-flexcompute: looks like this kind of operation is performed 3-4 times, I think it's worth putting it in a function something like:
```
doping_box_tmp = rotated_copy(doping_box)
new_boxes.append(doping_box_tmp.updated_copy(source=source))
...
new_structures.append(
    structure.updated_copy(
        geometry=rotated_copy(structure.geometry),
        medium=struct_medium[structure],
    )
)
...
new_monitors.append(rotated_copy(mnt))
...
```

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1778,3 +1786,129 @@ def _useHeatSourceFromConductionSim(self):
...
+
+                    # simulation size
+                    new_size = list(self.size)
+                    new_size[zero_dims[0]] = self.size[2]
+                    new_size[2] = self.size[zero_dims[0]]
```
dbochkov-flexcompute: do we need to switch simulation center too?

dbochkov-flexcompute (commented):
it seems like we don't really need to put this in the frontend? Just thinking that if we need to fix anything later, we would need make changes to frontend instead of just backend

marc-flex:
> it seems like we don't really need to put this in the frontend? Just thinking that if we need to fix anything later, we would need make changes to frontend instead of just backend

I think you're right. I'll do all modifications in this PR and then I'll think how to move this to backend

tidy3d/components/tcad/simulation/heat_charge.py
marc-flex: hmm not sure. Don't have experience with other type of structures so not sure. 

I guess I'd propose that we limit the type of structures to `Box` until we can generalize the backend?

marc-flex:
Closing this PR since coordinate transformation is dealt with entirely in the back end

PR #2611, opened 2025-06-27
Validating that simulation md5 stored in mesher task matches solver task

momchil-flex:


tidy3d/web/core/task_core.py
```diff
@@ -671,3 +672,33 @@ def abort(self):
...
+                        "was run must be provided."
+                    )
+                try:
+                    # get mesh task info
+                    mesh_task = SimulationTask.get(parent_tasks[0], verbose=False)
+                    assert mesh_task.task_type == "VOLUME_MESH"
+                    assert mesh_task.status == "success"
+                    # get up-to-date task info
+                    task = SimulationTask.get(self.task_id, verbose=False)
+                    if task.fileMd5 != mesh_task.childFileMd5:
```
marc-flex, line 696: I'm assuming the child corresponds to the simulation being passed on to the Volume mesher, correct?

tidy3d/web/core/task_core.py
```diff
@@ -671,3 +672,33 @@ def abort(self):
...
+                            "current simulation."
+                        )
+                except Exception as e:
+                    raise ValidationError(
+                        "The parent task must be a 'VolumeMesher' task which has been successfully "
+                        "run and is associated to the same 'HeatChargeSimulation' as provided here."
+                    ) from e
+
+            except Exception as e:
+                raise WebError(f"Provided 'parent_tasks' failed validation: {e!s}") from e
```
marc-flex, line 708: Is there an actual validation or is this a somewhat generic message?

marc-flex (approved):

tidy3d/web/core/task_core.py
momchil-flex: right

tidy3d/web/core/task_core.py
momchil-flex: So this error will append the error message raised in the preceeding `try` block to the error message shown here. So e.g. if the simulations are found to not match you'll get something like `"ERROR: Provided 'parent_tasks' failed validation: Simulation stored in parent task 'VolumeMesher' does not match the current simulation."`

dbochkov-flexcompute (approved):
That was fast! Looks great to me

PR #2618, opened 2025-07-02
Enhance Custom Validator Error Messages with Object Names

### LINKED ISSUE #2614: Enhance Custom Validator Error Messages with Object Names
**Problem Description**

Currently, error messages from our custom validators, such as those that inspect simulation components, identify objects by their system path (e.g., `simulation.monitor[0]`). This path is not intuitive for users, who then have to manually locate the object in their simulation setup.

For example, a monitor validation error might look like this:
`monitors: 'simulation.monitors[1]' is outside of the simulation domain.`

This forces the user to figure out which monitor corresponds to index `[1]`, which especially in GUI can be cumbersome.

**Proposed Solution**

We can improve these error messages. For objects that have a `name` attribute (like monitors), the error message should be updated to include the object's name.

The improved error message could look like this:

`Monitor 'field_monitor_1' is outside of the simulation domain.`

or include the path for reference:

`Monitor 'field_monitor_1' (simulation.monitors[1]) is outside of the simulation domain.`

**Scope**

This change applies to custom validators written in the python client for objects that have a user-definable `name`. The monitor validation is a primary example.

@e-g-melo FYI

jewettaijfc:
This PR is an attempt to address issue #2614 .

This PR changes the warnings error messages printed to the user for `Monitor`, `Source`, and `Structure`.  If the user specified a "name" argument when instantiating these classes, the error message will refer to that name instead of the internal expression that refers to it (which means nothing to the user).

## Example input:
```py
import tidy3d as td

# structure = td.Structure(
#     geometry=td.Box(center=(1.0, 1.0, 0.0), size=(0.5, 0.5, 0.5)), medium=td.Medium(permittivity=2.0), name="structure_1"
# ) 
#
# monitor = td.FieldMonitor(
#     center=(-1.0, 0, 0), size=(0.5, 0, 1), freqs=[100e14], name="field_monitor_1"
# )

source = td.UniformCurrentSource(
    center=(0, -1.0, 0),
    size=(1, 0, 0.5),
    polarization="Ex",
    source_time=td.GaussianPulse(
        freq0=100e14,
        fwidth=10e14,
    ),
    name="source_1",
)

# make simulation
sim = td.Simulation(
    size=(1, 1, 1),
    grid_spec=td.GridSpec.auto(wavelength=4),
        x=td.Boundary.pml(num_layers=10),
        y=td.Boundary.periodic(),
        z=td.Boundary.pml(num_layers=10),
    ),
    sources=[source],
    # monitors=[monitor],
    # structures=[structure],
    run_time=1e-12,
)
```
*(test file: [test_pr2618.ipynb.zip](https://github.com/user-attachments/files/21042410/test_pr2618.ipynb.zip)*

## Before this PR (example error message)
This generates:
```
simulation.sources[0] is outside of the simulation domain.
```
*Similarly cryptic errors are displayed for Structures and Monitors if you uncomment the `# structures=[structure].` and `# monitors=[monitor],` lines in the code above.*

## After this PR (example error message)
```
'source_1' (simulation.sources[0]) is outside of the simulation domain.
```
- If the `name` argument is omitted from the argument list, then a (vague) error messages is generated.  For example:
"simulation.sources[0] (no 'name' was specified) is outside of the simulation domain."
- This does not do any good unless users provide names for the various objects in their simulations.  ~~To encourage that, a warning messages is generated when a Source is created without a name argument.  For example: WARNING: UniformCurrentSource instantiated without a name argument.~~

## Greptile Summary

Enhances error message clarity by incorporating user-specified names for Monitor, Source, and Structure objects in validation messages.

- Added new `named_obj_descr` helper function in `components/validators.py` to generate more descriptive error strings
- Updated error messages to show both user-specified names and internal references (e.g., `'source_1' (simulation.sources[0])`)
- Modified validation messages in `components/simulation.py` to use component names when available
- Falls back to descriptive placeholder when no name specified (e.g., `'(no name was specified)'`)
- Renamed validation function to more descriptive `assert_objects_completely_in_sim_bounds`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Warnings are now generated (instead of errors) when instantiating `PML`, `StablePML`,
 or `Absorber` classes (or when invoking `pml()`, `stable_pml()`, or `absorber()` functions)
 with fewer layers than recommended.
+- Warnings and error messages originating from `Structure`, `Source`, or `Monitor` classes now refer to problematic objects by their user-supplied `name` attribute, alongside their index.
 
 ### Fixed
 - Arrow lengths are now scaled consistently in the X and Y directions, and their lengths no longer exceed the height of the plot window.
--- a/tests/test_components/test_simulation.py
+++ b/tests/test_components/test_simulation.py
@@ -3648,3 +3648,45 @@ def test_create_sim_multiphysics_with_incompatibilities():
                 ),
             ],
         )
+
+
+def test_messages_contain_object_names():
+    """Make sure that errors and warnings contain the name of the object."""
+    # Note: This function currently tests for out-of-bounds errors and warnings.
+    # Create an empty simulation.
+    sim = td.Simulation(
+        size=(1, 1, 1),
+        grid_spec=td.GridSpec.auto(wavelength=4),
+        run_time=1e-12,
+    )
+
+    # Test 1) Create a structure lying outside the simulation boundary.
+    # Check that a warning message is generated containing the structure's `name`.
+    name = "structure_123"
+    structure = td.Structure(
+        name=name,
+        geometry=td.Box(center=(1.0, 0.0, 0.0), size=(0.5, 0.5, 0.5)),
+        medium=td.Medium(permittivity=2.0),
+    )
+    with AssertLogLevel("WARNING", contains_str=name):
+        _ = sim.updated_copy(structures=[structure])
+
+    # Test 2) Create a source lying outside the simulation boundary.
+    # Check that an error message is generated containing the source's `name`.
+    name = "source_123"
+    source = td.UniformCurrentSource(
+        name=name,
+        center=(0, -1.0, 0),
+        size=(1, 0, 0.5),
+        polarization="Ex",
+        source_time=td.GaussianPulse(freq0=100e14, fwidth=10e14),
+    )
+    with pytest.raises(pydantic.ValidationError, match=name) as e:
+        _ = sim.updated_copy(sources=[source])
+
+    # Test 3) Create a monitor lying outside the simulation boundary.
+    # Check that an error message is generated containing the monitor's `name`.
+    name = "monitor_123"
+    monitor = td.FieldMonitor(name=name, center=(-1.0, 0, 0), size=(0.5, 0, 1), freqs=[100e14])
+    with pytest.raises(pydantic.ValidationError, match=name) as e:
+        _ = sim.updated_copy(monitors=[monitor])
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -117,6 +117,7 @@ from .types import (
 from .validators import (
     assert_objects_contained_in_sim_bounds,
     assert_objects_in_sim_bounds,
+    named_obj_descr,
     validate_mode_objects_symmetry,
 )
 from .viz import (
@@ -2961,8 +2962,9 @@ class Simulation(AbstractYeeGridSimulation):
                 for geom in flatten_groups(structure.geometry):
                     zero_dims = geom.zero_dims
                     if len(zero_dims) > 0:
+                        obj_descr = named_obj_descr(structure, "structures", i)
                         consolidated_logger.warning(
-                            f"Structure at 'structures[{i}]' has geometry with zero size along "
+                            f"Structure: {obj_descr} has geometry with zero size along "
                             f"dimensions {zero_dims}, and with a medium that is not a 'Medium2D'. "
                             "This is probably not correct, since the resulting simulation will "
                             "depend on the details of the numerical grid. Consider either "
@@ -3018,10 +3020,11 @@ class Simulation(AbstractYeeGridSimulation):
 
         with log as consolidated_logger:
 
-            def warn(istruct, side):
+            def warn(structure, istruct, side):
                 """Warning message for a structure too close to PML."""
+                obj_descr = named_obj_descr(structure, "structures", istruct)
                 consolidated_logger.warning(
-                    f"Structure at structures[{istruct}] was detected as being less "
+                    f"Structure: {obj_descr} was detected as being less "
                     f"than half of a central wavelength from a PML on side {side}. "
                     "To avoid inaccurate results or divergence, please increase gap between "
                     "any structures and PML or fully extend structure through the pml.",
@@ -3044,7 +3047,7 @@ class Simulation(AbstractYeeGridSimulation):
                             and struct_val > sim_val
                             and abs(sim_val - struct_val) < lambda0 / 2
                         ):
-                            warn(istruct, axis + "-min")
+                            warn(structure, istruct, axis + "-min")
 
                     zipped = zip(["x", "y", "z"], sim_bound_max, struct_bound_max, boundaries)
                     for axis, sim_val, struct_val, boundary in zipped:
@@ -3056,7 +3059,7 @@ class Simulation(AbstractYeeGridSimulation):
                             and struct_val < sim_val
                             and abs(sim_val - struct_val) < lambda0 / 2
                         ):
-                            warn(istruct, axis + "-max")
+                            warn(structure, istruct, axis + "-max")
 
         return val
 
@@ -3095,9 +3098,8 @@ class Simulation(AbstractYeeGridSimulation):
                             medium_str = "The simulation background medium"
                             custom_loc = ["medium", "frequency_range"]
                         else:
-                            medium_str = (
-                                f"The medium associated with structures[{medium_index - 1}]"
-                            )
+                            medium_descr = named_obj_descr(medium, "mediums", medium_index)
+                            medium_str = f"The medium associated with {medium_descr}"
                             custom_loc = [
                                 "structures",
                                 medium_index - 1,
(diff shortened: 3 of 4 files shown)


jewettaijfc:
Quick Update: At the moment, this PR appears to be working.  But this PR is failing some unit tests with confusing error messages.  This is a very short PR, so it's easy to start over, running unit tests after I make each modification.  That way, I can quickly see what line in my code is making pytest/pydantic unhappy.

jewettaijfc:
Done.  That narrowed down which pieces of code were triggering the failing unit tests.

But I don't yet understand some of the unit tests that are failing.

jewettaijfc:
Most of my time today was spent trying to understand and update our unit tests file so that this PR could pass.  The worst file is `tests/test_package/test_log.py`.  Some of the tests there are cryptic.  Excerpt:
```
assert len(warning_list) == 29
```
It looks like something (perhaps our logger?) is capturing stdin and stdout.  So it's hard to use print() or pdb inside these pytest functions.  And we've setup pytest to run in parallel.   (I've been disabling that using `pytest -n0`.)  There's some additional complexity due to pydantic.

I could use some help.  (How to debug pytest functions?)

jewettaijfc:
I found a way to get these unit tests passing.  I have a different branch where they are working.  (But they don't solve the problem that the issue requested.)  I will copy the files over to this branch.  At least that way I will have something that works as a starting point.

jewettaijfc:
Currently, the tests are passing.
But I still need to update these functions from tidy3d/components/simulation.py:
- `_structures_not_close_pml()`
- `_warn_monitor_mediums_frequency_range()`

jewettaijfc:
In the process of doing that, I reintroduced changes I made to `validate_name_str()` that should not have been made in the first place.
I promise I will remove those changes as soon as I've resolved the other errors and addressed the original issue.

jewettaijfc:
Passing unit tests.
Ready for review.

tidy3d/components/simulation.py
```diff
@@ -115,8 +115,9 @@
     annotate_type,
 )
 from .validators import (
-    assert_objects_contained_in_sim_bounds,
+    assert_objects_completely_in_sim_bounds,
```
yaugenst-flex: What's the reason for renaming this validator?

yaugenst-flex (commented):
Thanks @jewettaijfc I think the implementation looks good! Still needs some tests though.

yaugenst-flex:
@jewettaijfc you should be able to catch and match a logged warning using `AssertLogLevel` like for example [here](https://github.com/flexcompute/tidy3d/blob/96d64e37d6da5871ba5c0bb478484020b473b474/tests/test_package/test_log.py#L281-L282). Basically just add a test where you set up a simulation with some named object that's not in bounds, and use `AssertLogLevel` to check whether the warning is being raised. Probably this would also go into `test_simulation.py` rather than `test_log.py`.

jewettaijfc:
> @jewettaijfc you should be able to catch and match a logged warning using `AssertLogLevel` like for example [here](https://github.com/flexcompute/tidy3d/blob/96d64e37d6da5871ba5c0bb478484020b473b474/tests/test_package/test_log.py#L281-L282). Basically just add a test where you set up a simulation with some named object that's not in bounds, and use `AssertLogLevel` to check whether the warning is being raised.

Cool!  Do you want me to parse the content of the warning messages and error messages (to make sure the message contains the object name)?

> Probably this would also rather go into `test_simulation.py` rather than `test_log.py`.

Thanks!  (I was wondering this too.)

jewettaijfc:
> you should be able to catch and match a logged warning using AssertLogLevel like for example [here](https://github.com/flexcompute/tidy3d/blob/96d64e37d6da5871ba5c0bb478484020b473b474/tests/test_package/test_log.py#L281-L282). Basically just add a test where you set up a simulation with some named object that's not in bounds,

This PR changes 9 different errors and warnings, including the boundary violations you mentioned.  I confess I have not figured out how to run Tidy3D in a way that would trigger the other 8 failures.  I am hoping that the people who wrote that code already wrote unit tests that make sure the appropriate errors and warnings were generated.  But I can write a separate test which take one or two of these failures and verifies that the error message contains the object `name`.

jewettaijfc:
I guess that wasn't so bad!  I added a new test (`test_simulation.py::test_messages_contain_object_names`).  Yay!  I think I was partially driven insane by by my earlier experiences.  At the moment, all the tests are passing.

jewettaijfc:
Sorry.  I misunderstood what you were asking for.  (You were clear.  This is my fault.)

I will add `AssertLogLevel` to my unit test.  Hold on.

jewettaijfc:
Done.

jewettaijfc:
Now using `pytest.raises()`.
This PR is ready for review.

tests/test_components/test_simulation.py
```diff
@@ -3648,3 +3648,62 @@ def test_create_sim_multiphysics_with_incompatibilities():
...
+    # Note: This function currently tests for out-of-bounds errors and warnings.
+    sim_args = {
+        "size": (1, 1, 1),
+        "center": (0, 0, 0),
+        "grid_spec": td.GridSpec.auto(wavelength=4),
+        "boundary_spec": td.BoundarySpec(
+            x=td.Boundary.periodic(),
+            y=td.Boundary.periodic(),
+            z=td.Boundary.periodic(),
+        ),
```
yaugenst-flex: This is fine just note that `center` and `boundary_spec` are not required arguments, so you could simplify this a bit.
Note you could also just instantiate the simulation directly here and then do `sim.updated_copy(structures=new_structures)` later down in the tests. It's a bit more in line with how the rest of the tests are written.

tests/test_components/test_simulation.py
```diff
@@ -3648,3 +3648,62 @@ def test_create_sim_multiphysics_with_incompatibilities():
...
+        "run_time": 1e-12,
+    }
+
+    # Test 1) Create a structure lying outside the simulation boundary.
+    # Check that a warning message is generated containing the structure's `name`.
+    structure_args = {
+        "geometry": td.Box(center=(1.0, 0.0, 0.0), size=(0.5, 0.5, 0.5)),
+        "medium": td.Medium(permittivity=2.0),
+    }
```
yaugenst-flex: Doesn't seem necessary to create the `dict` indirection here? I think something like:
```py
name = "structure_123"
structure = td.Structure(..., name=name)
with AssertLogLevel("WARNING", containts_str=name):
  _ = td.Simulation(...)
```
would be easier to understand.

tests/test_components/test_simulation.py
```diff
@@ -3648,3 +3648,62 @@ def test_create_sim_multiphysics_with_incompatibilities():
...
+    with AssertLogLevel("WARNING", contains_str="structure_123"):
+        # Create a simulation with this structure.
+        _ = td.Simulation(
+            structures=[td.Structure(name="structure_123", **structure_args)],
+            **sim_args,
+        )
+
+    # Test 2) Create a source lying outside the simulation boundary.
+    # Check that an error message is generated containing the source's `name`.
```
yaugenst-flex, line 3675: Same as with the structures above.

tests/test_components/test_simulation.py
```diff
@@ -3648,3 +3648,62 @@ def test_create_sim_multiphysics_with_incompatibilities():
...
+            fwidth=10e14,
+        ),
+    }
+    with pytest.raises(pydantic.ValidationError) as e:
+        # Create a simulation with this source.
+        _ = td.Simulation(
+            sources=[td.UniformCurrentSource(name="source_123", **source_args)],
+            **sim_args,
+        )
+        assert "source_123" in str(e)
```
yaugenst-flex: If the test errors correctly then this assertion will never be reached, so should be removed.

tests/test_components/test_simulation.py
```diff
@@ -3648,3 +3648,62 @@ def test_create_sim_multiphysics_with_incompatibilities():
...
+    # Test 3) Create a monitor lying outside the simulation boundary.
+    # Check that an error message is generated containing the monitor's `name`.
+    monitor_args = {"center": (-1.0, 0, 0), "size": (0.5, 0, 1), "freqs": [100e14]}
+    with pytest.raises(pydantic.ValidationError) as e:
+        # Create a simulation with this monitor.
+        _ = td.Simulation(
+            monitors=[td.FieldMonitor(name="monitor_123", **monitor_args)],
+            **sim_args,
+        )
+        assert "monitor_123" in str(e)
```
yaugenst-flex: Same as above.

yaugenst-flex (commented):
Nice! Some minor test cleanup and this should be good to go!

tests/test_components/test_simulation.py
```diff
@@ -3648,3 +3648,48 @@ def test_create_sim_multiphysics_with_incompatibilities():
...
+    with pytest.raises(pydantic.ValidationError, match=name) as e:
+        # Create a simulation with this source.
+        _ = sim.updated_copy(sources=[source])
+
+    # Test 3) Create a monitor lying outside the simulation boundary.
+    # Check that an error message is generated containing the monitor's `name`.
+    name = "monitor_123"
+    monitor = td.FieldMonitor(name=name, center=(-1.0, 0, 0), size=(0.5, 0, 1), freqs=[100e14])
+    with pytest.raises(pydantic.ValidationError, match=name) as e:
+        # Create a simulation with this monitor.
```
yaugenst-flex: I'd say all of these "Create" comments are redundant (since that's just what the line below _does_)

yaugenst-flex (commented):
Great! One last nitpick, then we need a changelog entry (under "Changed"), squash & rebase and this is good to go!

jewettaijfc:
> we need a changelog entry

I can't believe I forgot to update the changelog (again).  Sorry about that!

All ready now.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Warnings are now generated (instead of errors) when instantiating `PML`, `StablePML`,
 or `Absorber` classes (or when invoking `pml()`, `stable_pml()`, or `absorber()` functions)
 with fewer layers than recommended.
+- Many warnings and error messages now refer to problematic objects by their (user-supplied) `name`.  This affects objects which are instances of `Structure`, `Source`, or `Monitor` classes (or their derived classes).
```
yaugenst-flex: Maybe: "Warnings and error messages originating from `Structure`, `Source`, or `Monitor` classes now refer to problematic objects by their user-supplied `name` attribute, alongside their index."

yaugenst-flex (approved):

yaugenst-flex (approved):
LGTM! Thanks @jewettaijfc

PR #2623, opened 2025-07-02
Feat: Add `PointDipole.from_angles()`

### LINKED ISSUE #2617: Dipole source, arbitrary polarization
In Lumerical it is possible to specify an arbitrary orientation for a dipole source by setting two angles. In Tidy3D there is only the option to set the polarization along the cartesian axes. Hence, when trying to simulate for arbitrary angles we need to rotate our structure. Having the possibility to rotate the source would make our life easier.
 
Allowing arbitrary angles in the `polarization` argument. 

**Describe alternatives you've considered**
Rotating the structure instead. Works but it's tedious.

saeedece:
Resolves #2617 

Class method that returns a list of `PointDipole` objects used to emulate a single dipole polarized in an arbitrary direction. The direction is specified using a polar and azimuthal angle.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add support for `np.unwrap` in `tidy3d.plugins.autograd`.
 - Add Nunley variant to germanium material library based on Nunley et al. 2016 data.
-- Add `PointDipole.from_angles()` that constructs a list of `PointDipole` objects needed to emulate a dipole oriented at a user-provided set of theta and azimuth angles.
+- Add `PointDipole.from_angles()` that constructs a list of `PointDipole` objects needed to emulate a dipole oriented at a user-provided set of polar and azimuthal angles.
 
 ### Changed
 - Switched to an analytical gradient calculation for spatially-varying pole-residue models (`CustomPoleResidue`).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add support for `np.unwrap` in `tidy3d.plugins.autograd`.
 - Add Nunley variant to germanium material library based on Nunley et al. 2016 data.
-- Add `PointDipole.from_angles()` that constructs a list of `PointDipole` objects needed to emulate a dipole oriented at a user-provided set of polar and azimuthal angles.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add support for `np.unwrap` in `tidy3d.plugins.autograd`.
 - Add Nunley variant to germanium material library based on Nunley et al. 2016 data.
+- Add `PointDipole.from_angles()` that constructs a list of `PointDipole` objects needed to emulate a dipole oriented at a user-provided set of theta and azimuth angles.
```
momchil-flex: I guess let's call it "polar" not "theta".

tidy3d/components/source/current.py
```diff
@@ -107,6 +109,40 @@ class PointDipole(CurrentSource, ReverseInterpolatedSource):
         units=MICROMETER,
     )
 
+    @classmethod
+    def from_angles(
+        cls,
+        source_time: SourceTimeType,
+        angle_theta: float,
+        angle_phi: float,
+        electrical_component: bool = True,
```
momchil-flex: I think I like `component: Literal["electric", "magnetic"] = "electric"` better as it is more explicit. But note that you need to raise an error if it happens to be neither because `Literal` here would just be a type hint but not a type check.

tidy3d/components/source/current.py
```diff
@@ -107,6 +109,40 @@ class PointDipole(CurrentSource, ReverseInterpolatedSource):
...
+        **kwargs,
+    ) -> list[PointDipole]:
+        """Returns a list of `PointDipole` objects used to emulate a single dipole polarized in an arbitrary direction. The direction is specificed using a polar and azimuthal angle."""
+        dipoles: list[PointDipole] = []
```
momchil-flex, line 144: Need to add description to the arguments, you can look at other methods in our documentation to see what the format is.

momchil-flex (commented):
Thanks for doing this! I have just a few minor comments, and then also you need to add a test before we can accept it.

tidy3d/components/source/current.py
```diff
@@ -107,6 +109,62 @@ class PointDipole(CurrentSource, ReverseInterpolatedSource):
         units=MICROMETER,
     )
 
+    @classmethod
+    def from_angles(
```
momchil-flex: We discussed this internally a bit and `from_angles` sounds like a method that would create a single `PointDipole` rather than a list of them. This is pretty nit-picky, but let's change the method name to `sources_from_angles` (here and in the tests correspondingly).

momchil-flex (approved):

PR #2624, opened 2025-07-03
Adding 'precision' option to Simulation

momchil-flex:
## Greptile Summary

Added new 'precision' field to Simulation class enabling control over floating-point precision with 'hybrid' (default) or 'double' options.

- Added field `precision` in `tidy3d/components/simulation.py` to control numerical precision in computations
- 'hybrid' mode balances speed and accuracy for typical use cases
- 'double' precision mode provides higher accuracy at 2x FlexCredit cost
- Particularly valuable for large-scale or long-running simulations requiring high numerical precision


daquinteroflex (approved):
Yep, we'll continue discussing on the backend PR how to get this to run and test correctly

PR #2626, opened 2025-07-03
Warning about symmetry expansion in ModeSolver and ModeSimulation

momchil-flex:
I wanted to add one more improvement related to the plotting, but I need to go now and I wanted to open this up for @dbochkov 's consideration especially because I ended up removing the 'symmetry' argument from 'subsection', because it wasn't really used anywhere (including backend), except in one test. So now the `symmetry` is always inherited from the original simulation - do you have any concerns with that?

Otherwise, the point of this PR is to address [this ticket](https://flow360.atlassian.net/browse/SCEM-10245?focusedCommentId=68509), and then I didn't want the warning to appear both when creating the mode simulation and when running it, so I added an option to supress it.

## Greptile Summary

Improves symmetry handling in mode simulations by adding warnings and simplifying the API. This PR focuses on notifying users when a mode plane crosses a symmetry plane without proper center alignment.

- Added validator `_warn_plane_crosses_symmetry` in `tidy3d/components/mode/mode_solver.py` to warn users about automatic simulation region expansion
- Removed unused 'symmetry' argument from `subsection()` method in `tidy3d/components/eme/simulation.py`, making symmetry always inherit from parent simulation
- Added new test cases in `tests/test_components/test_mode.py` to verify warning behavior for different plane-symmetry crossing scenarios
- Added option to suppress duplicate warnings when creating and running mode simulations

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -1701,6 +1701,7 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
         boundary_spec: BoundarySpec = None,
         grid_spec: Union[GridSpec, Literal["identical"]] = None,
         symmetry: Optional[tuple[Symmetry, Symmetry, Symmetry]] = None,
+        warn_symmetry_expansion: bool = True,
         sources: Optional[tuple[SourceType, ...]] = None,
         monitors: Optional[tuple[MonitorType, ...]] = None,
         remove_outside_structures: bool = True,
@@ -1728,6 +1729,8 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
             New simulation symmetry. If ``None``, then it is inherited from the original
             simulation. Note that in this case the size and placement of new simulation domain
             must be commensurate with the original symmetry.
+        warn_symmetry_expansion : bool = True
+            Whether to warn when the subsection is expanded to preserve symmetry.
         sources : Tuple[SourceType, ...] = None
             New list of sources. If ``None``, then the sources intersecting the new simulation
             domain are inherited from the original simulation.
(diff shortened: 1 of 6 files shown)


dbochkov-flexcompute:
makes sense and looks good to me. Indeed, I cannot recall a situation where this symmetry argument was used. In an old discussion Tyler actually suggested just to error is new domain is not centered around symmetry plane https://github.com/flexcompute/tidy3d/issues/1191, but I guess I left it as automatic expansion + error

momchil-flex:
@caseyflex I also added an update to the `ModeSimulation` plotting, and also passing a few more arguments around like `fill_structures`. Let me know if it makes sense to you!

caseyflex (approved):
Looks good! My only comment is that I'm not sure about removing the symmetry argument from subsection. What if you want a subsection of the simulation that has more symmetry than the full simulation? Like a mode solver around one waveguide that is symmetric, when the full simulation is not symmetric.

momchil-flex:
Then just do an updated copy after the reduction?

caseyflex:
> Then just do an updated copy after the reduction?

Sure I’m fine either way. I guess we could keep backward compatibility by keeping the parameter. And it helps to suggest it as something the user might want to specify

tidy3d/components/simulation.py
```diff
@@ -1784,33 +1782,32 @@ def subsection(
                     new_bounds[0][dim] = grids_1d[dim][grid_inds[dim][0]]
                     new_bounds[1][dim] = grids_1d[dim][grid_inds[dim][1]]
 
-        # if symmetry is not overriden we inherit it from the original simulation where is needed
-        if symmetry is None:
```
caseyflex: Actually this is another case where I slightly prefer the previous handling. Previously if the user specified symmetry, then it wouldn't be inferred or expanded at all. In the new handling, there is no way to prevent automatic symmetry expansion.

momchil-flex: Fair enough I reverted it!

caseyflex: thanks!

PR #2627, opened 2025-07-04
Move imaginary factor in `MonitorData.get_amplitude()` to internal adjoint calculation

yaugenst-flex:
## Greptile Summary

Modifies the complex amplitude handling in `MonitorData.get_amplitude()` by relocating the imaginary factor to the adjoint source calculation for improved logical separation.

- Removes factor of `1j` from `tidy3d/components/data/monitor_data.py:get_amplitude()`, moving it to `_adjoint_source_amp()`
- Changes flux calculations to use `abs()` instead of direct comparison for zero amplitude checks
- Updates documentation in `CHANGELOG.md` to reflect breaking change in amplitude calculations
- Potential breaking change for users relying on complex phase in amplitude calculations

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -185,9 +185,9 @@ class MonitorData(AbstractMonitorData, ABC):
         """Get the complex amplitude out of some data."""
 
         if isinstance(x, DataArray):
-            x = complex(x.values)
+            x = x.values
 
-        return 1j * complex(x)
+        return complex(x)
 
 
 class AbstractFieldData(MonitorData, AbstractFieldDataset, ABC):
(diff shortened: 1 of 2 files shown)


tidy3d/components/data/monitor_data.py
```diff
@@ -2115,7 +2115,7 @@ def _make_adjoint_sources_amps(self, fwidth: float) -> list[ModeSource]:
                 for mode_index in coords["mode_index"]:
                     amp_single = self.amps.sel(f=freq, direction=direction, mode_index=mode_index)
 
-                    if self.get_amplitude(amp_single) == 0.0:
+                    if abs(self.get_amplitude(amp_single)) == 0.0:
```
momchil-flex, line 2118: do we want an `isclose` here or could it be a problem in some weirdly scaled setups?

tidy3d/components/data/monitor_data.py
```diff
@@ -3404,7 +3404,7 @@ def _make_adjoint_sources_amps(self, fwidth: float) -> list[PlaneWave]:
 
                         # ignore any amplitudes of 0.0 or nan
                         amp_complex = self.get_amplitude(amp_single)
-                        if (amp_complex == 0.0) or np.isnan(amp_complex):
+                        if (abs(amp_complex) == 0.0) or np.isnan(amp_complex):
```
momchil-flex, line 3407: same here

momchil-flex (approved):

tidy3d/components/data/monitor_data.py
yaugenst-flex: Yeah I'd be conservative with this for now and revisit at some point. Don't want to break the way it currently works :smile:

PR #2628, opened 2025-07-04
ci: Enforce stricter `pr-requirements-pass`

daquinteroflex:
## Greptile Summary

Enhanced PR validation requirements in CI workflow by implementing stricter test controls and failure handling.

- Modified `.github/workflows/run_tests.yml` to cancel in-progress local test jobs, preventing stale test states
- Updated shell script syntax in pr-requirements-pass job from double to single quotes for better compatibility
- Added strict failure handling to ensure PRs cannot proceed without passing remote tests after approval

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -326,7 +326,11 @@ jobs:
 
   pr-requirements-pass:
     name: pr-requirements-pass
-    if: always() && ( needs.determine-test-scope.outputs.local_tests == 'true' ) || ( needs.determine-test-scope.outputs.remote_tests == 'true' ) # always will run if dependent job is skipped
+    if: |
+      always() && 
+      ( needs.determine-test-scope.outputs.pr_approval_state == 'true' ) && 
+      ( needs.determine-test-scope.outputs.local_tests == 'true' ) || 
+      ( needs.determine-test-scope.outputs.remote_tests == 'true' )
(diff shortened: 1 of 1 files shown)


daquinteroflex:
Skipped when in draft 
image

daquinteroflex:
Failing correctly as remote-tests were skipped due to approval missing in an open PR, this will block merging with just PR open
image

daquinteroflex:
@yaugenst-flex would be good to chat about this stricter merge logic, if this is what we want?
- `pr-requirements-pass` will only turn green when local, remote, lint and approval has been provided. Otherwise it will not be skipped and fail
    - Developers can still see if local tests are passing when PR is open
    - Draft PR will skip everything except verification it is a draft
    - When a PR is open, but not approved, it will skip the `pr-requirements-test` check
- On approval, it should run all tests per change and generate the `pr-requirements-tests` flag that will not be skipped

daquinteroflex:
Actually github will prevent merging without approval so we can probably skip the `pr-requiremnts-pass` check unless an approval has been provided, will update that

daquinteroflex:
Now it skips the final approval unless the PR has been approved
image

yaugenst-flex (approved):

yaugenst-flex (changes requested):
👎

yaugenst-flex (approved):
👍

PR #2632, opened 2025-07-07
Turning a few validators into root validators

momchil-flex:
Some validators were causing loading a `VolumeMesherData` to issue spurious warnings due to `skip_if_fields_missing`. One fix is to just these into root validators - and the issue should anyway be addressed in the pydantic v2 migration which completely eliminates `skip_if_fields_missing`.

image

## Greptile Summary

Converts several field validators to root validators to fix spurious warnings when loading VolumeMesherData caused by `skip_if_fields_missing`. This change affects validation logic in simulation data handling and unstructured grid datasets.

- Modified `data_monitors_match_sim` in `components/base_sim/data/sim_data.py` to use root validation instead of field validation
- Converted validation in `components/tcad/data/sim_data.py` to use root validators for monitor data correspondence
- Updated `UnstructuredGridDataset` in `components/data/unstructured/base.py` to handle point and value count validation at root level
- Serves as temporary solution until pydantic v2 migration removes need for `skip_if_fields_missing`

### DIFF
--- a/tidy3d/components/base_sim/data/sim_data.py
+++ b/tidy3d/components/base_sim/data/sim_data.py
@@ -52,15 +52,15 @@ class AbstractSimulationData(Tidy3dBaseModel, ABC):
         """Dictionary mapping monitor name to its associated :class:`AbstractMonitorData`."""
         return {monitor_data.monitor.name: monitor_data for monitor_data in self.data}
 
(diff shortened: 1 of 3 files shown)


yaugenst-flex (approved):
LGTM!

PR #2640, opened 2025-07-09
Add the effective DOS and the energy band gap models

damiano-flex:
## Greptile Summary

Refactors semiconductor material modeling by introducing modular DOS and band gap models, replacing hardcoded values with configurable model classes.

- Added new model classes in `tidy3d/__init__.py` including `ConstantEffectiveDOS`, `IsotropicEffectiveDOS`, `MultiValleyEffectiveDOS`, and various energy band gap models
- Modified `SemiconductorMedium` in `components/material/tcad/charge.py` to use model-based implementations for `N_c`, `N_v`, and `E_g` parameters
- Updated cSi material properties in `material_library/material_library.py` to use new model classes while preserving existing values
- Modified test cases in `test_heat_charge.py` to use the new model classes instead of hardcoded values


tidy3d/components/tcad/bandgap_energy.py
```diff
@@ -0,0 +1,111 @@
...
+    Models the temperature dependence of the energy band gap (Eg) using a
+    quadratic approximation.
+
+    Notes
+    -----
+    The model uses the following formula:
+
+        .. math::
+
+            E_g(T) = E_g(300) + \\alpha T + \\beta T^2
```
marc-flex, line 30: Can you maybe add a dot between the constant and the `T`? I.e., `\\alpha \\dot T`

tidy3d/components/tcad/bandgap_energy.py
```diff
@@ -0,0 +1,111 @@
...
+        title="Band Gap",
+        description="Energy band gap",
+        units=ELECTRON_VOLT,
+    )
+
+
+class QuadraticEnergyBandGap(Tidy3dBaseModel):
+    """
+    Models the temperature dependence of the energy band gap (Eg) using a
+    quadratic approximation.
```
marc-flex, line 22: When I get home I'll give you a reference for this.

tidy3d/material_library/material_library.py
```diff
@@ -2026,9 +2028,9 @@ def medium(self, optical_axis: Axis):
         optical=cSi_Green2008.medium,
         charge=SemiconductorMedium(
             permittivity=11.7,
-            N_c=2.86e19,
-            N_v=3.1e19,
-            E_g=1.11,
+            N_c=ConstantEffectiveDOS(N=2.86e19),
+            N_v=ConstantEffectiveDOS(N=3.1e19),
+            E_g=ConstantEnergyBandGap(eg=1.11),
```
marc-flex, line 2033: Maybe we can add the temperature-dependence models by default? Maybe we should first check that we get the same values at 300K, though

marc-flex (commented):
This is great. Thanks @damiano-flex

PR #2643, opened 2025-07-09
feat(web): add nexus environment (FXC-1750)

frederikschubertflex:
This adds a new function to our web client to enable using it with a Nexus deployment.

```py
import tidy3d.web as web

web.use_nexus()
web.configure(apikey="...") # obtained from Nexus instance
```

## Greptile Summary

Adds Nexus deployment support to Tidy3D web client via `web.use_nexus()` function, enabling connection to local Nexus instances.

- Function hardcodes ports (5000, 9000) and disables SSL verification, introducing potential security vulnerabilities
- Missing input validation for IP address parameter could lead to runtime issues
- Configuration sequence requires careful user documentation with explicit examples
- Changes should consider environment variable support for flexible deployment
- Implementation may need error handling for connection failures


daquinteroflex (approved):
Nice, I like it!

tidy3d/web/__init__.py
```diff
@@ -66,4 +67,5 @@
     "test",
     "tidy3d_cli",
     "upload",
+    "use_nexus",
```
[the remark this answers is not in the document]
daquinteroflex: We should talk this in the jira ticket

tidy3d/web/core/nexus.py
```diff
@@ -0,0 +1,19 @@
...
+    local = EnvironmentConfig(
+        name="local",
+        web_api_endpoint=f"http://{ip_address}:5000",
+        ssl_verify=False,
+        enable_caching=False,
+        s3_region="us-east-1",
+        website_endpoint=f"http://{ip_address}/tidy3d",
+    )
+    Env.set_current(local)
+    os.environ["AWS_ENDPOINT_URL_S3"] = f"http://{ip_address}:9000"
```
yaugenst-flex: What's the reason for doing this differently than the other envs in `environment.py`? Because you also have to set an env var here? I prefer this approach. Would be a good opportunity to get rid of the env instantiation on import for the other envs too, before we diverge too much?

yaugenst-flex (commented):
Thanks @frederikschubertflex, left a comment. It's not blocking so if this is urgent then feel free to merge.

tidy3d/web/core/nexus.py
frederikschubertflex: Yes, I wanted to encapsulate the whole setup (including the env var) in one function that can be explicitly called. Do you mean we should create functions like `use_prod`, `use_dev`, `use_uat` and `use_pre`? which modify the current `Env` instance?

yaugenst-flex: Yeah so dwelling on this some more there is actually a potential problem with this approach in that this env var is not cleared when switching envs. Granted this is probably not something you'd typically run into but still an edge case. So here is a design that could work:
- All envs as a dict (e.g. in `environment.py`)
- Expand `EnvironmentConfig` with a field that accepts env vars it should set on env activation (and caches old ones to restore on teardown) and `deactivate()` method
- Make `EnvironmentManager` check for and deactivate `_current` before activating new env
- Create a new env variable `TIDY3D_NEXUS_IP` or similar

And interface would be mostly unchanged:
```py
Env.set_current("prod")
Env.set_current("nexus")  # deactivates "prod" and initializes "nexus"
```

I haven't fully thought this through but basically if it's avoidable I think we shouldn't make a new module for an env that's just slightly different than the other ones if at all possible. We could keep the special `use_nexus()` function (or method) around when you want to pass the IP at runtime.

frederikschubertflex: I also thought about this issue with the env var persisting for the rest of the session and I agree, this is sub-optimal.
I like the generality of the environment variable solution and am happy to dispose of the `use_nexus` function.  `Env.set_current("nexus")` feels more coherent.

frederikschubertflex: Updated the code to match your proposal @yaugenst-flex

yaugenst-flex: Awesome yeah this looks great. Minimally invasive 😃

yaugenst-flex (approved):
LGTM!

PR #2649, opened 2025-07-10
download files atomically from s3

### LINKED ISSUE #2648: File downloads should be atomic
File downloads should be atomic.

That means if the file exists at the target location with the correct name, we can assume the file transferred successfully.

**Suggested implementation**
- Download the file to a temporary location *(in the same folder)* with a different name.  After download is successful, rename the file to the target name (eg. using `os.rename()`).

**Describe alternatives you've considered**
- More documentation to inform users how to check if their file download was successful.

## Additional context
- Solving this problem will make it easier to solve problems like issue #2600 .  All we have to do is check to see if the file exists at the target location.
- Currently, all Tidy3D downloads use [a single function](https://github.com/flexcompute/tidy3d/blob/bcf635fa428dc99020b47b9df76ad67bef660027/tidy3d/web/core/s3utils.py#L322)   So we only have to update this one function.  (That function is a wrapper for `S3.Client.download_file()` from amazon's `boto3` module.  We just have to add a few lines of logic to that function to rename the file afterwards.)  

## Priority
Moderate.  Issue 2600 is a higher priority.  Making file downloads atomic will make that issue easier to solve.  But if for some reason, solving this issue is harder than we thought to implement, then we can try and implement some kind of work-around or warning.

jewettaijfc:
## Summary
The `s3utils.download_file()` function is currently used to download all files from amazon S3.  After this PR, we can be confident that if a file exists at the target destination, then the download was successful.  Temporary files that are generated during the transfer process are cleaned up automatically.

## Implementation
- Files downloaded from amazon s3 are given temporary names while they are in transit.
- After they have arrived, they are renamed to the correct name (using `os.rename()`)
- If an exception is raised during transit, the temporary file is erased.
- ~~If the python process was interrupted, the temporary file will be erased the next time we try to download the file.~~  ***EDIT: Instead we now use randomly generated temporary file names to eliminate file-name collision risk.  Deleting interrupted downloads is the responsibility of the user.***

## Greptile Summary

Implements atomic file downloads from S3 to prevent corrupted or partial downloads by using temporary files and proper cleanup mechanisms.

- Modified `tidy3d/web/core/s3utils.py` to download files with '.in_transit' suffix and use `os.rename()` for atomic moves
- Added safety cleanup of temporary files during failed downloads or process interruptions
- Added comprehensive tests in `tests/test_web/test_s3utils.py` to verify download atomicity and cleanup behavior
- Updated CHANGELOG.md to document improved download reliability

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 or `Absorber` classes (or when invoking `pml()`, `stable_pml()`, or `absorber()` functions)
 with fewer layers than recommended.
 - Warnings and error messages originating from `Structure`, `Source`, or `Monitor` classes now refer to problematic objects by their user-supplied `name` attribute, alongside their index.
+- File downloads are atomic. Interruptions or failures during download will no longer result in incomplete files.
 
 ### Fixed
 - Arrow lengths are now scaled consistently in the X and Y directions, and their lengths no longer exceed the height of the plot window.
--- a/tests/test_web/test_s3utils.py
+++ b/tests/test_web/test_s3utils.py
@@ -1,0 +1,122 @@
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
+import pytest
+
+import tidy3d
+from tidy3d.web.core import s3utils
+
+
+@pytest.fixture
+def mock_S3STSToken(monkeypatch):
+    mock_token = MagicMock()
+    mock_token.cloud_path = ""
+    mock_token.user_credential = ""
+    mock_token.get_bucket = lambda: ""
+    mock_token.get_s3_key = lambda: ""
+    mock_token.is_expired = lambda: False
+    mock_token.get_client = lambda: tidy3d.web.core.s3utils.boto3.client()
+    monkeypatch.setattr(
+        target=tidy3d.web.core.s3utils, name="_S3STSToken", value=MagicMock(return_value=mock_token)
+    )
+    return mock_token
+
+
+@pytest.fixture
+def mock_get_s3_sts_token(monkeypatch):
+    def _mock_get_s3_sts_token(resource_id, remote_filename):
+        return s3utils._S3STSToken(resource_id, remote_filename)
+
+    monkeypatch.setattr(
+        target=tidy3d.web.core.s3utils, name="get_s3_sts_token", value=_mock_get_s3_sts_token
+    )
+    return _mock_get_s3_sts_token
+
+
+@pytest.fixture
+def mock_s3_client(monkeypatch):
+    """
+    Fixture that provides a generic mock S3 client.
+    Method-specific side_effects are omitted here and are specified later in the unit tests.
+    """
+    mock_client = MagicMock()
+    # Patch the `client` as it is imported within `tidy3d.web.core.s3utils.boto3` so that
+    # whenever it's invoked (for example with "s3"), it returns our `mock_client`.
+    monkeypatch.setattr(
+        target=tidy3d.web.core.s3utils.boto3,
+        name="client",
+        value=MagicMock(return_value=mock_client),
+    )
+    return mock_client
+
+
+def test_download_s3_file_success(mock_s3_client, mock_get_s3_sts_token, mock_S3STSToken, tmp_path):
+    """Tests a successful download."""
+    destination_path = tmp_path / "downloaded_file.txt"
+    expected_content = "abcdefg"
+
+    def simulate_download_success(Bucket, Key, Filename, Callback, Config, **kwargs):
+        with open(Filename, "w") as f:
+            f.write(expected_content)
+        return None
+
+    mock_s3_client.download_file.side_effect = simulate_download_success
+    mock_S3STSToken.get_bucket = lambda: "test-bucket"
+    mock_S3STSToken.get_s3_key = lambda: "test-key"
+
+    s3utils.download_file(
+        resource_id="1234567890",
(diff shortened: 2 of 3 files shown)


tidy3d/web/core/s3utils.py
```diff
@@ -312,12 +312,12 @@ def download_file(
     # set to_file if None
     if not to_file:
         path = pathlib.Path(resource_id)
-        to_file = path / remote_basename
+        to_path = path / remote_basename
```
jewettaijfc, line 317: Perhaps it's a personal peeve of mine, but I don't like changing the *type* of a variable after it is defined (eg. from `str` to `pathlib.Path`).  It makes it harder for me to remember what each variable is.

So I introduced new variable names for variables of type `pathlib.Path` (`to_file`->`to_path`).

tidy3d/web/core/s3utils.py
```diff
@@ -327,14 +329,24 @@ def _download(_callback: Callable) -> None:
...
+                Bucket=token.get_bucket(),
+                Filename=str(to_path_tmp),
+                Key=token.get_s3_key(),
+                Callback=_callback,
+                Config=_s3_config,
+            )
+        except Exception as e:
+            to_path_tmp.unlink(missing_ok=True)  # Delete incompletely downloaded file.
+            raise e
+        os.rename(str(to_path_tmp), str(to_path))  # Rename temporary file to the correct name.
```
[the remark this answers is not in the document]
yaugenst-flex: @jewettaijfc This is actually a good suggestion, `shutil.move` is more robust. E.g. when moving files from `/tmp` as we should probably be doing here one easily runs into cross-device moves.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 or `Absorber` classes (or when invoking `pml()`, `stable_pml()`, or `absorber()` functions)
 with fewer layers than recommended.
 - Warnings and error messages originating from `Structure`, `Source`, or `Monitor` classes now refer to problematic objects by their user-supplied `name` attribute, alongside their index.
+- File downloads are atomic.  Interruptions or failures during download will no longer result in incomplete files.
```
yaugenst-flex: There is an extra space after the period: `atomic.ᵛᵛInterruptions`

tests/test_web/test_s3utils.py
```diff
@@ -0,0 +1,123 @@
...
+        progress_callback=None,
+    )
+
+    # Check that mock_s3_client.download_file() was invoked with the correct arguments.
+    mock_s3_client.download_file.assert_called_once()
+    call_args, call_kwargs = mock_s3_client.download_file.call_args
+    assert call_kwargs["Bucket"] == "test-bucket"
+    assert call_kwargs["Key"] == "test-key"
+    assert call_kwargs["Filename"] == str(destination_path) + s3utils.IN_TRANSIT_SUFFIX
+    # Check that file exists.
```
yaugenst-flex:
```
# Check that file exists.
assert destination_path.exists()
```
Remove comment 😉

tidy3d/web/core/s3utils.py
yaugenst-flex: Yeah I agree that's good practice.

tidy3d/web/core/s3utils.py
```diff
@@ -327,14 +329,24 @@ def _download(_callback: Callable) -> None:
...
-            Key=token.get_s3_key(),
-            Callback=_callback,
-            Config=_s3_config,
-        )
+        # Caller can assume the existence of the file means download succeeded.
+        # So make sure this file does not exist until that assumption is true.
+        to_path.unlink(missing_ok=True)
+        # Download to a temporary file.
+        to_path_tmp = to_path.with_suffix(to_path.suffix + IN_TRANSIT_SUFFIX)
+        to_path_tmp.unlink(missing_ok=True)  # Delete file from an interrupted download attempt.
```
yaugenst-flex: Shall we use a truly unique name here instead? I understand we might not want to download to `/tmp` or similar but we could still use `tmpfile` to generate the name?
Something like:
```py
fd, tmp_file_path_str = tempfile.mkstemp(suffix=".in_transit", dir=to_path.parent)
os.close(fd)
```

yaugenst-flex (commented):
Thanks @jewettaijfc! Some minor questions/clarifications but overall looks great!

tests/test_web/test_s3utils.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def test_download_s3_file_raises_oserror(
+    mock_s3_client, mock_get_s3_sts_token, mock_S3STSToken, tmp_path
+):
+    """Tests download failing with an ``OSError`` (No space left on device)."""
+    destination_path = tmp_path / "downloaded_file.txt"
+    destination_path_tmp = destination_path.with_suffix(
+        destination_path.suffix + s3utils.IN_TRANSIT_SUFFIX
+    )
```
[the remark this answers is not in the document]
yaugenst-flex: Doesnt seem necessary here.

jewettaijfc:
I just deleted an empty file (named `tests/test_web/test_container.py`), but I made no other changes.  If you made any comments on the state of my code any time in the last 2 hours, they are still valid.  I did not change anything else.

*(Sorry.  It was created by my vscode editor which had an open file created in a different branch I'm working on.  I just noticed it now.)*

jewettaijfc:
I just rebased this branch on top of the most recent changes to origin/develop.  No other changes were made.

yaugenst-flex (approved):
LGTM!

PR #2650, opened 2025-07-10
improve support for lossy transmission lines and complex reference impedances

dmarek-flex:
~$Z_{ref}$ is now $Z_{L}^{\*}$ where $Z_{L}^{\*}$ would be the lumped port load impedance or the characteristic impedance of a transmission line mode in a `WavePort`.~

We stick with using the unconjugated value for reference impedance, but now we switch our default approach for extracting S parameters to the pseudo-wave formulation. 

Also includes a fix for calculating voltage from current + flux as well as calculating current from voltage + flux, when modes are lossy.

~We are still missing proper calculation of mode amps using the non conjugated version of the dot product for lossy modes.~
Completed

backend: https://github.com/flexcompute/compute/pull/2171

## Greptile Summary

Improved handling of lossy modes in WavePort by updating reference impedance calculations and fixing voltage/current computations for transmission line modes.

- Changed reference impedance in `tidy3d/plugins/smatrix/component_modelers/terminal.py` to use complex conjugate of load impedance (Z_L*) for power wave calculations
- Fixed voltage/current calculations in `tidy3d/plugins/smatrix/ports/wave.py` to properly handle directionality and complex flux values for lossy modes
- Updated port_reference_impedances method in TerminalComponentModeler to align with Pozar's Microwave Engineering theoretical framework


tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -55,13 +70,18 @@ class TerminalComponentModeler(AbstractComponentModeler[NetworkIndex, NetworkEle
...
-        "when there are reflections from simulation boundaries. ",
+        description="If ``True``, only the excited port is assumed to have a nonzero incident wave "
+        "amplitude power. This choice simplifies the calculation of the scattering matrix. "
+        "If ``False``, every entry in the vector of incident wave amplitudes (a) is calculated "
+        "explicitly. This choice requires a matrix inversion when calculating the scattering "
+        "matrix, but may lead to more accurate scattering parameters when there are "
+        "reflections from simulation boundaries. ",
+    )
+
+    S_def: SDef = pd.Field(
```
weiliangjin2021: trying to compare the following terminologies for the field name: `wave_choice`, `basis_choice`, `s_param_def`, and the current one.

dmarek-flex: Yea not sure `S_def` is the best, scikit-rf uses `s_def` which is nice. I had `wave_type` before, and `wave_def` or similar could work.

I don't like `basis_choice`, `s_param_def` might be the clearest.

weiliangjin2021: yeah, `s_param_def` is clearer.

dbochkov-flexcompute: could be even `s_matrix_def`

weiliangjin2021: that also sounds good

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -57,6 +92,45 @@ def _warn_rf_license(cls, values):
...
+            return s_param_def
+
+        assume_ideal_excitation = values.get("assume_ideal_excitation")
+        if assume_ideal_excitation is True:
+            raise SetupError(
+                "Setting `assume_ideal_excitation` to ``True`` is incompatible with "
+                "using the power wave definition of scattering parameters. "
+                "Either use the pseudo wave definition by setting 's_param_def' to 'pseudo', "
+                "or set 'assume_ideal_excitation' to ``False``. "
+            )
```
weiliangjin2021: could you educate me on why they are incompatible?

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: Went with `s_param_def` at the moment, because the definition is sort of more fundamental than the S matrix itself, because it tells you how to interpret each a,b amplitudes at ports.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -612,24 +864,26 @@ def get_antenna_metrics_data(
...
         # Retrieve associated simulation data
         combined_directivity_data = None
         for port, amplitude in port_dict.items():
             sim_data_port = self.batch_data[self._task_name(port=port)]
             radiation_data = sim_data_port[rad_mon.name]
 
-            a, b = self.compute_power_wave_amplitudes_at_each_port(
-                self.port_reference_impedances, sim_data_port
+            a, b = self.compute_wave_amplitudes_at_each_port(
+                self.port_reference_impedances, sim_data_port, s_param_def="pseudo"
```
weiliangjin2021: should `s_param_def` be an option in `get_antenna_metrics_data`?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -57,6 +92,45 @@ def _warn_rf_license(cls, values):
...
+
+        element_mappings = values.get("element_mappings")
+        if len(element_mappings) != 0:
+            raise SetupError(
+                "Computing scattering parameters using the power wave definition "
+                "requires running a number of simulations equal to the number of ports. "
+                "Either use the pseudo wave definition by setting 's_param_def' to 'pseudo', "
+                "or ensure all simulations are run by keeping 'element_mappings' as its "
+                "default value of an empty 'tuple'. "
+            )
```
weiliangjin2021: also for the two validations here, why is there additional constraint in power wave definition?

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: I had just written out a long explanation to justify this, and realized it is no longer true :sweat_smile: 

I think now that I switched back to using the reference impedance of the mode, and not the conjugate, this restriction is not needed for power waves.

dmarek-flex: I'll just test it out first, but likely we dont need this complicated validation step

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: same as above, so probably not needed after all

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: this is a good question... I think the calculations in this part are more convenient using power waves, which has a clearer distinction between incident and reflected power. In fact, my calculation for power is incorrect using "pseudo" wave amplitudes.

I think in this case we want to use "power", which makes the following incident and backward power calculations correct. Need to work out the exact difference between $$a_{pseudo}$$ and $$a_{power} $$, since the function asks the user to supply an $$a$$ value. I think we just need to clarify that the amplitude passed in represents a power wave amplitude

dmarek-flex: Alright, so for this method we simply use the power wave amplitudes since they are more convenient. It is also conceptually easier for the user to use power wave definition for this method, since the power flow can be decomposed into incident and reflected powers, which is not the case for pseudo waves. I added a short comment to explain the `port_amplitude` a bit more.

weiliangjin2021 (approved):

PR #2652, opened 2025-07-11
refactor: remove debug print statement from sim_data.py

yaugenst-flex:
Accidentally left this one in :grimacing: 

## Greptile Summary

Removes an accidentally committed debug print statement from the `_adjoint_src_width_broadband` function in `tidy3d/components/data/sim_data.py`.

- Removed debug print that was outputting source width calculations (decay_by_f0_fwidth and fwidth_to_min_f0) used in adjoint source calculations


groberts-flex (approved):
oops thank you for catching this!

PR #2654, opened 2025-07-14
Interrupted Batch downloads can be resumed

### LINKED ISSUE #2600: Downloading batch files should check to see if any data is already downloaded
# Problem Description 

After running a `web.Batch()` of simulations, it's time to download the results. 
This is typically not a problem, but particularly for large batches I've run into: 
1. My computer decides to restart, killing the download task (it's windows...)
2. There's some kind of connection issue and the download fails

I can easily use `batch.load()` or `Batch.download()` to get new files, but at the moment this re-downloads _every file in the batch_. 

# Desired Solution

I would love if there were an option to only download the "outstanding" files -- I'm already providing the download directory, so just check the set difference between the batch filenames and the filenames in the target directory and download the difference. 

An even more advanced version of this would check that the downloaded files are complete, and finish downloading those which got cut off.

jewettaijfc:
This PR addresses issue #2600 

This PR:
- Adds the `replace_existing` argument to `Batch.load()`, `Batch.download()`, and `BatchData.load()` functions from `container.py`.  This allows these functions to skip over previously downloaded files.
- For *these batch functions*, the default value of `replace_existing` is `False`.
- A single warning is generated if it looks like any files have already been downloaded.  It explains how to override the default behavior (using `replace_existing==True`).
- I refrained from making any modifications which did not address issue 2600, ***except:***
- I followed Tyler's advice and moved the call to `Batch.download()` out of `Batch.run()` and into `Batch.load()`.
- `BatchData.load_sim_data()` (which downloads only one file) now invokes `web.load()` with `replace_existing=True`, instead of `False`.  (This change was made to be consistent with the other single-file download functions.  See below.)

### Single-file download functions overwrite existing files

There are also at least 10 different functions in `container.py`, `webapi.py`, and `s3utils.py` which download only *a single simulation file*.  After this PR, *all* of these single-file functions have the *opposite* behavior:  They will *overwrite* a previously downloaded file.  (Previously, some did and others did not.)  

I think this should be the default behavior for functions that download single files: It is the responsibility of the user to decide whether or not to download the file.

### Discussion
*For batch simulations, I think the situation is different because it's unreasonable to ask users to specify all of the files they want to download manually.  That's why I only make it possible to automatically skip previously downloaded **batch** simulation files.*

## How to test this code
This code is particularly difficult code to write a unit test for *(since it requires mocking in a multi-threaded environment)*.  Instead I tested it manually.

Test strategy

- You cannot test this code in this branch because currently the web-API is not working.  Ideally, you should rebase the changes from this PR onto a working release branch  (v2.9.0rc1).  *(Unfortunately I had trouble rebasing this PR onto v2.9.0rc1 due to a large number of conflicts in unrelated files.)*
- Fortunately all the changes in this PR are in one file (container.py).  So we just need to copy that file over to the v2.9.0rc1 version and run the tests.  (See details below.)

### Detailed test instructions:
```
git checkout jewettaijfc/issue2600  # checkout this branch
cp -f tidy3d/web/api/container.py /tmp/
git checkout v2.9.0rc1  # web.api works in this branch
cp -f /tmp/container.py tidy3d/web/api/  # (this effectively rebases on v2.9.0rc1)
```
***Then*** follow the instructions in this file: [**test_pr2654.py.zip**](https://github.com/user-attachments/files/21302613/test_pr2654.py.zip)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- By default, batch downloads will skip files that already exist locally. To force re-downloading and replace existing files, pass the `replace_existing=True` argument to `Batch.load()`, `Batch.download()`, or `BatchData.load()`.
+- The `BatchData.load_sim_data()` function now overwrites any previously downloaded simulation files (instead of skipping them).
+
 ### Fixed
 - Giving opposite boundaries different names no longer causes a symmetry validator failure.
 
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -431,12 +431,7 @@ class BatchData(Tidy3dBaseModel, Mapping):
         task_id = self.task_ids[task_name]
         web.get_info(task_id)
 
-        return web.load(
-            task_id=task_id,
-            path=task_data_path,
-            replace_existing=False,
-            verbose=False,
-        )
+        return web.load(task_id=task_id, path=task_data_path, verbose=False)
 
     def __getitem__(self, task_name: TaskName) -> SimulationDataType:
         """Get the simulation data object for a given ``task_name``."""
@@ -451,7 +446,7 @@ class BatchData(Tidy3dBaseModel, Mapping):
         return len(self.task_paths)
 
     @classmethod
-    def load(cls, path_dir: str = DEFAULT_DATA_DIR) -> BatchData:
+    def load(cls, path_dir: str = DEFAULT_DATA_DIR, replace_existing: bool = False) -> BatchData:
         """Load :class:`Batch` from file, download results, and load them.
 
         Parameters
@@ -459,6 +454,8 @@ class BatchData(Tidy3dBaseModel, Mapping):
         path_dir : str = './'
             Base directory where data will be downloaded, by default current working directory.
             A `batch.hdf5` file must be present in the directory.
+        replace_existing : bool = False
+            Downloads the data even if path exists (overwriting the existing).
 
         Returns
         ------
@@ -469,7 +466,7 @@ class BatchData(Tidy3dBaseModel, Mapping):
 
         batch_file = Batch._batch_path(path_dir=path_dir)
         batch = Batch.from_file(batch_file)
-        return batch.load(path_dir=path_dir)
+        return batch.load(path_dir=path_dir, replace_existing=replace_existing)
 
 
 class Batch(WebContainer):
@@ -606,7 +603,6 @@ class Batch(WebContainer):
         self.upload()
         self.start()
         self.monitor()
-        self.download(path_dir=path_dir)
         return self.load(path_dir=path_dir)
 
     @cached_property
@@ -900,13 +896,15 @@ class Batch(WebContainer):
         """
         return os.path.join(path_dir, "batch.hdf5")
 
-    def download(self, path_dir: str = DEFAULT_DATA_DIR) -> None:
+    def download(self, path_dir: str = DEFAULT_DATA_DIR, replace_existing: bool = False) -> None:
         """Download results of each task.
 
         Parameters
         ----------
         path_dir : str = './'
             Base directory where data will be downloaded, by default the current working directory.
+        replace_existing : bool = False
+            Downloads the data even if path exists (overwriting the existing).
 
         Note
         ----
@@ -919,17 +917,36 @@ class Batch(WebContainer):
         self._check_path_dir(path_dir=path_dir)
         self.to_file(self._batch_path(path_dir=path_dir))
 
+        num_existing = 0
+        for _, job in self.jobs.items():
+            job_path_str = self._job_data_path(task_id=job.task_id, path_dir=path_dir)
+            if os.path.exists(job_path_str):
+                num_existing += 1
+        if num_existing > 0:
+            files_plural = "files have" if num_existing > 1 else "file has"
+            log.warning(
+                f"{num_existing} {files_plural} already been downloaded "
+                f"and will be skipped. To forcibly overwrite existing files, invoke "
+                "the load or download function with `replace_existing=True`.",
+                log_once=True,
+            )
+
         with ThreadPoolExecutor(max_workers=self.num_workers) as executor:
             fns = []
             for task_name, job in self.jobs.items():
-                job_path = self._job_data_path(task_id=job.task_id, path_dir=path_dir)
-
+                job_path_str = self._job_data_path(task_id=job.task_id, path_dir=path_dir)
+                if os.path.exists(job_path_str):
+                    if replace_existing:
+                        log.info(f"File '{job_path_str}' already exists. Overwriting.")
+                    else:
+                        log.info(f"File '{job_path_str}' already exists. Skipping.")
+                        continue
                 if "error" in job.status:
                     log.warning(f"Not downloading '{task_name}' as the task errored.")
                     continue
 
-                def fn(job=job, job_path=job_path) -> None:
-                    return job.download(path=job_path)
+                def fn(job=job, job_path_str=job_path_str) -> None:
+                    return job.download(path=job_path_str)
 
                 fns.append(fn)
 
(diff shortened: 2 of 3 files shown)


tidy3d/web/api/container.py
```diff
@@ -919,17 +919,33 @@ def download(self, path_dir: str = DEFAULT_DATA_DIR) -> None:
         self._check_path_dir(path_dir=path_dir)
         self.to_file(self._batch_path(path_dir=path_dir))
 
+        warned_existing = False
```
yaugenst-flex: Use `log_once=True` instead? https://github.com/flexcompute/tidy3d/blob/5c96f7a76ce0fb565849a7136a05460cf2147286/tidy3d/log.py#L316

tidy3d/web/api/container.py
```diff
@@ -919,17 +919,33 @@ def download(self, path_dir: str = DEFAULT_DATA_DIR) -> None:
...
+        for _, job in self.jobs.items():
+            job_path_str = self._job_data_path(task_id=job.task_id, path_dir=path_dir)
+            if os.path.exists(job_path_str) and not replace_existing and not warned_existing:
+                warned_existing = True  # Only print this warning message once.
+                log.warning(
+                    "Some files have already been downloaded and will be skipped. To forcibly "
+                    "overwrite existing files, invoke the load or download function with "
+                    "`replace_existing=True`."
+                )
```
yaugenst-flex: Wonder if it might be helpful to be more specific about which files have already been downloaded? And log for every one?

tidy3d/web/api/container.py
```diff
@@ -919,17 +919,33 @@ def download(self, path_dir: str = DEFAULT_DATA_DIR) -> None:
...
             fns = []
             for task_name, job in self.jobs.items():
-                job_path = self._job_data_path(task_id=job.task_id, path_dir=path_dir)
-
+                job_path_str = self._job_data_path(task_id=job.task_id, path_dir=path_dir)
+                if os.path.exists(job_path_str):
+                    if replace_existing:
+                        log.info(f'File "{job_path_str}" already exists.  Overwriting.')
+                    else:
+                        log.info(f'File "{job_path_str}" already exists.  Skipping.')
```
yaugenst-flex: Superfluous space in these strings. Also, use single quotes instead `'{job_path_str}'`, then the path will be highlighted.

tidy3d/web/api/container.py
```diff
@@ -431,12 +433,7 @@ def load_sim_data(self, task_name: str) -> SimulationDataType:
         task_id = self.task_ids[task_name]
         web.get_info(task_id)
 
-        return web.load(
-            task_id=task_id,
-            path=task_data_path,
-            replace_existing=False,
-            verbose=False,
-        )
+        return web.load(task_id=task_id, path=task_data_path, replace_existing=True, verbose=False)
```
yaugenst-flex: Do I understand correctly that this is the only place where the default behavior changed? If so, then I'd be more explicit about this change in the changelog. The change itself is fine I think.

CHANGELOG.md
```diff
@@ -28,12 +28,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- File downloads are atomic. Interruptions or failures during download will no longer result in incomplete files.
 - Warnings are now generated (instead of errors) when instantiating `PML`, `StablePML`, or `Absorber` classes (or when invoking `pml()`, `stable_pml()`, or `absorber()` functions) with fewer layers than recommended.
 - `Simulation.subsection` can no longer take `symmetry` as an argument - the symmetry is always taken from the original simulation.
 - If a mode simulation is crossing a symmetry plane of the larger simulation domain, but the mode plane is not symmetric, a warning is issued that it will be expanded symmetrically. Previously this warning only happened during the solver run.
 - Enhanced `PolySlab` and `Cylinder` gradient computation via adaptive field sampling along geometry boundaries instead of fixed-grid center sampling.
 - Shape derivatives have been sped up significantly, especially for simulations containing many structures in a `GeometryGroup`.
+- File downloads are atomic.  Interruptions or failures during download will no longer result in incomplete files.
```
yaugenst-flex: Looks like this one slipped in from the last PR

CHANGELOG.md
```diff
@@ -28,12 +28,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Functions that download a *single file from a single simulation* (like `BatchData.load_sim_data()`) will consistently overwrite any previously downloaded simulation files (instead of sometimes skipping them).
```
yaugenst-flex: This should be more explicit. I think this really affects only one function?

CHANGELOG.md
```diff
@@ -28,12 +28,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- However *batch* file downloads will skip over previously downloaded files by default.  *(This can be overridden by passing the `replace_existing=True` argument to the `Batch.load()`, `Batch.download()`, or `BatchData.load()` functions.)*
```
yaugenst-flex: Don't let changelog entries depend on other entries. Remove the text formatting (we don't use it for changelogs beyond inline code blocks and maybe URLs). Remove parantheses around full stand-alone sentences. If the sentence is optional, remove it. For example:
```
By default, batch downloads will skip files that already exist locally. To force re-downloading and replace existing files, pass the `replace_existing=True` argument to `Batch.load()`, `Batch.download()`, or `BatchData.load()`.
```

yaugenst-flex (commented):
Thanks @jewettaijfc! Left some (mostly minor) comments.

jewettaijfc:
I tested the latest version using [**test_pr2654.py.zip**](https://github.com/user-attachments/files/21302613/test_pr2654.py.zip).

The new log.warning() message contains a link to the list of skipped files.

This PR is ready for review again.

CHANGELOG.md
yaugenst-flex: @jewettaijfc

CHANGELOG.md
```diff
@@ -31,12 +31,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - `Simulation.subsection` can no longer take `symmetry` as an argument - the symmetry is always taken from the original simulation.
 - If a mode simulation is crossing a symmetry plane of the larger simulation domain, but the mode plane is not symmetric, a warning is issued that it will be expanded symmetrically. Previously this warning only happened during the solver run.
 - Enhanced `PolySlab` and `Cylinder` gradient computation via adaptive field sampling along geometry boundaries instead of fixed-grid center sampling.
 - Shape derivatives have been sped up significantly, especially for simulations containing many structures in a `GeometryGroup`.
+- File downloads are atomic.  Interruptions or failures during download will no longer result in incomplete files.
+- By default, batch downloads will skip files that already exist locally. To force re-downloading and replace existing files, pass the `replace_existing=True` argument to `Batch.load()`, `Batch.download()`, or `BatchData.load()`.
+- The `BatchData.load_sim_data()` function now overwrites any previously downloaded simulation files (instead of skipping them).
```
yaugenst-flex: There was a release yesterday so these all need to move up into the `[Unreleased]` section again

tidy3d/web/api/container.py
```diff
@@ -919,17 +920,45 @@ def download(self, path_dir: str = DEFAULT_DATA_DIR) -> None:
...
+                        "existing files, invoke the load or download function with "
+                        "`replace_existing=True`.",
+                        log_once=True,
+                    )
+                existing_files.add(job_path_str)
+        if filelist_fd is not None:
+            assert filelist_path
+            with os.fdopen(filelist_fd, "w") as filelist_file:
+                filelist_file.write("\n".join(list(existing_files)) + "\n")
+
```
yaugenst-flex: Hmm not sure if we should be writing this to file. I think the previous approach was better. I was just wondering if we should log all existing files but in practice it probably doesn't matter too much. I'd prefer logging just once over creating a separate files for this.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
```
yaugenst-flex: Looks like a line break slipped in

tidy3d/web/api/container.py
```diff
@@ -328,7 +328,9 @@ def load(self, path: str = DEFAULT_DATA_PATH) -> SimulationDataType:
             Object containing simulation results.
         """
         self._check_path_dir(path=path)
-        data = web.load(task_id=self.task_id, path=path, verbose=self.verbose)
+        data = web.load(
+            task_id=self.task_id, path=path, replace_existing=True, verbose=self.verbose
```
yaugenst-flex: I just noticed, but since `replace_existing=True` is just in place to match the default `web.load()` behavior, it's probably best to just remove this argument entirely and rely on `web.load()`'s default behavior? Same for other functions that just try to inherit the default behavior.

yaugenst-flex (commented):
Thanks @jewettaijfc! Minor comment about argument defaults but looks almost ready to go.

yaugenst-flex (approved):
LGTM!

PR #2657, opened 2025-07-14
Add support for uniform frequency and wavelength sampling in freqs() and ldas()

George-Guryev-flxcmp:
### 📝 Summary
This PR adds a `spacing` argument to the `freqs()` and `wvls()` methods in the `FreqRange` class to support two sampling modes:
- `'uniform_freq'`: samples are evenly spaced in the frequency domain
- `'uniform_wvl'`: samples are evenly spaced in the wavelength domain

### ✅ Changes
- Added `spacing` parameter to `freqs()` and `wvls()` functions
- Updated internal logic to compute either frequency- or wavelength-based grids
- Added validation for `spacing` values
- Improved docstrings for both functions

### 📌 Motivation
This change provides more flexibility in frequency/wavelength sampling...

### 🔍 Usage Example
```python
f = FreqRange(...)
f.freqs(num_points=100, spacing='uniform_freq')
f.wvls(num_points=100, spacing='uniform_wvl')
```

## Greptile Summary

This PR enhances the `FreqRange` class by adding a new `spacing` parameter to both `freqs()` and `wvls()` methods, enabling users to choose between uniform frequency sampling (`'uniform_freq'`) and uniform wavelength sampling (`'uniform_wvl'`). This is a meaningful addition as different applications may require different sampling strategies - uniform frequency sampling is often preferred for spectral analysis, while uniform wavelength sampling is more suitable for certain optical applications.

The implementation properly handles the non-linear relationship between frequency and wavelength (f = c/λ), includes appropriate test coverage, and maintains backward compatibility. The changes are well-documented with clear docstrings and example usage.

## Confidence score: 4/5

1. This PR is safe to merge with standard review scrutiny
2. The implementation is mathematically sound and well-tested, with proper handling of edge cases and validation
3. Key files needing attention:
   - `tidy3d/components/source/freq_range.py`: Verify edge case handling and parameter validation
   - `tests/test_components/test_freq_range.py`: Ensure test coverage is comprehensive for both sampling modes

<sub>2 files reviewed, 3 comments</sub>
<sub>[Edit PR Review Bot Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=tidy3d_2657)</sub>


tidy3d/components/source/freq_range.py
```diff
@@ -149,7 +149,7 @@ def from_wvl_interval(cls, wvl_min: float, wvl_max: float) -> FreqRange:
 
         return cls.from_freq_interval(fmin=fmin, fmax=fmax)
 
-    def freqs(self, num_points: int) -> NDArray[np.float64]:
+    def freqs(self, num_points: int, spacing: str) -> NDArray[np.float64]:
```
dbochkov-flexcompute: probably makes sense to provide default value `spacing: str = "uniform_freq"`

tidy3d/components/source/freq_range.py
```diff
@@ -159,20 +159,38 @@ def freqs(self, num_points: int) -> NDArray[np.float64]:
...
         Returns
         -------
         np.ndarray
             a numpy array of uniformly distributed frequency samples in a frequency range of interest.
         """
+        if spacing == "uniform_wvl":
+            return (
+                np.array([td_const.C_0 / self.lda0])
+                if num_points == 1
+                else np.sort(td_const.C_0 / self.ldas(num_points=num_points, spacing=spacing))
```
dbochkov-flexcompute: we know that wavelength will be in reverse order here, so no need to call sort. Instead we can do
```
td_const.C_0 / self.ldas(num_points=num_points, spacing=spacing)[::-1]

File: tidy3d/components/source/freq_range.py [hunk 3]
```diff
@@ -159,20 +159,38 @@ def freqs(self, num_points: int) -> NDArray[np.float64]:
...
         else:
-            # calculate frequency points and corresponding wavelengths
-            return np.linspace(self.fmin, self.fmax, num_points)
+            raise ValueError(
+                "Invalid `spacing` value in FreqRange.freqs(): expected 'uniform_freq' or 'uniform_wvl'. "
+                f"Received: {spacing!r}. Please provide a valid spacing type."
+            )
 
-    def ldas(self, num_points: int) -> NDArray[np.float64]:
+    def ldas(self, num_points: int, spacing: str) -> NDArray[np.float64]:
```
dbochkov-flexcompute: default value `spacing: str = "uniform_wvl"` here

dbochkov-flexcompute (changes requested):
a few minor comments

tidy3d/components/source/freq_range.py
George-Guryev-flxcmp: yes,  it makes sense.

tidy3d/components/source/freq_range.py
```diff
@@ -159,20 +159,38 @@ def freqs(self, num_points: int) -> NDArray[np.float64]:
         ----------
         num_points : int
             Number of frequency points in a frequency range of interest.
+        spacing: str
```
dbochkov-flexcompute: need to add default values in docstrings too:
```
spacing: str = 'uniform_freq`
```

yaugenst-flex (commented):
Thanks @George-Guryev-flxcmp! The code itself looks fine, but I have some high-level comments/questions:

1. I think we should not use the abbreviation `lda` anywhere. The established convention in the codebase is `wvl`, so introducing a new (equivalent) name is confusing for users, especially since this class actually uses both conventions.
2. `fwidth` and `wvl_width` are really half-widths - as a user, I would expect this to be the actual (full) width. What then adds to the confusion is that [td.GaussianPulse](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.GaussianPulse.html) already has a parameter `fwidth`, where `fwidth` actually means the standard deviation of the frequency. @momchil-flex ?
3. The above makes the method `to_gaussian_pulse` quite problematic, because if you initialize the `FreqRange` class with a certain `fwidth`, this will _not_ be the same thing as if you had constructed the `GaussianPulse` with that same `fwidth`.
4. We have `from_wavelength` and `from_wvl_interval` - both of these methods should use _either_ `wvl` _or_ `wavelength` for consistency.
5. Currently it is possible to construct unphysical ranges, e.g., by initializing with an `fwidth` >= `freq0` - we should probably have a validator for this.

momchil-flex:
> 2. `fwidth` and `wvl_width` are really half-widths - as a user, I would expect this to be the actual (full) width. What then adds to the confusion is that [td.GaussianPulse](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.GaussianPulse.html) already has a parameter `fwidth`, where `fwidth` actually means the standard deviation of the frequency. @momchil-flex ?
> The above makes the method to_gaussian_pulse quite problematic, because if you initialize the FreqRange class with a certain fwidth, this will not be the same thing as if you had constructed the GaussianPulse with that same fwidth.

Not sure if there's a question for me specifically but these concerns make sense. Can we just use the full width here?

yaugenst-flex:
> Not sure if there's a question for me specifically but these concerns make sense. Can we just use the full width here?

It'd be better, but how do we handle pulses constructed using this not being the same as when constructed directly? I.e., `FreqRange(freq0=a, fwidth=b).to_gaussian_pulse() != td.GaussianPulse(freq0=a, fwidth=b)`?

momchil-flex:
> > Not sure if there's a question for me specifically but these concerns make sense. Can we just use the full width here?
> 
> It'd be better, but how do we handle pulses constructed using this not being the same as when constructed directly? I.e., `FreqRange(freq0=a, fwidth=b).to_gaussian_pulse() != td.GaussianPulse(freq0=a, fwidth=b)`?

So the `GaussianPulse.from_frequency_range(fmin=self.fmin, fmax=self.fmax, **kwargs)` that `to_gaussian_pulse` uses makes more sense in practice. So I would suggest just adding a `Note` in the docstring of `to_gaussian_pulse` saying that it uses this method, and if you want to define the pulse through freq0 and fwidth instead, then do it manually.

Another thought when looking at the code: I think the place of the frequency range module is not in `tidy3d/components/source`. A frequency/wavelength sampling is just as relevant, maybe even more so, for monitors. Also seems like we already have `tidy3d/component/frequencies` with some other frequency utilities, should we move it there?

tidy3d/components/source/freq_range.py
George-Guryev-flxcmp: Fixed default arguments and corresponding docstrings

tidy3d/components/freq_range.py
```diff
@@ -49,22 +49,33 @@ class FreqRange(Tidy3dBaseModel):
...
     )
 
+    @pydantic.root_validator
+    def check_half_fwidth_less_than_freq0(cls, values):
+        freq0 = values.get("freq0")
+        fwidth = values.get("fwidth")
+        if freq0 is not None and fwidth is not None:
+            if (fwidth / 2) >= freq0:
+                raise ValueError(
+                    "frequency bandwidth `fwidth` must be strictly less than 2 * `freq0`."
```
yaugenst-flex: Capitalization. Also, I'd prefer `` `2 * freq0` `` over `` 2 * `freq0` ``

yaugenst-flex (commented):
Thanks @George-Guryev-flxcmp looks great! Just a minor comment. Also, I think it might make sense to move this _into_ `tidy3d/components/frequencies.py` rather than have it as a separate file in `tidy3d/components`, since these are all frequency utilities in the end.

tidy3d/components/freq_range.py
George-Guryev-flxcmp: Thanks, @yaugenst-flex! Moved content of  freq_range.py and test_freq_range.py to frequencies.py and test_frequencies.py, respectively; addressed minor comments on capitalization and docstring

dbochkov-flexcompute (approved):

dbochkov-flexcompute:
looks good to me, just a bit more nit-picking: ideally the missing coverage should be addressed. That is, adding tests with invalid spacing parameter in `freqs()` and `wvls()`, and for duplicate parameters in `to_gaussian_pulse()`

yaugenst-flex (approved):
Thanks @George-Guryev-flxcmp looks good from my side!

PR #2663, opened 2025-07-15
feat: `RF GUI <-> python client` interoperabilty

daquinteroflex:
Not particularly pleased with this implementation, but can't think of a cleaner option whilst mantaining backwards compatiblity. We should properly design this though and looking forward to the webapi refactor one day.

There are more things that maybe would be worth changing, but unless we truly break backwards compatibility and have to update several notebooks... Maybe that is TBC in a different PR, for now this is just about standardising the simulation and data flow.

So ultimately, I think this should be close to achieving the goal of having a self-contained `TerminalComponentModelerData` as a serializable and deserializable data structure. However, again, unless we break backwards compatibility with `batch_data` on `xComponentModeler`, we cannot have fully pure functions. However, I have tried to at least "purify" the batch_data usage wherever it is called so that we can replace it in the future from `simulation.batch_data`.

Open to chatting about the exact data structures chosen for the SMatrixData types, it was mainly done by chatting with Damian and enabling backwards compatibility. It was renamed `MicrowaveXData` to reduce the scope and compatibility of these changes, really we should revisit this with a holistic view and willing to break compatibility. However, having these strong data types should in any case support the migration effort whatever we do onwards.

Considerations:
- Implemement minimal changes, whilst enabling interoperability
- Eventually enable `web.run(TerminalComponentModeler) -> TerminalComponentModelerData`, by using a `local_run.run(TerminalComponentModeler) -> TerminalComponentModelerData `

Again, I'm not super happy with it, but feel it's a bit better and can help us move forward. Open to suggestions to improve or feel free to propose changes directly.

TODO
- [ ] Add coverage tests if we're all happy with the implementation structure
- [ ] We need to debate when we merge this within the release flow, 2.9.0 or 2.10.0rc1
- [ ] Update submodules to develop
- [ ] Add depreciation warnings on TerminalComponentModeler classmethods that run internally.
- `batch_cached` is a mutation aberration that we should remove sooner than later.

## Greptile Summary

This PR implements interoperability between the RF GUI and Python client for Terminal Component Modeling through a significant architectural restructuring. The changes focus on standardizing data flow and S-matrix computations while maintaining backward compatibility with existing implementations.

Key changes include:
- New data structures (MicrowavePortSimulationData, MicrowaveSMatrixData, TerminalComponentModelerData)
- Dedicated local_run.py module for S-matrix computations
- RF-specific web API endpoints and task types
- Port impedance validation utilities
- RF license warning validators
- Version bump to 2.10.0rc1

## Confidence score: 2/5

1. This PR requires significant attention before merging but provides a foundation for future improvements
2. Score reflects:
   - Acknowledged technical debt in implementation
   - Circular dependencies in data.py that need resolution
   - Mixing of web methods in plugin data types that violates architectural principles
   - However, core RF computations and data structures are well-implemented with proper physics validation
3. Files needing attention:
   - tidy3d/plugins/smatrix/data/data.py: Resolve circular dependencies with local_run.py
   - tidy3d/plugins/smatrix/component_modelers/terminal.py: Clean up web method coupling
   - tidy3d/plugins/smatrix/local_run.py: Complete batch_data decoupling


dmarek-flex:
The failing tests is due to the random data, which was previously circumvented using monkeypatch to ignore the check.

To fix the breaking tests, you need to update [this monkeypatch part](https://github.com/flexcompute/tidy3d/blob/3effb92f46980a491b5f25b4040d873f03b43c63/tests/test_plugins/smatrix/test_terminal_component_modeler.py#L45-L49) with something that will ignore that check.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -547,113 +288,157 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
...
+            solver_version=self.solver_version,
+            simulation_type="microwave",
         )
+
+    def _construct_smatrix(self) -> TerminalPortDataArray:
+        from tidy3d.plugins.smatrix.run import _construct_smatrix
+
+        return _construct_smatrix(self)
+
+    def _internal_construct_smatrix(self, batch_data: typing.Any) -> TerminalPortDataArray:
```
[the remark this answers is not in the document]
daquinteroflex: no

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,140 @@
...
+        a_matrix.loc[indexer] = a
+        b_matrix.loc[indexer] = b
+
+    s_matrix = ab_to_s(a_matrix, b_matrix)
+    return s_matrix
+
+
+def compute_power_wave_amplitudes_at_each_port(
+    simulation, port_reference_impedances: PortDataArray, sim_data: SimulationData
+) -> tuple[PortDataArray, PortDataArray]:
```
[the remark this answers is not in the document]
daquinteroflex: Intentional to avoid circular imports for backwards compatibility

tests/test_plugins/smatrix/test_terminal_component_modeler.py
```diff
@@ -43,9 +44,9 @@ def run_component_modeler(monkeypatch, modeler: TerminalComponentModeler):
         lambda matrix: 1.0 / (2.0 * np.sqrt(np.abs(matrix) + 1e-4)),
     )
     monkeypatch.setattr(
-        TerminalComponentModeler,
+        td.plugins.smatrix.utils,
         "_check_port_impedance_sign",
-        lambda self, Z_numpy: (),
+        lambda Z_numpy: np.ndarray([]),
```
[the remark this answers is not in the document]
daquinteroflex: No

schemas/TerminalComponentModeler.json
```diff
@@ -1,6 +1,6 @@
 {
```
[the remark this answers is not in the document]
daquinteroflex: It is not empty

tidy3d/plugins/smatrix/local_run.py
```diff
@@ -0,0 +1,188 @@
+"""
+Tool for generating an S matrix automatically from a Tidy3d simulation and lumped port definitions.
+
+One major problem that prevents a stronger use of typing in these functions is the backwards compatibility of
+the imports. Ideally, we would do
+
+run(simulation: MySimulation) -> MySimulationData
+
+but because these
+"""
```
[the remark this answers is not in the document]
daquinteroflex: Fixed

tidy3d/plugins/smatrix/local_run.py
```diff
@@ -0,0 +1,188 @@
...
+    batch_data: BatchData = None,
+    path_dir: str = DEFAULT_DATA_DIR,
+) -> TerminalComponentModelerData:
+    _ = simulation.get_path_dir(path_dir)
+
+    if batch_data is None:
+        # TODO Remove once fully decoupled simulation from batch_data
+        batch_data = simulation.batch_data
+
+    terminal_port_data = construct_smatrix(simulation=simulation)
```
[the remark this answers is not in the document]
daquinteroflex: Fixed

tests/test_plugins/smatrix/test_component_modeler.py
```diff
@@ -408,7 +408,8 @@ def test_to_from_file_batch(tmp_path, monkeypatch):
     modeler.to_file(fname)
     modeler2 = modeler.from_file(fname)
 
-    assert modeler2.batch_cached == modeler2.batch == batch
+    # BREAK this test because it introduces mutability which shouldn't exist
+    # assert modeler2.batch_cached == modeler2.batch == batch
```
yaugenst-flex: But that seems to be a problem with the implementation and not with the test...?

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -168,8 +174,10 @@ def to_file(self, fname: str) -> None:
         super(AbstractComponentModeler, self).to_file(fname=fname)  # noqa: UP008
```
yaugenst-flex: The `noqa` here should go, as should the parameters to `super()`

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -99,10 +104,11 @@ class AbstractComponentModeler(ABC, Tidy3dBaseModel):
         "If not supplied, uses default for the current front end version.",
     )
 
-    batch_cached: Batch = pd.Field(
+    # TODO properly refactor, plugins should not have web methods.
+    batch_cached: typing.Any = pd.Field(
```
yaugenst-flex: What's the `Any` here for?

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -168,8 +174,10 @@ def to_file(self, fname: str) -> None:
         super(AbstractComponentModeler, self).to_file(fname=fname)  # noqa: UP008
 
     @cached_property
-    def batch(self) -> Batch:
+    def batch(self):
```
yaugenst-flex: Any reason here and below for removing the return type hints?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -280,64 +249,139 @@ def _check_grid_size_at_wave_ports(simulation: Simulation, ports: list[WavePort]
...
+    @cached_property
+    def _wave_ports(self) -> list[WavePort]:
+        """A list of all wave ports in the ``TerminalComponentModeler``"""
+        return [port for port in self.ports if isinstance(port, WavePort)]
+
+    @staticmethod
+    def _set_port_data_array_attributes(data_array: PortDataArray) -> PortDataArray:
+        """Helper to set additional metadata for ``PortDataArray``."""
+        data_array.name = "Z0"
+        return data_array.assign_attrs(units=OHM, long_name="characteristic impedance")
```
yaugenst-flex: Ooof.. is this really necessary? Why does the array need to be mutated?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -280,64 +249,139 @@ def _check_grid_size_at_wave_ports(simulation: Simulation, ports: list[WavePort]
...
+        raise Tidy3dKeyError(f"No radiation monitor named '{monitor_name}'.")
+
+    ##### Backwards compatibility methods, should be external functions really#####
+    def run(self, path_dir: str = DEFAULT_DATA_DIR):
+        """Solves for the scattering matrix of the system."""
+        from tidy3d.plugins.smatrix.local_run import run
+
+        terminal_component_modeler_data = run(simulation=self, path_dir=path_dir)
+        return terminal_component_modeler_data
+
```
yaugenst-flex: It looks to me like all methods from this point onward are postprocessing? Shouldn't they be in the data instead?

tidy3d/plugins/smatrix/data/data.py
```diff
@@ -0,0 +1,224 @@
...
+        title="TerminalComponentModeler",
+        description="The original :class:`TerminalComponentModeler` object that defines the simulation setup "
+        "and from which this data was generated.",
+    )
+
+    smatrix_data: MicrowaveSMatrixData = pd.Field(
+        ...,
+        title="S-Matrix Data",
+        description="Stores the computed S-matrix and reference impedances for the terminal ports",
+    )
```
yaugenst-flex: Isn't the smatrix calculated based on `port_simulation_data`? If so, isn't it redundant to include it as a field? Instead, it should probably be a cached property?

tidy3d/plugins/smatrix/local_run.py
```diff
@@ -0,0 +1,256 @@
...
+)
+from tidy3d.plugins.smatrix.data.data_array import PortDataArray, TerminalPortDataArray
+from tidy3d.plugins.smatrix.ports.wave import WavePort
+from tidy3d.plugins.smatrix.utils import check_port_impedance_sign, compute_F, compute_port_VI
+from tidy3d.web import BatchData
+
+
+def run(
+    simulation: TerminalComponentModeler,
+    batch_data: BatchData = None,
```
yaugenst-flex: It's preeetty weird that you can pass a completed run into `run`...
It seems like there should be two separate functions, one that runs the simulations, and one that does the postprocessing.

tidy3d/plugins/smatrix/local_run.py
```diff
@@ -0,0 +1,256 @@
...
+        port_i: batch_data[simulation._task_name(port=port_i)] for port_i in simulation.ports
+    }
+    port_simulation_data = MicrowavePortSimulationData(data=port_to_sim_data_map)
+    smatrix_data = MicrowaveSMatrixData(data=terminal_port_data)
+    return TerminalComponentModelerData(
+        simulation=simulation, smatrix_data=smatrix_data, port_simulation_data=port_simulation_data
+    )
+
+
+def construct_smatrix(
```
yaugenst-flex: This seems like it should be a method of `ComponentModelerData`. Actually this is my impression for most of the functions in here.

tidy3d/plugins/smatrix/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        z_vals = np.matmul(AbstractComponentModeler.inv(LHS), RHS)
+    else:
+        # Simpler case when all port impedances are the same
+        z_vals = np.matmul(AbstractComponentModeler.inv(eye - s_vals), (eye + s_vals)) * reference
+
+    z_matrix.data = z_vals
+    return z_matrix
+
+
+def check_port_impedance_sign(Z_numpy: np.ndarray):
```
yaugenst-flex: This seems like it should be a validator somewhere.

tidy3d/web/core/task_core.py
```diff
@@ -37,15 +37,15 @@ class Folder(Tidy3DResource, Queryable, extra=Extra.allow):
     )
 
     @classmethod
-    def list(cls) -> []:
+    def list(cls, projects_endpoint: str = "tidy3d/projects") -> []:
```
yaugenst-flex, line 40: Return type hint should be `list`, not `[]`

yaugenst-flex (commented):
Thanks @daquinteroflex! Left a couple of comments but overall I would say if it works we'll go with it. Depends a bit on how many breaking API changes we can introduce here. In general I would probably try to put many more methods into the data model. In some sense these are pretty similar to the monitor data in components, which also do a lot of domain-specific postprocessing.

tidy3d/plugins/smatrix/component_modelers/base.py
daquinteroflex: So it depends whether we want to eventually support `web.run(TerminalComponentModeler)` or not, because if we have web methods within the classes then it causes a circular dependency in the web subproject. Re @dbochkov

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: This is from my original implementation. My reasoning was that there are many types of data that can be represented by the same `DataArray` types, since they have the same dimensions or coordinates.

Is it better to create a new `DataArray` type just for impedance values, or change this method to copy and modify the input?

tidy3d/plugins/smatrix/component_modelers/base.py
yaugenst-flex: But the type hint here is just a forward reference? Probably have to call a `model.update_forward_refs()` but otherwise should be ok (maybe not pretty)?

tidy3d/plugins/smatrix/component_modelers/terminal.py
yaugenst-flex: Yeah. In principle you're right of course but this is exactly the reason why we have so many different array types. We'll have to rethink that at some point but I think for now it'd be better to just introduce a separate type.

tidy3d/plugins/smatrix/component_modelers/base.py
daquinteroflex: Sure yeah we can try that

tidy3d/plugins/smatrix/component_modelers/terminal.py
daquinteroflex: So yeah, I agree. So this is mainly to support backwards compatibility that I've re-added them here unless we want to break the old `TerminalComponentModeler` methods now

tidy3d/plugins/smatrix/data/data.py
daquinteroflex: Yeah that's a better idea sure

tidy3d/plugins/smatrix/local_run.py
daquinteroflex: I agree, but the problem is the `self.batch_data` existing implemntation in ComponentModeler that runs the simulations, so unless we want to break that now this was in order to mantain compatibility broadly

tidy3d/plugins/smatrix/local_run.py
daquinteroflex: So I was thinking it'd be good if we can separate the "data construction" methods from the data containers and the simualtion definitions. ie. it'd be nice if they're self contained functions that just take in `local.run(ComponentModeler) -> ComponentModelerData` to avoid the mixing up of stuff in the original implementation. I've moved them to a separate file because this way we can reuse them for backwards and onwards compatibilty without having to recreate a `TerminalComponentModelerData` class inside `TerminalComponentModeler` to use this method directly for now.

tidy3d/plugins/smatrix/utils.py
daquinteroflex: Agreed, the problem is backwards compatibility on how this method gets used dynamically in construct_smatrix

tidy3d/web/core/task_core.py
daquinteroflex: So the endpoints code itself has not been finished because it depends on whatver implementation we choose to do, so should have clarfied this is still WIP or can remove it for now

tidy3d/plugins/smatrix/component_modelers/base.py
daquinteroflex: So I think a lot of this stuff is from the privous implementation

tests/test_plugins/smatrix/test_component_modeler.py
daquinteroflex: Sure maybe it's probably the type hint that can fix it when it rehydrates

tidy3d/plugins/smatrix/component_modelers/base.py
daquinteroflex: So was the circular dependency, but maybe the model.update_forward_refs will sort it

tests/test_plugins/smatrix/test_component_modeler.py
daquinteroflex: Renabled and passing

tidy3d/plugins/smatrix/component_modelers/base.py
daquinteroflex: Still an issue in https://github.com/flexcompute/tidy3d/pull/2663/commits/125237b052c1bd48afb668fc512de289a72433c4, but this needs to be refactored for the endpoint implementation anyway

tidy3d/web/api/container.py
```diff
@@ -673,7 +687,14 @@ def num_jobs(self) -> int:
 
     def upload(self) -> None:
         """Upload a series of tasks associated with this ``Batch`` using multi-threading."""
-        self._check_folder(self.folder_name)
+        if self.simulation_type == "microwave":
+            self._check_folder(
+                self.folder_name,
+                projects_endpoint="tidy3d/projects/rf",
+                project_endpoint="tidy3d/project",  # TODO update
```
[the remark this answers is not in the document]
shanguomagi: I think we no longer need this API now: "tidy3d/projects/rf".

tidy3d/plugins/smatrix/data/data.py
dmarek-flex, line 1: rename `modeler_data.py`?

tidy3d/plugins/smatrix/data/data.py
```diff
@@ -0,0 +1,227 @@
...
+        title="Solver Log",
+        description="A string containing the log information from the simulation run.",
+    )
+
+    @cached_property
+    def smatrix(self) -> MicrowaveSMatrixData:
+        "Stores the computed S-matrix and reference impedances for the terminal ports"
+        from tidy3d.plugins.smatrix.local_run import construct_smatrix
+
+        terminal_port_data = construct_smatrix(simulation=self.simulation)
```
dmarek-flex: Think it would be better to directly use `self.data` now

tidy3d/plugins/smatrix/utils.py
dmarek-flex: When a `WavePort` is present, the port impedance is calculated in a post-processing step. Does it still make sense to make this a validator?

tidy3d/plugins/smatrix/data/data.py
dmarek-flex: But I also see the problem, you would need to add a method for going from `dict[TerminalPortType, SimulationData]` to `dict[str, SimulationData]` which would be compatible to how the batch_data is used.

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -35,6 +35,11 @@
 class AbstractComponentModeler(ABC, Tidy3dBaseModel):
     """Tool for modeling devices and computing port parameters."""
 
+    name: str = pd.Field(
```
dmarek-flex, line 30: Just needs update for title and description

tidy3d/plugins/smatrix/data/data.py
```diff
@@ -0,0 +1,227 @@
...
+class MicrowaveSMatrixData(Tidy3dBaseModel):
+    """Stores the computed S-matrix and reference impedances for the terminal ports."""
+
+    port_reference_impedances: Optional[PortDataArray] = pd.Field(
+        None,
+        title="Port Reference Impedances",
+        description="Reference impedance for each port used in the S-parameter calculation. "
+        "This is optional and may not be present if not specified or computed.",
+    )
+    data: TerminalPortDataArray = pd.Field(
```
dmarek-flex: rename to `s`

tidy3d/plugins/smatrix/data/data.py
```diff
@@ -0,0 +1,227 @@
...
+from tidy3d.log import log
+from tidy3d.plugins.smatrix.component_modelers.base import TerminalPortType
+from tidy3d.plugins.smatrix.component_modelers.terminal import TerminalComponentModeler
+from tidy3d.plugins.smatrix.data.data_array import PortDataArray, TerminalPortDataArray
+
+
+class MicrowavePortSimulationData(Tidy3dBaseModel):
+    """Stores raw simulation data from each microwave port-specific simulation."""
+
+    data: dict[TerminalPortType, SimulationData] = pd.Field(
```
dmarek-flex: I think this should change to `dict[str, SimulationData]` for two reasons:

1. One to make it easier to take the place of `batch_data` in some of these helper functions for postprocessing. Since I think in every post processing function the `batch_data` is mainly used as a mapping from a task name to the simulation data.
2. More importantly, a single `Port` may be associated with more than one `SimulationData`, like when we support multimodal `WavePort`  where each mode will correspond with a different simulation.

tidy3d/plugins/smatrix/local_run.py
dmarek-flex: Also would have expected most of these methods to be part of `ComponentModelerData`, since that is how these things are mostly organized for `MonitorData`. I don't have a strong preference either way. But shouldn't you also move `get_antenna_metrics_data` and `_monitor_data_at_port_amplitude` to this file as well then?

tidy3d/plugins/smatrix/data/data.py
```diff
@@ -0,0 +1,227 @@
...
+class TerminalComponentModelerData(Tidy3dBaseModel):
+    """
+    Data associated with a :class:`TerminalComponentModeler` simulation run.
+
+    This class serves as a data container for the results of a component modeler simulation,
+    holding the original simulation definition, the resulting S-matrix or raw port data,
+    and the solver log.
+    """
+
+    simulation: TerminalComponentModeler = pd.Field(
```
dmarek-flex: I would lean to calling this field `modeler` or something else, since it represents a set of simulations that will be run.

dmarek-flex (commented):
Just made a first pass, and the basic idea looks good to me. My main uncertainty is the the usage of the post-processing functionality present in `local_run.py`. Ignoring backwards compatibility, can you explain why that would be better than having them as members in the TerminalComponentModelerData?

tidy3d/plugins/smatrix/local_run.py
weiliangjin2021: Do we have a general guideline on those separations that are (or will be) applied to the entire tidy3d?

tidy3d/plugins/smatrix/data/data.py
daquinteroflex: So how does the current implementation of sound like? Achieves that goal

```
PortReferenceType = Union[str, TerminalPortType]  # TODO Debate this

class PortSimulationData(Tidy3dBaseModel):  # TODO Debate typing.TypedDict instead, I prefer
    """Stores raw simulation data from each microwave port-specific simulation."""

    data: dict[PortReferenceType, SimulationData] = pd.Field(
```

tidy3d/plugins/smatrix/data/data.py
daquinteroflex: Yep done that

tidy3d/web/api/container.py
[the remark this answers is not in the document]
daquinteroflex: Yes you are right

tidy3d/plugins/smatrix/data/data.py
daquinteroflex: So aiming for somthing lke
```
sim_data = modeler_data.data[port_in]
```

and/or

```
def compose_terminal_modeler_data(
    modeler: TerminalComponentModeler,
    batch_data: BatchData = None,
) -> TerminalComponentModelerData:
    port_to_sim_data_map = {
        port_i: batch_data[modeler.get_task_name(port=port_i)] for port_i in modeler.ports
    }
    port_simulation_data = PortSimulationData(data=port_to_sim_data_map)
    return TerminalComponentModelerData(modeler=modeler, data=port_simulation_data)
```

tidy3d/plugins/smatrix/local_run.py
daquinteroflex: Will reply privately to both this and this, as related to future plans as far as I understand them

> Also would have expected most of these methods to be part of ComponentModelerData, since that is how these things are mostly organized for MonitorData. I don't have a strong preference either way. But shouldn't you also move get_antenna_metrics_data and _monitor_data_at_port_amplitude to this file as well then?

daquinteroflex:
Closing this PR as it was the PoC for https://github.com/flexcompute/tidy3d/pull/2738

tidy3d/plugins/smatrix/data/data.py
daquinteroflex: So we changed this to `SimulationMap` and `SimulationDataMap` in https://github.com/flexcompute/tidy3d/blob/dario/rf_endpoints_update/tidy3d/components/index.py and https://github.com/flexcompute/tidy3d/blob/dario/rf_endpoints_update/tidy3d/components/data/index.py

tidy3d/plugins/smatrix/component_modelers/base.py
daquinteroflex: Fixed in https://github.com/flexcompute/tidy3d/pull/2738

PR #2666, opened 2025-07-16
Add check for conduction

marc-flex:
## Greptile Summary

This PR adds critical validation checks to the HeatChargeSimulation class to ensure that conduction simulations are properly configured with appropriate mediums. The changes enforce two key requirements:

1. All structures in conduction simulations must use `MultiPhysicsMedium` instead of basic `Medium`
2. At least one structure must have a `ChargeConductorMedium` for conduction simulations

These validations are implemented through new error checks in the simulation class and are thoroughly tested with corresponding test cases. The changes help catch configuration errors early in the development process, preventing runtime failures and providing clear error messages to users about missing or incorrect medium types.

## Confidence score: 5/5

1. This PR is extremely safe to merge as it adds defensive validation without modifying core behavior
2. The changes are well-tested with explicit test cases covering both validation scenarios
3. Key files to review:
   - tidy3d/components/tcad/simulation/heat_charge.py
   - tests/test_components/test_heat_charge.py

<sub>2 files reviewed, no comments</sub>
<sub>[Edit PR Review Bot Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=tidy3d_2666)</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ModeSimulation.plot()` method that plots the mode simulation plane by default, or the containing FDTD simulation if any of ``x``, ``y``, or ``z`` is passed. 
 - Enable singularity correction at PEC and lossy metal edges.
 - New `VolumeMesher` simulation type and associated `VolumeMeshMonitor` and `VolumeMesherData`, which can be used to run the unstructured meshing for a `HeatChargeSimulation` separately before running the solver.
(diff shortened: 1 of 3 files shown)


momchil-flex (commented):
What would happen before this validator if a `Medium` was defined? The `conductivity` would still be used?

This may break some user's simulations - probably for good measure, and we should do it, but let's add a CHANGELOG item that clearly explains the change in behavior.

marc-flex:
> What would happen before this validator if a Medium was defined? The conductivity would still be used?

So currently to determine whether a simulation is a Conduction simulation we check that at least one structure has the field `charge` defined. So you could define a Conduction simulation by defining an Insulator while having all the rest defined with `Medium`. The issue is that in the back end we rely on the `conductivity` defined in conducting structures which left some users wondering why the simulation was failing.

momchil-flex (approved):

PR #2680, opened 2025-07-17
fix[adjoint]: fix diffraction data adjoint source creation

groberts-flex:
Fixes the adjoint sources created for DiffractionData.

The 1j comes from that no longer being in the `get_amplitude` function and the other corrections are for when the monitor has a "-" normal direction. This, in combination with Momchil's recently merged compute PR fixes all the cases I've been able to test so far.

## Greptile Summary

This PR fixes two critical issues in the adjoint source creation for DiffractionData calculations:

1. Restores a missing complex factor (`1j`) that was previously handled in the `get_amplitude` function
2. Adds proper handling for monitors with negative normal direction by introducing a `normal_factor` correction

The changes primarily affect gradient calculations in optimization workflows that use DiffractionData. The fix ensures correct phase relationships and proper handling of monitor orientations, which is essential for accurate gradient calculations.

The PR includes comprehensive test coverage across different configurations (transmission/reflection modes, various orders and polarizations) to validate the fixes. The tests compare adjoint gradients with finite difference gradients, ensuring they match within an acceptable error threshold.

## Confidence score: 4/5
1. This PR is safe to merge as it fixes incorrect behavior and includes thorough testing
2. High confidence due to comprehensive test coverage and clear understanding of the physics involved
3. Files needing attention:
   - `tidy3d/components/data/monitor_data.py`: Review the adjoint source generation logic
   - `tests/test_components/test_autograd_periodic_numerical.py`: Verify test coverage is complete

<sub>5 files reviewed, 2 comments</sub>
<sub>[Edit PR Review Bot Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=tidy3d_2680)</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Validate mode solver object for large number of grid points on the modal plane.
 
+### Fixed
+- Fixed missing amplitude factor and handling of negative normal direction case when making adjoint sources from `DiffractionMonitor`.
+
 ## [2.9.0] - 2025-08-04
 
 ### Added
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -3448,7 +3448,13 @@ class DiffractionData(AbstractFieldProjectionData):
         k0 = 2 * np.pi * freq0 / C_0
         bck_eps = self.medium.eps_model(freq0)
         grad_const = 0.5 * k0 / np.sqrt(bck_eps) * np.cos(angle_theta)
-        src_amp = 1j * grad_const * amp_complex
+
+        normal_factor = 1.0 if (self.monitor.normal_dir == "+") else -1.0
+        src_amp = 1j * grad_const * amp_complex * normal_factor
+        # the angular direction for sources and monitors when the normal is "-"
+        # differs by a sign, so we need to flip the angle here when the normal
+        # is "-"
+        src_angle_theta = normal_factor * angle_theta
 
         # construct plane wave source
         adj_src = PlaneWave(
@@ -3461,7 +3467,7 @@ class DiffractionData(AbstractFieldProjectionData):
                 fwidth=fwidth,
             ),
             direction=self.flip_direction(monitor.normal_dir),
-            angle_theta=angle_theta,
+            angle_theta=src_angle_theta,
             angle_phi=angle_phi,
             pol_angle=pol_angle,
         )
--- a/tests/test_components/test_autograd_mode_polyslab_numerical.py
+++ b/tests/test_components/test_autograd_mode_polyslab_numerical.py
@@ -457,7 +457,7 @@ def test_finite_difference_mode_data_polyslab(
     fd_mag = np.linalg.norm(fd_grad)
     adj_mag = np.linalg.norm(pattern_dot_adj_gradient)
     percentage_error = 100.0 * np.mean(
-        (fd_grad - pattern_dot_adj_gradient) / (fd_grad + np.finfo(np.float64).eps)
+        np.abs(fd_grad - pattern_dot_adj_gradient) / (np.abs(fd_grad) + np.finfo(np.float64).eps)
     )
 
     print("\n" * 3)
--- a/tests/test_components/test_autograd_periodic_numerical.py
+++ b/tests/test_components/test_autograd_periodic_numerical.py
@@ -15,12 +15,12 @@ import tidy3d.web as web
 
 PLOT_FD_ADJ_COMPARISON = False
 NUM_FINITE_DIFFERENCE = 10
-SAVE_FD_ADJ_DATA = False
+SAVE_FD_ADJ_DATA = True
 SAVE_FD_LOC = 0
 SAVE_ADJ_LOC = 1
 LOCAL_GRADIENT = False
 VERBOSE = False
-NUMERICAL_RESULTS_DATA_DIR = "./numerical_field_test/"
+NUMERICAL_RESULTS_DATA_DIR = "./numerical_periodic_test/"
 SHOW_PRINT_STATEMENTS = False
 
 RMS_THRESHOLD = 0.25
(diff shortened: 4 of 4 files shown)


groberts-flex:
Hi all, no rush on review on this one. I'll be on PTO tomorrow and next week, so I'll check back in on this when I'm back!

tidy3d/components/data/monitor_data.py
```diff
@@ -3448,7 +3448,9 @@ def adjoint_source_amp(self, amp: DataArray, fwidth: float) -> PlaneWave:
         k0 = 2 * np.pi * freq0 / C_0
         bck_eps = self.medium.eps_model(freq0)
         grad_const = 0.5 * k0 / np.sqrt(bck_eps) * np.cos(angle_theta)
-        src_amp = grad_const * amp_complex
+        normal_factor = 1.0 if (self.monitor.normal_dir == "+") else -1.0
+        src_amp = 1j * grad_const * amp_complex * normal_factor
+        src_angle_theta = normal_factor * angle_theta
```
momchil-flex, line 3457: To be honest I don't understand this at all. It seems that we flip both `theta` *and* the plane wave `direction`. But this then should mean that the injected wave is in a wrong direction. This is my understanding of it. So if the original wave is going in theta, direction +, why would the adjoint wave not be in theta, direction -? Is there something wrong sitll in our definitions?

```
                theta,
                direction +
             /
            /
--------------------------------
          /  \
         /    \
theta,          -theta,
direction -     direction -  
```

This also doesn't make sense to me w.r.t. the Bloch vector. As far as I recall (and what would make sense to me) is that in the adjoint simulation we flip the sign of the bloch vector? But this is what I get.

```
import numpy as np
import tidy3d as td

src_amp = 1.0
freq0 = 1e14
fwidth = freq0 / 10
theta = np.pi / 6
src = td.PlaneWave(
    size=[td.inf, td.inf, 0],
    source_time=td.GaussianPulse(
        amplitude=abs(src_amp),
        phase=np.angle(src_amp),
        freq0=freq0,
        fwidth=fwidth,
    ),
    direction="+",
    angle_theta=theta, 
)

bounds = td.BlochBoundary.from_source(src, domain_size=1, axis=0)
print(bounds.bloch_vec)

bounds = td.BlochBoundary.from_source(src.updated_copy(direction="-"), domain_size=1, axis=0)
print(bounds.bloch_vec)

bounds = td.BlochBoundary.from_source(src.updated_copy(angle_theta=-theta, direction="-"), domain_size=1, axis=0)
print(bounds.bloch_vec)
```

```
0.166782047599076
-0.16678204759907594
0.166782047599076
```

Are you getting warnings that the Bloch vector is not correctly defined? If not, then I'm probably just missing something, it's just that here in the code it looks like both `theta` and `direction` are flipped, and I'd expect only one of them to be.

CHANGELOG.md
```diff
@@ -47,6 +50,7 @@ with fewer layers than recommended.
 - `EMESimulationData.port_modes_tuple` is now symmetry-expanded.
 - Fixed `Medium2D` validation error message when invalid data is passed to `ss`.
 - The phase of the amplitudes of a `DiffractionMonitor` was correctly centered such that the origin is at the monitor center.
+- Fixed missing amplitude factor and handling of negative normal direction case when making adjoint sources from `DiffractionMonitor`.
```
momchil-flex: remove this duplciate

tidy3d/components/data/monitor_data.py
momchil-flex: Ok, actually I see what's going on. What I wrote about the source is correct, but the point comes from the monitor. In the monitor definitions, we have this:

```
            theta
             /
            /
--------------------------
          /  \
         /    \
   -theta    theta
```

It is actually not clear whether we should change this definition or it has its own reasons for being like this. I vaguely feel like we had discussed this initially for the monitor. The thing is, unlike the `PlaneWave` source, the angle is not defined in the monitor directly. Instead the angle is determined by the simulation domain size and boundary conditions (bloch vector). So currently, a DiffractionMonitor placed in a given simulation would return the same "theta" regardless of whether the normal is "+" or "-", as in the picture above (because the -theta direction would require a flip of the Bloch vector). So in some situations it might also be confusing if we flipped things and the monitor returns theta or -theta depending on its normal. Well, I guess the worst is that we might not have this documented, regardless of the choice. @alec-flexcompute I know you were working on extending the DiffractionMonitor documentation at some point, are you aware of anything about this?

alec-flexcompute: The explanation article I worked on more about the calculation that was happening, so a monitor returning a theta for adjoint was not included. However, this Tuesday I noticed that MC still hadn't posted it in our learning center since they're super swamped, so I can add it to the documentation and ask them to publish the updated version

momchil-flex: Sorry, what I'm referring to us having to explain is not something to do with adjoint. It's just about the difference in angle definitions one might find between a PlaneWave source (direction +/-) and a DiffractionMonitor (normal_dir +/-)

For a source:
```
                theta,
                direction +
             /
            /
--------------------------------
          /  \
         /    \
theta,          -theta,
direction -     direction -  
```

For a monitor:
```
            theta, normal_dir +
             /
            /
--------------------------
          /  \
         /    \
   -theta    theta
normal_dir -  normal_dir -
```

I kind of still feel like we might want to switch the monitor one but it's kind of a breaking change...

alec-flexcompute: I see, I thought that the angle would only need to be referenced in adjoint here, but yeah it's a good note to have regardless

groberts-flex:
Wanted to check back in on this PR now that the release is done. How do we want to proceed with the angle definition discrepancy in monitors and sources @momchil-flex @yaugenst-flex? The adjoint procedure here to flip the angle for the adjoint source is hidden from the user so it shouldn't create confusion for them provided they know how the monitor angles are defined for different normals. 

Maybe for now, we can keep it as so and then make a decision if we want to flip the monitor definition? If we do decide to do that at some point, is is a really quick fix to change the adjoint source angle accordingly.

momchil-flex:
> p it as so and then make a decision if we want to flip the monitor definition? If we do decide to do that at some point, is is a really quick fix to change the adjoint source angle accordingly.

Yeah, I don't think we should change the monitor definition for now - we should just be aware of this.

Sorry, @groberts-flex I lost track of this during vacation so I'm confused. My impression was that we wanted to have the Diffraction adjoint fix in 2.9. Does the fact that this PR is still open mean that it's still broken in 2.9.0? It's not the worst thing to happen but yeah let's get the fix in quickly.

momchil-flex (approved):

momchil-flex:
One reason I also think it's OK not to switch the monitor definition is that the DiffractionData has an .angles property that tells you exactly what angles are associated to the amplitudes, so when using the monitor you don't really have to guess or understand the definition.

yaugenst (approved):
Thanks @groberts-flex LGTM!

yaugenst-flex (approved):

groberts-flex:
@momchil-flex - sorry this one didn't get in for 2.9! It does mean that the adjoint is not working for diffraction data when the monitor normal is "-", but if I'm not mistaken then the other cases should work with the backend changes you made. Still, I think it would be good to try and get this out as soon as possible!

PR #2681, opened 2025-07-17
fix[adjoint]: update polyslab in simulation bounds check to allow for…

groberts-flex:
… slabs of zero thickness

### DIFF
--- a/tidy3d/components/geometry/polyslab.py
+++ b/tidy3d/components/geometry/polyslab.py
@@ -1449,11 +1449,13 @@ class PolySlab(base.Planar):
 
         sim_min, sim_max = map(np.asarray, derivative_info.bounds_intersect)
         extents = sim_max - sim_min
-        is_2d = np.isclose(extents[self.axis], 0.0)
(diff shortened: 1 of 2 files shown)


tidy3d/components/geometry/polyslab.py
```diff
@@ -1449,11 +1449,13 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
 
         sim_min, sim_max = map(np.asarray, derivative_info.bounds_intersect)
         extents = sim_max - sim_min
-        is_2d = np.isclose(extents[self.axis], 0.0)
+        polyslab_2d = np.isclose(np.diff(self.slab_bounds), 0.0)
+        sim_2d = np.isclose(extents[self.axis], 0.0)
+        is_2d = polyslab_2d and sim_2d
```
groberts-flex, line 1454: this is failing a unit test - actually I think maybe we don't want to add the polyslab_2d check in here for cases like the unit test that's failing where the simulation is 2d but the polyslab isn't. in that case, the polyslab should still be integrated as if it was 2d.

PR #2683, opened 2025-07-18
Add support for EME monitor data in outer_dot

caseyflex:
Previously, to take `outer_dot` with EME data, you had to select a single `eme_port_index`, `eme_cell_index`, `sweep_index`, etc. This had to be done one field component at a time and was a pain. This PR makes `outer_dot` work directly with these additional fields, ensuring the input fields are aligned and broadcast appropriately. The additional fields end up as coords in the overlap DataArray.

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -439,13 +439,26 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
         }
 
     @property
+    def _normal_axis(self) -> int:
+        """For a 2D monitor data, return the normal axis.
+        For an EMEModeSolverMonitor, return the propagation axis."""
+        # special treatment for EMEModeSolverMonitor
+        test_field = list(self.field_components.values())[0]
+        if "eme_cell_index" in test_field.coords:
+            for axis, dim in enumerate(list("xyz")):
+                if dim not in test_field.coords:
+                    return axis
+        elif len(self.monitor.zero_dims) != 1:
+            raise DataError("Data must be 2D to get tangential dimensions.")
+        return self.monitor.zero_dims[0]
+
+    @property
     def _tangential_dims(self) -> list[str]:
         """For a 2D monitor data, return the names of the tangential dimensions. Raise if cannot
         confirm that the associated monitor is 2D."""
-        if len(self.monitor.zero_dims) != 1:
-            raise DataError("Data must be 2D to get tangential dimensions.")
+
         tangential_dims = ["x", "y", "z"]
-        tangential_dims.pop(self.monitor.zero_dims[0])
+        tangential_dims.pop(self._normal_axis)
 
         return tangential_dims
 
@@ -514,7 +527,7 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
         coords = [bs.copy() for bs in self._plane_grid_centers]
 
         # Append the first and last boundary
-        _, plane_inds = self.monitor.pop_axis([0, 1, 2], self.monitor.size.index(0.0))
+        _, plane_inds = self.monitor.pop_axis([0, 1, 2], self._normal_axis)
         coords[0] = np.array([bounds[0][0], *coords[0].tolist(), bounds[0][-1]])
         coords[1] = np.array([bounds[1][0], *coords[1].tolist(), bounds[1][-1]])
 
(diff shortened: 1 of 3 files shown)


tidy3d/components/data/monitor_data.py
```diff
@@ -942,6 +954,49 @@ def fn(fields_1, fields_2):
...
+                    mode_index_dim2 = "mode_index_1"
+                else:
+                    dims1.append(dim)
+            dims2 = list(dims1)
+            dims1.append(mode_index_dim1)
+            dims2.append(mode_index_dim2)
+            field_1 = field_1.transpose(*dims1)
+            field_2 = field_2.transpose(*dims2)
+            fields_1[key] = field_1
+            fields_2[key] = field_2
```
yaugenst-flex, line 994: Is it intentional that this function mutates the input? That seems potentially unsafe..

tidy3d/components/data/monitor_data.py
```diff
@@ -942,6 +954,49 @@ def fn(fields_1, fields_2):
...
+                elif dim == "mode_index_0":
+                    mode_index_dim1 = "mode_index_0"
+                    mode_index_dim2 = "mode_index_1"
+                else:
+                    dims1.append(dim)
+            dims2 = list(dims1)
+            dims1.append(mode_index_dim1)
+            dims2.append(mode_index_dim2)
+            field_1 = field_1.transpose(*dims1)
+            field_2 = field_2.transpose(*dims2)
```
yaugenst-flex, line 992: This is quite tricky to follow along, but is there potentially a bug here? Say we have `field_1` with `('x', 'mode_index_0')` and `field_2` with `('x', 'mode_index_1')`. After the broadcast, both of these would be `('x', 'mode_index_0', 'mode_index_1')`. So when we now iterate here, wouldn't we get:
 - `dims1`: `['x', 'mode_index_1', 'mode_index_0']`
 - `dims2`: `['x', 'mode_index_1', 'mode_index_1']`

Which would then error for the transpose?

tidy3d/components/data/monitor_data.py
```diff
@@ -942,6 +954,49 @@ def fn(fields_1, fields_2):
 
         return result
 
+    @staticmethod
+    def _align_fields(
+        fields_1: dict[str, xr.DataArray], fields_2: dict[str, xr.DataArray], d_area: xr.DataArray
+    ) -> tuple[dict[str, xr.DataArray], dict[str, xr.DataArray], xr.DataArray]:
```
yaugenst-flex, line 960: I left two comments but overall I find this function to be pretty difficult to follow. Maybe we could look at breaking the function down into smaller helpers to clarify the main workflow and better distinguish between the 'dot' and 'outer_dot' cases?

yaugenst-flex (commented):
Thanks @caseyflex! Looks good overall but had some questions particularly about the `_align_fields` method.

tidy3d/components/data/monitor_data.py
caseyflex: thanks for checking it -- I think it's ok because `mode_index_1` is included in `exclude_dims`. I agree it's a bit tricky

tidy3d/components/data/monitor_data.py
caseyflex: I can try to clean this up, comment it, split it up.

In the end I didn't include this in `dot` but maybe I can take another look at that.

tidy3d/components/data/monitor_data.py
caseyflex: I see. It's fine with this usage but agreed not good practice unless made explicit in the function signature.

tidy3d/components/data/monitor_data.py
yaugenst-flex: Aah you're right. Thanks for pointing it out!

FilipeFcp:
Hi @caseyflex Casey,

Thanks! I took a quick look at the code. It seems fine to me, but I’m sure Yaugenst will have much better insights than I do.

I tried a simple example, using the taper from our [EME notebook](https://www.flexcompute.com/tidy3d/examples/notebooks/EMESolver/). I placed a yz [EMEFieldMonitor](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.EMEFieldMonitor.html) and tried to do a mode overlap with a Gaussian beam, in the same way as our [mode overlap example](https://www.flexcompute.com/tidy3d/examples/notebooks/ModeOverlap/).

The behavior seems the same as before: just calling eme_sim_data['field2'].outer_dot(gaussian_beam.field_data) returns the error: ValueError: operands could not be broadcast together with shapes (45,39,1,30,2) (45,39,1).

Maybe I’m doing something wrong, or this isn’t the goal of the PR and I misunderstood?

PR #2684, opened 2025-07-21
Optionally set voxel size for subpixel averaging

weiliangjin2021:
The size of the control volume for VPEP, CPEP, and volumetric averaging can be customized; default value to be determined.

## Greptile Summary

This PR enhances the subpixel averaging functionality in Tidy3D by introducing customizable control over voxel sizes. The changes introduce two key parameters:

1. `voxel_size_comp`: Controls voxel size along the component direction
2. `voxel_size_trans`: Controls voxel size along the transverse direction

Both parameters are measured in units of local grid step size and default to 1 to maintain backward compatibility.

The PR also improves code organization by introducing a new abstract base class `AbstractYeeCellDiscretizationMethod`, which better separates the concerns of Yee cell discretization from subpixel averaging methods. This architectural change provides a clearer hierarchy for different discretization methods while maintaining the existing functionality.

PR Description Notes:
- The description mentions that default values are 'to be determined', but the code shows they are set to 1

## Confidence score: 3/5

1. This PR is moderately safe to merge but requires careful testing of numerical results
2. While the code structure is solid, the impact on simulation accuracy with different voxel sizes needs validation
3. Files needing attention:
   - `tidy3d/components/subpixel_spec.py`: Verify that the default values of 1 are indeed the correct choice
   - Any test files that verify numerical accuracy of the subpixel averaging methods

### DIFF
--- a/tidy3d/components/subpixel_spec.py
+++ b/tidy3d/components/subpixel_spec.py
@@ -1,7 +1,7 @@
 # Defines specifications for subpixel averaging
 from __future__ import annotations
 
-from typing import Union
+from typing import Literal, Union
 
 import pydantic.v1 as pd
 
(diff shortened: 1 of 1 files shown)


weiliangjin2021:
it's probably too much flexibility to allow different voxel size for dielectric, metal, and lossy metal subpixel averaging. I'll revise it to restrict to only one type of voxel for the methods that involve volumetric averaging.

PR #2686, opened 2025-07-22
release: 2.9.0

daquinteroflex:


dev.Dockerfile
daquinteroflex, line 1: @frederikschubertflex @bryance-flexcompute After a story I'll tell you at some point next week, I've decided to make this to isolate testing environments further. Want to build it a lot more also to eventually use act etc. Just thought to get some feedback but works nicely for me locally

frederikschubertflex (commented):
I like having this! It would be ideal to use the same image for the CI as well. This is less critical as the system dependencies are not that important but nevertheless...

dev.Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM ghcr.io/astral-sh/uv:debian AS tidy3d-python-client-dev
```
frederikschubertflex: Maybe we should use the latest patch version `ghcr.io/astral-sh/uv:0.8-debian` as things are still changing in uv and we might want to be explicit about this.

daquinteroflex: Good idea!

tidy3d/plugins/smatrix/component_modelers/terminal.py
daquinteroflex, line 1: @weiliangjin2021 btw just in case you have any suggestions!

weiliangjin2021: nice!

dev.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    curl \
+    git \
+    xsel \
+    xclip
+
+ENV POETRY_HOME=/opt/poetry
+RUN curl -sSL https://install.python-poetry.org | python3 -
+ENV PATH="/root/.local/bin:${POETRY_HOME}/bin:${PATH}"
```
weiliangjin2021, line 15: just curious what "/root/.local/bin" is used for?

dev.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+RUN apt-get update && DEBIAN_FRONTEND="noninteractive" TZ="America/New_York" apt-get install -y curl \
+    && curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux-x86_64.tar.gz \
+    && tar -C /opt -xzf nvim-linux-x86_64.tar.gz \
+    && rm nvim-linux-x86_64.tar.gz
```
weiliangjin2021, line 20: why not just vim or emacs?

daquinteroflex: Ahh so it could be any we want to add! We can update if those are ones you want to add?

That said, normally what I do is run scripts inside the container, but edit the same mounted file directory externally using nvim, visual studio code, or whatever I want so it kind of is independent of this really

dev.Dockerfile
daquinteroflex: So I got a bit of Gemini help making the dockerfile but here's why:
> Adding /root/.local/bin to the PATH environment variable is a best practice in Python Dockerfiles. It ensures that any Python command-line tools you install later for the root user (e.g., pip install --user black) will be immediately runnable without having to specify their full path

daquinteroflex:
Actually I have to override because of the schema block, once tests pass.

Running remote tests here https://github.com/flexcompute/tidy3d/actions/runs/16717257693

PR #2692, opened 2025-07-24
fix: validate parameters for InverseDesignResult are in valid range

yaugenst-flex:
## Greptile Summary

This PR implements parameter validation for the `InverseDesignResult` class in the inverse design plugin. The change adds a pydantic validator that automatically clips parameter values to the valid [0,1] range when they fall outside these bounds.

The core modification is in `tidy3d/plugins/invdes/result.py`, where a new `_validate_and_clip_params` validator is added to the `InverseDesignResult` class. This validator processes the `params` tuple containing optimization history arrays, counts how many values are below 0 or above 1, clips them using `np.clip(param_array, 0.0, 1.0)`, and logs a warning message indicating the total number of out-of-bounds parameters that were corrected.

This validation addresses data integrity issues that can occur during inverse design optimization workflows. Parameters in topology optimization are expected to represent normalized design variables in the [0,1] range, but can sometimes fall outside this range due to numerical precision issues, optimizer behavior exploring bounds, or when loading legacy optimization result files. The validator ensures backward compatibility by automatically correcting invalid data rather than raising errors, while still alerting users to potential issues through warning messages.

The implementation includes comprehensive test coverage in `tests/test_plugins/test_invdes.py` with a new `test_result_params_out_of_bounds()` function that verifies the clipping behavior works correctly and that downstream functionality like `get_sim()` continues to operate normally after parameter correction.

## Confidence score: 4/5

- This is a safe defensive programming fix that improves data integrity without breaking existing functionality
- The automatic clipping approach with warnings is more user-friendly than throwing errors, maintaining backward compatibility
- The `tidy3d/plugins/invdes/result.py` file needs careful review to ensure the validator logic handles edge cases correctly and doesn't introduce performance issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Giving opposite boundaries different names no longer causes a symmetry validator failure.
+- Fixed issue with parameters in `InverseDesignResult` sometimes being outside of the valid parameter range.
 
 ## [2.9.0rc2] - 2025-07-17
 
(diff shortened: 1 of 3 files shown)


daquinteroflex (approved):
Thanks

PR #2693, opened 2025-07-24
fix: multifrequency adjoint performance by vectorizing over frequencies

yaugenst-flex:
## Greptile Summary

This PR implements a significant performance optimization for multi-frequency adjoint gradient computations by vectorizing operations over frequencies. The changes transform the adjoint system from processing frequencies sequentially to handling them simultaneously in batch operations.

The core architectural change involves modifying the `DerivativeInfo` class to use a `frequencies` array instead of a single `frequency` value, enabling vectorized processing. Key components updated include:

- **Derivative computation infrastructure**: Updated `derivative_utils.py` to handle 4D interpolation grids (x,y,z,f) and vectorized field projections
- **Medium handling**: Modified dispersive medium classes (`PoleResidue`, `CustomPoleResidue`) to iterate over frequencies and accumulate gradients rather than selecting specific frequencies
- **Geometry primitives**: Updated `Cylinder` class to compute discretization parameters using minimum wavelength from maximum frequency
- **Monitor data**: Added phase correction (`1j` factor) to adjoint source amplitude calculation in `DiffractionData`
- **Configuration**: Added `ADJOINT_FREQ_CHUNK_SIZE` constant for optional frequency chunking control

The implementation maintains mathematical correctness while enabling significant performance improvements for broadband optimization problems. Test infrastructure was updated to support the new vectorized API, including changes to mock objects and expected return types. The vectorization allows all frequencies to be processed in single operations rather than separate computations for each frequency point, which should provide substantial speedups for multi-frequency adjoint calculations commonly used in inverse design workflows.

## Confidence score: 4/5

- This is a well-structured performance optimization that should significantly improve multi-frequency adjoint performance
- The vectorization approach is mathematically sound and maintains API consistency
- The phase correction in monitor data and comprehensive test updates indicate thorough consideration of edge cases
- Minor concerns about the complexity of the 4D interpolation changes and potential for subtle bugs in frequency dimension handling

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -3448,7 +3448,7 @@ class DiffractionData(AbstractFieldProjectionData):
         k0 = 2 * np.pi * freq0 / C_0
         bck_eps = self.medium.eps_model(freq0)
         grad_const = 0.5 * k0 / np.sqrt(bck_eps) * np.cos(angle_theta)
-        src_amp = grad_const * amp_complex
+        src_amp = 1j * grad_const * amp_complex
 
         # construct plane wave source
         adj_src = PlaneWave(
--- a/tidy3d/components/geometry/primitives.py
+++ b/tidy3d/components/geometry/primitives.py
@@ -15,7 +15,7 @@ from tidy3d.components.autograd.constants import PTS_PER_WVL_MAT_CYLINDER_DISCRE
 from tidy3d.components.autograd.derivative_utils import DerivativeInfo
 from tidy3d.components.base import cached_property, skip_if_fields_missing
 from tidy3d.components.types import Axis, Bound, Coordinate, MatrixReal4x4, Shapely
-from tidy3d.constants import C_0, LARGE_NUMBER, MICROMETER
+from tidy3d.constants import LARGE_NUMBER, MICROMETER
 from tidy3d.exceptions import SetupError, ValidationError
 from tidy3d.packaging import verify_packages_import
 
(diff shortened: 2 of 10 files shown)


daquinteroflex (approved):
Looks well implmentd, you mentioned you tested.

tidy3d/components/data/monitor_data.py
```diff
@@ -3448,7 +3448,7 @@ def adjoint_source_amp(self, amp: DataArray, fwidth: float) -> PlaneWave:
         k0 = 2 * np.pi * freq0 / C_0
         bck_eps = self.medium.eps_model(freq0)
         grad_const = 0.5 * k0 / np.sqrt(bck_eps) * np.cos(angle_theta)
-        src_amp = grad_const * amp_complex
+        src_amp = 1j * grad_const * amp_complex
```
yaugenst-flex, line 3451: Note this is technically an unrelated change but necessary because the `1j` was removed from `get_amplitude()`. This was causing diffraction grads to be wrong in https://github.com/flexcompute/tidy3d-notebooks/blob/develop/Autograd2GradientChecking.ipynb

tidy3d/components/geometry/primitives.py
```diff
@@ -281,9 +281,9 @@ def _points_unit_circle(
     def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradFieldMap:
         """Compute the adjoint derivatives for this object."""
 
-        # compute number of points in the circumference of the polyslab using resolution info
-        wvl0 = C_0 / derivative_info.frequency
-        wvl_mat = wvl0 / max(1.0, np.sqrt(abs(derivative_info.eps_in)))
+        # compute circumference discretization
+        wvl0_min = C_0 / np.max(derivative_info.frequencies)
```
tylerflex: this comes up a few times, maybe we should make a wvl_min property in DerivativeInfo? or something?

tylerflex (approved):
OK to merge as is but I guess a few comments.
1. Wvl_min stuff seems to pop up a lot, not a big deal but could be simplified maybe.
2. We're getting to a complexity with these vips where we might want to just write a master VJP on a Triangle Mesh and then try to do something more general. I wonder how feasible this is.

tidy3d/components/geometry/primitives.py
yaugenst-flex: done!

yaugenst-flex:
> We're getting to a complexity with these vips where we might want to just write a master VJP on a Triangle Mesh and then try to do something more general. I wonder how feasible this is.

yeah we should do this, just a matter of when 😄

PR #2696, opened 2025-07-24
feat: Add natural convection BC

damiano-flex:
## Greptile Summary

This PR introduces a new `NaturalConvectionVerticalSpec` class to enhance thermal boundary condition modeling in the Tidy3D framework. The class enables automatic calculation of heat transfer coefficients for natural convection from vertical plates based on fluid properties rather than requiring manual coefficient specification.

The implementation follows standard heat transfer engineering practices by using Rayleigh number calculations and appropriate Nusselt number correlations to determine heat transfer coefficients. It supports both laminar and turbulent flow regimes based on critical Rayleigh number thresholds. The class accepts comprehensive fluid property inputs including thermal conductivity, viscosity, density, specific heat, and thermal expansion coefficient.

The integration is achieved by extending the existing `ConvectionBC.transfer_coeff` field to accept either a simple float (existing behavior) or the new `NaturalConvectionVerticalSpec` object, maintaining backward compatibility. The class is properly exported through the main package interface and integrated into the TCAD types system, following established patterns in the codebase for thermal boundary conditions.

## Confidence score: 2/5

- This PR has significant issues that could cause problems if merged, particularly around unit consistency and incomplete implementation
- The unit system mismatch between SI units and Tidy3D's micrometer-based system could lead to incorrect calculations, and the `_compute_parameters` method returns unused values suggesting incomplete integration
- Files `tidy3d/components/tcad/boundary/heat.py` needs immediate attention for unit consistency, parameter validation, and completion of the computation logic

### DIFF
--- a/tidy3d/__init__.py
+++ b/tidy3d/__init__.py
@@ -24,6 +24,7 @@ from tidy3d.components.spice.analysis.dc import (
 from tidy3d.components.spice.sources.dc import DCCurrentSource, DCVoltageSource
 from tidy3d.components.spice.sources.types import VoltageSourceType
 from tidy3d.components.tcad.analysis.heat_simulation_type import UnsteadyHeatAnalysis, UnsteadySpec
+from tidy3d.components.tcad.boundary.heat import VerticalNaturalConvectionCoeffModel
 from tidy3d.components.tcad.boundary.specification import (
     HeatBoundarySpec,
     HeatChargeBoundarySpec,
@@ -708,6 +709,7 @@ __all__ = [
     "UnsteadyHeatAnalysis",
     "UnsteadySpec",
     "Updater",
+    "VerticalNaturalConvectionCoeffModel",
     "VisualizationSpec",
     "VoltageBC",
     "VoltageSourceType",
--- a/tidy3d/components/material/tcad/heat.py
+++ b/tidy3d/components/material/tcad/heat.py
@@ -10,8 +10,11 @@ import pydantic.v1 as pd
 from tidy3d.components.base import Tidy3dBaseModel
 from tidy3d.constants import (
     DENSITY,
+    DYNAMIC_VISCOSITY,
+    SPECIFIC_HEAT,
     SPECIFIC_HEAT_CAPACITY,
     THERMAL_CONDUCTIVITY,
+    THERMAL_EXPANSIVITY,
 )
 
 
@@ -46,10 +49,104 @@ class FluidMedium(AbstractHeatMedium):
     """Fluid medium. Heat simulations will not solve for temperature
     in a structure that has a medium with this 'heat_spec'.
 
-    Example
-    -------
-    >>> solid = FluidMedium()
+    The full set of parameters is primarily intended for calculations involving natural
+    convection, where they are used to determine the heat transfer coefficient.
+    In the current version, these specific properties may not be utilized for
+    other boundary condition types.
+
+    Attributes
+    ----------
+    thermal_conductivity : float, optional
+        Thermal conductivity ($k$) of the fluid in $W/(\\mu m \\cdot K)$.
+    viscosity : float, optional
+        Dynamic viscosity ($\\mu$) of the fluid in $kg/(\\mu m \\cdot s)$.
+    specific_heat : float, optional
+        Specific heat ($c_p$) of the fluid in $\\mu m^2/(s^2 \\cdot K)$.
+    density : float, optional
+        Density ($\rho$) of the fluid in $kg/\\mu m^3$.
+    expansivity : float, optional
+        Thermal expansion coefficient ($\beta$) of the fluid in $1/K$.
+
+    Examples
+    --------
+    >>> # If you are using a boundary condition without a natural convection model,
+    >>> # the specific properties of the fluid are not required. In this common
+    >>> # scenario, you can instantiate the class without arguments.
+    >>> air = FluidMedium()
+
+    >>> # It is most convenient to define the fluid from standard SI units
+    >>> # using the `from_si_units` classmethod.
+    >>> # The following defines air at approximately 20°C.
+    >>> air_from_si = FluidMedium.from_si_units(
+    ...     thermal_conductivity=0.0257,  # Unit: W/(m*K)
+    ...     viscosity=1.81e-5,          # Unit: Pa*s
+    ...     specific_heat=1005,         # Unit: J/(kg*K)
+    ...     density=1.204,              # Unit: kg/m^3
+    ...     expansivity=1/293.15        # Unit: 1/K
+    ... )
+
+    >>> # One can also define the medium directly in Tidy3D units.
+    >>> # The following is equivalent to the example above.
+    >>> air_direct = FluidMedium(
+    ...     thermal_conductivity=2.57e-8,
+    ...     viscosity=1.81e-11,
+    ...     specific_heat=1.005e+15,
+    ...     density=1.204e-18,
+    ...     expansivity=1/293.15
+    ... )
     """
+
+    thermal_conductivity: pd.NonNegativeFloat = pd.Field(
+        default=None,
+        title="Fluid Thermal Conductivity",
+        description="Thermal conductivity (k) of the fluid.",
+        units=THERMAL_CONDUCTIVITY,
+    )
+    viscosity: pd.NonNegativeFloat = pd.Field(
+        default=None,
+        title="Fluid Dynamic Viscosity",
+        description="Dynamic viscosity (μ) of the fluid.",
+        units=DYNAMIC_VISCOSITY,
+    )
+    specific_heat: pd.NonNegativeFloat = pd.Field(
+        default=None,
+        title="Fluid Specific Heat",
+        description="Specific heat of the fluid at constant pressure.",
+        units=SPECIFIC_HEAT,
+    )
+    density: pd.NonNegativeFloat = pd.Field(
+        default=None,
+        title="Fluid Density",
+        description="Density (ρ) of the fluid.",
+        units=DENSITY,
+    )
+    expansivity: pd.NonNegativeFloat = pd.Field(
+        default=None,
+        title="Fluid Thermal Expansivity",
+        description="Thermal expansion coefficient (β) of the fluid.",
+        units=THERMAL_EXPANSIVITY,
+    )
+
+    def from_si_units(
+        thermal_conductivity: pd.NonNegativeFloat,
+        viscosity: pd.NonNegativeFloat,
+        specific_heat: pd.NonNegativeFloat,
+        density: pd.NonNegativeFloat,
+        expansivity: pd.NonNegativeFloat,
+    ):
+        thermal_conductivity_tidy = thermal_conductivity / 1e6  # W/(m*K) -> W/(um*K)
+        viscosity_tidy = viscosity / 1e6  # Pa*s -> kg/(um*s)
+        specific_heat_tidy = specific_heat * 1e12  # J/(kg*K) -> um**2/(s**2*K)
+        density_tidy = density / 1e18  # kg/m**3 -> kg/um**3
+        expansivity_tidy = expansivity  # 1/K -> 1/K (no change)
+
+        return FluidMedium(
+            thermal_conductivity=thermal_conductivity_tidy,
+            viscosity=viscosity_tidy,
+            specific_heat=specific_heat_tidy,
+            density=density_tidy,
+            expansivity=expansivity_tidy,
+        )
 
 
 class FluidSpec(FluidMedium):
--- a/tidy3d/components/tcad/boundary/heat.py
+++ b/tidy3d/components/tcad/boundary/heat.py
@@ -2,10 +2,21 @@
 
 from __future__ import annotations
 
+from typing import Union
+
 import pydantic.v1 as pd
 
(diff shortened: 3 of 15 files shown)


tidy3d/components/tcad/boundary/heat.py
```diff
@@ -40,6 +53,137 @@ class HeatFluxBC(HeatChargeBC):
     )
 
 
+class NaturalConvectionVerticalSpec(HeatChargeBC):
```
marc-flex: Given that this isn't the BC but a class that specifies the coefficients I think a name along the lines of `VerticalNaturalConvectionCoeffModel` may be more appropriate?

Also, since it isn't a BC itself, let's make it derive from `Tidy3dBaseModel` model instead.

tidy3d/__init__.py
```diff
@@ -405,6 +406,7 @@ def set_logging_level(level: str) -> None:
 Transformed.update_forward_refs()
 ClipOperation.update_forward_refs()
 GeometryGroup.update_forward_refs()
+ConvectionBC.update_forward_refs()
```
marc-flex: Is this needed?

tidy3d/components/tcad/boundary/heat.py
```diff
@@ -40,6 +53,137 @@ class HeatFluxBC(HeatChargeBC):
...
+    """
+    Specification for natural convection from a vertical plate.
+
+    This class calculates the heat transfer coefficient (h) based on fluid
+    properties and an expected temperature difference, then provides these
+    values as 'base' and 'exponent' for a generalized heat flux equation
+    q = base * (T_surf - T_fluid)^exponent.
+    """
```
marc-flex: Maybe add in the docstring that one can define the constants in SI with `from_si_units` and maybe add a couple of examples like with the rest of the BC classes: one using the standard constructor and another one with the `from_si_units` function?

marc-flex (commented):
Thanks @damiano-flex ! This is looking very good from the get-go!

Apart from the minor comments that I left we're missing unit tests. I think once you add these we'll be good to go

tidy3d/components/tcad/boundary/heat.py
```diff
@@ -40,6 +53,137 @@ class HeatFluxBC(HeatChargeBC):
...
+        units=THERMAL_EXPANSIVITY,
+    )
+    plate_L: pd.NonNegativeFloat = pd.Field(
+        title="Plate Characteristic Length",
+        description="Characteristic length (L), defined as the height of the vertical plate.",
+        units=MICROMETER,
+    )
+
+    gravity: pd.NonNegativeFloat = pd.Field(
+        default=9.80665 * 1e6,
```
dbochkov-flexcompute: does the user need to change this? maybe move it to constants? if it still might be useful to be able to change, it's still forth defining `9.80665` as a constant, and using `default=GRAV_ACC` (not sure what would be the best name)

tidy3d/components/tcad/boundary/heat.py
```diff
@@ -40,6 +53,137 @@ class HeatFluxBC(HeatChargeBC):
...
+    )
+    fluid_beta: pd.NonNegativeFloat = pd.Field(
+        title="Fluid Thermal Expansivity",
+        description="Thermal expansion coefficient (β) of the fluid.",
+        units=THERMAL_EXPANSIVITY,
+    )
+    plate_L: pd.NonNegativeFloat = pd.Field(
+        title="Plate Characteristic Length",
+        description="Characteristic length (L), defined as the height of the vertical plate.",
+        units=MICROMETER,
```
dbochkov-flexcompute: should all fluid properties be pulled automatically from medium properties that touches that boundary? Sounds like we should expand heat_spec with these additional properties. For other parameters I would use more explicit names: `plate_L` -> `plate_length`

tidy3d/components/tcad/boundary/heat.py
```diff
@@ -40,6 +53,137 @@ class HeatFluxBC(HeatChargeBC):
...
+            fluid_k=k_tidy,
+            fluid_mu=mu_tidy,
+            fluid_Cp=cp_tidy,
+            fluid_rho=rho_tidy,
+            fluid_beta=beta_tidy,
+            plate_L=L_tidy,
+            gravity=g_tidy,
+        )
+
+    def _compute_parameters(self):
```
dbochkov-flexcompute: probably don't need it in the frontend (for example, if you happen to need to fix a bug in here, you would need to change both frontend and backend, instead of just backend)

dbochkov-flexcompute (commented):
overall looks all makes sense, but I would add all required fluid parameters into `FluidMedium` (probably using more explicit names like `viscosity`, `expansivity`, etc, to be consistent with `SolidMedium`) and make this new convection bc automatically get those parameters from fluids it touches. It means that we also need to add a validator that checks that these pararmeters are defined for fluid mediums in our simulations if natural convection is used.

Additionally, I would also move `_compute_parameters` to the backend

dbochkov-flexcompute:
> I would add all required fluid parameters into FluidMedium (probably using more explicit names like viscosity, expansivity, etc, to be consistent with SolidMedium) and make this new convection bc automatically get those parameters from fluids it touches. It means that we also need to add a validator that checks that these pararmeters are defined for fluid mediums in our simulations if natural convection is used.

Some more out-loud thinking:

Probably still need to have a medium parameter in the natural convection if one wants to define it on a simulation boundary, something like
```
h_spec = NaturalConvectionVerticalSpec(medium=None, plate_length=1)  # automatically determine which mediums are touched
h_spec = NaturalConvectionVerticalSpec(medium=oil, plate_length=1)  # use parameters defined in oil
```

Might also make sense to restrict automatic `medium=None` to `StructureStructureInterface` and `MediumMediumInterface` boundary placements only so that we can easily validate that one medium is solid and the other is fluid with all necessary parameters. And always require `medium` for other boundary placements

What do you think?

tidy3d/components/tcad/boundary/heat.py
damiano-flex: No the user doesn't need to change this, unless they want to simulate a thermal simulation on the moon :smile: , which I hardly doubt it. I agree the gravity might be stored as a constant.

tidy3d/components/tcad/boundary/heat.py
damiano-flex: Agree

tidy3d/components/tcad/boundary/heat.py
damiano-flex: Yeah, probably it is better to move it in the backend.

tidy3d/__init__.py
damiano-flex: Not anymore.

tidy3d/components/tcad/boundary/heat.py
damiano-flex: Yes, it can help readability.

tidy3d/components/tcad/boundary/heat.py
damiano-flex: I made some thoughts before answer this comment. Ultimately, I agree, pulling medium properties seems indeed an appropriate choice. If one wants to add 4 natural convection walls the code would look like much more cleaner.

tidy3d/components/material/tcad/heat.py
```diff
@@ -46,11 +49,100 @@ class FluidMedium(AbstractHeatMedium):
     """Fluid medium. Heat simulations will not solve for temperature
     in a structure that has a medium with this 'heat_spec'.
 
-    Example
-    -------
-    >>> solid = FluidMedium()
```
marc-flex, line 51: As discussed in our chat, keep this in.

tidy3d/components/tcad/boundary/heat.py
```diff
@@ -40,6 +51,64 @@ class HeatFluxBC(HeatChargeBC):
...
 
+class VerticalNaturalConvectionCoeffModel(Tidy3dBaseModel):
+    """
+    Specification for natural convection from a vertical plate.
+
+    This class calculates the heat transfer coefficient (h) based on fluid
+    properties and an expected temperature difference, then provides these
+    values as 'base' and 'exponent' for a generalized heat flux equation
+    q = base * (T_surf - T_fluid)^exponent + base * (T_surf - T_fluid).
+    """
```
marc-flex, line 62: Maybe add an example of how to create this BC. It'll be good for our docs

marc-flex (approved):
This looks good to me! Thanks @damiano-flex 

@dbochkov-flexcompute do you want to have another go at it? Or is it OK to merge?

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -455,6 +457,89 @@ def check_voltage_array_if_capacitance(cls, values):
...
+
+            # Case 2: The fluid medium IS specified directly in the convection model.
+            else:
+                fluid_medium = natural_conv_model.medium
+                if not isinstance(fluid_medium, FluidMedium):
+                    raise SetupError(
+                        f"Boundary spec at index {i}: The medium '{fluid_medium.name}' specified in "
+                        f"'VerticalNaturalConvectionCoeffModel' must be a fluid, but it has a heat "
+                        f"spec of type '{type(fluid_medium).__name__}'."
+                    )
```
dbochkov-flexcompute: I think this is not needed as we have `medium: FluidMedium = pd.Field(...)`, meaning that pydantic will already ensure that medium is `FluidMedium`

tidy3d/components/tcad/boundary/heat.py
```diff
@@ -40,6 +51,64 @@ class HeatFluxBC(HeatChargeBC):
...
+
+    medium: FluidMedium = pd.Field(
+        default=None,
+        title="Interface medium",
+        description="Medium to use for heat transfer coefficient.",
```
dbochkov-flexcompute: mention in the description what `medium=None` means

dbochkov-flexcompute (changes requested):
Yeah, this looks good. Just a couple minor comments, and also it looks like it's missing frontend tests

tidy3d/components/material/tcad/heat.py
```diff
@@ -46,11 +49,105 @@ class FluidMedium(AbstractHeatMedium):
...
+    thermal_conductivity : float, optional
+        Thermal conductivity ($k$) of the fluid in $W/(\\mu m \\cdot K)$.
+    viscosity : float, optional
+        Dynamic viscosity ($\\mu$) of the fluid in $kg/(\\mu m \\cdot s)$.
+    specific_heat : float, optional
+        Specific heat ($c_p$) of the fluid in $\\mu m^2/(s^2 \\cdot K)$.
+    density : float, optional
+        Density ($\rho$) of the fluid in $kg/\\mu m^3$.
+    expansivity : float, optional
+        Thermal expansion coefficient ($\beta$) of the fluid in $1/K$.
```
marc-flex, line 68: I have always seen this as thermal expansion coefficient. `expansivity` sounds a bit artificial to me. Have you read it like this anywhere? If not, I'd be inclined to go for a name like `expansion_coeff` or something along those lines

damiano-flex: Discussed in slack :ok_hand:

dbochkov-flexcompute (approved):
Looks good! Also forgot to mention earlier: don't forget to add an entry in CHANGELOG and include in the API docs https://github.com/flexcompute/tidy3d/blob/damiano/natural_convection_BC/docs/api/heat/boundary_conditions.rst

damiano-flex:
> Looks good! Also forgot to mention earlier: don't forget to add an entry in CHANGELOG and include in the API docs https://github.com/flexcompute/tidy3d/blob/damiano/natural_convection_BC/docs/api/heat/boundary_conditions.rst

Thank you again for checking. I added a CHANGELOG entry. 

I was thinking that after I add a couple of coefficient models, like the radiation boundary conditions and the top plate natural convection, I could create a section **Coefficient models** section in [boundary_conditions.rst](https://github.com/flexcompute/tidy3d/blob/damiano/natural_convection_BC/docs/api/heat/boundary_conditions.rst) including all these models that must be passed to `ConvectionBC`.

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -9,6 +9,8 @@
 import numpy as np
 import pydantic.v1 as pd
 
+from tidy3d import FluidMedium, VerticalNaturalConvectionCoeffModel
```
daquinteroflex, line 12: @damiano-flex  @marc-flex Just be careful of importing this way _inside_ tidy3d because it can create circular dependencies in the `__init__.py` during initialization. They normally appear after someone extends a file and triggers this. https://docs.python.org/3/tutorial/modules.html#intra-package-references

daquinteroflex:
```
from tidy3d.components.material.tcad.heat import FluidMedium
from tidy3d.components.tcad.boundary.heat import VerticalNaturalConvectionCoeffModel
```

is what is should have been I think

marc-flex: Very true! Thanks for pointing that out, @daquinteroflex  . Would you mind opening a PR with these changes?

daquinteroflex: So the big RF PR will fix it but feel free if you can fix sooner

PR #2697, opened 2025-07-24
Fix for single frequency FieldMonitor

FilipeFcp:
I just added an `if` condition to not interpolate the coordinates if it is a single frequency monitor to avoid interpolation data. 

This issue was reported by a user today.

## Greptile Summary

This PR fixes a bug in the ZBF export functionality for single-frequency FieldMonitor data. The issue occurs in the `to_zbf_file` method within `monitor_data.py` where xarray's `interp()` method produces NaN values when interpolating single-frequency datasets.

The change adds a conditional check `if len(e_x.f) > 1:` before interpolating the E-field components (`e_x` and `e_y`) to a specific frequency. This prevents unnecessary interpolation when only one frequency point exists, since the data already exists at the target frequency. The fix specifically addresses lines 1169-1171 where E-field interpolation occurs.

This change integrates with the existing ZBF export pipeline, which is used to convert Tidy3D field monitor data into the ZBF (Zemax Beam File) format for interoperability with Zemax optical design software. The ZBF export functionality is part of the monitor data processing workflow that handles various electromagnetic field data types.

## Confidence score: 4/5

• This is a safe, targeted bugfix that resolves a specific interpolation issue without affecting other functionality
• The fix follows sound logic by avoiding unnecessary operations that introduce numerical artifacts, though there's a potential inconsistency with `mode_area` interpolation handling
• The `mode_area` interpolation on line 1166 should be reviewed for consistency, as it lacks the same conditional check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Polygon vertices cleanup in `ClipOperation.intersections_plane`.
 - Removed sources from `sim_inf_structure` simulation object in `postprocess_adj` to avoid source and background medium validation errors.
 - Revert overly restrictive validation of `freqs` in the `ComponentModeler` and `TerminalComponentModeler`.
+- Fixed `ElectromagneticFieldData.to_zbf()` to support single frequency monitors and apply the correct flattening order.
 
 ## [2.9.0] - 2025-08-04
 
--- a/tests/test_data/test_monitor_data.py
+++ b/tests/test_data/test_monitor_data.py
@@ -907,6 +907,17 @@ class TestZBF:
         return self.simdata(monitor)["fields"]
 
     @pytest.fixture(scope="class")
+    def field_data_single_frequency(self) -> td.FieldData:
+        """Make random field data with single frequency from an emulated simulation run."""
+        monitor = td.FieldMonitor(
+            size=(td.inf, td.inf, 0),
+            freqs=self.freqs[0],
+            name="fields",
+            colocate=True,
+        )
+        return self.simdata(monitor)["fields"]
+
+    @pytest.fixture(scope="class")
     def mode_data(self) -> td.ModeData:
         """Make random ModeData from an emulated simulation run."""
         monitor = td.ModeMonitor(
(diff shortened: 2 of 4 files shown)


FilipeFcp:
Regarding `mode_area` (bot reminded me that): The interpolation seems necessary to adjust the data type, and for some reason, the value is slightly different (on the numerical error) from the one assigned to the monitor, so it doesn't generate the nans.

lucas-flexcompute:
> Regarding `mode_area` (bot reminded me that): The interpolation seems necessary to adjust the data type, and for some reason, the value is slightly different (on the numerical error) from the one assigned to the monitor, so it doesn't generate the nans.

Either way, does it hurt to include the mode area under the same conditional? If it also works, I think it's the safer choice.

lucas-flexcompute (approved):

lucas-flexcompute:
It's probably a good idea to include a short test case with a single frequency to guarantee the problem doesn't happen again in the future…

FilipeFcp:
Hi @lucas-flexcompute,

I just included the mode_area, it works well (not sure why I thought it wouldn’t in the first place).

Anyway, the user added another frequency to the monitor and could load the file, but mentioned that “Tidy3D and Zemax have a 90-degree difference.”

I’m not sure if this is a true rotation or just a transposition. Do you think transposing the data would fix it?

e_x = self._tangential_fields["E" + dim1].T
e_y = self._tangential_fields["E" + dim2].T

Or is this likely specific to this case?

bzhangflex:
Thanks @FilipeFcp! It looks good, I agree with @lucas-flexcompute that it would be good to include a single frequency test case. The tests are [here](https://github.com/flexcompute/tidy3d/blob/b58ae5f56ad488d5bebaf0840ee96c5f90cb0978/tests/test_data/test_monitor_data.py#L875), let me know if you need any help.

> Anyway, the user added another frequency to the monitor and could load the file, but mentioned that “Tidy3D and Zemax have a 90-degree difference.”

By 90-degree difference do they mean the x and y coordinates are swapped? It might be case dependent but it's hard to tell without more context, would the user be able to share more information?

FilipeFcp:
Thanks @bzhangflex.
It is my first time implementing a test. I will take a look at it and give it a try. If I need any help, I will ask you, thanks!

I will ask the user for some more context and let you know. Hi is from Japan, so probably he will reply just on Monday.

lucas-flexcompute:
> Hi @lucas-flexcompute,
> 
> I just included the mode_area, it works well (not sure why I thought it wouldn’t in the first place).
> 
> Anyway, the user added another frequency to the monitor and could load the file, but mentioned that “Tidy3D and Zemax have a 90-degree difference.”
> 
> I’m not sure if this is a true rotation or just a transposition. Do you think transposing the data would fix it?
> 
> e_x = self._tangential_fields["E" + dim1].T e_y = self._tangential_fields["E" + dim2].T
> 
> Or is this likely specific to this case?

This is a good question, since we did not have the opportunity to test or compare this code at all so far! We should try to work with the user to find out more.

bzhangflex:
Regarding the coordinate rotation I think I might see the issue, [here](https://github.com/flexcompute/tidy3d/blob/72763e8af779a77333f3ee8633fc50cf6e085106/tidy3d/components/data/monitor_data.py#L1246) instead of
`e_flat = e.values.flatten(order="C")`

I think it might need to be
`e_flat = e.values.flatten(order="F")`

If that's the case then `read_zbf()` probably needs to be updated too.

One test is if the user sets different values for `n_x` and `n_y` in `to_zbf(),` do they run into any errors or weird field profiles when they load the zbf into Zemax? If so I think it might be the flatten ordering causing the problem.

FilipeFcp:
Hi @bzhangflex and @lucas-flexcompute.

I have added the tests, and also implemented the option to pass the argument for the flatten order to 'A' or 'F', which hopefully can help us to debug it in the future.
@bzhangflex, I have added the FlattenOrders type, following the `UnitsZBF` model. Let me know if I need to change something or use a different approach. And thanks for your help so far!

@tomflexcompute, this is the PR I mentioned today

lucas-flexcompute (commented):
Is this `flatten_order` something that we need to make available? My understanding was that, once we confirmed the expected ordering in Zenmax, we could just set it here and not have the user worrying about it…

bzhangflex (approved):
Thank you @FilipeFcp! Looks great to me. @lucas-flexcompute would you like to have a quick look as well?

lucas-flexcompute:
I think it looks good as well!

Issue #2550, opened 2025-06-06
Get and plot field decay values after a simulation


tomflexcompute:
Some users want to get and plot the field decay values after running the simulation to check the convergence. Currently they can print out the log file and visually check the field decay values but there is no way, as far as I know, to get the numerical values of the field decay. 

I used a pattern matching script to extract the values from the log string, which seems to work well but maybe we can provide a more user friendly way to do so?

<img width="492" alt="Image" src="https://github.com/user-attachments/assets/137724ac-fa21-4c4a-9136-c8d135aa6ca1" />

tylerflex:
can we store the time series as a xarray.DataArray perhaps? or pandas.DataFrame (although not sure if we can serialize those?)

momchil-flex:
The values are stored in a file `"output/solver_progress.csv"` which can be downloaded from S3. For historic reasons (I guess) the file is not annotated but this is what it looks like

![Image](https://github.com/user-attachments/assets/f7b93e80-128d-48de-9174-2549c90b9634)

The columns are: row index, % done, field decay, time step, total number of time steps

tylerflex:
is there a way to just load this file to an array without direct downloading? we could write a simple wrapper for this in the SimulationData

tylerflex:
I guess we could use a tmp file ?

gvnwst:
I was this user, and Tom's code snippet was great. When doing convergence tests and validating a simulation's behavior this is super valuable -- I had a small reflection (which I didn't care about at all) essentially doubling my runtime because it kept the shutoff level above 1e-5.  Plotting it I was able to see where the "desired" cutoff level really was.

mardolph-ruco:
Ideally, the user should have the field decays as another TimeDataArray inside the simulation data (also accessible from its serialized representation), right?
From my perspective, I see 3 alternatives to provide that:

**1)**
An obvious quick solution would be to just have a property attribute, parsing the log as already shown by @tomflexcompute.
But this seems unsafe as it would rely on the logging format staying like this.
On the other hand, the current code relies on this logging format anyway, as the final decay value is already inferred from it.

**2)**
Alternatively, one could read the field decay values from S3 as pointed out by @momchil-flex.
This still seems somewhat hacky and resource-intensive to me, as it would require the following steps:

- download the simulation data from the server
- deserialize the simulation data
- load the progress data from the server
- add the field decay values from the progress data to the simulation data
- serialize the simulation data

[...]

PR #2699, opened 2025-07-25
Add access to field decay values in SimulationData

mardolph-ruco:
Relates to #2550 

Changes:

- Parse log to access step and field decay information as property attribute `sim_data.field_decay`, return as `TimeDataArray`
- adapt `final_decay_value`: use `field_decay` field, take last entry

Example usage:
```
field_decay = sim_data.field_decay
field_decay.plot()
plt.xlabel("Time Step")
plt.ylabel("Field Decay")
plt.yscale("log")
plt.show()
```

image

## Greptile Summary

This PR adds programmatic access to field decay values from simulation logs in the Tidy3D framework. Previously, users could only visually inspect field decay information in log outputs, but had no way to extract numerical values for analysis or plotting.

The implementation introduces a new `field_decay` property to the `SimulationData` class that parses the simulation log using regex patterns to extract time step and field decay information. The regex pattern matches log entries like "- Time step 827 / time 4.13e-14s (4 % done), field decay: 0.110e+00" to extract both the step number and decay value. This data is returned as a `TimeDataArray` with time steps as coordinates, making it consistent with other time-series data in the framework.

The existing `final_decay_value` method is refactored to use this new `field_decay` property, maintaining backward compatibility while reducing code duplication. The implementation includes proper error handling for cases where the log is None (raises `DataError`) or when no field decay values are found (returns a warning and defaults to 1.0 for `final_decay_value`).

This change integrates well with the existing codebase by following established patterns for data access properties and using the same `TimeDataArray` structure used elsewhere in the framework. The functionality enables users to create convergence plots and programmatically assess simulation quality, which is valuable for validation and optimization workflows.

## Confidence score: 4/5

- This PR is generally safe to merge with minimal risk of introducing bugs or breaking existing functionality.
- The implementation follows good practices with proper error handling, maintains backward compatibility, and includes comprehensive tests covering normal operation and edge cases.
- The `tidy3d/components/data/sim_data.py` file needs careful review to ensure the regex pattern correctly handles all possible log formats and that the error handling is robust.

### DIFF
--- a/tidy3d/components/data/sim_data.py
+++ b/tidy3d/components/data/sim_data.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import pathlib
+import re
 from abc import ABC
 from collections import defaultdict
 from typing import Callable, Optional, Union
@@ -14,7 +15,7 @@ import pydantic.v1 as pd
 import xarray as xr
 
 from tidy3d.components.autograd.utils import split_list
-from tidy3d.components.base import JSON_TAG, Tidy3dBaseModel
+from tidy3d.components.base import JSON_TAG, Tidy3dBaseModel, cached_property
 from tidy3d.components.base_sim.data.sim_data import AbstractSimulationData
 from tidy3d.components.file_util import replace_values
 from tidy3d.components.monitor import Monitor
@@ -29,7 +30,7 @@ from tidy3d.constants import C_0, inf
 from tidy3d.exceptions import DataError, FileError, Tidy3dKeyError
 from tidy3d.log import log
 
-from .data_array import FreqDataArray
+from .data_array import FreqDataArray, TimeDataArray
 from .monitor_data import AbstractFieldData, FieldTimeData, MonitorDataType, MonitorDataTypes
 
 DATA_TYPE_MAP = {data.__fields__["monitor"].type_: data for data in MonitorDataTypes}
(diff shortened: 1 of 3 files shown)


tidy3d/components/data/sim_data.py
```diff
@@ -932,20 +933,28 @@ class SimulationData(AbstractYeeGridSimulationData):
...
+        decays = [float(m[1]) for m in matches]
+        return TimeDataArray(decays, coords={"t": steps})
+
+    @property
+    def final_decay_value(self) -> float:
+        """Returns value of the field decay at the final time step."""
+        field_decay = self.field_decay
+        if len(field_decay) == 0:
+            log.warning("No field decay values found, using 1.0 as final decay value.")
+            return 1.0
```
yaugenst-flex, line 956: Does `1.0` make sense in this case or should it rather be `None`? No strong opinion either way but wondering why you decided to go with this?

tidy3d/components/data/sim_data.py
```diff
@@ -932,20 +933,28 @@ class SimulationData(AbstractYeeGridSimulationData):
     )
 
     @property
```
yaugenst-flex: This should probably be a cached property. Note we use `td.components.base.cached_property`.

yaugenst-flex (commented):
Thanks @mardolph-ruco this looks great! Really have only some very minor nitpicks.

tidy3d/components/data/sim_data.py
mardolph-ruco: Had the same thought, chose 1.0 as this was the behavior before.
Actually, 1.0 could be misleading here https://github.com/flexcompute/tidy3d/blob/b58ae5f56ad488d5bebaf0840ee96c5f90cb0978/tidy3d/web/api/tidy3d_stub.py#L266 if the `shutoff_value` was reached before the first progress log. Not sure, maybe change to None?
Alternatively, the log could be checked for "Field decay smaller than shutoff factor, exiting solver."

tidy3d/components/data/sim_data.py
mardolph-ruco: makes sense!

tylerflex (approved):
Looks great to me, thanks for the attention to detail!

tylerflex:
@tomflexcompute might be good to add an FAQ mention to let users know they can use

```py
field_decay = sim_data.field_decay
field_decay.plot()
plt.xlabel("Time Step")
plt.ylabel("Field Decay")
plt.yscale("log")
plt.show()
```

from here on

tidy3d/components/data/sim_data.py
tylerflex: my understanding is that the field decay signifies the ratio of the current average field value to the maximum value seen so far. So even if it is not logged, I think 1.0 is a reasonable assumption. This is probably why we went with it originally.

I'd be ok with just leaving this as default and reassessing this later if we ever want to improve how we do the field decay logging, eg with a backend storage and potentially checking it more often on the backend.

yaugenst-flex: ~~Yes exactly. I think we should just be explicit and return `None` if there is no final decay value. Generally this would likely point to a problem, but I would say that's the caller's concern then 😄~~

For the sake of maintaining backward compatibility let's just stick to the old behavior. We should revisit this at some point though.

yaugenst-flex (approved):

yaugenst-flex:
@mardolph-ruco thanks this looks ready to go! Could you squash your commits and rebase the branch, and we can merge once CI passes!

tomflexcompute (approved):
This is great to have!

PR #2701, opened 2025-07-28
Adding structured monitor validator for Heat

marc-flex:


tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -968,6 +968,31 @@ def check_transient_heat(cls, values):
...
+        if not volumetric_monitor:
+            structured_monitors = [m for m in monitors if m.unstructured is False]
+            if len(structured_monitors) > 0:
+                # In this case we need to either add a volumetric monitor or change monitors to be unstructured.
+                # NOTE: for efficiency, it's preferable to change them to unstructured.
+                new_monitors = [m.updated_copy(unstructured=True) for m in monitors]
+                values["monitors"] = tuple(new_monitors)
+
+                log.warning(
+                    "No volumetric monitor has been defined in a Heat simulation with structured monitors. "
```
daquinteroflex, line 991: @yaugenst-flex fyi re this, re mutating the defined state the users define if it's incompatible with the solver. Don't want to break it with a validation warning because of existing user scripts possibly, but do want to make it run correctly

PR #2707, opened 2025-07-31
adding to_file after uploading batch

FilipeFcp:
Hi @momchil-flex,

I'm not sure if you're the right person to review this. If not, please feel free to add the correct reviwer.

While assisting a user, I realized it would be straightforward to save the batch file after uploading the tasks but **before** running it, so that the `batch.hdf5` file contains the correct task_ids. If, for some reason, the kernel crashes or there is an internet disconnection before the batch completes, the user can easily load the batch.hdf5 and call the .load method.

Currently, the batch.hdf5 file is created initially without task_ids. If something happens before the batch finishes, it becomes a headache to recover.

## Greptile Summary

This PR improves the batch workflow robustness by addressing a critical user experience issue where batch executions could become unrecoverable if interrupted. The main change adds a `to_file()` call in the `run()` method immediately after `upload()` but before `start()`, ensuring that the `batch.hdf5` file contains the necessary task_ids for recovery. Additionally, the `load()` method has been restructured to handle ModeSolver data patching before downloading.

The core problem being solved is that previously, if a batch execution was interrupted (due to kernel crashes, network disconnections, etc.) before completion, users were left with `batch.hdf5` files lacking task_ids, making recovery difficult. With this change, users can now easily recover from interruptions by loading the saved batch file and calling the `load()` method.

The secondary change in the `load()` method reorders operations to process ModeSolver simulations' data patching before downloading, which could optimize performance by avoiding unnecessary downloads for mode solver simulations that may have different data handling requirements.

## Important Files Changed

<details>
<summary>Files Changed</summary>

| tidy3d/web/api/container.py | 4/5 | Added batch file saving after upload but before execution, and reordered load method operations |

</details>

## Confidence score: 4/5

• This PR addresses a legitimate user experience problem and the solution is straightforward and low-risk
• The change follows a logical execution order and doesn't introduce breaking changes to existing functionality
• The file needs attention to ensure the reordering in the load method doesn't introduce subtle timing issues with ModeSolver data handling

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
    participant Batch
    participant Job
    participant WebAPI
    participant FileSystem

    User->>Batch: run(path_dir)
    Batch->>Batch: _check_path_dir(path_dir)
    Batch->>Batch: upload()
    Batch->>Batch: _check_folder(folder_name)
    
    loop For each simulation
        Batch->>Job: create job instance
        Batch->>Job: upload()
        Job->>WebAPI: upload simulation
        WebAPI-->>Job: return task_id
    end
    
    Note over Batch: NEW: Save batch with task_ids
    Batch->>Batch: to_file(_batch_path(path_dir))
    Batch->>FileSystem: save batch.hdf5 with task_ids
    
    Batch->>Batch: start()
    
    loop For each job
        Batch->>Job: start()
        Job->>WebAPI: start task
    end
    
    Batch->>Batch: monitor()
    Batch->>Batch: load(path_dir)
    Batch-->>User: return BatchData
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed missing amplitude factor and handling of negative normal direction case when making adjoint sources from `DiffractionMonitor`.
+- Improved the robustness of batch jobs. The batch state, including all `task_ids`, is now saved to `batch.hdf5` immediately after upload. This fixes an issue where an interrupted batch (e.g., due to a kernel crash or network loss) would be unrecoverable.
 
 ## [2.9.0] - 2025-08-04
 
--- a/tests/test_web/test_webapi.py
+++ b/tests/test_web/test_webapi.py
@@ -1,6 +1,8 @@
 # Tests webapi and things that depend on it
 from __future__ import annotations
 
+import os
+
 import numpy as np
 import pytest
 import responses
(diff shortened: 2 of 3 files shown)


yaugenst-flex (approved):
Thanks filipe, looks great! Thanks for the catch.

momchil-flex:
Looks good, seems like the failures were in the self-hosted runners and we can merge this, but I assume only after 2.9.0 to be safe?

Generally still need to work out a lot about how we control what gets merged when we're wrapping up a release @yaugenst-flex @daquintero , e.g. what if someone is not on the same page with the team and just merges something.

PR #2715, opened 2025-08-04
Immutable DataArrays for RF data

dmarek-flex:
WIP attempt at providing immutable `DataArray` for microwave quantities. The goal here is to get rid of methods like `ImpedanceCalculator._set_data_array_attributes`, which add some simple data attributes for users.

This can work, but it just seems like a lot of code to get some fairly simple functionality, namely providing units and name for a `DataArray`.


dmarek-flex:
@daquinteroflex @yaugenst-flex Do you think this approach is better for packaging microwave data arrays?

If yes, I will continue by improving some of the data arrays used and produced by the `TerminalComponentModeler`.

tidy3d/components/microwave/data/data_array.py
daquinteroflex, line 1: Until we reorganise https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/data/data_array.py it should go there. Note the other voltage, index spatial data arrays. Can we reuse?

daquinteroflex: I think it's fine to start formalizing these classes even if it's more code, as long as we can reuse or interoperate with the existing ones

daquinteroflex: Fyi @marc-flex @damiano-flex

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -253,12 +255,12 @@ def compute_voltage(self, em_field: MonitorDataTypes) -> IntegralResultTypes:
...
         Returns
         -------
-        :class:`.IntegralResultTypes`
+        :class:`.VoltageIntegralResultTypes`
             Result of voltage computation over remaining dimensions (frequency, time, mode indices).
         """
         AxisAlignedPathIntegral._check_monitor_data_supported(em_field=em_field)
         voltage = -1.0 * self.compute_integral(field="E", em_field=em_field)
-        voltage = VoltageIntegralAxisAligned._set_data_array_attributes(voltage)
+        voltage = VoltageIntegralAxisAligned._make_result_data_array(voltage)
```
daquinteroflex: Is it fine by you that we make this a standalone function rather than a class instantiaton? Like we do for the TerminalComponentModelerData methods https://github.com/flexcompute/tidy3d/blob/bfe45d89cbfd65ead7699b74441157f28f217dc0/tidy3d/plugins/microwave/path_integrals.py#L208C1-L215C69

tidy3d/plugins/microwave/impedance_calculator.py
```diff
@@ -112,17 +118,14 @@ def check_voltage_or_current(cls, val, values):
...
-        data_array.name = "Z0"
-        return data_array.assign_attrs(units=OHM, long_name="characteristic impedance")
+    def _make_result_data_array(result: DataArray) -> ImpedanceResultTypes:
+        """Helper for creating the proper result type."""
+        cls = ImpedanceFreqDataArray
+        if "t" in result.coords:
+            cls = ImpedanceTimeDataArray
+        if "f" in result.coords and "mode_index" in result.coords:
+            cls = ImpedanceFreqModeDataArray
+        return cls.assign_data_attrs(cls(data=result.data, coords=result.coords))
```
daquinteroflex: Can we move this to utils as a standalone function?

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -206,13 +220,14 @@ def _check_monitor_data_supported(em_field: MonitorDataTypes):
...
         """Helper for creating the proper result type."""
+        cls = FreqDataArray
         if "t" in result.coords:
-            return TimeDataArray(data=result.data, coords=result.coords)
+            cls = TimeDataArray
         if "f" in result.coords and "mode_index" in result.coords:
-            return FreqModeDataArray(data=result.data, coords=result.coords)
-        return FreqDataArray(data=result.data, coords=result.coords)
+            cls = FreqModeDataArray
+        return cls.assign_data_attrs(cls(data=result.data, coords=result.coords))
```
daquinteroflex: Same this should be a standalone function in utils?

tidy3d/components/microwave/data/data_array.py
dmarek-flex: Oh that could get confusing. Looks like those VoltageArrays are arrays that have coordinates of voltage, whereas the VoltageArrays I want to introduce hold values related to voltage (with coordinates of freq/time/mode_index)

marc-flex: I feel like this is somewhat related to this: https://github.com/flexcompute/tidy3d-core/issues/863
Maybe it is a good time to think about a holistic approach to all this?

tidy3d/plugins/microwave/custom_path_integrals.py
dmarek-flex: ran into some circular dependency issues, but this is closer to what you want

daquinteroflex:
Doing final testing and changes in https://github.com/flexcompute/tidy3d/pull/2741/files

PR #2726, opened 2025-08-08
Add taper support and refactor array-factor calculation

George-Guryev-flxcmp:
Add taper support to `RectangularAntennaArrayCalculator`; refactor array factor computation for clarity and efficiency.
- Implemented taper integration
- Refactored AF calculation
- Updated tests and docs

### DIFF
--- a/tidy3d/plugins/microwave/array_factor.py
+++ b/tidy3d/plugins/microwave/array_factor.py
@@ -7,7 +7,9 @@ from typing import Optional, Union
 
 import numpy as np
 import pydantic.v1 as pd
-from pydantic.v1 import NonNegativeFloat, PositiveInt
+from pydantic.v1 import NonNegativeFloat, PositiveInt, conint
+from scipy.signal.windows import blackman, blackmanharris, chebwin, hamming, hann, kaiser, taylor
+from scipy.special import j0, jn_zeros
 
 from tidy3d.components.base import Tidy3dBaseModel, skip_if_fields_missing
 from tidy3d.components.data.monitor_data import AbstractFieldProjectionData, DirectivityData
@@ -20,13 +22,21 @@ from tidy3d.components.monitor import AbstractFieldProjectionMonitor, MonitorTyp
 from tidy3d.components.simulation import Simulation
 from tidy3d.components.source.utils import SourceType
 from tidy3d.components.structure import MeshOverrideStructure, Structure
-from tidy3d.components.types import ArrayLike, Axis, Bound, Undefined
+from tidy3d.components.types import TYPE_TAG_STR, ArrayLike, Axis, Bound, Undefined
 from tidy3d.constants import C_0, inf
+from tidy3d.exceptions import Tidy3dNotImplementedError
 from tidy3d.log import log
 
 
 class AbstractAntennaArrayCalculator(Tidy3dBaseModel, ABC):
     """Abstract base for phased array calculators."""
+
+    taper: Union[RectangularTaper, RadialTaper] = pd.Field(
+        None,
+        discriminator=TYPE_TAG_STR,
+        title="Antenna Array Taper",
+        description="Amplitude weighting of array elements to control main lobe width and suppress side lobes.",
+    )
 
     @property
     @abstractmethod
@@ -484,15 +494,15 @@ class AbstractAntennaArrayCalculator(Tidy3dBaseModel, ABC):
     ) -> AbstractFieldProjectionData:
         """Apply the array factor to the monitor data of a single antenna.
 
-        Parameters:
+        Parameters
         ----------
         monitor_data : AbstractFieldProjectionData
             The monitor data of a single antenna.
         new_monitor : AbstractFieldProjectionMonitor = None
             The new monitor to be used in the resulting data.
 
-        Returns:
-        --------
+        Returns
+        -------
         AbstractFieldProjectionData
             The monitor data of the antenna array.
         """
@@ -550,13 +560,13 @@ class AbstractAntennaArrayCalculator(Tidy3dBaseModel, ABC):
 
         Note that any near-field monitor data will be ignored.
 
-        Parameters:
+        Parameters
         ----------
         antenna_data : SimulationData
             The far-field data of a single antenna.
 
-        Returns:
-        --------
+        Returns
+        -------
         SimulationData
             The far-field data of the antenna array.
         """
@@ -592,6 +602,112 @@ class AbstractAntennaArrayCalculator(Tidy3dBaseModel, ABC):
             log_once=True,
         )
         return values
+
+    def _rect_taper_array_factor(
+        self, exp_x: ArrayLike, exp_y: ArrayLike, exp_z: ArrayLike
+    ) -> ArrayLike:
+        """
+        Compute the array factor assuming a separable rectangular (Cartesian) taper.
+
+        This method evaluates the array factor using separable amplitude weights
+        along the x, y, and z dimensions.
+
+        Parameters
+        ----------
+        exp_x: ArrayLike
+            3D array of phases along x axis [N_x, N_theta, N_freq].
+        exp_y: ArrayLike
+            3D array of phases along x axis [N_y, N_theta, N_freq].
+        exp_z: ArrayLike
+            3D array of phases along x axis [N_z, N_theta, N_freq].
+
+        Returns
+        -------
+        ArrayLike
+            Array factor values for each combination of theta and phi.
+        """
+
+        # if taper is not defined set all amplitudes to 1.0
+        if self.taper is None:
+            amp_x = (
+                1.0 if self.amp_multipliers[0] is None else self.amp_multipliers[0][:, None, None]
+            )
+            amp_y = (
+                1.0 if self.amp_multipliers[1] is None else self.amp_multipliers[1][:, None, None]
+            )
+            amp_z = (
+                1.0 if self.amp_multipliers[2] is None else self.amp_multipliers[2][:, None, None]
+            )
+
+        # if rectangular taper is spacified
+        elif isinstance(self.taper, RectangularTaper):
+            # get amplitudes along x, y and z axes
+            amp_x, amp_y, amp_z = self.taper.amp_multipliers(self.array_size)
+
+            # broadcast amplitudes to [N_x,1,1], [N_y,1,1] and [Nz,1,1], respectively
+            amp_x = amp_x[:, None, None]
+            amp_y = amp_y[:, None, None]
+            amp_z = amp_z[:, None, None]
+
+        # Tapers with non-separable amplitude weights are not supported by this function
+        else:
+            raise ValueError(f"Unsupported taper type {type(self.taper)} was passed.")
+
+        # Calculate individual array factors in x, y, and z directions
+        af_x = np.sum(
+            amp_x * exp_x,
+            axis=0,
+        )
+        af_y = np.sum(
+            amp_y * exp_y,
+            axis=0,
+        )
+        af_z = np.sum(
+            amp_z * exp_z,
+            axis=0,
+        )
+
+        # Calculate the overall array factor
+        array_factor = af_x * af_y * af_z
+
+        return array_factor
+
+    def _general_taper_array_factor(
+        self, exp_x: ArrayLike, exp_y: ArrayLike, exp_z: ArrayLike
+    ) -> ArrayLike:
+        """
+        Compute the array factor assuming a non-separable (non-Cartesian) taper.
+
+        This method evaluates the array factor using non-separable amplitude weights
+        along the x, y, and z dimensions.
+
+        Parameters
+        ----------
+        exp_x: ArrayLike
+            3D array of phases along x axis [N_x, N_theta, N_freq].
+        exp_y: ArrayLike
+            3D array of phases along x axis [N_y, N_theta, N_freq].
+        exp_z: ArrayLike
+            3D array of phases along x axis [N_z, N_theta, N_freq].
+
+        Returns
+        -------
+        ArrayLike
+            Array factor values for each combination of theta / phi and frequency.
+        """
+        # get taper weights
+        amps = self.taper.amp_multipliers(self.array_size)
+
+        # ensure amplitude weights are in format tuple[ArrayLike, ]
+        if len(amps) != 1:
+            raise ValueError(
+                "Non-cartesian taper was expected. Please ensure a valid taper is used."
+            )
+
+        # compute array factor: AF(theta,f) = sum_{x,y,z} amp(x,y,z) * exp_x(x,theta,f)*exp_y(y,theta,f)*exp_z(z,theta,f)
+        array_factor = np.einsum("xpf,ypf,zpf,xyz->pf", exp_x, exp_y, exp_z, amps[0])
+
+        return array_factor
 
 
 class RectangularAntennaArrayCalculator(AbstractAntennaArrayCalculator):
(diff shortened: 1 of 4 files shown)


tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +875,394 @@ def array_factor(
...
-        array_factor = af_x * af_y * af_z
 
-        return array_factor
+class AbstractWindow(Tidy3dBaseModel, ABC):
+    """This class provides interface for window selection."""
+
+    class Config:
+        """Disallow extra fields not defined in the model."""
+
+        extra = "forbid"
```
dbochkov-flexcompute: I don't think we need this, `Tidy3dBaseModel` prohibits extra fields already

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +875,394 @@ def array_factor(
...
+    """Taylor window with configurable sidelobe suppression and sidelobe count."""
+
+    sll: pd.PositiveFloat = pd.Field(
+        default=30,
+        title="Sidelobe suppression level",
+        description="Desired suppression of sidelobe level relative to the DC gain.",
+        units="dB",
+    )
+
+    nbar: pd.PositiveInt = pd.Field(
```
dbochkov-flexcompute: is there an upper limit for this parameter?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +875,394 @@ def array_factor(
...
+    """Class for 2-D radial Taylor taper with configurable sidelobe suppression and sidelobe count."""
+
+    sll: pd.PositiveFloat = pd.Field(
+        default=30,
+        title="Sidelobe suppression level",
+        description="Desired suppression of sidelobe level relative to the DC gain.",
+        units="dB",
+    )
+
+    nbar: pd.PositiveInt = pd.Field(
```
dbochkov-flexcompute: same question, especially that we have a loop defined by it

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +875,394 @@ def array_factor(
...
+    """Class for radial taper."""
+
+    @pd.validator("window", check_fields=False)
+    def validate_window(cls, v):
+        if not isinstance(v, TaylorTaperc):
+            raise ValueError(
+                f"`{type(v).__name__}` is not supported by `RadialTaper`. "
+                f"Use `RectangularTaper` for this type of window."
+            )
+        return v
```
dbochkov-flexcompute: instead of creating a custom validator it's better to simply override window with specifying which types are allowed. That is:
```
    window: TaylorTaperc = pd.Field(
        ..., title="Window object", description="Window type used to taper array antenna."
    )
```
this will automatically check that window is only `TaylorTaperc`, and another advantage is that in API docs it will also show that only this window is allowed

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +875,394 @@ def array_factor(
...
+    """Class for rectangular taper."""
+
+    @pd.validator("window", check_fields=False)
+    def validate_window(cls, v):
+        if isinstance(v, TaylorTaperc):
+            raise ValueError(
+                "`TaylorTaperc` is not supported by `RectangularTaper`. "
+                "Use `RadialTaper` if you want to use a Taylor taper."
+            )
+        return v
```
dbochkov-flexcompute: similarly here, but let's also allow to specify different tapers along different dimensions. E.g.:
```
RectangularWindowType = Union[HammingWindow, BlackmanWindow, ...]
...
    window_x: Optional[RectangularWindowType] = pd.Field(
        None, 
        title="X Axis Window", 
        description="Window type used to taper array antenna along x dimension.",
        discriminator=TYPE_TAG_STR,
    )

    window_y: Optional[RectangularWindowType] = pd.Field(
        None, 
        title="Y Axis Window", 
        description="Window type used to taper array antenna along y dimension.",
        discriminator=TYPE_TAG_STR,
    )

    window_z: Optional[RectangularWindowType] = pd.Field(
        None, 
        title="Z Axis Window", 
        description="Window type used to taper array antenna along z dimension.",
        discriminator=TYPE_TAG_STR,
    )

    # validate that at least one window is provided
    ...
  
    # define class method to use the same taper along all dimensions
    @classsmethod
    def all_dims(window: RectangularWindowType) -> RectangularTaper:
    ...
```
note the use of `discriminator=TYPE_TAG_STR` for fields that accepts multiple types. This is to tell pydantic how it can quickly match provided object and type, so that it doesn't try to initialize provided object with all allowed types

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +875,394 @@ def array_factor(
...
+
+        Returns
+        -------
+        ArrayLike
+            1D array of Kaiser window weights.
+        """
+        return kaiser(N, self.beta)
+
+
+class TaylorTaperc(AbstractWindow):
```
dbochkov-flexcompute: How about we just make a single taylor window, which, depending on whether it is used in rectangular or radial taper, it will use either `taylor(N, self.nbar, self.sll)` or the continuous variant?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +875,394 @@ def array_factor(
...
+    window: AbstractWindow = pd.Field(
+        ..., title="Window object", description="Window type used to taper array antenna."
+    )
+
+    @abstractmethod
+    def amp_multipliers(
+        self, array_size: tuple[PositiveInt, PositiveInt, PositiveInt]
+    ) -> tuple[np.ndarray, ...]:
+        """
+        method ``amp_multiplier()`` computes taper amplitudes for phased array antennas.
```
dbochkov-flexcompute: can be simplified `method ``amp_multiplier()`` computes ...` -> `Compute ...`

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -28,6 +30,12 @@
 class AbstractAntennaArrayCalculator(Tidy3dBaseModel, ABC):
     """Abstract base for phased array calculators."""
 
+    taper: Union[RectangularTaper, RadialTaper] = pd.Field(
+        None,
+        title="Antenna Array Taper",
+        description="Amplitude weighting of array elements to control main lobe width and suppress side lobes.",
```
dbochkov-flexcompute, line 38: for fields that allow different types we add `discriminator=TYPE_TAG_STR,` to prevent ambiguous errors for reasons explained above

dbochkov-flexcompute (changes requested):
Great start! Still not sure about `at`, `sll`, `nbar`, etc vs `attenuation`, `side_lobe_level`, `num_constant_sidelobes`, etc. @yaugenst what's our policy on using pydantic alias feature on frontend?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +875,394 @@ def array_factor(
...
+        """
+        return chebwin(N, self.at)
+
+
+class TaylorWindow(AbstractWindow):
+    """Taylor window with configurable sidelobe suppression and sidelobe count."""
+
+    sll: pd.PositiveFloat = pd.Field(
+        default=30,
+        title="Sidelobe suppression level",
```
dbochkov-flexcompute: capitalize first letter of each word in `title` here and everywhere else

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: Added this feature to RactangularWindow.

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: Yes, that make code cleaner. Unified access to Taylor 1d and radial windows via ``TaylorWindow`` factory class.

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: This part is a bit tricky: 

1) This parameter is set before ``get_weights(N)`` method is called. In principle one would expect a number of side lobes to be upper bounded by ``(N-1)/2``. Thus, the value for N is not known before nbar is specified.

2) In principle, nbar could be arbitrarily large with little to no effect on window shape and sidelobes as it controls the number of expansion coefficients in a basis of Bessel function (for radial taper). If nbar exceeds the actual number of sidelobes, it has no effect on the observable part of the radiation pattern/array factor. However,  increasing nbar further may lead to increase in condition number.

3) In practice, nbar is typically chosen from interval [1,10]. Matlab implementation of Taylor Radial taper raises error if nbar exceeds 10. 

  
I opted in for option 3).

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: Done!

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -20,14 +23,22 @@
...
 
 class AbstractAntennaArrayCalculator(Tidy3dBaseModel, ABC):
     """Abstract base for phased array calculators."""
 
+    taper: Union[RectangularTaper, RadialTaper] = pd.Field(
+        None,
+        discriminator=TYPE_TAG_STR,
+        title="Antenna Array Taper",
+        description="Amplitude weighting of array elements to control main lobe width and suppress side lobes.",
+        # discriminator=
```
weiliangjin2021: to be removed?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -484,15 +495,15 @@ def monitor_data_from_array_factor(
     ) -> AbstractFieldProjectionData:
         """Apply the array factor to the monitor data of a single antenna.
 
-        Parameters:
-        ----------
+        Parameters
+        ---------
```
weiliangjin2021: one missing "-"?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +878,494 @@ def array_factor(
...
+        ArrayLike
+            1D array of Hann window weights.
+        """
+        return hann(N)
+
+
+class ChebWindow(AbstractWindow):
+    """Standard Chebyshev window for tapering with configurable sidelobe attenuation."""
+
+    at: pd.PositiveFloat = pd.Field(
```
weiliangjin2021: more informative to name the field "attenuation", unless "at" is widely used in the field.

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +878,494 @@ def array_factor(
...
+    window_z: Optional[RectangularWindowType] = pd.Field(
+        None,
+        title="Z Axis Window",
+        description="Window type used to taper array antenna along z axis.",
+        discriminator=TYPE_TAG_STR,
+    )
+
+    # validate that at least one window is provided
+    @classmethod
+    def all_dims(cls, window: RectangularWindowType) -> RectangularTaper:
```
weiliangjin2021: maybe `isotropic_window` for clarity?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +878,494 @@ def array_factor(
...
+        amps = (
+            self.window_x.get_weights(effective_size[0])
+            if self.window_x is not None
+            else np.ones(effective_size[0]),
+            self.window_y.get_weights(effective_size[1])
+            if self.window_y is not None
+            else np.ones(effective_size[1]),
+            self.window_z.get_weights(effective_size[2])
+            if self.window_z is not None
+            else np.ones(effective_size[2]),
```
weiliangjin2021: avoid repetition with a for loop, e.g.
```
amps = (window.get_weights(effective_size[ind]) if window is not None
            else np.ones(effective_size[ind]) for ind, window in enumerate([self.windows_x, self.windows_y, self.windows_z]))
```

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +878,494 @@ def array_factor(
...
-        return array_factor
+
+class RadialTaper(AbstractTaper):
+    """Class for Radial Taper."""
+
+    window: TaylorWindow = pd.Field(
+        ..., title="Window Object", description="Window type used to taper array antenna."
+    )
+
+    # Change TaylorWindow mode to call _TaylorRadial
```
weiliangjin2021: i think we should raise an error if TaylorWindow's mode is not radial, rather than changing it silently.

weiliangjin2021 (commented):
Very nice work!

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: Makes sense. Changed it to require explicit definition of "radial" mode in `TaylorWindow` and raise an error if TaylorWindow has a default "1d" mode.

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: Thank you for the suggestion! Looks cleaner now.

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: Yes, that would make it clearer.  Changed to `isotropic_window()`

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: Makes sense.

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: Great catch!

tidy3d/plugins/microwave/array_factor.py
George-Guryev-flxcmp: Same here.

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +877,496 @@ def array_factor(
...
+    Selects the appropriate implementation (1D or radial) based on the
+    context in which it is used—`RectangularTaper` or `RadialTaper`,
+    respectively.
+    """
+
+    mode: Literal["1d", "radial"] = pd.Field(
+        default="1d",
+        title="Mode of Taylor Window",
+        description="Desired mode of Taylor window.",
+    )
```
dbochkov-flexcompute: Hmm, is there a way we can avoid this `mode` switch? What if we just define
```
class AbstractWindow(Tidy3dBaseModel, ABC):
    """This class provides interface for window selection."""

    def _get_weights_discrete(self, N: int) -> ArrayLike:
        """Interface function for computing window weights at N points."""
        raise Tidy3DNotImplemented(f"Calculation of antenna amplitudes at a discrete number of points is not yet implemented for window type {self.type}.")

    def _get_weights_continuous(self, p_vec: ArrayLike) -> ArrayLike:
        """Interface function for computing window weights at given locations."""
        raise Tidy3DNotImplemented(f"Calculation of antenna amplitudes at arbitrary locations is not yet implemented for window type {self.type}.")
```
then `RectangularTaper` can call `_get_weights_discrete()` while `RadialTaper` can call `_get_weights_continuous()`.  `TaylorWindow` would have valid implementation for both, while all the others would only have `_get_weights_discrete()` for now. Would that work here? I think that would simplify the code

tidy3d/plugins/microwave/array_factor.py
weiliangjin2021: sorry, maybe `from_isotropic_window` make more sense

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +877,496 @@ def array_factor(
...
+class RadialTaper(AbstractTaper):
+    """Class for Radial Taper."""
+
+    window: TaylorWindow = pd.Field(
+        ..., title="Window Object", description="Window type used to taper array antenna."
+    )
+
+    # Change TaylorWindow mode to call _TaylorRadial
+    @pd.root_validator(pre=False)
+    def force_radial_taylor(cls, values):
```
weiliangjin2021: do we really need a root validator here?

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +877,496 @@ def array_factor(
...
+        if not values:
+            return values
+
+        w = values.get("window")
+
+        # Allow TaylorWindow instance; reject anything else early
+        if not isinstance(w, TaylorWindow):
+            raise TypeError(
+                f"Expected a TaylorWindow instance for 'window', got {type(w).__name__}."
+            )
```
weiliangjin2021: i think we don't need to validate this, as it should be validated already in the field's type declaration

weiliangjin2021 (approved):

tidy3d/plugins/microwave/array_factor.py
```diff
@@ -761,28 +892,413 @@ def array_factor(
...
+    )
+
+    window_z: Optional[RectangularWindowType] = pd.Field(
+        None,
+        title="Z Axis Window",
+        description="Window type used to taper array antenna along z axis.",
+        discriminator=TYPE_TAG_STR,
+    )
+
+    # validate that at least one window is provided
```
dbochkov-flexcompute: a stray code comment

dbochkov-flexcompute (approved):

PR #2731, opened 2025-08-11
remove sources from adjoint postprocessing dummy simulation

groberts-flex:
This fixes an error a user was seeing where they had an anisotropic medium in their structure. During adjoint postprocessing, when `sim_inf_structure` is made, the structure medium becomes the simulation background medium. This was failing a validator for having a plane wave inject into an anisotropic medium. We shouldn't need sources in this dummy simulation object since we are just using it to get permittivities.

I tested this on the failing notebook from the user to verify this fixes their problem with computing gradients with an anisotropic structure medium.

### DIFF
--- a/tidy3d/web/api/autograd/autograd.py
+++ b/tidy3d/web/api/autograd/autograd.py
@@ -22,6 +22,7 @@ from tidy3d.components.autograd.constants import (
 )
 from tidy3d.components.autograd.derivative_utils import DerivativeInfo
 from tidy3d.components.data.data_array import DataArray
+from tidy3d.components.grid.grid_spec import GridSpec
 from tidy3d.exceptions import AdjointError
 from tidy3d.web.api.asynchronous import DEFAULT_DATA_DIR
 from tidy3d.web.api.asynchronous import run_async as run_async_webapi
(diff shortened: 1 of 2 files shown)


yaugenst-flex (approved):
Nice thanks @groberts-flex, good catch!

tidy3d/web/api/autograd/autograd.py
```diff
@@ -1086,6 +1086,7 @@ def postprocess_adj(
                 structures=structs_inf_struct,
                 medium=structure.medium,
                 monitors=[],
```
momchil-flex, line 1093: Can we also remove monitors? Not sure if there's some monitor validator issue also lurking.

groberts-flex: the monitors have been getting removed until now, which shouldn't be a problem since we are just grabbing epsilon via the simulation. do you think there is a potential validator issue with removing them?

momchil-flex: Ah my bad - I misread the code. Looks good as is.

momchil-flex (approved):

momchil-flex (commented):
Actually hold on. Could there be a meshing discrepancy? The auto grid relies on the source wavelength.

It's not clear to me how this works currently - seems that the two simulations could have different grids already because of the different media?

momchil-flex (approved):

PR #2737, opened 2025-08-14
Fixed overlap_sort to use the correct value of conjugated_dot_product

dmarek-flex:
## Greptile Summary

This PR fixes a consistency issue in the `overlap_sort` method of the `ModeData` class in the Tidy3D computational electromagnetics framework. The fix ensures that the overlap sorting algorithm respects the `conjugated_dot_product` parameter from the `ModeMonitor` configuration, rather than using a hardcoded default value.

The changes address two specific instances where dot product calculations were not using the monitor's `conjugated_dot_product` setting:
1. In the main overlap calculation on line 1757, where `self.dot()` now receives `self.monitor.conjugated_dot_product` as a parameter
2. In the reduced amplitude calculation on lines 1770-1772, where `outer_dot()` now also receives this parameter

This fix is particularly important for mode analysis in lossy or gyrotropic materials, where the choice of conjugated vs non-conjugated dot products affects the physical interpretation of modal overlaps. In lossless systems, both formulations are equivalent up to a phase, but in lossy systems, the conjugated version represents power fraction carried by modes while the non-conjugated version maintains orthogonality. By ensuring consistency with the monitor configuration, the overlap sorting now properly follows the user's intended dot product formulation throughout the mode analysis workflow.

The PR also includes a minor documentation fix, correcting a typo from 'overlap_tresh' to 'overlap_thresh' in the docstring.

## Important Files Changed

<details>

| `tidy3d/components/data/monitor_data.py` | 5/5 | Fixed `overlap_sort` method to use monitor's `conjugated_dot_product` parameter in dot product calculations |
| `CHANGELOG.md` | 5/5 | Added changelog entry documenting the bug fix for `overlap_sort` consistency |

</details>

## Confidence score: 5/5

- This PR is safe to merge with minimal risk as it fixes a clear inconsistency bug without introducing breaking changes
- Score reflects a straightforward bug fix that improves consistency and correctness in mode analysis functionality
- No files require special attention as the changes are well-focused and address a specific technical issue

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
    participant ModeData
    participant ModeMonitor
    participant dot_method

    User->>ModeData: call overlap_sort()
    ModeData->>ModeData: _find_ordering_one_freq()
    ModeData->>dot_method: dot(data_to_sort, conjugated_dot_product)
    Note over dot_method: Previously: hardcoded True parameter
    Note over dot_method: Now: uses self.monitor.conjugated_dot_product
    ModeData->>ModeMonitor: get conjugated_dot_product field
    ModeMonitor-->>ModeData: return conjugated_dot_product value
    dot_method-->>ModeData: return dot product result
    ModeData->>ModeData: compute overlap and sorting
    ModeData-->>User: return sorted ModeData
```


dbochkov-flexcompute (commented):
Makes sense to me, but, actually, I wonder if we should do overlap sort using the non-conjugated dot-product no matter what. What do you think, @momchil-flex?

momchil-flex:
> Makes sense to me, but, actually, I wonder if we should do overlap sort using the non-conjugated dot-product no matter what. What do you think, @momchil-flex?

At first thought this would make sense to me. It should have very minimal / rare effect on typical photonic simulations, but could be good to compare results for some potentially problematic case, either in RF or in a lossy photonic waveguide.

momchil-flex:
Should we make some decision and get this in for rc1?

dmarek-flex:
> Should we make some decision and get this in for rc1?

Sorry for the delay, I havent tested it yet on one of these potentially problematic cases. Let me just try today to set up some kind of example, and see if any issues pop up. 

But, I think using the non-conjugated dot will always be superior.

dmarek-flex:
I tried a very lossy optical waveguide, and found no real difference when using the conjugated or unconjugated dot product for sorting. Haven't tried a transmission line example yet, but I don't expect it will change much.

My vote is to keep the existing behavior, so the overlap sort uses the value of `conjugated_dot_product` from the associated monitor. I'll just check to see if there are any surprises when sorting lossy transmission line modes

dmarek-flex:
Same conclusion for lossy transmission line, both dot product definitions work for the purposes of sorting. Unless any one can think of a particularly challenging case, I think both work well enough, even for lossy modes.

momchil-flex:
So what I'm a bit confused about now is why this fix was needed, as in what prompted its discovery? It must have been something where the results of the overlap sort differ depending on which product is used?

dmarek-flex:
> So what I'm a bit confused about now is why this fix was needed, as in what prompted its discovery? It must have been something where the results of the overlap sort differ depending on which product is used?

Ah well I simply noticed that the `overlap_sort` was using a different definition of the dot product then the field set in the associated monitor. I did not encounter any issue, so far, but I figured it might be an issue.

momchil-flex (approved):
Ah ok then this looks good to me!

dmarek-flex:
Does this look good to you @dbochkov-flexcompute?

dbochkov-flexcompute (approved):
Yeah, it does!

PR #2738, opened 2025-08-15
feat: `RF GUI <-> python` client interoperabilty

daquinteroflex:
## Greptile Summary

This PR implements RF GUI <-> Python client interoperability, enabling seamless integration between RF microwave simulation tools and the Python API. The changes introduce significant architectural improvements to the S-matrix plugin, including:

**Core Architecture Changes:**
- **Component Modeler Refactoring**: Split the original `ComponentModeler` class into `ModalComponentModeler` and `TerminalComponentModeler` for better separation of concerns between modal and terminal-based RF analysis
- **Data Container Pattern**: Introduced `ModalComponentModelerData` and `TerminalComponentModelerData` classes that separate simulation orchestration from results processing, following the established pattern in other Tidy3D solvers
- **Batch Processing Support**: Added comprehensive batch task management in the web API (`BatchTask`, `BatchStatus`, `BatchDetail`) to handle RF component modeling workflows that operate on collections of simulations

**Type System Reorganization:**
- Restructured the types module from a monolithic `types.py` file into a package with `base.py`, `third_party.py`, `utils.py`, and `simulation.py` submodules
- Standardized imports to use Python's standard library types (`typing.Literal`, `typing.Union`) instead of internal re-exports
- Added specialized data arrays (`VoltageArray`, `CurrentArray`, `ImpedanceArray`) and RF-specific port data structures

**RF-Specific Features:**
- **S-Matrix Analysis**: New analysis modules (`analysis/modal.py`, `analysis/terminal.py`) containing the mathematical core for scattering parameter calculations
- **Antenna Metrics**: Added antenna analysis capabilities for computing radiation patterns and efficiency metrics
- **Port Management**: Enhanced port handling with support for both lumped and wave ports, including impedance calculations and wave amplitude processing
- **Container Classes**: Introduced `SimulationMap` and `SimulationDataMap` for managing collections of simulations and their results with dictionary-like access patterns

**Web API Enhancements:**
- Extended task management to support RF batch operations with specialized endpoints and monitoring
- Added RF-specific task types (`COMPONENT_MODELER`, `TERMINAL_COMPONENT_MODELER`) to the web infrastructure
- Improved error handling with comprehensive JSON response parsing for better debugging
- Updated file handling to support component modeler artifacts (`cm_data.hdf5`)

The changes maintain backward compatibility where possible while introducing a deprecation warning for the `TerminalComponentModeler` class indicating breaking changes in version 2.10. The implementation follows established patterns from other Tidy3D solvers, ensuring consistency across the ecosystem.

## Important Files Changed

<details>
<summary>Click to expand file changes table</summary>

| `tidy3d/plugins/microwave/custom_path_integrals.py` | 1/5 | Critical bugs in utility functions that swap voltage and current result types |
| `tidy3d/plugins/microwave/path_integrals.py` | 1/5 | Type mismatches in utility functions will cause runtime errors |
| `tidy3d/plugins/smatrix/analysis/__init__.py` | 2/5 | Completely emptied file removing all terminal component analysis functions |
| `tidy3d/plugins/smatrix/run.py` | 2/5 | Incomplete implementation with FIXME comments and potential null pointer exceptions |
| `schemas/TerminalComponentModeler.json` | 2/5 | Entire JSON schema definition removed, leaving only blank file |
| `tidy3d/components/types/simulation.py` | 2/5 | Missing ModalComponentModeler and TerminalComponentModeler types that exist in stub |
| `tidy3d/components/tcad/generation_recombination.py` | 3/5 | Minor inconsistency with 'union=SECOND' instead of 'units=SECOND' in field definition |
| `tidy3d/plugins/smatrix/utils.py` | 3/5 | Missing type annotations and potential division by zero vulnerabilities |
| `tests/utils.py` | 4/5 | Added support for new monitor types following established testing patterns |
| `tidy3d/components/data/index.py` | 4/5 | New SimulationDataMap class with comprehensive docstrings and examples |
| `tidy3d/components/index.py` | 4/5 | New container classes with proper immutable mapping interface |
| `tidy3d/components/types/base.py` | 4/5 | Major refactoring from types.py to package structure with new features |
| `tidy3d/components/types/third_party.py` | 4/5 | Clean conditional import handling for optional trimesh dependency |
| `tidy3d/components/types/__init__.py` | 4/5 | Centralizes type exports following Python packaging best practices |
| `tidy3d/components/types/utils.py` | 4/5 | Extracted utility function to dedicated module improving organization |
| `docs/faq` | 4/5 | Submodule update likely incorporating new RF-related documentation |
| `docs/notebooks` | 4/5 | Submodule update likely including RF functionality examples |
| `tidy3d/plugins/smatrix/analysis/antenna.py` | 4/5 | New antenna analysis module with comprehensive metrics computation |
| `tidy3d/plugins/smatrix/analysis/modal.py` | 4/5 | Core S-matrix construction logic for modal component modeling |
| `tidy3d/plugins/smatrix/analysis/terminal.py` | 4/5 | Terminal analysis functions with proper electromagnetic theory implementation |
| `tidy3d/plugins/smatrix/component_modelers/base.py` | 4/5 | Refactored to remove web functionality and focus on core modeling logic |
| `tidy3d/plugins/smatrix/component_modelers/modal.py` | 4/5 | Renamed from ComponentModeler with S-matrix logic moved to analysis module |
| `tidy3d/plugins/smatrix/component_modelers/terminal.py` | 4/5 | Major simplification with computational logic moved elsewhere |
| `tidy3d/plugins/smatrix/data/data_array.py` | 4/5 | New specialized data arrays for RF measurements with license warnings |
| `tidy3d/plugins/smatrix/data/terminal.py` | 4/5 | Rewritten as comprehensive data container with analysis capabilities |
| `tidy3d/plugins/smatrix/ports/types.py` | 4/5 | Clean type definitions for different port categories |
| `tidy3d/plugins/smatrix/smatrix.py` | 4/5 | Backward compatibility layer removed as part of API restructuring |
| `tidy3d/plugins/smatrix/__init__.py` | 4/5 | Updated to use absolute imports and export new classes |
| `tidy3d/web/api/webapi.py` | 4/5 | Comprehensive batch task support for RF component modeling |
| `tidy3d/web/core/http_util.py` | 4/5 | Enhanced error handling with comprehensive JSON response parsing |
| `tidy3d/web/core/task_core.py` | 4/5 | Added configurable endpoints and BatchTask class for RF workflows |
| `tidy3d/web/core/task_info.py` | 4/5 | New batch-related data structures for component modelers |
| `tidy3d/web/api/tidy3d_stub.py` | 4/5 | Extended to handle new component modeler types |
| `tests/test_components/test_map.py` | 4/5 | Comprehensive test coverage for SimulationMap functionality |
| `tests/test_data/test_map_data.py` | 4/5 | Well-structured tests for SimulationDataMap container |
| `tests/test_plugins/smatrix/test_component_modeler.py` | 4/5 | Updated to new data container pattern and API structure |
| `tests/test_plugins/smatrix/test_terminal_component_modeler.py` | 4/5 | Major refactoring to align with new TerminalComponentModeler API |
| `tidy3d/components/base.py` | 5/5 | Clean import refactoring from local types to standard library |
| `tidy3d/components/data/monitor_data.py` | 5/5 | Import reorganization moving TYPE_TAG_STR to types module |
| `tidy3d/components/autograd/types.py` | 5/5 | Updated import path reflecting type system reorganization |
| `tidy3d/components/data/data_array.py` | 5/5 | Added voltage, current, and impedance data arrays following established patterns |
| `tidy3d/components/geometry/mesh.py` | 5/5 | Clean import path refactoring to types.third_party |
| `tidy3d/components/material/tcad/charge.py` | 5/5 | Standardized to use typing.Union directly |
| `tidy3d/components/medium.py` | 5/5 | Updated to import Literal from standard library |
| `tidy3d/components/monitor.py` | 5/5 | Clean refactoring to use standard typing imports |
| `tidy3d/components/tcad/doping.py` | 5/5 | Standardized Union import from typing |
| `tidy3d/components/tcad/simulation/heat_charge.py` | 5/5 | Import reorganization to more specific module locations |
| `tidy3d/components/tcad/types.py` | 5/5 | Refactored to use standard typing.Union directly |
| `tidy3d/plugins/adjoint/components/types.py` | 5/5 | Updated import reflecting type system reorganization |
| `tidy3d/plugins/adjoint/web.py` | 5/5 | Updated to use standard typing.Literal instead of custom import |
| `tidy3d/plugins/design/design.py` | 5/5 | Import reorganization moving TYPE_TAG_STR to types module |
| `tidy3d/plugins/microwave/impedance_calculator.py` | 5/5 | Refactored to use centralized utility functions |
| `tidy3d/plugins/smatrix/component_modelers/types.py` | 5/5 | New type definition creating Union of component modeler classes |
| `tidy3d/plugins/smatrix/data/modal.py` | 5/5 | New data container following established patterns |
| `tidy3d/plugins/smatrix/data/types.py` | 5/5 | Clean Union type for component modeler data classes |
| `tidy3d/plugins/smatrix/network.py` | 5/5 | Centralized type definitions for network elements |
| `tidy3d/plugins/smatrix/ports/modal.py` | 5/5 | Improved docstring for better user understanding |
| `tidy3d/web/api/connect_util.py` | 5/5 | Centralized timeout constants to common module |
| `tidy3d/web/api/container.py` | 5/5 | Added BatchCategoryType and configurable endpoints |
| `tidy3d/web/common.py` | 5/5 | New constants file centralizing timing configurations |
| `tidy3d/web/core/constants.py` | 5/5 | Added component modeler file artifact constants |
| `tidy3d/web/core/types.py` | 5/5 | Added new RF-related task types to enum |
| `tests/test_plugins/test_array_factor.py` | 5/5 | Removed verbose parameter for cleaner test output |
| `tests/test_plugins/smatrix/terminal_component_modeler_def.py` | 5/5 | Removed verbose parameter to align with API changes |
| `tidy3d/plugins/autograd/README.md` | 5/5 | Documentation correction updating class reference |
| `docs/api/plugins/smatrix.rst` | 5/5 | Updated documentation reflecting component modeler refactoring |
| `tidy3d/__init__.py` | 5/5 | Added new imports for RF GUI integration |
| `schemas/Simulation.json` | 5/5 | Updated type references reflecting module restructuring |

</details>

## Confidence score: 2/5

- This PR contains critical issues that will cause runtime failures, particularly in utility functions with swapped return types and incomplete implementations
- Score reflects multiple high-risk bugs including type mismatches, missing implementations, and removed essential functionality without proper migration
- Pay close attention to microwave path integral files, smatrix analysis modules, and the run.py module which all have significant implementation issues


tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -50,6 +38,11 @@
 class AbstractComponentModeler(ABC, Generic[IndexType, ElementType], Tidy3dBaseModel):
     """Tool for modeling devices and computing port parameters."""
 
+    name: str = pd.Field(
+        "",
+        title="Simulation",
+        description="Simulation describing the device without any sources present.",
+    )
```
momchil-flex, line 42: Title and description seem wrong

tidy3d/components/data/index.py
```diff
@@ -0,0 +1,27 @@
...
+        Returns:
+            The SimulationData object corresponding to the given port name.
+
+        Raises:
+            KeyError: If no port with the given name is found.
+        """
+        for i, index_i in enumerate(self.index):
+            if index_i == index:
+                return self.data[i]
+        raise KeyError(f"Index '{index}' not found.")
```
momchil-flex: Seems like we're basically going for an immutable dict? But it seems like in that case some validation is needed (well, it's meant to be kind of internal only, but some safeguards wouldn't hurt). Like `len(index) == len(data)` and `len(set(index)) == len(index)`

tidy3d/web/api/container.py
```diff
@@ -32,6 +32,8 @@
 DEFAULT_DATA_DIR = "."
 BATCH_MONITOR_PROGRESS_REFRESH_TIME = 0.02
 
+BatchCategoryType = Literal["tidy3d", "microwave", "tidy3d_design"]
```
momchil-flex, line 34: This doesn't allow e.g. `"autograd_fwd"` or `"autograd_bwd"` which I think might then error if an adjoint task is run through a Job? But I'm not 100% sure, but the point is that you're placing a Literal here but I know we have various `simulation_type`-s that may or may not be used with containers and that are not listed here.

tidy3d/web/api/container.py
```diff
@@ -44,9 +46,18 @@ def _check_path_dir(path: str) -> None:
         """Make sure local output directory exists and create it if not."""
 
     @staticmethod
-    def _check_folder(folder_name: str) -> None:
+    def _check_folder(
+        folder_name: str,
+        projects_endpoint: str = "tidy3d/projects",
+        project_endpoint: str = "tidy3d/project",
```
momchil-flex, line 51: Ehm... yikes? Looks very error-prone :)

tidy3d/web/api/webapi.py
```diff
@@ -278,27 +299,50 @@ def upload(
...
 
     task.upload_simulation(
         stub=stub,
         verbose=verbose,
         progress_callback=progress_callback,
         remote_sim_file=remote_sim_file,
     )
-    estimate_cost(task_id=task.task_id, solver_version=solver_version, verbose=verbose)
+
+    # TODO: cost estimation for RF task?
```
momchil-flex, line 355: Note that the main purpose of the estimate_cost call here is not to e.g. report the cost to the user. Instead this is just the function call to the metadata api, which should run upon task upload. I forget exactly why, but I think it had something to do with proper validation (and it's matching the GUI workflow). However, it seems that for RF something like this is done in the `start` function, so it might be fine not to do this here. A related question is: should the `"tidy3d/projects/terminal-component-modeler-split",` api be called here already, right on modeler upload? Probably again best to mirror what GUI is doing.

When it comes purely to defining an `estimate_cost` for a modeler, I guess it can only be called once all the sub-task are uploaded, but if so then we just iterate over those and sum up `estimate_cost(task_id=task.task_id, solver_version=solver_version)`?

tidy3d/web/api/webapi.py
```diff
@@ -829,6 +933,159 @@ def load(
...
+            return ("Validating", 2)
+        if status in ("Validate_Success", "Validate_Warn"):
+            return ("Validate", 3)
+        if status in ("Running",):
+            return ("Running", 4)
+        if status in ("Postprocess",):
+            return ("Postprocess", 5)
+        if status in ("Run_Success",):
+            return ("Success", 6)
+        return (status, 6)
```
momchil-flex: This is certainly not your fault but we've been plagued by non-unified status messages since the start. Here for some reason the status messages seem to start with a capital, which is not the case for all other tasks ("preprocess", "success", "running" etc. are used in solvers). Let's fix this if not too late?

tidy3d/web/api/webapi.py
```diff
@@ -829,6 +933,159 @@ def load(
...
+                time.sleep(REFRESH_TIME)
+    else:
+        terminal_errors = {
+            "Validate_Failed",
+            "Run_Failed",
+            "Run_Diverged",
+            "Blocked",
+            "Abort",
+            "Aborted",
+        }
```
momchil-flex, line 1068: Same comment about statuses. For reference see [here](https://github.com/flexcompute/tidy3d/blob/5681c5e7cb9bc4deba41a3d4969ef0f1cb580fa9/tidy3d/web/api/container.py#L775-L795). Indeed we've had to support things like "abort" and "aborted", now we're adding capitalization as well as some new ones. For example should "Run_Failed" exist at all, shouldn't it just be "error"? Similarly with "Run_Diverged", it can just be "diverged" which we have for FDTD.

tidy3d/web/api/webapi.py
```diff
@@ -829,6 +933,159 @@ def load(
...
+                if p < total and not postprocess_triggered:
+                    try:
+                        BatchTask(batch_id).postprocess(batch_type="RF_SWEEP")
+                    except Exception:
+                        pass
+                    postprocess_triggered = True
+                if p >= total:
+                    break
+            if s in terminal_errors:
+                raise WebError(f"Batch {batch_id} terminated: {s}")
```
momchil-flex: If it so happens that neither of these if statements are true at any time (for some reason), this will hang indefinitely.

momchil-flex (commented):
Some comments on a first pass.. The actual component modeler code is kind of hard to review as a lot of it is moving things around.

tidy3d/web/api/webapi.py
momchil-flex: I guess this is fine and we do it in other parts of the code, but the logic here is somewhat convoluted so a bit hard to feel safe.

tidy3d/components/data/index.py
daquinteroflex: Haha I was literally implementing that validator and tests when you commented, I've pushed now. Yeah ultimately it was this question of whether we wanted to move away from a dictionary in order to implement validation and indexing more powerfully as proposed during a conversation with Yannick too

tidy3d/web/api/webapi.py
daquinteroflex: @shanguomagi we might need your help on this, and unifying status messages. Not sure how this might affect the GUI side

tidy3d/plugins/smatrix/component_modelers/base.py
daquinteroflex: Fixed

tidy3d/web/api/webapi.py
yaugenst-flex: the batch tasks do have a separate endpoint that provides cost estimation which i think is what we'd end up calling here. but maybe for context, the way the webapi is implemented right now is essentially to expose an api that is "the same" as the other task types but under the hood, due to how that pipeline currently works, the implementation is quite messy. the goal is to remove/refactor a lot of this once the underlying pipeline issues are sorted out.

tidy3d/components/data/index.py
```diff
@@ -0,0 +1,47 @@
...
+import pydantic.v1 as pd
+
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.components.simulation_types import SimulationDataType
+
+
+class IndexSimulationData(Tidy3dBaseModel):
+    """Container for a set of simulation data, accessible by a string index."""
+
+    index: tuple[str, ...]
```
yaugenst-flex: index seems like a weird choice, i guess these are "keys"?

yaugenst-flex: i have a similar comment on the name of the class actually. maybe something like `SimulationDataMap`? not sure about that one either but i think it's better than "Index"

tidy3d/components/data/index.py
yaugenst-flex: along the same lines: this really is an immutable dict right? so there would be two ways to achieve this. for instance, you could actually just have this be a [`MappingProxyType`](https://docs.python.org/3/library/types.html#types.MappingProxyType), which is a read-only view into a dict. but since we like pydantic models, i'd do it like this:
```py
from collections.abc import Mapping
SimulationDataMap(Tidy3dBaseModel, Mapping):
  def __getitem__(self, key): ...
  def __iter__(self): ...
  def __len__(self): ...
```
this then behaves exactly like a dict

tidy3d/components/index.py
```diff
@@ -0,0 +1,49 @@
...
+
+from typing import Union
+
+import pydantic.v1 as pd
+
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.components.simulation_types import SimulationType
+
+
+class IndexSimulation(Tidy3dBaseModel):
```
yaugenst-flex: same as simulaton data comment

tidy3d/components/index.py
```diff
@@ -0,0 +1,49 @@
...
+    """Tuple of unique string identifiers for each simulation."""
+
+    simulation: tuple[SimulationType, ...]
+    """Tuple of :class:`.Simulation` objects corresponding to each index."""
+
+    @pd.root_validator()
+    def _validate_lengths_match(cls, values):
+        """Validate that index and simulation have the same length."""
+        index, simulation = values.get("index"), values.get("simulation")
+        if index is not None and simulation is not None and len(index) != len(simulation):
```
yaugenst-flex: are these optional? otherwise the `is None` check isnt necessary

tidy3d/components/data/index.py
```diff
@@ -0,0 +1,47 @@
...
+    """Tuple of unique string identifiers for each simulation data object."""
+
+    data: tuple[SimulationDataType, ...]
+    """Tuple of :class:`.SimulationData` objects corresponding to each index."""
+
+    @pd.root_validator()
+    def _validate_lengths_match(cls, values):
```
yaugenst-flex: another approach would be to actually just pass a dict `dict[str, SimulationData]` as the field, then you don't need the validator.

tidy3d/components/simulation_types.py
yaugenst-flex, line 1: this is an organizational nitpick but it might be better to have `/tidy3d/components/types/` and put the types there

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,273 @@
...
+)
+from tidy3d.plugins.smatrix.data.modal import IndexSimulationData, ModalComponentModelerData
+from tidy3d.plugins.smatrix.data.terminal import TerminalComponentModelerData
+from tidy3d.plugins.smatrix.data.types import ComponentModelerDataType
+from tidy3d.web import Batch, BatchData
+
+DEFAULT_DATA_DIR = "."
+
+
+def compose_simulation_data_index(port_task_map: dict[str, str]) -> IndexSimulationData:
```
yaugenst-flex: are these public functions or just for internal use? i'd probably underscore most of these

tidy3d/components/data/index.py
daquinteroflex: Yes I like this idea as you propose. Had looked at TypedDict before, but yeah validation is still the issue. `SimulationDataMap(Tidy3dBaseModel, Mapping)` is a good idea

tidy3d/components/data/index.py
momchil-flex: You mean the pydantic Field? Isn't the point that we don't allow mutable fields?

yaugenst-flex: yeah totally i was just thinking about mutable defaults for some resason. anyways separate fields are better either way

tidy3d/plugins/smatrix/run.py
daquinteroflex: I'm still debating if using some of them in the legacy examples so users know they exist to help them migrate

tidy3d/components/simulation_types.py
daquinteroflex: Yes this is a good idea

tidy3d/components/data/index.py
daquinteroflex: So with the new approach Yannick proposed we won't use this anyway

tidy3d/web/api/webapi.py
daquinteroflex: @shanguomagi  @momchil-flex  can you both review https://magicloud.larksuite.com/wiki/NUiOwqr4niXR9pkPcJ2uut9Lsge#share-UDtQd51heopJ85xhX1tukkcZsgI

momchil-flex: These make sense to me, thanks!

tidy3d/components/data/index.py
daquinteroflex: Should be implemented now nicely

tidy3d/components/data/index.py
daquinteroflex: Done

tidy3d/components/index.py
daquinteroflex: Done

tidy3d/components/data/index.py
daquinteroflex: Implemented with alias `keys` and `values` equivalent fields as those couldn't be set directly

tidy3d/components/simulation_types.py
daquinteroflex: https://github.com/flexcompute/tidy3d/pull/2754

tidy3d/web/api/webapi.py
daquinteroflex: https://github.com/flexcompute/tidy3d/pull/2756

CHANGELOG.md
```diff
@@ -1578,7 +1578,7 @@ which fields are to be projected is now determined automatically based on the me
 
 - `Simulation` symmetries now fully functional.
 - Ability to perform near-to-far transformations from multiple surface monitors oriented along the x, y or z directions using `tidy3d.plugins.Near2Far`.
-- `tidy3d.plugins.ComponentModeler` tool for scattering matrix calculations.
+- `tidy3d.plugins.ModalComponentModeler` tool for scattering matrix calculations.
```
momchil-flex: So I don't think we should be replacing this in the changelog. At the time of these versions, it was called `ComponentModeler`, not `ModalComponentModeler`.

On the other hand, we should have a pretty detailed 2.10 changelog item about the breaking changes that have been introduced.

tidy3d/components/data/index.py
```diff
@@ -0,0 +1,136 @@
...
+import pydantic.v1 as pd
+
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.components.simulation_types import SimulationDataType
+
+
+class SimulationDataMap(Tidy3dBaseModel, Mapping[str, SimulationDataType]):
+    """An immutable dictionary-like container for simulation data.
+
+    Iit provides standard dictionary
```
momchil-flex: It

tidy3d/components/index.py
```diff
@@ -0,0 +1,135 @@
...
+
+        This allows for using the built-in `len()` function, e.g., `len(my_map)`.
+
+        Returns
+        -------
+        int
+            The total number of items in the map.
+        """
+        assert len(self.keys_tuple) == len(self.values_tuple), "Internal state mismatch."
+        return len(self.keys_tuple)
```
momchil-flex, line 118: Maybe you're already aware of this but it feels like we could have something like a `ModelMap` or `ComponentMap` base class that both `SimulationMap` and `SimulationDataMap` inherit from, and all these validators won't have to be repeated.

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,275 @@
...
+    Args:
+        modeler: The `ModalComponentModeler` used to generate the simulations.
+        batch_data: The results obtained from running the simulation `Batch`.
+
+    Returns:
+        A `ModalComponentModelerData` object containing the results mapped to
+        their respective ports.
+    """
+    port_simulation_data = compose_simulation_data_index(port_task_map)
+    return ModalComponentModelerData(modeler=modeler, data=port_simulation_data)
```
momchil-flex: As far as I can tell we can remove these three functions (which are currently broken).

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,275 @@
...
+    """Selects the correct composer based on the modeler type and creates the data object.
+
+    This method acts as a dispatcher, inspecting the type of `modeler` to determine
+    which composer function (`compose_component_modeler_data` or
+    `compose_terminal_modeler_data`) to invoke.
+
+    Args:
+        modeler: The component modeler, which can be either a `ModalComponentModeler` or
+            a `TerminalComponentModeler`.
+        batch_data: The results obtained from running the simulation `Batch`.
```
momchil-flex: The docstring needs to be updated a bit. First of all, we use

```
        Parameters
        ----------
```

not `Args:`. Second, there's no `batch_data` input. And lastly, we don't call a composer function (`compose_component_modeler_data` or `compose_terminal_modeler_data`) but just the class constructor directly.

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,275 @@
...
+def compose_modeler_data(
+    modeler: ModalComponentModeler | TerminalComponentModeler,
+    indexed_sim_data: SimulationDataMap,
+) -> ComponentModelerDataType:
+    """Selects the correct composer based on the modeler type and creates the data object.
+
+    Returns:
+        The appropriate `ComponentModelerDataType` object containing the simulation results.
+
+    Raises:
```
momchil-flex: Here Parameters are missing, and same comment about docstring formatting here and all other functions in this module.

tidy3d/plugins/smatrix/run.py
daquinteroflex: We need to wait for https://github.com/flexcompute/tidy3d/pull/2755

tidy3d/components/index.py
daquinteroflex: Good idea doing this

tidy3d/plugins/smatrix/run.py
daquinteroflex: Also https://github.com/flexcompute/tidy3d/pull/2765

tidy3d/components/index.py
daquinteroflex: Done

tidy3d/plugins/smatrix/run.py
daquinteroflex: Is this mainly for docs you suggest? Google docstring style is args https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html and sphinx will also understand the traditional numpy style https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html too

I've actually realised we have not a "formal policy" of which style to use. Let me do a quick check but maybe you're right we use numpy style primarily.

In any case sphinx will understand both, but we should be consistent to choose only one across the codebase it's true

daquinteroflex: You're right we primarily use numpy style so I'll update and I'll add a predominant note

tidy3d/plugins/smatrix/run.py
daquinteroflex: Fixed

tidy3d/plugins/smatrix/run.py
daquinteroflex: Fixed

tidy3d/web/api/webapi.py
daquinteroflex: We've updated this

tidy3d/plugins/smatrix/data/terminal.py
```diff
@@ -1,28 +1,103 @@
...
+        """Stores the computed S-matrix and reference impedances
+        for the terminal ports"""
+        from tidy3d.plugins.smatrix.analysis.terminal import terminal_construct_smatrix
+
+        terminal_port_data = terminal_construct_smatrix(
+            modeler_data=self,
+            assume_ideal_excitation=assume_ideal_excitation,
+            s_param_def=s_param_def,
+        )
+        smatrix_data = MicrowaveSMatrixData(data=terminal_port_data)
```
dmarek-flex: could you add the port reference impedance array to `MicrowaveSMatrixData` at this point as well?

tidy3d/plugins/smatrix/data/terminal.py
```diff
@@ -32,29 +107,152 @@ def _warn_rf_license(cls, values):
         )
         return values
 
+    def _monitor_data_at_port_amplitude(
```
dmarek-flex, line 144: move this helper method to the analysis module

dmarek-flex: hmm maybe not worth the effort at this stage, but I think this method could be made much simpler in the future. Maybe add a todo to clean up this function to use the `self.data` field more directly

dmarek-flex (approved):
Looks good to me, just a few comments

tidy3d/plugins/smatrix/data/modal.py
```diff
@@ -0,0 +1,33 @@
...
+
+import pydantic.v1 as pd
+
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.components.data.index import SimulationDataMap
+from tidy3d.plugins.smatrix.component_modelers.modal import ModalComponentModeler
+from tidy3d.plugins.smatrix.data.data_array import ModalPortDataArray
+
+
+class ModalComponentModelerData(Tidy3dBaseModel):
```
dbochkov-flexcompute, line 13: Missing docstrings?

tidy3d/plugins/smatrix/data/modal.py
```diff
@@ -0,0 +1,33 @@
...
+    modeler: ModalComponentModeler = pd.Field(
+        ...,
+        title="ModalComponentModeler",
+        description="The original :class:`ModalComponentModeler` object that defines the simulation setup "
+        "and from which this data was generated.",
+    )
+
+    data: SimulationDataMap = pd.Field(
+        ...,
+        title="ModalComponentModeler",
```
dbochkov-flexcompute: Wrong title and description

tidy3d/plugins/smatrix/data/modal.py
```diff
@@ -0,0 +1,33 @@
...
+        description="The original :class:`ModalComponentModeler` object that defines the simulation setup "
+        "and from which this data was generated.",
+    )
+
+    def smatrix(self) -> ModalPortDataArray:
+        "Stores the computed S-matrix and reference impedances for the terminal ports"
```
dbochkov-flexcompute: No impedance here,  right?

tidy3d/plugins/smatrix/data/terminal.py
daquinteroflex: Yeah not too tricky actually

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,333 @@
...
+    Returns
+    -------
+    TerminalComponentModelerData
+        An object containing the results mapped to their respective ports.
+    """
+    port_simulation_data = compose_simulation_data_index(port_task_map)
+    return TerminalComponentModelerData(modeler=modeler, data=port_simulation_data)
+
+
+def compose_component_modeler_data(
```
dbochkov-flexcompute: Should this be compose_modal_modeler_data for consistency?

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,333 @@
...
+    TerminalComponentModelerData
+        An object containing the results mapped to their respective ports.
+    """
+    ports = [modeler.get_task_name(port=port_i) for port_i in modeler.ports]
+    data = [batch_data[modeler.get_task_name(port=port_i)] for port_i in modeler.ports]
+    port_simulation_data = SimulationDataMap(keys=tuple(ports), values=tuple(data))
+    return TerminalComponentModelerData(modeler=modeler, data=port_simulation_data)
+
+
+def compose_component_modeler_data_from_batch_data(
```
dbochkov-flexcompute: Similarly here?

tidy3d/web/core/task_core.py
```diff
@@ -706,3 +734,171 @@ def validate_post_upload(self, parent_tasks: Optional[list[str]] = None):
 
             except Exception as e:
                 raise WebError(f"Provided 'parent_tasks' failed validation: {e!s}") from e
+
+
+class BatchTask:
```
dbochkov-flexcompute, line 741: Methods of the class are missing docstrings

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,333 @@
...
+        simulations=modeler.sim_dict,
+        parent_tasks=parent_task_dict,
+        group_ids=group_id_dict,
+        **kwargs,
+    )
+    batch.to_file(filepath)
+    return batch
+
+
+def run(
```
dbochkov-flexcompute, line 306: Is this a non-gui compatible alternative to web.run?

dbochkov-flexcompute (commented):
Impressive refactoring, nothing much to add

tidy3d/plugins/smatrix/run.py
daquinteroflex: So we agreed to expose them in the migraton docs but not part of the smatrix module

tidy3d/components/index.py
daquinteroflex: Done

EDIT something went weird on tests

tidy3d/plugins/smatrix/data/modal.py
daquinteroflex: Fixed

tidy3d/plugins/smatrix/data/modal.py
daquinteroflex: Clarified

tidy3d/plugins/smatrix/run.py
daquinteroflex: Done consistently now

tidy3d/plugins/smatrix/run.py
daquinteroflex: Done consistently now

tidy3d/plugins/smatrix/run.py
daquinteroflex: So we intend to provide a `migration` section https://github.com/flexcompute/tidy3d/pull/2765/files#diff-177a553ebbc3e20960fd53f536b9644eb55cf347ff1282f5f0504d357fbb200eR105 that expose these standalone functions to the users

tidy3d/web/api/container.py
daquinteroflex: Should be fixed in https://github.com/flexcompute/tidy3d/pull/2745

tidy3d/web/api/webapi.py
daquinteroflex: We could write epic poetry on how we tried to avoid doing it this way, but alas... Reorganised https://magicloud.larksuite.com/wiki/NUiOwqr4niXR9pkPcJ2uut9Lsge to have a clear narrative of how this ended this way and us know what should be implemented for this

tidy3d/plugins/smatrix/data/terminal.py
daquinteroflex: So I was looking into this earlier and forgot to send. What exactly would you like here?

```
impedances: PortDataArray
```

or 

```
def port_reference_impedances() -> be the same as the analysis/terminal.py method
```

Just becuase https://github.com/flexcompute/tidy3d/blob/d8f336e6f82270f1b0bcb768cda2791a37e31814/tidy3d/plugins/smatrix/analysis/terminal.py#L108 is a function of modeler as well so it needs to be passed a parameter which is fine, but just checking

tidy3d/plugins/smatrix/data/modal.py
daquinteroflex: fixed now

tidy3d/plugins/smatrix/data/terminal.py
daquinteroflex: Feel free to add into another PR if you like, we're doing the refactor for a reason

tidy3d/plugins/smatrix/data/terminal.py
dmarek-flex: This is what I had in mind. Just so that the smatrix_data has everything it needs to be used elsewhere

```
smatrix_data = MicrowaveSMatrixData(
            data=terminal_port_data,
            port_reference_impedances=self.port_reference_impedances,
            s_param_def=s_param_def,
        )
```

daquinteroflex: Ah great then this is straightforward

daquinteroflex: Done

daquinteroflex:
Time for the greptile barrage

schemas/TerminalComponentModeler.json
```diff
@@ -1,6 +1,6 @@
 {
```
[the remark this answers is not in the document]
daquinteroflex: Avoid reviewing JSON files

tidy3d/plugins/smatrix/data/terminal.py
dmarek-flex: I added this to my PR

tidy3d/plugins/smatrix/utils.py
```diff
@@ -0,0 +1,306 @@
...
+    """Helper for creating the proper voltage array type."""
+    cls = CurrentFreqDataArray
+    if "t" in result.coords:
+        cls = CurrentTimeDataArray
+    if "f" in result.coords and "mode_index" in result.coords:
+        cls = CurrentFreqModeDataArray
+    return cls.assign_data_attrs(cls(data=result.data, coords=result.coords))
+
+
+def _make_current_data_array(result: DataArray) -> VoltageIntegralResultTypes:
```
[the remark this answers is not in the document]
dmarek-flex: fixed this in my PR

tidy3d/plugins/microwave/path_integrals.py
```diff
@@ -103,6 +101,9 @@ class AxisAlignedPathIntegral(AbstractAxesRH, Box):
 
     def compute_integral(self, scalar_field: EMScalarFieldType) -> IntegralResultTypes:
         """Computes the defined integral given the input ``scalar_field``."""
+        from tidy3d.plugins.smatrix.utils import (
+            _make_base_result_data_array,
+        )
```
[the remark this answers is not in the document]
dmarek-flex: fixed

tidy3d/plugins/smatrix/data/modal.py
```diff
@@ -0,0 +1,48 @@
+"""Tool for generating an S matrix automatically from a Tidy3d simulation and lumped port definitions."""
```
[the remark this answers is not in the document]
dmarek-flex: fixed

tidy3d/plugins/smatrix/utils.py
```diff
@@ -0,0 +1,306 @@
...
+from tidy3d.components.types import ArrayFloat1D
+from tidy3d.exceptions import Tidy3dError
+from tidy3d.plugins.smatrix.data.data_array import PortDataArray, TerminalPortDataArray
+from tidy3d.plugins.smatrix.network import SParamDef
+from tidy3d.plugins.smatrix.ports.types import LumpedPortType, TerminalPortType
+
+
+def port_array_inv(matrix: DataArray):
+    """Helper to invert a port matrix."""
+    return np.linalg.inv(matrix)
```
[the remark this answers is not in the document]
dmarek-flex: fixed

tidy3d/plugins/smatrix/utils.py
```diff
@@ -0,0 +1,306 @@
...
+    """Helper for creating the proper base result type."""
+    cls = FreqDataArray
+    if "t" in result.coords:
+        cls = TimeDataArray
+    if "f" in result.coords and "mode_index" in result.coords:
+        cls = FreqModeDataArray
+    return cls.assign_data_attrs(cls(data=result.data, coords=result.coords))
+
+
+def _make_voltage_data_array(result: DataArray) -> CurrentIntegralResultTypes:
```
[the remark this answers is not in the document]
dmarek-flex: fixed

tidy3d/components/index.py
```diff
@@ -0,0 +1,205 @@
...
+    ...     monitors=[
+    ...         FieldMonitor(
+    ...             size=(1, 1, 0),
+    ...             center=(0, 0, 0),
+    ...             freqs=[2e14],
+    ...             name='field'
+    ...         ),
+    ...     ],
+    ...     run_time=1e-12,
+    ...     boundary_spec=BoundarySpec.all_sides(boundary=PML()),
```
weiliangjin2021, line 182: we don't need `boundary_spec`, `monitors` and `grid_spec` in the minimal definition?

tidy3d/components/index.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+
+    keys_tuple: tuple[str, ...] = pd.Field(
+        description="A tuple of unique string identifiers for each simulation.", alias="keys"
+    )
+    values_tuple: tuple[Any, ...] = pd.Field(
+        description=(
+            "A tuple of `Simulation` objects, each corresponding to a key at the same index."
+        ),
+        alias="values",
```
weiliangjin2021, line 44: with `alias`, i see that you are using them interchangable: docstring and validator are using the alias. What's the rule of thumb here for developer and user?

tidy3d/components/index.py
```diff
@@ -0,0 +1,205 @@
...
+        """Returns the number of key-value pairs stored in the map.
+
+        This allows for using the built-in `len()` function, e.g., `len(my_map)`.
+
+        Returns
+        -------
+        int
+            The total number of items in the map.
+        """
+        assert len(self.keys_tuple) == len(self.values_tuple), "Internal state mismatch."
```
weiliangjin2021: it's validated already, so not needed?

tidy3d/components/index.py
```diff
@@ -0,0 +1,205 @@
...
+        return len(self.keys_tuple)
+
+
+class SimulationMap(ValueMap, Mapping[str, SimulationType]):
```
weiliangjin2021, line 121: Glad that we finally have this! But I don't find it used anywhere. Maybe modeler.sim_dict should return SimulationMap rather than `dict[str, Simulation]`?

docs/api/plugins/smatrix.rst
```diff
@@ -7,7 +7,7 @@ Scattering Matrix Calculator
    :toctree: ../_autosummary/
    :template: module.rst
 
-   tidy3d.plugins.smatrix.ComponentModeler
+   tidy3d.plugins.smatrix.ModalComponentModeler
```
weiliangjin2021, line 10: should run function be included here in the doc?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -1,47 +1,42 @@
...
 from tidy3d.components.viz import add_ax_if_none, equal_aspect
 from tidy3d.constants import C_0, OHM
-from tidy3d.exceptions import SetupError, Tidy3dError, Tidy3dKeyError, ValidationError
+from tidy3d.exceptions import SetupError, Tidy3dKeyError, ValidationError
 from tidy3d.log import log
-from tidy3d.plugins.smatrix.data.terminal import PortDataArray, TerminalPortDataArray
+from tidy3d.plugins.smatrix.component_modelers.base import (
+    FWIDTH_FRAC,
+    AbstractComponentModeler,
+    TerminalPortType,
```
weiliangjin2021: clearer if import from the initial place (types.py)?

tidy3d/plugins/smatrix/ports/types.py
```diff
@@ -0,0 +1,19 @@
...
+    VoltageFreqDataArray,
+    VoltageFreqModeDataArray,
+)
+from tidy3d.plugins.smatrix.ports.coaxial_lumped import CoaxialLumpedPort
+from tidy3d.plugins.smatrix.ports.rectangular_lumped import LumpedPort
+from tidy3d.plugins.smatrix.ports.wave import WavePort
+
+LumpedPortType = Union[LumpedPort, CoaxialLumpedPort]
+TerminalPortType = Union[LumpedPortType, WavePort]
+PortReferenceType = Union[str, TerminalPortType]  # TODO Debate this
```
weiliangjin2021: where is it used?

tidy3d/plugins/smatrix/data/terminal.py
```diff
@@ -1,55 +1,137 @@
...
+        "and from which this data was generated.",
+    )
+
+    data: SimulationDataMap = pd.Field(
+        ...,
+        title="Port-Simulation Data Map",
+        description="A read-only dictionary that maps simulation data to each microwave port name",
+    )
+
+    log: str = pd.Field(
```
weiliangjin2021, line 96: Should `ModalComponentModelerData` have this field too?

weiliangjin2021 (approved):
Wow, this is massive amount of effort! The flow looks good. A few minor comments and questions:

- How do we decide which function goes to analysis folder and utils.py?
- To make sure I understand correctly, now we'll need two lines to get smatrix: run and then compute smatrix from the data method?

tidy3d/components/index.py
```diff
@@ -0,0 +1,205 @@
...
+    ... )
+    >>>
+    >>> # Access a simulation like a dictionary
+    >>> # print(simulation_map["sim_1"])
+    """
+
+    keys_tuple: tuple[str, ...] = pd.Field(
+        description="A tuple of unique string identifiers for each simulation.", alias="keys"
+    )
+    values_tuple: tuple[SimulationType, ...] = pd.Field(
```
yaugenst-flex, line 199: I'm wondering if this `ValueMap` abstraction makes sense. I'd think for the `SimulationMap` you'd actually want to give meaningful names to the fields (not just `keys` and `values`). Do we subclass `ValueMap` elsewhere too?

tidy3d/components/index.py
yaugenst-flex: Unnecessary guard, that's what the validator is for.

tidy3d/components/index.py
daquinteroflex: So Momchil had suggested https://github.com/flexcompute/tidy3d/pull/2738#discussion_r2294060666 to share the validators and querying methods between SimulationMap and SimulationDataMap

In pydantic v2 we can have multiple aliases that can sort this issue out because right now the fields can only be aliased to keys and values

tidy3d/plugins/smatrix/data/terminal.py
daquinteroflex: You are correct

tidy3d/plugins/smatrix/ports/types.py
daquinteroflex: Removed it, originally it was so we could index simulations using port objects, but eventually just moved to names in `SimulationMap`

daquinteroflex: We can easily extend this in the future if we want

tidy3d/plugins/smatrix/component_modelers/terminal.py
daquinteroflex: Good catch!

docs/api/plugins/smatrix.rst
daquinteroflex: So we intend to provide a migration section https://github.com/flexcompute/tidy3d/pull/2765/files#diff-177a553ebbc3e20960fd53f536b9644eb55cf347ff1282f5f0504d357fbb200eR105 that expose these standalone functions to the users

tidy3d/components/index.py
daquinteroflex: Yes, for users the usage will be the same way because of the mapping but it should make our lives a lot easier if we do this. Let me try and see if it breaks tests

tidy3d/components/index.py
daquinteroflex: Correct

tidy3d/components/types/simulation.py
```diff
@@ -0,0 +1,37 @@
...
+
+SimulationType = Union[
+    Simulation,
+    HeatChargeSimulation,
+    HeatSimulation,
+    EMESimulation,
+    ModeSolver,
+    ModeSimulation,
+    VolumeMesher,
+]
```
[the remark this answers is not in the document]
daquinteroflex: This again was caused because we should reconsider consolidated domain scoping in my opinion. I've fixed this and the resulting circular import issues that were caused.

tidy3d/components/types/simulation.py
```diff
@@ -0,0 +1,37 @@
...
+SimulationDataType = Union[
+    SimulationData,
+    HeatChargeSimulationData,
+    HeatSimulationData,
+    EMESimulationData,
+    ModeSolverData,
+    ModeSimulationData,
+]
```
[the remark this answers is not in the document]
daquinteroflex: This again was caused because we should reconsider consolidated domain scoping in my opinion. I've fixed this and the resulting circular import issues that were caused.

tests/test_plugins/smatrix/test_terminal_component_modeler.py
```diff
@@ -17,41 +21,55 @@
...
+        "compute_F",
         lambda Z_numpy, s_param_def: 1.0 / (2.0 * np.sqrt(np.abs(Z_numpy) + 1e-4)),
     )
     monkeypatch.setattr(
-        TerminalComponentModeler,
-        "_check_port_impedance_sign",
-        lambda self, Z_numpy: (),
+        td.plugins.smatrix.analysis.terminal,
+        "check_port_impedance_sign",
+        lambda Z_numpy: np.ndarray([]),
```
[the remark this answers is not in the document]
daquinteroflex: This is correct

tidy3d/components/index.py
daquinteroflex: Done for all sim_dicts

tidy3d/components/index.py
```diff
@@ -0,0 +1,205 @@
...
+
+        Raises
+        ------
+        KeyError
+            If no simulation with the given key name is found in the map.
+        """
+        for i, current_key in enumerate(self.keys_tuple):
+            if current_key == key:
+                return self.values_tuple[i]
+        raise KeyError(f"Key '{key}' not found in the SimulationMap.")
```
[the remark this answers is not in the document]
daquinteroflex: Fixed with {self.type}

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+    ports = [modeler.get_task_name(port=port_i) for port_i in modeler.ports]
+    data = [batch_data[modeler.get_task_name(port=port_i)] for port_i in modeler.ports]
+    port_simulation_data = SimulationDataMap(keys=tuple(ports), values=tuple(data))
+    return TerminalComponentModelerData(modeler=modeler, data=port_simulation_data)
+
+
+def compose_modal_modeler_data_from_batch_data(
+    modeler: ModalComponentModeler,
+    batch_data: Optional[BatchData] = None,
```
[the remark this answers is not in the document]
daquinteroflex: Done

tidy3d/components/index.py
daquinteroflex: Maybe we can continue discussing in https://github.com/flexcompute/tidy3d/pull/2738#discussion_r2305678509

In the future we can have many aliases or key values names. For now, it seems stable to use key, values externally as that is unlikely to not be supported in the future

tidy3d/plugins/smatrix/data/data_array.py
```diff
@@ -0,0 +1,86 @@
...
+    @pd.root_validator(pre=False)
+    def _warn_rf_license(cls, values):
+        log.warning(
+            "ℹ️ ⚠️ RF simulations are subject to new license requirements in the future. You have instantiated at least one RF-specific component.",
+            log_once=True,
+        )
+        return values
+
+
+class ModalPortDataArray(DataArray):
```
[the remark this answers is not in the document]
daquinteroflex: It does not need it

tidy3d/plugins/smatrix/analysis/antenna.py
```diff
@@ -0,0 +1,141 @@
...
+        if amplitude is None:
+            # No scaling performed when amplitude is None
+            scaled_directivity_data = sim_data_port[rad_mon.name]
+            scale_factor = 1.0
+        else:
+            scaled_directivity_data = (
+                terminal_component_modeler_data._monitor_data_at_port_amplitude(
+                    port, sim_data_port, radiation_data, amplitude
+                )
+            )
```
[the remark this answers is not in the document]
daquinteroflex: Yes I agree let's fix this

tidy3d/components/index.py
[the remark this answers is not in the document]
daquinteroflex: Fixed

tidy3d/plugins/smatrix/analysis/antenna.py
```diff
@@ -0,0 +1,141 @@
...
+            dtype=complex,
+        ),
+        coords=coords,
+    )
+    b_sum = a_sum.copy()
+    # Retrieve associated simulation data
+    combined_directivity_data = None
+    for port, amplitude in port_dict.items():
+        if amplitude == 0.0:
+            continue
```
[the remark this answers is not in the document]
daquinteroflex: Fixed

tidy3d/plugins/smatrix/analysis/terminal.py
```diff
@@ -0,0 +1,279 @@
...
+        I_matrix = I_matrix._with_updated_data(data=I_out.data, coords=indexer)
+
+    V_numpy = V_matrix.values
+    I_numpy = I_matrix.values
+    Z_numpy = port_reference_impedances.values
+
+    # Check to make sure sign is consistent for all impedance values
+    check_port_impedance_sign(Z_numpy)
+
+    # # Check for negative real part of port impedance and flip the V and Z signs accordingly
```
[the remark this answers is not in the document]
daquinteroflex: Fixed

tests/utils.py
```diff
@@ -1419,6 +1419,36 @@ def make_field_projection_kspace_data(
...
+            symmetry=(0, 0, 0),
+            symmetry_center=simulation.center,
+            grid_expanded=grid,
+            **field_cmps,
+        )
+
+    def make_flux_time_data(monitor: td.FluxTimeMonitor) -> td.FluxTimeData:
+        """make a random ModeData from a ModeMonitor."""
+
+        coords = {"t": [0, 1, 2]}
```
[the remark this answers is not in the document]
daquinteroflex: doesnt matter

tidy3d/plugins/smatrix/run.py
```diff
@@ -0,0 +1,333 @@
...
+def compose_simulation_data_index(port_task_map: dict[str, str]) -> SimulationDataMap:
+    port_data_dict = {}
+    for _, _ in port_task_map.items():
+        pass
+        # FIXME: get simulationdata for each port
+        # port_data_dict[port] = sim_data_i
+
+    return SimulationDataMap(
+        keys=tuple(port_data_dict.keys()), values=tuple(port_data_dict.values())
+    )
```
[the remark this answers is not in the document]
daquinteroflex: In pending list in https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/plugins/smatrix/ports/types.py
daquinteroflex: I've added this to https://github.com/flexcompute/tidy3d/issues/2773

docs/api/plugins/smatrix.rst
daquinteroflex: I've added this to https://github.com/flexcompute/tidy3d/issues/2773 to fix later in the next PR

tidy3d/components/index.py
daquinteroflex: Added to revisit in https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/web/api/container.py
daquinteroflex: Added to follow up on https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/web/api/webapi.py
daquinteroflex: Added to follow up in https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/plugins/smatrix/run.py
daquinteroflex: Added to follow up https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/plugins/smatrix/run.py
daquinteroflex: Follow up in https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/plugins/smatrix/data/terminal.py
```diff
@@ -58,3 +136,153 @@ def _warn_rf_license(cls, values):
...
+
+    # Mirror Utils
+    # So they can be reused elsewhere without a class reimport
+    ab_to_s = ab_to_s
+    compute_F = compute_F
+    check_port_impedance_sign = check_port_impedance_sign
+    compute_port_VI = compute_port_VI
+    compute_power_wave_amplitudes = compute_power_wave_amplitudes
+    compute_power_delivered_by_port = compute_power_delivered_by_port
+    s_to_z = s_to_z
```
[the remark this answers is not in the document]
daquinteroflex: Will be fixed in docs PR but trackng https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/plugins/smatrix/component_modelers/modal.py
```diff
@@ -104,7 +139,25 @@ def to_monitor(self, port: Port) -> ModeMonitor:
     def to_source(
         self, port: Port, mode_index: int, num_freqs: int = 1, **kwargs
     ) -> list[ModeSource]:
```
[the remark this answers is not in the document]
daquinteroflex: Legacy will look into later

tidy3d/web/core/task_core.py
daquinteroflex: Done

tidy3d/web/core/task_info.py
```diff
@@ -167,3 +167,78 @@ def display(self):
...
+    status: str = None
+    sweepData: str = None
+    validateInfo: str = None
+    replaceData: str = None
+    protocolVersion: str = None
+    variable: str = None
+    createdAt: Optional[datetime] = None
+    updatedAt: Optional[datetime] = None
+    denormalizeStatus: str = None
+    summary: dict = None
```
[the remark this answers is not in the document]
daquinteroflex: tracked https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/web/core/task_info.py
```diff
@@ -167,3 +167,78 @@ def display(self):
...
+
+
+class BatchTaskBlockInfo(TaskBlockInfo):
+    """Extended block info for batch detail."""
+
+    accountLimit: float = None
+    taskBlockMsg: str = None
+    taskBlockType: str = None
+    blockStatus: str = None
+    taskStatus: str = None
```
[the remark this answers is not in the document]
daquinteroflex: https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/web/api/webapi.py
```diff
@@ -652,7 +816,47 @@ def download(
...
+                status_str = status.value
+                if status_str in {"error", "diverged", "blocked", "aborted", "aborting"}:
+                    raise WebError(
+                        f"Batch task {task_id} failed during postprocess: {status_str}"
+                    ) from None
+                if total > 0 and post_succ >= total:
+                    break
+                time.sleep(REFRESH_TIME)
+            if not _download_cm():
+                raise WebError("Failed to download 'cm_data' after postprocess completion.")
```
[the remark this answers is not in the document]
daquinteroflex: https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/components/index.py
daquinteroflex: Will investigate further in https://github.com/flexcompute/tidy3d/issues/2773 as the api is broadly stable independent of this

daquinteroflex:
> How do we decide which function goes to analysis folder and utils.py?

In a way it can be noted by the "types" that the function accepts. The idea of the refactor is to have a more structured extensible flow for complex workflows. In a way the `get_antenna_params` function illustrates this. It goes from a set of `[simulation, modeler data, port] -> device metrics` so its almost an information scope transformation. It's separated this way because you could conceive that this function could be useful as a static method of `AntennaMetricsData` for example, but our relationships are only going to become more complex as functionality increases, particularly true for RF nonlinear relationships, so we need to start becoming more modular in our software architecture.

We can also concieve other previous class-methods as really pure operations onto the state of a given class. In a way, this is what `TerminalComponentModeler` does with `port_reference_impedances` for example. It's a postprocessing operation onto the data stored in `TerminalComponentModelerData` now, but with a clear separation of the immutable state.

So ultimately, I'd suggest, going forward with RF we need to have a clear separation of the "schema / information containers" and the "operations / functions / methods" onto those containers. This should both make the API design a lot easier, and also make it more extensible and powerful moving onwards.

You could argue utils functions have a local scope limited to either the function itself, or used across multiple scopes so they should be standalone functions without any relation to the data containers or higher level functions they get used in.

>To make sure I understand correctly, now we'll need two lines to get smatrix: run and then compute smatrix from the data method?

Yep exactly! Again this conceptual separation between operations and data containers / schema.

tidy3d/plugins/smatrix/data/terminal.py
daquinteroflex: Lets do this later before prerelease https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/components/index.py
daquinteroflex: I'd suggest to move forward let's keep it like this

CHANGELOG.md
daquinteroflex: So I've fixed the old changes, but need to write a little overview of changes

daquinteroflex:
I've listed all pending items https://github.com/flexcompute/tidy3d/issues/2773 but this should be stable enough to merge to develop and do the RF GUI solver upgrade

tidy3d/components/index.py
daquinteroflex: Kept getting a bunch of validation errors, but this one works. Maybe we can come up with a few really small examples for the future

daquinteroflex: Added to revisit in https://github.com/flexcompute/tidy3d/issues/2773

tidy3d/web/api/container.py
daquinteroflex: Should be internal, the whole webapi needs a cleanup

daquinteroflex: Mainly was introduced bcauase a tidy3d/projects/rf was being considered at the time

daquinteroflex: We can bring the old change back but I don't necessarily consider it considerably better either. We've tested this so should be fine. Since we know this can change, maybe handy for the future as we know it works

daquinteroflex: Can revisit in https://github.com/flexcompute/tidy3d/issues/2773 too

momchil-flex (approved):

tidy3d/plugins/smatrix/data/terminal.py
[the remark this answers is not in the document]
daquinteroflex: Fixed by default turns out

PR #2740, opened 2025-08-19
poc: adjoint surface monitors

yaugenst-flex:


tidy3d/components/data/dataset.py
```diff
@@ -675,6 +678,71 @@ class TriangleMeshDataset(Dataset):
...
+        None,
+        title="Tangent 1",
+        description="Optional first tangent vector per sample; orthonormal to the normal.",
+    )
+    perp2: SurfaceVectorDataArray = pd.Field(
+        None,
+        title="Tangent 2",
+        description="Optional second tangent vector; completes an orthonormal basis.",
+    )
+
```
groberts-flex, line 710: we may want distance from points to the geometry edge - this would be useful for 2D PEC singularity correction in the gradient

tidy3d/components/data/dataset.py
```diff
@@ -675,6 +678,71 @@ class TriangleMeshDataset(Dataset):
...
+    """Geometry surface samples and associated geometric metadata.
+
+    - ``points``: Cartesian coordinates of each surface sample.
+    - ``normals``: Outward-pointing unit normals at each sample.
+    - ``perp1`` and ``perp2``: Optional orthonormal tangents spanning the local tangent plane.
+      If not provided, we can construct a consistent basis from ``normals``.
+    - ``weights``: Quadrature weights (e.g., differential area per sample).
+    """
+
+    points: SurfaceVectorDataArray = pd.Field(
```
groberts-flex, line 691: ultimately, to connect to the gradient, we will need to associate each of these points with the geometric parameter we are getting the gradient for. For PolySlab, for example, we will need a good way to sort these points as belonging to different edges for example. I wonder if possibly, we want the surface sample points broken down by edge. This would work well, I think, for Box and PolySlab, but it may not be as straightforward if we had something like a level set geometry in the future. See below as well - I think we may need each geometric edge to correspond to a list of interface edges where each interface edge is a continuous run of inside and outside materials being the same combination.

tidy3d/components/data/dataset.py
```diff
@@ -675,6 +678,71 @@ class TriangleMeshDataset(Dataset):
...
+    )
+    Et2: SurfaceScalarDataArray = pd.Field(
+        None, title="Et2", description="Second tangential component of E at samples."
+    )
+    Dn: SurfaceScalarDataArray = pd.Field(
+        None, title="Dn", description="Normal component of D at samples."
+    )
+
+
+class AdjointPECSurfaceComponents(Dataset):
```
groberts-flex, line 729: LossyMetal would look similar to this although we would include the Et1 and Et2 as well.

tidy3d/components/data/dataset.py
```diff
@@ -675,6 +678,71 @@ class TriangleMeshDataset(Dataset):
...
+        description="Optional first tangent vector per sample; orthonormal to the normal.",
+    )
+    perp2: SurfaceVectorDataArray = pd.Field(
+        None,
+        title="Tangent 2",
+        description="Optional second tangent vector; completes an orthonormal basis.",
+    )
+
+
+class AdjointDielectricSurfaceComponents(Dataset):
```
groberts-flex, line 712: If we want to support a more flexible integration scheme, instead of having surface components for a single type of material like this, we would want to have surface components specific to each type of interface (dielectric <-> dielectric, PEC <-> dielectric, lossy metal <-> dielectric, PEC <-> PEC (this one so we would know to ignore the integration here)). This would allow a dielectric geometry that was inside a PEC background to be integrated in the adjoint instead of assuming there was dielectric outside of it for example. With overlapping structures, we could potentially have multiple of these types of interfaces for a single geometry and each geometric edge could have different interfaces on it. So I would think that we could have something like

`AdjointDielectricDielectricSurfaceComponents`, `AdjointDielectricPECSurfaceComponents`, etc.

and then the samples would be for a single edge where an edge is a run of continuous interface. This edge maps to a geometric edge (i.e. - one of the sides of a `Box` or between two vertices in a `PolySlab`). Then, each geometric edge has a list of these interface edges each with field components corresponding to that type of interface. When we integrate each geometric edge, we can integrate/sum each interface edge in the list and they will all operate under their specific integration rule.

tidy3d/components/monitor.py
```diff
@@ -1544,6 +1543,57 @@ def _storage_size_solver(self, num_cells: int, tmesh: ArrayFloat1D) -> int:
         return BYTES_COMPLEX * num_cells * len(self.freqs) * 6
 
 
+class AdjointSurfaceMonitor(FreqMonitor):
```
groberts-flex, line 1546: See above - but I think a general `AdjointSurfaceMonitor` may make sense and the data returned will have all the information on what type of interface it is (dielectric <-> dielectric, dielectric <-> PEC, etc.)

PR #2745, opened 2025-08-20
autograd support for web.run(component_modeler)

yaugenst-flex:


tests/test_plugins/smatrix/test_component_modeler_autograd.py
```diff
@@ -0,0 +1,304 @@
...
+from tidy3d.plugins.smatrix.analysis import terminal as terminal_analysis
+from tidy3d.plugins.smatrix.component_modelers.modal import ComponentModeler
+from tidy3d.plugins.smatrix.component_modelers.terminal import TerminalComponentModeler
+from tidy3d.plugins.smatrix.data.data_array import TerminalPortDataArray
+from tidy3d.plugins.smatrix.ports.modal import Port as ModalPort
+from tidy3d.plugins.smatrix.ports.rectangular_lumped import LumpedPort as RectLumpedPort
+from tidy3d.web.api.autograd import autograd as web_ag
+
+
+def _run_emulated_minimal(simulation: td.Simulation, path=None, **kwargs) -> td.SimulationData:
```
tylerflex, line 19: could any of this be ported to the regular run_emulated? would be nice to have it in one place.

tidy3d/web/api/autograd/autograd.py
```diff
@@ -24,6 +24,7 @@
 from tidy3d.components.data.data_array import DataArray
 from tidy3d.components.grid.grid_spec import GridSpec
 from tidy3d.exceptions import AdjointError
+from tidy3d.plugins.smatrix.component_modelers.types import ComponentModelerType
```
tylerflex, line 27: at first glance, introducing this dependency only to support `tidy3d.web.run(ComponentModeler)` seems unnecessary to me. 

It would seem more clean to just add this extra code to `tidy3d.plugins.smatrix.web.run()` instead and call the relevant autograd compatible functions internally?

Maybe can we discuss this a bit?

tidy3d/web/api/autograd/autograd.py
```diff
@@ -54,37 +55,55 @@
...
+
+    # Simulation path
+    if isinstance(simulation, td.Simulation):
+        return _is_valid_sim(simulation)
+
+    # Component modeler path
+    if isinstance(simulation, ComponentModelerType):
+        try:
+            sims = simulation.sim_dict
+        except Exception:
```
tylerflex, line 99: when does this come up? Seems like something we should not handle silently?

tidy3d/web/api/autograd/autograd.py
```diff
@@ -54,37 +55,55 @@
...
+    Supports either a single td.Simulation or a ComponentModelerType. For a component
+    modeler, returns True if any of the underlying simulations are valid for autograd.
+    """
 
-    # if no frequency-domain data (e.g. only field time monitors), raise an error
-    if not simulation._freqs_adjoint:
-        raise AdjointError(
-            "No frequency-domain data found in simulation, but found traced structures. "
-            "For an autograd run, you must have at least one frequency-domain monitor."
+    def _is_valid_sim(sim: td.Simulation) -> bool:
```
tylerflex, line 65: one potential thing we could add is a private property `_is_valid_for_autograd()` to a few classes, such as `Simulation` and `ComponentModeler` and then just call that here. That would separate out this logic specific to the component modeler

tylerflex (commented):
See my comments. The main thing we should discuss is whether running the ComponentModeler from the regular tidy3d.web.run() is worth injecting all of this dependency of the web API on the component modeler logic. I would lean towards just making a plugins.smatrix.web to handle this stuff, and the non-autograd version can just call regular web.run().  We should discuss @yaugenst-flex / @daquinteroflex . What do we gain by using tidy3d.web.run()?

tests/test_plugins/smatrix/test_component_modeler_autograd.py
yaugenst-flex: these tests need to be reworked altogether, this is basically just in place so that we can have a sanity check that this chain is intact. the emulated run in `test_autograd.py` is super heavy and if we use it here then the component modeler test takes 10 seconds

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: yeah this can probably moved, this is basically proof of concept to show how to add this support

tylerflex: I see. yea so basically just splitting out the ComponentModeler -> sim_dict / running / batch_data -> ComponentModelerData out into their own functions and then the autograd one just does the full process client side? (no VJP hacking?). This works fine. In a sense, it's just like continuing to support the old ComponentModeler.run() functionality but wrapping it in a web.run() call?

tylerflex: I think what could be a potential hiccup is that the local autograd flow vs remote flow are switched based on whether any valid autograd simulations exist. I worry this might cause some confusion with users. eg. if they try inverse design and wonder why they can't see their results in the web UI? Another question is whether we should keep the old ComponentMOdeler.run() syntax and have it essentially be for dedicated autograd only while we work on supporting this more properly for later.

daquinteroflex:
Basically propagated latest changes from the PR branch https://github.com/flexcompute/tidy3d/pull/2755

I also updated the implementation a bit so that all web imports would be at the same file scope, as ideally was targeting no more web imports at any of the base classes and only can be done inside on-the-fly function calls to avoid the web circular import problem.

tidy3d/web/api/autograd/autograd.py
daquinteroflex: Implemented already for autograd support on modeler.run

daquinteroflex:
Just to add a future TODO we need to properly fix https://github.com/flexcompute/tidy3d/commit/d5c05558de3874e95aca3ada7b80fa7867792165

PR #2746, opened 2025-08-20
feat(adjoint): full adjoint support for dispersive models

yaugenst-flex:
Also includes a slight reorganization of the autograd tests (not exhaustive, but it's a start).

## Greptile Summary

This PR implements comprehensive adjoint (automatic differentiation) support for dispersive optical material models in Tidy3D, specifically adding gradient computation capabilities for Sellmeier, Lorentz, Drude, Debye, and PoleResidue models along with their custom spatially-varying counterparts. The implementation uses two complementary approaches: Tensor-Jacobian Products (TJP) through automatic differentiation for uniform models with scalar parameters, and analytical chain rule implementations for custom models with spatial data arrays.

The core changes are in `medium.py` where new gradient computation methods `_compute_derivatives` are added to each dispersive model class. These methods enable inverse design workflows by allowing optimization algorithms to compute gradients with respect to material parameters. The implementation includes shared helper functions for common operations, frequency-dependent weight functions for analytical derivatives, and proper handling of complex-valued parameters versus real-valued projections.

Additionally, the PR includes a reorganization of autograd tests, moving them from a flat structure into a hierarchical `autograd/` directory with numerical tests grouped under `numerical/`. New comprehensive test suites validate the VJP implementations through finite difference comparisons and monkeypatched derivative field computations to ensure mathematical correctness.

## Important Files Changed

<details>
<summary>Files Changed</summary>

| `tidy3d/components/medium.py` | 4/5 | Added comprehensive adjoint support for all dispersive models with dual TJP/analytical approaches |
| `tidy3d/components/autograd/utils.py` | 5/5 | Added utility functions for complex array handling and autograd-compatible operations |
| `tests/test_components/autograd/test_autograd_custom_dispersive_vjps.py` | 4/5 | New comprehensive test validating VJPs for custom dispersive models using monkeypatching |
| `tests/test_components/autograd/test_autograd_dispersive_vjps.py` | 4/5 | New finite difference validation tests for all dispersive model VJPs with tight tolerances |
| `tests/test_components/autograd/test_autograd.py` | 4/5 | Moved file with new custom dispersive model tests and updated import paths |
| `tests/test_components/autograd/numerical/test_autograd_conductivity_numerical.py` | 4/5 | Moved conductivity gradient validation test with some configuration concerns |
| `tests/test_components/autograd/numerical/test_autograd_numerical.py` | 5/5 | Moved core numerical validation test for autograd gradients |
| `tests/test_components/autograd/numerical/test_autograd_mode_polyslab_numerical.py` | 4/5 | Moved numerical polyslab test with some cleanup needed |
| `tests/test_components/autograd/numerical/test_autograd_periodic_numerical.py` | 4/5 | Moved periodic diffraction gradient validation test |
| `tests/test_components/autograd/numerical/test_autograd_symmetry_numerical.py` | 5/5 | Moved symmetry consistency test with minor code quality issues |
| `tests/test_components/autograd/test_autograd_polyslab.py` | 5/5 | Moved comprehensive polyslab derivative tests |
| `tests/test_components/autograd/test_autograd_rf_polyslab.py` | 5/5 | Moved RF polyslab numerical validation tests |
| `tests/test_components/autograd/test_autograd_rf_box.py` | 5/5 | Moved RF box gradient validation tests |
| `tests/test_components/autograd/__init__.py` | 5/5 | Emptied as part of test reorganization |
| `tests/test_components/autograd/numerical/__init__.py` | 5/5 | Emptied as part of test reorganization |
| `tests/test_plugins/test_invdes.py` | 5/5 | Updated import path to reflect test reorganization |
| `CHANGELOG.md` | 5/5 | Added appropriate changelog entry documenting new dispersive autograd support |

</details>

## Confidence score: 4/5

- This PR implements a significant new feature with comprehensive testing and appears mathematically sound
- Score reflects the complexity of dispersive model mathematics and potential for subtle numerical issues
- Pay close attention to the medium.py implementation and numerical test configurations


tylerflex (commented):
@yaugenst-flex I took a read through. 

I think at the end of the day, if it works it works and since we probably wont build on top of this, I'm not sure how much to stress about it.  

What kind of worries me again is the increasing complexity in all of the derivative calculation code. which I think we can address separately. 

But also in this instance there seems to be a good amount of duplication? For example, am I right in seeing that the ε(params, frequency) function is re-implemented in many of the `_compute_derivatives` methods and also in the tests themselves?  The worry is that it makes it a bit brittle. I wonder if we can instead re-use `eps_model()` functions, or do one VJP for pole residue and then re-use the `to_pole_residue()` functionality for the dispersive models to handle the derivatives automatically. Did you consider these approaches?

yaugenst-flex:
> What kind of worries me again is the increasing complexity in all of the derivative calculation code. which I think we can address separately.

I'm not sure about this. Some small repetition aside, I don't think this really introduces a lot of complexity, or at least unnecessary complexity. We are adding custom VJPs for 8 material classes here, which necessarily introduces more code. It's a +400 LOC diff, so basically 50 per VJP, which doesn't seem too bad? I agree we can probably move this elsewhere, but I don't think this would change much in terms of complexity.

> But also in this instance there seems to be a good amount of duplication? For example, am I right in seeing that the ε(params, frequency) function is re-implemented in many of the `_compute_derivatives` methods and also in the tests themselves? The worry is that it makes it a bit brittle. I wonder if we can instead re-use `eps_model()` functions

Right. I changed `ε(params, frequency)` for custom medium models to no longer inline the partials but instead introduced model-specific weights (`_w_B`, `_w_C` for Sellmeier for example) that can be reused by tests, this does cut down a lot on duplication there.

I also switched to using `eps_model(f)` for the uniform medium models, which works. I'm not sure it's necessarily better though because `eps_model()` reads parameters from `self`, i.e., it doesn't depend directly on the parameters that we're differentiating w.r.t.. The current approach uses autograd TJP to calculate the derivatives of the material models instead of hard-coding them, so for that to work the function that we're differentiating needs to depend on the parameters. I made this work using `self.updated_copy(params).eps_model()`. I don't know if this is necessarily better than to have a function that depends on the parameters directly, especially since it introduces an additional copy (which granted shouldn't really be a problem here). But basically that's what the pure helper was for.

> do one VJP for pole residue and then re-use the `to_pole_residue()` functionality for the dispersive models to handle the derivatives automatically. Did you consider these approaches?

So this works in some cases, and in some it doesn't, or at least I'm not sure if it would cause problems. For Drude and Debye this works well because they have a closed-form mapping to PoleResidue. But Lorentz and Sellmeier mappings are piecewise. For small parameter perturbations this can cause poles to be reordered, or for example switching between over-/underdamped branches. If you are in a region where this happens during the Lorentz <-> PoleResidue mapping, then I think there would be a jump in the gradient. So basically I think if we did `to_pole_residue()`, we could safely do that for Drude and Debye, but for Lorentz and Sellmeier I think the current approach is safer. And in the spirit of reducing complexity, I chose to go with the current approach because it works for all models.

That said, doing this via `to_pole_residue()` probably can work too, but I think it would need more testing to check what happens to the derivatives at those branching points.

tylerflex (approved):

groberts-flex (approved):
Looks good to me and seems well tested.  I think we can also add tests for this with simulations in the numerical test suite as well in the future.

PR #2747, opened 2025-08-20
feat(adjoint): sidewall angle gradients

yaugenst-flex:
gradients_axis_0_ref_bottom
gradients_axis_0_ref_middle
gradients_axis_0_ref_top
gradients_axis_1_ref_bottom
gradients_axis_1_ref_middle
gradients_axis_1_ref_top
gradients_axis_2_ref_bottom
gradients_axis_2_ref_middle
gradients_axis_2_ref_top

## Greptile Summary

Updated On: 2025-09-04 13:18:06 UTC

This PR implements sidewall angle gradients for automatic differentiation (adjoint optimization) in PolySlab and Cylinder geometries. The changes enable users to compute gradients with respect to sidewall angles, expanding Tidy3D's inverse design capabilities beyond existing vertex and bounds optimization.

The implementation consists of several key architectural changes:

**Core Infrastructure**: The `sidewall_angle` field in the `Planar` base class is upgraded from `float` to `TracedFloat`, enabling autograd tracking. A new `reference_axis_pos` property is added to standardize reference point calculations across different reference plane configurations.

**Gradient Computation**: A comprehensive `_compute_derivative_sidewall_angle()` method is implemented in PolySlab that calculates Vector-Jacobian Products (VJP) through surface integration over sidewall patches. The mathematical approach correctly handles the relationship between sidewall angle changes and surface normal velocities, with proper area element corrections for slanted surfaces.

**Cylinder Support**: The Cylinder class is enhanced to map sidewall angle gradients to its underlying PolySlab representation, with defensive programming to handle edge cases where derivative paths may not exist.

**Web Platform Integration**: New batch category types (`'tidy3d_autograd'`, `'tidy3d_autograd_async'`, `'autograd_fwd'`, `'autograd_bwd'`) are added to properly categorize and route different phases of autograd computations.

**Validation Compatibility**: Existing validation functions are updated to handle autograd-traced sidewall angles using `get_static()` and `getval()` functions, ensuring validation logic remains functional while preserving autograd traceability.

The feature integrates seamlessly with Tidy3D's existing adjoint framework, following established patterns for parameter tracing and gradient computation. The implementation includes extensive mathematical handling of sidewall geometry transformations and proper numerical integration techniques.

## Important Files Changed

| `tidy3d/web/api/container.py` | 5/5 | Adds new batch category types for autograd workflow classification |
| `tests/test_components/autograd/test_autograd.py` | 5/5 | Adds end-to-end test validating sidewall angle gradient propagation |
| `CHANGELOG.md` | 5/5 | Documents new autograd support for sidewall angles in Cylinder and PolySlab |
| `tidy3d/components/geometry/primitives.py` | 4/5 | Implements sidewall angle gradient mapping for Cylinder class with defensive programming |
| `tidy3d/components/geometry/base.py` | 5/5 | Upgrades sidewall_angle to TracedFloat and adds reference_axis_pos property |
| `tidy3d/components/geometry/utils.py` | 5/5 | Updates validation to handle autograd-traced sidewall angles using get_static |
| `tidy3d/components/geometry/polyslab.py` | 4/5 | Major implementation of sidewall angle gradient computation with VJP calculation |
| `tests/test_components/autograd/numerical/test_autograd_polyslab_sidewall_numerical.py` | 4/5 | New numerical validation test comparing adjoint vs finite difference gradients |
| `tests/test_components/autograd/test_autograd_polyslab_sidewall.py` | 4/5 | Comprehensive unit tests for sidewall angle gradient computation scenarios |
| `tests/test_components/autograd/test_sidewall_edge_cases.py` | 4/5 | Edge case tests covering geometric transformations and boundary conditions |

</details>

## Confidence score: 4/5

- This PR implements a complex mathematical feature with comprehensive testing, but the mathematical complexity and integration points create some risk
- Score reflects the thorough testing coverage and well-structured implementation, but lowered due to complex VJP calculations and potential for numerical precision issues
- Pay close attention to the PolySlab gradient computation method and numerical validation tests for mathematical correctness

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
    participant JaxPolySlab
    participant ModeSolver as "Mode Solver"
    participant WebAPI as "Web API"
    participant AutogradEngine as "Autograd Engine"
    participant DerivativeComputer as "Derivative Computer"

    User->>JaxPolySlab: "Create PolySlab with traced sidewall_angle"
    JaxPolySlab->>JaxPolySlab: "Validate sidewall_angle bounds"
    User->>ModeSolver: "Create simulation with JaxPolySlab"
    User->>AutogradEngine: "Define objective function"
    AutogradEngine->>WebAPI: "Run forward simulation"
    WebAPI-->>AutogradEngine: "Return SimulationData"
    AutogradEngine->>DerivativeComputer: "Compute gradient w.r.t. sidewall_angle"
    DerivativeComputer->>DerivativeComputer: "_compute_derivative_sidewall_angle()"
    DerivativeComputer->>DerivativeComputer: "_collect_sidewall_patches()"
    DerivativeComputer->>DerivativeComputer: "_adaptive_edge_samples()"
    DerivativeComputer->>DerivativeComputer: "evaluate_gradient_at_points()"
    DerivativeComputer->>DerivativeComputer: "Compute shape-derivative factors"
    DerivativeComputer-->>AutogradEngine: "Return VJP gradients"
    AutogradEngine-->>User: "Return objective value and gradients"
```

### DIFF
--- a/tidy3d/components/geometry/polyslab.py
+++ b/tidy3d/components/geometry/polyslab.py
@@ -209,7 +209,8 @@ class PolySlab(base.Planar):
         """
 
         # no need to validate anything here
-        if math.isclose(values["sidewall_angle"], 0):
+        # sidewall_angle may be autograd-traced; use static value for this check only
+        if math.isclose(getval(values["sidewall_angle"]), 0):
             return val
 
         # apply dilation
@@ -1009,7 +1010,8 @@ class PolySlab(base.Planar):
 
         # check for the maximum possible contribution from dilation/slant on each side
         max_offset = self.dilation
-        if not math.isclose(self.sidewall_angle, 0):
+        # sidewall_angle may be autograd-traced; unbox for this check
+        if not math.isclose(getval(self.sidewall_angle), 0):
             if self.reference_plane == "bottom":
                 max_offset += max(0, -self._tanq * self.finite_length_axis)
             elif self.reference_plane == "top":
@@ -1285,7 +1287,8 @@ class PolySlab(base.Planar):
             Shift along x and y direction.
         """
 
-        if math.isclose(dist, 0):
+        # 'dist' may be autograd-traced; unbox for the zero-check only
+        if math.isclose(getval(dist), 0):
             return vertices, np.zeros(vertices.shape[0], dtype=float), None
 
         def rot90(v):
@@ -1476,6 +1479,10 @@ class PolySlab(base.Planar):
                     derivative_info, sim_min, sim_max, is_2d, interpolators
                 )
 
+            elif path == ("sidewall_angle",):
+                vjps[path] = self._compute_derivative_sidewall_angle(
+                    derivative_info, sim_min, sim_max, is_2d, interpolators
+                )
             elif path[0] == "slab_bounds":
                 idx = path[1]
                 face_coord = self.slab_bounds[idx]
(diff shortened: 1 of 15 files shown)


tidy3d/components/autograd/types.py
```diff
@@ -27,9 +27,9 @@
 Box.__deepcopy__ = lambda v, memo: _deepcopy(v, memo)
 
 # Types for floats, or collections of floats that can also be autograd tracers
-TracedFloat = typing.Union[float, Box]
-TracedPositiveFloat = typing.Union[pd.PositiveFloat, Box]
-TracedSize1D = typing.Union[Size1D, Box]
+TracedFloat = typing.Union[Box, float]
```
tylerflex: curious what this is about? did the union order affect something?

tests/test_components/test_autograd.py
tylerflex, line 1: just going to trust you (or Claude?) on the tests. Also generally I had been adding every new structure type support to the `structure_keys` dictionary to run through the automated testing. but I suppose if re-hauling the autograd tests soon maybe we dont need to do that here.

tidy3d/components/base.py
tylerflex, line 1: are these changes necessary for this PR? if not we should do them separately I think

tidy3d/components/geometry/primitives.py
```diff
@@ -317,9 +317,26 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+            elif path == ("radius",):
+                ps_paths.add(("vertices",))
+            elif "center" in path:
+                _, center_index = path
+                _, (index_x, index_y) = self.pop_axis((0, 1, 2), axis=self.axis)
+                if center_index in (index_x, index_y):
+                    ps_paths.add(("vertices",))
+                else:
+                    ps_paths.update({("slab_bounds", 0), ("slab_bounds", 1)})
+            elif path == ("sidewall_angle",):
```
tylerflex, line 334: I think these two lines were the only things that needed to change here. but I suppose the new code does avoid unnecessary grad computation?

tidy3d/components/geometry/polyslab.py
```diff
@@ -209,12 +209,13 @@ def no_self_intersecting_polygon_during_extrusion(cls, val, values):
...
         # no need to validate anything here
-        if math.isclose(values["sidewall_angle"], 0):
+        angle_val = getval(values["sidewall_angle"]) if "sidewall_angle" in values else 0.0
+        if math.isclose(angle_val, 0):
             return val
 
         # apply dilation
         poly_ref = PolySlab._proper_vertices(val)
-        if not math.isclose(values["dilation"], 0):
+        if not math.isclose(getval(values["dilation"]) if "dilation" in values else 0.0, 0):
```
tylerflex: I worry that we may run into situations where we need more of these down the line. these can be quite a source of bugs.

maybe one of the tests with a traced PolySlab should involve something with dilation, erosion, or some edge cases just to check everything. (if not already)

tidy3d/components/geometry/polyslab.py
```diff
@@ -1500,6 +1510,307 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+        S, W = [], []
+        for i in range(segs):
+            a, b = edges_t[i], edges_t[i + 1]
+            S.append(
+                (0.5 * (b - a) * g_loc + 0.5 * (b + a)).astype(GRADIENT_DTYPE_FLOAT, copy=False)
+            )
+            W.append((w_loc * 0.5 * (b - a)).astype(GRADIENT_DTYPE_FLOAT, copy=False))
+        return np.concatenate(S), np.concatenate(W)
+
+    def _collect_sidewall_patches(
```
tylerflex, line 1588: these VJP calculation functions are getting really complicated. I wonder if we should move them to their own file as helper functions to not blow up polyslab.py.

tidy3d/components/geometry/polyslab.py
```diff
@@ -1500,6 +1510,307 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+        so each patch weight is w = L·dz·( -(z-z0) / cos^2θ ).
+        """
+        if interpolators is None:
+            interpolators = derivative_info.create_interpolators(dtype=GRADIENT_DTYPE_FLOAT)
+
+        # 2D sim ⇒ no dependence on θ (z_local=0)
+        if is_2d:
+            return 0.0
+
+        vertices = np.asarray(self.vertices, dtype=GRADIENT_DTYPE_FLOAT)
```
tylerflex: this logic is shared in the vertices VJP calculator. I dont know if it should just be moved into the `_collect_sidewall_patches` method? this kind of thing might come up in some other areas too.

tidy3d/components/geometry/polyslab.py
```diff
@@ -1500,6 +1510,307 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+        self,
+        derivative_info: DerivativeInfo,
+        sim_min: np.ndarray,
+        sim_max: np.ndarray,
+        is_2d: bool = False,
+        interpolators: Optional[dict] = None,
+    ) -> float:
+        """VJP for dJ/dθ where θ = sidewall_angle.
+
+        Use ∂J/∂θ = ∬_S g(x) V_n(x;θ) dA, with g(x) from
```
tylerflex: this is where we should really understand what's going on.

So the `g(z)` here is the derivative of objective `J` with respect to an outward, normal shift of the sidewall at z=z?

we want to compute `dJ/dθ = dJ/dg * dg/dθ`. integrated along z

`g(θ) = (z-z0) * tan(θ)`  so I think `dg/dθ` should be `(z-z0) /cos²(θ)` and then we multiply by the area element dA.

This area element is a bit weird in the derivation though, I dont understand

```
dA = (dz/cosθ)·dℓ,
so each patch weight is w = L·dz·( -(z-z0) / cos^2θ ).
```

what happened to the cosθ in dA? wouldn't it be `w = dA ( -(z-z0) / cos^2θ ) = - L dz (z-z0) / cos(θ)`

slightly confused here. This wouldn't affect much unless theta was substantial, but still worth understanding.

tidy3d/components/geometry/polyslab.py
```diff
@@ -1500,6 +1510,307 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+            sim_max=sim_max,
+            is_2d=False,
+            dx=dx,
+        )
+        if patch["centers"].shape[0] == 0:
+            return 0.0
+
+        cos_theta = np.cos(self.sidewall_angle)
+        inv_cos2 = 1.0 / (cos_theta * cos_theta)
+        z_mid = self.center_axis
```
tylerflex: definitely worth checking this. so as far as I know it's not the center axis necessarily, this `z_mid` or basically the hinge point for the sidewall angle, depends on the `PolySlab` field `.reference_plane`, which is middle by default but I think z_mid needs to take this into account. there should be a `PolySlab` property somewhere to return the z point that is the center of the θ rotation.

tylerflex (changes requested):
OK!  A few comments and suggestions. Thanks @yaugenst-flex

tidy3d/components/geometry/polyslab.py
yaugenst-flex: yeah good catch. i added it as a property to `Planar` and fixed the calculation

tidy3d/components/autograd/types.py
yaugenst-flex: yes it would wrongly coerce traced values to untraced ones, reordering fixes it

tidy3d/components/geometry/primitives.py
yaugenst-flex: yeah cylinder was doing a lot of extra computation so i think this is a good change?

tidy3d/components/base.py
yaugenst-flex: yeah this change was added because the current implementation was causing problems with sidewall angle grad tracking. the `dict()` conversion can drop traced vals

tidy3d/components/geometry/polyslab.py
yaugenst-flex: yeah we should but this is already planned so i think we should do that in a separate PR? this one is already quite big

tidy3d/components/geometry/polyslab.py
yaugenst-flex: good point, added some tests

tidy3d/components/geometry/polyslab.py
yaugenst-flex: > So the g(z) here is the derivative of objective J with respect to an outward, normal shift of the sidewall at z=z?

right, so `g(x) = dJ/du_n`

> `dJ/dθ = dJ/dg * dg/dθ`

so this is not quite right. `g` here is not an independent variable right? if you look at the units we want "objective per rad". but the `g` in the above expression has units of "objective per unit length" since it's the change in `J` per normal shift. so something like `dJ/dθ` if treated like above would give "rad per objective". instead, we need to multiply by the normal velocity and integrate over area. so `dJ/dθ = ∫ dJ/du_n * du_n/dθ dA`, which is what is implemented here.

> `g(θ) = (z − z0) tan θ, so dg/dθ = (z − z0) / cos^2 θ`

so this would mix up `g` with the geometric offset. `(z - z_ref) tan θ` is the in-plane radial offset that defines the sidewall shape, but it's not the same thing as `g`. the chain rule has to be applied to the normal displacement, not `g`. so it's:
```
d(z) = −(z − z_ref) tan θ
=> dd/dθ = −(z − z_ref) sec^2 θ
```

>  multiply by the area element dA (where `dA = (dz/cosθ)·dℓ`)

yup

> what happened to the cosθ in dA? wouldn't it be `w = dA ( -(z-z0) / cos^2θ ) = - L dz (z-z0) / cos(θ)`

this is in the projection to the surface normal. the normal velocity is not `dd/dθ`, the tangential change needs to be projected onto the sidewall normal so you get `V_n = (dd/dθ) · cos θ = −(z − z_ref) / cos θ`.

tidy3d/components/geometry/polyslab.py
yaugenst-flex: yeah good point. i didn't move it into `_collect_sidewall_patches` though but made a new helper instead. single responsibility and all

tidy3d/components/base.py
yaugenst-flex: i'll have another look at this though, stay tuned 😄

tidy3d/components/geometry/polyslab.py
tylerflex: sure, we can do it later

tidy3d/components/geometry/polyslab.py
tylerflex: I'm having trouble following all of this without an image to follow. but it sounds like you've double checked it so I feel satisfied enough.

tests/test_components/test_autograd_polyslab_sidewall_numerical.py
```diff
@@ -0,0 +1,81 @@
...
+
+import tidy3d as td
+import tidy3d.web as web
+
+LOCAL_GRADIENT = True
+VERBOSE = False
+
+
+@pytest.mark.numerical
+def test_polyslab_sidewall_fd_vs_adj():
```
groberts-flex: to validate a few more cases numerically, maybe this could be parameterized by angle?

groberts-flex (commented):
looks good to me, small note on testing!

tylerflex (approved):

tidy3d/components/base.py
yaugenst-flex: @tylerflex ok yeah had another look and got to the bottom of it. reverted all the base.py changes (and many of the other getval-related ones)

tests/test_components/test_autograd_polyslab_sidewall_numerical.py
yaugenst-flex: yup parametrized over a couple of angles. can't run the numerical tests currently though :/

yaugenst-flex:
@groberts-flex @tylerflex ok polished it up, i think this is ready to go. waiting for dario's merge so i can run the numerical tests but other than that i think it's done

groberts-flex (approved):
thanks for adding this! it looks good to me if numerical tests are looking good!

PR #2754, opened 2025-08-22
Consolidate files into `types` directory

daquinteroflex:


daquinteroflex:
Making a separate PR because it's overwhelming to review otherwise

yaugenst-flex (changes requested):
For organization I think this is a great first step but to make the change less invasive I would expose the types in `__init__.py`, then none of the imports have to change (and the diff becomes a lot smaller).

daquinteroflex:
Yes good idea, working on this now

daquinteroflex:
Actually @yaugenst-flex turns out we can't actually do this because it creates circular imports everywhere because `base` and other types are part of different scopes and other directories need to use both. So going to revert the last commit

This is related to doing the file structures based on scope are we were discussing with @tylerflex

PR #2790, opened 2025-09-03
fix: Update rf smatrix logic flow

daquinteroflex:
## Greptile Summary

Updated On: 2025-09-03 13:38:53 UTC

This PR improves the API consistency of the S-matrix functionality in the `TerminalComponentModeler` by modifying how default parameters are handled in the `smatrix()` method. Previously, the method had hardcoded default values (`assume_ideal_excitation=False` and `s_param_def="pseudo"`) that would override the modeler's configuration settings. Now, these parameters default to `None` and fall back to the modeler's own configuration when not explicitly provided.

The change affects the `TerminalComponentModelerData.smatrix()` method in `tidy3d/plugins/smatrix/data/terminal.py`. When users call `smatrix()` without specifying these parameters, the method will now respect the `assume_ideal_excitation` and `s_param_def` values that were set during modeler initialization, creating a more intuitive configuration inheritance pattern.

This modification aligns with common software design principles where child operations should respect parent configuration by default while still allowing method-level overrides. The implementation uses conditional logic to check if parameters are `None` and falls back to `self.modeler.assume_ideal_excitation` and `self.modeler.s_param_def` accordingly.

## Important Files Changed

<details>

| tidy3d/plugins/smatrix/data/terminal.py | 3/5 | Modified parameter defaults in `smatrix()` method to use optional parameters with fallback logic to modeler configuration |

</details>

## Confidence score: 3/5

- This PR introduces a potentially breaking change that could affect existing user code that relied on the previous default behavior
- Score reflects concern about backward compatibility despite the logical improvement in API design
- Pay close attention to `tidy3d/plugins/smatrix/data/terminal.py` and ensure proper testing of the fallback logic

### DIFF
--- a/tidy3d/plugins/smatrix/data/terminal.py
+++ b/tidy3d/plugins/smatrix/data/terminal.py
@@ -101,8 +101,8 @@ class TerminalComponentModelerData(Tidy3dBaseModel):
 
     def smatrix(
         self,
-        assume_ideal_excitation: bool = False,
-        s_param_def: SParamDef = "pseudo",
+        assume_ideal_excitation: Optional[bool] = None,
+        s_param_def: Optional[SParamDef] = None,
     ) -> MicrowaveSMatrixData:
         """Computes and returns the S-matrix and port reference impedances.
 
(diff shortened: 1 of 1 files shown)


dmarek-flex (approved):

momchil-flex (approved):
Makes sense to me and also tested that we can remove the explicit setting in the tests now.

PR #2793, opened 2025-09-04
FXC-2087 Implement automatic structure extrusion for boundary waveports

George-Guryev-flxcmp:
This PR adds automatic extrusion of structures intersecting waveports that are placed on boundaries. 

Key changes:
- Introduced the  field in  to mark ports for which material should be extruded.
- Implemented the extrusion logic in , ensuring mode sources, internal absorbers, and PEC frames are fully contained within the extruded structures.
- Extrusion occurs only when  is set to .

This improves simulation accuracy when waveports are located on structure boundaries and simplifies user workflow.

## Greptile Summary

Updated On: 2025-09-04 15:39:10 UTC

This PR implements automatic structure extrusion for waveports positioned on simulation boundaries. The main functionality adds a new `extrude_structures` boolean field to the `WavePort` class that, when set to `True`, triggers automatic extension of intersecting structures to ensure mode sources, internal absorbers, and PEC frames are fully contained within the material geometry.

The core implementation resides in the `TerminalComponentModeler` class with the addition of a new `_extrude_port_structures` method that performs the geometric operations. This method identifies structures intersecting the waveport plane and creates extended `PolySlab` geometries that stretch from the port location to cover all necessary simulation components. The extrusion logic handles complex geometries with holes using `ClipOperation` and integrates seamlessly into the existing simulation pipeline.

This feature addresses a common simulation accuracy issue where waveports placed directly on structure boundaries might not provide adequate material coverage for electromagnetic field calculations. By automating the extrusion process, the PR simplifies user workflow while ensuring proper containment of all port-related simulation elements.

## Important Files Changed

| tidy3d/plugins/smatrix/component_modelers/terminal.py | 3/5 | Implements core extrusion logic with potential bounds checking issues |
| tidy3d/plugins/smatrix/ports/wave.py | 5/5 | Clean addition of extrude_structures boolean field with proper defaults |
| tests/test_plugins/smatrix/test_terminal_component_modeler.py | 4/5 | Comprehensive test validating extrusion functionality for coaxial geometry |
| CHANGELOG.md | 4/5 | Documents new feature with clear user-facing descriptions |

</details>

## Confidence score: 3/5

- This PR requires careful review due to potential runtime errors in the extrusion logic
- Score lowered due to unsafe array access without bounds checking and complex geometric operations that could fail
- Pay close attention to tidy3d/plugins/smatrix/component_modelers/terminal.py, specifically the `_extrude_port_structures` method

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
    participant TerminalComponentModeler as TCM
    participant WavePort
    participant Simulation as Sim
    participant Structure
    participant GeometryGroup as GeoGroup
    participant PolySlab
    participant ClipOperation as ClipOp

    User->>WavePort: "Create WavePort with extrude_structures=True"
    User->>TCM: "Create TerminalComponentModeler with WavePort"
    
    User->>TCM: "Access sim_dict property"
    TCM->>TCM: "_add_source_to_sim(network_index)"
    TCM->>WavePort: "to_source(source_time, snap_center)"
    WavePort-->>TCM: "ModeSource"
    TCM->>Sim: "updated_copy(sources=[port_source])"
    Sim-->>TCM: "simulation with source"
    
    TCM->>TCM: "_extrude_port_structures(sim_with_src)"
    
    loop "For each WavePort with extrude_structures=True"
        TCM->>WavePort: "Check extrude_structures flag"
        WavePort-->>TCM: "True"
        
        TCM->>TCM: "Calculate mode source position and offset"
        TCM->>TCM: "Define extrusion bounding box"
        TCM->>TCM: "Get slice planes (left and right)"
        
        loop "For each structure in simulation"
            TCM->>Structure: "Get geometry intersections with slice planes"
            Structure-->>TCM: "Shapely polygons"
            
            loop "For each intersecting polygon"
                TCM->>PolySlab: "Create outer shell from exterior vertices"
                PolySlab-->>TCM: "Outer PolySlab"
                
                alt "Polygon has holes"
                    loop "For each hole"
                        TCM->>PolySlab: "Create hole PolySlab from interior vertices"
                        PolySlab-->>TCM: "Hole PolySlab"
                    end
                    TCM->>GeoGroup: "Create GeometryGroup(hole_polyslabs)"
                    GeoGroup-->>TCM: "Holes geometry group"
                    TCM->>ClipOp: "ClipOperation(difference, outer_shell, holes)"
                    ClipOp-->>TCM: "Extruded geometry with holes"
                else "No holes"
                    TCM->>TCM: "Use outer shell as extruded geometry"
                end
                
                TCM->>TCM: "Add extruded geometry to new_geoms list"
            end
            
            TCM->>TCM: "Add original structure geometry to new_geoms"
            TCM->>GeoGroup: "Create GeometryGroup(new_geoms)"
            GeoGroup-->>TCM: "Combined geometry group"
            TCM->>Structure: "Create new Structure with combined geometry"
            Structure-->>TCM: "Extended structure"
            TCM->>TCM: "Add to new_structures list"
        end
    end
    
    TCM->>Sim: "updated_copy(structures=new_structures)"
    Sim-->>TCM: "Simulation with extruded structures"
    TCM-->>User: "SimulationMap with extruded structures"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added S-parameter de-embedding to `TerminalComponentModelerData`, enabling recalculation with shifted reference planes.
+- Added optional automatic extrusion of structures intersecting with a `WavePort` via the new `extrude_structures` field, ensuring mode sources, absorbers, and PEC frames are fully contained.
 - Added support for `tidy3d-extras`, an optional plugin that enables more accurate local mode solving via subpixel averaging.
 - Added support for `symlog` and `log` scale plotting in `Scene.plot_eps()` and `Scene.plot_structures_property()` methods. The `symlog` scale provides linear behavior near zero and logarithmic behavior elsewhere, while 'log' is a base 10 logarithmic scale.
 - Added `LowFrequencySmoothingSpec` and `ModelerLowFrequencySmoothingSpec` for automatic smoothing of mode monitor data at low frequencies where DFT sampling is insufficient.
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -2151,29 +2151,9 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
         the frame is added around the injection plane. For internal absorbers, a backing pec
         plate is also added on the non-absorbing side.
         """
-        span_inds = np.array(self.grid.discretize_inds(obj))
 
-        coords = self.grid.boundaries.to_list
-        direction = obj.direction
-        if isinstance(obj, ModeSource):
-            axis = obj.injection_axis
-            length = obj.frame.length
-            if direction == "+":
-                span_inds[axis][1] += length - 1
-            else:
-                span_inds[axis][0] -= length - 1
-        else:
-            axis = obj.size.index(0.0)
-
-        box_bounds = [
-            [
-                c[beg],
-                c[end],
-            ]
-            for c, (beg, end) in zip(coords, span_inds)
-        ]
-
-        box = Box.from_bounds(*np.transpose(box_bounds))
+        # get pec frame bounding box, object's axis and direction
+        (box, axis, direction) = self._pec_frame_box(obj)
 
         surfaces = Box.surfaces(box.size, box.center)
         if isinstance(obj, ModeSource):
@@ -2192,6 +2172,42 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
         )
 
         return structure
+
+    def _pec_frame_box(
+        self, obj: Union[ModeSource, InternalAbsorber], expand: bool = False
+    ) -> tuple[Box, int, str]:
+        """Return pec bounding box, frame axis and object's direction"""
+
+        span_inds = np.array(self.grid.discretize_inds(obj))
+        coords = self.grid.boundaries.to_list
+        direction = obj.direction
+        if isinstance(obj, ModeSource):
+            axis = obj.injection_axis
+            length = obj.frame.length
+            if direction == "+":
+                span_inds[axis][1] += length - 1
+            else:
+                span_inds[axis][0] -= length - 1
+        else:
+            axis = obj.size.index(0.0)
+
+        # ensure that the pec frame is at least one cell larger than wave port plane
+        if expand:
+            for dim in range(3):
+                if dim != axis:
+                    grid_size = len(coords[dim])
+                    (beg, end) = span_inds[dim]
+                    span_inds[dim] = [np.maximum(0, beg - 1), np.minimum(grid_size - 1, end + 1)]
+
+        box_bounds = [
+            [
+                c[beg],
+                c[end],
+            ]
+            for c, (beg, end) in zip(coords, span_inds)
+        ]
+
+        return (Box.from_bounds(*np.transpose(box_bounds)), axis, direction)
 
     @cached_property
     def _modal_plane_frames(self) -> list[Structure]:
--- a/tidy3d/plugins/smatrix/component_modelers/terminal.py
+++ b/tidy3d/plugins/smatrix/component_modelers/terminal.py
@@ -7,12 +7,14 @@ from typing import Optional, Union
 import numpy as np
 import pydantic.v1 as pd
 
+from tidy3d import ClipOperation, GeometryGroup, GridSpec, PolySlab
 from tidy3d.components.base import cached_property
 from tidy3d.components.boundary import BroadbandModeABCSpec
 from tidy3d.components.frequency_extrapolation import (
     AbstractLowFrequencySmoothingSpec,
     LowFrequencySmoothingSpec,
 )
+from tidy3d.components.geometry.utils import _shift_object
 from tidy3d.components.geometry.utils_2d import snap_coordinate_to_grid
 from tidy3d.components.index import SimulationMap
 from tidy3d.components.microwave.base import MicrowaveBaseModel
@@ -21,7 +23,7 @@ from tidy3d.components.simulation import Simulation
 from tidy3d.components.source.time import GaussianPulse
 from tidy3d.components.types import Ax, Complex
 from tidy3d.components.viz import add_ax_if_none, equal_aspect
-from tidy3d.constants import C_0, OHM
+from tidy3d.constants import C_0, OHM, fp_eps
 from tidy3d.exceptions import SetupError, Tidy3dKeyError, ValidationError
 from tidy3d.log import log
 from tidy3d.plugins.smatrix.component_modelers.base import (
@@ -284,6 +286,7 @@ class TerminalComponentModeler(AbstractComponentModeler, MicrowaveBaseModel):
         # Now, create simulations with wave port sources and mode solver monitors for computing port modes
         for network_index in self.matrix_indices_run_sim:
             task_name, sim_with_src = self._add_source_to_sim(network_index)
+            # update simulation
             sim_dict[task_name] = sim_with_src
 
         # Check final simulations for grid size at ports
@@ -379,8 +382,14 @@ class TerminalComponentModeler(AbstractComponentModeler, MicrowaveBaseModel):
                 max_deviation=self.low_freq_smoothing.max_deviation,
             )
 
-        # This is the new default simulation will all shared components added
-        return sim_wo_source.copy(update=update_dict)
+        # update base simulation with updated set of shared components
+        sim_wo_source = sim_wo_source.copy(update=update_dict)
+
+        # extrude port structures
+        sim_wo_source = self._extrude_port_structures(sim=sim_wo_source)
+
+        # This is the new default simulation with all shared components added
+        return sim_wo_source
 
     def _add_source_to_sim(self, source_index: NetworkIndex) -> tuple[str, Simulation]:
         """Adds the source corresponding to the ``source_index`` to the base simulation."""
@@ -398,6 +407,7 @@ class TerminalComponentModeler(AbstractComponentModeler, MicrowaveBaseModel):
                 self._source_time, snap_center=new_port_center, grid=self.base_sim.grid
             )
         task_name = self.get_task_name(port=port, mode_index=mode_index)
+
         return (task_name, self.base_sim.updated_copy(sources=[port_source]))
 
     @cached_property
@@ -506,6 +516,130 @@ class TerminalComponentModeler(AbstractComponentModeler, MicrowaveBaseModel):
             if monitor.name == monitor_name:
                 return monitor
         raise Tidy3dKeyError(f"No radiation monitor named '{monitor_name}'.")
+
+    def _extrude_port_structures(self, sim: Simulation) -> Simulation:
+        """
+        Extrude structures intersecting a port plane when a wave port lies on a structure boundary.
+
+        This method checks wave ports with ``extrude_structures==True`` and automatically extends the boundary structures
+        to PEC plates associated with internal absorbers in the direction opposite to the mode source.
+        This ensures that mode sources and internal absorbers are fully contained within the extrusion.
+
+        Parameters
+        ----------
+        sim : Simulation
+            Simulation object containing mode sources, internal absorbers, and monitors,
+            after mesh overrides and snapping points are applied.
+
+        Returns
+        -------
+        Simulation
+            Updated simulation with extruded structures added to ``simulation.structures``.
+        """
+
+        # create list with extruded structures
+        new_structures = []
+        all_new_structures = []
+
+        # get all mode sources from TerminalComponentModeler that correspond to ports with ``extrude_structures`` flag set to ``True``.
+        for port in self.ports:
+            if isinstance(port, WavePort) and port.extrude_structures:
+                # compute snap_center and shift the internal absorber associated with the current port
+                snap_center = port.center[port.injection_axis] + self._shift_value_signed(port)
+                absorber = port.to_absorber(snap_center=snap_center)
+                shifted_absorber = _shift_object(
+                    obj=absorber,
+                    grid=sim.grid,
+                    bounds=sim.bounds,
+                    direction=absorber.direction,
+                    shift=absorber.grid_shift,
+                )
+
+                # get the PEC box with its face surfaces
+                (box, inj_axis, direction) = sim._pec_frame_box(shifted_absorber, expand=True)
+                surfaces = box.surfaces(box.size, box.center)
+
+                # get extrusion coordinates and a cutting plane for inference of intersecting structures.
+                sign = 1 if direction == "+" else -1
+                back_pec_plane = surfaces[2 * inj_axis + (1 if direction == "+" else 0)]
+
+                # get extrusion extent along injection axis
+                extrude_to = back_pec_plane.center[inj_axis]
+
+                # move cutting plane beyond the waveport plane along the `ModeSource` injection direction.
+                center = list(back_pec_plane.center)
+                center[inj_axis] = port.center[inj_axis] - sign * fp_eps * box.size[inj_axis]
+                cutting_plane = back_pec_plane.updated_copy(center=center)
+
+                # define extrusion bounds
+                extrusion_bounds = [cutting_plane.center[inj_axis], extrude_to][::sign]
+
+                # loop over structures and extrude those that intersect a waveport plane
+                for structure in sim.structures:
+                    # get geometries that intersect the plane on which the waveport is defined
+
+                    shapely_geom = cutting_plane.intersections_with(structure.geometry)
+
+                    polygon_list = []
+                    for geom in shapely_geom:
+                        polygon_list = polygon_list + ClipOperation.to_polygon_list(geom)
+
+                    new_geoms = []
+                    # loop over identified geometries and extrude them
+                    for polygon in polygon_list:
+                        # construct outer shell of an extruded geometry first
+                        exterior_vertices = np.array(polygon.exterior.coords)
+                        outer_shell = PolySlab(
+                            axis=inj_axis, slab_bounds=extrusion_bounds, vertices=exterior_vertices
+                        )
+
+                        # construct innner shells that represent holes
+                        hole_polyslabs = [
+                            PolySlab(
+                                axis=inj_axis,
+                                slab_bounds=extrusion_bounds,
+                                vertices=np.array(hole.coords),
+                            )
+                            for hole in polygon.interiors
+                        ]
+
+                        # construct final geometry by removing inner holes from outer shell
+                        if hole_polyslabs:
+                            holes = GeometryGroup(geometries=hole_polyslabs)
+                            extruded_slab_new = ClipOperation(
+                                operation="difference", geometry_a=outer_shell, geometry_b=holes
+                            )
+                        else:
+                            extruded_slab_new = outer_shell
+
+                        # append extruded geometry
+                        new_geoms.append(extruded_slab_new)
+                    if len(polygon_list) != 0:
+                        # update structure and add it to the list
+                        new_struct = structure.updated_copy(
+                            geometry=GeometryGroup(geometries=new_geoms)
+                        )
+                        new_structures.append(new_struct)
+
+                        # if current port does not intersect any structures raise error
+                if not new_structures:
+                    raise SetupError(
+                        f"The 'WavePort' '{port.name}' does not intersect any structures."
+                        f"Please ensure that it is located within or at the boundary of a structure."
+                    )
+
+                all_new_structures = all_new_structures + new_structures
+                new_structures = []
+
+        # if new structures are extruded (Lumped Port extrusion is ignored)
+        if all_new_structures:
+            # update structures in simulation while keeping the same grid
+            sim = sim.updated_copy(
+                grid_spec=GridSpec.from_grid(sim.grid),
+                structures=[*sim.structures, *all_new_structures],
+            )
+
+        return sim
 
 
 TerminalComponentModeler.update_forward_refs()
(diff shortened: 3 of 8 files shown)


tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -443,5 +446,239 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
                 return monitor
         raise Tidy3dKeyError(f"No radiation monitor named '{monitor_name}'.")
 
+    def get_antenna_metrics_data(
```
daquinteroflex: This needs to go into smatrix/data/terminal, will not work here anymore

George-Guryev-flxcmp: Thank you, it is helpful! It seems there's been a few changes in `terminal.py` as rebase to develop broke my tests and notebooks. Will backtrack changes now.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Add automatic structure extrusion for waveports defined on boundaries, controlled by the `extrude_structures` field in `WavePort`.
+- The extrusion method, implemented in `TerminalComponentModeler`, ensures that mode sources, absorbers, and PEC frames are fully contained within the extruded structures; extrusion occurs only when `extrude_structures` is set to `True`.
```
dbochkov-flexcompute: just one, combined in some way, point is sufficient

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -443,5 +446,142 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
...
+            slices = box.surfaces(box.size, box.center)
+            slice_plane_left = slices[2 * inj_axis]
+            slice_plane_right = slices[2 * inj_axis + 1]
+
+            # loop over structures and extrude those that intersect a waveport plane
+            for structure in sim.structures:
+                # get geometries that intersect the plane on which the waveport is defined
+                left_geom = slice_plane_left.intersections_with(structure.geometry)
+                right_geom = slice_plane_right.intersections_with(structure.geometry)
+                shapely_geom = left_geom or right_geom or []
```
dbochkov-flexcompute: is this done in this way so that it automatically determines from which side to extrude? I think it's better to rely on `WavePort.direction` as in general you can have structures on both sides of a port

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -231,7 +232,9 @@ def sim_dict(self) -> SimulationMap:
         # Now, create simulations with wave port sources and mode solver monitors for computing port modes
         for network_index in self.matrix_indices_run_sim:
             task_name, sim_with_src = self._add_source_to_sim(network_index)
-            sim_dict[task_name] = sim_with_src
+
+            # extrude structures if necessary and update simulation
+            sim_dict[task_name] = self._extrude_port_structures(sim_with_src)
```
dbochkov-flexcompute: is it possible to move this to [base_sim](https://github.com/flexcompute/tidy3d/blob/george/waveport/tidy3d/plugins/smatrix/component_modelers/terminal.py#L247)? so that we extrude structures only once and then re-use it for all simulation in the sim_dict? I guess it would be defined by whether the grid can change when [a source is added to base_sim](https://github.com/flexcompute/tidy3d/blob/george/waveport/tidy3d/plugins/smatrix/component_modelers/terminal.py#L324). It seems like it doesn't change, @dmarek-flex?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -443,5 +446,147 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
...
+
+                idx = np.abs(port_center[inj_axis] - coords[inj_axis]).argmin()
+                port_center[inj_axis] = coords[inj_axis][idx]
+
+                port = port.updated_copy(center=tuple(port_center))
+
+                mode_src_pos = port.center[port.injection_axis] + self._shift_value_signed(port)
+
+                # then convert to source
+                mode_sources.append(port.to_source(self._source_time, snap_center=mode_src_pos))
```
dbochkov-flexcompute: I am a bit unsure about this change of coordinates and determining bounds of extrusion overall. Not that it is incorrect, but more from the point of view that it feels somewhat disconnected from how the placement of backing PEC plates for absorbers is calculated and that could make it brittle and difficult to maintain in future. Given that bounds for extrusion should be given by either `[pec_plate_location, port.center[inj_axis]]` or `[port.center[inj_axis], pec_plate_location]`, depending on injection direction, what we only need is to calculate is `pec_plate_location`. Maybe what we could do is this:
- [when creating absorbers](https://github.com/flexcompute/tidy3d/blob/george/waveport/tidy3d/plugins/smatrix/component_modelers/terminal.py#L300) in `base_sim`, keep track of ports that request extrusions by recording created absorber and the port original location 
```
port_location = wave_port.center[wave_port.injection_axis]
extrude_to_absorber.append((port_absorber, port_location))
```
- pass `extrude_to_absorber` list to `_extrude_port_structures` with[ the current base sim](https://github.com/flexcompute/tidy3d/blob/george/waveport/tidy3d/plugins/smatrix/component_modelers/terminal.py#L322)
- in `Simulation` factor out the code that [determines placement of pec frames](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/simulation.py#L5643-L5667) into a separate function, e.g. `_pec_frame_box(self, obj: Union[ModeSource, InternalAbsorber])`
- then in `_extrude_port_structures` for each entry in `extrude_to_absorber` we can call `sim._pec_frame_box(absorber)` and infer where the backing PEC frame is expected. This would also give you the lateral dimensions of the extrusion.
- to get a plane for determining polygons for extrusion, you can take this pec_frame_box, shrink it to zero size in the injection direction (`absorber._normal_axis`), and place it at the `port_location` + small perturbation in the injection direction (opposite of `absorber.direction`)

Do you think that would work?

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -443,5 +446,147 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
...
+            for structure in sim.structures:
+                # get geometries that intersect the plane on which the waveport is defined
+                left_geom = slice_plane_left.intersections_with(structure.geometry)
+                right_geom = slice_plane_right.intersections_with(structure.geometry)
+                shapely_geom = left_geom or right_geom or []
+
+                new_geoms = []
+
+                # loop over identified geometries and extrude them
+                for polygon in shapely_geom:
```
dbochkov-flexcompute: does this work if shapely geom of multiple polygons?

tidy3d/plugins/smatrix/component_modelers/terminal.py
George-Guryev-flxcmp: I agree, this would be an elegant solution. The issue, however, is that  mode source and final absorber location are snapped to the updated grid:   `self.sim_dict` -> `self._add_source_to_sim()` -> `self.base_sim.updated_copy(sources=[port_source])`. The `updated_copy` in this case changes the grid.

tidy3d/plugins/smatrix/component_modelers/terminal.py
George-Guryev-flxcmp: Yes, that's a great point!

tidy3d/plugins/smatrix/component_modelers/terminal.py
George-Guryev-flxcmp: No, unfortunately, it does not work for shapely < 2.0 (that is a default version for Python 3.9).

tidy3d/plugins/smatrix/component_modelers/terminal.py
dbochkov-flexcompute: ah, I see

dmarek-flex: ok you are both right, but are you sure the grid changed after adding the source @George-Guryev-flxcmp? The goal is to not change the grid after adding the source, so that all simulation grids in `sim_dict` are identical up to the source.

I think the only way for the source to modify the grid is if a wavelength is not supplied to the grid spec, in which case the central wavelength of the source is used. Maybe we should add a validator or test somewhere that confirms this

dmarek-flex: Ah I think I missed the more important issue, extruding structures obviously changes the grid. But I think this is an important thing to consider. Should all simulations contain all of the extrusions? If not each simulation could have slightly different grids.

George-Guryev-flxcmp: yes, `base_sim` has a grid update based on the highest frequency. This update does occur if the initial grid was defined using (probably) different wavelength; this does happen in one of the test examples I worked on. Thus, the initial grid was changed  in `self._add_source_to_sim()` call inside the for loop in `sim_dict()`. 

Your suggestion of adding a validator is a good resolution to the problem. This way grid would be the same before and after the `tcm.sim_dict()` call and allow to extrude structures once for all simulations. 

Currently structures are extruded based on the updated grid (without changing it) and all simulations utilize the same updated grid (unless grid updates in `base_sim` are not numerically identical).

dmarek-flex:
Looking good, just testing it on a simple example though and I think I found a bug. In this 2 waveport example, the extrusion works properly, but the final list of structures is out of order. In this case, the extended substrate is added later in the list which overlaps previously extended strips.

image

[extrude_stripline.ipynb](https://github.com/user-attachments/files/22174760/extrude_stripline.ipynb)

tidy3d/plugins/smatrix/component_modelers/terminal.py
dmarek-flex: If possible, would you mind sharing the test you had where adding the source changed the grid? Recently, it has become more important that `base_sim` has the final grid, since that is how the web estimates the cost of the batch.

Not sure your final point is 100% accurate. Adding new structures might interact with the automatic layer refinement, to produce a new grid.

Now I am considering to make a PR that will enforce this, by setting the grid in `base_sim` using `CustomGrid`. Might be a performance boost too, since we will only need to perform automated meshing steps once, then the result will be stored in the `CustomGrid`.

dbochkov-flexcompute: yeah, I would expect grid not change from base_sim -> sim_dict given that we already provide the frequency to mesh at https://github.com/flexcompute/tidy3d/blob/george/waveport/tidy3d/plugins/smatrix/component_modelers/terminal.py#L254-L260. As for grid after adding extruded structures, George grabs the input mesh and converts it into custom mesh https://github.com/flexcompute/tidy3d/pull/2793/files#diff-141e8b2abd96198d1d7f10f1a6518111bea725814b7634466e00da1e11adeaaaR593 to prevent any changes

dmarek-flex: oh beauty!

dmarek-flex: @George-Guryev-flxcmp could you move that grid copying idea directly at the end of `base_sim`? Goal is to only generate the grid once and have it be identical for all simulations in `sim_dict`.

George-Guryev-flxcmp: Here is an example where grid is updated: 
[quick_test_3.ipynb](https://github.com/user-attachments/files/22175751/quick_test_3.ipynb)


here is a .json:
[Microstrip_modeler_draft_tidy3d-RF.json](https://github.com/user-attachments/files/22175693/Microstrip_modeler_draft_tidy3d-RF.json)


PS. I don't think .json is being parsed correctly after RF refactor. However, in this example `tcm.simulation.grid` and `sim.grid for sim in tcm.sim_dict` return different grids

dmarek-flex: I was able to get it to work using a slightly older commit. I think I see the confusion though. You are correct that 
`tcm.simulation` will have a different grid than `sim in tcm.sim_dict`. But recently we added the `base_sim` field, which modifies the user-supplied `tcm.simulation`. So in your example I get:

> print(tcm.simulation.grid == sims["smatrix_port_1_0"].grid)
> print(tcm.base_sim.grid == sims["smatrix_port_1_0"].grid)
> False
> True

George-Guryev-flxcmp: Oh, I see. That simplifies things a lot. Thank you both for the clarifications

tidy3d/plugins/smatrix/component_modelers/terminal.py
George-Guryev-flxcmp: Yes, that would be the cleanest way of doing that. 

The highlighted code was useful for symmetric extrusion of structures where ports' centers were forced to be snapped to a grid (in reality they are not perfectly aligned along extrusion direction). However, since we decided on a relative extrusion (i.e. extrude to cover mode source, absorber and PEC plate) this becomes messy and unnecessary.

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -443,5 +452,119 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
...
+                    extrude_to = cutting_plane.center[inj_axis]
+                    center = list(cutting_plane.center)
+
+                    # move cutting plane to waveport location and apply small perturbation to ensure cutting plane intersects structures
+                    center[inj_axis] = port.center[inj_axis] * (
+                        1 + 100 * np.sign(port.center[inj_axis]) * np.finfo(float).eps
+                    )
+                    # update cutting plane center and extrusion bounds
+                    cutting_plane = cutting_plane.updated_copy(center=center)
+                    extrusion_bounds = [extrude_to, center[inj_axis]]
```
dbochkov-flexcompute: these two large blocks are mostly repeating each other, let's consolidate along the lines of:
```
sign = 1 if direction == "+" else -1
cutting_plane = surfaces[2 * inj_axis + direction == "+"]
center[inj_axis] = port.center[inj_axis] * (1 - sign * ...)
extrusion_bounds = [center[inj_axis], extrude_to][::sign]
```

also, for perturbation let's use the size of the enclosing box in the injection direction (== grid size there) `box[2 * inj_axis + 1] - box[2 * inj_axis]` multiplied by `https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/constants.py#L67`

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -443,5 +452,119 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
...
+            after mesh overrides and snapping points are applied.
+
+        Returns
+        -------
+        Simulation
+            Updated simulation with extruded structures added to ``simulation.structures``.
+        """
+
+        # get all mode sources from TerminalComponentModeler that correspond to ports with ``extrude_structures`` flag set to ``True``.
+        for port, absorber in zip(self.ports, sim._shifted_internal_absorbers):
```
dbochkov-flexcompute: While I expect this to be uncommon, it is still possible to add internal absorbers into a simulation manually, not related to any port. In that case, self.ports and sim._shifted_internal_absorbers will have different lengths and not correspond to each other element by element. This needs to be taken into account

dbochkov-flexcompute (commented):
Almost there, a couple of comments + let's add an example where a waveguide assembly contains a geometry which is a union of two disjoint objects (something like `GeometryGroup(geometries=[box_left, box_right])`, like a differential strip

tidy3d/plugins/smatrix/component_modelers/terminal.py
George-Guryev-flxcmp: Good point. Fixed this one.

tidy3d/plugins/smatrix/component_modelers/terminal.py
George-Guryev-flxcmp: I agree. Modified code accordingly

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -315,8 +318,14 @@ def base_sim(self) -> Simulation:
...
-        # This is the new default simulation will all shared components added
-        return sim_wo_source.copy(update=update_dict)
+        # update base simulation with updated set of shared components
+        sim_wo_source = sim_wo_source.copy(update=update_dict)
+
+        # extrude port structures
+        sim_wo_source = self._extrude_port_structures(sim=sim_wo_source)
+
+        # This is the new default simulation with all shared components added
+        return sim_wo_source
```
dmarek-flex, line 392: thank you!

tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -98,6 +98,10 @@ class WavePort(AbstractTerminalPort, Box):
         "If ``ABCBoundary`` or ``ModeABCBoundary``, a mode absorber is placed in the port with the specified boundary conditions.",
     )
 
+    extrude_structures: bool = pd.Field(
+        False, title="Extrusion flag", description="Extrude structures attached to wave port."
```
dmarek-flex: I think the description can be improved a bit. Like in the absorber example above, explain briefly what happens when it is true (structures intersecting the mode plane are extruded by a few grid cells to improve mode injection). Also maybe just make the title=`"Extrude Structures"`, which follows our convention more.

Also I noticed the `frame` and `absorber` fields have periods in their title, can you fix those typos as well?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added optional automatic extrusion of boundary waveport structures via `WavePort.extrude_structures`, ensuring mode sources, absorbers, and PEC frames are fully contained.
```
dmarek-flex: " of structures intersecting with a `WavePort`  via the new `extrude_structures` field ,"

dmarek-flex (approved):
Thanks! Looking like a good feature, just a couple of minor style comments.

daquinteroflex (approved):
Looks good!

.pre-commit-config.yaml
```diff
@@ -4,15 +4,15 @@ default_install_hook_types:
...
+    rev: "v0.14.1"
     hooks:
      - id: ruff-check
        args: [ --fix ]
        stages: [pre-commit]
      - id: ruff-format
        stages: [pre-commit]
   - repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
-    rev: v9.22.0
+    rev: v9.23.0
```
dbochkov-flexcompute: should these changes be reverted?

tidy3d/components/simulation.py
```diff
@@ -5814,3 +5814,106 @@ def from_scene(cls, scene: Scene, **kwargs) -> Simulation:
         )
 
     _boundaries_for_zero_dims = validate_boundaries_for_zero_dims()
+
+    def _make_pec_frame(self, obj: Union[ModeSource, InternalAbsorber]) -> Structure:
```
dbochkov-flexcompute: ah, all these functions: `_make_pec_frame`, `_modal_plane_frames`, `_finalized`, and `_validate_finalized` were moved to `AbstractYeeGridSimulation`, you need to move any changes from here to there

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -507,5 +517,126 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
...
+                    # update structure and add it to the list
+                    new_struct = structure.updated_copy(
+                        geometry=GeometryGroup(geometries=new_geoms)
+                    )
+                    new_structures.append(new_struct)
+
+        # if new structures are extruded
+        if new_structures:
+            # update structures in simulation while keeping the same grid
+            print(type(sim.structures))
```
dbochkov-flexcompute: leftover print

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -507,5 +517,126 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
...
+
+                        # append extruded geometry
+                        new_geoms.append(extruded_slab_new)
+
+                    # if current port does not intersect any structures raise error
+                    if not new_geoms:
+                        raise SetupError(
+                            f"The 'WavePort' '{port.name}' does not intersect any structures."
+                            f"Please ensure that it is located within or at the boundary of a structure."
+                        )
```
dbochkov-flexcompute: this would error if the port intersects some structures, but not all structures, right? You should probably just skip non-intersecting structures (`len(shapely_geom) == 0`) and check `new_structures` in the end instead

dbochkov-flexcompute (changes requested):

tidy3d/plugins/smatrix/component_modelers/terminal.py
George-Guryev-flxcmp: That's a great catch!

tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -507,5 +517,125 @@ def get_radiation_monitor_by_name(self, monitor_name: str) -> DirectivityMonitor
...
+                        new_geoms.append(extruded_slab_new)
+                    if len(polygon_list) != 0:
+                        # update structure and add it to the list
+                        new_struct = structure.updated_copy(
+                            geometry=GeometryGroup(geometries=new_geoms)
+                        )
+                        new_structures.append(new_struct)
+
+                        # if current port does not intersect any structures raise error
+                if not new_structures:
```
dbochkov-flexcompute, line 625: almost there :) now I think if the first port intersects structures and the second port does not intersect anything, then it will not error because `new_structures` would be already non-empty from the first port

dbochkov-flexcompute (approved):

PR #2795, opened 2025-09-05
feat: initialize topology design regions from simulation geometry

yaugenst-flex:
## Greptile Summary

Updated On: 2025-09-05 10:24:39 UTC

This PR introduces a new topology optimization initialization feature to the Tidy3D inverse design module. The primary addition is the `initialize_params_from_simulation` function, which allows users to initialize design parameters for topology optimization by matching the permittivity distribution of an existing simulation within a specified design region.

The function addresses a common challenge in topology optimization where starting from random or uniform parameter values can lead to poor convergence or suboptimal results. Instead, this implementation uses L-BFGS-B optimization with automatic differentiation to find parameter values that best reproduce the permittivity distribution from a base simulation. The optimization minimizes the L2 error between the base simulation's permittivity and the parameterized structure's permittivity.

Key technical features include:
- Permittivity extraction and interpolation using scipy's zoom function to handle resolution differences
- Bounded optimization with L-BFGS-B that respects parameter constraints
- Early stopping mechanism using a sliding window to detect convergence based on relative improvement tolerance
- Comprehensive error handling that preserves the best solution even if optimization is interrupted
- Proper normalization to handle edge cases where the denominator might be zero

The function is integrated into the public API through the `__init__.py` file, making it easily discoverable for users. The implementation follows established patterns in the codebase and includes thorough documentation with examples. A comprehensive test suite validates the function's behavior, ensuring shape preservation, bounds compliance, and actual loss reduction during optimization.

This addition enhances the inverse design workflow by enabling users to start topology optimization from realistic initial conditions derived from existing simulations, potentially leading to faster convergence and better optimization outcomes.

## Important Files Changed

<details>

| `tidy3d/plugins/autograd/invdes/__init__.py` | 5/5 | Added import and export for the new `initialize_params_from_simulation` function to the public API |
| `CHANGELOG.md` | 4/5 | Added changelog entry documenting the new topology optimization initialization function |
| `tidy3d/plugins/autograd/invdes/parametrizations.py` | 4/5 | Implemented the core `initialize_params_from_simulation` function with L-BFGS-B optimization and early stopping |
| `tests/test_plugins/autograd/invdes/test_initialize_params_from_sim.py` | 4/5 | Added comprehensive test suite for the new initialization function with helper functions and validation checks |

</details>

## Confidence score: 4/5

- This PR is safe to merge with minimal risk as it adds a well-contained new feature without modifying existing functionality
- Score reflects solid implementation with proper optimization techniques, error handling, and comprehensive testing, but could benefit from additional edge case validation
- Pay close attention to the optimization logic in `parametrizations.py` to ensure the permittivity matching and early stopping mechanisms work correctly across different simulation types

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
```


tidy3d/plugins/autograd/invdes/parametrizations.py
```diff
@@ -96,3 +102,147 @@ def make_filter_and_project(
         filter_type=filter_type,
         padding=padding,
     )
+
+
+def initialize_params_from_simulation(
+    sim: td.Simulation,
+    param_to_structure: Callable[[np.ndarray], td.Structure],
+    params0: np.ndarray,
+    design_box: td.Box,
```
tylerflex: why do we need this extra required argument? There are two ways to approach this automatically I think:

1. simply make the design_box infinite size (potentially problematic but works)
2. use the `param_to_structure(params0).geometry` to get the design region box. Unfortunately when feeding this to `simulation.epsilon()` the data returned didnt span all of the data in the `CustomMedium` so that might be an edge case to think about (custom medium data can extend outside geometry, so maybe we need to also interpolate that back into the design box).

tidy3d/plugins/autograd/invdes/parametrizations.py
```diff
@@ -96,3 +102,147 @@ def make_filter_and_project(
...
+    ...     param_to_structure=param_to_structure,
+    ...     params0=params0,
+    ...     design_box=box,
+    ...     maxiter=3,
+    ... )
+    >>> params.shape
+    (3, 3)
+    """
+
+    eps_base = sim.epsilon(box=design_box).data
```
tylerflex: the `sim` grid with and without the design region Structure will have a different grid. So what I did in my branch was added the `param_to_structure(params0)` as a mesh override structure to `sim` so we had an equal grid. There are probably edge cases where this doesnt work, but do you think we need this kind of thing here?

tidy3d/plugins/autograd/invdes/parametrizations.py
```diff
@@ -96,3 +102,147 @@ def make_filter_and_project(
...
+    eps_param = param_to_structure(params0).medium.permittivity.data
+    factors = np.array(eps_param.shape) / np.array(eps_base.shape)
+    eps_base_interp = zoom(eps_base, factors, order=1)
```
tylerflex: working with the arrays directly (and not using the spatial coords) seems risky.. doesnt this assume that eps_param and eps_base have exactly the same spatial coords? which is not generally true if
* the design_box doesnt actually correspond to the `param_to_structure(params0).geometry`
* the `sim.grid` is different from the grid used in the design region custom medium
* the `eps_base` coordinates might extend outside of the design region (which is valid).

Maybe there's something about `zoom()` or the setup that handles this? but having bumped into a few of these issues in my notebook I'm a bit skeptical

tylerflex (commented):
Can test in my Smatrix notebook branch in a bit

tidy3d/plugins/autograd/invdes/parametrizations.py
yaugenst-flex: ooh yeah you're right, 2. does seem much nicer! i think we don't have to worry about the span too much though? i think as long as we make sure the array shapes are aligned, being off by a tiny bit shouldnt really matter in this context? the topopt initial design is never going to reconstruct the geometry completely faithfully anyways

tidy3d/plugins/autograd/invdes/parametrizations.py
yaugenst-flex: so yeah we can discuss. i excluded it because i think the mesh override should actually just be part of the base simulation. we can include the mesh override here too, but we can't (or shouldn't) modify the base simulation in this function so i'm not sure how we'd propagate that back to the user?

tidy3d/plugins/autograd/invdes/parametrizations.py
yaugenst-flex: when would the design box not correspond to (at least to good approximation) to the geometry that comes out of the parametrization?

basically i don't think the coordinates need to match exactly for this purpose here, or at least i can't really see a situation where this would cause significant problems? maybe you have an example?

basically here the coordinates are grabbed out of the simulation. the zoom is just in place so that the shape of the simulation epsilon matches that of the parametrized structure (since that will get resampeld by the simulation). the coordinates of the individual pixels won't be 100% accurate but basically i don't think it matters in this context..

tidy3d/plugins/autograd/invdes/parametrizations.py
yaugenst-flex: fixed!

tylerflex: so I was trying to interpolate the design region permittivity data into the `Simulation.epsilon(geometry)` coordinates to do 1-1 comparison. And because the `Simulation.epsilon()` didnt span the entire design region coordinates, I was getting Nans. Which I guess we could just ignore..

yaugenst-flex: i saw that too but i think the way it's implemented here that won't happen. have you tested it?

tidy3d/plugins/autograd/invdes/parametrizations.py
tylerflex: the mesh override structure was never intended to go back to the user. the point of it was just to make sure that the simulation mesh matched the mesh found in the design region when the Structure is added to the base simulation. But I think even this is brittle as it makes assumptions about the way the design region is constructed.

tidy3d/plugins/autograd/invdes/parametrizations.py
tylerflex: > when would the design box not correspond to (at least to good approximation) to the geometry that comes out of the parametrization?

my point is that what if the user just provides something random?

> basically i don't think the coordinates need to match exactly for this purpose here, or at least i can't really see a situation where this would cause significant problems? maybe you have an example?

yea I suppose it's fine as long as we do the interpolation properly. on 2nd thought.

>  basically here the coordinates are grabbed out of the simulation. the zoom is just in place so that the shape of the simulation epsilon matches that of the parametrized structure (since that will get resampeld by the simulation). the coordinates of the individual pixels won't be 100% accurate but basically i don't think it matters in this context..

ok so what you're saying is that you essentially 
1) take a `simulation.epsilon(design_box).data`
2)  take the permittivity of the `param_to_structure(params0).medium.permittivity.data`

and then resample 1) such that it has the same shape as 2?

There are some complications though. these data points done necessarily line up. It all depends on how Simulation.epsilon() determines where to evaluate permittivity from (the Simulation grid) and also how the user set up their design region permittivity DataArray.

I think the safest is to actually use coordinates and interpolate.

tylerflex (changes requested):
Worked great in my test notebook, and overall implementation is good. Thank @yaugenst-flex . A couple comments:

1. I see no reason to not import this also into the main `tidy3d.plugins.autograd` namespace. I forget why we decided to separate them into `.invdes`. do you remember?
2. In some cases the `param_to_structure` might take not just `params` but also `**kwargs`. In my notebook, it took `beta` so I had to passs `functools.partial(param_to_structure, beta=1)`, which is not a huge deal but I think it would be kind of nice to just allow passing kwargs into the `param_to_structure` what do you think?
3.

yaugenst-flex:
@tylerflex yeah both good points, addressed. the `invdes` separation is intended so that more "pure" autograd-related functionality (such as e.g. primitives) are there, while inverse-design specific functionality (such as autograd-diffable parametrizations and filters) are in `invdes`. but for simplicity we can expose most of those functions in the `plugins.autograd` namespace

tylerflex (approved):

PR #2798, opened 2025-09-05
fix pec frame lengths

dbochkov-flexcompute:
`span_inds = np.array(self.grid.discretize_inds(obj))` by itself gives frame of length 1 already

## Greptile Summary

Updated On: 2025-09-08 21:40:39 UTC

This PR fixes a bug in the PEC (Perfect Electric Conductor) frame length calculation for mode sources in the Tidy3D electromagnetic simulation framework. The issue was that the code was double-counting frame dimensions when discretizing objects on the computational grid.

The problem occurred in the `_get_pec_objects` method where `self.grid.discretize_inds(obj)` already returns grid indices that create a frame of length 1, but the original code was then adding/subtracting additional frame length on top of that base frame. This resulted in PEC frames being larger than intended.

The fix reorganizes the logic to:
1. Use the base frame from `discretize_inds()` for `InternalAbsorber` objects (which only need a single-cell frame)
2. Apply frame length adjustments only to `ModeSource` objects, but subtract 1 from the adjustment to account for the base frame already provided
3. Move the frame length adjustment logic inside the `ModeSource` conditional block to prevent it from affecting other object types

This change ensures that PEC frames have the correct dimensions as specified by the design, with `ModeSource` objects respecting their configurable `frame.length` parameter and `InternalAbsorber` objects using the appropriate single-cell frame.

## Important Files Changed

| tidy3d/components/simulation.py | 4/5 | Fixed double-counting bug in PEC frame length calculation by moving frame adjustment logic inside ModeSource conditional |

</details>

## Confidence score: 4/5

- This PR is safe to merge with low risk as it fixes a clear bug in frame dimension calculation
- Score reflects a straightforward bug fix that addresses double-counting in frame length calculations
- Pay close attention to tidy3d/components/simulation.py to ensure the frame length logic works correctly for all object types

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
    participant Simulation
    participant Grid
    participant PECFrame

    User->>Simulation: Create simulation with ModeSource + PECFrame
    Simulation->>Simulation: _make_pec_frame(mode_source)
    Simulation->>Grid: discretize_inds(mode_source)
    Grid-->>Simulation: span_inds (already length 1)
    
    Note over Simulation: Bug fix: span_inds already gives length 1
    Note over Simulation: Old code: span_inds[axis][1] += length
    Note over Simulation: New code: span_inds[axis][1] += length - 1
    
    Simulation->>Simulation: Calculate box_bounds from corrected span_inds
    Simulation->>PECFrame: Create Structure with PECMedium
    PECFrame-->>Simulation: PEC frame structure
    Simulation-->>User: Simulation with correctly sized PEC frame
```

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -5645,16 +5645,12 @@ class Simulation(AbstractYeeGridSimulation):
         if isinstance(obj, ModeSource):
             axis = obj.injection_axis
             length = obj.frame.length
+            if direction == "+":
+                span_inds[axis][1] += length - 1
+            else:
+                span_inds[axis][0] -= length - 1
         else:
             axis = obj.size.index(0.0)
-            length = 1
-
-        if direction == "+":
-            span_inds[axis][1] += length - 1
-            span_inds[axis][0] -= 1
-        else:
-            span_inds[axis][1] += 1
-            span_inds[axis][0] -= length - 1
 
         box_bounds = [
             [


George-Guryev-flxcmp:
Great catch!

In the previous implementation, the `ModeSource` PEC frame ended up one cell longer than frame.length, while the `InternalAbsorber` PEC frame was effectively 2 cells long. This caused a 2-cell overlap when a waveport was defined.

I do have a tangential question: when a `WavePort` is defined, the `ModeSource` is placed either 1.5 or 2.5 cells away from the monitor depending on rounding (`WavePort.center` doesn’t always align exactly with sim.grid.boundary, leading to near machine-precision differences that convert to 1 cell offset in `ModeSource` placement). This results in two cases:

- The PEC frame extends to the `ModeMonitor` (when the `ModeSource` is 1.5 cells away), or

- The PEC frame is one cell short of the `ModeMonitor`.

I assume the second case is unintended. I wounder what is a cleaner approach to fixing it: either ensuring that port center is snapped to sim.grid.boundaries exactly or extending 'ModeSource's PEC frame to a `ModeSource`?

Curious to hear your thoughts.

George-Guryev-flxcmp (commented):
Great catch!

In the previous implementation, the `ModeSource` PEC frame ended up one cell longer than `frame.length`, while the `InternalAbsorber` PEC frame was effectively 2 cells long. This caused a 2-cell overlap when a waveport was defined.

I do have a tangential question: when a `WavePort` is defined, the `ModeSource` is placed either 1.5 or 2.5 cells away from the monitor depending on rounding (`WavePort.center` doesn’t always align exactly with sim.grid.boundary, leading to near machine-precision differences that convert to 1 cell offset in `ModeSource` placement). This results in two cases:

The PEC frame extends to the `ModeMonitor` (when the `ModeSource` is 1.5 cells away), or

The PEC frame is one cell short of the `ModeMonitor`.

I assume the second case is unintended. I wounder what is a cleaner approach to fixing it: either ensuring that port center is snapped to `sim.grid.boundaries` exactly or extending 'ModeSource's PEC frame to a `ModeSource`?

Curious to hear your thoughts.

George-Guryev-flxcmp:
Just to follow up on this point: Damian and I believe the problem mentioned earlier stems from `_shifted_value_signed()` that is used in  TCM `base_sim()` when setting up `ModeSources` and absorbers.  In particular, line https://github.com/flexcompute/tidy3d/blob/da424f6bd76fd0eb7caef7693638e6dbbfb4aa6c/tidy3d/components/geometry/utils.py#L515  is susceptible to machine precision perturbations in port coordinates.

dbochkov-flexcompute:
> Great catch!
> 
> In the previous implementation, the `ModeSource` PEC frame ended up one cell longer than `frame.length`, while the `InternalAbsorber` PEC frame was effectively 2 cells long. This caused a 2-cell overlap when a waveport was defined.
> 
> I do have a tangential question: when a `WavePort` is defined, the `ModeSource` is placed either 1.5 or 2.5 cells away from the monitor depending on rounding (`WavePort.center` doesn’t always align exactly with sim.grid.boundary, leading to near machine-precision differences that convert to 1 cell offset in `ModeSource` placement). This results in two cases:
> 
> The PEC frame extends to the `ModeMonitor` (when the `ModeSource` is 1.5 cells away), or
> 
> The PEC frame is one cell short of the `ModeMonitor`.
> 
> I assume the second case is unintended. I wounder what is a cleaner approach to fixing it: either ensuring that port center is snapped to `sim.grid.boundaries` exactly or extending 'ModeSource's PEC frame to a `ModeSource`?
> 
> Curious to hear your thoughts.

Regarding first approach, while the tidy3d attempts to place a port exactly at a primal grid location, it is not guaranteed, so it might end up to be between primal grid locations, and snapping would not be appropriate in this case. As for extending all the way to the mode monitor location, from what I have seen it didn't seem to change the results either way. @yuanshen-flexcompute, in the examples that you ran so far, have you noticed any variations due to pec frame extending or not all the way to mode monitor locations?

yuanshen-flexcompute:
I have not tested it extensively, but the length of the frame does tend to affect results (since we're essentially inserting a long metal tube). I'd imagine it depends on the size of the wave port relative to the mode. For a "simple" RF case like a stripline I think it does not matter. But for micron-scale CPW it might make a difference. Could be worth testing

George-Guryev-flxcmp (approved):

PR #2800, opened 2025-09-08
MultiGaussian pulse and more accurate frequency range

weiliangjin2021:
This PR addresses a few existing issues:

- source time's frequency_range is accurate for pure Gaussian. For derivative Gaussian, it will deviates significantly with broader bandwidth.
- Similary, `offset` is only accurate for pure Gaussian, and derivative Gaussian with `freq0>fwidth`, otherwise there can be two peaks in time, as illustrated in the figure below. In this case, `offset` refers to the position of the 1st peak in this PR.
image

New features:

- MultiGaussian pulse is introduced, which is simply a tuple of Gaussian pulses. But it introduces lots of changes to the code (both frontend and solver), as it doesn't have `freq0` and `fwidth` fields.
- The important feature of MultiGaussian is its automatic generation from frequency range. The minimal amplitude of the pulse in frequency domain relative to the peak amplitude in the frequency range is roughly bounded by the user defined value. The number of Gaussians will increase with larger frequency range, and smaller lower bound.
- In the auto-generated MulitGaussian, we always remove DC component for each individual Gaussian pulse. This complicates the math, and `Brentq` solver is used in multiple places.

## Example
Let's define a source over a very broadband frequency range [0.1, 60] GHz, and source amplitude relative to the peak amplitude over this range should be above 30%. The code is as follows,
`source_time = td.MultiGaussianPulse.from_frequency_range(fmin=.1e9, fmax=60e9, min_rel_amp=0.3)`

The figure illustrates that the generated pulse contains 3 Gaussian pulses.
image

### DIFF
--- a/tidy3d/components/simulation.py
+++ b/tidy3d/components/simulation.py
@@ -814,7 +814,7 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
             bounds=self.simulation_bounds, x=x, y=y, z=z, hlim=hlim, vlim=vlim
         )
         if freq is None:
-            freq0s = [source.source_time.freq0 for source in self.sources]
+            freq0s = [source.source_time._freq0 for source in self.sources]
             if freq0s and all(math.isclose(freq0, freq0s[0]) for freq0 in freq0s):
                 freq = freq0s[0]
             else:
@@ -1982,7 +1982,7 @@ class AbstractYeeGridSimulation(AbstractSimulation, ABC):
 
         # some nonlinear materials depend on the central frequency
         # we update them with hardcoded freq0
-        freqs = np.array([source.source_time.freq0 for source in self.sources])
+        freqs = np.array([source.source_time._freq0 for source in self.sources])
         for i, structure in enumerate(new_structures):
             medium = structure.medium
             nonlinear_spec = medium.nonlinear_spec
@@ -3131,7 +3131,7 @@ class Simulation(AbstractYeeGridSimulation):
                     "Add at least one source or use parameter 'frequency' for 'ModeABCBoundary'.",
                 )
 
-            freq0s = [source.source_time.freq0 for source in sources]
+            freq0s = [source.source_time._freq0 for source in sources]
             if not all(math.isclose(freq0, freq0s[0]) for freq0 in freq0s):
                 log.warning(
                     "At least one 'ModeABCBoundary' does not specify frequency at which the absorbed mode must be evaluated. "
@@ -3261,7 +3261,7 @@ class Simulation(AbstractYeeGridSimulation):
                 struct_bound_min, struct_bound_max = structure.geometry.bounds
 
                 for source in sources:
-                    lambda0 = C_0 / source.source_time.freq0
+                    lambda0 = C_0 / source.source_time._freq0
 
                     zipped = zip(["x", "y", "z"], sim_bound_min, struct_bound_min, boundaries)
                     for axis, sim_val, struct_val, boundary in zipped:
@@ -3352,7 +3352,7 @@ class Simulation(AbstractYeeGridSimulation):
         if val is None:
             return val
 
-        source_ranges = [source.source_time.frequency_range() for source in values["sources"]]
+        source_ranges = [source.source_time.frequency_range_loose() for source in values["sources"]]
         if not source_ranges:
             # Commented out to eliminate this message from Mode real time log in GUI
             # TODO: Bring it back when it doesn't interfere with mode solver
@@ -3823,7 +3823,7 @@ class Simulation(AbstractYeeGridSimulation):
 
         with log as consolidated_logger:
             for source_index, source in enumerate(values.get("sources")):
-                freq0 = source.source_time.freq0
+                freq0 = source.source_time._freq0_for_mesher
 
                 for medium_index, medium in enumerate(mediums):
                     # min wavelength in PEC/PMC is meaningless and we'll get divide by inf errors
@@ -4260,7 +4260,7 @@ class Simulation(AbstractYeeGridSimulation):
     def _validate_nonlinear_specs(self) -> None:
         """Run :class:`.NonlinearSpec` validators that depend on knowing the central
         frequencies of the sources. Also print some warnings only once per unique medium."""
-        freqs = np.array([source.source_time.freq0 for source in self.sources])
+        freqs = np.array([source.source_time._freq0 for source in self.sources])
         for medium in self.scene.mediums:
             if medium.nonlinear_spec is not None:
                 for model in medium._nonlinear_models:
@@ -4423,7 +4423,7 @@ class Simulation(AbstractYeeGridSimulation):
                         center=source.center,
                         size=source.size,
                         name="tmp",
-                        freqs=[source.source_time.freq0],
+                        freqs=[source.source_time._freq0],
                         colocate=False,
                     )
                     msg_header = f"Mode source at sources[{src_ind}] "
@@ -4487,7 +4487,7 @@ class Simulation(AbstractYeeGridSimulation):
 
     def _validate_freq_monitors_freq_range(self) -> None:
         """Rise the error if any DFT monitors have frequencies outside of the simulation frequency range."""
-        source_ranges = [source.source_time.frequency_range() for source in self.sources]
+        source_ranges = [source.source_time.frequency_range_loose() for source in self.sources]
         if not source_ranges:
             return
 
@@ -4597,7 +4597,7 @@ class Simulation(AbstractYeeGridSimulation):
                 # side wall - the profiles must be the same along the injection axis, so we take
                 # a single "stripe" of epsilon as the reference and subtract it from all other
                 # stripes, which should result in zero if all the epsilon profiles are the same
-                freq0 = source.source_time.freq0
+                freq0 = source.source_time._freq0
                 _, plane_axs = source.pop_axis("xyz", axis=source.injection_axis)
                 ref_eps = self.epsilon(box=sidewall_surfaces[0], coord_key="centers", freq=freq0)
                 kwargs = {plane_axs[0]: 0, plane_axs[1]: 0}
@@ -4703,7 +4703,7 @@ class Simulation(AbstractYeeGridSimulation):
 
             # get the maximum refractive index evaluated over each of all the source central frequencies
             all_ref_inds = [
-                self.get_refractive_indices(src.source_time.freq0) for src in self.sources
+                self.get_refractive_indices(src.source_time._freq0) for src in self.sources
             ]
             avg_ref_inds = [np.mean(np.array(n)) for n in all_ref_inds]
             max_ref_ind = np.max(avg_ref_inds, initial=1)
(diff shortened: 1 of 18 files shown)


weiliangjin2021:
@yaugenst-flex a failing autograd test that I don't get the details (oh, it seems to pass on github action):

> _____________________________________ TestTidyArrayBox.test_is_tidy_box ______________________________________
> self = <tests.test_components.autograd.test_autograd.TestTidyArrayBox object at 0x74692a3d3cd0>
> 
>     def test_is_tidy_box(self):
> >       da = DataArray(tracer_arr, dims=map(str, range(tracer_arr.ndim)))
> 
> tests/test_components/autograd/test_autograd.py:2502: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> tidy3d/components/data/data_array.py:89: in __init__
>     super().__init__(data, *args, **kwargs)
> python3.10/site-packages/xarray/core/dataarray.py:450: in __init__
>     coords, dims = _infer_coords_and_dims(data.shape, coords, dims)
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> shape = (1, 1, 1), coords = None, dims = <map object at 0x746929d79390>
> 
>     def _infer_coords_and_dims(
>         shape, coords, dims
>     ) -> tuple[Mapping[Hashable, Any], tuple[Hashable, ...]]:
>         """All the logic for creating a new DataArray"""
>     
>         if (
>             coords is not None
>             and not utils.is_dict_like(coords)
>             and len(coords) != len(shape)
>         ):
>             raise ValueError(
>                 f"coords is not dict-like, but it has {len(coords)} items, "
>                 f"which does not match the {len(shape)} dimensions of the "
>                 "data"
>             )
>     
>         if isinstance(dims, str):
>             dims = (dims,)
>     
>         if dims is None:
>             dims = [f"dim_{n}" for n in range(len(shape))]
>             if coords is not None and len(coords) == len(shape):
>                 # try to infer dimensions from coords
>                 if utils.is_dict_like(coords):
>                     dims = list(coords.keys())
>                 else:
>                     for n, (dim, coord) in enumerate(zip(dims, coords)):
>                         coord = as_variable(coord, name=dims[n]).to_index_variable()
>                         dims[n] = coord.name
>             dims = tuple(dims)
> >       elif len(dims) != len(shape):
> E       TypeError: object of type 'map' has no len()
> 
> /python3.10/site-packages/xarray/core/dataarray.py:172: TypeError

weiliangjin2021:
- This failing test `tests/test_components/test_simulation.py::test_warn_lumped_elements_outside_sim_bounds ` is strange: I got `ℹ️ ⚠️ RF simulations are subject to new license requirements` on both develop and my branch, but the warning is not captured on develop branch, and thus passing the test
- The other failing test `tests/test_components/test_simulation.py::test_proj_monitor_distance` is pointing to a more concerning issue: the more rigorous `frequency_range_sigma` based on `sigma` can be slightly narrower, and thus some monitor frequency is considered out of range. Should we use the union of `frequency_range` and `frequency_range_sigma` for validation purpose?

dbochkov-flexcompute:
overall looks good to me, but will give another more detailed pass

tidy3d/components/simulation.py
```diff
@@ -3351,7 +3351,7 @@ def _warn_monitor_simulation_frequency_range(cls, val, values):
         if val is None:
             return val
 
-        source_ranges = [source.source_time.frequency_range() for source in values["sources"]]
+        source_ranges = [source.source_time.frequency_range_sigma() for source in values["sources"]]
```
yaugenst-flex: Agree with https://github.com/flexcompute/tidy3d/pull/2800#issuecomment-3268741118 that we should just use a union for validation.

yaugenst-flex (approved):
Thanks @weiliangjin2021 looks good to me. Regarding the failing autograd test: this seems most likely to be a local env problem if we're not reproducing it in CI. I think switching from `dims=map(...)` to `dims=tuple(map(...))` should fix the error in any case though.

tidy3d/components/source/time.py
```diff
@@ -253,7 +414,131 @@ def from_frequency_range(
...
+
+        Parameters
+        ----------
+        fmin : float
+            Lower bound of frequency of interest.
+        fmax : float
+            Upper bound of frequency of interest.
+        min_rel_amp : float
+            A soft lower bound of amplitude relative to the peak amplitude in the frequency range [fmin, fmax].
+        kwargs : dict
```
momchil-flex: This is unused

tidy3d/components/source/time.py
```diff
@@ -63,10 +70,24 @@ def plot_spectrum(
     def frequency_range(self, num_fwidth: float = 4.0) -> FreqBound:
         """Frequency range within plus/minus ``num_fwidth * fwidth`` of the central frequency."""
 
+    def frequency_range_sigma(self, sigma: float = 4.0) -> FreqBound:
+        """Frequency range where the source intensity is within ``exp(-sigma)`` of the peak intensity."""
+        return self.frequency_range(num_fwidth=sigma)
```
momchil-flex, line 85: Here and in other places you seem to be using `num_fwidth=sigma`. Note that the `num_fwidth` frequency range for a simple Gaussian will be where the amplitude is larger than `exp(-num_fwdith**2 / 2)`. But I think it's just the docstring that needs to be changed.

tidy3d/components/source/time.py
```diff
@@ -162,7 +216,71 @@ def end_time(self) -> Optional[float]:
...
+
+    def frequency_range_sigma(self, sigma: float = 4.0) -> FreqBound:
+        """Frequency range where the source intensity is within ``exp(-sigma)`` of the peak intensity."""
+        if not self.remove_dc_component:
+            return self.frequency_range(num_fwidth=sigma)
+
+        # With dc removed, we'll need to solve for the nonlinear equation to find the frequency range
+        def equation_for_sigma_frequency(freq):
+            """computes I / I_p - exp(-sigma)"""
+            return np.abs(self._rel_amp_freq(freq)) ** 2 - np.exp(-sigma)
```
momchil-flex: Actually apart from that docstring, you probably want to change this here to `np.exp(-sigma**2 / 2)`?

tidy3d/components/source/time.py
```diff
@@ -253,7 +414,131 @@ def from_frequency_range(
...
+        f_intervals = [fmin]
+        while f_intervals[-1] < fmax:
+            pulse, fmax_iter = GaussianPulse._from_min_frequency(f_intervals[-1], min_rel_amp)
+            if fmax_iter > fmax or np.isclose(fmax_iter, fmax):
+                pulse = GaussianPulse.from_frequency_range(
+                    f_intervals[-1], fmax, disable_warning=True
+                )
+                fmax_iter = fmax
+            f_intervals.append(fmax_iter)
+            pulses.append(pulse)
```
momchil-flex, line 606: Could this get stuck indefinitely?

momchil-flex (commented):
Thanks Weiliang.

One thing I don't really understand is if there's any guarantee that the created pulse will be well-bounded in time. How do you ensure that?

weiliangjin2021:
> One thing I don't really understand is if there's any guarantee that the created pulse will be well-bounded in time. How do you ensure that?

It's linear summation of Gaussian pulses, so the endtime of the multi-Gaussian is bounded by the longest endtime of its contained Gaussian pulses?

tidy3d/components/source/time.py
weiliangjin2021: that's right!

weiliangjin2021:
> I think switching from dims=map(...) to dims=tuple(map(...)) should fix the error in any case though.

Thanks, that fixes it locally.

weiliangjin2021:
To @all: aside from using unionized frequency range in validators, I'll make a few additional changes:

- allow `freq0=0` in Gaussian pulse, which is unmodulated Gaussian. The actual central frequency for meshing, and lowest frequency validation should be computed from `frequency_range_sigma(sigma=1)`. The unmodulated Gaussian (but still DC removal) can be useful in broadband source, as discussed with @dbochkov-flexcompute 
- Some improvement to multiGaussian.from_frequency

The main headache is that the derivative Gaussian complicates lots of our notations in broadband. `freq0` and `fwidth` don't refer to what they are in our field description. But they are used everywhere in our code.

momchil-flex:
> > One thing I don't really understand is if there's any guarantee that the created pulse will be well-bounded in time. How do you ensure that?
> 
> It's linear summation of Gaussian pulses, so the endtime of the multi-Gaussian is bounded by the longest endtime of its contained Gaussian pulses?

So to put it another way, do we know that there won't be a pulse with a very narrow fwidth created (very broad time width)?

weiliangjin2021:
> So to put it another way, do we know that there won't be a pulse with a very narrow fwidth created (very broad time width)?

For the one from `from_frequency_range`, the fwidth of the narrowest one is a few times `fmin`, depending on the value of `min_rel_amp`. In a way it makes sense that you'll need the runtime to be at least the order of `1/fmin` to simulate the response at `fmin` well.

momchil-flex:
> > So to put it another way, do we know that there won't be a pulse with a very narrow fwidth created (very broad time width)?
> 
> For the one from `from_frequency_range`, the fwidth of the narrowest one is a few times `fmin`, depending on the value of `min_rel_amp`. In a way it makes sense that you'll need the runtime to be at least the order of `1/fmin` to simulate the response at `fmin` well.

So on one hand that does make sense to me. On the other, if `fmin` is small, this could require quite the large run time. I thought that we were also discussing with @dbochkov-flexcompute that in some cases, one can get the low-frequency response while simulating only a fraction of the cycle (and that's why he's been exploring the non-Fourier based methods to get the frequency-domain fields)?

weiliangjin2021:
> So on one hand that does make sense to me. On the other, if `fmin` is small, this could require quite the large run time. I thought that we were also discussing with @dbochkov-flexcompute that in some cases, one can get the low-frequency response while simulating only a fraction of the cycle (and that's why he's been exploring the non-Fourier based methods to get the frequency-domain fields)?

The new plan now is to apply extrapolation to the very low-frequency response in terminal modeler. So in terminal modeler, we will need to automate the division of the frequency range into the extrapolation region, and reliable FDTD simulation region. `fmin` in the source setup should be defined to the FDTD simulation region.

tidy3d/components/source/time.py
weiliangjin2021: fmax/fmin in each iteration is quite large (like ~10), so it usually takes just a very few iterations. To prevent edge cases, now a few more stopping criteria have been added.

weiliangjin2021:
I reverted the change to terminal.py: we should have a separate PR once/while the extrapolation is merged on how to setup the source. Otherwise, it's ready for review again @dbochkov-flexcompute

dbochkov-flexcompute:
Looks good to me, I am just not sure about `frequency_range_loose`:

> The other failing test tests/test_components/test_simulation.py::test_proj_monitor_distance is pointing to a more concerning issue: the more rigorous frequency_range_sigma based on sigma can be slightly narrower, and thus some monitor frequency is considered out of range. Should we use the union of frequency_range and frequency_range_sigma for validation purpose?

To me, switching to more rigorous frequency range is more a "bugfix" than a "breaking change" , so maybe no need to use the union of ranges afterall?

PR #2807, opened 2025-09-10
refc: backwards-compatible componentmodeler run() and autograd-support

yaugenst-flex:
## Greptile Summary

Updated On: 2025-09-10 08:44:14 UTC

This PR implements backwards-compatible component modeler `run()` functionality with autograd support. The changes consolidate the component modeler execution path under the unified `tidy3d.web.run()` API while maintaining backwards compatibility for existing code.

The core changes involve three main areas:

1. **API Consolidation**: The autograd `run()` function in `tidy3d/web/api/autograd/autograd.py` is modified to accept `ComponentModeler` objects by broadening the simulation parameter type from `td.Simulation` to `typing.Any`. This enables the function to handle both regular simulations and component modelers through a unified interface.

2. **Component Modeler Routing**: New routing logic detects `ComponentModeler` instances and routes them to the appropriate smatrix local run function when they contain autograd-valid simulations or when `local_gradient` is enabled. This leverages the existing smatrix infrastructure while enabling autograd capabilities.

3. **Documentation Cleanup**: The migration documentation is updated to remove references to deprecated legacy functionality, specifically removing the "legacy run command" section that documented `tidy3d.plugins.smatrix.run.run`. This aligns the documentation with the deprecation of the legacy helper function.

The integration fits well with the existing codebase architecture by reusing established patterns for type checking and routing to different execution paths. The changes maintain the existing `td.Simulation` autograd path unchanged, ensuring no breaking changes for current users while extending functionality to component modelers. The test updates demonstrate that the new API works correctly, with `modeler.run()` now returning smatrix data directly instead of requiring an additional `smatrix()` call.

## Important Files Changed

| `docs/api/plugins/smatrix_migration.rst` | 5/5 | Removes deprecated legacy run command documentation to align with API modernization |
| `tests/test_plugins/smatrix/test_component_modeler_autograd.py` | 4/5 | Updates tests to use new web.run() API and validates autograd functionality with component modelers |
| `tidy3d/web/api/autograd/autograd.py` | 4/5 | Adds component modeler support to autograd run function with type broadening and routing logic |

</details>

## Confidence score: 4/5

- This PR is generally safe to merge with some attention needed for the type system changes
- Score reflects solid architectural design but introduces runtime type checking instead of compile-time validation
- Pay close attention to the autograd.py file for potential type safety implications

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
```

### DIFF
--- a/tidy3d/plugins/smatrix/component_modelers/base.py
+++ b/tidy3d/plugins/smatrix/component_modelers/base.py
@@ -227,7 +227,8 @@ class AbstractComponentModeler(ABC, Tidy3dBaseModel):
     ):
         log.warning(
             "'ComponentModeler.run()' is deprecated and will be removed in a future release. "
-            "Use 'web.run(modeler)' instead.",
+            "Use web.run(modeler) instead. 'web.run' returns a 'ComponentModelerData' object; "
+            "get the scattering matrix via 'data.smatrix()'.",
             log_once=True,
         )
         from tidy3d.plugins.smatrix.run import _run_local
--- a/tidy3d/plugins/smatrix/component_modelers/base.py
+++ b/tidy3d/plugins/smatrix/component_modelers/base.py
@@ -225,14 +225,14 @@ class AbstractComponentModeler(ABC, Tidy3dBaseModel):
         local_gradient: bool = False,
         max_num_adjoint_per_fwd: int = MAX_NUM_ADJOINT_PER_FWD,
     ):
-        """Run the component modeler and return modeler data.
+        log.warning(
+            "'ComponentModeler.run()' is deprecated and will be removed in a future release. "
+            "Use 'web.run(modeler)' instead.",
+            log_once=True,
+        )
+        from tidy3d.plugins.smatrix.run import _run_local
 
-        Delegates to `tidy3d.plugins.smatrix.run.run`, which selects between
-        an autograd-compatible path and the standard Batch path.
-        """
-        from tidy3d.plugins.smatrix.run import run
-
-        return run(
+        data = _run_local(
             self,
             path_dir=path_dir,
             folder_name=folder_name,
@@ -244,6 +244,7 @@ class AbstractComponentModeler(ABC, Tidy3dBaseModel):
             local_gradient=local_gradient,
             max_num_adjoint_per_fwd=max_num_adjoint_per_fwd,
         )
+        return data.smatrix()
 
 
 AbstractComponentModeler.update_forward_refs()
--- a/tidy3d/plugins/smatrix/run.py
+++ b/tidy3d/plugins/smatrix/run.py
@@ -4,6 +4,7 @@ import json
 
 from tidy3d.components.base import Tidy3dBaseModel
 from tidy3d.components.data.index import SimulationDataMap
+from tidy3d.log import log
 from tidy3d.plugins.smatrix.component_modelers.modal import ModalComponentModeler
 from tidy3d.plugins.smatrix.component_modelers.terminal import TerminalComponentModeler
 from tidy3d.plugins.smatrix.component_modelers.types import ComponentModelerType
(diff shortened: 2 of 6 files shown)


tidy3d/web/api/autograd/autograd.py
```diff
@@ -94,7 +94,7 @@ def is_valid_for_autograd_async(simulations: dict[str, td.Simulation]) -> bool:
 
 
 def run(
-    simulation: td.Simulation,
+    simulation: typing.Any,
```
groberts-flex: I think before, this was WorkflowType before from components/types/workflow.py. Does that work to use here or does it create something circular?

groberts-flex: and I guess similarly, would we want return type to be WorkflowDataType?

groberts-flex (approved):
looks good to me, just a couple small questions! Thanks @yaugenst-flex

tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -225,14 +225,14 @@ def run(
...
     ):
-        """Run the component modeler and return modeler data.
-
-        Delegates to `tidy3d.plugins.smatrix.run.run`, which selects between
-        an autograd-compatible path and the standard Batch path.
-        """
-        from tidy3d.plugins.smatrix.run import run
+        log.warning(
+            "'ComponentModeler.run()' is deprecated and will be removed in a future release. "
+            "Use 'web.run(modeler)' instead.",
```
tylerflex: I think we can make this a bit more descriptive? A few things that might be worth mentioning:
1. `web.run()` returns a `ComponentModelerData` object (linked to docs) and the scattering matrix can be found by calling `.smatrix()` on it.
2. a link to the migration guide for more info

tidy3d/plugins/smatrix/run.py
```diff
@@ -191,7 +177,19 @@ def run(
         S-parameter extraction and analysis.
     """
 
-    if _is_autograd_valid(modeler):
+    # autograd path if any sim is valid for autograd
+    from tidy3d.web.api.autograd import autograd as web_ag
+
+    sims = getattr(modeler, "sim_dict", None) or {}
```
tylerflex: can we not assume that `sim_dict` exists?

tidy3d/plugins/smatrix/run.py
```diff
@@ -191,7 +177,19 @@ def run(
...
+    if any(web_ag.is_valid_for_autograd(sim) for sim in sims.values()):
+        if len(modeler.element_mappings) > 0:
+            raise AdjointError(
+                "Element mappings are not supported with autograd. We recommend "
+                "enforcing S-Matrix symmetry in the geometry creation and objective function and "
```
tylerflex: I actually hadn't considered this. So it seems like as long as the  `params -> ε(r)` mapping respects the symmetry that is reflected in the element mappings maybe everything just works properly? If that's true maybe we dont need to error after all and could just raise a warning to say that the gradients done consider the s-matrix elements set by the element mapping. so if you want to use autograd, you should focus on setting up your symmetries in the base simulation and parameterization rather than relying on element mappings to handle that for you? @groberts-flex what do you think?

tidy3d/plugins/smatrix/run.py
```diff
@@ -191,7 +177,19 @@ def run(
...
+                "running the unique simulations via the `run_only` specification in the component "
```
tylerflex: this should be the regular single quotes I think `'`

tidy3d/web/api/autograd/autograd.py
```diff
@@ -199,6 +199,30 @@ def run(
     if priority is not None and (priority < 1 or priority > 10):
         raise ValueError("Priority must be between '1' and '10' if specified.")
 
+    # component modeler path: route autograd-valid modelers to local run
+    from tidy3d.plugins.smatrix.component_modelers.types import ComponentModelerType
+
+    if isinstance(simulation, typing.get_args(ComponentModelerType)):
+        sims = getattr(simulation, "sim_dict", None)
+        if local_gradient or (
```
tylerflex: not sure about this logic, so if `local_gradient=True` for some reason but the component modeler is being run normally (eg not in a function being `autograd.grad`-ed.) then shouldn't this just use the remote pipeline? seems like the flow should be independent to what `local_gradient` is? this just controls how the grads are computed for each of the sims in sim_dict.

tylerflex (changes requested):

tidy3d/plugins/smatrix/run.py
daquinteroflex: If sim_dict doesn't exist there's a much bigger problem in the modeler definition?

tidy3d/plugins/smatrix/run.py
groberts-flex: I think that could definitely work. The only thing I am thinking is if there is a reason a user would need the element_mappings  in that case since they wouldn't be using the mapped port values. If we disallow, then they at least are aware that they need something that looks more like `run_only` + geometry/objective function symmetry. However, I could see it being nice for users who want to use the same arguments for their autograd and non-autograd runs in which case they would want to be able to use element_mappings to fill in the ports.

tidy3d/plugins/smatrix/run.py
yaugenst-flex: yup done

tidy3d/plugins/smatrix/run.py
yaugenst-flex: i changed it to a warning, wanna have another look?

tidy3d/web/api/autograd/autograd.py
yaugenst-flex: yeah this was a remnant from me playing around how to better decouple this from the plugins. removed

tidy3d/plugins/smatrix/component_modelers/base.py
yaugenst-flex: migration guide currently doesnt exist, but i included the wording change.

tylerflex (approved):

PR #2808, opened 2025-09-10
feat(FXC-1655): Add SSAC voltage source

damiano-flex:
Add SSAC Voltage source class.

## Greptile Summary

Updated On: 2025-09-10 09:22:23 UTC

This PR adds support for Small Signal AC (SSAC) voltage sources to the Tidy3D simulation framework. The implementation introduces a new `SSACVoltageSource` class that enables AC analysis in heat-charge simulations by providing DC bias voltages, AC frequencies, and small-signal amplitudes.

The core change is the addition of the `SSACVoltageSource` class in a new module (`tidy3d/components/spice/sources/ac.py`), which follows the established patterns from the existing `DCVoltageSource` class. The class includes three main parameters: `voltage` (array of DC bias voltages), `frequency` (array of AC analysis frequencies), and `amplitude` (small-signal AC magnitude). All fields include proper validation to ensure finite values and positive frequencies.

The implementation integrates seamlessly with the existing SPICE sources framework by updating the `VoltageSourceType` union to include the new AC source alongside the existing DC source. A validation constraint has been added to the `HeatChargeSimulation` class to enforce that only a single AC source can be present in a simulation's boundary conditions, which prevents ambiguous analysis scenarios.

The new functionality is exposed through the main package interface by adding the class to the imports and `__all__` list in `tidy3d/__init__.py`. This follows the established pattern where SPICE-like analysis components are made available as part of the public API.

## Important Files Changed

<details>

| `tidy3d/components/spice/sources/ac.py` | 4/5 | New file implementing SSACVoltageSource class for small-signal AC analysis with proper validation |
| `tidy3d/components/tcad/simulation/heat_charge.py` | 4/5 | Added validator to enforce single AC source constraint in simulations |
| `tests/test_components/test_heat_charge.py` | 5/5 | Added comprehensive validation tests for SSACVoltageSource and multiple AC source constraints |
| `tidy3d/components/spice/sources/types.py` | 5/5 | Updated VoltageSourceType union to include new SSACVoltageSource |
| `tidy3d/__init__.py` | 5/5 | Added SSACVoltageSource to main package imports and exports |

</details>

## Confidence score: 4/5

- This PR is safe to merge with minimal risk as it adds new functionality without modifying existing behavior
- Score reflects well-structured implementation following established patterns, though the single AC source constraint may need documentation
- Pay close attention to `tidy3d/components/tcad/simulation/heat_charge.py` to ensure the validation logic is complete

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
    participant HeatChargeSimulation
    participant VoltageBC
    participant SSACVoltageSource
    participant ValidationEngine

    User->>HeatChargeSimulation: "Create simulation with VoltageBC"
    HeatChargeSimulation->>VoltageBC: "Initialize VoltageBC"
    VoltageBC->>SSACVoltageSource: "Initialize SSACVoltageSource"
    SSACVoltageSource->>ValidationEngine: "Validate voltage array"
    ValidationEngine-->>SSACVoltageSource: "Check for infinite values"
    ValidationEngine-->>SSACVoltageSource: "Validation result"
    SSACVoltageSource-->>VoltageBC: "Source initialized"
    VoltageBC-->>HeatChargeSimulation: "Boundary condition created"
    HeatChargeSimulation->>ValidationEngine: "Validate single SSAC source"
    ValidationEngine-->>HeatChargeSimulation: "Check boundary specs for multiple AC sources"
    ValidationEngine-->>HeatChargeSimulation: "Validation result"
    HeatChargeSimulation-->>User: "Simulation created with SSAC voltage source"
```


tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,83 @@
...
+class SSACVoltageSource(Tidy3dBaseModel):
+    """
+    Small-signal AC voltage source.
+
+    This source represents a small-signal AC excitation defined by a list of bias voltages,
+    a set of small-signal analysis frequencies, and a small-signal amplitude (linear scale).
+
+    Notes
+    -----
+    - The bias `voltage` refer to the DC operating point above the simulation ground.
```
momchil-flex: It should be RST syntax here
```
 ``voltage``
```
As a rule, we don't use markdown notation in the docs. It's either RST (two "\`"-s) in docstrings, or single quotes `'` in warning/error messages that will be printed to stdout.

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,83 @@
...
+    ...     amplitude=amplitude,
+    ... )
+    """
+
+    name: Optional[str]
+
+    voltage: ArrayFloat1D = pd.Field(
+        ...,
+        title="DC Bias Voltages",
+        description="List of DC operating point voltages (above ground) used with 'VoltageBC'.",
```
momchil-flex: Also double \` here as it gets converted to a docstring.

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,83 @@
...
+    Small-signal AC voltage source.
+
+    This source represents a small-signal AC excitation defined by a list of bias voltages,
+    a set of small-signal analysis frequencies, and a small-signal amplitude (linear scale).
+
+    Notes
+    -----
+    - The bias `voltage` refer to the DC operating point above the simulation ground.
+    - The `frequency` array lists the AC analysis frequencies (in Hz).
+    - The `amplitude` is the small-signal AC magnitude (in Volts).
```
momchil-flex: Seems like the "frequency" and "amplitude" Note are not really needed given that the information is also given in the field description.

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,83 @@
...
+        ...,
+        title="DC Bias Voltages",
+        description="List of DC operating point voltages (above ground) used with 'VoltageBC'.",
+        units=VOLT,
+    )
+
+    frequency: ArrayFloat1D = pd.Field(
+        ...,
+        title="AC Analysis Frequencies",
+        description="List of small-signal analysis frequencies (in Hz).",
```
momchil-flex: No need to say the units in the description if they are defined in the `units` argument below, as they should be. See for example how this is parsed for the existing VoltageBC:

image

tidy3d/components/tcad/data/sim_data.py
```diff
@@ -84,6 +85,12 @@ class DeviceCharacteristics(Tidy3dBaseModel):
         "is given in Ohms. Note that in 2D the resistance is given in :math:`\\Omega \\mu`.",
     )
 
+    ac_current_voltage: Optional[FreqPotentialDataArray] = pd.Field(
+        None,
+        title="AC current vs voltage and frequency",
+        description="Complex small-signal current I(v, f) as a FreqPotentialDataArray.",
```
momchil-flex: Again, no need to say `as a FreqPotentialDataArray` here. Since the field type is `Optional[FreqPotentialDataArray]`, this information will already appear in the parsed docstring, including a clickable link to the type.

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1917,3 +1931,22 @@ def _useHeatSourceFromConductionSim(self):
...
+    def _get_ssac_frequency_and_amplitude(self):
+        for bc in self.boundary_spec:
+            if isinstance(bc.condition, VoltageBC):
+                if isinstance(bc.condition.source, SSACVoltageSource):
+                    return (
+                        bc.condition.source.frequency.tolist(),
+                        bc.condition.source.amplitude,
+                    )
+        raise SetupError(
+            "'HeatSimulation' does not include any `SSACVoltageSource` in 'boundary_spec')."
```
momchil-flex: \' everywhere in error messages.

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,83 @@
...
+            if v == td_inf or v <= 0:
+                raise ValueError(
+                    f"Frequencies must be strictly positive and finite. Current frequency={val}."
+                )
+        return val
+
+    @pd.validator("amplitude")
+    def validate_amplitude(cls, val):
+        if val == td_inf:
+            raise ValueError(f"Small signal amplitude must be finite. Current amplitude={val}.")
```
momchil-flex: Thinking out loud here, we want the amplitude to not just be finite, but "small" in some sense. Is there something that we can test it against and at least issue a warning if it doesn't seem to be the case?

marc-flex: For this type of analysis the perturbed output varies linearly (i.e., perturbation n, p, phi) with the amplitude so in principle in doesn't matter the amplitude. A different story is whether the solution is physically meaningful. I think it is, in general, not straightforward to define a validity limit and it may indeed be the case that it is case-dependent. 

It won't hurt, however, to include some warning or note in the docstring mentioning the validity of the simulation.

tidy3d/components/data/data_array.py
```diff
@@ -532,6 +532,24 @@ class FreqDataArray(DataArray):
...
+class FreqPotentialDataArray(DataArray):
+    """Frequency-domain array.
+
+    Example
+    -------
+    >>> f = [2e14, 3e14]
+    >>> v = [0.1, 0.2, 0.3]
+    >>> coords = dict(f=f, v=v)
+    >>> fd = FreqPotentialDataArray((1+1j) * np.random.random((2, 3)), coords=coords)
+    """
```
marc-flex: Do we want `FreqPotentialDataArray` or `FreqVoltageDataArray`? I think it'd be more consistent with the other classes with voltage

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,80 @@
...
+class SSACVoltageSource(Tidy3dBaseModel):
+    """
+    Small-signal AC voltage source.
+
+    Notes
+    -----
+    This source represents a small-signal AC excitation defined by a list of bias voltages,
+    a set of small-signal analysis frequencies, and a small-signal amplitude (linear scale).
+
+    The bias ``voltage`` refer to the DC operating point above the simulation ground. Currently, electrical ports are not defined.
```
marc-flex: - `refer` -> refers
- We don't really define a ground, do we? We define a voltage difference. Or have we changed this for SSAC?

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,80 @@
...
+    ...     amplitude=amplitude,
+    ... )
+    """
+
+    name: Optional[str]
+
+    voltage: ArrayFloat1D = pd.Field(
+        ...,
+        title="DC Bias Voltages",
+        description="List of DC operating point voltages (above ground) used with :class:`VoltageBC`.",
```
marc-flex, line 46: I guess this is related to my previous question. Do we define a ground in SSAC?

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1917,3 +1931,22 @@ def _useHeatSourceFromConductionSim(self):
...
+    def _get_ssac_frequency_and_amplitude(self):
+        for bc in self.boundary_spec:
+            if isinstance(bc.condition, VoltageBC):
+                if isinstance(bc.condition.source, SSACVoltageSource):
+                    return (
+                        bc.condition.source.frequency.tolist(),
+                        bc.condition.source.amplitude,
+                    )
+        raise SetupError(
+            "'HeatSimulation' does not include any 'SSACVoltageSource' in 'boundary_spec')."
```
marc-flex: `HeatSimulation` is deprecated. We should use `HeatChargeSimulation` instead

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1917,3 +1931,22 @@ def _useHeatSourceFromConductionSim(self):
         """Returns True if 'HeatFromElectricSource' has been defined."""
 
         return any(isinstance(source, HeatFromElectricSource) for source in self.sources)
+
+    def _get_voltage_regime(self):
+        for bc in self.boundary_spec:
+            if isinstance(bc.condition, VoltageBC):
+                if isinstance(bc.condition.source, SSACVoltageSource):
+                    return "ac"
+        return "dc"
```
marc-flex: I think we could generalize this to something like ` _get_charge_type()`.
Right now we only need to modify the name while we keep the body of the function but I can see how this can be also helpful as we move into transient (time-domain) simulations.

marc-flex (commented):
Thanks @damiano-flex 

I left a few comments but this is looking good!

I guess a more general comment, Do we want to include field output? Such as maybe perturbation current, perturbation potential etc?

tidy3d/components/data/data_array.py
damiano-flex: fair enough.

tidy3d/components/spice/sources/ac.py
damiano-flex: Yes, I had a discussion with @daquinteroflex about this. We need to discuss about on how and where let the user set the ground.

tidy3d/components/spice/sources/ac.py
daquinteroflex, line 1: Looks good re what we discussed 
Screenshot from 2025-09-18 11-44-15

I guess for this type of analysis we're not specifying the type of AC signal shape?
Screenshot from 2025-09-18 11-47-26

and so you're not envisioning adding a type? Or should we introduce signal shape types even if say this defaults to sine?

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,80 @@
...
+        description="List of small-signal analysis frequencies.",
+        units=HERTZ,
+    )
+
+    amplitude: pd.FiniteFloat = pd.Field(
+        ...,
+        title="Small-Signal Amplitude",
+        description="Small-signal AC amplitude.",
+        units=VOLT,
+    )
```
daquinteroflex: So yeah wondering if we should specify this via something like

```
class SinusodialSignalType
    frequency: ArrayFloat1D = pd.Field(
        ...,
        title="AC Analysis Frequencies",
        description="List of small-signal analysis frequencies.",
        units=HERTZ,
    )

    amplitude: pd.FiniteFloat = pd.Field(
        ...,
        title="Small-Signal Amplitude",
        description="Small-signal AC amplitude.",
        units=VOLT,
    )
```

and then:
```
class ACVoltageSource(Tidy3dBaseModel):

... 

 signal_parameters: SignalType
```

Just hacked this quickly but maybe also need to think about names

Apart from this rest of the PR looks good

daquinteroflex: Feel this would be pretty useful for the future too, also suspect maybe we already anyways have somthing like this for some time domain sources used elsewhere in the API

momchil-flex: I think that makes some sense @daquinteroflex, similarly to how we have our FDTD sources take a `source_time` which just defines the time dependence of the source, while the other parameters are related to the spatial extent, specific details of the modes, etc. On the other hand, doesn't amplitude make more sense to be an `ACVoltageSource` parameter than a `SinusoidalSignal` (because it's nice if the signal itself is dimensionless). But then that just leaves `frequency` in the `SinusoidalSignal`, so it makes it a bit pointless?

I don't think there's anything from the existing API we can reuse directly. One other thing I'm thinking about is that while this is technically a *source*, it also more closely matches the FDTD *monitor*-s. You can define a list of frequencies, and you will get output data that has a dimension over that list. This is exactly like our MonitorData. One point here is that we've been already inconsistent about the frequency labeling: we have `Monitor.freqs`, but the MonitorData has a corresponding `f` dimension. Shall we at least call `frequency` -> `freqs` here to be in line with our FDTD monitors?

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,80 @@
...
+    >>> frequency = [1e3, 1e4, 1e5]
+    >>> amplitude = 1e-3
+    >>> voltage_source = td.ACVoltageSource(
+    ...     voltage=voltages,
+    ...     frequency=frequency,
+    ...     amplitude=amplitude,
+    ... )
+    """
+
+    name: Optional[str]
```
momchil-flex: What is the purpose of the `name`?

If it's needed, it should be formatted similarly to e.g. [this](https://github.com/flexcompute/tidy3d/blob/6e1ff2249daac150d12ffd043b52567ac2655db8/tidy3d/components/base_sim/monitor.py#L20-L25)

tidy3d/components/spice/sources/ac.py
daquinteroflex: So am thinking eventually we probably want to implement all these different ways of specifying time domain SPICE signal sources (either for voltage or current). 

But feel maybe we need to be more specific whether we want to use an `ACVoltageSource` or simply all the specific signal shapes we want to support directly. I think @damiano-flex made it this way because of the way voltage sources are specified in SPICE

```
VXXXXXXX N+ N- <<DC> DC/TRAN VALUE> <AC <ACMAG <ACPHASE>>>
+ <DISTOF1 <F1MAG <F1PHASE>>> <DISTOF2 <F2MAG <F2PHASE>>>

translates to

name positive_node negative_node
[this specifies a DC bias <<DC> DC/TRAN VALUE>]
[this specifies a custom AC source <AC <ACMAG <ACPHASE>>> + <DISTOF1 <F1MAG <F1PHASE>>> <DISTOF2 <F2MAG <F2PHASE>>>]
```
image

So in this sense feeling this is why a `ACSignalType` should be defined as a standalone class. See 4.1 in https://ngspice.sourceforge.io/docs/ngspice-44-manual.pdf These are some examples of how they can be defined:

Based on these types of signal definitions, in my opinion I actually do think ampltude does make sense to live say in `SinusosidalSignal` (not my favorite name yet just to be consistent), because say a `PulseSignal` can be specifed between two voltages rather than just a single magnitude for example, and other AC sources require different parameters independent of amplitude too yet are still "AC" source compatible.

Sounds good, was not sure either what exactly was reusable. I do agree for consistency we can use `freqs`.

Screenshot from 2025-09-19 11-08-33
Screenshot from 2025-09-19 11-08-23
Screenshot from 2025-09-19 11-08-09

tidy3d/components/spice/sources/ac.py
damiano-flex: I had no purposes for it in mind, I just used the same structure of `DCVoltageSource` (which has a field formatted like this one) . But in principle, having a name could be useful for future usage (e.g. a function that modify a source value through the name).

So, either I remove it in both, or I format it correctly here and in `DCVoltageSource` well.

momchil-flex: Yeah, let's format both, thanks.

tidy3d/components/spice/sources/ac.py
damiano-flex: OK to use `freqs` for consistency.

momchil-flex:
@damiano-flex @marc-flex I am getting confused with the `source` types and the `analysis` types. Would one still be defining e.g. `IsothermalSteadyChargeDCAnalysis` when there are `AC` type sources?

tidy3d/components/spice/sources/ac.py
damiano-flex: Regarding the signal aspect.

Devsim, in principle, can accept the phase of the ac source as well. So it could fit into the signal definition. To me, it'd feel more natural having frequency, amplitude and phase all parts of the signal definition. But I am biased by knowing how it is implemented internally.

This said, I am fine keeping them as separate parameters of `ACVoltageSource` for the moment.

damiano-flex:
> @damiano-flex @marc-flex I am getting confused with the `source` types and the `analysis` types. Would one still be defining e.g. `IsothermalSteadyChargeDCAnalysis` when there are `AC` type sources?

Initially, I had a different Analysis type for AC analysis. It was virtually a clone. Marc suggested to keep old analysis types, and define AC at source level, and this idea seemed good to me as well. If you have an AC source, you'll get everything from the DC simulation + a graph for AC current.

An option might be that `IsothermalSteadyChargeDCAnalysis` drops the `DC` keyword or change its name, since: 1) the DC analysis will be **always** performed 2) the AC analysis will be performed *after* the DC analysis **if** an AC source has been defined.

tidy3d/components/spice/sources/ac.py
momchil-flex: So like I mentioned, in our FDTD sources, we have separated the time-signal specific settings in [these types](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/source/time.py) of classes. Those are definitely related to what we're doing here, but I was saying that I don't think our existing API can be used, because there's still some assumptions there that this is geared towards EM simulations. For example, the `ContinuousWave` class is basically a `Sinusoidal`, but, first of all, it's complex-valued, and second of all, it includes a linear ramp-up based on the `fwidth` parameter.

That said, on second thought, if we do decide to separate time-only (unitless) definitions, this file could be a good place to put some new classes.

In what @daquinteroflex is showing I still see a number of parameters which relate to the time only (FREQ, TD, THETA, etc.) as well as one or two amplitudes. For something as simple as a base amplitude and an offset amplitude where the source is something like `V0 + V1*Sinusoidal`, it still makes sense to me to follow this approach of defining a time-only class. I don't know if some of the above examples are not so easily separable though.

momchil-flex:
> > @damiano-flex @marc-flex I am getting confused with the `source` types and the `analysis` types. Would one still be defining e.g. `IsothermalSteadyChargeDCAnalysis` when there are `AC` type sources?
> 
> Initially, I had a different Analysis type for AC analysis. It was virtually a clone. Marc suggested to keep old analysis types, and define AC at source level, and this idea seemed good to me as well. If you have an AC source, you'll get everything from the DC simulation + a graph for AC current.
> 
> An option might be that `IsothermalSteadyChargeDCAnalysis` drops the `DC` keyword or change its name, since: 1) the DC analysis will be **always** performed 2) the AC analysis will be performed _after_ the DC analysis **if** an AC source has been defined.

I think that makes sense. Note that, for backwards compatibility, we'd keep `IsothermalSteadyChargeDCAnalysis` but issue a deprecation warning when using it. So something like

```
class IsothermalSteadyChargeAnalysis():
    # takes everything that's currently in IsothermalSteadyChargeDCAnalysis

class IsothermalSteadyChargeDCAnalysis(IsothermalSteadyChargeAnalysis):
    # add deprecation warning
```

Refer to the [HeatSimulation](https://github.com/flexcompute/tidy3d/blob/6e1ff2249daac150d12ffd043b52567ac2655db8/tidy3d/components/tcad/simulation/heat.py#L50-L57).

tidy3d/components/spice/sources/ac.py
daquinteroflex: Ahh yes, sounds good. 

>  Those are definitely related to what we're doing here, but I was saying that I don't think our existing API can be used, because there's still some assumptions there that this is geared towards EM simulations. For example, the ContinuousWave class is basically a Sinusoidal, but, first of all, it's complex-valued, and second of all, it includes a linear ramp-up based on the fwidth parameter.  For something as simple as a base amplitude and an offset amplitude where the source is something like V0 + V1*Sinusoidal, it still makes sense to me to follow this approach of defining a time-only class. I
> That said, on second thought, if we do decide to separate time-only (unitless) definitions, this file could be a good place to put some new classes. 

This seems a good approach in terms of API consistency, basically these abstract time classes would be an equivalent representation, and we will be able to translate to SPICE types if we need eventually too. Also worth getting @nikoladjor-flex input here. So creating those abstract time-shared classes seems good, and they can be used within ACVoltageSource and or eventually ACCurrentSource too in this approach. I do think it makes sense to define signal classes separate from the kind of physical quantity like Voltage that they modulate for reusability. Not sure what you think about that @damiano-flex ?

tidy3d/components/data/data_array.py
daquinteroflex: I vaguely remember this was already introduced for RF hmm

daquinteroflex: Fyi https://github.com/flexcompute/tidy3d/blob/6e1ff2249daac150d12ffd043b52567ac2655db8/tidy3d/components/data/data_array.py#L1363-L1375

daquinteroflex: This was all recently added fyi https://github.com/flexcompute/tidy3d/blob/6e1ff2249daac150d12ffd043b52567ac2655db8/tidy3d/components/data/data_array.py#L1344-L1500

marc-flex: Thanks @daquinteroflex!
Last time I spoke to @dmarek-flex we kind of reserved "voltage" to whenever we're comparing potentials and "potential" when we talk about the field. But maybe @dmarek-flex can confirm

tidy3d/components/spice/sources/ac.py
damiano-flex: @daquinteroflex as we discussed in Slack, I don't have a strong opinion about what's the right thing to do. Here's what i think though.

Having a class `SinusoidalSignal` is not something inherently wrong to me, but providing a **vector** of frequencies to this class is. Classes in `time.py`  are made to generate a **single** time-signal each, either by specifying some parameters (`ContinousWave`) or by specifying the value of the signal **in time domain**. Providing a vector of frequencies here still seems a bit off to me.

daquinteroflex: I agree, a vector would be weird in this case since it doesn't make sense physical sense unless it is a collection of signals. Feel this is a byproduct of wanting to integrate a sweep at the same time as a source potentially in the initial implementation, maybe instead of separating it into the analysis. Personally of the opinion parameter sweeps analysis to be separate from the source definition anyway

marc-flex: That's a good point @daquinteroflex. Maybe we could allow to have an array of ac sources?

marc-flex:
@damiano-flex @momchil-flex I'm now having second thoughts about the current naming of sources and analysis types.

Right now we're saying that if we see an `ACVoltageSource` in a boundary condition we'll perform an SSAC simulation. But what if, in the future we can do non-linear simulations with AC sources? I guess we have two way forward:

1. We create small-signal equivalents of the `SteadyChargeAnalysis` and `IsothermalSteadyChargeAnalysis` as @damiano-flex suggested in the first place; or
2. We create `SSACVoltageSource`. I guess the only issue with this one is that it isn't very SPICE like

daquinteroflex:
These type of charge simulations are different from FDTD simulations in the sense that a given "scene" with sources etc. can have different types of analysis depending on what you're looking into like transient or DC operating point, and so the data generated is inherently different per analysis type.

In any case:
> We create small-signal equivalents of the SteadyChargeAnalysis and IsothermalSteadyChargeAnalysis as @damiano-flex suggested in the first place; or

would be then equivalent to this type of SS ac SPICE analysis directive that Damiano mentioned separately, but then the decision is whether the analysis type has the frequency sweep or the source. IMO it should be the analysis but use the source name to reference to it, and override any specified frequency in it.
```
.ac dec nd fstart fstop
.ac oct no fstart fstop
.ac lin np fstart fstop
```

tidy3d/components/data/data_array.py
dmarek-flex: yea, same units but conceptually fairly different. Not a strong feeling, but I do think that `FreqPotentialDataArray` is a better description that also differentiates it more from the `VoltageArray`s we introduced in RF.

Sadly  if we don't modify we will have `FreqVoltageDataArray` and `VoltageFreqDataArray` :disappointed:

dmarek-flex: We still don't have an enforced naming scheme for `DataArray`, but one idea is to have the leading term prefix relate to the quantity (Voltage/Current/Potential) and the remaining terms relate to the coordinates (Freq/Mode/Spatial). So maybe follow that approach

dmarek-flex: Actually I might be confused about the purpose of `FreqVoltageDataArray`. The coordinates are frequency and voltage? In that case, I agree with the current approach :smile:

damiano-flex: Yes, coordinates are frequency and voltage.

tidy3d/components/spice/sources/ac.py
nikoladjor-flex: Hey, the things I can add at the moment are more related to the application and I might be completely off-topic and wrong:

- small signal analysis is usually done to determine frequency behaviour when you do not know the impedance of the device and want to get the bandwidth. Two most obvious applications for PIC are p/n modulators and photodetectors. Having array of sources here might help if there is a step in the simulation with brings the system to a steady state and then you apply small signals separately.
- from the API point of view, this analysis makes sense on a single component, because you want three terminals the most. I tend to agree with @daquinteroflex about the abstract ACSource with added offset and frequency as list maybe. The point with the array of sources is the same as above. My main remark is that from the point of view of setting this analysis up, one needs to know which structures get what type of signal, and if we are already using BCs for that, it's fine, but we need to provide easy way to go from the circuit to the BC definition and from analysis result to impedance.

momchil-flex:
> @damiano-flex @momchil-flex I'm now having second thoughts about the current naming of sources and analysis types.

Yeah I don't really know the answer but we must 100% think about it right now before things are released. It does seem to me that the current approach will be becoming more and more tedious/confusing as new regimes are added.

Maybe try to sketch what other analysis we want to support and how what the API might look like - even if it's for one or two years from now. Maybe let's take the discussion off this repo, too.

docs/lectures/small_signal_analysis.rst
```diff
@@ -0,0 +1,89 @@
...
+    
+    ac_source = td.ACVoltageSource(
+        voltage=bias_voltages,
+        amplitude=ac_amplitude
+    )
+
+SteadyChargeDCAnalysis with SSAC
+------------------------------
+
+The :class:`SteadyChargeDCAnalysis` class now supports small signal analysis through the ``ssac_freqs`` parameter:
```
marc-flex: Here I'd remove "now". It seems new today but it won't in a few months?

docs/lectures/small_signal_analysis.rst
```diff
@@ -0,0 +1,89 @@
...
+    )
+    simulation = td.HeatChargeSimulation(
+        analysis_spec=analysis,
+        # ... other parameters
+    )
+
+Best Practices
+==============
+
+1. **Ground Reference**: Specify exactly one ground reference using :class:`GroundVoltageSource`. If no ground is specified, the smallest voltage source will be considered as the ground voltage.
```
marc-flex: This isn't entirely true, right? If no ground is provided, "ground" will be the BC without an array of voltages?

docs/lectures/small_signal_analysis.rst
```diff
@@ -0,0 +1,89 @@
...
+
+2. **Small Signal Amplitude**: Keep the AC amplitude small (typically 1-10 mV) to ensure linear operation.
```
marc-flex: The only reason why you'd keep the amplitude small is so that the solution has some physical meaning. I think you could say something like this instead: `Note that since this is a linear analysis tool, the simulated AC current and voltages will be proportional to the amplitude and may not have physical meaning for large enough amplitudes.`

docs/lectures/small_signal_analysis.rst
```diff
@@ -0,0 +1,89 @@
...
+4. **Convergence**: Use appropriate tolerance settings for the charge analysis to ensure convergence.
+
+5. **Validation**: The simulation will automatically validate that one :class:`ACVoltageSource` is present when ``ssac_freqs`` is specified.
+
+Common Use Cases
+================
+
+- **MOSFET Analysis**: Study transconductance and output conductance vs. frequency.
+- **Diode Characterization**: Analyze junction capacitance and conductance.
+- **BJT Small-Signal**: Compute h-parameters and frequency response.
```
marc-flex: My only problem with saying this is that the users may expect the corresponding notebook examples, so I would refrain from saying this, at least for now. 
what do you think @momchil-flex

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,67 @@
...
+class ACVoltageSource(Tidy3dBaseModel):
+    """
+    Small-signal AC voltage source.
+
+    Notes
+    -----
+    This source represents a small-signal AC excitation defined by a list of bias voltages,
+    and a small-signal amplitude (linear scale). The list of frequencies of interest must be supplied to the analysis class.
+
+    The bias ``voltage`` refers to the DC operating point above the simulation ground. Currently, full circuit simulation though electrical ports is not supported.
```
marc-flex: `full circuit simulation though electrical ports` -> `full circuit simulation through electrical ports`?

marc-flex: So this is a little tricky. Right now this class is called `ACVoltageSource` which in principle could be used for either linear or non-linear analysis but everything in this class is done assuming SSAC. So I guess we have two options:

1. We change the name to something like `SSACVoltageSource` or 
2. we keep the `ACVoltageSource` as pure AC source but we treat it as AC since at the moment it can only be defined in simulations where `ssac_freq` has been defined in the steady analysis spec

The problem with 2 is that in the long term we may want to support AC sources with arbitrary shapes so we'd need to modify this.

tidy3d/components/spice/sources/dc.py
```diff
@@ -68,6 +74,28 @@ def check_voltage(cls, val):
...
+
+    This source explicitly sets the ground reference (0V) for the simulation.
+    It is equivalent to DCVoltageSource(voltage=0) but more explicit about
+    establishing the ground reference.
+
+    Notes
+    -----
+    This establishes the ground reference for the simulation. All other voltage sources
+    will be referenced to this ground potential. If no GroundVoltageSource is specified,
+    the smallest voltage among all sources becomes the ground reference.
```
marc-flex: Similar comment to above. The ground (or reference) is the one without voltage arrays?

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -627,6 +648,48 @@ def check_only_one_voltage_array_provided(cls, val, values):
...
+
+    @pd.root_validator(skip_on_failure=True)
+    def check_ssac_freqs_requires_ac_source(cls, values):
+        """Ensure that if ssac_freqs is provided, at least one ACVoltageSource is present."""
+        analysis_spec = values.get("analysis_spec")
+        if (
+            analysis_spec
+            and hasattr(analysis_spec, "ssac_freqs")
+            and len(analysis_spec.ssac_freqs) > 0
+        ):
```
marc-flex, line 665: I think in general we kind of avoid `hasattr` though maybe @momchil-flex can comment.
As an alternative you could replace by a `isinstance` if?
```
if isinstance(analysis_spec, SteadyChargeAnalysis) and len(analysis_spec.ssac_freqs) > 0:
```

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1928,3 +1991,22 @@ def _useHeatSourceFromConductionSim(self):
...
+                if isinstance(bc.condition.source, ACVoltageSource):
+                    return "ac"
+        return "dc"
+
+    def _get_ssac_frequency_and_amplitude(self):
+        for bc in self.boundary_spec:
+            if isinstance(bc.condition, VoltageBC):
+                if isinstance(bc.condition.source, ACVoltageSource):
+                    return (
+                        bc.condition.source.freqs.tolist(),
```
marc-flex: `ACVoltageSource` no longer has frequencies, correct?

tidy3d/components/spice/sources/dc.py
damiano-flex: That is actually true. I was just thinking about the case with one single BC for each contact.

docs/lectures/small_signal_analysis.rst
damiano-flex: That's right.

docs/lectures/small_signal_analysis.rst
damiano-flex: I think that once this is merged, notebooks can be produced with these examples. But in principle, you could already do these things already.

I have a notebook with SSAC that can be released (after few modifications) after this comes out.

tidy3d/components/spice/sources/ac.py
damiano-flex: We could discuss API alternatives in the Lark document. Each approach has his own strength and weakness.

tidy3d/components/tcad/simulation/heat_charge.py
damiano-flex: right.

tidy3d/components/tcad/simulation/heat_charge.py
momchil-flex: Yes, writing it in this more manifest way is preferred.

docs/lectures/small_signal_analysis.rst
momchil-flex: I'm not necessarily opposed to mentioning use cases before we have an example, but these use cases (apart from the diode characterization) are not our target use cases and we're unlikely to work on an example showcasing our charge solver on a MOSFET for example. The relevant devices for us would be modulators and photodetectors, for example.

docs/lectures/small_signal_analysis.rst
```diff
@@ -0,0 +1,89 @@
+Small Signal Analysis Guide |:zap:|
```
momchil-flex: I like this file. However, it is not in the right place. Our [lectures](https://docs.flexcompute.com/projects/tidy3d/en/latest/lectures/index.html) page just links to the *video seminar* series that we have released. So far, we use a combination of the API reference and the main notebook introducing a given feature to provide the information that you have compiled here (e.g. the [HeatChargeSimulation](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.HeatChargeSimulation.html#tidy3d.HeatChargeSimulation) reference and the original [charge solver notebook](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/ChargeSolver.html).

Again, I like the idea of this file, and it could be nice if some day we have such "feature white papers". But given that it's a loner right now, probably this info should be moved to the components and the upcoming example.

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,67 @@
...
+        ...,
+        title="DC Bias Voltages",
+        description="List of DC operating point voltages (above ground) used with :class:`VoltageBC`.",
+        units=VOLT,
+    )
+
+    amplitude: pd.FiniteFloat = pd.Field(
+        default=1.0,
+        title="Small-Signal Amplitude",
+        description="Small-signal AC amplitude. If not specified, it is assumed to be unitary.",
```
momchil-flex: Maybe you mean `unity`?

Should we default to a small value like 0.001 instead?

tidy3d/components/spice/analysis/dc.py
```diff
@@ -78,6 +81,14 @@ class SteadyChargeDCAnalysis(Tidy3dBaseModel):
         "either the conduction or valence energy bands.",
     )
 
+    ssac_freqs: Optional[ArrayFloat1D] = pd.Field(
```
momchil-flex: So your suggestion is for this spec to eventually become `SteadyChargeAnalysis`?

If so, you could introduce it already:

```
class SteadyChargeDCAnalysis(Tidy3dBaseModel):
    # same as before, maybe a new deprecation message on init

class SteadyChargeAnalysis(SteadyChargeDCAnalysis):
    ssac_freqs: Optional[ArrayFloat1D] = pd.Field(...)
```

tidy3d/components/spice/sources/dc.py
```diff
@@ -78,7 +108,13 @@ class DCCurrentSource(Tidy3dBaseModel):
     >>> current_source = td.DCCurrentSource(current=0.4)
     """
 
-    name: Optional[str]
+    name: Optional[str] = pd.Field(
+        None,
+        title="Name",
+        description="Unique name for the DC current source",
```
momchil-flex, line 114: Do we validate that the names are unique?

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -468,6 +476,19 @@ def check_voltage_array_if_capacitance(cls, values):
...
 
+    @pd.root_validator(skip_on_failure=True)
+    def check_single_ssac(cls, values):
+        boundary_spec = values["boundary_spec"]
+        ssac_present = False
+        for bc in boundary_spec:
+            if isinstance(bc.condition, VoltageBC):
+                if isinstance(bc.condition.source, ACVoltageSource):
+                    if ssac_present:
+                        raise SetupError("Only a single AC source can be supplied")
```
momchil-flex: Missing period at the end of the message

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1928,3 +1987,38 @@ def _useHeatSourceFromConductionSim(self):
...
+                "to be defined in the 'boundary_spec', but none was found."
+            )
+
+        if len(self.analysis_spec.ssac_freqs) > 0:
+            return (self.analysis_spec.ssac_freqs, amplitude)
+        else:
+            raise SetupError(
+                "Frequency list for SSAC analysis is missing. "
+                "Please specify the frequencies to be analyzed, in the 'analysis_spec.ssac_freqs' attribute."
+            )
```
momchil-flex: I'm not sure about all the error checking here. Shouldn't these things have been caught already? I guess it doesn't *hurt* to have extra checks, but if any of those are hit then something's wrong with the initial validation?

momchil-flex (commented):
Some initial comments. But I think we should consider @daquinteroflex 's proposal in the Lark document, if not implementing fully then at least some aspect of it. I'll put down some more thoughts a bit later.

tidy3d/components/tcad/simulation/heat_charge.py
damiano-flex: the new version implements already the Marc's approach.

tidy3d/components/spice/sources/ac.py
damiano-flex: unitary does not look wrong to me, as it is an adjective that means "of or relating to a unit". But to be fair I just searched it now :smiley:. I just wrote it assuming it to be correct to me. But can switch to "unity" if you don't like it.

Yeah, we could actually default to 1mV.

tidy3d/components/spice/sources/dc.py
damiano-flex: No, since names are not used anywhere basically. They're just introduced for future (probably near) future. This said, validating the uniqueness regardless seems a good idea to me.

tidy3d/components/tcad/simulation/heat_charge.py
damiano-flex: That's a very good point linked to the API discussion (I hate it sounds like a Gemini response).

Currently, the error error cannot caught earlier because: 
- The AC source is provided as a BC.
- `ssac_freqs` is provided to the analysis type (and not to the AC source as in the earlier implementation).

Therefore, the only place where you can have this check is when you assemble the `HeatChargeSimulation`. Because if there's no AC source, you don't actually need `ssac_freqs`.

tidy3d/components/spice/analysis/dc.py
damiano-flex: Given the ongoing discussion in the lark document, I'll postpone this to the moment a final decision on SSAC analysis class is taken. If SSAC will eventually go to a new class, then we could keep DC actually.

docs/lectures/small_signal_analysis.rst
damiano-flex: That's right. OK, I'll create an example directly into `HeatChargeSimulation`, and will save the rest of the information for the upcoming notebook.

docs/lectures/small_signal_analysis.rst
damiano-flex: I see. Given the other discussion though, this file will go away.

tidy3d/components/spice/sources/ac.py
momchil-flex: To be honest you could just drop the sentence? The default is the parsed docstring so no need to say anything.

My problem with `unitary` is that to me it sounds like a unitary matrix, or related to that a unitary time evolution, which has a completely unrelated meaning (but e.g. in some contexts unitary kind of means "lossless", so just want to avoid confusion).

tests/test_components/test_frequencies.py
```diff
@@ -254,3 +254,57 @@ def test_gaussian_pulse():
...
+    Test the sweep_decade method for generating logarithmically spaced frequencies.
+    """
+    # Test basic functionality
+    freq_range = td.FreqRange.from_freq_interval(1e3, 1e6)  # 1 kHz to 1 MHz (3 decades)
+    freqs = freq_range.sweep_decade(10)  # 10 points per decade
+
+    # Check that frequencies are logarithmically spaced
+    assert len(freqs) > 0
+    assert np.isclose(freqs[0], freq_range.fmin)
+    assert np.isclose(freqs[-1], freq_range.fmax)
```
marc-flex: Here you aren't really checking that the frequencies are logarithmically spaced, right? You could for instance check something like this?
```python
for i, f  in enumerate(freqs):
    np.log10(f/freqs[i-1]) == 1/10  # 10 points per decade
np.log10(ffreq_rage.fmax) - np.log10(ffreq_rage.fmin)  == 3  # 3 decades
```

then the rest of assers below wouldn't be necessary?

tests/test_components/test_frequencies.py
```diff
@@ -254,3 +254,57 @@ def test_gaussian_pulse():
...
+def test_sweep_decade_edge_cases():
+    """
+    Test edge cases and error conditions for sweep_decade.
+    """
+    # Test with single decade
+    freq_range = td.FreqRange.from_freq_interval(1e3, 1e4)  # 1 decade
+    freqs = freq_range.sweep_decade(5)
+    assert len(freqs) >= 5  # Should have at least 5 points
+    assert np.isclose(freqs[0], freq_range.fmin)
+    assert np.isclose(freqs[-1], freq_range.fmax)
```
marc-flex, line 301: Similar to the above comment

tests/test_components/test_heat_charge.py
```diff
@@ -1482,6 +1638,41 @@ def test_charge_simulation(
...
+            condition_p_ground = bc_p.condition.updated_copy(source=td.GroundVoltageSource())
+            sim.updated_copy(
+                boundary_spec=[
+                    bc_n.updated_copy(condition=condition_n_ground),
+                    bc_p.updated_copy(condition=condition_p_ground),
+                ]
+            )
+
+        # Two AC sources cannot be defined
+        with pytest.raises(pd.ValidationError, match="Only a single AC source can be supplied."):
```
momchil-flex: Given that there's this can `test_heat_charge_multiple_ac_sources` be removed?

tidy3d/__init__.py
```diff
@@ -20,12 +20,14 @@
 from tidy3d.components.microwave.data.monitor_data import (
     AntennaMetricsData,
 )
+from tidy3d.components.spice.analysis.ac import IsothermalSSACAnalysis
```
momchil-flex: Help me understand how we can also enable SSAC for non-isothermal. Can we do non-isothermal DC and then isothermal SSAC on top of it, and then it is pretty straightforward? Or is it more complicated?

tidy3d/components/frequencies.py
```diff
@@ -436,6 +436,54 @@ def freqs(self, num_points: int, spacing: str = "uniform_freq") -> NDArray[np.fl
                 f"Received: {spacing!r}. Please provide a valid spacing type."
             )
 
+    def sweep_decade(self, num_points_per_decade: int) -> NDArray[np.float64]:
```
momchil-flex, line 439: So with this, we are still missing a `log` sweep option? Basically we can do a linear sweep with `.freqs` and a decade sweep with this new function. Should we add a `log` option to `.freqs` and `.wvls`?

tests/test_components/test_heat_charge.py
```diff
@@ -897,6 +897,113 @@ def test_heat_charge_bcs_validation(boundary_conditions):
...
+            ),
+            td.HeatChargeBoundarySpec(
+                placement=td.StructureStructureInterface(structures=["cathode", "silicon"]),
+                condition=td.VoltageBC(source=td.GroundVoltageSource()),
+            ),
+        ],
+        grid_spec=td.UniformUnstructuredGrid(dl=0.1),
+        monitors=[volt_monitor],
+        analysis_spec=isothermal_spec,
+    )
```
marc-flex, line 966: Maybe you can create this simulation object first so that in the test above (line 952, `with pytest...`) you can simply update the simulation with new BCs. That way we can save a little bit of code

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,112 @@
...
+
+class SinusoidalSignal(Tidy3dBaseModel):
+    """
+    A transient sinusoidal signal.
+
+    Notes
+    -----
+    The parameter ``frequency`` is set for future usage,
+     and is not employed in small-signal analysis
+    (See :class:`IsothermalSSACAnalysis`).
```
momchil-flex: I think you need a `.` for this to render correctly if the class name is not in the same file, i.e.

```
:class:`.IsothermalSSACAnalysis`
```

With the dot, it will find it in the `tidy3d` namespace.

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class SinusoidalSignal(Tidy3dBaseModel):
+    """
+    A transient sinusoidal signal.
+
+    Notes
+    -----
+    The parameter ``frequency`` is set for future usage,
+     and is not employed in small-signal analysis
```
momchil-flex: Let's not talk about future usage. Just say that the `frequency` parameter is not used in the SSAC.

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,112 @@
...
+        default=1.0,
+        title="Signal Amplitude",
+        description="Amplitude of the sinusoidal signal",
+        units=VOLT,
+    )
+
+    frequency: Optional[pd.PositiveFloat] = pd.Field(
+        default=None,
+        title="Frequency",
+        description="Frequency of the sinusoidal signal.",
```
momchil-flex: "When using SSAC analysis, the value is optional, and even if provided it is overwritten by the 'ssac_freqs' in the analysis specification."

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -1928,3 +1991,32 @@ def _useHeatSourceFromConductionSim(self):
...
+        for bc in self.boundary_spec:
+            if isinstance(bc.condition, VoltageBC):
+                if isinstance(bc.condition.source, ACVoltageSource):
+                    amplitude = bc.condition.source.signal.amplitude
+        if amplitude is None:
+            raise SetupError(
+                "AC voltage source not found. "
+                "A Small-Signal AC (SSAC) analysis requires one 'ACVoltageSource' "
+                "to be defined in the 'boundary_spec', but none was found."
+            )
```
momchil-flex: Sorry I'm still confused by these checks here. If the simulation has been correctly initialized and the validators have passed, how can these errors ever be hit?

tidy3d/components/spice/analysis/ac.py
```diff
@@ -0,0 +1,45 @@
...
+
+import pydantic.v1 as pd
+
+from tidy3d.components.spice.analysis.dc import IsothermalSteadyChargeDCAnalysis
+from tidy3d.components.types import ArrayFloat1D
+from tidy3d.constants import HERTZ
+from tidy3d.constants import inf as td_inf
+
+
+class IsothermalSSACAnalysis(IsothermalSteadyChargeDCAnalysis):
```
marc-flex: We should also include a non isothermal SSAC analysis type, right?

tidy3d/components/spice/sources/ac.py
marc-flex: Maybe just mention that in SS analysis this frequency is ignored in favor of the frequencies defined in the analysis spec.

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,112 @@
...
+
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.components.types import ArrayFloat1D
+from tidy3d.constants import HERTZ, VOLT
+from tidy3d.constants import inf as td_inf
+
+
+class SinusoidalSignal(Tidy3dBaseModel):
+    """
+    A transient sinusoidal signal.
```
marc-flex: Maybe just `Sinusoidal signal in the time domain`?

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,112 @@
...
+            raise ValueError(f"Signal frequency must be finite. Current frequency={val}.")
+        return val
+
+
+ACSignalType = Union[SinusoidalSignal]
+
+
+class ACVoltageSource(Tidy3dBaseModel):
+    """
+    Small-signal AC voltage source.
```
marc-flex: This isn't SS-specific, right? We should be able to use this in non linear transient, right?

tidy3d/__init__.py
damiano-flex: It just requires another class which is a copy of `SteadyChargeDCSimulation` the same way for isothermal analysis.

Let me add it.

tidy3d/components/frequencies.py
damiano-flex: Yes. I created a new function to create an exact counterpart for spice. In spice, each modality uses  `N` in different ways (points per decade, octave variation, etc.), so I concluded that the best porting of that would be to have separate functions.

damiano-flex: We could add logaritmic option, but that would not be related to this PR. It is typical for SSAC analysis to use points per decade (which isn't the same of just running logarithmic).

tidy3d/components/tcad/simulation/heat_charge.py
damiano-flex: - For the first error, you can still call `_get_ssac_frequency_and_amplitude()` on a IsothermalDCAnalysis, so you can still hit that error.
- For the second error, you're right.

tidy3d/components/spice/sources/ac.py
```diff
@@ -0,0 +1,112 @@
...
+        title="DC Bias Voltages",
+        description="List of DC operating point voltages (above ground) used with :class:`VoltageBC`.",
+        units=VOLT,
+    )
+
+    signal: ACSignalType = pd.Field(
+        default=SinusoidalSignal(),
+        title="Small-signal properties",
+        description="AC signal specification for small-signal analysis.",
+    )
```
marc-flex, line 55: This doesn't feel very generic because this is restricting this AC source to a single signal. This is fine for SS but may be troublesome for transient simulations. 

I think we should have instead a tuple of signals. In that case, `voltages` would come from the time-average of each signal.

marc-flex (commented):
I guess the only missing thing is SSAC for non isothermal. 

My other concern is the specification of the ACVoltage signal definition. Maybe we should discuss how we go about it with the team

tidy3d/components/spice/sources/ac.py
damiano-flex: It is not. I can remove SS to be more generic.

tidy3d/components/spice/sources/ac.py
damiano-flex: Considering that `ACSignalType` is yet not developed, we could create classes that include different signal components inside to convoy the idea that the signal is 1 composed by many components.

However, since it is not completely clear to me what transient AC simulations with multiple signals will look like, I will defer this to your judgment.

@momchil-flex what do you think?

marc-flex: Good point. In transient simulation it may not make sense to have an array of signals. So maybe we could accept an array of signals and for transient we could check that only on signal is provided per ACVoltage

tidy3d/components/frequencies.py
momchil-flex: Okay yeah sounds good.

tidy3d/components/spice/analysis/ac.py
damiano-flex: Yeah, the class would be an exact copy. I already have it in local now. But since the inheritance graph would look like this.

`SteadyChargeDCAnalysis` -----> `SSACAnalysis`
|
|
|
V
`IsothermalSteadyChargeDCAnalysis` -> `IsothermalSSACAnalysis`

I couldn't think of a better way.

tidy3d/components/spice/sources/ac.py
momchil-flex: I am not sure, presumably if we provide a few common signal types, it should cover almost all ground. We could also provide a CustomSignal with arbitrary time dependence for the ultimate freedom? Might be simpler than allowing for multiple sources @marc-flex ?

tidy3d/components/spice/analysis/ac.py
momchil-flex: Now that I think about it again, did we kind of discuss the option of just `SteadyChargeAnalysis` and `IsothermalSteadyChargeAnalysis` with `ssac_freqs` as an optional field - rather than introducing SSAC subtypes? And then deprecating the ones with `DC` in the name.

tests/test_components/test_frequencies.py
damiano-flex: With the test you supplied, if `fmax `and `fmin` were translated (let's say multiplied both by 10), that test would still pass. So better to have a test that checks the actual bounds of `fmin` and `fmax`. 

Regarding the logarithmic spacing you are right, probably I accidentally removed it at some point. Let me add it back.

tidy3d/components/spice/analysis/ac.py
damiano-flex: Yes, I added it back because your V2-alternative version includes an ad-hoc analysis class. The unification approach was proposed before your V2 alternative in the lark document.

damiano-flex:
@greptile check this PR again.

tidy3d/components/spice/analysis/ac.py
momchil-flex: So I am not sure the V2 (alternative) still necessitates it. Just thinking out loud here: is the main difference going to be whether we check things like `if isinstance()` with `if analysis_spec.ssac_freqs is not None`?

Anyway, I guess let's just go with the explicit separation. Note that you could introduce an `AbstractSSACAnalysis` class that adds the `ssac_freqs` field and then just `SSACAnalysis(SteadyChargeDCAnalysis, AbstractSSACAnalysis)`,  IsothermalSSACAnalysis(IsothermalSteadyChargeDCAnalysis, AbstractSSACAnalysis)`.

tidy3d/components/spice/sources/ac.py
marc-flex: A `CustomSignal` can definitely be useful in full unsteady non-linear transient simulations. What "feels" unnatural is to define `voltages` outside of the signal which is equivalent to separate time-average and oscillating part (assuming is time-periodic). 

Having said that, maybe it is OK to have `voltage` as an optional parameter for the Steady and SSAC cases? It is definitely convenient

tidy3d/components/spice/analysis/ac.py
damiano-flex: 1) Yes. An ad hoc class only allows you an explicit specification that you're gong to do SSAC. But given the choices we made, we have to keep it. If we remove it, the only way for `SteadyChargeDCAnalysis` to understand the user wants to do SSAC is because it finds a `ACVoltageSource`. But if we go to transient, we can keep the same class and just change analysis type.

2) Fair enough. Let's do like that.

marc-flex: I think the reason why we decided to go for different analysis types was so that we were more explicit and avoid more if statements. It also allows us to check for the stuff we need for the simulation in a more concise way. This is, if we have a SSAC type analysis we know we need to look for some sources, we need frequencies, etc; if we have a non isothermal type of simulation, we look for thermal BCs, etc etc

damiano-flex:
@greptile 

The PR was modified completely due to a change of API structure. So, rewrite the PR body with the correct information referring to the updated code.

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -459,6 +469,9 @@ def check_voltage_array_if_capacitance(cls, values):
                     if isinstance(bc.condition.source, DCVoltageSource):
                         if len(bc.condition.source.voltage) > 1:
                             voltage_array_present = True
+                    elif isinstance(bc.condition.source, SSACVoltageSource):
+                        if len(bc.condition.source.voltage) > 1:
+                            voltage_array_present = True
```
marc-flex, line 474: can we maybe unify these two?
```python
if isinstance(bc.condition.source, (DCVoltageSource, SSACVoltageSource)):
    if len(bc.condition.source.voltage) > 1:
           voltage_array_present = True
```

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -468,6 +481,21 @@ def check_voltage_array_if_capacitance(cls, values):
...
+        for bc in boundary_spec:
+            if isinstance(bc.condition, VoltageBC):
+                if isinstance(bc.condition.source, SSACVoltageSource):
+                    if ssac_present:
+                        raise SetupError(
+                            "Only a single 'SSACVoltageSource' source can be supplied."
+                        )
+                    else:
+                        ssac_present = True
+        return values
```
marc-flex, line 497: Wondering whether we should always run this test or only in the case of SS analysis types. I mean, should this raise error if, for some reason, someone adds two SSAC sources but they actually want to run a DC simulation? 

@momchil-flex what's the usual behavior in FDTD?

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -627,6 +655,47 @@ def check_only_one_voltage_array_provided(cls, val, values):
...
+
+        return val
+
+    @pd.root_validator(skip_on_failure=True)
+    def check_ssac_freqs_requires_ac_source(cls, values):
+        """Ensure that if ssac_freqs is provided, at least one ACVoltageSource is present."""
+        analysis_spec = values.get("analysis_spec")
+        if (
+            isinstance(analysis_spec, (SSACAnalysis, IsothermalSSACAnalysis))
+            and len(analysis_spec.ssac_freqs) > 0
```
marc-flex: This `len(analysis_spec.ssac_freqs) == 0` can never happen, right?

tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -627,6 +655,47 @@ def check_only_one_voltage_array_provided(cls, val, values):
...
+                if isinstance(bc_spec.condition.source, GroundVoltageSource):
+                    ground_count = ground_count + 1
+
+        if ground_count > 1:
+            raise SetupError(
+                f"Only one GroundVoltageSource can be used. Found {ground_count} "
+                "'GroundVoltageSource' instances."
+            )
+
+        return val
```
marc-flex: Why can only one be the ground? In principle you could have different parts of the device connected to ground?

marc-flex (approved):
A few minor comments that can be ignored. 

Only one question that I don't quite understand. Why only one ground? Couldn't you have several parts of the device connected to ground?

tidy3d/components/tcad/simulation/heat_charge.py
momchil-flex: We don't really have an "analysis" notion in FDTD, so no exact comparison. That said I would say it's probably a good idea to run this always, and I would actually go one step further: should we make sure that there are no SSAC sources if the analysis spec is not SSAC? I actually kind of assumed we do, I haven't checked the current code though.

On one hand, this would prevent someone setting an SSAC source but forgetting to set the analysis type, and then not realizing that the source has no effect.

On the other hand, it's a bit annoying for situations where you just want to change the analysis spec for some reason, and not have to add/remove sources too. But honestly this is a bit hard to imagine right now since SSAC includes DC.

So my impression is that we should just validate it out, what do you guys think?

docs/api/spice.rst
```diff
@@ -10,16 +10,21 @@ Sources
    :toctree: _autosummary/
    :template: module.rst
 
+   tidy3d.SSACVoltageSource
    tidy3d.DCVoltageSource
+   tidy3d.GroundVoltageSource
```
daquinteroflex: Because ground is the circuit reference in which voltages are defined it does not make sense to name it a voltage source, maybe it should just be `GroundNode` or `GroundReference`

tidy3d/components/spice/analysis/ac.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Examples
+    --------
+    >>> import tidy3d as td
+    >>> freq_range = td.FreqRange.from_freq_interval(start_freq=1e3, stop_freq=1e6)
+    >>> sweep_freqs = freq_range.sweep_decade(num_points_per_decade=10)
+    >>> ssac_spec = td.SSACAnalysis(ssac_freqs=sweep_freqs)
+    """
+
+    ssac_freqs: ArrayFloat1D = pd.Field(
```
daquinteroflex: Why ssac_ if it is already in the class name?

daquinteroflex:
So I've had a brief chat with Damiano about circling back to the specific SSAC API.

Again from what I understand the main issue of not having Signal and a generic AC source is simply the issue that we have a voltage array on DC sources (which is a API design problem manifesting here) since the parameter sweeps make sense to be self contained to each Analysis and it just reference a source name for example.

That said, because I was on holiday, you guys have already discussed this and I don't want to hold back this PR any further. So if you want to merge this assuming it will be deprecated eventually when remove parameter sweeps from source definitions to have more AC source generality that's ok. 

I don't agree this is a good approach since it's something we should avoid but if we're rushed to get this in then that's how it goes. We should avoid keeping introducing classes we know are going to be deprecated, and have to further support and just do it properly once. I understand the argument that the charge code-base is small enough to do this for now so it's just a matter of paying that price for the sake of introducing this functionality sooner.

Really in the future we should have discussed the API via a PRD and a TDD rather than jumping in to implement it and keeping discussing on circling back etc. and agree the user facing functionality early on.

 We're breaking compatibility in 2.10 in RF already and we could do it for these specific things in Charge if we really want to fix them, or transform it internally to a new API structure (maybe part of a separate conversation).
 
 It probably would be great to get @tylerflex 's thoughts on this.

momchil-flex:
@daquinteroflex if I understand things correctly, what we agreed on wouldn't require breaking compatibility. If we get to extend things to full transient anaysis, we will just keep separating the small-signal sources from the more general AC sources. The point is that the SSAC source is special and it's very constrained in the type of signal it supports. You can then think of it as a convenience component - it seemed like adding a more general component that can accommodate both SS and full transient in the future just adds verbosity to the SSAC setup, which will be the only thing we support for a while anyway.

Not sure I fully understand this though. Can you elaborate?

> So if you want to merge this assuming it will be deprecated eventually when remove parameter sweeps from source definitions to have more AC source generality that's ok.

marc-flex:
@daquinteroflex I don't think we'll deprecate the SSAC signal BC. It's very specific so I think it makes sense to keep it.

daquinteroflex:
So just for the record I think the solution as you describe will "work inline to what we have been doing before". So it's not different from our previous charge API approaches, and should provide the desired functionality for now. I just think we're keeping on building on previous implicit behaviour issues that may be confusing (already maybe they are) as the functionality increases, and we're not taking this as an opportunity to address known issues of the electrical setup and workflow user clarity that can become more confusing. It is fine by me if we want to merge it for the sake of getting this in, but I describe below why this could introduce technical debut and further lack of workflow clarity for the users even if it introduces more verbosity in the backend.

This is what I was proposing mainly through having a more explicit Analysis separation and configuration, and introduce nodal treatment to help those analysis specify what should run. I illustrate what I mean below on the line of your request Momchil.

> So if you want to merge this assuming it will be deprecated eventually when remove parameter sweeps from source definitions to have more AC source generality that's ok.

Currently we have parameter sweeps in the source definitions, which introduces multivariable input complexity to manage also depending on the type of analysis performed. For example, `DCVoltageSource(voltages=[....])` will have different treatments that might be unclear to the user according to the different analysis performed. Say someone uses this `SSACVoltageSource(voltages=[.....], freqs=[.....])` , then they're actually performing an "SSAC analysis" for each DC voltage point right? From a user perspective, we're adding ambiguity what exactly is running as a "source" is not and should not be a parameter sweep in my opinion in this context since it can confuse what exactly it does or occurs in the backend. Introducing the verbosity is not necessarily a bad thing, it can clarify to users exactly what they are defining even if it is harder for us to deal with it possibly?

We can make this more explicit by simply specifying source names, like nodes, and then have parameter sweeps only be defined at the analysis level. So the case above: 

```
SSACVoltageSource(voltages=[.....], freqs=[.....]) ----->  becomes something like (in the current architecture

[
HeatChargeSimulation(..., VoltageBC(DCVoltageSource(voltage=1, name="V0")), SSAnalysis(freqs=[...], name="V0")), 
HeatChargeSimulation(....),
HeatChargeSimulation(....)
]
```

Like this approach means it is explicit they know an SSAC is being run as a "single heat-charge simulation" configuration independent of the other DC voltage sweeps they are doing so it explains what's actually going on. We could do a similar thing for DCAnalysis where the parameter sweep is specified at the analysis class because that way it's clearer what's going on and how their configuration changes their final results. If it is not the case, then we probably want to make this very explicit. When we did the first charge API, management of what is actually run under the hood in the backend was not very explicit so trying to avoid the combinations of settings get unbounded and users don't know what's going on as the simulations become more complex.

> it seemed like adding a more general component that can accommodate both SS and full transient in the future just adds verbosity to the SSAC setup

So the approach above simply separates "analysis configuration" to the analysis class, rather than mixing it in the source class possibly. 

> You can then think of it as a convenience component - it seemed like adding a more general component that can accommodate both SS and full transient in the future just adds verbosity to the SSAC setup, which will be the only thing we support for a while anyway.

> I don't think we'll deprecate the SSAC signal BC. It's very specific so I think it makes sense to keep it.

So this all kind of depends if we want to keep embedding SSAC analysis-specific stuff into the simulation/scene definition rather than simply treat the analysis to be defined based on the "physical/electrical node scene" of the simulation. 

This will work, so if we just want to keep doing this it's ok. I just feel we're keeping on building implicit treatment rather than explict treatment of how the simulation is being processes. Say how this is going to work if people want to eventually reuse "physical scenes" or even as they begin adding sources that are incompatible with a given analysis just feels we might introduce a bunch of edge cases and not catch the users intention explicitly possibly. Again, it's what we've been doing already and "it works", but feel one day we need to deal with these issues of explicit behavior.

momchil-flex:
Note that this is *not* what this PR is introducing: `SSACVoltageSource(voltages=[.....], freqs=[.....])`

There are no `freqs` in the source itself. Instead the ssac frequencies go in the analysis spec. This is because eventually you could have multiple SSAC sources, and the analysis should to be done at the same frequency for all of them to get the admittance matrix at that frequency.

daquinteroflex:
Ahh, sorry, the API has been changing a lot and the diffs got me mixed up

> Note that this is not what this PR is introducing: SSACVoltageSource(voltages=[.....], freqs=[.....])
> There are no freqs in the source itself. Instead the ssac frequencies go in the analysis spec. This is because eventually you could have multiple SSAC sources, and the analysis should to be done at the same frequency for all of them to get the admittance matrix at that frequency.

Yeah this works well

PR #2813, opened 2025-09-11
Warn if different media have the same name

momchil-flex:
This was based on a discussion that GUI uses names for media more heavily than python does, and non-unique names are a bit problematic.

## Greptile Summary

Updated On: 2025-09-11 15:29:09 UTC

This PR adds validation logic to warn users when different media (mediums) in a simulation have the same name. The change addresses a specific integration issue where GUI tools rely more heavily on medium names than the Python API does, potentially causing confusion when different media share identical names.

The implementation modifies the `_validate_num_mediums` method in `tidy3d/components/scene.py`, renaming it to `_validate_mediums` and expanding its functionality. The new validation preserves the existing check for the maximum number of mediums while adding a name uniqueness check. When duplicate names are detected among different media, the system emits a warning message recommending unique names for distinct media.

The validation is implemented as a warning rather than an error, which is appropriate since medium names are optional in the Python API. This preserves backward compatibility while providing helpful guidance to users, especially those who work with both the Python API and GUI tools. The change runs whenever structures are set or validated through the `@pd.validator` decorator, ensuring timely feedback.

A corresponding test was added in `tests/test_components/test_simulation.py` to verify the warning behavior. The test creates a simulation with two structures containing different media (permittivity=2 and permittivity=3) that both use the name 'medium1', and validates that the expected warning is generated.

## Important Files Changed

| tidy3d/components/scene.py | 5/5 | Renamed and expanded `_validate_num_mediums` to `_validate_mediums`, adding duplicate medium name detection with warning |
| tests/test_components/test_simulation.py | 5/5 | Added `test_unique_medium_names()` to verify warning behavior for duplicate medium names |

</details>

## Confidence score: 5/5

- This PR is safe to merge with minimal risk
- Score reflects simple, well-tested validation logic that only adds warnings without breaking existing functionality

### DIFF
--- a/tidy3d/components/scene.py
+++ b/tidy3d/components/scene.py
@@ -138,8 +138,8 @@ class Scene(Tidy3dBaseModel):
     _unique_structure_names = assert_unique_names("structures")
 
     @pd.validator("structures", always=True)
-    def _validate_num_mediums(cls, val):
-        """Error if too many mediums present."""
+    def _validate_mediums(cls, val):
+        """Error if too many mediums present. Warn if different mediums have the same name."""
 
         if val is None:
             return val
@@ -149,6 +149,14 @@ class Scene(Tidy3dBaseModel):
             raise SetupError(
                 f"Tidy3D only supports {MAX_NUM_MEDIUMS} distinct mediums."
                 f"{len(mediums)} were supplied."
+            )
+
+        medium_names = [medium.name for medium in mediums if medium.name is not None]
+        if len(medium_names) != len(set(medium_names)):
+            log.warning(
+                "Different mediums with the same name were detected. "
+                "This may error in future Tidy3D versions, and using unique names for distinct "
+                "media is recommended."
             )
 
         return val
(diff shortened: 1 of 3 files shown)


tidy3d/components/scene.py
```diff
@@ -151,6 +151,13 @@ def _validate_num_mediums(cls, val):
                 f"{len(mediums)} were supplied."
             )
 
+        medium_names = [medium.name for medium in mediums if medium.name is not None]
+        if len(medium_names) != len(set(medium_names)):
+            log.warning(
+                "Different mediums with the same name were detected. "
+                "Using unique names for distinct media is recommended."
```
yaugenst-flex: Maybe this should mention something about GUI compatibility? Although I'm not sure what happens right now in GUI if you do this?

yaugenst-flex (approved):
Thanks @momchil-flex for breezing through these 😅
Does this one warrant a changelog?

e-g-melo (approved):
Thanks @momchil-flex!

It sounds good! I have created a ticket in GUI to make it consistent there.

tidy3d/components/scene.py
momchil-flex: I just added a note that this may error in the future, effectively a deprecation warning.

PR #2822, opened 2025-09-16
Make task names optional

### LINKED ISSUE #2819: Make task names optional
Since task names are not required to be unique, it would make sense to drop the requirement to supply them when running simulations.

Change
 - `task_name` becomes optional in `run` (and maybe `run_async`?) as well as related functions.
 - Automatically generate unique(-ish) name when absent, ideally with some useful human-readable identifiers.

Before:
```py
sim_data = run(simulation=sim, task_name="adjoint_quickstart")
```

After:
```py
sim_data = run(sim)
results = run_async([sim1, sim2, sim3])  # maybe?
```

Back-compat
 - Explicit `task_name` still supported.

marcorudolphflex:
Relates to https://github.com/flexcompute/tidy3d/issues/2819

**Change**

task_name becomes optional in `run` and `run_async` as well as in `Job` and `Batch`.
In  `run_async` and `Batch`, a list of simulations instead of a dictionary can be provided.
Task name is generated with simulation type and timestamp like `fdtd_2025-09-16_14-30-55`, trailed with `"_{index}"` for batch jobs.

**Usage Example:**

```
sim_data = run(sim)
results = run_async([sim1, sim2, sim3])
```
(similar with `Job` or `Batch`)

Explicit task_names are still supported.

## Greptile Summary

Updated On: 2025-09-16 11:20:45 UTC

This PR implements a significant user experience improvement by making task names optional across the Tidy3D web API. The changes affect multiple core functions (`run`, `run_async`, `Job`, and `Batch`) to allow users to submit simulations without explicitly providing task names.

The implementation adds automatic task name generation using the format `{simulation_type}_{timestamp}` (e.g., `fdtd_2025-09-16_14-30-55`). For batch operations, an index suffix is appended (`_{index}`) to ensure uniqueness. The core functionality is implemented through a new `get_default_task_name()` method in the `Tidy3dStub` class that combines the simulation type (lowercased) with a human-readable timestamp.

Key API changes include:
- `run(simulation, task_name=None)` - task_name parameter becomes optional
- `run_async(simulations)` - now accepts either a dictionary (existing) or list of simulations (new)
- `Job` class - task_name field becomes optional
- `Batch` class - simulations field accepts either dictionary or list format

The changes maintain full backward compatibility - existing code using explicit task names continues to work unchanged. The implementation correctly handles the conversion from list format to the internal dictionary format required by the underlying job management system.

## Important Files Changed

| `tidy3d/web/api/tidy3d_stub.py` | 4/5 | Adds `get_default_task_name()` method that generates task names from simulation type and timestamp |
| `tidy3d/web/api/container.py` | 4/5 | Makes task_name optional in Job class and enables list input for Batch simulations with auto-naming |
| `tidy3d/web/api/webapi.py` | 4/5 | Makes task_name parameter optional in `run()` and `upload()` functions with default name generation |
| `tidy3d/web/api/asynchronous.py` | 4/5 | Enables `run_async()` to accept list of simulations instead of requiring dictionary format |
| `tidy3d/web/api/autograd/autograd.py` | 4/5 | Updates autograd-specific `run()` and `run_async()` functions to support optional task names |
| `tests/test_web/test_webapi.py` | 4/5 | Adds parametrized tests covering both explicit and auto-generated task name scenarios |
| `tests/test_web/test_tidy3d_stub.py` | 4/5 | Adds test for default task name generation functionality |
| `tests/test_components/autograd/test_autograd.py` | 5/5 | Extends autograd async test to cover both dictionary and list simulation formats |
| `CHANGELOG.md` | 5/5 | Documents the new optional task names feature with clear user-facing description |

</details>

## Confidence score: 4/5

- This PR is generally safe to merge with good backward compatibility, but has some potential edge cases around timing
- Score reflects solid implementation and comprehensive testing, but concerns about timestamp collision edge cases and missing validation
- Pay close attention to the task name generation logic in `tidy3d_stub.py` and batch conversion logic in `container.py`

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
    participant run as web.run()
    participant Tidy3dStub
    participant upload as web.upload()
    participant Task as SimulationTask
    participant start as web.start()
    participant monitor as web.monitor()
    participant load as web.load()

    User->>run: run(simulation, task_name=None)
    
    alt task_name is None
        run->>Tidy3dStub: Tidy3dStub(simulation=simulation)
        Tidy3dStub->>Tidy3dStub: get_default_task_name()
        Note over Tidy3dStub: Generates name like "fdtd_2025-09-16_14-30-55"
        Tidy3dStub-->>run: generated_task_name
        run->>run: task_name = generated_task_name
    end
    
    run->>upload: upload(simulation, task_name, ...)
    upload->>Tidy3dStub: Tidy3dStub(simulation)
    upload->>Task: create(task_type, task_name, ...)
    Task-->>upload: task_id
    upload-->>run: task_id
    
    run->>start: start(task_id, ...)
    start->>Task: submit(task_id, ...)
    Task-->>start: success
    start-->>run: success
    
    run->>monitor: monitor(task_id, ...)
    monitor->>Task: get_status(task_id)
    loop Until completion
        Task-->>monitor: status
        alt status not complete
            monitor->>monitor: wait
        end
    end
    monitor-->>run: completed
    
    run->>load: load(task_id, ...)
    load->>Task: download and load data
    Task-->>load: simulation_data
    load-->>run: simulation_data
    
    run-->>User: simulation_data
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - New `MediumMonitor` that returns both permittivity and permeability profiles.
+- Task names are now optional when using `run(sim)` or `Job`. When running multiple jobs (via `run_async` or `Batch`), you can also provide simulations as a list without specifying task names. The previous dictionary-based format with explicit task names is still supported.
+
 ### Changed
 
 ### Fixed
--- a/tidy3d/web/api/autograd/autograd.py
+++ b/tidy3d/web/api/autograd/autograd.py
@@ -28,6 +28,7 @@ from tidy3d.exceptions import AdjointError
 from tidy3d.web.api.asynchronous import DEFAULT_DATA_DIR
 from tidy3d.web.api.asynchronous import run_async as run_async_webapi
 from tidy3d.web.api.container import DEFAULT_DATA_PATH, Batch, BatchData, Job
+from tidy3d.web.api.tidy3d_stub import Tidy3dStub
 from tidy3d.web.api.webapi import run as run_webapi
 from tidy3d.web.core.s3utils import download_file, upload_file
 from tidy3d.web.core.types import PayType
@@ -96,7 +97,7 @@ def is_valid_for_autograd_async(simulations: dict[str, td.Simulation]) -> bool:
 
 def run(
     simulation: WorkflowType,
-    task_name: str,
+    task_name: typing.Optional[str] = None,
     folder_name: str = "default",
     path: str = "simulation_data.hdf5",
     callback_url: typing.Optional[str] = None,
@@ -121,8 +122,8 @@ def run(
     ----------
     simulation : Union[:class:`.Simulation`, :class:`.HeatSimulation`, :class:`.EMESimulation`, :class:`.ModalComponentModeler`, :class:`.TerminalComponentModeler`]
         Simulation to upload to server.
-    task_name : str
-        Name of task.
+    task_name : Optional[str] = None
+        Name of task. If not provided, a default name will be generated.
     folder_name : str = "default"
         Name of folder to store task on web UI.
     path : str = "simulation_data.hdf5"
@@ -200,6 +201,10 @@ def run(
     if priority is not None and (priority < 1 or priority > 10):
         raise ValueError("Priority must be between '1' and '10' if specified.")
 
+    if task_name is None:
+        stub = Tidy3dStub(simulation=simulation)
+        task_name = stub.get_default_task_name()
+
     # component modeler path: route autograd-valid modelers to local run
     from tidy3d.plugins.smatrix.component_modelers.types import ComponentModelerType
 
@@ -261,7 +266,7 @@ def run(
 
 
 def run_async(
-    simulations: dict[str, td.Simulation],
+    simulations: typing.Union[dict[str, td.Simulation], tuple[td.Simulation], list[td.Simulation]],
     folder_name: str = "default",
     path_dir: str = DEFAULT_DATA_DIR,
     callback_url: typing.Optional[str] = None,
@@ -283,8 +288,8 @@ def run_async(
 
     Parameters
     ----------
-    simulations : Dict[str, Union[:class:`.Simulation`, :class:`.HeatSimulation`, :class:`.EMESimulation`]]
-        Mapping of task name to simulation.
+    simulations : Union[Dict[str, Union[:class:`.Simulation`, :class:`.HeatSimulation`, :class:`.EMESimulation`]], tuple[Union[:class:`.Simulation`, :class:`.HeatSimulation`, :class:`.EMESimulation`]], list[Union[:class:`.Simulation`, :class:`.HeatSimulation`, :class:`.EMESimulation`]]]
+        Mapping of task name to simulation or list of simulations.
     folder_name : str = "default"
         Name of folder to store each task on web UI.
     path_dir : str
@@ -328,6 +333,13 @@ def run_async(
     # validate priority if specified
     if priority is not None and (priority < 1 or priority > 10):
         raise ValueError("Priority must be between '1' and '10' if specified.")
+
+    if isinstance(simulations, (tuple, list)):
+        sim_dict = {}
+        for i, sim in enumerate(simulations, 1):
+            task_name = Tidy3dStub(simulation=sim).get_default_task_name() + f"_{i}"
+            sim_dict[task_name] = sim
+        simulations = sim_dict
 
     if is_valid_for_autograd_async(simulations):
         return _run_async(
--- a/tidy3d/web/api/connect_util.py
+++ b/tidy3d/web/api/connect_util.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import time
 from functools import wraps
+from typing import Optional
 
 from requests import ReadTimeout
 from requests.exceptions import ConnectionError as ConnErr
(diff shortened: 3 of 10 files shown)


tidy3d/web/api/container.py
```diff
@@ -532,7 +537,9 @@ class Batch(WebContainer):
         * `Inverse taper edge coupler <../../notebooks/EdgeCoupler.html>`_
     """
 
-    simulations: dict[TaskName, annotate_type(WorkflowType)] = pd.Field(
+    simulations: Union[
+        dict[TaskName, annotate_type(WorkflowType)], list[annotate_type(WorkflowType)]
```
yaugenst-flex: Should be `tuple` instead of `list` due to mutability. Lists will be coerced to `tuple`.

tidy3d/web/api/container.py
```diff
@@ -663,12 +670,21 @@ def jobs(self) -> dict[TaskName, Job]:
         if self.jobs_cached is not None:
             return self.jobs_cached
 
+        if isinstance(self.simulations, list):
```
yaugenst-flex: See above, `tuple`

tidy3d/web/api/webapi.py
```diff
@@ -129,8 +129,8 @@ def run(
     ----------
     simulation : Union[:class:`.Simulation`, :class:`.HeatSimulation`, :class:`.EMESimulation`]
         Simulation to upload to server.
-    task_name : str
-        Name of task.
+    task_name : Optional[str] = None
+        Name of task. If not provided, default name will be generated.
```
yaugenst-flex: **a** default name

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+* Task names are now optional when using `run(sim)` or `Job`. When running multiple jobs (via `run_async` or `Batch`), you can also provide simulations as a **list** without specifying task names. The previous dictionary-based format with explicit task names is still supported.
```
yaugenst-flex: `-` instead of `*` for the bullet

Generally we avoid text styling in the changelog beyond code blocks (bold list here).

tidy3d/web/api/autograd/autograd.py
```diff
@@ -261,7 +266,7 @@ def run(
 
 
 def run_async(
-    simulations: dict[str, td.Simulation],
+    simulations: typing.Union[dict[str, td.Simulation], list[td.Simulation]],
```
yaugenst-flex: Here we need to be more careful with `tuple`/`list` (see my comment on `Job`). I suppose this should probably be a `Sequence`. And then runtime check on `isinstance(obj, (list, tuple))`. Could also just add both types to the union.

tidy3d/web/api/container.py
yaugenst-flex, line 1: There is one somewhat nasty problem in `Job`. Since `task_name` is now optional, but we generate one, effectively the task name will change after-the-fact and the `Job`'s task name will stay `None`, and we can't adapt it because the `Job` is immutable (and should be). I think the best way to address this would be via a (probably root) validator that assigns the task name if it's `None` during model construction.

yaugenst-flex (changes requested):
Thanks @marcorudolphflex this is looking good!

One more general observation: I was considering that it might be nice if we return a list if a list was supplied. In the future we could generalize that to more data structures. However, many things currently assume that the output is a mapping, and this would become a behavior change. So with that in mind I think we should stick with the "always return a mapping" approach for now.

tidy3d/web/api/tidy3d_stub.py
```diff
@@ -173,6 +174,27 @@ def validate_pre_upload(self, source_required) -> None:
...
+        'fdtd_2025-09-16_14-30-55'
+
+        Returns
+        -------
+        str
+            Default task name, e.g. ``"fdtd_2025-09-16_14-30-55"``.
+        """
+        sim_type = self.get_type().lower()
+        timestamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+        return f"{sim_type}_{timestamp}"
```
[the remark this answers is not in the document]
yaugenst-flex: this is fine here, the task names are not required to be unique

tidy3d/web/api/connect_util.py
```diff
@@ -12,13 +13,14 @@
...
 from tidy3d.log import log
-from tidy3d.web.common import CONNECTION_RETRY_TIME, REFRESH_TIME
+from tidy3d.web import common
+from tidy3d.web.common import REFRESH_TIME
 
 
-def wait_for_connection(decorated_fn=None, wait_time_sec: float = CONNECTION_RETRY_TIME):
-    """Causes function to ignore connection errors and retry for ``wait_time_sec`` secs."""
+def wait_for_connection(decorated_fn=None, wait_time_sec: Optional[float] = None):
+    """Causes function to ignore connection errors and retry for `wait_time_sec` seconds."""
```
yaugenst-flex: Double backticks are correct here, docstrings use rst format

tidy3d/web/api/tidy3d_stub.py
```diff
@@ -173,6 +174,27 @@ def validate_pre_upload(self, source_required) -> None:
...
+    def get_default_task_name(self) -> str:
+        """
+        Generate a default task name based on the simulation type and
+        the current date and time.
+
+        The name is composed of the simulation type and a human-readable timestamp in the format ``YYYY-MM-DD_HH-MM-SS``
+
+        Example
+        -------
+        >>> stub.get_default_task_name()
```
yaugenst-flex: We have doctests so this will fail CI. I think `# doctest: +SKIP` is fine here.

yaugenst-flex (commented):
Sorry forgot a couple of formatting comments.

yaugenst-flex (approved):
Looks great! Could you squash into a single commit and rebase please, then it'll be ready to go.

PR #2829, opened 2025-09-19
Deprecation messages for semiconductor

marc-flex:
## Greptile Summary

Updated On: 2025-09-19 15:01:06 UTC

This PR implements deprecation warnings for the `SemiconductorMedium` class in the TCAD/charge material module. The changes transition the API from accepting simple float values to requiring structured model objects for key semiconductor material properties.

Specifically, the PR modifies five parameters (`N_c`, `N_v`, `E_g`, `N_d`, `N_a`) to accept both the new model types and legacy float values through Union type hints. Five new validators are added that detect when float values are provided, issue deprecation warnings, and automatically convert them to the appropriate model objects:

- `N_c` (effective density of states in conduction band) → `ConstantEffectiveDOS`
- `N_v` (effective density of states in valence band) → `ConstantEffectiveDOS` 
- `E_g` (energy band gap) → `ConstantEnergyBandGap`
- `N_d` (donor doping concentration) → `[ConstantDoping]`
- `N_a` (acceptor doping concentration) → `[ConstantDoping]`

This change fits into the broader Tidy3D framework's evolution toward more structured and extensible material definitions. The deprecation approach maintains backward compatibility while encouraging migration to the new API pattern. The automatic conversion ensures existing user code continues to function during the transition period, following established software engineering practices for API evolution.

## Important Files Changed

<details>

| tidy3d/components/material/tcad/charge.py | 4/5 | Added deprecation warnings and validators for semiconductor material parameters transitioning from float to model objects |

</details>

## Confidence score: 4/5

- This PR is safe to merge with minimal risk as it maintains full backward compatibility while implementing a clean deprecation strategy
- Score reflects well-structured deprecation handling with automatic conversion that preserves existing functionality
- Pay close attention to the validator implementations to ensure proper warning messages and type conversions are working as expected

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
    participant SemiconductorMedium
    participant Validator
    participant Logger
    participant ModelClass

    User->>SemiconductorMedium: "Create with N_c=float"
    SemiconductorMedium->>Validator: "check_nc_uses_model()"
    Validator->>Validator: "isinstance(val, EffectiveDOSModelType)?"
    alt Float provided
        Validator->>Logger: "log.warning(deprecation message)"
        Validator->>ModelClass: "ConstantEffectiveDOS(N=val)"
        ModelClass-->>Validator: "model instance"
        Validator-->>SemiconductorMedium: "model instance"
    else Model provided
        Validator-->>SemiconductorMedium: "val"
    end

    User->>SemiconductorMedium: "Create with N_v=float"
    SemiconductorMedium->>Validator: "check_nv_uses_model()"
    Validator->>Validator: "isinstance(val, EffectiveDOSModelType)?"
    alt Float provided
        Validator->>Logger: "log.warning(deprecation message)"
        Validator->>ModelClass: "ConstantEffectiveDOS(N=val)"
        ModelClass-->>Validator: "model instance"
        Validator-->>SemiconductorMedium: "model instance"
    else Model provided
        Validator-->>SemiconductorMedium: "val"
    end

    User->>SemiconductorMedium: "Create with E_g=float"
    SemiconductorMedium->>Validator: "check_eg_uses_model()"
    Validator->>Validator: "isinstance(val, EnergyBandGapModelType)?"
    alt Float provided
        Validator->>Logger: "log.warning(deprecation message)"
        Validator->>ModelClass: "ConstantEnergyBandGap(eg=val)"
        ModelClass-->>Validator: "model instance"
        Validator-->>SemiconductorMedium: "model instance"
    else Model provided
        Validator-->>SemiconductorMedium: "val"
    end

    User->>SemiconductorMedium: "Create with N_d=float"
    SemiconductorMedium->>Validator: "check_nd_uses_model()"
    Validator->>Validator: "isinstance(val, (float, int))?"
    alt Float/int provided
        Validator->>Logger: "log.warning(deprecation message)"
        Validator->>ModelClass: "ConstantDoping(concentration=val)"
        ModelClass-->>Validator: "doping instance"
        Validator-->>SemiconductorMedium: "[doping instance]"
    else Proper type provided
        Validator-->>SemiconductorMedium: "val"
    end

    User->>SemiconductorMedium: "Create with N_a=float"
    SemiconductorMedium->>Validator: "check_na_uses_model()"
    Validator->>Validator: "isinstance(val, (float, int))?"
    alt Float/int provided
        Validator->>Logger: "log.warning(deprecation message)"
        Validator->>ModelClass: "ConstantDoping(concentration=val)"
        ModelClass-->>Validator: "doping instance"
        Validator-->>SemiconductorMedium: "[doping instance]"
    else Proper type provided
        Validator-->>SemiconductorMedium: "val"
    end

    SemiconductorMedium-->>User: "SemiconductorMedium instance"
```

### DIFF
--- a/tidy3d/components/material/tcad/charge.py
+++ b/tidy3d/components/material/tcad/charge.py
@@ -8,15 +8,18 @@ import pydantic.v1 as pd
 
 from tidy3d.components.data.data_array import SpatialDataArray
 from tidy3d.components.medium import AbstractMedium
-from tidy3d.components.tcad.doping import DopingBoxType
+from tidy3d.components.tcad.doping import ConstantDoping, DopingBoxType
 from tidy3d.components.tcad.types import (
     BandGapNarrowingModelType,
+    ConstantEffectiveDOS,
+    ConstantEnergyBandGap,
     EffectiveDOSModelType,
     EnergyBandGapModelType,
     MobilityModelType,
     RecombinationModelType,
 )
 from tidy3d.constants import CONDUCTIVITY, ELECTRON_VOLT, PERCMCUBE, PERMITTIVITY
+from tidy3d.log import log
 
 
 class AbstractChargeMedium(AbstractMedium):
@@ -228,8 +231,8 @@ class SemiconductorMedium(AbstractChargeMedium):
         ...         c2=0.5,
         ...         min_N=1e15,
         ...     ),
-        ...     N_a=0,
-        ...     N_d=0
+        ...     N_a=[td.ConstantDoping(concentration=1e15)],
+        ...     N_d=[td.ConstantDoping(concentration=1e15)]
         ... )
 
 
@@ -253,21 +256,21 @@ class SemiconductorMedium(AbstractChargeMedium):
 
     """
 
-    N_c: EffectiveDOSModelType = pd.Field(
+    N_c: Union[EffectiveDOSModelType, pd.PositiveFloat] = pd.Field(
         ...,
         title="Effective density of electron states",
         description=":math:`N_c` Effective density of states in the conduction band.",
         units=PERCMCUBE,
     )
 
-    N_v: EffectiveDOSModelType = pd.Field(
+    N_v: Union[EffectiveDOSModelType, pd.PositiveFloat] = pd.Field(
         ...,
         title="Effective density of hole states",
         description=":math:`N_v` Effective density of states in the valence band.",
         units=PERCMCUBE,
     )
 
-    E_g: EnergyBandGapModelType = pd.Field(
+    E_g: Union[EnergyBandGapModelType, pd.PositiveFloat] = pd.Field(
         ...,
         title="Band-gap energy",
         description=":math:`E_g` Band-gap energy",
@@ -300,7 +303,7 @@ class SemiconductorMedium(AbstractChargeMedium):
     )
 
     N_a: Union[pd.NonNegativeFloat, SpatialDataArray, tuple[DopingBoxType, ...]] = pd.Field(
-        0,
+        (),
         title="Doping: Acceptor concentration",
         description="Concentration of acceptor impurities, which create mobile holes, resulting in p-type material. "
         "Can be specified as a single float for uniform doping, a :class:`SpatialDataArray` for a custom profile, "
@@ -309,10 +312,66 @@ class SemiconductorMedium(AbstractChargeMedium):
     )
 
     N_d: Union[pd.NonNegativeFloat, SpatialDataArray, tuple[DopingBoxType, ...]] = pd.Field(
-        0,
+        (),
         title="Doping: Donor concentration",
         description="Concentration of donor impurities, which create mobile electrons, resulting in n-type material. "
         "Can be specified as a single float for uniform doping, a :class:`SpatialDataArray` for a custom profile, "
         "or a tuple of geometric shapes to define specific doped regions.",
         units=PERCMCUBE,
     )
+
+    # DEPRECATION VALIDATORS
+    @pd.validator("N_c", always=True)
+    def check_nc_uses_model(cls, val, values):
+        """Issue deprecation warning if float is provided"""
+        if isinstance(val, (float, int)):
+            log.warning(
+                "Passing a float to 'N_c' is deprecated and will be removed in future versions. "
+                "Please use 'ConstantEffectiveDOS' instead."
+            )
+            return ConstantEffectiveDOS(N=val)
+        return val
+
+    @pd.validator("N_v", always=True)
+    def check_nv_uses_model(cls, val, values):
+        """Issue deprecation warning if float is provided"""
+        if isinstance(val, (float, int)):
+            log.warning(
+                "Passing a float to 'N_v' is deprecated and will be removed in future versions. "
+                "Please use 'ConstantEffectiveDOS' instead."
+            )
+            return ConstantEffectiveDOS(N=val)
+        return val
+
+    @pd.validator("E_g", always=True)
+    def check_eg_uses_model(cls, val, values):
+        """Issue deprecation warning if float is provided"""
+        if isinstance(val, (float, int)):
+            log.warning(
+                "Passing a float to 'E_g' is deprecated and will be removed in future versions. "
+                "Please use 'ConstantEnergyBandGap' instead."
+            )
+            return ConstantEnergyBandGap(eg=val)
+        return val
+
+    @pd.validator("N_d", always=True)
+    def check_nd_uses_model(cls, val, values):
+        """Issue deprecation warning if float is provided"""
+        if isinstance(val, (float, int)):
+            log.warning(
+                "Passing a float to 'N_d' is deprecated and will be removed in future versions. "
+                f"Please use a list of 'DopingBoxType' instead, e.g., [ConstantDoping(concentration={val})]."
+            )
+            return (ConstantDoping(concentration=val),)
+        return val
+
+    @pd.validator("N_a", always=True)
+    def check_na_uses_model(cls, val, values):
+        """Issue deprecation warning if float is provided"""
+        if isinstance(val, (float, int)):
+            log.warning(
+                "Passing a float to 'N_a' is deprecated and will be removed in future versions. "
+                f"Please use a list of 'DopingBoxType' instead, e.g., [ConstantDoping(concentration={val})]."
+            )
+            return (ConstantDoping(concentration=val),)
+        return val
--- a/tidy3d/components/scene.py
+++ b/tidy3d/components/scene.py
@@ -20,7 +20,12 @@ from tidy3d.components.material.tcad.charge import (
 )
 from tidy3d.components.material.tcad.heat import SolidMedium, SolidSpec
 from tidy3d.components.material.types import MultiPhysicsMediumType3D, StructureMediumType
-from tidy3d.components.tcad.doping import ConstantDoping, GaussianDoping
+from tidy3d.components.tcad.doping import (
+    ConstantDoping,
+    CustomDoping,
+    DopingBoxType,
+    GaussianDoping,
+)
 from tidy3d.components.tcad.viz import HEAT_SOURCE_CMAP
 from tidy3d.constants import CONDUCTIVITY, THERMAL_CONDUCTIVITY, inf
 from tidy3d.exceptions import SetupError, Tidy3dError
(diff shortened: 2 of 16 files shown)


tidy3d/components/material/tcad/charge.py
```diff
@@ -309,10 +312,66 @@ class SemiconductorMedium(AbstractChargeMedium):
     )
 
     N_d: Union[pd.NonNegativeFloat, SpatialDataArray, tuple[DopingBoxType, ...]] = pd.Field(
-        0,
+        (ConstantDoping(concentration=0, center=(0, 0, 0), size=(inf, inf, inf)),),
```
momchil-flex: How about we set this center and size as default in ConstantDoping?

tidy3d/components/material/tcad/charge.py
```diff
@@ -309,10 +312,66 @@ class SemiconductorMedium(AbstractChargeMedium):
...
+            return ConstantEnergyBandGap(eg=val)
+        return val
+
+    @pd.validator("N_d", always=True)
+    def check_nd_uses_model(cls, val, values):
+        """Issue deprecation warning if float is provided"""
+        if isinstance(val, (float, int)):
+            log.warning(
+                "Passing a float to 'N_d' is deprecated and will be removed in future versions. "
+                "Please use 'SpatialDataArray' or a list of 'DopingBoxType' instead."
```
momchil-flex: I feel like here we should just say use `ConstantDoping` instead because if the user is providing a float, that's the equivalent way. Or at least something like "use the equivalent  'ConstantDoping', or other 'DopingBoxType' classes or a 'SpatialDataArray' for more advanced doping definitions."

tidy3d/components/material/tcad/charge.py
```diff
@@ -300,7 +303,7 @@ class SemiconductorMedium(AbstractChargeMedium):
     )
 
     N_a: Union[pd.NonNegativeFloat, SpatialDataArray, tuple[DopingBoxType, ...]] = pd.Field(
```
momchil-flex, line 305: At some point we also discussed moving the `SpatialDataArray` to something like a `CustomDoping` doping box type. Do you foresee any issues or is it just a matter of implementing it? One usage difference is that then it will be additive with other doping boxes.

tidy3d/components/tcad/doping.py
```diff
@@ -20,12 +20,14 @@ def _normal_dim(self):
         """Returns the normal direction if the box is 2D. False otherwise"""
```
momchil-flex: It seems to returns *some* normal_dim when the box is 2D or 3D, and it raises otherwise.

tidy3d/components/tcad/doping.py
```diff
@@ -20,12 +20,14 @@ def _normal_dim(self):
...
+        inf_dims = 0
         for dim in range(3):
             if self.size[dim] == np.inf:
-                if normal_dim is not None:
-                    raise SetupError("Only 3D and 2D boxes are considered for doping.")
+                inf_dims += 1
                 normal_dim = dim
 
+        if normal_dim is not None and inf_dims < 3:
+            raise SetupError("Only 3D and 2D boxes are considered for doping.")
```
momchil-flex: I am actually very confused by this function now that I look at it.

- With this modification, it returns *the last* infinite dimension as `normal_dim`? So if I have [inf, 0, inf], `normal_dim` will be 2?
- Previously this would error in this case?
- What if I have a box which has finite dimensions?

I think we've discussed this function before but I'm still baffled by it.

momchil-flex (commented):
Thanks! We probably need a changelog item about this too, or to modify some of the existing changelog items that were introduced in the previous modification to remove float (if there were such).

tidy3d/components/tcad/doping.py
marc-flex: > With this modification, it returns the last infinite dimension as normal_dim? So if I have [inf, 0, inf], normal_dim will be 2?

In the case of `[inf, 0, inf]`, you'd have `inf_dims < 3` and the error would be raised.

> Previously this would error in this case?

Previously it would have raised the same error.

> What if I have a box which has finite dimensions?

`normal_dim` would be `None`

momchil-flex: Right. I'm still confused by the purpose of this / the definitions. Maybe a quick chat?

tidy3d/components/material/tcad/charge.py
marc-flex: I don't see any issues. We'd just need to implement it in the back end as well as the plotting functions. 

Also, the behavior would change, I think? Currently the `SpatialDataArray` is applied over all the structures which have this medium. If we add it to a doping box it'd be applied only within the box containing the `SpatialDataArray` regardless of the coordinates within that `SpatialDataArray`? 

Other than that, I think you're right we should probably add it to a "doping box". Is this the right PR or should I open a new one for this?

tidy3d/components/scene.py
```diff
@@ -1995,7 +1994,10 @@ def _pcolormesh_shape_doping_box(
             # Handle doping boxes
             if isinstance(doping, tuple):
                 for doping_box in doping:
-                    if isinstance(doping_box, DopingBoxType):
+                    # NOTE: python 3.9 does not support this type comparison: isinstance(doping_box, DopingBoxType)
+                    # instead I need to use a tuple of types
+                    # if isinstance(doping_box, DopingBoxType):
+                    if isinstance(doping_box, (ConstantDoping, GaussianDoping, CustomDoping)):
```
momchil-flex: Apparently you can also do `if isinstance(doping_box, DopingBoxType.__args__):` which should work on both. I think I have a slight preference for this (with a note we can remove `__args__` once python 3.9 is no longer supported, which will be soon). Just because the alternative is that we update `DopingBoxType` later on but forget to update the list here.

tidy3d/components/material/tcad/charge.py
```diff
@@ -309,10 +312,66 @@ class SemiconductorMedium(AbstractChargeMedium):
...
+            return ConstantEnergyBandGap(eg=val)
+        return val
+
+    @pd.validator("N_d", always=True)
+    def check_nd_uses_model(cls, val, values):
+        """Issue deprecation warning if float is provided"""
+        if isinstance(val, (float, int)):
+            log.warning(
+                "Passing a float to 'N_d' is deprecated and will be removed in future versions. "
+                f"Please use a list of 'DopingBoxType' instead, i.e., [ConstantDoping(concentration={val}, center=myCenter, size=mySize)]."
```
momchil-flex: I think you mean `e.g.` instead of `i.e.`

Also we can remove center and size now that they have defaults?

tidy3d/components/tcad/doping.py
```diff
@@ -6,27 +6,26 @@
...
-                normal_dim = dim
-
-        return normal_dim
+    # Override size so that we can set default values
+    size: TracedSize = pd.Field(
+        (inf, inf, inf),
+        title="Size",
+        description="Size in x, y, and z directions.",
+        units=MICROMETER,
+    )
```
momchil-flex, line 28: Wondering if we should set this for every doping box or just the constant one.

In fact for GaussianDoping having infinite size is probably a problem not great in the computation? Well, if it doesn't error or warn about some div by zero, it will just produce constant doping, right?

tidy3d/components/material/tcad/charge.py
```diff
@@ -309,10 +312,66 @@ class SemiconductorMedium(AbstractChargeMedium):
     )
 
     N_d: Union[pd.NonNegativeFloat, SpatialDataArray, tuple[DopingBoxType, ...]] = pd.Field(
-        0,
+        (ConstantDoping(concentration=0),),
```
momchil-flex: Should the default just be an empty tuple?

tidy3d/components/tcad/doping.py
```diff
@@ -316,11 +300,85 @@ def _get_contrib(self, coords: dict, meshgrid: bool = True):
...
+            interp_res = self.concentration.interp(
+                **interp_da, kwargs={"fill_value": 0, "bounds_error": False}
+            )
+            contrib[indices_in_box] = interp_res.values[indices_in_box]
+
+        # NOTE: this is required by Scene
+        if normal_axis is not None and meshgrid:
+            slices = [slice(None)] * X.ndim
+            slices[normal_axis] = 0
+            return contrib[tuple(slices)]
```
momchil-flex: As far as I can tell, the purpose of `normal_axis` now is just to squeeze dimensions along which there's a single coordinate.

Is it possible to completely remove normal_axis and just replace this whole thing

```
        if normal_axis is not None and meshgrid:
            slices = [slice(None)] * X.ndim
            slices[normal_axis] = 0
            return contrib[tuple(slices)]
```

with `return contrib.squeeze()`?

momchil-flex: I mean just a single return, no need for if check.

tidy3d/components/tcad/doping.py
marc-flex: yes, for a Gaussian doping you get the highest concentration everywhere

marc-flex: but happy to move it to the `ConstantDoping` class

momchil-flex: I guess let's leave it.

tidy3d/components/tcad/doping.py
```diff
@@ -316,11 +295,75 @@ def _get_contrib(self, coords: dict, meshgrid: bool = True):
...
+        ...,
+        title="Doping concentration data array.",
+        description="Doping concentration data array.",
+        units=PERCMCUBE,
+    )
+
+    def _get_contrib(self, coords: dict, meshgrid: bool = True):
+        """Returns the contribution to the doping a the locations specified in coords"""
+
+        indices_in_box, X, Y, Z, normal_axis = self._get_indices_in_box(
```
momchil-flex: Ok so now `normal_axis` is unused here; the only place where it is still used as far as I can tell is in the `GaussianDoping._get_contrib`. My preference would be to completely eliminate it because I still feel like it might not be designating what we *think* it's designating in all possible cases. Can you think about what exactly the GaussianDoping handling of [normal_axis] is doing and if it can be modified? I'm not just pushing for this for code brevity: I'd prefer removing the `normal_axis` because it's still not clear to me what exactly it designates.

marc-flex: Of course. I'll get on with that :+1:

marc-flex: It seems like we no longer need the `normal_axis` since we're forcing the boxes to be 3D. Even when the coordinates given are 2D, the `np.exp()` will evaluate to 1 which is equivalent to what we had before.

momchil-flex (approved):

PR #2830, opened 2025-09-19
Lazy loading of data

marcorudolphflex:
Introduce a lazy load path for any Objects so callers can opt into deferred disk reads.

Why?

- Arbitrary-container support for `web.run` requires that a returned SimulationData can stay lightweight until monitors are accessed. Without laziness, returning a nested structure from web.run would eagerly hydrate every simulation and blow up memory.

Solution:

- A proxy object is returned. If any field or method is requested, the object becomes the fully-loaded instance of it.
- You can define a callback `on_load` which is applied on the object after loading it. This can be used for convergence checks after postprocessing.

```
from tidy3d.components.data.sim_data import SimulationData

lazy_sim = SimulationData.from_file("./simulation_data.hdf5", lazy=True)
assert type(lazy_sim).__name__ == "SimulationDataProxy"

# object should only contain relevant metadata to actually load it
assert lazy_sim.__dict__ == {'_lazy_fname': './simulation_data.hdf5', '_lazy_group_path': None, '_lazy_parse_obj_kwargs': {}}

# First access should trigger load
data = lazy_sim.data

assert isinstance(lazy_sim, SimulationData)
```

## Greptile Summary

Updated On: 2025-09-19 15:23:31 UTC

This PR introduces a lazy loading mechanism for Tidy3D data objects, primarily targeting `SimulationData` and related classes. The implementation adds a proxy pattern that defers disk I/O operations until data is actually accessed, addressing memory concerns when working with large datasets or nested structures from `web.run`.

The core functionality is implemented in the `Tidy3dBaseModel` class through a new `_make_lazy_proxy` function that creates dynamic proxy classes. These proxies inherit from the target class but override `__getattribute__` to trigger lazy loading on first access. When any attribute is accessed, the proxy loads the actual data using `dict_from_file` and `parse_obj`, then transforms itself into the fully loaded instance.

Key components of the implementation:
- **Base Model Changes**: `from_file` method now accepts `lazy` and `on_load` parameters, with lazy loading handled by dynamically created proxy classes
- **API Integration**: The `web.load` function now defaults to `lazy=True`, automatically providing memory optimization for users
- **Callback Support**: An `on_load` callback mechanism allows post-processing operations (like convergence checks) to be deferred until data materialization
- **Backward Compatibility**: Existing functionality is preserved while new lazy loading capabilities are opt-in via parameters

The implementation follows a sentinel pattern using `_lazy_fname` in the object's `__dict__` to track loading state. The proxy classes are cached to avoid recreation overhead, and the system properly handles cleanup of cached properties during the loading transition.

**PR Description Notes:**
- There's a documentation error in `sim_data.py` where the example shows `lazy=True` with `to_file()` instead of `from_file()`

## Important Files Changed

<details>

| `CHANGELOG.md` | 5/5 | Added changelog entry documenting the new lazy loading feature for data objects |
| `tidy3d/components/base.py` | 4/5 | Core implementation of lazy loading proxy pattern with dynamic class creation and loading logic |
| `tidy3d/components/data/sim_data.py` | 4/5 | Added documentation for lazy loading but contains incorrect example syntax |
| `tidy3d/web/api/tidy3d_stub.py` | 4/5 | Extended from_file method with lazy loading support and refactored postprocess method |
| `tidy3d/web/api/webapi.py` | 4/5 | Added lazy parameter to load function with default value of True for automatic optimization |

</details>

## Confidence score: 4/5

- This PR introduces significant architectural changes but appears well-implemented with proper proxy patterns and backward compatibility
- Score reflects complex logic changes in critical base classes that require careful review to ensure no edge cases are missed
- Pay close attention to `tidy3d/components/base.py` for the proxy implementation and `tidy3d/components/data/sim_data.py` for the documentation error

## Sequence Diagram
```mermaid
sequenceDiagram
    participant User
    participant Tidy3dStubData
    participant _LazyProxy
    participant Target
    participant OnLoadCallback

    User->>Tidy3dStubData: from_file(fname, lazy=True, on_load=callback)
    Tidy3dStubData->>_LazyProxy: create proxy instance
    _LazyProxy->>_LazyProxy: store fname, group_path, parse_obj_kwargs
    _LazyProxy-->>User: return proxy object

    User->>_LazyProxy: access any field/method
    _LazyProxy->>Target: dict_from_file(fname, group_path)
    Target-->>_LazyProxy: model_dict
    _LazyProxy->>Target: parse_obj(model_dict)
    Target-->>_LazyProxy: target instance
    _LazyProxy->>_LazyProxy: replace __dict__ with target.__dict__
    _LazyProxy->>_LazyProxy: set __class__ to target class
    alt on_load callback provided
        _LazyProxy->>OnLoadCallback: execute callback(self)
        OnLoadCallback-->>_LazyProxy: callback complete
    end
    _LazyProxy-->>User: return requested attribute/method result
```

### DIFF
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -308,7 +308,12 @@ class Tidy3dBaseModel(pydantic.BaseModel):
 
     @classmethod
     def from_file(
-        cls, fname: str, group_path: Optional[str] = None, **parse_obj_kwargs
+        cls,
+        fname: str,
+        group_path: Optional[str] = None,
+        lazy: bool = False,
+        on_load: Optional[Callable] = None,
+        **parse_obj_kwargs,
     ) -> Tidy3dBaseModel:
         """Loads a :class:`Tidy3dBaseModel` from .yaml, .json, .hdf5, or .hdf5.gz file.
 
@@ -316,9 +321,17 @@ class Tidy3dBaseModel(pydantic.BaseModel):
         ----------
         fname : str
             Full path to the file to load the :class:`Tidy3dBaseModel` from.
-        group_path : str, optional
+        group_path : str | None = None
             Path to a group inside the file to use as the base level. Only for hdf5 files.
             Starting `/` is optional.
+        lazy : bool = False
+            Whether to load the actual data (``lazy=False``) or return a proxy that loads
+            the data when accessed (``lazy=True``).
+        on_load : Callable | None = None
+            Callback function executed once the model is fully materialized.
+            Only used if ``lazy=True``. The callback is invoked with the loaded
+            instance as its sole argument, enabling post-processing such as
+            validation, logging, or warnings checks.
         **parse_obj_kwargs
             Keyword arguments passed to either pydantic's ``parse_obj`` function when loading model.
 
@@ -331,8 +344,14 @@ class Tidy3dBaseModel(pydantic.BaseModel):
         -------
         >>> simulation = Simulation.from_file(fname='folder/sim.json') # doctest: +SKIP
         """
+        if lazy:
+            Proxy = _make_lazy_proxy(cls, on_load=on_load)  # staticmethod usage
+            return Proxy(fname, group_path, parse_obj_kwargs)
         model_dict = cls.dict_from_file(fname=fname, group_path=group_path)
-        return cls.parse_obj(model_dict, **parse_obj_kwargs)
+        obj = cls.parse_obj(model_dict, **parse_obj_kwargs)
+        if not lazy and on_load is not None:
+            on_load(obj)
+        return obj
 
     @classmethod
     def dict_from_file(cls, fname: str, group_path: Optional[str] = None) -> dict:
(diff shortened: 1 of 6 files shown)


tidy3d/components/base.py
```diff
@@ -319,6 +324,12 @@ def from_file(
         group_path : str, optional
             Path to a group inside the file to use as the base level. Only for hdf5 files.
             Starting `/` is optional.
+        lazy: bool, optional
+            Whether to load the actual data (lazy=False) or return only a proxy that loads the data when accessed (lazy=True).
+        on_load : callable, optional
+            A callback function to be executed once the model is fully materialized. Only used if ``lazy=True``.
+            The callback is invoked with the loaded instance as its sole argument, allowing post-processing such as
+            validation, logging, or warnings checks.
```
yaugenst-flex: Not numpydoc convention, but we do `name : type = default`
Also, use double backticks for inline code (seems to be used inconsistently here)

tidy3d/components/base.py
```diff
@@ -1201,3 +1215,62 @@ def to_sci(value: float, exponent: int, precision: int) -> str:
...
+        Must implement `dict_from_file` and `parse_obj`.
+    on_load : callable, optional
+        A function to call with the fully loaded instance once materialized.
+
+    Returns
+    -------
+    type
+        A class named `<TargetClsName>Proxy` with init args:
+        `(fname, group_path, parse_obj_kwargs)`.
+    """
```
yaugenst-flex, line 1244: Same formatting comments here too

yaugenst-flex (approved):
Thanks @marcorudolphflex overall I think this looks great and will work. Coupler minor nitpicks but the overall LGTM.

tidy3d/components/base.py
```diff
@@ -1201,3 +1215,62 @@ def to_sci(value: float, exponent: int, precision: int) -> str:
...
+            object.__setattr__(self, "_lazy_fname", fname)
+            object.__setattr__(self, "_lazy_group_path", group_path)
+            object.__setattr__(self, "_lazy_parse_obj_kwargs", dict(parse_obj_kwargs or {}))
+
+        def __getattribute__(self, name: str):
+            if name in (
+                "__class__",
+                "__dict__",
+                "__weakref__",
+                "__post_root_validators__",
```
frederikschubertflex, line 1268: There are more private fields that are set by `pydantic` that need to be taken are of, e.g. `__fields_set__` which is required for calling `.copy()`. So if someone copies a lazily loaded `SimulationData` instance it will fail.

tidy3d/components/base.py
```diff
@@ -1201,3 +1215,62 @@ def to_sci(value: float, exponent: int, precision: int) -> str:
...
+                group_path = d["_lazy_group_path"]
+                kwargs = d["_lazy_parse_obj_kwargs"]
+
+                model_dict = target_cls.dict_from_file(fname=fname, group_path=group_path)
+                target = target_cls.parse_obj(model_dict, **kwargs)
+
+                d.clear()
+                d.update(target.__dict__)
+                object.__setattr__(self, "__class__", target_cls)
+                object.__setattr__(self, "_cached_properties", {})
```
frederikschubertflex: So, regarding the comment above, we also need to copy all other private attributes over, i.e.

```py
object.__setattr__(self, "__fields_set__", set(target.__fields_set__))
private_attrs = getattr(target, "__private_attributes__", {}) or {}
for attr_name in private_attrs:
    object.__setattr__(self, attr_name, getattr(target, attr_name))
```

Replace `object.__setattr__(self, "_cached_properties", {})` with this snippet.

frederikschubertflex (changes requested):
I like this change!

@marcorudolphflex  Can you temporarily change the default for lazy to `True` and re-run our full test suite. Just as a kind of smoke test to catch anything that we might have missed.

marcorudolphflex:
> I like this change!
> 
> @marcorudolphflex Can you temporarily change the default for lazy to `True` and re-run our full test suite. Just as a kind of smoke test to catch anything that we might have missed.

good idea!
I just did that and the errors were "false positives":

```
def test_to_json(tmp_path):
    sim_data = make_sim_data()
    FNAME = str(tmp_path / "sim_data_refactor.json")
    sim_data.to_file(fname=FNAME)

    # saving to json does not store data, so trying to load from file will trigger custom error.
    with pytest.raises(pydantic.ValidationError):
        _ = SimulationData.from_file(fname=FNAME)
```
```
def test_io_json(tmp_path):
    """to json warns and then from json errors."""
    path = str(tmp_path / "custom_source.json")
    with AssertLogLevel("WARNING"):
        FIELD_SRC.to_file(path)
    with AssertLogLevel("WARNING"):
        FIELD_SRC2 = FIELD_SRC.from_file(path)
    assert FIELD_SRC2.field_dataset is None
```

No error here as not loaded yet.

frederikschubertflex (approved):

PR #2837, opened 2025-09-23
FXC-3517 add impedance calculations to mode solver

### LINKED ISSUE #1735: Automatic setup of path integrals

dmarek-flex:
add a ImpedanceSpec for controlling how characteristic impedance is calculated from modes

backend PR: https://github.com/flexcompute/compute/pull/2497

~- [ ] Finalize docs including the new Microwave components somewhere (monitors/mode_specs/path_specs)~
Now tracking [here](https://github.com/flexcompute/tidy3d/pull/2897)
- [x] Implement more precise mode plane bounds calculation raised by @dbochkov-flexcompute 
- [x] Support ModeSimulation

<h2>Greptile Overview</h2>

Updated On: 2025-10-07 20:11:19 UTC

<h3>Summary</h3>
This PR introduces a comprehensive impedance calculation system for microwave mode solvers in Tidy3D. The implementation adds the ability to calculate characteristic impedance from electromagnetic modes, which is essential for microwave and RF transmission line analysis.

**Key architectural changes:**

- **New impedance specification system**: Introduces `ImpedanceSpecTypes` union with `AutoImpedanceSpec` for automatic impedance calculation and `CustomImpedanceSpec` for user-defined voltage/current path integrals
- **Microwave mode specifications**: Adds `MicrowaveModeSpec` extending the base mode specification with impedance calculation capabilities for each mode
- **Enhanced mode solver integration**: Extends `ModeSpec`, `ModeSolver`, and related classes to support microwave specifications while maintaining backward compatibility
- **Path integral infrastructure**: Comprehensive system for voltage and current path integral specifications including axis-aligned, custom 2D, and composite current integrals
- **New monitor types**: Adds `MicrowaveModeMonitor` and `MicrowaveModeSolverMonitor` for microwave-specific data collection
- **Data extensions**: New `MicrowaveModeData` and `MicrowaveModeSolverData` classes that include transmission line characteristics (impedance, voltage/current coefficients)

**Implementation highlights:**

- **Mode plane analysis**: Automatic conductor detection and path generation for simple geometries through `ModePlaneAnalyzer`
- **Flexible impedance calculation**: Supports both direct calculation (when voltage and current integrals are available) and fallback methods using power flux
- **Comprehensive validation**: Multi-level validation ensures proper setup before expensive simulation runs
- **Clean refactoring**: Established microwave component hierarchy with `MicrowaveBaseModel` base class for consistent RF licensing and functionality
- **API consistency improvements**: Standardized naming conventions for path integral classes (e.g., `VoltageIntegralAxisAligned` → `AxisAlignedVoltageIntegral`)

The system integrates seamlessly with existing Tidy3D workflows - users can continue using standard mode specifications for basic electromagnetic analysis while accessing advanced impedance calculations through the new microwave specifications. The architecture properly separates automatic impedance calculation for common cases from custom path integral specifications for complex geometries.

**PR Description Notes:**
- The PR description references an incomplete checklist item about finalizing documentation, but the migration guide and API documentation appear comprehensive

<details>

| `tidy3d/components/microwave/path_integrals/impedance_spec.py` | 4/5 | New impedance specification system with Auto and Custom variants, comprehensive validation |
| `tidy3d/components/microwave/microwave_mode_spec.py` | 5/5 | Clean microwave mode specification implementation with proper caching and validation |
| `tidy3d/plugins/microwave/impedance_calculator.py` | 4/5 | Core impedance calculator with voltage/current integral support, minor style improvements needed |
| `tidy3d/components/mode_spec.py` | 3/5 | Contains syntax error in validation message that needs fixing |
| `tidy3d/components/microwave/path_integrals/mode_plane_analyzer.py` | 4/5 | Sophisticated conductor analysis for automatic impedance calculation |
| `tidy3d/components/mode/mode_solver.py` | 4/5 | Major extension to support microwave impedance calculations while maintaining compatibility |
| `tidy3d/components/microwave/path_integrals/base_spec.py` | 5/5 | Well-designed base classes for path integral specifications |
| `tidy3d/components/microwave/path_integrals/voltage_spec.py` | 4/5 | Voltage path integral specifications with minor spelling error in docstring |
| `tidy3d/components/microwave/path_integrals/current_spec.py` | 5/5 | Comprehensive current path integral specifications including composite support |
| `tidy3d/components/microwave/monitor.py` | 5/5 | Clean microwave monitor implementations extending existing monitors |
| `tidy3d/components/microwave/data/monitor_data.py` | 5/5 | Well-structured microwave data classes with proper inheritance |
| `tidy3d/components/microwave/base.py` | 5/5 | Clean base class for microwave components with RF licensing warnings |
| `tidy3d/components/simulation.py` | 4/5 | Added microwave mode specification validation with proper error handling |
| `tidy3d/components/geometry/utils.py` | 4/5 | Enhanced geometry utilities with new snapping behaviors and flattening functions |
| `tidy3d/plugins/microwave/path_integrals.py` | 4/5 | Refactored path integral execution classes following spec/execution separation |
| `tidy3d/plugins/microwave/custom_path_integrals.py` | 4/5 | Complex custom path integrals with phase analysis and composite current support |
| `docs/api/microwave/microwave_migration.rst` | 5/5 | Comprehensive migration guide for API changes |
| `tidy3d/__init__.py` | 5/5 | Clean addition of microwave components to public API |
| `CHANGELOG.md` | 4/5 | Comprehensive changelog entries with minor duplicate entry cleanup needed |
| Multiple schema files | 5/5 | Properly updated JSON schemas for all new microwave components |

</details>
<h3>Confidence score: 3/5</h3>

- This PR requires careful review due to a syntax error that will cause runtime failures
- Score lowered due to syntax error in validation message and extensive scope of changes requiring thorough testing
- Pay close attention to `tidy3d/components/mode_spec.py` for the missing closing quote in the validation error message

**Context used:**

Rule from `dashboard` - Do not use markdown formatting in exception or warning messages; use single quotes to highlight vari... ([source](https://app.greptile.com/review/custom-context?memory=9b45957a-2d0d-4c24-a5d3-eef6f721cbfd))
Rule from `dashboard` - Update the CHANGELOG.md file when making changes that affect user-facing functionality or fix bugs. ([source](https://app.greptile.com/review/custom-context?memory=a109c62a-aa8c-4cc4-b515-0ad947c47929))


tidy3d/components/microwave/path_integrals/impedance_spec.py
```diff
@@ -0,0 +1,67 @@
...
+class CustomImpedanceSpec(Tidy3dBaseModel):
+    """Specification for custom transmission line voltages and currents in mode solvers.
+
+    The :class:`.CustomImpedanceSpec` class specifies how quantities related to transmission line
+    modes are computed. For example, it defines the paths for line integrals, which are used to
+    compute voltage, current, and characteristic impedance of the transmission line.
+
+    Users may supply their own voltage and current path specifications to control where these integrals
+    are evaluated. If neither voltage nor current specifications are provided, an automatic choice of
+    paths will be made based on the simulation geometry and context.
```
weiliangjin2021: the last sentence is outdated?

tidy3d/components/microwave/path_integrals/specs/impedance.py
```diff
@@ -0,0 +1,67 @@
...
+
+    current_spec: Optional[CurrentPathSpecTypes] = pd.Field(
+        None,
+        title="Current Integration Path",
+        description="Path specification for computing the current associated with each mode. "
+        "The number of path specifications should equal the 'num_modes' field "
+        "in the 'ModeSpec'.",
+    )
+
+    @pd.validator("current_spec", always=True)
```
weiliangjin2021, line 65: @skip_if_fields_missing for voltage_spec

tidy3d/components/mode_spec.py
```diff
@@ -235,3 +244,22 @@ def angle_rotation_with_phi(cls, val, values):
...
+        num_modes = values["num_modes"]
+        valid_number_impedance_specs = val.num_impedance_specs == num_modes
+
+        if not valid_number_impedance_specs:
+            raise SetupError(
+                f"Given {val.num_impedance_specs} impedance specifications in the 'MicrowaveModeSpec', "
+                f"but the number of modes requested is {num_modes}. Please either ensure that the "
+                "number of impedance specifications is equal to the number of modes or leave the "
+                "'MicrowaveModeSpec' field as 'None', if impedances are not needed."
+            )
```
weiliangjin2021: Is it common to have the same or different `MicrowaveModeSpec` for each mode?

tidy3d/components/mode/mode_solver.py
```diff
@@ -1357,6 +1370,52 @@ def _filter_polarization(self, mode_solver_data: ModeSolverData):
...
+        """Wrapper for making path integrals from the MicrowaveModeSpec. Note: overriden in the backend to support
+        auto creation of path integrals."""
+        return make_path_integrals(
+            self.mode_spec.microwave_mode_spec,
+            self.to_monitor(name=MODE_MONITOR_NAME),
+        )
+
+    def _add_microwave_data(self, mode_solver_data: ModeSolverData) -> ModeSolverData:
+        """Calculate and add microwave data to ``mode_solver_data`` which uses the path specifications.
+        Note: this modifies ``mode_solver_data`` in-place."""
```
weiliangjin2021: does it modifies in-place?

weiliangjin2021 (commented):
Most went through code structures, and the structure looks in good shape. Some minor comments:

tidy3d/components/mode_spec.py
dmarek-flex: The intention here is to put any additional RF/Microwave specific settings which may control the overall results of the mode solver (like adding options for processing groups of degenerate modes). Right now it only has `impedance_specs` which is needed for each individual mode.

So the purpose of `MicrowaveModeSpec` is to mainly keep rf and photonics things a bit separate. Although we could add `impedance_specs` directly to `ModeSpec`, or `MicrowaveModeSpec` derive from `ModeSpec`. But then we might need like a `MicrowaveModeMonitor`. paging @daquinteroflex

tidy3d/components/mode/mode_solver.py
dmarek-flex: no it doesnt! :sweat_smile:

tidy3d/components/mode_spec.py
daquinteroflex: >  MicrowaveModeSpec derive (inherits) from ModeSpec

would be much better if you can do this @dmarek-flex ? I believe we have a goal to split things more cleanly also for all core fdtd and mode components

ie

```
class MicrowaveModeSpec(ModeSpec):
   microwave_parameters1:  ...
   microwave_parameters2:  ...
```

dmarek-flex: I guess I should do the same with `MicrowaveModeData`?

daquinteroflex: Yeah ideally, let me have a look across the PR iab then

tidy3d/components/microwave/data/dataset.py
```diff
@@ -0,0 +1,46 @@
+"""Post-processing data and figures of merit for antennas, including radiation efficiency,
+reflection efficiency, gain, and realized gain.
+"""
```
dbochkov-flexcompute: misplaced description?

tidy3d/components/microwave/microwave_mode_spec.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class MicrowaveModeSpec(ModeSpec):
+    """
+    The :class:`.MicrowaveModeSpec` class specifies how quantities related to transmission line
+    modes and microwave waveguides are computed. For example, it defines the paths for line integrals, which are used to
+    compute voltage, current, and characteristic impedance of the transmission line.
+    """
+
+    impedance_specs: tuple[Optional[annotate_type(ImpedanceSpecTypes)], ...] = pd.Field(
```
dbochkov-flexcompute: so, some of the elements can be `None`? Would be good to mention what that means in the description

tidy3d/components/microwave/path_integrals/base_spec.py
```diff
@@ -0,0 +1,248 @@
...
+    Given a set of vertices :math:`\\vec{r}_i`, this class approximates path integrals over
+    vector fields of the form :math:`\\int{\\vec{F} \\cdot \\vec{dl}}`
+    as :math:`\\sum_i{\\vec{F}(\\vec{r}_i) \\cdot \\vec{dl}_i}`,
+    where the differential length :math:`\\vec{dl}` is approximated using central differences
+    :math:`\\vec{dl}_i = \\frac{\\vec{r}_{i+1} - \\vec{r}_{i-1}}{2}`.
+    If the path is not closed, forward and backward differences are used at the endpoints.
+    """
+
+    axis: Axis = pd.Field(
+        2, title="Axis", description="Specifies dimension of the planar axis (0,1,2) -> (x,y,z)."
```
dbochkov-flexcompute: is it not dangerous to have a default value here?

tidy3d/components/microwave/path_integrals/specs/impedance.py
```diff
@@ -0,0 +1,70 @@
...
+        "in the 'ModeSpec'.",
+    )
+
+    current_spec: Optional[CurrentPathSpecTypes] = pd.Field(
+        None,
+        title="Current Integration Path",
+        description="Path specification for computing the current associated with each mode. "
+        "The number of path specifications should equal the 'num_modes' field "
+        "in the 'ModeSpec'.",
+    )
```
dbochkov-flexcompute: from the description, it sounds like these fields are tuples?

tidy3d/components/microwave/path_integrals/types.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
```
dbochkov-flexcompute, line 3: missing docstring at the beginning of the file

tidy3d/plugins/microwave/custom_path_integrals.py
```diff
@@ -123,8 +97,8 @@ def compute_integral(
         field2_interp = field2.interp(path_indexer, method="linear")
 
         # Determine the differential length elements along the path
-        dl_x = self._compute_dl_component(x_path, self.is_closed_contour)
-        dl_y = self._compute_dl_component(y_path, self.is_closed_contour)
+        dl_x = CustomPathIntegral2DSpec._compute_dl_component(x_path, self.is_closed_contour)
+        dl_y = CustomPathIntegral2DSpec._compute_dl_component(y_path, self.is_closed_contour)
```
dbochkov-flexcompute: for closed curves first and last points are the same, are we accounting twice the same point when performing integration here? should one of the points be excluded?

dbochkov-flexcompute (changes requested):
Very impressive work, finished the first pass and left some comments. Additionally, it seems like many new classes `AxisAlignedPathIntegralSpec`, `CustomPathIntegral2DSpec`, `CurrentIntegralAxisAlignedSpec`, `CustomCurrentIntegral2DSpec`, `CompositeCurrentIntegralSpec`, `CustomImpedanceSpec`, etc are missing examples in docstrings

tidy3d/components/microwave/path_integrals/base_spec.py
```diff
@@ -0,0 +1,248 @@
...
+        description="List of (d1, d2) defining the 2 dimensional positions of the path. "
+        "The index of dimension should be in the ascending order, which means "
+        "if the axis corresponds with ``y``, the coordinates of the vertices should be (x, z). "
+        "If you wish to indicate a closed contour, the final vertex should be made "
+        "equal to the first vertex, i.e., ``vertices[-1] == vertices[0]``",
+        units=MICROMETER,
+    )
+
+    @staticmethod
+    def _compute_dl_component(coord_array: xr.DataArray, closed_contour=False) -> np.array:
```
dbochkov-flexcompute: Is this the integration weight associated with each point? I wonder if using numpy gradient gives the right thing here at end points. Maybe it would be simpler to explicitly follow the trapezoidal integration? That is, $\int f_x dx = \sum 0.5 (f_x^{i} + f_x^{i+1}) \Delta x_i = \sum f_x^{i} 0.5 (\Delta x_{i-1} + \Delta x_i) = \sum f_x^{i} dl_x^i$, or, in other words, each points gets half the length of each existing neighboring segment. Here's a quick test code snippet:
```
import numpy as np

def _compute_dl_component(coord_array, closed_contour=False) -> np.array:
    """Computes the differential length element along the integration path."""
    dl = np.gradient(coord_array)
    if closed_contour:
        # If the contour is closed, we can use central difference on the starting/end point
        # which will be more accurate than the default forward/backward choice in np.gradient
        grad_end = np.gradient([coord_array[-2], coord_array[0], coord_array[1]])
        dl[0] = dl[-1] = grad_end[1]
    return dl

open_curve = np.logspace(0, 1, 10)
closed_curve = np.concatenate([open_curve, (11 - open_curve)[1:]])

dl_open = _compute_dl_component(open_curve, False)
dl_closed = _compute_dl_component(closed_curve, True)

# expected 9 and 0
print(np.sum(dl_open), np.sum(dl_closed), np.sum(dl_closed[:-1]))

def trap_dl(coord_array):
    dl = np.zeros_like(coord_array)
    diff_over_two = 0.5 * np.diff(coord_array)
    dl[1:] += diff_over_two
    dl[:-1] += diff_over_two
    return dl

dl_open_trap = trap_dl(open_curve)
dl_closed_trap = trap_dl(closed_curve)

# expected 9 and 0
print(np.sum(dl_open_trap), np.sum(dl_closed_trap))

# output:
# 10.274456419101808 -0.9829067540869239 -4.440892098500626e-16
# 9.0 -4.440892098500626e-16
```

tidy3d/plugins/microwave/custom_path_integrals.py
dbochkov-flexcompute: ah, I missed that you later do `.integrate()` which should take care of that. Ignore this comment

tidy3d/components/microwave/path_integrals/base_spec.py
dbochkov-flexcompute: ignore this comment as well

tests/test_components/test_geometry.py
```diff
@@ -1158,12 +1175,78 @@ def test_snap_box_to_grid(snap_location, snap_behavior):
...
+        location=[SnapLocation.Boundary] * 3, behavior=[SnapBehavior.StrictExpand] * 3
+    )
+
+    expanded_box = snap_box_to_grid(grid, box_coincident, snap_spec_strict_expand)
+
+    # StrictExpand should move bounds outwards even when already on grid
+    assert expanded_box.bounds[0][0] < 0.1  # Left bound moved left from 0.1
+    assert expanded_box.bounds[1][0] > 0.1  # Right bound moved right from 0.1
+    assert expanded_box.bounds[0][1] < 0.2  # Bottom bound moved down from 0.2
+    assert expanded_box.bounds[1][1] > 0.2  # Top bound moved up from 0.2
```
dbochkov-flexcompute: Does the box gets snapped to [0, 0.1, 0.2] x [0.2, 0.3, 0.4]? Would it make sense to make these checks stricter to check for exact bounds?

tidy3d/__init__.py
```diff
@@ -20,6 +20,22 @@
...
+    CurrentIntegralAxisAlignedSpec,
+    CustomCurrentIntegral2DSpec,
+)
+from tidy3d.components.microwave.path_integrals.impedance_spec import (
+    AutoImpedanceSpec,
+    CustomImpedanceSpec,
+)
+from tidy3d.components.microwave.path_integrals.voltage_spec import (
+    CustomVoltageIntegral2DSpec,
+    VoltageIntegralAxisAlignedSpec,
```
dbochkov-flexcompute: the naming seems a tiny bit inconsistent: if we take `VolumeIntegral` as the base name, should it be then `Custom2DVoltageIntegralSpec` and `AxisAlignedVoltageIntegralSpec`? and, similarly, for `CurrentIntegral`?

dmarek-flex:
Thanks @dbochkov-flexcompute for finding those docstring mistakes. It was a little bit harder than expected to get doc tests running correctly because of the log warning we have for rf and microwave components. I got it working by introducing the `MicrowaveBaseModel` class. @daquinteroflex  does this approach make sense for replacing the duplicated method for warning.

tidy3d/components/mode_spec.py
dmarek-flex: Ultimately, I had to revert back to a composition approach. Inheritance led to too many changes related to `ModeData`/`ModeSolverData`. 

But I did manage to allow for broadcasting a single impedance spec to all modes

docs/api/microwave/ports/wave.rst
```diff
@@ -46,27 +46,27 @@ If it is desired to only solve for the 2D port mode, one can use the ``to_mode_s
...
-   tidy3d.plugins.microwave.CurrentIntegralAxisAligned
-   tidy3d.plugins.microwave.CustomVoltageIntegral2D
-   tidy3d.plugins.microwave.CustomCurrentIntegral2D
+   tidy3d.plugins.microwave.AxisAlignedVoltageIntegral
+   tidy3d.plugins.microwave.AxisAlignedCurrentIntegral
+   tidy3d.plugins.microwave.Custom2DVoltageIntegral
+   tidy3d.plugins.microwave.Custom2DCurrentIntegral
    tidy3d.plugins.microwave.AxisAlignedPathIntegral
-   tidy3d.plugins.microwave.CustomPathIntegral2D
+   tidy3d.plugins.microwave.Custom2DPathIntegral
```
daquinteroflex, line 54: This was only introduced in rc1 right?

tidy3d/components/data/monitor_data.py
```diff
@@ -1645,11 +1646,18 @@ class ModeData(ModeSolverDataset, ElectromagneticFieldData):
 
     eps_spec: list[EpsSpecType] = pd.Field(
         None,
-        title="Permettivity Specification",
+        title="Permittivity Specification",
         description="Characterization of the permittivity profile on the plane where modes are "
         "computed. Possible values are 'diagonal', 'tensorial_real', 'tensorial_complex'.",
     )
 
+    microwave_data: Optional[MicrowaveModeDataset] = pd.Field(
```
daquinteroflex: Sorry wasn't it meant to be a MicrowaveModeData instead of adding this parameter?

tidy3d/components/lumped_element.py
```diff
@@ -464,7 +456,7 @@ def geometry(self) -> ClipOperation:
         return self.to_geometry()
 
 
-class NetworkConversions(Tidy3dBaseModel):
+class NetworkConversions(MicrowaveBaseModel):
```
daquinteroflex: I feel this should not be a class but a collection of functions instead but this was here previously though

tidy3d/components/microwave/base.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @pd.root_validator(pre=False)
+    def _warn_rf_license(cls, values):
+        # Skip warning during doctest runs
+        if not is_running_doctest():
+            log.warning(
+                "ℹ️ ⚠️ RF simulations are subject to new license requirements in the future. You have instantiated at least one RF-specific component.",
+                log_once=True,
+            )
+        return values
```
daquinteroflex, line 27: This sounds good! So I'm just curious how this is visualised in the notebooks but I guess will be the same

tidy3d/components/microwave/base.py
```diff
@@ -0,0 +1,40 @@
...
+        return True
+
+    # Check for standard doctest module execution
+    if "doctest" in sys.modules:
+        doctest_mod = sys.modules["doctest"]
+        # Check if doctest.master exists (indicates active doctest run)
+        if hasattr(doctest_mod, "master") and doctest_mod.master is not None:
+            return True
+
+    return False
```
daquinteroflex: This should probably live in tidy3d/utils.py

tidy3d/components/microwave/path_integrals/factory.py
```diff
@@ -0,0 +1,144 @@
+"""Factory functions for creating current and voltage path integrals from path specifications."""
```
daquinteroflex, line 1: Great! Are we intending to expose this to the user?

tidy3d/components/mode_spec.py
```diff
@@ -20,7 +21,7 @@
 
 class ModeSpec(Tidy3dBaseModel):
```
daquinteroflex: So yeah why simply not make a class MicrowaveModeSpec(ModeSpec) ?

tidy3d/plugins/microwave/__init__.py
```diff
@@ -17,29 +17,32 @@
...
     "ChebWindow",
-    "CurrentIntegralAxisAligned",
+    "CompositeCurrentIntegral",
     "CurrentIntegralTypes",
-    "CustomCurrentIntegral2D",
-    "CustomPathIntegral2D",
-    "CustomVoltageIntegral2D",
+    "Custom2DCurrentIntegral",
+    "Custom2DPathIntegral",
+    "Custom2DVoltageIntegral",
```
daquinteroflex, line 58: Agree with the standardisation just got to check it is not in 2.9

daquinteroflex (commented):
Looks pretty good @dmarek-flex ! Just a few questions on very few API design considerations

tidy3d/components/mode_spec.py
dmarek-flex: I did try this approach and it worked ok for the `ModeSpec`/`MicrowaveModeSpec` side of things. But then I felt like I would need to have a `MicrowaveModeData`/`MicrowaveModeSolverData` and that would require modifications all over the code base (frontend and backend) to provide the correct type hints and Pydantic fields. 

And then following that logic we would need a `MicrowaveModeMonitor`/`MicrowaveModeSolverMonitor`/`MicrowaveModeSimulation` and it felt like maintaining the proper type hints would be a lot of extra work. From trying this out, inheritance did not seem the easiest approach for maintenance, and I decided that composition would be a better approach.

tidy3d/components/data/monitor_data.py
dmarek-flex: See my other comment on ModeSpec

docs/api/microwave/ports/wave.rst
dmarek-flex: Path integrals have been since ~2.7, but we are taking the opportunity here to get in some breaking changes before RF release.

tidy3d/components/microwave/base.py
dmarek-flex: will check

tidy3d/components/lumped_element.py
dmarek-flex: Yes, I will avoid empty classes moving forward

tidy3d/components/mode_spec.py
daquinteroflex: @yaugenst-flex @tylerflex it'd be good to get your input here in terms of how we do separation of scope rather than mixing RF into the general ModeSpec as currently implemented in this PR per the following lines in this file. It is concievable that `MicrowaveModeSpec(ModeSpec)` keeps getting extended as well as future  classes that Damian mentioned in order to provide custom RF functionality eventually, but it is a larger effort in terms of the API restructure that we also have to consider. The benefit though is a stronger separation of API, which is related to previous conversations we've had privately about product scope.

https://github.com/flexcompute/tidy3d/blob/a4c3c40c07851002e908f5ca4ce443410133f35c/tidy3d/components/mode_spec.py#L167-L173

tylerflex: Yea I'd rather make a dedicated `MicrowaveModeSpec(ModeSpec)` in a separate place vs mixing it into the component as `microwave_spec: Optional[MicrowaveModeSpec] = pd.Field`

eventually I think all of the `RF`-specific components should be moved into their own separate namespace, like a tidy3d.rf, and that includes terminal component modeler, MicrowaveModeSpec, and probably a separate RFSimulation.

To me that seems like the long term direction (a separate rf python package that extends tidy3d, but tidy3d proper can be independent of it)

docs/api/microwave/ports/wave.rst
daquinteroflex: Hmm maybe we need a breaking changes section of 2.10

tidy3d/components/mode_spec.py
dmarek-flex: `MicrowaveModeSpec` makes sense but then I need to create new classes that accept it, which will lead to `MicrowaveModeData`/`MicrowaveModeSolverData`/`MicrowaveModeMonitor`/`MicrowaveModeSolverMonitor`/`MicrowaveModeSimulation` versions?

Or do I allow a `ModeMonitor` to accept both mode_specs but produce either `ModeData` or `MicrowaveModeData` depending on the spec?

daquinteroflex: Ultimately, sounds like we should go this way in the long run @dmarek-flex though, and try to have a cleaner separation for RF. Happy to chat to scope this out if we want to do this for 2.10 too before getting to MicrowaveSimulation? This said it'd be good for the product council to discuss the extra effort this might take to start doing this separation in the RF product plans @tylerflex ?

tylerflex: yea, it's a tough call, because we will start to have tons of classes, but I also feel it should happen eventually. Hopefully most of these could be very thin wrappers around the tidy3d objects that they inherit from?

Ideally we can get to a place where  `tidy3d` just contains stuff for basic fdtd + photonics. and `tidy3d.rf` extends `tidy3d` and adds  anything specific to microwave. I think we'll necessarily have to add more classes there but hopefully most things can be shared.

docs/api/microwave/ports/wave.rst
daquinteroflex: We also need to update a bunch of notebooks too

tidy3d/components/microwave/path_integrals/factory.py
dmarek-flex: No, users can make path integrals directly. I needed this to get around circular dependencies

docs/api/microwave/ports/wave.rst
dmarek-flex: I have a notebook branch incorporating these changes.

`Hmm maybe we need a breaking changes section of 2.10`
Like in the docs or changelog?

daquinteroflex: Maybe in Changelog given it's a special one

tidy3d/components/lumped_element.py
daquinteroflex: Can also take this opportunity we're breaking things now to do this before 2.10 too

dmarek-flex: :saluting_face:

tidy3d/components/microwave/base.py
dmarek-flex: looks fine to me

docs/api/microwave/ports/wave.rst
dmarek-flex: OK added a migration guide to the docs and mentioned the breaking changes in the changelog

dmarek-flex:
Alright lets have another go @greptileai. Also since @daquinteroflex is away, can @tylerflex or @yaugenst-flex take a look at this refactoring of the `MicrowaveModeSpec` and monitors/monitor data

CHANGELOG.md
```diff
@@ -33,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `LayerRefinementSpec` defaults to assuming structures made of different materials are interior-disjoint for more efficient mesh generation.
 - Removal of the Adjoint plugin.
 - Deprecation of Python 3.9 support.
+- Removal of the Adjoint plugin.
```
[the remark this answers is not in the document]
dmarek-flex: rebase mistake

tidy3d/components/microwave/data/dataset.py
```diff
@@ -0,0 +1,44 @@
...
+from tidy3d.components.data.dataset import Dataset
+
+
+class TransmissionLineDataset(Dataset):
+    """Holds mode data that is specific to transmission lines in microwave and RF applications, like characteristic impedance.
+
+    Notes
+    -----
+        The data in this class is only calculated when a :class:`MicrowaveModeSpec`
+        is provided to the :class:`ModeMonitor`, :class:`ModeSolverMonitor`, or mode solver.
```
weiliangjin2021: also :class:`ModeSolver`

tidy3d/components/microwave/data/dataset.py
```diff
@@ -0,0 +1,44 @@
...
+        "complex-valued mode amplitude.",
+    )
+
+    current_coeffs: CurrentFreqModeDataArray = pd.Field(
+        ...,
+        title="Mode Current Coefficients",
+        description="Quantity calculated for transmission lines, which associates "
+        "a current-like quantity with each mode profile that scales linearly with the "
+        "complex-valued mode amplitude.",
+    )
```
weiliangjin2021, line 46: Is the purpose of storing `voltage_coeffs` and `current_coeffs` to reconstruct Z0 if needed? If so, shall we also store flux when Z0 is computed through PV or PI approach?

tidy3d/components/geometry/utils.py
```diff
@@ -43,6 +50,44 @@
...
+    if isinstance(geoms, Shapely):
+        geoms = [geoms]
+
+    flat = []
+    for geom in geoms:
+        if geom.is_empty:
+            continue
+        if isinstance(geom, keep_types):
+            flat.append(geom)
+        elif isinstance(geom, BaseMultipartGeometry):
```
weiliangjin2021, line 86: I cannot find `BaseMultipartGeometry` on their doc page. Could you explain a bit here?

tidy3d/components/microwave/path_integrals/factory.py
```diff
@@ -0,0 +1,148 @@
...
+    Given a microwave mode specification and monitor, create the voltage and
+    current path integrals used for the impedance computation.
+
+    Parameters
+    ----------
+    microwave_mode_spec : MicrowaveModeSpec
+        Microwave mode specification for creating voltage and current path specifications.
+    auto_spec: Optional[CustomImpedanceSpec]
+        The automatically created impedance specification, if available.
+    Returns
```
weiliangjin2021, line 105: One line spacing above "Returns"

weiliangjin2021 (commented):
Very impressive work! All look solid. One question: all integral_specs are defined in component/microwave, but the actual integration computation is defined in plugin/microwave. Should we unify this by moving them from plugin to component?

tidy3d/components/geometry/utils.py
dmarek-flex: It seemed like a simple way to filter out all geometry types that are collections of geometries. It is inherited by `MultiPoint`/`MultiLine`/`MultiPolygon`/`GeometryCollection`.

https://github.com/shapely/shapely/blob/main/shapely/geometry/base.py#L1021-L1077

tidy3d/components/microwave/data/dataset.py
dmarek-flex: The purpose of those are to make the multimodal wave ports much easier to implement. We can also allow users to plot voltage/current at the `ModePlane` using this data. The way I did it was to always return a voltage and current pair no matter the PV/PI/PVI approach. In the PI/PV case, the voltage/current will be calculated from the flux and current/voltage.

weiliangjin2021: so do you think flux should be stored too?

dmarek-flex: Don't think it is needed, flux is not directly needed and it is in the `MonitorData` anyways

tylerflex:
@dmarek-flex @weiliangjin2021 

> One question: all integral_specs are defined in component/microwave, but the actual integration computation is defined in plugin/microwave. Should we unify this by moving them from plugin to component?

We should figure out a plan regarding the RF/Tidy3D separation and organization long term.

@weiliangjin2021 just for a sense, here are some related discussions https://github.com/flexcompute/tidy3d/pull/2837#discussion_r2393831503

My feeling is that the best long term solution is a complete separation of RF from the main tidy3d components. Meaning perhaps a `tidy3d.rf` namespace where all of the RF components, TerminalComponentModeler, microwave plugin, etc. can live.

But for this to be a reality, we would need to (at least):
* Figure out how we want to extend tidy3d components / schemas to RF, eg. that `ModeSpec` discussion. Perhaps inheritance for start? with RF then having it's own schema that extends the tidy3d schema
* Write the web API in a more generic way so we can easily extend it. Maybe using `Protocol`.
* We may need dedicated web endpoints for RF, which we should eventually have but will be a bigger effort.

Could you guys think about this a bit in terms of requirements from your side and any preferences? @weiliangjin2021 let's discuss together with yannick and Lucas at the tech council meeting monday to get a general feeling because the same issues arise in photonforge.

weiliangjin2021 (approved):

tylerflex:
I just started looking at this:

One idea occurred to me, which is that if we eventually do want to import `rf` components through a `tidy3d.rf` namespace, and this helps with organization, we could already create this namespace, eg `tidy3d/rf/__init__.py` that imports all RF related things directly, including `TerminalComponentModeler`.  And then deprecate or leave out of the main `tidy3d/__init__.py`

While underlying things would still be all mixed without any changes, this would at least give us an ability to start crafting the user-facing side of the package.

note sure if worth doing here since this is a needed feature, but wanted to make a note.

@yaugenst-flex @daquinteroflex   FYI

docs/api/microwave/microwave_migration.rst
```diff
@@ -0,0 +1,130 @@
+.. _microwave_migration:
+
+v2.10 Refactor Migration
```
daquinteroflex, line 3: This is very good thanks

tidy3d/web/api/tidy3d_stub.py
```diff
@@ -245,6 +246,7 @@ def from_file(
         supported_data_classes = [
             SimulationData,
             ModeSolverData,
+            MicrowaveModeSolverData,
```
daquinteroflex, line 249: Curious you've tried this works well via web?

tidy3d/components/microwave/base.py
```diff
@@ -0,0 +1,31 @@
...
+
+from tidy3d.components.base import Tidy3dBaseModel
+from tidy3d.log import log
+from tidy3d.utils import is_running_pytest
+
+
+class MicrowaveBaseModel(Tidy3dBaseModel):
+    """Base model that all RF and microwave specific components inherit from."""
+
+    suppress_instantiation_warnings: bool = pd.Field(
```
dbochkov-flexcompute: just curios, is this meant to be on instance-by-instance choice, or more like the entire classes?

tidy3d/components/boundary.py
```diff
@@ -1127,7 +1129,7 @@ def mode_abc(
         ----------
         plane: Box
             Cross-sectional plane in which the absorbed mode will be computed.
-        mode_spec: ModeSpec = ModeSpec()
+        mode_spec: ModeSpecTypes = ModeSpec()
```
daquinteroflex: Should be ModeSpecType not Types

tidy3d/components/lumped_element.py
```diff
@@ -464,98 +456,100 @@ def geometry(self) -> ClipOperation:
         return self.to_geometry()
 
 
-class NetworkConversions(Tidy3dBaseModel):
-    """Helper functionality for directly computing complex conductivity and permittivities using
-    equations in _`[1]`. Useful for testing the direct translations of lumped network parameters into
-    an equivalent PoleResidue medium.
+def network_complex_conductivity(
```
daquinteroflex, line 459: So happy to see this

tidy3d/components/types/mode_spec.py
```diff
@@ -0,0 +1,11 @@
...
+
+from __future__ import annotations
+
+from typing import Union
+
+from tidy3d.components.microwave.mode_spec import MicrowaveModeSpec
+from tidy3d.components.mode_spec import ModeSpec
+
+# Type aliases
+ModeSpecTypes = Union[ModeSpec, MicrowaveModeSpec]
```
daquinteroflex: ModeSpecType

daquinteroflex (commented):
So we've decided to refactor this in a separate PR into tidy3d.rf, for now feel free to merge but won't be like this for rc3

Will do another review pass tomorrow my morning.

tidy3d/components/data/data_array.py
```diff
@@ -208,6 +208,13 @@ def abs(self):
         """Absolute value of data array."""
         return abs(self)
 
+    @property
+    def angle(self):
+        """Angle or phase value of data array."""
+        values = np.angle(self.values)
+        SelfType = type(self)
+        return SelfType(values, coords=self.coords)
```
daquinteroflex: why out of curiosity?

tidy3d/components/microwave/monitor.py
```diff
@@ -0,0 +1,70 @@
...
+        We can also use the mode amplitudes recorded in the mode monitor to reveal the decomposition
+        of the radiated power into forward- and backward-propagating modes, respectively.
+
+        .. TODO give an example of how to extract the data from this mode.
+
+        .. TODO add derivation in the notebook.
+
+        .. TODO add link to method
+
+        .. TODO add links to notebooks correspondingly
```
daquinteroflex, line 31: Probably not worth completing this until we do the refactor, but we should make a jira ticket?

tidy3d/components/mode/data/sim_data.py
```diff
@@ -22,10 +23,11 @@ class ModeSimulationData(AbstractYeeGridSimulationData):
         ..., title="Mode simulation", description="Mode simulation associated with this data."
     )
 
-    modes_raw: ModeSolverData = pd.Field(
+    modes_raw: ModeSolverDataTypes = pd.Field(
```
daquinteroflex: ModeSolverDataType

daquinteroflex (approved):
Just the few minor comments

tidy3d/web/api/tidy3d_stub.py
dmarek-flex: I have tested running with `ModeSimulation` and also `ModeSolver` types through the web api run. Not sure how to test the web api through the mode solver plugin with a custom `solver_version`, but I assume that API is going to be deprecated at some point?

I'll just double check submitting a Simulation to the web that includes a `MicrowaveModeSolverMonitor`/`MicrowaveModeMonitor`

tidy3d/components/mode/data/sim_data.py
dmarek-flex: I changed some other "*Types" introduced in the RF refactor to follow this same cvonvention.

tidy3d/components/microwave/base.py
dmarek-flex: I guess it could be more like a config-level switch. I will explore briefly.

The intention is to be able to turn off this warning for doctests, since I was unable to find a away to get all doctests passing when this warning is present. And I don't want to ignore those tests!

tidy3d/components/data/data_array.py
dmarek-flex: I want to make sure the return DataArray has the correct type (e.g. if you call .angle on a `FreqDataArray` I want it to return `FreqDataArray`.

I found a better way I think though using `_replace`

tidy3d/components/microwave/monitor.py
dmarek-flex: These Todo's are also present in the normal `ModeMonitor`. Ill add it to Jira

dmarek-flex: https://flow360.atlassian.net/browse/FXC-3675

tidy3d/components/data/data_array.py
dmarek-flex: Well actually `_replace` is not working properly in some tests for some reason. Reverting to returning a new instance of the correct type.

tidy3d/components/microwave/base.py
dmarek-flex: The config approach worked nicely. But had to include a helper class method to instantiate default versions of classes in Pydantic models. Otherwise we get license warning as soon as we `import tidy3d`

dmarek-flex:
Just made some final tests with the deployed solver and everything looks good. I decided to submit the docs changes in a different PR so that we can take our time with that https://github.com/flexcompute/tidy3d/pull/2897

@dbochkov-flexcompute Let me know if you approve of the changes regarding the license warning.

tidy3d/components/microwave/base.py
dbochkov-flexcompute: Looks fine too me!

dbochkov-flexcompute (approved):

PR #2854, opened 2025-09-30
Fix subdivision for named 2D materials

caseyflex:
2D materials are subdivided when they have inhomogeneous substrates. The new structures were inheriting the name from the old structure, giving an error due to name reuse. Also, this validator was not run for EME simulations.

This PR:

- fixes the subdivision and adds a test
- checks `sim._finalized` for any `AbstractYeeGridSimulation` in `validate_pre_upload` (which must now be called by subclasses)

<h2>Greptile Overview</h2>

Updated On: 2025-09-30 20:38:03 UTC

<h3>Summary</h3>
This PR fixes a critical bug affecting named 2D materials with inhomogeneous substrates. When Tidy3D encounters 2D materials on inhomogeneous substrates, it automatically subdivides them into volumetric equivalents for simulation accuracy. However, the subdivision process was inheriting the original structure's name for all subdivided pieces, causing validation errors due to duplicate names within the simulation.

The fix implements two key changes: First, it modifies the subdivision logic in `simulation.py` to append unique suffixes (`_SUBDIVIDED[{i}]`) to each subdivided structure's name, preventing naming conflicts. Second, it restructures the validation hierarchy by moving the `_validate_finalized` call to the base `AbstractYeeGridSimulation.validate_pre_upload` method and ensuring EME simulations properly call the parent validation through `super().validate_pre_upload()`.

These changes ensure that all simulation types (FDTD, EME, etc.) that inherit from `AbstractYeeGridSimulation` consistently handle 2D material subdivision validation, maintaining the integrity of the simulation setup process while preventing solver errors.

## Important Files Changed

<details><summary>Changed Files</summary>

| Filename | Score | Overview |
|----------|-------|----------|
| `CHANGELOG.md` | 5/5 | Added changelog entry documenting the fix for named 2D materials with inhomogeneous substrates |
| `tidy3d/components/simulation.py` | 4/5 | Fixed subdivision naming conflicts by adding unique suffixes and moved validation to base class |
| `tests/test_components/test_simulation.py` | 4/5 | Enhanced test coverage by adding named structure and ensuring finalization triggers validation |
| `tidy3d/components/eme/simulation.py` | 5/5 | Added missing `super().validate_pre_upload()` call to inherit critical parent class validations |

</details>

## Confidence score: 4/5

- This PR is safe to merge with low risk as it addresses a specific bug with a targeted fix
- Score reflects solid implementation and thorough testing, though the subdivision logic is moderately complex
[...]

PR #2862, opened 2025-10-02
FXC-3354-Lint-commit-messages

mahlau-flex:
Added commitlint to pre-commit pipeline and github wokflow

<h2>Greptile Overview</h2>

Updated On: 2025-10-02 06:59:20 UTC

<h3>Summary</h3>
This PR implements commit message linting across the Tidy3D project by adding commitlint configuration and integrating it into both the pre-commit pipeline and GitHub Actions workflow. The changes introduce standardized commit message formatting based on conventional commit standards.

The implementation adds three key components:
1. **Configuration file** (`.commitlint.yml`) that defines strict rules for commit messages including mandatory commit types (feat, fix, docs, style, refactor, perf, test, build, ci, chore, revert), length constraints (12-72 characters for headers), and lowercase scoping requirements
2. **Pre-commit integration** that validates commit messages locally during the commit process using the alessandrojcm/commitlint-pre-commit-hook
3. **GitHub Actions job** that validates all commit messages in PRs or single commits in pushes using a Go-based commitlint tool

The enforcement follows a gradual rollout strategy - currently configured in informational mode to show warnings but not block merges, with plans to make it mandatory in the future. This approach allows developers to adapt to the new requirements without disrupting existing workflows. The GitHub Actions implementation intelligently handles different event types, checking all commits in a PR range versus just the head commit for pushes.

This change fits into the broader Tidy3D codebase quality infrastructure, complementing existing pre-commit hooks and automated testing pipelines. It establishes a foundation for improved commit history readability, automated changelog generation, and better project maintainability.

**PR Description Notes:**
- Minor typo: "wokflow" should be "workflow"

## Important Files Changed

<details>

| `.commitlint.yml` | 5/5 | New configuration file defining conventional commit rules with appropriate type enforcement and length constraints |
| `.pre-commit-config.yaml` | 4/5 | Added commitlint hook integration to validate commit messages during local development |
| `.github/workflows/tidy3d-python-client-tests.yml` | 4/5 | Added GitHub Actions job to validate commit messages in CI pipeline with proper PR/push handling |

</details>

## Confidence score: 4/5

- This PR is safe to merge with minimal risk as it introduces commit message validation without breaking existing functionality
- Score reflects well-structured implementation with proper gradual rollout strategy and comprehensive integration across development workflows
- Pay close attention to the GitHub Actions workflow implementation to ensure the commit message extraction logic works correctly for all event types

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Developer
    participant Git
    participant PreCommit as "Pre-commit Hooks"
    participant CommitLint as "CommitLint"
    participant GitHub as "GitHub Actions"
    participant RuffLinter as "Ruff Linter"
    participant TestRunner as "Test Runner"

    Developer->>Git: "git commit -m 'feat: add new feature'"
    Git->>PreCommit: "Trigger pre-commit hooks"
    PreCommit->>RuffLinter: "Run ruff-check and ruff-format"
    RuffLinter-->>PreCommit: "Code formatting results"
    PreCommit->>CommitLint: "Validate commit message format"
    CommitLint-->>PreCommit: "Commit message validation result"
    PreCommit-->>Git: "Pre-commit validation complete"
    Git-->>Developer: "Commit accepted/rejected"

    Developer->>GitHub: "Push commits to PR branch"
    GitHub->>GitHub: "Trigger tidy3d-python-client-tests workflow"
    GitHub->>GitHub: "Determine test scope based on PR status"
    
    par Linting Jobs
        GitHub->>RuffLinter: "Run ruff format --check --diff"
        RuffLinter-->>GitHub: "Format check results"
        GitHub->>RuffLinter: "Run ruff check tidy3d"
        RuffLinter-->>GitHub: "Lint check results"
    and Commit Message Linting
        GitHub->>CommitLint: "Check all PR commit messages"
        CommitLint-->>GitHub: "Commit message validation results"
    and Schema Verification
        GitHub->>GitHub: "Verify schema changes"
        GitHub-->>GitHub: "Schema validation results"
    end

    GitHub->>TestRunner: "Run local tests (Python 3.10, 3.13)"
    TestRunner-->>GitHub: "Local test results"

    alt PR is approved
        GitHub->>TestRunner: "Run remote tests (multiple platforms)"
        TestRunner-->>GitHub: "Remote test results"
    end

    GitHub->>GitHub: "Check pr-requirements-pass"
    GitHub-->>Developer: "PR status update"
```

### DIFF
--- a/.github/workflows/tidy3d-python-client-tests.yml
+++ b/.github/workflows/tidy3d-python-client-tests.yml
@@ -159,6 +159,41 @@ jobs:
         run: ruff format --check --diff
       - name: Run ruff check
         run: ruff check tidy3d
+  
+  lint-commit-messages:
+    needs: determine-test-scope
+    runs-on: ubuntu-latest
+    name: veryify-commit-linting
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # fetch all commits in the PR
+
+      - name: Setup node
+        uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+
+      - name: Install commitlint
+        run: npm install -D @commitlint/cli @commitlint/config-conventional
+
+      - name: Print versions
+        run: |
+          git --version
+          node --version
+          npm --version
+          npx commitlint --version
+        
+      - name: Check commit messages (merge_group)
+        if: github.event_name == 'merge_group'
+        run: |
+          # For merge groups, check the commits being merged
+          npx commitlint --from ${{ github.event.merge_group.base_sha }} --to ${{ github.event.merge_group.head_sha }} --verbose || {
+            echo "Commit message linting failed; please follow the conventional commits format at https://www.conventionalcommits.org/"
+            exit 1
+          }
+      
 
   verify-schema-change:
     name: verify-schema-change
@@ -480,7 +515,7 @@ jobs:
       (( needs.determine-test-scope.outputs.pr_approval_state == 'true' ) && 
       ( needs.determine-test-scope.outputs.local_tests == 'true' ) || 
       ( needs.determine-test-scope.outputs.remote_tests == 'true' ))
-    needs: [local-tests, remote-tests, lint, verify-schema-change]
+    needs: [local-tests, remote-tests, lint, verify-schema-change, lint-commit-messages]
     runs-on: ubuntu-latest
     steps:
       - name: check-passing-remote-tests
@@ -501,6 +536,9 @@ jobs:
             echo "❌ local-tests failed or were skipped."
           elif [[ "${{ needs.determine-test-scope.outputs.pr_approval_state }}" == 'true' && github.event.pull_request ]]; then
             echo "❌ PR requires approval."
+            exit 1
+          elif [[ github.event_name == 'merge_group' && "${{ needs.lint-commit-messages.result }}" != 'success' ]]; then
+            echo "❌ Linting of commit messages failed or was skipped."
             exit 1
           fi
           echo "✅ All required test jobs passed!"
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,0 +1,23 @@
+{
+  "rules": {
+    "header-min-length": [2, "always", 12],
+    "header-max-length": [2, "always", 72],
+    "footer-max-line-length": [2, "always", 72],
+    "type-empty": [2, "never"],
+    "type-enum": [2, "always", [
+      "feat",
+      "fix",
+      "docs",
+      "style",
+      "refactor",
+      "perf",
+      "test",
+      "build",
+      "ci",
+      "chore",
+      "revert"
+    ]],
+    "body-max-line-length": [2, "always", 72],
+    "scope-case": [2, "always", "lower-case"]
+  }
+}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,6 @@
+default_install_hook_types:
+  - pre-commit
+  - commit-msg
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: "v0.11.11"
@@ -5,4 +8,11 @@ repos:
      - id: ruff-check
        args: [ --fix ]
      - id: ruff-format
+  - repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
+    rev: v9.22.0
+    hooks:
+      - id: commitlint
+        stages: [commit-msg]
+        verbose: true
+        entry: bash -c 'commitlint --edit || exit 0'
 


frederikschubertflex:
I like this!  👍  However, if we always to squash and merge and are using the merge queue, couldn't we enforce the commit message linting only for the commit that lands in main (develop)? 
Or is this already how this is supposed to work?

yaugenst-flex:
> I like this! 👍 However, if we always to squash and merge and are using the merge queue, couldn't we enforce the commit message linting only for the commit that lands in main (develop)? Or is this already how this is supposed to work?

Great point. This should probably only be enforced once we queue for merge? And then for all commits (in case there is more than one, which does happen from time to time). Nice side-effect is that this would probably also catch cases where we forget to squash.

mahlau-flex:
What exactly do you mean that I should change then? Is there an option to run the commitlint-Pipeline only if the PR is merged? I feel like it would be better to immediately have the feedback as soon as the PR is opened rather than on merge.

mahlau-flex:
Currently, the github pipeline is not working at all, I will see how to fix this

mahlau-flex:
I added a check, which fails the PR if there is more than a single commit present. Is there any case where we might want to have more than a single commit in the PR? If this check is too strict, we can also remove it again. @yaugenst-flex could you give feedback here?

yaugenst-flex:
> I added a check, which fails the PR if there is more than a single commit present. Is there any case where we might want to have more than a single commit in the PR? If this check is too strict, we can also remove it again. @yaugenst-flex could you give feedback here?

So I think we should allow for more than one commit per PR, because sometimes it does make sense to split up to changes within one PR into two logical commits. That should not be the norm, but it does happen. Sometimes however we just forget to squash, and in those cases typically the individual commits will also not follow commitlint rules (might be commits added after review comments for example). And if we enforce commitlint on all commits, then we would catch these cases automatically.

I also think we might not want to enforce commitlint rules on PR open but instead only when queuing up for merge? Since that‘s really the only place where it matters.

daquinteroflex:
So yes, by default the merge queue will rebase the PR commits, and I guess we want to check every commit in the rebase, that said it can be quite annoying to have this check other than in the `merge_group` event, so it'd be good for this to be triggered on those only.

.github/workflows/tidy3d-python-client-tests.yml
```diff
@@ -159,6 +159,62 @@ jobs:
...
+            exit 1
+          }
+
+      - name: Check commit message (push)
+        if: github.event_name == 'push'
+        run: |
+          echo "${{ github.event.head_commit.message }}" | npx commitlint --verbose || {
+            echo "Commit message linting failed; please follow the conventional commits format at https://www.conventionalcommits.org/"
+            exit 1
+          }
```
daquinteroflex, line 195: Curious if something like a `pre-commit run` won't catch this for N amount of commits in a PR? Probably not just thought to check if that way people get a similar behaviour locally too.

mahlau-flex: I added a pre-commit config in the PR, which checks the name of every commit directly as soon as it is done. ```pre-commit run``` cannot catch this, because commit messages are only checked in the commit-msg stage. During ```pre-commit run```, the name of the commit message is not known yet. So with the current config you will get the error as soon as you put ```git commit -m "bad message"```.

mahlau-flex: though for this to work, you will need to install  ``pre-commit install --hook-type commit-msg `` , see the change in the docs.

mahlau-flex:
> So yes, by default the merge queue will rebase the PR commits, and I guess we want to check every commit in the rebase, that said it can be quite annoying to have this check other than in the `merge_group` event, so it'd be good for this to be triggered on those only.

I changed the github pipeline to only run on merge-group events

mahlau-flex:
Does it make sense to change the local pre-commit config now as well? I am not sure how this should be configured now

.github/workflows/tidy3d-python-client-tests.yml
daquinteroflex: Yeah guess I was curious if there was an equivalent check of `pre-commit run --all-files` (which we sometimes use locally when the ruff pre-commit has not been initialized previously) for git previous commit metadata but sounds like there's not anyways for reproducing the exact CI check behaviour locally. Well sounds good then.

There are options of writing custom scripts into the pre-commit which can check N number of commits locally but probably not worth it as also may require installing commitlint rather than it be packaged like you have already.

No worries though! Just thought to check

.github/workflows/tidy3d-python-client-tests.yml
```diff
@@ -159,6 +159,57 @@ jobs:
         run: ruff format --check --diff
       - name: Run ruff check
         run: ruff check tidy3d
+  
+  lint-commit-messages:
+    needs: determine-test-scope
+    if: |
+      github.event_name == 'merge_group' &&
+      (( needs.determine-test-scope.outputs.local_tests == 'true' ) || ( needs.determine-test-scope.outputs.remote_tests == 'true' ))
```
daquinteroflex: This line is not needed really? Since we always want to run in a merge group and the logic is already fulfilled https://github.com/flexcompute/tidy3d/blob/6f783bdde07a9094ee4953f65ace0537f704674f/.github/workflows/tidy3d-python-client-tests.yml#L117-L120

docs/development/introduction/index.rst
```diff
@@ -7,7 +7,7 @@ This page hopefully will get you started to develop Tidy3D.
 
 - Branch off of the target branch (usually ``develop`` or ``pre/x.x``), work on your branch, and submit a PR when ready.
 - Use isolated development environments with ``poetry``.
-- Use ``ruff`` to lint and format code, and install the pre-commit hook via ``pre-commit install`` to automate this.
+- Use ``ruff`` to lint and format code, and install the pre-commit hook via ``pre-commit install --hook-type commit-msg `` to automate this. The commit-msg hook is necessary, because the pre-commit pipeline also checks the formatting of commit messages.
```
daquinteroflex: Hmm I see, I tested locally. So many people will forget to do this I suspect since several sometimes don't use the pre-commit hooks. The problem as well is that they will not realize this is a problem until they try to merge and the merge_queue blocks them out which will be annoying. That said, this is just part of adding a new standard to the codebase and cannot do more than inform them via slack that they need to add this new pre-commit hook, and well if it blocks their merges it should be evident why.

Guess it's fine in the interim to have a bit of pain.

daquinteroflex:
Looks broadly good! Not sure if any further comments before moving forward?

daquinteroflex:
> Does it make sense to change the local pre-commit config now as well? I am not sure how this should be configured now

```
default_install_hook_types:
  - pre-commit
  - commit-msg
```

https://pre-commit.com/#top_level-default_install_hook_types

Maybe means `pre-commit install` will work automatically for people like before`

.github/workflows/tidy3d-python-client-tests.yml
mahlau-flex: True, I will change that

.github/workflows/tidy3d-python-client-tests.yml
```diff
@@ -502,5 +553,8 @@ jobs:
           elif [[ "${{ needs.determine-test-scope.outputs.pr_approval_state }}" == 'true' && github.event.pull_request ]]; then
             echo "❌ PR requires approval."
             exit 1
+          elif [[ "${{ needs.lint-commit-messages.result }}" != 'success' ]]; then
```
daquinteroflex: So I've realised you also need to make this line run only on a merge group check otherwise it will block PR checks from being added to the merge queue, can probably add an `needs.determine-test-scope.outputs.merge_tests` flag too if cleaner?

mahlau-flex:
> > Does it make sense to change the local pre-commit config now as well? I am not sure how this should be configured now
> 
> ```
> default_install_hook_types:
>   - pre-commit
>   - commit-msg
> ```
> 
> https://pre-commit.com/#top_level-default_install_hook_types
> 
> Maybe means `pre-commit install` will work automatically for people like before`

Good suggestions, this works locally for me

.github/workflows/tidy3d-python-client-tests.yml
mahlau-flex: I have added a check directly in this line now. In my opinion, this is cleaner, but if you want to see another approach please tell me

mahlau-flex:
I have changed the pre-commit pipeline, such that the user will get a warning if the commit message is bad, but the pre-commit pipeline does not fail. With this, we can make bad commit messages (temporarily) and squash them together later with a good commit message.

.github/workflows/tidy3d-python-client-tests.yml
daquinteroflex: Yep looks fine! In the future we could have other `merge_group` only checks and we can introduce  `needs.determine-test-scope.outputs.merge_tests ` then  but for now this is good, thanks!

.github/workflows/tidy3d-python-client-tests.yml
```diff
@@ -159,6 +159,55 @@ jobs:
...
+        run: |
+          COMMIT_COUNT=$(git rev-list --count ${{github.event.pull_request.base.sha}}..${{github.event.pull_request.head.sha}})
+          echo "Number of commits in PR: $COMMIT_COUNT"
+          if [ "$COMMIT_COUNT" -ne 1 ]; then
+            echo "❌ ERROR: This PR contains $COMMIT_COUNT commits, but only 1 commit is allowed."
+            echo "Please squash your commits into a single commit before merging."
+            exit 1
+          else
+            echo "✅ PR contains exactly 1 commit"
+          fi
```
daquinteroflex: So this still needs to change for multiple commits support right?

mahlau-flex: I am not sure, would it be good to have this check here as an automatic check that during merge the commits should be squashed into a single commit? Otherwise I can also remove this

yaugenst-flex: No we do _not_ want to force a single commit, see https://github.com/flexcompute/tidy3d/pull/2862#issuecomment-3359760675

.github/workflows/tidy3d-python-client-tests.yml
```diff
@@ -502,5 +551,8 @@ jobs:
           elif [[ "${{ needs.determine-test-scope.outputs.pr_approval_state }}" == 'true' && github.event.pull_request ]]; then
             echo "❌ PR requires approval."
             exit 1
+          elif [[github.event_name == 'merge_group' && "${{ needs.lint-commit-messages.result }}" != 'success' ]]; then
```
yaugenst-flex: There is a space missing here: `[[github...` --> `[[ github...` otherwise this will be evaluated as a missing command I think

.github/workflows/tidy3d-python-client-tests.yml
mahlau-flex: I have removed this check now

yaugenst-flex (changes requested):
So if I read this correctly, then `pr-requirements-pass` now depends on `lint-commit-messages`, but that depends on `if: github_event_name == 'merge_group'`. This will be skipped on every PR event, which means `pr-requirements-pass` will never run (correct my understanding if wrong, not a github actions expert...). I think the fix would be to drop the job-level `if` so that the required dependency is always satisfied and then make the actual commitlint work conditional inside the job (`if: github.event_name == 'merge_group'`).

.github/workflows/tidy3d-python-client-tests.yml
mahlau-flex: changed both in newest version

daquinteroflex:
@yaugenst-flex isn't it because of `always()` in 
https://github.com/flexcompute/tidy3d/blob/ea80d7c9bf52a2e388846e244896589ee7acd398/.github/workflows/tidy3d-python-client-tests.yml#L513-L517

this overrides the` needs.lint-commit-messages` skip event for `pr-requirements-pass`? It was intended to override the needs logic intentionally

I've used it in some other actions and works well, and should cover this situation I think

yaugenst-flex:
> @yaugenst-flex isn't it because of `always()` in
> 
> https://github.com/flexcompute/tidy3d/blob/ea80d7c9bf52a2e388846e244896589ee7acd398/.github/workflows/tidy3d-python-client-tests.yml#L513-L517
> 
> this overrides the` needs.lint-commit-messages` skip event for `pr-requirements-pass`? It was intended to override the needs logic intentionally
> 
> I've used it in some other actions and works well, and should cover this situation I think

I'm not sure, but we can try it. My understanding is that `always()` works well in cases where the job actually runs but fails. But here, `lint-commit-messages` would be a dependency that never runs (https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution#if-failure).

yaugenst-flex:
Ah but I guess we could of course let the commitlint run and fail, in which case your guard would catch it. Seems a bit wasteful though...

daquinteroflex:
So if we add the `if github.event == merge_group` here https://github.com/flexcompute/tidy3d/blob/ea80d7c9bf52a2e388846e244896589ee7acd398/.github/workflows/tidy3d-python-client-tests.yml#L163-L166

then it would just skip that job, and and also then be skipped by a similar if statement in the `pr-requirements-pass` and that'd be fine?

 Ideally though some flag like `needs.determine-workflow-scope.merge_group` can be helpful for managing that state though but as long as it's the same logic that's ok too

yaugenst-flex:
No in what i’m proposing, the job would pass (by skipping its steps), and only truly run when going in merge queue.

daquinteroflex:
So I think both approaches get ultimately the same PR-facing behaviour - happy to try them out! My personal preference is skipping it at the job level like we do other jobs, and simply skipping the check in a merge_group. Ultimately though, I think we're both intending to skip it until the check is meant to run from what I understand you're proposing, it's just where we skip

yaugenst-flex:
Yes let’s just try. What’s the worst that could happen 😄

mahlau-flex:
@yaugenst-flex @frederikschubertflex Do you want to see any other changes or this it PR ready to merge? If so please let me know.

frederikschubertflex (approved):

yaugenst-flex (approved):

yaugenst-flex:
> So I think both approaches get ultimately the same PR-facing behaviour - happy to try them out! My personal preference is skipping it at the job level like we do other jobs, and simply skipping the check in a merge_group. Ultimately though, I think we're both intending to skip it until the check is meant to run from what I understand you're proposing, it's just where we skip

So this is what I mean https://github.com/flexcompute/tidy3d/pull/2854

PR #2864, opened 2025-10-03
feat(FXC-3536): add log and symlog plot scale

damiano-flex:
<h2>Greptile Overview</h2>

Updated On: 2025-10-03 10:54:12 UTC

<h3>Summary</h3>
This PR adds logarithmic and symmetric logarithmic scaling support to Tidy3D's plotting functionality. The changes enhance the visualization capabilities by expanding the available scale options from just linear ('lin') and decibel ('dB') to include logarithmic ('log') and symmetric logarithmic ('symlog') scales.

The implementation involves two main components: First, the `PlotScale` type definition in `tidy3d/components/types/base.py` is updated to include the new scale options. Second, the plotting infrastructure in `tidy3d/components/scene.py` is modified to handle these new scales using matplotlib's normalization objects (`LogNorm` and `SymLogNorm`).

The core plotting methods like `plot_structures_eps` and `plot_structures_property` now accept a `scale` parameter and use appropriate matplotlib normalizations instead of simple `vmin`/`vmax` parameters. For symmetric logarithmic scaling, the implementation includes sophisticated logic to calculate appropriate linear threshold (`linthresh`) values, particularly for doping concentration data where values can span many orders of magnitude and include both positive and negative values.

This enhancement integrates seamlessly with Tidy3D's existing plotting architecture and follows matplotlib's standard conventions. The change is particularly valuable for electromagnetic simulations where field data, power quantities, and material properties often span wide dynamic ranges that benefit from logarithmic visualization.

<details>

| tidy3d/components/types/base.py | 5/5 | Added 'log' and 'symlog' options to PlotScale type definition |
| tidy3d/components/scene.py | 4/5 | Implemented log/symlog scaling in plotting methods with normalization logic and linthresh calculation |

</details>

- This PR is safe to merge with minimal risk of breaking existing functionality
- Score reflects well-structured implementation following matplotlib conventions, though the complex linthresh calculation logic needs attention
- Pay close attention to the symlog threshold calculation methods in scene.py

### DIFF
--- a/tidy3d/components/scene.py
+++ b/tidy3d/components/scene.py
@@ -59,6 +59,7 @@ from .types import (
     InterpMethod,
     LengthUnit,
     PermittivityComponent,
+    PlotScale,
     PriorityMode,
     Shapely,
     Size,
@@ -632,11 +633,19 @@ class Scene(Tidy3dBaseModel):
         return plot_params
 
     @staticmethod
-    def _add_cbar(vmin: float, vmax: float, label: str, cmap: str, ax: Ax = None) -> None:
+    def _add_cbar(
+        vmin: float,
+        vmax: float,
+        label: str,
+        cmap: str,
+        ax: Ax = None,
+        norm: mpl.colors.Normalize = None,
+    ) -> None:
         """Add a colorbar to plot."""
-        norm = mpl.colors.Normalize(vmin=vmin, vmax=vmax)
         divider = make_axes_locatable(ax)
         cax = divider.append_axes("right", size="5%", pad=0.15)
+        if norm is None:
+            norm = mpl.colors.Normalize(vmin=vmin, vmax=vmax)
         mappable = mpl.cm.ScalarMappable(norm=norm, cmap=cmap)
         plt.colorbar(mappable, cax=cax, label=label)
 
@@ -804,6 +813,8 @@ class Scene(Tidy3dBaseModel):
         ax: Ax = None,
         hlim: Optional[tuple[float, float]] = None,
         vlim: Optional[tuple[float, float]] = None,
+        eps_lim: tuple[Union[float, None], Union[float, None]] = (None, None),
+        scale: PlotScale = "lin",
     ) -> Ax:
         """Plot each of scene's components on a plane defined by one nonzero x,y,z coordinate.
         The permittivity is plotted in grayscale based on its value at the specified frequency.
@@ -828,6 +839,10 @@ class Scene(Tidy3dBaseModel):
             The x range if plotting on xy or xz planes, y range if plotting on yz plane.
         vlim : Tuple[float, float] = None
             The z range if plotting on xz or yz planes, y plane if plotting on xy plane.
+        eps_lim : Tuple[float, float] = None
+            Custom limits for eps coloring.
+        scale : PlotScale = "lin"
+            Scale for the plot. Either 'lin' for linear, 'log' for log10, 'symlog' for symmetric logarithmic (linear near zero, logarithmic elsewhere), or 'dB' for decibel scale.
 
         Returns
         -------
@@ -838,7 +853,17 @@ class Scene(Tidy3dBaseModel):
         hlim, vlim = Scene._get_plot_lims(bounds=self.bounds, x=x, y=y, z=z, hlim=hlim, vlim=vlim)
 
         ax = self.plot_structures_eps(
-            freq=freq, cbar=True, alpha=alpha, ax=ax, x=x, y=y, z=z, hlim=hlim, vlim=vlim
+            freq=freq,
+            cbar=True,
+            alpha=alpha,
+            ax=ax,
+            x=x,
+            y=y,
+            z=z,
+            hlim=hlim,
+            vlim=vlim,
+            eps_lim=eps_lim,
+            scale=scale,
         )
         ax = self._set_plot_bounds(bounds=self.bounds, ax=ax, x=x, y=y, z=z, hlim=hlim, vlim=vlim)
         return ax
@@ -855,6 +880,7 @@ class Scene(Tidy3dBaseModel):
         cbar: bool = True,
         reverse: bool = False,
         eps_lim: tuple[Union[float, None], Union[float, None]] = (None, None),
+        scale: PlotScale = "lin",
         ax: Ax = None,
         hlim: Optional[tuple[float, float]] = None,
         vlim: Optional[tuple[float, float]] = None,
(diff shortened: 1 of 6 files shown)


tidy3d/components/types/base.py
```diff
@@ -238,7 +238,7 @@ def __modify_schema__(cls, field_schema):
 PlotVal = Literal["real", "imag", "abs"]
 FieldVal = Literal["real", "imag", "abs", "abs^2", "phase"]
 RealFieldVal = Literal["real", "abs", "abs^2"]
-PlotScale = Literal["lin", "dB"]
+PlotScale = Literal["lin", "dB", "log", "symlog"]
```
momchil-flex, line 241: If you update this then you need to update all the methods that use it, specifically the ones for the various field plotting functions in SimulationData, etc. Because the type signature will now show that these methods allow these values.

The alternative is to introduce a separate type, or potentially to add a check if the scale is `lin` e.g. [here](https://github.com/flexcompute/tidy3d/blob/ef3d257581796280f6a4b031cec6a6ca8de512d9/tidy3d/components/data/sim_data.py#L543) (and other relevant places, if any), and then a final `else` that throws an error about the scale not being supported.

tidy3d/components/scene.py
```diff
@@ -1013,6 +1028,8 @@ def plot_structures_property(
                 structures=structures, x=x, y=y, z=z, hlim=hlim, vlim=vlim
             )
 
+        # default linthresh for symlog scale
+        linthresh = 10
```
momchil-flex: Seems like this default is only reset for doping plotting; should it be set to something like a fraction of the maximum absolute value for all properties?

tidy3d/components/types/base.py
damiano-flex: I did as you suggested, thanks!

tidy3d/components/scene.py
damiano-flex: I modified it so that the default value is not even set.

- If `property_min` is not specified, it will be 1e-2 times eps, or the value of the smallest doping property otherwise.
- If  `property min` is specified, then it will be equal to 1e-2 times the specified value (except if it is close or less than 0, the we pick 1e-3).

Doping might change a lot, so I cannot imagine a good way to set it as a fraction of the maximum.

tidy3d/components/scene.py
```diff
@@ -632,9 +633,15 @@ def _get_structure_plot_params(
...
+        label: str,
+        cmap: str,
+        ax: Ax = None,
+        norm: mpl.colors.Normalize = None,
+    ) -> None:
         """Add a colorbar to plot."""
-        norm = mpl.colors.Normalize(vmin=vmin, vmax=vmax)
         divider = make_axes_locatable(ax)
         cax = divider.append_axes("right", size="5%", pad=0.15)
         mappable = mpl.cm.ScalarMappable(norm=norm, cmap=cmap)
```
yaugenst-flex, line 649: Doesn't this change behavior? Previously a `norm = mpl.colors.Normalize(vmin=vmin, vmax=vmax)` was created, and passing `norm=None` to `mpl.cm.ScalarMappable` I think is not the same.

tidy3d/components/scene.py
```diff
@@ -1014,27 +1029,47 @@ def plot_structures_property(
...
+                linthresh = 1e-3
+            else:
+                linthresh = 1e-2 * np.abs(property_min)
+
+        if scale == "lin":
+            norm = mpl.colors.Normalize(vmin=property_min, vmax=property_max)
+        elif scale == "log":
+            norm = mpl.colors.LogNorm(vmin=property_min, vmax=property_max)
+        elif scale == "symlog":
+            norm = mpl.colors.SymLogNorm(linthresh=linthresh, vmin=property_min, vmax=property_max)
```
yaugenst-flex, line 1095: Can `property_min` be <0? In which case I think matplotlib would raise here.

tidy3d/components/scene.py
yaugenst-flex, line 1: I think polygons (ie non-custom media) are still shaded by `_get_structure_eps_plot_params` which still does a linear map between `eps_min` and `eps_max`, which means in such situations the actual face colors would be inconsistent with the colorbar. So for consistency the facecolor calculation needs to go through the same normalization.

yaugenst-flex (changes requested):

tidy3d/components/scene.py
damiano-flex: in symlog `property_min` can be less than 0, so no issues here.

yaugenst-flex: what about `LogNorm` though?

damiano-flex: For `LogNorm`, it raises the `SetupError` if `property_min <= 0 or property_max <= 0` (line 1086).

yaugenst-flex: oh ok, all good then!

damiano-flex:
@greptile Given the tests added in the new commit, check the PR another time.

marc-flex (approved):

tidy3d/components/scene.py
damiano-flex: In the current version of the function `_get_structure_eps_plot_params`, at line 1434, all the medium cases fall into the `else` at line 1448, where the correct `norm` is applied. 

Is it fixed now?

damiano-flex: @greptile Check if the issue pointed out by @yaugenst-flex has been solved.

damiano-flex: solved that issue with a fallback default value.

tidy3d/components/scene.py
```diff
@@ -1018,23 +1051,52 @@ def plot_structures_property(
...
-                    property_min = acceptor_limits[0]
-                    property_max = acceptor_limits[1]
+                    property_min = property_min if property_min else acceptor_limits[0]
+                    property_max = property_max if property_max else acceptor_limits[1]
+                    linthresh = acceptor_abs_min
                 elif property == "doping":
-                    property_min = -donor_limits[1]
-                    property_max = acceptor_limits[1]
+                    property_min = property_min if property_min else -donor_limits[1]
+                    property_max = property_min if property_max else acceptor_limits[1]
```
yaugenst-flex: Shouldn't this be `property_max = property_*max* if property_max`? Generally these checks should also use `x if x is None` instead of a generic "truthiness" check, since for example a limit of `0` will currently be treated the same as `None`, which I think wouldn't be correct behavior?

yaugenst-flex (commented):
Thanks @damiano-flex! Only one point left from my side. Sorry these excursions into the plotting module tend to get ... messy.

tidy3d/components/scene.py
damiano-flex: You're right. I solved it now.

damiano-flex: If things are fine I'll squash, rebase and merge. (the Changelog is updated).

yaugenst-flex (approved):
LGTM, thanks @damiano-flex!

PR #2867, opened 2025-10-06
Improve speed of Box derivative computation when in GeometryGroup

groberts-flex:
Connect box face derivatives to `evaluate_gradient_at_points` in `DerivativeInfo` so that interpolators can be shared when using inside a `GeometryGroup`.

<h2>Greptile Overview</h2>

Updated On: 2025-10-06 19:50:07 UTC

<h3>Summary</h3>
This PR optimizes the performance of Box geometry derivative computation when used inside a GeometryGroup by refactoring the implementation to leverage the existing `evaluate_gradient_at_points` infrastructure in `DerivativeInfo`. The key changes include:

- **Major refactoring of Box derivative computation**: The `_derivative_face` method in `tidy3d/components/geometry/base.py` was completely rewritten, removing 380+ lines of complex field integration code and replacing it with a simpler grid-based approach that builds spatial grids and evaluates gradients at specific points.

- **Performance optimization through interpolator sharing**: By connecting to the `evaluate_gradient_at_points` method, the new implementation allows interpolators to be shared when multiple Box geometries are processed together in a GeometryGroup, significantly improving computational efficiency.

- **Code simplification**: The refactoring eliminates the need for PEC-specific handling, complex epsilon data management, and manual interpolation logic, making the codebase more maintainable while leveraging existing gradient evaluation frameworks.

- **Comprehensive testing**: A new test file was added to validate that the optimized Box gradient implementation produces mathematically equivalent results to the established PolySlab approach across different dimensional configurations (2D and 3D cases).

- **Test cleanup**: Obsolete Box-specific test functions were removed since the functionality is now covered by the broader gradient evaluation system.

This change fits into the broader Tidy3D architecture by unifying Box derivative computation with the existing `DerivativeInfo` framework, enabling better resource sharing and performance optimization when working with multiple geometries in electromagnetic simulations.

<details>

| `CHANGELOG.md` | 5/5 | Added changelog entry documenting the Box derivative performance improvement |
| `tidy3d/components/geometry/base.py` | 4/5 | Major refactoring of Box `_derivative_face` method to use grid-based approach with `evaluate_gradient_at_points` |
| `tests/test_components/autograd/numerical/test_autograd_box_polyslab_numerical.py` | 5/5 | New comprehensive test validating Box vs PolySlab gradient equivalence across 2D/3D configurations |
| `tests/test_components/test_geometry.py` | 4/5 | Removed obsolete Box-specific test functions that are no longer needed after refactoring |

</details>

- This PR appears safe to merge with good performance benefits and comprehensive testing coverage
- Score reflects the substantial refactoring of core gradient computation logic which requires careful validation
- Pay close attention to the geometry base module changes and ensure the new grid-based approach handles all edge cases correctly

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Geometry
    participant DerivativeInfo
    participant Interpolators
    participant GradientEvaluator
    
    User->>Geometry: "_compute_derivatives(derivative_info)"
    Geometry->>DerivativeInfo: "create_interpolators(dtype=GRADIENT_DTYPE_FLOAT)"
    DerivativeInfo->>Interpolators: "create field data interpolators"
    Interpolators-->>DerivativeInfo: "interpolators"
    DerivativeInfo-->>Geometry: "interpolators"
    
    loop For each derivative path
        Geometry->>Geometry: "_derivative_faces(derivative_info)"
        
        loop For min/max faces and axes
            Geometry->>Geometry: "_derivative_face(min_max_index, axis_normal, derivative_info)"
            Geometry->>DerivativeInfo: "evaluate_gradient_at_points(spatial_coords, normals, perps1, perps2, interpolators)"
            DerivativeInfo->>GradientEvaluator: "interpolate fields at points"
            GradientEvaluator->>Interpolators: "get field values"
            Interpolators-->>GradientEvaluator: "field values"
            GradientEvaluator-->>DerivativeInfo: "gradient values"
            DerivativeInfo-->>Geometry: "gradient_at_points"
            Geometry->>Geometry: "sum(integration_weight * gradient_at_points)"
            Geometry-->>Geometry: "vjp_face"
        end
        
        Geometry->>Geometry: "_derivatives_center_size(vjps_faces)"
        Geometry-->>Geometry: "vjps_center_size"
    end
    
    Geometry-->>User: "derivative_map"
```


tidy3d/components/geometry/base.py
```diff
@@ -2465,380 +2465,133 @@ def _derivative_face(
...
-            log.info(
-                f"skipping VJP for 'Box' face '{dim_normal}{'-+'[min_max_index]}' "
-                "as it is entirely outside of the simulation domain."
-            )
-            return 0.0
 
-        # permittivity data
-        eps_xyz = [derivative_info.eps_data[f"eps_{dim}{dim}"] for dim in "xyz"]
+        if min_max_index == 0:
+            if coord_normal_face <= derivative_info.simulation_bounds[0][axis_normal]:
```
yaugenst-flex: I remember we discussed this at some point but I need a refresher 😆 So now we zero gradients for faces that sit _at_ the simulation boundaries, right? It's an edge case, but I feel like these should be strict comparisons?

tidy3d/components/geometry/base.py
```diff
@@ -2465,380 +2465,133 @@ def _derivative_face(
...
+            # build 1D grid for sampling points along the face, which is an edge in the 2D case
+            zero_dim = np.where(is_2d_map)[0][0]
+            # zero dim is one of the perpendicular directions, so the other perpendicular direction
+            # is the nonzero dimension
+            nonzero_dim = 1 - zero_dim
+
+            # clip at simulation bounds for integration dimension
+            integration_bounds_perp = (
+                np.maximum(bounds_perp[nonzero_dim][0], sim_bounds_perp[nonzero_dim][0]),
+                np.minimum(bounds_perp[nonzero_dim][1], sim_bounds_perp[nonzero_dim][1]),
```
yaugenst-flex: It might be good to check here if the resulting interval is valid. If a box lies completely outside of the domain I think we'd end up with `lower > upper` here? And then `spacing_to_grid_points` will create invalid data for the interpolators I think.

yaugenst-flex (commented):
`+99 / -346` love to see it 😄
Caught some minor issues but overall looks great, thanks @groberts-flex

tidy3d/components/geometry/base.py
groberts-flex: good point! yeah, if we say that faces that sit at the simulation boundaries can be nonzero, then these should be strict! I'll change those

tidy3d/components/geometry/base.py
groberts-flex: good point!

tidy3d/components/geometry/base.py
```diff
@@ -2465,380 +2465,153 @@ def _derivative_face(
...
-        )
-
-        integration_dims = dims_perp
-        integration_bounds = bounds_perp
-        if do_singularity_correction:
-            integration_dims, integration_bounds = self._trim_dims_and_bounds_edge(
-                dims_perp, bounds_perp, zero_dimension
+            # set up grid points to pass into evaluate_gradient_at_points
+            grid_points[:, axis_perp[nonzero_dim]] = grid_points_linear
+            grid_points[:, axis_perp[zero_dim]] = bounds_perp[0][zero_dim]
```
yaugenst-flex: Just noticed this, I think here since you always take the first perpendicular axis, this would break if the zero-extent axis is the second entry, right? I think the fix would just be `grid_points[:, axis_perp[zero_dim]] = bounds_perp[zero_dim][0]`

yaugenst-flex (commented):
One last thing I spotted but then it's ready to go I think.

groberts-flex:
made the changes for handling cases near the simulation bounds in Box and added more numerical test variants to check those cases. In the process, found those gradients were not lining up between polyslab and box and was able to trace down to a couple of fixes in the polyslab gradient computation

yaugenst-flex (approved):
Awesome thanks @groberts-flex

yaugenst-flex (approved):

Issue #1391, opened 2024-01-23
Native methods for filtering mode data


tomflexcompute:
A user reported that they spend a lot of time on either filtering modes manually or implementing their own filters. What could help them:
1.	a flag to automatically filter out modes that are marked as "not decayed at the boundary".
2.	filter by the loss, i.e. provide a threshold for how much of loss is allowed and remove all modes above that.
3.	filter by mode volume: similarly, provide a threshold value for mode volume.

And thanks @dbochkov-flexcompute for collecting the information from the user.

momchil-flex:
From discussion, few possible ways:
1. Define frontend post-processing methods that can be e.g. `ModeSolverData.filter(filter_what) -> FilteredModeSolverData`, where `FilteredModeSolverData` also keeps the indexing into the original mode solver data so that a user can e.g. select a mode for injection in a `ModeSource`. Pros: easier to implement and use directly, including different filtering in post-processing. Cons: a bit more confusing how to find modes in the original data; harder to incorporate in the GUI post-run viz.
2. Extend the filtering definition in `ModeSpec` (currently only `filter_pol`, convert to a `filter` dictionary or a `FilterSpec` that defines various filtering options) such that the produced modes are directly filtered as requested. Pros: better-defined, smaller chance to mistake definitions in a mode source or mode monitor, easier to incorporate in GUI. Cons: less flexible, requires a new mode solver for every filtering spec.
3. A combination of the two? But it can make things confusing.

I'm kinda leaning towards option 2. still - @dbochkov-flexcompute @tomflexcompute @xin-flex what do you think?

dbochkov-flexcompute:
yeah, I agree, I think option 2 is good. Are you thinking of filtering as elimination of modes (threshold for loss, threshold for volume, etc) or reordering of them (order by loss, order by volume, put non-decayed modes to the end, etc)?

momchil-flex:
> yeah, I agree, I think option 2 is good. Are you thinking of filtering as elimination of modes (threshold for loss, threshold for volume, etc) or reordering of them (order by loss, order by volume, put non-decayed modes to the end, etc)?

[...]

PR #2869, opened 2025-10-07
FXC-3547-Adding mode sorting to ModeSpec

momchil-flex:
Addresses the new suggestion in #1391 to add more advanced ways to sort modes in the mode solver (and mode sources and monitors).

Probably we should still keep the issue open as I should also implement the FilterSpec as discussed in that issue.

@yuanshen-flexcompute does the current PR help with some of your use cases that made you bring it up though?

<h2>Greptile Overview</h2>

Updated On: 2025-10-07 08:34:20 UTC

<h3>Summary</h3>
This PR adds advanced mode sorting functionality to the Tidy3D mode solver by introducing a new `ModeSortSpec` class and integrating it into the `ModeSpec` configuration. The implementation addresses user feedback from issue #1391 where users were spending significant time manually filtering and sorting modes after simulation.

The core addition is the `ModeSortSpec` class which allows users to sort modes by various physical properties including effective index (`n_eff`), loss coefficient (`k_eff`), TE polarization fraction (`TE_fraction`), waveguide TE/TM fractions (`wg_TE_fraction`, `wg_TM_fraction`), and mode area (`mode_area`). Users can specify sorting direction (ascending/descending) and optionally provide a reference value to sort by proximity to that value.

The implementation integrates seamlessly into the existing mode solver pipeline through a new `_sort_modes()` method that operates on `ModeSolverData` after polarization filtering but before frequency tracking. The sorting is applied per-frequency to handle cases where mode properties vary across the frequency spectrum. A mutual exclusivity validator ensures that the new `sort_spec` cannot be used simultaneously with the legacy `filter_pol` field to prevent conflicting behaviors.

To support the sorting functionality, three new property aliases were added to the `ModeData` class: `TE_fraction`, `wg_TE_fraction`, and `wg_TM_fraction`. These provide more convenient access to polarization fraction data and are used by the `modes_info` method for consistency. The `ModeSortSpec` class is properly exported through the main `__init__.py` file to make it available in the public API.

| tidy3d/components/mode_spec.py | 4/5 | Adds new `ModeSortSpec` class and `sort_spec` field to `ModeSpec` with mutual exclusivity validation |
| tidy3d/components/mode/mode_solver.py | 4/5 | Implements core sorting logic with `_sort_modes()` method and helper functions for mode reordering |
| tidy3d/components/data/monitor_data.py | 5/5 | Adds convenience property aliases for polarization fractions used by sorting functionality |
| tidy3d/__init__.py | 5/5 | Exports new `ModeSortSpec` class to public API |
| tests/test_plugins/test_mode_solver.py | 4/5 | Adds comprehensive tests for mode sorting functionality and mutual exclusivity validation |

</details>

- This PR is generally safe to merge with well-structured implementation and comprehensive testing
- Score reflects the complexity of the mode manipulation logic and integration with existing pipeline components
- Pay close attention to the mode solver implementation file and validator logic in the mode_spec file

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant ModeSpec
    participant ModeSolver
    participant ModeSolverData
    participant sort_methods as _sort_modes & _apply_mode_reorder

    User->>ModeSpec: "Create ModeSpec with sort_spec"
    ModeSpec->>ModeSpec: "Validate sort_spec parameters"
    Note over ModeSpec: Check key is in SORT_KEYS<br/>Ensure filter_pol and sort_spec not both used

    User->>ModeSolver: "Create ModeSolver with ModeSpec"
    ModeSolver->>ModeSolver: "Initialize and validate"

    User->>ModeSolver: "Call solve() or data_raw"
    ModeSolver->>ModeSolver: "_data_on_yee_grid()"
    ModeSolver->>ModeSolver: "_colocate_data() if needed"
    ModeSolver->>ModeSolver: "_normalize_modes()"
    ModeSolver->>ModeSolver: "_filter_polarization()"
    ModeSolver->>sort_methods: "_sort_modes(mode_solver_data)"
    
    sort_methods->>ModeSolverData: "getattr(mode_solver_data, sort_spec.key)"
    Note over sort_methods: Get metric values (n_eff, k_eff, etc.)
    
    sort_methods->>sort_methods: "Apply reference_value if provided"
    Note over sort_methods: metric = abs(metric - reference_value)
    
    sort_methods->>sort_methods: "Loop over frequencies"
    loop For each frequency
        sort_methods->>sort_methods: "np.argsort(metric values)"
        sort_methods->>sort_methods: "Reverse if direction='descending'"
        sort_methods->>sort_methods: "_apply_mode_reorder(data, freq_idx, sort_inds)"
        sort_methods->>ModeSolverData: "Reorder field components and n_complex"
        Note over ModeSolverData: data.values[..., freq_idx, :] = data.values[..., freq_idx, sort_inds]
    end
    
    alt If track_freq is set and multiple frequencies
        ModeSolver->>ModeSolverData: "mode_solver_data.overlap_sort(track_freq)"
        Note over ModeSolverData: Apply frequency-based mode tracking
    end

    ModeSolver-->>User: "Return sorted ModeSolverData"
```

### DIFF
--- a/tests/test_plugins/test_mode_solver.py
+++ b/tests/test_plugins/test_mode_solver.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from typing import get_args
+
 import matplotlib.pyplot as plt
 import numpy as np
 import pydantic.v1 as pydantic
@@ -12,6 +14,7 @@ from tidy3d import ScalarFieldDataArray
 from tidy3d.components.data.monitor_data import ModeSolverData
 from tidy3d.components.mode.derivatives import create_sfactor_b, create_sfactor_f
 from tidy3d.components.mode.solver import compute_modes
+from tidy3d.components.mode_spec import MODE_DATA_KEYS
 from tidy3d.exceptions import DataError, SetupError
 from tidy3d.plugins.mode import ModeSolver
 from tidy3d.plugins.mode.mode_solver import MODE_MONITOR_NAME
@@ -689,7 +692,7 @@ def test_mode_solver_angle_bend():
         bend_axis=0,
         angle_theta=np.pi / 3,
         angle_phi=np.pi,
-        track_freq="highest",
+        sort_spec=td.ModeSortSpec(track_freq="highest"),
     )
     # put plane entirely in the symmetry quadrant rather than sitting on its center
     plane = td.Box(center=(0, 0.5, 0), size=(1, 0, 1))
@@ -850,7 +853,7 @@ def test_group_index(mock_remote_api, local, tmp_path):
         num_modes=2,
         target_neff=3.0,
         precision="double" if local else "single",
-        track_freq="central",
+        sort_spec=td.ModeSortSpec(track_freq="central"),
     )
 
     if local:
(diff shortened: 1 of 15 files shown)


momchil-flex:
Actually @caseyflex @yuanshen-flexcompute hold off the review, I'm going to generalize this further to include filtering.

momchil-flex:
Okay now you can review. :) I've introduced advanced filter/sort options. You can select to filter by a given property so that values e.g. over or under a reference value will be put first in the list; then separately you can sort each of the two groups by some other criterion. I think with this #1391 can be closed - probably one thing that still would be nice to add though is some built-in that computes the field intensity close to the plane boundaries (maybe not just *on* them, but within some region), which can help with PML mode filtering. Or similarly some metric that computes the field intensity within a given region.

yuanshen-flexcompute (commented):
Nice! Generally very clean and easy to use. I have some comments:
* `sort_order` should default to `ascending` (currently `descending`) when `sort_reference` is not `None`. This way, values closest to the reference are returned first.
* ~~Sorting after filtering doesn't seem to work correctly in some cases. I tested filtering by `mode_area` under a certain value, then sorting by `n_eff` closest to a reference value. This resulted in the sorting "breaking" the filtered order in some cases. I have a test notebook if you would like to take a closer look.~~ This actually might be user error. Things look good for now.

Thanks!

momchil-flex:
Thanks!

> * `sort_order` should default to `ascending` (currently `descending`) when `sort_reference` is not `None`. This way, values closest to the reference are returned first.

I think that makes sense. Originally I set it to `descending`, because I thought the default value should be our current default mode sorting, which is by descending n_eff. However I think this is not really needed since that is still the default when `sort_key` is `None`, while for other applications `ascending` likely makes more sense as you point out, especially if you also supply a `sort_reference`. 

> * Sorting after filtering doesn't seem to work correctly in some cases. I tested filtering by `mode_area` under a certain value, then sorting by `n_eff` closest to a reference value. This resulted in the sorting "breaking" the filtered order in some cases. I have a test notebook if you would like to take a closer look.

Yeah, thanks, a minimally reproducible example would be great! I added a test for that but it might not be comprehensive enough so this will help fix and extend testing.

tidy3d/components/mode/mode_solver.py
```diff
@@ -1324,34 +1326,104 @@ def _filter_components(self, mode_solver_data: ModeSolverData):
...
+        """Apply a mode reordering along mode_index for a single frequency index."""
+        mode_solver_data._cached_properties = {}  # clear cache as we're modifying in-place
+        for data in [
+            *list(mode_solver_data.field_components.values()),
+            mode_solver_data.n_complex,
+            mode_solver_data.grid_primal_correction,
+            mode_solver_data.grid_dual_correction,
+        ]:
+            if hasattr(data, "values") and data.values.ndim >= 2:
+                data.values[..., ifreq, :] = data.values[..., ifreq, sort_inds]
```
caseyflex: two comments here:

indexing xarray values directly is not robust compared to something like this: `data.loc[dict(frequency=ifreq)] = data.sel(frequency=ifreq).isel(mode=sort_inds)`
so that might be preferred, unless the performance is a big issue

also, maybe we can be clear that we mutate the provided mode_solver_data

tidy3d/components/mode/mode_solver.py
```diff
@@ -1324,34 +1326,104 @@ def _filter_components(self, mode_solver_data: ModeSolverData):
...
+
+        for ifreq in range(num_freqs):
+            all_inds = np.arange(self.mode_spec.num_modes)
+
+            # Build groups according to filter if requested
+            if filter_metric is not None:
+                vals_filt = filter_metric.isel(f=ifreq)
+                # Boolean mask for modes in the first group
+                if filter_spec.filter_order == "over":
+                    mask_first = (vals_filt >= filter_spec.filter_reference).to_numpy()
```
caseyflex: since you're already operating on numpy arrays, it would be more efficient to convert everything to numpy at the top

tidy3d/components/mode/mode_solver.py
```diff
@@ -1324,34 +1326,104 @@ def _filter_components(self, mode_solver_data: ModeSolverData):
...
-            for data in [
-                *list(mode_solver_data.field_components.values()),
-                mode_solver_data.n_complex,
-                mode_solver_data.grid_primal_correction,
-                mode_solver_data.grid_dual_correction,
-            ]:
-                data.values[..., ifreq, :] = data.values[..., ifreq, sort_inds]
+            self._apply_mode_reorder(mode_solver_data, ifreq, sort_inds)
+
+    def _sort_modes(self, mode_solver_data: ModeSolverData) -> None:
```
caseyflex: Can we have a public function `sort_modes(mode_solver_data, filter_spec)`? That way we can apply it automatically, or the user can do additional filtering as a post-processing step?

tests/test_plugins/test_mode_solver.py
```diff
@@ -1299,3 +1302,104 @@ def test_translated_dot():
...
     assert np.allclose(data_translated.outer_dot(data2), data2.outer_dot(data2), atol=atol)
+
+
+def test_mode_spec_filter_pol_filter_spec_exclusive():
+    """Ensure ModeSpec errors when both filter_pol and filter_spec are set."""
+    with pytest.raises(pydantic.ValidationError, match="simultaneously"):
+        _ = td.ModeSpec(num_modes=1, filter_pol="te", filter_spec=td.ModeFilterSpec())
+
+
+def test_modes_filter_sort():
```
caseyflex, line 1313: does this test the case where we filter only and don't sort? maybe need to check test coverage again after filtering was added?

caseyflex (commented):
This is a great addition! It mostly looks good, just a few comments

tests/test_plugins/test_mode_solver.py
momchil-flex: I'll expand the testing, but generally test coverage does not check whether different pathways are tested together vs separately. It just tests whether every line is called by *some* test. So e.g. a test that does filtering only and a test that does sorting only will already pass test coverage.

tidy3d/components/mode/mode_solver.py
momchil-flex: So I was thinking about this myself and I was just worried about some potential issues / confusion, but I think after thinking about it more carefully now it could be worth it. I think the method could be a call on `ModeSolverData` directly, like `new_data = old_data.filter_sort(filter_spec=...)` (this one will not be modifying in-place). Some things to pay attention to:
- `ModeSolverData` contains a `ModeSolverMonitor` which stores the original `mode_spec.filter_spec`. So the sorting method should also update that.
- Something that can be slightly confusing, but I think it's ok, is that the user can use this method to explore various sortings locally, but then they can't use the sorted data as a mode source or monitor directly. They need to pass the desired filter spec to the mode spec of their object.
- For `ModeSimulationData`, they would have to do something like `new_modes = mode_sim_data.modes.filter_sort(...)`, i.e. we don't create new `ModeSimulationData`, just a new `ModeSolverData`.

Does this plan sound good? Any other caveats that come to mind?

tidy3d/components/mode/mode_solver.py
momchil-flex: So in general I agree with you but it turns out it's hard to mutate an xarray array in place. This

```
data.loc[{"f": freq}] = data.sel(f=freq).isel(mode_index=sort_inds)
```

fails with `IndexError: dimension coordinate 'mode_index' conflicts between indexed and indexing objects`. I tinkered around with it a bit and it seems like the only way (or at least the main way) would be to reassign a data array rather than sort it in-place. Then this is complicated further by the fact that we need to modify multiple data arrays, so I think it might have to go through a dict representation and then update the mode data with the updated dict.

Overall it seems a bit too much given that the current approach works... Btw note that I just separated it into a function, otherwise it's the same sorting approach we've been using forever in `filter_pol`.

tidy3d/components/mode/mode_solver.py
caseyflex: Oh yeah, this reminds me, what if the user wants to filter on multiple properties? Should we have an ordered list of filter_specs?

then if you already had a filter spec, and you filter again, it can concatenate to this list. Then the entire list can be used in a mode source.

otherwise sounds good. Although there’s no reason not to also allow calling this directly on ModeSimulationData I suppose

tests/test_plugins/test_mode_solver.py
caseyflex: I think line 1424 of mode_solver.py was not tested

tidy3d/components/mode/mode_solver.py
momchil-flex: Right, I guess I could add it to ModeSimulationData - I just need to update the associated `simulation` too...

List of filtering sounds... too much honestly :D

momchil-flex:
The issue that @yuanshen-flexcompute thought he was seeing was related to the fact that by default, we have `track_freq="central"`. We already explain in the description of `ModeSpec.filter_spec` that if it's used in combination with `track_freq`, the filter/sort will only be correct at the tracked frequency, while at other frequencies it could be overwritten by the sort. To me this sounds like the correct behavior, but it could be confusing to a user. To obtain strictly sorted modes at every frequency, they'd have to set `track_freq=None`. What do you think @caseyflex ?

tidy3d/components/mode/mode_solver.py
caseyflex: Maybe, but it’s needed if you want every use case to be translatable into a modeSource

caseyflex:
> The issue that @yuanshen-flexcompute thought he was seeing was related to the fact that by default, we have `track_freq="central"`. We already explain in the description of `ModeSpec.filter_spec` that if it's used in combination with `track_freq`, the filter/sort will only be correct at the tracked frequency, while at other frequencies it could be overwritten by the sort. To me this sounds like the correct behavior, but it could be confusing to a user. To obtain strictly sorted modes at every frequency, they'd have to set `track_freq=None`. What do you think @caseyflex ?

I see, that’s a good point. So this new sorting is frequency independent, while track_freq couples different frequencies. Currently, track_freq is applied second. Since they are both sorting, it could be nice to unify their api in mode_spec. Or could just validate that only one or the other is used.

momchil-flex:
> I see, that’s a good point. So this new sorting is frequency independent, while track_freq couples different frequencies. Currently, track_freq is applied second. Since they are both sorting, it could be nice to unify their api in mode_spec. Or could just validate that only one or the other is used.

So I don't think that's a good idea, because a typical use case could very well be that you roughly want a certain filtering, but you also want the modes to be continuously ordered across frequencies, that is to say the latter is still more important than the strict filtering. Like, I want to have my TM mode come first at my central frequency, but if there's a crossing somewhere I don't want to force the reshuffling.

By the way the reason I still filter all modes rather than the tracked frequency only, if `track_freq` is set, is to help with the reshuffling after. Since we don't compute the full outer_dot among each frequency (too expensive), the reshuffling works well only if the modes are already roughly ordered.

So having thought about this I think indeed moving `track_freq` into the `sort_freq` would make sense (deprecating the direct `mode_spec.track_freq`) because indeed it is *also* related to how modes are sorted. So at least if everything is kept in the same place, hopefully it will be less confusing.

caseyflex:
@momchil-flex this might be another case for having a list of filter_specs, so the order in which they are applied is transparent

momchil-flex:
Well tracking first and then sorting doesn't really make sense as an option, right? Because that would just sort. You just need to define whether you want to track in the end or not.

momchil-flex:
@greptile take a look

momchil-flex:
Alright so I did some major updates based on previous discussions @caseyflex :
- Probably most notably, I introduced `sort_spec.track_freq` and deprecated `mode_spec.track_freq`, because indeed the tracking can be thought of part of the sorting. This did lead to various headaches though specifically when leaving `mode_spec.track_freq` as allowed but deprecated. The way this works now is that `mode_spec.track_freq=None` by default, but if it *is* set, then it *will* take precedence (but a deprecation warning will be printed).
- I also exposed standalone `ModeData.sort_modes` and `ModeSimulationData.sort_modes` methods that the user can utilize to reorder modes after the solve.
- I refactored everything to avoid in-place operations. I also realized that we need a method that explicitly iterates over all data attributes of a dataset, because previously we were *not* reordering things like `n_group` or the `amps` if present, because we were manually iterating only over the field data and the grid corrections. The one place where this was problematic previously from the user side was `overlap_sort` performed *after* the mode solving @dbochkov-flexcompute . Now it should be all good and I also unified the mode reordering among the `filter_pol`, `track_freq` and the `sort_spec` methods.
- Probably the one remaining awkwardness that I'm not 100% happy with is that `overlap_sort` and `sort_modes` will always set the `mode_data.monitor.mode_spec.track_freq` to `None`, and set the `sort_spec.track_freq` to the one that matches the tracking that was done. This can lead to a situation where the `mode_data.monitor.mode_spec` is not the same as the one of the `monitor` that initially generated it if `track_freq` was defined there in the deprecated way. However, this resulted in the most straightforward usage for everything else.
- Expanded the test coverage quite a bit.

caseyflex (approved):
Great, thanks @momchil-flex !

> Probably the one remaining awkwardness that I'm not 100% happy with is that overlap_sort and sort_modes will always set the mode_data.monitor.mode_spec.track_freq to None, and set the sort_spec.track_freq to the one that matches the tracking that was done. This can lead to a situation where the mode_data.monitor.mode_spec is not the same as the one of the monitor that initially generated it if track_freq was defined there in the deprecated way. However, this resulted in the most straightforward usage for everything else.

At least this only happens if the deprecated api is used. Actually, I had a similar concern. Say the original `monitor.mode_spec` filters based on one key, and then the user calls `sort_modes` with another filter key for example. Then the updated `monitor.mode_spec` will only contain the second filter key, and if you use that monitor in a fresh simulation, it will give different results.

More broadly, this difficulty comes up repeatedly due to our concept of `MonitorData` containing a `Dataset` as well as a `Monitor`. For example, `ModeData._isel` is private because it results in inconsistency between the monitor and the data. Also in EMESimulationData, we have `_extract_mode_solver_data`, which makes an attempt to update the monitor in a reasonable way, but is not guaranteed that the `Monitor` would reproduce the `MonitorData` if used in a fresh simulation. It would be extremely useful to have `isel` functionality for all of our `Datasets`.

Probably to avoid this, we would have to shift a lot of functionality out of the `MonitorData` classes into the `Dataset` classes, so that then `MonitorData` is really just a `Dataset` together with a `Monitor` that gives rise to it. And the methods that process the `Dataset` return a new `Dataset` that doesn't necessarily correspond to any `Monitor`.

Not suggesting changing this here, just bringing it up as a recurring difficulty.

yuanshen-flexcompute (approved):
Latest changes looks good to me. Thanks!

momchil-flex:
> Actually, I had a similar concern. Say the original `monitor.mode_spec` filters based on one key, and then the user calls `sort_modes` with another filter key for example. Then the updated `monitor.mode_spec` will only contain the second filter key, and if you use that monitor in a fresh simulation, it will give different results.

Good thought - actually in this case, the sorting errors, because we explicitly error if both `filter_pol` and `sort_spec.filter_key` or `sort_spec.sort_key` are defined, so when the sorting function tries to update `monitor.mode_spec.sort_spec`, it errors. Which is good, right?

The reason I couldn't do the same for `track_freq` is that its default is not `None` but instead `"central"`. And I want to move this default to `sort_spec.track_freq`, but then still keep the deprecated `mode_spec.track_freq` option (which I set to None as default). Basically, the current defaults are `track_freq = None` and `sort_spec = td.ModeSortSpec(track_freq="central")`. If I add a validator that those two can't be defined simultaneously, it will slightly break down compatibility, because where users had just e.g. `mode_spec = td.ModeSpec(track_freq="lowest")`, they would then have to do one of these: `mode_spec = td.ModeSpec(track_freq="lowest", sort_spec=None)`, or `mode_spec = td.ModeSpec(sort_spec=td.ModeSortSpec(track_freq="lowest"))`. To be honest... it almost sounds worth it though to avoid the other internal headaches.

PR #2871, opened 2025-10-07
FXC-3294-add-opt-in-local-cache-for-simulation-results-hashed-by-simulation-runtime-context

marcorudolphflex:
**Summary**

Add opt-in local cache for simulation results hashed by simulation + runtime context

**Background**

Every repeated run of the same simulation re-uploads/downloads data. We want to cache the resulting .hdf5 locally (opt-in) and reuse it on identical runs. Cache key must go beyond _hash_self() because solver version, workflow path, environment, and runtime options also influence the artifact we return.

<h2>Greptile Overview</h2>

Updated On: 2025-10-07 11:35:12 UTC

<h3>Summary</h3>
This PR introduces a comprehensive opt-in local caching system for Tidy3D simulation results to eliminate redundant uploads and downloads when running identical simulations. The cache system stores simulation result HDF5 files locally using a composite cache key that combines simulation hash with runtime context (solver version, workflow type, environment variables) to ensure cache validity beyond simple simulation parameters.

The implementation centers around a new `SimulationCache` class in `tidy3d/web/cache.py` that provides thread-safe LRU caching with configurable size and entry limits. The cache integrates seamlessly into the existing web API workflow by intercepting `run()` calls early to check for cached results and storing successful simulation outputs in the `load()` function. Cache entries are validated using SHA256 checksums and support atomic file operations to prevent corruption during concurrent access.

Configuration is handled through a new `SimulationCacheSettings` class in `config.py` with sensible defaults (disabled by default, 10GB max size, 128 max entries, `~/.tidy3d/cache/simulations` directory). The feature is exposed through an optional `use_cache` parameter across all API entry points (`run()`, `run_async()`, autograd functions) allowing per-call override of global cache settings.

The cache system handles both individual `Job` objects and batch operations through `Batch` objects, with comprehensive error handling for edge cases like cache corruption, missing files, and network failures. The implementation follows enterprise software patterns with proper thread synchronization using a global singleton pattern and LRU eviction policies to manage cache size.

| tidy3d/web/cache.py | 4/5 | New comprehensive cache implementation with thread-safe LRU caching, atomic operations, and checksum verification |
| tidy3d/web/api/container.py | 4/5 | Core integration of cache logic into Job/Batch classes with cache-aware workflow methods |
| tidy3d/web/api/webapi.py | 4/5 | Integration of cache lookup/storage into main run() and load() functions with proper error handling |
| tidy3d/config.py | 4/5 | New SimulationCacheSettings configuration class with validation and sensible defaults |
| tests/test_web/test_simulation_cache.py | 4/5 | Comprehensive test coverage for cache functionality with extensive mocking of web pipeline |
| tidy3d/web/api/autograd/autograd.py | 5/5 | Clean addition of use_cache parameter to autograd-compatible run functions |
| tidy3d/web/api/autograd/engine.py | 5/5 | Simple addition of use_cache to job_fields list for proper parameter propagation |
| tidy3d/web/api/asynchronous.py | 5/5 | Addition of use_cache parameter to run_async function with proper documentation |

</details>

- This PR introduces significant new functionality with good architectural design and comprehensive error handling
- Score reflects complex caching logic that could have edge cases, but implementation follows solid engineering practices
- Pay close attention to tidy3d/web/cache.py and tidy3d/web/api/container.py for the core caching logic and integration points

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant WebAPI as "web.run()"
    participant Cache as "SimulationCache"
    participant Job as "Job"
    participant Server as "Server"
    participant Stub as "Tidy3dStub"

    User->>WebAPI: "run(simulation, use_cache=True)"
    WebAPI->>Cache: "resolve_simulation_cache(use_cache)"
    Cache-->>WebAPI: "SimulationCache instance or None"
    
    alt Cache enabled
        WebAPI->>Cache: "try_fetch(simulation)"
        Cache->>Cache: "build_cache_key(simulation_hash, workflow_type, version)"
        Cache->>Cache: "_fetch(cache_key)"
        alt Cache hit
            Cache-->>WebAPI: "CacheEntry"
            WebAPI->>Cache: "materialize(path)"
            Cache-->>WebAPI: "Cached data path"
            WebAPI->>Stub: "postprocess(path)"
            Stub-->>WebAPI: "SimulationData"
            WebAPI-->>User: "SimulationData (from cache)"
        else Cache miss
            Cache-->>WebAPI: "None"
            WebAPI->>Job: "upload(simulation)"
            Job->>Server: "Upload simulation"
            Server-->>Job: "task_id"
            WebAPI->>Job: "start(task_id)"
            Job->>Server: "Start simulation"
            WebAPI->>Job: "monitor(task_id)"
            Job->>Server: "Poll status"
            Server-->>Job: "Status updates"
            WebAPI->>Job: "download(task_id, path)"
            Job->>Server: "Download results"
            Server-->>Job: "Simulation data file"
            WebAPI->>Stub: "postprocess(path)"
            Stub-->>WebAPI: "SimulationData"
            WebAPI->>Cache: "store_result(stub_data, task_id, path, workflow_type)"
            Cache->>Cache: "_store(cache_key, task_id, source_path, metadata)"
            WebAPI-->>User: "SimulationData"
        end
    else Cache disabled
        WebAPI->>Job: "upload(simulation)"
        Job->>Server: "Upload simulation"
        Server-->>Job: "task_id"
        WebAPI->>Job: "start(task_id)"
        Job->>Server: "Start simulation"
        WebAPI->>Job: "monitor(task_id)"
        Job->>Server: "Poll status"
        Server-->>Job: "Status updates"
        WebAPI->>Job: "download(task_id, path)"
        Job->>Server: "Download results"
        Server-->>Job: "Simulation data file"
        WebAPI->>Stub: "postprocess(path)"
        Stub-->>WebAPI: "SimulationData"
        WebAPI-->>User: "SimulationData"
    end
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -168,6 +168,18 @@ This will produce the following plot, which visualizes the electromagnetic field
 
 You can now postprocess simulation data using the same python session, or view the results of this simulation on our web-based `graphical user interface (GUI) <https://tidy3d.simulation.cloud>`_.
 
+.. admonition:: Caching for repeated simulations
+   :class: tip
+
+   Repeated runs of the same simulation can reuse solver results by enabling the
+   local cache: ``td.config.local_cache.enabled = True``. You may configure the cache directory
+   with ``local_cache.directory``. If the number of entries (``local_cache.max_entries``) or the storage size
+   (``local_cache.max_size_gb``) is exceeded, cache entries are evicted by least-recently-used (LRU) order.
+   You can clear all stored artifacts with ``td.web.cache.clear()``.
+   Additionally, there is server-side caching controlled via ``td.config.web.enable_caching``
+   (enabled by default). While this can avoid recomputation on the server, it still requires
+   upload/download of results which is why we recommend enabling the local cache.
+
 .. `TODO: open example in colab <https://github.com/flexcompute/tidy3d>`_
 
 
@@ -259,7 +271,6 @@ Contents
   development/index
   changelog
   About our Solver <https://www.flexcompute.com/tidy3d/solver/>
-
 
 
 
--- a/tidy3d/config/sections.py
+++ b/tidy3d/config/sections.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import os
 import ssl
+from os import PathLike
 from pathlib import Path
 from typing import Any, Literal, Optional
 from urllib.parse import urlparse
@@ -11,7 +13,10 @@ import numpy as np
 from pydantic import (
     BaseModel,
     ConfigDict,
+    DirectoryPath,
     Field,
+    NonNegativeFloat,
+    NonNegativeInt,
     PositiveInt,
     SecretStr,
     field_serializer,
@@ -379,6 +384,67 @@ def apply_web(config: WebConfig) -> None:
     if manager is None:
         raise RuntimeError("Configuration manager not attached; cannot apply web env overrides.")
     manager.apply_web_env(dict(config.env_vars))
+
+
+def _default_cache_directory() -> Path:
+    """Determine the default on-disk cache directory respecting platform conventions."""
+
+    base_override = os.getenv("TIDY3D_BASE_DIR")
+    if base_override:
+        base = Path(base_override).expanduser().resolve()
+        return (base / "cache" / "simulations").resolve()
+    else:
+        xdg_cache = os.getenv("XDG_CACHE_HOME")
+        if xdg_cache:
+            base = Path(xdg_cache).expanduser().resolve()
+        else:
+            base = Path.home() / ".cache"
+    return (base / "tidy3d" / "simulations").resolve()
+
+
+@register_section("local_cache")
+class LocalCacheConfig(ConfigSection):
+    """Settings controlling the optional local simulation cache."""
+
+    enabled: bool = Field(
+        False,
+        title="Enable cache",
+        description="Enable or disable the local simulation cache.",
+        json_schema_extra={"persist": True},
+    )
+
+    directory: DirectoryPath = Field(
+        default_factory=_default_cache_directory,
+        title="Cache directory",
+        description="Directory where cached artifacts are stored.",
+        json_schema_extra={"persist": True},
+    )
+
+    max_size_gb: NonNegativeFloat = Field(
+        10.0,
+        title="Maximum cache size (GB)",
+        description="Maximum cache size in gigabytes. Set to 0 for no size limit.",
+        json_schema_extra={"persist": True},
+    )
+
+    max_entries: NonNegativeInt = Field(
+        0,
+        title="Maximum cache entries",
+        description="Maximum number of cache entries. Set to 0 for no limit.",
+        json_schema_extra={"persist": True},
+    )
+
+    @field_validator("directory", mode="before")
+    def _ensure_directory_exists(cls, v: PathLike) -> Path:
+        """Expand ~, resolve path, and create directory if missing before DirectoryPath validation."""
+        p = Path(v).expanduser().resolve()
+        p.mkdir(parents=True, exist_ok=True)
+        return p
+
+    @field_serializer("directory")
+    def _serialize_directory(self, value: Path) -> str:
+        """Persist directory as strings."""
+        return str(value)
 
 
 @register_section("plugins")
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -2,8 +2,13 @@
 
 from __future__ import annotations
 
+import atexit
 import concurrent
+import os
+import shutil
+import tempfile
 import time
+import uuid
 from abc import ABC
 from collections.abc import Mapping
 from concurrent.futures import ThreadPoolExecutor
@@ -12,6 +17,7 @@ from pathlib import Path
 from typing import Literal, Optional, Union
 
 import pydantic.v1 as pd
+from pydantic.v1 import PrivateAttr
 from rich.progress import BarColumn, Progress, TaskProgressColumn, TextColumn, TimeElapsedColumn
 
 from tidy3d.components.base import Tidy3dBaseModel, cached_property
@@ -32,6 +38,7 @@ from tidy3d.web.api.states import (
     STATE_PROGRESS_PERCENTAGE,
 )
 from tidy3d.web.api.tidy3d_stub import Tidy3dStub
+from tidy3d.web.api.webapi import restore_simulation_if_cached
 from tidy3d.web.core.constants import TaskId, TaskName
 from tidy3d.web.core.task_core import Folder
 from tidy3d.web.core.task_info import RunInfo, TaskInfo
@@ -241,6 +248,29 @@ class Job(WebContainer):
         "reduce_simulation",
     )
 
+    _stash_path: Optional[str] = PrivateAttr(default=None)
+
+    def _stash_path_for_job(self) -> str:
+        """Stash file which is a temporary location for the cached-restored file."""
+        stash_dir = Path(tempfile.gettempdir()) / "tidy3d_stash"
+        stash_dir.mkdir(parents=True, exist_ok=True)
+        return str(Path(stash_dir / f"{self._cached_task_id}.hdf5"))
+
+    def _materialize_from_stash(self, dst_path: os.PathLike) -> None:
+        """Atomic copy from stash to requested path."""
+        tmp = str(dst_path) + ".part"
+        shutil.copy2(self._stash_path, tmp)
+        os.replace(tmp, dst_path)
+
+    def clear_stash(self) -> None:
+        """Delete this job's stash file only."""
+        if self._stash_path:
+            try:
+                if os.path.exists(self._stash_path):
+                    os.remove(self._stash_path)
+            finally:
+                self._stash_path = None
+
     def to_file(self, fname: PathLike) -> None:
         """Exports :class:`Tidy3dBaseModel` instance to .yaml, .json, or .hdf5 file
 
@@ -258,7 +288,9 @@ class Job(WebContainer):
         super(Job, self).to_file(fname=fname)  # noqa: UP008
 
     def run(
-        self, path: PathLike = DEFAULT_DATA_PATH, priority: Optional[int] = None
+        self,
+        path: PathLike = DEFAULT_DATA_PATH,
+        priority: Optional[int] = None,
     ) -> WorkflowDataType:
         """Run :class:`Job` all the way through and return data.
 
@@ -274,17 +306,50 @@ class Job(WebContainer):
         :class:`WorkflowDataType`
             Object containing simulation results.
         """
-        self.upload()
-        if priority is None:
-            self.start()
-        else:
-            self.start(priority=priority)
-        self.monitor()
-        return self.load(path=path)
+        self._check_path_dir(path=path)
+
+        loaded_from_cache = self.load_if_cached
+        if not loaded_from_cache:
+            self.upload()
+            if priority is None:
+                self.start()
+            else:
+                self.start(priority=priority)
+            self.monitor()
+        data = self.load(path=path)
+
+        return data
+
+    @cached_property
+    def load_if_cached(self) -> bool:
+        """Checks if results are cached and (if yes) restores them into our shared stash file."""
+        # use temporary path as final destination is unknown
+        stash_path = self._stash_path_for_job()
+
+        restored = restore_simulation_if_cached(
+            simulation=self.simulation,
+            path=stash_path,
+            reduce_simulation=self.reduce_simulation,
+            verbose=getattr(self, "verbose", True),
+        )
+
+        if restored is None:
+            return False
+
+        self._stash_path = stash_path
+        atexit.register(self.clear_stash)
+        return True
+
+    @cached_property
+    def _cached_task_id(self) -> TaskId:
+        """The task ID for jobs which are loaded from cache."""
+        return "cached_" + self.task_name + "_" + str(uuid.uuid4())
 
     @cached_property
     def task_id(self) -> TaskId:
         """The task ID for this ``Job``. Uploads the ``Job`` if it hasn't already been uploaded."""
+        if self.load_if_cached:
+            return self._cached_task_id
         if self.task_id_cached:
             return self.task_id_cached
         self._check_folder(self.folder_name)
@@ -298,7 +363,9 @@ class Job(WebContainer):
         return task_id
 
     def upload(self) -> None:
-        """Upload this ``Job``."""
+        """Upload this ``Job`` if not already got cached results."""
+        if self.load_if_cached:
+            return
         _ = self.task_id
 
     def get_info(self) -> TaskInfo:
@@ -314,6 +381,8 @@ class Job(WebContainer):
     @property
     def status(self):
         """Return current status of :class:`Job`."""
+        if self.load_if_cached:
+            return "success"
         if web._is_modeler_batch(self.task_id):
             detail = self.get_info()
             status = detail.totalStatus.value
@@ -346,13 +415,16 @@ class Job(WebContainer):
         Note
         ----
         To monitor progress of the :class:`Job`, call :meth:`Job.monitor` after started.
+        Function has no effect if cache is enabled and data was found in cache.
         """
-        web.start(
-            self.task_id,
-            solver_version=self.solver_version,
-            pay_type=self.pay_type,
-            priority=priority,
-        )
+        loaded = self.load_if_cached
+        if not loaded:
+            web.start(
+                self.task_id,
+                solver_version=self.solver_version,
+                pay_type=self.pay_type,
+                priority=priority,
+            )
 
     def get_run_info(self) -> RunInfo:
         """Return information about the running :class:`Job`.
@@ -372,6 +444,8 @@ class Job(WebContainer):
         To load the output of completed simulation into :class:`.SimulationData` objects,
         call :meth:`Job.load`.
         """
+        if self.load_if_cached:
+            return
         web.monitor(self.task_id, verbose=self.verbose)
 
     def download(self, path: PathLike = DEFAULT_DATA_PATH) -> None:
@@ -386,6 +460,9 @@ class Job(WebContainer):
         ----
         To load the data after download, use :meth:`Job.load`.
         """
+        if self.load_if_cached:
+            self._materialize_from_stash(path)
+            return
         self._check_path_dir(path=path)
         web.download(task_id=self.task_id, path=path, verbose=self.verbose)
 
@@ -403,8 +480,11 @@ class Job(WebContainer):
             Object containing simulation results.
         """
         self._check_path_dir(path=path)
+        if self.load_if_cached:
+            self._materialize_from_stash(path)
+
         data = web.load(
-            task_id=self.task_id,
+            task_id=None if self.load_if_cached else self.task_id,
             path=path,
             verbose=self.verbose,
             lazy=self.lazy,
@@ -450,6 +530,8 @@ class Job(WebContainer):
         Cost is calculated assuming the simulation runs for
         the full ``run_time``. If early shut-off is triggered, the cost is adjusted proportionately.
         """
+        if self.load_if_cached:
+            return 0.0
         return web.estimate_cost(self.task_id, verbose=verbose, solver_version=self.solver_version)
 
     def postprocess_start(self, worker_group: Optional[str] = None, verbose: bool = True) -> None:
@@ -547,6 +629,11 @@ class BatchData(Tidy3dBaseModel, Mapping):
     verbose: bool = pd.Field(
         True, title="Verbose", description="Whether to print info messages and progressbars."
     )
+    cached_tasks: Optional[dict[TaskName, bool]] = pd.Field(
+        None,
+        title="Cached Tasks",
+        description="Whether the data of a task came from the cache.",
+    )
 
     lazy: bool = pd.Field(
         False,
@@ -554,13 +641,27 @@ class BatchData(Tidy3dBaseModel, Mapping):
         description="Whether to load the actual data (lazy=False) or return a proxy that loads the data when accessed (lazy=True).",
     )
 
+    is_downloaded: Optional[bool] = pd.Field(
+        False,
+        title="Is Downloaded",
+        description="Whether the simulation data was downloaded before.",
+    )
+
     def load_sim_data(self, task_name: str) -> WorkflowDataType:
         """Load a simulation data object from file by task name."""
         task_data_path = Path(self.task_paths[task_name])
         task_id = self.task_ids[task_name]
-        web.get_info(task_id)
+        from_cache = self.cached_tasks[task_name] if self.cached_tasks else False
+        if not from_cache:
+            web.get_info(task_id)
 
-        return web.load(task_id=task_id, path=task_data_path, verbose=False, lazy=self.lazy)
+        return web.load(
+            task_id=None if from_cache else task_id,
+            path=task_data_path,
+            verbose=self.verbose,
+            replace_existing=not (from_cache or self.is_downloaded),
+            lazy=self.lazy,
+        )
 
     def __getitem__(self, task_name: TaskName) -> WorkflowDataType:
         """Get the simulation data object for a given ``task_name``."""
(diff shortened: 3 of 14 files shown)


lucas-flexcompute (approved):
I believe this works for photonforge!
Just to make sure I understand correctly, we basically do:
```
cache = get_cache()
entry = cache.try_fetch(simulation)
if entry: ...
 # later, after loading results
cache.store_result(simulation_data, task_id, path)
```
Can I set path to be the actual cache path? I don't want the user having to worry about where the data will reside.

lucas-flexcompute:
I also just realized there's no way to retrieve the cached data without copying (https://github.com/flexcompute/tidy3d/blob/80d7e47a62fe4f28209038b73315aa4e298dfa5b/tidy3d/web/cache.py#L254)
Can we have the option to skip copying? Imagine doing this for hundreds of files in a single circuit simulation… Ideally, I'd like to just retrieve the data itself.

marcorudolphflex:
> I also just realized there's no way to retrieve the cached data without copying (
> 
> https://github.com/flexcompute/tidy3d/blob/80d7e47a62fe4f28209038b73315aa4e298dfa5b/tidy3d/web/cache.py#L254
> 
> )
> Can we have the option to skip copying? Imagine doing this for hundreds of files in a single circuit simulation… Ideally, I'd like to just retrieve the data itself.

I just added the option to specify no path (default).
So you could just use 
`load_simulation_if_cached(sim, use_cache=True)`
if `use_cache=None` (default), the default cache setting is used (config+environment)

marcorudolphflex:
> I believe this works for photonforge! Just to make sure I understand correctly, we basically do:
> 
> ```
> cache = get_cache()
> entry = cache.try_fetch(simulation)
> if entry: ...
>  # later, after loading results
> cache.store_result(simulation_data, task_id, path)
> ```
> 
> Can I set path to be the actual cache path? I don't want the user having to worry about where the data will reside.

I would just rely on 
`load_simulation_if_cached(sim)`
to have a common interface. ̶T̶̶h̶̶i̶̶s̶ ̶d̶̶o̶̶e̶̶s̶ ̶t̶̶h̶̶e̶ ̶s̶̶t̶̶o̶̶r̶̶i̶̶n̶̶g̶ ̶a̶̶u̶̶t̶̶o̶̶m̶̶a̶̶t̶̶i̶̶c̶̶a̶̶l̶̶l̶̶y̶.  If not found in cache, `None` is returned
Edit: storing is done with regular "run"

lucas-flexcompute:
> I would just rely on `load_simulation_if_cached(sim)` to have a common interface. This does the storing automatically.

Ah, I see! Sorry, I completely missed that function! No problems, then!

marcorudolphflex:
> > I would just rely on `load_simulation_if_cached(sim)` to have a common interface. This does the storing automatically.
> 
> Ah, I see! Sorry, I completely missed that function! No problems, then!

Perfect. One correction from my last comment: `load_simulation_if_cached` does not store in cache, but regular "run" does if cache is enabled or forced with "use_cache"

lucas-flexcompute:
> > > I would just rely on `load_simulation_if_cached(sim)` to have a common interface. This does the storing automatically.
> > 
> > 
> > Ah, I see! Sorry, I completely missed that function! No problems, then!
> 
> Perfect. One correction from my last comment: `load_simulation_if_cached` does not store in cache, but regular "run" does if cache is enabled or forced with "use_cache"

If I don't use `run` and load the results manually with `web.load`, I see that has the same argument and should work as well, right?

marcorudolphflex:
> > > > I would just rely on `load_simulation_if_cached(sim)` to have a common interface. This does the storing automatically.
> > > 
> > > 
> > > Ah, I see! Sorry, I completely missed that function! No problems, then!
> > 
> > 
> > Perfect. One correction from my last comment: `load_simulation_if_cached` does not store in cache, but regular "run" does if cache is enabled or forced with "use_cache"
> 
> If I don't use `run` and load the results manually with `web.load`, I see that has the same argument and should work as well, right?

Yup! Technically, storing is done in load.

tidy3d/web/api/webapi.py
```diff
@@ -312,6 +314,124 @@ def _task_dict_to_url_bullet_list(data_dict: dict) -> str:
...
+    simulation_cache = resolve_simulation_cache(use_cache)
+    retrieved_simulation_path = None
+    if simulation_cache is not None:
+        sim_for_cache = simulation
+        if isinstance(simulation, (ModeSolver, ModeSimulation)):
+            sim_for_cache = get_reduced_simulation(simulation, reduce_simulation)
+        entry = simulation_cache.try_fetch(simulation=sim_for_cache, verbose=verbose)
+        if entry is not None:
+            if path is not None:
+                retrieved_simulation_path = _copy_simulation_data_from_cache_entry(entry, path)
```
yaugenst-flex: Doesn't this make `restore_simulation_if_cached` return a `bool` instead of a path (or `None`)?

tidy3d/web/api/webapi.py
```diff
@@ -312,6 +314,124 @@ def _task_dict_to_url_bullet_list(data_dict: dict) -> str:
...
+    Returns
+    -------
+    Optional[WorkflowDataType]
+        The loaded simulation data if found in cache, otherwise None.
+    """
+    restored_path = restore_simulation_if_cached(simulation, path, use_cache, reduce_simulation)
+    if restored_path:
+        data = load(
+            task_id=None,
+            path=str(restored_path),
```
yaugenst-flex, line 425: Similar problem here, `restored_path` is a `bool`, no?

tidy3d/web/api/container.py
```diff
@@ -401,8 +486,18 @@ def load(self, path: str = DEFAULT_DATA_PATH) -> WorkflowDataType:
         Union[:class:`.SimulationData`, :class:`.HeatSimulationData`, :class:`.EMESimulationData`]
             Object containing simulation results.
         """
+        if self.load_if_cached:
+            self.move_cache_file(path=path)
+            data = Tidy3dStubData.postprocess(path)
+            return data
```
yaugenst-flex: Not sure whether the early return here is correct. Doesn't this skip both the `web.load(lazy)` flag as well as `ModeSolver._patch_data()`, which are runtime behaviors unrelated to caching?

tidy3d/web/api/container.py
```diff
@@ -385,9 +458,21 @@ def download(self, path: str = DEFAULT_DATA_PATH) -> None:
...
         """
+        if self.load_if_cached:
+            self.move_cache_file(path=path)
+            return
         self._check_path_dir(path=path)
         web.download(task_id=self.task_id, path=path, verbose=self.verbose)
 
+    def move_cache_file(self, path: str) -> None:
+        if self._cache_file_moved:
+            return
```
yaugenst-flex: So once `self._cache_file_moved` is set, changing the `path` passed here will have no effect right? It'll pass silently without actually moving the file to that location. Sounds like unintended behavior to me?

yaugenst-flex (changes requested):
Thanks @marcorudolphflex this will be a super useful feature to have!

tidy3d/web/api/container.py
marcorudolphflex: good point! I now used a stashed directory which persists and deletes on exit.
Also added a test which loads in another path than specified in run.

tidy3d/web/api/webapi.py
marcorudolphflex: Yes this was the first idea here. As Lucas requested having the data without copying it, I decided to return the path which may be the original cache path (if path=None) such that the function gets more informative without having redundancies in return like tuple[bool, Optional[str]]

tidy3d/web/api/webapi.py
marcorudolphflex: see above

tidy3d/web/api/webapi.py
```diff
@@ -1027,22 +1073,43 @@ def load(
     Union[:class:`.SimulationData`, :class:`.HeatSimulationData`, :class:`.EMESimulationData`]
         Object containing simulation data.
     """
+    assert from_cache or task_id, "Either task_id or from_cache must be provided."
```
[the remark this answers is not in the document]
momchil-flex: Generally I believe we don't do assert checks, here a ValueError should be raised instead right @yaugenst-flex ?

docs/index.rst
```diff
@@ -168,6 +168,13 @@ This will produce the following plot, which visualizes the electromagnetic field
 
 You can now postprocess simulation data using the same python session, or view the results of this simulation on our web-based `graphical user interface (GUI) <https://tidy3d.simulation.cloud>`_.
 
+.. tip::
+
+   Repeated runs of the same simulation can reuse solver results by enabling the optional
+   local cache: ``td.config.simulation_cache.enabled = True``. The cache location and limits are
+   configurable (see ``~/.tidy3d/config``), entries are checksum-validated, and you can clear
+   all stored artifacts with ``tidy3d.web.cache.clear()``.
```
momchil-flex: Seems like we don't currently have any explanation about the server-side cache in the docs. As a general comment we need to make this distinction clear here and everywhere else. Some questions that come to mind:

- After the config PR is merged, the server-side caching will be set [here](https://github.com/flexcompute/tidy3d/pull/2711/files#diff-e5d597ca35ca033e398905cb3a7c188f111b19a10fe3a4fd6e6dcfb3948fb933R250-R255). Do we want instead the two caching mechanisms to be toggled under the same config class?
- State in docs exactly what happens when both are set
- In this PR it seems that "local cache" is generally used explicitly in docstrings, but variables like `use_cache` or `SimulationCacheSettings` are not explicit about it - should they always say local explicitly?

tidy3d/web/api/webapi.py
```diff
@@ -1233,6 +1393,11 @@ def load(
         If ``True``, will print progressbars and status, otherwise, will run silently.
     progress_callback : Callable[[float], None] = None
         Optional callback function called when downloading file with ``bytes_in_chunk`` as argument.
+    use_cache: Optional[bool] = None
+        Whether to use local cache if identical simulation is rerun. If not provided, cache settings from config or
+        environment variables will be used.
+    from_cache: bool = None
+        Whether data will be loaded from cache.
```
momchil-flex: Can't we just eliminate `from_cache`, make TaskId optional and then if the simulation is not found in cache and no task id is provided, error correspondingly?

tidy3d/web/api/webapi.py
```diff
@@ -316,6 +318,126 @@ def _task_dict_to_url_bullet_list(data_dict: dict) -> str:
...
+            return True
+        except Exception:
+            return False
+    return False
+
+
+def restore_simulation_if_cached(
+    simulation: WorkflowType,
+    path: Optional[os.PathLike] = None,
+    use_cache: Optional[bool] = None,
```
momchil-flex: What's the benefit / use case of carrying `use_cache` around all sorts of functions as opposed to just setting it in the config, which can be dynamically adjusted before web calls - at least after the config refactor is merged in?

tidy3d/web/api/webapi.py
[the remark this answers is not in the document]
yaugenst-flex: Oh yeah right, missed that! Should definitely raise an error yeah.

tidy3d/web/api/webapi.py
marcorudolphflex: We just want to give the user a more convenient opportunity to configure that for single run calls. But I'm open to discuss that. Would surely clean the code.

docs/index.rst
yaugenst-flex: on web vs local cache in config: definitely we'll want them to be separate toggles, right? i agree it would be nice if they land next to each other in the config, but also local caching shouldnt really live in the web section... happy to discuss alternative section naming. for example i'm also not super sure about `simulation`.

tidy3d/web/api/webapi.py
momchil-flex: Yeah I feel like cleaning up is worth it given that the config toggle is already available, but I might not be appreciating all the use cases.

marcorudolphflex: that's true, just removed it

tidy3d/web/api/webapi.py
marcorudolphflex: Indeed, that's a little bit redundant. I now assume a from_cache implicitely if task_id is None

docs/index.rst
marcorudolphflex: Probably it would make sense to first merge the new config and then integrate this PR in there.

tidy3d/web/cache.py
```diff
@@ -0,0 +1,532 @@
...
+    Returns LocalCache instance if enabled.
+    Returns None if use_cached=False or config-fetched 'enabled' is False.
+    Deletes old cache directory if existing.
+    """
+    global _CACHE
+
+    if use_cache is False or (use_cache is not True and not config.local_cache.enabled):
+        return None
+
+    cfg_dir = Path(config.local_cache.directory).absolute()
```
yaugenst-flex: I think `~/.tidy3d/cache` would not expand correctly with `absolute()`, this should be `.expanduser().resolve()`.

tidy3d/web/cache.py
```diff
@@ -0,0 +1,532 @@
...
+
+    if _CACHE is not None:
+        current_root = Path(_CACHE._root).absolute()
```
yaugenst-flex: Same here, not sure `absolute()` is what we want?

tidy3d/config/sections.py
```diff
@@ -381,6 +381,47 @@ def apply_web(config: WebConfig) -> None:
...
+        json_schema_extra={"persist": True},
+    )
+
+    max_entries: int = Field(
+        512,
+        title="Maximum cache entries",
+        description="Maximum number of cache entries. Set to 0 for no limit.",
+        ge=0,
+        json_schema_extra={"persist": True},
+    )
```
yaugenst-flex: Is this necessary? Seems like it should be optional? Typically you'd only care about the cache size I think.

tidy3d/config/sections.py
```diff
@@ -381,6 +381,47 @@ def apply_web(config: WebConfig) -> None:
...
+    """Settings controlling the optional local simulation cache."""
+
+    enabled: bool = Field(
+        False,
+        title="Enable cache",
+        description="Enable or disable the local simulation cache.",
+        json_schema_extra={"persist": True},
+    )
+
+    directory: str = Field(
```
yaugenst-flex: This should be typed as a `Path` probably? The `_expand_directory` validator would also be unnecessary then.

tidy3d/web/cache.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def __init__(self, directory: os.PathLike, max_entries: int, max_size_gb: float) -> None:
+        self.max_size_gb = max_size_gb
+        self.max_entries = max_entries
+        self._root = Path(directory).expanduser().resolve()
+        self._lock = threading.RLock()
+
+    @property
+    def config(self) -> LocalCacheConfig:
+        return self._config
```
yaugenst-flex: I think this is never assigned? Where does this come from?

yaugenst-flex (changes requested):

tidy3d/web/cache.py
marcorudolphflex: changed now the validator to use .expanduser().resolve(), so no canonicalization is necessary here

tidy3d/config/sections.py
marcorudolphflex: Was my first approach, but Path causes problems in serialization to toml, changed to str

tidy3d/web/cache.py
marcorudolphflex: is now deleted, survived previous refactoring...

tidy3d/config/sections.py
yaugenst-flex: i see. in that case i think it'd be more natural to have a field serializer like [here](https://github.com/flexcompute/tidy3d/blob/c52a9a59f22a9075ba769b495da290d7b88a733e/tidy3d/config/sections.py#L245-L249)

tidy3d/config/sections.py
```diff
@@ -381,6 +381,52 @@ def apply_web(config: WebConfig) -> None:
...
+        title="Maximum cache entries",
+        description="Maximum number of cache entries. Set to 0 for no limit.",
+        ge=0,
+        json_schema_extra={"persist": True},
+    )
+
+    @field_validator("directory", mode="before")
+    @classmethod
+    def _expand_directory(cls, value: Any) -> Path:
+        """Expand ~ and ensure directory path is pathlib oath."""
```
yaugenst-flex: typo here `oath` -> `path`

tidy3d/config/sections.py
```diff
@@ -381,6 +381,52 @@ def apply_web(config: WebConfig) -> None:
...
+        json_schema_extra={"persist": True},
+    )
+
+    max_entries: Optional[int] = Field(
+        None,
+        title="Maximum cache entries",
+        description="Maximum number of cache entries. Set to 0 for no limit.",
+        ge=0,
+        json_schema_extra={"persist": True},
+    )
```
yaugenst-flex, line 435: Ohh I'm just realizing the docstring, `0` is no limit. So maybe the `Optional` doesn't make too much sense after all? Since `toml` doesn't support null, perhaps `0` is better?

tidy3d/config/sections.py
```diff
@@ -381,6 +381,52 @@ def apply_web(config: WebConfig) -> None:
...
+    """Settings controlling the optional local simulation cache."""
+
+    enabled: bool = Field(
+        False,
+        title="Enable cache",
+        description="Enable or disable the local simulation cache.",
+        json_schema_extra={"persist": True},
+    )
+
+    directory: Path = Field(
```
yaugenst-flex: I think if you use pydantic's [FilePath](https://docs.pydantic.dev/2.2/usage/types/file_types/) then you wouldn't have to write a custom validator.

tidy3d/config/sections.py
```diff
@@ -381,6 +381,52 @@ def apply_web(config: WebConfig) -> None:
...
+        title="Cache directory",
+        description="Directory where cached artifacts are stored.",
+        json_schema_extra={"persist": True},
+    )
+
+    max_size_gb: float = Field(
+        10.0,
+        title="Maximum cache size (GB)",
+        description="Maximum cache size in gigabytes. Set to 0 for no size limit.",
+        ge=0.0,
```
yaugenst-flex: Let's use `PositiveFloat`

tidy3d/config/sections.py
```diff
@@ -381,6 +381,52 @@ def apply_web(config: WebConfig) -> None:
...
+        json_schema_extra={"persist": True},
+    )
+
+    max_entries: Optional[int] = Field(
+        None,
+        title="Maximum cache entries",
+        description="Maximum number of cache entries. Set to 0 for no limit.",
+        ge=0,
```
yaugenst-flex: `PositiveInt`

tidy3d/config/sections.py
marcorudolphflex: went for Directory path and added a validator that ensures existence (otherwise validation fail)

tidy3d/config/sections.py
marcorudolphflex: went for NonNegativeInt

tidy3d/config/sections.py
marcorudolphflex: went for NonNegativeFloat to support no limit on 0

tidy3d/config/sections.py
marcorudolphflex: went for NonNegativeInt to support no limit on 0

yaugenst-flex: oops yeah

yaugenst-flex (approved):
Thanks @marcorudolphflex LGTM! Can you make sure all previous comments are resolved and then we can merge.

tidy3d/config/sections.py
```diff
@@ -385,6 +386,21 @@ def apply_web(config: WebConfig) -> None:
...
+    base_override = os.getenv("TIDY3D_BASE_DIR")
+    if base_override:
+        base = Path(base_override).expanduser().resolve() / ".cache"
+    else:
+        xdg_cache = os.getenv("XDG_CACHE_HOME")
+        if xdg_cache:
+            base = Path(xdg_cache).expanduser().resolve()
+        else:
+            base = Path.home() / ".cache"
+    return (base / "tidy3d" / "simulations").resolve()
```
marcorudolphflex, line 402: in case we take the base_dir from TIDY3D_BASE_DIR, the path would be then `TIDY3D_BASE_DIR / .cache / tidy3d / simulations` with tidy3d in it again? Do we want that?

tidy3d/config/sections.py
```diff
@@ -381,6 +386,66 @@ def apply_web(config: WebConfig) -> None:
...
+
+    max_entries: NonNegativeInt = Field(
+        0,
+        title="Maximum cache entries",
+        description="Maximum number of cache entries. Set to 0 for no limit.",
+        json_schema_extra={"persist": True},
+    )
+
+    @field_validator("directory", mode="before")
+    def ensure_directory_exists(cls, v: PathLike) -> Path:
```
momchil-flex: Should this just be private too?

docs/index.rst
```diff
@@ -168,6 +168,13 @@ This will produce the following plot, which visualizes the electromagnetic field
 
 You can now postprocess simulation data using the same python session, or view the results of this simulation on our web-based `graphical user interface (GUI) <https://tidy3d.simulation.cloud>`_.
 
+.. tip::
+
+   Repeated runs of the same simulation can reuse solver results by enabling the optional
+   local cache: ``td.config.local_cache.enabled = True``. The cache location and limits are
+   configurable (see ``~/.tidy3d/config``), entries are checksum-validated, and you can clear
+   all stored artifacts with ``tidy3d.web.cache.clear()``.
```
momchil-flex: It still seems to me like a detailed explanation about usage and interaction with the server cache is missing.

For example even I currently don't understand some details, like what happens if I exceed the max entries or the max memory limit. Looking at the tests, it seems like it's first-in first-out as one would expect, but we should make this more explicit either here or in the corresponding Config fields.

tidy3d/config/sections.py
```diff
@@ -381,6 +386,66 @@ def apply_web(config: WebConfig) -> None:
...
+        xdg_cache = os.getenv("XDG_CACHE_HOME")
+        if xdg_cache:
+            base = Path(xdg_cache).expanduser().resolve()
+        else:
+            base = Path.home() / ".cache"
+    return (base / "tidy3d" / "simulations").resolve()
+
+
+@register_section("local_cache")
+class LocalCacheConfig(ConfigSection):
```
momchil-flex, line 406: How do these things enter our documentation @yaugenst-flex ? Does this need to be added somewhere in `docs/`, like the api reference?

tidy3d/config/sections.py
yaugenst-flex: yeah good point, the config dir actually has the same "issue". i aligned these two now to treat `TIDY3D_BASE_DIR` as the "tidy3d home" with config under `config/` and cache under `cache/`

momchil-flex (approved):
Thanks, this is much cleaner now!

My main comments are about making sure this is well documented for the user.

tidy3d/config/sections.py
yaugenst-flex: yes there is an open docs PR for the config system, although the local cache doesnt have an entry there yet. we can add it there i think https://github.com/flexcompute/tidy3d/pull/2917

yaugenst-flex: i wonder though if we should maintain a list of config settings in the docs too? if you look at `~/.config/tidy3d/config.toml` you'll see that it's "self-documenting" (the toml includes comments for each setting)

momchil-flex: Well, many users wouldn't think to look there, but at least if we say this in our docs then maybe yeah. But any downside to having it explicitly in the docs?

yaugenst-flex: just maintenance overhead :smile:

momchil-flex: The `toml` file only shows things that are saved though. There's a bunch of other options that the user could set (even if not persisted) that are not documented there.

docs/index.rst
marcorudolphflex: good point, added some details and mentioned the server-side caching

tidy3d/config/sections.py
yaugenst-flex: yeah i'll address that in the docs PR

PR #2905, opened 2025-10-17
refactor(RF)!: edge singularity correction defaults to True

weiliangjin2021:
<h2>Greptile Overview</h2>

Updated On: 2025-10-17 17:29:07 UTC

### Greptile Summary

This PR changes the default behavior for edge singularity correction in PEC (Perfect Electric Conductor) simulations from `False` to `True`. The change affects the `edge_singularity_correction` field in the `PECConformal` class, which controls whether field correction models are applied at metallic edges where field singularities occur. This correction is particularly important for straight edges aligned with primal grids having wedge angles of 0 or 90 degrees.

The change improves simulation accuracy by default, especially for RF and microwave applications where PEC boundaries are common. By enabling this correction by default, users will automatically get more physically accurate results without needing to understand the technical details of edge singularity handling. The change is properly documented in the CHANGELOG as a breaking change, acknowledging that it affects default behavior and may impact existing simulations that relied on the previous default.

### Important Files Changed

<details>

| CHANGELOG.md | 5/5 | Documents the breaking change for edge singularity correction defaulting to True |
| tidy3d/components/subpixel_spec.py | 5/5 | Changes default value of `edge_singularity_correction` from False to True in PECConformal class |

</details>

### Confidence score: 5/5

- This PR is safe to merge with minimal risk as it's a straightforward default value change with proper documentation
- Score reflects the simple nature of the change, clear documentation, and appropriate breaking change labeling
- No files require special attention - both changes are well-implemented and documented

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Simulation
    participant SubpixelSpec
    participant PECConformal
    participant Solver

    User->>Simulation: "Create simulation with PEC structures"
    Simulation->>SubpixelSpec: "Initialize default SubpixelSpec()"
    SubpixelSpec->>PECConformal: "Create PECConformal() for pec field"
    Note over PECConformal: edge_singularity_correction=True (new default)
    PECConformal->>SubpixelSpec: "Return PECConformal instance"
    SubpixelSpec->>Simulation: "Return configured SubpixelSpec"
    
    User->>Simulation: "Run simulation"
    Simulation->>Solver: "Submit with subpixel configurations"
    Solver->>PECConformal: "Apply edge singularity correction at metal edges"
    Note over Solver: Field correction model applied at metallic edges where singularity occurs
    Solver->>Simulation: "Return results with improved accuracy"
    Simulation->>User: "Simulation completed"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unified run submission API: `web.run(...)` is now a container-aware wrapper that accepts a single simulation or arbitrarily nested containers (`list`, `tuple`, `dict` values) and returns results in the same shape.
 - `web.Batch(ComponentModeler)` and `web.Job(ComponentModeler)` native support
 - Simulation data of batch jobs are now automatically downloaded upon their individual completion in `Batch.run()`, avoiding waiting for the entire batch to reach completion.
+- Edge singularity correction at PEC and lossy metal edges defaults to `True`.
 
 ### Fixed
 - More robust `Sellmeier` and `Debye` material model, and prevent very large pole parameters in `PoleResidue` material model.
--- a/schemas/EMESimulation.json
+++ b/schemas/EMESimulation.json
@@ -8768,7 +8768,7 @@
           "type": "object"
         },
         "edge_singularity_correction": {
-          "default": false,
+          "default": true,
           "type": "boolean"
         },
         "timestep_reduction": {
@@ -11246,7 +11246,7 @@
         "lossy_metal": {
           "default": {
             "attrs": {},
-            "edge_singularity_correction": false,
+            "edge_singularity_correction": true,
             "timestep_reduction": 0.0,
             "type": "SurfaceImpedance"
           },
@@ -11294,7 +11294,7 @@
         "pec": {
           "default": {
             "attrs": {},
-            "edge_singularity_correction": false,
+            "edge_singularity_correction": true,
             "timestep_reduction": 0.3,
             "type": "PECConformal"
           },
@@ -11357,7 +11357,7 @@
           "type": "object"
         },
         "edge_singularity_correction": {
-          "default": false,
+          "default": true,
           "type": "boolean"
         },
         "timestep_reduction": {
@@ -12507,7 +12507,7 @@
         },
         "lossy_metal": {
           "attrs": {},
-          "edge_singularity_correction": false,
+          "edge_singularity_correction": true,
           "timestep_reduction": 0.0,
           "type": "SurfaceImpedance"
         },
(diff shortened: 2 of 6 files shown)


weiliangjin2021:
Not sure what is the correction commit specification: it's not a fix, chore, feat.

dbochkov-flexcompute (approved):

weiliangjin2021:
@yaugenst could you help on the errored branch name linting?

yaugenst-flex:
> @yaugenst could you help on the errored branch name linting?

Ah, it's because we currently actually require the branch name to start with the Jira ticket ID (if it's not a `hotfix/` branch). I think we can relax this constraint, we'll make a PR, and then this should pass once rebased.

yaugenst-flex:
Should be fixed in https://github.com/flexcompute/tidy3d/pull/2907

weiliangjin2021:
Thanks @yaugenst-flex ! I also got this error on the commit type
https://github.com/flexcompute/tidy3d/actions/runs/18658150131/job/53195297703 . What should be the type for a commit that changes the default value?

yaugenst-flex:
> Thanks @yaugenst-flex ! I also got this error on the commit type [flexcompute/tidy3d/actions/runs/18658150131/job/53195297703](https://github.com/flexcompute/tidy3d/actions/runs/18658150131/job/53195297703) . What should be the type for a commit that changes the default value?

I think you have to get rid of the exclamation mark

weiliangjin2021:
> > Thanks @yaugenst-flex ! I also got this error on the commit type [flexcompute/tidy3d/actions/runs/18658150131/job/53195297703](https://github.com/flexcompute/tidy3d/actions/runs/18658150131/job/53195297703) . What should be the type for a commit that changes the default value?
> 
> I think you have to get rid of the exclamation mark

I was following [this](https://www.conventionalcommits.org/en/v1.0.0/) to use exclamation to indicate breaking change.

yaugenst-flex:
I see. I wasn't even aware of that! The reason is that our `.commitlint.rc` does not strictly follow the commitlint spec (via `@commitlint/config-conventional` it could, but it would need to pull in node dependencies, which is why we decided not to go that route). so in this commit message there are currently two failures: we currently require all lowercase for the type, so it should be `(rf)` instead of `(RF)`, and we don't support the exclamation mark.
i'll create a pr tomorrow to fix both of these, but in the meantime i think if you rename then we could merge this?

weiliangjin2021:
> I see. I wasn't even aware of that! The reason is that our `.commitlint.rc` does not strictly follow the commitlint spec (via `@commitlint/config-conventional` it could, but it would need to pull in node dependencies, which is why we decided not to go that route). so in this commit message there are currently two failures: we currently require all lowercase for the type, so it should be `(rf)` instead of `(RF)`, and we don't support the exclamation mark. i'll create a pr tomorrow to fix both of these, but in the meantime i think if you rename then we could merge this?

Thanks, will just rename it and use the other style for now.

PR #2914, opened 2025-10-20
feat: custom source time in modeler

weiliangjin2021:
<h2>Greptile Overview</h2>

Updated On: 2025-10-21 00:00:39 UTC

### Greptile Summary

This PR adds the ability to specify custom source time dependence in component modelers (`ModalComponentModeler` and `TerminalComponentModeler`) through a new optional `custom_source_time` parameter in the `AbstractComponentModeler` base class. When provided, the custom source time overrides the default `GaussianPulse` behavior that was previously constructed automatically from frequency and bandwidth parameters. The feature integrates with the existing component modeler framework by checking `if self.custom_source_time is not None` in the `to_source` methods and using the custom value when present. This change maintains backward compatibility since the parameter defaults to `None`, preserving existing behavior for all code that doesn't explicitly provide a custom source time. The JSON schema for `TerminalComponentModeler` has been updated to reflect the three supported source time types: `ContinuousWave`, `CustomSourceTime`, and `GaussianPulse`.

### Important Files Changed

<details>

|----------|-------|-----------|
| tidy3d/plugins/smatrix/component_modelers/base.py | 5/5 | Added `custom_source_time` optional field to base class with proper type hints and documentation |
| tidy3d/plugins/smatrix/component_modelers/modal.py | 5/5 | Implemented custom source time logic in `to_source` method using conditional check |
| tidy3d/plugins/smatrix/component_modelers/terminal.py | 3/5 | Missing in PR context but likely contains similar implementation to modal.py |
| schemas/TerminalComponentModeler.json | 5/5 | Updated JSON schema to include `custom_source_time` field with three source time type options |
| CHANGELOG.md | 5/5 | Added clear changelog entry documenting the new feature |
| tests/test_plugins/smatrix/test_component_modeler.py | 1/5 | Test uses invalid `ContinuousWave` with non-zero `fwidth`, which violates source constraints |
| tests/test_plugins/smatrix/test_terminal_component_modeler.py | 2/5 | Test has same issue - uses `ContinuousWave` with `fwidth=1e8` instead of required `fwidth=0` |

</details>

### Confidence score: 2/5

- This PR requires careful review before merging due to critical test implementation issues that will likely cause validation errors or incorrect behavior
- Score is lowered because both new tests create invalid `ContinuousWave` sources with non-zero `fwidth` values (test_component_modeler.py uses `fwidth=1e13`, test_terminal_component_modeler.py uses `fwidth=1e8`), but `ContinuousWave` sources are monochromatic by design and should have `fwidth=0`; the core feature implementation in base.py and modal.py appears sound, but terminal.py is missing from context so cannot be verified; tests should use `GaussianPulse` instead or set `fwidth=0` for `ContinuousWave`
- Pay close attention to tests/test_plugins/smatrix/test_component_modeler.py and tests/test_plugins/smatrix/test_terminal_component_modeler.py which both contain invalid source time configurations that need correction

### DIFF
--- a/tidy3d/plugins/smatrix/component_modelers/base.py
+++ b/tidy3d/plugins/smatrix/component_modelers/base.py
@@ -8,9 +8,10 @@ from typing import TYPE_CHECKING, Literal, Optional, Union
 import pydantic.v1 as pd
 
 from tidy3d.components.autograd.constants import MAX_NUM_ADJOINT_PER_FWD
-from tidy3d.components.base import Tidy3dBaseModel, cached_property
+from tidy3d.components.base import Tidy3dBaseModel, cached_property, skip_if_fields_missing
 from tidy3d.components.geometry.utils import _shift_value_signed
 from tidy3d.components.simulation import Simulation
+from tidy3d.components.source.time import SourceTimeType
 from tidy3d.components.types import Complex, FreqArray
 from tidy3d.components.validators import (
     assert_unique_names,
@@ -94,6 +95,12 @@ class AbstractComponentModeler(ABC, Tidy3dBaseModel):
         "matrix element. If all elements of a given column of the scattering matrix are defined "
         "by ``element_mappings``, the simulation corresponding to this column is skipped automatically.",
     )
+    custom_source_time: Optional[SourceTimeType] = pd.Field(
+        None,
+        title="Custom Source Time",
+        description="If provided, this will be used as specification of the source time-dependence in simulations. "
+        "Otherwise, a default source time will be constructed.",
+    )
 
     @pd.validator("simulation", always=True)
     def _sim_has_no_sources(cls, val):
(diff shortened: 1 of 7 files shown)


tidy3d/plugins/smatrix/component_modelers/base.py
```diff
@@ -94,6 +95,12 @@ class AbstractComponentModeler(ABC, Tidy3dBaseModel):
         "matrix element. If all elements of a given column of the scattering matrix are defined "
         "by ``element_mappings``, the simulation corresponding to this column is skipped automatically.",
     )
+    custom_source_time: Optional[SourceTimeType] = pd.Field(
```
dmarek-flex, line 98: Is it possible to add validation or warning in the case that the `freqs` field is outside (or partially outside) the frequency range of the `custom_ source_time`?

dmarek-flex (approved):

tidy3d/plugins/smatrix/component_modelers/base.py
weiliangjin2021: yeah, good point; even though our default source_time right now might still trigger those warnings before broadband is merged.

weiliangjin2021: added

Issue #2530, opened 2025-05-31
Mode solver sweeps using interpolation


dbochkov-flexcompute:
**Is your feature request related to a problem? Please describe.**
Sometimes it is of interest to obtain the waveguide mode field profiles and/or propagation constants at a large number of frequency points. Given that waveguide modes often exhibit smooth frequency dependence, it could be much faster and still accurate to solve for waveguide modes only at a smaller number of selected frequencies and then interpolate found solutions onto frequency points in-between, as opposed to directly solving mode equations at each frequency point.

**Describe the solution you'd like**
Various types of interpolations such as piecewise linear, quadratic, spline, etc can be considered. In the first version of such a feature it probably makes most sense to use Chebyshev interpolation given its high accuracy. One approach to implement this consists in adding a field `sweep_spec` in the `ModeSolver` class, that could take values of `ModeSweepSpecType = Union[DirectSolveSpec, ChebyshevInterpSpec]` (default: `DirectSolveSpec()`). Internally, `ModeSolver` should wrap this part of code https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/mode/mode_solver.py#L463-L464 such that if  `sweep_spec` is an instance of class `ChebyshevInterpSpec`:
- it detects the range of frequency sample points where the mode solutions are requested
- create an auxiliary `ModeSolver` instance that differs from the original one only by value of field `freqs`, which now would be chebyshev nodes in the detected frequency range (the number of chebyshev nodes to use should be a field in `ChebyshevInterpSpec` class)
- retrieves `ModeSolverData` from this auxiliary `ModeSolver`
- applies mode tracking to  `ModeSolverData` to facilitate correct interpolation
- interpolates all quantities (propagation index, fields, grid correction factors, etc) onto the original frequency sample points and constructs a new `ModeSolverData` based on these interpolated values.

**Additional context**
- https://arxiv.org/pdf/1810.04282
- https://docs.simulation.cloud/projects/tidy3d/en/latest/notebooks/ModeSolver.html#Mode-tracking
- https://docs.simulation.cloud/projects/tidy3d/en/latest/notebooks/ModalSourcesMonitors.html#Waveguide-junction-(using-broadband-source)

doddgray:
[...]

PR #2846, opened 2025-09-25
FXC-3724 support multimodal waveports

dmarek-flex:
<h2>Greptile Overview</h2>

Updated On: 2025-10-16 20:12:05 UTC

### Greptile Summary

This PR implements multimodal waveport support in the S-matrix component modeler, enabling waveguide ports to handle multiple propagating modes simultaneously. The changes refactor the WavePort class to use `MicrowaveModeSpec` instead of manual impedance calculations, introduce a network indexing system that creates separate entries for each port-mode combination, and update the analysis pipeline to work with `NetworkIndex` identifiers rather than simple port references.

Key architectural changes include replacing the single `mode_index` field with a multimodal approach, switching from `TerminalPortType` to `NetworkIndex` for port identification, and implementing caching optimizations for voltage/current calculations to avoid redundant computations across modes of the same port. The refactoring leverages existing microwave components for transmission line calculations while maintaining backward compatibility with lumped ports.

### Important Files Changed

<details><summary>Changed Files</summary>

| Filename | Score | Overview |
|----------|-------|----------|
| `tidy3d/plugins/smatrix/ports/wave.py` | 4/5| Major refactoring to support multimodal operation using MicrowaveModeSpec and removing manual impedance calculations |
| `tidy3d/plugins/smatrix/component_modelers/terminal.py` | 4/5 | Core network indexing changes to create separate entries for each port-mode combination |
| `tidy3d/plugins/smatrix/analysis/terminal.py` | 4/5 | Added caching optimization for voltage/current calculations and mode_index parameter support |
| `tidy3d/plugins/smatrix/data/terminal.py` | 4/5 | Updated method signature from TerminalPortType to NetworkIndex for multimodal support |
| `tidy3d/plugins/smatrix/component_modelers/base.py` | 5/5 | Removed automatic mode_index assignment to require explicit specification |
| `tidy3d/plugins/smatrix/analysis/antenna.py` | 4/5 | Refactored to use NetworkIndex-based processing instead of port-based iteration |
| `tests/test_plugins/smatrix/test_terminal_component_modeler.py` | 3/5 | Updated test suite with incomplete validation migration and duplicate code blocks |
| `tests/test_plugins/smatrix/terminal_component_modeler_def.py` | 4/5 | Updated WavePort configuration to use MicrowaveModeSpec with CustomImpedanceSpec |
| `tests/utils.py` | 4/5 | Added data generator functions for microwave monitor data with transmission line support |

</details>

### Confidence score: 3/5

[...]

PR #2923, opened 2025-10-23
feat: downsample frequencies in mode monitors (FXC-3351)

dbochkov-flexcompute:
Addresses https://github.com/flexcompute/tidy3d/issues/2530. 

Notebooks re-run examples:
- photonics: https://github.com/flexcompute/tidy3d-notebooks/blob/d6109a7182701e0e487e1cf4d35434f5063c7bca/RingResonator.ipynb (1001 mode solves -> 21 mode solves with no visible change)
- RF: https://github.com/flexcompute/tidy3d-notebooks/blob/d6109a7182701e0e487e1cf4d35434f5063c7bca/DifferentialStripline.ipynb

Still not sure about making it default in WavePorts

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR implements frequency downsampling/interpolation for mode monitors and mode solvers to reduce computational cost in broadband simulations. The implementation introduces a new `ModeInterpSpec` class with three interpolation methods (linear, cubic, Chebyshev polynomial), allowing modes to be computed at fewer strategically-chosen frequencies and interpolated to the full frequency set.

**Key Changes:**
- New `ModeInterpSpec` class with `UniformSampling`, `ChebSampling`, and `CustomSampling` strategies for flexible frequency selection
- Integration throughout the mode computation pipeline: `ModeSpec`, `AbstractModeMonitor`, `ModeSolver`, and `ModeSolverData`
- Comprehensive interpolation implementation in `ModeSolverData.interp()` supporting all three methods with proper handling of complex-valued data
- Requires mode tracking to be enabled (`mode_spec.sort_spec.track_freq`) to maintain consistent mode ordering across frequencies
- Extensive test coverage with 965 lines of well-structured tests
- Extrapolation warnings when interpolating outside the computed frequency range

**Notable Implementation Details:**
- Chebyshev nodes use barycentric formula for optimal polynomial interpolation
- Complex data is interpolated by treating real and imaginary parts independently
- `reduce_data` flag allows storing only sampled frequencies to minimize storage costs
- Grid correction factors can be recalculated or interpolated based on user preference

- This PR is generally safe to merge with minor issues that should be addressed
- Score reflects well-designed core functionality with comprehensive testing, but deducted one point for: (1) missing upper bound validation on `num_points` in `ModeInterpSpec` that could exceed total frequencies, (2) f-string usage in warning message violating style guide rule 9b45957a-2d0d-4c24-a5d3-eef6f721cbfd, (3) potential edge case with single-frequency inputs not explicitly handled, and (4) extrapolation tolerance using multiplicative factor may behave unexpectedly near zero frequencies
- Pay closest attention to `tidy3d/components/mode_spec.py` (validation gaps) and `tidy3d/components/monitor.py` (style guide violation)

File Analysis

| tidy3d/components/mode_spec.py | 4/5 | Introduces `ModeInterpSpec` with three interpolation methods and sampling strategies; validation ensures `num_points` meets method requirements but lacks upper bound check and single-frequency edge case handling |
| tidy3d/components/monitor.py | 4/5 | Adds `interp_spec` field to `AbstractModeMonitor` with tracking requirement validation; f-string usage in warning violates style guide |
| tidy3d/components/mode/mode_solver.py | 5/5 | Implements core interpolation logic in `ModeSolver`; debug print statements have been removed (no longer present in latest commit) |
| tidy3d/components/data/monitor_data.py | 4/5 | Adds comprehensive `interp()` method to `ModeSolverData` supporting all interpolation methods; extrapolation tolerance uses multiplicative factor which may be problematic near zero frequencies |
| tests/test_components/test_mode_interp.py | 5/5 | Comprehensive test coverage (965 lines) for all interpolation functionality with well-structured test cases |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant ModeSolver
    participant ModeInterpSpec
    participant ModeSolverData
    participant compute_modes

    User->>ModeSolver: Create with interp_spec
    ModeSolver->>ModeInterpSpec: Validate num_points
    ModeInterpSpec-->>ModeSolver: Validation complete
    
    User->>ModeSolver: Call data_raw property
    
    alt interp_spec is None
        ModeSolver->>compute_modes: Solve at all frequencies
        compute_modes-->>ModeSolver: Return raw data
        ModeSolver-->>User: ModeSolverData (all freqs)
    else interp_spec provided
        ModeSolver->>ModeInterpSpec: sampling_points(freqs)
        
        alt method is linear/cubic
            ModeInterpSpec->>ModeInterpSpec: Generate uniform points
        else method is cheb
            ModeInterpSpec->>ModeInterpSpec: Generate Chebyshev nodes
        end
        
        ModeInterpSpec-->>ModeSolver: Return freqs_reduced
        ModeSolver->>ModeSolver: Create mode_solver_reduced
        ModeSolver->>compute_modes: Solve at reduced freqs
        compute_modes-->>ModeSolver: Return data_reduced
        
        ModeSolver->>ModeSolverData: interp(freqs, method)
        
        alt method is cheb
            ModeSolverData->>ModeSolverData: Barycentric interpolation
        else method is linear/cubic
            ModeSolverData->>ModeSolverData: xarray interpolation
        end
        
        ModeSolverData-->>ModeSolver: Interpolated data
        ModeSolver-->>User: ModeSolverData (all freqs)
    end
```

### DIFF
--- a/tidy3d/__init__.py
+++ b/tidy3d/__init__.py
@@ -339,7 +339,14 @@ from .components.mode.data.sim_data import ModeSimulationData
 from .components.mode.simulation import ModeSimulation
 
 # modes
-from .components.mode_spec import ModeSortSpec, ModeSpec
+from .components.mode_spec import (
+    ChebSampling,
+    CustomSampling,
+    ModeInterpSpec,
+    ModeSortSpec,
+    ModeSpec,
+    UniformSampling,
+)
 
 # monitors
 from .components.monitor import (
@@ -524,6 +531,7 @@ __all__ = [
     "ChargeDataArray",
     "ChargeInsulatorMedium",
     "ChargeToleranceSpec",
+    "ChebSampling",
     "ClipOperation",
     "CoaxialLumpedResistor",
     "CompositeCurrentIntegral",
@@ -558,6 +566,7 @@ __all__ = [
     "CustomLorentz",
     "CustomMedium",
     "CustomPoleResidue",
+    "CustomSampling",
     "CustomSellmeier",
     "CustomSourceTime",
     "Cylinder",
@@ -702,6 +711,7 @@ __all__ = [
     "ModeAmpsDataArray",
     "ModeData",
     "ModeIndexDataArray",
+    "ModeInterpSpec",
     "ModeMonitor",
     "ModeSimulation",
     "ModeSimulationData",
@@ -809,6 +819,7 @@ __all__ = [
     "UniformCurrentSource",
     "UniformGrid",
     "UniformHeatSource",
+    "UniformSampling",
     "UniformUnstructuredGrid",
     "UnsteadyHeatAnalysis",
     "UnsteadySpec",
--- a/tidy3d/components/data/dataset.py
+++ b/tidy3d/components/data/dataset.py
@@ -3,14 +3,14 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Optional, Union, get_args
+from typing import Any, Callable, Literal, Optional, Union, get_args
 
 import numpy as np
 import pydantic.v1 as pd
 import xarray as xr
 
 from tidy3d.components.base import Tidy3dBaseModel
-from tidy3d.components.types import Axis, xyz
+from tidy3d.components.types import Axis, FreqArray, xyz
 from tidy3d.constants import C_0, PICOSECOND_PER_NANOMETER_PER_KILOMETER, UnitScaling
 from tidy3d.exceptions import DataError
 from tidy3d.log import log
@@ -48,6 +48,139 @@ class Dataset(Tidy3dBaseModel, ABC):
             if isinstance(data, DataArray):
                 data_arrs[key] = data
         return data_arrs
+
+
+class FreqDataset(Dataset, ABC):
+    """Abstract base class for objects that store collections of `:class:`.DataArray`s."""
+
+    def _interp_in_freq_update_dict(
+        self,
+        freqs: FreqArray,
+        method: Literal["linear", "cubic", "poly"] = "linear",
+        assume_sorted: bool = False,
+    ) -> dict[str, DataArray]:
+        """Interpolate mode data to new frequency points.
+
+        Interpolates all stored mode data (effective indices, field components, group indices,
+        and dispersion) from the current frequency grid to a new set of frequencies. This is
+        useful for obtaining mode data at many frequencies from computations at fewer frequencies,
+        when modes vary smoothly with frequency.
+
+        Parameters
+        ----------
+        freqs : FreqArray
+            New frequency points to interpolate to. Should generally span a similar range
+            as the original frequencies to avoid extrapolation.
+        method : Literal["linear", "cubic", "poly"]
+            Interpolation method. ``"linear"`` for linear interpolation (requires 2+ source
+            frequencies), ``"cubic"`` for cubic spline interpolation (requires 4+ source
+            frequencies), ``"poly"`` for polynomial interpolation using barycentric
+            formula (requires 3+ source frequencies).
+            For complex-valued data, real and imaginary parts are interpolated independently.
+        assume_sorted: bool = False,
+            Whether to assume the frequency points are sorted.
+
+        Returns
+        -------
+        ModeSolverData
+            New :class:`ModeSolverData` object with data interpolated to the requested frequencies.
+
+        Note
+        ----
+            Interpolation assumes modes vary smoothly with frequency. Results may be inaccurate
+            near mode crossings or regions of rapid mode variation. Use frequency tracking
+            (``mode_spec.sort_spec.track_freq``) to help maintain mode ordering consistency.
+
+            For polynomial interpolation, source frequencies at Chebyshev nodes provide
+            optimal accuracy within the frequency range.
+
+        Example
+        -------
+        >>> # Compute modes at 5 frequencies
+        >>> import numpy as np
+        >>> freqs_sparse = np.linspace(1e14, 2e14, 5)
+        >>> # ... create mode_solver and compute modes ...
+        >>> # mode_data = mode_solver.solve()
+        >>> # Interpolate to 50 frequencies
+        >>> freqs_dense = np.linspace(1e14, 2e14, 50)
+        >>> # mode_data_interp = mode_data.interp(freqs=freqs_dense, method='linear')
+        """
+        freqs = np.array(freqs)
+
+        modify_data = {}
+        for key, data in self.data_arrs.items():
+            modify_data[key] = self._interp_dataarray_in_freq(data, freqs, method, assume_sorted)
+
+        return modify_data
+
+    @staticmethod
+    def _interp_dataarray_in_freq(
+        data: DataArray,
+        freqs: FreqArray,
+        method: Literal["linear", "cubic", "poly", "nearest"],
+        assume_sorted: bool = False,
+    ) -> DataArray:
+        """Interpolate a DataArray along the frequency coordinate.
+
+        Parameters
+        ----------
+        data : DataArray
+            Data array to interpolate. Must have a frequency coordinate ``"f"``.
+        freqs : FreqArray
+            New frequency points.
+        method : Literal["linear", "cubic", "poly", "nearest"]
+            Interpolation method (``"linear"``, ``"cubic"``, ``"poly"``, or ``"nearest"``).
+            For ``"poly"``, uses barycentric formula for polynomial interpolation.
+        assume_sorted: bool = False,
+            Whether to assume the frequency points are sorted.
+
+        Returns
+        -------
+        DataArray
+            Interpolated data array with the same structure but new frequency points.
+        """
+        # Map 'poly' to xarray's 'barycentric' method
+        xr_method = "barycentric" if method == "poly" else method
+
+        # Use xarray's built-in interpolation
+        # For complex data, this automatically interpolates real and imaginary parts
+        interp_kwargs = {"method": xr_method}
+
+        if method == "nearest":
+            return data.sel(f=freqs, method="nearest")
+        else:
+            if method != "poly":
+                interp_kwargs["kwargs"] = {"fill_value": "extrapolate"}
+            return data.interp(f=freqs, assume_sorted=assume_sorted, **interp_kwargs)
+
+
+class ModeFreqDataset(FreqDataset, ABC):
+    """Abstract base class for objects that store collections of `:class:`.DataArray`s."""
+
+    def _apply_mode_reorder(self, sort_inds_2d):
+        """Apply a mode reordering along mode_index for all frequency indices.
+
+        Parameters
+        ----------
+        sort_inds_2d : np.ndarray
+            Array of shape (num_freqs, num_modes) where each row is the
+            permutation to apply to the mode_index for that frequency.
+        """
+        num_freqs, num_modes = sort_inds_2d.shape
+        modify_data = {}
+        for key, data in self.data_arrs.items():
+            if "mode_index" not in data.dims or "f" not in data.dims:
+                continue
+            dims_orig = data.dims
+            f_coord = data.coords["f"]
+            slices = []
+            for ifreq in range(num_freqs):
+                sl = data.isel(f=ifreq, mode_index=sort_inds_2d[ifreq])
+                slices.append(sl.assign_coords(mode_index=np.arange(num_modes)))
+            # Concatenate along the 'f' dimension name and then restore original frequency coordinates
+            data = xr.concat(slices, dim="f").assign_coords(f=f_coord).transpose(*dims_orig)
+            modify_data[key] = data
+        return self.updated_copy(**modify_data)
 
 
 class AbstractFieldDataset(Dataset, ABC):
@@ -492,7 +625,7 @@ class AuxFieldTimeDataset(AuxFieldDataset):
     )
 
 
-class ModeSolverDataset(ElectromagneticFieldDataset):
+class ModeSolverDataset(ElectromagneticFieldDataset, ModeFreqDataset):
     """Dataset storing scalar components of E and H fields as a function of freq. and mode_index.
 
     Example
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -18,7 +18,7 @@ from tidy3d.components.base import cached_property, skip_if_fields_missing
 from tidy3d.components.base_sim.data.monitor_data import AbstractMonitorData
 from tidy3d.components.grid.grid import Coords, Grid
 from tidy3d.components.medium import Medium, MediumType
-from tidy3d.components.mode_spec import ModeSortSpec
+from tidy3d.components.mode_spec import ModeSortSpec, ModeSpec
 from tidy3d.components.monitor import (
     AuxFieldTimeMonitor,
     DiffractionMonitor,
@@ -45,8 +45,10 @@ from tidy3d.components.types import (
     ArrayFloat1D,
     ArrayFloat2D,
     Coordinate,
+    Direction,
     EMField,
     EpsSpecType,
+    FreqArray,
     Numpy,
     PolarizationBasis,
     Size,
@@ -78,6 +80,7 @@ from .data_array import (
     MixedModeDataArray,
     ModeAmpsDataArray,
     ModeDispersionDataArray,
+    ModeIndexDataArray,
     ScalarFieldDataArray,
     ScalarFieldTimeDataArray,
     TimeDataArray,
@@ -103,6 +106,7 @@ AXIAL_RATIO_CAP = 100
 MIN_ANGULAR_SAMPLES_SPHERE = 10
 # Threshold for cos(theta) to avoid unphysically large amplitudes near grazing angles
 COS_THETA_THRESH = 1e-5
+MODE_INTERP_EXTRAPOLATION_TOLERANCE = 1e-2
 
 
 class MonitorData(AbstractMonitorData, ABC):
@@ -1672,11 +1676,11 @@ class ModeData(ModeSolverDataset, ElectromagneticFieldData):
     )
 
     @pd.validator("eps_spec", always=True)
-    @skip_if_fields_missing(["monitor"])
+    @skip_if_fields_missing(["n_complex"])
     def eps_spec_match_mode_spec(cls, val, values):
         """Raise validation error if frequencies in eps_spec does not match frequency list"""
         if val:
-            mode_data_freqs = values["monitor"].freqs
+            mode_data_freqs = values["n_complex"].coords["f"].values
             if len(val) != len(mode_data_freqs):
                 raise ValidationError(
                     "eps_spec must be provided at the same frequencies as mode solver data."
@@ -1719,7 +1723,8 @@ class ModeData(ModeSolverDataset, ElectromagneticFieldData):
         """
         if len(self.field_components) == 0:
             return self.copy()
-        num_freqs = len(self.monitor.freqs)
+
+        num_freqs = len(self.monitor._stored_freqs)
         num_modes = self.monitor.mode_spec.num_modes
 
         if track_freq == "lowest":
@@ -1759,7 +1764,7 @@ class ModeData(ModeSolverDataset, ElectromagneticFieldData):
                 # Get next frequency to sort
                 data_to_sort = data_expanded._isel(f=[freq_id])
                 # Assign to the base frequency so that outer_dot will compare them
-                data_to_sort = data_to_sort._assign_coords(f=[self.monitor.freqs[f0_ind]])
+                data_to_sort = data_to_sort._assign_coords(f=[self.monitor._stored_freqs[f0_ind]])
 
                 # Compute "sorting w.r.t. to neighbor" and overlap values
                 sorting_one_mode, amps_one_mode = data_template._find_ordering_one_freq(
@@ -1777,8 +1782,8 @@ class ModeData(ModeSolverDataset, ElectromagneticFieldData):
                 for mode_ind in list(np.nonzero(overlap[freq_id, :] < overlap_thresh)[0]):
                     log.warning(
                         f"Mode '{mode_ind}' appears to undergo a discontinuous change "
-                        f"between frequencies '{data_expanded.monitor.freqs[freq_id]}' "
-                        f"and '{data_expanded.monitor.freqs[freq_id - step]}' "
+                        f"between frequencies '{self.monitor._stored_freqs[freq_id]}' "
+                        f"and '{self.monitor._stored_freqs[freq_id - step]}' "
                         f"(overlap: '{overlap[freq_id, mode_ind]:.2f}')."
                     )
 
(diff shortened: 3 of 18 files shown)


tidy3d/components/mode_spec.py
```diff
@@ -86,6 +87,117 @@ class ModeSortSpec(Tidy3dBaseModel):
...
+
+        Example
+        -------
+        >>> import numpy as np
+        >>> freqs = np.linspace(1e14, 2e14, 100)
+        >>> interp_spec = ModeInterpSpec(num_points=10, method='cheb')
+        >>> sampling_freqs = interp_spec.sampling_points(freqs)
+        """
+        freqs_array = np.asarray(freqs)
+        f_min, f_max = float(freqs_array.min()), float(freqs_array.max())
```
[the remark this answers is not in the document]
dbochkov-flexcompute: this shouldn't happen because in this case interp_spec will be bypassed

tidy3d/__init__.py
```diff
@@ -693,6 +693,7 @@ def set_logging_level(level: str) -> None:
     "ModeAmpsDataArray",
     "ModeData",
     "ModeIndexDataArray",
+    "ModeInterpSpec",
```
momchil-flex, line 714: need to also add it in "__all__" later on in the file

tidy3d/components/data/monitor_data.py
```diff
@@ -2388,6 +2391,211 @@ def normalize(self, source_spectrum_fn: Callable[[float], complex]) -> ModeSolve
...
+
+        # Handle eps_spec if present - use nearest neighbor interpolation
+        if self.eps_spec is not None:
+            update_dict["eps_spec"] = list(
+                self._interp_dataarray(
+                    FreqDataArray(self.eps_spec, coords={"f": self.monitor.freqs}),
+                    freqs,
+                    "nearest",
+                ).data
+            )
```
momchil-flex, line 2651: In my sort_spec modification I introduced a better way to iterate all relevant arrays without the risk of forgetting something (or forgetting to add it if added in the future): https://github.com/flexcompute/tidy3d/blob/19df1ca082bbfefa4cd70f7c0b81e8f242be43d0/tidy3d/components/data/monitor_data.py#L2222-L2233

tidy3d/components/mode_spec.py
```diff
@@ -86,6 +87,117 @@ class ModeSortSpec(Tidy3dBaseModel):
...
+    --------
+
+    :class:`ModeSolver`:
+        Mode solver that can use this specification for efficient broadband computation.
+
+    :class:`ModeSolverMonitor`:
+        Monitor that can use this specification to reduce mode computation cost.
+
+    :class:`ModeMonitor`:
+        Monitor that can use this specification to reduce mode computation cost.
```
momchil-flex, line 261: My first thought was that the interp spec belongs in `ModeSpec`. 

However, reading through I think I see why you took this route. We might want to set different default values in different objects, which could get quite awkward if we only have a single shared place to set the default value. Furthermore, `ModeSource`-s should have this off (either error or ignore it if set) as they already have their `num_freqs` for Chebyshev interpolation.

On the other hand, I'm not a huge fan of having it outside of `ModeSpec` either. I think because of that, you've missed adding the parameter in at least two places: `ModalPort` (commented below) and `ModeSimulation`.

All things considered though it might be best to be separate (just be careful about adding it everywhere...)

tidy3d/plugins/smatrix/ports/wave.py
```diff
@@ -108,6 +110,17 @@ class WavePort(AbstractTerminalPort, Box):
         description="Extrudes structures that intersect the wave port plane by a few grid cells when ``True``, improving mode injection accuracy.",
     )
 
+    interp_spec: Optional[ModeInterpSpec] = pd.Field(
```
momchil-flex: We want to have this as an option in ModalPort-s too even if the default is still None.

momchil-flex (commented):
Nice! I didn't know this was coming!

I think we *might* want to enable it by default for ModeMonitor-s, or at least for all ports. But we could also sit on it for a bit, maybe let @lucas-flexcompute test this out as a default in photonforge first?

tidy3d/components/mode_spec.py
caseyflex: I would also prefer this in `ModeSpec`. We can still set the defaults separately in different places, like `WavePort.mode_spec` defaults to `ModeSpec(interp_spec=DEFAULT_WAVE_PORT_INTERP_SPEC)`

tidy3d/components/mode/mode_solver.py
```diff
@@ -531,6 +602,9 @@ def data_raw(self) -> ModeSolverDataType:
         if self.mode_spec.group_index_step > 0:
             return self._get_data_with_group_index()
 
+        if self.interp_spec is not None and self.interp_spec.num_points < len(self.freqs):
+            return self._get_data_with_interp()
```
caseyflex: I'm wondering about the interaction with `group_index_step`. Maybe if both are set, we should use the interpolated data to obtain the group index?

tidy3d/components/mode_spec.py
momchil-flex: The problem I'm anticipating is when you want to set a non-default one. Say default ModeSpec doesn't interpolate but default `ModeMonitor.mode_spec` does:

```
ms = td.ModeSpec(angle_phi=np.pi / 4)
monitor = td.ModeMonitor(mode_spec=ms) # now this won't interpolate anymore
```

tidy3d/components/mode_spec.py
```diff
@@ -86,6 +87,117 @@ class ModeSortSpec(Tidy3dBaseModel):
...
+    """
+
+    num_points: int = pd.Field(
```
caseyflex: I have a use case which is not quite covered by the current API. I want to know the mode profiles at a single frequency, say the central frequency; but I don't care about how they change wrt frequency. I do care about how the dispersion relation changes wrt frequency, and ideally I would compute that with interpolation.

Currently, I would need to do two separate mode solves; one at the central frequency recording the fields, and one over the full frequency range just storing the `n_complex`. This is a bit inefficient and perhaps cumbersome.

I wonder if we can add a flag to `mode_spec` like `store_fields_at_all_freqs`. If `False`, we just store the fields at the central frequency, while storing the index at all frequencies? This introduces some misalignment between the field and n_complex data arrays, but it can be handled reasonably by broadcasting where needed?

Also I think it would have been better if `fields` and `conjugated_dot_product` were part of `mode_spec` too.

@momchil-flex what do you think?

caseyflex (commented):
This is awesome @dbochkov-flexcompute !

tidy3d/components/mode_spec.py
caseyflex: oh I see. Well, we  already have `ModeSpecType = Union[ModeSpec, MicrowaveModeSpec]`. Maybe we should introduce more subclasses like `WavePortModeSpec`, if we want different default mode specification behavior? I still think everything related to the mode solver settings should be in the shared class `ModeSpec`

caseyflex: we actually have `EMEModeSpec` too already

tidy3d/components/data/monitor_data.py
dbochkov-flexcompute: btw, regarding grid correction factors, do you think it's ok to interpolate or better recalculate after interpolation?

tidy3d/components/mode_spec.py
dbochkov-flexcompute: yeah, no strong preference here from me, just unsure about having this information in `mode_spec` when some objects, `ModeSource` and `ModeABCBoundary`, will not use it. Another option to make it a bit cleaner while not putting `ModeInterpSpec` into `ModeSpec` is to create a mixin `ModeInterpolationModel` and make all revelant classes (`ModeMonitor`, `ModeSolverMonitor`, `WavePort`, `ModalPort`, `ModeSimulation`) to inherit from it

tidy3d/components/mode_spec.py
```diff
@@ -86,6 +87,117 @@ class ModeSortSpec(Tidy3dBaseModel):
...
+        >>> interp_spec = ModeInterpSpec(num_points=10, method='cheb')
+        >>> sampling_freqs = interp_spec.sampling_points(freqs)
+        """
+        freqs_array = np.asarray(freqs)
+        f_min, f_max = float(freqs_array.min()), float(freqs_array.max())
+
+        if self.method in ("linear", "cubic"):
+            # Uniformly spaced points
+            return np.linspace(f_min, f_max, self.num_points)
+        elif self.method == "cheb":
```
weiliangjin2021: pylint would complain no else is needed

tidy3d/components/mode/mode_solver.py
dbochkov-flexcompute: I need to improve on it indeed. What currently happens, if `group_index_step` is not `False` and `interp_spec` is not `None`:
- `_get_data_with_group_index` intercepts original mode solver and creates a new mode solver with additional frequency points
- `_get_data_with_interp` intercepts this modified mode solver and interpolates modes to all frequency points (including the additional aones)
- `_get_data_with_group_index` uses interpolated values to calculate group index and dispersion 

I'm thinking of modifying the behavior in the following way:
- `group_index_step` is not `False` and `interp_spec` is `linear`: error, because linear interpolation is not enough for computing second order derivatives for dispersion
- if `group_index_step=True` and `interp_spec` is not `None`: obtain group index and dispersion by differentiating underlying interpolators directly, without first interpolating to additional points
- if `group_index_step` is a specific number and `interp_spec` is not `None`: obtain group index and dispersion as it's currently done, that is, by using interpolated values

Maybe @lucas-flexcompute would also have some opinion on that?

tidy3d/components/data/monitor_data.py
```diff
@@ -2388,6 +2391,211 @@ def normalize(self, source_spectrum_fn: Callable[[float], complex]) -> ModeSolve
...
+        mode_interp_spec = ModeInterpSpec(method="cheb", num_points=len(freqs))
+        expected_freqs = mode_interp_spec.sampling_points(freqs)
+
+        # Sort both arrays for comparison (Chebyshev nodes are naturally sorted in descending order)
+        freqs_sorted = np.sort(freqs)
+        expected_sorted = np.sort(expected_freqs)
+
+        # Check relative error
+        freq_range = np.abs(expected_freqs[-1] - expected_freqs[0])
+        max_error = np.max(np.abs(freqs_sorted - expected_sorted)) / freq_range
```
weiliangjin2021: will `np.allclose` with custom atol and rtol be more robust here?

weiliangjin2021 (commented):
A very pleasant PR to review: both feature-rich, but not so many lines of code change!

As a future step, maybe it's useful to support adaptive sampling, as the mode profile probably changes more rapidly in high-frequency compared to low-frequency range. Or perhaps expose an API for custom sampling frequencies for now?

tidy3d/components/data/monitor_data.py
```diff
@@ -2388,6 +2391,211 @@ def normalize(self, source_spectrum_fn: Callable[[float], complex]) -> ModeSolve
...
+        # Build update dictionary
+        update_dict = {}
+
+        # Interpolate n_complex (required field)
+        update_dict["n_complex"] = self._interp_dataarray(self.n_complex, freqs, method)
+
+        # Interpolate field components if present
+        for field_name, field_data in self.field_components.items():
+            if field_data is not None:
+                update_dict[field_name] = self._interp_dataarray(field_data, freqs, method)
```
caseyflex: This runs the risk of being very slow, did you benchmark it at all?

tidy3d/components/mode_spec.py
caseyflex: or `assume_constant_fields`. Basically that could be the API for interpolating the `n_complex` but only storing one copy of the mode fields

tidy3d/components/mode/mode_solver.py
```diff
@@ -498,6 +546,29 @@ def _get_data_with_group_index(self) -> ModeSolverData:
...
+
+        # Create reduced frequency set based on interpolation method
+        freqs_reduced = self.interp_spec.sampling_points(self.freqs)
+
+        # Create a copy of the mode solver with reduced frequencies and no interp_spec
+        # (to prevent recursion)
+        mode_solver_reduced = self.copy(update={"freqs": freqs_reduced, "interp_spec": None})
+
+        # Get data at reduced frequencies
+        data_reduced = mode_solver_reduced.data_raw
```
caseyflex: Can you use `_data_on_yee_grid` here instead, and then do the processing later in `data_raw`?

tidy3d/components/mode/mode_solver.py
momchil-flex: Actually I don't really see the benefit of your proposed approach. In the current approach, the group index will be computed relatively correctly at the sampling frequencies, or at least equally correctly to the way we're computing it now. Then we just interpolate that. The group index is usually smooth. The GVD might have stronger dependence but even that shouldn't be too bad.

In your proposal, this would be the default way people do the computation.

```
if group_index_step=True and interp_spec is not None: obtain group index and dispersion by differentiating underlying interpolators directly, without first interpolating to additional points
```

Are you suggesting it for higher accuracy or for increased speed compared to the current approach? I think we could consider it, but only if the accuracy of the GVD is higher.

momchil-flex:
> Or perhaps expose an API for custom sampling frequencies for now?

Yes I was thinking about suggesting the same thing. Another application of custom sampling frequencies could be when you have a mode crossing, you could put more frequencies around it so that the tracking works.

tidy3d/components/mode_spec.py
momchil-flex: I think it's a useful suggestion. As another example I think in photonforge @lucas-flexcompute has to get the fields, but can probably do all that's needed using the central frequency fields only. This could drastically reduce the amount of data that's downloaded. I mean in principle there's a workaround to this already: you could place a ModeSolverMonitor with a single frequency at the same position as the ModeMonitor. That was kind of the original purpose of the ModeSolverMonitor but we eventually kind of unified the two.

It would be nice to have a more native support yeah, although it definitely increases the complexity. So in any case I think that's something that should be done in a separet PR.

> Also I think it would have been better if fields and conjugated_dot_product were part of mode_spec too

I guess this, as the current interp spec, has to do with the fact that they don't take effect in a ModeSource. Actually, `conjugated_dot_product` is only really used in a `ModeMonitor`. It seems that it is an attribute to the `ModeSolver` just so that it can be passed in `to_monitor`. But I feel like it should instead be passed as an argument to that method, i.e. add a `kwargs` as in `to_source`? Looping @dmarek-flex here too. It's not yet released officially so we could remove it from the ModeSolver.

tidy3d/components/mode_spec.py
momchil-flex: > oh I see. Well, we already have ModeSpecType = Union[ModeSpec, MicrowaveModeSpec]. Maybe we should introduce more subclasses like WavePortModeSpec, if we want different default mode specification behavior? I still think everything related to the mode solver settings should be in the shared class ModeSpec

Just trying to go with this to see how it feels: 
- Say that we ignore interpolation spec in ModeSource (this in itself is a bit confusing but potentially OK). Then I can still use regular ModeSpec in ModeSource, ModeMonitor, ModeSolverMonitor, ModeSolver, ModeSimulation, ModalPort.
- For all microwave objects like MicrowaveModeMonitor, WavePort, we use MicrowaveModeSpec. By the way @dmarek-flex I notice that WavePort [expects a ModeSpec](https://github.com/flexcompute/tidy3d/blob/7d63acee11a2079442552ffd7913495b84b40963/tidy3d/plugins/smatrix/ports/wave.py#L50-L54) rather than a MicrowaveModeSpec, is that a mistake?

> we actually have EMEModeSpec too already

- In EME, we've been talking about a different setting to only simulate at the central frequency. But I guess separately from that it could benefit from this frequency interpolation as an approach that would be more accurate than central frequency only but not as computationally heavy as all freqs. So potentially the interpolation is still used in EMEModeSpec, with even different defaults if needed.

I guess overall it doesn't sound *too* bad, especially if the user is already forced to use `MicrowaveModeSpec` or `EMEModeSpec` in the corresponding objects, which I think should be the case?


Then for Daniil's suggestion:
> Another option to make it a bit cleaner while not putting ModeInterpSpec into ModeSpec is to create a mixin ModeInterpolationModel and make all revelant classes (ModeMonitor, ModeSolverMonitor, WavePort, ModalPort, ModeSimulation) to inherit from it

Not opposed to that either, but apart from inheriting the field, there could be various places where we have to be careful that it's correctly passed. At least when it's in a ModeSpec, it should be passed around by default in most places.

dmarek-flex: > * By the way @dmarek-flex I notice that WavePort [expects a ModeSpec](https://github.com/flexcompute/tidy3d/blob/7d63acee11a2079442552ffd7913495b84b40963/tidy3d/plugins/smatrix/ports/wave.py#L50-L54) rather than a MicrowaveModeSpec, is that a mistake?

It is part of these bigger changes:
https://github.com/flexcompute/tidy3d/pull/2846

tidy3d/components/mode_spec.py
dmarek-flex: > Actually, `conjugated_dot_product` is only really used in a `ModeMonitor`. It seems that it is an attribute to the `ModeSolver` just so that it can be passed in `to_monitor`. But I feel like it should instead be passed as an argument to that method, i.e. add a `kwargs` as in `to_source`? Looping @dmarek-flex here too. It's not yet released officially so we could remove it from the ModeSolver.

I might have jumped the gun adding `conjugated_dot_product` to the `ModeSolver` and `ModeSimulation`, but I will need to add some Microwave-specific features for post-processing modes that I think is best done in the backend mode solver, so I do think it might be convenient to have that field around in the future. On the other hand, maybe internally I don't need to consider the conjugated version for my uses.

tidy3d/components/mode/mode_solver.py
momchil-flex: Sorry, I just reread your description of what the *current* setup does, and I see the issue. The reason I posted the comment above was because in my mind something else was happening. This is what I thought:

- Interpolation frequencies at which modes will be computed are selected
- If group index is enabled, the additional frequencies around these are added
- The group index and GVD are computed with the same accuracy as it is now at those frequencies
- They are then interpolated to the other solver frequencies just as n_complex, etc.

Does that make sense as the workflow to implement?

dbochkov-flexcompute: yeah, I see, I think this is the matter of simply swapping the order of `_get_data_with_group_index` and `_get_data_with_interp`. The only concerns is that we are tripling number of mode solves, so being able to, say, use the underlying chebyshev interpolator could be nice as an advanced option

tidy3d/components/mode_spec.py
lucas-flexcompute: >I think it's a useful suggestion. As another example I think in photonforge @lucas-flexcompute has to get the fields, but can probably do all that's needed using the central frequency fields only.

PF needs the fields at a single frequency indeed only for computing mode number remapping (if the user uses symmetry or `filter_pol`) so that we can connect to the correct modes in other parts of the circuit where those might not have been used. But we build this as a projection operation (not index search) because there's no guarantee that the "same" modes will be found with and without symmetry due to degenerate mode mixing.
In the end, I think we might just keep using the full frequency data (at least in the short term) because it is easy to implement.

lucas-flexcompute (commented):
Sorry if this is a dumb question, but I need to understand it first: is the `ModeInterpSpec` a superset of what `ModeSource.num_freqs` already does? How do they combine?

tidy3d/components/mode/mode_solver.py
lucas-flexcompute: > Maybe @lucas-flexcompute would also have some opinion on that?

Not strongly… My understanding is that the interpolation error should probably already be less than the error we get from the effective index due to discretization already. If interpolating over more points and getting the derivatives from the interpolation helps, solving for 3× the number of frequencies is not a big deal from the user perspective.

dbochkov-flexcompute:
> Sorry if this is a dumb question, but I need to understand it first: is the `ModeInterpSpec` a superset of what `ModeSource.num_freqs` already does? How do they combine?

There is definitely a big overlap between them, but `ModeInterpSpec` was not intended as a replacement for `num_freqs`. More like `ModeInterpSpec` is analogue of broadband source's `num_freqs` for monitors with more options. I think we could consider using `ModeInterpSpec` in sources as well and limit it there to only `method="cheb"` (and also maybe rename it to `FreqInterpSpec` as it would also apply to gaussian beam sources).

@momchil-flex @caseyflex this could actually make the decision whether to put `ModeInterpSpec` into `ModeSpec` or not easier. Because, if sources use `ModeInterpSpec` too, then no issue with them ignoring it in `ModeSpec`. However, for gaussian beams we still need to pass it separately. Btw, I don't think there is any benefit in using profile interpolation in incoming gaussian monitors, right?

tidy3d/components/mode/mode_solver.py
momchil-flex: @dbochkov-flexcompute I think we should go with the approach of getting the "equivalent accuracy" group index at the sampling points (even if it triples the number of solves, as it has until now) just because it matches legacy usage and just overall keeps the usage straightforward. I'm not a huge fan for example of not allowing group_index_step=True with interpolation=linear. Maybe just in the future we could experiment with using the interpolators to compute the derivatives if `group_index_step=True` and interpolation is not linear.

momchil-flex:
> @momchil-flex @caseyflex this could actually make the decision whether to put `ModeInterpSpec` into `ModeSpec` or not easier. Because, if sources use `ModeInterpSpec` too, then no issue with them ignoring it in `ModeSpec`. However, for gaussian beams we still need to pass it separately.

So are you driving towards separating it also because of this argument? And I guess deprecating `ModeSource.num_freqs` for a `FreqInterpSpec`?

> Btw, I don't think there is any benefit in using profile interpolation in incoming gaussian monitors, right?

I don't think so as the field computation is just analytical. Maybe a tiny benefit to interpolating.

caseyflex:
@dbochkov-flexcompute I think it makes sense to incorporate it into `ModeSpec` and use it in mode sources as well

dbochkov-flexcompute:
I think I will move `ModeInterpSpec` into `ModeSpec`, but not switch broadband sources to use it. This is because we have broadband gaussian beams and even plane wave, and it would also be weird to require `method="cheb"` only

dbochkov-flexcompute:
Did a revision of this (turned out a bit more than expected):
- moved interp spec into mode spec
- added renormalization after interpolation
- slight reorg of basic dataset classes so that we can easily reorder and interpolate both main mode data and additional RF data attached to it (there is an option just to recalculate RF priorities though)
- swapped order of interpretors in data_raw so that group index and dispersion are calculated as previously

Still not sure whether we need to worry much about interpolating grid correction factor or better to recompute them.

Another question is whether we want to implement any sort of automatic fall back strategy "cheb" -> "cubic" -> "linear" -> "nearest" of we detect a mode discontinuity

caseyflex:
I think it's better to recompute the grid correction factor, interpolating the complex exponential can lead to bad behavior (abs < 1)

I'm not sure about a fallback, but I am going to add `method="poly"` where the sampling points are closely concentrated around the central frequency (for more reliable tracking), and polynomial extrapolation is used

momchil-flex:
One thing I remembered, you should add the `ModeInterpSpec` to the api docs. And while your at it could you also add the `ModeSortSpec` which I forgot in my PR?

dbochkov-flexcompute:
> I think it's better to recompute the grid correction factor, interpolating the complex exponential can lead to bad behavior (abs < 1)

Agree. I did have to add some new fields (`grid_distances_primal` and `grid_distances_dual`) into `ModeSolverData` though to enable that.

`ModeSolverData.interpolated_data` is working now too and I followed the following logic:
```
mode_solver_full_mnt = td.ModeSolverMonitor(..., reduce_data=False, name="full")
mode_solver_reduced_mnt = td.ModeSolverMonitor(..., reduce_data=True, name="reduced")
...
assert sim_data["full"] == sim_data["reduced"].interpolated_copy
```

momchil-flex:
> > I think it's better to recompute the grid correction factor, interpolating the complex exponential can lead to bad behavior (abs < 1)
> 
> Agree. I did have to add some new fields (`grid_distances_primal` and `grid_distances_dual`) into `ModeSolverData` though to enable that.

Curious about this - the grid should be the same at all frequencies, no?

dbochkov-flexcompute:
> > > I think it's better to recompute the grid correction factor, interpolating the complex exponential can lead to bad behavior (abs < 1)
> > 
> > 
> > Agree. I did have to add some new fields (`grid_distances_primal` and `grid_distances_dual`) into `ModeSolverData` though to enable that.
> 
> Curious about this - the grid should be the same at all frequencies, no?

yeah, just after n_complex is interpolated onto new set of frequencies, we need to recompute `grid_factor = |d2| / (|d1| + |d2|) * exp(1j * k * n * d1) + |d1| / (|d1| + |d2|) * exp(1j * k * n * d2)` for each new frequency. Currently, grid factors are computed in `ModeSolver` where distances in normal direction d1 and d2 are known from the `simulation`. To allow this recalculation to happen in `ModeSolverData.interp_in_freq() -> ModeSolverData` we need to keep d1 and d2 in `ModeSolverData` as well

tidy3d/components/mode/mode_solver.py
caseyflex: @dbochkov-flexcompute I still can't use this directly in EME unless you use `_data_on_yee_grid` instead of `data_raw` here. It might take reworking the group index handling a bit too, but I think it is worth it

tidy3d/components/mode/mode_solver.py
```diff
@@ -505,6 +514,41 @@ def _get_data_with_group_index(self) -> ModeSolverData:
 
         return mode_solver.data_raw._group_index_post_process(self.mode_spec.group_index_step)
 
+    def _get_data_with_interp(self) -> ModeSolverData:
```
caseyflex: @dbochkov-flexcompute Maybe it would be better to directly incorporate this method into `_data_on_yee_grid`. That is, define a property `ModeSolver.sampling_points` which is either `freqs` or `sampling_points(freqs)`; then `_data_on_yee_grid` always computes at `sampling_points`. Then `interpolated_data` can be called, depending on `reduce_data`, either there or in `data_raw`.

tidy3d/components/base.py
```diff
@@ -1397,3 +1397,16 @@ def __getattribute__(self, name: str):
 
     _LazyProxy.__name__ = proxy_name
     return _LazyProxy
+
+
+class InterpolatableMixin(Tidy3dBaseModel):
```
caseyflex: I don't understand why this has to be a mixin, instead of an extra field in `ModeInterpSpec`

tidy3d/components/mode/mode_solver.py
```diff
@@ -257,6 +263,9 @@ def _warn_plane_crosses_symmetry(cls, val, values):
                         )
         return val
 
+    _warn_interp_num_points = validate_interp_num_points()
```
caseyflex: Can we put this warning in `data_raw` instead?

caseyflex (approved):
@dbochkov-flexcompute I left a few more comments, but I'd like to get this in soon, so I can clean up my EME PR. I guess the important thing is to get the API stabilized -- maybe we can do that and merge, and then I can change the internals while working on EME. To that end, I'd just consider moving `reduce_data` into `ModeInterpSpec` as the primary API change, then see if we can get this merged

tidy3d/components/base.py
caseyflex: also you may need a unit test for reduce_data, not sure it works currently

dbochkov-flexcompute: Since `ModeSpec`/`ModeInterpSpec` are used in so many places I think it might get a bit confusing for the user to remember what options in those specs apply to which objects. With including `interp_spec` into `ModeSpec` it is not so bad as distinction there is simple: it applies only to monitors. But distinction for `reduce_data` is more fine grained: it applies to `ModeSolverMonitor` and `ModeSolver` (`ModeSimulation` and `EMEModeSolverMonitor` in future?), but not `ModeMonitor`. It also avoids accidentally turning on this feature where we do not want/anticipate it.

tidy3d/components/mode/mode_solver.py
dbochkov-flexcompute: Yeah, I think this should be doable now with `.interpolated_copy` property now, because we do need to wait until mode tracking is done and until group_index/dispersion are calculated to do interpolation

tidy3d/components/base.py
caseyflex: Can we not just have different defaults / validation in these different places?

It's not the end of the world, but for example in EME, it is convenient to have all of the arguments for the mode solver kept in `ModeSpec` -- then we can simply have one mode spec per EME cell, rather than tracking additional data for every cell too. It is cleaner from my perspective to have everything in one place

tidy3d/components/data/dataset.py
```diff
@@ -50,6 +50,139 @@ def data_arrs(self) -> dict:
...
+        -------
+        DataArray
+            Interpolated data array with the same structure but new frequency points.
+        """
+        # Map 'cheb' to xarray's 'barycentric' method
+        xr_method = "barycentric" if method == "cheb" else method
+
+        # Use xarray's built-in interpolation
+        # For complex data, this automatically interpolates real and imaginary parts
+        interp_kwargs = {"method": xr_method}
```
caseyflex, line 147: maybe it would be faster with `assume_sorted=True`

tidy3d/components/mode_spec.py
```diff
@@ -86,6 +87,126 @@ class ModeSortSpec(Tidy3dBaseModel):
...
+        >>> freqs = np.linspace(1e14, 2e14, 100)
+        >>> interp_spec = ModeInterpSpec(num_points=10, method='cheb')
+        >>> sampling_freqs = interp_spec.sampling_points(freqs)
+        """
+        freqs_array = np.asarray(freqs)
+        f_min, f_max = float(freqs_array.min()), float(freqs_array.max())
+
+        if self.method in ("linear", "cubic"):
+            # Uniformly spaced points
+            return np.linspace(f_min, f_max, self.num_points)
```
caseyflex: can we override the default sampling points by manually specifying them, instead of just using `num_points`? In case the frequency tracking fails, maybe we need them to be closer together than the default

dbochkov-flexcompute: yeah, I think that could be useful. Not sure what would be the best API. Maybe something like this?
```
interp_spec=td.ModeInterpSpec(
    method="linear"/"cubic"/"poly",
    sampling_spec=td.UniformSampling(num_points=3)/td.ChebSampling(num_points)/td.CustomSampling(freq_arr),
)
```
where any combination of method + sampling_spec is valid, and a few convenience class methods
```
interp_spec=td.ModeInterpSpec.uniform(method="linear"/"cubic"/"poly", num_points)
interp_spec=td.ModeInterpSpec.cheb(num_points)
interp_spec=td.ModeInterpSpec.custom(method="linear"/"cubic"/"poly", freqs)
```

caseyflex: That could work - but I wonder, since overriding is a more specialized use case, maybe it’s better to have a new field “override_sampling_points” that accepts the raw sampling frequencies and takes precedence over auto sampling point generation

caseyflex: Sorry for coupling this so closely to my EME PR, but I feel they are related. In my use case, I actually would not use polynomial interpolation (doesn’t always work so well in practice), but maybe spline with extrapolation but maybe with default sampling points concentrated around the central frequency?

dbochkov-flexcompute: No worries at all, I think that's the best kind of feedback. Comparing the two options, I think I do slightly prefer the sampling_spec route better because it nicely generalizes cheb method to poly. Additionally, I think we might want to introduce some sort of `LogSampling` in future for cases where propagation constant is expected to be singular as we approach zero frequency. For your application, maybe you could introduce `GaussianSampling` (the naming maybe not the best though)? Overall feels like a more future-proof organization. I gave a quick stab in the last commit, see if it looks ok to you

caseyflex (approved):
This looks great to me, very flexible and extensible. thanks @dbochkov-flexcompute !

dbochkov-flexcompute:
@greptile

tidy3d/components/data/monitor_data.py
```diff
@@ -2455,10 +2465,257 @@ class ModeSolverData(ModeData):
...
+
+    grid_distances_dual: Union[tuple[float], tuple[float, float]] = pd.Field(
+        (0.0,),
+        title="Distances to the Dual Grid",
+        description="Relative distances to the dual grid locations along the normal direction in "
+        "the original simulation grid. Needed to recalculate grid corrections after "
+        "interpolating in frequency.",
+    )
+
+    reduced_data: bool = pd.Field(
```
caseyflex: do we need this field, or can we just use ModeMonitorData.monitor.reduce_data (and maybe set this flag differently after `interpolated_copy`)?

dbochkov-flexcompute: Yeah, I actually wanted to go with this approach in the beginning, but then had to switched to an additional flag (similar in nature to `symmetry` parameter of monitor data classes). However, midway I simplified the internal mode solver monitor, which allowed me know to remove the additional flag and rely on `monitor.reduce_data` (although changing the underlying monitor may be not the most ideal thing)

PR #2927, opened 2025-10-24
FXC-3351-Updates to speed up some mode data operations

momchil-flex:
As discussed [here](https://github.com/flexcompute/compute/pull/2686#issuecomment-3443156203) I made some changes to speed up the overlap sort.

- In that comment, I mention that my recent sorting updates did not degrade performance, but I was actually wrong (I was testing without those...) The new reorder I introduces no longer used in-place operations on the data arrays directly, which is more robust, but it was also slower. So I replaced this with a function that still acts on the data directly, but unfortunately to keep the robustness this introduces some complexity in rearranging dimensions.
- Apply symmetry expansion before starting the frequency tracking to avoid doing it repeatedly.
- Add fast pathways in some methods when coordinates already match (as would be the case for doing dot products of mode data with itself at a different frequency or mode index).
- Remove `monitor_data._updated` as we now support `updated_copy(deep=False)`, which is both better documented and seems to work faster.

<h2>Greptile Overview</h2>

Updated On: 2025-10-24 13:20:46 UTC

<h3>Greptile Summary</h3>

Optimized mode data operations by replacing the internal `_updated()` method with direct calls to `updated_copy(deep=False, validate=False)` to avoid unnecessary deep copying and validation overhead.

Key changes:
- **Removed `_updated()` method** from `MonitorData` class which used `dict()` + `parse_obj()` approach
- **Replaced with `updated_copy()`** using `deep=False` and `validate=False` flags in 5 locations across `monitor_data.py` and 1 location in `mode_solver.py`
- **Added early-exit optimization** in `_interpolated_tangential_fields()` that checks if coordinates already match using `np.allclose()` before interpolation
- **Improved `overlap_sort()` efficiency** by using `symmetry_expanded` data once and removing unnecessary coordinate reassignment via `_assign_coords()`

The changes maintain the same shallow-copy semantics while using the more standard `updated_copy()` API with appropriate flags, resulting in cleaner code and improved performance.

- This PR is safe to merge with minor considerations around the early-exit optimization
- The changes are well-focused performance optimizations that replace a custom `_updated()` method with the standard `updated_copy()` API. The refactoring maintains the same shallow-copy semantics and improves code consistency. However, the new early-exit optimization in `_interpolated_tangential_fields()` uses `np.allclose()` for coordinate comparison, which is appropriate for floating-point comparisons but should be monitored to ensure it doesn't skip interpolation when coordinates are very close but not identical enough for the use case.
- Pay attention to `monitor_data.py:800-808` where the new coordinate matching logic was added

File Analysis

| tidy3d/components/data/monitor_data.py | 4/5 | Replaced `_updated()` method with `updated_copy(deep=False, validate=False)` for performance; added early-exit optimization in `_interpolated_tangential_fields()`; optimized `overlap_sort()` to use `symmetry_expanded` and removed unnecessary coordinate reassignment |
| tidy3d/components/mode/mode_solver.py | 5/5 | Replaced `_updated()` method call with `updated_copy(deep=False, validate=False)` for consistency with monitor_data.py changes |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client
    participant ModeData
    participant AbstractFieldData
    participant ModeSolver
    
    Note over ModeData: overlap_sort() optimization
    Client->>ModeData: overlap_sort(track_freq, overlap_thresh)
    ModeData->>AbstractFieldData: symmetry_expanded
    AbstractFieldData-->>ModeData: data_expanded (self or new copy)
    
    alt No symmetry (returns self)
        Note over AbstractFieldData: Returns self directly
    else Has symmetry
        AbstractFieldData->>AbstractFieldData: updated_copy(deep=False, validate=False)
        Note over AbstractFieldData: Creates shallow copy with expanded fields
    end
    
    ModeData->>ModeData: dot(data_expanded, conjugate)
    Note over ModeData: Compute self-overlap
    
    loop For each frequency direction
        ModeData->>ModeData: _isel(f=[freq_id])
        Note over ModeData: Extract frequency slice
        ModeData->>ModeData: _find_ordering_one_freq()
        Note over ModeData: Uses dot() instead of outer_dot()
    end
    
    ModeData->>ModeData: updated_copy(deep=False, validate=False)
    ModeData-->>Client: Sorted mode data
    
    Note over ModeSolver: _colocate_data() optimization
    Client->>ModeSolver: _colocate_data(mode_solver_data)
    ModeSolver->>ModeSolver: updated_copy(deep=False, validate=False)
    Note over ModeSolver: Replaces _updated() method
    ModeSolver-->>Client: Colocated data
    
    Note over AbstractFieldData: _interpolated_tangential_fields() optimization
    Client->>AbstractFieldData: _interpolated_tangential_fields(coords)
    AbstractFieldData->>AbstractFieldData: Check if coords match using np.allclose()
    
    alt Coords already match
        AbstractFieldData-->>Client: Return fields directly (early exit)
    else Coords don't match
        AbstractFieldData->>AbstractFieldData: Interpolate fields
        AbstractFieldData-->>Client: Return interpolated fields
    end
```

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -149,20 +149,6 @@ class MonitorData(AbstractMonitorData, ABC):
 
         return self.normalize(amplitude_fn)
 
-    def _updated(self, update: dict) -> MonitorData:
-        """Similar to ``updated_copy``, but does not actually copy components, for speed.
-
-        Note
-        ----
-            This does **not** produce a copy of mutable objects, so e.g. if some of the data arrays
-            are not updated, they will point to the values in the original data. This method should
-            thus be used carefully.
-
-        """
-        data_dict = self.dict()
-        data_dict.update(update)
-        return type(self).parse_obj(data_dict)
-
     def _make_adjoint_sources(self, dataset_names: list[str], fwidth: float) -> list[Source]:
         """Generate adjoint sources for this ``MonitorData`` instance."""
 
@@ -261,7 +247,7 @@ class AbstractFieldData(MonitorData, AbstractFieldDataset, ABC):
         if all(sym == 0 for sym in self.symmetry):
             return self
 
-        return self._updated(self._symmetry_update_dict)
+        return self.updated_copy(**self._symmetry_update_dict, deep=False, validate=False)
 
     @property
     def symmetry_expanded_copy(self) -> AbstractFieldData:
@@ -780,22 +766,50 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
             fields_self = {key: field.conj() for key, field in fields_self.items()}
 
         fields_other = field_data._interpolated_tangential_fields(self._plane_grid_boundaries)
-
-        # Drop size-1 dimensions in the other data
-        fields_other = {key: field.squeeze(drop=True) for key, field in fields_other.items()}
-
-        # Cross products of fields
         dim1, dim2 = self._tangential_dims
-        e_self_x_h_other = fields_self["E" + dim1] * fields_other["H" + dim2]
-        e_self_x_h_other -= fields_self["E" + dim2] * fields_other["H" + dim1]
-        h_self_x_e_other = fields_self["H" + dim1] * fields_other["E" + dim2]
-        h_self_x_e_other -= fields_self["H" + dim2] * fields_other["E" + dim1]
+        d_area = self._diff_area
+
+        # After interpolation, the tangential coordinates should match. However, the two arrays
+        # may either have the same shape along other dimensions, or be broadcastable.
+        if (
+            fields_self[next(iter(fields_self))].shape
+            == fields_other[next(iter(fields_other))].shape
+        ):
+            # Arrays are same shape, so we can use numpy
+            e_self_x_h_other = fields_self["E" + dim1].values * fields_other["H" + dim2].values
+            e_self_x_h_other -= fields_self["E" + dim2].values * fields_other["H" + dim1].values
+            h_self_x_e_other = fields_self["H" + dim1].values * fields_other["E" + dim2].values
+            h_self_x_e_other -= fields_self["H" + dim2].values * fields_other["E" + dim1].values
+            integrand = xr.DataArray(
+                e_self_x_h_other - h_self_x_e_other, coords=fields_self["E" + dim1].coords
+            )
+            integrand *= d_area
+        else:
+            # Broadcasting is needed, which may be complicated depending on the dimensions order.
+            # Use xarray to handle robustly.
+
+            # Drop size-1 dimensions in the other data
+            fields_other = {key: field.squeeze(drop=True) for key, field in fields_other.items()}
+
+            # Cross products of fields
+            e_self_x_h_other = fields_self["E" + dim1] * fields_other["H" + dim2]
+            e_self_x_h_other -= fields_self["E" + dim2] * fields_other["H" + dim1]
+            h_self_x_e_other = fields_self["H" + dim1] * fields_other["E" + dim2]
+            h_self_x_e_other -= fields_self["H" + dim2] * fields_other["E" + dim1]
+            integrand = (e_self_x_h_other - h_self_x_e_other) * d_area
 
         # Integrate over plane
-        d_area = self._diff_area
-        integrand = (e_self_x_h_other - h_self_x_e_other) * d_area
-
         return ModeAmpsDataArray(0.25 * integrand.sum(dim=d_area.dims))
+
+    def _tangential_fields_match_coords(self, coords: ArrayFloat2D) -> bool:
+        """Check if the tangential fields already match given coords in the tangential plane."""
+        for field in self._tangential_fields.values():
+            for idim, dim in enumerate(self._tangential_dims):
+                if field.coords[dim].values.size != coords[idim].size or not np.all(
+                    field.coords[dim].values == coords[idim]
+                ):
+                    return False
+        return True
 
     def _interpolated_tangential_fields(self, coords: ArrayFloat2D) -> dict[str, DataArray]:
         """For 2D monitors, interpolate this fields to given coords in the tangential plane.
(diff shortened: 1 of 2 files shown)


momchil-flex:
@greptile give this another go buddy

tidy3d/components/data/monitor_data.py
```diff
@@ -812,6 +817,19 @@ def _interpolated_tangential_fields(self, coords: ArrayFloat2D) -> dict[str, Dat
...
+            # If coords already match, just return the tangential fields directly
+            for field in fields.values():
+                for idim, dim in enumerate(self._tangential_dims):
+                    if field.coords[dim].values.size != coords[idim].size or not np.all(
+                        field.coords[dim].values == coords[idim]
+                    ):
+                        raise ValueError("Coordinates do not match.")
+            return fields
+        except:  # noqa: E722
+            # Catching a broad exception here in case anything went wrong in the check.
```
dbochkov-flexcompute: do we even need try-catch here?

tidy3d/components/data/monitor_data.py
```diff
@@ -2215,20 +2234,58 @@ def _apply_mode_reorder(self, sort_inds_2d):
...
+            dims_orig = tuple(data.dims)
+            # Preserve coords (as numpy)
+            coords_out = {
+                k: (v.values if hasattr(v, "values") else np.asarray(v))
+                for k, v in data.coords.items()
+            }
+            f_axis = data.get_axis_num("f")
+            m_axis = data.get_axis_num("mode_index")
+
+            # Move axes so array is (..., f, mode)
```
dbochkov-flexcompute: can we move axes to `(f, ..., mode)` to begin with?

dbochkov-flexcompute (approved):
Looks good, just a couple minor (?) comments

tidy3d/components/data/monitor_data.py
momchil-flex: The idea is to not assume some specific axes order for robustness. Is your suggestion just so we can avoid the reorder after the sorting?

tidy3d/components/data/monitor_data.py
momchil-flex: Good point, but I do like the simplicity of try-except for control flow. Since I have a nested loop here, it's kind of verbose to control with a boolean and breaks.

I did change the except to explicitly only catch ValueError though. How does that sound?

dmarek-flex: I am wary of about using `try/except` for control flow. What about a helper method or 2 to simplify the steps?

tidy3d/components/data/monitor_data.py
```diff
@@ -811,6 +817,19 @@ def _interpolated_tangential_fields(self, coords: ArrayFloat2D) -> dict[str, Dat
...
+            # Using try: except for flow control.
+            for field in fields.values():
+                for idim, dim in enumerate(self._tangential_dims):
+                    if field.coords[dim].values.size != coords[idim].size or not np.all(
+                        field.coords[dim].values == coords[idim]
+                    ):
+                        raise ValueError("Coordinates do not match.")
+            return fields
+        except ValueError:
+            pass
```
dmarek-flex: Also here I think bool logic is preferable. From my understand raising an exception can also be slower. In this case, the interpolation is the expensive part so not a big problem, but good to avoid I think.

dmarek-flex (approved):
Looks good, but I do think trying to avoid those try/except usages would be better.

tidy3d/components/data/monitor_data.py
dmarek-flex: A method that can determine whether two monitor data have matching tangential fields would be useful when I get back to the `dot_yee`/`flux_yee` stuff as well

momchil-flex: I could switch this to bool and split this out in a separate method. What do you think about the other one? There I'm more worried about something erroring that I haven't anticipated (like e.g. due to dimensions order) so I kind of want it to fallback to the xarray handling.

dmarek-flex: Well I guess that would indicate a bug we need to fix. I am trying to think how it could fail, from your check the `coords` must be identical right? Same shape and order I would think.

dmarek-flex: Oh I see, only the tangential dims are checked

momchil-flex: Yeah because this is to speed up e.g. overlap computations of mode data at one frequency with the same mode data at a different frequency.

dmarek-flex: Ultimately, I will let you decide!

tidy3d/components/data/monitor_data.py
dbochkov-flexcompute: It seems like there could be up to 4 reorders: (orig) -> (..., f, mode) -> (f, ..., mode) -> (..., f, mode) -> (orig). So, I was wondering if it can be reduced to 2: (orig) -> (f, ..., mode) -> (orig).

momchil-flex: Ah got it will give it another look tomorrow!

tidy3d/components/data/monitor_data.py
dbochkov-flexcompute: yeah, I think switching this one to if-based, but leaving the other one as try-catch seems like a good compromise at the moment. 

But here's some idea for the one falling back to xarray: if we know fairly broad, but maybe not exhaustive, conditions under which we expect with certainty the optimized calculation to complete successfully maybe it's worth to switching to still if-based branching. That is,
```
if (broad but not exhaustive ok):
    # direct calc
else:
    # xarray calc
```
Yes, we won't get the most performance in all possible cases, but hopefully in overwhelming majority. Again, not insisting on this though.

momchil-flex: That's basically how I had it previously, I guess I just decided to be somewhat defensive against something randomly erroring in a way that I haven't anticipated and was not captured by tests. But I do agree with you guys that hiding errors is not great either.

tidy3d/components/data/monitor_data.py
momchil-flex: Updated, thanks!

momchil-flex:
Alright, I removed the try: except logic and made a standalone method `_tangential_fields_match_coords` to test if a given field data's tangential coords match a provided set. Now in `dot` I just test if the shapes are the same, because the tangential coordinates have to be the same after the call to the interpolation. Previously, I was attempting to fast-path the computation even when the shapes are not the same (broadcasting) but that was actually not always working and it was masked by the try - except. Doing the broadcasting with numpy is pretty nontrivial because we might have to do it on freqs and/or mode indexes, so I just fast-track if shapes are identical, and let xarray handle the more complicated cases for robustness.

Thanks for the useful suggestions!

PR #2936, opened 2025-10-28
Add interpolation to EME

caseyflex:


tidy3d/components/data/monitor_data.py
```diff
@@ -1660,6 +1660,9 @@ def eps_spec_match_mode_spec(cls, val, values):
         """Raise validation error if frequencies in eps_spec does not match frequency list"""
         if val:
             mode_data_freqs = values["monitor"].freqs
+            interp_spec = values["monitor"].mode_spec.interp_spec
+            if interp_spec is not None:
```
caseyflex, line 1664: I should change this to check that it matches the sampling_points

tidy3d/components/eme/data/sim_data.py
```diff
@@ -197,8 +197,8 @@ def smatrix_in_basis(
             modes1 = port_modes1
         if not modes2_provided:
             modes2 = port_modes2
-        f1 = list(modes1.field_components.values())[0].f.values
-        f2 = list(modes2.field_components.values())[0].f.values
+        f1 = list(modes1.monitor.freqs)
```
caseyflex, line 200: the monitor freqs are the full frequencies, even though the data is only stored at the sampling_points

tidy3d/components/eme/data/sim_data.py
```diff
@@ -259,6 +259,11 @@ def smatrix_in_basis(
                 overlaps1 = modes1.outer_dot(port_modes1, conjugate=False)
                 if not modes_in_1:
                     overlaps1 = overlaps1.expand_dims(dim={"mode_index_0": mode_index_1}, axis=1)
+                interp_spec1 = modes1.monitor.mode_spec.interp_spec
+                if interp_spec1 is not None:
+                    overlaps1 = modes1._interp_dataarray(
```
caseyflex, line 264: interpolating the overlaps is more efficient than interpolating the modes, but it can only be done if the frequencies agree. If the sampling_points don't agree, then the modes have to be interpolated first

tidy3d/components/mode/mode_solver.py
```diff
@@ -565,6 +565,7 @@ def _get_data_with_interp(self) -> ModeSolverData:
 
         # Get data at reduced frequencies
         data_reduced = mode_solver_reduced.data_raw
+        return data_reduced
```
caseyflex, line 568: @dbochkov-flexcompute in this version, the data is not interpolated when it is computed -- it can be interpolated as-needed later

caseyflex (commented):
@dbochkov-flexcompute @momchil-flex 

still a work in progress, just wanted to show what I'm thinking and some of the benefits of not interpolating until last minute in the interpolation PR. Feel free to comment at this point. Still need to add group index / GVD option, and other changes as per my comments

caseyflex:
Closed in favor of https://github.com/flexcompute/tidy3d/pull/2973

PR #2956, opened 2025-11-03
fix(tidy3d): FXC-3948-fix-caching-for-autograd

marcorudolphflex:
**Previous:** Caching does not work with autograd and non-local gradients because some data access via web is tried to made with a dummy task id for cached simulations.

**Solution:** Get original task id from originally cached simulation to get the trace-relevant data.

<h2>Greptile Overview</h2>

Updated On: 2025-11-04 12:14:42 UTC

<h3>Greptile Summary</h3>

Fixes caching for autograd with non-local gradients by retrieving and using the original task_id from cached simulations.

**Key changes:**
- Modified `restore_simulation_if_cached` to return both the restored path and the cached task_id
- Updated autograd forward pass to check cache and use cached task_id when available for backward pass parent_tasks dependency
- Changed stash path generation to use `uuid.uuid4()` instead of `_cached_task_id` to avoid None values
- Added early exit in `Batch.monitor` for cached jobs

**Issues found:**
- Critical bug in `autograd.py:583-589`: `load()` is called with `path=None` instead of using `restored_path`, which will cause the load to fail
- Performance issue in `container.py:1128`: `load_if_cached` is a `@cached_property` but is accessed inside the monitor loop, causing repeated cache checks
- Missing docstring update in `webapi.py` for the new tuple return value

<h3>Confidence Score: 2/5</h3>

- This PR has critical bugs that will break the caching functionality for autograd
- Score reflects a critical bug where `load()` is called with `path=None` instead of `restored_path`, which will cause autograd cache loading to fail. Additionally, there's a performance issue with repeated cached property access in the monitor loop.
- Critical attention needed for `tidy3d/web/api/autograd/autograd.py` (incorrect path parameter). Performance issue in `tidy3d/web/api/container.py` (inefficient cached property access).

File Analysis

| tidy3d/web/api/webapi.py | 3/5 | Modified `restore_simulation_if_cached` to return tuple of (path, task_id); has incorrect return type annotation and outdated docstring |
| tidy3d/web/api/container.py | 3/5 | Updated cache handling with `_cached_task_id` attribute and uuid-based stash paths; inefficient cached property access in monitor loop |
| tidy3d/web/api/autograd/autograd.py | 2/5 | Added cache restoration for non-local gradients; critical bug passing None as path to `load()` instead of using `restored_path` |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant AG as Autograd
    participant Cache as restore_simulation_if_cached
    participant Load as load()
    participant Web as WebAPI
    
    Note over AG: Non-local gradient case
    AG->>Cache: restore_simulation_if_cached(sim_original)
    Cache-->>AG: (restored_path, cached_task_id)
    
    alt Cache Hit
        Note over AG: BUG: passes path=None instead of restored_path
        AG->>Load: load(task_id=None, path=None)
        Note over Load: Will fail - cannot load from None path
        Load--xAG: Error
    else Cache Miss
        AG->>Web: _run_tidy3d(sim_original)
        Web-->>AG: (sim_data_orig, task_id_fwd)
    end
    
    Note over AG: aux_data[FWD_TASK_ID] = task_id_fwd
    Note over AG: Later: Backward pass uses cached task_id
```

### DIFF
--- a/tests/test_web/test_local_cache.py
+++ b/tests/test_web/test_local_cache.py
@@ -7,16 +7,23 @@ import re
 from pathlib import Path
 from types import SimpleNamespace
 
+import autograd as ag
 import pytest
+import xarray as xr
+from autograd.core import defvjp
 from rich.console import Console
 
 import tidy3d as td
 from tests.test_components.autograd.test_autograd import ALL_KEY, get_functions, params0
 from tests.test_web.test_webapi_mode import make_mode_sim
 from tidy3d import config
+from tidy3d.components.autograd.field_map import FieldMap
 from tidy3d.config import get_manager
 from tidy3d.web import Job, common, run, run_async
 from tidy3d.web.api import webapi as web
+from tidy3d.web.api.autograd import autograd, engine, io_utils
+from tidy3d.web.api.autograd.autograd import run as run_autograd
+from tidy3d.web.api.autograd.constants import SIM_VJP_FILE
 from tidy3d.web.api.container import Batch, WebContainer
 from tidy3d.web.api.webapi import load_simulation_if_cached
 from tidy3d.web.cache import CACHE_ARTIFACT_NAME, clear, get_cache_entry_dir, resolve_local_cache
@@ -38,6 +45,24 @@ def _reset_fake_maps():
 class _FakeStubData:
     def __init__(self, simulation: td.Simulation):
         self.simulation = simulation
+
+    def __getitem__(self, key):
+        if key == "mode":
+            params = self.simulation.attrs["params_autograd"]
+            return SimpleNamespace(
+                amps=xr.DataArray(params, dims=["x"], coords={"x": list(range(len(params)))})
+            )
+
+    def _strip_traced_fields(self, *args, **kwargs):
+        """Fake _strip_traced_fields: return minimal valid autograd-style mapping."""
+        return {"params": self.simulation.attrs["params"]}
+
+    def _insert_traced_fields(self, field_mapping, *args, **kwargs):
+        self.simulation.attrs["params_autograd"] = field_mapping["params"]
+        return self
+
+    def _make_adjoint_sims(self, **kwargs):
+        return [self.simulation.updated_copy(run_time=self.simulation.run_time * 2)]
 
 
 @pytest.fixture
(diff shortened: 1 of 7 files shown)


tests/test_web/test_local_cache.py
yaugenst-flex, line 1: I wonder if this wouldn't be simpler if you just use the emulated run from the autograd tests? So for example a test like this, but with the local cache enabled: https://github.com/flexcompute/tidy3d/blob/8da1ce8d5ca87e7ab476c5c82a0d809c01efa72f/tests/test_components/autograd/test_autograd.py#L3013-L3042

Would that help simplify things?

tidy3d/web/api/autograd/io_utils.py
```diff
@@ -39,6 +39,24 @@ def get_vjp_traced_fields(task_id_adj: str, verbose: bool) -> AutogradFieldMap:
     try:
         download_file(task_id_adj, SIM_VJP_FILE, to_file=fname, verbose=verbose)
         field_map = FieldMap.from_file(fname)
+        from tidy3d.web.cache import resolve_local_cache
```
yaugenst-flex: Does this have to be a lazy import?

tidy3d/web/api/autograd/io_utils.py
```diff
@@ -39,6 +39,24 @@ def get_vjp_traced_fields(task_id_adj: str, verbose: bool) -> AutogradFieldMap:
...
+
+        simulation_cache = resolve_local_cache()
+        if simulation_cache is not None:
+            from tidy3d.web.api.webapi import get_info, load_simulation
```
yaugenst-flex: Same question here

yaugenst-flex (approved):
Thanks @marcorudolphflex!

tests/test_web/test_local_cache.py
marcorudolphflex: Thought about that, but does not really help. The interesting parts I want to test are mocked there. And also collides with my other cache-designed mocks.

tidy3d/web/api/autograd/io_utils.py
marcorudolphflex: was somehow necessary during development, but is not anymore - made a top-level import now

PR #2963, opened 2025-11-05
feat(tidy3d): FXC-3961-faster-convolutions-for-tidy-3-d-plugins-autograd-filters

marcorudolphflex:
<h2>Greptile Overview</h2>

Updated On: 2025-11-05 16:45:19 UTC

<h3>Greptile Summary</h3>

This PR replaces the autograd scipy convolution implementation with a custom FFT-based convolution to improve tracing performance. The main changes include:

- Removed dependency on `autograd.scipy.signal.convolve` and replaced with custom `_fft_convolve_general` function using FFT convolution
- Added `_normalize_axes` helper function to handle axis specification and validation
- Added comprehensive tests for convolution with explicit axes specification
- Updated CHANGELOG to document the performance improvement

**Issues found:**
- Dead code in `_normalize_axes` (lines 74-82): bounds checking occurs after modulo normalization, so checks will never trigger
- Extra blank line in CHANGELOG.md

- This PR is mostly safe to merge but contains dead code that should be addressed
- The FFT-based convolution implementation appears algorithmically sound and includes appropriate tests. However, the dead code in bounds checking (lines 74-82) indicates incomplete validation logic that could mask invalid inputs. While this won't cause runtime errors due to the modulo operation preventing out-of-bounds access, it means the function won't properly reject invalid axis specifications, which could lead to confusing behavior.
- Pay close attention to `tidy3d/plugins/autograd/functions.py` - the bounds checking logic in `_normalize_axes` should be fixed or removed

File Analysis

| tidy3d/plugins/autograd/functions.py | 3/5 | replaced autograd.scipy.signal.convolve with custom FFT-based implementation for performance; added helper functions `_normalize_axes` and `_fft_convolve_general`; contains dead code in bounds checking (lines 74-82) |
| tests/test_plugins/autograd/test_functions.py | 5/5 | added test class `TestConvolveAxes` with parametrized tests for convolution with explicit axes, including validation against numpy and gradient checks |
| CHANGELOG.md | 4/5 | added entry for improved autograd convolution speed; contains extra blank line on line 57 |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant convolve
    participant _normalize_axes
    participant pad
    participant _fft_convolve_general
    participant fftn/ifftn

    User->>convolve: array, kernel, axes, mode, padding
    convolve->>convolve: validate kernel dimensions (must be odd)
    convolve->>_normalize_axes: array.ndim, kernel.ndim, axes
    _normalize_axes->>_normalize_axes: normalize axes with modulo
    _normalize_axes->>_normalize_axes: validate axes (length match, uniqueness)
    _normalize_axes-->>convolve: axes_array, axes_kernel
    
    alt mode is "same" or "full"
        loop for each axis pair
            convolve->>pad: working_array, pad_width, padding mode
            pad-->>convolve: padded array
        end
        convolve->>convolve: set effective_mode ("valid" for "same", "full" for "full")
    end
    
    convolve->>_fft_convolve_general: working_array, kernel, axes, effective_mode
    _fft_convolve_general->>_fft_convolve_general: reorder axes (batch vs conv)
    _fft_convolve_general->>_fft_convolve_general: expand dimensions for broadcasting
    _fft_convolve_general->>fftn/ifftn: compute FFT convolution
    fftn/ifftn-->>_fft_convolve_general: full convolution result
    
    alt effective_mode is "valid"
        _fft_convolve_general->>_fft_convolve_general: slice valid region
    end
    
    _fft_convolve_general->>_fft_convolve_general: convert to real if inputs were real
    _fft_convolve_general-->>convolve: result
    convolve-->>User: convolved array
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `web.Batch(ComponentModeler)` and `web.Job(ComponentModeler)` native support
 - Simulation data of batch jobs are now automatically downloaded upon their individual completion in `Batch.run()`, avoiding waiting for the entire batch to reach completion.
 - Port names in `ModalComponentModeler` and `TerminalComponentModeler` can no longer include the `@` symbol.
+- Improved speed of convolutions for large inputs.
 
 ### Fixed
 - Ensured the legacy `Env` proxy mirrors `config.web` profile switches and preserves API URL.
--- a/tests/test_plugins/autograd/test_functions.py
+++ b/tests/test_plugins/autograd/test_functions.py
@@ -28,6 +28,7 @@ from tidy3d.plugins.autograd import (
     threshold,
     trapz,
 )
+from tidy3d.plugins.autograd.functions import _normalize_axes
 from tidy3d.plugins.autograd.types import PaddingType
 
 _mode_to_scipy = {
@@ -37,6 +38,15 @@ _mode_to_scipy = {
     "symmetric": "reflect",
     "wrap": "wrap",
 }
+
+CONV_MODES = ["full", "same", "valid"]
+
+_CONVOLVE_AXES_CASES = [
+    ([0], [0]),
+    ([1], [1]),
+    ([1], [0]),
+    ([-1], [-1]),
+]
 
 
 @pytest.mark.parametrize("mode", PaddingType.__args__)
@@ -94,7 +104,7 @@ class TestPadExceptions:
             pad(self.array, (1, 1), axis=-3)
 
 
-@pytest.mark.parametrize("mode", ["full", "valid", "same"])
+@pytest.mark.parametrize("mode", CONV_MODES)
 @pytest.mark.parametrize("padding", PaddingType.__args__)
 @pytest.mark.parametrize(
     "ary_size", [7, 8, (7, 7), (8, 8), (7, 8), (7, 7, 7), (8, 8, 8), (7, 8, 9)]
@@ -117,22 +127,10 @@ class TestConvolve:
         """Test convolution values against SciPy for various modes, padding, array sizes, and kernel sizes."""
         x, k = self._ary_and_kernel(rng, ary_size, kernel_size, square_kernel)
 
-        if mode in ("full", "same"):
-            pad_widths = [(k // 2, k // 2) for k in k.shape]
-            x_padded = x
-            for axis, pad_width in enumerate(pad_widths):
-                x_padded = pad(x_padded, pad_width, mode=padding, axis=axis)
-            conv_sp = convolve_sp(x_padded, k, mode="valid" if mode == "same" else mode)
-        else:
-            conv_sp = convolve_sp(x, k, mode=mode)
-
         conv_td = convolve(x, k, padding=padding, mode=mode)
+        conv_sp = _reference_convolution(x, k, mode, padding, axes=None)
 
-        npt.assert_allclose(
-            conv_td,
-            conv_sp,
-            atol=1e-12,  # scipy's "full" somehow is not zero at the edges...
-        )
+        npt.assert_allclose(conv_td, conv_sp, atol=1e-12)
 
     def test_convolve_grad(self, rng, mode, padding, ary_size, kernel_size, square_kernel):
         """Test gradients of convolution function for various modes, padding, array sizes, and kernel sizes."""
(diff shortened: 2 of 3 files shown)


tidy3d/plugins/autograd/functions.py
```diff
@@ -37,6 +37,139 @@
...
 
+def _normalize_axes(
+    ndim_array: int,
+    ndim_kernel: int,
+    axes: Union[tuple[Iterable[int], Iterable[int]], None],
+) -> tuple[tuple[int, ...], tuple[int, ...]]:
+    """Normalize the axes specification for convolution."""
+
+    def _normalize_single_axis(ax: int, ndim: int, kind: str) -> int:
+        if not isinstance(ax, int):
```
yaugenst-flex, line 50: Are we sure we want to be this strict with the `int` type here? This would break if you pass something like `axes=np.arange(3)` for example because those are `np.int64`. I think this would be a pretty common case.

tests/test_plugins/autograd/test_functions.py
```diff
@@ -168,6 +171,52 @@ def test_kernel_array_dimension_mismatch(self):
...
+        """Test gradients of convolution when specific axes are provided."""
+        array = rng.random((2, 5))
+        kernel = rng.random((3, 3))
+        check_grads(convolve, modes=["rev"], order=2)(
+            array,
+            kernel,
+            padding="constant",
+            mode="valid",
+            axes=([1], [1]),
+        )
```
yaugenst-flex, line 274: Not really sure I understand this test. Wouldn't we want to parametrize over `axes`? That's what I would expect from the naming of the test at least. And we'd want to parametrize over `array` and `kernel` too? Why are these things parametrized in the value but not in the gradient test?

More generally, isn't this test replicating much of `TestConvolve`? Wouldn't we just want to add a parametrization over `axes` there?

tidy3d/plugins/autograd/functions.py
```diff
@@ -37,6 +37,139 @@
...
+            start = int(min(array_dim, kernel_dim) - 1)
+            length = int(abs(array_dim - kernel_dim) + 1)
+            axis = full_result.ndim - num_conv_axes + axis_offset
+            valid_slices[axis] = slice(start, start + length)
+        result = full_result[tuple(valid_slices)]
+    else:
+        raise ValueError(f"Unsupported convolution mode '{mode}'.")
+
+    if not np.iscomplexobj(array) and not np.iscomplexobj(kernel):
+        result = np.real(result)
```
yaugenst-flex: Wonder if this check is really necessary or we just return `np.real(result)`?

tidy3d/plugins/autograd/functions.py
```diff
@@ -180,19 +313,27 @@ def convolve(
...
+
+    working_array = array
+    effective_mode = mode
+
+    if mode in ["same", "full"]:
+        for ax_array, ax_kernel in zip(axes_array, axes_kernel):
+            if mode == "same" and kernel.shape[ax_kernel] % 2 == 0:
+                raise ValueError(
+                    f"Even-sized kernel along axis {ax_kernel} not supported for 'same' mode."
+                )
```
yaugenst-flex: Does numpy also complain about this?

CHANGELOG.md
```diff
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unified run submission API: `web.run(...)` is now a container-aware wrapper that accepts a single simulation or arbitrarily nested containers (`list`, `tuple`, `dict` values) and returns results in the same shape.
 - `web.Batch(ComponentModeler)` and `web.Job(ComponentModeler)` native support
 - Simulation data of batch jobs are now automatically downloaded upon their individual completion in `Batch.run()`, avoiding waiting for the entire batch to reach completion.
+- Improved speed of autograd tracing for convolutions.
```
yaugenst-flex: This change is not only related to autograd tracing right?

yaugenst-flex (changes requested):
Thanks @marcorudolphflex this is great! I guess as a general observation we should keep in mind that this new implementation is slower for small arrays/kernels and faster for larger ones. Since for small convolutions both methods are fast in absolute terms and for large cases (where the cost of the convolution becomes noticeable) FFT will be significantly faster I think this is definitely a better default. It could be interesting to check with a little bit of benchmarking where this crossover lies and compare that to typical sizes in topopt problems where the arrays are generally on the order of a few hundred pixels in x and y and kernels are probably on the order of ~10 pixels.

tidy3d/plugins/autograd/functions.py
marcorudolphflex: no, it silently returns a result that is shifted by half a sample (floor division), i.e. it assumes center = floor(M/2) for kernel length M - which I dislike... But guess we want to be compatible with that?

tests/test_plugins/autograd/test_functions.py
marcorudolphflex: Yes, would make sense to vary the axes/kernel param here. 
The reason for this test is to specifically test that this works with specified axes. We could do a unified test with axes parameters being None and non-None, but in this case we would probably just skip some tests if they would not make sense regarding the parameter combination? Alternatively, we could use some shared helpers to check the results to avoid duplicate code. Not sure what is more elegant here.

tidy3d/plugins/autograd/functions.py
yaugenst-flex: Hmm yeah I think it's probably fine as-is, we're not really trying to match the scipy or numpy interface here

marcorudolphflex:
> Thanks @marcorudolphflex this is great! I guess as a general observation we should keep in mind that this new implementation is slower for small arrays/kernels and faster for larger ones. Since for small convolutions both methods are fast in absolute terms and for large cases (where the cost of the convolution becomes noticeable) FFT will be significantly faster I think this is definitely a better default. It could be interesting to check with a little bit of benchmarking where this crossover lies and compare that to typical sizes in topopt problems where the arrays are generally on the order of a few hundred pixels in x and y and kernels are probably on the order of ~10 pixels.

image
 You are right, thats really size-dependent. In this test I used quadratic 2D inputs/kernels.

marcorudolphflex:
tweaked it a little bit (direct real fft and scipy.fft.next_fast_len), got this
image

yaugenst-flex:
> tweaked it a little bit (direct real fft and scipy.fft.next_fast_len)

Nice! I'd say for any realistic use case this would be strictly better than before!

yaugenst-flex (approved):

PR #2964, opened 2025-11-05
fix(rf): FXC-3994 simplify task and network naming scheme in TCM

dmarek-flex:
Main change here is fixing a missed issue when adding the multimodal waveports last week (using _ instead of @). Now network_index will be identical to task name. But also tried to reduce code duplication and make things more explicit in the `get_task_name`.

I still want to keep `network_index` because it makes more sense to me when indexing into matrices, rather than task name. I don't want to break backwards compatibility by removing or renaming `get_task_name` either.

<h2>Greptile Overview</h2>

Updated On: 2025-11-05 16:59:42 UTC

<h3>Greptile Summary</h3>

This PR fixes an inconsistency in the network naming scheme for the Terminal Component Modeler (TCM). The main change is standardizing the separator between port names and mode indices from underscore (`_`) to `@` symbol for WavePorts, making `network_index` consistent with `task_name`.

**Key Changes:**
- Refactored `network_index()` in `terminal.py` to delegate to `get_task_name()` instead of duplicating logic, eliminating the underscore vs `@` inconsistency
- Enhanced `get_task_name()` in `base.py` to enforce stricter validation: RF format now requires `mode_index` for non-lumped ports and prohibits it for lumped ports
- Added `PortType` union type to support the broader type signature needed in `get_task_name()`
- Updated all tests to reflect the new `@` separator format (`port1@0` instead of `port1_0`)

**Issues Found:**
- Docstring for `get_task_name()` is outdated and doesn't accurately describe the new RF format behavior or document all ValueError cases

- This PR is safe to merge with minor documentation improvements needed
- The logic changes are sound and well-tested, successfully fixing the naming inconsistency. Code duplication was eliminated through proper refactoring. However, the docstring for `get_task_name()` needs updating to accurately reflect the new behavior, which prevents a score of 5. All tests were updated correctly and error handling follows the style guide.
- The docstring in `tidy3d/plugins/smatrix/component_modelers/base.py` should be updated before merge

File Analysis

| tidy3d/plugins/smatrix/component_modelers/base.py | 3/5 | Refactored `get_task_name` to enforce mode_index requirement for non-lumped ports in RF format; docstring needs updating to reflect new behavior |
| tidy3d/plugins/smatrix/component_modelers/terminal.py | 5/5 | Simplified `network_index` to delegate to `get_task_name`, eliminating code duplication and fixing underscore vs `@` inconsistency |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant TCM as TerminalComponentModeler
    participant Base as AbstractComponentModeler
    
    Note over TCM,Base: Old flow (underscore format)
    User->>TCM: network_index(waveport, mode=0)
    TCM->>TCM: isinstance check
    TCM-->>User: "port1_0"
    
    Note over TCM,Base: New flow (@ format)
    User->>TCM: network_index(waveport, mode=0)
    TCM->>Base: get_task_name(port, mode=0, format="RF")
    Base->>Base: Check if LumpedPortType
    Base->>Base: Check mode_index not None
    Base-->>TCM: "port1@0"
    TCM-->>User: "port1@0"
    
    Note over TCM,Base: Lumped port handling
    User->>TCM: network_index(lumped_port, None)
    TCM->>Base: get_task_name(port, None, format="RF")
    Base->>Base: Check if LumpedPortType
    Base->>Base: Validate no mode_index
    Base-->>TCM: "port1"
    TCM-->>User: "port1"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unified run submission API: `web.run(...)` is now a container-aware wrapper that accepts a single simulation or arbitrarily nested containers (`list`, `tuple`, `dict` values) and returns results in the same shape.
 - `web.Batch(ComponentModeler)` and `web.Job(ComponentModeler)` native support
 - Simulation data of batch jobs are now automatically downloaded upon their individual completion in `Batch.run()`, avoiding waiting for the entire batch to reach completion.
+- Port names in `ModalComponentModeler` and `TerminalComponentModeler` can no longer include the `@` symbol.
 
 ### Fixed
 - Ensured the legacy `Env` proxy mirrors `config.web` profile switches and preserves API URL.
--- a/tests/test_plugins/smatrix/test_component_modeler.py
+++ b/tests/test_plugins/smatrix/test_component_modeler.py
@@ -434,17 +434,9 @@ def test_get_task_name():
     task_name = ModalComponentModeler.get_task_name(port=port, mode_index=1)
     assert task_name == "port1@1"
 
-    # Test with format="PF"
-    task_name = ModalComponentModeler.get_task_name(port=port, format="PF")
+    # Test with mode_index unspecified
+    task_name = ModalComponentModeler.get_task_name(port=port)
     assert task_name == "port1@0"
-
-    # Test with format="RF"
-    task_name = ModalComponentModeler.get_task_name(port=port, format="RF")
-    assert task_name == "port1"
-
-    # Test with invalid format
-    with pytest.raises(ValueError):
-        ModalComponentModeler.get_task_name(port=port, format="invalid")
 
 
 def test_custom_source_time(monkeypatch):
--- a/tests/test_plugins/smatrix/test_terminal_component_modeler.py
+++ b/tests/test_plugins/smatrix/test_terminal_component_modeler.py
@@ -1263,8 +1263,12 @@ def test_antenna_parameters(monkeypatch, port_type):
     modeler_data = run_component_modeler(monkeypatch, modeler)
 
     # Make sure network index works for single mode / multimode cases
-    port_1_network_index = modeler.network_index(modeler.ports[0], 0)
-    port_2_network_index = modeler.network_index(modeler.ports[1], 0)
+    if port_type == "lumped":
+        port_1_network_index = modeler.network_index(modeler.ports[0])
+        port_2_network_index = modeler.network_index(modeler.ports[1])
+    else:
+        port_1_network_index = modeler.network_index(modeler.ports[0], 0)
+        port_2_network_index = modeler.network_index(modeler.ports[1], 0)
     _ = modeler_data.get_antenna_metrics_data({port_1_network_index: 1.0})
     _ = modeler_data.get_antenna_metrics_data({port_2_network_index: None})
     antenna_params = modeler_data.get_antenna_metrics_data()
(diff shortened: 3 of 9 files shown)


dmarek-flex:
@lucas-flexcompute @momchil-flex I know this is super simple, but want to get this right before release. @momchil-flex expressed some concerns about having 2 formats for `get_task_name`. Right now we have two formats because RF users prefer no trailing `@0` but photon forge uses the trailing  `@0` . Any suggestions or opinions?

momchil-flex (approved):
Thanks @dmarek-flex. Personally I don't have a very strong preference either way, just hoping that we can simplify this. @lucas-flexcompute shouldn't photonforge just call some function that returns the port names whatever we decide them to be, rather than rely on the existence of something like a `@0` suffix?

daquinteroflex (approved):

dmarek-flex:
I added one restriction to port names, so that "@" is a reserved symbol. This way we can guarantee the task names are actually unique.

lucas-flexcompute:
> Thanks @dmarek-flex. Personally I don't have a very strong preference either way, just hoping that we can simplify this. @lucas-flexcompute shouldn't photonforge just call some function that returns the port names whatever we decide them to be, rather than rely on the existence of something like a `@0` suffix?

This is deeply embedded in PF… Because waveguides can be multimode, they always generate `f"{port_name}@{mode_number}"` ports. Adding a conditional clause everywhere this is used (to check the number of modes) would not be easy (or possible, to be honest) and would break backwards compatibility at this point.
I prefer to have a conversion wrapper if we ever use the TCM.

dmarek-flex:
> This is deeply embedded in PF… Because waveguides can be multimode, they always generate `f"{port_name}@{mode_number}"` ports. Adding a conditional clause everywhere this is used (to check the number of modes) would not be easy (or possible, to be honest) and would break backwards compatibility at this point.
> I prefer to have a conversion wrapper if we ever use the TCM.

Is there a plan for PhotonForge to allow/manage setting up lumped ports?

I am a little confused why we need two formats then if you are not using lumped ports. If photon forge always uses Modal ports and uses task names like `f"{port_name}@{mode_number}"`, it should have always worked?

lucas-flexcompute:
> Is there a plan for PhotonForge to allow/manage setting up lumped ports?
> 
> I am a little confused why we need two formats then if you are not using lumped ports. If photon forge always uses Modal ports and uses task names like `f"{port_name}@{mode_number}"`, it should have always worked?

We will use them in the (near?) future. I'm just not rushing any integration (incl. component modeler, lumped ports, and heat/charge) until all interfaces have stabilized so that we don't have compatibility problems like in 2.10 now.

dmarek-flex:
> We will use them in the (near?) future. I'm just not rushing any integration (incl. component modeler, lumped ports, and heat/charge) until all interfaces have stabilized so that we don't have compatibility problems like in 2.10 now.

Ah ok very reasonable :smile: . In that case though I propose getting rid of the "format" option, since photon forge always supplies port and mode index, the task name will always be as expected `f"{port_name}@{mode_number}"`. Once/if photon forge adds support for lumped ports, it will be up to PhotonForge to correctly leave the `mode_index` parameter unset in `get_task_name`. 

@lucas-flexcompute Do you foresee any issues in PhotonForge if we enforce port names to not include the `@` symbol?

lucas-flexcompute:
> @lucas-flexcompute Do you foresee any issues in PhotonForge if we enforce port names to not include the `@` symbol?

Not really, it will probably be handled automatically (by PF import/export functions) when we use a TCM. One question I have, though, is whether it is possible to mix (multi-)modal and lumped ports in either CM or TCM, but that's unrelated to this discussion.

dmarek-flex:
> One question I have, though, is whether it is possible to mix (multi-)modal and lumped ports in either CM or TCM

`ModalComponentModeler` only supports the modal ports. The `TerminalComponentModeler` allows for having a mixture of port types, currently lumped ports and wave ports (which are the RF version of modal ports with a lot more features around impedance calculation)

momchil-flex (approved):
Ok, I think this looks better now, thanks for these improvements @dmarek-flex !

dmarek-flex:
@daquinteroflex please take a quick glance at the `get_task_name`. I removed the format option that was introduced in the refactor. From the discussion above, it seems like @lucas-flexcompute is not using this method anyway and assumes the format of `f{port.name}@{mode_index} ` which will work fine now for both `Port` and `WavePort`. I should have updated this stuff when introducing multimodal wave ports, but anyways I think much simpler now and consolidated.

daquinteroflex:
Yep this sounds good if it will be managed by the wrapper at the pf side then!

PR #2965, opened 2025-11-05
feat: 2.10 nexus support

daquinteroflex:
Seems to be working fine.
```
(tidy3d-A) flexdaemon@dxps:~$ tidy3d configure --apikey xxx  --nexus-url http://localhost
Nexus configuration saved successfully to profile: profiles/nexus.toml
  API endpoint: http://localhost/tidy3d-api
  Website endpoint: http://localhost/tidy3d
  S3 endpoint: http://localhost:9000

Default profile set to 'nexus'. Tidy3D will now use these endpoints by default.
To switch back to production, use: config.set_default_profile(None) in Python
(tidy3d-A) flexdaemon@dxps:~$ cat ~/.config/tidy3d/config.toml 
default_profile = "nexus"

[web]
apikey = "xxx" # Tidy3D API key.
(tidy3d-A) flexdaemon@dxps:~$ ls ~/.config/tidy3d
config.toml  profiles
(tidy3d-A) flexdaemon@dxps:~$ ls ~/.config/tidy3d/profiles/
nexus.toml
(tidy3d-A) flexdaemon@dxps:~$ cat .config/tidy3d/profiles/nexus.toml 
[web]
website_endpoint = "http://localhost/tidy3d" # Tidy3D website URL.
api_endpoint = "http://localhost/tidy3d-api" # Tidy3D API base URL.

[web.env_vars]
AWS_ENDPOINT_URL_S3 = "http://localhost:9000"

```

```
 default_profile                                                                              │
│ │   └── 'nexus'                                                                                  │
│ ├── local_cache                                                                                  │
│ │   └── {                                                                                        │
│ │           'enabled': False,                                                                    │
│ │           'directory': '/home/flexdaemon/.cache/tidy3d/simulations',                           │
│ │           'max_size_gb': 10.0,                                                                 │
│ │           'max_entries': 0                                                                     │
│ │       }                                                                                        │
│ ├── logging                                                                                      │
│ │   └── {                                                                                        │
│ │           'level': 'WARNING',                                                                  │
│ │           'suppression': True                                                                  │
│ │       }                                                                                        │
│ ├── microwave                                                                                    │
│ │   └── {                                                                                        │
│ │           'suppress_rf_license_warning': False                                                 │
│ │       }                                                                                        │
│ ├── plugins                                                                                      │
│ │   └── {}                                                                                       │
│ ├── simulation                                                                                   │
│ │   └── {                                                                                        │
│ │           'use_local_subpixel': None                                                           │
│ │       }                                                                                        │
│ └── web                                                                                          │
│     └── {                                                                                        │
│             'apikey': 'xxx',                        │
│             'ssl_verify': False,                                                                 │
│             'enable_caching': False,                                                             │
│             'api_endpoint': 'http://localhost/tidy3d-api',                                       │
│             'website_endpoint': 'http://localhost/tidy3d',                                       │
│             's3_region': 'us-east-1',                                                            │
│             'timeout': 120,                                                                      │
│             'ssl_version': None,                                                                 │
│             'env_vars': {                                                                        │
│                 'AWS_ENDPOINT_URL_S3': 'http://localhost:9000'                                   │
│             }                                                                                    │
│         }                          
```

image
image

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR adds Nexus on-premises server support for Tidy3D, enabling enterprise customers to connect to custom deployments.

**Key changes:**
- Extended CLI `configure` command with `--nexus-url`, `--api-endpoint`, `--website-endpoint`, `--s3-region`, `--s3-endpoint`, and SSL/caching flags
- Added `default_profile` persistence to `config.toml` so Nexus configuration persists across sessions
- Extended legacy config migration to handle Nexus-related fields (`web_api_endpoint`, `s3_endpoint`, etc.)
- Updated S3 client to use custom endpoint URL from `web.env_vars` for MinIO/S3-compatible storage
- Added comprehensive documentation and tests

**Architecture:**
- Base config (`config.toml`) stores API key and `default_profile` setting
- Profile-specific settings are stored in `profiles/nexus.toml`
- Profile resolution priority: explicit parameter → environment variables → `default_profile` → "default"

- This PR is safe to merge with one minor edge case to address in the S3 endpoint URL derivation
- Score reflects well-tested feature with comprehensive documentation, but there's a minor URL parsing edge case in the CLI that could produce invalid S3 endpoints when nexus-url contains trailing slashes or ports
- tidy3d/web/cli/app.py - S3 endpoint derivation logic needs to handle edge cases

File Analysis

| tidy3d/web/cli/app.py | 3/5 | Extended configure command for Nexus support; S3 endpoint derivation has edge case issues with URLs containing ports or trailing slashes |
| tidy3d/config/manager.py | 5/5 | Added set_default_profile/get_default_profile methods and updated profile resolution logic; well-implemented with proper validation |
| tidy3d/config/loader.py | 5/5 | Added default profile persistence and auto-migration logic; clean implementation with proper error handling |
| tidy3d/config/legacy.py | 5/5 | Extended legacy config migration to support Nexus fields; correctly maps old flat keys to new nested structure |
| tidy3d/web/core/s3utils.py | 5/5 | Updated get_client() to configure custom S3 endpoint from web.env_vars; clean and correct implementation |
| tests/config/test_nexus_migration.py | 5/5 | Comprehensive tests for Nexus config migration covering legacy config parsing, auto-migration, and profile handling |
| docs/configuration/nexus.rst | 5/5 | Comprehensive documentation for Nexus environment configuration with examples for CLI, Python, and troubleshooting |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant CLI as tidy3d CLI
    participant Config as ConfigManager
    participant Loader as ConfigLoader
    participant Profile as profiles/nexus.toml
    participant Base as config.toml
    participant S3Client as boto3 S3 Client

    User->>CLI: tidy3d configure --nexus-url
    CLI->>CLI: Derive endpoints from nexus_url
    CLI->>Config: update_section(web, apikey)
    Config->>Base: Save API key
    CLI->>Config: switch_profile(nexus)
    CLI->>Config: update_section(web, endpoints)
    Config->>Profile: Save nexus-specific settings
    CLI->>Config: set_default_profile(nexus)
    Config->>Loader: set_default_profile(nexus)
    Loader->>Base: Write default_profile
    
    Note over User,S3Client: On next session
    
    User->>Config: import tidy3d
    Config->>Loader: init
    Loader->>Base: Read default_profile
    Loader-->>Config: profile is nexus
    Config->>Profile: Load nexus settings
    
    User->>S3Client: Upload or Download
    S3Client->>Config: Read web.env_vars
    S3Client->>S3Client: Configure endpoint_url from env_vars
```

### DIFF
--- a/docs/configuration/nexus.rst
+++ b/docs/configuration/nexus.rst
@@ -1,0 +1,594 @@
+.. currentmodule:: tidy3d
+
+Nexus Environment Configuration
+================================
+
+Overview
+--------
+
+The Nexus environment enables Tidy3D to connect to custom on-premises or private cloud deployments for enterprise customers.
+
+.. note::
+   This feature is for enterprise customers with custom Tidy3D deployments. Most users should use the standard API key setup with the default production environment. See the :doc:`../install` guide for standard setup instructions.
+
+Quick Start
+-----------
+
+**Option 1: Automatic Configuration** (Recommended)
+
+When you configure Nexus using the CLI, Tidy3D automatically sets it as your default profile:
+
+.. code-block:: bash
+
+   tidy3d configure --apikey YOUR_KEY \
+     --nexus-url http://your-server
+
+This automatically:
+
+- Sets API endpoint: ``http://your-server/tidy3d-api``
+- Sets Website endpoint: ``http://your-server/tidy3d``
+- Sets S3 endpoint: ``http://your-server:9000``
+- **Sets Nexus as the default profile** (persisted across sessions)
+
+After configuration, all your scripts will automatically use Nexus without needing ``config.switch_profile("nexus")``.
+
+**Option 2: Manual Profile Switching** (for temporary use):
+
+For local development on localhost without persistent configuration:
+
+.. code-block:: python
+
+   from tidy3d.config import config
+   config.switch_profile("nexus")
+
+This uses the built-in nexus profile with localhost endpoints (127.0.0.1:5000 for API, 127.0.0.1:9000 for S3).
+
+**Advanced setup** with individual endpoints:
+
+.. code-block:: bash
+
+   tidy3d configure --apikey YOUR_KEY \
+     --api-endpoint http://your-server/tidy3d-api \
+     --website-endpoint http://your-server/tidy3d
+
+Or configure only Nexus endpoints (preserves existing API key):
+
+.. code-block:: bash
+
+   tidy3d configure --nexus-url http://your-server
+
+Configuration
+-------------
+
+Command Syntax
+~~~~~~~~~~~~~~
+
+.. code-block:: bash
+
+   tidy3d configure [--apikey <key>] [--nexus-url <url> | --api-endpoint <url> --website-endpoint <url>] [OPTIONS]
+
+**Options:**
+
+* ``--apikey <key>``: API key (prompts if not provided and no Nexus options given)
+* ``--nexus-url <url>``: Nexus base URL (automatically sets api=/tidy3d-api, web=/tidy3d, s3=:9000)
+* ``--api-endpoint <url>``: Nexus API server URL (e.g., http://server/tidy3d-api)
+* ``--website-endpoint <url>``: Nexus web interface URL (e.g., http://server/tidy3d)
+* ``--s3-region <region>``: S3 region (default: us-east-1)
+* ``--s3-endpoint <url>``: S3 storage URL (e.g., http://server:9000)
+* ``--ssl-verify`` / ``--no-ssl-verify``: SSL verification (default: enabled for HTTPS)
+* ``--enable-caching`` / ``--no-caching``: Server-side result caching
+
+Examples
+~~~~~~~~
+
+.. code-block:: bash
+
+   # Simple configuration using nexus-url (recommended)
+   tidy3d configure --apikey YOUR_KEY \
+     --nexus-url http://nexus.company.com
+
+   # Configure individual endpoints
+   tidy3d configure --apikey YOUR_KEY \
+     --api-endpoint http://nexus.company.com/tidy3d-api \
+     --website-endpoint http://nexus.company.com/tidy3d
+
+   # Add Nexus to existing configuration (preserves API key)
+   tidy3d configure --nexus-url http://nexus.company.com
+
+   # Full configuration with all options
+   tidy3d configure --apikey YOUR_KEY \
+     --api-endpoint https://api.company.com/tidy3d-api \
+     --website-endpoint https://tidy3d.company.com \
+     --s3-region eu-west-1 \
+     --s3-endpoint https://s3.company.com:9000 \
+     --ssl-verify \
+     --enable-caching
+
+Configuration File
+~~~~~~~~~~~~~~~~~~
+
+**Location:**
+
+When you configure Nexus, your custom settings are saved to a profile file:
+
+- Linux/macOS: ``~/.config/tidy3d/profiles/nexus.toml``
+- Windows: ``C:\Users\username\.config\tidy3d\profiles\nexus.toml``
+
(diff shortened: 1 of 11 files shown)


frederikschubertflex:
> `To switch back to production, use: config.set_default_profile(None) in Python`

Should we also provide a way to switch back with the CLI?

frederikschubertflex:
~~I find it a bit strange that we configure the nexus backend and S3 storage differently (arg vs. env var).~~ Just saw that this was only in your example.

docs/configuration/nexus.rst
```diff
@@ -0,0 +1,583 @@
...
+Nexus Environment Configuration
+================================
+
+Overview
+--------
+
+The Nexus environment enables Tidy3D to connect to custom on-premises or private cloud deployments for enterprise customers.
+
+.. note::
+   This feature is for enterprise customers with custom Tidy3D deployments. Most users should use the standard API key setup with the default production environment.
```
frederikschubertflex: Maybe provide a link to the standard API key setup docs?

frederikschubertflex (approved):

docs/configuration/nexus.rst
daquinteroflex: done and also added a cli option

tidy3d/config/manager.py
yaugenst-flex, line 1: Let's add a message in e.g. `ConfigManager.__init__()` after profile resolution with something like
```py
if self._profile != "default":
  log.info(f"Using configuration profile: '{self._profile}'", log_once=True)
```
to make sure users are aware if they're using a non-default profile.

yaugenst-flex (commented):
Thanks @daquinteroflex overall LGTM

tidy3d/config/manager.py
daquinteroflex: oh I forgot, thanks

tidy3d/config/sections.py
yaugenst-flex, line 1: Oh hold on, I'm just realizing this adds a ton of new persisted fields! I think we should not do that, this will make it really hard to migrate the config later, and these specifically seem like fields that are likely to change. If someone calls `config.save()`, this will cause all of the prod endpoint setup to leak into the default `config.toml`, we don't want that behavior. The nexus-specific values should only exist in the nexus-specific profile, not clutter the base config.

yaugenst-flex (changes requested):

tidy3d/config/sections.py
yaugenst-flex: I think it makes more sense to have the nexus cli path simply write the nexus profile and set it as the default without going through `config.save()`?

yaugenst-flex: The more I think about it, the current approach maybe isn't that bad.. the CLI does handle it correctly, so the problem really only happens if you manually call `config.save()` on the default profile after loading the nexus endpoints somehow..

yaugenst-flex (approved):

tidy3d/config/sections.py
daquinteroflex: yeah I wasn't sure how we wanted to handle persisting the configuration in the profile. I felt this was a good compromise given it's the same behaviour as what we released in 2.9.2.

If we can find a better approach that still persists the profile would be down to change it! I was trying to reduce the code changes whilst broadly mantaining web config compatibility. Because then it can still reuse the loading logic to some degree. But yeah forwards compatibility is a good point, just not sure how that can be evaded when you're still writing to files and persisting values from a custom config.

tidy3d/config/manager.py
yaugenst-flex: could you point me to where this is implemented now? i can't find it

daquinteroflex: turns out I hadn't pushed this change! Good catch, thanks. So we sure only `log.info` is enough as only `log.warning` will print. Guess it doesn't matter if they check the web error traceback properly though

PR #2969, opened 2025-11-05
feat(source_time): broadband pulse

weiliangjin2021:
- [x] add tests

<h2>Greptile Overview</h2>

Updated On: 2025-11-05 20:52:36 UTC

<h3>Greptile Summary</h3>

This PR introduces `BroadbandPulse`, a new source time dependence class that enables exciting simulations across a wide frequency spectrum. The feature is implemented as a wrapper around the tidy3d-extras licensed implementation.

**Key changes:**
- Added `BroadbandPulse` class in `tidy3d/components/source/time.py` with configurable frequency range and minimum amplitude
- Refactored tidy3d-extras checking logic into reusable helper functions (`_check_tidy3d_extras_available` and `check_tidy3d_extras_licensed_feature`)
- Updated documentation and CHANGELOG to reflect the new feature

**Issues found:**
- **Syntax error** in `tidy3d/packaging.py:244` - missing 'f' prefix on f-string causes literal `{exc!s}` to appear in error message instead of being interpolated

- This PR is mostly safe to merge with one syntax error that needs fixing
- The implementation is well-structured with proper licensing checks and documentation. However, there's a syntax error in the error message formatting that will cause incorrect error messages to be displayed to users when tidy3d-extras fails to load. This is a user-facing bug that should be fixed before merge.
- Pay attention to `tidy3d/packaging.py` which contains a syntax error on line 244

File Analysis

| tidy3d/components/source/time.py | 4/5 | Added `BroadbandPulse` class that wraps tidy3d-extras implementation for wide frequency spectrum excitation |
| tidy3d/packaging.py | 2/5 | Refactored tidy3d-extras checking logic with new helper functions; contains f-string formatting bug on line 244 |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant BroadbandPulse
    participant Validator
    participant PackagingModule
    participant TidyExtras

    User->>BroadbandPulse: Instantiate with freq_range, minimum_amplitude
    BroadbandPulse->>Validator: _check_broadband_pulse_available()
    Validator->>PackagingModule: check_tidy3d_extras_licensed_feature("BroadbandPulse")
    PackagingModule->>PackagingModule: _check_tidy3d_extras_available()
    
    alt tidy3d-extras not loaded
        PackagingModule->>TidyExtras: import tidy3d_extras
        alt Import fails
            TidyExtras-->>PackagingModule: ImportError
            PackagingModule-->>Validator: Tidy3dImportError
            Validator-->>User: Error: Install tidy3d-extras
        else Import succeeds
            TidyExtras-->>PackagingModule: tidy3d_extras_mod
            PackagingModule->>PackagingModule: Check version and features
            alt BroadbandPulse not in features
                PackagingModule-->>Validator: Tidy3dImportError
                Validator-->>User: Error: Feature not licensed
            else Feature available
                PackagingModule->>PackagingModule: Cache tidy3d_extras["mod"]
                Validator-->>BroadbandPulse: Validation passed
            end
        end
    else tidy3d-extras already loaded
        PackagingModule->>PackagingModule: Check features list
        alt BroadbandPulse not in features
            PackagingModule-->>Validator: Tidy3dImportError
            Validator-->>User: Error: Feature not licensed
        else Feature available
            Validator-->>BroadbandPulse: Validation passed
        end
    end
    
    BroadbandPulse->>User: Instance created
    
    User->>BroadbandPulse: Call amp_time(time) or amp_freq(freq)
    BroadbandPulse->>BroadbandPulse: Access _source property
    BroadbandPulse->>TidyExtras: extension.BroadbandPulse(...)
    TidyExtras-->>BroadbandPulse: BroadbandPulse instance
    BroadbandPulse->>TidyExtras: _source.amp_time(time)
    TidyExtras-->>BroadbandPulse: Complex amplitude
    BroadbandPulse-->>User: Return amplitude
```

### DIFF
--- a/tidy3d/components/source/time.py
+++ b/tidy3d/components/source/time.py
@@ -21,6 +21,7 @@ from tidy3d.components.viz import add_ax_if_none
 from tidy3d.constants import HERTZ
 from tidy3d.exceptions import ValidationError
 from tidy3d.log import log
+from tidy3d.packaging import check_tidy3d_extras_licensed_feature, tidy3d_extras
 
 # how many units of ``twidth`` from the ``offset`` until a gaussian pulse is considered "off"
 END_TIME_FACTOR_GAUSSIAN = 10
@@ -605,4 +606,80 @@ class CustomSourceTime(Pulse):
         return np.max(t_non_zero)
 
 
-SourceTimeType = Union[GaussianPulse, ContinuousWave, CustomSourceTime]
+class BroadbandPulse(SourceTime):
+    """A source time injecting significant energy in the entire custom frequency range."""
+
+    freq_range: FreqBound = pydantic.Field(
+        ...,
+        title="Frequency Range",
+        description="Frequency range where the pulse should have significant energy.",
+        units=HERTZ,
+    )
+    minimum_amplitude: float = pydantic.Field(
+        0.3,
+        title="Minimum Amplitude",
+        description="Minimum amplitude of the pulse relative to the peak amplitude in the frequency range.",
+        gt=0.05,
+        lt=0.5,
+    )
+    offset: float = pydantic.Field(
+        0.0,
+        title="Offset",
+        description="An automatic time delay of the peak value of the pulse has been applied under the hood "
+        "to ensure smooth ramping up of the pulse at time = 0. This offfset is added on top of the automatic time delay "
+        "in units of 1 / [``2pi * (freq_range[1] - freq_range[0])``].",
+    )
+
+    @pydantic.validator("freq_range", always=True)
+    def _validate_freq_range(cls, val):
+        """Validate that freq_range is positive and properly ordered."""
+        if val[0] <= 0 or val[1] <= 0:
+            raise ValidationError("Both elements of 'freq_range' must be positive.")
+        if val[1] <= val[0]:
+            raise ValidationError(
+                f"'freq_range[1]' ({val[1]}) must be greater than 'freq_range[0]' ({val[0]})."
+            )
+        return val
+
+    @pydantic.root_validator()
+    def _check_broadband_pulse_available(cls, values):
+        """Check if BroadbandPulse is available."""
+        check_tidy3d_extras_licensed_feature("BroadbandPulse")
+        return values
+
+    @cached_property
+    def _source(self):
+        """Implementation of broadband pulse."""
+        return tidy3d_extras["mod"].extension.BroadbandPulse(
+            fmin=self.freq_range[0],
+            fmax=self.freq_range[1],
+            minRelAmp=self.minimum_amplitude,
+            amp=self.amplitude,
+            phase=self.phase,
+            offset=self.offset,
+        )
+
+    def end_time(self) -> float:
+        """Time after which the source is effectively turned off / close to zero amplitude."""
+        return self._source.end_time(END_TIME_FACTOR_GAUSSIAN)
+
+    def amp_time(self, time: float) -> complex:
+        """Complex-valued source amplitude as a function of time."""
+        return self._source.amp_time(time)
+
+    def amp_freq(self, freq: float) -> complex:
+        """Complex-valued source amplitude as a function of frequency."""
+        return self._source.amp_freq(freq)
+
+    def frequency_range_sigma(self, sigma: float = DEFAULT_SIGMA) -> FreqBound:
+        """Frequency range where the source amplitude is within ``exp(-sigma**2/2)`` of the peak amplitude."""
+        return self._source.frequency_range(sigma)
+
+    def frequency_range(self, num_fwidth: float = DEFAULT_SIGMA) -> FreqBound:
+        """Delegated to `frequency_range_sigma(sigma=num_fwidth)` for computing the frequency range where the source amplitude
+        is within ``exp(-num_fwidth**2/2)`` of the peak amplitude.
+        """
+        return self.frequency_range_sigma(num_fwidth)
+
+
+SourceTimeType = Union[GaussianPulse, ContinuousWave, CustomSourceTime, BroadbandPulse]
--- a/tidy3d/packaging.py
+++ b/tidy3d/packaging.py
@@ -16,7 +16,6 @@ import numpy as np
 from tidy3d.config import config
 
 from .exceptions import Tidy3dImportError
-from .log import log
 from .version import __version__
 
 vtk = {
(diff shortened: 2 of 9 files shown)


momchil-flex (approved):

tidy3d/packaging.py
```diff
@@ -193,47 +261,15 @@ def _fn(*args: Any, **kwargs: Any):
...
-
-                else:
-                    version = tidy3d_extras_mod.__version__
-
-                    if version is None:
-                        tidy3d_extras["mod"] = None
-                        tidy3d_extras["use_local_subpixel"] = False
-                        raise Tidy3dImportError(
-                            "The package 'tidy3d-extras' did not initialize correctly, "
-                            "likely due to an invalid API key."
```
caseyflex: @yaugenst-flex some of the messages here got a bit worse in this commit:

https://github.com/flexcompute/tidy3d/commit/8bdf76ea084f70e4d05bf90be391f9ab231cbf86#diff-1f3286e98a479fce43d65741bd9d499d913a667b9cd64569be1d092df4548459

tidy3d/packaging.py
```diff
@@ -193,47 +261,15 @@ def _fn(*args: Any, **kwargs: Any):
...
-
-                    tidy3d_extras["mod"] = tidy3d_extras_mod
-                    tidy3d_extras["use_local_subpixel"] = "local_subpixel" in features
+            try:
+                _check_tidy3d_extras_available()
+            except Tidy3dImportError as exc:
+                tidy3d_extras["use_local_subpixel"] = False
+                if preference is True:
+                    raise Tidy3dImportError(
+                        f"Failed to load 'tidy3d-extras' for local subpixel support. {exc!s}"
```
caseyflex: Thanks @weiliangjin2021 ! I think the new handling is a nice generalization and improvement of the old handling. Can you add in this error message that `config.simulation.use_local_subpixel` can be used to supress this error, like here:

https://github.com/flexcompute/tidy3d/commit/8bdf76ea084f70e4d05bf90be391f9ab231cbf86#diff-1f3286e98a479fce43d65741bd9d499d913a667b9cd64569be1d092df4548459

tidy3d/packaging.py
```diff
@@ -193,47 +261,15 @@ def _fn(*args: Any, **kwargs: Any):
...
-                            "the version of 'tidy3d'. You can install the correct "
-                            "version using 'pip install tidy3d[extras]'."
-                        )
-
-                    features = tidy3d_extras_mod.extension._features()
-
-                    tidy3d_extras["mod"] = tidy3d_extras_mod
-                    tidy3d_extras["use_local_subpixel"] = "local_subpixel" in features
+            try:
+                _check_tidy3d_extras_available()
```
caseyflex: in this case, you need to set `tidy3d_extras["use_local_subpixel"] = True`, right? The backend tidy3d-extras tests should catch this?

caseyflex: also, can you just make this use your `check_tidy3d_extras_licensed_feature` function too?

caseyflex (changes requested):

tidy3d/packaging.py
yaugenst-flex: aaah good catch, will fix

tidy3d/packaging.py
```diff
@@ -183,6 +183,74 @@ def get_numpy_major_version(module=np):
...
+                    "likely due to an invalid API key."
+                )
+
+            if version != __version__:
+                log.warning(
+                    "The package 'tidy3d-extras' is required for this "
+                    "operation. The version of 'tidy3d-extras' should match "
+                    "the version of 'tidy3d'. You can install the correct "
+                    "version using 'pip install tidy3d[extras]'."
+                )
```
yaugenst-flex: Are we sure this should be a warning? What if the the user is using a version of tidy3d-extras where `BroadbandPulse` doesn't exist? At least the function docstring suggests version matching, so why would we not want to raise an `ImportError` here?

tidy3d/packaging.py
```diff
@@ -183,6 +183,74 @@ def get_numpy_major_version(module=np):
...
 
+def _check_tidy3d_extras_available():
+    """Helper function to check if 'tidy3d-extras' is available and version matched.
+
+    Raises
+    ------
+    Tidy3dImportError
+        If tidy3d-extras is not available or not properly initialized.
+    """
+    if tidy3d_extras["mod"] is None:
```
yaugenst-flex: All logic is under this branch, so every useful path is double-indented. This should just early return.

tidy3d/packaging.py
```diff
@@ -183,6 +183,74 @@ def get_numpy_major_version(module=np):
...
+    ------
+    Tidy3dImportError
+        If the feature is not available with your license.
+    """
+
+    try:
+        _check_tidy3d_extras_available()
+    except Tidy3dImportError as exc:
+        raise Tidy3dImportError(f"Failed to load 'tidy3d-extras'. {exc!s}") from exc
+    else:
```
yaugenst-flex: This `else` is redundant, the above already re-raises.

tidy3d/packaging.py
```diff
@@ -183,6 +183,74 @@ def get_numpy_major_version(module=np):
...
+        features = tidy3d_extras["mod"].extension._features()
+        if feature_name not in features:
+            raise Tidy3dImportError(
+                f"The feature '{feature_name}' is not available with your license. "
+                "Please contact Tidy3D support, or upgrade your license."
+            )
```
yaugenst-flex: Any `AttributeError` coming from this is currently raises as a license issue, but this is misleading when the specific build of tidy3d-extras predates the feature. These errors should be caught and properly converted for version mismatches.

tidy3d/components/source/time.py
```diff
@@ -605,4 +606,83 @@ def end_time(self) -> Optional[float]:
...
+        title="Minimum Amplitude",
+        description="Minimum amplitude of the pulse relative to the peak amplitude in the frequency range.",
+        gt=0.05,
+        lt=0.5,
+    )
+    offset: float = pydantic.Field(
+        5.0,
+        title="Offset",
+        description="Time delay of the maximum value of the "
+        "pulse in units of 1 / (``2pi * fwidth``).",
```
yaugenst-flex: Looks like this was copy-pasted from `Pulse`, but `BroadbandPulse` doesn't have an `fwidth`?

tidy3d/components/source/time.py
```diff
@@ -605,4 +606,83 @@ def end_time(self) -> Optional[float]:
...
+        """Complex-valued source amplitude as a function of frequency."""
+        return self._source.amp_freq(freq)
+
+    def frequency_range_sigma(self, sigma: float = DEFAULT_SIGMA) -> FreqBound:
+        """Frequency range where the source amplitude is within ``exp(-sigma**2/2)`` of the peak amplitude."""
+        return self._source.frequency_range(sigma)
+
+    def frequency_range(self, num_fwidth: float = DEFAULT_SIGMA) -> FreqBound:
+        """Frequency range where the source amplitude is within ``exp(-sigma**2/2)`` of the peak amplitude."""
+        return self.frequency_range_sigma(num_fwidth)
```
yaugenst-flex, line 682: This just delegates to `frequency_range_sigma` but it keeps the sigma-based docstring, which is misleading. This should provide the correct description for the function.

yaugenst-flex (changes requested):

tidy3d/packaging.py
weiliangjin2021: This part is copied over. @caseyflex could you comment on why it's a warning?

For BroadbandPulse, if it's a version where this feature doesn't exist, the feature check below will raise an error.

caseyflex: I think version matching can be recommended but not required, but not sure. Seems like it will usually work to use a different version, and if it doesn’t, at least we raised a warning. I always recommend people leave warnings on and pay attention to them, but I understand they don’t always do so. So there may be a reason to raise an error.

caseyflex: Yeah in fact let’s just change this to an error

tidy3d/packaging.py
weiliangjin2021: `_features()` also includes license check. A feature can be missing for 2 possible reasons:

- Doesn't exist in older version
- User's license doesn't include this feature.

Maybe we should have a separate `_feature_all()` function to check if the feature exists regardless of the license check?

tidy3d/packaging.py
weiliangjin2021: Yeah, I think it'll work in most cases without version matching. One confusing scenario is that if we updated binary in a new version, users might see different subpixel/pulse behavior locally from the server run. So perhaps we should enforce this?

tidy3d/packaging.py
weiliangjin2021: but if we enforce version matching, a separation function wouldn't be necessary.

tidy3d/components/source/time.py
weiliangjin2021: Actually, that's the part I have been working on since yesterday: offset is not well defined in this pulse. I need to find a better definition.

tidy3d/packaging.py
caseyflex: Yeah exactly - it could lead to obscure differences in behavior or hidden bugs

tidy3d/packaging.py
weiliangjin2021: In your current code, it doesn't raise an error if the subpixel is not licensed: it just silently disables local subpixel. Do you think we should instead raise an error here?

weiliangjin2021: (the error is raised in the revised version)

tidy3d/components/source/time.py
weiliangjin2021: Now it's "Delegated to `frequency_range_sigma(sigma=num_fwidth)` for computing the frequency range where the source amplitude is within ``exp(-num_fwidth**2/2)`` of the peak amplitude"

tidy3d/components/source/time.py
weiliangjin2021: Updated

caseyflex (approved):
Great! So now if `use_local_subpixel` is None, then the feature will be used if everything works (package installed + licensing for that feature), and otherwise it will not be used. I think this makes sense going forward.

@daquinteroflex The install option `tidy3d[extras]` is still commented out in `pyproject.toml`. It should be uncommented before release. However I think it will cause an issue with `poetry lock` because the version is not yet available on pypi. Did we ever decide how to resolve this?

tidy3d/packaging.py
weiliangjin2021: Is the change here fine @caseyflex

daquinteroflex:
Hi @caseyflex I was actually testing just that yesterday https://github.com/flexcompute/tidy3d/pull/2967/files

yaugenst-flex (approved):

tidy3d/packaging.py
caseyflex: I think so, thanks!

PR #2978, opened 2025-11-08
SCRF-1630: Refactor RF webapi for new workflow

momchil-flex:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR successfully refactors the RF webapi to use a simplified workflow for Component Modeler batch tasks, migrating from legacy `tidy3d/tasks/{id}/batch-*` endpoints to new `rf/task/{id}/*` endpoints and removing ~400 lines of complex async split/validation logic.

## Key Changes

- **Simplified batch workflow**: Removed the `_upload_component_modeler_subtasks()` function that handled async job splitting and multi-phase validation (~170 lines), as this logic now happens server-side
- **Unified task handling**: Introduced `WebTask` base class with `BatchTask` and `SimulationTask` subclasses, enabling consistent routing through `TaskFactory.get()`
- **Endpoint migration**: 
  - Task creation: `POST rf/task` (new) vs `POST tidy3d/projects/{folder}/tasks` (old)
  - Batch detail: `GET rf/task/{id}/statistics` (new) vs `GET tidy3d/tasks/{id}/batch-detail` (old)
  - Batch submit: `POST rf/task/{id}/submit` (new) vs `POST tidy3d/projects/{id}/batch-submit` (old)
- **Removed postprocessing logic**: Eliminated separate postprocess phase and auto-trigger logic from `Job` and `Batch.monitor()` classes
- **Validation deferral**: Component modeler simulations now use `validate=False` in intermediate `updated_copy()` calls to defer validation, with final validation happening in the `base_sim` property getter (terminal.py:541)
- **Proper error handling**: `BatchTask.submit()` now raises `NotImplementedError` for unsupported `pay_type`/`priority` parameters instead of silently ignoring them

## Implementation Notes

The validation deferral pattern is correctly implemented: intermediate simulation copies skip validation for performance, but the `base_sim` property (accessed during `validate_pre_upload()`) performs full validation before upload. Grid size checks were moved earlier to run on the validated `base_sim` rather than on individual unvalidated simulations.

- This PR is reasonably safe to merge with proper testing - the refactor is well-structured and removes significant complexity
- Score of 4 reflects: (1) Large refactor removing ~400 lines and migrating to new endpoints represents moderate risk despite clean implementation. (2) The previously reported critical bug with `pay_type`/`priority` parameters has been properly fixed with `NotImplementedError` exceptions. (3) Validation deferral pattern is correctly implemented with proper final validation. (4) Limited visibility into backend API changes and how new `rf/task/*` endpoints handle edge cases. (5) Removal of complex async split/validation logic appears sound but should be verified with integration tests.
- All files look good - no specific files require special attention. Integration testing recommended to verify new RF endpoint behavior.

File Analysis

| tidy3d/web/api/webapi.py | 4/5 | Removed complex async split/validation logic (~200 lines), simplified RF batch handling to use new `rf/task/*` endpoints, removed `_upload_component_modeler_subtasks()` function |
| tidy3d/web/core/task_core.py | 5/5 | Introduced `WebTask` base class, migrated `BatchTask` to use `rf/task/{id}/*` endpoints, properly raises `NotImplementedError` for unsupported `pay_type`/`priority` params |
| tidy3d/web/api/container.py | 5/5 | Removed Component Modeler-specific postprocess logic and auto-trigger functionality (~100 lines cleanup), simplified status handling |
| tidy3d/plugins/smatrix/component_modelers/terminal.py | 4/5 | Added `validate=False` to intermediate `updated_copy()` calls to defer validation, moved grid size checks to run on validated `base_sim`, final validation happens in `base_sim` property (line 541) |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant webapi
    participant WebTask
    participant BatchTask
    participant API as RF API (rf/task/*)

    User->>webapi: upload(component_modeler)
    webapi->>webapi: Check task type (RF)
    webapi->>WebTask: create(task_type="RF")
    WebTask->>API: POST rf/task (create group)
    API-->>WebTask: batchId/groupId
    WebTask->>API: Upload modeler.hdf5.gz to S3
    webapi->>webapi: estimate_cost(resource_id)
    webapi-->>User: return batch_id

    User->>webapi: start(batch_id)
    webapi->>BatchTask: get(batch_id)
    BatchTask->>API: GET rf/task/{id}/statistics (verify exists)
    API-->>BatchTask: BatchDetail
    webapi->>BatchTask: submit()
    Note right of BatchTask: pay_type/priority raise<br/>NotImplementedError if provided
    BatchTask->>API: POST rf/task/{id}/submit
    API-->>BatchTask: Response

    User->>webapi: monitor(batch_id)
    
    loop Poll Status
        webapi->>BatchTask: detail()
        BatchTask->>API: GET rf/task/{id}/statistics
        API-->>BatchTask: BatchDetail
        Note right of webapi: No auto-postprocess<br/>triggering (removed)
    end

    User->>webapi: download(batch_id)
    webapi->>BatchTask: get_data_hdf5()
    BatchTask->>API: Download modeler_data.hdf5.gz from S3
    API-->>User: Data file
```


tidy3d/plugins/smatrix/component_modelers/terminal.py
```diff
@@ -382,18 +384,17 @@ def matrix_indices_run_sim(self) -> tuple[NetworkIndex, ...]:
     def sim_dict(self) -> SimulationMap:
         """Generate all the :class:`.Simulation` objects for the port parameter calculation."""
 
+        # Check base simulation for grid size at ports
+        TerminalComponentModeler._check_grid_size_at_ports(self.base_sim, self._lumped_ports)
+        TerminalComponentModeler._check_grid_size_at_wave_ports(self.base_sim, self._wave_ports)
+
```
dmarek-flex, line 390: good

dmarek-flex (commented):
All your changes to TCM seem good to me

tidy3d/web/api/webapi.py
```diff
@@ -1229,39 +1023,16 @@ def download(
     if _is_modeler_batch(task_id):
         # Use a more descriptive default filename for component modeler downloads.
         # If the caller left the default as 'simulation_data.hdf5', prefer 'cm_data.hdf5'.
+        # TODO: seems like the default should then be maybe set to None and defined somewhere else
+        # per task type?
```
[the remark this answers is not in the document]
momchil-flex: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/flexcompute/tidy3d/pull/2978#discussion_r2514728156)

tidy3d/web/__init__.py
```diff
@@ -30,7 +30,6 @@
     load,
     load_simulation,
     monitor,
-    postprocess_start,
```
daquinteroflex, line 33: :tada: :tada:

tidy3d/web/api/webapi.py
```diff
@@ -164,167 +156,6 @@ def _batch_detail_error(resource_id: str) -> Optional[WebError]:
         return WebError(error_msg)
 
 
-def _upload_component_modeler_subtasks(
```
daquinteroflex, line 167: :tada:

tidy3d/web/api/webapi.py
```diff
@@ -1593,36 +1349,10 @@ def _status_to_stage(status: str) -> tuple[str, int]:
...
-            elif postprocess_status == "queued":
-                progress.update(p_post, completed=0.22)
-            elif postprocess_status == "preprocess":
-                progress.update(p_post, completed=0.33)
-            elif postprocess_status == "running":
-                progress.update(p_post, completed=0.55)
-            elif postprocess_status in ERROR_STATES:
-                raise WebError(
-                    f"Batch {batch_id} terminated. Please contact customer support and provide this Component Modeler batch ID: '{batch_id}'"
-                )
```
daquinteroflex, line 1623: That's great a single status

tidy3d/web/core/http_util.py
```diff
@@ -42,7 +41,8 @@ class ResponseCodes(Enum):
 
 def get_version() -> str:
     """Get the version for the current environment."""
-    return core_config.get_version()
+    # return core_config.get_version()
+    return "2.10.0rc2.1"
```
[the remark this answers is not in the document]
daquinteroflex: webapi implementation issue?

tidy3d/web/core/task_core.py
```diff
@@ -246,20 +243,24 @@ def create(
...
-
-        resp = http.post(f"{projects_endpoint}/{folder.folder_id}/tasks", payload)
+
+        if task_type == "RF":
+            payload = {
+                "groupName": task_name,
+                "folderId": folder.folder_id,
+                "fileType": file_type,
+            }
+            resp = http.post("rf/task", payload)
```
daquinteroflex, line 209: Feel the compromise in the naming

tidy3d/web/core/task_info.py
```diff
@@ -295,14 +234,14 @@ class BatchDetail(TaskBase):
...
         tasks: A list of `BatchMember` objects, one for each task in the batch.
+        taskType: The type of tasks contained in the batch.
     """
 
     refId: str = None
     optimizationId: str = None
     groupId: str = None
     name: str = None
     status: str = None
-    totalStatus: BatchStatus = None
```
daquinteroflex, line 305: wohoo!

daquinteroflex:
So wanted to just check a few final things before merging:

image

So we're not giving them a way to check a batch sub-simulation log if an error occurs out of curiosity? Maybe this could be a future thing? I guess the expectation is that this does not happen in general if it passed the validation, but do you think it will propagate cases where it diverges etc?

It looks like it's handling properly a failing initial simulation in terms of the monitoring, but this is maybe something worth discussing 
image

image
It looks really nice other than those questions, and probably good to merge

daquinteroflex (approved):

marc-flex (approved):

tidy3d/components/tcad/mesher.py
```diff
@@ -24,3 +24,6 @@ class VolumeMesher(Tidy3dBaseModel):
 
     def _get_simulation_types(self) -> list[TCADAnalysisTypes]:
         return [TCADAnalysisTypes.MESH]
+
+    def validate_pre_upload(self):
+        return
```
yaugenst-flex, line 32: Would be nice if this had a comment for why this is an empty override

tidy3d/web/api/states.py
yaugenst-flex, line 100: Isn't this missing the new error states? If the webapi returns any of the new error states then this will lead to a `KeyError`

tidy3d/web/api/webapi.py
```diff
@@ -889,30 +706,13 @@ def start(
...
-        return
-
     if priority is not None and (priority < 1 or priority > 10):
         raise ValueError("Priority must be between '1' and '10' if specified.")
-    task = SimulationTask.get(task_id)
+
+    if _is_modeler_batch(task_id):
+        task = BatchTask(task_id)
+    else:
+        task = SimulationTask.get(task_id)
```
yaugenst-flex: Wouldn't we need similar handling also for e.g. `delete`?

tidy3d/web/api/webapi.py
```diff
@@ -1780,11 +1504,17 @@ def estimate_cost(
     console = get_logging_console() if verbose else None
 
     if _is_modeler_batch(task_id):
-        d = _batch_detail(task_id)
-        status = d.totalStatus.value
+        batch = BatchTask(task_id)
+        _ = batch.check(solver_version=solver_version)
+        detail = batch.detail()
+        status = detail.status.lower()
+        while status not in ALL_POST_VALIDATE_STATES:
```
yaugenst-flex, line 1479: This will also hang for the new error states I think? Wouldn't this need to check `ERROR_STATES` too?

yaugenst-flex (commented):
Since we have this branching between regular and component modeler tasks, wouldn't all the webapi-related functions now need to gate behind `_is_modeler_batch`, e.g. `download_simulation()`, `get_run_info()`, etc, to then call the RF endpoints? Currently only `download()` and `upload()` seem to be handled correctly?

momchil-flex:
> Since we have this branching between regular and component modeler tasks, wouldn't all the webapi-related functions now need to gate behind `_is_modeler_batch`, e.g. `download_simulation()`, `get_run_info()`, etc, to then call the RF endpoints? Currently only `download()` and `upload()` seem to be handled correctly?

Yeah you are right.

Maybe before making any changes in that direction, we can discuss what this should look like? Either in a perfect world or, and more relevant, given what we currently have as backend and endpoints.

On one hand, it feels like we shouldn't be working with task id-s directly but rather with something like the existing `SimulationTask` and `BatchTask` objects, i.e. passing around something like a `WebTask` subclass. However, for the user-facing functions, using a `task_id` is quite deeply ingrained, and one could argue is the right approach as e.g. the user can copy the task id directly from GUI? On the other hand wrapping it in a `SimulationTask(task_id)` before using all the web functions is maybe not that much to ask for - but definitely a breaking change.

The other thing I'm wondering about is avoiding all the api calls just to determine if something is a batch or not based on the task id. One option that comes to mind is to use the prefixes (e.g. currently modeler tasks are `sid-***`) but that doesn't strike me as a particularly good choice.

tidy3d/web/api/webapi.py
momchil-flex: Introduced the delete endpoint which MC has implemented and raised NotImplementedError-s everywhere else.

tidy3d/web/api/webapi.py
momchil-flex: Indeed, and it seems like it was a problem in the `develop` version too - when a metadata pipeline errors, this would just hang. My best guess why we haven't encountered this is that it was introduced in the recent rc-s, and for existing solvers where metadata lambdas exist, the error is thrown by the api (while it is not directly thrown if it goes to pipeline processing...)

tidy3d/web/api/states.py
momchil-flex: added, thanks

momchil-flex:
So apart from addressing all review comments, I added some big updates along the lines of unifying SimulationTask and BatchTask and avoiding too many api calls to determine if something is a `batch` or not. These were introduced [in this commit](https://github.com/flexcompute/tidy3d/pull/2978/commits/24fe116f9f44a57e5d917403acb3df5076d7fe44), notably a [TaskFactory](https://github.com/flexcompute/tidy3d/pull/2978/commits/24fe116f9f44a57e5d917403acb3df5076d7fe44#diff-2f024019772f548396269e338b610c794d176828a11efc814eeed37abe1efa40R972-R986) getter that creates the appropriate SimulationTask or BatchTask, and stores a registry mapping of task id-s to whether they are batch or simulation, so the web api should only be accessed once for task.

I'm done with chaages as far as I'm concerned, but the problem is that some web tests seem to pretty reliably fail with an API key or connection error, indicating that for some reason something is trying to use the server rather than the mock. What's confusing though is that I can't reproduce this locally, even after turning of my internet and removing the config file with my API key: locally tests still pass for me. Any idea @yaugenst-flex ?

tidy3d/web/api/webapi.py
```diff
@@ -1646,9 +1368,9 @@ def delete(task_id: TaskId, versions: bool = False) -> TaskInfo:
         Object containing information about status, size, credits of task.
 
     """
-    if not task_id:
+    task = TaskFactory.get(task_id, verbose=False)
```
yaugenst-flex: `TaskFactory.get()` now always runs before checking empty-id, so `delete()` will always hit `/tidy3d/tasks/detail`

tidy3d/web/api/webapi.py
```diff
@@ -941,10 +739,21 @@ def get_run_info(task_id: TaskId) -> tuple[Optional[float], Optional[float]]:
         Average field intensity normalized to max value (1.0).
         Is ``None`` if run info not available.
     """
-    task = SimulationTask(taskId=task_id)
+    task = TaskFactory.get(task_id)
```
yaugenst-flex, line 714: Doing this through `TaskFactory.get` always makes an extra `/tidy3d/tasks/{task_id}/detail` which the fixtures currently don't mock

yaugenst-flex:
> So apart from addressing all review comments, I added some big updates along the lines of unifying SimulationTask and BatchTask and avoiding too many api calls to determine if something is a `batch` or not. These were introduced [in this commit](https://github.com/flexcompute/tidy3d/pull/2978/commits/24fe116f9f44a57e5d917403acb3df5076d7fe44), notably a [TaskFactory](https://github.com/flexcompute/tidy3d/pull/2978/commits/24fe116f9f44a57e5d917403acb3df5076d7fe44#diff-2f024019772f548396269e338b610c794d176828a11efc814eeed37abe1efa40R972-R986) getter that creates the appropriate SimulationTask or BatchTask, and stores a registry mapping of task id-s to whether they are batch or simulation, so the web api should only be accessed once for task.
> 
> I'm done with chaages as far as I'm concerned, but the problem is that some web tests seem to pretty reliably fail with an API key or connection error, indicating that for some reason something is trying to use the server rather than the mock. What's confusing though is that I can't reproduce this locally, even after turning of my internet and removing the config file with my API key: locally tests still pass for me. Any idea @yaugenst-flex ?

left a couple of comments which are what i suspect is breaking. as to why this passes locally for you, i suspect this has to do with how tests are executed because i think the `TaskFactory` makes things depend on how tests are scheduled. so if `TaskFactory.get()` is called before anything has registered the `TASK_ID="1234"` then this will do the http call, but if an earlier test like `test_start` runs (which do use `mock_get_info`) then later tests will skip the http.

this is bad :smile: we now have (even more) stateful tests. i think cleanest would be to itnroduce a fixture that is automatically used that clears the task factory registry

yaugenst-flex (commented):
Here you go

tidy3d/web/api/webapi.py
momchil-flex: Should I just add a `return None if task_id is None` kind of thing in `task.get`?

momchil-flex:
I could resolve most failing tests easily in the end (thanks @yaugenst-flex) but there is one that's pointing to a bit of a problem, the local cache one. The problem is that to use the local cache, the user would still need internet connection currently because the check of whether a task is a simulation or a batch task is done through a web api call. The problem existed already since rc2 due to the `_is_modeler_batch` call in `web.load`, but it is only caught by the test now because TaskFactory calls SimulationTask.get, which calls the `detail` api directly. The local cache tests mock the `webapi.get_info` function which ws previously used in the `_is_modeler_batch`.

I can of course mock the `SimulationTask.get` response too, but it seems to me that it's good that this was caught because do we want to have the local cache run fine without internet connection? @marcorudolphflex what do you think?

tidy3d/web/api/webapi.py
yaugenst-flex: something like that. but it shouldn't be in `TaskFactory.get()` (if that's what you mean?) because this is used by a bunch of other APIs too. if this just returns `None` then you have to re-add the error handling everywhere else. i would do this in `delete()`, which should just check the task id and then error if there is none. this check existed before but now it moved below the http call, so just reverting that would be simplest? or am i missing something

momchil-flex: No, you're right.

momchil-flex:
Fixed the tests (let's see the actions...), added a fixture to clear the TaskFactory registry.

I did end up monkeypatching `SimulationTask.get` to go around the cache load failure - this was already an issue in the existing code and maybe not needed to fix right away, but if we don't fix now then I'll make a ticket to follow up.

marcorudolphflex:
> I could resolve most failing tests easily in the end (thanks @yaugenst-flex) but there is one that's pointing to a bit of a problem, the local cache one. The problem is that to use the local cache, the user would still need internet connection currently because the check of whether a task is a simulation or a batch task is done through a web api call. The problem existed already since rc2 due to the `_is_modeler_batch` call in `web.load`, but it is only caught by the test now because TaskFactory calls SimulationTask.get, which calls the `detail` api directly. The local cache tests mock the `webapi.get_info` function which ws previously used in the `_is_modeler_batch`.
> 
> I can of course mock the `SimulationTask.get` response too, but it seems to me that it's good that this was caught because do we want to have the local cache run fine without internet connection? @marcorudolphflex what do you think?

@momchil-flex I agree that this should run without internet connection. Perhaps we should store the information in the cache metadata if a task is a modeler batch? Unfortunately, this would produce ugly code add-ons as we have pass around the cache entry or is_modeler_batch information.

momchil-flex:
> > I could resolve most failing tests easily in the end (thanks @yaugenst-flex) but there is one that's pointing to a bit of a problem, the local cache one. The problem is that to use the local cache, the user would still need internet connection currently because the check of whether a task is a simulation or a batch task is done through a web api call. The problem existed already since rc2 due to the `_is_modeler_batch` call in `web.load`, but it is only caught by the test now because TaskFactory calls SimulationTask.get, which calls the `detail` api directly. The local cache tests mock the `webapi.get_info` function which ws previously used in the `_is_modeler_batch`.
> > I can of course mock the `SimulationTask.get` response too, but it seems to me that it's good that this was caught because do we want to have the local cache run fine without internet connection? @marcorudolphflex what do you think?
> 
> @momchil-flex I agree that this should run without internet connection. Perhaps we should store the information in the cache metadata if a task is a modeler batch? Unfortunately, this would produce ugly code add-ons as we have pass around the cache entry or is_modeler_batch information.

This sounds related to the other discussion about storing a task type to get a default file name. It's actually 100% related because if we have the task type we can know whether it's a batch or not.

marcorudolphflex:
> This sounds related to the other discussion about storing a task type to get a default file name. It's actually 100% related because if we have the task type we can know whether it's a batch or not.

Yup. Should I handle this in a separate PR?

momchil-flex:
Yeah I think separate PR makes sense!

momchil-flex:
@daquinteroflex I added one commit to remove `compose_modeler` in favor of the new `Tidy3dBaseModel.from_file` (*another* place where we had something like that implemented locally). I also added `Tidy3dBaseModel` to the td namespace because it's awkward to use otherwise if it has to be imported from `components.base` every time. Does that sound good @yaugenst-flex ?

yaugenst-flex:
> @daquinteroflex I added one commit to remove `compose_modeler` in favor of the new `Tidy3dBaseModel.from_file` (_another_ place where we had something like that implemented locally). I also added `Tidy3dBaseModel` to the td namespace because it's awkward to use otherwise if it has to be imported from `components.base` every time. Does that sound good @yaugenst-flex ?

Well is `Tidy3dBaseModel` public or not? :smile: It's fine for now, at least until we better define that

momchil-flex:
> > @daquinteroflex I added one commit to remove `compose_modeler` in favor of the new `Tidy3dBaseModel.from_file` (_another_ place where we had something like that implemented locally). I also added `Tidy3dBaseModel` to the td namespace because it's awkward to use otherwise if it has to be imported from `components.base` every time. Does that sound good @yaugenst-flex ?
> 
> Well is `Tidy3dBaseModel` public or not? 😄 It's fine for now, at least until we better define that

I think it's convenient for users to be able to do `td.Tidy3dBaseModel.from_file()` on anything. Whether this means making `Tidy3dBaseModel` public or that we should move this somewhere else can be discussed?

yaugenst-flex:
yeah it's fine. i guess as a user facing thing it's a bit cumbersome (naming) but still nice to have

momchil-flex:
@copilot what do you think

yaugenst-flex:
@greptile review again

momchil-flex:
@daquinteroflex just pushed two more commits with further cleanup and generalization and looking ahead towards properly handling taskType again. Take a look if you want.

PR #2990, opened 2025-11-12
FXC-4111-Enable Tidy3dBaseModel.from_file() for various subclasses

momchil-flex:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR refactors polymorphic deserialization by centralizing type registry and parsing logic in `Tidy3dBaseModel`, enabling automatic type resolution when loading files.

**Key changes:**
- Moved `TYPE_TO_CLASS_MAP` from Expression plugin to base class `__init_subclass__()`
- Added `parse_obj()` override to dispatch to correct subclass based on `type` field
- Simplified stub implementations by delegating to base class methods
- Removed 100+ lines of duplicate type mapping code

**Critical issue found:**
- Line 219 in `tidy3d/components/base.py` has a logical error that breaks the intended behavior. The condition compares a `ModelField` object to a string, which will always be False, causing `parse_obj()` to never use the polymorphic dispatch logic.

**Minor issues:**
- Debug print statement left in test file

- This PR has a critical logic bug that breaks core functionality
- The condition on line 219 of `tidy3d/components/base.py` contains a logic error that compares a `ModelField` object to a string literal "Tidy3dBaseModel", which will always evaluate to False. This means the polymorphic `parse_obj()` behavior will never be triggered when calling `Tidy3dBaseModel.parse_obj()`, breaking the core feature this PR is intended to add. The fix is simple (change to `cls.__name__ == "Tidy3dBaseModel"`), but this is a critical functional bug that needs to be resolved before merging.
- tidy3d/components/base.py requires immediate attention due to the critical logic bug on line 219

File Analysis

| tidy3d/components/base.py | 2/5 | Added `TYPE_TO_CLASS_MAP` registry and `parse_obj()` override with critical bug in line 219 condition check |
| tidy3d/web/api/tidy3d_stub.py | 4/5 | Simplified stub implementation by delegating to `Tidy3dBaseModel.from_file()`. Clean refactoring with minor variable naming issue |
| tests/test_web/test_webapi.py | 3/5 | Added debug print statement that should be removed before merging |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Tidy3dBaseModel
    participant TYPE_TO_CLASS_MAP
    participant Subclass as Simulation/HeatSimulation/etc

    Note over Tidy3dBaseModel,Subclass: Registration Phase (at import time)
    Subclass->>Tidy3dBaseModel: __init_subclass__()
    Tidy3dBaseModel->>Tidy3dBaseModel: add_type_field()
    Tidy3dBaseModel->>Tidy3dBaseModel: generate_docstring()
    Tidy3dBaseModel->>TYPE_TO_CLASS_MAP: Register type -> class mapping
    Note over TYPE_TO_CLASS_MAP: {"Simulation": Simulation,<br/>"HeatSimulation": HeatSimulation, ...}

    Note over User,Subclass: Loading Phase (from_file)
    User->>Tidy3dBaseModel: from_file("sim.json")
    Tidy3dBaseModel->>Tidy3dBaseModel: dict_from_file("sim.json")
    Tidy3dBaseModel->>Tidy3dBaseModel: parse_obj({"type": "HeatSimulation", ...})
    Note over Tidy3dBaseModel: BUG: Line 219 condition always False
    Tidy3dBaseModel->>TYPE_TO_CLASS_MAP: Get class for "HeatSimulation"
    TYPE_TO_CLASS_MAP-->>Tidy3dBaseModel: HeatSimulation class
    Tidy3dBaseModel->>Subclass: HeatSimulation(**obj)
    Subclass-->>User: HeatSimulation instance
```

### DIFF
--- a/tests/test_plugins/expressions/test_dispatch.py
+++ b/tests/test_plugins/expressions/test_dispatch.py
@@ -1,0 +1,19 @@
+from __future__ import annotations
+
+import pytest
+
+from tidy3d.plugins.expressions.base import Expression
+from tidy3d.plugins.expressions.variables import Constant
+
+
+def test_expression_parse_obj_round_trip():
+    expr = Constant(3.14)
+    parsed = Expression.parse_obj(expr.dict())
+    assert isinstance(parsed, Constant)
+    assert parsed.value == pytest.approx(3.14)
+
+
+def test_expression_parse_obj_rejects_unrelated_types():
+    # Simulation registers a distinct type in the global map; parsing via Expression should fail.
+    with pytest.raises(ValueError, match="Cannot parse type"):
+        Expression.parse_obj({"type": "Simulation"})
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -42,6 +42,7 @@ JSON_TAG = "JSON_STRING"
 MAX_STRING_LENGTH = 1_000_000_000
 FORBID_SPECIAL_CHARACTERS = ["/"]
 TRACED_FIELD_KEYS_ATTR = "__tidy3d_traced_field_keys__"
+TYPE_TO_CLASS_MAP: dict[str, type[Tidy3dBaseModel]] = {}
 
 
 def cache(prop):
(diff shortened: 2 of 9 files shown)


yaugenst-flex:
@marcorudolphflex could you have a look at the failing lazy loading test :pray:

momchil-flex:
> @marcorudolphflex could you have a look at the failing lazy loading test 🙏

I have this print statement about the types currently. In both develop and this branch I get
`[<class 'tidy3d.components.base._make_lazy_proxy.<locals>._LazyProxy'>, <class 'dict'>, <class 'tuple'>]`
but somehow on this branch the lazy proxy is not manifested as SimulationData in the `isinstance` check.

momchil-flex:
I guess because of [this change](https://github.com/flexcompute/tidy3d/pull/2990/files#diff-a1c5b495b9c2be138a1e9be6c70d83ed70de00d55cb1d1ca4f06d46c1ba27010R174) the data[0].__class__.__name__  is now `'Tidy3dBaseModelProxy'`. Will wait to hear from Marco before digging through more.

marcorudolphflex:
If I catch that correctly, the fundamental issue is that it tries to create a `Tidy3DBaseModel` instead of the "correct" simulation class.
This fails for me without lazy loading:
```
from tests.test_web.test_tidy3d_stub import make_sim_data
from tidy3d.web.api.tidy3d_stub import Tidy3dStubData

tmp_path = "test.hdf5"
sim_data = make_sim_data()
sim_data.to_file(tmp_path)
Tidy3dStubData.postprocess(tmp_path, lazy=False)
```

```
pydantic.v1.error_wrappers.ValidationError: 5 validation errors for Tidy3dBaseModel
data
  extra fields not permitted (type=value_error.extra)
diverged
  extra fields not permitted (type=value_error.extra)
log
  extra fields not permitted (type=value_error.extra)
simulation
  extra fields not permitted (type=value_error.extra)
type
  extra fields not permitted (type=value_error.extra)
```

momchil-flex:
Sorry @marcorudolphflex there was a bug in that state you tried. Now the test failures are exclusively due to lazy loading.

marcorudolphflex:
> Sorry @marcorudolphflex there was a bug in that state you tried. Now the test failures are exclusively due to lazy loading.

No problem.

So the issue with lazy loading is the following: In the current state of `Tidy3dStubData.from_file`, we cannot detect the individual class when creating the lazy object as we do not look into the file when working lazy. Therefore, we simply cannot inherit from the original class which let `isinstance` checks fail. The type name itself should not be really critical, that is more a gimmick when you print the type - but it is failing for the same reason.

Now the question is if we want to support `isinstance` on the simulation data type which I would generally tend to and it looks like we do some checks like that currently.

So I see 2 alternative solutions:
- check the json for the type as before - at least in the lazy loading case
- materialize on isinstance checks (delete `__class__` from list of non-materialize-attributes)

What do you think?

marcorudolphflex:
Edit:

> materialize on isinstance checks (delete __class__ from list of non-materialize-attributes)

this option just does not work in general. The object would still be a Tidy3DBaseModel after materialization such that it would be not usable. I would prefer the other option which seems more stable.

momchil-flex:
Yes if the first option works it sounds better, are you going to give it a try?

marcorudolphflex:
I can try to do that, but actually it would be better to do that on the pydantic v2 branch as we already observed varied behavior there.
Maybe we should postpone that if it does not work out of the box? Will post an update tomorrow.

momchil-flex:
Thanks, sounds good, this is not exactly urgent it's just a nice cleanup.

tidy3d/components/base.py
```diff
@@ -195,6 +196,76 @@ def __init_subclass__(cls) -> None:
...
+        """Return the type tag from a raw dictionary."""
+        if not isinstance(obj, dict):
+            raise TypeError("Input must be a dict")
+        try:
+            type_value = obj[TYPE_TAG_STR]
+        except KeyError as exc:
+            raise ValueError(f'Missing "{TYPE_TAG_STR}" in data') from exc
+        if not isinstance(type_value, str) or not type_value:
+            raise ValueError(f'Invalid "{TYPE_TAG_STR}" value: {type_value!r}')
+        return type_value
```
momchil-flex, line 214: Do we really need all these protections LLM's like to put...?

yaugenst-flex: haha i guess this is more of a fundamental question on error handling and how afraid we are of more verbose code :wink:

momchil-flex:
Cleaned up and rebased, @yaugenst-flex you wanna approve?

tests/test_plugins/expressions/test_dispatch.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import annotations
+
+import pytest
+
+from tidy3d.components.simulation import Simulation  # noqa: F401  # ensure type registration
```
yaugenst-flex: Pretty sure this import isn't necessary.

yaugenst-flex (approved):
LGTM, minor test nit

PR #3001, opened 2025-11-13
feat(medium): track parent medium from which the medium is derived

weiliangjin2021:
Not elegant, especially the quote in the type.

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

Added a `super_medium` field to `AbstractMedium` that tracks the parent medium from which a derived medium was created. This is set in the `perturbed_copy()` methods of `PerturbationMedium` and `PerturbationPoleResidue`, allowing users to trace back to the original perturbation medium.

**Key Changes:**
- Added `super_medium: Optional["MediumType"]` field to `AbstractMedium` with forward reference
- Updated `AbstractMedium.update_forward_refs()` call (though it currently resolves to `Any` instead of the full `MediumType` union)
- Modified `perturbed_copy()` in both `PerturbationMedium` and `PerturbationPoleResidue` to set `super_medium = self` before returning derived mediums
- Added comprehensive test coverage for the new field

**Issues Found:**
- Forward reference resolution uses `Any` instead of the actual `MediumType`, which reduces type safety
- Missing CHANGELOG entry for this user-facing feature

- This PR is generally safe to merge with one logical issue that should be addressed
- The implementation correctly adds the `super_medium` field and sets it appropriately in all code paths. The test coverage is excellent and validates the expected behavior. However, the forward reference resolution at line 1166 uses `Any` instead of the full `MediumType` union, which undermines type safety for the new field. Additionally, a CHANGELOG entry is missing per project guidelines.
- Pay close attention to `tidy3d/components/medium.py` line 1166 for the forward reference resolution issue

File Analysis

| tidy3d/components/medium.py | 4/5 | Added `super_medium` field to track parent medium and updated forward reference resolution; needs CHANGELOG entry |
| tests/test_components/test_perturbation_medium.py | 5/5 | Comprehensive test coverage for `super_medium` field on both `PerturbationMedium` and `PerturbationPoleResidue` |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant PerturbationMedium
    participant Medium
    participant CustomMedium
    
    User->>PerturbationMedium: perturbed_copy()
    alt No perturbation data
        PerturbationMedium->>PerturbationMedium: Create new_dict excluding perturbation fields
        PerturbationMedium->>PerturbationMedium: Set super_medium = self
        PerturbationMedium->>Medium: parse_obj(new_dict)
        Medium-->>User: Return Medium with super_medium set
    else With perturbation data
        PerturbationMedium->>PerturbationMedium: Apply perturbations to fields
        PerturbationMedium->>PerturbationMedium: Set super_medium = self
        PerturbationMedium->>CustomMedium: parse_obj(new_dict)
        CustomMedium-->>User: Return CustomMedium with super_medium set
    end
```

**Context used:**

- Rule from `dashboard` - Require a changelog entry for any PR that is not purely an internal refactor. ([source](https://app.greptile.com/review/custom-context?memory=b72c7231-b258-4d03-aed2-51a50a9fe757))

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -91,6 +91,7 @@ from .types import (
     PermittivityComponent,
     PoleAndResidue,
     TensorReal,
+    annotate_type,
 )
 from .validators import _warn_potential_error, validate_name_str, validate_parameter_perturbation
 from .viz import VisualizationSpec, add_ax_if_none
@@ -873,6 +874,12 @@ class AbstractCustomMedium(AbstractMedium, ABC):
         "``Simulation``'s field ``subpixel`` for this type of material on the "
         "interface of the structure, including exterior boundary and "
         "intersection interfaces with other structures.",
+    )
+
+    derived_from: Optional[annotate_type(PerturbationMediumType)] = pd.Field(
+        None,
+        title="Parent Medium",
+        description="If not ``None``, it records the parent medium from which this medium was derived.",
     )
 
     @cached_property
@@ -6748,7 +6755,7 @@ class PerturbationMedium(Medium, AbstractPerturbationMedium):
         electron_density: CustomSpatialDataType = None,
         hole_density: CustomSpatialDataType = None,
         interp_method: InterpMethod = "linear",
-    ) -> Union[Medium, CustomMedium]:
+    ) -> Union[PerturbationMedium, CustomMedium]:
         """Sample perturbations on provided heat and/or charge data and return 'CustomMedium'.
         Any of temperature, electron_density, and hole_density can be 'None'. If all passed
         arguments are 'None' then a 'Medium' object is returned. All provided fields must have
@@ -6780,9 +6787,13 @@ class PerturbationMedium(Medium, AbstractPerturbationMedium):
 
         Returns
         -------
-        Union[Medium, CustomMedium]
+        Union[PerturbationMedium, CustomMedium]
             Medium specification after application of heat and/or charge data.
         """
+
+        # in the absence of perturbation
+        if all(x is None for x in [temperature, electron_density, hole_density]):
+            return self
 
         new_dict = self.dict(
             exclude={
@@ -6792,10 +6803,6 @@ class PerturbationMedium(Medium, AbstractPerturbationMedium):
                 "type",
             }
         )
-
-        if all(x is None for x in [temperature, electron_density, hole_density]):
-            new_dict.pop("subpixel")
-            return Medium.parse_obj(new_dict)
 
         permittivity_field = self.permittivity + ParameterPerturbation._zeros_like(
             temperature, electron_density, hole_density
@@ -6836,6 +6843,7 @@ class PerturbationMedium(Medium, AbstractPerturbationMedium):
         new_dict["permittivity"] = permittivity_field
         new_dict["conductivity"] = conductivity_field
         new_dict["interp_method"] = interp_method
+        new_dict["derived_from"] = self
 
         return CustomMedium.parse_obj(new_dict)
 
@@ -6959,7 +6967,7 @@ class PerturbationPoleResidue(PoleResidue, AbstractPerturbationMedium):
         electron_density: CustomSpatialDataType = None,
         hole_density: CustomSpatialDataType = None,
         interp_method: InterpMethod = "linear",
-    ) -> Union[PoleResidue, CustomPoleResidue]:
+    ) -> Union[PerturbationPoleResidue, CustomPoleResidue]:
         """Sample perturbations on provided heat and/or charge data and return 'CustomPoleResidue'.
         Any of temperature, electron_density, and hole_density can be 'None'. If all passed
         arguments are 'None' then a 'PoleResidue' object is returned. All provided fields must have
@@ -6991,17 +6999,17 @@ class PerturbationPoleResidue(PoleResidue, AbstractPerturbationMedium):
 
         Returns
         -------
-        Union[PoleResidue, CustomPoleResidue]
+        Union[PerturbationPoleResidue, CustomPoleResidue]
             Medium specification after application of heat and/or charge data.
         """
+
+        # in the absence of perturbation
+        if all(x is None for x in [temperature, electron_density, hole_density]):
+            return self
 
         new_dict = self.dict(
             exclude={"eps_inf_perturbation", "poles_perturbation", "perturbation_spec", "type"}
         )
-
-        if all(x is None for x in [temperature, electron_density, hole_density]):
-            new_dict.pop("subpixel")
-            return PoleResidue.parse_obj(new_dict)
 
         zeros = ParameterPerturbation._zeros_like(temperature, electron_density, hole_density)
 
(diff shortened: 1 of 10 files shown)


tidy3d/components/medium.py
```diff
@@ -197,6 +197,12 @@ class AbstractMedium(ABC, Tidy3dBaseModel):
         description="Plotting specification for visualizing medium.",
     )
 
+    super_medium: Optional["MediumType"] = pd.Field(  # noqa: UP037
+        None,
+        title="Super Medium",
+        description="If not ``None``, it records the super medium from which this medium was derived.",
+    )
```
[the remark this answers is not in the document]
weiliangjin2021: only internal

weiliangjin2021:
Adding the class itself as its field looks messy with pydantic. Probably an `identifier` field is a cleaner implementation, unless @momchil-flex you have better idea to improve the current PR.

dbochkov-flexcompute:
Would it help if we add the `super_medium` (or `parent_medium`) only to `AbstractCustomMedium`? I don't think there is a case when we end up in the same perturbed simulation with two different mediums, one `Medium` and one `CustomMedium`, derived from the same `PerturbationMedium`. They should either be all `Medium`'s (if no perturbation fields are passed) or all of them are `CustomMedium`'s. This should also allow us to easily add `CustomMedium.perturbed_copy` for resampling on a different set perturbation fields

weiliangjin2021:
> Would it help if we add the `super_medium` (or `parent_medium`) only to `AbstractCustomMedium`? I don't think there is a case when we end up in the same perturbed simulation with two different mediums, one `Medium` and one `CustomMedium`, derived from the same `PerturbationMedium`. They should either be all `Medium`'s (if no perturbation fields are passed) or all of them are `CustomMedium`'s. This should also allow us to easily add `CustomMedium.perturbed_copy` for resampling on a different set perturbation fields

That's right. On the other hand, we'll also want to keep track when there is no perturbation, so that the two generated materials are still considered the same. Then with perturbation applied, the subpixel should be applied in the same way to avoid abrupt change.

dbochkov-flexcompute:
> > Would it help if we add the `super_medium` (or `parent_medium`) only to `AbstractCustomMedium`? I don't think there is a case when we end up in the same perturbed simulation with two different mediums, one `Medium` and one `CustomMedium`, derived from the same `PerturbationMedium`. They should either be all `Medium`'s (if no perturbation fields are passed) or all of them are `CustomMedium`'s. This should also allow us to easily add `CustomMedium.perturbed_copy` for resampling on a different set perturbation fields
> 
> That's right. On the other hand, we'll also want to keep track when there is no perturbation, so that the two generated materials are still considered the same. Then with perturbation applied, the subpixel should be applied in the same way to avoid abrupt change.

actually, if only `AbstractCustomMedium` carries `parent_medium`, then if you do `psim = sim.perturbation_mediums_copy(temperature=None, electron_density=None, hole_density=None)`, then there would be no way to apply perturbations to `psim`. But maybe we should change `AbstractPerturbationMedium.perturbed_copy()` to return itself if no perturbation are provided. This should resolve both issues.

weiliangjin2021:
> actually, if only AbstractCustomMedium carries parent_medium, then if you do psim = sim.perturbation_mediums_copy(temperature=None, electron_density=None, hole_density=None), then there would be no way to apply perturbations to psim. But maybe we should change AbstractPerturbationMedium.perturbed_copy() to return itself if no perturbation are provided. This should resolve both issues.

That's a good idea!

tidy3d/components/medium.py
```diff
@@ -7050,12 +7058,28 @@ def perturbed_copy(
...
+    """Recursively get all subclasses of a class."""
+    all_subclasses = []
+    for subclass in cls.__subclasses__():
+        all_subclasses.append(subclass)
+        all_subclasses.extend(_get_all_subclasses(subclass))
+    return all_subclasses
+
+
+for _custom_medium_cls in _get_all_subclasses(AbstractCustomMedium):
+    _custom_medium_cls.update_forward_refs()
```
weiliangjin2021, line 7082: Still need to update forward refs for all custom materials

tidy3d/components/medium.py
```diff
@@ -875,6 +876,12 @@ class AbstractCustomMedium(AbstractMedium, ABC):
         "intersection interfaces with other structures.",
     )
 
+    super_medium: Optional[annotate_type(PerturbationMediumType)] = pd.Field(
```
momchil-flex: This sounds a bit weird to me as a name for the field. How about `derived_from` or something like that?

momchil-flex (approved):

tidy3d/components/medium.py
weiliangjin2021: changed to `derived_from` now

tidy3d/components/simulation.py
```diff
@@ -2216,15 +2216,55 @@ def _modal_plane_frames(self) -> list[Structure]:
     @cached_property
     def _finalized(self) -> Simulation:
         """Return the finalized version of the simulation setup. That is, including automatic frames around mode sources and internal absorbers, and 2d strutures converted into volumetric analogues."""
+        if (
+            len(self._modal_plane_frames) == 0
+            and not self._contains_converted_volumetric_structures
+        ):
+            return self
+        return self.updated_copy(
+            grid_spec=GridSpec.from_grid(self.grid),
```
momchil-flex, line 2225: This is not really related to this PR but I just noticed that here too we are losing the information about the original grid_spec, like we've been discussing for the component modeler speed up updates. Where is `_finalized` used? Can it lead to user confusion?

tidy3d/components/simulation.py
```diff
@@ -2216,15 +2216,55 @@ def _modal_plane_frames(self) -> list[Structure]:
...
+            structure._optical_medium for structure in self._finalized_volumetric_structures
+        }
+        medium_set.add(Structure._get_optical_medium(self.medium))
+        return {medium: index for index, medium in enumerate(medium_set)}
 
-        return self.updated_copy(grid_spec=GridSpec.from_grid(self.grid), structures=structures)
+    @cached_property
+    def _finalized_full_optical_medium_map(self) -> dict[MediumType, pydantic.NonNegativeInt]:
+        """Returns dict mapping medium to index in material in finalized simulation, including
+        parent medium.
```
momchil-flex: Maybe just for some extra clarity instead of "including parent medium" -> "including media from which custom media have been derived."

momchil-flex (approved):

tidy3d/components/simulation.py
weiliangjin2021: so far I think it's only used in backend @dbochkov-flexcompute

dbochkov-flexcompute: correct, and for understanding on the frontend what exactly you end up simulating

dbochkov-flexcompute: That is, the returned `SimulationData` will contain the original simulation, not sim._finalized

dbochkov-flexcompute (approved):

PR #3013, opened 2025-11-19
Hotfix/fix scene doping limits

marc-flex:
<h3>Greptile Summary</h3>

- Fixed `doping_bounds()` to return `[0, 0]` instead of sentinel values `[1e50, -1e50]` when no doping structures are present
- This prevents invalid visualization limits when scenes lack semiconductor doping configurations

- This PR fixes a visualization bug but uses direct float equality which could fail in edge cases
- The fix correctly handles the empty doping case but violates the floating-point comparison best practice (custom rule 4a0cb6d3). Direct equality `== 1e50` should use `np.isclose()` for robustness. The logic is sound but the implementation needs a minor correction.
- Pay attention to `tidy3d/components/scene.py` - the sentinel value checks need tolerance-based comparison

| Filename | Overview |
|----------|----------|
| tidy3d/components/scene.py | Added sentinel value checks to handle empty doping case, but uses direct float equality instead of tolerance-based comparison |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Scene
    participant doping_bounds
    participant Structures

    User->>Scene: "Visualize doping properties (N_a, N_d, or doping)"
    Scene->>doping_bounds: "Calculate doping bounds"
    doping_bounds->>Structures: "Loop through all_structures"
    alt Structures with SemiconductorMedium found
        Structures-->>doping_bounds: "Update acceptors_lims and donors_lims"
    else No structures with doping
        Structures-->>doping_bounds: "Limits remain at sentinel values [1e50, -1e50]"
    end
    doping_bounds->>doping_bounds: "Check if limits are still at sentinel values"
    alt Limits are sentinel values
        doping_bounds->>doping_bounds: "Set limits to [0, 0]"
    end
    doping_bounds-->>Scene: "Return acceptors_lims, donors_lims"
    Scene-->>User: "Use limits for visualization"
```

### DIFF
--- a/tidy3d/components/scene.py
+++ b/tidy3d/components/scene.py
@@ -1979,8 +1979,8 @@ class Scene(Tidy3dBaseModel):
     def doping_bounds(self):
         """Get the maximum and minimum of the doping"""
 
-        acceptors_lims = [1e50, -1e50]
-        donors_lims = [1e50, -1e50]
+        acceptors_lims = [np.inf, -np.inf]
+        donors_lims = [np.inf, -np.inf]
 
         for struct in self.all_structures:
             if isinstance(struct.medium.charge, SemiconductorMedium):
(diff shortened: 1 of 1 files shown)


tidy3d/components/scene.py
```diff
@@ -2020,6 +2020,15 @@ def doping_bounds(self):
...
                                     limits[1] = max_value
+        # make sure we have recorded some values. Otherwise, set to 0
+        if np.isclose(acceptors_lims[0], np.inf):
+            acceptors_lims[0] = 0
+        if np.isclose(acceptors_lims[1], -np.inf):
+            acceptors_lims[1] = 0
+        if np.isclose(donors_lims[0], np.inf):
+            donors_lims[0] = 0
+        if np.isclose(donors_lims[1], -np.inf):
+            donors_lims[1] = 0
```
yaugenst-flex: Nit, but I'd suggest using [np.isinf](https://numpy.org/doc/stable/reference/generated/numpy.isinf.html)

marc-flex: Thanks!

momchil-flex (approved):

PR #3030, opened 2025-11-25
fix(mode): fixed bug where mode solver was not respecting location of PEC

dmarek-flex:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR fixes a bug where the mode solver grid would extend beyond simulation domain boundaries, causing incorrect PEC boundary condition application. The fix implements grid truncation at simulation boundaries and zero-pads fields outside the domain.

Key changes:
- Added `_sim_boundary_positions` method to detect PEC boundaries and determine where to truncate the computational grid
- Implemented truncation logic in `compute_modes` that clips coordinates, permittivity, permeability, and basis fields to simulation bounds
- Added zero-padding to restore fields to original grid size after solving on truncated domain
- Added warnings for unsupported boundary types (PMC, Periodic, Bloch) when mode plane intersects them
- Added comprehensive tests verifying zero fields outside simulation and PEC boundary conditions

The implementation correctly handles the core issue, but there's a minor inconsistency in the boundary intersection detection logic that could cause edge cases.

- This PR is generally safe to merge with one logical inconsistency that should be addressed
- The implementation is well-designed with comprehensive tests, proper documentation, and correct changelog entry. However, there's a logical inconsistency in the boundary intersection detection (using `np.isclose` vs `fp_eps` threshold) that could cause edge case failures. The core truncation and zero-padding logic appears sound
- Pay close attention to `tidy3d/components/mode/mode_solver.py` for the boundary intersection logic inconsistency

File Analysis

| CHANGELOG.md | 5/5 | Added changelog entry describing the bug fix for mode solver grid extending beyond simulation bounds |
| tests/test_plugins/test_mode_solver.py | 4/5 | Added comprehensive tests for PEC boundary truncation and zero-padding, including warning tests for unsupported boundary types. Tests verify field values outside simulation bounds and at PEC boundaries |
| tidy3d/components/mode/mode_solver.py | 4/5 | Added `_sim_boundary_positions` method to detect PEC boundaries and determine truncation positions. Passes boundary positions to solver for grid truncation. Minor inconsistency in boundary checking logic |
| tidy3d/components/mode/solver.py | 4/5 | Implements core truncation logic: truncates grid/eps/mu/fields to simulation bounds when needed, then zero-pads results. Adds `_truncate_medium_data` helper method |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant MS as ModeSolver
    participant MSC as ModeSolver._solve_single_freq
    participant SBP as ModeSolver._sim_boundary_positions
    participant CM as compute_modes (solver.py)
    participant TMD as _truncate_medium_data
    
    MS->>MS: User calls .data property
    MS->>MSC: _solve_single_freq(freq, coords, symmetry)
    MSC->>SBP: Get PEC boundary positions
    SBP->>SBP: Check boundary types (PEC, PMC, etc.)
    SBP->>SBP: Determine if solver grid intersects boundaries
    SBP-->>MSC: Return (pos_min, pos_max) tuples
    
    MSC->>CM: compute_modes(eps_cross, coords, ..., sim_pec_pos_min, sim_pec_pos_max)
    
    CM->>CM: Calculate truncation indices from PEC positions
    CM->>CM: Check if truncation needed (trim_min != [0,0] or trim_max != [Nx,Ny])
    
    alt Truncation needed
        CM->>CM: Truncate coords arrays
        CM->>TMD: Truncate eps_cross
        TMD-->>CM: Truncated eps
        CM->>TMD: Truncate mu_cross (if present)
        TMD-->>CM: Truncated mu
        CM->>CM: Truncate split_curl_scaling (if present)
        CM->>CM: Truncate solver_basis_fields (if present)
    end
    
    CM->>CM: Solve eigenvalue problem on truncated grid
    CM->>CM: Reshape fields to (2, 3, Nx_trunc, Ny_trunc, 1, num_modes)
    
    alt Truncation was applied
        CM->>CM: Zero-pad fields back to original size
        Note over CM: np.pad with trim_min/trim_max offsets
    end
    
    CM-->>MSC: Return (n_complex, fields, eps_spec)
    MSC-->>MS: Return mode data with correct boundary conditions
```

---

> [!NOTE]
> **Medium Risk**
> Touches core mode-solver boundary handling and field interpolation/truncation logic; mistakes could subtly affect mode fields/impedances near boundaries despite added coverage via new regression tests.
> 
> **Overview**
> Fixes a long-standing mismatch where `ModeSolver` could extend its solver grid beyond the simulation domain, causing PEC boundary conditions to be enforced at the wrong locations for interior ports.
> 
> `ModeSolver` now computes PEC positions by snapping the mode plane to the simulation grid, passes these bounds into `compute_modes` to *truncate* coordinates/material data (eps/mu/basis fields) during the eigensolve, and then zero-pads the solved fields back to the original shape. Colocation now slices fields to the PEC-bounded region and interpolates with `fill_value="extrapolate"`.
> 
> Adds warnings when the mode plane intersects simulation boundaries that are not supported for the mode solver’s PEC enforcement (PMC/Periodic/Bloch), updates `ModePlaneAnalyzer` to filter conductors using the same PEC-snapped bounds (fixing ground-plane shorting behavior), introduces the `Bound2D` type, and expands tests to cover truncation/zero-padding, boundary warnings, interior-plane snapping, and degenerate-axis (2D) cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b5eb390064c9b6694caf801c14121dbdc07016ad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


dmarek-flex:
@greptileai another round

tidy3d/components/mode/mode_solver.py
```diff
@@ -1171,6 +1181,99 @@ def _reduced_simulation_copy_with_fallback(self) -> ModeSolver:
...
+                    )
+            elif isinstance(bc_plus, (Periodic, BlochBoundary)):
+                if intersects_max:
+                    log.warning(
+                        f"Mode solver plane intersects simulation '{axis_name}+' boundary with "
+                        f"unsupported periodic/Bloch boundary condition. "
+                        f"Fields will not wrap around periodically."
+                    )
+            # ABC boundaries: no truncation, no warning (fields can extend)
+        return (tuple(pos_min), tuple(pos_max))
```
momchil-flex, line 1315: So this fix seems to be specific to the case where the enclosing `Simulation` has e.g. PEC boundaries.

I am wondering if this still excludes problematic cases where the enclosing Simulation is large, but the mode plane itself does not impose the PEC boundaries on the analytical plane boundaries, but instead on whatever the discretized version of the plane turns out to be. You guys might have dealt with something like that already, not sure.

If memory serves me right, at least in the new `ModeSimulation` class we've made a small step towards handling this correctly in the following sense:
1. If the `ModeSimulation` is defined with a 3D `center`, `size`, as well as a `plane` argument, then the inaccuracy is technically there. The plane will be discretized based on the larger 3D "simulation" grid, and the discretization and hence where boundaries are applied may not match the analytical plane definition.
2. If however the `ModeSimulation` is defined with a 2D `size` and no extra `plane` argument, then the grid matches the plane, and the boundaries match the expected location.

So I guess what I'm wondering is if your approach to pass the `sim_pec_pos_max`, etc. should also be applied in situations like case 1. above. Although I'm not sure what should happen if the analytical plane position just lies somewhere in between the mode plane grid points.. Here your fix works because the boundary location does match a grid point, there just might be an extra pixel, right?

dmarek-flex: > I am wondering if this still excludes problematic cases where the enclosing Simulation is large, but the mode plane itself does not impose the PEC boundaries on the analytical plane boundaries, but instead on whatever the discretized version of the plane turns out to be. You guys might have dealt with something like that already, not sure.

Yes, so far I was trying to match existing behavior, but fix the bug. So if the mode plane is smaller than the `Simulation` plane bounds, there is no change in behavior when compared to  `develop`. It is not clear to me what to do if the mode plane boundaries don't line up with grid boundaries. I think changing the grid is never a good idea, especially since these modes might be injected into an FDTD simulation. And if we choose the closest grid boundary, or say the closest grid boundary that encompasses the mode plane, then basically we have the same problem (adding additional space), just potentially not quite as bad as the one complete cell in the bug. 

> Here your fix works because the boundary location does match a grid point, there just might be an extra pixel, right?

Correct, `sim_pec_pos_max` might as well be an index into the grid boundaries. I only want to support the PEC boundary matching an existing grid boundary, because otherwise the mode solver grid would have to change slightly to respect a new location of the PEC boundary.

tidy3d/components/mode/mode_solver.py
```diff
@@ -1171,6 +1181,99 @@ def _reduced_simulation_copy_with_fallback(self) -> ModeSolver:
...
+    def _sim_boundary_positions(
+        self,
+    ) -> tuple[tuple[float, float], tuple[float, float]]:
+        """Get simulation boundary positions for the mode plane's tangential axes.
+
+        This method checks the simulation's boundary conditions and returns positions
+        only for sides that have supported boundaries. For unsupported boundary types,
+        a warning is logged only if the mode solver plane intersects that boundary.
+
+        The mode solver supports:
```
caseyflex: The PML in mode solver is handled separately from the FDTD PML. I think previously the PEC was also handled separately in the mode solver vs. the FDTD boundary condition.

It would be nice to unify things, as long as it works in all use cases. Can PML handling be unified in the same way?

I'm not sure about all the new warnings -- I guess they are because those boundary conditions are not supported in the mode solver. But I'm also a bit worried about having the mode solver behave so qualitatively differently depending on the underlying FDTD simulation (PEC vs high-conductivity medium, for example).

dmarek-flex: > I think previously the PEC was also handled separately in the mode solver vs. the FDTD boundary condition.

Yea, that is what was causing issues. In RF simulations you might use a PEC boundary condition to take the place of a ground plane, but the exact location of the PEC boundary in the FDTD simulation and the ModeSolver solve did not line up. So the ModeSolver was calculating modes with effectively one extra cell before the PEC.

dmarek-flex: I guess the big picture question is what should we do if the ModeSolver plane bounds intersects with the Simulation bounds:

1. Override the BCs in the ModeSolver with the Simulation's BCs 
2. Keep PECs or whatever BCs are set in the ModeSolver

I think the PML handling could also be unified, it seems a little harder since we probably should use all of PML layers from the Tidy3D simulation, instead of the subset of layers that might exist in an intersection of the mode plane and the simulation bounds.

Yes there are a lot of new warnings, but I would say that the current behavior of ignoring the Simulation BCs is ripe for confusion!

weiliangjin2021:
> Yes, so far I was trying to match existing behavior, but fix the bug. So if the mode plane is smaller than the Simulation plane bounds, there is no change in behavior when compared to develop. And if we choose the closest grid boundary, or say the closest grid boundary that encompasses the mode plane, then basically we have the same problem (adding additional space), just potentially not quite as bad as the one complete cell in the bug.

I think it's quite common to have mode plane smaller than the simulation domain, and it'll be very useful to fix that too?

Regarding closest grid boundary, i guess it should work in most cases, as it's expected to have the mode plane boundary to lie at the metal surface, and metal surface is snapped to grid.

tidy3d/components/mode/mode_solver.py
caseyflex: Can you try running some simple EME simulations and make sure there aren’t excessive warnings?

dmarek-flex:
> I think it's quite common to have mode plane smaller than the simulation domain, and it'll be very useful to fix that too?

It's not really an issue though if the mode plane is smaller than the simulation domain. All that will happen is an extra cell will be added that is outside the monitor bounds. But since in the simulation there is no PEC boundary at the monitor bounds anyways, it is an accurate representation. A structure just outside/touching the monitor bounds will still be captured by the mode solver, and is the current workaround for this bug.

caseyflex (approved):
I think the warnings make sense if the boundary won’t be used.

the only unclear case is PML, which you don’t earn for just because it would be annoying. I guess that can be a separate discussion about unifying PML treatment

tidy3d/components/mode/mode_solver.py
dmarek-flex: > 2\. If however the `ModeSimulation` is defined with a 2D `size` and no extra `plane` argument, then the grid matches the plane, and the boundaries match the expected location.

I just tried this case, and it has the same issue without this PR's fix. The mode solver grid created by `ModeSimulation` will have grid points outside the `ModeSimulation` bounds

dmarek-flex: The goal of this PR fix is not necessarily to impose PEC boundaries exactly on `ModeSolver` plane bounds, but to make sure that when the `ModeSolver` is used in conjunction with a tidy3d `Simulation` that they both agree on the exact location of the PEC boundary.

dmarek-flex: > So I guess what I'm wondering is if your approach to pass the sim_pec_pos_max, etc. should also be applied in situations like case 1. above. Although I'm not sure what should happen if the analytical plane position just lies somewhere in between the mode plane grid points.

Yea, it is a bit awkward, but in the absence of Simulation boundary conditions, I think it is fine to have the `ModeSolver` BCs placed where the `ModeSolver` subgrid ends (which happens to be outside the `ModeSolver` plane). The problem is that the `ModeSolver` does not have control of the grid boundary locations, so it is hard to enforce a boundary condition exactly on the `ModeSolver` plane bounds.

tidy3d/components/mode/solver.py
```diff
@@ -111,6 +119,66 @@ def compute_modes(
...
+        original_Ny = len(coords[1]) - 1
+
+        # Find indices where coords fall within simulation PEC boundaries
+        trim_min = [0, 0]
+        trim_max = [len(coords[0]), len(coords[1])]
+
+        if sim_pec_pos_min is not None:
+            # Find first coord index >= sim_pec_pos_min (for boundary coords)
+            for i in range(2):
+                idx = np.searchsorted(coords[i], sim_pec_pos_min[i])
```
weiliangjin2021: would be good to add `side="left"`, even though it's the default

weiliangjin2021 (approved):

tidy3d/components/mode/mode_solver.py
momchil-flex: > I just tried this case, and it has the same issue without this PR's fix. The mode solver grid created by ModeSimulation will have grid points outside the ModeSimulation bounds

I see. Yeah fine by me to proceed with this fix for now. It is pretty hard to come up with a bigger refactor that fixes some of the known issues without also (potentially) breaking some of the current functionality / results.

tidy3d/components/mode/solver.py
```diff
@@ -93,6 +95,12 @@ def compute_modes(
             The center of the mode plane along the tangential axes of the global simulation. Used
             in case of bend modes to offset the coordinates correctly w.r.t. the bend radius, which
             is assumed to refer to the distance from the bend center to the mode plane center.
+        sim_pec_pos_min : Optional[tuple[float, float]]
+            Simulation PEC boundary positions (x_min, y_min). If the solver grid extends beyond a PEC
+            position, it will be truncated and fields outside will be zero-padded.
+        sim_pec_pos_max : Optional[tuple[float, float]]
+            Simulation PEC boundary positions (x_max, y_max). If the solver grid extends beyond a PEC
+            position, it will be truncated and fields outside will be zero-padded.
```
momchil-flex: Why have two separate arguments for `min/max` but then not have them separate for x vs. y boundary?

Maybe just a single argument for all 4 sides?

momchil-flex (approved):

dmarek-flex:
Superseded by https://github.com/flexcompute/compute/pull/2955

PR #3046, opened 2025-12-01
ci: tidy3d-extras integration tests and docs/dev setup (FXC-4441)

daquinteroflex:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR introduces comprehensive `tidy3d-extras` integration testing and Docker-based development environment setup. The changes add a new CI workflow for testing the optional extras package across multiple platforms and Python versions, with two test modes: basic (4 configurations for fast smoke testing on PRs) and full (10 configurations covering all architectures for releases). The PR also includes a complete Docker development environment with detailed documentation.

**Major Changes:**
- New `tidy3d-extras-python-client-tests-integration.yml` workflow with dual test modes (basic/full) covering Windows, Linux x86_64/aarch64, and macOS x86_64/arm64 on Python 3.10 and 3.13
- Integration of extras tests into main test workflow with automatic enablement on merge queue events
- Updated release workflow to run full extras integration tests by default
- Daily workflow now runs a draft release (`daily-0.0.0`) instead of just submodule tests, providing broader CI validation
- New `dev.Dockerfile` providing isolated development environment with Poetry, AWS CLI, and development tools
- Comprehensive Docker development guide (`docs/development/docker.rst`) with setup instructions, troubleshooting, and AWS CodeArtifact authentication
- New test file `_test_tidy3d_extras_license.py` to verify license validation and error handling
- Main test workflows now properly exclude `_test_tidy3d_extras_license.py` to prevent redundant execution (this test is run separately in the extras integration workflow)
- Poetry lockfile updated to 2.1.1 with minor dependency version bumps

**Architecture:**
The extras integration tests are invoked through the main test workflow, which determines the test type based on the event trigger. PRs and merge queue events use "basic" mode for fast feedback, while the release workflow uses "full" mode for comprehensive validation before deployment.

<h3>Confidence Score: 5/5</h3>

- This PR is safe to merge with minimal risk - it adds well-structured CI infrastructure and development tooling without modifying core application logic.
- The changes are primarily additive (new workflows, documentation, and development tools) with minimal modifications to existing code. The workflow integration follows established patterns in the repository, properly handles test exclusions to avoid redundancy, and includes comprehensive documentation. The Docker setup and extras integration tests improve developer experience and CI coverage without introducing breaking changes.
- No files require special attention - all changes follow best practices and are well-documented.

<h3>Important Files Changed</h3>

File Analysis

| Filename | Score | Overview |
|----------|-------|----------|
[...]

PR #3053, opened 2025-12-03
chore: regenerate lockfile (FXC-4403)

daquinteroflex:
The daily-updated lockfile upgrade PR is fixed on the https://github.com/flexcompute/tidy3d/pull/3046

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -71,25 +71,24 @@ files = [
 
 [[package]]
 name = "anyio"
-version = "4.11.0"
+version = "4.12.0"
 description = "High-level concurrency and networking framework on top of asyncio or Trio"
 optional = true
 python-versions = ">=3.9"
 groups = ["main"]
 markers = "extra == \"dev\" or extra == \"docs\""
 files = [
-    {file = "anyio-4.11.0-py3-none-any.whl", hash = "sha256:0287e96f4d26d4149305414d4e3bc32f0dcd0862365a4bddea19d7a1ec38c4fc"},
-    {file = "anyio-4.11.0.tar.gz", hash = "sha256:82a8d0b81e318cc5ce71a5f1f8b5c4e63619620b63141ef8c995fa0db95a57c4"},
+    {file = "anyio-4.12.0-py3-none-any.whl", hash = "sha256:dad2376a628f98eeca4881fc56cd06affd18f659b17a747d3ff0307ced94b1bb"},
+    {file = "anyio-4.12.0.tar.gz", hash = "sha256:73c693b567b0c55130c104d0b43a9baf3aa6a31fc6110116509f27bf75e21ec0"},
 ]
 
 [package.dependencies]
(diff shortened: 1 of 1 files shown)


mahlau-flex:
@daquinteroflex Since the fonttools dependency is not optional, would it make sense to explicitly include it in the pyproject.toml with a minimum required version? This would not lead to an additional dependency for a user and explicitly restrict the version for everyone installing tidy3d.

mahlau-flex:
We have matplotlib as a dependency, but it might take a while for them to update the fonttools dependency

daquinteroflex:
@mahlau-flex interesting so is this an error you found when using tidy3d via pip or just from the lockfile you identified this is an issue?

mahlau-flex:
when pushing my pydanticV2 rebases, git automatically gave me the following notice on the commandline since we have setup the dependabot alerts:

```
remote: Resolving deltas: 100% (274/274), completed with 268 local objects.
remote:
remote: GitHub found 1 vulnerability on flexcompute/tidy3d's default branch (1 moderate). To find out more, visit:
remote:      https://github.com/flexcompute/tidy3d/security/dependabot/40
remote:
```

daquinteroflex:
So it looks like our lockfile upgrades to "fonttools-4.61.0-cp" which complies with "Upgrade fonttools to version 4.60.2 or later. ". 

So this should be enough for now? Internal code should use the lockfile dependencies anyway and I prefer matplotlib to handle their own dependencies since every dependency we explicitly add to the pyproject.toml is equivalent to a proportional amount of maintenance pain.

mahlau-flex:
@daquinteroflex If we only update the lockfile we only fix the security vulnerability for our own developers. Any user that installs tidy3d may still install the problematic fonttools library through the dependencies, such that their system may become vulnerable indirectly through installing our library. I think this should be adressed as well.

If we include a direct dependency to fonttools >= 4.61.0 in our pyproject.toml then any package manager should know only to install the correct version. Ideally, matplotlib should update their dependencies and we update the dependency to the corresponding matplotlib version, but this may take time.

daquinteroflex:
Will leave it up to @yaugenst-flex judgment, I'd think any pip installation of matplotlib would simply use the latest fonttools
https://github.com/matplotlib/matplotlib/blob/3323161b83c56f90cf63d2b30c24a65b8b12c953/pyproject.toml#L36

it can just become a slippery slope of trying to patch our dependencies' dependencies for third party vulnerability issues and becomes a lot harder to mantain imo when they do upgrade their dependencies (as our own project uses those packages indirectly really)

yaugenst-flex:
I think @daquinteroflex is right, there is no code in tidy3d that ingests user-provided font files so the vulnerability is effectively not exploitable through tidy3d, we don't let users upload custom fonts. I agree we shouldn't be pinning transitive dependencies if not absolutely necessary.

mahlau-flex (approved):

PR #3057, opened 2025-12-04
fix(mode): ensure degenerate modes are orthogonal from mode solver

dmarek-flex:
Remaining considerations:

- ~bi-orthogonality for fully anisotropic media?~ bi-orthogonality is only violated for non-reciprocal media
- ~better way to identify degenerate modes~ two step procedure now
- ~should we make the user aware of detected degenerate modes?~ no
- ~how does bi-orthogonality interact with bend or angled transforms~ should apply in transformed system

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR fixes a mathematical issue with degenerate modes in the mode solver. When eigenmodes have identical or nearly identical eigenvalues (degenerate modes), the eigensolver doesn't guarantee orthogonality between them. This PR ensures degenerate modes respect the bi-orthogonality condition by:

- Adding `TOL_DEGENERATE` constant for identifying degenerate modes
- Implementing `_identify_degenerate_modes()` to group modes with similar effective indices
- Adding `_dot()` and `_outer_dot()` for computing bi-orthogonal overlap integrals
- Adding `_normalize_modes()` to normalize modes to unit self-overlap
- Implementing `_make_orthogonal_basis_for_degenerate_modes()` using a complex-symmetric eigendecomposition approach

The implementation correctly handles the complex-symmetric (not Hermitian) nature of the bi-orthogonal inner product, using unconjugated dot products for eigenvector normalization.

- This PR is safe to merge with minor considerations around edge case handling.
- The implementation is mathematically sound and follows the correct approach for bi-orthogonal mode orthogonalization. The test verifies the core functionality. Minor concern about potential numerical edge cases with near-zero eigenvalues in the orthogonalization transformation.
- `tidy3d/components/mode/solver.py` - the `_make_orthogonal_basis_for_degenerate_modes` method could benefit from a check for near-zero eigenvalues.

File Analysis

| CHANGELOG.md | 5/5 | Added changelog entry for degenerate mode handling fix under the "Fixed" section, following project conventions. |
| tests/test_plugins/test_mode_solver.py | 5/5 | Added test `test_degenerate_mode_processing` that verifies degenerate modes are bi-orthogonal by checking off-diagonal overlap values are below threshold. |
| tidy3d/components/mode/solver.py | 4/5 | Added `TOL_DEGENERATE` constant and implemented degenerate mode orthogonalization via `_identify_degenerate_modes`, `_dot`, `_outer_dot`, `_normalize_modes`, and `_make_orthogonal_basis_for_degenerate_modes` methods. The implementation handles complex symmetric matrices correctly. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant ModeSolver
    participant EigSolver
    participant compute_modes

    User->>ModeSolver: data_raw
    ModeSolver->>compute_modes: compute_modes(...)
    compute_modes->>EigSolver: solver_em()
    EigSolver-->>compute_modes: E, H, neff, keff
    
    Note over compute_modes: Transform back to original axes
    
    compute_modes->>EigSolver: _normalize_modes(E, H, dl_f, dl_b, all_modes)
    Note over EigSolver: Normalize each mode to unit self-overlap
    EigSolver-->>compute_modes: E, H (normalized)
    
    compute_modes->>EigSolver: _identify_degenerate_modes(n_complex, TOL_DEGENERATE)
    Note over EigSolver: Group modes with similar n_complex
    EigSolver-->>compute_modes: degenerate_sets: [{0,1}, ...]
    
    compute_modes->>EigSolver: _make_orthogonal_basis_for_degenerate_modes(...)
    
    loop For each degenerate group
        EigSolver->>EigSolver: _outer_dot() → compute overlap matrix S
        EigSolver->>EigSolver: np.linalg.eig(S) → diagonalize
        EigSolver->>EigSolver: Renormalize eigenvectors (v.T @ v = 1)
        EigSolver->>EigSolver: W = V @ Λ^(-1/2) @ V.T
        EigSolver->>EigSolver: Transform E, H by W
    end
    
    EigSolver-->>compute_modes: E, H (orthogonalized)
    compute_modes-->>ModeSolver: fields, n_complex, eps_spec
    ModeSolver-->>User: ModeSolverData
```

### DIFF
--- a/tests/test_plugins/test_mode_solver.py
+++ b/tests/test_plugins/test_mode_solver.py
@@ -13,7 +13,7 @@ import tidy3d.plugins.mode.web as msweb
 from tidy3d import ScalarFieldDataArray
 from tidy3d.components.data.monitor_data import ModeSolverData
 from tidy3d.components.mode.derivatives import create_sfactor_b, create_sfactor_f
-from tidy3d.components.mode.solver import compute_modes
+from tidy3d.components.mode.solver import TOL_DEGENERATE_CANDIDATE, EigSolver, compute_modes
 from tidy3d.components.mode_spec import MODE_DATA_KEYS
 from tidy3d.exceptions import DataError, SetupError
 from tidy3d.plugins.mode import ModeSolver
@@ -963,7 +963,16 @@ def test_mode_solver_nan_pol_fraction():
 
     md = ms.solve()
     check_ms_reduction(ms)
+    # Inject NaN at mode_index=5 for selected field components
+    nan_fields = {}
+    for field_name in ["Ex", "Ez", "Hx", "Hz"]:
+        field = getattr(md, field_name)
+        data = field.values.copy()
+        data[..., 5] = np.nan
+        nan_fields[field_name] = field.copy(data=data)
 
+    md = md.updated_copy(**nan_fields)
+    md = ms._filter_polarization(md)
     assert list(np.where(np.isnan(md.pol_fraction.te))[1]) == [9]
 
 
@@ -1503,3 +1512,52 @@ def test_sort_spec_track_freq():
     assert np.allclose(modes_lowest.Ex.abs, modes_lowest_retracked.Ex.abs)
     assert np.all(modes_lowest.n_eff == modes_lowest_retracked.n_eff)
     assert np.all(modes_lowest.n_group == modes_lowest_retracked.n_group)
+
+
+def test_degenerate_mode_processing():
+    """Ensure degenerate modes returned by mode solver are bi-orthogonal."""
+    freq0 = td.C_0
+    sim_size = (0, 2, 2)
+    inf = 10
+    W1 = 0.3
+    n = 1.5
+    num_modes = 4
+    mode_spec = td.ModeSpec(num_modes=num_modes)
+    medium = td.Medium(permittivity=n**2)
+    geom1 = td.Box.from_bounds((-inf, -W1 / 2, -W1 / 2), (inf, W1 / 2, W1 / 2))
+    wg1 = td.Structure(geometry=geom1, medium=medium)
+
+    grid_spec = td.GridSpec.uniform(dl=0.2)
+
+    sim = td.Simulation(
+        size=sim_size,
+        structures=[wg1],
+        grid_spec=grid_spec,
+        run_time=10 / freq0,
+    )
+
+    ms = ModeSolver(
+        simulation=sim,
+        plane=sim.geometry,
+        mode_spec=mode_spec,
+        freqs=[freq0],
+        direction="+",
+    )
+
+    mode_data = ms.data_raw
+
+    degen_sets = EigSolver._identify_degenerate_modes(
+        mode_data.n_complex.values[0, :], TOL_DEGENERATE_CANDIDATE
+    )
+    assert len(degen_sets) == 1
+
+    S = mode_data.outer_dot(mode_data, conjugate=False).isel(f=0).values
+    threshold = 1e-7
+    off_diag_mask = ~np.eye(S.shape[0], dtype=bool)
+    large_vals = np.abs(S) > threshold
+    problem_mask = off_diag_mask & large_vals
+
+    indices = np.argwhere(problem_mask)
+    msg = f"Found {len(indices)} off-diagonal values > {threshold}:\n"
+    msg += "\n".join(f"  |S[{i},{j}]| = {np.abs(S[i, j]):.4e}" for i, j in indices)
+    assert not np.any(problem_mask), msg
--- a/tidy3d/components/mode/solver.py
+++ b/tidy3d/components/mode/solver.py
@@ -18,6 +18,10 @@ from .transforms import angled_transform, radial_transform
 TOL_COMPLEX = 1e-10
 # Tolerance for eigs
 TOL_EIGS = fp_eps / 10
+# Relative tolerance of complex effective index for finding candidate degenerate modes
+TOL_DEGENERATE_CANDIDATE = 1e3 * TOL_EIGS
+# Tolerance used to determine whether possible degenerate modes need manual orthogonalization
+TOL_NEEDS_ORTHOGONALIZATION = 1e2 * TOL_EIGS
 # Tolerance for deciding on the matrix to be diagonal or tensorial
 TOL_TENSORIAL = 1e-6
 # shift target neff by this value, both rel and abs, whichever results in larger shift
@@ -285,6 +289,7 @@ class EigSolver(Tidy3dBaseModel):
         E = E.reshape((3, Nx, Ny, 1, num_modes))
         H = np.sum(jac_h[..., None] * H[:, None, ...], axis=0)
         H = H.reshape((3, Nx, Ny, 1, num_modes))
+
         fields = np.stack((E, H), axis=0)
 
         neff = neff * np.linalg.norm(kp_to_k)
@@ -382,6 +387,13 @@ class EigSolver(Tidy3dBaseModel):
         mu_offd = np.abs(mu_tensor[off_diagonals])
         is_tensorial = np.any(eps_offd > TOL_TENSORIAL) or np.any(mu_offd > TOL_TENSORIAL)
 
+        # Determine if ``eps`` and ``mu`` represent reciprocal media
+        is_reciprocal = True
+        if is_tensorial:
+            is_reciprocal = cls._check_reciprocity(
+                eps_tensor, TOL_TENSORIAL
+            ) and cls._check_reciprocity(mu_tensor, TOL_TENSORIAL)
+
         # initial vector for eigensolver in correct data type
         vec_init = cls.set_initial_vec(Nx, Ny, is_tensorial=is_tensorial)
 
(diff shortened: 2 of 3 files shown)


tests/test_plugins/test_mode_solver.py
```diff
@@ -1503,3 +1503,52 @@ def test_sort_spec_track_freq():
...
+    )
+
+    mode_data = ms.data_raw
+
+    degen_sets = EigSolver._identify_degenerate_modes(
+        mode_data.n_complex.values[0, :], TOL_DEGENERATE
+    )
+    assert len(degen_sets) == 1
+
+    S = mode_data.outer_dot(mode_data, conjugate=True).isel(f=0).values
```
caseyflex: shouldn't this be conjugate=False?

caseyflex (commented):
Still testing with EME, sent you a slack message. Just a preliminary comment, will give more feedback tomorrow. This is great though, thanks @dmarek-flex !

tidy3d/components/mode/solver.py
```diff
@@ -1089,6 +1112,248 @@ def mode_plane_contain_good_conductor(material_response) -> bool:
...
+
+        # Make the differential area elements
+        Ex_Hy_dS = np.outer(dl_primal[0], dl_dual[1])
+        Ey_Hx_dS = np.outer(dl_dual[0], dl_primal[1])
+
+        term1 = Ex[..., mode_1] * Hy[..., mode_2] + Ex[..., mode_2] * Hy[..., mode_1]
+        term1 *= Ex_Hy_dS[..., np.newaxis]
+        term2 = Ey[..., mode_1] * Hx[..., mode_2] + Ey[..., mode_2] * Hx[..., mode_1]
+        term2 *= Ey_Hx_dS[..., np.newaxis]
+        return np.sum(term1 - term2)
```
caseyflex: I think we want a factor of 1/2 here for consistency with the usage elsewhere in tidy3d (e.g. outer_dot with conj=False)

caseyflex (approved):
This is great, thanks! Very exciting!

caseyflex:
Added @momchil-flex as a reviewer because he will be interested

dmarek-flex:
@dbochkov-flexcompute @momchil-flex I had to change one test because after normalization the nan disappeared. So I modified it a bit to find a mode exactly at cut off, real and imaginary part of n_complex are 0. Any slight change though and the mode shifts away from the exact cutoff.

momchil-flex:
> * ~bi-orthogonality for fully anisotropic media?~ bi-orthogonality is only violated for non-reciprocal media

Gemini helped me understand this better. Indeed the usual bi-orthogonality is dependent on reciprocity. But bi-orthogonal modes can still be defined if you do two solves, the second one being for the backward modes in the transposed system. Not saying we should do this for now just good to understand. :) Is your plan basically to ignore non-reciprocal media for now?

Need a bit more time to go through the code, but also, as a general question, is the plan to leave all the definitions in the solver.py file? I think it probably makes sense. As usual I feel a bit weird about having multiple definitions of the same operations (dot, outer_dot) but I do see the utility of having them defined on the raw eigenvectors. For the sake of argument, have you considered an implementation on the non-colocated fields of a ModeSolverData? Because even if the mode solver requires colocation, we first solve for the fields on the yee grid before we apply further post-processing, and I think there's not that much difference between those and the ones in the solver file?

tidy3d/components/mode/solver.py
```diff
@@ -1089,6 +1110,340 @@ def mode_plane_contain_good_conductor(material_response) -> bool:
...
+        dl_primal: np.ndarray,
+        dl_dual: np.ndarray,
+        mode_idx: int,
+    ) -> complex:
+        """Calculates the upper bound for the self-dot (overlap) using the Cauchy-Schwarz inequality.
+
+        Hard to predict how the normalization of the eigenvectors from the linear algebra solver
+        translates to normalization in terms of the EM overlap integral,
+        so we use this upper bound to determine whether a mode should be considered as self-orthogonal.
+        """
```
momchil-flex, line 1258: I don't really understand how this cauchy schwartz bound is defined. I mean, the way it is used, it doesn't need to be exact, you just need some decent reference value, but I have a few points of confusion:

- It seems like you are using something like `|E|^2` here instead of an `E x H` kind of product. But it doesn't really seem to be using the EM energy definition of something like `eps * |E|^2 + mu * |H|^2`. It rather seems to be taking a Cauchy-Schwartz as if `|E|^2` is the inner product of the left vector and `|H|^2` is the one for the right vector, which I don't understand.
- Also, the `ExH` dot product which is used everywhere else is not positive-definite, and Cauchy-Schwartz wouldn't hold.

momchil-flex (commented):
Generally looks good! I have some confusion about the Cauchy-Schwartz bound.

And yeah just generally wondering what the plan is for changing the dot product to avoid colocation in general. Likely just leave the current additioins to `solver.py` as they are, and then work on also modifying the `FieldData.dot` method separately and together with flux and projection modifications? Maybe worth exposing some sort of way to use the non-colocated `dot` already though (separate PR). I know Casey is looking forward to testing it in EME, and I'm also curious if it might improve the modal decomposition accuracy in some cases, e.g. with bends + angles which have been a problem for Xanadu at the accuracy they are after.

tidy3d/components/mode/solver.py
caseyflex: The dot product might be positive definite on the forward modes of a single plane. It was when comparing modes of different cross-sections (like at an interface) where we get overlaps >= 1 (even analytically)

dmarek-flex: I'd say it is a pretty loose usage of the cauchy schwarz inequality, but so far simplest idea for checking self-orthogonality. It should work because ultimately I compare to the absolute value of the `dot`, which makes it positive definite (at least when I compare it). 

Then using an argument like (thanks gemini):

$$\left| \iint (\mathbf{E} \times \mathbf{H}) \cdot \hat{\mathbf{z}} ) \, dA \right| \leq \iint \left| (\mathbf{E} \times \mathbf{H}) \cdot \hat{\mathbf{z}} \right| \, dA$$

$$|(\mathbf{E} \times \mathbf{H}) \cdot \hat{\mathbf{z}}| \leq |\mathbf{E} \times \mathbf{H}| \leq |\mathbf{E}| |\mathbf{H}|$$

allows us to use the actual inequality

$$\iint |\mathbf{E}| |\mathbf{H}| \, dA \leq \sqrt{\iint |\mathbf{E}|^2 \, dA} \sqrt{\iint |\mathbf{H}|^2 \, dA}$$

dmarek-flex: And the reason I use this is because the eigenvectors produced by the linear algebra packages will be normalized differently depending on the type of problem, like diagonal produces normalized transverse E vector (|Et| = 1), but I believe the anisotropic version will probably be a combination of Et and Ht. And these normalizations don't include the transverse grid units.

So this seemed like a sort of physical way to gauge how self-orthogonal an eigenmode is. Honestly probably a bit overkill, because I noticed that when a mode with n_complex=0 is found, the returned fields will include NaN

momchil-flex: Ah ok this makes sense thanks!

dmarek-flex:
>  Is your plan basically to ignore non-reciprocal media for now?

Yes, but now that you mention it I should add a step that checks if the `eps` and `mu` are reciprocal, if not I will skip the orthogonalization steps.

> Likely just leave the current additioins to solver.py as they are, and then work on also modifying the FieldData.dot method separately and together with flux and projection modifications?

Yes this is my plan, I also agree with what you said in your earlier comment regarding multiple definitions. In this case I think the data structures are different enough to warrant separate definitions of these functions. See my next response though.

> For the sake of argument, have you considered an implementation on the non-colocated fields of a ModeSolverData?

You mean reusing the `EigSolver._dot` to implement a `ModeSolverData.dot`? I haven't seriously considered it, but it is definitely possible to reuse those methods. Just need to prepare the correct numpy arrays and handle multiple frequencies, but I can probably reuse the code in EigSolver to implement the dots.

> Maybe worth exposing some sort of way to use the non-colocated dot already though (separate PR).

Yes, I'll submit a PR today that enables non-colocated dot so that @caseyflex can test.

momchil-flex:
> > For the sake of argument, have you considered an implementation on the non-colocated fields of a ModeSolverData?
> 
> You mean reusing the `EigSolver._dot` to implement a `ModeSolverData.dot`? I haven't seriously considered it, but it is definitely possible to reuse those methods. Just need to prepare the correct numpy arrays and handle multiple frequencies, but I can probably reuse the code in EigSolver to implement the dots.

Actually, I meant it the other way round - move `EigSolver._dot` outside of the solver code, to a place where it can be used by the ModeFieldData directly, if it can produce the same results for un-colocated data (especially the numerical precision level overlap for modes that should be orthogonal). We always call [data_on_yee_grid](https://github.com/flexcompute/tidy3d/blob/e33c7e4d2a141ab95d7b35b9bad39460925a880c/tidy3d/components/mode/mode_solver.py#L1174) first which does not do much massaging of the raw data coming from the solver. So I could potentially imagine doing the degeneracy fixing after this function, rather than inside the solver. There could be some caveats I'm missing, just bringing it up for discussion. One thing for example to be careful with is the fact that the integration boundaries should *not* be truncated to the monitor plane in that case but instead the full dl-s should be used at the edge pixels too.

dmarek-flex:
> There could be some caveats I'm missing, just bringing it up for discussion. One thing for example to be careful with is the fact that the integration boundaries should not be truncated to the monitor plane in that case but instead the full dl-s should be used at the edge pixels too.

I see, but yea I like the idea of "cleaning" the modes directly in the mode solver for reasons like your point about handling edge pixels near the monitor boundary. I think another reason is that it is potentially more robust to apply the bi-orthogonality integral in the plane normal to the propagation direction, so when coordinate transformations are involved (bent and slanted waveguides), to get best results we need to orthogonalize degenerate modes in the transformed coordinate system. Not so sure about this second point to be honest.

momchil-flex (approved):
Indeed, that also makes sense, thanks for bearing with me. :)

PR #3064, opened 2025-12-09
fix(web): Add 'warning' to post validate states

momchil-flex:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR adds a new "warning" state to `POST_VALIDATE_STATES` to handle an additional valid post-validation state from the backend.

- Added "warning" to `POST_VALIDATE_STATES` set which propagates to `RUNNING_STATES` and `ALL_POST_VALIDATE_STATES` through set unions
- **Missing update**: The "warning" state was not added to `STATE_PROGRESS_PERCENTAGE` dictionary, which can cause `KeyError` exceptions in `webapi.py` when displaying progress bars for tasks with "warning" status

<h3>Confidence Score: 2/5</h3>

- This PR introduces a potential KeyError bug due to incomplete propagation of the new state across related dictionaries
- Score of 2 reflects a likely runtime error: 'warning' state added to POST_VALIDATE_STATES but not to STATE_PROGRESS_PERCENTAGE, causing KeyError when webapi.py accesses progress percentage for tasks with 'warning' status
- tidy3d/web/api/states.py needs STATE_PROGRESS_PERCENTAGE update for 'warning' state

File Analysis

| tidy3d/web/api/states.py | 3/5 | Adds 'warning' to POST_VALIDATE_STATES, but 'warning' is missing from STATE_PROGRESS_PERCENTAGE which can cause KeyError in webapi.py |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant webapi.py
    participant states.py
    participant Backend

    User->>webapi.py: run_batch() / monitor task
    webapi.py->>Backend: get task status
    Backend-->>webapi.py: status = "warning"
    webapi.py->>states.py: POST_VALIDATE_STATES check
    states.py-->>webapi.py: "warning" in set ✓
    webapi.py->>states.py: STATE_PROGRESS_PERCENTAGE["warning"]
    states.py-->>webapi.py: KeyError ✗
```

**Context used:**

- Rule from `dashboard` - When modifying a piece of logic, ensure the change is propagated to all independent functions or cal... ([source](https://app.greptile.com/review/custom-context?memory=809457d4-6b33-46ed-b49b-37057d02807e))

### DIFF
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -37,10 +37,11 @@ from tidy3d.web.api import webapi as web
 from tidy3d.web.api.states import (
     COMPLETED_PERCENT,
     COMPLETED_STATES,
+    DRAFT_STATES,
     END_STATES,
     ERROR_STATES,
     PRE_ERROR_STATES,
-    PRE_VALIDATE_STATES,
+    QUEUED_STATES,
     RUNNING_STATES,
     STATE_PROGRESS_PERCENTAGE,
 )
@@ -1068,7 +1069,7 @@ class Batch(WebContainer):
                 status_part = f"→ [red]{status:<{status_width}}"
             elif status in COMPLETED_STATES:
                 status_part = f"→ [green]{status:<{status_width}}"
-            elif status in (PRE_ERROR_STATES | PRE_VALIDATE_STATES):
+            elif status in (PRE_ERROR_STATES | DRAFT_STATES | QUEUED_STATES):
                 status_part = f"→ [yellow]{status:<{status_width}}"
             elif status in RUNNING_STATES:
                 status_part = f"→ [blue]{status:<{status_width}}"
--- a/tidy3d/web/api/states.py
+++ b/tidy3d/web/api/states.py
@@ -1,5 +1,18 @@
 from __future__ import annotations
 
+# progression order for a typical run
+PROGRESSION_ORDER = (
+    "draft",
+    "queued",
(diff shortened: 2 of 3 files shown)


momchil-flex:
@daquintero the comment from AI above made me completely revisit the `states.py` file. I think it was doing too much. After the CM refactor, we are back to where a typical run (even for CMs) would be `draft -> queued -> preprocess -> running -> postprocess -> success`. For example, validate statuses still go in `metadataStatus` instead of the main `status` field, and in general validation shouldn't be part of the progression: it should map to 0% done, because the idea of a workflow is that a user could modify the uploaded simulation and validate it multiple times before finally running, i.e. we should only increment the percentage once the task is *submitted*, and then it goes into queued, etc. Let me know what you think.

@copilot  and how about you buddy?

momchil-flex:
@greptile what do you think?

momchil-flex:
Note as in my previous comment that I actually do think "warning" should still be 0% since no part of the solver has been started / run. Note also that previously these `"validating"`, etc. state progressions were not really being hit, maybe only in the CM workflow in rc2, but not after the refactor or in other solvers. We use the `metadataStatus`, not the full task `status`, which just sits in "draft" when validation happens. I'm keeping all the possible options though just in case even if `status` should now never be e.g. `validating`.

daquinteroflex (approved):
Looks like it's going where it should go!

PR #3082, opened 2025-12-12
feat(tidy3d): FXC-4520-add-autograd-support-for-sphere

marcorudolphflex:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR adds autograd support for the `Sphere` geometry class, enabling automatic differentiation for inverse design optimization. The implementation follows the same pattern as `Cylinder` and other geometries by overriding the `radius` field with `TracedSize1D` and implementing the `_compute_derivatives` method.

**Key changes:**
- Overrides `radius` field with `TracedSize1D` to enable autograd tracking
- Implements `_compute_derivatives` method that discretizes the sphere surface using icosphere triangulation and computes gradients with respect to `radius` and `center` parameters
- Adds helper methods `to_triangle_mesh`, `_triangulated_surface`, `_discretization_wavelength`, and `_tangent_basis_from_normals`
- Includes comprehensive numerical validation tests comparing autograd gradients with finite-difference approximations
- Adds basic integration test to verify non-zero gradients

**Issues found:**
- Division by zero risk in normal computation (line 373 in primitives.py) - missing safety check that polyslab.py uses
- Test pollution from global variable modification in `test_triangle_sphere_fd_step_sweep_ref` (line 530-531)

- This PR requires fixes for two logic issues before merging - division by zero risk and test pollution
- Score reflects solid implementation following established patterns, but two critical logic issues need addressing: (1) unprotected division by zero when computing normals could cause NaN/inf values, and (2) global variable modification in tests causes test pollution. Both are straightforward to fix.
- Pay close attention to `tidy3d/components/geometry/primitives.py` lines 372-373 (division by zero) and `tests/test_components/autograd/numerical/test_autograd_sphere_trianglemesh_numerical.py` line 530-531 (global modification)

File Analysis

| tidy3d/components/geometry/primitives.py | 4/5 | Core implementation of Sphere autograd support with gradient computation for radius and center parameters |
| tests/test_components/autograd/numerical/test_autograd_sphere_trianglemesh_numerical.py | 4/5 | Comprehensive numerical validation tests comparing finite-difference with autograd gradients for Sphere |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Sphere
    participant _compute_derivatives
    participant _triangulated_surface
    participant TriangleMesh
    participant DerivativeInfo
    
    User->>Sphere: Create Sphere(center, radius)
    User->>Sphere: Compute gradients
    Sphere->>_compute_derivatives: derivative_info
    _compute_derivatives->>_compute_derivatives: Validate derivative paths
    _compute_derivatives->>_triangulated_surface: Discretize sphere surface
    _triangulated_surface->>_triangulated_surface: Get unit sphere triangles
    _triangulated_surface-->>_compute_derivatives: Physical triangles
    _compute_derivatives->>TriangleMesh: Convert to trimesh object
    TriangleMesh-->>_compute_derivatives: vertices, faces, areas
    _compute_derivatives->>_compute_derivatives: Compute normals from vertices
    _compute_derivatives->>_compute_derivatives: Filter vertices inside bounds
    _compute_derivatives->>_compute_derivatives: Compute tangent basis
    _compute_derivatives->>DerivativeInfo: Evaluate gradient at points
    DerivativeInfo-->>_compute_derivatives: Gradient values
    _compute_derivatives->>_compute_derivatives: Weight by face areas
    _compute_derivatives->>_compute_derivatives: Integrate to get vjps
    _compute_derivatives-->>Sphere: Gradients for radius/center
    Sphere-->>User: Autograd gradients
```

---

> [!NOTE]
> Adds automatic differentiation for `Sphere` and aligns discretization logic.
> 
> - Implements autograd for `Sphere` (`primitives.py`): traced `radius`, `to_triangle_mesh`, surface triangulation, tangent-basis helper, and `_compute_derivatives` for `radius`/`center`, including 2D (collapsed-axis) handling via `Cylinder` mapping
> - Introduces shared `discretization_wavelength()` used by `Sphere` and refactors `Cylinder` to use it
> - Updates JSON schemas (`Simulation*.json`, `ModeSimulation.json`, etc.) to allow autograd tracer for `Sphere.radius`
> - Expands tests: numerical gradient checks (FD vs autograd) in 3D/2D, Sphere↔Cylinder 2D consistency, FD step sweeps, and zero-radius gradient behavior; enables previously skipped numerical test
> - Updates changelog to mention autograd support for `Sphere`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit df19d272711a285f766231270b7d2e91fb480722. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/schemas/TerminalComponentModeler.json
+++ b/schemas/TerminalComponentModeler.json
@@ -16498,8 +16498,15 @@
           ]
         },
         "radius": {
-          "minimum": 0,
-          "type": "number"
+          "anyOf": [
+            {
+              "minimum": 0,
+              "type": "number"
+            },
+            {
+              "type": "autograd.tracer.Box"
+            }
+          ]
         },
         "type": {
           "default": "Sphere",
--- a/tests/test_components/autograd/test_autograd.py
+++ b/tests/test_components/autograd/test_autograd.py
@@ -28,6 +28,7 @@ from tidy3d.components.autograd.field_map import FieldMap
 from tidy3d.components.autograd.utils import get_static, is_tidy_box
 from tidy3d.components.base import TRACED_FIELD_KEYS_ATTR
 from tidy3d.components.data.data_array import DataArray
+from tidy3d.components.geometry.primitives import discretization_wavelength
 from tidy3d.config import config
 from tidy3d.exceptions import AdjointError
 from tidy3d.plugins.polyslab import ComplexPolySlab
@@ -569,6 +570,14 @@ def make_structures(params: anp.ndarray) -> dict[str, td.Structure]:
         medium=td.Medium(permittivity=mesh_eps),
     )
 
+    # use first 4 params for radius/center, rest for eps
+    cx, cy, cz = params[1:4]
+    sphere_geom = td.Sphere(radius=params[0] + 1, center=(cx, cy, cz))
+    sphere = td.Structure(
+        geometry=sphere_geom,
+        medium=td.Medium(permittivity=anp.mean(params[4:]) + 2),
+    )
+
     return {
         "medium": medium,
         "center_list": center_list,
@@ -584,6 +593,7 @@ def make_structures(params: anp.ndarray) -> dict[str, td.Structure]:
         "custom_pole_res": custom_pole_res,
         "cylinder": cylinder,
         "triangle_mesh": triangle_mesh,
+        "sphere": sphere,
     }
 
 
@@ -682,6 +692,7 @@ structure_keys_ = (
     "custom_pole_res",
     "cylinder",
     "triangle_mesh",
+    "sphere",
 )
 monitor_keys_ = ("mode", "diff", "field_vol", "field_point")
 
@@ -1991,10 +2002,8 @@ def test_cylinder_discretization(eps_real):
     with AssertLogLevel(
         "WARNING", contains_str="The minimum wavelength inside the cylinder material"
     ):
-        cylinder = td.Cylinder(axis=2, length=info.wavelength_min, radius=2 * info.wavelength_min)
-
         expected_wvl_mat = info.wavelength_min * config.adjoint.min_wvl_fraction
-        wvl_mat = cylinder._discretization_wavelength(derivative_info=info)
+        wvl_mat = discretization_wavelength(info, "cylinder")
 
         assert np.isclose(expected_wvl_mat, wvl_mat), (
             "Unexpected wavelength for discretizing cylinder!"
@@ -3336,3 +3345,33 @@ def test_geometry_group_passes_intersected_bounds_to_children():
     assert object.__getattribute__(big_box, "recorded_bounds_intersect") == group.bounds, (
         f"got {object.__getattribute__(big_box, 'recorded_bounds_intersect')} and {group.bounds}"
     )
+
+
+@pytest.mark.parametrize("monitor_key", ("mode",))
+def test_autograd_sphere_0_radius(use_emulated_run, monitor_key):
+    """Integration test that Sphere gradients are non-zero (mirrors cylinder check)."""
+
+    monitor, postprocess = make_monitors()[monitor_key]
+
+    def make_sphere(radius, x0, y0, z0):
+        return td.Sphere(center=(x0, y0, z0), radius=radius)
+
+    def make_sim(params):
+        geometry = make_sphere(*params)
+        structure = td.Structure(geometry=geometry, medium=td.Medium(permittivity=2))
+        return SIM_BASE.updated_copy(structures=[structure], monitors=[monitor])
+
+    p0 = [0.0, 0.0, 0.0, 0.0]
+
+    def objective(params):
+        sim = make_sim(params)
+        if PLOT_SIM:
+            plot_sim(sim, plot_eps=True)
+        data = run(sim, task_name="autograd_test", verbose=False)
+        return anp.sum(anp.abs(data[monitor.name].amps)).item()
+
+    with AssertLogLevel("WARNING", contains_str="cannot be computed"):
+        val_sphere, grad_sphere = ag.value_and_grad(objective)(p0)
+    # first 4 parameters are related to the geometry
+    geom_grad = np.asarray(get_static(grad_sphere[:4]), dtype=float)
+    assert np.allclose(geom_grad, 0.0)
--- a/tidy3d/components/geometry/primitives.py
+++ b/tidy3d/components/geometry/primitives.py
@@ -12,7 +12,7 @@ import shapely
 from pydantic.v1 import PrivateAttr
 from shapely.geometry.base import BaseGeometry
 
-from tidy3d.components.autograd import AutogradFieldMap, TracedSize1D
+from tidy3d.components.autograd import AutogradFieldMap, TracedSize1D, get_static
 from tidy3d.components.autograd.derivative_utils import DerivativeInfo
 from tidy3d.components.base import cached_property, skip_if_fields_missing
 from tidy3d.components.geometry import base
@@ -90,6 +90,25 @@ def _base_icosahedron() -> tuple[np.ndarray, np.ndarray]:
 _ICOSAHEDRON_VERTS, _ICOSAHEDRON_FACES = _base_icosahedron()
 
 
+def discretization_wavelength(derivative_info: DerivativeInfo, geometry_label: str) -> float:
+    """Choose reference wavelength for surface discretization."""
+    wvl0_min = derivative_info.wavelength_min
+    wvl_mat = wvl0_min / np.max([1.0, np.max(np.sqrt(abs(derivative_info.eps_in)))])
+
+    grid_cfg = config.adjoint
+
+    min_wvl_mat = grid_cfg.min_wvl_fraction * wvl0_min
+    if wvl_mat < min_wvl_mat:
+        log.warning(
+            f"The minimum wavelength inside the {geometry_label} material is {wvl_mat:.3e} μm, which would "
+            f"create a large number of discretization points for computing the gradient. "
+            f"To prevent performance degradation, the discretization wavelength has "
+            f"been clipped to {min_wvl_mat:.3e} μm.",
+            log_once=True,
+        )
+    return max(wvl_mat, min_wvl_mat)
+
+
 class Sphere(base.Centered, base.Circular):
     """Spherical geometry.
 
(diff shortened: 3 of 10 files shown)


marcorudolphflex:
@greptile

tidy3d/components/geometry/primitives.py
```diff
@@ -278,6 +331,133 @@ def unit_sphere_triangles(
...
+
+    def _triangulated_surface(
+        self,
+        *,
+        max_edge_length: Optional[float] = None,
+        subdivisions: Optional[int] = None,
+    ) -> tuple[np.ndarray, np.ndarray]:
+        """Return physical and unit triangles for the surface discretization."""
+
+        unit_tris = self._unit_sphere_triangles(
```
yaugenst-flex, line 541: `self._unit_sphere_triangles()` checks `if target_edge_length is not None and subdivisions is not None`, and here both `subdivisions` and `max_edge_length` are always non-`None` so I think this will always raise?

tests/test_components/autograd/test_autograd.py
```diff
@@ -3199,3 +3199,30 @@ def record_method(self, derivative_info):
...
+
+    monitor, postprocess = make_monitors()[monitor_key]
+
+    def make_sphere(radius, x0, y0, z0):
+        return td.Sphere(center=(x0, y0, z0), radius=radius)
+
+    def make_sim(params):
+        geometry = make_sphere(*params)
+        structure = td.Structure(geometry=geometry, medium=td.Medium(permittivity=2))
+        return SIM_BASE.updated_copy(structures=[structure], monitors=[monitor])
```
yaugenst-flex, line 3362: `SIM_BASE` is actually 2D by default (check top of file). Since `Sphere._compute_derivatives` errors for 2D currently this should fail? Probably unintended? That being said, I feel like we should handle the 2D case, I know that I've set up 2d simulations using spheres instead of cylinders in the past...

tidy3d/components/geometry/primitives.py
```diff
@@ -312,6 +492,42 @@ def _subdivisions_for_edge(self, target_edge_length: Optional[float]) -> int:
...
+    def _tangent_basis_from_normals(normals: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+        """Construct orthonormal tangential bases for each normal vector."""
+
+        dtype = normals.dtype
+        basis1 = np.zeros_like(normals)
+        basis2 = np.zeros_like(normals)
+        ref_vectors = np.array([[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]], dtype=dtype)
+        tol = np.finfo(dtype).eps
+
+        for idx, normal in enumerate(normals):
```
yaugenst-flex: The per-vertex loop here could get pretty slow for high mesh resolutions. Might consider a vectorized basis construction here.

yaugenst-flex (changes requested):
Overall LGTM, everything makes sense! Spotted a couple things to iron out.

tidy3d/components/geometry/primitives.py
marcorudolphflex: mh I do not see a case where we directly set both non-None - but sure does not make sense to make target_edge_length always non-none and then reject calls with non-none subdivisions.

tidy3d/components/geometry/primitives.py
```diff
@@ -225,6 +259,25 @@ def intersections_plane(
             return []
         return [shapely.Point(x0, y0).buffer(0.5 * intersect_dist, quad_segs=quad_segs)]
 
+    def _discretization_wavelength(self, derivative_info: DerivativeInfo) -> float:
```
groberts-flex: possibly since we have one of these in cylinder as well, there is a way to combine them together into a general function for choosing the discretization wavelength and issuing the warning?

tidy3d/components/geometry/primitives.py
```diff
@@ -278,14 +331,226 @@ def unit_sphere_triangles(
...
+                axis_idx=axis_idx,
+                plane_value=plane_value,
+            )
+
+        trimesh_obj = TriangleMesh._triangles_to_trimesh(triangles)
+        vertices = np.asarray(trimesh_obj.vertices, dtype=grid_cfg.gradient_dtype_float)
+        center = np.asarray(self.center, dtype=grid_cfg.gradient_dtype_float)
+        verts_centered = vertices - center
+        norms = np.linalg.norm(verts_centered, axis=1, keepdims=True)
+        norms = np.where(norms == 0, 1, norms)
```
groberts-flex, line 385: is this catching the case where you have a vertex point right at the sphere center?

tidy3d/components/geometry/primitives.py
```diff
@@ -278,14 +331,226 @@ def unit_sphere_triangles(
...
+        normals = verts_centered / norms
+
+        if vertices.size == 0:
+            return dict.fromkeys(derivative_info.paths, 0.0)
+
+        # get vertex weights
+        faces = np.asarray(trimesh_obj.faces, dtype=int)
+        face_areas = np.asarray(trimesh_obj.area_faces, dtype=grid_cfg.gradient_dtype_float)
+        weights = np.zeros(len(vertices), dtype=grid_cfg.gradient_dtype_float)
+        np.add.at(weights, faces[:, 0], face_areas / 3.0)
```
groberts-flex, line 395: this is my unfamiliarity with how the trimesh works, but checking my understanding that each face names three vertices that make up the face?

groberts-flex: are these weights connecting gradients between the faces and the vertices? would it be possible to identify the face center and normal directly and then just multiply by the face_area?

marcorudolphflex: Yup, each face has 3 vertices, forming a triangle. This weight calculation is about calculating the "area" of vertices which is done via face areas.
Not sure what you are proposing... You mean just take the face centers as samples and having the weight directly by their area? Should be more efficient, right?

tidy3d/components/geometry/primitives.py
marcorudolphflex: Ah, I guess not - good point

marcorudolphflex: does this case make sense at all? The area and thus the weights would be zero then. Take an epsilon radius in this case?

marcorudolphflex: we also might just raise an exception here?

tidy3d/components/geometry/primitives.py
marcorudolphflex: Thinking about it again, I think this face-center approach is less precise compared to the sampling using vertices from a subdivided icosphere (which is locally approx. planar). Planes can be most efficiently sampled with a hexagonal pattern (see https://en.wikipedia.org/wiki/Circle_packing#Densest_packing) which we approx. archieve locally using the vertices - using face centers would sample the center between the circles which should be less uniform and inefficient.

marcorudolphflex: Another question is how to define efficiency here. Given a icosphere, the face-centered approach is more efficient in terms of sampling points as we have approx. 2x more faces than vertices. Still it is less uniform. Actually the better approach in that sense would be taking face centers AND vertices which would raise the question of weighting (cut faces in quarters)? But I guess this is overthinking at that point :D do we want to stick to the original approach?

tidy3d/components/geometry/primitives.py
groberts-flex: yeah good point! the area would be negligibly small I think. I would suspect this case would only happen for a small or zero radius with respect to the mesh. I could see something like this happening in an optimization if the gradient is saying to reduce the sphere size. do you think a one-time warning would make sense here?

tidy3d/components/geometry/primitives.py
groberts-flex: thanks for the analysis on this one! I think the original approach is good especially since it seems to be lining up well numerically. so is my understanding correct that the vertices are more uniformly sampling points on the actual sphere and so they end up being more accurate in terms of gradient calculation?

tidy3d/components/geometry/primitives.py
```diff
@@ -278,14 +331,226 @@ def unit_sphere_triangles(
...
+
+        derivative_info_cyl = derivative_info.updated_copy(
+            paths=list(cyl_paths),
+            bounds=bounds,
+            bounds_intersect=(intersect_min, intersect_max),
+        )
+
+        vjps_cyl = cylinder._compute_derivatives(derivative_info_cyl)
+        result = dict.fromkeys(derivative_info.paths, 0.0)
+        vjp_radius = float(vjps_cyl.get(("radius",), 0.0)) if need_radius else 0.0
```
groberts-flex, line 508: curious why we need the need_radius here? will "radius" just not be in the derivative_info.paths if it isn't needed?

tidy3d/components/geometry/primitives.py
```diff
@@ -278,14 +331,226 @@ def unit_sphere_triangles(
...
+            if path == ("radius",):
+                result[path] = vjp_radius * (radius / radius_plane)
+            elif path == ("center", axis_idx):
+                result[path] = vjp_radius * (delta / radius_plane)
+            elif path[0] == "center" and path[1] != axis_idx:
+                result[path] = float(vjps_cyl.get(("center", path[1]), 0.0))
+
+        return result
+
+    def _edge_length_on_unit_sphere(self, max_edge_length: Optional[float]) -> Optional[float]:
```
groberts-flex: should we have a default value for max_edge length if it is optional?

schemas/TerminalComponentModeler.json
```diff
@@ -16456,8 +16456,15 @@
           ]
         },
         "radius": {
-          "minimum": 0,
-          "type": "number"
+          "anyOf": [
```
groberts-flex, line 16501: I ran into a similar thing here in my custom vjp PR. Talking with @yaugenst-flex , I think they are going to open up schema breaking changes again Jan 5th. So the PR might need to wait to get merged until then

groberts-flex (approved):
looking good @marcorudolphflex!

tidy3d/components/geometry/primitives.py
marcorudolphflex: this is because the cylinder radius can effectively change with the sphere center (on the sim-orthogonal axis) which causes a different intersection circle/cylinder with the simulation plane

tidy3d/components/geometry/primitives.py
marcorudolphflex: added a warning and zero grad return

schemas/TerminalComponentModeler.json
marcorudolphflex: that's fine :)

yaugenst-flex (approved):

PR #3080, opened 2025-12-12
feat(tidy3d): FXC-3296-autograd-support-for-anisotropic-medium-and-custom-anisotropic-medium

marcorudolphflex:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR adds autograd support for diagonal `AnisotropicMedium` and `CustomAnisotropicMedium`, enabling gradient-based inverse design with anisotropic materials. The implementation delegates derivative computation to individual component media (xx, yy, zz) with proper field projection to ensure only the corresponding axis contributes to each gradient.

Key changes:
- Implemented `_compute_derivatives` method in `AnisotropicMedium` that projects electric field maps to individual axes and delegates to component media
- Added helper methods `_project_field_map_to_axis` and `_component_derivative_info` for field filtering
- Included comprehensive unit tests with synthetic fields and numerical validation tests comparing adjoint gradients with finite differences
- Updated CHANGELOG to document the new feature and improved error handling for unsupported anisotropic configurations

The implementation correctly inherits to `CustomAnisotropicMedium` through the class hierarchy. Error handling raises `NotImplementedError` for unsupported paths (e.g., off-diagonal components), though explicit test coverage for this error case would strengthen the validation.

<h3>Confidence Score: 4/5</h3>

- This PR is safe to merge with minor recommendation for additional test coverage
- The implementation is technically sound with proper delegation pattern and field projection. Comprehensive numerical tests validate gradient correctness across multiple scenarios. Score reduced by 1 point due to missing explicit test for the error handling mentioned in CHANGELOG (NotImplementedError for unsupported anisotropic configurations)
- tests/test_components/autograd/test_autograd_anisotropic_medium.py could benefit from explicit error handling test

<h3>Important Files Changed</h3>

File Analysis

| Filename | Score | Overview |
|----------|-------|----------|
| CHANGELOG.md | 5/5 | Added documentation for new autograd support for diagonal anisotropic media and improved error handling for unsupported cases |
| tidy3d/components/medium.py | 4/5 | Implemented `_compute_derivatives` method for `AnisotropicMedium` that delegates to component media with proper field projection - minor test coverage gap for error cases |
| tests/test_components/autograd/test_autograd_anisotropic_medium.py | 5/5 | Added unit tests for gradient computation with synthetic fields, covering both `AnisotropicMedium` and `CustomAnisotropicMedium` |
[...]

PR #3113, opened 2025-12-22
fix(tidy3d): FXC-4641-fix-gradients-in-custom-medium

marcorudolphflex:
Fixed interpolation and boundary handling in gradient computation of custom medium.
Note that the new testfile [test_autograd_medium_numerical.py](https://github.com/flexcompute/tidy3d/compare/FXC-4641-fix-gradients-in-custom-medium?expand=1#diff-68a43822d122debf7337fca502f0103741f7d6a7d57a5deae8fd6a3c5778caae) will be extended to anisotropic medium for this PR https://github.com/flexcompute/tidy3d/pull/3080 which is why some test branches are already prepared.

<h3>Greptile Summary</h3>

This PR fixes critical bugs in gradient computation for `CustomMedium` by replacing the previous xarray-based interpolation approach with a custom transpose-accumulation implementation. The old implementation had issues with interpolation handling and boundary conditions that led to incorrect gradients.

**Key changes:**
- Added cached properties (`_permittivity_sorted`, `_conductivity_sorted`, `_eps_components_sorted`) to ensure spatial coordinates are sorted before gradient computation, preventing errors during interpolation
- Replaced `_derivative_field_cmp` with `_derivative_field_cmp_custom` that implements a custom transpose-based accumulation strategy instead of relying on xarray's `.interp()` method
- Implemented proper boundary handling via `searchsorted` to clip field data to intersection bounds before processing
- Added explicit volume element scaling based on field grid coordinates rather than parameter grid coordinates
- Implemented custom nearest and linear interpolation using `npo.add.at` for proper gradient accumulation (using regular numpy `npo` instead of autograd numpy `np` for operations that need in-place accumulation)
- Added comprehensive numerical validation tests comparing adjoint gradients to finite-difference gradients

**Technical improvements:**
- The new implementation correctly handles the transpose (adjoint) operation needed for gradient backpropagation by accumulating field values onto parameter grid points using weighted contributions
- Volume elements are now computed from field coordinates and applied before interpolation, fixing the previous normalization issues
- The approach works for both "nearest" and "linear" interpolation methods with proper validation

- This PR is safe to merge with minor suggestions for improvement
- The implementation is mathematically sound and includes comprehensive numerical validation tests. The refactor addresses real bugs in gradient computation. Score reduced from 5 to 4 due to one style suggestion regarding floating-point comparison that could improve robustness
- No files require special attention - the implementation is solid with good test coverage

| Filename | Overview |
|----------|----------|
| CHANGELOG.md | Added changelog entry documenting the interpolation bug fix for CustomMedium gradients |
| tests/test_components/autograd/numerical/test_autograd_medium_numerical.py | New comprehensive test file for numerical gradient validation using finite differences; includes parametrized tests for isotropic and custom media with flux/intensity objectives |
| tidy3d/components/medium.py | Major refactor of gradient computation: replaced xarray interpolation with custom implementation using transpose accumulation for proper boundary handling; added sorted data caching |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client as Autograd System
    participant CM as CustomMedium
    participant CDC as _compute_derivatives
    participant DFCC as _derivative_field_cmp_custom
    participant TIA as _transpose_interp_axis
    
    Client->>CM: Request gradients for permittivity/conductivity/eps_dataset
    CM->>CM: Access cached _permittivity_sorted, _conductivity_sorted, _eps_components_sorted
    CM->>CDC: _compute_derivatives(derivative_info)
    
    loop For each field_path
        CDC->>CDC: Get sorted spatial_data from cache
        
        alt permittivity or conductivity
            loop For each dimension (x, y, z)
                CDC->>DFCC: _derivative_field_cmp_custom(E_der_map, spatial_data, dim, freqs, bounds, component)
                DFCC->>DFCC: Extract field coordinates and values
                DFCC->>DFCC: Apply bounds filtering (searchsorted)
                DFCC->>DFCC: Compute volume element scaling (_axis_sizes)
                DFCC->>DFCC: Multiply values by scale
                
                loop For each axis (x, y, z)
                    DFCC->>TIA: _transpose_interp_axis(arr, field_axis, param_axis)
                    
                    alt method == "nearest"
                        TIA->>TIA: Compute midpoints between param coords
                        TIA->>TIA: Use searchsorted to find indices
                        TIA->>TIA: Accumulate using npo.add.at
                    else method == "linear"
                        TIA->>TIA: Find upper/lower indices via searchsorted
                        TIA->>TIA: Compute interpolation weights
                        TIA->>TIA: Accumulate weighted contributions using npo.add.at
                    end
                    
                    TIA-->>DFCC: Return interpolated array
                end
                
                DFCC->>DFCC: Apply component extraction (real/imag/sigma/complex)
                DFCC->>DFCC: Sum over frequencies
                DFCC-->>CDC: Return gradient array
            end
            CDC->>CDC: Sum contributions from all dimensions
        else eps_dataset component
            CDC->>DFCC: _derivative_field_cmp_custom for specific component
            DFCC-->>CDC: Return gradient array
        end
        
        CDC->>CDC: Store in vjps dictionary
    end
    
    CDC-->>Client: Return vjps (gradient dictionary)
```

### DIFF
--- a/tidy3d/components/medium.py
+++ b/tidy3d/components/medium.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import functools
 from abc import ABC, abstractmethod
 from math import isclose
-from typing import Any, Callable, Literal, Optional, Union
+from typing import Any, Callable, Literal, Optional, Union, get_args
 
 import autograd.numpy as np
 
@@ -14,6 +14,7 @@ import numpy as npo
 import pydantic.v1 as pd
 import xarray as xr
 from autograd.differential_operators import tensor_jacobian_product
+from numpy.typing import NDArray
 
 from tidy3d.components.autograd.utils import pack_complex_vec
 from tidy3d.components.material.tcad.heat import ThermalSpecType
@@ -107,6 +108,8 @@ LOSSY_METAL_DEFAULT_SAMPLING_FREQUENCY = 20
 LOSSY_METAL_SCALED_REAL_PART = 10.0
 LOSSY_METAL_DEFAULT_MAX_POLES = 5
 LOSSY_METAL_DEFAULT_TOLERANCE_RMS = 1e-3
+
+ALLOWED_INTERP_METHODS = get_args(InterpMethod)
 
 
 def ensure_freq_in_range(eps_model: Callable[[float], complex]) -> Callable[[float], complex]:
@@ -1912,6 +1915,29 @@ class CustomMedium(AbstractCustomMedium):
         return self._medium.is_spatially_uniform
 
     @cached_property
+    def _permittivity_sorted(self) -> SpatialDataArray | None:
+        """Cached copy of permittivity sorted along spatial axes."""
+        if self.permittivity is None:
+            return None
+        return self.permittivity._spatially_sorted
+
+    @cached_property
+    def _conductivity_sorted(self) -> SpatialDataArray | None:
+        """Cached copy of conductivity sorted along spatial axes."""
+        if self.conductivity is None:
+            return None
+        return self.conductivity._spatially_sorted
+
+    @cached_property
+    def _eps_components_sorted(self) -> dict[str, ScalarFieldDataArray]:
+        """Cached copies of dataset components sorted along spatial axes."""
+        if self.eps_dataset is None:
+            return {}
+        return {
+            key: comp._spatially_sorted for key, comp in self.eps_dataset.field_components.items()
+        }
+
+    @cached_property
     def freqs(self) -> np.ndarray:
         """float array of frequencies.
         This field is to be deprecated in v3.0.
@@ -2320,37 +2346,49 @@ class CustomMedium(AbstractCustomMedium):
 
         for field_path in derivative_info.paths:
             if field_path[0] == "permittivity":
+                spatial_data = self._permittivity_sorted
+                if spatial_data is None:
+                    continue
                 vjp_array = 0.0
                 for dim in "xyz":
-                    vjp_array += self._derivative_field_cmp(
+                    vjp_array += self._derivative_field_cmp_custom(
                         E_der_map=derivative_info.E_der_map,
-                        spatial_data=self.permittivity,
+                        spatial_data=spatial_data,
                         dim=dim,
                         freqs=derivative_info.frequencies,
+                        bounds=derivative_info.bounds_intersect,
                         component="real",
                     )
                 vjps[field_path] = vjp_array
 
             elif field_path[0] == "conductivity":
+                spatial_data = self._conductivity_sorted
+                if spatial_data is None:
+                    continue
                 vjp_array = 0.0
                 for dim in "xyz":
-                    vjp_array += self._derivative_field_cmp(
+                    vjp_array += self._derivative_field_cmp_custom(
                         E_der_map=derivative_info.E_der_map,
-                        spatial_data=self.conductivity,
+                        spatial_data=spatial_data,
                         dim=dim,
                         freqs=derivative_info.frequencies,
+                        bounds=derivative_info.bounds_intersect,
                         component="sigma",
                     )
                 vjps[field_path] = vjp_array
 
             elif field_path[0] == "eps_dataset":
                 key = field_path[1]
+                spatial_data = self._eps_components_sorted.get(key)
+                if spatial_data is None:
+                    continue
                 dim = key[-1]
-                vjps[field_path] = self._derivative_field_cmp(
+                vjps[field_path] = self._derivative_field_cmp_custom(
                     E_der_map=derivative_info.E_der_map,
-                    spatial_data=self.eps_dataset.field_components[key],
+                    spatial_data=spatial_data,
                     dim=dim,
                     freqs=derivative_info.frequencies,
+                    bounds=derivative_info.bounds_intersect,
                     component="complex",
                 )
             else:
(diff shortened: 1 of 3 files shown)


marcorudolphflex:
@greptile

tidy3d/components/medium.py
```diff
@@ -2320,37 +2345,49 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
 
         for field_path in derivative_info.paths:
             if field_path[0] == "permittivity":
+                spatial_data = self._permittivity_sorted
```
groberts-flex, line 2349: do we have cases where the permittivity is not sorted? are these cases where it is specified from high to low spatial values?

tidy3d/components/medium.py
```diff
@@ -2360,96 +2397,132 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+        E_der_dim = E_der_map.get(f"E{dim}")
+        if E_der_dim is None or np.all(E_der_dim.values == 0):
+            return np.zeros(eps_shape, dtype=dtype_out)
+
+        field_coords = {axis: np.asarray(E_der_dim.coords[axis]) for axis in "xyz"}
+        values = E_der_dim.values
+
+        if bounds is not None:
+            (xmin, ymin, zmin), (xmax, ymax, zmax) = bounds
+            ix0 = np.searchsorted(field_coords["x"], xmin, side="left")
```
groberts-flex: nice! I didn't know about this numpy function, looks convenient

tidy3d/components/medium.py
```diff
@@ -2360,96 +2397,132 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+            iy1 = np.searchsorted(field_coords["y"], ymax, side="right")
+            iz0 = np.searchsorted(field_coords["z"], zmin, side="left")
+            iz1 = np.searchsorted(field_coords["z"], zmax, side="right")
+
+            field_coords = {
+                "x": field_coords["x"][ix0:ix1],
+                "y": field_coords["y"][iy0:iy1],
+                "z": field_coords["z"][iz0:iz1],
+            }
+            values = values[ix0:ix1, iy0:iy1, iz0:iz1, :]
```
groberts-flex: is it possible for coordinate axes to end of with zero size? if you have a single dimensional structure on one dimension for example or both bounds end up landing at the same index for an axis.

tidy3d/components/medium.py
```diff
@@ -2360,96 +2397,132 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+    def _derivative_field_cmp_custom(
         self,
         E_der_map: ElectromagneticFieldDataset,
-        spatial_data: CustomSpatialDataTypeAnnotated,
+        spatial_data: SpatialDataArray,
         dim: str,
         freqs: np.ndarray,
+        bounds: Optional[Bound] = None,
         component: str = "real",
+        interp_method: Optional[InterpMethod] = None,
```
groberts-flex, line 2409: the interp method is specified as part of the class right? is there a reason to allow a different to be specified here?

tidy3d/components/medium.py
```diff
@@ -2360,96 +2397,132 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
+        values = _interp_axis(values, 0, field_coords["x"], param_coords["x"])
+        values = _interp_axis(values, 1, field_coords["y"], param_coords["y"])
+        values = _interp_axis(values, 2, field_coords["z"], param_coords["z"])
+
+        freqs_da = np.asarray(E_der_dim.coords["f"])
+        if component == "sigma":
+            values = values.imag * (-1.0 / (2.0 * np.pi * freqs_da * EPSILON_0))
+        elif component == "imag":
+            values = values.imag
+        elif component != "complex":
```
groberts-flex: might be clearer here to check for 
`component == "real"`

tidy3d/components/medium.py
```diff
@@ -2360,96 +2397,132 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
-            )
-        vjp_array = E_der_dim_interp.values
-        vjp_array = vjp_array.reshape(eps_coordinate_shape)
+            # Flatten to 2D: (n_field, n_rest)
+            n_field = arr.shape[0]
+            a2 = arr.reshape(n_field, -1)
+
+            if method == "nearest":
+                midpoints = (param_sorted[1:] + param_sorted[:-1]) / 2.0
+                indices = np.searchsorted(midpoints, field_axis)
```
groberts-flex: does this ensure we get to the nearest index in the array? I think searchsorted is enforcing an ordering, so it may not pick the closest index for each midpoint.

tidy3d/components/medium.py
```diff
@@ -2360,96 +2397,132 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
...
-            d_vol = functools.reduce(np.outer, sizes_list)
-        else:
-            # if sizes_list is empty, then reduce() fails
-            d_vol = np.array(1.0)
+        method = interp_method or self.interp_method
 
-        E_der_dim_interp_complex = E_der_dim_interp.interp(
-            **coords_interp, assume_sorted=True
-        ).fillna(0.0)
+        def _transpose_interp_axis(arr, field_axis, param_axis):
```
groberts-flex: in this function, I think it would be great to have a little more documentation of what's going on at each step and some more descriptive variable names. It's a great catch that we need this and I want to make sure it's easy for people to understand in the future and modify if needed!

groberts-flex (commented):
thanks @marcorudolphflex this is great! overall, looks good and thanks for adding some good tests! I left a few comments on it, excited to have this change!

tidy3d/components/medium.py
marcorudolphflex: Yes it does, as we compare to the next midpoint indices, which are the "thresholds" between the values when we have to choose the nearest neighbor. Just made a small test with dummy data to confirm that.

tidy3d/components/medium.py
marcorudolphflex: totally right in this case. But we might have component-individual `interp_method` for `CustomAnisotropicMedium` in the future where we want to overload that. Passing `None` will default to the `self.interp_method`

tidy3d/components/medium.py
marcorudolphflex: good catch! Technically that would be possible if min and max of sim_bounds are between two neighboring field coordinates - which should not happen in practice? Still added a fallback for that with a warning.

tidy3d/components/medium.py
marcorudolphflex: Absolutely! Did a makeover.

tidy3d/components/medium.py
marcorudolphflex: right now the user might define it unsorted and we do not catch that...
```
import numpy as np
from tidy3d import SpatialDataArray, CustomMedium
dim_coords = np.array([-1.0, 1.0, 0.0])
coords = dict(x=dim_coords, y=dim_coords, z=dim_coords)
m = CustomMedium(permittivity=SpatialDataArray(np.ones((3, 3, 3)), coords=coords))
print("x field:", m.permittivity.x.values) # x field: [-1.  1.  0.]
```

marcorudolphflex:
> thanks @marcorudolphflex this is great! overall, looks good and thanks for adding some good tests! I left a few comments on it, excited to have this change!

Thanks for the careful review! Hope it is now more clear overall.

tidy3d/components/medium.py
groberts-flex: aww I see! good catch

groberts-flex (approved):
thanks for the changes @marcorudolphflex, looks good to me!

PR #3123, opened 2026-01-05
fix: avoid scale-dependent zero gradient filtering

yaugenst-flex:
<h3>Greptile Summary</h3>

This PR fixes a critical scale-dependent bug in adjoint gradient computation where `np.allclose()` with default tolerance (`atol=1e-8`) was incorrectly filtering out small but valid gradients. The fix replaces tolerance-based comparisons with exact zero checks using `np.count_nonzero() == 0`.

**Key Changes:**
- In `backward.py`: replaced `np.allclose(v, 0)` with exact zero check to prevent valid small gradients from being filtered
- In `optimizer.py`: changed from warning with `anp.allclose(grad, 0.0)` to error with exact zero check using `np.count_nonzero()`
- Added comprehensive test `test_scaled_objective_grad_not_filtered()` that verifies scaled objectives (e.g., `1e-10 * value`) work correctly
- Updated existing test to expect `SetupError` instead of warning

**Impact:**
This fix allows optimization with scaled objective functions where gradients may be legitimately small (e.g., 1e-10) without being incorrectly treated as zero. The change from warning to error for true zero gradients also improves user feedback for genuine setup issues.

- This PR is safe to merge with no identified issues
- The fix correctly addresses a well-defined bug with appropriate tests. The change from tolerance-based to exact zero checking is the right approach for gradient filtering. The new test case specifically validates the bug fix scenario. All changes are logically consistent and improve the robustness of the gradient computation

| Filename | Overview |
|----------|----------|
| tidy3d/web/api/autograd/backward.py | Fixed gradient filtering to check for exact zeros instead of using scale-dependent `np.allclose`, properly handling small but non-zero gradients |
| tidy3d/plugins/invdes/optimizer.py | Changed zero gradient check from warning to error with exact zero checking, ensuring true setup issues are caught while allowing scaled objectives |
| tests/test_plugins/test_invdes.py | Updated test to expect `SetupError` instead of warning, added comprehensive test for scaled objective gradients not being filtered |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Optimizer as AbstractOptimizer
    participant Autograd as Autograd (val_and_grad)
    participant Backward as backward.setup_adj()
    participant Filter as Gradient Filter
    
    Note over Optimizer: User calls optimizer.run()
    Optimizer->>Autograd: val_and_grad_fn(params)
    Autograd->>Backward: Compute VJP with data_fields_vjp
    
    Note over Backward: OLD: np.allclose(v, 0, atol=1e-8)
    Note over Backward: NEW: np.count_nonzero(v) == 0
    
    Backward->>Filter: Filter gradient fields
    alt Gradient has small values (e.g., 1e-10)
        Filter-->>Backward: OLD: Filtered out (treated as zero)
        Filter-->>Backward: NEW: Kept (non-zero)
    else Gradient is exactly zero
        Filter-->>Backward: Both: Filtered out
    end
    
    Backward->>Autograd: Return adjoint simulations (or empty list)
    Autograd->>Optimizer: Return (val, grad)
    
    Note over Optimizer: OLD: anp.allclose(grad, 0.0)
    Note over Optimizer: NEW: np.count_nonzero(grad) == 0
    
    alt All gradient elements exactly zero
        Optimizer->>Optimizer: Raise SetupError
    else Gradient has non-zero elements
        Optimizer->>Optimizer: Continue optimization
    end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed `AutoImpedanceSpec` validation to check path intersections against all conductors, not just filtered ones, as well as the mode plane bounds.
+- Fixed adjoint gradients being treated as zero due to scale-dependent `np.allclose(..., atol=1e-8)` checks, which could skip adjoint simulations and return zero gradients.
 
 ## [2.10.0] - 2025-12-18
 
--- a/tests/test_plugins/test_invdes.py
+++ b/tests/test_plugins/test_invdes.py
@@ -8,6 +8,7 @@ import pytest
 
 import tidy3d as td
 import tidy3d.plugins.invdes as tdi
+from tidy3d.exceptions import SetupError
 from tidy3d.plugins.expressions import ModeAmp, ModePower
 from tidy3d.plugins.invdes.initialization import (
     CustomInitializationSpec,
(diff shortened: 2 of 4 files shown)


yaugenst-flex:
@greptileai

groberts-flex (approved):
thanks @yaugenst-flex looks good!

momchil-flex (approved):

momchil-flex:
I approved, but just had a second thought about this wondering why this check was added in the first place and if we are still addressing the original reason for it.

Like you mentioned in the chat, the most reasonable thing might be to compare to the OF magnitude or something like that?

yaugenst-flex:
> I approved, but just had a second thought about this wondering why this check was added in the first place and if we are still addressing the original reason for it.
> 
> Like you mentioned in the chat, the most reasonable thing might be to compare to the OF magnitude or something like that?

Yeah so the check was added because zero or very close to zero gradients usually point to a problem with the setup. However after thinking about this some more, I think only true zeros are actually safe to strictly filter out. Comparing to OF magnitude could work, but this information is not available in the generic adjoint pipeline. And while it is available in invdes, it's not _necessarily_ wrong to have small gradient magnitudes, compared to the objective function value, it could also just point to a converged optimization. The only other generic thing I can think of right now is that we skip the adjoint simulation and warn the user if the adjoint source magnitude drops below some value where simulating doesn't really make sense. Thoughts?

momchil-flex:
> > I approved, but just had a second thought about this wondering why this check was added in the first place and if we are still addressing the original reason for it.
> > Like you mentioned in the chat, the most reasonable thing might be to compare to the OF magnitude or something like that?
> 
> Yeah so the check was added because zero or very close to zero gradients usually point to a problem with the setup. However after thinking about this some more, I think only true zeros are actually safe to strictly filter out. Comparing to OF magnitude could work, but this information is not available in the generic adjoint pipeline. And while it is available in invdes, it's not _necessarily_ wrong to have small gradient magnitudes, compared to the objective function value, it could also just point to a converged optimization. The only other generic thing I can think of right now is that we skip the adjoint simulation and warn the user if the adjoint source magnitude drops below some value where simulating doesn't really make sense. Thoughts?

Yeah, I think all these are valid points. I guess let's go with strict 0 and see if anything else pops up in the future. I agree it's hard to come up with something precise, I mean the adjoint source magnitude will also be small if the scale of the OF is small right, but this might not be a bad setup, apart from potentially underflow issues. I do think that there are various dangers lurking if the OF is too poorly scaled but yeah we can't really warn about that in a gradient computation right?

yaugenst-flex:
> we can't really warn about that in a gradient computation right?

More likely it'd have to be a bunch of heuristics that take into account also the resulting changes in the parameters because it also depends on the interaction with the optimizer...

PR #3132, opened 2026-01-07
fix: pydanticv2 speed improvements

mahlau-flex:
fix hashing and double validation issues, which gives a speedup

---

> [!NOTE]
> Improves Pydantic v2 performance and logging by eliminating double-validation and capturing warnings hierarchically during model construction.
> 
> - Integrates capture via `@model_validator`: add wrap validator to `begin_capture/abort` on errors and inject an `after` validator in subclasses to `end_capture`; remove old `__init__` capture path
> - Adds `log.abort_capture()` and clarifies capture semantics; supports explicit capture regions (e.g., `validate_pre_upload`)
> - Rewrites `__hash__` with recursive hashing for `numpy` arrays, `xarray` objects, lists, and stable digest for mutable `attrs`
> - Refactors simulation validation: drops `_validate_scene` aggregator; converts `_validate_internal_abc_no_fully_anisotropic` to `@model_validator(mode="after")` and fixes call to `Scene.intersecting_media(..., tuple(...))`
> - Tests: add hierarchical warning-capture test for nested models and update end-to-end capture checks
> - Minor: add type-ignore for abstract `Metric` filtering
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 831cd1f54d9ec30d3fb7628b76186687df126257. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

<h3>Greptile Summary</h3>

This PR optimizes Pydantic v2 performance by removing redundant validation and fixing type handling issues. The changes address hashing of mutable attributes, eliminate double validation via consolidating model validators, and correct type signatures.

**Key Changes:**
- **base.py**: Removed `__init__` method (no longer needed in Pydantic v2) and rewrote `__hash__` to explicitly handle mutable attributes (`attrs`), numpy arrays, xarray objects, and lists. Potential issue identified: list hashing assumes all items are hashable and may fail on nested collections.
- **simulation.py**: Removed redundant `_validate_scene` aggregator method and converted `_validate_internal_abc_no_fully_anisotropic` to a standalone `@model_validator`. Fixed type signature by converting list to tuple when calling `Scene.intersecting_media(abc, tuple(...))`.
- **metrics.py**: Added type checker suppression comment for abstract base class usage.

**Testing Notes:** Changes align with Pydantic v2 validator patterns. All validation methods remain independent and will execute during model construction.

- PR contains a potential hash function issue that could cause runtime failures with nested collections in fields.
- The PR makes valid Pydantic v2 optimizations and the core changes are sound. However, the new `__hash__` implementation (line 244-246 in base.py) has a logical flaw: it assumes all list items are hashable but does not handle nested collections (lists, dicts) that appear in some Tidy3d component fields. This could cause silent hash failures or TypeErrors at runtime. The validation refactoring is correct and all validators remain active through independent `@model_validator` decorators. The metrics.py change is trivial and safe.
- tidy3d/components/base.py - The hash function needs defensive handling for unhashable items in lists, particularly for complex nested types used in Tidy3d components.

| Filename | Overview |
|----------|----------|
| tidy3d/components/base.py | Hash implementation refactored to handle mutable attributes and special types (numpy arrays, xarray objects). Removed __init__ method (Pydantic v2 no longer requires it). Potential issue: list items in line 246 are hashed individually but if any list contains unhashable items (nested lists, dicts), hashing will fail despite the comment's assumption. |
| tidy3d/components/simulation.py | Removed _validate_scene method and split its responsibilities into individual @model_validator methods. Fixed type signature by converting list to tuple when calling Scene.intersecting_media. All validation methods now independently decorated with @model_validator(mode='after'). |
| tidy3d/plugins/expressions/metrics.py | Added type: ignore[type-abstract] comment to suppress type checker warning when passing abstract Metric class to filter method. Minimal, non-functional change. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Pydantic
    participant OldValidation as Old: _validate_scene
    participant NewValidators as New: @model_validator methods
    
    User->>Pydantic: Create Simulation()
    
    alt Old Approach (Removed)
        Pydantic->>OldValidation: _validate_scene()
        OldValidation->>OldValidation: _validate_no_structures_pml()
        OldValidation->>OldValidation: _validate_tfsf_nonuniform_grid()
        OldValidation->>OldValidation: _validate_tfsf_aux_sources()
        OldValidation->>OldValidation: _validate_nonlinear_specs()
        OldValidation->>OldValidation: _validate_custom_source_time()
        OldValidation->>OldValidation: _validate_mode_objects()
        OldValidation->>OldValidation: _warn_rf_license()
        OldValidation->>OldValidation: _validate_internal_abc_no_fully_anisotropic()
    end
    
    alt New Approach (Current)
        Pydantic->>NewValidators: @model_validator(mode="after")
        par
            NewValidators->>NewValidators: _validate_no_structures_pml()
        and
            NewValidators->>NewValidators: _validate_tfsf_nonuniform_grid()
        and
            NewValidators->>NewValidators: _validate_internal_abc_no_fully_anisotropic()
        and
            NewValidators->>NewValidators: (other validators)
        end
    end
    
    Pydantic->>User: Return Simulation instance
```

### DIFF
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -201,12 +201,6 @@ class Tidy3dBaseModel(BaseModel):
                 )
         return name
 
-    def __init__(self, **kwargs: Any) -> None:
-        """Init method, includes post-init validators."""
-        log.begin_capture()
-        super().__init__(**kwargs)
-        log.end_capture(self)
-
     def __init_subclass__(cls: type[T], **kwargs: Any) -> None:
         """Injects a constant discriminator field before Pydantic builds the model.
 
@@ -235,10 +229,33 @@ class Tidy3dBaseModel(BaseModel):
 
     def __hash__(self) -> int:
         """Hash method."""
-        try:
-            return super().__hash__(self)
-        except TypeError:
-            return hash(self.model_dump_json())
+        # This function needs to take special care because of mutable attributes inside of frozen pydantic models
+        to_hash_list = []
+        for k, v in dict(self).items():
+            if k == "attrs":
+                continue
+            if isinstance(v, np.ndarray):
+                # numpy arrays are not hashable by default, use byte representation
+                v_hash = hashlib.md5(v.tobytes()).hexdigest()
+            elif isinstance(v, (xr.DataArray, xr.Dataset)):
+                # we choose to not hash data arrays as this would require a lot of careful handling of units, metadata.
+                # technically this is incorrect, but should never lead to bugs in current implementation
+                v_hash = str(v.__class__.__name__)
+            elif isinstance(v, list):
+                # this assumes all objects in lists are hashable by default and do not require special handling
+                v_hash = tuple([hash(vi) for vi in v])
+            else:
+                v_hash = hash(v)
+            to_hash_list.append((k, v_hash))
+
+        # attrs is mutable, use serialized output as safe hashing option
+        if self.attrs:
+            attrs_json = self.model_dump_json(include={"attrs"})
+            attrs_hash = hash(attrs_json)
+            to_hash_list.append(("attrs", attrs_hash))
+
+        result_hash = hash(tuple(to_hash_list))
+        return result_hash
 
     def _hash_self(self) -> str:
         """Hash this component with ``hashlib`` in a way that is the same every session."""
--- a/tidy3d/components/base.py
+++ b/tidy3d/components/base.py
@@ -262,28 +262,33 @@ class Tidy3dBaseModel(BaseModel):
         for k, v in dict(self).items():
             if k == "attrs":
                 continue
-            if isinstance(v, np.ndarray):
-                # numpy arrays are not hashable by default, use byte representation
-                v_hash = hashlib.md5(v.tobytes()).hexdigest()
-            elif isinstance(v, (xr.DataArray, xr.Dataset)):
-                # we choose to not hash data arrays as this would require a lot of careful handling of units, metadata.
-                # technically this is incorrect, but should never lead to bugs in current implementation
-                v_hash = str(v.__class__.__name__)
-            elif isinstance(v, list):
-                # this assumes all objects in lists are hashable by default and do not require special handling
-                v_hash = tuple([hash(vi) for vi in v])
-            else:
-                v_hash = hash(v)
+            v_hash = self._recursive_hash(v)
             to_hash_list.append((k, v_hash))
 
         # attrs is mutable, use serialized output as safe hashing option
         if self.attrs:
-            attrs_json = self.model_dump_json(include={"attrs"})
-            attrs_hash = hash(attrs_json)
+            attrs_str = self._attrs_digest()
+            attrs_hash = hash(attrs_str)
             to_hash_list.append(("attrs", attrs_hash))
 
         result_hash = hash(tuple(to_hash_list))
         return result_hash
+
+    @staticmethod
+    def _recursive_hash(value: Any) -> int:
+        if isinstance(value, np.ndarray):
+            # numpy arrays are not hashable by default, use byte representation
+            v_hash = hashlib.md5(value.tobytes()).hexdigest()
+            return hash(v_hash)
+        if isinstance(value, (xr.DataArray, xr.Dataset)):
+            # we choose to not hash data arrays as this would require a lot of careful handling of units, metadata.
+            # technically this is incorrect, but should never lead to bugs in current implementation
+            return hash(str(value.__class__.__name__))
+        if isinstance(value, list):
+            # this assumes all objects in lists are hashable by default and do not require special handling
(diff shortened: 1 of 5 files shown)


tidy3d/components/base.py
```diff
@@ -235,10 +229,33 @@ def __pydantic_init_subclass__(cls: type[T], **kwargs: Any) -> None:
...
+            elif isinstance(v, list):
+                # this assumes all objects in lists are hashable by default and do not require special handling
+                v_hash = tuple([hash(vi) for vi in v])
+            else:
+                v_hash = hash(v)
+            to_hash_list.append((k, v_hash))
+
+        # attrs is mutable, use serialized output as safe hashing option
+        if self.attrs:
+            attrs_json = self.model_dump_json(include={"attrs"})
```
marcorudolphflex: Wouldn't it make sense to unify the `attr` part here with `_attrs_digest` which serves as a guard for `_json_string`?
...
while typing this I just discovered that we don't really use `_json_string` anymore (only in `material_library`, also use `model_dump_json` here?)
Still I'm wondering why we don't need `make_json_compatible` here anymore, I guess the fixes from there are now natively supported by v2?
Could it make sense to cache this hash? Do you know how often it is usually called by pydantic? But this should be guarded by `attr` again (see below).

------ not related to this function/PR, but to the related mentions by me

I'm wondering if we have still any kind of caching in `_hash_self` (as we had it before by caching `_json_string`)? Note that this still should be guarded by `attr` changes as in develop.

tidy3d/components/base.py
```diff
@@ -201,12 +201,6 @@ def _validate_name_no_special_characters(cls: type[T], name: Optional[str]) -> O
                 )
         return name
 
-    def __init__(self, **kwargs: Any) -> None:
```
marcorudolphflex, line 204: what was the idea behind this log capturing and why do we discard that here?

tidy3d/components/base.py
```diff
@@ -235,10 +229,33 @@ def __pydantic_init_subclass__(cls: type[T], **kwargs: Any) -> None:
...
+            elif isinstance(v, (xr.DataArray, xr.Dataset)):
+                # we choose to not hash data arrays as this would require a lot of careful handling of units, metadata.
+                # technically this is incorrect, but should never lead to bugs in current implementation
+                v_hash = str(v.__class__.__name__)
+            elif isinstance(v, list):
+                # this assumes all objects in lists are hashable by default and do not require special handling
+                v_hash = tuple([hash(vi) for vi in v])
+            else:
+                v_hash = hash(v)
+            to_hash_list.append((k, v_hash))
```
marcorudolphflex, line 266: do we want to have some fallback here if some of the functions above fail? E.g. `model_dump_json` as before

marcorudolphflex (commented):
LGTM overall, just a couple of questions/nits and things that may be worth to investigate. And some side notes on hashing/caching.

tidy3d/components/base.py
mahlau-flex: This was necessary in pydanticV1, but in v2 this would lead to double checking. I did not do a thorough test yet if the logs are still captured correctly in all cases

tidy3d/components/base.py
mahlau-flex: problem is that model_dump_json is very slow and hashing is used quite a lot in v2. We could add a try-except around everything and do model-dump in case of failure, but this might obscure also errors. thoughts?

tidy3d/components/base.py
mahlau-flex: I made a separate issue for the logging, because I think this might be something that requires a bit more thought and consolidation:
https://flow360.atlassian.net/jira/software/c/projects/FXC/boards/1054?selectedIssue=FXC-4706

tidy3d/components/base.py
mahlau-flex: I agree that this PR should re-use _attrs_digest, good point!

about the other point, I am not sure I understand what you mean

tidy3d/components/base.py
marcorudolphflex: Also not sure about that. From my side, we can also go with the risky variant. Tests should show violations and users may adapt to it when flooding `attrs` with strange types.

tidy3d/components/base.py
yaugenst-flex: yeah this one seemed suspicious so i spent some time going over it in detail but it does seem like we don't need the log capture on init

yaugenst-flex: aah it does have some implications for the backend

yaugenst-flex (approved):
Let's try it!

PR #3177, opened 2026-01-15
shape gradient cleanup and consolidation

groberts-flex:
In this PR:
1. the inside/outside permittivity handling is consolidated to all use the eps_inf_structure and eps_no_structure treatment for eps_in and eps_out respectively. This includes a couple of fixes to the treatment that are especially relevant to Box geometries (which consequently fix the autograd regression we saw with the TMM notebook): (a) the outside permittivity grabber is enlarged slightly to make sure we can sample outside permittivity for eps_no_structure and (b) the interpolation method is changed to 'nearest' for permittivity and coordinates are adjusted to make sure we are on the correct side of the boundary after this interpolation.
2. dielectric structures embedded in PEC materials or ones that are partially embedded in PEC are enabled via setting the background_medium to PECMedium() in the structure. In the future, it will be nice to automatically detect this, but for now I think this is a good intermediate solution. The main thing this helps with is that we only collect H fields in relevant cases (either a PEC structure or one where there is PEC around some part of the edges). 
3. three numerical tests are added: (a) test_flexible_dielectric_pec_boundary.py checks equivalence between gradients when choosing the same shared boundary on two different structures (this allows us to test both dielectric-dielectric and hybrid dielectric-PEC interfaces for gradients), (b) test_autograd_geometry_group_numerical.py ensures we get the same gradients for a group of structures when we use or don't use a GeometryGroup and we don't have overlapping structures, and (c) test_tmm_gradients.py captures the test from the TMM gradient checking notebook and augments it to also make sure it works when the boxes are replaced with equivalent polyslabs

Notes:
1. I wanted to get this posted before the end of the day so the review could get started and we could hopefully get this in for the patch release early next week. Tomorrow, I'll keep running more numerical tests to make sure they pass and then try out some autograd notebooks to make sure they still work as well.
2. I'll also do another pass myself through the code to check for places that are still in need of more documentation/explanation.

---

> [!NOTE]
> **Unifies shape-gradient permittivity handling and adds PEC embedding support**
> 
> - Consolidates inside/outside permittivity via `DerivativeInfo.eps_in/eps_out` (`ScalarFieldDataArray`), using `eps_no_structure` (outside) and infinite-structure (inside); switches permittivity/field interpolation to `nearest` where needed and snaps coordinates to the correct side of boundaries
> - Adds PEC-aware integration: detect PEC via `adjoint.pec_detection_threshold`; support dielectric structures partially/fully embedded in PEC via `Structure.background_medium=td.PECMedium()`; collect `H*` only when structure or background is PEC; blend PEC vs dielectric boundary contributions pointwise
> - Improves adjoint pipeline: enlarge `eps_no_structure` sampling window (`Grid.get_bounding_*`), provide `eps_in` for 2D/3D (including non-traceable custom media), and block mixed `CustomMedium` + traced geometry (`Simulation._check_custom_medium_geometry_overlap`)
> - New adjoint config: `adjoint.boundary_snapping_fraction`, `adjoint.pec_detection_threshold`; adaptive spacing now derived from `eps_data`
> - Geometry/grid updates: `PolySlab` normal/perp arrays use gradient dtype; new `Grid.get_bounding_indices/get_bounding_values`
> - Testing: add `test_flexible_dielectric_pec_boundary.py`, `test_autograd_geometry_group_numerical.py`, `test_tmm_gradients.py`; broaden numerical suites and add `redirect_stdout_to_stderr` fixture
> - **Fix**: enlarge outside-permittivity sampling and use nearest interpolation to resolve Box gradient regression
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d7387fa156d59b6801f03d724282716b5b5c4593. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

<h3>Greptile Summary</h3>

- Consolidates shape gradient permittivity handling to use `eps_inf_structure` and `eps_no_structure` for inside/outside permittivity values respectively, fixing Box geometry regression
- Enables PEC gradient computation for dielectric structures embedded in PEC materials through `Structure.background_medium` field
- Adds comprehensive numerical test suites for gradient validation including TMM comparison, geometry group consistency, and flexible PEC-dielectric boundary testing

| Filename | Overview |
|----------|----------|
| `tidy3d/components/medium.py` | Contains debug print statements and commented code that must be removed before merging |
| `tidy3d/components/autograd/derivative_utils.py` | Major refactor consolidating permittivity handling with PEC detection, dual interpolation modes, and coordinate snapping |
| `tidy3d/web/api/autograd/backward.py` | Updated to use consolidated epsilon arrays with enlarged sampling regions and improved 2D geometry handling |
| `tests/test_components/autograd/numerical/test_flexible_dielectric_pec_boundary.py` | New test for dielectric-PEC boundary gradient consistency with potential undefined constants |

<h3>Confidence score: 2/5</h3>

- This PR has critical issues that prevent safe merging due to debug print statements in production code
- Score significantly lowered due to temporary debugging code in `tidy3d/components/medium.py` that will pollute production output  
- Pay close attention to removing all debug/print statements and validating numerical test stability before merging

**Context used:**

- Rule from `dashboard` - Remove temporary debugging code (print() calls), commented-out code, and other workarounds before fi... ([source](https://app.greptile.com/review/custom-context?memory=f6a669d8-0060-4f11-9cac-10ac7ee749ea))


tidy3d/web/api/autograd/backward.py
```diff
@@ -304,16 +318,15 @@ def postprocess_adj(
                 H_fwd=H_fwd_chunk,
                 H_adj=H_adj_chunk,
                 eps_data=eps_data_chunk,
-                eps_in=eps_in_chunk,
-                eps_out=eps_out_chunk,
-                eps_background=eps_background_chunk,
+                eps_in=eps_inf_structure_chunk,
```
yaugenst-flex, line 321: Since `eps_inf_structure` is now unconditionally `None` for custom medium, this would break `_evaluate_dielectric_gradient_at_points` which assumes it to be a data array

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -784,29 +921,33 @@ def adaptive_vjp_spacing(
             if min_allowed_spacing_fraction is None:
                 min_allowed_spacing_fraction = config.adjoint.minimum_spacing_fraction
 
-        # handle FreqDataArray or scalar eps_in
-        if isinstance(self.eps_in, FreqDataArray):
+        def spacing_by_permittivity(eps_array):
```
yaugenst-flex, line 935: Another comment on this function: do we want exterior materials to influence sampling distance? It seems like we should?

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -375,24 +379,78 @@ def evaluate_gradient_at_points(
...
+        # dielectric-dielectric interface.
+        vjps_dielectric = self._evaluate_dielectric_gradient_at_points(
+            spatial_coords,
+            normals,
+            perps1,
+            perps2,
+            interpolators,
+            self.eps_in,
+            self.eps_out,
+        )
```
yaugenst-flex, line 390: Just to make sure, this is what we discussed right? That we now unconditionally calculate the dielectric path and accept the potential performance hit?

tidy3d/components/grid/grid.py
```diff
@@ -205,6 +205,45 @@ def _interp_from_unstructured(
 
         return interp_array
 
+    def get_bounding_indices(
+        self, coordinate: Coordinate, side: str, buffer: Optional[int] = 0
```
marcorudolphflex: just int instead of `Optional[int]`? see also `get_bounding_values`
`Literal["left", "right"]` for side

tidy3d/config/sections.py
```diff
@@ -145,6 +145,16 @@ class AdjointConfig(ConfigSection):
         ge=0.0,
     )
 
+    pec_detection_threshold: float = Field(
```
marcorudolphflex: `NonPositiveFloat`?

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -448,92 +517,157 @@ def _evaluate_dielectric_gradient_at_points(
...
 
-            # this edge distance is useful when correcting for edge singularities from the PEC material
-            # and is used when the PEC PolySlab structure has zero thickness
-            edge_distance = np.sqrt(edge_distance)
+        eps_adjusted_all = [
+            component_interpolator(adjusted_coords)
+            for _, component_interpolator in interpolator.items()
+        ]
+        eps_detect_pec = reduce(np.minimum, eps_adjusted_all)
```
marcorudolphflex, line 659: Physics question: np.minimum just operates on real part, does it still makes sense then?

tidy3d/components/autograd/derivative_utils.py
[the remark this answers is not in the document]
groberts-flex: great catch!

groberts-flex: yes, I think we do want them to influence sampling distance since it should be based on materials on either side of the boundary. this was a typo on my part to keep using self.eps_in. I've changed to use eps_array and below we call this function with both eps_in and eps_out and take the minimum of the two spacings.

tidy3d/web/api/autograd/backward.py
groberts-flex: That's a good point, it would error if we called that function for a custom medium. But do we have cases where a custom medium is also being used in a shape gradient? Maybe the best thing to do is catch the case where the medium is custom but there are tracers in the geometry since that's not really a supported case.

tidy3d/components/autograd/derivative_utils.py
groberts-flex: I think this is the safest thing to do, but might be covering a case that is more uncommon. 

- For dielectric structures with background_medium not PEC, this needs to happen regardless and there isn't a change in performance (this covers all of the non-RF autograd cases)
- For dielectric structures with background_medium PEC, this needs to happen if we are assuming the structure can be partially embedded in PEC and partially embedded in dielectric. I think this is a good case to cover.
- For PEC structures, this is helpful if the user also has a dielectric structure overlapping with the PEC one. This means there is a run of the border that is dielectric-dielectric (and should contribute zero gradient). Of the three scenarios, this would be the one where we are paying a little bit of a performance hit to make sure we cover a case that may not be highly common. I think it's safer to leave it in for now because it'll only affect RF cases and shouldn't slow down any of our current functionality and this logic can be improved if we can, in the future, get more metadata on the boundary type from the solver in something like a surface monitor.

tidy3d/components/grid/grid.py
groberts-flex: thank you!

tidy3d/components/autograd/derivative_utils.py
groberts-flex: good question! in this case the PEC material shows up as a having a very large (effectively infinite) negative permittivity value so we can detect it via the real part.  for other metallic detection, depending on the frequency range, we might want to instead look at the imaginary part to see the material loss.

groberts-flex:
Here is some additional context that might be helpful that I mean to post yesterday as well. Before coding, I spent some time trying to map out all of the derivative paths and make sure we have things covered under this new system. The main things we eliminate from before are (1) eps_approx which we were using inside geometry groups (I've tested we no longer need this  with the added numerical test), (2) separate eps_in from eps_inf_structure (we now just get permittivity for inside via eps_inf_structure strategy) and (3) separate eps_out from eps_no_structure (we now just get permittivity for outside via eps_no_structure strategy).

image

One note on this is it is showing LossyMetal paths too but these aren't in place yet (those will be added in the future when I revive the LossyMetal support PR)

groberts-flex:
Numerical test test_flexible_dielectric_pec_boundary.py

BEFORE CHANGE:

image


AFTER CHANGE:

image

groberts-flex:
Numerical test test_tmm_gradients.py:

BEFORE CHANGE:

image


AFTER CHANGE:

image

groberts-flex:
Numerical test: test_autograd_geometry_group_numerical.py

image

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -448,92 +515,157 @@ def _evaluate_dielectric_gradient_at_points(
...
-                with a grid center and are outside the PEC surface
-                (N,) array of distances from the nearest interpolation points to the desired surface
-                edge points specified by `spatial_coords`
+        Parameters
+        ----------
+        spatial_coords : np.ndarray
+            (N, 3) array of surface evaluation points.
+        normals : np.ndarray
+            (N, 3) array of outward-pointing normal vectors at each surface point.
+        normal_direction: float
```
tylerflex: should this just be a `is_outside: bool`?

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -448,92 +515,157 @@ def _evaluate_dielectric_gradient_at_points(
...
+        for idx, dim in enumerate("xyz"):
+            expanded_adjusted_coords = np.expand_dims(adjust_spatial_coords[:, idx], axis=1)
+            grid_centers_select = grid_centers[dim]
+
+            # find nearest grid point from the adjusted coordinates
+            diff = np.abs(expanded_adjusted_coords - grid_centers_select)
+            nearest_grid = np.argmin(diff, axis=-1)
+
+            # compute edge distance from the nearest interpolated point to the boundary edge
+            edge_distance += np.abs(spatial_coords[:, idx] - grid_centers_select[nearest_grid]) ** 2
```
tylerflex: feel free to ignore, minor nitpick / comment: I got slightly tripped up here because I thought it was missing a sqrt(), but now I see. might be better to call this `edge_distance_squared` or `edge_distance_squared_sum` and then `edge_distance = np.sqrt(edge_distance_squared)`

tidy3d/components/autograd/derivative_utils.py
```diff
@@ -580,10 +723,18 @@ def _interpolate_field_components(interp_coords, field_name):
         H_adj_coords_adjusted = _snap_coordinate_outside(self.H_adj)
 
         # using the adjusted coordinates, evaluate all field components at surface points
-        E_fwd_at_coords = _interpolate_field_components(E_fwd_coords_adjusted, "E_fwd")
-        E_adj_at_coords = _interpolate_field_components(E_adj_coords_adjusted, "E_adj")
-        H_fwd_at_coords = _interpolate_field_components(H_fwd_coords_adjusted, "H_fwd")
-        H_adj_at_coords = _interpolate_field_components(H_adj_coords_adjusted, "H_adj")
+        E_fwd_at_coords = _interpolate_field_components(E_fwd_coords_adjusted, "E_fwd_nearest")
```
tylerflex: call with keywords?

tylerflex (approved):
I noted a couple things as reading through the code that caught my eye, but overall the implementation seems sound. And the unification is great! Will be worth running the full numerical gradient tests if not already done. I wasn't able to follow all of the subtleties but if there's a specific area where you need me to explain some historical artifact let me know in the comments. Thanks @groberts-flex this is awesome 💪

tidy3d/components/autograd/derivative_utils.py
groberts-flex: good point, I think that's much more readable that way! changed to `edge_distance_squared_sum`

tidy3d/components/autograd/derivative_utils.py
yaugenst-flex: yeah sounds good, just wanted to make sure i understand

yaugenst-flex (approved):
awesome!

tidy3d/components/autograd/derivative_utils.py
groberts-flex: agreed! changed

tidy3d/components/autograd/derivative_utils.py
groberts-flex: yeah, I like this idea! much clearer what's going on with that. changed

PR #3183, opened 2026-01-19
FXC-4925: Remove BatchDetail and BatchTask in webapi

momchil-flex:
This is an attempt simplify the webapi components that were introduced in the first component modeler iteration. Specifically, I have removed BatchDetail and BatchTask in favor of everything still being TaskInfo and SimulationTask, respectively. One advantage is that now we just store which task types correspond to a "batch" and the `task._is_batch_type():` can be evaluated simply based on the stored `task.taskType`, rather than calling the detail API every time (previously, this call was cached when the TaskFactory was used, but it was still awkward and would create unnecessary API call repetitions in some cases). 

I think I can completely remove BatchTask and TaskFactory (no need to keep backward compatibility of internally used components I would say, especially those that were added very recently for the CM) - but before doing more I wanted to get your opinion @daquintero (maybe also fyi @yaugenst )

I would say this is still not in an ideal state, but looks better to me. Generally, I am not opposed to having the different subclasses, but this was a bit of a mess. For example, `WebTask.create` was always creating a `SimulationTask` instead of a `BatchTask`, and it was a bit hard to figure out what was needed where and why.

I think the main downside of this change is that the TaskInfo now has fields relevant to both regular simulations and to component modelers, so in some cases one part will be None while in other cases another part will be None. But seems OK to me.

> [!NOTE]
> Unifies task handling by folding batch/modeler APIs into existing simulation abstractions.
> 
> - Removes `BatchDetail` and `BatchTask`; introduces `BatchTask = SimulationTask` alias for backward compatibility
> - Extends `SimulationTask` to detect batch types (`_is_batch_type`) and route to `rf/task/...` endpoints for `detail`, `submit`, `abort`, and `check`
> - Simplifies `TaskFactory` to always return `SimulationTask`; `get()` now probes simulation API then falls back to batch API with proper 404 handling
> - Reworks `TaskInfo` into a unified schema including batch-specific fields; `get_info` now returns only `TaskInfo`
> - Updates webapi functions to use `_is_batch_type()` checks instead of `isinstance(BatchTask)`, adjust monitoring, status, download, load, and cost/real_cost flows
> - Improves delete flow to prefer group/version deletion when available; tests adjusted to mock new endpoints and fields (e.g., `groupId`, `version`) and remove deprecated paths
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c3b1eb463860ea94ff7532b8992e15fae5afb5ab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tidy3d/web/core/task_core.py
+++ b/tidy3d/web/core/task_core.py
@@ -830,6 +830,10 @@ class SimulationTask(WebTask):
 class BatchTask(WebTask):
     """Interface for managing a batch task on the server."""
 
+    task_type: Optional[str] = Field(
+        None, title="task_type", description="The type of task.", alias="taskType"
+    )
+
     @classmethod
     def get(cls, task_id: str, verbose: bool = True) -> BatchTask:
         """Get batch task by id.
@@ -851,8 +855,11 @@ class BatchTask(WebTask):
         except WebNotFoundError as e:
             td.log.error(f"The requested batch ID '{task_id}' does not exist.")
             raise e
-        # We only need to validate existence; store id on the instance.
-        return BatchTask(taskId=task_id) if resp else None
+        # Extract taskType from response if available
+        if resp:
+            task_type = resp.get("taskType") if isinstance(resp, dict) else None
+            return BatchTask(taskId=task_id, taskType=task_type)
+        return None
 
     def detail(self) -> BatchDetail:
         """Fetches the detailed information and status of the batch.
--- a/tidy3d/web/core/task_core.py
+++ b/tidy3d/web/core/task_core.py
@@ -154,6 +154,13 @@ class WebTask(ResourceLifecycle, Submittable, extra=Extra.allow):
         alias="taskId",
     )
 
+    def _is_batch_type(self) -> bool:
+        """Check if this task uses the batch/modeler API.
+
+        Default implementation returns False. Overridden in SimulationTask.
+        """
+        return False
+
     @classmethod
     def create(
         cls,
@@ -222,7 +229,7 @@ class WebTask(ResourceLifecycle, Submittable, extra=Extra.allow):
 
     def get_url(self) -> str:
         base = str(config.web.website_endpoint or "")
-        if isinstance(self, BatchTask):
+        if self._is_batch_type():
             return "/".join([base.rstrip("/"), f"rf?taskId={self.task_id}"])
         return "/".join([base.rstrip("/"), f"workbench?taskId={self.task_id}"])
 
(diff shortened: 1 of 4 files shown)


tidy3d/web/core/task_core.py
```diff
@@ -450,14 +469,24 @@ def get(cls, task_id: str, verbose: bool = True) -> SimulationTask:
...
+            return SimulationTask(**resp)
+
+        # Fall back to batch/modeler API
+        # Use suppress_404 and catch all exceptions to handle unmocked endpoints in tests
         try:
-            resp = http.get(f"tidy3d/tasks/{task_id}/detail")
-        except WebNotFoundError as e:
-            td.log.error(f"The requested task ID '{task_id}' does not exist.")
-            raise e
+            resp = http.get(f"rf/task/{task_id}/statistics", suppress_404=True)
```
marcorudolphflex, line 481: why is that in a try block at all when we suppress 404? Do we want to let other exceptions pass?

tidy3d/web/core/task_core.py
```diff
@@ -450,14 +469,24 @@ def get(cls, task_id: str, verbose: bool = True) -> SimulationTask:
...
+            if resp:
+                task_type = resp.get("taskType") if isinstance(resp, dict) else None
+                return SimulationTask(taskId=task_id, taskType=task_type)
+        except Exception:
+            pass
 
-        task = SimulationTask(**resp) if resp else None
-        return task
+        td.log.error(f"The requested task ID '{task_id}' does not exist.")
+        raise WebNotFoundError("Resource not found (HTTP 404).")
```
marcorudolphflex, line 489: 404 is not guaranteed here, is it?

tidy3d/web/core/task_core.py
```diff
@@ -450,14 +469,24 @@ def get(cls, task_id: str, verbose: bool = True) -> SimulationTask:
             :class:`.SimulationTask` object containing info about status,
              size, credits of task and others.
         """
+        # Try simulation API first (most common case)
+        # Use suppress_404 to avoid error logging when falling back to batch API
+        resp = http.get(f"tidy3d/tasks/{task_id}/detail", suppress_404=True)
```
marcorudolphflex, line 474: can't we infer the endpoint from the task type?

marcorudolphflex (commented):
As you already pointed out, it may be not be the most elegant solution - but still not bad to burn code. But don't have a strong opinion on that.

yaugenst-flex (commented):
The practical benefit of this was already addressed in https://github.com/flexcompute/tidy3d/pull/3229, and I think the remaining changes add quite a bit of risk. I would treat this as a deferred design idea. Let's revisit this as an RFC?

momchil-flex:
Yeah I'm fine to defer this and tbh have too much on my plate to work towards a "better" fix, which probably involves a broader refactor, so we'll just live with this for a while.

PR #3213, opened 2026-01-28
FXC-5153: making trimesh and matplotlib lazy

momchil-flex:
`trimesh` was not being lazy because of TrimeshType. That was an easy fix.

The bulk of the changes were needed to make matplotlib lazy. This doesn't look too bad, what do you think @yaugenst ?

On my PC import time dropped from 1.9s to 1.2s.

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR implements lazy imports for `trimesh` and `matplotlib` to reduce initial import overhead for the tidy3d package.

**Key Changes:**
- Fixed `trimesh` lazy loading by moving `TrimeshType` import check from module-level to usage-time
- Implemented comprehensive lazy loading for `matplotlib` across visualization components
- Changed `arrow_style` from module-level constant to lazy-loading function
- Added `_ensure_tidy3d_style()` mechanism to defer matplotlib style configuration until first plot
- Moved all matplotlib imports to either `TYPE_CHECKING` blocks (for type hints) or function-level imports
- Added comprehensive test suite using subprocess isolation to verify lazy import behavior

**Impact:**
The changes successfully defer heavy optional dependencies until they're actually needed, improving import performance for users who don't require visualization features.

- This PR is safe to merge with one minor style improvement suggested
- The implementation is well-structured and comprehensive, with proper TYPE_CHECKING usage and function-level imports. Tests validate the behavior. Minor inefficiency in `is_valid_color` that could be optimized but doesn't affect correctness.
- Check `tidy3d/components/viz/visualization_spec.py` for the redundant import optimization

| Filename | Overview |
|----------|----------|
| tests/test_package/test_lazy_imports.py | New comprehensive test suite for lazy imports using subprocess isolation |
| tidy3d/components/types/base.py | Changed matplotlib Axes import to TYPE_CHECKING pattern for lazy loading |
| tidy3d/components/viz/__init__.py | Added lazy style application mechanism to defer matplotlib initialization |
| tidy3d/components/viz/styles.py | Converted arrow_style from module-level constant to lazy-loading function |
| tidy3d/components/viz/visualization_spec.py | Made is_color_like import lazy but imports it twice per validation call |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Tidy3d
    participant VizModule as viz/__init__.py
    participant MakeAx as make_ax()
    participant EnsureStyle as _ensure_tidy3d_style()
    participant Matplotlib

    User->>Tidy3d: import tidy3d
    Note over Tidy3d,VizModule: No matplotlib import at this stage
    Tidy3d-->>User: Module loaded (fast)
    
    User->>Tidy3d: sim.plot(z=0)
    Tidy3d->>MakeAx: ax is None, create one
    MakeAx->>EnsureStyle: Check if style applied
    
    alt First plot call
        EnsureStyle->>Matplotlib: import matplotlib
        EnsureStyle->>Matplotlib: apply_tidy3d_params()
        Note over EnsureStyle: Set _tidy3d_style_applied = True
    else Subsequent calls
        Note over EnsureStyle: Style already applied, skip
    end
    
    MakeAx->>Matplotlib: plt.subplots()
    Matplotlib-->>MakeAx: ax created
    MakeAx-->>Tidy3d: Return ax
    Tidy3d->>Matplotlib: Render plot
    Matplotlib-->>User: Display plot
```

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes import-time behavior and plotting paths across multiple modules; failures would surface as runtime ImportErrors or subtle plotting/style differences rather than at import time.
> 
> **Overview**
> **Makes optional heavy deps truly lazy.** Removes/avoids module-level `matplotlib` imports across plotting-related code and defers style initialization by introducing `_ensure_tidy3d_style()` (no longer calling `apply_tidy3d_params()` at `tidy3d.components.viz` import).
> 
> Moves `matplotlib` imports into plot-time code paths (e.g., `Simulation`/EME plotting, triangular dataset plotting, geometry arrow rendering, colorbar helpers) and updates arrow styling to a lazy `arrow_style()` factory. Also adjusts `VisualizationSpec` color validation to check availability via `find_spec` and lazily import `is_color_like`, and updates `check_import()` to use `find_spec` to avoid importing modules just to check presence.
> 
> Fixes `TrimeshType` so `trimesh` isn’t imported during `tidy3d` import (including handling “installed but broken” cases), and adds subprocess-based tests ensuring `matplotlib`/`trimesh`/`vtk` (and `networkx`) are not imported on `import tidy3d` while `matplotlib` *is* imported when calling `plot()`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fdd9e22783c006d78430c088d24751d801c48e42. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/test_package/test_lazy_imports.py
+++ b/tests/test_package/test_lazy_imports.py
@@ -1,0 +1,158 @@
+"""Tests that heavy optional packages are not imported at top-level tidy3d import.
+
+These tests ensure that packages like matplotlib, scipy, trimesh, and vtk
+are only imported when actually needed (lazy imports), not at module load time.
+This keeps the initial import time low for users who don't need these features.
+"""
+
+from __future__ import annotations
+
+import os
+import subprocess
+import sys
+
+import pytest
+
+
+def subprocess_env() -> dict[str, str]:
+    """Return a clean environment for subprocess calls.
+
+    Removes coverage-related variables that could cause additional imports
+    in the subprocess and interfere with lazy import tests.
+    """
+    env = dict(os.environ)
+    # Remove coverage-related variables that might cause extra imports
+    for key in list(env.keys()):
+        if "COV" in key.upper() or "COVERAGE" in key.upper():
+            del env[key]
+    return env
+
+
+# List of packages that should NOT be imported on top-level tidy3d import
+# Note: scipy is excluded because xarray (a core dependency) imports its
+# scipy backend automatically via entrypoints
+LAZY_PACKAGES = [
+    "matplotlib",
+    "trimesh",
+    "vtk",
+    "networkx",  # transitive dependency of trimesh
+]
+
+
+@pytest.mark.parametrize("package", LAZY_PACKAGES)
+def test_package_not_imported_on_tidy3d_import(package: str) -> None:
+    """Test that a package is not imported when importing tidy3d.
+
+    We run this in a subprocess to ensure a clean Python environment
+    without any prior imports that might pollute sys.modules.
+    """
+    # Create a script that imports tidy3d and checks if the package is in sys.modules
+    script = f"""
+import sys
+import tidy3d
+if "{package}" in sys.modules:
+    print(f"FAIL: {package} was imported")
+    sys.exit(1)
+else:
+    print(f"OK: {package} was not imported")
+    sys.exit(0)
+"""
+    # Use -E to ignore PYTHON* env vars and -s to disable site customization
+    # that might be installed by coverage tools
+    result = subprocess.run(
+        [sys.executable, "-E", "-s", "-c", script],
+        capture_output=True,
+        text=True,
+        env=subprocess_env(),
+    )
+
+    # Print output for debugging
+    if result.stdout:
+        print(result.stdout)
+    if result.stderr:
+        print(result.stderr, file=sys.stderr)
+
+    assert result.returncode == 0, (
+        f"Package '{package}' was imported on top-level tidy3d import. "
+        f"This increases import time. Consider making the import lazy."
+    )
+
+
+def test_all_lazy_packages_not_imported() -> None:
+    """Test that none of the lazy packages are imported when importing tidy3d.
+
+    This is a combined test that checks all packages at once, which is faster
+    than running separate subprocesses for each package.
+    """
+    packages_str = ", ".join(f'"{p}"' for p in LAZY_PACKAGES)
+    script = f"""
+import sys
+# Import tidy3d
+import tidy3d
+# Check which packages were imported
+lazy_packages = [{packages_str}]
+imported = [p for p in lazy_packages if p in sys.modules]
+if imported:
+    print(f"FAIL: These packages were imported: {{imported}}")
+    sys.exit(1)
+else:
+    print(f"OK: None of the lazy packages were imported")
+    sys.exit(0)
+"""
+    # Use -E to ignore PYTHON* env vars and -s to disable site customization
+    result = subprocess.run(
+        [sys.executable, "-E", "-s", "-c", script],
+        capture_output=True,
+        text=True,
+        env=subprocess_env(),
+    )
+
+    # Print output for debugging
+    if result.stdout:
+        print(result.stdout)
+    if result.stderr:
+        print(result.stderr, file=sys.stderr)
+
+    assert result.returncode == 0, (
+        f"Some lazy packages were imported on top-level tidy3d import: {result.stdout}"
+    )
+
+
+def test_matplotlib_imported_on_plot() -> None:
+    """Test that matplotlib IS imported when a plot function is called."""
+    script = """
+import sys
+import tidy3d as td
+
+# matplotlib should not be imported yet
+assert "matplotlib" not in sys.modules, "matplotlib imported too early"
+
+# Create a simple simulation and try to plot it
+sim = td.Simulation(
+    size=(1, 1, 1),
+    grid_spec=td.GridSpec.auto(wavelength=1.0),
+    run_time=1e-12,
+)
+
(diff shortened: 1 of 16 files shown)


tests/test_package/test_lazy_imports.py
```diff
@@ -0,0 +1,138 @@
...
+sim = td.Simulation(
+    size=(1, 1, 1),
+    grid_spec=td.GridSpec.auto(wavelength=1.0),
+    run_time=1e-12,
+)
+
+# This should trigger matplotlib import
+import matplotlib
+matplotlib.use('Agg')  # Use non-interactive backend for testing
+ax = sim.plot(z=0)
```
yaugenst-flex: This doesn't actually prove plotting triggers the import because matplotlib is imported explicitly _before_ the plotting call?

tidy3d/components/viz/styles.py
```diff
@@ -1,11 +1,17 @@
 from __future__ import annotations
 
-try:
-    from matplotlib.patches import ArrowStyle
+_arrow_style = None
+
+
+def arrow_style():
```
yaugenst-flex: This changed from value to function and is sort of a public-ish function since it's exposed in `viz/__init__.py`. Could just not do that instead, I doubt anyone uses it..

tidy3d/components/viz/axes_utils.py
```diff
@@ -99,6 +99,10 @@ def tick_values(self, vmin, vmax):
 
 def make_ax() -> Ax:
     """makes an empty ``ax``."""
+    from tidy3d.components.viz import _ensure_tidy3d_style
+
+    _ensure_tidy3d_style()
```
yaugenst-flex, line 116: Maybe this is on purpose, but it constitutes a behavioral change. This means that the matplotlib style is no longer applied at import time and only happens when this function is used, meaning if a user passes an existing `ax`, our style won't be applied. This behavior might actually be better though, not sure.

tidy3d/components/types/__init__.py
```diff
@@ -68,7 +68,6 @@
     tidycomplex,
     xyz,
 )
-from tidy3d.components.types.third_party import TrimeshType
```
yaugenst-flex, line 66: Technically a breaking change...

yaugenst-flex (commented):
Overall looks like a good direction, cleaner overall than before. Some comments to iron out.

marcorudolphflex (commented):
LGTM overall. I do not really have something to add but I wouldn't really mind the theoretical user-facing changes for these minor cases too much. Would also fix the testing issue pointed out by @yaugenst-flex

marc-flex (approved):

tidy3d/components/types/__init__.py
momchil-flex: Yes but like... should be fine? :)

tidy3d/components/viz/axes_utils.py
momchil-flex: Hm, it does sound better to me.

tidy3d/components/viz/styles.py
momchil-flex: Right, sounds very benign. But do you mean I should rather remove it from init?

tests/test_package/test_lazy_imports.py
momchil-flex: fixed

tidy3d/components/viz/axes_utils.py
yaugenst-flex: Agreed. Will need to make sure though the notebooks look ok.

tidy3d/components/viz/styles.py
yaugenst-flex: Yeah I think since it's pretty clearly an internal function we probably don't want to have it init

tidy3d/components/viz/axes_utils.py
momchil-flex: Hm actually on second thought maybe it's not better. You are right that I often pass axes (and our notebooks do) and we still want the style.

tidy3d/components/viz/styles.py
momchil-flex: Done

tidy3d/components/viz/axes_utils.py
momchil-flex: I reverted it for consistency. I think at the end of the day that's more important since it might be common for a user to mix passing axes and not passing them, without any explicit style goal in doing so.

momchil-flex:
@yaugenst-flex I pushed one last fix to the scipy import test; it seems the code coverage was interfering with the environment so now the lazy import tests are using an entirely clean env.

The only thing left imo is to hear your opinion on the last unresolved comment above. And if you want to take another look, go ahead.

tidy3d/components/viz/axes_utils.py
yaugenst-flex: sounds good yeah. tbh we could probably design the whole visualization thing with a bit more intent 😄 but for now this sounds good

yaugenst-flex (approved):
yeah LGTM!

marcorudolphflex (approved):

momchil-flex:
Finally figured out why scipy lazy import test keeps failing in CI: recent versions of xarray auto-load their backends via entrypoints, which includes the scipy backend. I had to upgrade xarray locally to reproduce the error finally.

Since xarray is a core dependence I think we have no choice but to accept that scipy will not be imported lazily if available. However many things should still work without scipy if not available (there's a separate test for that).

Sounds good @yaugenst-flex ?

PR #3216, opened 2026-01-28
reduce computation time for adaptive_vjp_spacing via caching for GeometryGroup

groberts-flex:
`adapative_vjp_spacing` takes longer to run now that we use the full `eps_data` inside of `DerivativeInfo`. For geometry groups this was running for every geometry and adding a significant overhead. Just caching the property didn't work because the `derivative_info` variable is copied and updated with new derivative paths in `GeometryGroup`. Instead, we allow the variable to be cached inside of `DerivativeInfo`.

---

> [!NOTE]
> **Low Risk**
> Low risk performance optimization: changes are localized to spacing computation/caching and are covered by new tests asserting unchanged derivative outputs and cache cleanup.
> 
> **Overview**
> Reduces autograd shape-derivative overhead for `GeometryGroup` by caching the permittivity-derived minimum spacing used by `DerivativeInfo.adaptive_vjp_spacing`, so it’s computed once per group instead of once per child geometry.
> 
> This refactors the permittivity spacing calculation into a reusable `DerivativeInfo.min_spacing_from_permittivity` property, adds a `cache_min_spacing_from_permittivity()` context manager to control cache lifetime, and wraps `GeometryGroup._compute_derivatives()` in that cache scope; tests were updated to assert identical VJP results with/without caching and that the cache is cleared after derivative computation. The changelog notes the performance improvement.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a4b2f2fb7b937b1ce35480643c38d6bdcf133bd7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR optimizes the performance of `adaptive_vjp_spacing` for `GeometryGroup` by implementing a caching mechanism for the permittivity-based spacing calculation.

**Key Changes:**
- Extracted the core permittivity-based spacing computation into a new `@cached_property` method `min_spacing_from_permittivity` in `DerivativeInfo`
- Added a `cached_min_spacing_from_permittivity` field to enable explicit caching that survives `dataclasses.replace()` operations
- Added helper methods `cache_min_spacing_from_permittivity()` and `uncache_min_spacing_from_permittivity()` to control cache lifecycle
- Modified `GeometryGroup._compute_derivatives()` to cache the value once before the loop and uncache after, so all child geometries share the same cached value

**Implementation Details:**
The caching approach cleverly handles the fact that `GeometryGroup` creates copies of `DerivativeInfo` for each child geometry using `updated_copy()` (which uses `dataclasses.replace()`). Since `@cached_property` stores its cache in the instance's `__dict__`, it doesn't survive across `replace()` operations. The solution is to store the computed value in an explicit field (`cached_min_spacing_from_permittivity`) which DOES get copied during `replace()`. The `@cached_property` decorator checks this field first, so copied instances use the pre-computed value without recalculation.

**Performance Impact:**
Previously, `adaptive_vjp_spacing` would compute the permittivity-based spacing for every geometry in a `GeometryGroup`, even though all geometries share the same `eps_data`. With this change, it's computed once and reused across all geometries in the group.

- This PR is safe to merge with low risk - it's a performance optimization with a well-thought-out caching mechanism
- The implementation is sound and the caching logic correctly handles the interaction between `@cached_property` and `dataclasses.replace()`. The confidence is 4 (not 5) because this is a subtle performance optimization involving cached properties and dataclass copies, which could benefit from additional testing to verify the performance improvement and ensure no edge cases were missed.
- No files require special attention - all changes are straightforward

| Filename | Overview |
|----------|----------|
| CHANGELOG.md | Added performance improvement entry for `adaptive_vjp_spacing` caching in `GeometryGroup` |
| tidy3d/components/autograd/derivative_utils.py | Refactored `adaptive_vjp_spacing` to cache permittivity-based spacing via new `cached_min_spacing_from_permittivity` field and `@cached_property` decorator |
| tidy3d/components/geometry/base.py | Added caching calls around geometry loop in `GeometryGroup._compute_derivatives` to enable cache sharing across geometries |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant GG as GeometryGroup
    participant DI as DerivativeInfo (original)
    participant DI_copy as DerivativeInfo (copy)
    participant G1 as Geometry 1
    participant G2 as Geometry 2
    
    Note over GG: _compute_derivatives called
    GG->>DI: cache_min_spacing_from_permittivity()
    activate DI
    DI->>DI: Access min_spacing_from_permittivity (@cached_property)
    DI->>DI: Compute spacing from eps_data
    DI->>DI: Store in cached_min_spacing_from_permittivity field
    deactivate DI
    
    loop For each geometry in GeometryGroup
        GG->>DI: updated_copy(paths, bounds, ...)
        DI-->>DI_copy: Create shallow copy via dataclasses.replace()
        Note over DI_copy: cached_min_spacing_from_permittivity<br/>is copied to new instance
        
        GG->>G1: _compute_derivatives(DI_copy)
        activate G1
        G1->>DI_copy: adaptive_vjp_spacing()
        DI_copy->>DI_copy: Access min_spacing_from_permittivity
        Note over DI_copy: Check cached_min_spacing_from_permittivity<br/>Return cached value immediately
        DI_copy-->>G1: Return cached spacing
        G1-->>GG: Return gradient
        deactivate G1
    end
    
    GG->>DI: uncache_min_spacing_from_permittivity()
    Note over DI: Set cached_min_spacing_from_permittivity = None
    
    Note over GG: All geometries used cached value,<br/>avoiding redundant computation
```

### DIFF
--- a/tidy3d/components/geometry/base.py
+++ b/tidy3d/components/geometry/base.py
@@ -3593,26 +3593,28 @@ class GeometryGroup(Geometry):
         # create interpolators once for all geometries to avoid redundant field data conversions
         interpolators = derivative_info.interpolators or derivative_info.create_interpolators()
 
-        for field_path in derivative_info.paths:
-            _, index, *geo_path = field_path
-            geo = self.geometries[index]
-            # pass pre-computed interpolators if available
-            geo_info = derivative_info.updated_copy(
-                paths=[tuple(geo_path)],
-                bounds=geo.bounds,
-                bounds_intersect=self.bounds_intersection(
-                    geo.bounds, derivative_info.simulation_bounds
-                ),
-                deep=False,
-                interpolators=interpolators,
-            )
+        with derivative_info.cache_min_spacing_from_permittivity():
+            for field_path in derivative_info.paths:
+                _, index, *geo_path = field_path
 
-            vjp_dict_geo = geo._compute_derivatives(geo_info)
+                geo = self.geometries[index]
+                # pass pre-computed interpolators if available
+                geo_info = derivative_info.updated_copy(
+                    paths=[tuple(geo_path)],
+                    bounds=geo.bounds,
+                    bounds_intersect=self.bounds_intersection(
+                        geo.bounds, derivative_info.simulation_bounds
+                    ),
+                    deep=False,
+                    interpolators=interpolators,
+                )
 
-            if len(vjp_dict_geo) != 1:
-                raise AssertionError("Got multiple gradients for single geometry field.")
+                vjp_dict_geo = geo._compute_derivatives(geo_info)
 
-            grad_vjps[field_path] = vjp_dict_geo.popitem()[1]
+                if len(vjp_dict_geo) != 1:
+                    raise AssertionError("Got multiple gradients for single geometry field.")
+
+                grad_vjps[field_path] = vjp_dict_geo.popitem()[1]
 
         return grad_vjps
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added deprecation warning for ``TemperatureMonitor`` and ``SteadyPotentialMonitor`` when ``unstructured`` parameter is not explicitly set. The default value of ``unstructured`` will change from ``False`` to ``True`` after the 2.11 release.
 - Added validation to `GaussianDoping` to ensure `ref_con < concentration`, validate `source` face identifier, and warn the user when the box size is not sufficient for the specified transition width.
 - Local caching is now enabled by default (set `td.config.local_cache.enabled=False` to opt out).
+- Reduced computation time of `adaptive_vjp_spacing` for `GeometryGroup` by allowing permittivity based spacing value to be cached.
 
 ### Fixed
 - Fixed intermittent "API key not found" errors in parallel job launches by making configuration directory detection race-safe.
--- a/tests/test_components/autograd/test_autograd.py
+++ b/tests/test_components/autograd/test_autograd.py
@@ -5,6 +5,7 @@ import copy
 import cProfile
 import typing
 import warnings
+from contextlib import nullcontext
 from dataclasses import dataclass
 from importlib import reload
 from os.path import join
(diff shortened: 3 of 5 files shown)


tidy3d/components/geometry/base.py
```diff
@@ -3606,6 +3608,8 @@ def _compute_derivatives(self, derivative_info: DerivativeInfo) -> AutogradField
 
             grad_vjps[field_path] = vjp_dict_geo.popitem()[1]
 
+        derivative_info.uncache_min_spacing_from_permittivity()
```
marcorudolphflex: to simplify the cache/uncache workflow, I would prefer a context manager like
```
@contextmanager
def cached_min_spacing(self) -> Iterator[None]:
    """
    Cache min_spacing_from_permittivity for the duration of the block.
    Cache is always cleared on exit.
    """
    self.cached_min_spacing_from_permittivity = self.min_spacing_from_permittivity
    try:
        yield
    finally:
        self.cached_min_spacing_from_permittivity = None
```
which can be used like

```
with derivative_info.cached_min_spacing():
    # operations which need cache
```

marcorudolphflex (commented):
Good to have that! How many impact do we observe from that caching?

yaugenst-flex (approved):
Thanks @groberts-flex LGTM! Agree with @marcorudolphflex's comment that this might be nicer with a context manager, alternatively a `try/finally` to ensure things are cleaned up.
Only nit would be that it might be good to add a unit test to ensure that cached vs uncached `adaptive_vjp_spacing` results are identical. Helps catch future regressions.

tidy3d/components/geometry/base.py
groberts-flex: thank you, I really like this!

groberts-flex:
> Good to have that! How many impact do we observe from that caching?

It is really beneficial in geometry groups with a lot of geometries. For example, in the performance profiling test, I have a 35x35 array of geometries in a geometry group. Without caching, the `postprocess_adj` function takes 602sec. With caching it takes 327sec.

groberts-flex:
> Thanks @groberts-flex LGTM! Agree with @marcorudolphflex's comment that this might be nicer with a context manager, alternatively a `try/finally` to ensure things are cleaned up. Only nit would be that it might be good to add a unit test to ensure that cached vs uncached `adaptive_vjp_spacing` results are identical. Helps catch future regressions.

great point, I'll definitely add a test!

PR #3218, opened 2026-01-29
FXC-1617: switch tensorial mode solver to tidy3d-extras

momchil-flex:
The tensorial mode solver in tidy3d-extras has been made more accurate. Notably, for nominally lossless, fully tensorial media, the mode solver should return very close to lossless modes.

To avoid confusion, we are removing the legacy local tensorial mode solver. Local runs are still fully supported (and more accurate!) they just require installing tidy3d-extras.

---

> [!NOTE]
> **Medium Risk**
> Changes affect the mode-solver execution path and error behavior for tensorial/angled/aniso cases, and rely on optional `tidy3d-extras` availability; diagonal/local and server paths should remain unchanged but need regression coverage.
> 
> **Overview**
> Local mode solving now **selects the best available solver at runtime**: `ModeSolver`/`ModeSimulation.run_local()` will use `tidy3d-extras`’ mode solver when present, otherwise fall back to the base solver.
> 
> The legacy in-core **fully tensorial eigenmode implementation is removed**; attempting a tensorial solve (non-zero `ModeSpec.angle_theta` or fully anisotropic media) without `tidy3d-extras` now raises a clear `NotImplementedError` instructing users to install `tidy3d[extras]` or run via the server.
> 
> Tests and docs are updated accordingly: new coverage asserts the error/fallback behavior and skips tensorial-local tests when extras aren’t available, and plugin tests that directly exercised `compute_modes`/angled tensorial paths are removed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3cb8539978efdf4590c576b7cc389842ba2a18a7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tidy3d/components/mode/mode_solver.py
```diff
@@ -1644,10 +1685,9 @@ def _solve_single_freq(
         The fields are rotated from propagation coordinates back to global coordinates.
         """
 
-        if not LOCAL_SOLVER_IMPORTED:
-            raise ImportError(IMPORT_ERROR_MSG)
+        solver_func = _get_solver_func()
```
caseyflex, line 1688: Why not use the `@supports_local_subpixel` decorator and related logic like in `_solve_all_freqs`?

caseyflex: I think that handling is somewhat more flexible actually, in addition to already being factored out into a subroutine. It lets the user control the behavior: config.use_local_subpixel=None means use extras if available, False means don't use it, True means always use it. I would do the same thing here, and just raise an error if tensorial mode solve is required and `tidy3d_extras["use_local_subpixel"]` is False.

caseyflex (changes requested):

tests/test_components/test_mode.py
```diff
@@ -510,3 +510,102 @@ def test_filter_pol_with_default_sort_spec():
...
+    from tidy3d.components.mode import mode_solver as mode_solver_module
+    from tidy3d.components.mode.solver import compute_modes
+    from tidy3d.packaging import tidy3d_extras
+
+    # Mock _get_solver_func to always return the base solver (simulating no tidy3d-extras)
+    monkeypatch.setattr(mode_solver_module, "_get_solver_func", lambda: compute_modes)
+
+    # Also disable local subpixel to prevent tidy3d-extras from bypassing our patched code path
+    # when tidy3d-extras is installed and licensed
+    monkeypatch.setitem(tidy3d_extras, "use_local_subpixel", False)
```
caseyflex, line 556: on a related note, not sure about this test logic. Seems we want the error when this setting is None too.

tests/test_components/test_mode.py
```diff
@@ -510,3 +510,102 @@ def test_filter_pol_with_default_sort_spec():
...
+    assert result.modes_raw.n_eff is not None
+
+    # Test with fully anisotropic medium
+    sim_aniso = _make_tensorial_mode_sim(fully_anisotropic=True)
+    result_aniso = sim_aniso.run_local()
+    assert result_aniso is not None
+    assert result_aniso.modes_raw.n_eff is not None
+
+
+def test_diagonal_mode_solver_still_works():
```
caseyflex, line 593: This test seems redundant

tests/test_components/test_mode.py
caseyflex: To take advantage of the integration tests, in addition to the mock test for the error message, you could also have a test that checks whether tidy3d_extras is available (I think there are others like that) and has different behavior in the two cases

tests/test_plugins/test_mode_solver.py
```diff
@@ -660,125 +645,6 @@ def test_mode_solver_unstructured_custom_medium(nx, cond_factor, interp, tol, tm
     assert error_up < tol
 
 
-@td.packaging.disable_local_subpixel
-def test_mode_solver_straight_vs_angled():
```
caseyflex, line 664: Interesting, so this test previously had local subpixel disabled, presumably because subpixel would change the results of the test. I guess your new tests cover this use case? We could have one of them in the frontend but only run if tidy3d_extras is present?

tidy3d/components/mode/mode_solver.py
momchil-flex: Yep, great suggestion, implemented and simplifies things.

tests/test_components/test_mode.py
momchil-flex: added

tests/test_plugins/test_mode_solver.py
momchil-flex: So the problem is that some of the tidy3d_frontend CI was running without extras installed, and then this would error. Now that tidy3d_frontend is essentially moved to compute, I think your comment about having a test in the frontend doesn't make that much sense? They're all in compute just in different folders.

tests/test_components/test_mode.py
momchil-flex: To be honest I am a bit confused about all the possible combinations of extra installed vs. not installed in CI, and what the existing tests would cover. I feel like better to leave it explicitly just in case?

tests/test_plugins/test_mode_solver.py
caseyflex: @daquinteroflex 

I’m not quite clear on what is happening with frontend tests, solver tests, integration tests in CI now. Ideally we would still have fast tests, slow tests, and for all combinations: frontend with or without tidy3d-extras installed, and on all supported platforms?

daquinteroflex: Hi Casey, yes, very soon we will have remote tests like before. I had to disable them. We will have it as you describe but not yet

 https://enreal.slack.com/archives/C037SK4U96Z/p1772044936049529?thread_ts=1772024685.975309&cid=C037SK4U96Z

PR #3225, opened 2026-02-03
feat(tidy3d): FXC-5275 schema-versioned config migrations for tidy3d config

marcorudolphflex:
- Introduces schema-versioned config files (config_version) with a migration registry and auto in-memory upgrades, including optional write-back and forward-compat handling.
- Adds a tidy3d config upgrade CLI for dry-run diffs, CI checks, and manual schema upgrades.
- Enforces deprecation metadata (deprecated_in, removed_in, replaced_by) during config load.
- Updates docs and CHANGELOG to describe versioning behavior and new CLI usage.

---

> [!NOTE]
> **Medium Risk**
> Touches core config load/validation and CLI behavior, with new auto-migration/write-back paths and forward-compat parsing that could affect startup or persisted config files if edge cases slip through.
> 
> **Overview**
> Adds **schema-versioned configuration** via a root `config_version`, a `register_migration` registry (`tidy3d.config.migrations`), and automatic in-memory upgrades with optional write-back (keeps `.bak` snapshots) plus forward-compat handling (`best-effort` filtering vs `strict` via env vars).
> 
> Refactors config loading/validation: `ConfigLoader` now queues/commits migration writes after validating merged base+profile trees, ignores unknown env overrides, enforces base-only metadata (`default_profile`) placement, and introduces centralized deprecation/removed-key enforcement during validation.
> 
> Extends the CLI: moves `configure` into `tidy3d.web.cli.config` and adds `tidy3d config upgrade` to preview diffs (`--dry-run`), enforce CI checks (`--check`), and apply upgrades transactionally. Documentation/CHANGELOG are updated accordingly, and extensive new tests cover migration, forward-compat, rollback, and legacy Nexus behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit aa0c6be298f3531f1f8f359ae45236bbe73e605f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tidy3d/config/loader.py
yaugenst-flex: If the parsing fails, we set `document = tomlkit.document()` and continue, and then `_apply_schema_migrations()` will run migrations against this empty document. This can drop all user config keys, and then when we write back a file that will contain only `config_version`. So in this case there will be data loss for the user from a failed parse.
Not sure how to handle.. error? Or skip the migration?

yaugenst-flex: Nit: at this point, `legacy` has already been passed through `_migrate_legacy_payload()`, so the returned data is migrated.

tidy3d/web/cli/config.py
```diff
@@ -0,0 +1,543 @@
...
+                diff = _format_diff(path, result["before"], result["after"])
+                if diff:
+                    click.echo(diff)
+            elif not check and result["document"] is not None:
+                loader.write_document(path, result["document"])
+
+    if check:
+        if changed_paths:
+            raise click.ClickException("Configuration upgrade required.")
+        if forward_paths:
```
yaugenst-flex, line 429: is this unreachable? since forward-version files already raise when `check=True`?

tidy3d/config/manager.py
```diff
@@ -21,6 +21,7 @@
 from tidy3d.log import log
 
 from .loader import ConfigLoader, deep_diff, deep_merge, load_environment_overrides
+from .migrations import _resolve_model_type, check_deprecations
```
yaugenst-flex: Can we avoid importing `_resolve_model_type` from `migrations`? This couples the manager to the manager internals. Maybe we can create a helper module like `schema_utils`?

tidy3d/config/migrations.py
```diff
@@ -0,0 +1,372 @@
...
+    if isinstance(value, bool) or not isinstance(value, int):
+        log.warning(f"Ignoring invalid {label}={value!r} on '{path}'.")
+        return None
+    if value < 0:
+        log.warning(f"Ignoring invalid {label}={value!r} on '{path}'.")
+        return None
+    return value
+
+
+def check_deprecations(
```
yaugenst-flex: This is used by config loading and model building, but it lives in the migrations module. Could we move it to a dedicated `deprecations.py` so `migrations.py` stays focused on .. migrations 😆

tidy3d/web/cli/config.py
```diff
@@ -0,0 +1,543 @@
...
+    base_path = config_dir / "config.toml"
+    if base_path.exists():
+        targets.append(base_path)
+
+    if profiles_dir.exists():
+        targets.extend(sorted(profiles_dir.glob("*.toml")))
+    return targets
+
+
+def _preview_schema_upgrade(path: Path) -> dict[str, Any]:
```
yaugenst-flex: This seems to mostly reimplement the migration/validation that's already used in runtime loading, so I think a shared helper could reduce drift here?

yaugenst-flex (changes requested):
Thanks @marcorudolphflex this is great overall! There are not really any blocking issues (except maybe the toml parsing thing) but I noticed that this introduces a bit of coupling between modules that hopefully could be avoided. While reviewing I also noticed that probably the config module could be further refactored to reduce coupling overall because there is quite a bit, however that seems out of scope for now. I just commented on some low hanging fruit.

tidy3d/config/loader.py
marcorudolphflex: changed to "Using legacy data without migration."

tidy3d/web/cli/config.py
marcorudolphflex: it was. Should now respect the non-strict mode and raise here

tidy3d/config/manager.py
marcorudolphflex: makes sense, added the new module

tidy3d/config/loader.py
marcorudolphflex: would handle with error log
```
f"Automatic configuration migration failed for '{path}' "
f"(from config_version {version} to {CURRENT_CONFIG_VERSION}): {exc}. "
"Continuing with the original configuration (no migration write-back). "
"Retry manually with 'tidy3d config upgrade' after fixing the issue."
```
and skip.

marcorudolphflex:
Additional to your comments, I did some refactorings and error-handling-related tests/fixes to address bugbot-reviews.

tidy3d/config/loader.py
```diff
@@ -313,6 +505,63 @@ def _clean_data(data: Any) -> Any:
...
+
+        payload = data.get(name, {})
+        if not isinstance(payload, dict) and coerce_non_dict:
+            payload = {}
+        yield SectionPayload(name, schema, payload, (name,), None)
+
+
+def _validate_config_data(data: dict[str, Any]) -> None:
+    errors: list[Exception] = []
+    for item in iter_section_payloads(data, coerce_non_dict=True):
```
yaugenst-flex: I reproduced a validation mismatch here that can make `tidy3d config upgrade --check` return a false green.

`_validate_config_data()` iterates with `coerce_non_dict=True`, so malformed section payloads like `logging = "oops"` are coerced to `{}` and pass validation. Runtime loading (`ConfigManager._build_models`) uses `coerce_non_dict=False`, so the same file fails at startup.

Minimal repro:
- `config.toml`:
  - `config_version = 1`
  - `logging = "oops"`
- `ConfigManager(config_dir=...)` -> raises `TypeError` (`**` expects mapping)
- `tidy3d config upgrade --check` -> exit `0`, `Configuration files are up to date.`

Could we align upgrade-check validation semantics with runtime for `config_version <= CURRENT_CONFIG_VERSION` (or route both through one shared strict validation path)?

tidy3d/web/cli/config.py
```diff
@@ -0,0 +1,494 @@
...
+                    click.echo(diff)
+            elif not check and result["document"] is not None:
+                loader.write_document(path, result["document"])
+
+    if check:
+        if changed_paths:
+            raise click.ClickException("Configuration upgrade required.")
+        if forward_paths:
+            raise click.ClickException("Configuration files target a newer schema version.")
+        click.echo("Configuration files are up to date.")
```
yaugenst-flex, line 431: Given the loader-side coercion in preview validation, this success path can currently fire for invalid current-version configs.

In my repro (`config_version = 1`, `logging = "oops"`), runtime init fails, but this command still returns exit 0 with "Configuration files are up to date.".

I think `config upgrade --check` should fail whenever runtime would fail for the same file, otherwise CI may report a clean state while startup still breaks.

yaugenst-flex (commented):
I did another local pass and found one concrete regression tied to the validation path used by `config upgrade --check`.

Summary:
- `ConfigManager` runtime load and `ConfigLoader.preview_schema_upgrade()` currently disagree on malformed non-table section payloads.
- This allows `tidy3d config upgrade --check` to return success on files that runtime rejects.

I left inline comments on:
- `tidy3d/config/loader.py` (coercion behavior in `_validate_config_data`)
- `tidy3d/web/cli/config.py` (false-green success path in `--check` mode)

I also validated with a minimal repro (`config_version = 1`, `logging = "oops"`) showing runtime failure vs check-mode success.

tidy3d/web/cli/config.py
marcorudolphflex: thanks, fixed and added direct test on such case

tidy3d/config/loader.py
marcorudolphflex: should be covered now

yaugenst-flex (approved):

PR #3194, opened 2026-01-22
fix(tidy3d): FXC-5005-web-estimate-cost-printing-duplicate-output

marcorudolphflex:
> [!NOTE]
> Eliminates duplicate cost estimation output when uploading a `Job` by separating implicit vs. explicit verbosity and plumbing a dedicated flag.
> 
> - Suppresses `estimate_cost` during implicit `Job.task_id` resolution by calling `_upload(verbose_estimate_cost=False)`; `upload()` now explicitly calls `estimate_cost(verbose=True)` to print once
> - Adds `verbose_estimate_cost` to `web.upload` and threads it through `Job._upload`
> - Extends test fakes to honor verbosity and adds `test_job_estimate_cost_logging` to assert correct print counts
> - Updates `CHANGELOG.md` to note the fix
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0eeacefb01a8bef07c3bc602d71544fbd9ba28c1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

<h3>Greptile Summary</h3>

This PR fixes duplicate cost estimation output when uploading a `Job`. The issue occurred because accessing `job.task_id` (which implicitly uploads) and calling `job.upload()` both triggered cost estimation printing.

**Key Changes:**
- Modified `Job.task_id` property to call `_upload(verbose=False)` instead of `_upload()`, suppressing the implicit cost estimation print
- Updated `_upload()` method to accept optional `verbose` parameter and forward it to `web.upload()`
- Modified `Job.upload()` to explicitly call `estimate_cost(verbose=True)` before accessing `task_id`, ensuring single controlled output
- Enhanced test mocks (`fake_estimate_cost` and `fake_upload`) to respect the `verbose` parameter
- Added comprehensive test `test_job_estimate_cost_logging` that verifies cost estimation prints exactly once during upload and respects verbose flags

The fix elegantly separates concerns: implicit operations (accessing `task_id`) run silently, while explicit operations (`upload()`) provide user feedback when appropriate.

<h3>Confidence Score: 5/5</h3>

- This PR is safe to merge with minimal risk
- The changes are well-contained and surgical, addressing a specific UX issue without affecting core logic. The implementation correctly uses the existing verbose parameter pattern, and comprehensive tests verify the behavior. No breaking changes or edge cases were identified.
- No files require special attention

<h3>Important Files Changed</h3>

| Filename | Overview |
[...]

PR #3229, opened 2026-02-04
fix(tidy3d): FXC-5311-minimize-http-calls-for-web-load

marcorudolphflex:
- Reduced redundant server lookups when loading batch and task results.
- Cached task kind as class in TaskFactory and used it to avoid unnecessary fetches.
- Simplified load logging message.
- Added regression test to ensure batch item loads skip task lookups when filename isn’t default.

---

> [!NOTE]
> **Medium Risk**
> Touches core web loading/monitoring paths and task-kind detection; risk is mainly incorrect task classification or missed lookups leading to wrong filenames or load behavior, but changes are small and covered by regression tests.
> 
> **Overview**
> Reduces redundant web API calls during result loading by caching task “kind” in `TaskFactory` (as the task class) and using `get_kind()` in `monitor()`/`load()` to decide batch vs simulation behavior without fetching full task objects.
> 
> `BatchData.load_sim_data()` no longer performs an eager `web.get_info()` lookup before calling `web.load()`, and `web.load()` simplifies its verbose output while still handling component-modeler batch default filenames based on task kind. Tests were updated/added to assert batch loads don’t trigger web lookups, and the changelog notes the fix.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e939322376fe4656bb3900ec867222c3802b5188. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -682,8 +682,6 @@ class BatchData(Tidy3dBaseModel, Mapping):
         task_data_path = Path(self.task_paths[task_name])
         task_id = self.task_ids[task_name]
         from_cache = self.cached_tasks[task_name] if self.cached_tasks else False
-        if not from_cache:
-            web.get_info(task_id)
 
         data = web.load(
             task_id=None if from_cache else task_id,
--- a/tidy3d/web/api/webapi.py
+++ b/tidy3d/web/api/webapi.py
@@ -790,8 +790,8 @@ def monitor(task_id: TaskId, verbose: bool = True, worker_group: Optional[str] =
     """
 
     # Batch/modeler monitoring path
-    task = TaskFactory.get(task_id)
-    if isinstance(task, BatchTask):
+    task_kind = TaskFactory.get_kind(task_id)
+    if task_kind is BatchTask:
         return _monitor_modeler_batch(task_id, verbose=verbose)
 
     console = get_logging_console() if verbose else None
@@ -1147,32 +1147,26 @@ def load(
         Object containing simulation data.
     """
     path = Path(path)
-    task = TaskFactory.get(task_id) if task_id else None
+    from_cache = task_id is None  # for readability
     # For component modeler batches, default to a clearer filename if the default was used.
-    if (
-        task_id
-        and isinstance(task, BatchTask)
-        and path.name in {"simulation_data.hdf5", "simulation_data.hdf5.gz"}
-    ):
-        path = path.with_name(path.name.replace("simulation", "cm"))
+    if not from_cache and path.name in {"simulation_data.hdf5", "simulation_data.hdf5.gz"}:
+        if TaskFactory.get_kind(task_id) is BatchTask:
+            path = path.with_name(path.name.replace("simulation", "cm"))
 
-    if task_id is None:
+    if from_cache:
         if not path.exists():
             raise FileNotFoundError("Cached file not found.")
     elif not path.exists() or replace_existing:
         download(task_id=task_id, path=path, verbose=verbose, progress_callback=progress_callback)
 
-    if verbose and task_id is not None:
+    if verbose and not from_cache:
         console = get_logging_console()
-        if isinstance(task, BatchTask):
-            console.log(f"Loading component modeler data from {path}")
-        else:
-            console.log(f"Loading simulation from {path}")
+        console.log(f"Loading results from {path}")
 
     stub_data = Tidy3dStubData.postprocess(path, lazy=lazy)
 
     simulation_cache = resolve_local_cache()
-    if simulation_cache is not None and task_id is not None:
+    if simulation_cache is not None and not from_cache:
         info = get_info(task_id, verbose=False)
         workflow_type = getattr(info, "taskType", None)
         if (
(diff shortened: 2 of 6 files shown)


tidy3d/web/api/webapi.py
```diff
@@ -1148,32 +1148,26 @@ def load(
...
         download(task_id=task_id, path=path, verbose=verbose, progress_callback=progress_callback)
 
-    if verbose and task_id is not None:
+    if verbose and not from_cache:
         console = get_logging_console()
-        if isinstance(task, BatchTask):
-            console.log(f"Loading component modeler data from {path}")
-        else:
-            console.log(f"Loading simulation from {path}")
+        console.log(f"Loading results from {path}")
```
marcorudolphflex, line 1164: Changed that to this generic log. Was there a strong reason to have the information in more detail?

tidy3d/web/api/container.py
```diff
@@ -622,8 +622,6 @@ def load_sim_data(self, task_name: str) -> WorkflowDataType:
         task_data_path = Path(self.task_paths[task_name])
         task_id = self.task_ids[task_name]
         from_cache = self.cached_tasks[task_name] if self.cached_tasks else False
-        if not from_cache:
-            web.get_info(task_id)
```
marcorudolphflex, line 686: Could not find a reason to do this and got no problems in online tests after removing it. Does anybody know a good reason why we did this here?

marcorudolphflex: @momchil-flex do you know something about this? Maybe it is related to https://github.com/flexcompute/tidy3d/pull/3194#discussion_r2716576978 ?
But as we call estimate cost for batches, `get_info` should be triggered anyways.

marcorudolphflex: the link does not really work? :D 
@daquinteroflex  said:

> So the issue with this is that if verbose is false, it should still run estimate_cost to trigger the metadata

momchil-flex: No, I can't think of a reason for this and shouldn't be related to the estimate cost comment. `get_info` shouldn't have any side effects.

momchil-flex (commented):
Thanks, note that this was also achieved in [my PR that we decided to postpone]. That PR is much more invasive, but the approach it used to avoid repeated api calls for batch checking is still an alternative to the approach you've taken here. Namely, instead of caching the kind, we use the `WebTask.task_type` (which now every task should have) and enumerate [which task types are batch kind](https://github.com/flexcompute/tidy3d/pull/3183/changes#diff-2f024019772f548396269e338b610c794d176828a11efc814eeed37abe1efa40R402).

Currently, there is no way to introduce a new task type without modification both to the code here and to the server code. I wonder how we might want this to look like eventually? Should the task types be defined here only? And if so should the distinction between batch and regular also be defined here?

marcorudolphflex:
> Thanks, note that this was also achieved in [my PR that we decided to postpone]. That PR is much more invasive, but the approach it used to avoid repeated api calls for batch checking is still an alternative to the approach you've taken here. Namely, instead of caching the kind, we use the `WebTask.task_type` (which now every task should have) and enumerate [which task types are batch kind](https://github.com/flexcompute/tidy3d/pull/3183/changes#diff-2f024019772f548396269e338b610c794d176828a11efc814eeed37abe1efa40R402).
> 
> Currently, there is no way to introduce a new task type without modification both to the code here and to the server code. I wonder how we might want this to look like eventually? Should the task types be defined here only? And if so should the distinction between batch and regular also be defined here?

That's what I realized at the very end, initially I assumed that to be already merged.
Maybe we can mix some concepts of the PRs?

Not sure how to handle this. Couldn't we also just let the server define this and get this batch kind information as response?

momchil-flex:
> > Thanks, note that this was also achieved in [my PR that we decided to postpone]. That PR is much more invasive, but the approach it used to avoid repeated api calls for batch checking is still an alternative to the approach you've taken here. Namely, instead of caching the kind, we use the `WebTask.task_type` (which now every task should have) and enumerate [which task types are batch kind](https://github.com/flexcompute/tidy3d/pull/3183/changes#diff-2f024019772f548396269e338b610c794d176828a11efc814eeed37abe1efa40R402).
> > Currently, there is no way to introduce a new task type without modification both to the code here and to the server code. I wonder how we might want this to look like eventually? Should the task types be defined here only? And if so should the distinction between batch and regular also be defined here?
> 
> That's what I realized at the very end, initially I assumed that to be already merged. Maybe we can mix some concepts of the PRs?
> 
> Not sure how to handle this. Couldn't we also just let the server define this and get this batch kind information as response?

Yes, I'm fine with that.

I also don't necessarily need to merge my PR, it could be left for a bigger refactor some day.

momchil-flex (approved):

daquinteroflex:
Do you think it's possible to verify this with at least:
* one RF notebook (the GroundCPW one is pretty fast)
* one Charge notebook or one Heat notebook (like ThermalRingResonator) should run FDTD too

Just because the webapi is really not well tested/covered at all. Any change impact is more visible in the notebooks

marcorudolphflex:
> Do you think it's possible to verify this with at least:
> 
> * one RF notebook (the GroundCPW one is pretty fast)
> * one Charge notebook or one Heat notebook (like ThermalRingResonator) should run FDTD too
> 
> Just because the webapi is really not well tested/covered at all. Any change impact is more visible in the notebooks

Good idea!
I just did that and the Jobs were successful.
Still I got an unrelated error after loading the data which I also observed in develop:
```
14:29:02 CET Loading component modeler data from data/tcm_data_exp.hdf5         
             ERROR: If the 'ModeSimulation' geometry is not planar, then 'plane'
             must be specified.                                                 
             ERROR: If the 'ModeSimulation' geometry is not planar, then 'plane'
             must be specified.   
```
Will create a ticket for that.

momchil-flex:
Sorry. accidentally closed, reopening.

PR #3234, opened 2026-02-04
feat(tidy3d): FXC-5311-enable-cached-loading-from-batch-data

marcorudolphflex:
**What changed**

- Added in-memory caching for `BatchData` access when results are already downloaded and the total size of all batch data files stays under a configurable threshold.
- Introduced `config.batch_data_cache.max_total_size_gb` (with enabled) to control this behavior.
- Updated docs and changelog to document the new configuration.
- Added tests for cache-hit and cache-miss behavior based on total batch size.

**Why**

- Avoids repeated `web.load` calls when users iterate or index batch results multiple times, while preventing large batches from staying in memory.

---

> [!NOTE]
> **Medium Risk**
> Changes data-loading behavior in `BatchData` and adds stateful caching, which could impact memory usage and freshness if edge cases aren’t covered; guarded by a size threshold and explicit config flags.
> 
> **Overview**
> Speeds up repeated access to `td.web.BatchData` results by **optionally caching loaded task data objects in memory** when the batch’s local result files are already present and their combined size is below a configurable threshold.
> 
> Introduces a new config section `config.batch_data_cache` (with `enabled` and `max_total_size_gb`) and wires `BatchData.load_sim_data()` / `__getitem__` to reuse cached objects when eligible, while explicitly disabling the behavior for WASM builds and for large/non-downloaded batches. Documentation, changelog, and new unit tests cover cache-hit vs cache-miss behavior based on total file size.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4ad9e26c45fcc135df270b66cd5fa9edc660d69d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tidy3d/config/sections.py
+++ b/tidy3d/config/sections.py
@@ -517,6 +517,25 @@ class LocalCacheConfig(ConfigSection):
         return str(value)
 
 
+class BatchDataCacheConfig(ConfigSection):
+    """Settings controlling in-memory caching for batch data."""
+
+    enabled: bool = Field(
+        True,
+        title="Enable batch data cache",
+        description="Cache batch results in memory when files are below the size threshold.",
+    )
+
+    max_total_size_gb: NonNegativeFloat = Field(
+        1.0,
+        title="Maximum total batch data size (GB)",
+        description=(
+            "Cache batch task data only when the combined size of all task data files is at or "
+            "below this threshold. Set to 0 to disable."
+        ),
+    )
+
+
 @register_section("plugins")
 class PluginsContainer(ConfigSection):
     """Container that holds plugin-specific configuration sections."""
@@ -530,10 +549,12 @@ if not WASM_BUILD:
     register_section("web")(WebConfig)
     register_handler("web")(apply_web)
     register_section("local_cache")(LocalCacheConfig)
+    register_section("batch_data_cache")(BatchDataCacheConfig)
 
 
 __all__ = [
     "AdjointConfig",
+    "BatchDataCacheConfig",
     "LocalCacheConfig",
     "LoggingConfig",
     "MicrowaveConfig",
--- a/tidy3d/web/api/container.py
+++ b/tidy3d/web/api/container.py
@@ -18,10 +18,12 @@ from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 from pydantic import Field, PositiveInt, PrivateAttr, model_validator
 from rich.progress import BarColumn, Progress, TaskProgressColumn, TextColumn, TimeElapsedColumn
 
+from tidy3d._runtime import WASM_BUILD
 from tidy3d.components.base import Tidy3dBaseModel, cached_property
 from tidy3d.components.mode.mode_solver import ModeSolver
 from tidy3d.components.types.base import discriminated_union
 from tidy3d.components.types.workflow import WorkflowType
+from tidy3d.config import config
 from tidy3d.exceptions import DataError
 from tidy3d.log import get_logging_console, log
 from tidy3d.web.api import webapi as web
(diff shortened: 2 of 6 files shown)


tidy3d/config/sections.py
```diff
@@ -514,6 +514,27 @@ def _serialize_directory(self, value: Path) -> str:
...
+    """Settings controlling in-memory caching for batch data."""
+
+    enabled: bool = Field(
+        True,
+        title="Enable batch data cache",
+        description="Cache batch results in memory when files are below the size threshold.",
+        json_schema_extra={"persist": True},
+    )
+
+    max_total_size_gb: NonNegativeFloat = Field(
```
yaugenst-flex, line 529: I guess in principle this could be more automatic via some fraction of free memory?

marcorudolphflex: Would make sense. But do we want to add psutil as a non-optional dependency for that? At least it's not heavy.

marcorudolphflex: [will leave it like this as discussed]

tidy3d/config/sections.py
```diff
@@ -517,6 +517,27 @@ def _serialize_directory(self, value: Path) -> str:
...
 
 
+class BatchDataCacheConfig(ConfigSection):
+    """Settings controlling in-memory caching for batch data."""
+
+    enabled: bool = Field(
+        True,
+        title="Enable batch data cache",
+        description="Cache batch results in memory when files are below the size threshold.",
+        json_schema_extra={"persist": True},
```
yaugenst-flex: Let's not add persisted fields right now, at least not until we have the migrations properly sorted.

tidy3d/config/sections.py
```diff
@@ -517,6 +517,27 @@ def _serialize_directory(self, value: Path) -> str:
...
+    )
+
+    max_total_size_gb: NonNegativeFloat = Field(
+        1.0,
+        title="Maximum total batch data size (GB)",
+        description=(
+            "Cache batch task data only when the combined size of all task data files is at or "
+            "below this threshold. Set to 0 to disable."
+        ),
+        json_schema_extra={"persist": True},
```
yaugenst-flex: Ditto.

tidy3d/config/sections.py
```diff
@@ -530,10 +551,12 @@ class PluginsContainer(ConfigSection):
     register_section("web")(WebConfig)
     register_handler("web")(apply_web)
     register_section("local_cache")(LocalCacheConfig)
+    register_section("batch_data_cache")(BatchDataCacheConfig)
```
yaugenst-flex, line 552: Wonder if we should just have a caching section?

tidy3d/web/api/container.py
```diff
@@ -629,24 +631,80 @@ class BatchData(Tidy3dBaseModel, Mapping):
         description="Whether the simulation data was downloaded before.",
     )
 
+    _data_cache: dict[TaskName, WorkflowDataType] = PrivateAttr(default_factory=dict)
+    _cache_enabled: Optional[bool] = PrivateAttr(default=None)
```
yaugenst-flex, line 635: I feel a bit uneasy about these, but I understand they are probably necessary. As long they don't in any way affect serialization or hashing?

marcorudolphflex: made a test, does not affect hashing (both `hash` and `_hash_self`) and serialization

tidy3d/config/sections.py
marcorudolphflex: [leave it for now as discussed]

yaugenst-flex (approved):
Nice!

PR #3237, opened 2026-02-04
fix(adjoint): fixes for CustomMedium and FieldMonitor gradients

groberts-flex:
- handle numerical precision errors at CustomMedium boundaries
- correct adjoint grid_spec to fully align grid between forward and adjoint simulations

Here are the fixes to get the custom medium numerical test working again. I'm running the full test for it right now including both interpolation types in the custom medium.

---

> [!NOTE]
> **Medium Risk**
> Touches adjoint simulation construction and `CustomMedium` gradient cropping, which can change gradient values and convergence behavior; test coverage is expanded to reduce regression risk but downstream optimization workflows may be sensitive to these numerical changes.
> 
> **Overview**
> Fixes autograd gradient regressions for `CustomMedium`/`FieldData` by **forcing adjoint simulations to reuse the exact forward simulation grid** (`GridSpec.from_grid(sim_original.grid)`) rather than relying on wavelength-based remeshing.
> 
> Hardens `CustomMedium` parameter-gradient computation against numerical precision at geometry boundaries by adding a small coordinate tolerance when cropping field data, and avoids mutating cached derivative arrays by copying values before slicing.
> 
> Updates the numerical autograd regression test to hold the grid fixed across forward and finite-difference runs and to run coverage for both `CustomMedium` interpolation modes (`nearest`/`linear`), with optional improved plotting output. Changelog entries are added for both fixes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d9f967836c73f83572197b6eda06f00d9e364e9a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/test_components/autograd/numerical/test_autograd_numerical.py
+++ b/tests/test_components/autograd/numerical/test_autograd_numerical.py
@@ -117,7 +117,21 @@ def make_base_sim(
     return sim_base
 
 
-def create_objective_function(geometry, create_sim_base, eval_fn, sim_path_dir):
+def create_objective_function(
+    geometry, create_sim_base, eval_fn, sim_path_dir, perm_init, cm_interp_method
+):
+    block_structure = td.Structure.from_permittivity_array(
+        eps_data=perm_init,
+        geometry=geometry,
+    )
+
+    sim_base = create_sim_base()
+
+    sim_with_block = sim_base.updated_copy(structures=(*sim_base.structures, block_structure))
+
+    # use a fixed grid for all forward and finite difference simulations
+    grid_fixed = sim_with_block.grid
+
     def objective(perm_arrays):
         sim_base = create_sim_base()
 
@@ -128,8 +142,13 @@ def create_objective_function(geometry, create_sim_base, eval_fn, sim_path_dir):
                 geometry=geometry,
             )
 
+            block_structure = block_structure.updated_copy(
+                medium=block_structure.medium.updated_copy(interp_method=cm_interp_method)
+            )
+
             sim_with_block = sim_base.updated_copy(
-                structures=(*sim_base.structures, block_structure)
+                structures=(*sim_base.structures, block_structure),
+                grid_spec=td.GridSpec.from_grid(grid_fixed),
             )
 
             simulation_dict[f"numerical_field_testing_{idx}"] = sim_with_block.copy()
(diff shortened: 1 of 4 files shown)


tidy3d/config/sections.py
```diff
@@ -179,6 +180,15 @@ class AdjointConfig(ConfigSection):
         ),
     )
 
+    custom_medium_boundary_coordinate_tolerance: PositiveFloat = Field(
```
marcorudolphflex: Not sure if we really need another config setting for that instead of a constant. Do we really expect that 1e-12 could be too big or too small for some cases?

tests/test_components/autograd/numerical/test_autograd_numerical.py
```diff
@@ -332,12 +358,23 @@ def test_finite_difference_field_data(
...
     print(f"FD, Adj magnitudes: {fd_mag}, {adj_mag}")
     print(f"Percentage Error: {percentage_error}")
     print("-" * 20)
     print("\n" * 3)
 
+    if PLOT_FD_ADJ_COMPARISON:
+        plt.plot(pattern_dot_adj_gradient, color="g", linewidth=2.0)
+        plt.plot(fd_grad, color="b", linewidth=1.5, linestyle="--")
+        plt.title(f"Gradient for objective: {eval_fn_name}")
+        plt.legend(["Finite difference", "Adjoint"])
```
marcorudolphflex: I think that didn't work in the plots? Just use `labels=...` in `plt.plot` and then `plt.legend()`?

marcorudolphflex (commented):
Looks great overall, thanks! Just some minor comments.

tidy3d/config/sections.py
groberts-flex: good point, I had the same thought and wasn't sure if I should put in the config or not. I think you're right, though, that nobody would need to really configure this. I'll move it back to a constant in the file!

tests/test_components/autograd/numerical/test_autograd_numerical.py
groberts-flex: I had an extra plt.legend() call after this for some reason, but changed to label anyway since I think that looks a little nicer! thanks for the catch

marcorudolphflex (approved):

PR #3250, opened 2026-02-06
FXC-4546: add baseband source time classes for transient simulations

dmarek-flex:
> [!NOTE]
> **Medium Risk**
> Touches core source-time typing and behavior (including `amp_time` return types and validation semantics) which could affect downstream code expecting previous scalar/None behaviors. New baseband types also expand schema unions, so compatibility depends on consumers handling the updated discriminator options.
> 
> **Overview**
> Adds **baseband (unmodulated) `SourceTime` profiles** for transient RF workflows: `BasebandStep`, `BasebandGaussianPulse`, `BasebandRectangularPulse`, and `BasebandCustomSourceTime`, including dataset-based interpolation, FFT-derived `frequency_range`, and phase locked to `0`.
> 
> Wires these new types into JSON schemas (`Simulation`, `ModeSimulation`, `TerminalComponentModeler`) and exports them via `tidy3d.__init__` and `tidy3d.rf`, while centralizing `SourceTimeType` into a new `components/types/time.py` and updating imports accordingly.
> 
> Tightens and standardizes source-time behavior: `amp_time` now accepts scalar or array inputs and returns arrays, `CustomSourceTime` validates strictly increasing time coordinates, rejects non-positive `dt`, raises `SetupError` when datasets are missing, and returns `None` for `end_time` on all-zero data; extensive new tests cover baseband sources and these edge cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b7d872572408e83dc61e0e177a5f5afcb125b911. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tidy3d/components/microwave/time.py
+++ b/tidy3d/components/microwave/time.py
@@ -1,0 +1,350 @@
+"""Baseband (unmodulated) source time classes for transient RF simulations."""
+
+from __future__ import annotations
+
+from abc import ABC
+from typing import TYPE_CHECKING, Any, Literal, Optional
+
+import numpy as np
+from pydantic import Field, PositiveFloat, field_validator
+
+from tidy3d.components.base import cached_property
+from tidy3d.components.data.data_array import TimeDataArray
+from tidy3d.components.data.dataset import TimeDataset
+from tidy3d.components.data.validators import validate_no_nans
+from tidy3d.components.source.time import DEFAULT_SIGMA, END_TIME_FACTOR_GAUSSIAN, SourceTime
+from tidy3d.components.validators import warn_if_dataset_none
+from tidy3d.exceptions import SetupError, ValidationError
+
+if TYPE_CHECKING:
+    from typing import Union
+
+    from tidy3d.components.types import ArrayFloat1D, FreqBound
+
+# Factor converting 10%-to-90% rise time to Gaussian sigma (standard deviation).
+# rise_time = RISE_TIME_10_90_FACTOR * sigma
+# Derived from: 2 * sqrt(2) * erfinv(0.8) ≈ 2.5631
+RISE_TIME_10_90_FACTOR = 2.5631031310892006
+
+
+class BasebandSourceTime(SourceTime, ABC):
+    """Abstract base class for baseband (unmodulated) source time profiles.
+
+    These source times have no carrier frequency and are intended for transient
+    RF simulations (TDR, step excitations, arbitrary time-domain profiles).
+    The ``phase`` parameter is locked to 0 since baseband signals have no carrier.
+    """
+
+    phase: Literal[0] = Field(
+        0,
+        frozen=True,
+        title="Phase",
+        description="Phase is locked to 0 for baseband source times (no carrier frequency).",
+    )
+
+
+class BasebandStep(BasebandSourceTime):
+    """Step function source time profile using an error function (erf).
+
+    The signal ramps from 0 to ``amplitude`` with ``rise_time`` specifying the
+    10%-to-90% rise time.
+
+    Example
+    -------
+    >>> step = BasebandStep(rise_time=1e-9)
+    """
+
+    rise_time: PositiveFloat = Field(
+        title="Rise Time",
+        description="The 10%-to-90% rise time in seconds. "
+        "Related to the Gaussian standard deviation (sigma) by ``rise_time ≈ 2.563 * sigma``.",
+    )
+
+    offset: float = Field(
+        5.0,
+        title="Offset",
+        description="Delay of the step center in units of ``rise_time``.",
+        ge=2.5,
+    )
+
+    @cached_property
+    def _sigma(self) -> float:
+        """Gaussian standard deviation derived from the 10-90% rise time."""
+        return self.rise_time / RISE_TIME_10_90_FACTOR
+
+    @cached_property
+    def t_center(self) -> float:
+        """Time of the step center in seconds."""
+        return self.offset * self.rise_time
+
+    @cached_property
+    def fwidth_equiv(self) -> float:
+        """Equivalent frequency-domain standard deviation (sigma_f) in Hz.
+
+        Defined as the frequency where the Gaussian spectral envelope drops
+        to ``exp(-1/2)`` of its peak value."""
+        return 1.0 / (2 * np.pi * self._sigma)
+
+    def amp_time(self, time: Union[float, ArrayFloat1D]) -> ArrayFloat1D:
+        """Real-valued source amplitude as a function of time."""
+        from scipy.special import erf
+
+        time = np.atleast_1d(np.asarray(time, dtype=float))
+        return (
+            self.amplitude * 0.5 * (1.0 + erf((time - self.t_center) / (np.sqrt(2) * self._sigma)))
+        )
+
+    def frequency_range(self, num_fwidth: float = DEFAULT_SIGMA) -> FreqBound:
+        """Frequency range based on equivalent bandwidth."""
+        return (0.0, num_fwidth * self.fwidth_equiv)
+
+    def end_time(self) -> Optional[float]:
+        """Step never decays, so no end time."""
+        return None
+
+
+class BasebandGaussianPulse(BasebandSourceTime):
+    """Unmodulated Gaussian pulse source time profile.
+
+    The signal is a Gaussian envelope centered at ``offset * twidth`` with
+    temporal width ``twidth``.
+
+    Example
+    -------
+    >>> gauss = BasebandGaussianPulse(twidth=1e-9)
+    """
+
+    twidth: PositiveFloat = Field(
+        title="Temporal Width",
+        description="Temporal width (standard deviation) of the Gaussian pulse in seconds.",
+    )
+
+    offset: float = Field(
+        5.0,
+        title="Offset",
+        description="Delay of the pulse peak in units of ``twidth``.",
+        ge=2.5,
+    )
+
+    @cached_property
+    def t_center(self) -> float:
+        """Time of the pulse peak in seconds."""
+        return self.offset * self.twidth
+
+    @cached_property
+    def fwidth(self) -> float:
+        """Standard deviation of the frequency content of the Gaussian pulse."""
+        return 1.0 / (2 * np.pi * self.twidth)
+
+    def amp_time(self, time: Union[float, ArrayFloat1D]) -> ArrayFloat1D:
+        """Real-valued source amplitude as a function of time."""
+        time = np.atleast_1d(np.asarray(time, dtype=float))
+        return self.amplitude * np.exp(-((time - self.t_center) ** 2) / (2 * self.twidth**2))
+
+    def frequency_range(self, num_fwidth: float = DEFAULT_SIGMA) -> FreqBound:
+        """Frequency range based on equivalent bandwidth."""
+        return (0.0, num_fwidth * self.fwidth)
+
+    def end_time(self) -> Optional[float]:
+        """Time after which the source is effectively zero."""
+        return self.t_center + END_TIME_FACTOR_GAUSSIAN * self.twidth
+
+
+class BasebandRectangularPulse(BasebandSourceTime):
+    """Smoothed rectangular pulse source time profile.
+
+    The signal is a flat-top pulse constructed from two error functions,
+    producing a smooth rise and fall. The ``rise_time`` parameter specifies
+    the 10%-to-90% rise/fall time.
+
+    Example
+    -------
+    >>> pulse = BasebandRectangularPulse(rise_time=1e-10, twidth=1e-9)
+    """
+
+    rise_time: PositiveFloat = Field(
+        title="Rise Time",
+        description="The 10%-to-90% rise/fall time in seconds. "
+        "Related to the Gaussian standard deviation (sigma) by ``rise_time ≈ 2.563 * sigma``.",
+    )
+
+    twidth: PositiveFloat = Field(
+        title="Temporal Width",
+        description="Duration of the flat-top region in seconds.",
+    )
+
+    offset: float = Field(
+        5.0,
+        title="Offset",
+        description="Delay of the pulse start (rising edge center) in units of ``rise_time``.",
+        ge=2.5,
+    )
+
+    @cached_property
+    def _sigma(self) -> float:
+        """Gaussian standard deviation derived from the 10-90% rise time."""
+        return self.rise_time / RISE_TIME_10_90_FACTOR
+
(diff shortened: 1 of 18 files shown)


dmarek-flex:
Here are some nice pictures of the new baseband signals
demo_baseband

dmarek-flex:
This one is basically ready to go. I opted for creating new classes to handles these signals with DC components, instead of modifying the existing ones to support `freq0=0`, I think the distinction is clearer this way. But open to your thoughts.

tidy3d/components/microwave/time.py
```diff
@@ -0,0 +1,308 @@
...
+    Example
+    -------
+    >>> step = BasebandStep(rise_time=1e-9)
+    """
+
+    rise_time: PositiveFloat = Field(
+        title="Rise Time",
+        description="Characteristic rise time in seconds. "
+        "The 10%-to-90% rise time is approximately 2.56 times this value.",
+    )
```
George-Guryev-flxcmp, line 61: I believe the variable ``rise_time`` is misleading in this context: the rise time is typically a time it takes to go from 10% to 90%, while the variable ``rise_time`` is used here as ``sigma`` in step function below. The actual rise time is approximately 2.56 * ``sigma``.  I wonder one can rename ``rise_time`` to ``standard_deviation`` or something similar? Or alternatively use the actual rise time for this variable. 

Image

tidy3d/components/microwave/time.py
```diff
@@ -0,0 +1,308 @@
...
+class BasebandSourceTime(SourceTime, ABC):
+    """Abstract base class for baseband (unmodulated) source time profiles.
+
+    These source times have no carrier frequency and are intended for transient
+    RF simulations (TDR, step excitations, arbitrary time-domain profiles).
+    The ``phase`` parameter is locked to 0 since baseband signals have no carrier.
+    """
+
+    phase: Literal[0] = Field(
+        0,
```
George-Guryev-flxcmp, line 39: I guess the default nominal value makes sense. However, I wonder if one could have a scenario where a simulation has multiple ports (say two) whose sources/waveforms should have a constant phase offset to each other? One can surely do phase scaling in post-processing, however, I wonder if you think this might be a useful behavior here (i.e. default to ``0`` but allow for any constant phase offset for a given waveform?)

tidy3d/components/microwave/time.py
```diff
@@ -0,0 +1,317 @@
...
+
+    @cached_property
+    def t_center(self) -> float:
+        """Time of the step center in seconds."""
+        return self.offset * self.rise_time
+
+    @cached_property
+    def fwidth_equiv(self) -> float:
+        """Equivalent bandwidth in Hz."""
+        return 1.0 / (2 * np.pi * self.rise_time)
```
George-Guryev-flxcmp: I believe this uses the RMS bandwidth; however, multiple alternative definitions exist. Making this explicit in the docstring would remove any ambiguity.

tidy3d/components/microwave/time.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def amp_time(self, time: float) -> complex:
```
George-Guryev-flxcmp: All `amp_time()` methods give a type hint `float` for `time`. However, in most cases `time` is treated as a numpy array. I wonder you could use `time: Union[float, ArrayLike]` or something similar?

George-Guryev-flxcmp (commented):
Thanks, Damian — this looks solid! I have a few minor comments and suggestions.

tidy3d/components/microwave/time.py
```diff
@@ -0,0 +1,317 @@
...
+    """Abstract base class for baseband (unmodulated) source time profiles.
+
+    These source times have no carrier frequency and are intended for transient
+    RF simulations (TDR, step excitations, arbitrary time-domain profiles).
+    The ``phase`` parameter is locked to 0 since baseband signals have no carrier.
+    """
+
+    phase: Literal[0] = Field(
+        0,
+        frozen=True,
```
weiliangjin2021, line 40: Is this needed?

tidy3d/components/microwave/time.py
dmarek-flex: For phase to be meaningful it requires like a carrier frequency. I have time offsets in all but the CustomSourceTime, which I think is the best that we can do for this kind of signal.

George-Guryev-flxcmp: Good point. I guess I was thinking about the envelope of the modulated signal and this feature as a simplification for analysis of RF signals in a rotating frame. But these are not envelopes but the actual signals, so yes, you're right.

tidy3d/components/microwave/time.py
dmarek-flex: FOr the field I will stay with rise_time, but make it correspond with the actualy 10%-90% rise time,

tidy3d/components/microwave/time.py
dmarek-flex: Let me know if the new description is clearer

George-Guryev-flxcmp: Looks great to me!

George-Guryev-flxcmp (approved):

tidy3d/components/source/time.py
```diff
@@ -76,6 +78,48 @@ def plot_spectrum(
...
+        carrier).
+
+        Parameters
+        ----------
+        dt : float
+            Uniform time step between samples.
+        values : np.ndarray
+            Real-valued signal samples.
+        num_fwidth : float
+            Number of equivalent bandwidths for the cutoff threshold.
```
weiliangjin2021, line 96: this is the sigma? Would be good to add a formula here for clarity

tidy3d/components/source/time.py
```diff
@@ -76,6 +78,48 @@ def plot_spectrum(
...
+        -------
+        Tuple[float, float]
+            ``(fmin, fmax)`` frequency range.
+        """
+        spectrum = np.abs(np.fft.rfft(np.real(values)))
+        freqs = np.fft.rfftfreq(len(values), d=dt)
+
+        peak = np.max(spectrum)
+        if peak == 0:
+            return (0.0, 1.0)
```
weiliangjin2021: maybe return (freqs[0], freqs[-1]), as the frequency 0 Hz and 1Hz is weird.

tidy3d/components/source/time.py
```diff
@@ -76,6 +78,48 @@ def plot_spectrum(
...
+
+        cutoff = np.exp(-(num_fwidth**2) / 2)
+        above_cutoff = spectrum >= cutoff * peak
+        if not np.any(above_cutoff):
+            return (0.0, freqs[-1])
```
weiliangjin2021: why not freqs[0] for lower range.

tidy3d/components/source/time.py
```diff
@@ -76,6 +78,48 @@ def plot_spectrum(
...
+
+        indices = np.where(above_cutoff)[0]
+        fmin = float(freqs[indices[0]])
+        fmax = float(freqs[indices[-1]])
+        # Ensure at least one frequency bin of bandwidth
+        fmax = max(fmax, fmin + freqs[1])
```
weiliangjin2021: should this be max(fmax, freqs[1])?

tidy3d/components/microwave/time.py
```diff
@@ -0,0 +1,340 @@
...
+    The signal is defined by a ``source_time_dataset`` containing the time-domain
+    envelope.
+
+    Example
+    -------
+    >>> cst = BasebandCustomSourceTime.from_values(values=np.linspace(0, 1, 100), dt=1e-10)
+    """
+
+    source_time_dataset: Optional[TimeDataset] = Field(
+        None,
```
weiliangjin2021, line 236: what does it mean when it's None?

tidy3d/components/microwave/time.py
```diff
@@ -0,0 +1,340 @@
...
+        envelope[mask] = values.sel(t=times[mask], method="nearest").to_numpy()
+        if not all(mask):
+            envelope[~mask] = values.interp(t=times[~mask]).to_numpy()
+
+        return self.amplitude * envelope
+
+    def frequency_range(self, num_fwidth: float = DEFAULT_SIGMA) -> FreqBound:
+        """Frequency range computed from FFT of the source envelope."""
+        if self.source_time_dataset is None:
+            return (0.0, 1.0)
```
weiliangjin2021: again, I find (0Hz, 1Hz) confusing

tidy3d/components/microwave/time.py
```diff
@@ -0,0 +1,340 @@
...
+        return cls(source_time_dataset=source_time_dataset, **kwargs)
+
+    @property
+    def data_times(self) -> ArrayFloat1D:
+        """Times of envelope definition."""
+        if self.source_time_dataset is None:
+            return []
+        return self.source_time_dataset.values.coords["t"].values.squeeze()
+
+    def amp_time(self, time: Union[float, ArrayFloat1D]) -> Union[float, ArrayFloat1D]:
```
weiliangjin2021: A comment from Claude, quite minor though

> scalar input returns 1-d array, not 0-d array
>   (microwave/time.py:295)
> 
>   times = np.array([time] if isinstance(time, (int, float)) else time)
> 
>   This wraps a scalar in a list, so amp_time(0.5e-9) returns a 1-element 1-d array. The other
>   baseband classes use np.asarray(time) which preserves 0-d arrays for scalar input. The test even
>   reveals this inconsistency:
> 
>   amp_mid = cst.amp_time(0.5e-9)
>   assert np.isclose(np.real(amp_mid[0]), 0.25, ...)  # needs [0] indexing
> 
>   while other classes would give a scalar-like 0-d result. Should unify to handle scalars
>   consistently (return a scalar-like value for scalar input, array for array input).
>

tidy3d/components/microwave/time.py
```diff
@@ -0,0 +1,340 @@
...
+
+        data_times = self.data_times
+        values = self.source_time_dataset.values.to_numpy().squeeze()
+        dts = np.diff(data_times)
+        dt = float(np.min(dts))
+
+        # Resample to uniform grid if times are non-uniform
+        if not np.allclose(dts, dt):
+            uniform_times = np.arange(data_times[0], data_times[-1] + dt * 0.5, dt)
+            values = np.interp(uniform_times, data_times, np.real(values))
```
weiliangjin2021: from Claude

>   values = np.interp(uniform_times, data_times, np.real(values))
> 
>   When times are non-uniform, np.real(values) is used for resampling, but then np.real(values) is
>   also passed on line 325. If the user provides complex-valued data, the imaginary part is silently
>   dropped with no warning. The from_values docstring says values: ArrayComplex1D, suggesting complex
>    data is expected. Either warn or document that only the real part is used for frequency range
>   estimation.
>

weiliangjin2021 (approved):
Nice work. Just some minor comments.

tidy3d/components/microwave/time.py
dmarek-flex: `Literal[0]` validates at construction, but `frozen=True` prevents post-creation mutation (e.g. `obj.phase = 5`). Defensive but can remove if you prefer.

tidy3d/components/source/time.py
dmarek-flex: Added the formula to the docstring: `amplitude >= exp(-num_fwidth**2 / 2) * peak`.

tidy3d/components/source/time.py
dmarek-flex: Good point — changed to `(freqs[0], freqs[-1])` so the fallback is data-driven rather than a magic constant.

tidy3d/components/source/time.py
dmarek-flex: Changed to `freqs[0]` for consistency (same value for rfft, but cleaner).

tidy3d/components/source/time.py
dmarek-flex: This is intentional — `fmin + freqs[1]` ensures at least one frequency bin of bandwidth *above fmin*. If fmin > 0 (no DC), using just `freqs[1]` would not guarantee that.

tidy3d/components/microwave/time.py
dmarek-flex: Added clarification to the field description: "If None, the source produces no signal and must be populated before use."

tidy3d/components/microwave/time.py
dmarek-flex: Added a log warning for the None-dataset case. The (0, 1) is still returned as a placeholder since we have no data to compute from, but the warning makes it explicit.

tidy3d/components/microwave/time.py
dmarek-flex: Fixed — now uses `np.atleast_1d(np.asarray(time))` and returns `result.squeeze()` for scalar input, matching the behavior of the other baseband classes.

tidy3d/components/microwave/time.py
dmarek-flex: Good catch — added a note in the docstring that only the real part is used for frequency estimation. Also cleaned up the code to take `np.real()` once at the top rather than in two places.

tidy3d/components/microwave/time.py
dmarek-flex: Some of the behavior I copied from the modulated versions of these signals. `CustomSourceTime` has this field.

dmarek-flex: I'll remove this optional behavior of the field though since I dont know why it is supported.

tidy3d/components/source/time.py
dmarek-flex: I added an explicit df calculation to make it clearer

tidy3d/components/microwave/time.py
dmarek-flex: Ok this was a rabbit hole, basically though the return type will always np.ndarray (either 0D or 1D) so we just hint it as ArrayFloat1D.

tidy3d/components/microwave/time.py
dmarek-flex: AH actually it is needed because of how the data is serialized/deserialized from HDF5

weiliangjin2021: ah, I recall that now!

tidy3d/components/microwave/time.py
dmarek-flex: Yea I will switch the behavior here and in the existing code to raise an error when the source_time_dataset is None

PR #3252, opened 2026-02-06
chore: Improve local subpixel warnings and docstrings

caseyflex:
Fixes unclear warnings / docstrings reported by Tyler.
Related FAQ PR: https://github.com/flexcompute/tidy3d-faq/pull/60

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to warning emission and docstrings, with lightweight test coverage; minimal risk to solver correctness aside from potential warning visibility changes.
> 
> **Overview**
> Updates local mode-solving behavior so `ModeSolver.solve()` is wrapped with `@supports_local_subpixel` and **only emits the accuracy warning when local subpixel averaging is not active**, reducing noise for users running with `tidy3d-extras`.
> 
> Expands/clarifies documentation for `config.simulation.use_local_subpixel` across `Simulation.epsilon*`, `ModeSimulation.run_local()`, `ModeSolver.solve()`, the configuration reference, and the extras docs, including install guidance and the `True`/`False`/`None` semantics. Adds regression tests to verify the warning suppression/emission behavior under different `tidy3d-extras` availability/config settings, and records the change in the changelog.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 290e288779cb5217c15ffe3fd170c8abf36f457d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed redundant server lookups when loading simulation results.
 - Updated docstrings for `DerivativeInfo` to more accurately reflect dataclass fields.
 - Fixed local cache race conditions causing `FileNotFoundError`.
+- Improved `ModeSolver.solve()` to suppress the accuracy warning when local subpixel averaging is enabled via `tidy3d-extras`, and expanded docstrings for `ModeSolver.solve()`, `ModeSimulation.run_local()`, `Simulation.epsilon()`, and `Simulation.epsilon_on_grid()` to document the `config.simulation.use_local_subpixel` option.
 
 ## [2.10.2] - 2026-01-21
 
--- a/tidy3d/components/mode/mode_solver.py
+++ b/tidy3d/components/mode/mode_solver.py
@@ -493,6 +493,7 @@ class ModeSolver(Tidy3dBaseModel):
         and will thus be creating :class:`.MicrowaveModeSolverData`."""
         return isinstance(self.mode_spec, MicrowaveModeSpec)
 
+    @supports_local_subpixel
     def solve(self) -> ModeSolverData:
         """:class:`.ModeSolverData` containing the field and effective index data.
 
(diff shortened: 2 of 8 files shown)


tidy3d/components/mode/mode_solver.py
```diff
@@ -493,19 +493,34 @@ def _has_microwave_mode_spec(self) -> bool:
...
+<tidy3d.config.sections.SimulationConfig.use_local_subpixel>`
+        for details.
         """
-        log.warning(
-            "Use the remote mode solver with subpixel averaging for better accuracy through "
-            "'tidy3d.web.run(...)' or the deprecated 'tidy3d.plugins.mode.web.run(...)'.",
-            log_once=True,
-        )
+        # only warn if tidy3d-extras local subpixel is not enabled
+        if not tidy3d_extras["use_local_subpixel"]:
```
tylerflex, line 516: won't this also warn If this is left to default of `None`? should it be `if ... is False`?

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Changed
+- Improved `ModeSolver.solve()` to suppress the accuracy warning when local subpixel averaging is enabled via `tidy3d-extras`, and expanded docstrings for `ModeSolver.solve()`, `ModeSimulation.run_local()`, `Simulation.epsilon()`, and `Simulation.epsilon_on_grid()` to document the `config.simulation.use_local_subpixel` option.
```
tylerflex: might be better under "fixed"

tylerflex (commented):
a couple comments, also same thing about the use of quotes in the pip install extras, I'm not sure if it's just a Mac thing. It trips me up from time to time, but now that I know about it, I know how to fix it. just worried some users might get ```no matches found: tidy3d[extras]``` which is what I see on Mac if I copy that syntax.

tidy3d/components/mode/mode_solver.py
caseyflex: I actually think I want this to warn if left at the default of None, if tidy3d-extras isn’t installed (because then subpixel won’t work out the box, and the warning then is the current desired behavior). The warning is updated to recommend extras as well though

caseyflex: Note that this value im checking is whether extras package loaded, not the config flag

caseyflex: Although if the config flag is false, this decorator will explicitly unload extras

caseyflex: Basically this value tells you exactly whether subpixel is actually being used in the current function

tylerflex: ok I see. yea I thought this was the config flag.

From my perspective as a user, it should just be a warning if for whatever reason I'm not using subpixel.  If I am, it should be quiet? is that consistent with your thinking too and the logic here?

tylerflex: I guess another question is whether the current logic now prints 2 warnings (one if tidy3d extras not installed and one if subpixel is not used?) that would be less ideal

caseyflex: Yeah it should only warn here if subpixel isn’t being used.

There usually should not be two warnings. Only this one if subpixel is not being used. There will be another warning if tidy3d-extras is installed but failed to load for another reason like user permissions, but I think that’s good

tylerflex (approved):

PR #3255, opened 2026-02-08
FXC-3763: Added more lossy dielectrics and lossy metals with frequency parametrization to RF material library

George-Guryev-flxcmp:
## Summary

Adds frequency-parametrized support for lossy dielectrics and lossy metals to the RF material library, enabling users to request material models for specific frequency ranges.

### Core Functionality
- **Added frequency-parametrized materials**:
  - `VariantItemFreqRangeDielectric`: For lossy dielectric materials with `PoleResidue` models
  - `VariantItemFreqRangeMetal`: For lossy metal materials with `LossyMetalMedium` models

- **Enhanced `medium()` methods**:
  - Both classes support optional `frequency_range` parameter
  - `VariantItemFreqRangeDielectric`: Validates if requested range is within stored measurements
    - Returns stored model with updated metadata when within range
    - Creates new `PoleResidue` model using constant loss tangent fitter with averaged properties when outside range
    - Issues informative warnings when measurements are unavailable
  - `VariantItemFreqRangeMetal`: Creates `LossyMetalMedium` fitted for the requested frequency range

- **Added fields to `VariantItemFreqRangeDielectric`**:
  - `loss_tangent`: Loss tangent values (single or list)
  - `eps_real`: Real permittivity values (single or list)
  - `measurement_frequencies`: Frequencies at which measurements were taken

### Testing
- Added comprehensive unit tests covering:
  - Frequency range validation for dielectrics (inside/outside/overlapping ranges)
  - Metal material creation with frequency ranges
  - Warning verification using `AssertLogLevel`
  - Averaged values verification for dielectrics

## Behavior Changes

**Before**: RF material library only supported fixed-frequency material models.

**After**: Users can now:
1. Use default frequency range: `variant.medium()` returns model for stored/default range
2. Request custom frequency range: `variant.medium(frequency_range)` returns model fitted for requested range
3. For dielectrics outside stored measurements: Receive warnings and models created with averaged properties
4. For metals: Always get models fitted specifically for the requested frequency range

## Testing

Ran the following checks:
- `pytest tests/test_plugins/test_rf_material_library.py` - All tests pass

## Documentation

- Updated docstrings for both `VariantItemFreqRangeDielectric` and `VariantItemFreqRangeMetal`
- Warning messages clearly explain the situation and approach taken
- Examples demonstrate usage of both material types

---

> [!NOTE]
> **Medium Risk**
> Adds new RF material item/variant types and introduces on-demand fitting behavior (including warnings and default fit parameters) for `LossyMetalMedium`, which could affect simulation material behavior if misused. Changes are localized to the RF material library plus docs/tests and a new constant.
> 
> **Overview**
> Adds **frequency-range parametrized** entries to the RF material library by introducing `VariantItemFreqRangeDielectric` and `VariantItemFreqRangeMetal`, plus `MaterialItemFreqRange` with a new `medium_in_range(frequency_range)` API alongside the existing `.medium` default.
> 
> Dielectrics can now return the stored `PoleResidue` when the requested range is within the fitted bounds, or emit a warning and create a new constant-loss-tangent fit using averaged measured properties when out of range; metals now generate `LossyMetalMedium` models fitted for the requested range (defaulting to a new `MICROWAVE_FREQUENCY_RANGE` of 0.3–300 GHz with enhanced fit params).
> 
> Updates generated docs and microwave docs pages to document the new access patterns and expanded RF material catalog (additional laminates/ceramics and many common metals), and adds a comprehensive new `test_rf_material_library.py` covering validation, range handling, warnings, and model creation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bd194d8235097379fccd08f88e170a3d3c4178c7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -203,6 +467,420 @@
...
+        eps_inf=2.019212989211342,
+        poles=[
+            ((-881114136668.1382 + 0j), (306501347.27997166 - 0j)),
+            ((-55285601084.21439 + 0j), (9937348.543665012 - 0j)),
+            ((-164820074.0670714 + 0j), (1195084.7305780055 - 0j)),
+            ((-133999177246.56708 + 0j), (27607743.236883767 - 0j)),
+        ],
+        frequency_range=(1e9, 1e11),
+    ),
+    reference=[rf_material_refs["PTFE"]],
```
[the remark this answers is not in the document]
George-Guryev-flxcmp: Nice catch! Fixed the issue.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,275 @@
...
+        # Check if requested frequency_range is within the stored pole_residue frequency_range
+        stored_freq_range = self.pole_residue.frequency_range
+        if stored_freq_range is not None:
+            f_min_stored, f_max_stored = stored_freq_range
+            f_min_requested, f_max_requested = frequency_range
+
+            # Check if requested range is within stored range
+            if f_min_requested >= f_min_stored and f_max_requested <= f_max_stored:
+                # Within range: return stored model with updated frequency_range
+                return self.pole_residue
```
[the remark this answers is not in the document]
George-Guryev-flxcmp: Good point. Updated docstring to reflect implemented logic

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,275 @@
...
+            A LossyMetalMedium fitted for the specified frequency range.
+
+        Notes
+        -----
+        The ``LossyMetalMedium`` internally fits a pole-residue model for the surface impedance
+        over the specified frequency range. This fitting happens on-the-fly when the medium
+        is created, ensuring accuracy for the requested frequency range.
+        """
+        if frequency_range is None:
+            raise ValueError("frequency_range is required for LossyMetalMedium")
```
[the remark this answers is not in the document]
George-Guryev-flxcmp: Fixed.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,275 @@
...
+    variants: dict[str, VariantItemFreqRange] = Field(
+        ...,
+        title="Dictionary of available variants for this material",
+        description="A dictionary of available variants for this material "
+        "that maps from a key to the variant model.",
+    )
+
+    def __getitem__(self, variant_name: str) -> VariantItemFreqRange:
+        """Helper function to easily access a variant."""
+        return self.variants[variant_name]
```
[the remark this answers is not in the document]
George-Guryev-flxcmp: Added `medium` property to `MaterialItemFreqRange` to make `.medium` consistent with `MaterialItem`.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -203,6 +514,420 @@
...
+            ),
+            (
+                (-25161091901.745766 - 3359786740.934802j),
+                (130260064.80755003 + 364684745.84531015j),
+            ),
+        ],
+        frequency_range=(1e9, 1e10),
+    ),
+    reference=[rf_material_refs["Megtron6_R5775_R5670(N)"]],
+)
```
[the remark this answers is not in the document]
George-Guryev-flxcmp: Nice catch!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added frequency-parametrized lossy dielectrics and lossy metals to the RF material library. User can choose between the material fit for the default frequency range via ``medium()`` or get a fit for the desired frequency range via ``medium(frequency_range)``. When the specified range is outside the default one, a new model is created with averaged properties and a warning is issued.
```
dmarek-flex: in case you missed it should be at bottom

weiliangjin2021:
Could you also update doc-related scripts and take a look at how the doc page looks like?

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+from tidy3d.plugins.dispersion.fit_fast import FastDispersionFitter
+
+if TYPE_CHECKING:
+    from tidy3d.components.types import FreqBound
 
 from .rf_material_reference import rf_material_refs
 
+
+class VariantItemFreqRange(AbstractVariantItem, ABC):
```
weiliangjin2021: named it as `AbstractVariantItemFreqRange`

weiliangjin2021:
A few comments by Claude first:

  ---
  tidy3d/plugins/microwave/rf_material_library.py:761

  Megtron6_R5775_N.frequency_range appears too narrow

  This variant has measurement data spanning 1 GHz to 50 GHz (10 data points at lines 751-753), but
  pole_residue.frequency_range is (1e9, 1e10) — only up to 10 GHz. A user requesting e.g. (1e9,
  50e9) would get the averaged fallback model despite measurement data existing up to 50 GHz.

  Should this be (1e9, 5e10) to match the measurement range? (Compare with Megtron6_R5670_N at line
  776 which correctly uses (1e9, 5e10).)

  ---
  tidy3d/plugins/microwave/rf_material_reference.py:133-138

  Cobalt reference appears to cite the wrong paper

  The reference for Cobalt cites "Benon H. J. Bielski, et al., Reactivity of HO2/O Radicals in
  Aqueous Solution" — this is a paper on radical chemistry, not electrical conductivity. The JPCRD
  volume/issue looks correct but the specific paper seems wrong.

  ---
  tidy3d/plugins/microwave/rf_material_library.py:780

  "Neclo" should be "Nelco"

  The variable names and library keys (Neclo_N4000_6, Neclo_N4000_13EP, etc.) misspell the brand —
  it's Nelco (Park Electrochemical). The reference datasheet_title at rf_material_reference.py:101
  spells it correctly as "Nelco". Since these are user-facing keys in rf_material_library.keys(),
  worth fixing before this ships.

  ---
  tidy3d/plugins/microwave/rf_material_library.py:1060-1085

  Duplicate display names across distinct library entries

  Several pairs share the same name field:
  - "Megtron6_R5775_KG" and "Megtron6_R5670_KG" both → name="Megtron6 R5775_R5670(KG)"
  - "Megtron6_R5775_N" and "Megtron6_R5670_N" both → name="Megtron6 R5775_R5670(N)"
  - "Neclo_N4000_13EP" and "Neclo_N4000_13EP_SI" both → name="Neclo N4000-13"

  Could be confusing in listings/summaries. Consider differentiating the display names (e.g.
  "Megtron6 R5775(KG)" vs "Megtron6 R5670(KG)").

  ---
  tidy3d/plugins/microwave/rf_material_reference.py:90-91

  Duplicated substring in datasheet_title

  Python implicit string concatenation produces:
  "...Multi-layer circuit board materialsMulti-layer circuit board materials"
  Missing a space or the second line is an accidental duplicate.

  ---
  tidy3d/plugins/microwave/rf_material_library.py:314-319

  Nit: dead else branch

  VariantItemFreqRange is sealed to VariantItemFreqRangeDielectric and VariantItemFreqRangeMetal —
  this fallback with getattr(variant, "medium", None) can never be reached. Consider removing it.

  ---
  tidy3d/plugins/microwave/rf_material_library.py:1144-1150

  Nit: Aluminum/Aluminium dual library entries

  Both "Aluminium" and "Aluminum" exist as separate entries pointing to the same variant object.
  This is user-friendly but it means both show up in iterations/listings. A note in the docstring or
   a comment here would help clarify the intent.

  ---

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+        description="Real permittivity for lossy dielectric materials.",
+    )
+
+    measurement_frequencies: Union[float, list[float]] = Field(
+        ...,
+        title="Measurement Frequencies",
+        description="Frequencies at which the material properties were measured.",
+    )
+
+    pole_residue: PoleResidue = Field(
```
weiliangjin2021: how about `prefitted_medium` (or other better name) to be clearer?

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+        ...,
+        title="Pole-Residue Model",
+        description="Pole-residue model for lossy dielectric materials.",
+    )
+
+    def medium(self, frequency_range: Optional[FreqBound] = None) -> PoleResidue:
+        """
+        Generate ``PoleResidue`` medium with specified ``frequency_range``.
+        If ``frequency_range`` is not provided, returns the original ``PoleResidue`` model.
```
weiliangjin2021: We don't need to emphasize PoleResidue here, as this function is more user-facing. So instead of "original ``PoleResidue`` model", prefitted material model should be sufficient.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+            f_min_stored, f_max_stored = stored_freq_range
+            f_min_requested, f_max_requested = frequency_range
+
+            # Check if requested range is within stored range
+            if f_min_requested >= f_min_stored and f_max_requested <= f_max_stored:
+                # Within range: return stored model
+                return self.pole_residue
+
+        # Outside range (or stored_freq_range is None): create new model using constant loss tangent fitter
+        # Calculate average values
```
weiliangjin2021, line 199: This part (how an array of eps/loss tangent data are handled) should be added to docstring, e.g. the description of loss_tangent/eps_real fields

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+            loss_tangent_avg = float(self.loss_tangent)
+
+        log.warning(
+            f"Requested frequency_range {frequency_range} is outside the stored "
+            f"pole_residue frequency_range {stored_freq_range}. "
+            f"The RF material library does not have measurements for loss_tangent and eps_real "
+            f"in the desired frequency range. A new PoleResidue model will be created using "
+            f"the constant loss tangent fitter with averaged material properties "
+            f"(eps_real={eps_real_avg:.6f}, loss_tangent={loss_tangent_avg:.6f}). "
+            f"The resulting fit may be less accurate."
```
weiliangjin2021: would be good to add that a more accurate approach is to call the fast fitter directly.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+
+class VariantItemFreqRangeMetal(VariantItemFreqRange):
+    """Frequency-range parametrized variant item for lossy metal materials.
+    This class stores conductivity and optional parameters for creating a ``LossyMetalMedium``.
+    """
+
+    conductivity: PositiveFloat = Field(
+        ...,
+        title="Conductivity, S/μm",
+        description="Electrical conductivity for lossy metal materials (S/μm).",
```
weiliangjin2021: units: https://github.com/flexcompute/tidy3d/blob/9ff44504c399eb47070754e8e6c5951d915b9c55/tidy3d/components/medium.py#L1486

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+    roughness: Optional[SurfaceRoughnessType] = Field(
+        None,
+        title="Surface Roughness Model",
+        description="Surface roughness model for lossy metal.",
+    )
+
+    thickness: Optional[PositiveFloat] = Field(
+        None,
+        title="Conductor Thickness",
+        description="Conductor thickness for lossy metal (μm).",
```
weiliangjin2021: More detailed description, e.g. the one in lossy metal medium.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+        Raises
+        ------
+        ValueError
+            If ``frequency_range`` is None.
+
+        Notes
+        -----
+        The ``LossyMetalMedium`` internally fits a pole-residue model for the surface impedance
+        over the specified frequency range. This fitting happens on-the-fly when the medium
+        is created, ensuring accuracy for the requested frequency range.
```
weiliangjin2021: actually, the fitting is not happening when the medium is created.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+        return self.variants[variant_name]
+
+    @property
+    def medium(self) -> Union[PoleResidue, LossyMetalMedium]:
+        """The default medium for the default variant.
+
+        Returns the medium for the default variant using its default frequency range.
+        For dielectrics, this returns the stored PoleResidue model (equivalent to calling
+        ``variant.medium()`` without arguments).
+        For metals, raises ValueError as frequency_range is required.
```
weiliangjin2021: this makes me wonder if we should add a default frequency range to `VariantItemFreqRangeMetal`, which is based on measured data or known valid range. With that both Variant will have a default frequency range.

tidy3d/plugins/microwave/rf_material_library.py
dmarek-flex: yea and maybe just leave the title as "Conductivity" the units can be in the description too I suppose

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+        # This is consistent with the limitation that LossyMetalMedium requires frequency_range
+        return {}
+
+
+class MaterialItemFreqRange(MaterialItem):
+    """A material that includes several frequency-range parametrized variants."""
+
+    variants: dict[str, VariantItemFreqRange] = Field(
+        ...,
+        title="Dictionary of available variants for this material",
```
dmarek-flex, line 380: make title just Variants or similar, and keep the rest in the description

dmarek-flex: hmm I guess the MaterialItem has a long title. Still think it should be short though @weiliangjin2021

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,323 @@
...
+                f"frequency_range is required for {variant.__class__.__name__}.medium(). "
+                f"Please use {self.name}['{self.default}'].medium(frequency_range) instead, "
+                "where frequency_range is a tuple (f_min, f_max) in Hz."
+            )
+        else:
+            # Fallback: try to call if it's callable, otherwise return as-is
+            medium_attr = getattr(variant, "medium", None)
+            if callable(medium_attr):
+                return medium_attr()
+            return medium_attr
```
dmarek-flex: Probably just raise an error here unless I am missing something. The variant can only be `VariantItemFreqRangeDielectric` or `VariantItemFreqRangeMetal`?

dmarek-flex (approved):
Looks like a nice contribution! Just a bit confused by that part where it falls back to checking if the variant is callable. From the Pydnatic model it seems like it has to be either one of the previous classes? VariantItemFreqRangeDielectric/VariantItemFreqRangeMetal

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: Good point!

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: Yes. The fallback was mostly to potentially accommodate  new medium types, but currently it supports only the two types you've mentioned. I've removed the clause for now.

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: Done! 

Is there a reason for preferring explicit specification of `json_schema_extra={"units":CONDUCTIVITY}`  instead of `units=CONDUCTIVITY`? Is this done to affect only tidy3d_extra-related Jsons and not `Simulation.json`?

CHANGELOG.md
George-Guryev-flxcmp: Thanks!

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: Changed dosctring to: 

"The ``LossyMetalMedium`` internally fits a pole-residue model for the scaled surface impedance over the specified frequency range. This fitting happens lazily when the ``scaled_surface_impedance_model`` property is first accessed, ensuring accuracy for the requested frequency range."

Please let me know if you have any suggestions.

tidy3d/plugins/microwave/rf_material_reference.py
```diff
@@ -40,4 +40,107 @@
...
+        journal="Matula, R.A. Journal of Physical and Chemical Reference Data. 8 (4): 1147 (1979).",
+        url="https://srd.nist.gov/JPCRD/jpcrd155.pdf",
+        doi="https://doi.org/10.1063/1.555614",
+    ),
+    "Annealed_Copper": ReferenceData(
+        journal="Copper wire tables, U.S. National Bureau of Standards, (1914).",
+        url="https://archive.org/details/copperwiretables31unituoft/page/10/mode/2up",
+    ),
+    "Alum_293K": ReferenceData(
+        journal="P.D. Desai, H.M. James, and C.Y. Ho. Electrical Resistivity of Aluminum and Manganese,"
```
yaugenst-flex: Similar spacing issue here.

yaugenst-flex (changes requested):
This change introduces callable `variant.medium(...)` for RF frequency-range variants, but doc-generation logic still assumes `var.medium` is a medium object. Repro:

```py
from tidy3d.plugins.microwave import rf_material_library as rf_lib
for _, mat in sorted(rf_lib.items()):
    for _, var in sorted(mat.variants.items()):
        medium = var.medium
        _ = len(medium.poles)
```

This fails with `AttributeError: 'function' object has no attribute 'poles'`. Could we update `docs/generate_doc.py` accordingly?

tidy3d/plugins/microwave/rf_material_reference.py
```diff
@@ -40,4 +40,107 @@
...
+    ),
+    "Isola_FR408": ReferenceData(
+        manufacturer="Isola Group",
+        datasheet_title="FR408 Standard Loss, High Tg Epoxy Laminate and Prepreg",
+        url="https://www.isola-group.com/wp-content/uploads/data-sheets/fr408.pdf",
+    ),
+    "Megtron6_R5775_R5670(KG)": ReferenceData(
+        manufacturer="Panasonic Industry",
+        datasheet_title="Ultra-low transmission loss Highly heat resistant"
+        "Multi-layer circuit board materials",
```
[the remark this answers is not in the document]
yaugenst-flex: needs to be fixed

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: Revisited this: it now raises an error, and I added a unit test to validate it.

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: Changed it to `Variants`, but I am flexible on the title.

tidy3d/plugins/microwave/rf_material_library.py
weiliangjin2021: @yaugenst-flex i have similar questions too. This seems to be introduced in pydantic v1->v2 PR?

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -203,6 +515,414 @@
...
+            ((-51917915632.41483 + 0j), (540455950.6398327 - 0j)),
+            ((-11738278467.08503 - 4376627321.665409j), (-36521596.20551511 + 168696656.94222957j)),
+        ],
+        frequency_range=(1e9, 5e10),
+    ),
+    reference=[rf_material_refs["Megtron6_R5775_R5670(N)"]],
+)
+Megtron6_R5670_N = VariantItemFreqRangeDielectric(
+    loss_tangent=[0.002, 0.003, 0.004, 0.004, 0.004, 0.005, 0.005, 0.005, 0.005, 0.005],
+    eps_real=[3.47, 3.40, 3.39, 3.39, 3.39, 3.58, 3.58, 3.58, 3.58, 3.58],
```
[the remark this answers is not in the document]
George-Guryev-flxcmp: Good catch! Fixed it.

tidy3d/plugins/microwave/rf_material_library.py
yaugenst-flex: That's right, these are json schema metadata fields and now all go into `json_schema_extra` as they are not relevant at runtime. The old syntax has been deprecated for a while now (we just used to ignore warnings). This was changed in the v2 migration but would've been changed either way.

tidy3d/plugins/microwave/rf_material_reference.py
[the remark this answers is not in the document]
George-Guryev-flxcmp: Fixed this.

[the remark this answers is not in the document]
George-Guryev-flxcmp: Fixed this.

tidy3d/plugins/microwave/rf_material_reference.py
George-Guryev-flxcmp: Fixed.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,353 @@
...
+        ...,
+        title="Measurement Frequencies",
+        description="Frequencies at which the material properties were measured.",
+    )
+
+    prefitted_medium: PoleResidue = Field(
+        ...,
+        title="Pole-Residue Model",
+        description="Pole-residue model for lossy dielectric materials.",
+    )
```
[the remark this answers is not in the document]
George-Guryev-flxcmp: Added a model validator and updated unit tests.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -252,4 +998,279 @@
...
+        },
+        default="standard",
+    ),
+    "Nelco_N4000_13EP_SI": MaterialItemFreqRange(
+        name="Nelco N4000-13",
+        variants={
+            "SI": Nelco_N4000_13EP_SI,
+        },
+        default="SI",
+    ),
```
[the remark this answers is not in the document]
George-Guryev-flxcmp: Updated material names.

yaugenst-flex (changes requested):
Thanks for the substantial RF material-library expansion and test coverage. I validated this locally and have two remaining issues before approval:

1. `docs/api/microwave/rf_material_library.rst` is the page actually included in docs (`docs/api/microwave/index.rst`), and it is still inconsistent with current behavior/examples:
- It still advertises `rf_material_library['mat'].medium` as a property, but for `MaterialItemFreqRange` entries (e.g. `RT_duroid5880`) this is now a method (`.medium()`).
- Many examples in that page still use `material_library[...]` instead of `rf_material_library[...]`.

2. The abstract contract docstring in `tidy3d/plugins/microwave/rf_material_library.py` is now inconsistent:
- `AbstractVariantItemFreqRange.medium()` says metals require `frequency_range` and raise if `None`.
- `VariantItemFreqRangeMetal.medium()` actually defaults `None` to `MICROWAVE_FREQUENCY_RANGE`.

Please align the docs page + abstract docstring with implemented behavior, then I can re-review quickly.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,395 @@
...
+
+    This class stores a ``PoleResidue`` model and provides a ``medium()`` method that
+    updates the ``frequency_range`` attribute for the requested range.
+    """
+
+    loss_tangent: Union[float, list[float]] = Field(
+        ...,
+        title="Loss Tangent",
+        description="Loss tangent for lossy dielectric materials. "
+        "If the desired frequency range is outside of the default frequency range, "
```
weiliangjin2021: default frequency range sounds vague: do you mean prefitted material frequency range or measured frequency range?

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,395 @@
...
+        title="Dictionary of available variants for this material",
+        description="A dictionary of available variants for this material "
+        "that maps from a key to the variant model.",
+    )
+
+    def __getitem__(self, variant_name: str) -> Union[PoleResidue, LossyMetalMedium]:
+        """Helper function to easily access the medium of a variant."""
+        return self.variants[variant_name].medium()
+
+    def medium(
```
yaugenst-flex: This changes a base-class contract: `MaterialItem.medium` is a property, but `MaterialItemFreqRange.medium` is callable. Polymorphic code that does `isinstance(x, MaterialItem)` and then accesses `x.medium` can now get a bound method instead of a medium object. Could we keep `medium` as property and use a separate range selector (`at(...)` / `medium_at(...)` / `medium_in_range(...)`)?

yaugenst-flex (changes requested):
The frequency-parameterized RF direction looks good overall. Remaining concern is API consistency/backward compatibility: `MaterialItem.medium` is a property in the base class, but `MaterialItemFreqRange.medium` is callable.

Could we preserve `medium` as the default property and add a separate range-selector API? Suggested names for discussion: `at(frequency_range=...)`, `medium_at(frequency_range=...)`, or `medium_in_range(frequency_range=...)`.

I would avoid `.medium.at(...)` because that likely requires turning `.medium` into a proxy object instead of an actual medium value.

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: Yes, your suggestion to keep `MaterialItem.medium` as a property makes sense and I agree that having a separate range selector makes API more consistent/uniform. 

In this regard I want to bring up a closely related issue: the proposed frequency parameterization largely follows the axis-parameterized class `MaterialItemUniaxial` used in the `tidy3d/material_library.py`. If we decide to follow the path outlined above it would make sense to update `MaterialItemUniaxial`  (and maybe `VariantItemUniaxial`) accordingly.   

Open to your suggestions @weiliangjin2021, @yaugenst-flex, @dbochkov-flexcompute

yaugenst-flex: I agree. I think we should preserve the base `MaterialItem` contract in this PR: `medium` should stay a property, and range selection should use a separate API (`medium_in_range(...)` / `at(...)`).

Concretely for `MaterialItemFreqRange`:
- `material.medium` -> default medium (property)
- `material.medium_in_range(freq_range)` -> parametrized medium

On the uniaxial point: I agree that `MaterialItemUniaxial`/`VariantItemUniaxial` already have the same inconsistency today. I’d keep that as a follow-up cleanup (with a deprecation path), rather than expanding scope in this PR.

docs/api/microwave/rf_material_library.rst
```diff
@@ -23,7 +22,17 @@ For example, Rogers3010 laminate can be loaded as:
 
 You can also import the default variant of a material by:
 
->>> medium = rf_material_library['mat'].medium
+For frequency-range parametrized materials (MaterialItemFreqRange), medium is a property:
+
+>>> medium = rf_material_library['RT_duroid5880'].medium
+
+For standard materials (MaterialItem), medium is also a property:
```
weiliangjin2021: On the doc page, MaterialItemFreqRange and MaterialItem are probably not meaningful enough to users. They are more internally oriented. It can be more helpful to mark the materials in the table on whether it's frequency-range parametrized or not.

docs/api/microwave/rf_material_library.rst
```diff
@@ -23,7 +22,17 @@ For example, Rogers3010 laminate can be loaded as:
...
+
+>>> medium = rf_material_library['RO3010'].medium
+
+To get a medium for a specific frequency range, use medium_at_range():
+
+>>> medium = rf_material_library['RT_duroid5880'].medium_at_range(frequency_range=(5e9, 10e9))
```
weiliangjin2021: The wording above can be improved. I guess the main point is that for frequency-range parametrized, it offers `medium_at_range`.

weiliangjin2021 (commented):
Looks in good shape! Just a few last comments.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,446 @@
...
+
+        If ``frequency_range`` is not provided (None), returns the ``medium`` property directly,
+        avoiding duplicate medium creation.
+        """
+        # If no frequency range specified, return the default medium property
+        if frequency_range is None:
+            return self.medium
+
+        # Use provided fit_param or None for default
+        fit_param = self.fit_param
```
yaugenst-flex, line 354: Quick clarification question: is the default-range divergence intentional here? `medium` uses enhanced fit params for the microwave default range (`12` poles / `50` sampling points), while `medium_at_range(MICROWAVE_FREQUENCY_RANGE)` goes through the normal defaults (`5` / `20`) when `fit_param` is unset. I’m not requesting a change, just checking whether that difference is by design.

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,446 @@
...
+            "fit_param": fit_param,  # fit_param is always set (either from self.fit_param or created above)
+        }
+        if self.roughness is not None:
+            kwargs["roughness"] = self.roughness
+        if self.thickness is not None:
+            kwargs["thickness"] = self.thickness
+
+        return LossyMetalMedium(**kwargs)
+
+    def medium_at_range(self, frequency_range: Optional[FreqBound] = None) -> LossyMetalMedium:
```
yaugenst-flex: Naming question: do we want `medium_at_range` as the long-term public name, or would something like `medium_in_range`/`medium_for_range` better match that this takes an interval `(f_min, f_max)` rather than a point? No blocker either way; just asking for API clarity/intent.

yaugenst-flex (commented):
Two non-blocking questions before merge:
1. Is the default-range fit-parameter divergence between `medium` and `medium_at_range(MICROWAVE_FREQUENCY_RANGE)` intentional?
2. Is `medium_at_range` the preferred API name, or should we consider a name that more explicitly conveys interval semantics?

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: That's a great catch! The reason for using the enhanced fit parameters is that the default `MICROWAVE_FREQUENCY_RANGE` is too wide to obtain a reasonable fit with only `12` poles and `50` sampling points, and therefore requires a higher-order approximation. In practice, the expected frequency range is usually much narrower—typically on the order of a few GHz.

That said, `medium_at_range(MICROWAVE_FREQUENCY_RANGE)` should not downgrade the fit_param. I’ve added a check for the default frequency range to ensure that the enhanced parameters are used in this case.

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: I agree - `medium_in_range()` sounds more natural. Changed `medium_at_range()` to `medium_in_range()` in both the `.py` files and in the `.rst` documentation.

docs/api/microwave/rf_material_library.rst
```diff
@@ -34,142 +37,861 @@ To access the details of a variant, including material model and references, use
...
+.. rubric:: Alumina AO479U ("Alumina_AO479U")
+
+.. table::
+   :widths: auto
+
+   ======================== ============================ ============= ============= ===========
+   Variant                  Type                         Valid for     Model Info    Reference  
+   ======================== ============================ ============= ============= ===========
+   ``'standard'`` (default) Frequency-range parametrized 0.0 - 8.5 GHz 1-pole, lossy [1]        
+   ======================== ============================ ============= ============= ===========
```
weiliangjin2021, line 91: how does the doc page look like with the added column?

docs/api/microwave/rf_material_library.rst
```diff
@@ -34,142 +37,861 @@ To access the details of a variant, including material model and references, use
...
-   ====================== ============== ============= ===========
-   ``'design'`` (default) 1.0 - 30.0 GHz 5-pole, lossy [1]
-   ``'process'``          1.0 - 30.0 GHz 5-pole, lossy [1]
-   ====================== ============== ============= ===========
+   ====================== ======== ============== ============= ===========
+   Variant                Type     Valid for      Model Info    Reference  
+   ====================== ======== ============== ============= ===========
+   ``'design'`` (default) Standard 1.0 - 30.0 GHz 5-pole, lossy [1]        
+   ``'process'``          Standard 1.0 - 30.0 GHz 5-pole, lossy [1]        
+   ====================== ======== ============== ============= ===========
```
weiliangjin2021: Type "Standard" sounds too general, and you also have some material variant that is "(standard)". How about "Pre-fitted", or "Fixed frequency range"

weiliangjin2021 (approved):

docs/api/microwave/rf_material_library.rst
George-Guryev-flxcmp: `Pre-fitted` sounds reasonable.

docs/api/microwave/rf_material_library.rst
George-Guryev-flxcmp: Screenshot from 2026-02-24 20-27-32
Screenshot from 2026-02-24 20-27-43

George-Guryev-flxcmp: Seems to be reasonable.

weiliangjin2021: looks good

tidy3d/plugins/microwave/rf_material_library.py
```diff
@@ -2,11 +2,448 @@
...
+        json_schema_extra={"units": HERTZ},
+    )
+
+    prefitted_medium: PoleResidue = Field(
+        ...,
+        title="Pole-Residue Model",
+        description="Pole-residue model for lossy dielectric materials.",
+    )
+
+    @model_validator(mode="after")
```
yaugenst-flex, line 115: `loss_tangent`, `eps_real`, and `measurement_frequencies` are validated for equal lengths, but empty triples still pass construction. That creates a deferred failure path: `medium_in_range()` averages empty arrays to `nan` and then fails deep in dispersion fitting with a much less actionable error.

Evidence:
- `tidy3d/plugins/microwave/rf_material_library.py:115-134` validates only length equality.
- `tidy3d/plugins/microwave/rf_material_library.py:188-195` computes means that become `nan` for empty arrays.
- Repro at current PR head:
  - `poetry run pytest -q /tmp/test_rf_material_empty_lists_fit_trace.py -q --maxfail=1`
  - warning logs `eps_real=nan, loss_tangent=nan`, then fitting raises a `ValidationError` with NaN inputs.

Could we fail fast here with a clear `ValidationError` for empty paired inputs?

yaugenst-flex (commented):
Nice expansion of the RF material library and coverage.

One correctness gap remains: `VariantItemFreqRangeDielectric` accepts empty paired measurement arrays and then fails later in `medium_in_range()` via NaN propagation into dispersion fitting, instead of rejecting invalid input at construction.

Could you add a fail-fast validation for empty paired inputs and a regression test in `tests/test_plugins/test_rf_material_library.py` for that case?

tidy3d/plugins/microwave/rf_material_library.py
George-Guryev-flxcmp: Sure, this is a great catch! Added checks for empty tuples and NaN's.

PR #3258, opened 2026-02-09
FXC-5512 perf: stream outer_dot overlap kernel to reduce runtime and memory

yaugenst-flex:
## Summary

- Replaced the `outer_dot` hot path in `ElectromagneticFieldData.outer_dot(...)` with an adaptive streaming kernel:
  - `unblocked` path for smaller working sets.
  - `blocked` term-by-term path for large working sets to cap temporary allocations.
- Removed the runtime `try/except` fallback to `_outer_fn_summation` from `outer_dot`.
- Kept existing preprocessing and output semantics intact:
  - tangential-dim checks,
  - conjugation behavior,
  - interpolation/alignment behavior,
  - frequency intersection with left-order semantics,
  - mode add/rename/drop behavior (`mode_index_0`, `mode_index_1`).
- Added parity coverage for blocked vs unblocked streaming kernels.

## Why This Is Safe

- No public API/signature changes.
- No schema/data-model architecture changes.
- EME call sites unchanged.
- Numerical parity verified against legacy/current implementations in benchmark harness (checksums matched within floating-point tolerance).

## Benchmark (Legacy vs Current vs Final)

Environment: local macOS, median of 3 runs per case.
Memory metric: `tracemalloc` peak allocations during kernel call.

| Case | Layout | Legacy wall (s) | Current wall (s) | Final wall (s) | Final speedup vs legacy | Legacy peak alloc | Current peak alloc | Final peak alloc |
|---|---|---:|---:|---:|---:|---:|---:|---:|
| small (`48x48`, `f=3`, `m=12x12`) | mode last | 0.027 | 0.002 | 0.002 | 11.08x | 1.6 MiB | 744.9 KiB | 745.1 KiB |
| eme_like (`64x64`, `f=3`, `sweep=4`, `cell=6`, `m=20x20`) | mode last | 8.230 | 0.114 | 0.108 | 76.17x | 68.0 MiB | 2.0 MiB | 2.0 MiB |
| mode_axis_middle (`128x128`, `f=2`, `m=72x72`) | mode middle | 2.434 | 0.082 | 0.109 | 22.42x | 7.7 MiB | 162.5 MiB | 50.4 MiB |
| **Total (sum)** | mixed | **10.691** | **0.198** | **0.219** | **48.80x** | - | - | - |

Notes:
- `mode_axis_middle` is the case that reproduces the current memory regression (current unblocked path spikes to 162.5 MiB peak alloc); final blocked path reduces that to 50.4 MiB (~3.2x lower) while still being ~22x faster than legacy.
- In mode-last workloads, final stays on the unblocked path and remains effectively non-regressing vs current.

## Validation

Executed locally:

- `poetry run pytest -q tests/test_data/test_monitor_data.py`
- `poetry run pytest -q tests/test_plugins/test_mode_solver.py`
- `poetry run pytest -q tests/test_components/test_eme.py`
- `poetry run ruff check tidy3d/components/data/monitor_data.py tests/test_data/test_monitor_data.py`

Results:

- `tests/test_data/test_monitor_data.py`: 354 passed
- `tests/test_plugins/test_mode_solver.py`: 49 passed
- `tests/test_components/test_eme.py`: 18 passed

## Known Limitations

- Benchmark scenarios are synthetic but chosen to cover small, EME-like preserved-dim, and memory-regression layouts.
- `tracemalloc` measures Python allocator peaks (good for NumPy temporaries), not full process RSS.

---

> [!NOTE]
> **Medium Risk**
> Touches a performance-critical numerical kernel and reshapes how overlaps are computed; while API/output intent is preserved, errors in dimension handling or blocking could cause subtle numerical or shape regressions.
> 
> **Overview**
> Replaces the `ElectromagneticFieldData.outer_dot()` hot path with a new streaming NumPy kernel (`_outer_dot_numpy_kernel_streaming`) that iterates over frequency/preserved dims and performs blockwise mode matrix products to avoid materializing large mode-pair broadcast tensors, with new tunables like `OUTER_DOT_BLOCK_TARGET_BYTES`.
> 
> Adds a strict slicing helper (`_slice_mode_xy_matrix`) and updates the implementation to keep frequency ordering and mode-dimension semantics while reducing peak allocations. Extends `tests/test_monitor_data.py` with targeted parity tests covering preserved-dimension outputs and cases where the mode axis is not last.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit db498b4f2f4f4e6588406640cba137244837b44c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -1081,7 +1081,14 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
         tangential_dims: tuple[str, str],
         mode_slice: slice = SLICE_ALL,
     ) -> NDArray:
-        """Slice one preserve/frequency point and return ``(mode, xy)`` matrix."""
+        """Return a ``(mode, xy)`` view for one preserve/frequency point.
+
+        This helper is intentionally strict about expected remaining dimensions:
+        after indexing preserved dims and frequency, the data must reduce to
+        ``(mode_dim, tangential_dim_0, tangential_dim_1)`` in any order.
+        Singleton preserved/frequency axes are accepted to support alignment-
+        generated broadcast dimensions.
+        """
         dim_to_axis = {dim: axis for axis, dim in enumerate(dims)}
         idx = [slice(None)] * arr.ndim
         indexed_dims = set()
@@ -1142,10 +1149,17 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
         h_2: str,
         d_area: NDArray,
     ) -> DataArray:
-        """Blocked streaming kernel for outer-dot mode overlaps."""
+        """Compute ``outer_dot`` with frequency/preserve streaming and mode blocking.
+
+        The kernel avoids materializing mode-pair broadcast tensors of shape
+        ``(mode_left, mode_right, x, y)`` by iterating over frequency/preserved
+        indices and accumulating blockwise matrix products over mode chunks.
+        """
         data_array_temp_1 = list(fields_1.values())[0]
         data_array_temp_2 = list(fields_2.values())[0]
 
+        # Keep output coord ordering consistent with the existing outer_dot contract:
+        # preserved dims + ``f`` from fields_1, then ``mode_index_0`` and ``mode_index_1``.
         coords = {key: val.to_numpy() for key, val in data_array_temp_1.coords.items()}
         for dim in tangential_dims:
             coords.pop(dim)
@@ -1196,6 +1210,8 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
 
         num_grid_points = d_area.size
         itemsize = np.dtype(np.result_type(fields_1[e_1].dtype, fields_2[e_1].dtype)).itemsize
+        # Heuristic: choose a mode block size that targets bounded temporary allocations,
+        # then clamp to practical limits to avoid tiny or overly large chunks.
         if num_grid_points == 0:
             block_size = OUTER_DOT_BLOCK_MAX_SIZE
         else:
@@ -1203,6 +1219,9 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
             block_size = max(OUTER_DOT_BLOCK_MIN_SIZE, block_size)
             block_size = min(OUTER_DOT_BLOCK_MAX_SIZE, block_size)
 
+        # Decompose:
+        # 0.25 * (E1 x H2 - E2 x H1 - H1 x E2 + H2 x E1)
+        # into signed matrix products that can be accumulated blockwise.
         term_specs = (
             (1.0, e_1, h_2),
             (-1.0, e_2, h_1),
--- a/tidy3d/components/data/monitor_data.py
+++ b/tidy3d/components/data/monitor_data.py
@@ -121,6 +121,11 @@ MIN_ANGULAR_SAMPLES_SPHERE = 10
 # Threshold for cos(theta) to avoid unphysically large amplitudes near grazing angles
 COS_THETA_THRESH = 1e-5
 MODE_INTERP_EXTRAPOLATION_TOLERANCE = 1e-2
+OUTER_DOT_UNBLOCKED_MAX_BYTES = 128 * 1024**2
+OUTER_DOT_BLOCK_TARGET_BYTES = 64 * 1024**2
+OUTER_DOT_BLOCK_MIN_SIZE = 8
+OUTER_DOT_BLOCK_MAX_SIZE = 64
+SLICE_ALL = slice(None)
 
 GRID_CORRECTION_TYPE = Union[
     float,
@@ -1018,7 +1023,6 @@ class ElectromagneticFieldData(AbstractFieldData, ElectromagneticFieldDataset, A
         freq_self = Index(coords[0]["f"].values)
         freq_other = Index(coords[1]["f"].values)
         common_freqs = freq_self.intersection(freq_other, sort=False)
-        f = common_freqs.to_numpy()
         # Keep frequency order consistent with the current data while aligning the other dataset.
         isel1 = freq_self.get_indexer(common_freqs)
         isel2 = freq_other.get_indexer(common_freqs)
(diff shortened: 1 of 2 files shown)


caseyflex:
Thanks Yannick, this is a helpful PR. It's helpful to dig into this sort of thing.

I did some benchmarking in frontend, backend, and EME. The current implementation has a memory issue for our primary use case (large transverse grids, many modes, few frequencies), but it's fixable without giving up the BLAS speedup.

Let M = num_modes, F = num_freqs, S = num_transverse_grid_points.

**What the "streaming" restructuring actually trades off**

The old approach loops over mode pairs (M^2 iterations), processing all frequencies per iteration. The new approach loops over frequencies (F iterations), processing all modes per iteration via BLAS. This "streams" over frequency -- removing F from per-iteration memory -- but vectorizes over modes -- adding M:

| Loop | Per-iter work | Per-iter kernel memory |
| --- | --- | --- |
| Old (M^2 iterations) | O(F\*S) element-wise | O(S\*F) -- independent of M |
| New (F iterations) | O(M^2\*S) BLAS gemm | O(M\*S) -- independent of F |

The speed improvement comes from replacing M^2 Python loop iterations with a single BLAS matmul -- same total arithmetic, but much better cache utilization, SIMD, and lower per-iteration overhead. For the primary use case (M >> F), this trades a large loop count for a small one.

But the memory trade-off goes the wrong way for the same case: when M >> F, replacing O(S\*F) temporaries with O(M\*S) temporaries is a net increase.

**Benchmarks**

At benchmark scale (kernel-only, excluding shared field extraction overhead):

| Config | Old time | New time | Speedup | Old mem | New mem |
| --- | --- | --- | --- | --- | --- |
| M=30, S=2961, F=1 | 38 ms | 2.3 ms | 16x | 313 KiB | 7,155 KiB |
| M=30, S=4800, F=3 | 230 ms | 8.6 ms | 27x | 1,219 KiB | 11,537 KiB |

At production scale (M=50, S=250,000, F=1), measured via tracemalloc: the current kernel peaks at ~760 MiB vs ~31 MiB for the old approach. The ~760 MiB comes from materializing all 4 weighted right-side matrices of shape (M, S) simultaneously inside the frequency loop. The speed gain is real (8.7s to 376ms), but the ~25x memory increase is prohibitive for large-grid workloads -- especially under parallel execution where per-worker memory matters.

**Suggested fix**

The memory increase is incidental, not fundamental to the BLAS approach. Two changes reduce peak memory to O(block\*S) -- independent of M -- while retaining the BLAS speedup:

1. **Process the 4 cross-product terms one at a time** instead of holding all weighted matrices simultaneously. Nearly free -- reduces peak from 760 to 191 MiB with negligible speed cost.
2. **Block both sides of the mode dimension** -- compute (block, S) @ (S, block) chunks. The left block is a view into the source field data (zero alloc for tidy3d's standard (tan1, tan2, f, mode) dim ordering); only the weighted right block allocates block \* S \* 16 bytes per inner iteration. BLAS remains efficient because S is large.

Benchmarked at M=50, S=250,000, F=1 (tracemalloc peak):

| Strategy | Time | Speedup | Peak alloc |
| --- | --- | --- | --- |
| Old (M^2 loop) | 8.7s | -- | 31 MiB |
| Current impl (4 weighted alive) | 376ms | 23x | 763 MiB |
| + term-by-term | 403ms | 22x | 191 MiB |
| + term-by-term + block=8 | 1.6s | 5.4x | 31 MiB |
| + term-by-term + block=16 | 1.1s | 8.0x | 61 MiB |
| + term-by-term + block=32 | 587ms | 14.8x | 122 MiB |

At block=8, peak memory matches the old approach while still being 5x faster. The block size is a tunable speed-memory knob.

**Memory regression test**

I'd also suggest adding a production-scale memory regression test -- a tracemalloc-guarded call at M>=30, S>=50,000 that asserts peak kernel memory stays within a bound proportional to block \* S, not M \* S. This would have caught the current issue.

**Summary**

The core idea (BLAS matmul over mode pairs) is the right direction and the speed gains are significant. I'd like to see the double-sided blocking optimization and a memory test before merging.

yaugenst-flex:
Posted updated benchmark comparison with commit-hash columns for clarity:
- `af528fad` = **before** blocked/adaptive kernel changes ("current" in earlier discussion)
- `287209c0` = **after** blocked/adaptive kernel changes (final)

Environment:
- local macOS
- median walltime over 3 runs
- peak allocation from `tracemalloc` during kernel call

| Case | Layout | Legacy wall (s) | `af528fad` wall (s) | `287209c0` wall (s) | `af528fad` peak alloc | `287209c0` peak alloc |
|---|---|---:|---:|---:|---:|---:|
| small (`48x48`, `f=3`, `m=12x12`) | mode-last | 0.027 | 0.00244 | 0.00245 | 744.9 KiB | 745.1 KiB |
| eme_like (`64x64`, `f=3`, `sweep=4`, `cell=6`, `m=20x20`) | mode-last | 8.230 | 0.11375 | 0.10805 | 2.0 MiB | 2.0 MiB |
| mode_axis_middle (`128x128`, `f=2`, `m=72x72`) | mode-middle | 2.434 | 0.08209 | 0.10856 | 162.5 MiB | 50.4 MiB |

Key point on regression case:
- In `mode_axis_middle`, `af528fad` shows the high temporary-allocation footprint (`162.5 MiB`).
- `287209c0` reduces that to `50.4 MiB` (~3.2x lower), which is the intended memory fix from the PR discussion.

Aggregate runtime (sum of listed cases):
- `af528fad`: `0.198s`
- `287209c0`: `0.219s`
- Overall: `~0.90x` vs `af528fad` (tradeoff accepted to remove memory spike on mode-middle layout).

tidy3d/components/data/monitor_data.py
```diff
@@ -1080,6 +1068,218 @@ def fn(fields_1: dict[str, NDArray], fields_2: dict[str, NDArray]) -> NDArray:
...
+                                right_meta[right_comp][1],
+                                preserve_dims=preserve_dims,
+                                preserve_index=preserve_index,
+                                freq_dim="f",
+                                freq_index=freq_index,
+                                mode_dim=outer_dim_2,
+                                tangential_dims=tangential_dims,
+                                mode_slice=slice(i1, i1_end),
+                            )
+                            weighted_right = right_block * d_area
```
caseyflex, line 1272: maybe this can be moved out of the inner loop, by moving it to the left factor or swapping loop ordering

caseyflex (changes requested):
I think the speed-memory tradeoff is good overall. And I think this is an algorithmic improvement even when we do try to keep memory usage low.

A couple more comments:

- This will conflict heavily with @dmarek-flex 's outer dot colocation PR. Whoever merges second will need to carefully incorporate the logic improvements from the other
- I'd still like a unit test that explicitly checks that the memory usage is as expected. This could be here or backend.
- Do we need to update any memory estimation logic?
- Can we lower OUTER_DOT_BLOCK_TARGET_BYTES to 8, and make the min block size 1 instead of 8? This may be safer for parallel execution.
- I also talked with @dmarek-flex about this, but maybe you have thoughts -- you keep the old `_outer_fn_summation` because it is more general with an arbitrary function and used elsewhere, but the flexibility is at the cost of performance. Do you have any ideas about how to handle all use cases with minimal code duplication?

here is the other PR:
https://github.com/flexcompute/tidy3d/pull/3100

tidy3d/components/data/monitor_data.py
```diff
@@ -120,6 +121,10 @@
 # Threshold for cos(theta) to avoid unphysically large amplitudes near grazing angles
 COS_THETA_THRESH = 1e-5
 MODE_INTERP_EXTRAPOLATION_TOLERANCE = 1e-2
+OUTER_DOT_BLOCK_TARGET_BYTES = 64 * 1024**2
```
caseyflex, line 124: Actually, maybe we can use a simpler logic for block size. We could just have some default block size and add it as an optional parameter to the private outer dot implementation function

tidy3d/components/data/monitor_data.py
```diff
@@ -1080,6 +1068,218 @@ def fn(fields_1: dict[str, NDArray], fields_2: dict[str, NDArray]) -> NDArray:
...
+        block_size_right = min(num_modes_right, block_size)
+
+        for preserve_index in preserve_indices:
+            for freq_index in range(n_freq):
+                out = np.zeros((num_modes_left, num_modes_right), dtype=dtype)
+
+                for sign, left_comp, right_comp in term_specs:
+                    for i0 in range(0, num_modes_left, block_size_left):
+                        i0_end = min(i0 + block_size_left, num_modes_left)
+                        left_block = ElectromagneticFieldData._slice_mode_xy_matrix(
```
caseyflex, line 1242: Not sure we need to optimize further now, but maybe we don't need to slice the data 4 times here

dmarek-flex:
Correct me if I am wrong, but I think the most effort here is spent figuring out the dimensions and their order and prepping the arrays for the matrix multiply. So I think this optimization could be easily added to the numpy part of my branch.

https://github.com/flexcompute/tidy3d/blob/e147035e97a22ed39d58462cc80d8ae6d97692b3/tidy3d/components/data/utils.py#L229-L301

caseyflex:
> Correct me if I am wrong, but I think the most effort here is spent figuring out the dimensions and their order and prepping the arrays for the matrix multiply. So I think this optimization could be easily added to the numpy part of my branch.

Yeah I'm not completely sure. Seems a mix of that, and then the actual performance gain is from changing the python looping / BLAS matrix multiply tradeoffs. But I also didn't study your handling in detail. So I think it would be critical to have some unit tests that control the memory at least of the implementation, as there's just so much that can go wrong here -- one extra line of python can double the memory usage

yaugenst-flex:
Closing this in favor of https://github.com/flexcompute/tidy3d/pull/3318

PR #3262, opened 2026-02-10
fix(tidy3d): FXC-5400-disable-cache-racing-tests

marcorudolphflex:
The newly introduced tests for local cache race conditions seem to run instable in CI as processes do not spawn properly. This is why I would skip those tests by default. They might be temporally activated for local testing.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to test code, primarily skipping flaky CI tests and tightening a temp-directory safety check, with no production behavior impact.
> 
> **Overview**
> Improves local-cache test safety and CI stability.
> 
> The local-cache isolation fixture now only allows cache directory deletion within the test `tmp_path` (tightening the guard to avoid touching parent paths). Two multiprocessing race-condition tests for cache stats are now **skipped on GitHub Actions** via `pytest.mark.skipif` due to CI process-spawn/resource instability, while remaining available for targeted local runs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5ce1c02bb1c969776734e6a00ce4bac89e83a74d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/test_web/test_local_cache.py
+++ b/tests/test_web/test_local_cache.py
@@ -70,7 +70,7 @@ def _isolate_local_cache(tmp_path, monkeypatch):
 
     def _safe_remove_cache_dir(path, *, recreate):
         target = Path(path).resolve()
-        allowed_root = tmp_path.resolve().parent
+        allowed_root = tmp_path.resolve()
         try:
             target.relative_to(allowed_root)
         except ValueError:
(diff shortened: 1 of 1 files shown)


yaugenst-flex (changes requested):
I think we could just disable them in CI? In the root conftest I believe there is already a mechanism to detect that, so we could do [skipif](https://docs.pytest.org/en/stable/how-to/skipping.html#id1)

marcorudolphflex:
> I think we could just disable them in CI? In the root conftest I believe there is already a mechanism to detect that, so we could do [skipif](https://docs.pytest.org/en/stable/how-to/skipping.html#id1)

Ah I did not know that, thanks. Did that.

yaugenst-flex (approved):

PR #3265, opened 2026-02-10
refactor(tidy3d): FXC-5286-pydantic-v-2-refactor-validator-sprawl

marcorudolphflex:
Refactors validator hotspots by consolidating multiple @model_validator(mode="after") hooks into a single orchestrator validator that invokes ordered helper methods. This preserves existing warning/error ordering and messages while improving readability and maintainability. Tests are added/updated to lock in validation order where needed.

---

> [!NOTE]
> **Medium Risk**
> Broad refactor of validation ordering across core `Simulation`/`Medium` models; while intended to preserve messages, subtle ordering changes can affect which warnings/errors users see first and may break edge-case setups.
> 
> **Overview**
> Refactors Pydantic v2 post-init validation across the `Simulation` and `Medium` families by replacing many separate `@model_validator(mode="after")` hooks with a single `_run_after_validators()` orchestrator that runs checks in an explicit, dependency-aware order.
> 
> Introduces `call_wrapped_validator(...)` to invoke existing validator factories deterministically, updates several simulations (`Simulation`, `AbstractSimulation`, `ModeSimulation`, `EMESimulation`, `HeatChargeSimulation`) and many medium subclasses accordingly, and adds tests/policy enforcement to lock in warning/error ordering (including new log-order assertions) and prevent new ad-hoc after-validators in these families.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 26c1fbee7c0e6cc3d4fe89948c7a16373cda8968. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tidy3d/components/tcad/simulation/heat_charge.py
+++ b/tidy3d/components/tcad/simulation/heat_charge.py
@@ -408,7 +408,7 @@ class HeatChargeSimulation(AbstractSimulation):
 
     @field_validator("structures")
     @classmethod
-    def check_unsupported_geometries(cls, val: tuple[Structure, ...]) -> tuple[Structure, ...]:
+    def _check_unsupported_geometries(cls, val: tuple[Structure, ...]) -> tuple[Structure, ...]:
         """Error if structures contain unsupported yet geometries."""
         for ind, structure in enumerate(val):
             bbox = structure.geometry.bounding_box
@@ -486,6 +486,30 @@ class HeatChargeSimulation(AbstractSimulation):
         return obj_do_not_cross_solid_idx, obj_do_not_cross_cond_idx
 
     @model_validator(mode="after")
+    def _run_after_validators(self) -> Self:
+        """Run post-init validations in an explicit, dependency-aware order."""
+        super()._run_after_validators()
+        self._structures_not_at_edges()
+        self._validate_scene()
+        self._monitors_cross_solids()
+        self._check_voltage_array_if_capacitance()
+        self._names_exist_bcs()
+        self._check_natural_convection_bc()
+        self._check_freqs_requires_ac_source()
+        self._check_charge_simulation()
+        self._not_all_neumann()
+        self._names_exist_grid_spec()
+        self._warn_if_minimal_mesh_size_override()
+        self._names_exist_sources()
+        self._check_medium_specs()
+        self._check_coupling_source_can_be_applied()
+        self._check_heat_sim()
+        self._check_conduction_sim()
+        self._estimate_charge_mesh_size()
+        self._check_transient_heat()
+        self._check_non_isothermal_is_possible()
+        return self
+
     def _monitors_cross_solids(self) -> Self:
         """Error if monitors does not cross any solid medium."""
         val = self.monitors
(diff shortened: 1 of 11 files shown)


tidy3d/components/tcad/simulation/heat_charge.py
```diff
@@ -491,6 +491,28 @@ def _check_cross_solids(self, objs: tuple[Box, ...]) -> tuple[int, ...]:
...
+        self.warn_if_minimal_mesh_size_override()
+        self.names_exist_sources()
+        self.check_medium_specs()
+        self.check_coupling_source_can_be_applied()
+        self.check_heat_sim()
+        self.check_conduction_sim()
+        self.estimate_charge_mesh_size()
+        self.check_transient_heat()
+        self.check_non_isothermal_is_possible()
+        return self
```
marc-flex, line 511: Maybe a naive question. Does this means every time we add a new (root) validator we need to remember to add it to this model validator?

marcorudolphflex: Yup. I also see some trade-off here.

marc-flex: I'm struggling to see the up side of doing this this way. Could you explain?

marcorudolphflex: In this way, we have a stable ordering and deterministic outputs.
There is also some planning on a more advanced validation setup which would call all validators in any case. In this context, we could also handle different levels of validations easier and better to read.

yaugenst-flex: @marc-flex You don't want to know how much time I spent reordering validators for the pydantic v2 migration to make sense of what's going on and why things were failing 😉

This is an attempt to fix the following problem:
1. Validation logic is currently scattered across many separate "run after init" hooks
2. The run order is implicit (driven by decorator registration order and inheritance) and is not obvious from reading the class
3. Some checks depend on other checks having run first, so hidden ordering oftentimes leads to confusing behavior (and in fact also unintended and uncaught behavior)

A concrete example: Currently the `_check_natural_convection_bc` validator is called _before_ `_names_exist_bcs`, so bad names will throw `KeyError` instead of the intended `ValidationError/SetupError`. Similar issues happen all over the place because the execution order is not obvious.

So this PR makes things explicit, inheritance order is easier to reason about, and the behavior is more testable. That is well worth adding a line to `_run_after_validators` I think. If we decide on a naming convention for validators (e.g., `_check_*` or `_validate_*`) then we can also automatically enforce this, but then you'll have to remember to name your validators accordingly.

The long term fix will be to split validation into distinct, well structured phases.

yaugenst-flex (commented):
@marcorudolphflex thanks, this looks good directionally. But as evidenced by [this thread](https://github.com/flexcompute/tidy3d/pull/3265#discussion_r2789088457) i think we need to improve the process design here too. At the very least there should be some docstrings, and also a contributor doc (could add this to AGENTS.md too) that documents the correct way to add validators in this style.

marcorudolphflex:
> @marcorudolphflex thanks, this looks good directionally. But as evidenced by [this thread](https://github.com/flexcompute/tidy3d/pull/3265#discussion_r2789088457) i think we need to improve the process design here too. At the very least there should be some docstrings, and also a contributor doc (could add this to AGENTS.md too) that documents the correct way to add validators in this style.

That makes definitely sense! Added some notes to `Agents.md` and `docs/development/introduction/code_quality_principles.rst`

yaugenst-flex (approved):

PR #3273, opened 2026-02-11
FXC-0000: simplify testmon cache promotion path

yaugenst-flex:
## Summary
- remove in-run promotion (`promote-testmon-cache`) from `tidy3d-python-client-tests.yml` and its workflow-validation gate
- add a single promotion entrypoint triggered by `workflow_run` after successful `merge_group` runs
- simplify `tidy3d-testmon-cache-promote.yml` to always require and use `source_run_id` (source-run artifacts only)

## Why
- avoid cache promotion writes in merge-queue refs
- keep one deterministic promotion path instead of dual-path logic

## Validation
- `poetry run pre-commit run --all-files`

---

> [!NOTE]
> **Medium Risk**
> Changes alter CI control flow for cache promotion and rely on `workflow_run` context and artifact download behavior; misconfiguration could stop cache promotion or unexpectedly skip it, but production code is unaffected.
> 
> **Overview**
> **Testmon cache promotion is decoupled from the main CI workflow.** The `promote-testmon-cache` job and its `workflow-validation` gating were removed from `tidy3d-python-client-tests.yml`, so the test workflow no longer performs cache writes during merge-queue execution.
> 
> **Cache promotion is now a single, deterministic path.** `tidy3d-testmon-cache-promote.yml` is updated to run on `workflow_run` completion for the tests workflow (restricted to merge-queue branches) and to *always* resolve and require a numeric `source_run_id`, downloading artifacts only from that source run and guarding execution to successful `merge_group` runs from the same repo.
> 
> **CI hardening tweaks.** The `zizmor` job now pins `zizmor==1.19.0`, uses `${{ github.token }}` for `GH_TOKEN`, and explicitly sets `contents: read` permissions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 70b57e0bc286fdad5b93484b2c6f4eda7e5a790d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/tidy3d-python-client-tests.yml
+++ b/.github/workflows/tidy3d-python-client-tests.yml
@@ -334,6 +334,7 @@ jobs:
     needs: determine-test-scope
     if: needs.determine-test-scope.outputs.code_quality_tests == 'true'
     permissions:
+      contents: read
       security-events: write
     steps:
       - name: Checkout repository
@@ -345,9 +346,9 @@ jobs:
         uses: astral-sh/setup-uv@b75a909f75acd358c2196fb9a5f1299a9a8868a4 # v6.7.0
 
       - name: Run zizmor 🌈
-        run: uvx zizmor .github/workflows/*.y* --format=sarif . > results.sarif
+        run: uvx --from "zizmor==1.19.0" zizmor .github/workflows/*.y* --format=sarif . > results.sarif
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
+          GH_TOKEN: ${{ github.token }}
 
       - name: Upload SARIF file
         uses: github/codeql-action/upload-sarif@192325c86100d080feab897ff886c34abd4c83a3 # v3.30.3
@@ -356,9 +357,9 @@ jobs:
           category: zizmor
       
       - name: run zizmor directly  # this gets a success or fail result
-        run: uvx zizmor  .github/workflows/*.y*
+        run: uvx --from "zizmor==1.19.0" zizmor .github/workflows/*.y*
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
+          GH_TOKEN: ${{ github.token }}
   
   lint-branch-name:
     needs: determine-test-scope
@@ -1337,25 +1338,6 @@ jobs:
       test_type: ${{ needs.determine-test-scope.outputs.test_type }}
     secrets: inherit # zizmor: ignore[secrets-inherit]
 
-  promote-testmon-cache:
-    name: promote-testmon-cache
-    needs:
-      - determine-test-scope
-      - local-tests
-      - remote-tests
-    if: >-
-      needs.determine-test-scope.outputs.local_tests == 'true' &&
-      needs.determine-test-scope.outputs.remote_tests == 'true' &&
-      github.event_name == 'merge_group' &&
-      needs.local-tests.result == 'success' &&
-      needs.remote-tests.result == 'success'
-    permissions:
-      actions: write
-      contents: read
-    with:
-      source_event: ${{ github.event_name }}
-    uses: ./.github/workflows/tidy3d-testmon-cache-promote.yml
-
   workflow-validation:
     name: workflow-validation
     if: always()
@@ -1374,7 +1356,6 @@ jobs:
       - verify-version-consistency
       - test-submodules
       - extras-integration-tests
-      - promote-testmon-cache
     runs-on: ubuntu-latest
     steps:
       - name: move-type-imports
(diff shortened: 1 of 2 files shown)


.github/workflows/tidy3d-testmon-cache-promote-on-workflow-run.yml
```diff
@@ -0,0 +1,24 @@
+name: "public/tidy3d/promote-testmon-cache-from-merge-group"
+
+on: # zizmor: ignore[dangerous-triggers]
+  workflow_run:
+    workflows:
+      - "public/tidy3d/python-client-tests"
+    types:
+      - completed
+
```
daquinteroflex: I would strongly avoid this (zizmor is warning correctly) because it causes a racing condition sometimes and makes them very annoying to cancel. Also completed does not mean success. Why not just add an if statement as part of the larger workflow that depends on the completion outcome of the other `tidy3d-workflow-validate` step in tests but only triggers on merge queues?

daquinteroflex: You should be able to do this without creating a new action file and we reduce the bloating on the action panel too (since there are a lot already in the other repos)

.github/workflows/tidy3d-python-client-tests.yml
```diff
@@ -1337,25 +1337,6 @@ jobs:
...
-    permissions:
-      actions: write
-      contents: read
-    with:
-      source_event: ${{ github.event_name }}
-    uses: ./.github/workflows/tidy3d-testmon-cache-promote.yml
-
   workflow-validation:
     name: workflow-validation
     if: always()
```
daquinteroflex, line 1343: Like you could introduce that step here, and check that workflow-validation was successful, it should achieve the logic you desire right?

daquinteroflex: Or add all the corresponding if statements here accordingly?

.github/workflows/tidy3d-testmon-cache-promote-on-workflow-run.yml
yaugenst-flex: ok sure i can fold it into the existing one

.github/workflows/tidy3d-python-client-tests.yml
yaugenst-flex: i can't move this back inline because that writes cache entries under merge queue refs which is exactly what i am intending to fix here. the promotion needs to be run in a default-branch workflow context, otherwise the cache isnt reusable

.github/workflows/tidy3d-testmon-cache-promote.yml
```diff
@@ -39,40 +44,52 @@ permissions:
 
 jobs:
   discover-cache-candidates:
+    if: >-
+      github.event_name != 'workflow_run' ||
+      (
+        github.event.workflow_run.event == 'merge_group' &&
+        github.event.workflow_run.conclusion == 'success' &&
```
daquinteroflex, line 53: sure sounds good then

daquinteroflex (approved):

.github/workflows/tidy3d-python-client-tests.yml
daquinteroflex: Actually wait why don't you just make it trigger on a push to `develop` instead then? Effectively that's what the merge queue does when it merges

.github/workflows/tidy3d-testmon-cache-promote.yml
```diff
@@ -1,15 +1,20 @@
 name: "public/tidy3d/promote-testmon-cache"
 
-on:
+on: # zizmor: ignore[dangerous-triggers]
+  workflow_run:
+    workflows:
+      - "public/tidy3d/python-client-tests"
+    types:
+      - completed
```
daquinteroflex, line 8: Because this will trigger on PR on every branch otherwise

.github/workflows/tidy3d-testmon-cache-promote.yml
daquinteroflex: So this will just be trigggered on every tests run completion and then skipped

.github/workflows/tidy3d-python-client-tests.yml
yaugenst-flex: i considered this but we need deterministc access to the source run artifacts, which you get via `workflow_run.id` directly. for `push` i don't know how we would do the lookup exactly

.github/workflows/tidy3d-testmon-cache-promote.yml
yaugenst-flex: changed `workflow_run` to only `branches: gh-readonly-queue/develop/**` so it won't trigger on this workflow

.github/workflows/tidy3d-testmon-cache-promote.yml
yaugenst-flex: commented above

.github/workflows/tidy3d-python-client-tests.yml
daquinteroflex: Fair enough, this should work fine as is now too anyways. 

Actually do wonder if there's a way to query the actions a given SHA has run possibly? Normally they all show in the commit history, but maybe as a future improvement if needed

daquinteroflex (approved):

.github/workflows/tidy3d-python-client-tests.yml
daquinteroflex: Technically looks like it's possible but dealing with the API in an action is a recipe for timeout problems and connection/auth issues `gh run list --commit <YOUR_SHA>` . Your solution should work fine anyways

yaugenst-flex: yeah it's possible in principle but adds a lot of (i think) unnecessary logic...

